PR #571, opened 2015-11-23
[macros] Disable italic correction for monospaced italics

tkoeppe:
This helps align code columns vertically in the presence of italics.


tkoeppe:
Solved alternatively by 5ca060c.

Issue #430, opened 2015-01-21
Choose whether to put a space on the left or right of `...` and apply consistently throughout library


StephanTLavavej:
Search N4296 for `class ...`

There are a couple dozen occurrences of `class ... Meow` in Core, with similar function parameter packs. That's a questionable style, but not outright toxic.

However, there are 4 occurrences of `class ...Args` in Library, which is yuck and bad and wrong. They are:

30.3.1 [thread.thread.class]:
`template <class F, class ...Args> explicit thread(F&& f, Args&&... args);`

30.3.1.2 [thread.thread.constr]/3:
`template <class F, class ...Args> explicit thread(F&& f, Args&&... args);`

30.4 [thread.mutex]:

```
template<class Callable, class ...Args>
  void call_once(once_flag& flag, Callable&& func, Args&&... args);
```

30.4.4.2 [thread.once.callonce]:

```
template<class Callable, class ...Args>
  void call_once(once_flag& flag, Callable&& func, Args&&... args);
```

zygoloid:
> However, there are 4 occurrences of class ...Args in Library, which is yuck and bad and wrong.

I'm curious what your argument is here; is it just aesthetic, or is there some more fundamental reason?

`...` consistently binds to the thing on its right, so space on the left is the style most consistent with the grammar. That said, the same argument suggests that `*`, `&`, and `&&` should have the space on the left, and we don't do that, largely because some prefer to think of the type being "the thing on the left of the name" rather than "the thing surrounding the name". But the `...` is not part of the type; that argument doesn't apply here.

[My general feeling is that the formatting of code samples in the C++ standard should reflect the C++ language as well as possible. To this end, the core language deliberately uses a mixture of styles to emphasize that whitespace doesn't matter. The library specification lives in a somewhat different world; it should stick to a single consistent style. All else being equal, I think it would make sense for that style to match the C++ grammar (rather than only working in common but special cases -- one declarator per declaration, no postfix declarators, no paren declarators), because that would improve the internal consistency of the standard.]

StephanTLavavej:
It's purely an aesthetic argument. In the Library, we think of `typename...` meaning "lots of types", regardless of how the grammar is technically formed.

I see your point about Core mixing styles, and I don't think that it needs to change. If you are tempted to change Library to say `class ...Args` (which would make me regret mentioning this!), please ask the LWG reflector first, as I suspect there will be wide agreement that `class... Args` is preferable. We Library implementers have to read those clauses, after all. :->

[...]

PR #577, opened 2015-12-05
Whitespace fixes around punctuators (library)

tkoeppe:


source/basic.tex
```diff
@@ -991,7 +991,7 @@
 
 \begin{codeblock}
 typedef int N;
-template<N X, typename N, template<N Y> class T> struct A;
+template<N X, class N, template<N Y> class T> struct A;
```
jwakely: @zygoloid has said elsewhere that the core parts of the standard intentionally use a mix of formatting styles to reinforce the point that whitespace (or the  choice of `class` or `typename` here) has no semantic significance. So changing things to be consistent breaks that.

jwakely:
The library parts look good, but not the rest.

@zygoloid has said elsewhere that the core parts of the standard intentionally use a mix of formatting styles to reinforce the point that whitespace (or the  choice of `class` or `typename` in template parameter lists) has no semantic significance. So changing things to be consistent breaks that intentional inconsistency.

tkoeppe:
You mean Issue #430? Hm, that's unfortunate. I can split the library changes out if you like those.

jwakely:
Yes. I'll merge the library parts, but I'm not touching the rest :-)

tkoeppe:
OK, I restricted the branch to only the library changes!

source/utilities.tex
```diff
@@ -5421,13 +5421,13 @@
 
 \begin{itemdescr}
 \pnum
-\requires Let \tcode{CT} be \tcode{common_type<unique_ptr<T1, D1>::pointer,}
+\requires Let \tcode{CT} be \tcode{common_type<unique_ptr<T1, D1>::pointer},
```
jwakely: This comma is part of the type and should be in code font.

tkoeppe: Oh, right, this is a manual line break. Fixed.

PR #578, opened 2015-12-05
Replace raw string 'C++' with the appropriate macro

tkoeppe:


source/templates.tex
```diff
@@ -117,7 +117,7 @@
 units.
 A template, a template explicit specialization~(\ref{temp.expl.spec}), and a class
 template partial specialization shall not have C linkage. Use of a linkage specification
-other than C or C++ with any of these constructs is conditionally-supported, with
+other than C or \Cpp with any of these constructs is conditionally-supported, with
```
zygoloid: These should probably say `\tcode{"C"}` or `\tcode{"C++"}` instead.

tkoeppe: Seems reasonable, changing.

PR #580, opened 2015-12-05
Use textual angle brackets rather than relational operators

tkoeppe:
Angle brackets look better than less-than/greater-than operators when used to typeset parentheticals.


tkoeppe:
Example:

image

PR #584, opened 2015-12-11
[atomics] Reword table headers, removing mention of "inttypes.h"

tkoeppe:
`inttypes.h` is not the primary header that defines `int_least8_t` etc. That's `cstdint`. The only new thing `inttypes.h` adds is `idivmax_t`, but we don't have an atomic version of that.

### DIFF
--- a/source/atomics.tex
+++ b/source/atomics.tex
@@ -612,9 +612,9 @@ corresponding specialization. If it is a base class, it shall support the same
 member functions as the corresponding specialization.
 
 \begin{floattablebase}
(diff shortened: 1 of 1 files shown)


tkoeppe:
@zygoloid: Is this change sufficiently editorial, or will this require a broader discussion?

tkoeppe:
@zygoloid: PTAL. I can remove the word "corresponding" if you think it's too much, or move it into the table caption (e.g. "named atomic types and their corresponding non-atomic counterparts").

Issue #586, opened 2015-12-16
[diff.library] NULL <cstdio> missing in Table 149


cubbimew:
`Table 133 — Header <cstdio> synopsis` mentions `NULL <cstdio>`
`C.5.2.4[diff.null]` mentions "The macro `NULL`, defined in .. `<cstdio>` .."
however
`Table 149 — Standard macros` lists only six NULLs. `NULL <cstdio>` is missing in this table.

jwakely:
Fixed by merging the pull request - thanks.

PR #587, opened 2015-12-17
[diff.library] add missing NULL <cstdio> to Table 149

cubbimew:
Add `NULL <cstdio>` and reflow Table 149 to maintain alphabetical order (for issue #586)

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -1554,7 +1554,7 @@ declarations, or behavior from the Standard C library
 noted in other subclauses (\ref{headers}, \ref{support.types}, \ref{c.strings}).
 
 \pnum
-The \Cpp standard library provides 57 standard macros from the C library,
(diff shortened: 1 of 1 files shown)


tkoeppe:
Should we also change the first `57` to `51`, and the second `57` to `45`?

cubbimew:
Also, "20 standard types" is actually 15, although "209 standard functions" seems correct.
Sounds like a separate editorial issue.

tkoeppe:
Haha, yes, counting is hard :-) Will you make the changes?

cubbimew:
Okay, updated the counts.

tkoeppe:
Great, can you squash too?

cubbimew:
...and squashed

tkoeppe:
One single diff (LHS: before, RHS: after):

image

jwakely:
Thanks! I know how much of a pain it is to reflow those tables.

PR #588, opened 2015-12-18
[macros] Add space into \range, reflow [alg.transform]

tkoeppe:
The usual typographic convention is to have a space after (but not before) a comma. This change inserts a comma into the `\range` macro to print ranges as `[a, b)` rather than the current `[a,b)`.

This has about 90 pdfdiffs, all looking like improvements.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -1581,12 +1581,12 @@ or
 \requires
 \tcode{op} and \tcode{binary_op}
 shall not invalidate iterators or subranges, or modify elements in the ranges
-\crange{first1}{last1},
-\crange{first2}{first2 + (last1 - first1)},
-and
-\crange{result}
-{result + (last1 - first1)}.\footnote{The use of fully
+\begin{itemize}
+\item \crange{first1}{last1},
+\item \crange{first2}{first2 + (last1 - first1)}, and
+\item \crange{result}{result + (last1 - first1)}.\footnote{The use of fully
 closed ranges is intentional.}
+\end{itemize}
 
 \pnum
 \returns
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -234,7 +234,7 @@
 \newcommand{\mname}[1]{\tcode{\unun\ungap#1\ungap\unun}}
 
 %% Ranges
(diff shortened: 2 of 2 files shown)


source/macros.tex
```diff
@@ -234,7 +234,7 @@
 \newcommand{\mname}[1]{\tcode{\unun\ungap#1\ungap\unun}}
 
 %% Ranges
-\newcommand{\Range}[4]{\tcode{#1\brk{}#3,\brk{}#4\brk{}#2}\xspace}
+\newcommand{\Range}[4]{\tcode{#1\brk{}#3, \brk{}#4\brk{}#2}\xspace}
```
zygoloid: Please use `~` not ` ` here; we don't want line breaks within these.

tkoeppe: I thought `\brk{}` means "allow linebreak here"?

tkoeppe: The change adds some line breaks inside ranges. Here's one example:

image

zygoloid:
OK, after some discussion: I think we should allow linebreaks here, but with the hyphenation penalty like we had before (we should prefer to break on a "real" space rather than after the comma). Something like `\discretionary{}{}{ }` might work.

tkoeppe:
Changed to just `,~\brk{}` which is tighter than `, `.

tkoeppe:
Note to self: investigate 25.4.3/2.

tkoeppe:
Added a change to [alg.transform] to look a bit more palatable.

source/algorithms.tex
```diff
@@ -1581,12 +1581,12 @@
...
 \tcode{op} and \tcode{binary_op}
 shall not invalidate iterators or subranges, or modify elements in the ranges
-\crange{first1}{last1},
-\crange{first2}{first2 + (last1 - first1)},
-and
-\crange{result}
-{result + (last1 - first1)}.\footnote{The use of fully
+\begin{itemize}
+\item \crange{first1}{last1},
+\item \tcode{[first2, first2 + (last1 - first1)]}, and
```
zygoloid: Why is this one not a \crange?

tkoeppe: Left-over from experimenting; reverted to `\crange`.

PR #591, opened 2015-12-21
[macros, atomic] Align placeholders

tkoeppe:
[deleted; need to rebuild]


tkoeppe:
5 pdf diffs. Screenshots (LHS: before, RHS: after):

image

image

image

image

image


@zygoloid: PTAL?

Issue #596, opened 2016-01-02
[c.files] table 134 lists SCNX* macros, which do not exist


cubbimew:
[c.files] Table 134 - Header `<cinttypes>` synopsis mentions
`SCN{d i o u x X}[FAST LEAST]{8 16 32 64}`
`SCN{d i o u x X}{MAX PTR}`
but there are no SCNX* macros in C's `inttypes.h`, only SCNx*.

This contradicts [c.files]/4 "The contents of header `<cinttypes>` are the same as the Standard C Library header `<inttypes.h>`", and looks like a copy-paste error from the PRI* macros just above, which do come in both PRIx and PRIX variants.

timsong-cpp:
I actually sent an email to lwgchair@ yesterday, since I wasn't sure if this counts as editorial. But if it can be handled editorially, all the better :)

jwakely:
Resolved by merging #597

PR #597, opened 2016-01-02
[c.files] cinttypes synopsis refers to SCNX* macros which do not exist

cubbimew:
This addresses #596 by removing the uppercase X from the SCN* rows of Table 134


mclow:
This looks to be editorial to me. Those macros don't exist in C99.

tkoeppe:
@jwakely: For your consideration.

PR #604, opened 2016-01-19
Remove redundant constraint in [class.static.data]

Arcoth:
The fact that local classes shall not have static data members is already stated in [class.local]/4 with the exact same wording.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -1207,10 +1207,9 @@ or indirectly within unnamed classes shall not contain \tcode{static}
 data members.
 
 \pnum
-\indextext{restriction!static@\tcode{static} member local~class}%
-\tcode{Static} data members of a class in namespace scope have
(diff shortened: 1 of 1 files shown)


zygoloid:
Seems like we would benefit from more massaging here. The previous sentence "static data members of a class in namespace scope have the linkage of that class (3.5)" makes one wonder what happens for static data members of classes in other scopes. Again, that's normatively covered by wording elsewhere.

Maybe we can make this whole paragraph a note, and add a cross-reference from the second sentence to [class.local] instead? (Or maybe delete the whole paragraph.)

Arcoth:
I agree; a note seems appropriate. See the new commit.

tkoeppe:
Squash, rebase and push?

Arcoth:
@tkoeppe Good call.

tkoeppe:
@zygoloid PTAL

PR #619, opened 2016-02-13
Uniform "ones' complement" and "two's complement"

k-satoda:
"one's complement" is inconsistent with the C standard.
"1's complement" and "2's complement" are also inconsistent.

FYI, I found that there was a lengthy discussion on Wikipedia at 2011
about "one's complement". The C++ standard was mentioned as a notable
use of "one's complement". There seems no consensus in the discussion,
but the main article has been settled as "ones' complement" for years
since then.
https://en.wikipedia.org/wiki/Ones'_complement
https://en.wikipedia.org/wiki/Wikipedia_talk:WikiProject_Computer_science/Archive_10#Ones.27_complement_or_one.27s_complement.3F

For me, a quote from "Art of Computer Programming, Volume 2" by
Donald E. Knuth is convincing.
https://books.google.com/books?id=Zu-HAwAAQBAJ&pg=PT343
> ... A two's complement number is complemented with respect to a single
> power of 2, while ones' complement number is complemented with respect
> to a long sequence of 1s. ...

While it is still debatable in general, I think inconsistency with the
C standard is enough for the C++ standard to justify this change.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3802,8 +3802,8 @@ and 1, in which the values represented by successive bits are additive,
 begin with 1, and are multiplied by successive integral power of 2,
 except perhaps for the bit with the highest position. (Adapted from the
 \doccite{American National Dictionary for Information Processing Systems}.)}
-\enterexample this International Standard permits 2's complement, 1's
-complement and signed magnitude representations for integral types.
+\enterexample this International Standard permits two's complement,
+ones' complement and signed magnitude representations for integral types.
 \exitexample
 
 \pnum
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -1984,7 +1984,7 @@ the enumeration are the values of the underlying type. Otherwise,
 for an enumeration where $e_\mathit{min}$ is the smallest enumerator and
 $e_\mathit{max}$ is the largest, the values of the enumeration are the
 values in the range $b_{min}$ to $b_{max}$, defined as follows: Let $K$
-be 1 for a two's complement representation and 0 for a one's complement
+be 1 for a two's complement representation and 0 for a ones' complement
 or sign-magnitude representation. $b_{max}$ is the smallest value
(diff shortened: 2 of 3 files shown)


sigfpe:
Is the inconsistency with the C++ standard enough to justify that the C
standards text be modified?

k-satoda:
@sigfpe I think it might be, for example, if the intent of the word in
question was to be compatible with the predecessor in C++.

sigfpe:
Should the modifications be applied to the C standards draft instead?-------- Original message --------From: Kazutoshi SATODA <notifications@github.com> Date: 2/13/2016  8:06 AM  (GMT-08:00) To: cplusplus/draft <draft@noreply.github.com> Cc: sigfpe <sig.fpe@axiomatics.org> Subject: Re: [draft] Uniform "ones' complement" and "two's complement" (#619) 
@sigfpe I think it might be, for example, if the intent of the word in

question was to be compatible with the predecessor in C++.

k-satoda:
@sigfpe I personally don't think so. I believe my above comments are
enough to explain why. If you think the C standard should be changed
instead, please explain why.

sigfpe:
You asserted they are enough.

k-satoda:
@sigfpe Sorry if I made the confusion. But no, I didn't.

zygoloid:
It looks like "one's complement" is more common than "ones' complement" by a fairly large factor:

https://books.google.com/ngrams/graph?content=one%27s+complement%2Cones%27+complement%2Cones+complement&year_start=1800&year_end=2000&corpus=15&smoothing=3&share=&direct_url=t1%3B%2Cone%20%27s%20complement%3B%2Cc0%3B.t1%3B%2Cones%20%27%20complement%3B%2Cc0%3B.t1%3B%2Cones%20complement%3B%2Cc0

However, I'm inclined to take TAOCP as authority here, and I buy the argument that this is really the result of taking "ones's complement" and applying the usual English rule of Xs's -> Xs'.

PR #626, opened 2016-02-26
[futures.shared_future] Remove duplicated word "shared"

jwilk:


tkoeppe:
@K-ballo dug up this history: In [N2997](http://wg21.link/n2997), the original wording was:

>  but they synchronize on the shared associated state

It seems that "associated state" was renamed to "shared state" later, and so the construction became "shared shared state".

zygoloid:
Can we find a better word than "shared" for the first occurrence? Maybe "the common shared state"?

K-ballo:
I think dropping the first "shared" should be fine, there is only (at most) one shared state involved.

K-ballo:
Anthony Williams: "That looks like a simple editorial fix to me, especially since it's in a note."
Lawrence Crowl: "Agreed."

PR #627, opened 2016-02-27
Delete the notionally redundant 'update' when describing actions upon ...

faisalv:
... independent memory locations.  The definition of 'access' includes modifying a location per 1.3.1/access, so we don't need to state 'update and access' we should be able to just state 'access'

(p.s. My first pull request - please let me know if I'm not doing this right or I should be doing something better.  Thank you!)


zygoloid:
Regarding your ps: we aim for editorial commits to the draft to start with the section name they're modifying, for instance "[intro.memory] Delete the [...]". No big deal if you forget, I can do it as part of the merge, but it saves me a couple of seconds.

PR #238, opened 2013-11-27
Fixed nesting bugs for notes and examples

godbyk:
I replaced `\enterexample` and `\exitexample` with `\begin{example}` and `\end{example}`, respectively. Similarly for `\enternote` and `\exitnote`.

By using environments, LaTeX can detect missing `\end{...}` markers and incorrectly nested environments (i.e., closing the outer environment before closing the inner environment).

The side effects of this change are that any new markup should make the following changes:

* Change `\enterexample` to `\begin{example}`.
* Change `\exitexample` to `\end{example}`.
* Change `\enternote` to `\begin{note}`.
* Change `\exitnote` to `\end{note}`.
* Change `\note` to `\remark`. (Because `\note` conflicts with `\begin{note}`. The output is the same as it was previously.)
* Change `\notes` to `\remarks`. (For consistency. The output is the same as it was previously.)

PR #633, opened 2016-03-08
Replaced \note with \remark and \notes with \remarks.

tkoeppe:
This is a single commit extracted from PR #238 which seems abandoned. This commit had been deemed useful, so I copied it and rebased it so it can be applied.

No PDF diffs.


godbyk:
I rebased PR #238 against the current master branch again.

PR #635, opened 2016-03-09
Travis CI support

godbyk:
This allows the PDF to be built using [Travis CI](https://travis-ci.org/godbyk/draft). Handy for catching syntax errors in pull requests.

The build script could be expanded to add other checks in the future, too (for example, raise errors on too overfull lines, missing cross-reference labels).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,64 @@
+#
+# Builds the C++ standard document on Travis CI <https://travis-ci.org/cpluscplus/draft>
+#
+
+env:
+  - BUILD_TYPE=latexmk   # build using latexmk
+  - BUILD_TYPE=make      # build using Makefile
+  - BUILD_TYPE=manual    # build manually
+  - BUILD_TYPE=complete  # build manually and regenerate figures, grammer, and cross-references
+
+script:
+  # Build std.pdf
+  - pushd source
+  - if [ "$BUILD_TYPE" = "latexmk" ]; then
+      latexmk -pdf std;
+    fi
+  - if [ "$BUILD_TYPE" = "make" ]; then
+      make -j2;
+    fi
+  - if [ "$BUILD_TYPE" = "complete" ]; then
+      for FIGURE in *.dot; do
+        dot -o$(basename $FIGURE .dot).pdf -Tpdf $FIGURE;
+      done;
+      ../tools/makegram;
+      ../tools/makexref;
+    fi
+  - if [ "$BUILD_TYPE" = "manual" -o "$BUILD_TYPE" = "complete" ]; then
+      pdflatex std;
+      pdflatex std;
+      pdflatex std;
+      makeindex generalindex;
+      makeindex libraryindex;
+      makeindex grammarindex;
+      makeindex impldefindex;
+      pdflatex std;
+      pdflatex std;
+    fi
+  - popd
+  # Check to see if generated files are out-dated
+  - pushd source
+  - ../tools/makegram && git status --porcelain grammar.tex
+  - ../tools/makexref && git status --porcelain xref.tex
+  - for FIGURE in *.dot; do
+      dot -o$(basename $FIGURE .dot).pdf -Tpdf $FIGURE;
+      git status --porcelain $(basename $FIGURE .dot).pdf;
+    done
+  - popd
+
+sudo: false
+
+addons:
+  apt:
+    packages:
+    - latexmk
+    - poppler-utils
+    - texlive-binaries
+    - texlive-fonts-recommended
+    - texlive-latex-base
+    - texlive-latex-extra
+    - texlive-latex-recommended
+    - texlive-binaries
+    - graphviz
+    - lmodern
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,4 +61,3 @@ addons:
     - texlive-binaries
     - graphviz
     - lmodern
-


tkoeppe:
Please squash.

.travis.yml
```diff
@@ -0,0 +1,24 @@
...
+    - ppa:texlive-backports/ppa
+    packages:
+    - latexmk
+    - poppler-utils
+    - texlive-binaries
+    - texlive-fonts-recommended
+    - texlive-latex-base
+    - texlive-latex-extra
+    - texlive-latex-recommended
+    - texlive-xetex
```
tkoeppe: I don't think we need xetex.

godbyk:
@tkoeppe I've squashed the commits. I also removed the unnecessary package and unused repository. Thanks!

.travis.yml
```diff
@@ -0,0 +1,21 @@
+script:
+  - pushd source
+  - make -j2
```
godbyk: @tkoeppe Only because I know `make` works and have never tried building it with `latexmk`. If `latexmk` works and is preferred, I'm happy to use it instead.

tkoeppe: Well, either use `make` and don't install `latexmk`, or use `latexmk`, I'd say...

godbyk: @tkoeppe I'm not picky. Which method is preferred?

tkoeppe: `latexmk`

godbyk: @tkoeppe I'll give `latexmk` a try. I think I'll have to duplicate some of the effort from the `Makefile` to the `.travis.xml` file (e.g., building the diagram PDFs with `dot`, generating the grammar and cross-references appendices) since `latexmk` won't do that for us.

Is it worth the effort or is it better to rely on the `Makefile`?

tkoeppe: Ah, it maybe that `latexmk` doesn't go the whole way and is only useful for the common case of small changes. In that case stick with `make`, sure.

.travis.yml
```diff
@@ -0,0 +1,21 @@
...
+  - popd
+
+sudo: false
+
+addons:
+  apt:
+    packages:
+    - latexmk
```
tkoeppe: So, why are you installing `latexmk` then?

godbyk: Because I started by copying the list of required packages from the README.

We could set up a build matrix and have it build using `latexmk` and with the `Makefile` to exercise both build options, too, I s'pose.

Installing the `latexmk` package isn't terribly time-consuming or onerous in any case, though.

.travis.yml
tkoeppe: On second thought, please do use `latexmk` for the test. The regeneration of the xrefs and grammar and plots is not a transient operation; and thus `make` is a tool for the author, not the user (despite what we say in the readme). If `make` were a user-facing tool, then we wouldn't have the grammar and xrefs checked in.

Note also that just `make` doesn't regenerate anything, since it only runs `rebuild` by default.

In summary, the semantics of `latexmk -pdf std` seem to be exactly what we want.

.travis.yml
```diff
@@ -0,0 +1,21 @@
...
+    - poppler-utils
+    - texlive-binaries
+    - texlive-fonts-recommended
+    - texlive-latex-base
+    - texlive-latex-extra
+    - texlive-latex-recommended
+    - texlive-binaries
+    - graphviz
+    - lmodern
+
```
tkoeppe: Gratuitous newline, remove.

tkoeppe:
@zygoloid: What do you think of adding Travis testing? I think it may require you to link the repository to Travis somehow and give them access. Running `latexmk` seems like a decent smoke test, though on the other hand I don't know how much lifting power that would have.

zygoloid:
Sounds like a good idea to me. What sort of access do they require? Presumably push access isn't required for them to do this integration?

In addition to latexmk, it might be interesting to run the various regeneration steps (makegram etc) and check that the checked-in outputs of these steps don't change (though I do wonder how much sense it makes to check in a generated file in the first place...).

tkoeppe:
Setup instructions here: https://docs.travis-ci.com/user/getting-started/#To-get-started-with-Travis-CI%3A

I agree in principle that checking in generated files is iffy, but at least the current setup separates LaTeX editing from our own homebrew scripts. I wouldn't want to *mandate* that any user be able to execute our scripts. However, I would be quite happy to have a test that a simple LaTeX edit does't *change* the output of the generated files, as you suggested, so it would indeed be nice if we could make Travis run that test.

godbyk:
I would recommend setting up a matrix and having Travis CI build the files with both latexmk and using the build scripts.

godbyk:
I've rebased this PR against master and updated the `.travis.yml` file to do the following:

Travis CI will build `std.pdf` in four different ways:

  1. Using `latexmk`.
  2. Using `make`.
  3. Manually per the instructions in `README.rst`.
  4. Manually (same as 3), but also regenerating the grammar annex, cross-references, and figures.

It also regenerates the grammar annex, cross-references, and figures and compares them to the pre-generated versions checked into the repository.

You can view the results of this PR at <https://travis-ci.org/godbyk/draft/builds/137449847>.

.travis.yml
tkoeppe: @godbyk: Ping? While you're at it, might as well clean this up...

godbyk: @tkoeppe Whoops! I missed that one. I'll fix it. Thanks!

tkoeppe:
@godbyk: Are you sure this testing catches missing xref and grammar regen? I just pushed a branch that was missing the regeneration, and all four tests passed.

godbyk:
@tkoeppe What do you mean by 'catches missing xref and grammar regen'?

It just runs `git status` on the generated files to see if they've changed from what's in the repository. The output of that command appears in the console output, but it doesn't throw a build error or anything.

(We should probably write better tests that do throw errors.)

tkoeppe:
@godbyk Oh - but who reads the 5MB of console output? I thought it should be a test failure if you end up affecting the xrefs without checking them in yourself. Is that possible?

godbyk:
@tkoeppe I think it should be possible. Lemme take a look at the Travis CI docs and get back to you.

(The `dot`-generated figures seem to always differ from the files in the git repository, but I think that's just due to the PDF metadata. We should find a better test for those.)

Are there any other tests we should run?

godbyk:
@tkoeppe See https://github.com/cplusplus/draft/pull/780.

PR #645, opened 2016-03-13
(may be wrong) [tuple.apply] Use existing std::invoke function rather than magic INVOKE

tkoeppe:


K-ballo:
I believe this might not be editorial as `std::invoke` is not constexpr? That seems a bit inconsistent though.

tkoeppe:
Oh god, right. Why isn't it constexpr? And why isn't it appropriately noexcept? :-(

K-ballo:
Beats me. I think the lack of `constexpr` deserves an issue, it's inconsistent with `apply`. As for `noexcept`, implementations might add those if they wish to do so.

tkoeppe:
[N4169](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4169) discusses `constexpr`.

AlisdairM:
I believe we had this discussion when designing the feature for the TS, and the current design of using the INVOKE magic was deliberately chosen, as it provides all the necessary magic without accidentally mis-specifying a subtle language feature in the function signature - any change here should definitely go through a working group.

That said, I think there was some interest in cleaning the magic out of the standard once 'std::invoke' landed - although it would definitely require a paper to happen, and may be more subtle than expected.  I got burned in the past trying to clean up mem_fn, for example.

tkoeppe:
@AlisdairM: Thanks, that's good to know. I might just close this PR then and we leave the wording as is.

AlisdairM:
I should add that I am tentatively in favor of such a paper that I have no plans to write myself ;)  I do think it will be a non-trivial amount of work to get right though - we still need better language support for forwarding functions that implicitly preserve the properties of what they are forwarding to.

tkoeppe:
OK, noted :-)

PR #646, opened 2016-03-16
Move removal of bool++ from C++14 compatibility annex to C++17 annex

AlisdairM:
This change was clearly introduced in C++17, at the same meeting
as removing the meaning of the register keyword.  I can see no
Core issue tied to this removal being resolved as a DR against
14 or earlier.

I have confirmed that paper p0002r1 applied this to the C++17
annex, but was listed as a change in annex D rather than clause
5 - the latter change seeming the consistent editorial policy
for removed Core features, so retained.

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -1369,17 +1369,6 @@ might match a predefined usual (non-placement)
 program is ill-formed, as it was for class member allocation functions and
 deallocation functions~(\ref{expr.new}).
 
-\rSec2[diff.cpp11.expr]{Clause \ref{expr}: expressions}
-
-\ref{expr.post.incr}, \ref{expr.pre.incr}
-\change
-Remove increment operator with \tcode{bool} operand.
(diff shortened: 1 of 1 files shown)


tkoeppe:
The commit message should start with "[compatibility]". (That can be fixed in merging, but it would expedite the process if you amended it yourself.)

AlisdairM:
Thanks - my first pull request and I was not aware of the conventions, are they documented somewhere that I missed?

jwakely:
I don't think so, but that's the format used in most of the existing commit log messages.

AlisdairM:
OK, I'll check the log for precedent and update my commit message when I'm home from the office tonight.  I'd rather clean this up myself now, so I am better educated for future pull requests.

AlisdairM:
OK, following the log, I may have got it wrong again, but I retitled with the stable-ref of the clause that I changed - better?

tkoeppe:
Looks great, thanks!

PR #650, opened 2016-03-17
[dcl.attr] Fix wrong source encoding

tkoeppe:
Please always check that your source code uses only basic characters:

    grep -P -n "[\x80-\xFF]" *.tex

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -3832,10 +3832,10 @@ to the name or entity. \exitnote
 
 \pnum
 The \grammarterm{attribute-token} \tcode{fallthrough}
-may be applied to a null statement~(\ref{stmt.expr});
+may be applied to a null statement~(\ref{stmt.expr});
 \indextext{statement!fallthrough}
 such a statement is a fallthrough statement.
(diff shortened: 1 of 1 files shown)


source/declarations.tex
```diff
@@ -3843,7 +3843,7 @@
 The next statement that would be executed after a fallthrough statement
 shall be a labeled statement whose label is a case label or
 default label for the same \tcode{switch} statement.
-The program is ill­formed if there is no such statement.
+The program is ill-formed if there is no such statement.
```
tkoeppe: Note: github isn't displaying the original character here (a dash).

source/declarations.tex
```diff
@@ -3859,17 +3859,17 @@
 \enterexample
 \begin{codeblock}
 void f(int n) {
-  void g(), h(), i();
```
tkoeppe: Please note that the original character was GREEK QUESTION MARK.

tkoeppe: @tomazos: Do you know anything about greek question marks? :-)

PR #657, opened 2016-03-21
[class.mem] Convert nonstatic to non-static since the hyphenated use …

AaronBallman:
…is the more common term.

(We use non-static data member later in this same paragraph, too.)

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -1230,10 +1230,9 @@ or indirectly within unnamed classes shall not contain \tcode{static}
 data members.
 
 \pnum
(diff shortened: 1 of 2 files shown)


source/classes.tex
```diff
@@ -653,7 +653,7 @@
 
 \pnum
 \indextext{layout!class~object}%
-Nonstatic data members of a (non-union) class
+Non-static data members of a (non-union) class
```
tkoeppe: Should "nonstatic function" on line 860 also be changed?

PR #658, opened 2016-03-21
[class.mem, class.mfct.non-static] Convert nonstatic to non-static si…

AaronBallman:
…nce the hyphenated use is the more common term.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -860,7 +860,7 @@ Also see~\ref{temp.arg}.
 \indextext{member function!non-static}
 
 \pnum
-A \grammarterm{non-static} member function may be called for an object of
+A non-static member function may be called for an object of
 its class type, or for an object of a class derived
 (Clause~\ref{class.derived}) from its class type, using the class member
 access syntax~(\ref{expr.ref},~\ref{over.match.call}). A non-static
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -653,7 +653,7 @@ of the \tcode{tword} member of the \tcode{right} subtree of \tcode{s}.
 
 \pnum
 \indextext{layout!class~object}%
-Nonstatic data members of a (non-union) class
+Non-static data members of a (non-union) class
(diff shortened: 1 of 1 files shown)


source/classes.tex
```diff
@@ -856,8 +856,8 @@
 Also see~\ref{temp.arg}.
 \exitnote
 
-\rSec2[class.mfct.non-static]{Nonstatic member functions}%
-\indextext{member function!nonstatic}
+\rSec2[class.mfct.non-static]{Non-static member functions}%
+\indextext{member function!non-static}
 
 \pnum
 A \grammarterm{non-static} member function may be called for an object of
```
zygoloid: Can you remove the bogus \grammarterm while you're here?

AaronBallman:
I've removed the bogus grammar term. I didn't squash because they seem like unrelated changes, but I can try to squash if you would really like me to. :-P

tkoeppe:
No, the two commits are fine now.

Does this change require regenerating the grammar and xrefs?

AaronBallman:
When I ran latexmk, it did say that it was generating a file for
grammarindex.idx, so I would imagine it does, but am not 100% certain (I'm
not overly familiar with latexmk).

tkoeppe:
There are "regenerate" guides in the Readme.

AaronBallman:
Thank you for the hint about regenerating from the readme, there were indeed modifications which I've added.

tkoeppe:
The way `makegram` works depends on the particular flavour of `sed` on your system. For now, can you edit the grammar changes so that they only contain the relevant diff?

tkoeppe:
Richard has fixed something, can you fetch and rebase? Then the makegram script should work better.

AaronBallman:
I fetched and rebased, which seems to have removed all the inserted whitespace from f586dad.

tkoeppe:
My bad, the change doesn't affect the grammar at all, so no need to regenerate anything. As you were.

PR #660, opened 2016-03-21
[inclusive.scan] Add missing template parameter.

timsong-cpp:
The `T init` overload for `inclusive_scan` is missing a template parameter `class T`. Also fixed the `<numeric>` synopsis in [numeric.ops.overview].

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -8718,7 +8718,7 @@ namespace std {
     OutputIterator inclusive_scan(InputIterator first, InputIterator last,
                                   OutputIterator result,
                                   BinaryOperation binary_op);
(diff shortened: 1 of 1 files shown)


source/numerics.tex
```diff
@@ -8731,7 +8731,7 @@
                                   InputIterator first, InputIterator last,
                                   OutputIterator result,
                                   BinaryOperation binary_op);
-  template<class ExecutionPolicy, class InputIterator, class OutputIterator, class BinaryOperation>
+  template<class ExecutionPolicy, class InputIterator, class OutputIterator, class BinaryOperation, class T>
```
zygoloid: Does this fit horizontally on the page?

PR #662, opened 2016-03-24
[intro.object] For object names, don't refer to 'name' grammar rule, …

Eelis:
…because that one is from [path.generic].

### DIFF
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -540,7 +540,7 @@ created by a \grammarterm{definition}~(\ref{basic.def}), by a
 \grammarterm{new-expression}~(\ref{expr.new}) or by the
 implementation~(\ref{class.temporary}) when needed. The properties of an
 object are determined when the object is created. An object can have a
-\grammarterm{name}~(Clause~\ref{basic}). An object has a \term{storage
+\term{name}~(Clause~\ref{basic}). An object has a \term{storage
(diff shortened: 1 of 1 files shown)


tkoeppe:
Surely the right thing would be to fix the filesystem grammar name rather than meddle with the fundamental object model?

Also, why doesn't the filesystem grammar get indexed? Maybe it's missing from the list of indexable files.

tkoeppe:
Oh, I see, the grammar in iostreams.tex is "non-copied".

Eelis:
The thing is, as far as I can see, there is no "name" grammar rule other than the one in [path.generic]. So even if we convince ourselves it doesn't really refer to that one, then what _does_ it refer to? :)

jwakely:
For consistency with the rest of the paragaph it should use `\term{name}`, rather than simply dropping all formatting from that word.

tkoeppe:
Yes, I can see that. Maybe the commit should be phrased like https://github.com/cplusplus/draft/commit/5c8435cbf3721eefc02eae58543adc73b51d889f

jwakely:
And clearly it refers to the definition of "name" in Clause 3, as indicated by the cross-reference.

Eelis:
Thanks, that makes sense. Updated branch

tkoeppe:
Much better, thanks!

PR #665, opened 2016-03-28
Index entries for [hardware.interference]

jfbastien:


source/support.tex
```diff
@@ -2252,6 +2252,8 @@
 \end{itemdescr}
 
 \rSec2[hardware.interference]{Hardware interference size}
+\indextext{cats!interfering with canines}%
+\indextext{dogs!obliviousness to interference}%
```
zygoloid: Move these to right before the example, so they are kept on the same page.

jfbastien: Done.

PR #666, opened 2016-03-29
[cpp.replace] Remove a bogus grammar term

AaronBallman:
function-like and object-like were being defined as grammar terms when they are not, and this is confusing with the term of art definitions in p10.

### DIFF
--- a/source/preprocessor.tex
+++ b/source/preprocessor.tex
@@ -592,18 +592,15 @@ where all white-space separations are considered identical.
 An identifier currently defined as an
 \indextext{object-like macro|see{macro, object-like}}%
 \indextext{macro!object-like}%
-\grammarterm{object-like}
-macro may be redefined by another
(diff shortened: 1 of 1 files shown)


source/preprocessor.tex
```diff
@@ -592,18 +592,15 @@
 An identifier currently defined as an
 \indextext{object-like macro|see{macro, object-like}}%
 \indextext{macro!object-like}%
-\grammarterm{object-like}
-macro may be redefined by another
+object-like macro may be redefined by another
```
zygoloid: Please add a parenthetical "(see below)" to each of these uses preceding the definition. Please also convert the `\grammarterm`s on lines 668 and 692 to `\defn`s.

AaronBallman:
0b1e789 should be updated properly (and squashed with previous commit).

PR #668, opened 2016-03-30
[sf.cmath] Use \indextext and \indexlibrary instead of \index.

Eelis:


source/numerics.tex
```diff
@@ -10085,9 +10085,9 @@
 and for positive infinity.
 
 \rSec2[sf.cmath.Lnm]{Associated Laguerre polynomials}%
-\index{assoc_laguerre@\tcode{assoc_laguerre}}%
-\index{Laguerre polynomials $\mathsf{L}_n^m$}%
-\index{L nm@$\mathsf{L}_n^m$ (associated Laguerre polynomials)}%
+\indexlibrary{\idxcode{assoc_laguerre}}%
+\indextext{Laguerre polynomials!$\mathsf{L}_n^m$}%
```
zygoloid: Should this really be a `!` rather than a space? IIUC, this isn't specifying a particular kind of Laguerre polynomials. I'd prefer to drop the `L_n^m` entirely rather than add it under a `!`.

source/numerics.tex
```diff
@@ -10085,9 +10085,9 @@
 and for positive infinity.
 
 \rSec2[sf.cmath.Lnm]{Associated Laguerre polynomials}%
-\index{assoc_laguerre@\tcode{assoc_laguerre}}%
-\index{Laguerre polynomials $\mathsf{L}_n^m$}%
-\index{L nm@$\mathsf{L}_n^m$ (associated Laguerre polynomials)}%
+\indexlibrary{\idxcode{assoc_laguerre}}%
```
zygoloid: Would you mind also adding index entries for the `l`- and `f`-suffixed forms here?

Eelis:
Added index entries for the l- and f-suffixed forms :)

PR #636, opened 2016-03-09
Improve consistency of complexity descriptions

tkoeppe:
This is a rebasing of the abandoned PR #305 by @cassioneri.

PR #673, opened 2016-04-02
[string.view, alg.random.sample, numerics] Use \bigoh.

Eelis:


Eelis:
Spotted a few more. Think this is all of them.

tkoeppe:
Looks like this is in the same spirit as #636.

PR #677, opened 2016-04-06
[cpp.predefined] Use \xname in index entry for __cplusplus.

Eelis:
To be consistent, and so that we don't miss an \ungap.


zygoloid:
Does this affect the collation order in the index?

Eelis:
No.

zygoloid:
I went with a somewhat different change that explicitly specifies the collation order for all of these __blah names.

Issue #544, opened 2015-10-08
Inconsistent formatting of `void` in [meta.unary.prop]


timsong-cpp:
Some rows of Table 49 in [meta.unary.prop] uses `\tcode{void}`. Others just use plain `void` without the `\tcode`. They should be made consistent.

PR #684, opened 2016-04-12
[meta.unary.prop] Consistent formatting for 'void'

AlisdairM:
This patch addresses https://github.com/cplusplus/draft/issues/544.
Update all use of 'void' in type traits tables to use the codified
form \tcode{void}.  There was one use of the phrase "void types"
that I deliberately chose to not touch.

A second issue I stumbled over in this edit was inconsistent use
of cv-qualified vs. \cv-qualfied.  The latter seemed preferable,
so I applied that formatting consistently through this file for
all uses of cv-qualified that were not otherwise participating
in mark-up.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -2200,7 +2200,7 @@ namespace std {
 
 \pnum
 A program that necessitates the instantiation of template \tcode{optional} for
-a reference type, or for possibly \cv-qualified types \tcode{in_place_t} or
+a reference type, or for possibly cv-qualified types \tcode{in_place_t} or
 \tcode{nullopt_t} is ill-formed.
 
 \rSec2[optional.object]{\tcode{optional} for object types}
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -2133,7 +2133,7 @@
 
 \pnum
 A program that necessitates the instantiation of template \tcode{optional} for
-a reference type, or for possibly cv-qualified types \tcode{in_place_t} or
+a reference type, or for possibly \cv-qualified types \tcode{in_place_t} or
```
zygoloid: cv should not be italicized except when referencing a metavariable representing a set of cv-qualifiers. That is: "*cv*-qualified" means "qualified with the set of qualifiers named *cv*", whereas "cv-qualified" means "qualified with some unspecified cv-qualifiers". See [basic.type.qualifier]/5.

AlisdairM:
So I should consistently strike that leading '\' on all the cv-qualified terms in those tables, rather than consistently add it?  (i.e., not just for the one line that you highlighted - think I am stating the obvious, but checking before I push an update).

AlisdairM:
Rebased, and consistently struct the \ on "(potentially \cv-qualified)" throughout this file.  Force-pushed an update that should be clean to land, assuming it passes muster.

PR #700, opened 2016-04-14
[dcl.enum] This is not where layout-compatible is defined

AaronBallman:
There's a spurious definition of layout-compatible hiding in dcl.enum when it really lives in basic.types.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -757,7 +757,8 @@ The common initial sequence of \tcode{A} and \tcode{E} is empty.
 \exitexample
 
 \pnum
-Two standard-layout struct (Clause~\ref{class}) types are layout-compatible if
+Two standard-layout struct (Clause~\ref{class}) types are
+\defnx{layout-compatible classes}{layout-compatible!class} if
 their common initial sequence comprises all members and bit-fields of
 both classes~(\ref{basic.types}).
 
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -2017,8 +2017,8 @@ expression of enumeration type from having a value that falls outside
 this range.}
 
 \pnum
-Two enumeration types are \defn{layout-compatible} if they have the same
-underlying type.
+Two enumeration types are \defnx{layout-compatible enumerations}{layout-compatible!enumeration}
+if they have the same underlying type.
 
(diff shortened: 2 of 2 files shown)


zygoloid:
This is the definition of layout-compatible for enumerations, and is referenced by the definition of layout-compatible in general in [basic.types]/11:

"Two types cv1 T1 and cv2 T2 are layout-compatible types if T1 and T2 are the same type, layout-compatible enumerations (7.2), or layout-compatible standard-layout class types (9.2)."

How about this instead:

> Two enumeration types are \defn{layout-compatible enumerations} if they have the same

zygoloid:
It looks like a similar change is warranted in [class.mem]. We probably also want the index entry to be "layout-compatible!enumerations" rather than "layout-compatible enumerations".

AaronBallman:
I thought about that approach, but since we don't appear to use "layout-compatible enumeration" (or "layout-compatible class") anywhere else in the standard aside from this instance, it seemed more sensible to remove the definition from dcl.enum instead of adding one-time use definitions.

zygoloid:
Well, [dcl.enum] is where we define what "layout-compatible" means for enumeration types. The wording in [basic.types] is really just saying that the definition is split across two other clauses. It makes much more sense for the index entries to point to the useful definition, in my opinion.

AaronBallman:
I've come around to your logic, and I think I've implemented what you were suggesting (though I freely admit that index entries are new to me, so I may not have done the entries properly).

source/classes.tex
```diff
@@ -757,8 +757,9 @@
 \exitexample
 
 \pnum
-Two standard-layout struct (Clause~\ref{class}) types are layout-compatible if
-their common initial sequence comprises all members and bit-fields of
+\indextext{layout-compatible!classes}%
+Two standard-layout struct (Clause~\ref{class}) types are \defn{layout-compatible classes}
```
zygoloid: I think the term we put in the index here should be "layout-compatible class" (or maybe "layout-compatible!class") "layout-compatible classes". Use \defnx to specify different index text from the body text.

tkoeppe:
Looks good! @zygoloid: PTAL.

PR #683, opened 2016-04-11
[allocator.adaptor] consolidate memory utilities

AlisdairM:
This patch simply moves the scoped allocator clause to
be adjacent to the other memory and allocator clauses
for a more consistent sub-organization of clause 20.

The patch looks pretty awful thanks to the diff
algorithm treating it a lot of interspersed edits,
where in fact is way a single cut/paste for the
bulk of the test (plus a second one-liner for the
clause 20 table of contents).

Issue #581, opened 2015-12-08
replace `typedef` declarations with *alias-declaration*s in the library


zygoloid:
In library synopses, it would be better to use *alias-declaration*s rather than `typedef` declarations as our "house style".

(As discussed with various parties elsewhere.)

sigfpe:
Amen.-------- Original message --------From: Richard Smith <notifications@github.com> Date: 12/8/2015  2:54 PM  (GMT-08:00) To: cplusplus/draft <draft@noreply.github.com> Subject: [draft] replace `typedef` declarations with *alias-declaration*s in
  the library (#581) 
In library synopses, it would be better to use alias-declarations rather than typedef declarations as our "house style".

(As discussed with various parties elsewhere.)

AlisdairM:
Is this something we want to do for C++17?  It is something I am, personally, keen to see happen, and if we believe this is entirely editorial, I am prepared to do the work of submitting many pull requests between now and Oulu.  The main question, once we confirm that we want to proceed, is what granularity should I exercise for a single pull request?

I would want to go fine-grained enough that a single pull request is (relatively) easy to review, but not so fine-grained that it swamps the queue.  The obvious granularity would be one clause at a time, as that is one file/request - but I fear that may be a little too big a bite to get right.

There is also the question of whether we should run it past LWG, or whether your previous discussion with various parties gives you confidence to proceed already?

sigfpe:
Go for it! :-)-------- Original message --------From: Alisdair Meredith <notifications@github.com> Date: 4/11/2016  4:56 PM  (GMT-08:00) To: cplusplus/draft <draft@noreply.github.com> Cc: sigfpe <sig.fpe@axiomatics.org> Subject: Re: [cplusplus/draft] replace `typedef` declarations with
  *alias-declaration*s in the library (#581) 
Is this something we want to do for C++17?  It is something I am, personally, keen to see happen, and if we believe this is entirely editorial, I am prepared to do the work of submitting many pull requests between now and Oulu.  The main question, once we confirm that we want to proceed, is what granularity should I exercise for a single pull request?

I would want to go fine-grained enough that a single pull request is (relatively) easy to review, but not so fine-grained that it swamps the queue.  The obvious granularity would be one clause at a time, as that is one file/request - but I fear that may be a little too big a bite to get right.

There is also the question of whether we should run it past LWG, or whether your previous discussion with various parties gives you confidence to proceed already?

keryell:
[...]

PR #704, opened 2016-04-15
[18-30] Replace typedefs with alias

AlisdairM:
This set of changes replace (almost) all use of 'typedef' in the library with a type alias instead:
   typedef Original NewName;   ->    using NewName = Original;

Attention was paid to retaining table-like formatting where present, which is worth reviewing in case I made idiosyncratic choices.

Clause 20 has NOT been applied, as that clause is churning a bit at the moment, and I don't want a conflict there to hold up a this review.  The most likely conflict is my own pull request:
   https://github.com/cplusplus/draft/pull/683
I will happily rebase and complete the work once that branch lands (or is deferred again until this branch completes)

A second set of changes that should follow are to update the text that refers to typedefs to perhaps use a term more consistent with the type alias syntax applied here - but I defer to the project editors to suggest such phrasing.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -2639,17 +2639,18 @@ namespace std {
   template <class T, size_t N>
   struct array {
     //  types:
-    typedef T&                                    reference;
-    typedef const T&                              const_reference;
-    typedef @\impdefx{type of \tcode{array::iterator}}\itcorr[-1]@                iterator;
-    typedef @\impdefx{type of \tcode{array::const_iterator}}\itcorr[-1]@                const_iterator;
-    typedef size_t                                size_type;
-    typedef ptrdiff_t                             difference_type;
-    typedef T                                     value_type;
-    typedef T*                                    pointer;
-    typedef const T*                              const_pointer;
-    typedef std::reverse_iterator<iterator>       reverse_iterator;
-    typedef std::reverse_iterator<const_iterator> const_reverse_iterator;
+    using value_type             = T;
+    using pointer                = T*;
+    using const_pointer          = const T*;
+    using reference              = T&;
+    using const_reference        = const T&;
+    using size_type              = size_t;
+    using difference_type        = ptrdiff_t;
+    using iterator               = @\impdefx{type of \tcode{array::iterator}}@;
+    using const_iterator         = @\impdefx{type of \tcode{array::const_iterator}}@;
+    using reverse_iterator       = std::reverse_iterator<iterator>;
+    using const_reverse_iterator = std::reverse_iterator<const_iterator>;
+
 
     T       elems[N];           // \expos
 
@@ -2885,18 +2886,18 @@ namespace std {
   class deque {
   public:
     // types:
-    typedef value_type&                                         reference;
-    typedef const value_type&                                   const_reference;
-    typedef @\impdefnc@                              iterator;        // see \ref{container.requirements}
-    typedef @\impdefnc@                              const_iterator;  // see \ref{container.requirements}
-    typedef @\impdefnc@                              size_type;       // see \ref{container.requirements}
-    typedef @\impdefnc@                              difference_type; // see \ref{container.requirements}
-    typedef T                                                   value_type;
-    typedef Allocator                                           allocator_type;
-    typedef typename allocator_traits<Allocator>::pointer       pointer;
-    typedef typename allocator_traits<Allocator>::const_pointer const_pointer;
-    typedef std::reverse_iterator<iterator>                     reverse_iterator;
-    typedef std::reverse_iterator<const_iterator>               const_reverse_iterator;
+    using value_type             = T;
+    using allocator_type         = Allocator;
+    using pointer                = typename allocator_traits<Allocator>::pointer;
+    using const_pointer          = typename allocator_traits<Allocator>::const_pointer;
+    using reference              = value_type&;
+    using const_reference        = const value_type&;
+    using size_type              = @\impdef@; // see \ref{container.requirements}
+    using difference_type        = @\impdef@; // see \ref{container.requirements}
+    using iterator               = @\impdef@; // see \ref{container.requirements}
+    using const_iterator         = @\impdef@; // see \ref{container.requirements}
+    using reverse_iterator       = std::reverse_iterator<iterator>;
+    using const_reverse_iterator = std::reverse_iterator<const_iterator>;
 
     // \ref{deque.cons}, construct/copy/destroy:
     deque() : deque(Allocator()) { }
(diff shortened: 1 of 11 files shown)


AlisdairM:
This pull request addresses https://github.com/cplusplus/draft/issues/581

AlisdairM:
Note that I can break this down into a dozen distinct file-specific pull requests if that is preferred; the suggestion in the issue is that a single pull large pull request was the way to go though.

The are no changes to clause 19 and 25, and I deliberately stayed out of clause 29 as that clause is deliberately written using C idioms.

jwakely:
Thanks. I think a single pull request makes more sense. I'll go through this asap and merge.

source/containers.tex
```diff
@@ -2888,18 +2889,18 @@
...
-    typedef typename allocator_traits<Allocator>::const_pointer const_pointer;
-    typedef std::reverse_iterator<iterator>                     reverse_iterator;
-    typedef std::reverse_iterator<const_iterator>               const_reverse_iterator;
+    using value_type             = T;
+    using allocator_type         = Allocator;
+    using pointer                = typename allocator_traits<Allocator>::pointer;
+    using const_pointer          = typename allocator_traits<Allocator>::const_pointer;
+    using reference              = value_type&;
+    using const_reference        = const value_type&;
+    using size_type              = @\impdefnc@; // see \ref{container.requirements}
```
tkoeppe: We can get rid of the non-correcting forms here now that the alignment at the ragged edge doesn't matter anymore.

source/locales.tex
```diff
@@ -4689,9 +4689,9 @@
     class InputIterator = istreambuf_iterator<charT> >
   class money_get : public locale::facet {
   public:
-    typedef charT               char_type;
-    typedef InputIterator       iter_type;
-    typedef basic_string<charT> string_type;
+    using   char_type = charT;
```
tkoeppe: Why right-align? Isn't that a bit cute? We used to left-align everything previously.

source/containers.tex
```diff
@@ -2639,17 +2639,18 @@
...
-    typedef std::reverse_iterator<iterator>       reverse_iterator;
-    typedef std::reverse_iterator<const_iterator> const_reverse_iterator;
+    using value_type             = T;
+    using pointer                = T*;
+    using const_pointer          = const T*;
+    using reference              = T&;
+    using const_reference        = const T&;
+    using size_type              = size_t;
+    using difference_type        = ptrdiff_t;
+    using iterator               = @\impdefx{type of \tcode{array::iterator}}\itcorr[-1]@;
```
tkoeppe: No more need for the italic correction at the end of the line now!

source/containers.tex
AlisdairM: If I understand correctly, you are saying replace '\impdefnc' with '\impdef'?
If so, is that a general rule, or applies only to header synopses and other code snippets?

source/locales.tex
AlisdairM: It is a bit cute. I largely threw it in to see the reaction, as we work out how to align the new syntax.  I like it, but suspected it might not be the preferred choice.  Unless I hear support, I will left-align any remaining right-aligned identifiers, but retain the alignment on the '='.

source/containers.tex
AlisdairM: I'll admit, I am mostly moving text around and don't recognize all of the escape codes ;)
Will update, thanks.

source/locales.tex
sigfpe: Left-align is good.
-------- Original message --------From: Alisdair Meredith <notifications@github.com> Date: 4/16/2016  9:34 AM  (GMT-08:00) To: cplusplus/draft <draft@noreply.github.com> Subject: Re: [cplusplus/draft] [18-30] Replace typedefs with alias (#704) 
In source/locales.tex:
> @@ -4689,9 +4689,9 @@
>      class InputIterator = istreambuf_iterator<charT> >
>    class money_get : public locale::facet {
>    public:
> -    typedef charT               char_type;
> -    typedef InputIterator       iter_type;
> -    typedef basic_string<charT> string_type;
> +    using   char_type = charT;

It is a bit cute. I largely threw it in to see the reaction, as we work out how to align the new syntax.  I like it, but suspected it might not be the preferred choice.  Unless I hear support, I will left-align any remaining right-aligned identifiers, but retain the alignment on the '='.

AlisdairM:
Final (I hope!) patch adds the updates for <utility>.  I will now wait for this pull request to land before once-again rebasing https://github.com/cplusplus/draft/pull/683

source/containers.tex
tkoeppe: Correct. "nc" stands for "no (italic) correction", which affects the column alignment. You ought to be able to see the difference (look for how characters (fail to) align on the grid without the correction). But now that the relevant text is at the end of the line, there's nothing left to be aligned, and we don't need to control the italic correction.

That was one of the benefits of changing `typedef` to `using` that @zygoloid pointed out in his first email about this subject a few months ago if I recall correctly.

tkoeppe:
Can you squash the few fixup commits you have into the respective main commits?

AlisdairM:
Squash them *all* as a single commit, or just the last couple?

tkoeppe:
Just the ones that are fixups - 0f6d327 and 8933e27.

lv-zheng:
typo in `source/support.tex` line 132: `using` is misspelt as `ussing`

AlisdairM:
Fix the typo (thanks lv-zheng) and force-pushed a single squash-commit of the whole branch.  No other changes.

zygoloid:
I want to go ahead with this, but not until after Oulu. Sorry if this requires extra rebasing as a result!

tkoeppe:
@zygoloid: Last standard, you didn't want to make major formatting changes after the CD went out (e.g. I remember fixing up the vertical whitespace problems with the old heading styles). Will there still be room for changes like this after Oulu? Will you only create the CD after the meeting?

AlisdairM:
If we plan to land this after Oulu, is there anything I can do to land it /before/ papers start landing from the meeting.  It will be much easier to update the standard for incoming papers with this in place (and editorial discretion when applying updates) than taking another pass to totally redo the work, at an undetermined time after all Oulu papers have landed again.

zygoloid:
If we're going to land this before we land the Oulu motions, we may as well do so now. I think you're right, handling the merge as Oulu motions land is not going to be very difficult. (Sorry for not merging this sooner, I'd missed that the pre-Oulu mailing deadline was 2pm not end-of-day, and was putting together the new working draft in a bit of a rush.)

source/containers.tex
```diff
@@ -8345,11 +8346,11 @@
...
-    typedef typename Container::value_type            value_type;
-    typedef typename Container::reference             reference;
-    typedef typename Container::const_reference       const_reference;
-    typedef typename Container::size_type             size_type;
-    typedef          Container                        container_type;
+    using value_type      = typename Container::value_type;
+    using reference       = typename Container::reference;
+    using const_reference = typename Container::const_reference;
+    using size_type       = typename Container::size_type;
+    using container_type  =          Container;
```
zygoloid: Maybe remove the whitespace before `Container` here. Now that we're not aligning anything else after the `=`, this looks a bit odd.

source/future.tex
```diff
@@ -1271,7 +1271,7 @@
 To enable old function adaptors to manipulate function objects
 that take one or two arguments,
 many of the function objects in this standard
-correspondingly provide typedefs
+correspondingly provide type aliases
```
zygoloid: Maybe `\grammarterm{typedef-name}{s}`, to be clear that we don't care whether they're introduced by a `typedef` declaration or by an *alias-declaration*? "Type alias" is not a defined term in the standard.

AlisdairM: Thanks - I think I may want to apply this recommendation in several places that I did not touch, as I did not have a better way to update the existing wording talking about typedefs.  I'll take another pass tonight.

source/containers.tex
AlisdairM: I like my 2-dimensional code :)  But defer to the actual project editor on style choice - will update this tonight.

AlisdairM:
Corrected Richard's issues above, and deployed the \grammarterm{typedef-name}{s} in several other places.  Tried to avoid reformatting around long lines, but at least one paragraph was re-wrapped.

Rebased, squashed, and --force pushed again.

source/iterators.tex
```diff
@@ -3289,15 +3294,15 @@
...
-    typedef basic_ostream<charT,traits>   ostream_type;
+    using iterator_category = output_iterator_tag;
+    using value_type        = void;
+    using difference_type   = void;
+    using pointer           = void;
+    using reference         = void;
+    using char_type         = charT;
+    using traits_type       = traits;
+    using streambuf_type    = basic_streambuf<charT,traits>;
+    using ostream_type = basic_ostream<charT,traits>;
```
zygoloid: Please align this.

AlisdairM: Rebased again (just to be safe) and aligned the line highlighted above - should land cleanly.

Issue #709, opened 2016-04-17
[meta.rel] New 'is_callable' entries overflow the middle cell of table


AlisdairM:
The pack-expansion ellipsis is overflowing the 'Condition' column of table 54 - Type Relation Predicates in [meta.rel] for the two new traits added in Jacksonville, is_callable, and is_nothrow_callable.  I don't see any obvious way to improve matters either, so filing an issue rather than a pull request.

K-ballo:
Using a shorter name for `ArgTypes` perhaps?

jwakely:
Fixed by 535ff921a4c61da91716c9e18d42ffb26cc53c24

PR #710, opened 2016-04-17
[meta.rel] Fix for for INVOKE in table 54

AlisdairM:
Fix the font for the use of INVOKE in the two new additions
to table 54, is_callable and is_nothrow_callable.  Considered
adding a new \invoke command and applying that consistently,
but decided I am not a LaTeX hacker yet, and took the easy
way out.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -13288,7 +13288,7 @@ with a BaseCharacteristic of
 
 \tcode{template <class Fn, class... ArgTypes, class R>}\br
  \tcode{struct is_callable<Fn(ArgTypes...), R>;}                      &
- The expression \tcode{INVOKE(declval<Fn>(), declval<ArgTypes>()..., R)}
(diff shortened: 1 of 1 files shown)


AlisdairM:
I opened a separate issue, https://github.com/cplusplus/draft/issues/709, for the overflowing of the table column.

tkoeppe:
@zygoloid, @jwakely: PTAL

jwakely:
N.B. we already have:

    \newcommand{\EXPLICIT}{\textit{\texttt{EXPLICIT}}}

And for the Networking TS I defined:

    \newcommand{\DEDUCED}{\textit{\texttt{DEDUCED}}}

So I would be in favour of adding `\INVOKE` too. I didn't check, but it might need to be slightly different, because EXPLICIT and DEDUCED are only(?) used inside codeblocks, whereas we sometimes use INVOKE in the middle of prose.

PR #717, opened 2016-04-27
[ext.manip] fix typo in put_money description

timsong-cpp:
There's no way `os << put_money(...)` could be a formatted *input* function; presumably a formatted *output* function is meant. Also add cross reference to [ostream.formatted.reqmts].


tkoeppe:
@zygoloid, @jwakely: PTAL

PR #724, opened 2016-05-16
[class.abstract] The intro para seems like it should be a note - noth…

faisalv:
…ing seems prescriptive about it.


tkoeppe:
@jwakely, @zygoloid: Please have a look.

jwakely:
LGTM but I'll leave @zygoloid to decide.

PR #726, opened 2016-05-17
[namespace.udecl] tweak the example in para 17 so that the comment is…

faisalv:
… clear as to the reason for the ambiguity

  - the current comment could be interpreted as suggesting that the result of member-name-lookup (i.e. the declaration-set of S(x,D)) is ambiguous, which is certainly not the case -- the reason for the ambiguity is specified in [expr.ref]/5 -- and hopefully better reflected in the new comment.
  - add some additional examples that highlight the nuances of how the wording rejects or resolves certain member  ambiguity.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -2920,12 +2920,13 @@ visible in the scope of at least one of the direct base classes of the
 class where the \grammarterm{using-declaration} is specified. \enternote
 Because a \grammarterm{using-declaration} designates a base class member
 (and not a member subobject or a member function of a base class
-subobject), a \grammarterm{using-declaration} cannot be used to resolve
+subobject), a \grammarterm{using-declaration} cannot be used to resolve all
 inherited member ambiguities. For example,
 
 \begin{codeblock}
-struct A { int x(); };
-struct B : A { };
+struct A { int x(); int y(); int z(); };
+struct A1 : A { int y(); int z(); };
+struct B : A, A1 { using A1::z; };
 struct C : A {
   using A::x;
   int x(int);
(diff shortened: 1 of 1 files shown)


tkoeppe:
Can you please attach a diffpdf screenshot?

tkoeppe:
@faisalv: Ping.

source/declarations.tex
```diff
@@ -2935,8 +2935,12 @@
   using C::x;
   int x(double);
 };
-int f(D* d) {
-  return d->x();    // ambiguous: \tcode{B::x} or \tcode{C::x}
+void f(D* d) {
+   d->x();       // ambiguous: \tcode{x}'s direct parent \tcode{A} is an ambiguous base of the naming class \tcode{D}
```
tkoeppe: We never use the word "parent" in the context of classes. (We only use it in the context of threads and directories.) "Parent-child" is a *terrible* metaphor for class inheritance. A better metaphor is "base-derived".

faisalv:
just seeing this - never done it before - will look into it - thanks!

Faisal Vali

source/declarations.tex
faisalv: agreed - that parent sinifies the container of the declaration - so in
struct A { void f(); }; - regardless of how member lookup finds 'f' in
whichever hierarchy - the direct parent would be 'A'.  Should I use direct
containing class?

Faisal Vali

faisalv: Hmm - what about x's declaring class 'A' - i think i like prefer that in many ways

tkoeppe: I'm not really sure. Technically correct terminology would be good, but brevity is also good.

@zygoloid, @villevoutilainen: Wording masters, any suggestions on improving this example code?

villevoutilainen: the class \tcode{A} where \tcode{x} is declared is an ambiguous base of the naming class \tcode{D} ?

zygoloid: ambiguous: \tcode{x} is a member of the ambiguous base class \tcode{A} of the naming class \tcode{D}

villevoutilainen: The reason why I didn't want to go with talking about members is that it's not so clear, as some people might think inherited members are members. :) Hence the talk about the declaration.

zygoloid: We could use "direct member" instead. We'll have a CWG issue defining that term landing any day now, and we already have two uses of it anticipating its addition.

villevoutilainen: Sounds good to me.

faisalv: Thanks for the feedback.  I made some changes.  Let me know if you folks think this is better or worse or too much?  Also do you still need a diffpdf (if so, if you could point me to any resources that show a non-cumbersome way to do it in windows would certainly be appreciated)?

faisalv:
2016-06-09 20_43_31-clipboard

OK - I added a pdf preview - does this help?

source/declarations.tex
faisalv: Hmm - still trying to figure these comment threads out - so apologies if this is double spam but here is a snapshot to give you a visual:
2016-06-09 20_43_31-clipboard

zygoloid: The explanation and local terms are too much for my tastes, and I think they detract from the example as they're not directly linked to the terminology in the normative wording.

faisalv:
I tried to rebase and squash commits - hopefully this worked.  I also amended the note and added an example to show how using-declarations can indeed be used to disambiguate inherited lookup ambiguities.

This is what the preview looks like:
2016-06-09 20_43_31-clipboard

source/declarations.tex
faisalv: Fair enough - I'm not entirely unsympathetic to that POV.  I might get motivated to re-style it at a future date - but if anyone sees any value in the general idea, feel free to take a crack at it.  Thanks for the feedback.

tkoeppe:
Thanks a lot for making the screenshot! That's really nice. With something that is difficult to lay out, it's always nice to see not just the source but also how it turns out.

PR #727, opened 2016-05-17
Fix incorrect variable references

mclow:
Two move operations for the allocators refer to `a` or `a1`, though the code uses `u`.  Fix the description to match the code.
Also, change the phrase "equals" to "is equal to", which matches other uses in the standard (see [unique.ptr]/4.1 for an example).


source/lib-intro.tex
```diff
@@ -1821,12 +1821,12 @@
 \tcode{X u = std::move(a);} &
```
zygoloid: Can you also fix the two similar cases above (`X u = a;` and `X u(b);`)?

PR #733, opened 2016-05-29
Define \defnnote for notes in "Terms and definitions"

18192658826:


source/macros.tex
```diff
@@ -507,3 +507,4 @@
 \subsubsection[#1]{\hfill[#2]}\vspace{-.3\onelineskip}\label{#2}\textbf{#1}\\%
 }
 \newcommand{\defncontext}[1]{\textlangle#1\textrangle}
+\newcommand{\defnnote}[2]{{\small Note #1 to entry: #2}}
```
tkoeppe: We had discussed this, and as much as we all love ISO drafting directive normative annex D, zygoloid called this "overreaching" if I recall correctly.

tkoeppe:
If it's all the same to everyone, I'd recommend closing this PR. If a national body asks for this format, fine, but otherwise it seems like busywork.

jwakely:
Agreed. There's no rationale given for why we might want this (I think I left it unmerged in that branch for a reason).

PR #735, opened 2016-06-02
[lex.pptoken] fix term

timsong-cpp:
"preprocessor token" -> "preprocessing token"


tkoeppe:
Seems sensible; we never say "preprocessor token" anywhere else.

PR #740, opened 2016-06-08
[18-28] Fix index entry for constructors

AlisdairM:
The preferred way to list constructors in the index is under the
non-code font term 'constructor', but many classes retain old
text that lists the constructor under the class name as the same
class name, in a code font.  Worse, there are quite a few classes
that have fallen into a hybrid approach, with both listings, which
can be confusing. This change-set should catch every constructor
indexed in the old style and consistently transform it to the new.

One additional fix is that allocator_traits mis-indexed the static
member functions 'construct' and 'destroy' as the plain-text
constructor and destructor.  This patch correctly indexes them as
their correct name, in a code-font.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -3005,8 +3005,7 @@ namespace std {
 
 \rSec3[deque.cons]{\tcode{deque} constructors, copy, and assignment}
 
-\indexlibrary{\idxcode{deque}!\idxcode{deque}}%
-\indexlibrary{\idxcode{deque}!\idxcode{deque}}%
+\indexlibrary{\idxcode{deque}!constructor}%
 \begin{itemdecl}
 explicit deque(const Allocator&);
 \end{itemdecl}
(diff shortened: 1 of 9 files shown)


AlisdairM:
Two particular features I will call attention to:
1) The constructor for the 'ios_base' nested classes now have their classes qualified with 'ios_base::'.  Previously the qualifier was only on the constructor, so that the unqualified name also gained a place in the index.  I think this form is preferred, but my judgement is suspect!
2) many index entries have a trailing '%' after the closing '}', but many do not.  I don't really know LaTeX, so I left these entries as I found them - although I suspect there may be a desire to make this convention consistent one way or the other.  I observed no clear distinction for when the '%' was present or absent.

tkoeppe:
@AlisdairM If it's *after* the closing brace, then it's not part of the index. The `%` is the line-comment character for LaTeX, and putting it at the end of the line means that there is no whitespace. Without it, the newline counts as a whitespace. To see the effect, look for where the index macro sits in the body text and see if there's too much space; that may be a sign that you need to comment out the newline.

AlisdairM:
Rebases and conflicts resolved - apparently you can't make edits to adjacent lines in different branches without confusing the diff algorithm.  Manually reviewed diff to confirm rebase did not do anything unexpected - ready for review again.

jwakely:
I also fixed `scoped_allocator_adaptor::destroy` which was indexed as "destructor".

PR #741, opened 2016-06-09
[18-27] Remove tcode from index entries

AlisdairM:
Several old index entries use '\tcode' to force a code
font on their sub-entry.  This is no longer necessary,
as the LaTeX scripts produce the correct font for an
'\idxcode' entry.  However, using '\tcode' inconsistently
will produce duplicate entries, and entries that do not
sort correctly.  This change set consistently applies
'\idxcode' in preference to '\tcode' to eliminate the
redundant and mis-sorted entries.

Note that this change-set does NOT change '\tcode'
entries for constructors, as those are handled separely
by https://github.com/cplusplus/draft/pull/740

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -3085,7 +3085,7 @@ using the specified allocator.
 
 \rSec3[deque.capacity]{\tcode{deque} capacity}
 
-\indexlibrary{\idxcode{resize}!\tcode{deque}}%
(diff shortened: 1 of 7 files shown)


tkoeppe:
Looks good - this is a highly desirable change. Thanks a lot!

Issue #629, opened 2016-03-03
[meta.trans.other] Table 57 uses the layout for type predicates (Tables 47-49,51) instead of type transformations/queries (Tables 50,52-56)


cubbimew:
Table 57 "Other transformations" in 20.10.7.6[meta.trans.other] has three columns named "Template", "Condition", "Comments", which is how predicate tables (Table 47,48,49,51) are laid out. However, the traits listed in Table 57 are not predicates and do not test any condition. These are transformation traits, and the rest of those traits use two-column layout "Template", "Comments" (Tables 50,52,53,54,55,56)
The contents of the column named "Condition" in Table 57 where not empty appear to be the requirements on the template type parameters, which means they should move into Comments after the word "Requires:" as done for other transformation traits, namely in Tables 50 and 54

jwakely:
Fixed by #747

PR #747, opened 2016-06-10
[meta.trans.other] Reformat Other Transformations table

AlisdairM:
This PR applies the change recommended by https://github.com/cplusplus/draft/issues/629.

So far, I like the change giving the descriptions more room to
breathe, but for come reason the first column has become a
little narrower, which is messing with the format a little.

Generally, I think this looks like a good fix, but may use a
little more polish to get the presentation just right.


jwakely:
Fixed at 863cbd1add491eb3cfef66d438f5ca79e3d602f4

PR #749, opened 2016-06-11
[namespace.udecl]/17 Fix the note and comment to better reflect that …

faisalv:
…while using declarations can aid in resolving certain member-lookup ambiguities, they are not always sufficient.

I admittedly got a little carried away by my prior attempt at this - hopefully this one is more restrained (pls let me know if the comment is still too wordy, and whether you want me to truncate it further or have suggestions)

2016-06-11 10_25_51-f__c -draft_draft_source_declarations tex - texstudio

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -2937,10 +2937,10 @@ struct D : B, C {
 };
 
 void f(D* d) {
-  d->x();    // ambiguous: selected overload \tcode{x()} from the result of name lookup of \tcode{x} in implicit naming class \tcode{D}
-             // (i.e. \{A::x(), C::x(int), D::x(double)\}) is a direct member of class \tcode{A} which is an ambiguous base 
-             // of the naming class \tcode{D}
-  d->C::x(); // OK
+  d->x();    // ambiguous: while member lookup of \tcode{x} in implicit naming class \tcode{D} is unambiguous, the selected 
+             // overload of \tcode{x} is a direct member of \tcode{A}, which is an ambiguous base of the naming class \tcode{D}
+  d->C::x(); // OK: the selected \tcode{x} is a direct member of \tcode{A} which is an unambiguous base of the naming class \tcode{C}
+             // which in turn is an unambiguous base of the object expression's type \tcode{D}
 }
 \end{codeblock}
 \exitnote
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -2921,7 +2921,7 @@ class where the \grammarterm{using-declaration} is specified. \enternote
 Because a \grammarterm{using-declaration} designates a base class member
 (and not a member subobject or a member function of a base class
 subobject), a \grammarterm{using-declaration} cannot be used to resolve
-inherited member ambiguities. For example,
+inherited member ambiguities by itself. For example,
 
 \begin{codeblock}
 struct A { int x(); };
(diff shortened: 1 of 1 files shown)


source/declarations.tex
```diff
@@ -2935,8 +2935,12 @@
   using C::x;
   int x(double);
 };
-int f(D* d) {
-  return d->x();    // ambiguous: \tcode{B::x} or \tcode{C::x}
+
+void f(D* d) {
+  d->x();    // ambiguous: selected overload \tcode{x()} from the result of name lookup of \tcode{x} in implicit naming class \tcode{D}
+             // (i.e. \{A::x(), C::x(int), D::x(double)\}) is a direct member of class \tcode{A} which is an ambiguous base
```
tkoeppe: Suggestions:

1. Every `x()` should probably just be `x` -- distinguish the name from the call expression. This example is about names.

2. Make each term inside the braces appear in code font, e.g. `\tcode{A::x}` etc.

tkoeppe:
(@faisalv: Squash and force-push?)

source/declarations.tex
```diff
@@ -2935,8 +2935,12 @@
   using C::x;
   int x(double);
 };
-int f(D* d) {
-  return d->x();    // ambiguous: \tcode{B::x} or \tcode{C::x}
+
+void f(D* d) {
+  d->x();    // ambiguous: while member lookup of \tcode{x} in implicit naming class \tcode{D} is unambiguous, the selected
```
zygoloid: This is true regardless of the presence or absence of the *using-declaration*s in this example.

source/declarations.tex
```diff
@@ -2921,7 +2921,7 @@
 Because a \grammarterm{using-declaration} designates a base class member
 (and not a member subobject or a member function of a base class
 subobject), a \grammarterm{using-declaration} cannot be used to resolve
-inherited member ambiguities. For example,
+inherited member ambiguities by itself. For example,
```
zygoloid: This seems to be suggesting that the using-declaration may help when combined with something else. Is that actually true? Can we extend the example to show that, instead?

source/declarations.tex
faisalv: Yes - the using declaration simply helps to guide member lookup so that x() is not hidden.

source/declarations.tex
faisalv: I suppose what I was trying to get at was the following:
struct A { int x(); };
struct B : A { int x(B); };
struct C : A {  using A::x; int x(C); };
struct D: B, C {
  using C::x;
  using B::x;
};
D d;
d.x(C{}); // OK - because using declaration helps resolve the inherited member lookup ambiguities
d.x(); // NOT OK - because of conversion of D to base class
d.C::x(); // OK - using declaration brings x() into C's scope and and A is unambiguous base of C which is unambiguous of D

Does this seem reasonable - or should we drop the 'by itself'

tkoeppe:
@zygoloid, @faisalv: can you two figure out what to do with this PR?

jensmaurer:
@faisalv: Please rebase.

tkoeppe:
@jensmaurer: This has been stuck for a long time. Could you perhaps take an unbiased look and check whether a) the original problem is actually bad, and b) perhaps draft some alternative wording that we could use instead?

jensmaurer:
It seems to me what we really want to say here is that the using-declaration does help with name lookup issues, but doesn't help with the ambiguous base class conversion (from D to A). See pull request #1254.

PR #754, opened 2016-06-14
[strings,containers] Add iterators to index of implemenetation define…

AlisdairM:
Add iterators to index of implemenetation defined behavior

This patch slavishly copies the tag used for array::iterator, without
understanding if a better markup might be available.  One consequence
is that we get two entries for basic_string_view::const_iterator.
Another is that the index around type names appears as two sorted
subsequences, rather than one sorted sequence.

An additional patch will follow for the remaining implementation-defined
types that are not present in the index, once the preferred markup is
reviewed in this pull request.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -2646,8 +2646,8 @@ namespace std {
     using const_reference        = const T&;
     using size_type              = size_t;
     using difference_type        = ptrdiff_t;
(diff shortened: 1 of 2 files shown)


zygoloid:
The index collation issues are pre-existing; we can fix those separately.

PR #758, opened 2016-06-17
Remove std:: from normative wording

tkoeppe:


source/strings.tex
```diff
@@ -895,8 +895,8 @@
 
     using iterator               = @\impdef@; // See \ref{container.requirements}
     using const_iterator         = @\impdef@; // See \ref{container.requirements}
-    using reverse_iterator       = std::reverse_iterator<iterator>;
-    using const_reverse_iterator = std::reverse_iterator<const_iterator>;
+    using reverse_iterator       = reverse_iterator<iterator>;
+    using const_reverse_iterator = reverse_iterator<const_iterator>;
     static const size_type npos  = -1;
```
AlisdairM: The std:: is actually required here, as the typedef has the same name as the thing it is aliasing, in a class from the same namespace - omitting it flags an error on g++, but is ignored by Clang.

tkoeppe: Excellent point, reverting!

PR #762, opened 2016-06-22
[impldef] Fixed index entry collation to support keys containing code

tkoeppe:
image


zygoloid:
Relative orders of "type of filesystem [...]" and "meaning of attribute [...]" still look incorrect, even with this change. Is the screenshot out of date?

tkoeppe:
Complete nonsense.

PR #763, opened 2016-06-23
[macros,intro] Present example/note formatting better

tkoeppe:


tkoeppe:
One single diff:

image

source/macros.tex
```diff
@@ -144,18 +144,10 @@
...
-\newenvironment{example}[1][Example]{%
-  [\,\textit{#1:}\space
-}{%
-  \textit{\,---\,end example}\,]\xspace
-}
-
+\newcommand{\noteintro}[1]{[\,\textit{#1:}\space}
+\newcommand{\noteoutro}[1]{\textit{\,---\,end #1}\,]}
+\newenvironment{note}[1][Note]{\noteintro{#1}}{\noteoutro{note}\xspace}
+\newenvironment{example}[1][Example]{\noteintro{#1}}{\noteoutro{example}\xspace}
```
zygoloid: Do we really need the xspace? How often are we relying on that?

tkoeppe: The present change is non-semantic and urgent, because it fixes a regression.

I plan to investigate improvements to the environment right after this goes in.

Issue #52, opened 2013-04-15
Inconsistent null values throughout the draft


sdutoit:
From an email to cxxeditor@ (and pointed out by various people in the past):

I noticed following inconsistent usage of null and NULL in the current C++ standard (checked against N3337)

20.7.2.2.1/16 Uses NULL when i think it should be null to be consistent.
20.8.11.2.1/8 Same issue.

20.8.11.2.1/15 & 22 Uses NULL to compare where in other places in the std 0 and nullptr are used.

For example
    20.7.1.2.5 compares against nullptr
    20.7.2.2.1/22 compares against 0 (there are many more such)

There are many other NULL uses in the standard that are not clear that they should be NULL.

tkoeppe:
Does that fix this issue? Can it be closed?

PR #769, opened 2016-06-23
Replace NULL with "null pointer value"

tkoeppe:
Fixes issue #52.

Diff:

image

image

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -14541,9 +14541,9 @@ Table~\ref{tab:iostreams.hdr.cstdio} describes header \tcode{<cstdio>}.
 \end{libsyntab6}
 
 \pnum
-Calls to the function \tcode{tmpnam} with an argument of \tcode{NULL} may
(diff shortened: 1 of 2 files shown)


zygoloid:
This is still wrong -- we should say "an argument that is a null pointer value". We don't care how the null pointer value was written in the source code. But maybe fixing that is a lost cause. Thoughts?

tkoeppe:
PTAL, will upload screenshot shortly.

tkoeppe:
New screenshots:

image

image

timsong-cpp:
@zygoloid But neither `NULL` nor `nullptr` is a null pointer value. The *parameter* has a null pointer value because it is initialized by a null-pointer-constant *argument*, but the argument need not be a null pointer value.

tkoeppe:
@timsong-cpp: yeah... I'm not sure. Even the wording "an `mbstate_t*` argument" is odd. It suggests that the argument already has pointer type. Meh, maybe that's good enough.

PR #771, opened 2016-06-27
[utility,iterator] Apply p0174r2 deprecating vestigial library compon…

AlisdairM:
…ents

Moves text (with minor tweaks, per p0174r2) to Annex D
for deprecation, and updates the indices accordingly.

A drive-by fix in allocator so that operator==/!= in
the synopsis use the same parameter names in the
definition.

Moved another example using std::iterator into the deprecated clause,
as it is the appropriate location, but was missed in the original paper.

### DIFF
--- a/source/future.tex
+++ b/source/future.tex
@@ -1628,3 +1628,404 @@ template <class Predicate>
 \begin{itemdescr}
 \pnum \returns \tcode{binary_negate<Predicate>(pred)}.
 \end{itemdescr}
+
+\rSec1[depr.default.allocator]{The default allocator}
+
+\pnum
+The following members and explicit class template specialization are defined in
+addition to those specified in \ref{default.allocator}:
+
+\indexlibrary{\idxcode{allocator}}%
+\begin{codeblock}
+namespace std {
+  // specialize for \tcode{void}:
+  template <> class allocator<void> {
+  public:
+    using value_type    = void;
+    using pointer       = void*;
+    using const_pointer = const void*;
+    // reference-to-\tcode{void} members are impossible.
+
+    template <class U> struct rebind { using other = allocator<U>; };
+  };
+
+  template <class T> class allocator {
+   public:
+    using size_type       = size_t;
+    using difference_type = ptrdiff_t;
+    using pointer         = T*;
+    using const_pointer   = const T*;
+    using reference       = T&;
+    using const_reference = const T&;
+    template <class U> struct rebind { using other = allocator<U>; };
+
+    T* address(reference x) const noexcept;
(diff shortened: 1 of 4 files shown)


source/future.tex
```diff
@@ -1628,3 +1628,404 @@
...
+\pnum
+\returns
+The actual address of the object referenced by \tcode{x}, even in the presence of an
+overloaded operator\&.
+\end{itemdescr}
+
+\indexlibrary{\idxcode{allocate}!\idxcode{allocator}}
+\indexlibrary{\idxcode{allocator}!\idxcode{allocate}}
+\begin{itemdecl}
+T* allocate(size_type n, void* hint);
```
timsong-cpp: The synopsis above has `T* allocate(size_t, const void* hint);`.

source/future.tex
```diff
@@ -1628,3 +1628,404 @@
...
+   public:
+    using size_type       = size_t;
+    using difference_type = ptrdiff_t;
+    using pointer         = T*;
+    using const_pointer   = const T*;
+    using reference       = T&;
+    using const_reference = const T&;
+    template <class U> struct rebind { using other = allocator<U>; };
+
+    T* address(reference x) const noexcept;
```
timsong-cpp: Seems a little odd to stop using `pointer`/`const_pointer` but continue to use `reference`/`const_reference`. Is that intended?

tkoeppe:
@AlisdairM: Did you check with @zygoloid? Usually formal motions aren't applied as editorial pull requests, but rather, the editor has his own mechanism for creating and applying the changes directly. I'm sure the help is appreciated, but maybe double-check first, because until now motion applications have never been handled via pull requests.

AlisdairM:
I mentioned in Oulu I would put up pull requests for my 3 library motions, hopefully for the end of Monday.  No other scheme was suggested, although I am happy to take guidance.  Trying to simplify everyone's work at a busy time, not make more process, so just point me to whatever will work most easily.

source/future.tex
AlisdairM: Thanks - the synopsis is correct and I need to fix this.

source/future.tex
AlisdairM: That makes perfect sense and I wish I had noticed.  The wording approved by LWG has the same error, but do we agree this fix should be editorial?  We are talking about the nested 'reference' type of a specific allocator, not some generic customization point.

tkoeppe:
OK, that's fine, if you've mentioned this to Richard already and Dawn knows to not duplicate the work, then I'm sure the help is well appreciated. :-)

PR #772, opened 2016-06-27
[impldef] Fixed index entry collation to support keys containing code

tkoeppe:
OK, I reworked this. The new solution involves stripping out certain well-known commands from the index text and using the bare text for collation, while retaining the original text for the index entry. The stripped commands are `\tcode`, `\_`, `\mname` and `\grammarterm`.

For advanced entries that add further index structure, this duplication mechanism fails, so we provide a new macro `\impldefplain` that has the original functionality. It's only used for one set of entries (value of bit-fields).

Only two pages of diffs.


tkoeppe:
One consequence of the auto-generation of the collation order is that nested magic underscores (provided by the `underscore` package) don't work inside index entries. Magic underscores can always be replaced by explicit underscores (`\_`), though, which is what I've done here.

tkoeppe:
This change will impose the following permanent maintenance cost:

   * Anything inside an `\impldef` or `\impdefx` needs to explicitly escape underscores: `\_`
   * If text inside an `\impldef` or `\impdefx` uses macros (e.g. `\Cpp`), we either need to add a replacement for that macro to `macros.tex`, or we need to use the plain index form instead and manually provide the collation key.

I think that's manageable, but it's something that might catch you out as you add new content and it suddenly doesn't compile.

tkoeppe:
The two pages of diffs are now:

image

image

tkoeppe:
Superseded by PR #834.

PR #773, opened 2016-06-27
[utility,iterator] Apply p0174r2 deprecating vestigial library compon…

AlisdairM:
…ents

Moves text (with minor tweaks, per p0174r2) to Annex D
for deprecation, and updates the indices accordingly.

A drive-by fix in allocator so that operator==/!= in
the synopsis use the same parameter names in the
definition.

Fix use of 'reference' with 'T&' in std::allocator as an
editorial fix missed by the original paper.

### DIFF
--- a/source/future.tex
+++ b/source/future.tex
@@ -1628,3 +1628,404 @@ template <class Predicate>
 \begin{itemdescr}
 \pnum \returns \tcode{binary_negate<Predicate>(pred)}.
 \end{itemdescr}
+
+\rSec1[depr.default.allocator]{The default allocator}
+
+\pnum
+The following members and explicit class template specialization are defined in
+addition to those specified in \ref{default.allocator}:
+
+\indexlibrary{\idxcode{allocator}}%
+\begin{codeblock}
+namespace std {
+  // specialize for \tcode{void}:
+  template <> class allocator<void> {
+  public:
+    using value_type    = void;
+    using pointer       = void*;
+    using const_pointer = const void*;
+    // reference-to-\tcode{void} members are impossible.
+
+    template <class U> struct rebind { using other = allocator<U>; };
+  };
+
+  template <class T> class allocator {
+   public:
+    using size_type       = size_t;
+    using difference_type = ptrdiff_t;
+    using pointer         = T*;
+    using const_pointer   = const T*;
+    using reference       = T&;
+    using const_reference = const T&;
+    template <class U> struct rebind { using other = allocator<U>; };
+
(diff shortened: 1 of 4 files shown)


tkoeppe:
@zygoloid, @burblebee: Just a heads up that Alisdair has provided a PR to apply one of the motions.

source/future.tex
```diff
@@ -1628,3 +1628,404 @@
...
+    using pointer         = T*;
+    using const_pointer   = const T*;
+    using reference       = T&;
+    using const_reference = const T&;
+    template <class U> struct rebind { using other = allocator<U>; };
+
+    T* address(T& x) const noexcept;
+    const T* address(const T& x) const noexcept;
+
+    T* allocate(size_t, const void* hint);
```
timsong-cpp: Should the first parameter get a name here?

source/future.tex
```diff
@@ -1628,3 +1628,404 @@
...
+\pnum
+\returns
+The actual address of the object referenced by \tcode{x}, even in the presence of an
+overloaded operator\&.
+\end{itemdescr}
+
+\indexlibrary{\idxcode{allocator}!\idxcode{allocate}}
+\indexlibrary{\idxcode{allocate}!\idxcode{allocator}}
+\begin{itemdecl}
+T* allocate(size_type n, const void* hint);
```
timsong-cpp: Should this be `size_t` instead of `size_type`?

source/future.tex
AlisdairM: Good catch - this error was present in the original standard too.  FIXED

AlisdairM:
Force-pushed an amended branch to address Tim's comments, and apply with a clean history.
The missing parameter name was a problem in the original standard, and has been corrected in both utility and the deprecated annex.

tkoeppe:
@zygoloid, @burblebee, @jwakely: Until now, motion applications had commits titled `PnnnnRk Title of paper here`. Should we stick to that, or should the commit messages follow the pattern of editorial changes the way Alisdair has done here?

jwakely:
My preference would be without the [stable.name] for non-editorial changes. The point of putting that in the commit message is to know where changes happened, but with a paper number the paper provides that info. Not my call though.

tkoeppe:
I absorbed this into a branch on cplusplus/draft as commit b086b8a224ea1463aa60a9ae51e8db0e0989f924.

tkoeppe:
@burblebee Nice summary - that should go on the wiki!

PR #774, opened 2016-06-28
[utility,containers] Apply p0181r1, a default order

AlisdairM:
Added support to pmr container typedefs which were missed in the proposal
as it was based off an earlier draft, but were always inteded to be
included.  Could back that part out if we prefer to handle as a technical
issue against the CD.


tkoeppe:
I've cloned this branch into the main repository as commit 9044ac484ad0d8b425428c8746fc041e95af3625.

PR #776, opened 2016-06-28
P0305R1 Selection statements with initializer

tkoeppe:
Apply CWG Motion 18: P0305R1, "Selection statements with initializer".

Editorially moves the grammar production *condition* up to the top of Clause 6, along with the moved *init-statement*.

Do not merge until CWG Motion 12 (P0292R2, "if constexpr") has been applied, since the present application will need to be rebased first.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -772,7 +772,7 @@ is local to the
 \grammarterm{handler}.
 
 \pnum
-Names declared in the \grammarterm{for-init-statement}, the \grammarterm{for-range-declaration}, and in the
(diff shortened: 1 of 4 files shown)


zygoloid:
Can you move the changes to *condition* into a separate commit, please? I'd prefer to have separate commits for the change-as-moved and substantive editoral rework.

As part of moving *condition*, we should also move the text describing its semantics, which currently reside in [stmt.select], into [stmt.stmt].

tkoeppe:
@zygoloid: OK, I've done that, but now the description of *condition* makes a lot of references to things that haven't yet been described. Do you think that's OK?

PR #777, opened 2016-06-28
P0292R2 constexpr if: a slightly different syntax

tkoeppe:
This should not be merged before P0391R0 (motion 16).

### DIFF
--- a/source/statements.tex
+++ b/source/statements.tex
@@ -128,8 +128,8 @@ Selection statements choose one of several flows of control.
 %
 \begin{bnf}
 \nontermdef{selection-statement}\br
-    \terminal{if (} condition \terminal{)} statement\br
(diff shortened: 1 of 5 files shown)


source/statements.tex
```diff
@@ -245,6 +245,50 @@
...
 
+\pnum
+If the parenthesized \grammarterm{condition} is prefixed with
+\tcode{constexpr}, the value of the condition shall be a contextually
+converted constant expression of type bool (\ref{expr.const}); this
+form is called a \term{constexpr if} statement. If the value of the
+converted condition is \tcode{false}, the first substatement is a
+\term{discarded statement}, otherwise the second substatement, if
+present, is a discarded statement. During the instantation of an
+enclosing templated entity, if the condition is not value-dependent
```
tkoeppe: Do you think "templated entity" should come with a cross reference to "Clause 14"?

PR #779, opened 2016-06-30
[ratio.si] fix typo

cpplearner:


tkoeppe:
Thanks, will merge after the motions have been applied.

PR #780, opened 2016-06-30
Travis CI: Fail build if regenerated xref and grammar files don't match those in repository

godbyk:
This patch causes the Travis CI builds to fail if they detect that the newly generated grammar and xref files differ from those committed to the git repository.

Figure PDFs are currently ignored as regenerated PDFs will always differ from the committed PDFs due to changed metadata (build date/time, primarily). If a better solution for diffing the PDFs is found, we can add this test in the future.

This PR follows the discussion near the end of https://github.com/cplusplus/draft/pull/635.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,11 +38,11 @@ script:
   - popd
   # Check to see if generated files are out-dated
   - pushd source
-  - ../tools/makegram && git status --porcelain grammar.tex
-  - ../tools/makexref && git status --porcelain xref.tex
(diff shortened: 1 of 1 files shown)


godbyk:
One alternative that you may prefer would be to have these tests as a separate build type (their own row in the matrix). This would allow the other builds to complete and only fail if building `std.pdf` failed. The test build would fail only if the generated files differed from those in the repository. Which do you prefer?

godbyk:
An example of what having the tests as a separate matrix item looks like is at https://travis-ci.org/godbyk/draft/builds/141772422.

tkoeppe:
This is obsolete after PR #816.

PR #783, opened 2016-07-01
[19-26] Consistent use of _v

AlisdairM:
The library clauses have started to make inconsistent use of the _v
variable templates for traits, in preferene to directly querying
trait::value.  This change set applies that change consistently over
all existing wording, where there is a choice of which form might
be used.  It does not make any changes where the _v variable is not
aviable, or is being defined in terms of the corresponding ::value.

This may want a second pass by someone better able to pay attention
to desirable line-breaking.


tkoeppe:
Thanks. We'll look at that after the motions have been applied. Expect some rebasing :-)

AlisdairM:
rebased off current master (now that all Oulu motions have landed) and force-pushed

PR #789, opened 2016-07-03
[diff.basic] Fix format and inexact wording

FrankHB:
Though Annex C is informative, it's better to keep the terminology similar to normative text, which is easier to refer.

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -155,7 +155,7 @@ Because \tcode{const} objects may be used as values during translation in
 \Cpp, this feature urges programmers to provide an explicit initializer
 for each \tcode{const} object.
 This feature allows the user to put \tcode{const} objects in headers or source files that are included
-in more than one translation units.
+in more than one translation unit.
 \effect
 Change to semantics of well-defined feature.
 \difficulty
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -152,9 +152,9 @@ The latter is probably rare.
 declared \tcode{extern}, has internal linkage, while in C it would have external linkage
 \rationale
 Because \tcode{const} objects may be used as values during translation in
-\Cpp, this feature urges programmers to provide explicit initializer
+\Cpp, this feature urges programmers to provide an explicit initializer
 for each \tcode{const} object.
-This feature allows the user to put \tcode{const} objects in headers that are included
+This feature allows the user to put \tcode{const} objects in headers or source files that are included
 in more than one translation units.
 \effect
 Change to semantics of well-defined feature.
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -151,11 +151,11 @@ The latter is probably rare.
 \change A name of file scope that is explicitly declared \tcode{const}, and not explicitly
 declared \tcode{extern}, has internal linkage, while in C it would have external linkage
 \rationale
-Because \tcode{const} objects can be used as compile-time values in
+Because \tcode{const} objects may be used as values during translation in
 \Cpp, this feature urges programmers to provide explicit initializer
-values for each \tcode{const}.
-This feature allows the user to put \tcode{const}objects in header files that are included
-in many compilation units.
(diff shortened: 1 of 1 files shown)


tkoeppe:
@FrankHB: Can you please rebase (and sorry for the force-pushes on master)?

FrankHB:
Done.

source/compatibility.tex
```diff
@@ -151,11 +151,11 @@
 \change A name of file scope that is explicitly declared \tcode{const}, and not explicitly
 declared \tcode{extern}, has internal linkage, while in C it would have external linkage
 \rationale
-Because \tcode{const} objects can be used as compile-time values in
+Because \tcode{const} objects may be used as values during translation in
 \Cpp, this feature urges programmers to provide explicit initializer
-values for each \tcode{const}.
-This feature allows the user to put \tcode{const}objects in header files that are included
-in many compilation units.
+for each \tcode{const} object.
```
zygoloid: The grammar doesn't work here: "provide explicit initializer for each `const` object." Maybe "provide _an_ explicit initializer[...]"?

source/compatibility.tex
```diff
@@ -151,11 +151,11 @@
...
+This feature allows the user to put \tcode{const} objects in headers that are included
```
zygoloid: The change from "header files" to "headers" is an improvement, but is still wrong. "Headers" are the <blah> items provided by the C++ standard library. Non-C++ standard library headers, included by `#include "blah"` syntax, are (surprisingly) called "source files" within the standard.

tkoeppe: That made my day. *Of course* the standard would use "headers" and "source files" in a completely different way from the colloquial use :-)

source/compatibility.tex
FrankHB: Fixed.

source/compatibility.tex
```diff
@@ -151,11 +151,11 @@
...
-Because \tcode{const} objects can be used as compile-time values in
-\Cpp, this feature urges programmers to provide explicit initializer
-values for each \tcode{const}.
-This feature allows the user to put \tcode{const}objects in header files that are included
-in many compilation units.
+Because \tcode{const} objects may be used as values during translation in
+\Cpp, this feature urges programmers to provide an explicit initializer
+for each \tcode{const} object.
+This feature allows the user to put \tcode{const} objects in headers or source files that are included
+in more than one translation units.
```
tkoeppe: Shouldn't it say "more than one *unit*", singular?

source/compatibility.tex
FrankHB: Ah, my fault. I should not have missed "source file". I think it is not so much surprisingly (literally) because a "source file" is not a "header" (which is not guaranteed as a "file") or "object/binary/... file" (which is not file of "source" code). It is a common convention that some source files may be provided as the implementation of API exposed by some header files, but actually the latter are also in source code form (and probably in the same source tree), so they are also source files. Perhaps we'd better avoid such daily confusion :) I have already made the rules explicit about _non-header (source) files_ in projects I work on.

source/compatibility.tex
FrankHB: You are right.

source/compatibility.tex
```diff
@@ -151,11 +151,11 @@
...
 \rationale
-Because \tcode{const} objects can be used as compile-time values in
-\Cpp, this feature urges programmers to provide explicit initializer
-values for each \tcode{const}.
-This feature allows the user to put \tcode{const}objects in header files that are included
-in many compilation units.
+Because \tcode{const} objects may be used as values during translation in
+\Cpp, this feature urges programmers to provide an explicit initializer
+for each \tcode{const} object.
+This feature allows the user to put \tcode{const} objects in headers or source files that are included
```
tkoeppe: Say just "objects in source files"; @zygoloid says that "users have no control over headers". Then I'll merge this.

PR #791, opened 2016-07-04
[valarray.members] add valarray index entry for named member functions

AlisdairM:
The index currently lists the named member-functions of valarray as
if they were free functions - this adds the corresponding entry under
the valarray listing in the index, alongside the existing entries for
the many operator functions.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -7429,6 +7429,7 @@ void swap(valarray& v) noexcept;
 \end{itemdescr}
 
 \indexlibrary{\idxcode{length}!\idxcode{valarray}}%
+\indexlibrary{\idxcode{valarray}!\idxcode{length}}%
 \begin{itemdecl}
 size_t size() const;
 \end{itemdecl}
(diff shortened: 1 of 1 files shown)


AlisdairM:
sorry - looks like I have a pull request based on a reverted version of 'master'?

tkoeppe:
@AlisdairM master got force-pushed, so you too should fetch, rebase and force push.

AlisdairM:
rebased off the current master and force-pushed

tkoeppe:
(Don't tell anyone, but we had one or two force-pushes to master in the last week. Sorry about the inconvenience.)

AlisdairM:
Don't worry, I expected to rebase once everything landed anyway, as this touches so many places.  In fact, I was surprised the final rebase was as clean as it turned out to be :)

source/numerics.tex
```diff
@@ -7429,6 +7429,7 @@
 \end{itemdescr}
 
 \indexlibrary{\idxcode{length}!\idxcode{valarray}}%
+\indexlibrary{\idxcode{valarray}!\idxcode{length}}%
```
tkoeppe: Why does this say "length" when the function is called "size"?

tkoeppe:
I stole this PR; it is now #808.

PR #806, opened 2016-07-09
Improve cross references

tkoeppe:


README.rst
```diff
@@ -51,6 +51,7 @@ If you can't use latexmk or make for some reason, you can run LaTeX manually ins
 #. run ``makeindex libraryindex``
 #. run ``makeindex grammarindex``
 #. run ``makeindex impldefindex``
+#. run ``makeindex -s basic.gst -o xrefindex.gls xrefindex.glo``
```
godbyk: We should add this line to the `manual` build steps in `.travis.yml`, too.

tkoeppe: @godbyk: That's a good point. Can you maybe fold that into your pending change?

Issue #805, opened 2016-07-08
"// see below" vs @\seebelow@


tkoeppe:
In synopses, we often use `@\seebelow@` to create the code-italic magic token *see below*. But we also use this in comments to provide a reference.

I think when used as a comment, "see below" should just be normal, italic comment font. The code-italic style should only be used when it replaces an actual normative definition:

    using foo_t = @\seebelow@;  // see \ref{abc}
    using bar_t = @\seebelow@;  // see below

PR #809, opened 2016-07-09
[input.output,strings,utilities] Don't format informative 'see below' as code

tkoeppe:
Fixes Issue #805.


tkoeppe:
Diffs:

image


 
image

PR #811, opened 2016-07-09
[variant.visit] Add missing std:: qualification on uses of std::forward.

Eelis:


tkoeppe:
There's another one in Table 29.

Eelis:
That one is possibly intentional, because 17.6.3.5p2 just before it says: "Within Tables 28 and 29, the use of move and forward always refers to std::move and std::forward, respectively."

tkoeppe:
Oh, right, that's true.

tkoeppe:
Could you perhaps make a diffpdf screenshot?

Eelis:
Why?

tkoeppe:
Anything that's overly long and contains things like `\brk` concerns me, so I'd like to see what the result looks like.

tkoeppe:
While you're at it, I think the specification of `try_emplace`/`insert_or_assign` may also be using naked `forward`s in a few places. I'm sure the author would be grateful for the fixes :-)

tkoeppe:
@Eelis: ping?

Eelis:
Oh, yeah, I saw a bunch of other unqualified things as well, and then I got discouraged by the prospect of trying to figure out for each individual case whether the qualification is desired. :(  Feel free to close if a partial fix is not worthwhile.

tkoeppe:
Hm, generally, only `move` and `forward` are qualified, except when they appear in certain requirements tables which have a blanket rule saying that all unqualified uses are understood to be qualified. Otherwise, nothing is qualified except when it's syntactically required (e.g. `using reverse_iterator = std::reverse_iterator<T>;`, because it's the same name).

tkoeppe:
Best left to a systematic review.

PR #812, opened 2016-07-09
[/tools] Delete vestigial support files

tkoeppe:


tkoeppe:
@jwakely, @zygoloid: Does anyone still use any of these to-be-deleted files?

zygoloid:
As far as I'm aware, these aren't used for anything.

PR #816, opened 2016-07-10
[gram] Use 'extract' package for grammar summary

tkoeppe:


source/std.tex
```diff
@@ -49,6 +49,15 @@
            ]{hyperref}
 \usepackage{memhfixc}     % fix interactions between hyperref and memoir
 
+
+%% Grammar extraction.
+\IfFileExists{./std-gram.ext}{
+\usepackage[extract-env={bnftab,simplebnf,bnf,bnfkeywordtab}]{extract}
+\def\PrintGrammar{\input{std-gram.ext}}}{
+\usepackage[active,header=false,handles=false,copydocumentclass=false,generate=std-gram.ext,extract-env={bnftab,simplebnf,bnf,bnfkeywordtab}]{extract}
+\def\PrintGrammar{Run again to include grammar summary}
```
zygoloid: Do we really need this two-pass approach? Is there no way to flush the grammar at the end of the normal clauses so we can do this in one go? It seems easier to get this wrong than the current approach, at least in some ways. If we can't make this automatic, then at least we should use something like \ref{grammar.not.extracted.yet} here to trigger latexmk to do another iteration, and we should always regenerate the file even if we already have a copy.

tools/grambase.tex
```diff
@@ -1,66 +0,0 @@
-\infannex{gram}{Grammar summary}
```
zygoloid: Don't we still need this file?

source/std.tex
tkoeppe: There might be a way to close the current file by hacking into the guts of the extract package to make it close the output file and stop extracting. Then we'll get the desired output on every pass. I can try that if you'd entertain that route.

tools/grambase.tex
tkoeppe: No, why? It was only used as the starting point for `source/grammar.tex`, which now contains the contents from this file.

source/std.tex
tkoeppe: Changed, please take another look. I'm just calling into `extract` now to close the output file and reopen it under a new dummy name. The dummy file is discarded, but the point is that the main content is now always generated in the same run.

tools/grambase.tex
zygoloid: Oh, I see, you kept this in the formerly-generated grammar.tex file. Never mind, then.

source/basic.tex
```diff
@@ -1,8 +1,9 @@
 %!TEX root = std.tex
 \rSec0[basic]{Basic concepts}
 
-%gram: \rSec1[gram.basic]{Basic concepts}
-%gram:
+\begin{extract}
```
zygoloid: Can you create a wrapper environment for this with a more semantically-meaningful name?

tkoeppe: Do you mean something integrated like `\grammarheading{gram.basic}{Basic concepts}`?

tools/grama.sed
```diff
@@ -1,6 +0,0 @@
-# remove index "see" entries
-/\\index.*|see/d
```
zygoloid: Is this unnecessary now? (How? Do we simply not have any of these in extracted groups?)

zygoloid:
This looks great, thanks!

tools/grama.sed
tkoeppe: That's right, this construction doesn't exist at the moment. (You can see what's extracted in `std-gram.ext`; and there are no PDF diffs from this PR.) In the future, if such a situation comes up, we can perhaps work around with with `extract` and `extractskip` environments, or we can build another special-purpose command that changes meaning in the grammar.

PR #820, opened 2016-07-11
[clause 27] Index all member functions in clause 27

AlisdairM:
Some work remains to properly index all of the free (namespace scope) functions.

Will review other clauses later, especially those clauses with lots of additions for C++17.

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -936,6 +936,7 @@ Constructs an object of class
 \rSec4[ios::fmtflags]{Type \tcode{ios_base::fmtflags}}
 
 \indexlibrary{\idxcode{fmtflags}!\idxcode{ios_base}}%
+\indexlibrary{\idxcode{ios_base}!\idxcode{fmtflags}}%
 \begin{itemdecl}
 using fmtflags = @\textit{T1}@;
 \end{itemdecl}
@@ -998,8 +999,8 @@ also defines the constants indicated in Table~\ref{tab:iostreams.fmtflags.consta
 
 \rSec4[ios::iostate]{Type \tcode{ios_base::iostate}}
 
-\indexlibrary{\idxcode{iostate}!\idxcode{ios_base}}%
 \indexlibrary{\idxcode{ios_base}!\idxcode{iostate}}%
+\indexlibrary{\idxcode{iostate}!\idxcode{ios_base}}%
 \begin{itemdecl}
 using iostate = @\textit{T2}@;
 \end{itemdecl}
@@ -1036,6 +1037,7 @@ the value zero.
 \rSec4[ios::openmode]{Type \tcode{ios_base::openmode}}
 
 \indexlibrary{\idxcode{openmode}!\idxcode{ios_base}}%
+\indexlibrary{\idxcode{ios_base}!\idxcode{openmode}}%
 \begin{itemdecl}
 using openmode = @\textit{T3}@;
 \end{itemdecl}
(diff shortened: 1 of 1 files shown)


source/iostreams.tex
```diff
@@ -1173,8 +1177,8 @@
 The format control information for both input and output.
 \end{itemdescr}
 
-\indexlibrary{\idxcode{fmtflags}!\idxcode{ios_base}}%
 \indexlibrary{\idxcode{ios_base}!\idxcode{fmtflags}}%
+\indexlibrary{\idxcode{fmtflags}!\idxcode{ios_base}}%
```
tkoeppe: Is this entry correct? It looks like the function is called `flags`, no` fmtflags`?!

source/iostreams.tex
```diff
@@ -4335,7 +4405,8 @@
 explicit basic_istream(basic_streambuf<charT, traits>* sb);
 \end{itemdecl}
 
-\indexlibrary{\idxcode{init}!\idxcode{basic_ios}}%
+\indexlibrary{\idxcode{init}!\idxcode{basic_istream}}%
+\indexlibrary{\idxcode{basic_istream}!\idxcode{init}}%
```
tkoeppe: What is this entry indexing? I don't quite see where this belongs.

AlisdairM: Reverted to the original, it is indexing a reference to base class initialization in the following text.

source/iostreams.tex
AlisdairM: Fixed - thanks, thought I had caught all of these (there were not many).

source/iostreams.tex
```diff
@@ -13249,6 +13479,7 @@
 These functions enable use of \tcode{recursive_directory_iterator}
 with range-based for statements.
 
+\indexlibrary{\idxcode{begin}!\idxcode{recursive_directory_iterator}}%
```
tkoeppe: Why is there only one version of this index entry? Is this not a member function?

source/iostreams.tex
```diff
@@ -13258,6 +13489,7 @@
 \returns \tcode{iter}.
 \end{itemdescr}
 
+\indexlibrary{\idxcode{begin}!\idxcode{recursive_directory_iterator}}%
```
tkoeppe: Double-check the name?

source/iostreams.tex
AlisdairM: I believe this is for the free-function overloads.

source/iostreams.tex
AlisdairM: Oops - good catch

source/iostreams.tex
```diff
@@ -6029,7 +6114,7 @@
 defines a class that is responsible for doing exception safe prefix and suffix
 operations.
 
-\indexlibrary{\idxcode{sentry}!constructor}%
```
tkoeppe: Why are you deleting this here, but you left both "sentry" and "basic_ostream::sentry" for the destructor?

AlisdairM: As far as I can tell, I am consistent for constructor and destructor - the difference is between basic_istream::sentry and basic_ostream::sentry.  In one case, I retain the old index entry and add the new one, in the other, I just rename the old to the new.

I agree they should be consistent, but not sure which way we think is the best to resolve.  I mildly lean towards retaining both 'sentry!constructor' and 'basic_Xstream::sentry!constructor' (and similarly for destructors), but I lean the other way for 'operator bool' - with is terribly inconsistent of me again.

Preferences?

tkoeppe: None at all. We can leave it as is :-)

PR #823, opened 2016-07-12
[support] clean up index for clause 18

AlisdairM:
Fix a couple of mis-indexed operators, and add index entries
for a few missing functions, most notably the new launder
function.

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -2104,7 +2104,7 @@ void* operator new[](std::size_t size, void* ptr) noexcept;
 Intentionally performs no other action.
 \end{itemdescr}
 
-\indexlibrary{\idxcode{new}!\idxcode{operator}}%
+\indexlibrary{\idxcode{delete}!\idxcode{operator}}%
 \begin{itemdecl}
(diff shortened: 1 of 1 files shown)


source/support.tex
```diff
@@ -2127,7 +2127,7 @@
 terminates by throwing an exception~(\ref{expr.new}).
 \end{itemdescr}
 
-\indexlibrary{\idxcode{new}!\idxcode{operator}}%
+\indexlibrary{\idxcode{delete}!\idxcode{operator}}%
```
tkoeppe: Why not array-delete?

AlisdairM: I believe this is consistent with the other index entries for array-delete, and did not want to be too creative at this point.  I can consistently index 'delete[]' if preferred?

tkoeppe: @zygoloid: Any preference how we index array-delete?

AlisdairM: and for consistency, array-new as well.

tkoeppe: Actually, don't worry. This is definitely an improvement, and it's consistent. Thanks.

PR #829, opened 2016-07-15
[diagnostics] audit index entries for clause 19

AlisdairM:
The only issues that turned up are missing '%' continuations
at the end of around half of the index entry lines.  This
patch consistently cleans that up, so that every index entry
terminates in a '%', but it is not clear that this has any
effect on the final rendering, and the patch could safely be
rejected.  The primary benefit is consistency, should we ever
want to write some follow-up tooling to analyze index entries.

Audit confirmed that every \begin{itemdecl} is indexed, and the
index entry matches the declared name(s).

### DIFF
--- a/source/diagnostics.tex
+++ b/source/diagnostics.tex
@@ -971,8 +971,8 @@ virtual ~error_category();
 \effects Destroys an object of class \tcode{error_category}.
 \end{itemdescr}
 
-\indexlibrary{\idxcode{name}!\idxcode{error_category}}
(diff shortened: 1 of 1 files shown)


tkoeppe:
You could run `diffpdf` to see if there's any effect.

tkoeppe:
OK, no diffs. I have no particular opinion on this change, but I suppose it can't hurt, and it's "leading by example" regarding consistency.

PR #830, opened 2016-07-15
Update the reference to the C standard in intro.scope

bourguet:
intro.scope references the C99 edition of the C standard while intro.refs
references the C11 edition, I assume it was not updated when the change was
made.


tkoeppe:
This is not editorial, and this is a known issue.

PR #832, opened 2016-07-18
Add back function lvalues to [dcl.init.ref]/5.2.1.2

bogiord:
I suppose this was omitted when applying [P0135R1](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0135r1.html), as it's present in the paper.


zygoloid:
Thank you!

PR #834, opened 2016-07-19
[impldef] Collate index correctly

tkoeppe:
This involves modifying the `\indeximpldef` macro to strip out the bare arguments from a selection of commands that we use inside index entries, and using the bare text to create the collation key. Complex index commands need to use the new `\impldefplain` commands.


tkoeppe:
Two pages of diffs:

image


image

PR #835, opened 2016-07-20
[18,19] Add and use new indexlibrarymember macro

AlisdairM:
Add a new macro, indexlibrarymember, that takes two arguments,
a class-name and a member name, which creates two index entries
for:
   class-name, member
and
   member, class-name

This should simplify many such double-declarations in the
library clauses, and reduce errors of inconsistency.  It
will also ease fixing up existing index entries that were
(accidentally) not doubled up.

Ths new macro is applied consistently throughout clauses
18 and 19 as a proof-of-concept, and follow-up patches
will deal with successive library clauses.  This is
deliberately broken up to ease the review burden.

Finally, some minor consistency issues regarding use of
% to ensure index macros are not consuming unexpected
whitespace.

### DIFF
--- a/source/diagnostics.tex
+++ b/source/diagnostics.tex
@@ -971,8 +971,7 @@ virtual ~error_category();
 \effects Destroys an object of class \tcode{error_category}.
 \end{itemdescr}
 
-\indexlibrary{\idxcode{name}!\idxcode{error_category}}%
-\indexlibrary{\idxcode{error_category}!\idxcode{name}}%
+\indexlibrarymember{error_category}{name}%
 \begin{itemdecl}
 virtual const char* name() const noexcept = 0;
 \end{itemdecl}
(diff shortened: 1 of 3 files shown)


tkoeppe:
How about you put all the changes into this pull request, but as a sequence of successive commits: one for macros.tex, and then one for each clause. Then we can merge the entire PR, but still review it commit by commit.

AlisdairM:
My main concern is that there will be a lot of moving parts, and the risk of conflicts making the whole sequence unlandable seems very high.  Unlike the last time I hit every clause (typedef -> using) I am not making a clear user-visible change to the document, so I think that incremental upgrades to the index are easier to land - although it is important to land this one first before I can base later clause updates on the right commit.

tkoeppe:
Well, there's nobody merging anything at the moment apart from me, so I don't think there'll be a big risk. And you are still incremental, just on the same branch.

AlisdairM:
My concern is that the branch is likely to be long-lived, as I don't commit to reviewing/resolving all clauses in a given time-frame, while the CD review is also ongoing.

More to the point - I would rather do a clause-by-clause index-review where this new macro is one tool available to simplify the job, rather than apply this as a separate pass and then review each clause again, or muddle the simple introduction of this macro, with a much more thorough index review.

tkoeppe:
Leave this change with me. I can do it in a few minutes in multiple commits.

PR #839, opened 2016-07-20
Add constexpr to optional::value_or(U&&) &&

CaseyCarter:
This constexpr specifier is present in the `<optional>` synopsis, but omitted from the detailed specification in [optional.object.observe]. (I understand I'm stretching the definition of "editorial" here. Please let me know if the editors feel this change is substantive, and I'll file a defect report with LWG.)

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -2908,7 +2908,7 @@ the program is ill-formed.
 \end{itemdescr}
 
 \begin{itemdecl}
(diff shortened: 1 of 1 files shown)


zygoloid:
This is OK to fix editorially; the detailed specification is only supposed to specify which function from the synopsis we're referring to, and it's not uncommon for elements that don't affect the signature to be omitted (exception specifiers, `constexpr`, `override`, ...).

PR #840, opened 2016-07-21
[language.support] review library indexing of clause 18

AlisdairM:
This change consistently applies to new indexlibrarymember
macro in places missed by the first pass - mostly for cases
where either the class or member-name has an underscore.

It consistently uses the notation of {class}{member} when
using the new macro.  This is purely a consistency concern,
as the macro will expand to the same results either way.

It consistently uses a % to eliminate potential whitespace
being introduced by an index entry macro.

It removes a duplicate entry for 'numeric_limits' causes
by the use of the \_ command.  This appears to be the
correct fix, on inspection of the index.  The alternative
would have been consistent use of \_ on every numeric_limits
entry, so trying the simpler fix first.

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -230,7 +230,7 @@ namespace std {
 
 \rSec3[numeric.limits]{Class template \tcode{numeric_limits}}
 
-\indexlibrary{\idxcode{numeric\_limits}}%
+\indexlibrary{\idxcode{numeric_limits}}%
 \begin{codeblock}
 namespace std {
   template<class T> class numeric_limits {
@@ -295,8 +295,7 @@ the specialization on the unqualified type \tcode{T}.
 
 \rSec3[numeric.limits.members]{\tcode{numeric_limits} members}
 
-\indexlibrary{\idxcode{min}!\idxcode{numeric_limits}}
-\indexlibrary{\idxcode{numeric_limits}!\idxcode{min}}
+\indexlibrarymember{numeric_limits}{min}%
 \begin{itemdecl}
 static constexpr T min() noexcept;
 \end{itemdecl}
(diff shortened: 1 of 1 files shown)


tkoeppe:
Hang on, there are more of these issues. The problem was that I only searched for things ending in a `%`, so there's more coming. I'll fix it and let you know so you can rebase.

AlisdairM:
OK.

Just so you know what I am looking at: I plan to perform a clause-by-clause review, as time permits, checking every itemdecl has an index entry, and all index entries have a trailing %.  For good measure, I will promote existing members to indexlibrarymember if they are not already (i.e., were not duplicated to begin with).

If I find more entries introduced by codeblock rather than itemdecl, I will fix those too - but I don't have an easy way to audit for those - that is more spot it by luck as I go.

tkoeppe:
Rebase please! :-)

AlisdairM:
OK - it looks like the overwhelming majority of uses of indexlibrarymember are:
   \indexlibrarymember{member}{class}
rather than
   \indexlibrarymember{class}{member}

While the latter is my preference, I will apply the former consistently (distinctly fewer changes), and make that part of my clause-by-clause audit.

tkoeppe:
It doesn't matter at all; either version works. I just went with the search-and-replace that was easy (picking the first line). If you're enjoying this sort of no-op work, feel free to switch the arguments around :-)

AlisdairM:
I would rather have a consistency (it aids later tooling - if any) and want to minimize churn on make-work tasks - some folks may be sensitive to the doc history,  It seems putting a class second (unlike when handling constructor/destructor) is the default, based on the pre-existing order when you ran your regex.

tkoeppe:
Well, I think the only thing that's consistent so far is that nothing is consistent :-) The previous commit, which used the same regex, had it the other way round. I really don't care whether a) we do anything at all or b) which way round you make it; do whatever you think is useful.

tkoeppe:
Are you reusing this PR?

tkoeppe:
Please reopen when it's ready.

Issue #566, opened 2015-11-20
Synopses should have numbered headers


tkoeppe:
Following [PR 560](https://github.com/cplusplus/draft/pull/560), @zygoloid noticed that there are still several synopses that don't have numbered headers (e.g. strings, algorithms). Uses of `\synopsis` should be replaced with:

```
\rSec2[foo.syn]{Header \tcode{<foo>} synopsis}
```

See also #542.

tkoeppe:
Question: Should all synopses appear at the top-most level, as a sibling to "General", as we currently do with the containers library? Or is it appropriate to keep them within their respective subsections?

For example, consider [diagnostics]. We could either list the synopses of `<stdexcept>` and `<system_error>` at the top, right after "General", as the new 19.2, or we could keep them in their current place as 19.2.1 and 19.5.1.

I'm mildly in favour of the latter approach, since we'd otherwise be left with text that wouldn't have a section any more (such as the text after the `<system_error>` synopsis at the moment).

tkoeppe:
@tkoeppe: Ping. What should we do with all the remaining `\synopsis`s?

PR #843, opened 2016-07-21
[cstdint, cinttypes.syn] Reorganize <cstdint> presentation

tkoeppe:
- Demote heading "18.4.1 Header `<cstdint>` synopsis" to `\synopsis`.
- Show MIN/MAX macros in synopsis.
- Remove stray namespace comment.
- Add See Also for the relevant C clause.
- Add cross reference from `<cinttypes>` synopsis.

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -1872,42 +1872,41 @@ namespace std {
   using uintptr_t      = @\textit{unsigned integer type}@; // optional
 }
 
-#define INT@\placeholdernc{N}@_MIN         @\seebelow@
-#define INT@\placeholdernc{N}@_MAX         @\seebelow@
-#define UINT@\placeholdernc{N}@_MAX        @\seebelow@
+#define INT[8, 16, 32, 64]_MIN @\seebelow@         // optional
+#define INT[8, 16, 32, 64]_MAX @\seebelow@         // optional
+#define UINT[8, 16, 32, 64]_MAX @\seebelow@        // optional
 
-#define INT_FAST@\placeholdernc{N}@_MIN    @\seebelow@
-#define INT_FAST@\placeholdernc{N}@_MAX    @\seebelow@
-#define UINT_FAST@\placeholdernc{N}@_MAX   @\seebelow@
+#define INT_LEAST[8, 16, 32, 64]_MIN @\seebelow@
+#define INT_LEAST[8, 16, 32, 64]_MAX @\seebelow@
+#define UINT_LEAST[8, 16, 32, 64]_MAX @\seebelow@
 
-#define INT_LEAST@\placeholdernc{N}@_MIN   @\seebelow@
-#define INT_LEAST@\placeholdernc{N}@_MAX   @\seebelow@
-#define UINT_LEAST@\placeholdernc{N}@_MAX  @\seebelow@
+#define INT_FAST[8, 16, 32, 64]_MIN @\seebelow@
+#define INT_FAST[8, 16, 32, 64]_MAX @\seebelow@
+#define UINT_FAST[8, 16, 32, 64]_MAX @\seebelow@
 
-#define INTMAX_MIN       @\seebelow@
-#define INTMAX_MAX       @\seebelow@
-#define UINTMAX_MAX      @\seebelow@
+#define INTPTR_MIN @\seebelow@                     // optional
+#define INTPTR_MAX @\seebelow@                     // optional
+#define UINTPTR_MAX @\seebelow@                    // optional
 
-#define INTPTR_MIN       @\seebelow@              // optional
-#define INTPTR_MAX       @\seebelow@              // optional
-#define UINTPTR_MAX      @\seebelow@              // optional
+#define INTMAX_MIN @\seebelow@
+#define INTMAX_MAX @\seebelow@
+#define UINTMAX_MAX @\seebelow@
 
-#define PTRDIFF_MIN      @\seebelow@
-#define PTRDIFF_MAX      @\seebelow@
-#define SIZE_MAX         @\seebelow@
+#define PTRDIFF_MIN @\seebelow@
+#define PTRDIFF_MAX @\seebelow@
+#define SIG_ATOMIC_MIN @\seebelow@
+#define SIG_ATOMIC_MAX @\seebelow@
+#define SIZE_MAX @\seebelow@
 
-#define SIG_ATOMIC_MIN   @\seebelow@
-#define SIG_ATOMIC_MAX   @\seebelow@
+#define WCHAR_MIN @\seebelow@
+#define WCHAR_MAX @\seebelow@
+#define WINT_MIN @\seebelow@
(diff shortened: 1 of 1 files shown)


source/support.tex
```diff
@@ -1211,31 +1211,52 @@
...
+
+#define INT@\placeholdernc{N}@_C(value) @\seebelow@
+#define UINT@\placeholdernc{N}@_C(value) @\seebelow@
+
+#define INTMAX_C(value) @\seebelow@
+#define UINTMAX_C(value) @\seebelow@
 \end{codeblock}
 
 \pnum
 The header defines all types and macros the same as
```
zygoloid: I think this has changed the meaning, but I'm not sure whether the old or new wording matches the intent. We used to definitively say that the C++ header definitely defines these macros for N in 8, 16, 32, 64, and in no other cases. C says something quite different -- those macros are defined if there happen to be types of those widths, and other macros can also be defined if there are types of other widths.

tkoeppe: Does C++ not actually allow other values of *N* at all?

tkoeppe: @zygoloid: updated version pushed that lists the four sizes that we support explicitly.

tkoeppe: @zygoloid: Ping

source/support.tex
```diff
@@ -1210,32 +1210,53 @@
...
   using uintmax_t      = @\textit{unsigned integer type}@;
-  using uintptr_t      = @\textit{unsigned integer type}@; // optional
-} // namespace std
-\end{codeblock}
+  using uintptr_t      = @\textit{unsigned integer type}@;  // optional
+}
 
-\pnum
-The header also defines numerous macros of the form:
+#define INT[8, 16, 32, 64]_MIN @\seebelow@         // optional
```
zygoloid: These did not appear to be optional before.

I do not think we can fix this editorially; we need to ask LWG what they meant.

tkoeppe: These macros are supposed to expand to expressions of their corresponding type. Since the corresponding type is optional, surely the macro must also be optional?

But indeed the LEAST/FAST macros should absolutely not be optional, that's a mistake that I will fix at once.

tkoeppe: Fixed.

tkoeppe: I will open an LWG issue to have the question settled of whether the macros are optional.

tkoeppe:
Pending resolution of [LWG 2764](http://wg21.link/lwg2764).

source/iostreams.tex
```diff
@@ -14984,7 +14984,7 @@
 \indexlibrary{\idxcode{SCNuPTR}}%
 \indexlibrary{\idxcode{SCNxPTR}}%
 \begin{codeblock}
-#include <cstdint>
+#include <cstdint> // see \ref{cstdint}
```
jensmaurer: Is that comment aligned to one of your preferred columns?

source/support.tex
```diff
@@ -1306,7 +1306,7 @@
 
 \rSec1[cstdint]{Integer types}
 
-\rSec2[cstdint.syn]{Header \tcode{<cstdint>} synopsis}
+\synopsis{Header \tcode{<cstdint>} synopsis}
 \indextext{\idxhdr{cstdint}}%
```
jensmaurer: \synopsis is gone; see #566 and #1231.

source/iostreams.tex
tkoeppe: Let me check. Preferred columns are 20, 32, 48, 24.

source/support.tex
tkoeppe: OK, I'll rebase this.

source/support.tex
```diff
@@ -1374,32 +1374,53 @@
...
+#define UINTMAX_C(value) @\seebelow@
 \end{codeblock}
 
 \pnum
 The header defines all types and macros the same as
 the C standard library header \tcode{<stdint.h>}.
 
+\xref ISO C~7.20.
+
 \rSec1[support.start.term]{Start and termination}
```
jensmaurer: This seems a good addition that can go in ahead of time.

tkoeppe: Done in edd2c5fb29b2feb6a6acc23f843d351a8f06b2ec.

source/iostreams.tex
tkoeppe: Fixed.

source/support.tex
```diff
@@ -1346,10 +1346,10 @@
...
 \begin{codeblock}
 namespace std {
-  using int8_t         = @\textit{signed integer type}@;  // optional
-  using int16_t        = @\textit{signed integer type}@;  // optional
-  using int32_t        = @\textit{signed integer type}@;  // optional
-  using int64_t        = @\textit{signed integer type}@;  // optional
+  using int8_t         = @\textit{signed integer type}@;    // optional
+  using int16_t        = @\textit{signed integer type}@;    // optional
+  using int32_t        = @\textit{signed integer type}@;    // optional
+  using int64_t        = @\textit{signed integer type}@;    // optional
```
jensmaurer: Could you please put in the whitespace fixes right away? This makes the "hard" parts easier to spot.

source/support.tex
```diff
@@ -1380,26 +1380,45 @@
...
-  [U]INT{MAX PTR}_MAX
-  {PTRDIFF SIG_ATOMIC WCHAR WINT}{_MAX _MIN}
-  SIZE_MAX
-\end{codeblock}
+#define INT_LEAST[8, 16, 32, 64]_MIN @\seebelow@
+#define INT_LEAST[8, 16, 32, 64]_MAX @\seebelow@
+#define UINT_LEAST[8, 16, 32, 64]_MAX @\seebelow@
 
-plus function macros of the form:
+#define INT_FAST[8, 16, 32, 64]_MIN @\seebelow@
```
cubbimew: This seems worse: the never-defined pattern-matching language of "INT_[FAST LEAST]{8 16 32 64}_MIN" is both changed and used in what would otherwise be a preprocessor directive (rather than prose).

source/support.tex
tkoeppe: Sure, will do.

tkoeppe:
I think this is pending some LWG issue ([LWG 2820](http://wg21.link/lwg2820)). I'll check with @mclow where we are on this.

source/support.tex
tkoeppe: Reverting those changes for now, since they don't take italic correction into account, either. Once we have direction, I'll separate cosmetic and substantive change.

jensmaurer:
@tkoeppe, ping. It seems LWG 2820 is ready for a -lib vote and will make some of the changes in this pull request moot. I'd suggest you ask @mclow.

tkoeppe:
Any day now, seriously.

tkoeppe:
@mclow, @jensmaurer: This has been resolved by LWG2820 via 2020-11 Motion LWG-3. Thank you all for your continued support!

PR #850, opened 2016-07-22
[container.node] Move sub-clause to after [sequence.reqmts]

### LINKED ISSUE #842: 23.1.1 Node handles [container.node] in the wrong place?
It seems wrong to me that [container.node] is the very first thing in clause 23. It comes before the general container requirements, even.

I would suggest that they get moved to a less prominent location.
How about between 23.2.3 and 23.2.4?

jwakely:
Resolves #842

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -31,254 +31,6 @@ Table~\ref{tab:containers.lib.summary}.
                              &                                  & \tcode{<stack>}         \\ \rowsep
 \end{libsumtab}
 
-\rSec2[container.node]{Node handles}
-
-\rSec3[container.node.overview]{\tcode{node_handle} overview}
-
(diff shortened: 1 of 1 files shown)


tkoeppe:
That seems like a good place for it. Do you need anyone's final approval for this? @zygoloid, any objections to this move?

zygoloid:
If we ever want node handles for `list` and `forward_list`, we'll probably want it one subclause earlier, but this is fine for now.

jwakely:
We can already splice with lists, and there are none of the problems with keys being const when using a list, so there's little motivation to extract a node in order to muck about with it. If you really want to, you can extract a node into a separate list (i.e. `std::list<T>::node_type` is `std::list<T>`)

PR #851, opened 2016-07-22
[atomics.order] Remove redundant typedef-name for memory_order

jwakely:


source/atomics.tex
```diff
@@ -202,10 +202,10 @@
 
 \begin{codeblock}
 namespace std {
-  typedef enum memory_order {
+  enum memory_order {
     memory_order_relaxed, memory_order_consume, memory_order_acquire, 
     memory_order_release, memory_order_acq_rel, memory_order_seq_cst
-  } memory_order;
```
tkoeppe: What's this C code doing in our C++? Kill it!

cpplearner:
The definition of atomic_flag is presented [in similar way](https://github.com/cplusplus/draft/blob/a8a89e1a8d69fcd2219a3d54e523afb4d860bd2e/source/atomics.tex#L1225).

Presumably that typedef-name should also be removed?

PR #853, opened 2016-07-23
[intro.progress] Add missing impldef index entry for concurrent forward progress guarantees provided by threads of execution.

Eelis:
Just an index entry, no visual change in the section itself.


Eelis:
Made obsolete by #899

PR #854, opened 2016-07-23
[basic.life] Fix modal verb

FrankHB:
Change "must" to "shall" in normative text.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3485,7 +3485,7 @@ implicitly---upon exit from the block for an object with
 automatic storage duration, upon exit from the thread for an object with
 thread storage duration, or upon exit from the program for an object
 with static storage duration.}
-the program must ensure that an object of the original type occupies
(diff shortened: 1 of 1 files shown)


tkoeppe:
@zygoloid: PTAL

tkoeppe:
I spoke with zygoloid, and the conclusion is two-fold: 1) is that we should rephrase this paragraph to say "If X, the behaviour is undefined", but 2) is that we should clarify the wording to make it clear that the following is not allowed:

    struct A {
      const int n;
      ~A() { printf("%d", n");
    };

    void f() {
      A a = {4};
      a.~A();
      new (&a) A{5};
    }

It should do so by being clear that `a.~A()` refers to the "original pointer" (and not a laundered version thereof). With that in place, the paragraph in question might become redundant.

We should perhaps file a CWG issue for this.

zygoloid:
This has been converted to a core issue.

PR #855, opened 2016-07-25
[utilities] review of library index

AlisdairM:
This review handles several topic related to the index
of library names:
  fix errors of omission: every itemdecl is now indexed
  all headers have an idxhdr entry


AlisdairM:
Amended commit with the last few \indexlibrarymember uses that were {class}{identifier} rather than the other way around.

tkoeppe:

PR #856, opened 2016-07-26
[language.support] review of library index

AlisdairM:

    apply indeglibrarymember for all member functions other than constructors/destructors


tkoeppe:

PR #857, opened 2016-07-26
[diagnostics] review of library index

AlisdairM:

      ensure every &lt;header> has an idxdhr entry in the library index
      properly index has functor specialization for type
      consistently apply \indexlibrarymember for operator!= overloads
      (drive by fix for 'override' on 'what' function)


tkoeppe:

PR #858, opened 2016-07-26
[strings] review of library index

AlisdairM:

      indexed the user-defined literal operators
      index the hash functor specializations by type
      consistent apply \indexlibrarymember for all member functions, including previous omissions


tkoeppe:

PR #859, opened 2016-07-26
[iterators] review of library index

AlisdairM:

    indexed all of the container free-function overloads
    consistent apply indexlibrarymember{identifier}{class-name} for all member functions
    clear out a little redundancy and unnecessary whitespace


tkoeppe:

PR #860, opened 2016-07-26
[numerics] review of library index

AlisdairM:

    consistently indexlibrarymember{identifier}{class-name} for all members and operators
    consistent sub-indexing of special member functions for complex and valarray
    index the user-defined literal operators
    ensure every class definition is introduced by an \indexlibrary{\idxcode{class-name}}%


tkoeppe:
Thank you! I'm collecting all your work into https://github.com/cplusplus/draft/pull/861. Are there any more changes coming up?

PR #866, opened 2016-07-28
[re.syn] Synchronize regex_constants type declarations with those in [re.const].

Eelis:


source/regex.tex
```diff
@@ -264,7 +264,9 @@
 
   // \ref{re.const}, regex constants:
   namespace regex_constants {
-    enum error_type;
+    using syntax_option_type = @\textit{T1}@;
```
tkoeppe: Don't we have something like `\placeholder` for this?

source/regex.tex
```diff
@@ -264,7 +264,9 @@
...
+    using match_flag_type = @\textit{T2}@;
+    using error_type = @\textit{T3}@;
   } // namespace regex_constants
```
tkoeppe: Please remove the namespace comment. We don't really do this in the Standard, but some people have smuggled this in.

Eelis:
Re-pushed with suggested changes

PR #867, opened 2016-07-28
[re.traits] Remove excessive newlines from codeblocks.

Eelis:
These two newlines are visible in the pdf as awkward whitespace.

I also tried a larger patch which removed trailing newlines and spliced adjacent codeblocks as well, but that seemed to have no visible effect.


jwakely:
It looks like there's another one on line 1229 that doesn't need to be there

Eelis:
Ok, added that one as well

jwakely:
thanks

PR #868, opened 2016-07-28
[cpp.predefined] Make description for __cplusplus non-redundant and consistent with rest.

Eelis:
On the point of redundancy: please correct me if I'm wrong, but my understanding is that /all/ the requirements in the C++ standard only apply to implementations compiling C++ translation units, in which case spelling it out in this instance seems unnecessary. :)


tkoeppe:
@zygoloid: PTAL

PR #870, opened 2016-07-30
[dcl.attr.deprecated, depr.ostream.members] Fix trailing whitespace in \tcode.

Eelis:
(Of the two changes, only the change in [depr.ostream.members] has a visible effect.)

### DIFF
--- a/source/future.tex
+++ b/source/future.tex
@@ -976,7 +976,7 @@ strstreambuf* rdbuf() const;
 \begin{itemdescr}
 \pnum
 \returns
-\tcode{(strstreambuf*)\&sb }.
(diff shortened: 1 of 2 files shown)


source/future.tex
```diff
@@ -976,7 +976,7 @@
 \begin{itemdescr}
 \pnum
 \returns
-\tcode{(strstreambuf*)\&sb }.
+\tcode{(strstreambuf*)\&sb}.
```
tkoeppe: Eew, shouldn't this be a `static_cast`?

jwakely: It's Annex D, so meh.

PR #871, opened 2016-07-30
[depr.strstream.dest] Move rdbuf() to [depr.strstream.oper].

Eelis:
diff

### DIFF
--- a/source/future.tex
+++ b/source/future.tex
@@ -1123,6 +1123,8 @@ Destroys an object of class
 \tcode{strstream}.
 \end{itemdescr}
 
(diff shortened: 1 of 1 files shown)


tkoeppe:
Thanks. In the future, can you please either say which side of the screenshot is before and which is after, or otherwise make this visible in the screenshot, or otherwise put the "before" on the left-hand side? That removes a few extra moments of surprise.

Eelis:
You find "left old, right new" surprising? It is also what Github uses for its side by side diffs. Do Github's side-by-side diffs surprise you? :)

tkoeppe:
@Eelis: No, sorry, I was confused, the screenshot was the right way round already. Never mind!

PR #872, opened 2016-08-01
Replace 0 with nullptr or 'null pointer value'.

Eelis:
I manually went over all occurrences of 0 in the document, so modulo human error this should be all of them.

[visual diff](http://eelis.net/nullptrdiffs.pdf)

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -2520,7 +2520,7 @@ characters of null-terminated multibyte strings (\ntmbs
 s)~(\ref{multibyte.strings}) and \tcode{argv[0]} shall be the pointer to
 the initial character of a \ntmbs that represents the name used to
 invoke the program or \tcode{""}. The value of \tcode{argc} shall be
-non-negative. The value of \tcode{argv[argc]} shall be 0. \begin{note} It
+non-negative. The value of \tcode{argv[argc]} shall be \tcode{nullptr}. \begin{note} It
 is recommended that any further (optional) parameters be added after
 \tcode{argv}. \end{note}
 
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -2520,7 +2520,7 @@ characters of null-terminated multibyte strings (\ntmbs
 s)~(\ref{multibyte.strings}) and \tcode{argv[0]} shall be the pointer to
 the initial character of a \ntmbs that represents the name used to
 invoke the program or \tcode{""}. The value of \tcode{argc} shall be
-non-negative. The value of \tcode{argv[argc]} shall be \tcode{nullptr}. \begin{note} It
+non-negative. The value of \tcode{argv[argc]} shall be the null pointer value. \begin{note} It
 is recommended that any further (optional) parameters be added after
 \tcode{argv}. \end{note}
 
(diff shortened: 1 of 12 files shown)


source/basic.tex
```diff
@@ -2520,7 +2520,7 @@
 s)~(\ref{multibyte.strings}) and \tcode{argv[0]} shall be the pointer to
 the initial character of a \ntmbs that represents the name used to
 invoke the program or \tcode{""}. The value of \tcode{argc} shall be
-non-negative. The value of \tcode{argv[argc]} shall be 0. \begin{note} It
+non-negative. The value of \tcode{argv[argc]} shall be \tcode{nullptr}. \begin{note} It
 is recommended that any further (optional) parameters be added after
```
tkoeppe: Hm, well, `argv[argc]` is of type `char*`, not `std::nullptr_t`, so I'm not sure this is accurate. It wasn't accurate in the first place, but perhaps we should fix that instead. ("... shall be a null"?)

source/expressions.tex
```diff
@@ -2412,7 +2412,7 @@
 structure of the underlying machine.
 \end{note} A value of type \tcode{std::nullptr_t} can be converted to an integral
 type; the conversion has the same meaning and validity as a conversion of
-\tcode{(void*)0} to the integral type. \begin{note} A \tcode{reinterpret_cast}
+\tcode{(void*)nullptr} to the integral type. \begin{note} A \tcode{reinterpret_cast}
```
cpplearner: I'm in doubt about this change. Won't this become a circular definition?

source/basic.tex
Eelis: Good point, I've added a second commit to address that.

source/support.tex
```diff
@@ -85,7 +85,7 @@
 and
 \tcode{0L},
 but not
-\tcode{(void*)0}.}
```
tkoeppe: This should probably stay.

Eelis: Fair enough, reverted that one.

tkoeppe:
@jwakely, @zygoloid: PTAL

source/basic.tex
```diff
@@ -2520,7 +2520,7 @@
 s)~(\ref{multibyte.strings}) and \tcode{argv[0]} shall be the pointer to
 the initial character of a \ntmbs that represents the name used to
 invoke the program or \tcode{""}. The value of \tcode{argc} shall be
-non-negative. The value of \tcode{argv[argc]} shall be 0. \begin{note} It
+non-negative. The value of \tcode{argv[argc]} shall be the null pointer value. \begin{note} It
```
zygoloid: the -> a, please

source/iostreams.tex
```diff
@@ -1509,7 +1509,7 @@
 \tcode{parray[idx]}.
 On failure a valid
 \tcode{void*\&}
-initialized to 0.
+initialized to the null pointer value.
```
zygoloid: This doesn't make any sense, with or without your change. You can't initialize a `void*&` to a null pointer value. To what object is this reference supposed to be bound? This seems like we need an LWG issue to say what this actually does.

source/basic.tex
zygoloid: Also in the other occurrences below. We can have "the null pointer value of some type", but "the null pointer value" without a type doesn't make sense, whereas "a null pointer value" seems to work.

tkoeppe:
@Eelis: ping

PR #879, opened 2016-08-03
[input.output] Review library index for clause 27

AlisdairM:

  apply indexlibrarymember for all member functions other than constructors/destructors
  ensure headers are indexed with synopsis
  ensure every itemdecl has a library index entry
  fix some mislabeled index entries
  minor reordering of paragraphs so istream/ostream have consistent layout
  add a missing entry to index of implementation defined behavior

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -95,8 +95,9 @@ are other types is
 \pnum
 In the classes of Clause~\ref{input.output}, a template parameter with name
 \tcode{charT} represents a member of the set of types containing \tcode{char}, \tcode{wchar_t},
-and any other implementation-defined character types that satisfy the requirements for
-a character on which any of the iostream components can be instantiated.
+and any other \impldef{set of character types that iostreams templates can be instantiated for}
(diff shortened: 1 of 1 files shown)


tkoeppe:
Hm, I think I would prefer these distinct changes as separate commits. They can be in the same PR, just as separate commits. I'll take this one and fix it up for now.

source/iostreams.tex
```diff
@@ -5914,6 +5916,19 @@
 \postcondition
 \tcode{rdbuf() == sb}.
 
+\indexlibrary{\idxcode{basic_ostream}!constructor}%
+\begin{itemdecl}
+basic_ostream(basic_ostream&& rhs);
```
tkoeppe: Why did you add a new constructor here? Has that one been missing erroneously?

tkoeppe: This constructor does not appear in the synopsis. Why is this?

tkoeppe:
Superseded by PR #891.

PR #881, opened 2016-08-04
[system_error.syn] [syserr.compare] Declare all nonmembers in the synopsis

### LINKED ISSUE #880: operator< for error_condition in wrong(?) place
`std::error_condition` is LessThanComparable, but the `operator<` overload does not appear in the header synopsis in [system_error.syn].

It is tucked away at  the end of the class synopsis in [syserr.errcondition.overview] and then defined in [syserr.errcondition.nonmembers], when [syserr.compare] would seem more sensible.

jwakely:
Move definitions of less-than operators to [syserr.compare]

Fixes #880

### DIFF
--- a/source/diagnostics.tex
+++ b/source/diagnostics.tex
@@ -880,10 +880,19 @@ namespace std {
   
   template <> struct is_error_condition_enum<errc> : true_type {};
 
+  // \ref{syserr.errcode.nonmembers} non-member functions:
   error_code make_error_code(errc e) noexcept;
+
+  template <class charT, class traits>
+    basic_ostream<charT, traits>&
+      operator<<(basic_ostream<charT, traits>& os, const error_code& ec);
+
+  // \ref{syserr.errcondition.nonmembers} non-member functions:
   error_condition make_error_condition(errc e) noexcept;
 
   // \ref{syserr.compare}, comparison operators:
+  bool operator<(const error_code& lhs, const error_code& rhs) noexcept;
+  bool operator<(const error_condition& lhs, const error_condition& rhs) noexcept;
   bool operator==(const error_code& lhs, const error_code& rhs) noexcept;
   bool operator==(const error_code& lhs, const error_condition& rhs) noexcept;
   bool operator==(const error_condition& lhs, const error_code& rhs) noexcept;
@@ -1192,7 +1201,6 @@ namespace std {
 
   // \ref{syserr.errcode.nonmembers} non-member functions:
   error_code make_error_code(errc e) noexcept;
-  bool operator<(const error_code& lhs, const error_code& rhs) noexcept;
 
   template <class charT, class traits>
     basic_ostream<charT, traits>&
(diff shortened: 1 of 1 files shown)


source/diagnostics.tex
```diff
@@ -1192,7 +1201,6 @@
 
   // \ref{syserr.errcode.nonmembers} non-member functions:
   error_code make_error_code(errc e) noexcept;
-  bool operator<(const error_code& lhs, const error_code& rhs) noexcept;
 
   template <class charT, class traits>
     basic_ostream<charT, traits>&
```
zygoloid: Did you mean to delete this one when adding the one above?

source/diagnostics.tex
```diff
@@ -1560,6 +1551,22 @@
...
+\begin{itemdecl}
+bool operator<(const error_code& lhs, const error_code& rhs) noexcept;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\returns
+\begin{codeblock}
+lhs.category() < rhs.category() ||
+(lhs.category() == rhs.category() && lhs.value() < rhs.value()).
```
zygoloid: It'd be nice to fix this weird non-code period in code font.

source/diagnostics.tex
jwakely: I'm pretty sure I did mean to, because I moved it to line 894.

We could have it in both places (as we do for `make_error_code` and `operator<<`), but if we do that then we should also have `operator==` and the other overloads in both places. IMHO it makes no sense for _only_ `operator<` to be in the class synopsis.

source/diagnostics.tex
jwakely: Changed to a semi-colon.

source/diagnostics.tex
zygoloid: Sorry, I wasn't very clear. The "this one" I'm referring to is the `operator<<` whose first line is highlighted here (github doesn't allow adding comments outside the diff context, annoyingly). I think that should be deleted, because you added it to the synopsis, but it hasn't been.

jwakely: Gotcha. I'll do that.

jwakely: Ah, I think I did it that way because `make_error_code(errc e)` is also in both places, and `operator<<` is grouped with that (both before and after my change) in [syserr.errcode.nonmembers], but `operator<<` was missing from the header synopsis, and after my change is not in [syserr.errcondition.nonmembers].

If I remove `operator<<` from the class synopsis I should also remove `make_error_code(errc)`. I think I prefer not to remove either.

I moved the relational operators to be alongside the equality operators, which are only in the header synopsis not the class synopses. That's because the equality ops are in [syserr.compare] which is at the same level as `error_code` and `error_condition`, rather than being a "non-member functions" subclause.

jwakely: I feel that everything should be in the header synopsis, otherwise that synopsis is incomplete. Whether some non-members are also duplicated in the class synopses is less important, but I made the call based on whether they are defined in a subclause of the class, or in a sibling clause.

PR #884, opened 2016-08-05
Remove redundant `this->` in library specification

timsong-cpp:
in [string::compare] and [any.modifiers].

Also fix a missing `\tcode` in [variant.assign] discovered on the way.

### DIFF
--- a/source/strings.tex
+++ b/source/strings.tex
@@ -3617,7 +3617,7 @@ int compare(size_type pos1, size_type n1,
 \effects
 Equivalent to:
 \begin{codeblock}
-return basic_string_view<charT, traits>(this.data(), pos1, n1).compare(sv);
+return basic_string_view<charT, traits>(data(), pos1, n1).compare(sv);
 \end{codeblock}
 \end{itemdescr}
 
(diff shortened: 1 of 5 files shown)


source/strings.tex
```diff
@@ -3617,7 +3617,7 @@
 \effects
 Equivalent to:
 \begin{codeblock}
-return basic_string_view<charT, traits>(this.data(), pos1, n1).compare(sv);
```
tkoeppe: Do we ever say `this->` elsewhere? Wouldn't we just say `data()`?

timsong-cpp: The other `any::emplace`, for one. A quick search shows a few dozen `this->`s in the library clauses.

zygoloid: Almost all of them appear to be very recent additions and redundant. I would prefer to remove all of these occurrences that are neither necessary ([filebuf.assign]/1's `this->` is necessary since `close` is a member of a dependent base class) nor significantly improve clarity.

timsong-cpp: OK, I'll prepare a commit removing them all instead.

What dependent base, though? `close` is a member of `basic_filebuf`, not `basic_streambuf`.

timsong-cpp:
I left four `this->` unchanged; two are in [bitset.members] (non-const `operator[]`), one in [unique.ptr.single.modifiers] (`reset()`), and one in [memory.resource.pool.ctor] (note in constructor's description).

source/strings.tex
tkoeppe: @zygoloid, @jwakely: PTAL

PR #885, opened 2016-08-05
[thread.thread.id] Consistently use an lvalue reference for operator<<'s first parameter.

Eelis:
(Note how the parameter type /is/ specified correctly in the synopsis at the start of the section.)


jwakely:
Good catch, thanks.

PR #886, opened 2016-08-05
g/special math functions/s//mathematical special functions/

W-E-Brown:
There's nothing special about the math; mathematicians have long termed these functions as "special functions".  Because C++ also uses "special functions" as a different term of art (referring to copy c'tors, d'tors, etc.), we disambiguate by prefixing "mathematical".

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -10074,7 +10074,7 @@ namespace std {
   int isunordered(double x, double y);
   int isunordered(long double x, long double y);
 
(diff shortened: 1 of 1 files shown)


jwakely:
Yes I was going to make this change myself. This also makes the subclause title consistent with the IS that the functions came from, "Extensions to the C++ Library to Support Mathematical Special Functions."

Thanks for finding all the spots to change.

PR #889, opened 2016-08-06
[futures.promise] Make swapping std::promises consistently noexcept.

Eelis:


K-ballo:
The `function` bits are LWG2062

Eelis:
Ok, removed the function bits. The remaining one for std::promise is probably more editorial, because that particular std::swap /is/ specified as noexcept in the synopsis at the start of the section.

PR #891, opened 2016-08-06
Editorial review of Clause 27 [input.output]

tkoeppe:
This is @AlisdairM's work, just repackaged.

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -12102,7 +12102,7 @@ namespace std::filesystem {
 
     const path& path1() const noexcept;
     const path& path2() const noexcept;
-    const char* what() const noexcept;
+    const char* what() const noexcept override;
   };
 }
 \end{codeblock}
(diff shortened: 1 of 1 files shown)


source/iostreams.tex
```diff
@@ -5916,6 +5916,19 @@
 \postcondition
 \tcode{rdbuf() == sb}.
 
+\indexlibrary{\idxcode{basic_ostream}!constructor}%
```
tkoeppe: @AlisdairM: Same question here, how come the synopsis does not list the move constructor?

AlisdairM: Looks like a mis-application of http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2005/n1862.html after the Oxford 2007 meeting that no-one had noticed yet!  I think the declaration in the class definition could be added editorially, given it was present in the original paper.

AlisdairM: Ah, and the other half of the mystery resolved! LWG 911 made these members protected, which is why they show up at the /end/ of the class definition - but the declarations are present and correct.

tkoeppe: Ahhh - thanks! I added reference comments to the synopsis, too.

PR #892, opened 2016-08-06
Use 'if and only if' for conditions that are sufficient as well as necessary.

Eelis:


jensmaurer:
We believe some of the changes in the core section are actively harmful (and wrong), since we're often not establishing equivalence, but defining a term.

For some of these, moving the "only" might be worthwhile, but that would be a different change.

PR #893, opened 2016-08-06
[depr.func.adaptor.typedefs] Clarify that reference_wrapper does not define argument_type for non-nullary member function pointer types.

Eelis:


source/future.tex
```diff
@@ -1468,8 +1468,7 @@
 only if the type \tcode{T} is any of the following:
 \begin{itemize}
 \item a function type or a pointer to function type taking one argument of type \tcode{T1}
-% FIXME: Should this be R T0::f() ?
-\item a pointer to member function \tcode{R T0::f} \cv{} (where \cv{} represents the member function's cv-qualifiers); the type \tcode{T1} is \cv{} \tcode{T0*}
+\item a pointer to member function \tcode{R T0::f()} \cv{} (where \cv{} represents the member function's cv-qualifiers); the type \tcode{T1} is \cv{} \tcode{T0*}
```
Eelis: I think it makes sense with the fix. As an example of what the item is talking about, consider the member function:

    size_t string::size() const

(R=size_t, T0=string, f=size, cv=const)

The item now says that if T (reference_wrapper's parameter) is a pointer to this thing, then T1 (the resulting argument_type) is `cv T0*`, which is `const string*`.

So it is saying that `reference_wrapper<size_t(string::*)()const>::argument_type` is `const string*`, which is also what I get with libstdc++ and libc++.

Eelis: (I agree it's sloppy and confusing phrasing though.)

zygoloid: This is talking about a type `T`. There's no way a type can be "a pointer to member function `size_t string::size() const`" -- that's not a property of a type.

My concerns here are twofold:
1) This may be a normative change, depending on whether an `argument_type` was also supposed to be defined for a pointer to member function with parameters or not. (With your change we seem to say only the no-parameter case is covered, whereas without it, this appears to apply regardless of whether the function has parameters.)
2) This change still leaves the wording meaningless.

I think we do want to make a change in this direction (and I suspect the right change would be to use `R T0::*f()`*`cv`*), but I'd like LWG to look at this and decide what exactly to do.

tkoeppe:
@zygoloid: What's the status of this pull request? Did anyone ever ask LWG for input?

@mclow: Could you perhaps advise on whether a) this is something worth revising, and b) whether this is editorial?

source/future.tex
jwakely: I think it is supposed to be `R T0::f()` _`cv`_, and should be read as:

> a pointer, to a member function of the form `R T0::f()` _`cv`_

i.e.

> a pointer to member function that points to a member function `R T0::f()` _`cv`_


It certainly would be clearer to talk about the type of the pointer to member function, not show a declaration of a member function.

jwakely: i.e. Eelis's suggestion is not a normative change, the intention, and I believe what all implementations do, is to only define `argument_type` when it's a pointer to member function for a member function taking no parameters.

For a member function taking exactly one parameter `first_argument_type` and `second_argument_type` are defined.

For member functions taking two or more parameters none of those typedefs is defined.

jwakely: `tr1::reference_wrapper` says:

> — a pointer to member function type with cv-qualifier _cv_ and no arguments; the type `T1` is _cv_ `T*` and `R` is the return type of the pointer to member function

and then for the `R T0::f(T2)` case:

> — a pointer to member function with cv-qualifier _cv_ and taking one argument of type `T2`; the type `T1` is _cv_ `T*` and `R` is the return type of the pointer to member function

So at some point we rephrased that and mangled it.

jwakely: It was changed by http://www.open-std.org/jtc1/sc22/wg21/docs/lwg-defects.html#521 and it was meant to have the `()` but the editor didn't apply the DR correctly.

Eelis:
Rebased

jwakely:
As I said in the now-collapsed comments on the original commit, the proposed change simply fixes the missing `()` from the resolution of [LWG DR 521](http://www.open-std.org/jtc1/sc22/wg21/docs/lwg-defects.html#521).

I can't speak for @mclow but my recommendation would be to merge this and not spend time trying to improve the wording for these deprecated typedefs.

tkoeppe:
@jwakely: OK, indeed. Merge that noise.

mclow:
> OK, indeed. Merge that noise.
Works for me.

PR #896, opened 2016-08-06
[set.cons, multiset.cons] Specify correct constructed type and add missing \tcode.

Eelis:


source/containers.tex
```diff
@@ -6946,7 +6946,7 @@
 \begin{itemdescr}
 \pnum
 \effects
-Constructs an empty set using the specified comparison objects and allocator.
+Constructs an empty \tcode{set} using the specified comparison objects and allocator.
```
tkoeppe: I believe the original wording may be as intended here. We also say "map" elsewhere to refer to both ordered and unordered maps. It seems OK to use the abstract concept rather than the C++ typename here. I'd leave it as it is.

Eelis: Well, the second std::set constructor in [set.cons] /does/ have the \tcode. I was merely trying to make it consistent.

tkoeppe: OK, I see. That's fine then. Thanks.

PR #898, opened 2016-08-11
[expr.call]p1 Use a more idiomatic way to specify the expression has undefined behavior

AaronBallman:
I find the phrasing to be a little weird to just say "is undefined" and think it's nominally more clear to state that the behavior of the call is what's undefined.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -1484,7 +1484,7 @@ conversion~(\ref{conv.func}) is suppressed on the postfix expression),
 or it shall have function pointer type. Calling a function through an
 expression whose function type has a language linkage that is different
 from the language linkage of the function type of the called function's
-definition is undefined~(\ref{dcl.link}). For a call to a non-static
+definition results in undefined behavior~(\ref{dcl.link}). For a call to a non-static
 member function,
 the postfix expression shall be an
 implicit~(\ref{class.mfct.non-static},~\ref{class.static}) or explicit


Eelis:
Makes sense to me. There's another one here: http://eel.is/c++draft/util.dynamic.safety#10

"Hence indirection through a pointer located there **is undefined** if the object it points to was created by global operator new and not previously declared reachable."

AaronBallman:
If the editors like this direction, I can do a more thorough pass through the standard and correct them all at once in a "big" patch. Or, I can correct them with individual patches.

Eelis:
I'm no editor, but I think that's a great idea, because there's also dubious stuff like (for valarray::min(), http://eel.is/c++draft/valarray.members#6):

> The value returned for an array of length 0 is undefined.

This makes it _sound_ like the undefinedness could be confined to the return value itself (so that other behavior would still be defined), but in reality, this is full-blown undefined behavior (libstd++ just gives an assertion failure). :)

AlisdairM:
Again, not an editor :)

Looking at 1.9p4  [intro.exection] we see that operations may be described as undefined, and the original wording is consistent with that form of specification.  If you want to make a pass for more consistent usage of the term, I would start there - and make sure the standard remains consistent with that approach (so we don't editorially break something).

It seems like a lot of work, and I am only mildly in favor (but still in favor of you doing the work, not opposed ;))

Eelis:
Yeah, I see a lot of "the effect is undefined", and that seems fine. But "the value is undefined" sounds different to me.

AaronBallman:
Alisdair, you bring up a good point about 1.9p4 that kind of deflates this particular change. However, I still find the wording to be kind of obtuse, whereas "results in undefined behavior" is a bit easier to intuit for [expr.call], IMO.

I'll let the editors make the call -- if they like the new wording, great, and if they don't, feel free to close the pull request. :-)

tkoeppe:
@zygoloid: What do you think of Aaron's suggestion of a general wording review regarding how we say that the behaviour is undefined?

zygoloid:
This change is good. If a general wording review would find more like this, that'd be nice too!

PR #899, opened 2016-08-11
[whole standard] Audit index of implementation-defined behavior

AlisdairM:
Review all uses of the terms 'implementation-defined' and
'implementation defined' in the standard, and replace with
\impldef entries in the index of implementation defined
behavior where appropriate.

Clean up some older index entries that did not have a clear
reference, and appear to predate the index of implementation
defined behavior being added by Pete Becker for C++11.

Changes may appear more disruptive than in practice, due to
word-wrapping reflowing a paragraph or two in the doc source,
but not on the rendered pdf.


source/algorithms.tex
```diff
@@ -1277,7 +1277,8 @@
 If the threads of execution created by \tcode{std::thread} provide concurrent
 forward progress guarantees~(\ref{intro.progress}), then a thread of execution
 implicitly created by the library will provide parallel forward progress guarantees;
-otherwise, the provided forward progress guarantee is implementation-defined.
+otherwise, the provided forward progress guarantee is
+\impldef{forward progress guarantees for implicits thread of parallel agorithms (if not defined for \tcode{std::thread})}.
```
zygoloid: Typos: "implicits thread" -> "implicit threads", "agorithms"

source/algorithms.tex
```diff
@@ -1395,7 +1396,8 @@
 
 \pnum
 The semantics of parallel algorithms invoked with an execution policy object of
-implementation-defined type are implementation-defined.
+\impldef{additional execution policies supported by parallel algorithms} type are
```
zygoloid: I assume this form of implementation-defined behavior is introduced elsewhere. Can we check that we use the same index entry for both places?

source/basic.tex
```diff
@@ -2889,8 +2889,8 @@
 passing an invalid pointer value to a deallocation function
 have undefined behavior.
 Any other use of an invalid pointer value
-has implementation-defined behavior.%
-\footnote{Some implementations might define that
+has \impldef{any use of an invalid pointer other than to dereference or deallcate}
```
zygoloid: "deallcate"

zygoloid: Maybe dereference -> perform indirection?

source/compatibility.tex
```diff
@@ -1482,8 +1482,9 @@
 Valid \CppXIV code that uses trigraphs may not be valid or may have different
 semantics in this International Standard. Implementations may choose to
 translate trigraphs as specified in \CppXIV if they appear outside of a raw
-string literal, as part of the implementation-defined mapping from physical
-source file characters to the basic source character set.
+string literal, as part of the \impldef{mapping physical source file characters
```
zygoloid: Maybe add "of" or "from" after "mapping"?

source/conversions.tex
```diff
@@ -176,7 +176,8 @@
 
 \item Otherwise, if the object to which the glvalue refers contains an invalid
 pointer value~(\ref{basic.stc.dynamic.deallocation},
-\ref{basic.stc.dynamic.safety}), the behavior is implementation-defined.
+\ref{basic.stc.dynamic.safety}), the behavior is
+\impldef{lvalue-to-rvalue conversion of an invalid pointer}.
```
zygoloid: pointer -> pointer value

source/lex.tex
```diff
@@ -181,7 +181,8 @@
 the members of the basic source character set are intended to
 identify characters from the subset of ISO/IEC 10646 which corresponds to the ASCII
 character set. However, because the mapping from source file characters to the source
-character set (described in translation phase 1) is specified as implementation-defined,
+character set (described in translation phase 1) is specified as
+\impldef{mapping physical source file characters to basic source character set},
```
zygoloid: Keep this matching the other occurrence of this phrase if we add "of" or "from" there.

source/lex.tex
```diff
@@ -474,12 +475,13 @@
...
-is conditionally-supported with implementation-defined semantics,
-as is the appearance of the
-character \tcode{"} in an \grammarterm{h-char-sequence}.\footnote{Thus, a
-sequence of characters that resembles an escape sequence might result in an
-error, be interpreted as the character corresponding to the escape sequence, or
-have a completely different meaning, depending on the implementation.}%
+is conditionally-supported with \impldef{meaning of \tcode{'}, \tcode{\textbackslash},
+\tcode{/*}, or \tcode{//} in a \grammarterm{q-char-sequence} or an
+\grammarterm{h-char-sequence}} semantics, as is the appearance of the character
+\tcode{"} in an \grammarterm{h-char-sequence}.\footnote{Thus, a sequence of characters
```
zygoloid: We should presumably have some entry for the `"`-in-*h-char-sequence* case, perhaps add words for this case to the prior \impldef?

source/utilities.tex
```diff
@@ -11018,7 +11018,8 @@
 from the upstream memory resource~(\ref{memory.resource.monotonic.buffer})
 to replenish a pool.
 If the value of \tcode{max_blocks_per_chunk} is zero or
-is greater than an implementation-defined limit, that limit is used instead.
+is greater than an \impldef{maximum maximum number of blocks to replenish a pool}
```
zygoloid: "maximum maximum"

source/utilities.tex
```diff
@@ -11036,7 +11037,8 @@
 Attempts to allocate a single block larger than this threshold
 will be allocated directly from the upstream memory resource.
 If \tcode{largest_required_pool_block} is zero or
-is greater than an implementation-defined limit, that limit is used instead.
+is greater than an \impldef{largest allocation satisfied directly by a pool}
```
zygoloid: I don't think this is quite right; the thing that's implementation-defined is the largest value that is supported as the largest allocation to be satisfied by a pool. But some less-confusing way to word that would be preferred :)

source/lex.tex
AlisdairM: This is a slightly more complex rewording than I, as a non-Core-competent drafter, am comfortable with.  I agree a change for the index is important, but without "implementation-defined" occurring as a phrase in the current paragraph, I would need to substantially reword in order to introduce an \impldef{}.  As the rest of this patch is entirely such simple substitutions, I would prefer to open an issue, and land the rest of this patch.

source/basic.tex
AlisdairM: Done and Done

source/compatibility.tex
AlisdairM: Chose 'from'

source/conversions.tex
AlisdairM: Done

source/lex.tex
AlisdairM: Done

source/utilities.tex
AlisdairM: maximum maximum is actually correct, but probably too cute - reworded.

source/algorithms.tex
AlisdairM: Done.

PR #903, opened 2016-08-17
chmod 644 source/{containers,lib-intro}.tex

wilhelmtell:
Minor filesystem permissions fix on a couple of source tex files. :) It's not a very good idea for a source file to be executable. :)


jwakely:
It looks like I made them executable, though I have no idea how. Thanks for fixing it.

tkoeppe:
@jwakely: I was assuming that you misread "executive summary" as "executable summary". :-)

PR #907, opened 2016-08-25
[class.local, facet.num.{get,put}.virtuals, temp.expl.spec] Remove stray whitespace.

Eelis:


Eelis:
Oh don't merge yet, spotted another one.

PR #917, opened 2016-09-06
[dcl.type.auto.deduct] Use void() instead of the ill-formed void{}.

timsong-cpp:


lessness:
Interesting. This looks the same as reported on [StakOverflow](http://stackoverflow.com/questions/39354065/is-void-legal-or-not) and on [isocpp](https://groups.google.com/a/isocpp.org/forum/embed/?place=forum/std-discussion&showsearch=true&showpopout=true&parenturl=https%3A%2F%2Fisocpp.org%2Fforums%2Fiso-c-standard-discussion#!topic/std-discussion/g6Rpo1NSHOw) by the same author.
Is this a duplicate or a complete different subject?

timsong-cpp:
edadb9b fixed this.

Issue #905, opened 2016-08-23
type traits do not show in the index of library names


AlisdairM:
The type traits are defined by a sequence of table, which do not have appropriate \indexlibrary tags to create entries in the index of library names.

PR #922, opened 2016-09-09
[meta] Add index entries for each type trait

AlisdairM:
Create an index entry for each row in the type traits tables,
indexing the corresping trait.  Where a trait is defined outside
the table, add a second index reference to the latter.  This
causes an annoying duplication, as the current software sees the
index entry inside the table as in some way NOT the same as the
entry outside the table.

Disambiguate the is_empty function from the filesystem library,
and the is_empty trait.  The issue for is_signed being a trait
and a member of numberic_limits resolves itself.

This is intended to resolve https://github.com/cplusplus/draft/issues/905

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -14698,6 +14698,10 @@ namespace std {
 }
 \end{codeblock}
 
+\indexlibrary{\idxcode{integral_constant}}%
+\indexlibrary{\idxcode{bool_constant}}%
+\indexlibrary{\idxcode{true_type}}%
+\indexlibrary{\idxcode{false_type}}%
 \pnum
 The class template \tcode{integral_constant},
 alias template \tcode{bool_constant}, and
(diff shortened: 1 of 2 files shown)


Eelis:
I see there's currently no index entry for void_t. Maybe worth adding?

AlisdairM:
I think I would rather handle this under a follow-up patch.

Adding an entry for 'void_t' makes sense, but currently there are no entries for any other '_t' alias in this patch.  I am not currently aware of precedent for sprinkling indexing markup into header synopses.  'void_t' is definitely a noteworthy special case though.

Eelis:
Ok makes sense, thanks

source/utilities.tex
```diff
@@ -15739,6 +15847,7 @@
  needed when only explicit conversions are desired among the template arguments.
  \end{note} \\ \rowsep
 
+\indexlibrary{\idxcode{underlying_type}}%
```
zygoloid: Seems to be missing the end of the line.

zygoloid: Looks like this might be a github bug? Some other views of this file show the file ending here!

zygoloid:
Looks like we hit a github bug here; manually checked and merged as d5a02cebd5b393f52896b362e436d3c42c4af310

PR #927, opened 2016-09-20
[stmt.return] Change 'void' to 'cv void'

jlaire:
Using "cv void" here is consistent with 3.9.1p9.

It's not clear to me whether the _cv_ of the expression and the _cv_ of the function are independent meta variables or whether they should match, both in 3.9.1p9 and 6.6.3p2. Can a function with any _cv_-qualified `void` return type return a `void` with a different _cv_-qualifier?

```
const volatile void fcv() {}
void f() { return fcv(); }
```

A note about formatting: basic.tex uses `\term{cv}` but statements.tex uses `\cv`. There are also instances of `\cvqual{cv}` in other files.

### DIFF
--- a/source/statements.tex
+++ b/source/statements.tex
@@ -810,7 +810,7 @@ no operand shall be used only in a function whose return type is
 destructor~(\ref{class.dtor}).
 \indextext{\idxcode{return}!constructor~and}%
 \indextext{\idxcode{return}!constructor~and}%
-A return statement with an operand of type \tcode{void} shall be used only
+A return statement with an operand of type \cv\ \tcode{void} shall be used only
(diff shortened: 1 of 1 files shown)


zygoloid:
An expression cannot have type cv void. All void-typed expressions are prvalues, and prvalues of non-class, non-array type are never cv-qualified; the type of `fcv()` in your example is just `void`.

jlaire:
So should 3.9.1p9 be changed to remove the cv? "An expression of type cv void shall be used only ..."

jensmaurer:
"So should 3.9.1p9 be changed to remove the cv?"

We agree this is technically redundant, but given the context in which it appears (see prior sentence), we decided not to make any change.

PR #930, opened 2016-09-24
[diff.cpp14.special] Clarify example comment

kpfleming:
The comment describing the reason for the new behavior was subject to
misinterpretation; a reader may have thought that the substition
failure was the cause of A(int) being called, which is not the case.
This patch reorders the clauses of the sentence so that they follow
a logical cause-and-effect order.

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -1606,9 +1606,9 @@ struct B : A {
   using A::A;
   B(int);
 };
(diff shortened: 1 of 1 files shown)


zygoloid:
That isn't a misinterpretation: a substitution failure *was* the reason that `A(int)` was called rather than the derived class inherited constructor calling the corresponding base class constructor.

PR #931, opened 2016-09-24
[basic.link] consistent indexing of 'translation unit'

AlisdairM:
Ensure all definitions of 'translation unit' sort together
in the index, with a common spelling of the white-space.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -78,7 +78,7 @@ the same literal suffix identifier.
 \end{itemize}
 
 \pnum
-\indextext{translation~unit!name~and}%
+\indextext{translation unit!name and}%
 \indextext{linkage}%
 A name used in more than one translation unit can potentially
 refer to the same entity in these translation units depending on the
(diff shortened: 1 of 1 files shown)


source/basic.tex
```diff
@@ -2233,8 +2233,8 @@
 
 \pnum
 \indextext{program}%
-A \defn{program} consists of one or more \defn{translation
-units} (Clause~\ref{lex}) linked together. A translation unit consists
+A \defn{program} consists of one or more \defn{translation units}
```
tkoeppe: Do we really need a separate entry for "translation units" (plural)? Can we not index this also under "translation unit"?

AlisdairM: I thought about changing this, but the term is a \defn today, and I don't think we have precedent for moving the 's' outside the \defn.  I certainly did not want to rewrite core wording so that this term might be re-rendered as singular, so opted for the lesser among evils,  leaving the term as defined.

tkoeppe: The usual solution would be `\defnx{plural}{singular}`. We've done that or similar things elsewhere. I think that'd be appropriate here.

tkoeppe: Examples:

- `\defnx{encoded character types}{encoded character type}`
- `\defnx{makes progress}{make progress}` :-)

tkoeppe:
@AlisdairM: Ping

AlisdairM:
Sorry, got distracted with a few other issues, will try to push an updated patch tomorrow.  Thanks for following up :)

jensmaurer:
Alisdair, we're still waiting for the updated patch.

AlisdairM:
Updated patch available, using the \defnx macro as suggested.

PR #934, opened 2016-09-25
Consistently use the stdlib designations defined in [intro.refs]/2 and [library.general]/1.

Eelis:


source/lib-intro.tex
```diff
@@ -2781,7 +2781,7 @@
 
 \pnum
 Except where explicitly specified in this standard, it is \impldef{which functions in
-Standard C++ library may be recursively reentered} which functions in the Standard \Cpp
+the C++ standard library may be recursively reentered} which functions in the \Cpp standard
```
zygoloid: Can you change this to \Cpp while you're here?

Eelis:
Good idea, done

source/lib-intro.tex
Eelis: Good idea, done

PR #942, opened 2016-10-03
[class, class.derived, special] Write space in 'access control' consistently in index keys.

Eelis:
Other places already use a space instead of a tilde, and this distinction currently results in two separate index key nodes.


tkoeppe:
Can you make a diffpdf of the relevant change to the index?

Eelis:
First page of diffs: diff
Rest is just reflow.

PR #943, opened 2016-10-03
Fix lots of see/seealso references, especially regarding operators.

Eelis:
To fix dead "links" and make the operator index entries more consistent.


tkoeppe:
Can we see some representative diffs?

Eelis:
It's spread out a bit, but the two most important pages are these

PR #948, opened 2016-10-13
[basic, stmt, dcl.dcl] Move surrounding punctuation out of \grammarterm arguments.

Eelis:
Btw, you may notice the inconsistency in whether parentheses are used after "see". That happens in a couple other places in the document as well, and I'll submit a separate patch for that.


Eelis:
Oh, don't merge yet, found a few more occurrences that I'll include.

Eelis:
There, that's all the ones I can find.

PR #951, opened 2016-10-13
[optional.nullopt] Add missing 'the'.

Eelis:
(Pure coincidence that I spotted two of these today. Unfortunately I have no way to search for these systematically.)

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -2969,7 +2969,7 @@ In particular, \tcode{optional<T>} has a constructor with \tcode{nullopt_t} as a
 this indicates that an optional object not containing a value shall be constructed.
 
 \pnum
-Type \tcode{nullopt_t} shall not have a default constructor. It shall be a literal type. Constant \tcode{nullopt} shall be initialized with an argument of literal type.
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -2969,7 +2969,7 @@
 this indicates that an optional object not containing a value shall be constructed.
 
 \pnum
-Type \tcode{nullopt_t} shall not have a default constructor. It shall be a literal type. Constant \tcode{nullopt} shall be initialized with an argument of literal type.
+Type \tcode{nullopt_t} shall not have a default constructor. It shall be a literal type. The constant \tcode{nullopt} shall be initialized with an argument of literal type.
```
tkoeppe: No definite article here?

Can you locate the paper that provided the wording, just for double-checking?

Eelis: The "the" is missing here as well: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3793.html

Eelis: I guess it's technically grammatically correct without the "the", too...

PR #954, opened 2016-10-18
[stmt.stmt] condition contains brace-or-equal

lichray:


tkoeppe:
I take it this is just meant to be a simplifying reuse of an existing intermediate production?

lichray:
Yes.  Looks like something predates brace-or-equal, and the "elaborated" BNFs are not used in *condition*'s specification.

PR #957, opened 2016-10-25
Special math functions editorial issues found by Matwey V. Kornilov f…

apolukhin:
…rom Sternberg Astronomical Institute, Lomonosov Moscow State University, Russia:
- Clause "Associated Legendre polynomials" is wrongly entitled. "Associated Legendre functions" would be more appropriate here. Though "Associated Legendre polynomials" term is sometimes used it is formally wrong term. A polynomial (by definition) is a particular kind of function which can be represented using only finite number of additions, multiplications and exponentiations to a non-negative power, i.e. in canonical form of `SUM(AiX^i)`. Obviously, some of P^m_l are not polynomials. For instance, for m=l=1, `P11(x) == sqrt(1 − x*x)` is not representable as `SUM(AiX^i)`. See for reference: Abramowitz and Stegun, Chapter 8 "Legendre Functions".
- "[sf.cmath.cyl_bessel]" is a bad name for the tag. "[sf.cmath.cyl_bessel]" sounds like "Bessel functions" and when people say "Bessel functions" they usually mean Jν from [sf.cmath.cyl_bessel_j]. Replaced "[sf.cmath.cyl_bessel]" with "[sf.cmath.cyl_bessel_i]".
- "[sf.cmath.cyl_bessel_k]" misses references to "[sf.cmath.cyl_bessel_j]" and "[sf.cmath.cyl_neumann]" in the "See also" section. In [sf.cmath.cyl_bessel_j] Jv(x) is defined, in [sf.cmath.cyl_neumann] Nν(x) is defined - both of them are used in the "Returns:" section of the [sf.cmath.cyl_bessel_k].

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -10081,7 +10081,7 @@ namespace std {
   float        assoc_laguerref(unsigned n, unsigned m, float x);
   long double  assoc_laguerrel(unsigned n, unsigned m, long double x);
 
-  // \ref{sf.cmath.assoc_legendre}, associated Legendre polynomials:
+  // \ref{sf.cmath.assoc_legendre}, associated Legendre functions:
   double       assoc_legendre(unsigned l, unsigned m, double x);
   float        assoc_legendref(unsigned l, unsigned m, float x);
   long double  assoc_legendrel(unsigned l, unsigned m, long double x);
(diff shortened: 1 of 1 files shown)


zygoloid:
@W-E-Brown This looks correct to me; what do you think?

tkoeppe:
[Walter says some correspondence has got lost here. I'll try to chase up @apolukhin in person.]

W-E-Brown:
> On Nov 9, 2016, at 1:04 PM, Thomas Köppe <notifications@github.com> wrote:
> 
> [Walter says some correspondence has got lost here. I'll try to chase up @apolukhin in person.]

Here's my previous response:

>> Begin forwarded message:
>> 
>> Subject: Re: [cplusplus/draft] Special math functions editorial issues found by Matwey V. Kornilov f… (#957)
>> From: W Brown <webrown.cpp@gmail.com>
>> Date: 2016-10-25 at 6:40:27 PM CDT
>> To: cplusplus/draft <reply@reply.github.com>
>> 
>>

tkoeppe:
@apolukhin checked the wording in ISO-80000-2, and apparently they also changed the term from "associated Legendre polynomial" to "associated Legendre function", so this change does indeed keep us in sync with both ISO and factual reality.

tkoeppe:
For the record, ISO-80000-2 also uses the correct term "function":

image

PR #967, opened 2016-11-06
[string.view.io] Update Effects: with P0254R2 changes

johelegp:


johelegp:
This is actually covered by the LWG issue [2755](http://cplusplus.github.io/LWG/lwg-active.html#2755). Closing.

PR #969, opened 2016-11-06
[associative.reqmts, unord.req] insert_return_type only for unique keys

danieljames:
The requirements for associative and unordered associative containers require `insert_return_type` for all containers, but it's only defined for containers with unique keys.

I'm not sure about the wording, I followed the example of other entries in the table which say things like (map and multimap only), although it might be better to say it's only for containers that support unique keys.

I think this is an editorial change, because it just changes the requirements to match what's specified elsewhere.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -1615,7 +1615,7 @@ and \tcode{nh} denotes a non-const rvalue of type \tcode{X::node_type}.
  see~\ref{container.node} &
  compile time \\ \rowsep
 
-\tcode{X::insert_return_type} &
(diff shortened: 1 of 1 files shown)


jwakely:
I agree this is an issue, but I don't think it's an editorial one. It's possible that the return type will be changed to `pair<iterator, bool>` anyway, but if that doesn't happen I agree we should fix this. An LWG defect report would be the correct avenue.

danieljames:
OK, thanks. I'll submit one soon.

jwakely:
Great, thanks. You could mention in the issue that it depends on the resolution to [LWG 2772](http://wg21.link/lwg2772).

PR #970, opened 2016-11-07
Add index entries for ones' and two's complements

jfbastien:


source/basic.tex
```diff
@@ -3899,6 +3899,8 @@
 in~\ref{conv.prom}.}
 A synonym for integral type is
 \indextext{integer~type}%
+\indextext{complement!ones'}%
+\indextext{complement!two's}%
```
tkoeppe: Why not also "signed magnitude"?

tkoeppe: Yes, since it appears that you are indexing the various possible representations of signed integers here. So it would be strange to index only some of them, don't you think?

jfbastien: Should the index entry not be "complex" then? Maybe change it to "Signed integer representation", and have sub-entries "Ones' complement", "Two's complement", and "Signed magnitude"?

Are there other places where "signed magnitude" should be indexed, or other representations that should be indexed?

jfbastien:
I updated this patch as suggested.

tkoeppe:
Thanks, that looks nice.

tkoeppe:
@AlisdairM: What's your opinion as the index expert?

source/atomics.tex
```diff
@@ -1148,6 +1148,7 @@
 \pnum
 \returns Atomically, the value pointed to by \tcode{object} or by \tcode{this} immediately before the effects.
 
+\indextext{signed integer representation!two's complement}%
```
tkoeppe: Actually, I think the index entries are better placed *after* the paragraph:

    \pnum
    \indextext{...}%
    Real text starts here

jfbastien: Done.

AlisdairM: Looks good to me :)

tkoeppe:
Thanks!

PR #971, opened 2016-11-09
OK, error, ambiguous consistency of comments in examples.

vgvassilev:
Current stats:
```bash
git grep '// OK,' | wc -l
      60
git grep '// OK:' | wc -l
     246
git grep '// error,' | wc -l
      22
git grep '// error:' | wc -l
     257
git grep '// Error' | wc -l
       7
```

This PR introduces:
```
  // OK, -> // OK:
  // Error -> // error
  // error[,] -> // error:
  // ambiguous[,] -> ambiguous:
  // error[:] ambiguous[,]-> error: ambiguous:
```

If you consider this as an improvement please merge.

### DIFF
--- a/source/access.tex
+++ b/source/access.tex
@@ -88,8 +88,8 @@ public:
 };
 
 void f() {
-  A::BB x;          // OK, typedef name \tcode{A::BB} is public
-  A::B y;           // access error, \tcode{A::B} is private
+  A::BB x;          // OK: typedef name \tcode{A::BB} is public
+  A::B y;           // access error: \tcode{A::B} is private
 }
 \end{codeblock}
 \end{note}
@@ -187,7 +187,7 @@ protected:
 template <class U, class V = typename U::TT>
 class D : public U { };
 
-D <C<B> >* d;       // access error, C::TT is protected
+D <C<B> >* d;       // access error: C::TT is protected
 \end{codeblock}
 \end{example}
 
@@ -657,7 +657,7 @@ class X {
 };
 
 class Y {
-  int v[X::a];      // OK, \tcode{Y} is a friend of \tcode{X}
+  int v[X::a];      // OK: \tcode{Y} is a friend of \tcode{X}
 };
 
 class Z {
@@ -840,15 +840,15 @@ void f() {
   class Y;
   extern void b();
   class A {
-  friend class X;   // OK, but \tcode{X} is a local class, not \tcode{::X}
+  friend class X;   // OK: but \tcode{X} is a local class, not \tcode{::X}
   friend class Y;   // OK
-  friend class Z;   // OK, introduces local class \tcode{Z}
-  friend void a();  // error, \tcode{::a} is not considered
+  friend class Z;   // OK: introduces local class \tcode{Z}
+  friend void a();  // error: \tcode{::a} is not considered
   friend void b();  // OK
   friend void c();  // error
   };
-  X* px;            // OK, but \tcode{::X} is found
-  Z* pz;            // error, no \tcode{Z} is found
+  X* px;            // OK: but \tcode{::X} is found
+  Z* pz;            // error: no \tcode{Z} is found
 }
 \end{codeblock}
 \end{example}
--- a/source/access.tex
+++ b/source/access.tex
@@ -840,14 +840,14 @@ void f() {
   class Y;
   extern void b();
   class A {
-  friend class X;   // OK: but \tcode{X} is a local class, not \tcode{::X}
+  friend class X;   // OK, but \tcode{X} is a local class, not \tcode{::X}
   friend class Y;   // OK
   friend class Z;   // OK: introduces local class \tcode{Z}
   friend void a();  // error: \tcode{::a} is not considered
   friend void b();  // OK
   friend void c();  // error
   };
-  X* px;            // OK: but \tcode{::X} is found
+  X* px;            // OK, but \tcode{::X} is found
   Z* pz;            // error: no \tcode{Z} is found
 }
 \end{codeblock}
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3594,7 +3594,7 @@ int  arr[10];                   // now the type of \tcode{arr} is complete
 
 X x;
 void bar() {
-  xp = &x;                      // OK; type is ``pointer to \tcode{X}''
+  xp = &x;                      // OK: type is ``pointer to \tcode{X}''
   arrp = &arr;                  // ill-formed: different types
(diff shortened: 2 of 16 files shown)


source/access.tex
```diff
@@ -840,15 +840,15 @@
   class Y;
   extern void b();
   class A {
-  friend class X;   // OK, but \tcode{X} is a local class, not \tcode{::X}
+  friend class X;   // OK: but \tcode{X} is a local class, not \tcode{::X}
```
zygoloid: This one should stay as a comma

source/access.tex
```diff
@@ -840,15 +840,15 @@
...
-  friend class Z;   // OK, introduces local class \tcode{Z}
-  friend void a();  // error, \tcode{::a} is not considered
+  friend class Z;   // OK: introduces local class \tcode{Z}
+  friend void a();  // error: \tcode{::a} is not considered
   friend void b();  // OK
   friend void c();  // error
   };
-  X* px;            // OK, but \tcode{::X} is found
-  Z* pz;            // error, no \tcode{Z} is found
+  X* px;            // OK: but \tcode{::X} is found
```
zygoloid: This one too.

source/basic.tex
```diff
@@ -945,7 +945,7 @@
 
 typedef int I;
 class D {
-  typedef I I;                      // error, even though no reordering involved
+  typedef I I;                      // error: even though no reordering involved
```
zygoloid: This one should stay as-is.

source/declarations.tex
```diff
@@ -905,7 +905,7 @@
 \begin{codeblock}
 constexpr int f(bool b)
   { return b ? throw 0 : 0; }               // OK
-constexpr int f() { return f(true); }       // ill-formed, no diagnostic required
+constexpr int f() { return f(true); }       // ill-formed: no diagnostic required
```
zygoloid: This one should stay as-is, and likewise for other "ill-formed, no diagnostic required" cases.

source/declarations.tex
```diff
@@ -1683,8 +1683,8 @@
 body.
 \begin{example}
 \begin{codeblock}
-auto  f() { } // OK, return type is \tcode{void}
-auto* g() { } // error, cannot deduce \tcode{auto*} from \tcode{void()}
+auto  f() { } // OK: return type is \tcode{void}
```
zygoloid: This one should stay as a comma. (We're not saying it's OK because the return type is `void`, we're saying both (a) it's OK, and (b) the return type is `void`.)

source/declarators.tex
```diff
@@ -620,7 +620,7 @@
 or allow it to be changed through a cv-unqualified pointer later, for example:
 
 \begin{codeblock}
-*ppc = &ci;         // OK, but would make \tcode{p} point to \tcode{ci} ...
+*ppc = &ci;         // OK: but would make \tcode{p} point to \tcode{ci} ...
```
zygoloid: This one should stay as a comma.

source/declarators.tex
```diff
@@ -2271,7 +2271,7 @@
 
 \begin{codeblock}
 struct onlydouble {
-  onlydouble() = delete;              // OK, but redundant
+  onlydouble() = delete;              // OK: but redundant
```
zygoloid: This one should be left alone.

source/declarators.tex
```diff
@@ -2706,7 +2706,7 @@
 \begin{codeblock}
   int f(bool b) {
     unsigned char c;
-    unsigned char d = c; // OK, \tcode{d} has an indeterminate value
+    unsigned char d = c; // OK: \tcode{d} has an indeterminate value
```
zygoloid: This one should probably be left alone, since it's really saying "OK, but `d` has an indeterminate value"

vgvassilev:
That makes sense to me. Is the list of comments what should be reverted exhaustive?

tkoeppe:
Here's some tangential guidance: We all agreed that `error:` is the right way to state errors, because the cause of the error is something objective. If you like, you could factor out a change that only makes the error comments consistent.

By contrast, "OK" doesn't generally have a natural explanation other than "it's OK because it follows the rules of the standard". Any additional explanation that you choose to call out is subjective. So we're actually fine with `OK, ..." in general, but we can discuss individual change suggestions. But expect that part of the change to take more work.

tkoeppe:
@vgvassilev: ping - this PR needs rebasing.

vgvassilev:
Rebased, addressing @zygoloid's comments.

vgvassilev:
@tkoeppe thanks for the comments. Who is "we"? I am happy to remove the OK part of the PR if requested. I feel some of these substitutions are valid.

tkoeppe:
@vgvassilev: Please rebase and rescan the draft for new occurrences of the style you're changing.

vgvassilev:
@tkoeppe Done. It seems somebody did push -f and rewrote the history of the master, sigh...

There are a few occurrences of `ill-formed;` which I want to change to `ill-formed:` however I don't feel I should be piling more things on this PR.

tkoeppe:
@vgvassilev: I'm sorry. We do that *extremely* rarely, and never after more than five or so minutes after a push. Very sorry if you got caught in a force-push.

tkoeppe:
@zygoloid: I think you liked this PR in general. Could you review the updated situation?

source/compatibility.tex
```diff
@@ -1578,12 +1578,12 @@
...
 struct base {
   friend struct derived;
-private: 
+private:
   base();
 };
 struct derived : base {};
 
-derived d1{};       // Error. The code was well-formed before.
+derived d1{};       // error: the code was well-formed before.
```
jwakely: This change looks wrong.

jwakely:
I don't think any of the `s/OK,/OK:/` changes are improvements. Most of the capitalization changes and removing trailing whitespace changes are fine. The `s/error,/error:/` changes are generally OK in isolation, but when they occur in a group that contains both "error," and "OK," changing only the errors to use colons looks visually inconsistent (as stated above, the "OK" ones usually shouldn't have colons).

Overall, I think this is an attempt to impose an unnecessarily rigid rule on free-text comments, which don't need to be globally consistent to some fixed rule. Local consistency and clarity of meaning is probably more important. As the current patch actually changes meaning in some cases and decreases legibility in other cases, I think it would make the standard worse, not better.

Individual pieces of this change would be fine where they improve clarity, but I don't like trying to address every instance at once just in the name of consistency.

jensmaurer:
See #5080 for further cleanup for "OK"

PR #972, opened 2016-11-09
Editorial: execpol.vec was renamed to unsequenced

jfbastien:
Update missed updates.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -18319,7 +18319,7 @@ namespace std::execution {
   // \ref{execpol.par}, parallel execution policy:
   class parallel_policy;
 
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -18386,7 +18386,7 @@
 a parallel algorithm's execution may be parallelized.
 \end{itemdescr}
 
-\rSec2[execpol.vec]{Parallel+Vector execution policy}
+\rSec2[execpol.parunseq]{Parallel+Unsequenced execution policy}
```
tkoeppe: I appreciate that the original wording already does this, but this is typographically gross. Can we do better here, without  a textual `+`?

PR #976, opened 2016-11-10
Editorial: remove '+' from title

jfbastien:


jfbastien:
As promised in #972.

tkoeppe:
Thanks! Much better.

PR #977, opened 2016-11-10
[whole draft] Use \tcode{true} and \tcode{false} consistently

jlaire:


source/regex.tex
```diff
@@ -3113,7 +3113,7 @@
 \\ \rowsep
 \tcode{m.suffix().matched}
 &
-false 
+\tcode{false}
 \\ \rowsep
 \tcode{m[0].first}
 &
```
zygoloid: Can you apply `\tcode` to all the `first`s and `last`s in this table too?

zygoloid:
Looks good other than the couple of issues noted above.

source/locales.tex
```diff
@@ -2945,7 +2945,7 @@
 function determines the integral conversion specifier as indicated in
 Table~\ref{tab:localization.integer.conversions.in}.
 The table is ordered.
-That is, the first line whose condition is true applies.
+That is, the first line whose condition is \tcode{true} applies.
```
jlaire: True. I'll just leave this unchanged.

PR #980, opened 2016-11-11
Use \term and \defnx more correctly and consistently

Eelis:
No visual difference in the PDF.

### DIFF
--- a/source/atomics.tex
+++ b/source/atomics.tex
@@ -234,10 +234,10 @@ given atomic access to a particular atomic object be indivisible with respect
 to all other atomic accesses to that object. \end{note}
 
 \pnum
-An atomic operation \term{A} that performs a release operation on an atomic
-object \term{M} synchronizes with an atomic operation \term{B} that performs
-an acquire operation on \term{M} and takes its value from any side effect in the
-release sequence headed by \term{A}.
+An atomic operation \placeholder{A} that performs a release operation on an atomic
+object \placeholder{M} synchronizes with an atomic operation \placeholder{B} that performs
+an acquire operation on \placeholder{M} and takes its value from any side effect in the
+release sequence headed by \placeholder{A}.
 
 \pnum
 There shall be a single total order \textit{S} on all \tcode{memory_order_seq_cst}
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -144,9 +144,8 @@ at least one of which is not deleted.
 virtual functions or virtual base classes.\end{note}
 
 \indextext{class!standard-layout}%
-\indextext{standard-layout~class}%
 \pnum
-A class \tcode{S} is a \grammarterm{standard-layout class} if it:
+A class \tcode{S} is a \defnx{standard-layout class}{standard-layout~class} if it:
 \begin{itemize}
 \item has no non-static data members of type non-standard-layout class
 (or array of such types) or reference,
@@ -211,14 +210,12 @@ in \tcode{X}. \end{note}
 \end{example}
 
 \indextext{struct!standard-layout}%
-\indextext{standard-layout~struct}%
 \indextext{union!standard-layout}%
-\indextext{standard-layout~union}%
 \pnum
-A \grammarterm{standard-layout struct} is a standard-layout class
+A \defnx{standard-layout struct}{standard-layout~struct} is a standard-layout class
 defined with the \grammarterm{class-key} \tcode{struct} or the
 \grammarterm{class-key} \tcode{class}.
-A \grammarterm{standard-layout union} is a standard-layout class
+A \defnx{standard-layout union}{standard-layout~union} is a standard-layout class
 defined with the
 \grammarterm{class-key} \tcode{union}.
 
@@ -228,14 +225,13 @@ code written in other programming languages. Their layout is specified
 in~\ref{class.mem}.\end{note}
 
 \pnum
-\indextext{POD struct}\indextext{POD class}\indextext{POD union}%
-A \term{POD struct}\footnote{The acronym POD stands for ``plain old data''.}
+A \defn{POD struct}\footnote{The acronym POD stands for ``plain old data''.}
 is a non-union class that is both a trivial class and a
 standard-layout class, and has no non-static data members of type non-POD struct,
 non-POD union (or array of such types). Similarly, a
-\term{POD union} is a union that is both a trivial class and a standard-layout
+\defn{POD union} is a union that is both a trivial class and a standard-layout
 class, and has no non-static data members of type non-POD struct, non-POD
-union (or array of such types). A \term{POD class} is a
+union (or array of such types). A \defn{POD class} is a
 class that is either a POD struct or a POD union.
 
 \begin{example}
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -1498,7 +1498,7 @@ Ordinary string literals and UTF-8 string literals are
 also referred to as narrow
 string literals. A narrow string literal has type
 \indextext{literal!string!type~of}%
-``array of \term{n} \tcode{const char}'', where \term{n} is the size of
+``array of \placeholder{n} \tcode{const char}'', where \placeholder{n} is the size of
 the string as defined below, and has static storage
 duration~(\ref{basic.stc}).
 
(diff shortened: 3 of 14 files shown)


Eelis:
Please hold, spotted a few more

Eelis:
Added a commit that also fixes abuse of \term for things that are not terms at all.

Only visual difference now is erasure of some inappropriate trailing whitespace resulting from the fact that in addition to making text italic, \term also includes a trailing \xspace.

zygoloid:
Looks good other than that we would prefer the more semantic `\placeholder` be used rather than `\textit` in the incorrect uses of `\term` for placeholders.

Eelis:
Ah, thanks for the review, will fix (and rebase).

Eelis:
Fixed and rebased. (Still no visual difference apart from the trailing whitespace.)

Eelis:
Bleh, missed a few, so not completely fixed yet

Eelis:
Added fixes for a whole bunch more occurrences

source/atomics.tex
```diff
@@ -238,71 +238,71 @@
...
 \tcode{memory_order_seq_cst}.
 \end{itemize}
 
-\begin{note} Although it is not explicitly required that \textit{S} include locks, it can
+\begin{note} Although it is not explicitly required that \placeholder{S} include locks, it can
 always be extended to an order that does include lock and unlock operations, since the
 ordering between those is already included in the ``happens before'' ordering. \end{note}
 
 \pnum
-For an atomic operation \textit{B} that reads the value of an atomic object \textit{M},
```
Eelis: Ah no worries, I'm glad to hear there's coherent plans being formed for the use of all these things going forward. :)  I'll split out the \term stuff tonight, cheers!

jensmaurer:
Sorry, hadn't noticed this before. #1061 and #1062 as well as #1060 might conflict.

Eelis:
All right, I've split it up more, and now it only uses \placeholder to replace bad uses of \term.

tkoeppe:
@jensmaurer: As the standing expert on all things `\term`, could you please review this PR?

source/lex.tex
```diff
@@ -1726,93 +1726,93 @@
...
-operator "" @\term{X}@(@\term{f}@L)
+operator "" @\placeholder{X}@(@\placeholder{f}@L)
 \end{codeblock}
 
-Otherwise, \term{S} shall contain a raw literal operator or a literal operator
-template~(\ref{over.literal}) but not both. If \term{S} contains a raw literal operator,
-the \term{literal} \term{L} is treated as a call of the form
+Otherwise, \placeholder{S} shall contain a raw literal operator or a literal operator
+template~(\ref{over.literal}) but not both. If \placeholder{S} contains a raw literal operator,
+the \placeholder{literal} \placeholder{L} is treated as a call of the form
```
jensmaurer: \term{literal} should be \grammarterm{literal}, not \placeholder

source/lib-intro.tex
```diff
@@ -604,20 +604,20 @@
...
 
-static const @\term{enumerated C0}@ (@\term{V0}@);
-static const @\term{enumerated C1}@ (@\term{V1}@);
-static const @\term{enumerated C2}@ (@\term{V2}@);
-static const @\term{enumerated C3}@ (@\term{V3}@);
+static const @\placeholder{enumerated C0}@ (@\placeholder{V0}@);
+static const @\placeholder{enumerated C1}@ (@\placeholder{V1}@);
+static const @\placeholder{enumerated C2}@ (@\placeholder{V2}@);
+static const @\placeholder{enumerated C3}@ (@\placeholder{V3}@);
   .....
```
jensmaurer: Those are two distinct placeholders. Please try separating them (i.e. have two \placeholder's next to each other). Does that impair the spacing?

source/lib-intro.tex
```diff
@@ -637,56 +637,56 @@
...
 
-constexpr @\term{bitmask C0}@(@\term{V0}{}@);
-constexpr @\term{bitmask C1}@(@\term{V1}{}@);
-constexpr @\term{bitmask C2}@(@\term{V2}{}@);
-constexpr @\term{bitmask C3}@(@\term{V3}{}@);
+constexpr @\placeholder{bitmask C0}@(@\placeholder{V0}{}@);
+constexpr @\placeholder{bitmask C1}@(@\placeholder{V1}{}@);
+constexpr @\placeholder{bitmask C2}@(@\placeholder{V2}{}@);
+constexpr @\placeholder{bitmask C3}@(@\placeholder{V3}{}@);
   .....
```
jensmaurer: Same here: two \placeholders, once for each meta-variable.

source/regex.tex
```diff
@@ -3916,7 +3916,7 @@
 sequence \end{note}
 
 \pnum
-The \textit{current match} is \tcode{(*position).prefix()} if \tcode{subs[N] == -1}, or
+The \term{current match} is \tcode{(*position).prefix()} if \tcode{subs[N] == -1}, or
 \tcode{(*position)[subs[N]]} for any other value of \tcode{subs[N]}.
 
 \rSec3[re.tokiter.cnstr]{\tcode{regex_token_iterator} constructors}
```
jensmaurer: This patch adds \term, which this issue tries to eradicate (admittedly morphed from even worse \textit). Please drop this patch from here for now; I expect those \textit to mostly become \defn or \defnx.

source/classes.tex
```diff
@@ -228,14 +225,13 @@
...
 standard-layout class, and has no non-static data members of type non-POD struct,
 non-POD union (or array of such types). Similarly, a
-\term{POD union} is a union that is both a trivial class and a standard-layout
+\defn{POD union} is a union that is both a trivial class and a standard-layout
 class, and has no non-static data members of type non-POD struct, non-POD
-union (or array of such types). A \term{POD class} is a
+union (or array of such types). A \defn{POD class} is a
 class that is either a POD struct or a POD union.
 
 \begin{example}
```
jensmaurer: Do we have any rules when to use "POD~struct" vs. "POD struct" in the index? We seem to prefer the variant with the tilde = non-breaking space.  If true, the stuff above needs to become \defnx.

source/lex.tex
```diff
@@ -870,26 +870,21 @@
...
+An \defnx{octal}{literal!octal} integer
 literal (base eight) begins with the digit \tcode{0} and consists of a
 sequence of octal digits.\footnote{The digits \tcode{8} and \tcode{9} are not octal digits. }
-A \term{decimal}
+A \defnx{decimal}{literal!decimal}
 integer literal (base ten) begins with a digit other than \tcode{0} and
 consists of a sequence of decimal digits.
-A \term{hexadecimal} integer literal (base sixteen) begins with
+A \defnx{hexadecimal}{literal!hexadecimal} integer literal (base sixteen) begins with
 \tcode{0x} or \tcode{0X} and consists of a sequence of hexadecimal
```
jensmaurer: Please put all of "XXX integer literal" into the first argument of \defnx; we're not actually defining the term "decimal" here, but the phrase "decimal integer literal".

source/regex.tex
Eelis: Ah ok, I didn't know there was a plan to eradicate \term.

source/utilities.tex
```diff
@@ -12224,30 +12224,24 @@
 The following definitions apply to this Clause:
 
 \pnum
-\indexdefn{call signature}%
-A \term{call signature} is the name of a return type followed by a
+A \defn{call signature} is the name of a return type followed by a
 parenthesized comma-separated list of zero or more argument types.
```
jensmaurer: For this (and those that follow), the index entries should probably use tilde instead of space.

jensmaurer:
Review complete; see comments above. Going to bed now.

source/classes.tex
Eelis: (If that's an actual goal, though, there are hundreds more occurrences that would need to be replaced.)

Eelis:
Hm, this pull request has really lost focus. I originally just wanted to use \defn* for \terms that already had index entries, a change which I could make without affecting those index entries or the formatting of the text, but I don't feel qualified to make the kind of changes to the index entries themselves (by changing their use of tildes) or to the formatting of the text (by changing which parts of the text are italicized) that Jens is requesting, so probably better to just go with his patches. :)

PR #997, opened 2016-11-12
NB JP-21 (C++17 CD): [algorithms] Order elements consistently

tkoeppe:


tkoeppe:
@AlisdairM, @jwakely: I'd appreciate your review of this change, since it is pretty large and repetitive and details may have slipped by.

AlisdairM:
nth_element has what should be an Effects: paragraph (but is not labeled) preceding a Requires: paragraph.

The set operations algorithms all lead with a requirement that the resulting range does not overlap with either of the input sequences, although the length of that range is not obvious until the end of the algorithm description.  In context of the whole algorithm it can be understood, but is a minor step back in clarity.  I wonder if this requirement can be pushed into the front matter where we say that these algorithms require two sorted sequences?

Finally, most algorithms in clause 26.8 have a similar problem with disordered Requires: paragraphs.

After seeing the edits, this seems like a step forward, but also highlights a few places I would like to go back and be a bit more careful again, especially with some of the other disordered paragraphs.

tkoeppe:
Thanks!

I added the `\effects` for `nth_element`. Clause 26.8 is not mentioned by the NB comment, so we don't need to change it at this point. I think the set operations are understandable with and without this change, so I don't think we need to modify the proposed change at this point.

PR #1000, opened 2016-11-12
[lib] Make Remark: and Postcondition: plural

### LINKED ISSUE #282: Replace "Remark", "Postcondition", "Precondition" with their plural equivalents
These have special meaning and there's no reason to have the singular case be different; it only causes more editing thrash when these go from single cases to multiple cases.

It would be good to look for any other singular forms of special library requirements paragraph introductions.

jensmaurer:
Fixes #282.


tkoeppe:
I stole and rebased these commits. Thanks!

PR #1004, opened 2016-11-12
[diff.cpp11.lex] fix example for digit separators

### LINKED ISSUE #306: Invalid example in [diff.cpp11.lex]
[diff.cpp11.lex] has this example:

```
#define M(x, ...) __VA_ARGS__
int x[2] = { M(1’2,3’4) };
// int x[2] = {};      — C ++ 2011
// int x[2] = { 3’4 }; — this International Standard
```

However, that isn't valid C++11, [cpp.replace]/4 says that for a variadic macro:

>  there shall be more arguments in the invocation than there are parameters in the macro definition (excluding the `...`).

jensmaurer:
Fixes #306.

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -1354,9 +1354,9 @@ separators in this International Standard:
 
 \begin{codeblock}
 #define M(x, ...) __VA_ARGS__
-int x[2] = { M(1'2,3'4) };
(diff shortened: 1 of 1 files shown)


tkoeppe:
Ironically, 306 is also the number of the paper that will fix the original problem :-)

PR #1012, opened 2016-11-13
[containers] 'const iterator' should be 'constant iterator'

### LINKED ISSUE #386: "const iterator" and "constant iterator" used interchangeably
The standard formally defines "constant iterator", but then several places refer to them instead by the name "const iterator".

jensmaurer:
Fixes #386.


tkoeppe:
@AlisdairM: Could you please review this change?

AlisdairM:
Looks good to me

PR #1018, opened 2016-11-13
[vector.bool] Excise use of undefined term 'conversion operator'

### LINKED ISSUE #444: "conversion operator" denotes functions or operators?
[vector.bool] §23.3.7/4 refers to `std::vector<bool>::reference::operator bool` as a conversion operator, whereas it's a conversion function. The index indicates that "conversion operator" is the same as "conversion, user-defined." However [expr.cast] §5.4/2 uses the term to refer to static_cast, const_cast, dynamic_cast and reinterpret_cast, which seems more appropriate.

jensmaurer:
Fixes #444.

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -1289,7 +1289,7 @@ binary representation of the required template specializations of
 \ref{istream::sentry},
 \ref{ostream::sentry},
 \ref{iostate.flags}
-\change Specify use of explicit in existing boolean conversion operators
+\change Specify use of explicit in existing boolean conversion functions
 \rationale Clarify intentions, avoid workarounds.
 \effect
 Valid \CppIII code that relies on implicit boolean conversions will fail to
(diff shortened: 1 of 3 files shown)


source/compatibility.tex
```diff
@@ -1289,7 +1289,7 @@
 \ref{istream::sentry},
 \ref{ostream::sentry},
 \ref{iostate.flags}
-\change Specify use of explicit in existing boolean conversion operators
+\change Specify use of explicit in existing boolean conversion functions
```
tkoeppe: Should "explicit" appear in code font, since it's actually a keyword?

jensmaurer: Yes, but I shouldn't make that change here, which deals with "conversion operator" vs. "conversion function".

I've created #1019 to deal with the "explicit" code font issue.

tkoeppe: OK, sure, but I would have had no problem with a minor drive-by fix like that if this is a single, isolated problem. You can stick it into a separate commit in the same pull request, and I could retain the separate commits (i.e. not squash them), so we still get nicely atomic commits.

PR #1022, opened 2016-11-13
Ensure PDF files are not compared with text-based diff

### LINKED ISSUE #484: Forcing git to consider PDFs as binary files
It seems that git is incorrectly considering some PDFs as text files, e.g. `papers/n4431.pdf`. One can force git to consider files with certain endings to be binary using the `.gitattributes` file. According to the [docs](http://git-scm.com/docs/gitattributes#_marking_files_as_binary), the line

```
*.pdf -diff
```

should solve the issue.

jensmaurer:
Fixes #484.


tkoeppe:
Please double-check that this is really a problem that needs to be solved. See discussion in the Issue. Re-open this PR if you think we should really do this.

PR #1025, opened 2016-11-13
Harmonize formatting of 'cv-qualified'

### LINKED ISSUE #798: Inconsistent formatting of cv-qualified
The core and library clauses oscillate between cv-qualified, _cv_-qualified and (in one case) _cv-qualified_.

The library is also inconsistent about parenthesising "possibly cv-qualified".

jensmaurer:
Fixes #798.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -4146,7 +4146,7 @@ A type mentioned in~\ref{basic.fundamental} and~\ref{basic.compound} is
 a \term{cv-unqualified type}. Each type which is a
 cv-unqualified complete or incomplete object type or is
 \tcode{void}~(\ref{basic.types}) has three corresponding cv-qualified
-versions of its type: a \grammarterm{const-qualified} version, a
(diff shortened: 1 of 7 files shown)


source/basic.tex
```diff
@@ -4146,7 +4146,7 @@
 a \term{cv-unqualified type}. Each type which is a
 cv-unqualified complete or incomplete object type or is
 \tcode{void}~(\ref{basic.types}) has three corresponding cv-qualified
-versions of its type: a \grammarterm{const-qualified} version, a
+versions of its type: a \term{const-qualified} version, a
```
tkoeppe: Do we have an index entry for this? Should this be `\defnx` instead of `\term`?

jensmaurer: We don't have an index entry for "const-qualified".
All the other terms defined in that same sentence are just \term{volatile-qualified}, for example, so I just wanted to fix this particular bump.
I suggest to go forward with this (it's an obvious improvement).

PR #1034, opened 2016-11-14
[containers] Consistent semicolons in tables

jlaire:
The tables that list requirements contain both expressions and variable definitions. Most of the standard is consistent in using semicolons after the variable definitions, but not after expressions. This patch fixes the deviations from that.


tkoeppe:
@jwakely: Could you please review this PR?

tkoeppe:
Never mind, I think this is a good change.

PR #1037, opened 2016-11-14
[containers] Fix typos

jlaire:


tkoeppe:
Please retain only containers.tex. I'm fixing the other issues by other means.

jlaire:
Done.

source/containers.tex
```diff
@@ -1989,7 +1989,7 @@
 For associative containers with unique keys the stronger condition holds,
```
tkoeppe: Change to "holds:" (colon), and add to the previous paragraph: "...is positive, the following condition holds:"

tkoeppe:
Thanks, looks good, just one small comment.

source/containers.tex
jlaire: Done.

PR #1038, opened 2016-11-14
[atomics] Clean up indexing of atomics clause

AlisdairM:
Reverse class/member names in indexlibrarymember macros to be
consistent with the prefered style in other clauses.

Expand in the index several functions that are documented as a
pattern-match, such as fetch_add and fetch_sub.

Use 'atomic<type>' rather than 'atomic type' for index references
to the atomic template.  This allows for a more intuitive indexing
of the conversion operator.

### DIFF
--- a/source/atomics.tex
+++ b/source/atomics.tex
@@ -379,6 +379,16 @@ value~(\ref{intro.multithread}).
 
 \rSec1[atomics.lockfree]{Lock-free property}
 
+\indexlibrary{\idxcode{ATOMIC_BOOL_LOCK_FREE}}%
+\indexlibrary{\idxcode{ATOMIC_CHAR_LOCK_FREE}}%
+\indexlibrary{\idxcode{ATOMIC_CHAR16_T_LOCK_FREE}}%
+\indexlibrary{\idxcode{ATOMIC_CHAR32_T_LOCK_FREE}}%
+\indexlibrary{\idxcode{ATOMIC_WCHAR_T_LOCK_FREE}}%
+\indexlibrary{\idxcode{ATOMIC_SHORT_LOCK_FREE}}%
+\indexlibrary{\idxcode{ATOMIC_INT_LOCK_FREE}}%
+\indexlibrary{\idxcode{ATOMIC_LONG_LOCK_FREE}}%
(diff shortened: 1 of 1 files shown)


source/atomics.tex
```diff
@@ -427,6 +427,7 @@
 
 \rSec1[atomics.types.generic]{Atomic types}
 
+\indexlibrary{\idxcode{atomic<\placeholder{type}>}}%
```
zygoloid: For the primary template, we just use `\indexlibrary{\idxcode{atomic}}%`

zygoloid: Maybe index entries for `atomic<`*`integral`*`>` and `atomic<T*>` might make sense, though.

source/atomics.tex
```diff
@@ -763,7 +808,7 @@
 operations on non-volatile objects become volatile. Thus, volatile qualified operations
 on non-volatile objects may be merged under some conditions. \end{note}
 
-\indexlibrary{\idxcode{atomic type}!constructor}%
+\indexlibrary{\idxcode{atomic<\placeholder{type}>}!constructor}%
```
zygoloid: Likewise here, just `\indexlibrary{\idxcode{atomic}!constructor}%` would be in line with what we do elsewhere.

tkoeppe:
Please also fix the commit message to refer to the clause's stable name.

source/atomics.tex
AlisdairM: Replaced all uses of atomic<type> with just atomic, where appropriate to that type. Added fresh entries for atomic<integral> and atomic<T*>, to better index their larger interfaces. Then removed any remaining references to atomic<type>.

AlisdairM:
Rebased onto current master and pushed again, with updated headline as requested by @tkoeppe.

This patch has some additional index entries that I had missed on the first pass, and makes some (but not exhaustive) tweaks in the main text to use \placeholder for consistency with the indexed use.

It is probably worth a fresh review, rather than just comparing to changes in the last one.

tkoeppe:
Are you still around for a final rebase?

AlisdairM:
Rebased and pushed.

PR #1044, opened 2016-11-14
[container.requirements.general] Remove redundant Requires

timsong-cpp:
The allocator requirements already require move construction to not throw, so there's no need to repeat it here.


tkoeppe:
@jwakely: As the allocator guy, could you please review this PR?

PR #1045, opened 2016-11-14
[cpp.replace] Delete obsolete footnote

timsong-cpp:
non-directive is not a thing anymore (renamed to conditionally-supported-directive by CWG 2001) and the new name makes it clear that it's a directive.

### DIFF
--- a/source/preprocessor.tex
+++ b/source/preprocessor.tex
@@ -732,7 +732,7 @@ are separated by comma preprocessing tokens,
 but comma preprocessing tokens between matching
 inner parentheses do not separate arguments.
 If there are sequences of preprocessing tokens within the list of
-arguments that would otherwise act as preprocessing directives,\footnote{Despite the name, a non-directive is a preprocessing directive.}
+arguments that would otherwise act as preprocessing directives,\footnote{A \grammarterm{conditionally-supported-directive} is a preprocessing directive regardless of whether the implementation supports it.}
 the behavior is undefined.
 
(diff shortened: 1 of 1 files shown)


tkoeppe:
@hubert-reinterpretcast: Could you please review this PR?

hubert-reinterpretcast:
The footnote is indeed obsolete in its clarification on the terminology.

I believe a note remains necessary to clarify that tokens that would otherwise be instances of unsupported _conditionally-supported-directive_ still "otherwise act as preprocessing directives" (because they would have required a diagnostic). In the alternative, implementations should not fail compilation on what may be interpreted as an unsupported _conditionally-supported-directive_ within the list of arguments; they would either need to treat the tokens in question as "plain" tokens, or (to fail compilation) they would instead have to document support for all forms of _conditionally-supported-directive_ which they do not intend to support as having semantics amounting to "causes the program to be ill-formed". @zygoloid: comments?

zygoloid:
Hubert, would you be looking for an alternative footnote such as "A *conditionally-supported-directive* is a directive regardless of whether the implementation supports it." ?

I think that makes sense given the context ("[a sequence of tokens] that would act as a preprocessing directive").

hubert-reinterpretcast:
Yes, something like the suggested alternative footnote would be fine.

timsong-cpp:
Force-pushed a new version that replaces the footnote with @zygoloid's wording.

PR #1046, opened 2016-11-14
[clause 21] Index cstd header synopses

AlisdairM:
Adds entries to the library index for every cstd... header
in the strings clause, pointing to their synopsis.


tkoeppe:
Please use stable names to refer to clauses and subclauses in the future. "21" is not very meaningful.

PR #1049, opened 2016-11-15
[cstdlib.syn, csignal.syn] Introduce exposition-only function types with C and C++ language linkage to improve the presentation of C library functions that take callbacks.

### LINKED ISSUE #1002: Improve synopsis presentation for function pointers with different language linkages
We should clean up the synopses for `at_exit`, `qsort`, `signal` and such like to be clear that the language linkage of the function itself is not specified, but rather that we have overloads for the callback function parameter only. This may require some acrobatics, such as:

    extern "C" using PredC = int(const void*, const void*);  // exposition only
    extern "C++" using PredCXX = int(const void*, const void*);  // exposition only
    void qsort(void*, size_t, size_t, PredC*);
    void qsort(void*, size_t, size_t, PredCXX*);

tkoeppe:
This change improves correctness, since we never meant to specify the language linkage of the function names, but only of the callback parameter type.

Fixes #1002.

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -581,6 +581,23 @@ Certain types defined in Clause~\ref{input.output} are used to describe implemen
 \indextext{types!implementation-defined}%
 They are based on other types, but with added constraints.
 
+\rSec4[expos.only.types]{Exposition-only types}
+
+\pnum
+Several types defined in Clauses~\ref{\firstlibchapter} through~\ref{\lastlibchapter}
+and Annex~\ref{depr} that are used as function parameter or return types are defined
+for the purpose of exposition only in order to capture their language linkage. The
+declarations of such types are followed by a comment ending in \textit{exposition only}.
+\begin{example}
+\begin{codeblock}
+namespace std {
+  extern "C" using @\placeholdernc{some-handler}@ = int(int, void*, double);  // \expos
+}
+\end{codeblock}
+The type placeholder \tcode{\placeholder{some-handler}} can now be used to specify a function
+that takes a callback parameter with C language linkage.
+\end{example}
+
 \rSec4[enumerated.types]{Enumerated types}
 
 \pnum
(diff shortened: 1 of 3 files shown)


source/lib-intro.tex
```diff
@@ -3039,12 +3039,18 @@
 #define MB_CUR_MAX @\seebelow@
 
 namespace std {
+  // Exposion-only function type aliases
```
timsong-cpp: Typo.

tkoeppe: Thanks, fixed!

source/lib-intro.tex
```diff
@@ -3039,12 +3039,18 @@
 #define MB_CUR_MAX @\seebelow@
 
 namespace std {
+  // Exposition-only function type aliases
+  extern "C" using c_exit_t = void();                                // \expos
```
zygoloid: [objects.within.classes]/2 only defines what the "// exposition only" comment means for private members of classes. This isn't the only place where we would be using it for something else, but it might be a good idea to move the definition of "exposition only" to somewhere more general, and just refer to it from [objects.within.classes]/2.

source/lib-intro.tex
```diff
@@ -3039,12 +3039,18 @@
...
+  extern "C++" using cxx_exit_t = void();                            // \expos
```
zygoloid: This is the only use of "cxx" to refer to C++ in the standard; we use "cpp" in the few places where this comes up. Perhaps we could just omit the `cxx_` prefix and maybe also the `extern "C++"`?

I would also avoid an `_t` name, since that makes it look more like this is a name exposed by the library. Maybe even avoiding a real identifier entirely, and/or using an alternative typographical convention, would help to emphasize that this is not an exposed name:

> `using `*`atexit-handler`*` = void();              // `*`exposition only`*
> `extern "C" using `*`c-atexit-handler`*` = void(); // `*`exposition only`*
> `int atexit(`*`atexit-handler`*`* func);`
> `int atexit(`*`c-atexit-handler`*`* func);`
> `int at_quick_exit(`*`atexit-handler`*`* func);`
> `int at_quick_exit(`*`c-atexit-handler`*`* func);`

... although I note that none of our existing exposition-only names use such a typographical convention.

source/lib-intro.tex
```diff
@@ -3039,12 +3039,18 @@
...
+  extern "C++" using cxx_compar_t = int(const void* , const void*);  // \expos
+
   // \ref{support.start.term}, start and termination
   [[noreturn]] void abort();
-  extern "C" int atexit(void (*func)());
-  extern "C++" int atexit(void (*func)());
-  extern "C" int at_quick_exit(void (*func)());
-  extern "C++" int at_quick_exit(void (*func)());
+  int atexit(c_exit_t* fuc);
+  int atexit(cxx_exit_t* fuc);
```
zygoloid: Looks like you accidentally lost an 'n' here.

source/lib-intro.tex
tkoeppe: I removed `cxx_` and renamed the typedefs to `_fn`.

I like the placeholder style, but take note that we we will need to refer to that name in the itemdescls as well, which are quite far way (see updated PR).

tkoeppe: I retained the `extern "C++"`, because it makes the columns line up nicely :-S

tkoeppe:
image

source/lib-intro.tex
tkoeppe: Added a new subclause to define exposition-only types.

source/lib-intro.tex
tkoeppe: I implemented the placeholder types.

Issue #1006, opened 2016-11-13
Clean up C.5


tkoeppe:
We should resolve [LWG 2201](http://wg21.link/lwg2201) editorially.

I propose deleting the tables of names, now that P0175 added full synopses. We also need to update a few cross-references from C.5 back to the main text that were forgotten in P0175.

tkoeppe:
@zygoloid recommends describing the *differences* with C.

PR #1051, opened 2016-11-16
[diff.mods.to.headers] Explain what happened to meaningless C headers

tkoeppe:
A section called "modifications to headers" should not fail to explain what happened to all the other C headers.

Part of Issue #1006.

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -2152,10 +2152,30 @@ as shown in Table~\ref{tab:diff.standard.functions}.
 \rSec2[diff.mods.to.headers]{Modifications to headers}
 
 \pnum
-For compatibility with the C standard library,
-\indextext{library!C standard}%
+For compatibility with the C standard library\indextext{library!C standard},
 the \Cpp standard library provides the C headers enumerated
 in~\ref{depr.c.headers}, but their use is deprecated in \Cpp.
+
+\pnum
+There are no \Cpp headers for the C headers
+\tcode{<stdatomic.h>}\indextext{\idxhdr{stdatomic.h}},
+\tcode{<stdnoreturn.h>}\indextext{\idxhdr{stdnoreturn.h}},
+and \tcode{<threads.h>}\indextext{\idxhdr{threads.h}},
+nor are the C headers themselves part of \Cpp.
+
+\pnum
+The headers \tcode{<ccomplex>}\indextext{\idxhdr{ccomplex}}~(\ref{ccomplex.syn})
+and \tcode{<ctgmath>}\indextext{\idxhdr{ctgmath}}~(\ref{ctgmath.syn}), as well
+as their corresponding C headers \tcode{<complex.h>}\indextext{\idxhdr{complex.h}}
+and \tcode{<tgmath.h>}\indextext{\idxhdr{tgmath.h}}, do not contain any of the
+content from the C standard library and instead merely include other headers
+from the \Cpp standard library.
+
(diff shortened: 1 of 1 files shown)


tkoeppe:
Screenshot:

image

source/compatibility.tex
```diff
@@ -2152,11 +2152,29 @@
...
 
 \pnum
-For compatibility with the C standard library,
-\indextext{library!C standard}%
+For compatibility with the C standard library\indextext{library!C standard},
 the \Cpp standard library provides the C headers enumerated
 in~\ref{depr.c.headers}, but their use is deprecated in \Cpp.
 
+\pnum
+\begin{note}
```
zygoloid: You don't need a note here, this whole clause is non-normative.

source/compatibility.tex
```diff
@@ -2152,11 +2152,29 @@
...
+nor are the C headers themselves part of \Cpp.
+The headers \tcode{<ccomplex>}\indextext{\idxhdr{ccomplex}}~(\ref{ccomplex.syn})
+and \tcode{<ctgmath>}\indextext{\idxhdr{ctgmath}}~(\ref{ctgmath.syn}), as well
+as their corresponding C headers \tcode{<complex.h>}\indextext{\idxhdr{complex.h}}
+and \tcode{<tgmath.h>}\indextext{\idxhdr{tgmath.h}}, do not contain any of the
+content from the C standard library and instead merely include other headers
+from the \Cpp standard library.
+The headers \tcode{<ciso646>}, \tcode{<cstdalign>}~(\ref{cstdalign.syn}),
+and \tcode{<cstdbool>}~(\ref{cstdbool.syn}) are meaningless in \Cpp. Use of
+the \Cpp headers \tcode{<ccomplex>}, \tcode{<cstdalign>}, \tcode{<cstdbool>},
```
zygoloid: Hmm, I wonder why `<ciso646>` is not deprecated... maybe we should file an issue for this?

source/compatibility.tex
```diff
@@ -2152,11 +2152,29 @@
...
 in~\ref{depr.c.headers}, but their use is deprecated in \Cpp.
 
+\pnum
+\begin{note}
+There are no \Cpp headers for the C headers
+\tcode{<stdatomic.h>}\indextext{\idxhdr{stdatomic.h}},
+\tcode{<stdnoreturn.h>}\indextext{\idxhdr{stdnoreturn.h}},
+and \tcode{<threads.h>}\indextext{\idxhdr{threads.h}},
+nor are the C headers themselves part of \Cpp.
+The headers \tcode{<ccomplex>}\indextext{\idxhdr{ccomplex}}~(\ref{ccomplex.syn})
```
zygoloid: New paragraph here (between "headers in C and not in C++" and "headers in C++ that are not the same as the corresponding C header").

source/compatibility.tex
tkoeppe: I just emailed lwgchair and opened an issue :-)

In fact, I wonder if we cannot *remove* the deprecated C++ headers in C++2a and move the synopses into Annex D.

source/compatibility.tex
tkoeppe: Done, and also below.

source/compatibility.tex
tkoeppe: Ah, good point, removed.

tkoeppe:
Update:

image

PR #1054, opened 2016-11-16
Move specification of deprecated headers to Annex D

tkoeppe:


tkoeppe:
@AlisdairM, what do you think of this change?

source/lib-intro.tex
```diff
@@ -1061,55 +1061,44 @@
 
 
 \pnum
-The facilities of the C standard library are provided in 26
-\indextext{library!C standard}%
-additional headers, as shown in Table~\ref{tab:cpp.c.headers}.%
+The facilities of the C standard library\indextext{library!C standard}
+are provided in 22 additional headers, as shown in Table~\ref{tab:cpp.c.headers}.%
```
AlisdairM: I thought we were trying to avoid keeping accurate counts, as they are too much effort to maintain? Prefer just 'additional headers' without the 22?

source/future.tex
```diff
@@ -134,6 +135,71 @@
...
+\begin{codeblock}
+#define @\xname{alignas_is_defined}@ 1
+\end{codeblock}
+
+\pnum
+\indexlibrary{\idxhdr{cstdalign}}%
+\indexlibrary{\idxhdr{stdalign.h}}%
+The contents of the header \tcode{<cstdalign>} are the same as the C
+standard library header \tcode{<stdalign.h>}, with the following changes:
+The header \tcode{<cstdalign>} and the header \tcode{<stdalign.h>} shall not
```
AlisdairM: My biggest concern is providing the specification for the C header in the deprecated C++ header section, and similarly for stdbool.h.  It is also deprecated to use any of the C headers by name in C++, but it still makes me more nervous about this change.

source/lib-intro.tex
tkoeppe: I don't think we have a blanket rule for this. In the present case, it was easy enough to update the number, and we've previously updated (but kept) counts of things.

source/future.tex
tkoeppe: Then we have a conundrum: both headers `<cstdbool>` and `<stdbool.h>` are deprecated, yet you want the specification to be in the main text? So the specification is not deprecated, only the use is?

source/lib-intro.tex
```diff
@@ -1295,7 +1284,7 @@
 \ref{support.rtti}  & Type identification          &   \tcode{<typeinfo>}      \\ \rowsep
 \ref{support.exception} & Exception handling      &   \tcode{<exception>}     \\ \rowsep
 \ref{support.initlist}  & Initializer lists & \tcode{<initializer_list>} \\ \rowsep
-\ref{support.runtime} & Other runtime support     &   \tcode{<cstdalign>} \tcode{<cstdarg>} \tcode{<cstdbool>}       \\ \rowsep
```
tkoeppe: Good point. We can perhaps keep the headers in the free-standing implementation by adding a suitable sentence to Annex D?

source/future.tex
AlisdairM: As I said, this was the one migration from the main text that gave me most concern - even though both uses are deprecated for different reasons.  I' would be for this patch to proceed unless we have a better idea though.

source/lib-intro.tex
AlisdairM: If everything else is now in Annex D, I would suggest the "add this to freestanding" text belongs in Annex D as well.

source/lib-intro.tex
AlisdairM: Happy to go with whatever you decide, just calling it out so that we make a deliberate decision, rather than accidentally fall into the old pattern by inertia.

source/future.tex
```diff
@@ -134,6 +135,71 @@
 the namespace \tcode{std}.
 \end{example}
 
+\rSec2[ccomplex.syn]{Header \tcode{<ccomplex>} synopsis}
```
tkoeppe: Done.

source/lib-intro.tex
tkoeppe: I "fixed" this problem by reverting the changes to Clause 17. I noticed that we already list `<strstream>` as part of the library, even though we only ever mention it again in Annex D. We can do the same with these four headers.

source/lib-intro.tex
```diff
@@ -1285,18 +1285,19 @@
...
+\ref{support.types}      & Types                     & \tcode{<cstddef>}          \\ \rowsep
+\ref{support.limits}     & Implementation properties & \tcode{<cfloat>} \tcode{<limits>} \tcode{<climits>} \\ \rowsep
+\ref{cstdint}            & Integer types             & \tcode{<cstdint>}          \\ \rowsep
+\ref{support.start.term} & Start and termination     & \tcode{<cstdlib>}          \\ \rowsep
+\ref{support.dynamic}    & Dynamic memory management & \tcode{<new>}              \\ \rowsep
+\ref{support.rtti}       & Type identification       & \tcode{<typeinfo>}         \\ \rowsep
+\ref{support.exception}  & Exception handling        & \tcode{<exception>}        \\ \rowsep
+\ref{support.initlist}   & Initializer lists         & \tcode{<initializer_list>} \\ \rowsep
+\ref{support.runtime}    & Other runtime support     & \tcode{<cstdarg>}          \\ \rowsep
+\ref{depr.cstdalign.syn}, \ref{depr.cstdbool.syn} & Deprecated headers & \tcode{<cstdalign>} \tcode{<cstdbool>} \\ \rowsep
```
AlisdairM: This row should probably drop to last, to preserve section-ordering in the table.

tkoeppe: Good idea, done. Thanks!

source/future.tex
```diff
@@ -51,6 +51,15 @@
 \rSec1[depr.c.headers]{C standard library headers}
 
 \pnum
+For compatibility with the C standard library\indextext{library!C standard},
+the \Cpp standard library provides headers
+\tcode{<ccomplex>}~(\ref{depr.ccomplex.syn}),
```
jensmaurer: We're still in a section "C standard library headers" here, yet we keep talking about C++ headers.
Either rename this section to "Standard library headers" or add a section "C++ standard library headers" that contains the C++-only content. (I'm unsure what remains in the existing "C headers" section in the latter case.)

jensmaurer:
The general approach of moving the definitions of the deprecated headers into annex D is good. Annex C is about incompatibilities with C, so we can be non-normative and terse there, with plenty of \refs.

source/future.tex
tkoeppe: @jensmaurer: OK, snark alert: If you squint at it right, you could think of "C standard library headers" to be those headers in the standard library that start with "C". Then it would kind of make sense.

Ultimately, *everything* we describe in this document is about C++. The name "C header" is just a mnemonic.

jensmaurer: @tkoeppe: The entire clause talks about compatibility features.
Maybe just say "Headers"? And &lt;ccomplex> and friends aren't really for compatibility with C, despite your edit saying so.

tkoeppe: Hm. "Headers" sounds oddly bland, but maybe that's the only accurate description of this part...

jensmaurer:
You need to rebase.

jensmaurer:
This interacts with [LWG 2835](http://cplusplus.github.io/LWG/lwg-active.html#2835) "LWG 2536 seems to misspecify `<tgmath.h>`".

tkoeppe:
Yes, absolutely.

tkoeppe:
LWG agrees with this direction; I'll make a few tweaks and then ask for a review.

tkoeppe:
I decided to add a new section "C++ standard library headers [depr.cpp.headers]", as a sibling section to the existing "C standard library headers [depr.c.headers]", and the new section comes first so that the existing one can refer to it.

@zygoloid: Ready for review and application.

@mclow: Could you take a final look to see if this is in accord with LWG guidance?

mclow:
This looks good to me.

Issue #329, opened 2014-06-18
All definitions should be indexed


burblebee:
It was decided at the Rap. mtg that all definitions should be indexed, i.e., definitions which use \term{} should be fixed to use \defn{} or \defnx{}.

jensmaurer:
The fun part of this is that some definitions use \term and have an \indextext next to it. I agree \defnx is technically superior, though. For what it's worth, there are 684 mentions of \term, which should probably all become either \defn or \placeholder or something else.

tkoeppe:
I think PR #980  addresses part of that. It's pending review.

jensmaurer:
I would have been nice to mention #329 in PR #980 to avoid duplicate work...

tkoeppe:
That's true. I find it useful to browser the pending pull requests periodically to see what's pending. Feel free to edit pending PRs with issue links as you find them.

Eelis:
Update to quantify the "big" label: there's now 216 uses of \term left.

jwakely:
As I mentioned in #3435, these look wrong:

````
declarations.tex:Except for this special case, a parameter shall not have type \term{cv}
declarations.tex:if any of its \term{namespace-definition}{s} is exported,
expressions.tex:\tcode{std::type_info} or \tcode{const} \term{name} where \term{name} is an
locales.tex:\term{explicit},
````

AlisdairM:
Bump.  Is this something we should consider for C++26? or maybe an early project after the final C++26 doc is shipped to publish?

My quick `grep | wc` suggests we have made slow progress, down to 188 occurrences --- but some of those are in new wording!

I think there are some low-hanging fruit for obvious `\term` -> `\defn` changes, but the majority remaining are less clear to me.

Are we interesting in partial fixes of just the most obvious cases?

Issue #1060, opened 2016-11-17
Separate textual and code placeholders; use slanted font for the latter


tkoeppe:
We currently use the `\placeholder` macro (and its non-italic-correcting variant `\placeholdernc`) for both textual and code placeholders, and we typeset both in "italic".

This has two problems:

1. Monospace italics don't always look that great. Slanted "faux-italics" would look better for meta-variables. (But true italics may still be nice for things like `see below`.)

2. There is no bold italic monospace font, so a code placeholder in a heading, say, will fall back to bold slanted monospace.

My general idea here is that we should perhaps typeset *all* code placeholders in slanted, rather than true italics. Not only would that make placeholders look the same in headings and body text, but it is also generally nicer.

![image](https://cloud.githubusercontent.com/assets/6378233/20394034/1543753a-acd6-11e6-8151-18cecd2c2040.png)

This raises a new problem: We use `\placeholder` both in text and in code. Obviously, in normal text placeholders should still be decorated as true italics. So we need to distinguish the two use cases, either by laboriously replacing text placeholders with `\textit` or some new macro, or by making `\placeholder` smart and recognizing its context.

jensmaurer:
Let me object to introducing \textit for anything new. If we clearly mark all semantic things appropriately, we can easily fix the font later. Going \textit gets us back to where we've been all along: unable to add index entries for term definitions, for example.

In any case, I'd like to point out that we sometimes use the code placeholder and mention it in regular text to explain what it means. That's probably an argument against different fonts for these.

tkoeppe:
I think that's a red herring: code placeholders should always appear in code placeholder font (typewriter + decoration), whether that's in a code block or in body text. The things I want to distinguish are conceptual placeholders (like "a memory access *A*") and meta variables (like *`type`*).

jensmaurer:
Ok. Whatever the outcome, we should be clear what to use in "A is sequenced before B" vs. "offsetof(type, member-designator" vs. "atomic<integral>" (the latter doesn't even say "T"). If we can clearly differentiate these uses, let's introduce a \metavar for the first case NOW and discuss formatting later.

tkoeppe:
The other option that I hinted at is that we can use `\placeholder` for both, and since the latter usage always appears in a code font setting, we could make the macro smarter to distinguish the two use cases and select the desired style accordingly. In some sense, both uses are kinds of placeholders (just with somewhat different usage).

jensmaurer:
[...]

PR #1062, opened 2016-11-17
[intro] Replace \term with \placeholder or \defn as appropriate

jensmaurer:
Partially addresses #329.

### DIFF
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -39,7 +39,7 @@ requirement appear at various places within this International Standard.
 \Cpp is a general purpose programming language based on the C
 programming language as described in ISO/IEC 9899:2011
 \doccite{Programming languages --- C} (hereinafter referred to as the
-\indexdefn{C!standard}\term{C standard}). In addition to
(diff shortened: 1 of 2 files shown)


source/intro.tex
```diff
@@ -950,16 +950,16 @@
 \pnum
 \defn{Sequenced before} is an asymmetric, transitive, pair-wise relation between
 evaluations executed by a single thread~(\ref{intro.multithread}), which induces
-a partial order among those evaluations. Given any two evaluations \term{A} and
-\term{B}, if \term{A} is sequenced before \term{B}
-(or, equivalently, \term{B} is \defn{sequenced after} \term{A}),
+a partial order among those evaluations. Given any two evaluations \placeholder{A} and
+\placeholder{B}, if \placeholder{A} is sequenced before \placeholder{B}
```
tkoeppe: This is OK for now. See Issue #1060, I may eventually wish to move away from using `\placeholder` both in code and in textual situations, but I'm not sure yet how.

Issue #605, opened 2016-01-21
[facet.num.put.virtuals] Table 89 refers to an undefined variable and reapplies a bitwise &


cubbimew:
The text of Stage 1 begins by introducing local variables `flags`, `basefield`, `uppercase`, `floatfield`, `showpos`, and `showbase`, note that `showbase` is initialized with `flags & (ios_base::showbase)`

Tables 86 and 87 use `basefield`, `uppercase`, and `floatfield` as one would expect. However, Table 89 uses `showbase` in the expression `flags & showbase`, re-applying the bitwise &, and then uses undefined variable `showpoint` in the expression `flags & showpoint`. 

Either the definition of showpoint should be added to the list of local variables in Stage 1 and Table 89 modified to say `showpos` and `showbase` without the repeated bitwise &, or `showbase` should be removed from the list of local variables and Table 89 modified to say `flags & ios_base::showpos` and `flags & ios_base::showpoint`. The former option seems more consistent, the latter more concise.

PR #1065, opened 2016-11-18
[facet.num.put.virtuals] define showpoint and don't apply & twice

cubbimew:
Proposed fix for #605


tkoeppe:

Issue #693, opened 2016-04-13
"Underfull" and "Overfull" warnings emitted from Latex.


burblebee:
We get a ton of "Underfull" and "Overfull" warnings emitted during the Latex build of the spec.  What causes them?  Can they be fixed?

godbyk:
The underfull and overfull warnings mean that TeX was unhappy with the spacing in a line—either the line ended up being too short or too long.

For small values of points, this is rarely noticeable to the casual reader and can be ignored. For larger values, there will be white space at the end of a line or a line may overrun its margin. 

The amount that you let TeX stretch or shrink the spaces in a line to make it fit the margins can be modified to make TeX more or less lax.

Generally, I just look for overfull or underfull lines that are more than a few points and visually inspect them to see if I need to make any adjustments (e.g., teach TeX how to hyphenate a new word, reword a sentence so it can be typeset more easily, adjust TeX's line breaking parameters to better accommodate that paragraph).

zygoloid:
@tkoeppe went through and fixed a bunch of these, but I don't think he got through all of the document. It would be good to get this clean before we produce the C++17 IS.

burblebee:
I tracked down one of the Underfull cases - it's caused by a section name which is too long in xref.tex:
`func.searchers.boyer_moore_horspool.creation\quad\ref{func.searchers.boyer_moore_horspool.creation}\\`
which ends up spilling over to two lines in the spec:
```
    func.searchers.boyer_moore_horspool.creation
    20.12.13.3.1
```
The warning is:
```
    Underfull \hbox (badness 10000) in paragraph at lines 642--787
    \T1/lmr/m/n/10 (+20) func.searchers.boyer_-moore_-hor-spool.creation
```
although the actual line is 759.
This long section name also causes the section title to spill over:
```
    20.12.13.3.1 boyer_moore_horspool_searcher creation functions
                 [func.searchers.boyer_moore_horspool.creation]
```


I also tracked down an overfull case.  The warning is:
```
    Overfull \hbox (34.36942pt too wide) in paragraph at lines 262--264
    []\T1/lmr/m/n/10 (-20) whether stack is un-wound be-fore call-ing \T1/lmtt/m/n/
    10 std::terminate()
```
caused by an impldef being too long in exceptions.tex line 1296:
`\impldef{whether stack is unwound before calling \tcode{std::terminate()}`
and results in an over-long line in the spec:
```
    whether stack is unwound before calling std::terminate()
[...]

PR #1068, opened 2016-11-18
Resolve some 'Overfull \hbox' warnings.

jensmaurer:
Partially addresses #693.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -3209,8 +3209,8 @@ constexpr const T* data() const noexcept;
 
 \begin{itemdescr}
 \pnum \returns
-A pointer such that \range{data()}{data() + size()} is a valid range, and
-\tcode{data() == addressof(front())}.
+A pointer such that \tcode{data() == addressof(front())},
+and \range{data()}{data() + size()} is a valid range.
 \end{itemdescr}
 
 \rSec3[array.fill]{\tcode{array::fill}}
@@ -6338,7 +6338,7 @@ template <class M> iterator insert_or_assign(const_iterator hint, const key_type
 \pnum
 \requires
 \tcode{is_assignable_v<mapped_type\&, M\&\&>} shall be \tcode{true}.
-\tcode{value_type} shall be \tcode{EmplaceConstructible} into \tcode{map}
+\tcode{value_type} shall be \tcode{Emplace\-Constructible} into \tcode{map}
 from \tcode{k}, \tcode{forward<M>(obj)}.
 
 \pnum
@@ -6373,7 +6373,7 @@ template <class M> iterator insert_or_assign(const_iterator hint, key_type&& k, 
 \pnum
 \requires
 \tcode{is_assignable_v<mapped_type\&, M\&\&>} shall be \tcode{true}.
-\tcode{value_type} shall be \tcode{EmplaceConstructible} into \tcode{map}
+\tcode{value_type} shall be \tcode{Emplace\-Constructible} into \tcode{map}
 from \tcode{move(k)}, \tcode{forward<M>(obj)}.
 
 \pnum
@@ -7830,7 +7830,7 @@ template <class M> iterator insert_or_assign(const_iterator hint, const key_type
 \pnum
 \requires
 \tcode{is_assignable_v<mapped_type\&, M\&\&>} shall be \tcode{true}.
-\tcode{value_type} shall be \tcode{EmplaceConstructible} into \tcode{unordered_map}
+\tcode{value_type} shall be \tcode{Emplace\-Constructible} into \tcode{unordered_map}
 from \tcode{k}, \tcode{forward<M>(obj)}.
 
 \pnum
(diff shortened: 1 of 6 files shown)


source/containers.tex
```diff
@@ -6338,7 +6338,7 @@
 \pnum
 \requires
 \tcode{is_assignable_v<mapped_type\&, M\&\&>} shall be \tcode{true}.
-\tcode{value_type} shall be \tcode{EmplaceConstructible} into \tcode{map}
+\tcode{value_type} shall be \tcode{Emplace\-Constructible} into \tcode{map}
```
tkoeppe: Can you try adding `EmplaceConstructible` to the global hyphenation hints (in `std.tex`) and see if that works? If the issue is just that it's an unknown word, then that's a more sustainable approach.

source/declarations.tex
```diff
@@ -227,12 +227,12 @@
...
 \pnum
 A \grammarterm{nodeclspec-function-declaration} shall declare a
-constructor, destructor, or conversion function.  \footnote{The
+constructor, destructor, or conversion function.\footnote{The
 ``implicit int'' rule of C is no longer supported.}
 \begin{note}
 A \grammarterm{nodeclspec-function-declaration} can only be used in a
 \grammarterm{template-declaration}~(Clause~\ref{temp}),
-\grammarterm{explicit-instantiation}~(\ref{temp.explicit}), or
+\grammarterm{explicit-instan\-tiation}~(\ref{temp.explicit}), or
```
tkoeppe: My more natural reaction would be to remove the non-breaking space first. I always thought that those were tenuous at best; there's very little reason to prevent a break before the reference. Either the reference fits on the same line and all is well, or it doesn't, and then that's still by far the best place to break compared to anywhere else.

jwakely:
The results look very, very good to me.

source/containers.tex
jwakely: I have just tried using this and get no changes (according to diffpdf):

`\hyphenation{tem-plate ex-am-ple name-space name-spaces emplace-constructible copy-constructible less-than-comparable in-put-it-er-a-tor out-put-it-er-a-tor for-ward-iter-a-tor rand-om-acc-ess-iter-a-tor}`

jensmaurer: I am unable to get LaTeX to hyphenate EmplaceConstructible using \hyphenation. Maybe because the word has a capital letter in the middle.

tkoeppe: I think it may be because the text is inside a `\tcode`. We could probably also change that somehow. The `hyphenat` package does something similar.

source/declarations.tex
jensmaurer: Done, although I notice that we do usually attach cross-references to the preceding term.

source/containers.tex
jensmaurer: So, what's the next step here? Adding the manual hyphenation fixes the non-hyphenation of the term in the four places where it's an "overfull \hbox" problem.

tkoeppe: Yes, keep it manual for now. It's what we've been doing and it's an OK solution. We can research better solutions offline.

source/declarations.tex
tkoeppe: Right, but as I said that's rather strange if it requires hyphenation in the previous term. I think the non-breaking space is only an approximation, and what we really want is a "discouraging" space that says "please try not to break here if possible".

But surely it's *much* better to break at a real space than introduce awkward hyphenation. What do you think? In real-world publishing you would probably try to rephrase the text to achieve the desired presentation, but given the constant changes we are making, I would strongly prefer the simple and maintainable solution that doesn't require manual intervention.

source/containers.tex
tkoeppe: Ah, I just tried, and we probably really don't want to monkey with global `tcode`, because then we'd end up hyphenating things like `1.2e-34`. :-)

jensmaurer: regarding "very brittle": We have lots of paragraphs, and only if this particular paragraph changes again in the future do we need to reconsider the manual hyphenation.

source/declarations.tex
jensmaurer: Fine with me. As I said above, the newest update has the "make it breaking whitespace" change.

source/utilities.tex
```diff
@@ -1941,7 +1941,8 @@
 template <size_t I, class... Types>
   constexpr tuple_element_t<I, tuple<Types...>>&& get(tuple<Types...>&& t) noexcept; // Note A
 template <size_t I, class... Types>
-  constexpr tuple_element_t<I, tuple<Types...>> const& get(const tuple<Types...>& t) noexcept; // Note B
+  constexpr tuple_element_t<I, tuple<Types...>> const&
```
tkoeppe: Just for visual harmony, would it be possible to break all three declarations in the same way?

source/utilities.tex
```diff
@@ -13769,7 +13779,7 @@
 \requires \tcode{F} shall be \tcode{CopyConstructible}.
 
 \pnum
-\remarks This constructor shall not participate in overload resolution unless
+\remarks This constructor template shall not participate in overload resolution unless
```
tkoeppe: Nice :-)

source/tables.tex
```diff
@@ -457,7 +457,7 @@
 {
  \begin{LongTable}
  {#1}{#2}
- {x{.35\hsize}x{.28\hsize}x{.30\hsize}}
+ {x{.35\hsize}x{.28\hsize}x{.29\hsize}}
 }
```
tkoeppe: Interesting. I wonder were those original values came from. I suppose they were too big because they didn't account for cell padding?

tkoeppe:
I agree, a very nice change. I wonder if we get to merge it before the motions; will check with @zygoloid.

source/utilities.tex
jensmaurer: I don't think this is obviously an improvement, but done anyway.

PR #1072, opened 2016-11-18
LWG2754 The in_place constructors and emplace functions added by P0032R3 don't require CopyConstructible

jwakely:
Also includes the changes to [ans.cons] p9 from P0504R0 (LWG Motion 19)
which modify the same paragraph.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -5048,10 +5048,10 @@ public:
 
   template <class ValueType> any(ValueType&& value);
 
-  template <class T, class... Args>
-    explicit any(in_place_type_t<T>, Args&&...);
-  template <class T, class U, class... Args>
-    explicit any(in_place_type_t<T>, initializer_list<U>, Args&&...);
+  template <class ValueType, class... Args>
+    explicit any(in_place_type_t<ValueType>, Args&&...);
+  template <class ValueType, class U, class... Args>
+    explicit any(in_place_type_t<ValueType>, initializer_list<U>, Args&&...);
 
   ~any();
 
@@ -5062,9 +5062,9 @@ public:
   template <class ValueType> any& operator=(ValueType&& rhs);
 
   // \ref{any.modifiers}, modifiers
-  template <class T, class... Args>
+  template <class ValueType, class... Args>
     void emplace(Args&& ...);
-  template <class T, class U, class... Args>
+  template <class ValueType, class U, class... Args>
     void emplace(initializer_list<U>, Args&&...);
   void reset() noexcept;
   void swap(any& rhs) noexcept;
@@ -5151,7 +5151,6 @@ Let \tcode{T} be equal to \tcode{decay_t<ValueType>}.
 \pnum
 \requires
 \tcode{T} shall satisfy the \tcode{CopyConstructible} requirements.
-If \tcode{is_copy_constructible_v<T>} is \tcode{false}, the program is ill-formed.
 
 \pnum
 \effects
@@ -5159,9 +5158,10 @@ Constructs an object of type \tcode{any} that contains an object of type \tcode{
 
 \pnum
 \remarks
-This constructor shall not participate in overload resolution
-if \tcode{decay_t<ValueType>} is the same type as \tcode{any} or
-if \tcode{ValueType} is a specialization of \tcode{in_place_type_t}.
+This constructor shall not participate in overload resolution unless
+\tcode{T} is not the same type as \tcode{any},
+\tcode{T} is not a specialization of \tcode{in_place_type_t},
+and \tcode{is_copy_constructible_v<T>} is \tcode{true}.
 
 \pnum
 \throws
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -5151,17 +5151,17 @@
...
 \effects
 Constructs an object of type \tcode{any} that contains an object of type \tcode{T} direct-initialized with \tcode{std::forward<ValueType>(value)}.
 
 \pnum
 \remarks
-This constructor shall not participate in overload resolution
-if \tcode{decay_t<ValueType>} is the same type as \tcode{any} or
-if \tcode{ValueType} is a specialization of \tcode{in_place_type_t}.
+This constructor shall not participate in overload resolution unless
+\tcode{T} is the same type as \tcode{any},
```
burblebee: is *not* the same??

source/utilities.tex
```diff
@@ -5170,12 +5170,18 @@
...
 \begin{itemdecl}
-template <class T, class... Args>
-  explicit any(in_place_type_t<T>, Args&&... args);
+template <class ValueType, class... Args>
+  explicit any(in_place_type_t<ValueType>, Args&&... args);
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
+Let \tcode{T} be equal to \tcode{decay_t<ValueType>}.
```
burblebee: I would leave out "equal to" - it's not in the paper, it's inconsistent with other "Let"s, and it adds nothing.

source/utilities.tex
```diff
@@ -5186,18 +5192,26 @@
 \throws Any exception thrown by the selected constructor of \tcode{T}.
 
 \pnum
-\remarks This constructor shall not participate in overload resolution
-unless tcode{is_constructible_v<T, Args...>} is \tcode{true}.
+\remarks
+This constructor shall not participate in overload resolution unless
+\tcode{is_copy_constructible_v<T>} is \tcode{true}
+and \tcode{is_constructible_v<T, Args...>} is \tcode{true}.
```
burblebee: fyi, folks usually put the conjunction ("and" in this case) at the end of the previous line.  not that you need to change it here, but I try to keep the spec coding style consistent (I'm sure you've noticed we have a long way to go there!)

source/utilities.tex
```diff
@@ -5186,18 +5192,26 @@
...
 \begin{itemdecl}
-template <class T, class U, class... Args>
-  explicit any(in_place_type_t<T>, initializer_list<U> il, Args&&... args);
+template <class ValueType, class U, class... Args>
+  explicit any(in_place_type_t<ValueType>, initializer_list<U> il, Args&&... args);
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
+Let \tcode{T} be equal to \tcode{decay_t<ValueType>}.
```
burblebee: same with "equal to" here

source/utilities.tex
```diff
@@ -5302,13 +5319,17 @@
...
 \begin{itemdecl}
-template <class T, class... Args>
+template <class ValueType, class... Args>
   void emplace(Args&&... args);
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
-\requires \tcode{is_constructible_v<T, Args...>} is \tcode{true}.
+Let \tcode{T} be equal to \tcode{decay_t<ValueType>}.
```
burblebee: ditto for "equal to"

source/utilities.tex
```diff
@@ -5325,16 +5346,26 @@
...
 \indexlibrarymember{emplace}{any}%
 \begin{itemdecl}
-template <class T, class U, class... Args>
+template <class ValueType, class U, class... Args>
   void emplace(initializer_list<U> il, Args&&... args);
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
+Let \tcode{T} be equal to \tcode{decay_t<ValueType>}.
```
burblebee: ditto for "equal to"

source/utilities.tex
jwakely: Oops, thanks.

source/utilities.tex
jwakely: And it _is_ in LWG 2754, for better or worse.

source/utilities.tex
jwakely: FWIW I think that's the wrong choice, because if we have an Oxford comma then it seems natural to put the line break after the comma, and start the next line with the conjunction. 

It also involves more editing for cases like:

    case one,
    case two,
    case three, and
    case four.

If you want to add a new case between three and four you have to edit an existing line, not just insert a new one:

    case one,
    case two,
    case three, <DEL>and<DEL>
    <INS>case three-point-five, and</INS>
    case four.

source/utilities.tex
jwakely: Changed and force pushed.

source/utilities.tex
jwakely: Changed and force pushed.

source/utilities.tex
burblebee: It's NOT in 2754 - you must be looking at the superceded versions.  See http://wg21.cmeerw.net/lwg/issue2754 (or skip past the superceded versions in http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0165r3.html).

source/utilities.tex
burblebee: You make a good point.  I was simply following existing style - I will switch to doing things your way from now on, assuming others agree.  Any one else have a preference?

source/utilities.tex
jwakely: Oops, you're right, I was looking at the superseded version. Although confusingly, the final version says "Let `T` be `decay_t<ValueType>`." even for the existing text that isn't changed (which is wrong). Bah. Anyway, I changed it.

burblebee:
This is ready to go now right?  Can you merge it into the motions branch?  Or maybe I should ask: if I clicked "squash and merge", would it go to the motions branch or master?

burblebee:
Hmm - probably shouldn't have closed until you committed right?

burblebee:
I see - you can tell the base from the line at the top:

"wants to merge 1 commit into cplusplus:motions-2016-11-lwg-11 from jwakely:motions-2016-11-lwg-11"

So this will go to the motions branch.  Great - will merge then...

jwakely:
Yep, I created the pull request so it would only add to the branch, not master - thanks!

PR #1076, opened 2016-11-19
[class] Rephrase definition of M(X) used to define a standard-layout …

### LINKED ISSUE #496: [class] (9)/7.8 has ambiguous wording
> If `X` is a non-union class type, the set `M(X)` is empty if `X` has no (possibly inherited (Clause 10)) non-static data members; otherwise, it consists of the type of the first non-static data member of `X` (where said member may be an anonymous union), `X0`, and the elements of `M(X0)`.

would be clearer as something like:

> If `X` is a non-union class type with no (possibly inherited (Clause 10)) non-static data members, the set `M(X)` is empty.
> If `X` is a non-union class type with first non-static data member `X0` (where said member may be an anonymous union), the set `M(X)` consists of `X0` and the elements of `M(X0)`.

jensmaurer:
…class.

Fixes #496.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -173,12 +173,14 @@ address~(\ref{expr.eq}).}
 
 $M(\mathtt{X})$ is defined as follows:
 \begin{itemize}
-\item If \tcode{X} is a non-union class type, the set $M(\mathtt{X})$ is
-empty if \tcode{X} has no (possibly inherited (Clause~\ref{class.derived}))
-non-static data members;
-otherwise, it consists of the type of the first non-static data member
-of \tcode{X} (where said member may be an anonymous union), $\mathtt{X}_0$,
-and the elements of $M(\mathtt{X}_0)$.
+\item If \tcode{X} is a non-union class type with no (possibly
+inherited (Clause~\ref{class.derived})) non-static data members, the set
+$M(\mathtt{X})$ is empty.
+
+\item If \tcode{X} is a non-union class type whose first non-static data
+member has type $\mathtt{X}_0$ (where said member may be an anonymous union),
+the set $M(\mathtt{X})$ consists of $\mathtt{X}_0$ and the elements of
+$M(\mathtt{X}_0)$.
 
 \item If \tcode{X} is a union type, the set $M(\mathtt{X})$ is
 the union of all $M(\mathtt{U}_i)$ and the set containing all $\mathtt{U}_i$,


source/classes.tex
```diff
@@ -173,12 +173,14 @@
...
-non-static data members;
-otherwise, it consists of the type of the first non-static data member
-of \tcode{X} (where said member may be an anonymous union), $\mathtt{X}_0$,
-and the elements of $M(\mathtt{X}_0)$.
+\item If \tcode{X} is a non-union class type with no (possibly
+inherited (Clause~\ref{class.derived})) non-static data members, the set
+$M(\mathtt{X})$ is empty.
+
+\item If \tcode{X} is a non-union class type with the first non-static data
+member having type \tcode{X0} (where said member may be an anonymous union),
```
zygoloid: "with the first non-static data member having type" -> "whose first non-static data member has type"

... would seem a bit more natural to me.

jensmaurer:
Various fixes applied, including the one above plus indexing confusion resolved.

source/classes.tex
```diff
@@ -173,12 +173,14 @@
...
-non-static data members;
-otherwise, it consists of the type of the first non-static data member
-of \tcode{X} (where said member may be an anonymous union), $\mathtt{X}_0$,
-and the elements of $M(\mathtt{X}_0)$.
+\item If \tcode{X} is a non-union class type with no (possibly
+inherited (Clause~\ref{class.derived})) non-static data members, the set
+$M(\mathtt{X})$ is empty.
+
+\item If \tcode{X} is a non-union class type whose first non-static data
+member has type $\mathtt{X}_0$ (where said member may be an anonymous union),
```
tkoeppe: Everywhere else where you've been making changes you used a style like `\tcode{X}$_0$` -- any reason to prefer one over the other? (Personally, I prefer this, one, though I would perhaps make a macro `\mcode` for this use.)

jensmaurer: I'm using the style otherwise used locally. I'd like to go forward with this change and then:

I agree we're inconsistent here, and the other changes should probably have used \mathtt (or rather, \mcode). It seems we agree that a "T"-style placeholder for a type should be \tcode (not \placeholder) and indexing should be in math mode, with $\mcode{T}_i$ the preferred style of indexing. I'd suggest to open a new issue for exactly this and fix that aspect globally. (This should be a "no visual difference" change.)

tkoeppe: I don't even think this is a question of consistency. Both ways work and either is fine. I just wanted to point out that you've done this differently elsewhere, but I think going any further is busywork with limited value.

tkoeppe: (Leaving the actual review to @zygoloid.)

PR #1077, opened 2016-11-19
[class.copy] Introduce three subsections.

### LINKED ISSUE #490: Add subclauses to 12.8 for constructors, assignment, copy elision
This subclause is unwieldy and would benefit from further structure.

jensmaurer:
Fixes #490.

This should probably be reviewed by Richard. Note that I moved the "accessible" paragraph to the top, although I think that's redundant nowadays that we always perform overload resolution for copy/move, and access checking is on of the last steps there. Also, the "accessible" paragraph should say in which context the access is checked (existing defect).

### DIFF
--- a/source/special.tex
+++ b/source/special.tex
@@ -2424,8 +2424,6 @@ B::B(V* v, A* a) {
 \rSec1[class.copy]{Copying and moving class objects}%
 \indextext{copy!class~object|see{constructor, copy; assignment, copy}}%
 \indextext{move!class~object|see{constructor, move; assignment, move}}%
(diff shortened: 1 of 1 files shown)


tkoeppe:
@zygoloid: For you.

jensmaurer:
Pointed Mike Miller (CWG chair) to the changes.

tkoeppe:
Maybe we should call out these larger changes in the editor's report separately. That way we can give a clear message that we have made some important changes, please review, and here are five hundred changes of punctuation, capitalziation and whitespace that you probably don't need to look at.

PR #1078, opened 2016-11-19
[17-30] Create index entries for all namespaces

AlisdairM:
Index each library namespace, and the reserved posix namespace.
Note that many entries for literals namespaces are added after
the header syonpsis, as typically they are the last entries in
that synposis, and the index will find them much more easily
here.  Indexing inside the synopsis is avoided, as it leads to
duplicate index entries.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -381,6 +381,9 @@ namespace std {
   }
 }
 \end{codeblock}
+\indexlibrary{\idxcode{std}!\idxcode{literals}}%
+\indexlibrarymember{std}{complex_literals}%
+\indexlibrarymember{std::literals}{complex_literals}%
 
 \rSec2[complex]{Class template \tcode{complex}}
 
@@ -1314,6 +1317,9 @@ ensure, for a call with at least one argument of type \tcode{complex<T>}:
 
 \rSec2[complex.literals]{Suffixes for complex number literals}
 
+\indexlibrary{\idxcode{std}!\idxcode{literals}}%
+\indexlibrarymember{std}{complex_literals}%
+\indexlibrarymember{std::literals}{complex_literals}%
 \indexlibrary{complex!literals}%
 \indexlibrary{literals!complex}%
 \pnum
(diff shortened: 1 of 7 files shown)


tkoeppe:
Hm, do you think it's useful to index "namespace std"? Isn't *all* of the library in that namespace. I'm not sure... let me know you think this really adds value.

AlisdairM:
I index the clause 17 wording on use and requirements on namespace std.  It is a standard library (reserved) name and belongs in the index, without going crazy and listing every library page.

Collecting all the subnamespaces under 'std' in one place (in the index) also seemed quite useful to me.  Finally, indexing the literals namespaces was a little more interesting, given their structure, but I think I hit a happy medium.

source/numerics.tex
```diff
@@ -381,6 +381,9 @@
   }
 }
 \end{codeblock}
+\indexlibrary{\idxcode{std}!\idxcode{literals}}%
```
zygoloid: Hmm, I can understand where you're coming from, trying to group the sub-namespaces together under the `std` index entry, but I don't think that's useful to someone looking up the namespace (I would look for `complex_literals`, not for `std`, when looking for the `complex_literals` namespace), and I don't think a particular entry in the index of library names is the right place for a description of the namespace structure of the library. If we want such a description, it should probably go in a subclause of [organization] instead (but it's not obvious to me who the audience for such a description would be).

This approach creates five index entries:

 * std, literals
 * std, complex_literals
 * complex_literals, std
 * std::literals, complex_literals
 * complex_literals, std::literals

I don't really like any of these, and we don't index `literals` at all this way. I think what we want would be either:

 * literals, complex_literals
 * complex_literals, literals

or

 * literals
 * complex_literals, literals

I would index this as just `\indexlibrarymember{literals}{complex_literals}`; that seems most consistent with what we do for class members.

tkoeppe:
@AlisdairM: Do you have any plans to revise and rework this, or should we just close it?

AlisdairM:
Sorry, I plan to revise this when I get back from current road trip.

tkoeppe:
@AlisdairM: How's the road trip going? Are you still interested in revising this?

AlisdairM:
Sorry - yes, I want to revive it, but working out the best way to respond to feedback.
I may break it apart to simpler pull requests.  I think indexing namespace 'std' and 'posix' in clause 17 should be simple, and I hope non-controversial.  Indexing all of the literals namespaces is probably trickier to get the right balance.  The other namespaces fall somewhere in-between.

tkoeppe:
Please reopen if needed and when ready (or send new, smaller PRs instead).

PR #1079, opened 2016-11-20
Multiple global style improvements

tkoeppe:
* Bulleted lists spaced uniformly consistently with paragraphs.
* Bulleted lists work regardless of newlines in the source.
* All indents are now the same (1em): grammar, codeblocks, itemdecls.
* Grammar spaced uniformly consistently with paragraphs. (Removed use of minipages.)
* Removed unnecessary `simplebnf` environment. "Simple" grammar now uses less space and generally flows better with its surrounding paragraph.
* Removed excessive whitespace at the end of examples.

### DIFF
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -413,25 +413,23 @@
  \newcommand{\bnfindentfirst}{\BnfIndent}
  \newcommand{\bnfindentinc}{\BnfInc}
  \newcommand{\bnfindentrest}{\BnfRest}
- \begin{minipage}{.9\hsize}
  \newcommand{\br}{\hfill\\}
  \frenchspacing
  }
  {
  \nonfrenchspacing
- \end{minipage}
  }
 
 \newenvironment{BnfTabBase}[1]
 {
  \begin{bnfbase}
  #1
- \begin{indented}
+ \begin{adjustwidth}{1em}{}
  \begin{tabbing}
- \hspace*{\bnfindentfirst}\=\hspace{\bnfindentinc}\=\hspace{.6in}\=\hspace{.6in}\=\hspace{.6in}\=\hspace{.6in}\=\hspace{.6in}\=\hspace{.6in}\=\hspace{.6in}\=\hspace{.6in}\=\hspace{.6in}\=\hspace{.6in}\=\kill}
+ \hspace*{1em}\=\hspace{\bnfindentinc}\=\hspace{.6in}\=\hspace{.6in}\=\hspace{.6in}\=\hspace{.6in}\=\hspace{.6in}\=\hspace{.6in}\=\hspace{.6in}\=\hspace{.6in}\=\hspace{.6in}\=\hspace{.6in}\=\kill}
 {
  \end{tabbing}
- \end{indented}
+ \end{adjustwidth}
  \end{bnfbase}
 }
 
@@ -464,18 +462,13 @@
 
 \newenvironment{bnf}
 {
- \begin{bnfbase}
- \list{}
-	{
-	\setlength{\leftmargin}{\bnfindentrest}
-	\setlength{\listparindent}{-\bnfindentinc}
-	\setlength{\itemindent}{\listparindent}
-	}
- \BnfNontermshape
- \item\relax
+ \begin{bnfbase}%
+ \begin{adjustwidth}{2em}{0em}%
+ \BnfNontermshape%
+ \hspace*{-1.5em}%
 }
 {
- \endlist
+ \end{adjustwidth}
  \end{bnfbase}
 }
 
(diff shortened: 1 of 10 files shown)


source/macros.tex
```diff
@@ -424,32 +424,30 @@
 \newcommand{\nonterminal}[1]{{\BnfNontermshape #1}}
 
 \newenvironment{bnfbase}
- {
+ {%
```
burblebee: Does % do anything here?

source/preprocessor.tex
```diff
@@ -271,14 +271,12 @@
 
 \pnum
 Preprocessing directives of the forms
-
 \begin{ncbnftab}
 \indextext{\idxcode{\#if}}%
-\terminal{\# if}\>\>constant-expression new-line group\opt\br
+\terminal{\# if} constant-expression new-line group\opt\br
 \indextext{\idxcode{\#elif}}%
-\terminal{\# elif}\>\>constant-expression new-line group\opt
```
tkoeppe: This is a tab stop in a tabular setting. But I figured that that was completely unnecessary here (and I have also changed the tab stops to fit better with the main grammars, so this wasn't quite working anymore.)

zygoloid:
I'm currently looking through the standard with this applied. Mostly it looks really good. A raw stream of consciousness of some thoughts:

1) Lists look a little cramped right now, but I'm not sure if that's just overfamiliarity with the old style leading me astray.
2) [lex.charset]/1's table of characters looks a little silly now the extra whitespace above and below the character set list is gone -- there's significantly more space between rows than above and below the table.
3) Grammar snippets are not indented enough for my tastes, both text -> grammar and non-terminal -> alternatives indents aren't enough to clearly separate them from the surroundings. Can we retain bullet indent for these? (That said, the indent of [lex.pptoken]/3.1's snippet looks nice as-is.)
4) [lex.pptoken]/3's example still has imbalanced vertical whitespace, but I'm not sure if that's a source problem for this particular example or something more general.
5) Removal of minipages is permitting grammar productions to be split across pages, which we previously were trying hard to avoid (that's why we had them), eg pp-number production in 2.9. The problem of productions split across page boundaries is worsened by the reduced grammar indent.
6) We have a lot of vert whitespace around tables, and it's a little jarring now there's not much anywhere else.
7) Footnote 23 has been laid out on top of the bottom margin / section number.
8) Going back to point 1, I think there is some value in inter-paragraph and inter-list-item spacing being different. It's less obvious where the paragraph boundaries are with the new spacing (even though the paragraph spaces used to be /less/ than the list-item spacing). I wonder whether adding a little extra vertical space with each \pnum would help.
9) Page 90 has some text flowing slightly into the footnote area.
10) 5.2.11/7 has a lot of vertical whitespace (block math mode). Something about that extra whitespace comes as a relief after 100 pages of dense text with little vertical rest :)
11) 5.20/2.10 (nested list bullets) have bad vertical whitespace
[skipping clause 7-12]
12) inline grammar snippets still looking really nice with the codeblock indent (13.3.1.1.2 has some good examples)
13) 13.3.1.8/1.2 again has bad vspace with nested lists, and before 13.3.3.2/3.3
14) 13.5/1: *operator-function-id:* definition is slightly more indented than *operator:* definition
15) 13.5.6 has inconsistent indentation for the two lines of grammar snippet
[skipping to clause 20]
16) 20.2.6: where did the paragraph number and lack of indent here come from?
17) I think I'm generally ok with less  itemdescr indentation, but when the itemdecl ends with an indented continuation line, there is now sometimes no difference in indentation between the declaration "heading" and the pieces below it. I wonder if we can do better there. Crazy idea: move itemdecls left a bit (into the left margin)
18) 20.5.2.1/17: something bad happened to the list here, and in 20.5.2.6/6, and throughout 20.7.2.3
19) There's more vertical space above itemdecls than other vertical whitespace elsewhere. I like that distinction; these act more like "little headings".
20) 20.5.2.7/7: someone sure felt excited that that case was not valid

tkoeppe:
Thanks a lot for the review! I'm going to close this and start extracting smaller parts, keeping the things we like and discarding the ones we don't.

jwakely:
IMHO...

The reduced whitespace after codeblocks is nice.

While the denser itemize envs are certainly more consistent with paragraph spacing I don't always find them easier to read. In complex cases like [dcl.init] I think it is an improvement, so overall I like it. 

Making the grammar use the same indentation as codeblocks looks nice, although I'm less keen on the second level for indentation for the grammar expressions also reducing. Not a big deal though.

tkoeppe:
@jwakely: Thanks! I will be sending individual PRs to implement each of the things we like.

PR #1082, opened 2016-11-20
[temp.variadic] Move example so that it attaches to the correct parag…

### LINKED ISSUE #964: Example in [temp.variadic] attached to wrong paragraph
[temp.variadic]/4 defines a pack expansion and enumerates the contexts in which it can occur.

[temp.variadic]/5 contains a resolution of core issue 1662 about parameter pack instantiations being considered entities, and also this example:

    template<class ... Types> void f(Types ... rest);
    template<class ... Types> void g(Types ... rest) {
        f(&rest ...); // “&rest ...” is a pack expansion; “&rest” is its pattern
    }

This example used to be attached to [temp.variadic]/4, to which it directly relates. The example should be moved up a paragraph. As is, it's a little odd as the example doesn't relate at all to the paragraph it is attached to.

jensmaurer:
…raph.

Fixes #964.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -1790,12 +1790,6 @@ the pattern is the \grammarterm{cast-expression}
 that contains an unexpanded parameter pack.
 \end{itemize}
 
-\pnum
-For the purpose of determining whether a parameter pack satisfies a rule
-regarding entities other than parameter packs, the parameter pack is
-considered to be the entity that would result from an instantiation of
-the pattern in which it appears.
-
 \begin{example}
 
 \begin{codeblock}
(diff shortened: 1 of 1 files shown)


source/templates.tex
```diff
@@ -1808,6 +1802,13 @@
 \end{example}
 
 \pnum
+For the purpose of determining whether a parameter pack satisfies a rule
+regarding entities other than parameter packs, the parameter pack is
+considered to be the entity that would result from an instantiation of
+the pattern in which it appears.
+
```
tkoeppe: Just one newline is needed here.

jensmaurer:
Removed extra empty line.

PR #1084, opened 2016-11-21
[intro.execution] Adjust example for unsequenced side effects

hubert-reinterpretcast:
Provides a case where side effects to the same memory location are unsequenced. Also removes an unreferenced function declaration.

### DIFF
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -1007,7 +1007,6 @@ potentially concurrent computations.
 \begin{example}
 
 \begin{codeblock}
-void f(int, int);
 void g(int i) {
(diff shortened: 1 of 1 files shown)


tkoeppe:
@jensmaurer: Please take a look.

hubert-reinterpretcast:
My bad on the behavior is undefined line. The i++ + i was quite subtle.

jensmaurer:
No objection to removing the declaration of unused "f".
Regarding the other change: We already have an "undefined behavior" case, so I don't know if there's a particular need to add a second one, in particular since it doesn't fit with the "i = something" style in the other lines.

hubert-reinterpretcast:
I've revised the branch to only remove the unused `f`.

PR #1085, opened 2016-11-21
[lib] Remove 'std::' prefix from library names.

### LINKED ISSUE #431: Qualifying std::move, std::forward, etc. is unnecessary
Library Standardese almost universally qualifies `std::move` and `std::forward`. It occasionally qualifies other things, like `std::pair`, `std::is_constructible`, etc. (especially in TR1-era stuff). All of this is unnecessary, due to 17.6.1.1 [contents]/3:

"Whenever a name `x` defined in the standard library is mentioned, the name `x` is assumed to be fully qualified as `::std::x`, unless explicitly described otherwise. For example, if the Effects section for library function `F` is described as calling library function `G`, the function `::std::G` is meant."

Observe how /3 correctly uses `::std::` (which is what implementers have to do, for functions where ADL is a concern) while the rest of the Library casually says `std::` which isn't even correct. And the occurrences of `std::` do not correspond to where implementers must defend against ADL - e.g. `pair` is never vulnerable (it's a class template, so implementers can say it unqualified), but `equal()` is, yet the Library qualifies `std::pair` but says `equal()` unqualified. Madness!

To avoid useless verbosity, I recommend relying on /3 to the greatest extent possible. The Library should eliminate **all** occurrences of `std::` unless something is especially ambiguous to humans (in which case we should consider formally saying `::std::`). Currently, I believe the only human-ambiguous cases are those where ADL is involved elsewhere - for example, because we rely on ADL swap in some places, it is good and desirable to say `::std::swap` when we mean that.

This means that I disagree with #285, at least when it comes to saying `std::raise`. (A citation link whenever `raise` is mentioned would be entirely fine, though.)

jensmaurer:
The standard library specifies that references to its names are assumed to be prefixed by '::std::'. Therefore, we can remove any explicit 'std::' prefixes.

[iterator.range] was not touched, because it is unclear whether argument-dependent lookup was intended to be disabled here.

Fixes #431.

### DIFF
--- a/source/regex.tex
+++ b/source/regex.tex
@@ -139,7 +139,7 @@ and \tcode{loc} is an object of type \tcode{X::locale_type}.
     template \tcode{basic_regex}.
   \\ \rowsep
 \tcode{X::string_type}
-  & \tcode{std::basic_string<charT>}
+  & \tcode{basic_string<charT>}
   &
   \\ \rowsep
 \tcode{X::locale_type}
@@ -153,7 +153,7 @@ and \tcode{loc} is an object of type \tcode{X::locale_type}.
    \indextext{\idxcode{char_class_type}!regular expression traits}%
  \\ \rowsep
 \tcode{X::length(p)}
-  & \tcode{std::size_t}
+  & \tcode{size_t}
   & Yields the smallest \tcode{i} such that \tcode{p[i] == 0}. Complexity is
     linear in \tcode{i} .
   \\ \rowsep
@@ -1020,7 +1020,7 @@ expression could match the specified character sequence.  \\
 \rSec1[re.badexp]{Class \tcode{regex_error}}
 \indexlibrary{\idxcode{regex_error}}%
 \begin{codeblock}
-class regex_error : public std::runtime_error {
+class regex_error : public runtime_error {
   public:
     explicit regex_error(regex_constants::error_type ecode);
     regex_constants::error_type code() const;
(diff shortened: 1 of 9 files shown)


source/threads.tex
```diff
@@ -3666,7 +3666,7 @@
 
 \item these actions will not block for the shared state to become ready, except that it
 may block if all of the following are true: the shared state was created by a call to
-\tcode{std::async}, the shared state is not yet ready, and this was the last reference
```
tkoeppe: I think we've generally stayed away from anything `async`, because SG1 has been crafting that wording painstakingly and with great deliberation. For example, there's a case to be made that it should be easy for the reader to disinguish `std::async` from `std::launch::async`.

tkoeppe:

source/regex.tex
```diff
@@ -3893,8 +3893,8 @@
     position_iterator position;                                 // \expos
     const value_type* result;                                   // \expos
     value_type suffix;                                          // \expos
-    std::size_t N;                                              // \expos
-    std::vector<int> subs;                                      // \expos
+    size_t N;                                                   // \expos
+    ector<int> subs;                                           // \expos
```
jwakely: "ector"

source/threads.tex
jwakely: We should always qualify `launch::async` to avoid such problems, and then strictly speaking we don't need to qualify the function. But if it helps the reader significantly then we could say "a call to the function template `async`" or something else to disambiguate it.

jwakely:
Apart from the "ector" case where one too many characters was removed this looks good to me. I'm ambivalent about the `std::async` cases.

source/threads.tex
jensmaurer: To err on the side of caution, I'm reverting this change.

jensmaurer:
Fixed patch pushed.

tkoeppe:
Please rebase.

jensmaurer:
Rebased.

PR #1089, opened 2016-11-21
Fix typos and add missing \tcode{}s

jlaire:


tkoeppe:
Can you split this into two commits? "Fix typos", and "Add missing \tcode"? It can stay in the same PR.

jlaire:
Sure, done.

PR #1093, opened 2016-11-21
[library] Use \effects etc. when referring to them in text

### LINKED ISSUE #1091: [lib] Referring to \requires, \effects and others
Function descriptions can contain elements introduced with \requires, \effects, \sync, etc., and in some places the standard text refers to these elements. It does this in several different ways.

- 17.5.1.4p4 uses the same macros that are used to define these elements when mentioning them. This is a bit ugly because it includes the colon.

- A couple of places use `\synopsis{Effects}` or `\synopsis{Throws}`, which looks very wrong because it's bold.

- There's at least one instance of ` ``Requires''`.

- There's at least one instance of `\emph{Returns}`.

Visually, I like \emph the most, since it's rendered the same way as the definitions, but without the colon.

There are also many different nouns used when talking about these: at least element, paragraph, section, clause and subclause. I wonder whether clause and subclause are correct.

I can create a pull request making these consistent if someone decides which form to use. I think either the macros themselves (already used in many places, but come with the colon) or \emph would be best. I can leave the nouns as-is or replace uses of clause/subclause with element or section.

jlaire:
Fixes #1091.

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -487,7 +487,7 @@ preconditions, there will be no \requires paragraph.}
 \item \throws any exceptions thrown by the function, and the conditions that would cause the exception
 \item \complexity the time and/or space complexity of the function
 \item \remarks additional semantic constraints on the function
-\item \errors the error conditions for error codes reported by the function.
+\item \errors the error conditions for error codes reported by the function
 \item \realnotes non-normative comments about the function
 \end{itemize}
 
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -476,7 +476,7 @@ Descriptions of function semantics contain the following elements (as
 appropriate):\footnote{To save space, items that do not apply to a function are omitted.
 For example, if a function does not specify any
 further
-preconditions, there will be no ``Requires'' paragraph.}
+preconditions, there will be no \requires paragraph.}
 
 \begin{itemize}
 \item \requires the preconditions for calling the function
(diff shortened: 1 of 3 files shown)


source/lib-intro.tex
```diff
@@ -1592,15 +1592,15 @@
 \begin{codeblock}
 #include <utility>
 
-// Requires: \tcode{std::forward<T>(t)} shall be swappable with \tcode{std::forward<U>(u)}.
+// \requires \tcode{std::forward<T>(t)} shall be swappable with \tcode{std::forward<U>(u)}.
```
tkoeppe: I wouldn't change these two comments. They're in an example, not in normative wording, and the status quo seems just fine.

source/lib-intro.tex
```diff
@@ -476,7 +476,7 @@
 appropriate):\footnote{To save space, items that do not apply to a function are omitted.
 For example, if a function does not specify any
 further
-preconditions, there will be no ``Requires'' paragraph.}
+preconditions, there will be no \requires paragraph.}
```
tkoeppe: Hm, we should probably also (but separately) decide one only one of paragraph, section or clause.

jlaire: I think section is best, because these can contain more than one paragraph, and my understanding is that clause and subclause have specific ISO meanings.

"element" is also used, but I think it sounds more vague than section.

But I agree that this can be done separately.

source/lib-intro.tex
jlaire: Agreed.

source/lib-intro.tex
jwakely: "Element" is the term used when they are defined, and the term LWG has been using for them recently.

PR #1094, opened 2016-11-21
[library] Change program-defined to user-defined

### LINKED ISSUE #1090: Meaning of "program-defined"
There are 4 instances of the term "program-defined" in the standard. Does it have the same meaning as "user-defined"?

jlaire:
Fixes #1090.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -1026,7 +1026,7 @@ namespace std {
 \pnum
 All of the algorithms are separated from the particular implementations of data structures and are
 parameterized by iterator types.
-Because of this, they can work with program-defined data structures, as long
+Because of this, they can work with user-defined data structures, as long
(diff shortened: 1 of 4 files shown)


tkoeppe:

timsong-cpp:
See http://cplusplus.github.io/LWG/lwg-active.html#2139; this P/R seems to be going in the opposite direction.

tkoeppe:
Absolutely, which is why I wanted Jon to comment on this proposal.

jlaire:
Ah, I wasn't aware that. But given that program-defined isn't currently defined and it's used only a few times (compared to over a hundred hits for user-defined), the situation is a bit silly.

jwakely:
This is a step backwards, I'm trying to get the library to stop using that term. Changing from one undefined term to a different undefined term that means two different things in different parts of the standard is not an improvement.

PR #1096, opened 2016-11-21
[class.copy] Rephrase rule preferring a move constructor

### LINKED ISSUE #712: class.copy rules for overloading as rvalue
The rule for potentially selecting the the constructor as if the object were an rvalue currently reads as:

> When the criteria for elision of a copy/move operation are met, but not for an _exception-declaration_, and the object to be copied is designated by an lvalue, or when the _expression_ in a `return` statement is a (possibly parenthesized) _id-expression_ that names an object with automatic storage duration declared in the body or _parameter-declaration-clause_ of the innermost enclosing function or _lambda-expression_, overload resolution to select the constructor for the copy is first performed as if the object were designated by an rvalue.

This is a very long sentence with lots of internal boolean operators, and every three months I come across this sentence and mis-parse it. It would be great to come with with a rewording of this such that it's obvious what the two cases are. Something to the effect of:

> Overload resolution to select the constructor for the copy is first performed as if the object were designated by an rvalue if either:
> - the criteria for elision of a copy/move operation are met, but not for an _exception-declaration_, and the object to be copied is designated by an lvalue, or  
> - the _expression_ in a `return` statement is a (possibly parenthesized) _id-expression_ that names an object with automatic storage duration declared in the body or _parameter-declaration-clause_ of the innermost enclosing function or _lambda-expression_

Or really any other kind of editorial wizardry to clearly separate the two conditions that lead to the rvalue overload being considered.

jensmaurer:
in a throw-expression or a return statement.

Fixes #712.

### DIFF
--- a/source/special.tex
+++ b/source/special.tex
@@ -3203,17 +3203,23 @@ move construction from the local automatic object to \tcode{t2} that is elided.
 \end{example}
 
 \pnum
-When the criteria for elision of a copy/move operation are met,
-but not for an \nonterminal{exception-declaration},
-and the object
-to be copied is designated by an lvalue,
-or when the \grammarterm{expression} in a \tcode{return} statement
+In the following copy-initialization contexts, a move operation might be used instead of a copy operation:
+\begin{itemize}
+\item If the \grammarterm{expression} in a \tcode{return} statement~(\ref{stmt.return})
 is a (possibly parenthesized) \grammarterm{id-expression}
 that names an object with automatic storage duration declared in the body
 or \grammarterm{parameter-declaration-clause} of the innermost enclosing
-function or \grammarterm{lambda-expression},
+function or \grammarterm{lambda-expression}, or
+
+\item if the operand of a \grammarterm{throw-expression}~(\ref{expr.throw})
+is the name of a non-volatile automatic object
+(other than a function or catch-clause parameter)
+whose scope does not extend beyond the end of the innermost enclosing
+\grammarterm{try-block} (if there is one),
+\end{itemize}
 overload resolution to select the constructor
-for the copy is first performed as if the object were designated by an rvalue.
+for the copy is first performed as if the object were designated by an
+rvalue.
 If the first overload resolution fails or was not performed,
 or if the type of the first parameter of the selected
 constructor is not an rvalue reference to the object's type (possibly cv-qualified),
@@ -3224,6 +3230,8 @@ of whether copy elision will occur. It determines the constructor to be called i
 elision is not performed, and the selected constructor must be accessible even if
 the call is elided.
 \end{note}
+
+\pnum
 \begin{example}
 \begin{codeblock}
 class Thing {


tkoeppe:
@zygoloid: This looks like your area.

source/special.tex
```diff
@@ -3231,17 +3231,20 @@
 \end{example}
 
 \pnum
-When the criteria for elision of a copy/move operation are met,
+Overload resolution to select the constructor
+for the copy is first performed as if the object were designated by an
```
tkoeppe: @jensmaurer: Ping.

jensmaurer: @tkoeppe: Pong.

jensmaurer:
Updated.

tkoeppe:
This needs to be rebased for some reason.

jensmaurer:
Rebased.

source/special.tex
```diff
@@ -3203,17 +3203,23 @@
 \end{example}
 
 \pnum
-When the criteria for elision of a copy/move operation are met,
-but not for an \nonterminal{exception-declaration},
-and the object
-to be copied is designated by an lvalue,
-or when the \grammarterm{expression} in a \tcode{return} statement
+In the following copy-initialization contexts, a move constructor might be used instead of a copy constructor:
```
zygoloid: I don't think this is entirely accurate: the constructor we end up using might not be a move constructor (in particular, it might be instantiated from a constructor template). Can we make this a little more vague (given that the precise rules are later)?

source/special.tex
```diff
@@ -3203,17 +3203,23 @@
...
+\item If the \grammarterm{expression} in a \tcode{return} statement~(\ref{stmt.return})
 is a (possibly parenthesized) \grammarterm{id-expression}
 that names an object with automatic storage duration declared in the body
 or \grammarterm{parameter-declaration-clause} of the innermost enclosing
-function or \grammarterm{lambda-expression},
+function or \grammarterm{lambda-expression}, or
+
+\item if the operand of a \grammarterm{throw-expression}~(\ref{expr.throw})
+is the name of a non-volatile automatic object
+(other than a function or catch-clause parameter)
```
zygoloid: Interesting, I agree this is a correct rephrasing of the rule, but it sure does seem strange that this would not apply to function and catch-clause parameters. May be worth filing a CWG issue, even if it only gets dumped to P3.

source/special.tex
```diff
@@ -3203,17 +3203,23 @@
...
+whose scope does not extend beyond the end of the innermost enclosing
+\grammarterm{try-block} (if there is one),
```
zygoloid: This also seems like a defect (perhaps part of the same one): the *try-block* includes the *handler*s, which means you don't get to do copy-to-move conversion in this case, where it would obviously be safe:

```
void f() {
  try { /*...*/ }
  catch (my_exception a) {
    if (cond) throw a; // no move here =(
    my_wrapper_exception b(a);
    throw b; // no move here either =(
  }
}
```

There doesn't seem to be a technical reason to not move in the above cases.

source/special.tex
zygoloid: Maybe:

  "In the following copy-initialization contexts, a move operation might be used instead of a copy operation:"

We say "A copy operation [...] may be [...] converted to a move operation" in [stmt.return]/2, and mention copy/move operations in [class.copy.elision]/1 and in a few other contexts where we don't necessarily mean a copy/move constructor is used, but do mean a constructor that's constructing from another object of the same type, so this seems like consistent use of terminology (even if we don't define the term).

jensmaurer: Fixed.

source/special.tex
jensmaurer: I might be dense here, but "b" is an automatic object whose scope does not extend beyond the innermost enclosing "try", right? So why don't I get a move for "b"?

jensmaurer:
Updated for "copy/move operation".

source/special.tex
jensmaurer: If we have a try...catch in the same function, we surely can't move away the function parameter or catch-clause. For the function parameter, we could say that we need to be in the outermost try-block (if any) to be able to move from a function parameter. For a cache-clause parameter, maybe the requirement "scope does not extend beyond..." already covers it.

source/special.tex
jensmaurer: Send e-mail to -core pointing out these issues.

source/special.tex
zygoloid: I think the "scope does not extend beyond" rule would already do the right thing in all the relevant cases.

source/special.tex
zygoloid: Err, right, bad example, a correct example would be more like this:
```
void f() {
  exception_thrown_on_failure f;
  try { /*...*/ }
  catch (...) {
    throw f; // no move here =(
  }
}
```
I see no reason we couldn't use a move here. Anyway, yes, I'll take this to the core reflector.

source/special.tex
zygoloid: Wonderful. Every possible combination of can/cannot move function parameters and can/cannot move catch-clause parameters is provided by one of the big four implementations.

tkoeppe: @zygoloid: Isn't that what they mean by "has it been implemented"? :-)

PR #1097, opened 2016-11-22
[styles] Give lists uniform vertical spacing.

tkoeppe:
This change makes it so that the vertical space surrounding lists is independent of whether there is a paragraph break before the list (by setting partopsep to zero).

Moreover, all nested lists now use the same vertical spacing, which in turn is the same as the paragraph spacing.

There is a bug in memoir that may cause second-level lists (lists within a top-level list) to be rendered too tightly (because the itemsepii length is ignored). This can be fixed locally, though.

### DIFF
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -353,7 +353,7 @@
 %%--------------------------------------------------
 %% Indented text
 \newenvironment{indented}
-{\list{}{}\item\relax}
+{\list{}{\setlength{\parsep}{\parskip}}\item\relax}
 {\endlist}
 
 %%--------------------------------------------------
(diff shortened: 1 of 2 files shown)


tkoeppe:
@zygoloid, @burblebee: I extracted just the list changes; please have a look.

tkoeppe:
@zygoloid: This change should avoid all the problems with the previous proposal. The question now boils down to a simpler one of whether we like the tigher and more uniform list spacing.

There are two reasons that make me think so:

1. The ease of editing, not having to worry about awkward paragraph breaks to trigger negative spacing.
2. Lists within paragraphs look better with this, because the list now at least the list doesn't cause *more* whitespace inside a single paragraph than a paragraph break would. Given that our lists have "paragraph numbers", it feels apppropriate to space them exactly like paragraphs.

Please remember that second-level lists may look too tight (just as in the status quo in fact) unless you fix up your memoir.cls.

tkoeppe:
I'm closing this PR i favour of PF #1134, which uses the `enumitem` package and does not suffer from the memoir bug.

tkoeppe:
image

PR #1101, opened 2016-11-22
[iterator.requirements.general] Rephrase definition of 'writable to'

### LINKED ISSUE #698: Fix the definition of "writable to"
The definition of "writable" to is problematic as Richard explains below, but also the definition should be for "writable to", not just "writable", and it should be indexable.

e-mail from Richard:

> It's defined in [iterator.requirements.general]/1: "All output iterators support the expression *i = o where o is a value of some type that is in the set of types that are writable to the particular iterator type of i."
> 
> That said, this wording seems fairly poor: it suggests that *i = o only has to work for one type in the set of "writable to" types, not for _all_ such types, and is backwards: this is defining what "writable to" means, not adding a constraint on output iterators. Perhaps it would be clearer as:
> 
> "All output iterators i have a set of types that are writable to the iterator; for each such type T, the expression *i = o is valid where o is an expression of type T."

jensmaurer:
for output iterators.

Fixes #698.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -37,28 +37,24 @@ Iterators are a generalization of pointers that allow a \Cpp program to work wit
 To be able to construct template algorithms that work correctly and
 efficiently on different types of data structures, the library formalizes not just the interfaces but also the
 semantics and complexity assumptions of iterators.
-All input iterators
+An input iterator
 \tcode{i}
-support the expression
+supports the expression
 \tcode{*i},
 resulting in a value of some object type
 \tcode{T},
 called the
 \term{value type}
 of the iterator.
-All output iterators support the expression
-\tcode{*i = o}
-where
-\tcode{o}
-is a value of some type that is in the set of types that are
-\term{writable}
-to the particular iterator type of
-\tcode{i}.
-All iterators
+An output iterator \tcode{i} has a non-empty set of types that are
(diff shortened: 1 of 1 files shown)


source/iterators.tex
```diff
@@ -46,14 +46,10 @@
...
 of the iterator.
-All output iterators support the expression
-\tcode{*i = o}
-where
-\tcode{o}
-is a value of some type that is in the set of types that are
-\term{writable}
-to the particular iterator type of
-\tcode{i}.
+All output iterators \tcode{i} have a non-empty set of types that are
```
tkoeppe: I don't think the plural works here. How about "An output iterator `i` has a non-empty set..."?

jensmaurer:
I've pushed an update that addresses your concern, @tkoeppe.  I do want to point out that the surrounding text uses "All ... iterators...", although I do agree singular would be better everywhere.

tkoeppe:
Indeed it does... maybe we can change those ambient uses, too? They sound pretty weird at the moment. Cross-check with C++98 if this got grandfathered in; I'd be happy to improve the text at this point.

jensmaurer:
Here you are. I haven't checked C++98 yet, but the LaTeX there was one-word-per-line, which suggests troff heritage.

tkoeppe:
Thanks, much better!

PR #1102, opened 2016-11-23
[basic.scope.class] Use a regular itemize environment instead of enumeraten

Eelis:
diffs

PR #1105, opened 2016-11-23
[macros, basic, streambuf] Retire enumeraten environment in favour of…

Eelis:
Only visual difference between enumeraten (which is only used twice) and enumerate is that the former shows item numbers as "1)" while the latter shows them as "1."

Also see #1102


tkoeppe:
@zygoloid, @jwakely: Any objections to this simplification?

jwakely:
No objection from me.

tkoeppe:
Retired with extreme prejudice.

PR #1106, opened 2016-11-23
[basic.lookup.unqual] Rephrase unqualified lookup in a function defin…

### LINKED ISSUE #451: [basic.lookup.unqual]/6 is hard to parse
> A name used in the definition of a function following the function’s _declarator-id_ that is a member of namespace `N`

What does "that is a member of namespace `N`" refer to, the name or the function?

Normal English grammar would suggest it refers to the name, but the rest of the paragraph only seems to make sense if it refers to the function.

If I'm reading it correctly (which I'm still not sure about) "following the function's declarator-id" refers to where the name is used, but "is a member of namespace N" refers to the function, and yet neither of those clauses is adjacent to the thing they refer to!

This is like "The grass, in my garden, which is green, behind my house, ..."

jensmaurer:
…ition.

Fixes #451.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -1183,13 +1183,13 @@ any function or class shall be declared before its use in that namespace
 or before its use in a namespace enclosing its namespace.
 
 \pnum
-A name used in the definition of a function following the function's
+In the definition of a function that is a member of namespace \tcode{N},
+a name used after the function's
 \grammarterm{declarator-id}\footnote{This refers to unqualified names
 that occur, for instance, in
 a type or default argument in the
 \grammarterm{parameter-declaration-clause} or used in the function body.}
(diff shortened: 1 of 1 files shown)


source/basic.tex
```diff
@@ -1183,13 +1183,15 @@
 or before its use in a namespace enclosing its namespace.
 
 \pnum
-A name used in the definition of a function following the function's
+In the definition of a function that is a member of namespace \tcode{N}
+(where, only for the purpose of
```
zygoloid: This change looks like a significant improvement to me already, but can we also remove this parenthetical? The global scope *is* a namespace ([basic.scope.namespace]/3), so we don't need to say that it's one here just for exposition.

jensmaurer:
Parens removed.

PR #1107, opened 2016-11-23
[sequence.reqmts] Remove redundant 'forward iterator' requirement

### LINKED ISSUE #461: 23.2.3[sequence.reqmts]/5 is redundant with Table 96
`23.2.3[sequence.reqmts]/5` says "`iterator` and `const_iterator` types for sequence containers shall be at least of the forward iterator category", but Table 96 up in `23.2.1[container.requirements.general]` already said "any iterator category that meets the forward iterator requirements" for both `X::iterator` and `X::const_iterator`, for all containers. Shouldn't it be said only once?

The redundancy was [introduced](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2008/n2798.pdf) in 2008, when [n2776](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2008/n2776.pdf) was applied to the working draft, changing general container requirements from input iterators to forward iterators (then "ForwardIterator concept")

jensmaurer:
for sequence containers.

Any container is nowadays required to have forward iterators;
see the table entries for X::iterator and X::const_iterator in
[container.requirements.general].

Fixes #461.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -926,12 +926,6 @@ The complexities of the expressions are sequence dependent.
 \end{libreqtab3}
 
 \pnum
-\tcode{iterator}
-and
-\tcode{const_iterator}
-types for sequence containers shall be at least of the forward iterator category.
-
-\pnum
 The iterator returned from
 \tcode{a.insert(p, t)}
 points to the copy of


jwakely:
Looks good to me.

PR #1111, opened 2016-11-23
[streambuf.virtuals] Simplify the logic of exposition; remove several unneeded lists

tkoeppe:
Some of the logic in this section is needlessly convoluted. A slight rewrite makes the exposition more compact and fluent.

Before:

image

After:

image

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -3711,48 +3711,32 @@ to indicate failure.
 \pnum
 The
 \term{pending sequence}
-of characters is defined as the concatenation of:
-\begin{enumeratea}
-\item
-If
-\tcode{gptr()}
-is non-null, then the
+of characters is defined as the concatenation of
+\begin{itemize}
+\item the empty sequence if \tcode{gptr()} is null, otherwise the
 \tcode{egptr() - gptr()}
 characters starting at
 \tcode{gptr()},
-otherwise the empty sequence.
+followed by
 \item
-Some sequence (possibly empty) of characters read from the input sequence.
-\end{enumeratea}
+some (possibly empty) sequence of characters read from the input sequence.
+\end{itemize}
 
 \pnum
 The
 \term{result character}
 is
-\begin{enumeratea}
-\item
-If the pending sequence is non-empty, the first character of the sequence.
-\item
-If the pending sequence
-is
-empty then the next character that would be read from the input sequence.
-\end{enumeratea}
+the first character of the pending sequence if it is non-empty,
+otherwise
+the next character that would be read from the input sequence.
 
 \pnum
 The
 \term{backup sequence}
-is defined as the concatenation of:
-\begin{enumeratea}
-\item
-If
-\tcode{eback()}
-is null then empty,
-\item
-Otherwise the
+is the empty sequence if \tcode{eback()} is null, otherwise the
 \tcode{gptr() - eback()}
 characters beginning at
 \tcode{eback()}.
-\end{enumeratea}
 
 \pnum
 \effects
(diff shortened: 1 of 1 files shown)


tkoeppe:
@jwakely, @jensmaurer: Reviews welcome :-)

source/iostreams.tex
```diff
@@ -3712,97 +3712,73 @@
...
-\begin{enumeratea}
+are non-null then the function is not constrained as to their contents, but the ``usual backup condition'' is that either
+\begin{itemize}
 \item
-If the backup sequence contains at least
+the backup sequence contains at least
 \tcode{gptr() - eback()}
-characters, then the
+characters, and then the
 \tcode{gptr() - eback()}
```
jensmaurer: "and then" -> "in which case"

jensmaurer:
This seems to be correct (i.e. just a reshuffling of words), but I'd appreciate a second opinion.

That said, this section is a prime example for the following unrelated defects (that should be fixed in a separate pull request):
 - \term instead of \defn
 - \pnum in the middle of library descriptions, where it's unclear whether the paragraph introduces something new or is part of the previous item.

jensmaurer:
Btw, this talk about "backup sequence" seems slightly rubbish. We define the backup sequence to be gptr() - eback() characters starting at eback(). And then, in the "usual backup conditions" p13, we talk about "agreeeing" of the backup sequence with exactly these characters (first bullet). I'm confused.

tkoeppe:
I can change the `\term`s to `\defn`s, but do we really want those terms to be globally defined concepts? Maybe we can index them as "streambuf!pending sequence"?

jensmaurer:
Re: \term to \defn: As I said, not in this pull request. For instance, we'll need to check the scope of those terms. If they're just local terms (used in these five paragraphs), maybe we do not want them to be italics to start with. Or we might need a \localterm thing that keeps the italics, doesn't raise a red flag with a "grep -w term", and (optionally?) allows an index entry.

tkoeppe:
Makes sense, thanks. Do you want me to do anything about the "backup sequence" business? I didn't check the non-local context of the section. My goal was to get rid of the `enumeratea` environment.

jensmaurer:
Don't mess with the "backup sequence" or anything else in that section other than shuffling words around. Essentially, "don't ask, don't tell", I'd say. You seem to have achieved the goal of getting rid of the "enumeratea" environment, so let's leave it at that for now. Anything beyond that probably requires an editorial and/or LWG review of all of iostreams.

tkoeppe:
Agreed, thanks!

jwakely:
Late to the party, but I like this.

tkoeppe:
Thanks! This is part of the battle against enumerations :-)

jwakely:
@jensmaurer I _think_ the usual backup condition stuff is confusing because it talks about the post-Effects backup sequence in terms of the previous backup sequence. If I understand it correctly, "the backup sequence" is defined as the values in the original range `[eback(), gptr())`, so imagine that was `memcpy`'d out to some side buffer for later use, call it `buf[len]` where `len` is `gptr() - eback()`. Then the Effects are saying that the new contents of `[eback(), gptr())` are `[buf, buf+n)` where `n` is `min(gptr() - eback(), len)`

i.e. if the previous backup sequence was empty, or was only a few characters, the final backup sequence can be less than `gptr() - eback()` (so isn't required to be filled with characters plucked from thin air).

But we should ask an iostreams expert, maybe it needs an LWG issue, or maybe we just need to clarify that "backup sequence" means "ye olde backup sequence"

Independent of that, I've created pull #1157 for some other tweaks to the `underflow` wording. Please take a look.

PR #1118, opened 2016-11-24
[locale.facet] Don't bother itemizing a single item.

Eelis:
First two pages of diffs:

diffs

Rest is reflow.


tkoeppe:
(Note that diffpdf has different comparison modes; there's a more aggressive "visual" mode, too.)

PR #1121, opened 2016-11-24
[dcl.decl] Turn very large footnote into ordinary note.

Eelis:
This footnote
- is very large
- has very dense vertical layout in the pdf (as a result of being a footnote)
- uses unusual commands for its code formatting (\indent for indentation, \textrm for comments) 
- forgets to escape curly brackets in its tcode fragments, so they're missing in the pdf
- uses unusual commands for its prose formatting (\noindent)
- has unusual code formatting in the pdf (non-italic comments)

Some or all of these issues might be addressable individually, but the easiest way to fix them all in one fell swoop is to just turn this weird creature into a normal note that needs no special treatment or commands. :)

diff

(I tried diffpdf's other comparison modes, but they produced even less useful coloring.)

### DIFF
--- a/source/declarators.tex
+++ b/source/declarators.tex
@@ -55,51 +55,45 @@ initializers are discussed in~\ref{dcl.init} and~\ref{class.init}.
 \pnum
 Each
 \grammarterm{init-declarator}
-in a declaration is analyzed separately as if it was in a declaration by
-itself.\footnote{A declaration with several declarators is usually equivalent
-to the corresponding sequence of declarations each with a single declarator.
-That is
-
-\tcode{T  D1, D2, ... Dn;}
-
-\noindent is usually equivalent to
-
-\tcode{T  D1; T D2; ... T Dn;}
-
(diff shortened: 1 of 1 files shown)


source/declarators.tex
```diff
@@ -55,51 +55,45 @@
...
+\begin{codeblock}
+T D1, D2, ... Dn;
+\end{codeblock}
+is usually equivalent to
+\begin{codeblock}
+T D1; T D2; ... T Dn;
+\end{codeblock}
+where \tcode{T} is a \grammarterm{decl-specifier-seq}
+and each \tcode{Di} is an \grammarterm{init-declarator}.
 An exception occurs when a name introduced by one of the
```
zygoloid: If we're promoting this to a note, the bar for clarity and correctness is a little higher, and we shouldn't be saying "an exception occurs when" unless we mean that an exception is thrown. How about "One exception is when" here and "Another exception is when" below?

PR #1135, opened 2016-11-27
[macros] Reduce whitespace at the end of code blocks

tkoeppe:
@zygoloid, @jwakely: Please have a look. I think we can afford to lose that whitespace.


tkoeppe:
Before:

image


After:

image

PR #1137, opened 2016-11-28
[basic.scope.class] Break up list

tkoeppe:


languagelawyer:
The removal of
> The following rules describe the scope of names declared in classes.

seems to change the meaning of the fourth and the fifth paragraphs: now they talk about all names  whose scope cover a class or past the end of a class definition (e.g. a global variable declared before the class definition), not only names declared in classes.
This looks like a normative change which was applied editorially.

tkoeppe:
Thanks, that's a good point. Perhaps we should insert "The potential scope of a declaration <ins>in a class</ins> that extends ..." to p4? @zygoloid?

zygoloid:
Yes, I agree.

PR #1140, opened 2016-11-28
[except.spec] missing linebreak

alfmin:
It seems "noexcept throw()" is ment to be split as different possibilities


tkoeppe:
Thank you!

Issue #1139, opened 2016-11-28
Format indexed variables with code and maths properly


tkoeppe:
We still have a number of "indexed" variables, like `Ti` or `TiD`, that should be formatted with the variable name in code and the index in maths, either `\tcode{T}$_i$` or `$\mathtt{T}_i$` or `$\mcode{T}_i$` provided someone adds an `\mcode` macro, or `$\text{\tcode{T}}_i$`.

Please also look out for concrete indexes like `$T_0$` etc.

tkoeppe:
48efdfe014549e6d116ce12ca427f7df0e0f8c70 contributed to this, but there's more work left to be done in [tuple.tuple].

jensmaurer:
Waiting for #1142 to be applied to avoid merge conflicts.

tkoeppe:
c2144122667f95b65859887012c8bfed4b795225 also contributed to this.

jensmaurer:
@tkoeppe: Do you have a suggestion how to format `TiD = decay_t<Ti>` ? My guess would be `T_i^D = decay_t<T_i>`, but maybe you've got another idea.

tkoeppe:
@jensmaurer: How about `TD_i = decay_t<T_i>`? It's very unusual to have the index in the middle.

zygoloid:
Any further problems can be addressed by point issues; having an omnibus issue is no longer valuable.

PR #1142, opened 2016-11-29
s/(possibly cv-qualified)/\\cv{}~/ and allied changes.

W-E-Brown:
As previously discussed with the Project Editor, changed (mainly in <type_traits> tables but also in other places) occurrences of "(possibly cv-qualified)" followed by a type (usually void, but also occasionally bool, etc.) by the \cv{} macro.  To avoid a line break after \cv{}, replaced the space by a tilde (i.e., by a non-breaking space).

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -400,7 +400,7 @@ to an object of type \tcode{T}~(\ref{conv.lval}), or
 type \tcode{T} (Clause~\ref{conv}, \ref{expr.type.conv},
 \ref{expr.dynamic.cast}, \ref{expr.static.cast}, \ref{expr.cast}), or
 \item an expression that is not a null pointer constant, and has type
-other than \cv{} \tcode{void*}, is converted to the type pointer to \tcode{T}
+other than \cv{}~\tcode{void*}, is converted to the type pointer to \tcode{T}
 or reference to \tcode{T} using a standard conversion
 (Clause~\ref{conv}), a \tcode{dynamic_cast}~(\ref{expr.dynamic.cast}) or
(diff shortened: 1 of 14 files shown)


tkoeppe:
@jensmaurer: Could you please take a look? Weren't you working in that field recently?

source/basic.tex
```diff
@@ -401,7 +401,7 @@
 type \tcode{T} (Clause~\ref{conv}, \ref{expr.type.conv},
 \ref{expr.dynamic.cast}, \ref{expr.static.cast}, \ref{expr.cast}), or
 \item an expression that is not a null pointer constant, and has type
-other than \cv{} \tcode{void*}, is converted to the type pointer to \tcode{T}
+other than \cv{}~\tcode{void*}, is converted to the type pointer to \tcode{T}
 or reference to \tcode{T} using a standard conversion
```
jensmaurer: In general (not a comment on this change), instead of "cv void *", I'd like to see phrasing such as "pointer to cv void" (in double quotes). The latter is the "official" type language introduced in 8.3, and we should use it, in particular when we do not-quite-C++ things such as \cv with it.

jensmaurer:
This looks good to me, i.e. this patch does what the description above says. The bulk of the changes is adding non-breaking spaces between \cv{} and the stuff that follows; this has the chance of introducing more overfull/underfull hboxes, but we can fix those later.

It would be good to get this in sooner rather than later, because the changes overlap with the changes planned for #1139.

W-E-Brown:
Rebased. Anything else for me to do?

W-E-Brown:
Ping.

PR #1146, opened 2016-11-29
[containers] Add a space after "Requires:"

jlaire:


tkoeppe:
Can you please attach a screenshot of the diffpdf?

jlaire:
I will try to do that later today. I just noticed that the spacing looks bad and all other uses of `\requires` inside tables are followed by the escaped space.

jwakely:
Where is the spacing bad? I see no changes from this change (visually, or using diffpdf), so it doesn't seem to have any effect.

tkoeppe:
Indeed, since the macro uses xspace I expect there to not be an effect. I also don't see anything wrong with the status quo.

jlaire:
Darn, I was reading n4606.pdf. It's getting quite old compared to master.

tkoeppe:
@jlaire: the current working paper is n4618.pdf.

PR #1148, opened 2016-11-29
[whole standard] Put commas outside quotes

jlaire:
Grep shows that putting commas and periods outside quotes is much more common.


tkoeppe:
Putting small punctuators inside quotes is an old American tradition (so that the flimsy letters don't break off when they were on the outside of the block). Given the dominant tradition we're following is that of our US friends, I will leave them for them to decide.

@zygoloid: Any preferences?

tkoeppe:
Majority precedence seems like a good enough reason.

Issue #516, opened 2015-07-02
[tuple.helper] Inconsistent formatting of BaseCharacteristic etc.


jwakely:
[tuple.helper] uses code font for UnaryTypeTrait, BaseCharacteristic and TransformationTrait, I think it should be the normal font.

jwakely:
Same problem in [meta.unary.prop.query]

tkoeppe:
@jwakely: Can you fix this? I imagine you have an idea how it should look like, and how we do it elsewhere?

AlisdairM:
I believe that this issue was resolved by the terminology introduced for C++20 concepts, if not before, and can be closed now.

PR #1151, opened 2016-11-29
[utilities] Use code font for named requirements

jlaire:
https://github.com/cplusplus/draft/issues/516

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -7345,7 +7345,7 @@ template <class T, class Alloc> struct uses_allocator;
 \begin{itemdescr}
 \pnum
 \remarks Automatically detects whether \tcode{T} has a nested \tcode{allocator_type} that
-is convertible from \tcode{Alloc}. Meets the BinaryTypeTrait
+is convertible from \tcode{Alloc}. Meets the \tcode{BinaryTypeTrait}
 requirements~(\ref{meta.rqmts}). The implementation shall provide a definition that is
 derived from \tcode{true_type} if the \grammarterm{qualified-id} \tcode{T::allocator_type}
 is valid and denotes a type~(\ref{temp.deduct}) and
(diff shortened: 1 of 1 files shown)


tkoeppe:
@jwakely: Looks like your area.

jlaire:
`TrivialClock` is similar, and it's always in code font. A global decision about what to do with these would be useful.

zygoloid:
Named sets of requirements are usually formatted in code font. See [utility.requirements] for a bunch of examples, in addition to `TrivialClock`. UnaryTypeTrait, BinaryTypeTrait, and TransformationTrait are outliers here.

I would be fine with having a different typographical convention for them, but it should be applied consistently.

jlaire:
I'll update this PR to change them to code font for consistency tomorrow.

jwakely:
I agree these concept-ish terms should all be typographically distinct, and currently we do that with code font.

I think the one occurrence of "Clock" in code font is wrong, IMHO "Clock requirements" is more like "Associative container requirements" than a concept like "CopyConstructible". (Although it could be argued the other way, `Lockable` and `TimedLockable` are counterexamples.)

As #1154 says, BaseCharacteristic probably shouldn't be one of these terms, so don't change that.

jwakely:
And then there's "Allocator requirements" which is sometimes code font and sometimes not.

jlaire:
I'm not sure if the \defn instances should have \tcode inside them, too.

tkoeppe:
@jwakely: Are you waiting for some changes from the author here?

tkoeppe:
Can you please rebase this? I think we're ready to apply this.

jlaire:
Rebased. "Allocator requirements" and some others are still inconsistent like pointed out by @jwakely, but I don't have time to look into them at the moment. I think they can be done in a separate PR.

PR #1153, opened 2016-11-29
Do not use code font for constexpr constructor and constexpr function

### LINKED ISSUE #825: "constexpr constructor" and "constexpr function" are defined terms and should not be formatted in code font
Some of the references to the defined terms "constexpr constructor" and "constexpr function" incorrectly use code font for the "constexpr" keyword. (We probably make this mistake for some other terms that include keywords, too.)

jensmaurer:
Fixes #825.

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -2746,7 +2746,7 @@ with two or more member functions with equivalent behavior,
 or add additional signatures for a member function name.
 \end{note}
 
-\rSec3[constexpr.functions]{\tcode{constexpr} functions and constructors}
+\rSec3[constexpr.functions]{Constexpr functions and constructors}
 
 \pnum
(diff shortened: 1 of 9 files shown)


source/lib-intro.tex
```diff
@@ -2746,7 +2746,7 @@
 or add additional signatures for a member function name.
 \end{note}
 
-\rSec3[constexpr.functions]{\tcode{constexpr} functions and constructors}
+\rSec3[constexpr.functions]{constexpr functions and constructors}
```
tkoeppe: Since this is now a normal word, should it be capitalized?

jensmaurer: Yes.

jensmaurer:
Capitalized "constexpr" in section heading.

PR #1157, opened 2016-11-30
[streambuf.virt.get] use half-open ranges in underflow effects

jwakely:
underflow1
underflow2

We should discuss this before merging, I'm ambivalent about it.

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -3726,9 +3726,8 @@ The
 of characters is defined as the concatenation of
 \begin{itemize}
 \item the empty sequence if \tcode{gptr()} is null, otherwise the
-\tcode{egptr() - gptr()}
-characters starting at
-\tcode{gptr()},
+characters in
+\range{gptr()}{egptr()},
 followed by
 \item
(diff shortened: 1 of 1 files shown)


jensmaurer:
Oh well. This seems to be a presentational improvement. Is that the only place in streambufs where we have that style of description? I wouldn't want to address only one of possibly N adjacent places.
[I still find this "backup sequence" stuff totally obtuse, but that's a different story.]

zygoloid:
I like it. Looks good to merge once @tkoeppe's comment is addressed.

jwakely:
Rebased and added the comma.

jwakely:
I should probably also look at `overflow` in [streambuf.virt.put] which uses a similar style:

> the the `pptr() - pbase()` characters beginning at `pbase()`

tkoeppe:
@jwakely: Yes, sounds good. We don't come past this section very often, so getting all the cleanup done in one go seems like a good idea

zygoloid:
Taking this part for C++17; I'd rather get this in than wait for more cleanups to be added to it.

PR #1158, opened 2016-11-30
[expr.prim.lambda] Split specification of lambda expressions into sub…

### LINKED ISSUE #1155: [expr.prim.lambda] p 6, 7 obnoxiously long
Paragraphs 6 and 7 of [expr.prim.lambda] are very long. P 7 contains 2 notes and 3 examples!

Could we please rewrite those to become multiple paragraphs? Consider having long examples be paragraphs on their own.

@jensmaurer, @zygoloid, any takers?

jensmaurer:
…sections.

Fixes #1155.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -602,45 +602,6 @@ by the \grammarterm{nested-name-specifier}.
 \end{bnf}
 
 \begin{bnf}
-\nontermdef{lambda-capture}\br
-    capture-default\br
-    capture-list\br
-    capture-default \terminal{,} capture-list
-\end{bnf}
-
-\begin{bnf}
-\nontermdef{capture-default}\br
-    \terminal{\&}\br
-    \terminal{=}
-\end{bnf}
-
-\begin{bnf}
-\nontermdef{capture-list}\br
-    capture \terminal{...\opt}\br
-    capture-list \terminal{,} capture \terminal{...\opt}
-\end{bnf}
-
-\begin{bnf}
-\nontermdef{capture}\br
-    simple-capture\br
-    init-capture
-\end{bnf}
-
-\begin{bnf}
-\nontermdef{simple-capture}\br
-    identifier\br
-    \terminal{\&} identifier\br
-    \terminal{this}\br
-    \terminal{* this}
-\end{bnf}
-
-\begin{bnf}
-\nontermdef{init-capture}\br
-    identifier initializer\br
-    \terminal{\&} identifier initializer
-\end{bnf}
-
-\begin{bnf}
 \nontermdef{lambda-declarator}\br
     \terminal{(} parameter-declaration-clause \terminal{)} decl-specifier-seq\opt\br
     \hspace*{\bnfindentinc}noexcept-specifier\opt attribute-specifier-seq\opt trailing-return-type\opt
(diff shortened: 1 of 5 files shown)


jensmaurer:
I have tried to restrain myself to reshuffling text and introducing subsection headings.

source/expressions.tex
```diff
@@ -706,14 +706,35 @@
...
-This class type is not an aggregate type~(\ref{dcl.init.aggr}).
+whose properties are described in section~\ref{expr.prim.lambda.closure}.
+
+\pnum
+If a \grammarterm{lambda-expression} does not include a
+\grammarterm{lambda-declarator}, it is as if the \grammarterm{lambda-declarator} were
+\tcode{()}.
+The lambda return type is \tcode{auto}, which is replaced by the
+type specified by the
+\grammarterm{trailing-return-type} if provided and/or deduced from
```
tkoeppe: How about just "or" instead of "and/or"?

jensmaurer: This is unchanged text, just moved around. Not my invention.

source/expressions.tex
```diff
@@ -788,7 +792,9 @@
 q();                                                   // OK: outputs \tcode{1a3.14}
 \end{codeblock}
 \end{example}
-This function call operator or operator template is declared
+
```
tkoeppe: (I cannot comment on unchanged code, so this comment is out of place.)

The example seems to be specific to packs only, not to the entire paragraph. What do you think of the following: In lines old-762/new-766, take the sentence "The invented type ..." out and end the paragraph after the next sentence, just before the example, Then put the removed sentence back at the start of the now new paragraph, followed by the example. That way the new paragraph concentrates on one specific detail (packs), together with example.

jensmaurer: First, the example also has non-pack things, e.g. glambda at the top.
Second, "The invented type template-parameter ..." refers to the template-parameter introduced in the preceding sentence. Moving this to a separate paragraph, with something else in between, would lose the connection.

tkoeppe: OK, the example is a good point. I wouldn't find losing the connection too worrying in general; I think it's acceptable for later things to refer to earlier things. That's more of a judgement call. But we can keep it as is.

jensmaurer: Yes, but if you read the intermediate sentence, it goes off on a tangent, and returning to the template-parameter later is odd.

jensmaurer:
Taking no action on the review comments, as explained above.

@zygoloid: You're next.

jensmaurer:
I've updated some cross-references to now refer to a more specific subsection, where appropriate.

jensmaurer:
No-change rebase.

zygoloid:
1. [expr.prim.lambda]p2 seems out of place, and breaks the natural flow from p1 to p3. Maybe we can move it down next to where we talk about how these *decl-specifier*s map into properties of the closure type's `operator()`? Or if not there, then at least to after p5.

2. I think [expr.prim.lambda]p4 would be better as the first paragraph of [expr.prim.lambda.closure], as it's a good summary of what we're about to describe in there, slightly tweaked to something like:

> The type of a *lambda-expression* (which is also the type of the closure object) is a unique, unnamed non-union class type, called the *closure type*, whose properties are described below.

3. The start of [expr.prim.lambda.capture] is quite jarring. We could do with an introductory sentence explaining what on earth a capture is before we get into the nitty-gritty details. Maybe something like:

> The body of a *lambda-expression* may refer to automatic storage duration variables and the `*this` object of enclosing block scopes by capturing those entities, as described below.

4. We might also want to consider moving the grammar description of explicit captures down into the captures subclause.

jensmaurer:
re 1: Moved normative sentence from [expr.prim.lambda]p2 to the start of [expr.prim.lambda.closure]p3. Moved example as a new paragraph to after [expr.prim.lambda.closure]p3.

re 2: Done.

re 3: Done.  I'm not super-happy with the "*this object". I added "(if any)", since we might not be in a member function.

re 4: Why just the explicit captures? Seems like all of the capture grammar could go there. Done.

jensmaurer:
Subject: 	Re: editorial restructuring of [class.copy]
Date: 	Tue, 31 Jan 2017 14:56:02 -0500
From: 	William M. (Mike) Miller <william.m.miller@gmail.com>

[...]
The ones for expr.prim.lambda are involved enough that I think I'd like to have CWG take a look.

jensmaurer:
Uploaded to CWG wiki for Kona, as agreed with Mike Miller:
[expr.prim.lambda](http://wiki.edg.com/pub/Wg21kona2017/CoreWorkingGroup/expr-prim-lambda.pdf)

jensmaurer:
CWG suggested to move the first sentence of closure types p4 to the general section and otherwise approved this change.

jensmaurer:
Rebased and implemented CWG guidance.

tkoeppe:
@jensmaurer: Rebase, just to be sure?

PR #1159, opened 2016-11-30
[limits] Integrate section [limits.numeric] into section [numeric.lim…

### LINKED ISSUE #785: Sections 18.3.2.1 and 18.3.2.3 have the same name
Both are named "Class template numeric_limits".
Maybe one of them should say "synopsis"?

jensmaurer:
…its].

Fixes #785.

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -306,54 +306,17 @@ operators~(\ref{expr.add}), Free store~(\ref{class.free}), and ISO C~7.19.
 
 \rSec1[support.limits]{Implementation properties}
 
-\rSec2[support.limits.general]{In general}
+\rSec2[support.limits.general]{General}
 
 \pnum
 The headers
-\tcode{<limits>}~(\ref{limits}),
+\tcode{<limits>}~(\ref{limits.syn}),
 \tcode{<climits>}~(\ref{climits.syn}), and
 \tcode{<cfloat>}~(\ref{cfloat.syn})
 supply characteristics of implementation-dependent
 arithmetic types~(\ref{basic.fundamental}).
 
-\rSec2[limits]{Numeric limits}
-
-\rSec3[limits.numeric]{Class template \tcode{numeric_limits}}
-
-\pnum
-The
-\indexlibrary{\idxcode{numeric_limits}}%
-\tcode{numeric_limits}
-class template provides a \Cpp program with information about various properties of
-the implementation's representation of the
-arithmetic types.
-
-\pnum
-Specializations shall be provided for each
-arithmetic type,
-both floating-point and integer, including
-\tcode{bool}.
-The member
-\tcode{is_specialized}
-shall be
-\tcode{true}
-for all such specializations of
-\tcode{numeric_limits}.
-
-\pnum
-For all members declared
-\tcode{static} \tcode{constexpr}
-in the
-\tcode{numeric_limits}
-template, specializations shall define these values in such a way
-that they are usable as
-constant expressions.
-
-\pnum
-Non-arithmetic standard types, such as
-\tcode{complex<T>}~(\ref{complex}), shall not have specializations.
-
-\rSec3[limits.syn]{Header \tcode{<limits>} synopsis}
+\rSec2[limits.syn]{Header \tcode{<limits>} synopsis}
 \indextext{\idxhdr{limits}}%
 \indexlibrary{\idxhdr{limits}}%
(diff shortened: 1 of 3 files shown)


tkoeppe:
This would be a nice cleanup. @jwakely, @zygoloid, any objections to the loss of a stable name?

jwakely:
Fine by me.

W-E-Brown:
This cleanup seems to lose the introductory nature of the first paragraph.  Perhaps let's promote the original [limits.numeric]/1 to [limits]/1 instead of moving it so far down?

jensmaurer:
@W-E-Brown: Hm.. But the introductory sentence in the first paragraph only mentions numeric_limits, but the synopsis also has float_round_style and float_denorm_style. At this point, it's not clear these are related to the numeric_limits template. That said, I do admit there is little besides numeric_limits in the [limits] section.

tkoeppe:
The entire structure of 18.3.2 is weird. Anywhere else in the Standard we would structure it like this:

* 18.3.2: Numeric something something
  * 18.3.2.1: Header `<numeric>` synopsis
  * 18.3.2.2: Class template `numeric_limits`
    * Class template definition
    * 18.3.2.2.1: `numeric_limits` members
    * 18.3.2.2.2: `numeric_limits` specializations
  * 18.3.2.3: Type `float_round_style`
  * 18.3.2.4: Type `float_denorm_style`

That is, the synopsis is the first sibling in a subclause that describes the things that are listed in the synopsis. That's how we do it for most other library components. A "General" section could optionally precede the synopsis, but only if there's something useful to say. (I think the origin of "General" sections lies in the ISO drafting directives which say never to have "mixed" content in a subclause, i.e. paragraphs *and* sub-subclauses.)

tkoeppe:
In a hypothetical world, and given that we now have full synopses for the C library, I could even imagine the following structure:

* 18.3: Implementation properties
  * 18.3.1: General
  * 18.3.2: Header `<limits>` synopsis
  * 18.3.3: Header `<climits>` synopsis
  * 18.3.4: Header `<cfloat>` synopsis
  * 18.3.5: Class template `numeric_limits`
    * 18.3.5.1: members [etc]
  * 18.3.5: Floating point rounding and extreme values
  * 18.3.6: C library

jensmaurer:
A mild issue with all the presentation options so far is that float_round_style and float_denorm_style are used in the definition of numeric_limits. It would be good to introduce/explain them _before_ the numeric_limits template. Also, when extracting the <cXXX> synopses, 18.3.6 "C library" becomes empty.

I'd also like to point out that we have a subsection "C library" in clause 27, too, and we just have the synopses under this subsection with a one-sentence "same as in C" statement. Ok, but elsewhere (e.g. in the "strings" clause), the  C library stuff is just adjoint to the C++ stuff, without particularly highlighting it's from C.

jensmaurer:
Ok, I've done something along @tkoeppe 's "hypothetical world". I like it now. (Note that I expressly only moved sections, added/removed subsection headings, and fixed cross-references.)

W-E-Brown:
No longer hypothetical, eh?  I like it.

jensmaurer:
Rebased.

source/support.tex
```diff
@@ -444,18 +498,43 @@
 \end{codeblock}
 
 \pnum
+For all members declared
+\tcode{static} \tcode{constexpr}
+in the
+\tcode{numeric_limits}
+template, specializations shall define these values in such a way
+that they are usable as
+constant expressions.
```
jensmaurer: Well, there are member functions declared "static constexpr" (such as `infinity()`), and users can specialize numeric_limits for their own types. Since a function can be declared "constexpr", yet not be usable in a constant expression, my guess is that this essentially requires that "T" doesn't access global or volatile variables when copied. (Such as copy is required for the return value of infinity(), for example.)

zygoloid: Ah, this applies to user specializations. Seems weird to put a "shall" on that, but OK.

tkoeppe:
Tip for the future: Please always check the overfull box status for your proposed change:

    fgrep -e Overfull -e .tex std.log

source/support.tex
jensmaurer: See also [LWG 2833](https://cplusplus.github.io/LWG/lwg-active.html#2833) "Library needs to specify what it means when it declares a function constexpr".

jensmaurer:
Alerted Marshall Clow (LWG chair) to this change.

PR #1160, opened 2016-12-01
[lib] Replace 'Notes:' elements with regular [ Note: ... -- end note]

### LINKED ISSUE #492: Get rid of Notes: elements in library clauses
17.5.1.4 [structure.specifications] defines what a _Notes:_ element means in the detailed specification for a function.

The only correct use of this in the entire standard is 20.12.5.7 [time.duration.cast] (which is missing a paragraph number! it should be paragraph 3).

23.2.1 [container.requirements.general] p4 has a "Notes:" following table 96 but it's not the same thing, because it's not italicized, and isn't a detailed specification where [structure.specifications] applies (so just seems wrong to me, maybe it should be "[Note: the algorithm equal() is defined in Clause 25.]"

Daniel added:

> There exits some incorrectly spelled "Notes:" written as "Note:" such
> as in 20.4.2.4p13, [tuple.elem] p4, [unique.ptr.single.observers] p5
> and maybe elsewhere, but I don't really like this additional flavour
> of notes and would happily replace them all by our usual [Note: - end
> note]

jensmaurer:
Remove the \realnotes and \realnote macros.
No longer mention 'Notes:' in [structure.specifications].

Fixes #492.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -272,13 +272,13 @@ constant                        \\
 
 \end{libreqtab5}
 
-Notes: the algorithm
-\tcode{equal()}
-is defined in Clause~\ref{algorithms}.
 Those entries marked ``(Note A)'' or ``(Note B)''
 have linear complexity for \tcode{array} and have constant complexity
 for all other standard containers.
-
+\begin{note}
+The algorithm \tcode{equal()} is defined in Clause~\ref{algorithms}.
+\end{note}
+ 
 \pnum
 The member function \tcode{size()} returns the number of elements in the container.
 The number of elements is defined by the rules of
@@ -3230,15 +3230,14 @@ void swap(array& y) noexcept(is_nothrow_swappable_v<T>);
 
 \begin{itemdescr}
 \pnum
-\effects As if by \tcode{swap_ranges(begin(), end(), y.begin())}.
+\effects Equivalent to \tcode{swap_ranges(begin(), end(), y.begin())}.
 
 \pnum
-\throws Nothing unless one of the element-wise swap calls throws an exception.
-
-\pnum
-\realnote Unlike the \tcode{swap} function for other containers, \tcode{array::swap}
+\begin{note}
+Unlike the \tcode{swap} function for other containers, \tcode{array::swap}
 takes linear time, may exit via an exception, and does not cause iterators to
 become associated with the other container.
+\end{note}
 \end{itemdescr}
 
 \rSec3[array.zero]{Zero sized arrays}
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -2838,18 +2838,22 @@ cond.notify_all();
 all objects with thread storage duration associated with the current thread.
 
 \pnum
-\realnote The supplied lock will be held until the thread exits, and care
+\begin{note}
+The supplied lock will be held until the thread exits, and care
 must be taken to ensure that this does not cause deadlock due to lock
 ordering issues. After calling \tcode{notify_all_at_thread_exit} it is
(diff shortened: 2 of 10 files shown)


tkoeppe:
Looks good, but I'll let @zygoloid handle it, since it's a fairly substantial change.

source/containers.tex
```diff
@@ -3236,9 +3237,11 @@
 \throws Nothing unless one of the element-wise swap calls throws an exception.
 
 \pnum
-\realnote Unlike the \tcode{swap} function for other containers, \tcode{array::swap}
+\begin{note}
+Unlike the \tcode{swap} function for other containers, \tcode{array::swap}
 takes linear time, may exit via an exception, and does not cause iterators to
 become associated with the other container.
```
zygoloid: Where are these properties normatively stated?

source/threads.tex
```diff
@@ -2838,18 +2838,20 @@
...
-\realnote The supplied lock will be held until the thread exits, and care
+\begin{note}
+The supplied lock will be held until the thread exits, and care
 must be taken to ensure that this does not cause deadlock due to lock
 ordering issues. After calling \tcode{notify_all_at_thread_exit} it is
 recommended that the thread should be exited as soon as possible, and
 that no blocking or time-consuming tasks are run on that thread.
 
 \pnum
-\realnote It is the user's responsibility to ensure that waiting threads
```
zygoloid: Use two separate notes here? Generally I'd like to avoid `\pnum`s within a note.

source/threads.tex
```diff
@@ -3339,6 +3341,13 @@
 \pnum\effects Unblocks all threads that are blocked waiting for \tcode{*this}.
 \end{itemdescr}
 
+\pnum
+\begin{note}
+If any of the \tcode{wait} functions exits via an exception, it is
+unspecified whether the \tcode{Lock} is held. One can use a \tcode{Lock} type
+that allows to query that, such as the \tcode{unique_lock} wrapper.
+\end{note}
```
zygoloid: I'm not entirely comfortable positioning this between the `notify*` functions and the `wait*` functions -- it seems pretty easy to miss here.

source/threads.tex
```diff
@@ -4554,13 +4560,15 @@
 
 \begin{itemdescr}
 \pnum
-\realnote as described above, the template and its two required specializations differ only in
+\begin{note}
+As described above, the template and its two required specializations differ only in
 the return type and return value of the member function \tcode{get}.
 
 \pnum
-\realnote access to a value object stored in the shared state is
```
zygoloid: Again, please avoid the `\pnum` within a note here.

source/containers.tex
jensmaurer: All I have is Note A and Note B in [container.requirements.general] after the table.

Suggestion: This says "As if by swap_ranges" for the effects. Let's change that to "Equivalent to" and we get all the missing pieces for free, and can leave the note as a note.

source/threads.tex
jensmaurer: Ok.

source/threads.tex
jensmaurer: Ok, reverted to the previous position.

source/threads.tex
jensmaurer: Ok.

jensmaurer:
Updated.

source/containers.tex
jensmaurer: (The updated version adds the "Equivalent to" a few lines earlier.)

PR #1161, opened 2016-12-01
[util.smartptr] Cleanse of italics infestation.

### LINKED ISSUE #533: [util.smartptr.shared] misuses italics
Throughout [util.smartptr.shared] terms are italicised on each use, not only when defined, e.g. [util.smartptr.shared.const]/1 uses _empty_ which is defined [util.smartptr.shared]/1

> _Effects:_ Constructs an _empty_ shared_ptr object.

jensmaurer:
Technical terms should not be in italics everywhere they are referenced.
Common terms such as 'owns' or 'empty' that are used as technical terms
in this subsection only are not italicized or entered into the index.

Fixes #533.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -9334,7 +9334,7 @@ The \tcode{shared_ptr} class template stores a pointer, usually obtained
 via \tcode{new}. \tcode{shared_ptr} implements semantics of shared ownership;
 the last remaining owner of the pointer is responsible for destroying
 the object, or otherwise releasing the resources associated with the stored pointer. A
-\tcode{shared_ptr} object is \term{empty} if it does not own a pointer.
+\tcode{shared_ptr} is said to be empty if it does not own a pointer.
 
 \begin{codeblock}
 namespace std {
@@ -9488,7 +9488,7 @@ a pointer type \tcode{T*} when either
 
 \pnum
 In the constructor definitions below,
-\term{enables \tcode{shared_from_this} with \tcode{p}},
+enables \tcode{shared_from_this} with \tcode{p},
 for a pointer \tcode{p} of type \tcode{Y*},
 means that if \tcode{Y} has an unambiguous and accessible base class
 that is a specialization of \tcode{enable_shared_from_this}~(\ref{util.smartptr.enab}),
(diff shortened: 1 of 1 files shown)


tkoeppe:
Do you not even want to retain the italics in the definition of the term? I kind of feel that it would be a bit more readable to have that meta-syntactic disctinction.

jensmaurer:
That's what I understood Richard's position to be.

In any case, we talk about "empty" in [container.requirements.general] all over the place, but we never italicize the term.

Also, we're not defining "empty", we're (at best) defining "empty shared_ptr" as the term of art here. And you don't want to look at italics \tcode text, ever.

tkoeppe:
Well, that may be so, but all the same I am now left to wonder:

> A `shared_ptr` object is empty if it does not own a pointer.

Is this a normative requirement on `shared_ptr`'s behaviour when it doesn't own a pointer? Oh, no, it's *defining* the meaning of "empty". Hm. I mean, I don't dispute that we use this style, I'm just saying that it's an ambiguity that requires additional attention.

zygoloid:
On 2 Dec 2016 7:59 am, "Thomas Köppe" <notifications@github.com> wrote:

Well, that may be so, but all the same I am now left to wonder:

A shared_ptr object is empty if it does not own a pointer.

Changing that to:

> A `shared_ptr` is said to be empty if it does not own a pointer.

would help.

Is this a normative requirement on shared_ptr's behaviour when it doesn't
own a pointer? Oh, no, it's *defining* the meaning of "empty". Hm. I mean,
I don't dispute that we use this style, I'm just saying that it's an
ambiguity that requires additional attention.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/cplusplus/draft/pull/1161#issuecomment-264488533>, or mute
the thread

tkoeppe:
I like that!

jensmaurer:
Updated as desired.

W-E-Brown:
This makes me sad. I rather like italics for terms being defined. I don't mind the "is said to be" phrasing, but for me it's not a substitute. Unless you want to have a separate "Definitions" subclause, of course.  (I don't, but could live with it.)

PR #1170, opened 2016-12-03
[macros, styles] Make BNF vertical spacing like that of normal paragraphs

tkoeppe:
Removes minpages; adds custom enumitem list.

BNF entries are now spaced like normal paragraphs. Individual paragraphs are kept on a single page by setting a high widow penalty.

It is now immaterial whether multiple nontermdefs are contained in a single bnf environment, or whether each def is in its own environment. Pages can break between elements, not within.

### DIFF
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -399,13 +399,12 @@
  \newcommand{\bnfindentfirst}{\BnfIndent}
  \newcommand{\bnfindentinc}{\BnfInc}
  \newcommand{\bnfindentrest}{\BnfRest}
- \begin{minipage}{.9\hsize}
  \newcommand{\br}{\hfill\\}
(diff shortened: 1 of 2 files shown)


jensmaurer:
At least the resulting LaTeX looks a lot cleaner... Do you have before/after screenshots?

tkoeppe:
I can make those, hang on.

tkoeppe:
Here's a small example. The diffs are pervasive, which makes diffpdf not a great tool; I recommend building your own version and seeing for yourself. (Especially long grammars like in Clause 6 are interesting in the way the widow/orphan control works.)

image

jensmaurer:
Thanks. It seems this change is not harmful.

PR #1175, opened 2016-12-07
Class members are introduced by a member-declaration, not member-de…

jabelloc:
…clarator.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -531,7 +531,7 @@ or declare at least one unnamed bit-field.
 
 \pnum
 A \defn{data member} is a non-function member introduced by a
-\grammarterm{member-declarator}.
+\grammarterm{member-declaration}.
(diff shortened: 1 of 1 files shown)


jwakely:
This seems incorrect. A member-declaration can also introduce typedef names and other things which are non-functions. Your change would mean they are considered data members.

jabelloc:
@jwakely I didn't say: "class **data members** are introduced by a _member-declaration_". What I said was:  **class members** are introduced by _member-declarations_. `typedef` names are not data members but they are class members.

But that's not relevant here. The important point is that you cannot introduce a class data member with a _member-declarator_. That doesn't make sense. You've got to use a _member-declaration_ to introduce a data member into a class.

jabelloc:
@jwakely I take back what I said earlier. You're right, and I was wrong. Class data members have to be introduced into a class with a _member-declarator_.

PR #1179, opened 2016-12-07
[function.objects] Introduce a new subsection for the <functional> sy…

### LINKED ISSUE #192: std::hash<> could have a clearer index entry
Given that there are many specializations of std::hash<>, the index should clearly point to [unord.hash] for the general description.

jensmaurer:
…nopsis.

Also add a cross-reference from [unord.hash] to the synopsis, which
is the only place where the primary template is declared.

Fixes #192.


tkoeppe:
Very nice, thanks.

Also addresses #566.

PR #1180, opened 2016-12-07
[temp.class] Reorganize examples into separate paragraph

tkoeppe:
Replaces #1177.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -1362,21 +1362,16 @@ A class
 \term{template}
 defines the layout and operations
 for an unbounded set of related types.
+
+\pnum
 \begin{example}
-a single class template
+A single class template
 \tcode{List}
-might provide a common definition for
-list of
-\tcode{int},
-list of
-\tcode{float},
-and list of pointers to
-\tcode{Shape}s.
-\end{example}
-
-\begin{example}
-An array class template might be declared like this:
-
+might provide an unbounded set of class definitions:
+one class \tcode{List<T>} for every type \tcode{T},
+each describing a linked list of elements of type \tcode{T}.
+Similarly, a class template \tcode{Array} describing a contiguous,
+dynamic array might be defined like this:
 \begin{codeblock}
 template<class T> class Array {
   T* v;
@@ -1387,16 +1382,11 @@ public:
   T& elem(int i) { return v[i]; }
 };
 \end{codeblock}
-
-\pnum
-The prefix
-\tcode{template}
-\tcode{<class}
-\tcode{T>}
+The prefix \tcode{template<class T>}
 specifies that a template is being declared and that a
 \grammarterm{type-name}
 \tcode{T}
-will be used in the declaration.
+may be used in the declaration.
 In other words,
 \tcode{Array}
 is a parameterized type with


Eelis:
Nice! I was going to submit a pull request similar to #1177 for the example spanning [class.access]#6-7, but maybe you want to reorganize that one as well?

source/templates.tex
```diff
@@ -1390,16 +1388,14 @@
...
-\pnum
 The prefix
 \tcode{template}
 \tcode{<class}
 \tcode{T>}
 specifies that a template is being declared and that a
 \grammarterm{type-name}
 \tcode{T}
-will be used in the declaration.
+will be used in the definition.
```
zygoloid: This change seems unnecessary; a template parameter can be used in a non-defining declaration too.

source/templates.tex
```diff
@@ -1375,11 +1377,7 @@
 \tcode{float},
 and list of pointers to
 \tcode{Shape}s.
```
zygoloid: This sentence seems strange; it appears to be suggesting that a `List` template might only work for some finite set of specified types. Maybe we could remove this (or rephrase it) rather than incorporating it into the somewhat-more-reasonable `Array` example.

tkoeppe: How about:

> A `List` template or an `Array` template might provide a common definition for, respectively, a dynamic linked list or a contiguous array of elements of an arbitrary type. The `Array` template might be [...]

source/templates.tex
tkoeppe: But the example doesn't do that. The example should help make things clear, not list unrelated, though un-wrong, statements. Should we maybe delete the sentence?

tkoeppe: Or we could say "in the declaration or definition".

timsong-cpp: All definitions are also declarations...?

tkoeppe: @timsong-cpp: Yes, but not all declarations are definitions. To repeat, I think the example should be *useful*, otherwise we can just remove it. This change is an attempt to make the sentence relevant to the example and help the illustraion, rather than be a general statement that happens to be true.

tkoeppe: @timsong-cpp: OK, I see what you mean. It may be sufficient to leave the status quo.

timsong-cpp: Yes, this example is a definition; it's a class template definition, too, and it's also a declaration. I'm not sure why you consider "definition" to be the right level of generality. In any event, "declaration" matches "a template is being *declared*"; just saying "definition" here without changing that seems questionable to me.

Eelis: The "WILL" part seems dubious to me. I'd go for "MAY be used in its declaration", because the sentence does sound like a general-ish description of the meaning of the "template <class T>" prefix to me.

tkoeppe: @Eelis: How about "is", since the example is indeed using it?

source/templates.tex
```diff
@@ -1390,16 +1388,14 @@
...
-\pnum
 The prefix
 \tcode{template}
 \tcode{<class}
 \tcode{T>}
 specifies that a template is being declared and that a
 \grammarterm{type-name}
 \tcode{T}
-will be used in the declaration.
+may be used in the definition.
```
timsong-cpp: "specifies that a template is being **declared** and that a *type-name* `T` may be used in the **definition**" simply doesn't make much sense to me, sorry. Similar to the argument for "may" here, at this point in time we don't know if the prefix is being followed by a definition or a (non-defining) declaration yet, so I still think "declaration" makes more sense. But at least the two highlighted words should match.

tkoeppe: OK, you're right, I'll back those changes out for now. It seems like the status quo is fine, and this isn't making it better.

source/templates.tex
tkoeppe: @zygoloid: Ping

zygoloid: I think it's lacking something -- I think we want to more explicitly say that a single `List` template can simultaneously provide definitions for lists of type `T`, for all possible types `T`.

tkoeppe: @zygoloid: How about:

> A single class template `List` might provide an unbounded set of class definitions: one class `List<T>` for each type `T` describing a linked list of elements of type `T`. Similarly, a class template `Array` describing a contiguous, dynamic array might be defined as follows:

zygoloid: Sure, let's go with that.

tkoeppe: Done, now it looks like this:

image

tkoeppe:
https://cloud.githubusercontent.com/assets/6378233/21205536/a3cc3894-c254-11e6-91c8-300ee0d6e91f.png

tkoeppe:
I just changed "The prefix `template <class T>` to "The prefix `template<class T>`" for local consistency.

PR #1181, opened 2016-12-07
[lib] Adjust whitespace to attach & to the type, not the declarator-id.

### LINKED ISSUE #166: Placement of "*" and "&" inconsistent
The prevailing style is "int& foo" and "int\* foo". I fixed pointers to be more consistent in ce29631dccd86d1e1eefff24d1dd3420c598258d; fix references too.

jensmaurer:
Fixes #166.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -286,7 +286,7 @@ namespace std {
 
   template <class ForwardIterator, class Searcher>
     ForwardIterator search(ForwardIterator first, ForwardIterator last,
-                           const Searcher &searcher);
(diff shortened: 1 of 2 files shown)


tkoeppe:
Is that all -- does that resolve the *entire* issue?!

jensmaurer:
`grep "[a-zA-Z] &[a-zA-Z]" *.tex` doesn't find anything else in the library parts.

tkoeppe:
OK, great - who'd have thought we'd ever converge on a style for this.

PR #1182, opened 2016-12-07
[class.virtual] Remove awkward 'but'.

### LINKED ISSUE #170: Wording change suggestion in function override rules
I think 10.3/11 (C++11; no change in N3690) is currently slightly rubbish:

> A virtual function declared in a class shall be defined, or declared pure (10.4) in that class, or both; but no diagnostic is required (3.2).

Though the following _can_ be deduced, I think the passage would be clearer and easier to read if it went something like this:

> A virtual function declared in a class shall be defined, or declared pure (10.4) in that class, or both; however, no diagnostic is required if neither is true (3.2).

I also fixed the start-of-phrase/post-semicolon conjunction though, admittedly, that is something of a subjective style issue.

jensmaurer:
Fixes #170.

### DIFF
--- a/source/derived.tex
+++ b/source/derived.tex
@@ -800,7 +800,7 @@ another class.
 \pnum
 \indextext{definition!virtual~function}%
 A virtual function declared in a class shall be defined, or declared
-pure~(\ref{class.abstract}) in that class, or both; but no diagnostic is
+pure~(\ref{class.abstract}) in that class, or both; no diagnostic is
 required~(\ref{basic.def.odr}).
 \indextext{friend!\tcode{virtual}~and}%
 


source/derived.tex
```diff
@@ -800,7 +800,7 @@
 \pnum
 \indextext{definition!virtual~function}%
 A virtual function declared in a class shall be defined, or declared
-pure~(\ref{class.abstract}) in that class, or both; but no diagnostic is
+pure~(\ref{class.abstract}) in that class, or both; no diagnostic is
```
jensmaurer: The "pure" thing is explained a bit more in [class.abstract]. Nothing in the normative text suggests a function can be abstract.

tkoeppe: Never mind, I misread that, "abstract" only appears in the label, not in the text.

PR #1183, opened 2016-12-07
[lib] More & and && adjustments for #166

timsong-cpp:


tkoeppe:
@jensmaurer: Ah, I knew we were far from done :-)

PR #1188, opened 2016-12-08
[class.bit] Add missing "bit-" to "field declaration" in index entry

morinmorin:
For reference, the C standard says `bit-field declaration`.


jensmaurer:
Yes, these things are called bit-fields.

tkoeppe:
@jensmaurer: Do you want this merged?

jensmaurer:
Yes.

PR #1191, opened 2016-12-08
[depr] Add missing parentheses and fix formatting

jlaire:


tkoeppe:
Thanks! Operator precedence is hard.

PR #1192, opened 2016-12-08
[numeric.special] Remove redundant 'inline' specifiers in example.

Eelis:
'inline' is doubly redundant here because these member functions are defined inside the class definition as well as constexpr.


tkoeppe:
@jensmaurer, @jwakely: Does this look legit?

jensmaurer:
Looks good to me. "inline" is mostly noise on the semantic level, anyway.

PR #1194, opened 2016-12-08
[rand.req.adapt] Fix syntax error

jlaire:
The `void` return type in the constructor is clearly an error, and I think adding `A::` to the member functions is good for clarity.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -2133,7 +2133,7 @@ A::A(result_type s);
 
 
 \begin{itemdecl}
-template<class Sseq> void A::A(Sseq& q);  
+template<class Sseq> A::A(Sseq& q);
 \end{itemdecl}
 
 \begin{itemdescr}


source/numerics.tex
```diff
@@ -2143,7 +2143,7 @@
 \end{itemdescr}
 
 \begin{itemdecl}
-void seed();  
+void A::seed();
```
tkoeppe: This is weird. In most itemdecls we *never* say the class name. This "random" business is very idiosyncratic, though. But I'd call this change a step backwards.

jlaire: It is weird, but the constructors have `A::A`.

tkoeppe: `tuple`, `pair` and `vector` don't use this pattern.

Yeah, maybe drop this commit except the `void` fix. Feel free to open an issue if you like.

jlaire: I don't think `tuple`, `pair` and `vector` are really comparable. They are standard types defined with a synopsis and a detailed description of each function like a lot of others.

This section lists "adapter requirements" that any type can meet, and is more similar to the iterator categories, Allocator requirements, etc., which we specify by listing expressions in a table that must be valid and have some semantics. In fact the surrounding subclauses [rand.req.seedseq], [rand.req.urng], [rand.req.eng] and  [rand.req.dist] all use this approach too. I have no idea why [rand.req.adapt] is like this. For now I'll leave it alone, just removing the `void`.

PR #1196, opened 2016-12-08
Clean up duplicate 'itemdescr' items.

### LINKED ISSUE #288: [alg.lex.comparison] has two 'Remarks' paragraphs with an unexplained code snippet between them
There's an unexplained chunk of code between [alg.lex.comparison]p3 and p4. This would benefit from some wording introducing the code and explaining that it's a sample implementation.

p3 and p4 are both 'Remarks:' paragraphs. It's rare to have two 'Remarks:' paragraphs for the same entity; the only other cases are [util.smartptr.shared.atomic]p36 and p37, and the mess in [time.duration]p2-p7. Perhaps this should be cleaned up somehow.

jensmaurer:
Fixes #288.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -5272,7 +5272,7 @@ applications of the corresponding comparison.
 \pnum
 \remarks
 If two sequences have the same number of elements and their corresponding
-elements are equivalent, then neither sequence is lexicographically
+elements (if any) are equivalent, then neither sequence is lexicographically
 less than the other.
 If one sequence is a prefix of the other, then the shorter sequence is
 lexicographically less than the longer sequence.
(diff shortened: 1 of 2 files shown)


source/algorithms.tex
```diff
@@ -5272,25 +5272,30 @@
...
   if (*first1 < *first2) return true;
   if (*first2 < *first1) return false;
 }
 return first1 == last1 && first2 != last2;
 \end{codeblock}
+\end{example}
 
 \pnum
-\remarks\ An empty sequence is lexicographically less than any non-empty sequence, but
+\begin{note} An empty sequence is lexicographically less than any non-empty sequence, but
```
tkoeppe: When do we use the normal "note" environment and when do we use the `\Notes` element?

jensmaurer: The \Notes element will be removed with #1160.

tkoeppe: Oh RIP

PR #1197, opened 2016-12-08
[lib] Add main index entries for all headers.

### LINKED ISSUE #288: [alg.lex.comparison] has two 'Remarks' paragraphs with an unexplained code snippet between them

### LINKED ISSUE #940: Not all headers are indexed in the main index.
The main index (not the library names index) only has index entries for the C library headers and some freestanding ones such as `<exception>`, `<initializer_list>` and `<new>`.

It has entries for headers that aren't even part of C++, such as `<stdatomic.h>` and <threads.h>`but not  anything like`<vector>`.

Is that intentional?

jensmaurer:
Fixes #940.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -2875,6 +2875,7 @@ requirements for sequence containers.
 
 \rSec2[array.syn]{Header \tcode{<array>} synopsis}
 
+\indextext{\idxhdr{array}}%
 \indexlibrary{\idxhdr{array}}%
 \begin{codeblock}
 #include <initializer_list>
@@ -2916,7 +2917,8 @@ namespace std {
 
 \rSec2[deque.syn]{Header \tcode{<deque>} synopsis}
 
-\indexlibrary{\idxhdr{deque}}
+\indextext{\idxhdr{deque}}%
+\indexlibrary{\idxhdr{deque}}%
 
 \begin{codeblock}
 #include <initializer_list>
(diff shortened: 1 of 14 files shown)


source/containers.tex
```diff
@@ -5840,7 +5846,9 @@
 \end{codeblock}
 
 \rSec2[associative.set.syn]{Header \tcode{<set>} synopsis}%
-\indexlibrary{\idxhdr{set}}
+
+\indextext{\idxhdr{set}}%
```
tkoeppe: Escaping the newline is a bit pointless if the whole thing is surrounded by whitespace :-S

I would mildly suggest to not bother escaping newlines for index entries in between blocks... eh, I don't really care, it's just a bit of a pointless change.

jensmaurer: Ok. We seem to generally escape newlines for \indexXXX things.

source/iostreams.tex
```diff
@@ -15058,7 +15082,9 @@
 \end{codeblock}
 
 \pnum
+\indextext{\idxhdr{inttypes.h}}%
```
tkoeppe: (By contrast, it is crucial that these index entries not introduce whitespace, so that the paragraph begins on the left margin.)

tkoeppe:
I'm happy to merge this. Did you double-check the index that it all comes out as desired?

jensmaurer:
I used diffpdf on the index (which is semi-useful in this case) and it all kinds of new `<blah>` entries, as expected.

PR #1198, opened 2016-12-08
[expr.ass] The right operand is not necessarily an expression.

### LINKED ISSUE #720: [expr.ass] CWG 2399: What's "the expression"?
[expr.ass]/2,3,6 contains several references to "the expression":

> -2- In simple assignment (=), the value of the expression replaces that of the object referred to by the left operand.
> 
> -3- If the left operand is not of class type, the expression is implicitly converted (Clause [conv]) to the cv-unqualified type of the left operand.
> 
> -6- When the left operand of an assignment operator is a bit-field that cannot represent the value of the expression, the resulting value of the bit-field is implementation-defined.

Clearly it's meant to refer to the right operand, i.e., the _initializer-clause_, but that's not necessarily an expression, and in any event the reference is unclear just reading the text.

jensmaurer:
Fixes #720.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -4734,12 +4734,12 @@ single compound assignment operator.
 \end{bnf}
 
 \pnum
-In simple assignment (\tcode{=}), the value of the expression replaces
+In simple assignment (\tcode{=}), the value of the right operand replaces
 that of the object referred to by the left operand.
 
 \pnum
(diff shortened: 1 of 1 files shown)


tkoeppe:
This isn't good; please refer to the issue, I will paste @zygoloid's notes.

PR #1199, opened 2016-12-09
[allocator] Format OUTERMOST in code font

jlaire:
Other similar terms like *`INVOKE`*, *`DECAY_COPY`*, and *`GENERALIZED_SUM`* are always in code font and italics.

diff-1
diff-2


tkoeppe:
Thanks. This is a start; it's certainly not making things worse. I might play with the italic correction and linebreaking, but leave that to me.

Issue #1200, opened 2016-12-09
[diff] Capitalization of headings


jlaire:
Most clause headings mentioned in Annex C are not capitalized. For example:

**`C.2.11 Clause 20: general utilities library   [diff.cpp03.utilities]`**
**`C.3.7  Clause 27: input/output library [diff.cpp11.input.output]`**

But some of them are.

**`C.2.16 Clause 27: Input/output library [diff.cpp03.input.output]`**
**`C.4.9  Clause 20: General utilities library [diff.cpp14.utilities]`**

I think they should be all capitalized to match the style in the actual clause headings.

tkoeppe:
Headings are generally set in sentence case, and our sentence style does not generally capitalize after a colon. So the right solution is to fix the inappropriate capitalization. I'll handle this.

PR #1201, opened 2016-12-09
[diff] Capitalize clause headings

jlaire:
Fixes https://github.com/cplusplus/draft/issues/1200.

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -8,7 +8,7 @@
 This subclause lists the differences between \Cpp and
 ISO C, by the chapters of this document.
 
-\rSec2[diff.lex]{Clause~\ref{lex}: lexical conventions}
+\rSec2[diff.lex]{Clause~\ref{lex}: Lexical conventions}
 
 \ref{lex.key}
 \change New Keywords\\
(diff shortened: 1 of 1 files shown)


jensmaurer:
Agreed on the general plan as outlined in #1200; I haven't reviewed the patch itself.

zygoloid:
We generally use sentence case for headings, not title case, which suggests not capitalizing after a colon. We only appear to get this wrong in 6 sections out of nearly 2000:

compatibility.tex:\rSec2[diff.cpp03.input.output]{Clause \ref{input.output}: Input/output library}
compatibility.tex:\rSec2[diff.cpp14.utilities]{Clause \ref{utilities}: General utilities library}
future.tex:\rSec1[depr.meta.types]{Deprecated Type Traits}
future.tex:\rSec1[depr.iterator.primitives]{Deprecated Iterator primitives}
iostreams.tex:\rSec3[fs.op.weakly_canonical]{Weakly Canonical}
threads.tex:\rSec2[thread.req.lockable]{Requirements for Lockable types}

I would suggest fixing those six instead of the proposed change. (Note that the last one should presumably be changed to "\tcode{Lockable}" instead of "lockable".)

PR #1203, opened 2016-12-09
[limits] Use 'subnormal number' as defined by IEEE 754-2008 = ISO 60559.

### LINKED ISSUE #1017: Subnormal instead of denormal
As a follow-up to #1009: IEEE-754 2008 uses "subnormal" instead of "denormal". Since C++ refers to the 2008 version, we should use the same naming.

See its section 2.1.51 for the definition. See 2.1.15 for "denormalized number: See: subnormal number."

jensmaurer:
Also add index entries for 'denormalized value' and 'value,
denormalized', pointing to 'number, subnormal'.

Fixes #1017.

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -468,8 +468,9 @@ static constexpr T min() noexcept;
 Minimum finite value.\footnote{Equivalent to \tcode{CHAR_MIN}, \tcode{SHRT_MIN},
 \tcode{FLT_MIN}, \tcode{DBL_MIN}, etc.}
 
+\indextext{number!subnormal}%
 \pnum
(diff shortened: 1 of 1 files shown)


zygoloid:
It seems truly strange to be talking about subnormals here, but not mentioning the possibility of them existing at all, or what they might mean, anywhere within the core language. We don't even have a normative reference to ISO/IEC/IEEE 60559, except perhaps indirectly via ISO/IEC 9899.

But that's all a pre-existing problem. If we're going to have this mess, using the latest 60559 terminology to describe it seems reasonable to me.

PR #1204, opened 2016-12-09
For 'immediate context', refer to [temp.deduct].

### LINKED ISSUE #272: Verbiage about compilation errors in the non-immediate context spammed out everywhere
We need to centralize the definition of "immediate context" wrt template instantiation so we can eliminate spammy verbiage like the following in the description of `is_assignable`:

> Only the validity of
> the immediate context of
> the assignment expression
> is considered. [ Note: The
> compilation of the
> expression can result in
> side effects such as the
> instantiation of class
> template specializations
> and function template
> specializations, the
> generation of
> implicitly-defined
> functions, and so on. Such
> side effects are not in the
> “immediate context” and
> can result in the program
> being ill-formed. —end
> note ]

In addition to all the current places where "immediate context" is described in the current standard, similar wording can be found in proposals for SFINAE-friendly `iterator_traits`, `common_type`, and `invocation_traits`.

### LINKED ISSUE #445: 20.10.6 [meta.rel] says "return-statement"
Should probably be "return statement"

jensmaurer:
Drop the duplicated notes explaining ill-formed specializations
in other than the immediate context.

Fixes #445.
Fixes #272.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -16312,11 +16312,11 @@ To test() {
 \end{codeblock}
 
 \begin{note} This requirement gives well defined results for reference types, void
-types, array types, and function types.\end{note} Access checking is performed as
-if in a context unrelated to \tcode{To} and \tcode{From}. Only the validity of
-the immediate context of the expression of the \grammarterm{return-statement}
-(including conversions to the return type) is considered. \begin{note} The
-evaluation of the conversion can result in side effects such as the
+types, array types, and function types.\end{note} Access checking is performed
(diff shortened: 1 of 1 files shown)


cpplearner:
[LWG 2290](http://wg21.link/lwg2290) is closely related.

jensmaurer:
Rebased.

jensmaurer:
The library parts of this are covered by LWG 2290. For the narrow focus of #445, I'll create a new pull request.

PR #1207, opened 2016-12-11
[dcl.fct.default] Restore }, again

timsong-cpp:


tkoeppe:
Thanks, I fixed it slightly differently: a367d459e6e9cf548936876cac615989b1ae4b08

PR #1210, opened 2016-12-11
[{ex,in}clusive.scan]: Rephase 'for each *j' in terms of indexes

tkoeppe:
Addresses #693 for this part, but is also clearer.

**Exclusive:**

image

Note that the range is *empty* when `K` is zero.

**Inclusive:**

image

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -9155,11 +9155,12 @@ ranges \range{first}{last} or \range{result}{result + (last - first)}.
 
 \pnum
 \effects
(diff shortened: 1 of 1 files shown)


tkoeppe:
@zygoloid, @jfbastien, @jwakely: Could you please review this proposal?

timsong-cpp:
Any changes made here should be applied to the `transform_*` versions as well.

tkoeppe:
@timsong-cpp: Thanks, done!

PR #1212, opened 2016-12-12
[pairs.spec], [tuple.creation] Avoid 'equality' of types phrasing.

### LINKED ISSUE #491: [pairs.spec] talks about "if Ui equals reference_wrapper<X>"
We shouldn't talk about equality of types.

Reword it as:

> If `Ui` is a specialization of `reference_wrapper` then `Vi` is `Ui::type&`, otherwise `Vi` is `Ui`

[tuple.creation] has the same problem.

jensmaurer:
Fixes #491.

(I agree that the formatting of the U_i indexed type-variables is dreadful in either case, but that is the subject of another issue.)


tkoeppe:
This is much better, thanks!

Issue #574, opened 2015-12-05
Avoid hanging paragraphs as well as 'in general' / 'overview' subclauses


zygoloid:
Within the library, it's relatively common for a clause with subclauses to contain no other content. The content that would most naturally be directly included in that clause is instead in a subclause named "General" or "In general" or "<name of parent clause> overview".

These additional clauses seem to be without merit, and are inconsistent (as the core language doesn't use them, and the library doesn't consistently use them either). We should consider removing them and folding the content into the containing clause.

Removing a clause and its corresponding section label is not something that should be done lightly, since people use the section labels as stable names for the corresponding content, so this would need some care. We might want to keep the subclause section labels and discard the labels that used to have no content, to avoid breaking too many external references to particular paragraphs, even though that will leave us with somewhat bizarre section labels in some cases.

tkoeppe:
Would it be possible to assign *multiple* stable lables to a heading? Then we could have, say, "Diagnostics library [diagnostics, diagnostics.general]".

zygoloid:
We could have a mapping from old labels to new labels in Annex F. I'd rather keep the legacy stuff out of the main text if possible.

zygoloid:
The core language section actually does have a handful of these sections: [expr.prim.general] (which is a horrible mishmash that should be divided into subsections), [dcl.fct.def.general], [dcl.attr.grammar], [lex.literal.kinds], and perhaps [basic.scope.declarative].

[basic.start] also contains no text, but there's not really anything for it to say, so it should remain as-is.

AlisdairM:
I believe this is inspired by Pete Becker's time as the editor.  There was an understanding that ISO rules for documents required text could only occur in 'leaf' nodes of the [a.b.c] tree, and while our standard clearly violated that rule all over, there was an attempt to bring it closer to conformance.  If that rule/convention has since been rescinded, there could be an argument for simplifying out the empty tags again.

jensmaurer:
I think the section label issue should work out. We can certainly fix all references in the standard and in the LWG issues list. Are there large quantities of references elsewhere?

tkoeppe:
@jensmaurer: Printed books, lecture notes, ...

jensmaurer:
We have 39 section labels with ".general". We have 28 references to such section, with a substantial chunk referring to [iterator.requirements.general] and [containers.requirements.general]. Both sections contain much more than introductory material and should not be folded in any event.

[...]

PR #1213, opened 2016-12-12
[utility], [bitset] Introduce a separate heading for the synopsis.

jensmaurer:
Partially addresses #566.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -40,12 +40,7 @@ throughout the rest of the library.
 \indextext{\idxhdr{utility}}%
 \indexlibrary{\idxhdr{utility}}%
 \indexlibrary{\idxcode{rel_ops}}%
-\synopsis{Header \tcode{<utility>} synopsis}
-
-\pnum
-The header \tcode{<utility>} defines several types and function templates
-that are described in this Clause. It also defines the template \tcode{pair}
-and various function templates that operate on \tcode{pair} objects.
+\rSec2[utility.syn]{Header \tcode{<utility>} synopsis}
 
 \begin{codeblock}
 #include <initializer_list>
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -40,7 +40,7 @@
 \indextext{\idxhdr{utility}}%
 \indexlibrary{\idxhdr{utility}}%
 \indexlibrary{\idxcode{rel_ops}}%
-\synopsis{Header \tcode{<utility>} synopsis}
+\rSec2[utility.syn]{Header \tcode{<utility>} synopsis}
```
tkoeppe: I had thought about this, too, but until now no "Synopsis" section that has a numbered heading has an introductory paragraph. That's why I haven't made this change yet, I wonder if we need some more rewording.

tkoeppe: Right, with this change we would have a numbered section called "Synopsis" that does *not* start with a code block. That'd be new, and moreover it seems that all these introductory sentences don't really add any normative value. So maybe we could revise this on at a slightly larger scale?

jensmaurer: Ah, this is talking about <utility>, which somehow made it into the [bitset] patch.

jensmaurer:
Ok, I've moved the introductory comment for <utility> at the bottom of the synopsis. This is the same as for bitset.

source/utilities.tex
tkoeppe: Yeah, maybe -- I have never managed to make up my mind about this one. (And I have looked at this case before, and others awkward cases like `<algoritms>`.) Not sure what the best way is to present this.

jensmaurer: This interacts somewhat with #574. If we remove those "in general" clauses everywhere and move their contents to level x-1, the introductory sentence for &lt;utility> can simply go at that level, too.

Issue #1071, opened 2016-11-18
[valarray] wording needs major overhaul


AlisdairM:
This is a large piece of work.
The valarray section does not use the standard section labels,
Requires/Effects/Returns/etc. using free-form text to try to
establish the same guarantees.  Other than looking like bad
style, this form of presentation has grown to grant explicit
meaning through these terms, and the valarray specification
would be much clearer if adapted.

jensmaurer:
I can't believe this to be editorial. Rephrasing large bodies of text always risks a subtle semantic change, so this should really be an LWG issue.

jensmaurer:
See also #232.

AlisdairM:
I think editorially, we can incrementally improve these clauses in various ways without triggering a large paper and asking library to review it - as I believe LWG would rather focus its limited resources elsewhere.  Editorially, I think this subclause is far from satisfactory in its current shape, given the improving structure of library clauses over time.  Continuous refinement to avoid normative changes though, definitely makes this a 'big' task.

sigfpe:
Whatever we do, I recommend extreme caution. Be careful about edtorial fever overreach.
-------- Original message --------From: Alisdair Meredith <notifications@github.com> Date: 11/20/16  11:19 AM  (GMT-08:00) To: cplusplus/draft <draft@noreply.github.com> Subject: Re: [cplusplus/draft] [valarray] wording needs major overhaul (#1071) 
I think editorially, we can incrementally improve these clauses in various ways without triggering a large paper and asking library to review it - as I believe LWG would rather focus its limited resources elsewhere.  Editorially, I think this subclause is far from satisfactory in its current shape, given the improving structure of library clauses over time.  Continuous refinement to avoid normative changes though, definitely makes this a 'big' task.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub, or mute the thread.


  
  




[...]

PR #1215, opened 2016-12-12
[numarray] Add standard description elements.

jensmaurer:
Partially addresses #1071.

This is a first step. I've only inserted description elements such as \effects and \returns and removed boilerplate introductory phrases where appropriate.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -6748,6 +6748,8 @@ is a
 one-dimensional smart array, with elements numbered sequentially from zero.
 It is a representation of the mathematical concept
 of an ordered set of values.
+For convenience, an object of type \tcode{valarray<T>} is referred
+to as an ``array'' throughout the remainder of~\ref{numarray}.
 The illusion of higher dimensionality
 may be produced by the familiar idiom of computed indices, together
 with the powerful subsetting capabilities provided
@@ -6777,11 +6779,8 @@ valarray();
 \begin{itemdescr}
 \pnum
 \effects
-Constructs an object of class
-\tcode{valarray<T>}\footnote{For convenience, such objects are referred
-to as ``arrays'' throughout the
-remainder of~\ref{numarray}.}
-which has zero length.\footnote{This default constructor is essential,
+Constructs a \tcode{valarray}
+that has zero length.\footnote{This default constructor is essential,
 since arrays of
 \tcode{valarray}
 may be useful.
@@ -6792,37 +6791,41 @@ member function.}
 
 \indexlibrary{\idxcode{valarray}!constructor}%
 \begin{itemdecl}
-explicit valarray(size_t);
+explicit valarray(size_t n);
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
-The array created by this constructor has a length equal to the value of the argument.
-The elements of the array are value-initialized~(\ref{dcl.init}).
+\effects
+Constructs a \tcode{valarray} that has length \tcode{n}.
+Each element of the array is value-initialized~(\ref{dcl.init}).
 \end{itemdescr}
 
 \indexlibrary{\idxcode{valarray}!constructor}%
 \begin{itemdecl}
-valarray(const T&, size_t);
+valarray(const T& v, size_t n);
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
-The array created by this constructor has a length equal to the second
-argument.
-The elements of the array are initialized with the value of the first argument.
+\effects
+Constructs a \tcode{valarray} that has length \tcode{n}.
+Each element of the array is initialized with \tcode{v}.
 \end{itemdescr}
 
 \indexlibrary{\idxcode{valarray}!constructor}%
 \begin{itemdecl}
-valarray(const T*, size_t);
+valarray(const T* p, size_t n);
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
-The array created by this constructor has a length equal to the second
-argument
-\tcode{n}.
+\requires
+\tcode{p} points to an array~(\ref{dcl.array}) of at least
+\tcode{n} elements.
+
+\effects
+Constructs a \tcode{valarray} that has length \tcode{n}.
 The values of the elements of the array are initialized with the
 first
 \tcode{n}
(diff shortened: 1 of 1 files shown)


source/numerics.tex
```diff
@@ -6776,10 +6778,7 @@
 \begin{itemdescr}
 \pnum
 \effects
-Constructs an object of class
-\tcode{valarray<T>}\footnote{For convenience, such objects are referred
-to as ``arrays'' throughout the
-remainder of~\ref{numarray}.}
+Constructs a \tcode{valarray}
 which has zero length.\footnote{This default constructor is essential,
```
zygoloid: which -> that (here and below), as this is a restrictive clause.

source/numerics.tex
```diff
@@ -6874,7 +6874,7 @@
 
 \begin{itemdescr}
 \pnum
-\effects Same as \tcode{valarray(il.begin(), il.size())}.
+\effects Equivalent to \tcode{valarray(il.begin(), il.size())}.
```
zygoloid: We have been reluctant to add "Equivalent to"s in the past, since "Equivalent to" also notionally populates the other elements of the description whereas "Same as" does not; see [structure.specifications]/4.

@jwakely Can you check that the change to "Equivalent to" here doesn't have an (unintended) normative effect?

source/numerics.tex
jensmaurer: Pre-existing text... Will do.

jensmaurer:
Updated for "which" -> "that".

source/numerics.tex
```diff
@@ -7279,40 +7250,42 @@
...
+\returns
+\tcode{*this}.
 
 \pnum
-If the value of an element in the left-hand side of a valarray compound
-assignment operator depends on the value of another element in that left
-hand side, the behavior is undefined.
+\remarks
+The appearance of an array on the left-hand side of a compound assignment
+does \emph{not} invalidate references or pointers.
```
jwakely: Why are we emphasising "not" here at all?

source/numerics.tex
```diff
@@ -7279,40 +7250,42 @@
...
 \begin{itemdescr}
 \pnum
+\requires
+\tcode{size() == v.size()}.
 Each of these operators may only be instantiated for a type \tcode{T}
-to which the indicated operator can be applied.
-Each of these operators
-performs the indicated operation on each of its elements and the
-corresponding element of the argument array.
+to which the indicated operator can be applied.
```
jwakely: Do we want to be more precise in these requirements? Being able to apply the operator doesn't mean much if we don't say what types the operands are. A type could overload `T::operator*=(std::string)` but that doesn't mean we can use `valarray<T>{} *= T{}`.

Maybe something like "For each operator function, the operator that it implements is invocable with two operands of type `T`"

source/numerics.tex
```diff
@@ -6946,23 +6953,22 @@
 
 \begin{itemdescr}
 \pnum
-\effects As if by: \tcode{*this = valarray(il);}
-
-\pnum
-\returns \tcode{*this}.
+\effects Equivalent to: \tcode{return *this = valarray(il);}
```
jwakely: I don't see any negative effect from using "Equivalent to" here either, although it's again arguably not editorial.

source/numerics.tex
```diff
@@ -6791,60 +6790,61 @@
...
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
-The array created by this constructor has a length equal to the second
-argument.
-The elements of the array are initialized with the value of the first argument.
+\effects
+Constructs a \tcode{valarray} that has length \tcode{n}.
+The elements of the array are initialized with \tcode{v}.
```
jensmaurer: Fixed, also in the signature before that ("Each element of the array is value-initialized.")

source/numerics.tex
```diff
@@ -6791,60 +6790,61 @@
...
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
-The array created by this constructor has a length equal to the second
-argument
-\tcode{n}.
+\requires
+The number of values pointed to by \tcode{p} is
+equal to or greater than \tcode{n}.
```
jensmaurer: We could add a cross-reference to 8.3.4 [dcl.array] after this particular "array" to disambiguate. Done.

source/numerics.tex
```diff
@@ -6976,100 +6982,65 @@
...
-is less than the length of
-\tcode{a}
-and
-\tcode{j}
-is less than the length of
-\tcode{b}.
-This property indicates an absence of aliasing and may be used to
-advantage by optimizing compilers.\footnote{Compilers may take advantage
+\remarks
+The expression \tcode{\&a[i+j] == \&a[i] + j}
```
jensmaurer: Yes, please.

source/numerics.tex
jensmaurer: I'd like to point out that this is pre-existing text; and I think LWG failed in distant history to produce a specification for std::complex that would allow T's other than built-in floating-point types. I'm reluctant to attempt to come up with a better specification here. (Although I admit the std::complex case was probably more about trigonometric functions thatn simple stuff.)

jensmaurer: I've added a slight rewording, highlighting the "two operands" issue.

source/numerics.tex
jensmaurer: Removed \emph (twice).

jensmaurer:
I've updated the patch.

jensmaurer:
@zygoloid: Back in your front yard; @jwakely has approved the change.

source/numerics.tex
AlisdairM: [numeric.defns] bullet 1.9 mean that we can ignore the usual concerns leading to addressof in this case, so I think there is no issue at all.

jwakely: Ah yes, not the first time I've forgotten about that provision.

tkoeppe:
@jensmaurer: rebase?

zygoloid:
Jens, can you rebase this please? (I'm also waiting on Marshall to confirm that we're OK to do this without further LWG signoff.)

mclow:
This looks good to me.   The only place I had a concern was around (replaced) link 7266, but after a  re-read I'm good with that, too.  Thanks, Jens!

jensmaurer:
Rebased as requested.

PR #1217, opened 2016-12-13
[strings, algorithms, diff] Replace 'routine' with 'function'.

Eelis:
These are the only occurrences.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -1352,7 +1352,7 @@ be invoked by user code called from \tcode{execution::parallel_unsequenced_polic
 algorithms.
 \begin{note}
 Implementations must ensure that internal synchronization inside standard
-library routines does not prevent forward progress when those routines are
+library functions does not prevent forward progress when those functions are
 executed by threads of execution with weakly parallel forward progress guarantees.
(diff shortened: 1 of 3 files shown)


jensmaurer:
Looks good to me.

tkoeppe:
@jensmaurer: Are you sure that "function" doesn't subtly introduce a more constrained notion than "routine"? I'd be happy to merge if you are confident that this is correct.

Eelis:
For fun, I searched for occurrences of "routine" in the C99 standard. Of the five I found, three appear as synonym for "function", one occurs in "objects referred to by interrupt service routines activated by the signal function" (which I guess refers to callback functions), and the fifth and most interesting one is:

> Aprincipal use of the sizeof operator is in communication with routines such as storage allocators and I/O systems.

Sounds old-timey to me, haven't heard of an I/O system referred to as a "routine" recently. :)

jensmaurer:
@tkoeppe: "routine" is an undefined concept in C++. I've reviewed the patch again, and in all cases, we're directly talking about "functions" as C++ understands them. Mentioning a "routine" is just a colloquialism that crept in.

PR #1220, opened 2016-12-13
[any] Rename the ValueType template parameter to T.

### LINKED ISSUE #1202: The ValueType parameter of std::any_cast is confusingly named
```
template<class ValueType>
  ValueType any_cast(const any& operand);
template<class ValueType>
  ValueType any_cast(any& operand);
template<class ValueType>
  ValueType any_cast(any&& operand);
```

This template parameter is confusing, because it's really the return type of the function, _not_ the type of the contained value in the `any` (as shown by the fact the requirements say it must be a reference, or copy constructible). When the parameter is a reference the functions return by reference, not by value, and `ValueType` is not the type of the value in the `any`. The parameter can be `const` but the value in the `any` is not a const-qualified type.

Elsewhere in the library "value type" is used for containers, iterators, and allocators, and must be an object type, not a reference type.

I suggest renaming this to make it clearer that it can be a reference, and is not the type of the value in the `any`.  `Result` or `ResultType` might be better. Or just `Type` or `T`.

For the `any_cast` overloads taking pointers the `ValueType` parameter is OK, as in that case it does refer to the type of the value in the `any`.

jensmaurer:
The parameter has a confusing name: It is not actually the value type
contained in the 'any'.
Where the new T clashes with an existing use of T, rename the latter to VT,
since this is actually the value type of the 'any'.

Fixes #1202.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -5338,17 +5338,17 @@ namespace std {
   template <class T, class U, class... Args>
     any make_any(initializer_list<U> il, Args&& ...args);
 
-  template<class ValueType>
-    ValueType any_cast(const any& operand);
-  template<class ValueType>
-    ValueType any_cast(any& operand);
-  template<class ValueType>
-    ValueType any_cast(any&& operand);
+  template<class T>
+    T any_cast(const any& operand);
+  template<class T>
+    T any_cast(any& operand);
+  template<class T>
+    T any_cast(any&& operand);
   
-  template<class ValueType>
-    const ValueType* any_cast(const any* operand) noexcept;
-  template<class ValueType>
-    ValueType* any_cast(any* operand) noexcept;
+  template<class T>
+    const T* any_cast(const any* operand) noexcept;
+  template<class T>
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -5467,92 +5467,92 @@
 
 \indexlibrary{\idxcode{any}!constructor}%
 \begin{itemdecl}
-template<class ValueType>
-any(ValueType&& value);
+template<class T>
+any(T&& value);
```
tkoeppe: This should be indented.

source/utilities.tex
```diff
@@ -5610,21 +5610,21 @@
 
 \indexlibrarymember{operator=}{any}%
 \begin{itemdecl}
-template<class ValueType>
-any& operator=(ValueType&& rhs);
+template<class T>
+any& operator=(T&& rhs);
```
tkoeppe: Likewise here.

jensmaurer:
Done.

PR #1227, opened 2016-12-13
[localization] Fix indentation in class template synopses.

### LINKED ISSUE #1205: Synopses in [localization] use non-standard indentation style
A lot of the synopses in Clause 22 don't indent the second line of a class template, i.e. the say

    template <typename T>
    class foo {
    public:

instead of our usual:

    template <typename T>
      class foo {
      public:

This might be worth cleaning up. While you're there, there's also ample opportunity to look out for other idiosyncratic spacing and indentation.

jensmaurer:
Fixes #1205.

### DIFF
--- a/source/locales.tex
+++ b/source/locales.tex
@@ -1073,7 +1073,8 @@ std::cout << mbstring;
 namespace std {
 template <class Codecvt, class Elem = wchar_t,
     class Wide_alloc = allocator<Elem>,
-    class Byte_alloc = allocator<char>> class wstring_convert {
+    class Byte_alloc = allocator<char>>
+  class wstring_convert {
   public:
     using byte_string = basic_string<char, char_traits<char>, Byte_alloc>;
     using wide_string = basic_string<Elem, char_traits<Elem>, Wide_alloc>;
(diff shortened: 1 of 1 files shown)


source/locales.tex
```diff
@@ -1947,54 +1948,55 @@
 \indexlibrary{\idxcode{ctype<char>}}%
 \begin{codeblock}
 namespace std {
-  template <> class ctype<char>
-    : public locale::facet, public ctype_base {
-  public:
-    using char_type = char;
+  template <>
+    class ctype<char>
+      : public locale::facet, public ctype_base {
```
tkoeppe: Bring the base class up on the same line like everywhere else while you're at it.

source/locales.tex
```diff
@@ -3194,48 +3196,48 @@
...
+      using char_type = charT;
+      using iter_type = OutputIterator;
+
+      explicit num_put(size_t refs = 0);
+
+      iter_type put(iter_type s, ios_base& f, char_type fill, bool v) const;
+      iter_type put(iter_type s, ios_base& f, char_type fill, long v) const;
+      iter_type put(iter_type s, ios_base& f, char_type fill, long long v) const;
+      iter_type put(iter_type s, ios_base& f, char_type fill,
+                    unsigned long v) const;
```
tkoeppe: I think we don't need these next couple of linebreaks. If you remove them now, I will never have to go back to Clause 22 again, which I would appreciate.

tkoeppe:
Thanks a lot!

jensmaurer:
Fixed as desired.

PR #1229, opened 2016-12-13
[localization] Add numbered headings for synopses.

jensmaurer:
Partially addresses #566.


tkoeppe:
The change is good, but the commit message is a bit odd: you're mostly just *removing* bogus "synopses" that aren't any, right? I can reword that on merging.

jensmaurer:
There are two \synopsis removals for class templates.
And two conversions of header \synopsis into proper numbered sections.

I've adjusted the commit comment to include the first part, too.

PR #1231, opened 2016-12-14
Remove \synopsis macro.

### LINKED ISSUE #566: Synopses should have numbered headers

jensmaurer:
[res.on.functions] Do not use \synopsis for emphasis.

Fixes #566.


tkoeppe:
Hurray!

PR #1235, opened 2016-12-14
[string.view.hash] Add a note about equality to string hashes.

jensmaurer:
From http://lists.isocpp.org/lib/2016/12/1531.php.

### DIFF
--- a/source/strings.tex
+++ b/source/strings.tex
@@ -5685,6 +5685,10 @@ template<> struct hash<wstring_view>;
 \begin{itemdescr}
 \pnum
 The specialization is enabled~(\ref{unord.hash}).
+\begin{note}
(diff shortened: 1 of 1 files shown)


source/strings.tex
```diff
@@ -5685,6 +5685,11 @@
 \begin{itemdescr}
 \pnum
 The specialization is enabled~(\ref{unord.hash}).
+\begin{note}
+The hash value of a string view object is equal to the hash value of
+the corresponding string object (\ref{basic.string.hash}).
+\end{note}
+
```
tkoeppe: Lose the newline?

Otherwise good.

jensmaurer:
Updated.

PR #1236, opened 2016-12-14
[path.non-member] Structure multi-paragraph note using an itemization instead.

Eelis:
Diff:
diff

This removes the most egregious instance in the document of a note spanning multiple numbered paragraphs, and I think it actually reads and looks nicer as an itemization (and with the examples marked as examples).

Thoughts? :)


tkoeppe:
I like it!

I haven't even got to iostreams yet. I have no doubt that there will be a plethora of editorial improvements we'll make to the new parts of it.

PR #1237, opened 2016-12-14
[over.built] Rework formatting of placeholders.

### LINKED ISSUE #1233: Inconsistent placeholder styling in [over.built]
Subclause [over.built] uses type placeholders, which are styles as body text italics in the body text and as code italics in code. They should always be the latter, using `\tcode{\placeholder{...}}` in the body and `\placeholder` or `\placeholdernc` in codeblocks.

jensmaurer:
Fixes #1233.

### DIFF
--- a/source/overloading.tex
+++ b/source/overloading.tex
@@ -3421,17 +3421,19 @@ is hidden and is not included in the set of candidate functions.
 \end{note}
 
 \pnum
+\indextext{type!integral!promoted}%
+\indextext{type!arithmetic!promoted}%
 In this subclause, the term
-\term{promoted integral type}
+\defn{promoted integral type}
 is used to refer to those integral types which are preserved by
-integral promotion (including e.g.
+integral promotion~(\ref{conv.prom}) (including e.g.
 \tcode{int}
 and
 \tcode{long}
 but excluding e.g.
 \tcode{char}).
 Similarly, the term
-\term{promoted arithmetic type}
+\defn{promoted arithmetic type}
 refers to floating types plus promoted integral types.
 \begin{note}
 In all cases where a promoted integral type or promoted arithmetic type is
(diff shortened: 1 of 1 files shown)


source/overloading.tex
```diff
@@ -3440,331 +3442,295 @@
 \end{note}
 
 \pnum
+In the remainder of this section, \cvqual{vq} represents either
+\tcode{volatile} or the absence of \tcode{volatile}.
```
tkoeppe: Maybe "or no cv qualifier"? Because currently it sounds like you could use "vq cv" to take an arbitrary cv-qualifier and constrain it to not include `volatile`.

tkoeppe: Maybe reuse the "is either `volatile` or empty" wording from below?"

source/overloading.tex
```diff
@@ -3440,331 +3442,295 @@
...
 \end{codeblock}
 
 \pnum
 For every quintuple
-(\textit{C1},
-\textit{C2},
-\textit{T},
-\textit{CV1},
-\textit{CV2}),
+(\tcode{\placeholder{C1}},
```
tkoeppe: *Now* we could start thinking about subscripting, e.g. `\tcode{\placeholder{C}}$_1$`. But we can do that in a separate iteration.

tkoeppe:
Thanks!

source/overloading.tex
```diff
@@ -3421,17 +3421,19 @@
...
+integral promotion~(\ref{conv.prom}) (including e.g.
 \tcode{int}
 and
 \tcode{long}
 but excluding e.g.
 \tcode{char}).
 Similarly, the term
-\term{promoted arithmetic type}
+\defn{promoted arithmetic type}
 refers to floating types plus promoted integral types.
```
jensmaurer: No. 5p11.4 says we don't promote "float" to "double" in expressions, in the way we promote char to int.

tkoeppe: Oh, right, I was thinking of the default argument promotions, not the arithmetic promotions.

source/overloading.tex
jensmaurer: Yes, the subscripting is separate. And no, we want to use $\mcode{\placeholder{C}}_1$.
The thing inside the math environment should be a well-formed mathematical term. Just the subscript looks broken. (Maybe that rule helps LaTeX keep the spacing right.)

jensmaurer:
Updated.

source/overloading.tex
jensmaurer: error: missing closing brace

I'd really like to introduce \mcode = \text{\tcode{...}} otherwise the LaTeX source gets unreadable.

PR #1238, opened 2016-12-14
[library] Add full stops to \returns etc.

jlaire:
My understanding is that we want to always end \returns and other elements in a full stop. The wiki (https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines) seems to agree with this, although it doesn't talk about all elements and there is one conflicting "*Returns:* `T{0}`" example.

I haven't gone through the whole draft so there are likely a lot more remaining. These are just ones I've noticed so far.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -1281,7 +1281,7 @@ template <class T1, class T2>
 tuple_element<0, pair<T1, T2>>::type
 \end{itemdecl}
 \begin{itemdescr}
-\pnum\textit{Value:} the type \tcode{T1}.
+\pnum\textit{Value:} The type \tcode{T1}.
 \end{itemdescr}
 
 \indexlibrary{\idxcode{tuple_element}}%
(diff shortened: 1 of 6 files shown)


jensmaurer:
I've updated the style guidelines to fix the T{0} bug.

jensmaurer:
Looks like an improvement to me. The commit comment for the first one might want to mention that capitalization (first letter after element designation) is also adjusted.

jlaire:
I reworded the commit message and added a few more fixes.

source/utilities.tex
```diff
@@ -18990,7 +18990,7 @@
 \begin{example}
 \begin{codeblock}
 using namespace std;
-vector<int> v = ...
+vector<int> v = ...;
```
tkoeppe: This isn't valid syntax. Could you perhaps use `vector<int> v = @\commentellip@;`?

source/utilities.tex
```diff
@@ -1281,15 +1281,15 @@
 tuple_element<0, pair<T1, T2>>::type
 \end{itemdecl}
 \begin{itemdescr}
-\pnum\textit{Value:} the type \tcode{T1}.
+\pnum\textit{Value:} The type \tcode{T1}.
```
tkoeppe: The *value* is a *type*? Eww. This is weird.

jlaire: Yeah, it is.

source/utilities.tex
jlaire: Thanks, I didn't know about that. There seem to be many uses of literal `...` that could be changed, but I'll leave the others for a different PR.

Issue #781, opened 2016-07-01
\pnums in examples and notes are ugly and confusing


burblebee:
Update: the issue with \pnums in effects/return/etc. clauses was separated out and is now issue #821.

Adding \pnums added in the middle of examples and notes makes it hard to see where the example/note starts/ends, and \pnums added in \effects (or other) clauses causes the wording to spill over into a new paragraph making it look like it's no longer part of the Effects (or other).

As an example of \pnums in an example, have a look at 27.10.8.6.2 [path.factory]/p4 - it has new paragraphs as they appeared in the original wording, but it is ugly!  See also the Note above in 27.10.8.6 [path.non-member]/p8.

Inserting a blank line in the Latex before the text in the examples/notes/effects/etc. seems to work - is there a better way?  In any case, I think we should fix these as we come across them.  Or in one big swoop if someone has the patience.

Having examples embedded in the paragraph is ugly also - see 1.9 [intro.execution]/p16.  It would be nice to address those while we're at it, perhaps also by inserting a blank line before/after the example?  \linebreak doesn't work because it causes the previous line to get stretched, and \linebreak after a blank line results in an error.

Here's a few examples of embedded \pnums that I think we may want to fix:
- \pnums in clauses:
  *\* [ios.members.static]/p3
  *\* [streambuf.virt.pback]/p3
  *\* [istream::sentry]/p3
  *\* [istream::extractors]/p7
  *\* [istream::extractors]/p14
  *\* [istream.unformatted]/p8
  *\* [istream.unformatted]/p14
  *\* [istream.unformatted]/p18
  *\* [ostream.inserters.arithmetic]/p1
  *\* [ostream.inserters]/p7 (this is as far as I got thru Clause 27)
- \pnums in Examples:
  *\* [path.factory]/p4
  *\* [fs.op.copy]/p6
  *\* [fs.op.system_complete]/p5
  *\* [fs.op.temp_dir_path]/p3
- \pnums in Notes:
  *\* [path.non-member]/p8

Here's a few examples of embedded exampes/notes that I think we may want to fix:
- examples embedded in the paragraphs:
  *\* [intro.execution]/p16
- notes embedded in the paragraphs:
  *\*  [path.io]/p1

godbyk:
Could you be a bit more specific about the aspects of the current formatting that you find ugly? How would you prefer it to be formatted?

burblebee:
[...]

PR #1240, opened 2016-12-14
[class.derived] Don't bother splitting last remark in note off into separate paragraph.

Eelis:
While I think this change is definitely merited on its own, its real significance is that it gets rid of the very last (as far as I can tell) multi-paragraph note in the document, thereby fixing half of #781.

### DIFF
--- a/source/derived.tex
+++ b/source/derived.tex
@@ -144,19 +144,16 @@ derived object~(\ref{intro.object}) is unspecified.
 \begin{note}
 \indextext{directed acyclic graph|see{DAG}}%
 \indextext{lattice|see{DAG, subobject}}%
-a derived class and its base class subobjects can be represented by a
+A derived class and its base class subobjects can be represented by a
 directed acyclic graph (DAG) where an arrow means ``directly derived
-from''. A DAG of subobjects is often referred to as a ``subobject
-lattice''.
+from''. An arrow need not have a physical representation in memory.
+A DAG of subobjects is often referred to as a ``subobject lattice''.
 
 \begin{importgraphic}
 {Directed acyclic graph}
 {fig:dag}
 {figdag.pdf}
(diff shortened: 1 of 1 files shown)


source/derived.tex
```diff
@@ -147,16 +147,13 @@
 a derived class and its base class subobjects can be represented by a
```
tkoeppe: Capitalize beginning of note.

source/derived.tex
```diff
@@ -147,16 +147,13 @@
...
 directed acyclic graph (DAG) where an arrow means ``directly derived
 from''. A DAG of subobjects is often referred to as a ``subobject
-lattice''.
+lattice''. The arrows need not have a physical representation in memory.
```
tkoeppe: I'd delete "The" and just say "Arrows need not...".

Eelis: Oh, sorry, I re-pushed before addressing this second review comment. I'll address this one too.

Eelis: I've moved the sentence about the arrows not being physical to directly after the sentence where the arrows are first introduced. The association does indeed change, but I agree that's fine, especially with this rearrangement.

source/derived.tex
```diff
@@ -144,19 +144,16 @@
 \begin{note}
 \indextext{directed acyclic graph|see{DAG}}%
 \indextext{lattice|see{DAG, subobject}}%
-a derived class and its base class subobjects can be represented by a
+A derived class and its base class subobjects can be represented by a
 directed acyclic graph (DAG) where an arrow means ``directly derived
-from''. A DAG of subobjects is often referred to as a ``subobject
-lattice''.
+from''. The arrows need not have a physical representation in memory.
```
tkoeppe: I'd still change it to either "An arrow" or "Arrows", non?

PR #1245, opened 2016-12-14
[algorithms] Move non-modifying algorithms away

### LINKED ISSUE #289: is_partitioned in mutating sequence operations
Stephen Howe writes:

> is_partitioned() is under in section 25.3.13, which is under 25.3 “Mutating sequence operations”
> 
> Well there is something wrong with this classification as is_partitioned() does not change the sequence.
> is_partitioned() should be under 25.2 “Non-modifying sequence operations”

jensmaurer:
from [alg.modifying.operations].

For is_partitioned, create a new subsection [alg.is_partitioned]
under [alg.nonmodifying].
For partition_point, create a new subsection [partition.point]
under [alg.binary.search].

Fixes #289.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -3337,171 +3337,6 @@ randomness.
 
 \end{itemdescr}
 
-\rSec2[alg.partitions]{Partitions}
-
-\indexlibrary{\idxcode{is_partitioned}}%
-\begin{itemdecl}
-template <class InputIterator, class Predicate>
-  bool is_partitioned(InputIterator first, InputIterator last, Predicate pred);
-template <class ExecutionPolicy, class InputIterator, class Predicate>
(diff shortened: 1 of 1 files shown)


zygoloid:
I'm OK with this as-is, but I wonder if perhaps we should instead move all of [alg.partitions] into [alg.sorting], since partitioning is certainly a "related operation" of sorting.

jensmaurer:
@jwakely: What's your opinion?

jwakely:
I think it would be better to keep all the partition-related algos together, so if moving them under [alg.sorting] allows that then that seems good to me.

jensmaurer:
Ok, I've moved [alg.partitions] into [alg.sorting], right after [alg.binary.search].

PR #1251, opened 2016-12-15
[tuple.creation] Simplify the introductory notation, and then use notation that matches the introduction.

tkoeppe:
Here is a slight simplification of [tuple.creation]. I bugged me that the introduction was defining things that were only tangentially useful. I changed it to introduce the concept of indexing into a pack only once, and I used the "variable-pack" notation (e.g. `TTypes`, `VTypes`) in all the functions in this subclause.

I will work on `tuple_cat` separately.

Before:

image

After:

image

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -1394,16 +1394,17 @@ namespace std {
   // \ref{tuple.creation}, tuple creation functions
   constexpr @\unspec@ ignore;
 
-  template <class... Types>
-    constexpr tuple<@\placeholder{VTypes}@...> make_tuple(Types&&...);
-  template <class... Types>
-    constexpr tuple<Types&&...> forward_as_tuple(Types&&...) noexcept;
+  template <class... TTypes>
+    constexpr tuple<VTypes...> make_tuple(TTypes&&...);
 
-  template<class... Types>
-    constexpr tuple<Types&...> tie(Types&...) noexcept;
+  template <class... TTypes>
+    constexpr tuple<TTypes&&...> forward_as_tuple(TTypes&&...) noexcept;
+
+  template<class... TTypes>
+    constexpr tuple<TTypes&...> tie(TTypes&...) noexcept;
 
   template <class... Tuples>
-    constexpr tuple<@\placeholder{Ctypes}@...> tuple_cat(Tuples&&...);
+    constexpr tuple<CTypes...> tuple_cat(Tuples&&...);
 
   // \ref{tuple.apply}, calling a function with a tuple of arguments
   template <class F, class Tuple>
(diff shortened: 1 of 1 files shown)


tkoeppe:
If you like this general direction, I could rephrase `tuple_cat` to have template parameter `TupleTypes` and say that `Tuple`<sub>*i*</sub> is `tuple<ArgTypes_i>`, etc.

tkoeppe:
@jwakely: I'd appreciate your feedback.

jensmaurer:
Seems like an improvement to me.

jensmaurer:
@tkoeppe: We have merge conflicts here.

jwakely:
Yes, I think this is an improvement. I'm not keen on putting "where indexing is zero-based" in parentheses though, I preferred the original form.

tkoeppe:
@jwakely: Yes, I agree, I'll unparenthesize that.

tkoeppe:
@jwakely: Done.

source/utilities.tex
```diff
@@ -2047,7 +2040,7 @@
 \indexlibrary{\idxcode{tuple_cat}}
 \begin{itemdecl}
 template <class... Tuples>
-  constexpr tuple<@\placeholder{CTypes}@...> tuple_cat(Tuples&&... tpls);
+  constexpr tuple<CTypes...> tuple_cat(Tuples&&... tpls);
 \end{itemdecl}
 
 \begin{itemdescr}
```
zygoloid: The following description uses italic teletype *`T`_i* for elements of `Tuples`, which seems inconsistent with the changes made above.

tkoeppe: @zygoloid: I hadn't revised `tuple_cat` at all, and that was still the original formatting. I've now added a complete review of this description to the PR.

source/utilities.tex
```diff
@@ -2047,39 +2040,44 @@
...
+In the following paragraphs, let $\tcode{T}_i$ be the $i^\text{th}$ type in \tcode{Tuples},
+$\tcode{U}_i$ be \tcode{remove_reference_t<T$_i$>}, and $\tcode{tp}_i$ be the $i^\text{th}$
 parameter in the function parameter pack \tcode{tpls}, where all indexing is
 zero-based.
 
 \pnum
 \requires For all $i$, $\tcode{U}_i$ shall be the type
-$\cv_i$ \tcode{tuple<$Args_i...$>}, where $\cv_i$ is the (possibly empty) $i^{th}$
-cv-qualifier-seq and $Args_i$ is the parameter pack representing the element
-types in $\tcode{U}_i$. Let ${A_{ik}}$ be the ${k_i}^{th}$ type in $Args_i$. For all
```
tkoeppe: Please note that "k_i^th type" doesn't make sense - it should be "k^th type".

source/utilities.tex
```diff
@@ -2047,39 +2040,44 @@
 \indexlibrary{\idxcode{tuple_cat}}
 \begin{itemdecl}
 template <class... Tuples>
-  constexpr tuple<@\placeholder{CTypes}@...> tuple_cat(Tuples&&... tpls);
```
zygoloid: `CTypes` is still italicized in the synopsis. (Likewise for `make_tuple`.)

source/utilities.tex
```diff
@@ -2047,39 +2040,44 @@
...
 \requires For all $i$, $\tcode{U}_i$ shall be the type
-$\cv_i$ \tcode{tuple<$Args_i...$>}, where $\cv_i$ is the (possibly empty) $i^{th}$
-cv-qualifier-seq and $Args_i$ is the parameter pack representing the element
-types in $\tcode{U}_i$. Let ${A_{ik}}$ be the ${k_i}^{th}$ type in $Args_i$. For all
-$A_{ik}$ the following requirements shall be satisfied: If $\tcode{T}_i$ is
-deduced as an lvalue reference type, then
-\tcode{is_constructible_v<$A_{ik}$, $cv_i$ $A_{ik}$\&> == true}, otherwise
-\tcode{is_constructible_v<$A_{ik}$, $cv_i A_{ik}$\&\&> == true}.
+$\cv_i$ \tcode{tuple<$\tcode{Args}_i$...>}, where $\cv_i$ is the (possibly empty) $i^\text{th}$
+cv-qualifier-seq and $\tcode{Args}_i$ is the parameter pack representing the element
```
zygoloid: \grammarterm for `cv-qualifier-seq`, please.

tkoeppe: Done.

source/utilities.tex
tkoeppe: Ahh, yes; I change the synopsis to match all the changes in this section.

source/utilities.tex
zygoloid: ... but that doesn't make sense, since there is no such grammar element here. Perhaps it means that `$\cv_i$ are the cv-qualifiers of $\tcode{T}_i$`, but I'm not really sure. Let's make the formatting fix to italicize the grammar term cv-qualifier-seq here and file an LWG issue asking for this description to be revisited.

tkoeppe: OK, that's already applied then.

source/utilities.tex
zygoloid: While I agree, I'm also happier leaving that alone for now as you have done. I think using `k_i` is merely confusing rather than wrong.

source/utilities.tex
```diff
@@ -2047,39 +2041,44 @@
...
-deduced as an lvalue reference type, then
-\tcode{is_constructible_v<$A_{ik}$, $cv_i$ $A_{ik}$\&> == true}, otherwise
-\tcode{is_constructible_v<$A_{ik}$, $cv_i A_{ik}$\&\&> == true}.
+$\cv_i$ \tcode{tuple<$\tcode{Args}_i$...>}, where $\cv_i$ is the (possibly empty) $i^\text{th}$
+\grammarterm{cv-qualifier-seq} and $\tcode{Args}_i$ is the parameter pack representing the element
+types in $\tcode{U}_i$. Let $\tcode{A}_{ik}$ be the ${k}^\text{th}$ type in $\tcode{Args}_i$. For all
+$\tcode{A}_{ik}$ the following requirements shall be satisfied:
+\begin{itemize}
+\item If $\tcode{T}_i$ is deduced as an lvalue reference type, then
+      \tcode{is_constructible_v<$A_{ik}$, $cv_i\;A_{ik}$\&> == true}, otherwise
```
zygoloid: Missing tcode for A here and on the line below.

tkoeppe: Fixed.

PR #1253, opened 2016-12-15
[meta.rel] Fix incorrect use of \grammarterm{return-statement}.

### LINKED ISSUE #445: 20.10.6 [meta.rel] says "return-statement"

jensmaurer:
Fixes #445.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -16312,11 +16312,11 @@ To test() {
 \end{codeblock}
 
 \begin{note} This requirement gives well defined results for reference types, void
-types, array types, and function types.\end{note} Access checking is performed as
-if in a context unrelated to \tcode{To} and \tcode{From}. Only the validity of
-the immediate context of the expression of the \grammarterm{return-statement}
-(including conversions to the return type) is considered. \begin{note} The
-evaluation of the conversion can result in side effects such as the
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -16312,11 +16312,11 @@
...
 \begin{note} This requirement gives well defined results for reference types, void
-types, array types, and function types.\end{note} Access checking is performed as
-if in a context unrelated to \tcode{To} and \tcode{From}. Only the validity of
-the immediate context of the expression of the \grammarterm{return-statement}
-(including conversions to the return type) is considered. \begin{note} The
-evaluation of the conversion can result in side effects such as the
+types, array types, and function types.\end{note} Access checking is performed
+in a context unrelated to \tcode{To} and \tcode{From}. Only the validity of
+the immediate context of the \grammarterm{expression} of the \tcode{return} statement
+(including conversions to the returned object or reference) is considered. \begin{note} The
```
tkoeppe: Did you mean to retain "conversion to" here? @zygoloid suggested "initialization of" instead. It seems more accurate to "initialize a reference" than to "convert to a reference".

jensmaurer: Right.

jensmaurer:
Updated to use "initialization of" instead of "conversion".

PR #1261, opened 2016-12-16
[dcl.array], [temp.deduct.type] Split notes spanning numbered paragra…

jensmaurer:
…phs.

Partially addresses #781.

### DIFF
--- a/source/declarators.tex
+++ b/source/declarators.tex
@@ -1138,7 +1138,7 @@ is interpreted
 in such a way that
 \tcode{E1[E2]}
 is identical to
-\tcode{*((E1)+(E2))}.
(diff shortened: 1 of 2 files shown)


source/declarators.tex
```diff
@@ -1183,6 +1185,7 @@
 is applied to this pointer,
 the result is the pointed-to $(n-1)$-dimensional array,
 which itself is immediately converted into a pointer.
+\end{note}
```
tkoeppe: The alternative is to put the example inside the note, all in one paragraph. It depends on whether the example illustrates the note, or the general theme of this section.

jensmaurer:
Updated as desired.

PR #1262, opened 2016-12-16
Avoid \pnum inside examples.

### LINKED ISSUE #781: \pnums in examples and notes are ugly and confusing

jensmaurer:
With the exception of [facets.examples].

Fixes #781.

### DIFF
--- a/source/access.tex
+++ b/source/access.tex
@@ -116,8 +116,9 @@ being declared and, if the entity is a class, the definitions of members of
 the class appearing outside the class's \grammarterm{member-specification}{.}
 \begin{note} This access also applies to implicit references to constructors,
 conversion functions, and destructors. \end{note}
-\begin{example}
 
+\pnum
+\begin{example}
 \begin{codeblock}
 class A {
   typedef int I;    // private member
@@ -140,7 +141,6 @@ template<A::I> struct R { };
 struct D: A::B, A { };
 \end{codeblock}
 
-\pnum
 Here, all the uses of
 \tcode{A::I}
 are well-formed because
@@ -968,7 +968,6 @@ class C : public A, public B {
 };
 \end{codeblock}
 
-\pnum
 Since
 \tcode{W::f()}
 is available to
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -3918,14 +3918,12 @@ void h(int i) {
 }
 \end{codeblock}
 
-\pnum
 The \tcode{carries_dependency} attribute on function \tcode{f} means that the
 return value carries a dependency out of \tcode{f}, so that the implementation
 need not constrain ordering upon return from \tcode{f}. Implementations of
 \tcode{f} and its caller may choose to preserve dependencies instead of emitting
 hardware memory ordering instructions (a.k.a. fences).
 
-\pnum
 Function \tcode{g}'s second parameter has a \tcode{carries_dependency} attribute,
 but its first parameter does not. Therefore, function \tcode{h}'s first call to
 \tcode{g} carries a dependency into \tcode{g}, but its second call does not. The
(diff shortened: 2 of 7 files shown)


source/access.tex
```diff
@@ -116,8 +116,9 @@
 the class appearing outside the class's \grammarterm{member-specification}{.}
 \begin{note} This access also applies to implicit references to constructors,
 conversion functions, and destructors. \end{note}
-\begin{example}
 
+\pnum
+\begin{example}
```
tkoeppe: Consider here (and everywhere) whether the example illustrates the note or the section in general. (I'll take a look, too, but it's surprisingly hard to make GitHub show more context of the change.)

jensmaurer: There a star-like thing on the upper left corner of each diff hunk that, when pressed, shows more context.

Anyway, I did consider for each change whether I should create a new numbered paragraph for an example or not, and I decided on an individual basis. For this particular case, the example clearly does not belong to the preceding note, and I figured a separate paragraph would be more appropriate.

tkoeppe: Yeah, the star doesn't work well for interior changes :-S

OK, good, if this is all analyzed case-by-case, then I'm happy with it.

source/declarations.tex
```diff
@@ -3918,14 +3918,12 @@
 }
 \end{codeblock}
 
-\pnum
 The \tcode{carries_dependency} attribute on function \tcode{f} means that the
```
tkoeppe: (To be clear, these *are* still paragraph breaks, they're just not numbered. I'd like to call these "underhanded paragraphs". We do this in a few places.)

jensmaurer: Fully agreed. For this and a few other cases, I saw no way other than removing the number, but keeping the paragraph break.

PR #1268, opened 2016-12-17
Define 'object type' only once.

### LINKED ISSUE #511: "object type" has three definitions
_object type_ is defined in 3.9/8. It's also defined differently in 1.8/1, and not-defined-but-italicised in 3.9.3/1.

jensmaurer:
There were only a few references to the definition in
1.8 [intro.object], so drop that definition and refer
to 'an object's type' instead.

Fixes #511.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -391,7 +391,7 @@ type \tcode{T} must be complete if:
 \item an object of type \tcode{T} is defined~(\ref{basic.def}), or
 \item a non-static class data member of type \tcode{T} is
 declared~(\ref{class.mem}), or
-\item \tcode{T} is used as the object type or array element type in a
+\item \tcode{T} is used as the allocated type or array element type in a
 \grammarterm{new-expression}~(\ref{expr.new}), or
 \item an lvalue-to-rvalue conversion is applied to
 a glvalue referring
@@ -3596,7 +3596,6 @@ void bar() {
 contexts incomplete types are prohibited. \end{note}
 
 \pnum
-\indextext{object type}%
 An \defn{object type} is a (possibly cv-qualified) type that is not
 a function type, not a reference type, and not \cv{}~\tcode{void}.
 
@@ -4060,8 +4059,8 @@ cv-unqualified complete or incomplete object type or is
 \tcode{void}~(\ref{basic.types}) has three corresponding cv-qualified
 versions of its type: a \defn{const-qualified} version, a
 \defn{volatile-qualified} version, and a
-\defn{const-volatile-qualified} version. The term
-\term{object type}~(\ref{intro.object}) includes the cv-qualifiers
+\defn{const-volatile-qualified} version. The
+type of an object~(\ref{intro.object}) includes the \grammarterm{cv-qualifier}{s}
 specified in the \grammarterm{decl-specifier-seq}~(\ref{dcl.spec}), 
 \grammarterm{declarator} (Clause~\ref{dcl.decl}),
 \grammarterm{type-id}~(\ref{dcl.name}), or 
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -554,8 +554,7 @@ object are determined when the object is created. An object can have a
 name~(Clause~\ref{basic}). An object has a storage
 duration~(\ref{basic.stc}) which influences its
 lifetime~(\ref{basic.life}). An object has a
-type~(\ref{basic.types}). The term \defn{object type} refers to
-the type with which the object is created.
+type~(\ref{basic.types}).
 Some objects are
 polymorphic~(\ref{class.virtual}); the implementation
 generates information associated with each such object that makes it


source/basic.tex
```diff
@@ -391,7 +391,7 @@
 \item an object of type \tcode{T} is defined~(\ref{basic.def}), or
 \item a non-static class data member of type \tcode{T} is
 declared~(\ref{class.mem}), or
-\item \tcode{T} is used as the object type or array element type in a
+\item \tcode{T} is used as the object's type or array element type in a
```
zygoloid: It seems a little awkward to use the possessive in only one of the two cases here. Maybe "used as the allocated type or array element type" would be better?

source/basic.tex
```diff
@@ -4059,8 +4058,8 @@
 \tcode{void}~(\ref{basic.types}) has three corresponding cv-qualified
 versions of its type: a \defn{const-qualified} version, a
 \defn{volatile-qualified} version, and a
-\defn{const-volatile-qualified} version. The term
-\term{object type}~(\ref{intro.object}) includes the cv-qualifiers
+\defn{const-volatile-qualified} version. An
+object's type~(\ref{intro.object}) includes the cv-qualifiers
```
zygoloid: Here, perhaps "The type of an object" instead of "An object's type"?

While we're looking at this line, is there a reason why "cv-qualifiers" here isn't `\grammarterm`'d? We definitely seem to be talking about the grammar element.

source/basic.tex
jensmaurer: Yes, sounds better.

source/basic.tex
jensmaurer: Yes, we're talking about the grammar element, although the entire paragraph is odd:
```
using T = const int;
T x1;
const int x2;
const T x3;
```
All of the x1, x2, x3 objects have the same const-qualified type, even though there is no "const" cv-qualifier in the decl-specifier-seq of x1. But that fix is for another day.

I've added the "grammarterm" and change the possessive.

jensmaurer: Uh, our definition of \grammarterm has serious trouble handling
\grammarterm{cv-qualifier}s
(plural "s" after singular grammar non-terminal). There's lots of whitespace in between. That's also visible in 7.1.1 [dcl.stc] p5 (the note at the end; twice).

jensmaurer: Ah, we have

\newcommand{\grammarterm}[1]{\textit{#1}\xspace}

and the \xspace is probably the culprit (why do we need that in the first place?)

jensmaurer:
See also http://tex.stackexchange.com/questions/86565/drawbacks-of-xspace with the first answer from the package author of \xspace. Quote "So, if you find it useful, fine, it's there. But personally I wouldn't recommend it."

jensmaurer:
I've added a separate commit to this pull request that removes \xspace from the definition of \grammarterm. A diffpdf in "appearance" mode shows 239 differences. I've looked through the first of them and found only mildly reduced whitespace between a \grammarterm and a following punctuator. For a full stop (period), this actually seems to improve matters; there was awkward extra space between the r and the period in
\grammarterm{identifier}.

source/basic.tex
```diff
@@ -4060,8 +4059,8 @@
 \tcode{void}~(\ref{basic.types}) has three corresponding cv-qualified
 versions of its type: a \defn{const-qualified} version, a
 \defn{volatile-qualified} version, and a
-\defn{const-volatile-qualified} version. The term
-\term{object type}~(\ref{intro.object}) includes the cv-qualifiers
+\defn{const-volatile-qualified} version. The
+type of an object~(\ref{intro.object}) includes the \grammarterm{cv-qualifier}s
```
zygoloid: Use {cv-qualifier}{s} to avoid the \xspace inside \grammarterm inserting a space before the 's'.

source/macros.tex
```diff
@@ -160,7 +160,7 @@
 \newcommand{\defnx}[2]{\indexdefn{#2}\textit{#1}\xspace}
 \newcommand{\defn}[1]{\defnx{#1}{#1}}
 \newcommand{\term}[1]{\textit{#1}\xspace}
-\newcommand{\grammarterm}[1]{\textit{#1}\xspace}
+\newcommand{\grammarterm}[1]{\textit{#1}}
```
zygoloid: Ah, OK :)

Have you checked that this does not cause any change in formatting anywhere else?

zygoloid:
I'm not prepared to take a macro change the day before a working paper goes out unless someone has looked at all the resulting formatting changes; a survey of the first few is not sufficient at this point. I'll take this if either you've looked at all 239 and they are all OK, or if you change the "\grammarterm{cv-qualifier}s" to "\grammarterm{cv-qualifier}{s}", or even if you don't add the \grammarterm at all (since it was a drive-by anyway). I'm also a bit surprised we see any differences here in cases where \xspace was not adding a space; do we really have 239 bogus spaces inserted by this, or does \xspace also somehow change the formatting even when it doesn't add a space?

If we don't make the macro change now, I'd still like to make it at some point. Generally I'd like for us to get rid of the \xspaces entirely. (See http://tex.stackexchange.com/questions/86565/drawbacks-of-xspace for the reason why, and note that even the original author of the package does not recommend it.)

godbyk:
@zygoloid I wouldn't be surprised if `\xspace` interferes with the kerning between two characters (_i.e._, preventing kerning altogether). This would result in minute spacing differences.

zygoloid:
@godbyk Thanks, that makes sense.

I don't think that changes my position, though; I'd like to check that none of the 239 changes are losing a space that we want; that could happen pretty subtly:

    the \grammarterm{foo-name}%
    \indextext{...}%
    is the name of a foo

... would I think lose the space after *foo-name* with this change. I can check this myself if no-one else does so first, but I have a few other things I need to get done before the pre-meeting mailing deadline.

godbyk:
@zygoloid To clarify, my comment wasn't intended to change your position—I agree with it, given the inconsistencies I've seen in the code—but rather to provide an explanation for some of the minor changes you may see.

jensmaurer:
Updates:
- Removed change to macros.tex; this is handled by issue #1433.
- Used \grammarterm{cv-qualifier}{s}

PR #1269, opened 2016-12-18
Fix typos in [mem.poly.allocator.mem]/9.1 and /9.5

CaseyCarter:
(Obviously a typo since there is no "T" in scope.)

Also trim trailing whitespace in utilities.tex.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -65,7 +65,7 @@ namespace std {
     T exchange(T& obj, U&& new_val);
 
   // \ref{forward}, forward/move
-  template <class T> 
+  template <class T>
     constexpr T&& forward(remove_reference_t<T>& t) noexcept;
   template <class T>
     constexpr T&& forward(remove_reference_t<T>&& t) noexcept;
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -11226,7 +11226,7 @@
 If \tcode{uses_allocator_v<T1,memory_resource*>} is \tcode{false}
 \\
 and
-\tcode{is_constructible_v<T,Args1...>} is \tcode{true},
+\tcode{is_constructible_v<T1,Args1...>} is \tcode{true},
```
CaseyCarter: True non-whitespace change 1 of 1.

source/utilities.tex
```diff
@@ -11253,7 +11253,7 @@
 If \tcode{uses_allocator_v<T2,memory_resource*>} is \tcode{false}
 \\
 and
-\tcode{is_constructible_v<T,Args2...>} is \tcode{true},
+\tcode{is_constructible_v<T2,Args2...>} is \tcode{true},
```
CaseyCarter: True non-whitespace change 2 of 2

PR #1270, opened 2016-12-18
[dcl.init.aggr] Add example for array of unknown bound

### LINKED ISSUE #396: [dcl.init.aggr] resolve core issue 1985: array of unknown bound vs. default member initializer
Add an example indicating that a `brace-or-equal-initializer` for a non-static data member is not actually an initializer, and so does not confer a bound on a member with 'array of unknown bound' type.

jensmaurer:
as non-static data member.
    
Addresses core issue 1985.
    
Fixes #396.

Consistently refer to 'array of unknown bound' not 'array of unknown size'; see [dcl.array].

### DIFF
--- a/source/declarators.tex
+++ b/source/declarators.tex
@@ -2992,7 +2992,7 @@ An aggregate that is a class can also be initialized with a single
 expression not enclosed in braces, as described in~\ref{dcl.init}.
 
 \pnum
-An array of unknown size initialized with a
+An array of unknown bound initialized with a
 brace-enclosed
 \grammarterm{initializer-list}
 containing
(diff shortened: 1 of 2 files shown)


source/declarators.tex
```diff
@@ -3021,6 +3021,17 @@
 for an array of unknown bound.\footnote{The syntax provides for empty
 \grammarterm{initializer-list}{s},
 but nonetheless \Cpp does not have zero length arrays.}
+\begin{note}
+A \grammarterm{brace-or-equal-initializer} for a non-static data
+member does not determine the bound for an array of unknown bound.
```
zygoloid: It would be useful to explain why this follows from the normative wording.

jensmaurer: I've rephrased this.

PR #1274, opened 2016-12-19
Use \tcode for indexed type placeholders (also in math mode)

### LINKED ISSUE #1139: Format indexed variables with code and maths properly

jensmaurer:
The canonical way to format an indexed type placeholder is now $\mcode{T}_i$. This works inside an outside a \tcode area.

Fixes #1139.

### DIFF
--- a/source/conversions.tex
+++ b/source/conversions.tex
@@ -280,16 +280,16 @@ if the following conditions are satisfied,
 % NB: forbid line break between 'where' and 'cv'
 % to stop superscript j from running into
 % descender of p on the previous line.
-where~$cv_i^j$ denotes the cv-qualifiers in the cv-qualification signature of $T_j$:%
+where~$\text{\cv}_i^j$ denotes the cv-qualifiers in the cv-qualification signature of $\tcode{T}_j$:%
 \footnote{These rules ensure that const-safety is preserved by the conversion.}
 
 \begin{itemize}
 \item \tcode{T1} and \tcode{T2} are similar.
 
-\item For every $i > 0$, if \tcode{const} is in $\mathit{cv}_i^1$ then \tcode{const} is in $\mathit{cv}_i^2$, and similarly for \tcode{volatile}.
+\item For every $i > 0$, if \tcode{const} is in $\text{\cv}_i^1$ then \tcode{const} is in $\text{\cv}_i^2$, and similarly for \tcode{volatile}.
 
-\item If the $\mathit{cv}_i^1$ and $\mathit{cv}_i^2$ are different,
-then \tcode{const} is in every $\mathit{cv}_k^2$ for $0 < k < i$.
+\item If the $\text{\cv}_i^1$ and $\text{\cv}_i^2$ are different,
+then \tcode{const} is in every $\text{\cv}_k^2$ for $0 < k < i$.
 \end{itemize}
 
 \begin{note}
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -6554,15 +6554,15 @@ If
 \tcode{A}
 is a type
 \begin{indented}
-$\mathit{cv}_{1,0}$ ``pointer to $\ldots$'' $\mathit{cv}_{1,n-1}$ ``pointer to''
-$\mathit{cv}_{1,n}$ \tcode{T1}
+$\text{\cv}_{1,0}$ ``pointer to $\ldots$'' $\text{\cv}_{1,n-1}$ ``pointer to''
+$\text{\cv}_{1,n}$ \tcode{T1}
 \end{indented}
 and
 \tcode{P}
 is a type
 \begin{indented}
-$\mathit{cv}_{2,0}$ ``pointer to $\ldots$'' $\mathit{cv}_{2,n-1}$ ``pointer to''
-$\mathit{cv}_{2,n}$ \tcode{T2},
+$\text{\cv}_{2,0}$ ``pointer to $\ldots$'' $\text{\cv}_{2,n-1}$ ``pointer to''
+$\text{\cv}_{2,n}$ \tcode{T2},
 \end{indented}
 then the cv-unqualified
 \tcode{T1}
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -1577,13 +1577,13 @@ is zero-based.
 \pnum
 \remarks
 This constructor shall not participate in overload resolution unless
-\tcode{is_default_construct\-ible_v<T$_i$>} is \tcode{true} for all $i$.
+\tcode{is_default_construct\-ible_v<$\tcode{T}_i$>} is \tcode{true} for all $i$.
 \begin{note} This behavior can be implemented by a constructor template
 with default template arguments. \end{note}
 The constructor is explicit if and only if \tcode{T}$_i$ is not implicitly
 default-constructible for at least one $i$.
 \begin{note} This behavior can be implemented with a trait that checks whether
-a \tcode{const T$_i$\&} can be initialized with \tcode{\{\}}. \end{note}
+a \tcode{const $\tcode{T}_i$\&} can be initialized with \tcode{\{\}}. \end{note}
 \end{itemdescr}
 
 \indexlibrary{\idxcode{tuple}!constructor}%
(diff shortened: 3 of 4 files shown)


jensmaurer:
"diffpdf" is mostly silent on this change, except that Ti -> $T_i$ changes are detected.

source/macros.tex
```diff
@@ -156,6 +156,7 @@
 
 % Code and definitions embedded in text.
 \newcommand{\tcode}[1]{\CodeStylex{#1}}
+\newcommand{\mcode}[1]{\text{\tcode{#1}}}   % \tcode in math mode
```
godbyk: The `\text` command works in non-math mode as well. You could simply adjust the definition of `\tcode`:

    \newcommand{\tcode}[1]{\text{\CodeStylex{#1}}}

and it would work just fine in both math mode and text mode.

The only potential downside to using `\text` in text mode, it becomes the equivalent to `\mbox`. The contents of `\mbox` won't be split across lines. If that's a problem, we can further adjust the definition to work around this:

```latex
\newcommand{\tcode}[1]{%
  \ifmmode\text{\CodeStylex{#1}}\else\CodeStylex{#1}\fi
}
```

The benefit to this approach is that it reduces the cognitive load on the author. The author can use a single macro instead of having to decide between two different macros depending on the mode they're in.

source/templates.tex
```diff
@@ -6554,15 +6554,15 @@
...
+$\text{\cv{}}_{1,0}$ ``pointer to $\ldots$'' $\text{\cv{}}_{1,n-1}$ ``pointer to''
+$\text{\cv{}}_{1,n}$ \tcode{T1}
 \end{indented}
 and
 \tcode{P}
 is a type
 \begin{indented}
-$\mathit{cv}_{2,0}$ ``pointer to $\ldots$'' $\mathit{cv}_{2,n-1}$ ``pointer to''
-$\mathit{cv}_{2,n}$ \tcode{T2},
+$\text{\cv{}}_{2,0}$ ``pointer to $\ldots$'' $\text{\cv{}}_{2,n-1}$ ``pointer to''
```
godbyk: I don't see any benefit to typing `\cv{}` instead of `\cv`.

source/macros.tex
jensmaurer: Thanks for the information.

@tkoeppe: Ok to redefine \tcode to the second form (allowing line breaks)? (In non-math-mode, we definitely want linebreaks inside \tcode.)

jensmaurer: We actually have existing evidence that \tcode works in math mode as-is; see e.g. algorithms.tex:3665:
> \bigoh{N\log(N)} comparisons, where $N = \tcode{last - first}$.

\tcode is defined as \CodeStylex, which is defined as \texttt. Maybe that does work in math mode after all.

So, we could simply use `\tcode` instead of `\mcode` everywhere, knowing (through this issue) that we can differentiate between math mode and non-math-mode with `\ifmmode`, as shown above, should the need ever arise. That seems the simplest solution to me.

tkoeppe: @jensmaurer: I agree. If `\tcode` already works out of the box, we should just use that, never mind the name. So for example we can say `$\tcode{T}_i$` rather that `\tcode{T}$_i$`.

godbyk: @jensmaurer A tangent to nitpick the example you raised. I would write it as `$N = \tcode{last} - \tcode{first}$` instead of `$N = \tcode{last - first}$` so that the minus sign gets rendered as a minus sign instead of a code-style hyphen.

example output

On a more general note, I'm happy to help with any LaTeX code or questions.

tkoeppe: @godbyk: I don't think that's correct. There is no mathematical concept of subtracting iterators. However, we do have magic library wording that gives the minus-expression meaning with the library (even if it's not defined for the actual iterator type!).

source/templates.tex
jensmaurer: Fixed.

jensmaurer:
Updated; now using \tcode everywhere.

source/conversions.tex
```diff
@@ -280,16 +280,16 @@
 % NB: forbid line break between 'where' and 'cv'
 % to stop superscript j from running into
 % descender of p on the previous line.
-where~$cv_i^j$ denotes the cv-qualifiers in the cv-qualification signature of $T_j$:%
+where~$\text{\cv}_i^j$ denotes the cv-qualifiers in the cv-qualification signature of $\mcode{T}_j$:%
```
jensmaurer: Uh, sorry, I must have missed that one \mcode. Probably a rebase confusion.

jensmaurer:
Each of the three patches should now be LaTeX-valid, with no traces of `\mcode` left.

tkoeppe:
I may be missing something, but what's the change of the "cv"s about? Is that in line with what we do elsewhere? Is it related to formatting indexed types?

jensmaurer:
For "cv", we use \cv and/or \cv{} elsewhere. Except that we haven't updated the uses inside math mode, yet, which this patch does. It's unrelated to indexed types except that both use math mode, and it's a fairly small drive-by fix here.

source/utilities.tex
```diff
@@ -1249,10 +1249,10 @@
 \begin{itemdescr}
 \pnum
 \returns \tcode{pair<V1, V2>(std::forward<T1>(x), std::forward<T2>(y))},
-where \tcode{V1} and \tcode{V2} are determined as follows: Let \tcode{Ui} be
-\tcode{decay_t<Ti>} for each \tcode{Ti}. If \tcode{Ui} is a specialization
-of \tcode{reference_wrapper}, then \tcode{Vi} is \tcode{Ui::type\&},
-otherwise \tcode{Vi} is \tcode{Ui}.
+where \tcode{V1} and \tcode{V2} are determined as follows: Let $\tcode{U}_i$ be
+\tcode{decay_t<$\tcode{T}_i$>} for each $\tcode{T}_i$. If $\tcode{U}_i$ is a specialization
```
tkoeppe: I'm not sure about this change. I have considered this section myself, and I deliberately didn't want to call the same type both `T1` and `T_1`. We should consistently stick with one name (which is `T1` in this case). It's OK to define *new* types `U_i` as the elements of a pack, say, but unless you say "Let T_1 be T1 and T_2 be T2", I'm concerned that this is making things less clear.

source/utilities.tex
```diff
@@ -14022,10 +14022,10 @@
 \begin{itemize}
 \item \tcode{FD} is the type \tcode{decay_t<F>},
 \item \tcode{fd} is an lvalue of type \tcode{FD} constructed from \tcode{std::forward<F>(f)},
-\item \tcode{Ti} is the $i^{th}$ type in the template parameter pack \tcode{BoundArgs},
-\item \tcode{TiD} is the type \tcode{decay_t<Ti>},
+\item $\tcode{T}_i$ is the $i^{th}$ type in the template parameter pack \tcode{BoundArgs},
+\item $\tcode{TD}_i$ is the type \tcode{decay_t<$\tcode{T}_i$>},
 \item \tcode{ti} is the $i^{th}$ argument in the function parameter pack \tcode{bound_args},
```
tkoeppe: I think here we would also like `\tcode{t}_i` for the arguments.

I'm beginning to think that this changeset is getting too large. I had already been considering each of these passages individually, and each of them requires more care and finetuning that this blanket search-and-replace approach, I think. My recommendation would be to separate the strictly non-observable changes (`\tcode{T}$_i$` to `$\tcode{T}_i$`) from the rest, and revise each section separately for the more invasive changes. For example, #1251 is such a revision of one particular section.

jensmaurer:
Ok, let's close this pull request then. I'll submit a fresh one that only does the non-observable changes, in particular:
- `\tcode{T}$_i$` to `$\tcode{T}_i$`
- `$T_i$` to `$\tcode{T}_i$`
The rest can then be done manually section-by-section.

Issue #1088, opened 2016-11-21
[lib] Inconsistent style in \complexity


jlaire:
Wording of complexity requirements is inconsistent in a couple of ways.

### Noun-phrase or sentence?

"Exactly N applications of `f`."
vs.
"Applies `f` exactly N times.`"
vs.
"It takes approximately N comparisons."

The first form seems to be the most common.

### Time or space complexity, or both?

It is not always made explicit whether the requirement is referring to time or space complexity, or both.

"Linear time." vs. "Linear."
"Constant time." vs. "Constant."

17.5.1.4 says that the *Complexity:* element specifies "the time and/or space complexity of the function", so being explicit about this would be good.

### N log N vs. N log(N)

This is minor, but `N \log(N)`, `N \log N` and `N \log{N}` are all used.

jensmaurer:
Editorial meeting consensus: Use parens for log only when the argument is long and could be ambiguous, e.g. log(N+1). "It takes" is cringe-worthy. 

"Exactly N applications" and related phrases are all ok, no need to harmonize.

Time or space complexity is LWG territory.

jensmaurer:
The remaining question "Time or space complexity, or both?" was sent to LWG.

jensmaurer:
[LWG issue 2949](http://cplusplus.github.io/LWG/lwg-active.html#2949)

PR #1275, opened 2016-12-20
[lib] For showing complexity, use $N \log N$

jensmaurer:
and not $N \log(N)$ or other variants.

Partially addresses #1088.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -3502,7 +3502,7 @@ Sorts the elements in the range
 
 \pnum
 \complexity
-\bigoh{N\log(N)} comparisons, where $N = \tcode{last - first}$.
+\bigoh{N \log N} comparisons, where $N = \tcode{last - first}$.
 \end{itemdescr}
 
 \rSec3[stable.sort]{\tcode{stable_sort}}
@@ -3541,7 +3541,7 @@ Sorts the elements in the range \range{first}{last}.
 \complexity
 At most $N \log^2(N)$
 comparisons, where
-$N = \tcode{last - first}$, but only $N \log(N)$ comparisons if there is enough extra memory.
+$N = \tcode{last - first}$, but only $N \log N$ comparisons if there is enough extra memory.
 
 \pnum
 \remarks Stable~(\ref{algorithm.stable}).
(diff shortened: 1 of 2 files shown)


tkoeppe:
Why drop the parentheses? Mathematical function application notation usually uses parentheses; I would find this change both unconventional and unsightly. Note also that we sometimes say things like "log(a - b)" where we need the parentheses anyway.

jensmaurer:
We say $N \log N$ fairly often (in the containers section); the algorithms section uses $N \log(N)$ more often. This is an attempt to harmonize towards $N \log N$.
My impression is that for single-token arguments in complexity expressions, it is sort-of common to omit the parenthesis, but maybe I'm mistaken.

godbyk:
If you end up using nested parentheses (e.g., `\bigoh{log(a - b)}$`), then you should consider using larger parentheses on the outside (e.g., `\mathscr{O}\big(log(a - b)\big)$)`.

jensmaurer:
@godbyk: Yes, except I thought math mode would auto-scale such things? Anyway, the uses I've come across in our standard use just "N" or "N log N" and then define what N means in an extra phrase. I haven't come across nested parens, yet, except possibly for \bigoh{\log(N)}, which the proposed patch will fix (by removing the inner set of parens).

godbyk:
@jensmaurer I'm afraid there's no such autoscaling. The closest autoscaling would be using `\left(` and `\right)`, but those won't scale up unless the material inside them exceeds the vertical height of the normal parentheses—and `$log(a - b)$` wouldn't. If you wrote `$\left( log ( a - \frac{(n-1)n}{2} ) \right)$` then the outer parentheses would autoscale.

tkoeppe:
You can use `\bigl(`, `\Bigl(`, etc. (but only the first one fits on a normal line). But... that should be considered with utmost care and restraint.

jensmaurer:
Rebased.

burblebee:
I prefer log(N) to log N so as not to run into ambiguity issues with what log a + b means

tkoeppe:
@burblebee: See #1088 for the editorial direction.

@zygoloid: This seems ready to merge.

PR #1285, opened 2016-12-21
[func.bind.bind] Reformat to use math-style indexing.

jensmaurer:
Partially addresses #1139.


jensmaurer:
Before:
page

After:
page

tkoeppe:
This looks like a good idea. @zygoloid: I'll leave it to you to apply this since it's fairly substantial.

@W-E-Brown: for your information, some presentation improvements for index pack elements.

jwakely:
Looks good to me.

My only complaint is that the i subscripts are almost too small to see at some scales, or look like commas, but I still prefer it to the current `TiD` version.

Issue #701, opened 2016-04-15
grammar terms aren't \grammarterm'd


burblebee:
Many references to grammar terms defined in [lex](such as digit) are formatted in roman rather than italics.  This issue is cloned from #296 for the other grammar terms mentioned in that issue.

Note that we have separate issues for:
- universal-character-name #296
- literal #322

tkoeppe:
How much of this is still a problem?

jensmaurer:
> `grep -- - *.tex | fgrep -w '\term' `

still shows a few (very few) occurrences that need fixing (assuming that grammarterms often have a hyphen in their spelling).

PR #1286, opened 2016-12-21
[except] Replace \term and \textit

jensmaurer:
with \grammarterm or \defn/x as appropriate.

Also add an index entry.

Partially addresses #701.
Partially addresses #329.

### DIFF
--- a/source/exceptions.tex
+++ b/source/exceptions.tex
@@ -114,10 +114,10 @@ lab:  try {
   T1 t1;
   try {
     T2 t2;
-    if (@\textit{condition}@)
+    if (@\grammarterm{condition}@)
       goto lab;
-    } catch(...) { /* @\textit{handler 2}@ */ }
-  } catch(...) { /* @\textit{handler 1}@ */ }
+    } catch(...) { @\tcode{/* handler 2 */}@ }
+  } catch(...) { @\tcode{/* handler 1 */}@ }
 \end{codeblock}
 
 Here, executing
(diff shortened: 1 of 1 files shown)


source/exceptions.tex
```diff
@@ -114,10 +114,10 @@
   T1 t1;
   try {
     T2 t2;
-    if (@\textit{condition}@)
+    if (@\grammarterm{condition}@)
       goto lab;
-    } catch(...) { /* @\textit{handler 2}@ */ }
-  } catch(...) { /* @\textit{handler 1}@ */ }
+    } catch(...) { /* handler 2 */ }
```
tkoeppe: The comments should be in code font here. Have a look at the new `\commentellip` macro I made for inspiration on how to achieve this.

jensmaurer: The following text refers to "handler 1" and "handler 2" in italics. If we make the text in the comments \tcode, should I adjust the following text to \tcode, too?

jensmaurer: Assuming the answer is "yes"...

tkoeppe: The main idea here is to make the `/*` appear as code. So it should be something like `@\tcode{/* ... */}@` I think.

jensmaurer: Fixed.

tkoeppe:
That's probably good. For good measure, could you please attach a diff screenshot?

jensmaurer:
For extra fun, this part has a page break in the middle...

page1
page2

tkoeppe:
Thanks. By the way, you should probably switch your diff mode to "appearance".

PR #1287, opened 2016-12-21
[utilities] Tweaks for better linebreaks

jensmaurer:
avoiding 'Overfull hbox' warnings.

Partially addresses #693.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -2896,11 +2896,11 @@ The constructor is explicit if and only if
 \begin{itemdescr}
 \pnum
 \effects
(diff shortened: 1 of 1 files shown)


tkoeppe:
All these overfull boxes in utilities.tex are extremely mild. I'll take some of the straight-forward improvements and fixes, but let's leave the remainder until just before publication.

jensmaurer:
I understand that the usual approach is to leave Overfull hboxes "to just before publication". However, I don't have the impression that we have a lot of motion-induced churn in the majority of our text, so addressing them incrementally will at least a large pile of them "just before publication".

PR #1288, opened 2016-12-22
[string.view] Remove unnecessary std:: prefixes.

Eelis:
Same rationale as #1085.

### DIFF
--- a/source/strings.tex
+++ b/source/strings.tex
@@ -4661,8 +4661,8 @@ In the rest of this section, the type of the char-like objects held in a \tcode{
 
 \pnum
 \begin{note}
-The library provides implicit conversions from \tcode{const charT*} and \tcode{std::basic_string<charT, ...>} to \tcode{std::basic_string_view<charT, ...>} so that user code can accept just \tcode{std::basic_string_view<charT>} as a non-templated parameter wherever a sequence of characters is expected.
-User-defined types should define their own implicit conversions to \tcode{std::basic_string_view} in order to interoperate with these functions.
+The library provides implicit conversions from \tcode{const charT*} and \tcode{basic_string<charT, ...>} to \tcode{basic_string_view<charT, ...>} so that user code can accept just \tcode{basic_string_view<charT>} as a non-templated parameter wherever a sequence of characters is expected.
+User-defined types should define their own implicit conversions to \tcode{basic_string_view} in order to interoperate with these functions.
 \end{note}
 
 \pnum


source/strings.tex
```diff
@@ -4661,8 +4661,8 @@
 
 \pnum
 \begin{note}
-The library provides implicit conversions from \tcode{const charT*} and \tcode{std::basic_string<charT, ...>} to \tcode{std::basic_string_view<charT, ...>} so that user code can accept just \tcode{std::basic_string_view<charT>} as a non-templated parameter wherever a sequence of characters is expected.
-User-defined types should define their own implicit conversions to \tcode{std::basic_string_view} in order to interoperate with these functions.
+The library provides implicit conversions from \tcode{const charT*} and \tcode{basic_string<charT, ...>} to \tcode{basic_string_view<charT, ...>} so that user code can accept just \tcode{basic_string_view<charT>} as a non-templated parameter wherever a sequence of characters is expected.
+User-defined types should define their own implicit conversions to \tcode{basic_string_view} in order to interoperate with these functions.
 \end{note}
```
jensmaurer: Hm... This is a note explaining how to interoperate with string views in user code. Assuming no "using namespace std" is present, user code would need the `std::` prefix when mentioning `basic_string_view`.

On the other hand, talking about `basic_string_view<charT>` as a "non-templated parameter" certainly means this is not verbatim user code (because `charT` is used as a template parameter in this section).

Net result: I'm fine with the proposed change.

tkoeppe: I had similar misgivings about this change. Since it's in a note, and it is talking about user code (implicitly using `charT` as a user-chosen parameter), I'd be happy with the status quo.

Eelis: @tkoeppe: That applies to the last 2 of the 4 removals, right? If so, would the patch be more acceptable to you if it only removed the first 2 occurrences?

jensmaurer: @Eelis: Changing the `std::` convention mid-sentence seems like the worst of all options.

Eelis: @jensmaurer It's not "changing the convention" if the convention is "no prefix unless the occurrence in question is describing user code" and if only the last two occurrences in the sentence describe user code.

jensmaurer: @Eelis: Agreed, but it looks really odd to have the same standard library name prefixed by `std::` and not prefixed in the same sentence. (That would be the result with your "2 of 4" proposal.)

tkoeppe: Let me close this PR in the interest of not diluting attention too much for what are basically subjective issues. Feel free to reopen if you feel really strongly, but in that case I'd leave this for the project editor to process.

PR #1289, opened 2016-12-22
[strings] Remove remarks that repeat what was stated a few lines before.

Eelis:
This removes remarks that literally repeat what was just stated, but leaves intact instances where the same remarks do _not_ repeat what was just stated, e.g. http://eel.is/c++draft/string.cons#28.

Representative diff:
diff


jensmaurer:
This looks like an improvement to me. If someone feels strongly about the removed text, we could convert it into notes instead.

PR #1292, opened 2016-12-27
[meta.trans.other] add missing cv for common_type

timsong-cpp:
P0435R1 says 'cv void', but the cv part was dropped.


jensmaurer:
This looks good to me. [P0435R1](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0435r1.pdf) clearly shows the surviving "cv" here, amid some strikeout, so this is very likely an editorial oversight when applying the paper.

P0435R1 was voted into the working draft in Issaquah via LWG motion 29.

W-E-Brown:
I agree that this seems to have been "an editorial oversight when applying the paper."

tkoeppe:
@W-E-Brown: Thanks, merging.

PR #1300, opened 2016-12-30
[expr] Split off definition of 'composite pointer type' into dedicated paragraph.

Eelis:
This splits up a long paragraph that defines two complex terms ("cv-combined type" and "composite pointer type") using itemized lists. It reads and looks so much better to me this way that I suspect the \pnum was accidentally lost at some point.

diff


jensmaurer:
Looks good to me.

PR #1302, opened 2016-12-31
Replace 'sub-clause' by 'subclause'

### LINKED ISSUE #497: Editorial: "sub-clause" vs "subclause"
The standard uses the term "subclause" 107 times, and "sub-clause" 7 times.
_It also uses "section" 102 times_.

At the very least, the uses of "sub-clause" should be changed to "subclause", but changing the mentions of "section" might be considered as well.

alexanderkscott:
Fixes #497.


jensmaurer:
Looks good to me. I'm not sure how I missed these in #1040.

PR #1303, opened 2016-12-31
[limits] Use normal font in "try block"

alexanderkscott:
This is the only place where "try block" uses code font for the first word.


jensmaurer:
This change looks good.

[except]p2 says in a note that "try block" (without code font) means _try-block_ or _function-try-block_ (grammar terms).

PR #1316, opened 2017-01-01
Replace \nonterminal with \grammarterm or \cvqual, as appropriate.

### LINKED ISSUE #1312: Replace \nonterminal with \grammarterm or \cv or \cvqual or ...
... and remove the macro afterwards, taking care of the use(s) in macros.tex itself.

\nonterminal in running text should be \grammarterm these days.

jensmaurer:
Fixes #1312.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -609,7 +609,7 @@ exception specification of that constructor.
 
 \pnum
 A member shall not be declared with the \tcode{extern}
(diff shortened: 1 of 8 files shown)


tkoeppe:
Can you run diffpdf?

jensmaurer:
Yes. There are 14 pages with differences, all of them due to the fact that \nonterminal uses a slightly smaller font than \grammarterm or \cvqual.

jensmaurer:
(I've rebased the patch.)

tkoeppe:
Oh, very good, we should fix that. Thanks!

PR #1317, opened 2017-01-01
[sf.cmath] Fix 'where' associations of Greek variable names.

### LINKED ISSUE #1291: [sf.cmath] association of Greek math variables with parameters fails
We say, for example, `where $nu$ is \tcode{nu} and $x$ is \tcode{x}`, but we should say `where $\nu$ is \tcode{nu} and $x$ is \tcode{x}` so that we associate the Greek symbol used in the formula with the parameter, not some italics "nu" that appears nowhere else.

This affects the entire section.

jensmaurer:
Fixes #1291.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -10684,7 +10684,7 @@ of their respective arguments
 \]
 where
 $k$ is \tcode{k} and
-$nu$ is \tcode{nu}.
+$\nu$ is \tcode{nu}.
 
 \pnum See also \ref{sf.cmath.ellint_3}.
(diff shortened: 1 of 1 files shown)


tkoeppe:
@W-E-Brown: Do you have any objections to this change?

W-E-Brown:
Thx for asking; as I wrote earlier (attached to #1291): ".... I'd rather see all these phrases gone because I find them redundant.... But yes, if we must keep them, then we must certainly correct these phrases."

PR #1318, opened 2017-01-01
[conv.lval] Split off description of conversion result into dedicated paragraph.

Eelis:
This is a long paragraph that describes two things:

- the conditions under which the conversion does not access the referenced object's value;
- the rules for determining the result of the conversion.

Doing this in two paragraphs seems clearer to me:

diff


jensmaurer:
Fine with me.

PR #1319, opened 2017-01-01
Add missing hyphen in 'floating-point'.

Eelis:
For consistency. Addresses #440.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3836,7 +3836,7 @@ ones' complement and signed magnitude representations for integral types.
 \end{example}
 
 \pnum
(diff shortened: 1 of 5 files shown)


jensmaurer:
Looks good to me.

tkoeppe:
Does the index need updating, too?

Eelis:
Ah, I would say so, yes. Updated.

tkoeppe:
Thanks, that looks good to me, too. I'll leave it to @zygoloid to make changes to the fundamental C++ types, though.

jensmaurer:
This looks good to me.

PR #1322, opened 2017-01-02
[stmt.stmt] Remove stray {} after \grammarterm.

### LINKED ISSUE #1320: wrongly added {}
[stmt.stmt/2](https://github.com/cplusplus/draft/blob/master/source/statements.tex#L36):
there is a redundant `{}` in `\grammarterm{}{condition}{s}`.
Also in `\grammarterm{}{declarator}`.
Also in stmt.stmt/3-5.
Also in stmt.select/1, for `\grammarterm{}{statement}`.

I don't know whether these are intended (to temporarily disable the `\grammarterm` macro) or by mistakes.

In stmt.select/1, the statement before [Example, are the `selection-statement` and `compound-statement` need to be marked by `\grammarterm`?

jensmaurer:
Fixes #1320.

### DIFF
--- a/source/statements.tex
+++ b/source/statements.tex
@@ -33,9 +33,9 @@ The optional \grammarterm{attribute-specifier-seq} appertains to the respective 
 
 \pnum
 \indextext{\idxgram{condition}{s}!rules for}%
-The rules for \grammarterm{}{condition}{s} apply both to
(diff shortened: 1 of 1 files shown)


jensmaurer:
diffpdf shows no differences.

PR #1323, opened 2017-01-02
[re.submatch.op] Remove excessive parentheses in "Returns:" element.

Eelis:
(... and remove an invisible but unnecessary pair of curlies.)


jensmaurer:
Looks good to me.

PR #1324, opened 2017-01-02
[complex.ops] Don't repeat declaration in "Effects:" element.

Eelis:
diff


jensmaurer:
Since "Equivalent to:" has special (and fairly broad) meaning, we usually shy away from introducing it editorially. I'd support this change if you could leave the introductory phrase "inserts the complex number ..." untouched (except for a capital "I" at the start) and only delete the declaration and the function-body curly braces.

Eelis:
Makes sense, fixed.

PR #1327, opened 2017-01-04
[dcl.init] Fix a typo in definition of const-default-constructible

k-satoda:
The typo was from [p0490r0](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0490r0.html).


jwakely:
@jensmaurer this looks right to me.

PR #1329, opened 2017-01-04
[rand.util.seedseq] Consistently format T(...) as ordinary math, not code.

Eelis:
For consistency with the start of the paragraph, where T is defined as an ordinary math function.

Using `\left(` and `\right)` instead of plain `(` and `)` here seems to have no effect other than introducing unwanted whitespace between the T and opening parenthesis.

Diff:
diff

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -4067,10 +4067,10 @@ template<class RandomAccessIterator>
    calculate values
    \begin{eqnarray*}
      r_1 & = &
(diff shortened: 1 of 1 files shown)


jensmaurer:
Looks good to me.

(As a side note, for "t" I would much appreciate a math-style open brace similar to what is used for r_2 instead of the conditional-C++-operator-in-math style.

PR #1333, opened 2017-01-05
[string.io] Don't refer to sentry by name it was not given.

Eelis:
[[string.io]#2](http://eel.is/c++draft/string.io#2) and [#7](http://eel.is/c++draft/string.io#7) refer to "sentry object k", but the sentry object was never given that name.


jensmaurer:
Looks good to me.

PR #1334, opened 2017-01-05
[class.gslice.overview] Replace $$...$$ with \[...\].

Eelis:
The use of double dollars for math is an old TeX practice that is discouraged in LaTeX.

This patch removes the only two uses of double dollar math in the document.

No visual difference (because `\[...\]` is equivalent here).

(Incidentally, this helps my HTML version of the standard, because cxxdraft-htmlgen uses HaTeX, a Haskell LaTeX parsing library that does not support archaic double dollar math.)

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -8023,10 +8023,8 @@ It is useful for building multidimensional array classes using
 the
 \tcode{valarray}
 template, which is one-dimensional.
-The set of one-dimensional index values specified by a
-\tcode{gslice}
-are $$k = s + \sum_ji_jd_j$$
-% \$k = s + sum from j \{ i sub j d sub j \}\$
+The set of one-dimensional index values specified by a \tcode{gslice} are
+\[ k = s + \sum_j i_j d_j \]
 where the multidimensional indices $i_j$ range in value from
 0 to $l_{ij} - 1$.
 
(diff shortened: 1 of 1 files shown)


source/numerics.tex
```diff
@@ -8025,7 +8025,7 @@
 template, which is one-dimensional.
 The set of one-dimensional index values specified by a
 \tcode{gslice}
-are $$k = s + \sum_ji_jd_j$$
+are \[k = s + \sum_ji_jd_j\]
```
tkoeppe: While you're here, make this a bit easier to read:

    \[ k = s + \sum_j i_j d_j \]

Also add spacing after `\[` below.

Eelis: Sure thing. Shall I also remove the comment on line 8029? Doesn't look too useful.

tkoeppe: Yes please. That must be a left-over from troff.

Eelis: Done.

source/numerics.tex
```diff
@@ -8023,10 +8023,10 @@
 the
 \tcode{valarray}
 template, which is one-dimensional.
-The set of one-dimensional index values specified by a
-\tcode{gslice}
-are $$k = s + \sum_ji_jd_j$$
-% \$k = s + sum from j \{ i sub j d sub j \}\$
+The set of one-dimensional index values specified by a \tcode{gslice} are
+
```
tkoeppe: Wait, we didn't use to have paragraph breaks here originally, why did you introduce them?

Eelis: Oh, I just tried to make it consistent with the other one, which is also an equation in the middle of a sentence, and already had the empty lines around it. If empty lines around \[...\] introduce paragraph breaks that are unwanted in those cases, I guess the empty lines should be removed there, as well?

tkoeppe: Yes, they should both go. Both expressions are part of the single paragraph here, no breaks. (Even if the breaks don't show visually.)

Eelis: Roger that, updated.

PR #1338, opened 2017-01-06
[string.insert] Typo fix for argument name

Bekenn:


jwakely:
I was about to merge this, but I think it would actually be better to rename the argument from `pos1` to `pos` because  there is only one position argument, so we don't need to distinguish `pos1` and `pos2`.

Bekenn:
There are several other examples throughout basic_string (and basic_string_view) that also decorate parameter names in order to match what's used in the more versatile overloads, but it's not at all consistent; this was the simplest fix.

Bekenn:
There are a bunch of `n1`s floating around, too...

PR #1339, opened 2017-01-07
[dcl.enum, conv.prom] Style min/max subscripts as labels, not variables.

Eelis:


source/declarations.tex
```diff
@@ -2129,14 +2129,14 @@
 the enumeration are the values of the underlying type. Otherwise,
 for an enumeration where $e_\mathit{min}$ is the smallest enumerator and
 $e_\mathit{max}$ is the largest, the values of the enumeration are the
-values in the range $b_{min}$ to $b_{max}$, defined as follows: Let $K$
+values in the range $b_\mathit{min}$ to $b_\mathit{max}$, defined as follows: Let $K$
```
tkoeppe: I would *strongly* prefer `b_\text{min}`. The "min" is not a variable, but a label. It should not be styled as a variable.

jensmaurer:
Looks good to me.

PR #1340, opened 2017-01-07
[rand.req.eng] Omit superfluous dollar-math wrapping inside \bigoh.

Eelis:
Compare: [line 2070](https://github.com/cplusplus/draft/blob/master/source/numerics.tex#L2070)

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -1912,21 +1912,21 @@ according to Clause~\ref{strings} and Clause~\ref{input.output}.
     with the same initial state
     as all other default-constructed engines
     of type \tcode{E}.
-  & \bigoh{$\mbox{size of state}$}
+  & \bigoh{\mbox{size of state}}
   \\ \rowsep
 \tcode{E(x)}
 \indextext{copy constructor!random number engine requirement}
   &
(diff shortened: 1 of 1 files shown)


source/numerics.tex
```diff
@@ -1912,21 +1912,21 @@
     with the same initial state
     as all other default-constructed engines
     of type \tcode{E}.
-  & \bigoh{$\mbox{size of state}$}
```
tkoeppe: Why `\mbox` and not `\text`?

Eelis: Beats me, I tried reading https://tex.stackexchange.com/questions/70632/difference-between-various-methods-for-producing-text-in-math-mode but it gave me a headache :)

tkoeppe:
Let's leave this for now. I can revise the section systematically if and when I get to it.

Eelis:
I'm curious, is that due to a policy that an issue in a line of code may only be fixed if all other potential issues on the same line are fixed at the same time?

tkoeppe:
Just triaging the deluge of pull requests. I would make this change differently and probably will at some point, so I figured time is better spend on other things.

PR #1342, opened 2017-01-08
Preserve whitespace after \shr and \shl where appropriate.

Eelis:
Diffs:
diff

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -2633,7 +2633,7 @@ can be called by
 the function signature
 \tcode{basic_ostream\& stream::operator\shl(ios_base\& (*)(ios_base\&))}
 to permit expressions of the form
-\tcode{cout \shl dec}
+\tcode{cout \shl{} dec}
 to change the format flags stored in
 \tcode{cout}.}.
 \end{itemdescr}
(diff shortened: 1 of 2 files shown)


source/iostreams.tex
```diff
@@ -2633,7 +2633,7 @@
 the function signature
 \tcode{basic_ostream\& stream::operator\shl(ios_base\& (*)(ios_base\&))}
 to permit expressions of the form
-\tcode{cout \shl dec}
+\tcode{cout \shl{} dec}
```
tkoeppe: What is the point of this macro. Just say `<<` here (and everywhere).

Eelis: Alright, I'll give it a try

Eelis: The macros are defined as:

  \newcommand{\shl}{<{<}}
  \newcommand{\shr}{>{>}}

I tried removing the extra curlies, and it changes the order of some index entries. So I guess that's what they're for.

jwakely:
Don't they also prevent LaTeX replacing them with guillemets?

source/iostreams.tex
jwakely: Don't they also prevent LaTeX replacing them with guillemets in some contexts?

Eelis: I did an "appearance" check with diffpdf, and no guillemets showed up. Maybe because the macros are only ever used inside \tcode and similar.

tkoeppe: Yeah, I think the idea that anyone needs to prevent guillemets is archaic and pointless, and so we really should just get rid of those two macros.

Eelis: Here's an example of the index entry order difference I mentioned:
diff
If that change is deemed acceptable, then I can submit a pull request to remove them altogether.

tkoeppe: Either index looks fine.

jwakely: I've seen them causing problems in the standard, but maybe it was just the one use in lex.tex or something like that.

tkoeppe: @jwakely: Maybe they arose when a `\tcode` or equivalent was missing? There are no guillemets now.

Eelis:
Fixed by #1346

PR #1345, opened 2017-01-09
[rand.util.seedseq] Replace nested ternary expressions with definition by cases.

Eelis:
As suggested by @jensmaurer in #1329. Diff:

diff

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -4057,9 +4057,17 @@ template<class RandomAccessIterator>
    let $p = (n - t) / 2$
    and let $q = p + t$,
    where
-   \[%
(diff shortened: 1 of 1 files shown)


tkoeppe:
This should be styled roughly as follows:

    t = { a <tab> for ..., \n b <tab> for ..., and \n c <tab> for ... . }

(if you see what I mean)

Eelis:
You mean for all such definitions by cases in the document?

tkoeppe:
No, just in this case. I don't want the second column to start with a dangling comma.

In pseudo-latex:

    t = \begin{cases} 1 & \text{for $a = 1$,}\\ 2 & \text{for $a > 3$, and}\\ 3 & \text{otherwise.} \end{cases}

just to demonstrate how to integrate sentence punctuation into cases.

PR #1346, opened 2017-01-09
Remove \shl and \shr.

Eelis:
Visible differences are:

1) inserts missing whitespace:

ws

2) reorders some index entries, e.g:

index


tkoeppe:
Are those two the only visible differences?

Eelis:
According to DiffPDF, yes.

zygoloid:
These existed to avoid the formation of guillemets, which we're disabling through a different mechanism these days.

PR #1355, opened 2017-01-11
Do not italicize colons when defining grammar non-terminals or

### LINKED ISSUE #1351: Inconsistent italicization of \nontermdef's colon in bnf and bnfkeywordtab envs
Compare:
- [hexadecimal-digit](https://github.com/cplusplus/draft/blob/master/source/lex.tex#L897), defined with a bnf environment:
hexdigit
- [preprocessing-op-or-punc](https://github.com/cplusplus/draft/blob/master/source/lex.tex#L794), defined with a bnfkeywordtab environment:
ppoporpunc

jensmaurer:
following the "[ Note" and "[ Example" introducers.

Fixes #1351.

In "Appearance" mode, diffpdf shows everything (including spacing differences caused by the upright colon). In "characters" mode, there are just six pages of differences where a word was hyphenated the other way.

Sample change:

page

An additional "Overfull hbox" appeared in exceptions.tex, which took a reformulation of a note to fix. (Hyphenation hints were ineffective, for unknown reasons.)

### DIFF
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -198,7 +198,7 @@
 \newcommand{\leftshift}[1]{\ensuremath{\, \mathsf{lshift}_#1 \,}}
 
 %% Notes and examples
-\newcommand{\noteintro}[1]{[\,\textit{#1:}\space}
+\newcommand{\noteintro}[1]{[\,\textit{#1}:\space}
 \newcommand{\noteoutro}[1]{\textit{\,---\,end #1}\,]}
 \newenvironment{note}[1][Note]{\noteintro{#1}}{\noteoutro{note}\xspace}
 \newenvironment{example}[1][Example]{\noteintro{#1}}{\noteoutro{example}\xspace}
@@ -393,9 +393,9 @@
 
 \newenvironment{bnfbase}
  {
- \newcommand{\nontermdef}[1]{\nonterminal{##1}\indexgrammar{\idxgram{##1}}:}
+ \newcommand{\nontermdef}[1]{\nonterminal{##1\itcorr}\indexgrammar{\idxgram{##1}}\textnormal{:}}
  \newcommand{\terminal}[1]{{\BnfTermshape ##1}\xspace}
- \newcommand{\descr}[1]{\normalfont{##1}}
+ \newcommand{\descr}[1]{\textnormal{##1}}
  \newcommand{\bnfindentfirst}{\BnfIndent}
  \newcommand{\bnfindentinc}{\BnfInc}
  \newcommand{\bnfindentrest}{\BnfRest}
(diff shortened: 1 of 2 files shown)


tkoeppe:
@jensmaurer: No, we never reached a conclusion on that. It's one of my big plans for C++20.

tkoeppe:
I think a change of this nature will require some careful fine tuning. For example, the colon after `q-char` currently clashes with the "r"; it might be good to think about a pragmatic solution that still looks good.

jensmaurer:
Well, there is a tiny amount of space between the "r" and the colon...

page

source/macros.tex
```diff
@@ -393,9 +393,9 @@
 
 \newenvironment{bnfbase}
  {
- \newcommand{\nontermdef}[1]{\nonterminal{##1}\indexgrammar{\idxgram{##1}}:}
+ \newcommand{\nontermdef}[1]{\nonterminal{##1}\indexgrammar{\idxgram{##1}}\textnormal{:}}
```
tkoeppe: Try inserting an italic correction, see if that looks better (`\itcorr[1]`).

source/macros.tex
```diff
@@ -198,7 +198,7 @@
 \newcommand{\leftshift}[1]{\ensuremath{\, \mathsf{lshift}_#1 \,}}
 
 %% Notes and examples
-\newcommand{\noteintro}[1]{[\,\textit{#1:}\space}
+\newcommand{\noteintro}[1]{[\,\textit{#1}:\space}
```
tkoeppe: I've also always wanted to reduce the excessive space after the `[` and before the `]`. Instead of `\,`, I figured something like a `\kern0.25pt` or so would suffice. But I'll need to convince @zygoloid  of that first.

source/macros.tex
jensmaurer: Yeah, I had the same idea. And adding \/ or \itcorr indeed fixes it. Screenshot forthcoming.

source/macros.tex
jensmaurer: Hm... We could make the spacing slightly smaller, but I think having some spacing is good.

tkoeppe: Yes, some, but less. I can show you some variations at some point. But let's separate that from the bnf-colon issue.

jensmaurer:
Added italics correction:
page

tkoeppe:
Much better!

tkoeppe:
I committed just the `\nontermdef` part as ffb9408d01061c336807333bf2cd2743acb8df14. Let's talk about the Notes/Examples business in person. Thanks!

tkoeppe:
I'm hijacking this PR to post some screenshots:

**Original (`\,`):**

image

**No space:**

image

**Small space (itcorr):**
image

**No space at the beginning, itcorr at end:**

image

PR #1357, opened 2017-01-13
Remove stray whitespace around scope resolution operator.

Eelis:
diff

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -1483,7 +1483,7 @@ of the type of the object for which the member function is called.
 the
 \grammarterm{template-argument}
 for
-\tcode{Array<T>\,::\,op\-er\-a\-tor\,[]\,()}
+\tcode{Array<T>::op\-er\-a\-tor\,[]\,()}
 will be determined by the
 \tcode{Array}
 to which the subscripting operation is applied.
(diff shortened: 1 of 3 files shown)


source/templates.tex
```diff
@@ -1483,7 +1483,7 @@
 the
 \grammarterm{template-argument}
 for
-\tcode{Array<T>\,::\,op\-er\-a\-tor\,[]\,()}
+\tcode{Array<T>::op\-er\-a\-tor\,[]\,()}
```
Eelis: Unrelatedly, this line has an instance of a function being referred to as "f()" even though it doesn't have zero parameters and cannot be called with zero arguments. That practice seems a bit sloppy for the standard (and certainly isn't used consistently in the document). Thoughts?

jensmaurer: Please open an editorial issue for this (I thought we already had one, but maybe not.)

jwakely: Yes that's definitely wrong (but probably not the only place we do it).

jwakely:
We aren't (currently) benefiting from adding those hints in those locations, so removing them to get rid of the whitespace seems better.

jensmaurer:
Looks good to me.

PR #1358, opened 2017-01-13
[re] Remove whitespace before \colcol.

cpplearner:


jensmaurer:
Wait. We usually use "::", but sometimes we use \colcol, for no apparent reason.

So, I'd much prefer to globally replace \colcol with :: (and remove the whitespace in that process) instead of just removing whitespace, but leaving \colcol in.

jensmaurer:
The patch as proposed looks good to me.

tkoeppe:
This seems like churn. I'd say if you want to make progress here, replace the `\colcol` with plain double colons while you're at it.

tkoeppe:
Looks good, thanks. Can you just check quickly that this doesn't introduce any paragraph composition problems (overlong lines)? (E.g. by comparing the output of `fgrep std.log -e Overfull -e .tex` before and after.)

cpplearner:
No. The output is the same before and after.

tkoeppe:
Thanks!

PR #1364, opened 2017-01-14
Remove stray whitespace after opening parentheses.

Eelis:


tkoeppe:
Any visual diffs?

Eelis:
Yeah, these spaces are visible.

tkoeppe:
Yes, right, do you have a screenshot? It looks sane.

source/utilities.tex
```diff
@@ -15881,7 +15881,7 @@
   \tcode{is_constructible_v<T,}\br
   \tcode{Args...>} is \tcode{true} and the variable
   definition for \tcode{is_constructible}, as defined below, is known to call
-  no operation that is not trivial (~\ref{basic.types},~\ref{special}). &
+  no operation that is not trivial (\ref{basic.types},~\ref{special}). &
```
jwakely: We usually do `blah~(\ref` so it looks as though the `~` just got put in the wrong place here. Do we want to move it rather than removing it?

tkoeppe: Oh yes, that's where that comes from. I think we can do without those non-breaking spaces in general, though: Either there's no linebreak and all is well, or we do need a line break, and in that case it's better to break before the paren than have an overfull box.

jensmaurer: Let me disagree here. LaTeX might very well reflow the entire paragraph (without causing an overfull hbox) to avoid a linebreak at the tilde. I think it's good practice not to have a cross-reference at the start of a line. Is there a way to tell LaTeX "I don't want you to linebreak here, but before you produce an Overfull hbox, do it nonetheless"? If we can't do that, we have to consider the few cases where ~\ref produces an Overfull hbox on an individual basis.

tkoeppe: You could make a space with a positive but non-infinite penalty. But that's going too far; we can keep the non-breaking space and remove it on a case-by-case basis.

Eelis:
diff

source/utilities.tex
Eelis: @jwakely Good point, thanks. Patch amended.

PR #1365, opened 2017-01-15
consistent whitespace in \tcode types

AlisdairM:
Where a type contains embedded whitespace, such as 'long long',
the tex files have been inconsistent in rendering this as
'\tcode{long long}' or '\tode{long} \tcode{long}', even within
a list of such types.

This patch consistently reduces the numbwr of \tcode macros
by combining whitespaced keywords into a single \txode element.
There may be some minimal kerning adjustments through the use of
non-proportional whitespace, but I don't immediately see any
tezt reflowing.  The end result is source for the standard that
is easier to grep.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3481,11 +3481,11 @@ or functions (\ref{dcl.fct}).
 For any object (other than a base-class subobject) of trivially copyable type
 \tcode{T}, whether or not the object holds a valid value of type
 \tcode{T}, the underlying bytes~(\ref{intro.memory}) making up the
-object can be copied into an array of \tcode{char} or \tcode{unsigned}
-\tcode{char}.\footnote{By using, for example, the library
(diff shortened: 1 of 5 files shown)


jensmaurer:
Looks good to me. The patch performs as advertised.

Eelis:
Typos in the commit text:
- "numbwr"
- "tezt"
- "\txode"

AlisdairM:
Fixed typos in commit message, and re-pushed.  No change of content.

tkoeppe:
Do you have a diffpdf?

tkoeppe:
@AlisdairM: Ping

tkoeppe:
I'm not convinced this actually looks better -- a proportional space looks tidier in body text --, but there is certainly something to be said for maintainability, and the majority of use cases already put multiword types into a single `\tcode`, so I'll accept this.

PR #1370, opened 2017-01-16
[temp] Replace \term with \defn or \placeholder

jensmaurer:
Thereby also improving the index.

Partially addresses #329.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -8,7 +8,7 @@
 \indextext{type generator|see{template}}
 
 \pnum
-A \term{template} defines a family of classes, functions, or variables, or an alias for a
+A \defn{template} defines a family of classes, functions, or variables, or an alias for a
 family of types.
 
 \indextext{\idxcode{template}}%
@@ -56,10 +56,8 @@ A \grammarterm{template-declaration} is also a definition
 if its \grammarterm{declaration} defines a function, a class, a variable, or a
 static data member. A declaration introduced by a template declaration of a
 \indextext{variable template!definition of}%
-\indextext{template!variable}%
-\indextext{template!static data member}%
-variable is a \term{variable template}. A variable template at class scope is a
-\term{static data member template}.
+variable is a \defnx{variable template}{template!variable}. A variable template at class scope is a
+\defnx{static data member template}{template!static data member}.
 
 \begin{example}
 \begin{codeblock}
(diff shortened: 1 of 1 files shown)


source/templates.tex
```diff
@@ -3556,11 +3561,11 @@
 \rSec2[temp.dep]{Dependent names}
 
 \pnum
-\indextext{name!dependent}%
+\indextext{dependent name|see{name!dependent}}%
```
Eelis: Should this be `see{name, dependent}`?

jensmaurer: Yes, good catch.

jensmaurer:
Updated and rebased.

source/templates.tex
```diff
@@ -3704,7 +3708,8 @@
 
 \pnum
 A name refers to the
-\term{current instantiation}
+\indextext{current instantiation|see{instantiation, current}}
+\defnx{current instantiation}{instantiation!current}
```
zygoloid: I don't think grouping this under "instantiation" is appropriate; the "current instantiation" is a notion separate from instantiation itself (in fact, it's most relevant when checking a template definition prior to performing any instantiation).

jensmaurer: You're right.

jensmaurer:
Fixed as requested.

source/templates.tex
```diff
@@ -3808,9 +3812,9 @@
 \end{note}
 
 \pnum
-\indextext{instantiation!member of the current}%
+\indextext{member of the current instantiation|see{instantiation, member of the current}}%
 A name is a
-\term{member of the current instantiation}
+\defnx{member of the current instantiation}{instantiation!member of the current}
```
zygoloid: This should be `{current instantiation!member of}` or similar.

source/templates.tex
```diff
@@ -3863,14 +3867,14 @@
 \end{codeblock}
 \end{example}
 
-\indextext{instantiation!dependent member of the current}%
-A name is a \term{dependent member of the current instantiation} if it is a
+\indextext{dependent member of the current instantiation|see{instantiation, dependent member of the current}}%
+A name is a \defnx{dependent member of the current instantiation}{instantiation!dependent member of the current} if it is a
```
zygoloid: Likewise.

jensmaurer:
Fixed the two "member of the current instantiation" issues. The index looks sane now (in this regard).

PR #1371, opened 2017-01-17
[dcl.ambig.res] Describe example ambiguous case more clearly.

Eelis:
First I thought "hey that semicolon should be styled like the one on line 368", but then I noticed that it shouldn't be a semicolon at all.


jensmaurer:
Looks good to me.

Issue #1373, opened 2017-01-17
[expr.cond]/4.3 simplification


cubbimew:
Should the simplification from https://groups.google.com/a/isocpp.org/forum/#!topic/std-discussion/hFSv84WY14o (which pointed out awhile ago that [expr.cond]/4.3.1 is redundant) be applied to [expr.cond]/4.3 as an editorial? That wall of text could use any simplification it can get.

zygoloid:
Closing; we don't need to have both this and #1374 open. A pull request is as good as a (non-PR) issue for this.

PR #1374, opened 2017-01-17
[expr.cond] drop redundant subclause (CWG 2316)

cubbimew:
proposed fix for #1373

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -6451,24 +6451,13 @@ directly\iref{dcl.init.ref} to a glvalue.
 subject to the constraint that the reference must bind directly.
 
 \item If \tcode{E2} is a prvalue or if neither of the conversion sequences above can be
-formed and at least one of the operands has (possibly cv-qualified) class type:
-\begin{itemize}
-\item if \tcode{T1} and \tcode{T2} are the same class type
-(ignoring cv-qualification) and
-\tcode{T2} is at least as cv-qualified as \tcode{T1},
-the target type is \tcode{T2},
-
-\item otherwise, if \tcode{T2} is a base class of \tcode{T1},
-the target type is \cvqual{cv1} \tcode{T2}, where \cvqual{cv1}
-denotes the cv-qualifiers of \tcode{T1},
-
-\item otherwise, the target type is the type that \tcode{E2} would have
+formed and at least one of the operands has (possibly cv-qualified) class type, the
+target type is the type that \tcode{E2} would have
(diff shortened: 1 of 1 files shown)


jensmaurer:
These words are quick to anger... For @zygoloid to decide whether this is editorial or needs CWG review.

zygoloid:
I'm confident that the old and new words have the same meaning, but I'd like CWG review to double-check that we don't *intend* for these two cases to be handled differently any more. (As I recall, they did actually do different things before we reformulated this in terms of implicit conversion sequences in CWG1895.)

cubbimew:
resolved a conflict with https://github.com/cplusplus/draft/commit/cf283d86d2fecd1190fa9c437d73da70a35060bd

tkoeppe:
@jensmaurer, @zygoloid: Did this PR ever come up in CWG?

cpplearner:
I suspect that [CWG2316](https://wg21.link/cwg2316) is related, though I can't see the detailed description.

tkoeppe:
@cubbimew Rebase this regardless please if you don't mind, so that CWG can look at current wording.

tkoeppe:
@cubbimew: rebase please?

cubbimew:
rebased

jensmaurer:
Approved by CWG telecon 2020-08-17.

hubert-reinterpretcast:
Not that anyone implemented it that I know of, but this reverts the resolution to CWG 2321 (http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1113r0.html) adopted by the June 2018 Rapperswil meeting.

Compiler Explorer link demonstrating (at this time) non-adoption of the resolution to CWG 2321: https://godbolt.org/z/f8KqaM.

opensdh:
I don't understand, now, how that resolution was necessary anyway: if you fall through to the other bullet in the issue's example, isn't the target type `A`, to which `const B` can be implicitly converted?

jensmaurer:
The change was reverted in 02474121ba0ecde1871a52e409ba5174fb79c89f because it reverts CWG 2321. Such action is not editorial, and furthermore unintended.

hubert-reinterpretcast:
> I don't understand, now, how that resolution was necessary anyway: if you fall through to the other bullet in the issue's example, isn't the target type `A`, to which `const B` can be implicitly converted?

Maybe that's why MSVC accepts but with different cv-qualification for the result than CWG 2321 would give.

PR #1377, opened 2017-01-18
[cpp.concat] In example code comment, remove unnecessary line break and format code as code.

Eelis:
The comment easily fits on one line:

diff

(See [expr.dynamic.cast]/5 for a code comment that already has identical formatting.)

### DIFF
--- a/source/preprocessor.tex
+++ b/source/preprocessor.tex
@@ -866,8 +866,7 @@ operators is unspecified.
 #define mkstr(a) # a
 #define in_between(a) mkstr(a)
 #define join(c, d) in_between(c hash_hash d)
-char p[] = join(x, y);          // equivalent to
-                                // char p[] = "x \#\# y";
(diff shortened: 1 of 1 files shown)


tkoeppe:
Thanks, but I'm already working on making changes like that on a larger scale. I'll get to that file eventually. This is definitely on the radar.

Eelis:
Ok cool, is there a relevant issue already?

tkoeppe:
No, I don't think so. It's not like anything is actually *wrong*...

Eelis:
Well, we very routinely use issues to track cases of inconsistent formatting, right?

tkoeppe:
I'm kind of looking at everything. That's why there's no issue. "Perform continual editorial review", if you will.

jensmaurer:
@tkoeppe: I'm sorry, I don't understand why you closed this. I appreciate you're going through the whole standard fixing formatting, but this particular change seems fine on its own, does not introduce inconsistencies, and does materially improve the formatting for this example.

So, let's apply it.

PR #1380, opened 2017-01-18
Consistently use 'this International Standard'.

Eelis:
For consistency with all other occurrences.

A related question is whether all occurrences of "this Standard" should really say "this International Standard". (IMHO saying the latter over and over sounds pointless and pompous, but if it's to satisfy some legal requirement, perhaps we ought to do it consistently.)

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -1203,7 +1203,7 @@ operations on containers and other sequences in parallel.
 
 \rSec2[algorithms.parallel.defns]{Terms and definitions}
 \pnum
-A \defn{parallel algorithm} is a function template listed in this standard with
+A \defn{parallel algorithm} is a function template listed in this International Standard with
 a template parameter named \tcode{ExecutionPolicy}.
 
 \pnum
--- a/source/atomics.tex
+++ b/source/atomics.tex
@@ -1369,7 +1369,7 @@ The \tcode{atomic_flag} type provides the classic test-and-set functionality. It
 Operations on an object of type \tcode{atomic_flag} shall be lock-free. \begin{note} Hence
 the operations should also be address-free. No other type requires lock-free operations,
 so the \tcode{atomic_flag} type is the minimum hardware-implemented type needed to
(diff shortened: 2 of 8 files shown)


jensmaurer:
Yes, we want to say "this International Standard". Always. (We're not mentioning it too often, so it's ok to be formally correct in all cases.)

tkoeppe:
Either seems fine ("this Standard" or "this International Standard"); you guys figure out which one you want, and leave a message when it's ready.

Eelis:
I'll update the patch to make it long form everywhere :)

jensmaurer:
The
[ISO Directives Part 2](http://www.iec.ch/members_experts/refdocs/iec/isoiecdir-2%7Bed7.0%7Den.pdf), section 3.1.4, say that "International Standard" refers to a standard issued by ISO or IEC, as opposed to any random standard. That, to me, is a clear indication we should not use simply "standard", which has a broader definition.

Eelis:
@jensmaurer Cool, thanks for looking it up! Patch amended.

jensmaurer:
Looks good to me.

Leaving a message for @tkoeppe, as requested.

tkoeppe:
Thanks. @jensmaurer, would you mind adding a note to the wiki about this so we remember in the future?

jensmaurer:
@tkoeppe: Done.

timsong-cpp:
@CaseyCarter yesterday pointed to ISO/IEC directives part 2 §10.6 "References in a document to itself" regarding [a similar issue I reported for the Ranges TS](https://github.com/ericniebler/stl2/pull/304):

> For an individual document, the form "this document" shall be used.

This is a change from the sixth edition and explicitly called out in the foreword of the seventh edition of the directives.

Eelis:
@timsong-cpp Ah damn, I wish you'd been here half an hour earlier :)

tkoeppe:
Well, it's still an improvement over the status quo ante, and presumably we'd never even have heard of this discrepancy if it hadn't been for this PR.

PR #1385, opened 2017-01-18
Remove space between period/comma and footnote reference.

Eelis:
diff

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -2299,8 +2299,7 @@ global function signature declared with
 external linkage in a header is reserved to the
 implementation to designate that function signature with
 \indextext{linkage!external}%
-external linkage.
-\footnote{The list of such reserved function
+external linkage.\footnote{The list of such reserved function
 signatures with external linkage includes
 \indexlibrary{\idxcode{setjmp}}%
 \tcode{setjmp(jmp_buf)},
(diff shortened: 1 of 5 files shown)


source/lib-intro.tex
```diff
@@ -2338,8 +2337,7 @@
 and
 \indextext{\idxcode{extern ""C++""}}%
 \tcode{extern "C++"}
-linkage,
-\footnote{
+linkage,\footnote{
```
tkoeppe: Also unbreak the line after `{`

Eelis: Nice catch, thanks! Patch amended, with 1 new visual difference, namely the removal of the space at the start of this footnote.

source/numerics.tex
```diff
@@ -3869,8 +3869,8 @@
  nondeterministic uniform random bit generator object.
  The semantics and default value of the \tcode{token}
  parameter are
- \impldef{semantics and default value of \tcode{token} parameter to \tcode{random_device} constructor}.
- \footnote{The parameter is intended
+ \impldef{semantics and default value of \tcode{token} parameter to \tcode{random_device} constructor}.%
+\footnote{The parameter is intended
```
tkoeppe: Yarkhs, why is there a weird blank column at the start here... never mind, that can be revisited at a later point.

PR #1386, opened 2017-01-18
Hyphenate 'pointer-to-member' when it is an adjective.

### LINKED ISSUE #1369: "pointer to member" vs "pointer-to-member"
Pick one.

e.g. just in overloading.tex:

````
overloading.tex:a pointer to member, or \tcode{std::nullptr_t}
overloading.tex:pointer-to-member-function type.
overloading.tex:pointer to member function type.
overloading.tex:function name is required to have the form of a pointer to member as
overloading.tex:                                // pointer to member
overloading.tex:For every pointer to member type \tcode{\placeholder{T}} or type \tcode{std::nullptr_t} there
overloading.tex:is an enumeration or pointer to member type,
overloading.tex:is a pointer, pointer-to-member, or scoped enumeration type, there exist candidate operator
````

### PR DESCRIPTION
**Pull request by jensmaurer**
Fixes #1369.

### DIFF
--- a/source/conversions.tex
+++ b/source/conversions.tex
@@ -22,7 +22,7 @@ conversion.
 \item Zero or one conversion from the following set: integral
 promotions, floating-point promotion, integral conversions, floating-point
 conversions, floating-integral conversions, pointer conversions,
-pointer to member conversions, and boolean conversions.
+pointer-to-member conversions, and boolean conversions.
 
 \item Zero or one function pointer conversion.
 
@@ -326,7 +326,7 @@ of \tcode{X} of type \cvqual{cv2} \tcode{T}'' if ``\cvqual{cv2}
 
 \pnum
 \begin{note}
-Function types (including those used in pointer to member function
+Function types (including those used in pointer-to-member-function
 types) are never cv-qualified (\ref{dcl.fct}).
 \end{note}
 \indextext{conversion!qualification|)}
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by tkoeppe** [comment]
Looks good, will leave to @zygoloid to merge.

File: source/conversions.tex [hunk 1]
```diff
@@ -524,21 +524,21 @@
 object. The null pointer value is converted to the null pointer value of
 the destination type.
 
-\rSec1[conv.mem]{Pointer to member conversions}
+\rSec1[conv.mem]{Pointer-to-member conversions}
```
zygoloid: I'm concerned that this change suggests a false parallel: previously, all the "X-to-Y conversion"s converted an X into a Y. A pointer to member conversion definitely does not convert a pointer into a member. I'm also still trying to decide if the rules for adjectives really apply here; "pointer to member type" seems more like "office chair store" (which would not be written as "office-chair store") than like, say, "office-themed store".

We have quite a few other cases (such as "pointer to function types") that follow a similar non-hyphenation pattern. These seem quite natural to me as-is.

jensmaurer: We have 15 appearances where "pointer-to-member" is, in fact, hyphenated, so I presume you would accept a patch that removes the hyphens there?

zygoloid: Let's discuss this at Kona and come up with a rule, then apply it consistently.

tkoeppe: We decided that we can get over this. Fine as written.

tkoeppe:
I stole this commit, rebased it, added the definition of "pointer-to-member type", and committed this as cbd25c4b3e6f83aaf2fa9fe894ff52b1f428a8c5.

PR #1387, opened 2017-01-18
[cpp.pragma.op] Add missing \pnums and remove unwanted paragraph breaks.

Eelis:
[[cpp.pragma.op]](http://eel.is/c++draft/cpp#pragma.op) currently lacks numbered paragraphs altogether.

### DIFF
--- a/source/preprocessor.tex
+++ b/source/preprocessor.tex
@@ -1386,6 +1386,7 @@ underscore.
 \indextext{macro!pragma operator}%
 \indextext{operator!pragma|see{macro, pragma operator}}
 
+\pnum
 A unary operator expression of the form:
 
 \begin{ncbnf}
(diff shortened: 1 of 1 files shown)


source/preprocessor.tex
```diff
@@ -1387,6 +1387,7 @@
 \indextext{macro!pragma operator}%
 \indextext{operator!pragma|see{macro, pragma operator}}
 
+\pnum
```
tkoeppe: Also move the example into the same paragraph, otherwise we don't really have just "one paragraph".

tkoeppe: Or add another `\pnum`.

Eelis: Good idea, done

source/preprocessor.tex
```diff
@@ -1402,6 +1403,7 @@
 \grammarterm{pp-tokens} in a pragma directive. The original four preprocessing
 tokens in the unary operator expression are removed.
 
+\pnum
 \begin{example}
```
tkoeppe: Delete this pointless newline while you're at it. We got over whatever made us have those back in the day.

Eelis: Done

source/preprocessor.tex
tkoeppe: And also everywhere else below please. Sorry, I hadn't spotted that right away.

Eelis: Makes sense, done. No additional visual changes.

PR #1393, opened 2017-01-20
Comment out ellipses when they mean "more code here" rather than an actual ellipsis token.

### LINKED ISSUE #1390: Comment out ellipses when they mean "more code here" rather than an actual ellipsis token.
Most places already do this, but a few don't:

- [[iostream.forward.overview]/9](http://eel.is/c++draft/iostream.forward.overview#9)
- [[dcl.decl]/3](http://eel.is/c++draft/dcl.decl#3)
- [[expr.prim.lambda]/7](http://eel.is/c++draft/expr.prim.lambda#7)

Interestingly, there are also a couple uses of mega-ellipses consisting of five dots (`.....`):

- [[enumerated.types]/2](http://eel.is/c++draft/enumerated.types#2)
- [[bitmask.types]/2](http://eel.is/c++draft/bitmask.types#2)

But maybe these are fine, because at least `.....` is not the C++ ellipsis token.

Eelis:
Fixes #1390

### DIFF
--- a/source/declarators.tex
+++ b/source/declarators.tex
@@ -74,12 +74,12 @@ One exception is when a name introduced by one of the
 \grammarterm{decl-specifiers} are used in a subsequent declaration,
 they do not have the same meaning, as in
 \begin{codeblock}
-struct S { ... };
+struct S { /* ... */ };
 S S, T;                 // declare two instances of \tcode{struct S}
 \end{codeblock}
 which is not equivalent to
 \begin{codeblock}
-struct S { ... };
+struct S { /* ... */ };
 S S;
 S T;                    // error
 \end{codeblock}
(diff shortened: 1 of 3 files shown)


jwakely:
Because two of those cases become well-formed code with this change, so it could be argued that the ellipsis implies there's nothing but a comment there :-) What we have today isn't valid C++ and so is obviously meant to indicate something missing from the example. I'm not objecting to the change, but I don't think it's strictly necessary either.

Eelis:
I have no preference beyond consistency :)

tkoeppe:
I agree, the status quo doesn't seem to be bad, broken or confusing.

Regardless, please provide a screenshot.

tkoeppe:
Seems like there's no consensus for change nor anyone violently fighting for this change. I'll close this; please reopen if you feel strongly.

PR #1395, opened 2017-01-21
[namespace.memdef]/3

jabelloc:
The word **first** is confusing and unnecessary in [namespace.memdef]/3. It gives the impression that if the name in the `friend` declaration is not the first in its TU, the friend will not need to be a member of the innermost enclosing namespace, which is clearly incorrect as the snippet below shows:

`void f(); namespace A { struct S { friend void f(); }; }`

The name `f` is **not** first declared in the `friend` declaration. Nevertheless, the snippet will compile if the function `f` is declared in namespace `A`, either before or after the friend declaration.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -1203,7 +1203,7 @@ operations on containers and other sequences in parallel.
 
 \rSec2[algorithms.parallel.defns]{Terms and definitions}
 \pnum
-A \defn{parallel algorithm} is a function template listed in this standard with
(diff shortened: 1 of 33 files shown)


jensmaurer:
This is core issue 138 "Friend declaration name lookup". As such, and given the repeated discussions around it, this does not seem to be editorial at all.

jabelloc:
@jensmaurer 

See also [this discussion](http://stackoverflow.com/questions/31348475/the-member-function-outerf-is-not-a-friend-of-class-outerinner-why) in SO.

PR #1400, opened 2017-01-22
[tuple.creation] Fix formatting errors and capitalization in description of tuple_cat.

Eelis:
diff


tkoeppe:
I think this overlaps with my #1251 too much. Please reopen after that one has been resolved.

Issue #1403, opened 2017-01-24
Term "functor" is used but never defined


grisumbras:
Section 20.14 mentions the term "functor" several times but it is never defined. It should either be changed to "function object" as in all other places, or defined (probably as an alternative name to function objects, codifying a popular convention).

jwakely:
I only see two mentions of it, both in non-normative text (a subclause title, and a reference to that subclause in a comment). It doesn't need to be defined to be referred to in that way.

jwakely:
In fact if we _did_ define it, it would be unused in any normative text, and so the definition should be removed again.

grisumbras:
But wouldn't it be confusing for the reader to see this term there? I can send a PR to change those lines to refer to "function objects".

tkoeppe:
Closing this, since we are going to remove `default_order`.

PR #1405, opened 2017-01-25
[function.objects] removes uses of term functor

grisumbras:
Fixes issue #1403 by removing all mentions of the term "functor". While
there is a convention to use "functor" to mean "function object" it
seems to be fading away. Also, the word "functor" has very different
meaning in Category Theory and (as a consequence of that) in many
Functional Programming languages. As a result of that use of the word
"functor" in the Standard could be confusing to the reader.


jwakely:
`default_order` is almost certain to be removed from the draft in the near future anyway, so this isn't going to matter soon.

PR #1406, opened 2017-01-26
s/0-based/zero-based/

karies:
Conforming with other occurrences in the standard.


tkoeppe:
Thank you. Please provide commit messages in our usual style in the future. In particular, a) start with the relevant section reference, and b) use English, not code.

PR #1408, opened 2017-01-27
[dcl.attr.depend] Fix spacing in /**/ comments in example.

Eelis:
diff

(See [special.tex#1837](https://github.com/cplusplus/draft/blob/master/source/special.tex#L1837) for an example where the same trick is used for the same reason.)

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -3888,7 +3888,7 @@ program, but may result in generation of more efficient code. \end{note}
 \pnum
 \begin{example}
 \begin{codeblock}
-/* Translation unit A. */
+/* @\textrm{\textit{Translation unit A.}}@ */
 
 struct foo { int* a; int* b; };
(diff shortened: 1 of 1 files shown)


zygoloid:
Why do we get a wide space here and not in other comments? Is this a bug in the listings package?

tkoeppe:
It's the usual end-of-sentence space in text mode I think.

Eelis:
I think it is an effect of the "columns=flexible" listings option.
See 2.10 in http://mirrors.ctan.org/macros/latex/contrib/listings/listings.pdf:

> The normal flexible format will insert make-up space to fix the alignment at spaces,
> before and after identifiers, and before and after sequences of other characters

I'm experimenting with using columns=fullflexible. Preliminary results look good.

Eelis:
Closing, will submit a patch based on `columns=fullflexible` imminently.

PR #1411, opened 2017-01-30
Use `value_type` instead of `charT` in the `string_view` declarations.

mclow:


source/strings.tex
```diff
@@ -4760,8 +4760,8 @@
   constexpr basic_string_view() noexcept;
   constexpr basic_string_view(const basic_string_view&) noexcept = default;
   constexpr basic_string_view& operator=(const basic_string_view&) noexcept = default;
-  constexpr basic_string_view(const charT* str);
-  constexpr basic_string_view(const charT* str, size_type len);
+  constexpr basic_string_view(const value_type* str);
+  constexpr basic_string_view(const value_type* str, size_type len);
```
jwakely: I think this change means a deduction guide would be required to deduce the template args when using these constructors.

mclow: Ok, then.   Let's not do that. :-).  What I really want is the typedefs at the top.  If we can agree that those are OK (and editorial), then I'll make a new PR with just those.

jwakely: As long as there are no constructors taking the `pointer` or `reference` types I think that is editorial.

jensmaurer:
Is there any particular reason to prefer one or the other style? Or is it just that template parameters are uglified by the implementation, so not having them appear frequently is a good idea?

mclow:
It's both an uglified thing, and a consistency thing.  `basic_string` has a reference type of `value_type &`, and `basic_string_view` has a reference type of `charT&` (ok, const).  Unless there's a good reason for them to be different, I'd rather they be the same.

mclow:
I've created https://github.com/cplusplus/draft/pull/1416 with just the typedef changes.  Closing this one.

PR #1413, opened 2017-01-30
[valarray.members] Style unary minus sign as math symbol, not as hyphen.

Eelis:
By not abusing \emph for math. Diff:

diff

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -7490,7 +7490,7 @@ This function returns an object of class
 \tcode{valarray<T>}
 of length
 \tcode{size()}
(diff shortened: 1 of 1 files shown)


tkoeppe:
What's a "unary minus sign"? Does typography have arity now? :-)

godbyk:
@tkoeppe It does when it comes to spacing. Compare $-n$ with $m-n$. :-)

tkoeppe:
@godbyk: Hm, colour me crazy, but that's a unary *space*, not a unary *minus*, right? It's the same glyph!

godbyk:
@tkoeppe The minus sign glyphs are the same, that's correct. The only difference is the spacing. Internally to TeX, the glyphs are categorized and the spacing is inserted based on the types of adjacent glyphs.

PR #1416, opened 2017-01-31
Change the `reference`, `pointer` etc typedefs in `string_view` to us…

mclow:
…e `value_type` rather than `charT`

This is a cut down version of https://github.com/cplusplus/draft/pull/1411, to ensure that it is editorial

### DIFF
--- a/source/strings.tex
+++ b/source/strings.tex
@@ -4744,10 +4744,10 @@ public:
   // types
   using traits_type            = traits;
   using value_type             = charT;
-  using pointer                = charT*;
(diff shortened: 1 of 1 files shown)


jensmaurer:
Looks good to me, although I don't really understand the point of this change.  There's charT all over the place here; why fixing the typedefs at the start is important, escapes me.

zygoloid:
I agree that this change by itself seems minor, but it does make the `basic_string_view` specification more similar to the `basic_string` specification. And I think the LWG chair gets special dispensation to ask for such things :)

PR #1426, opened 2017-02-04
Use columns=fullflexible listings option.

Eelis:
1. Removes the need to use `\textit{\textrm{...}}`to prevent the `columns=flexible` listings option from inserting unwanted spaces.
2. Improves spacing in a few places where that workaround was not yet used.

[diffs.pdf](http://eel.is/fullflexiblediffs.pdf)


zygoloid:
Thanks for the diffs, that makes this very easy to approve :)

PR #1434, opened 2017-02-05
[expr.mptr.oper] Use defined term 'null member pointer value'.

Eelis:
See the definition at [conversions.tex#L536](https://github.com/cplusplus/draft/blob/master/source/conversions.tex#L536)


jensmaurer:
Looks good to me.

PR #1438, opened 2017-02-07
[dcl.stc] Remove stray whitespace in plural \grammarterm

jensmaurer:


jensmaurer:
@tkoeppe: Can you apply this? It seems typo-level.

PR #1439, opened 2017-02-07
Remove \xspace

### LINKED ISSUE #1433: Review use of \xspace in macro definitions
See http://tex.stackexchange.com/questions/86565/drawbacks-of-xspace with the first answer from the package author of \xspace. Quote:

> So, if you find it useful, fine, it's there. But personally I wouldn't recommend it.

\xspace in \grammarterm caused spacing issues while addressing #1268.

jensmaurer:
Fixes #1433.

This removes lots of kerning ugliness when (e.g.) a \grammarterm is followed by a punctuator. It's also a surprise to notice all the places where we actually had superfluous whitespace in the document.

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -4,7 +4,7 @@
 \rSec1[input.output.general]{General}
 
 \pnum
-This Clause describes components that \Cpp programs may use to perform
+This Clause describes components that \Cpp{} programs may use to perform
 input/output operations.
 
 \pnum
@@ -1347,7 +1347,7 @@ locale getloc() const;
 \begin{itemdescr}
 \pnum
 \returns
-If no locale has been imbued, a copy of the global \Cpp locale,
+If no locale has been imbued, a copy of the global \Cpp{} locale,
 \tcode{locale()},
 in effect at the time of construction.
 Otherwise, returns the imbued locale, to be used to
@@ -2721,7 +2721,7 @@ ios_base& hexfloat(ios_base& str);
 \pnum
 \begin{note} The more obvious use of
 \tcode{ios_base::hex} to specify hexadecimal floating-point format would
-change the meaning of existing well-defined programs. \CppIII
+change the meaning of existing well-defined programs. \CppIII{}
 gives no meaning to the combination of \tcode{fixed} and
 \tcode{scientific}.\end{note}
 
@@ -10835,7 +10835,7 @@ For Windows-based operating systems, the
 native narrow encoding is determined by calling a Windows API function.
 \end{note}
 \begin{note}
-This results in behavior identical to other C and \Cpp
+This results in behavior identical to other C and \Cpp{}
 standard library functions that perform file operations using narrow character
 strings to identify paths. Changing this behavior would be surprising and error
 prone.
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -10109,8 +10109,8 @@ entry for the file under discussion.
 A sequence of elements that identify
 the location of a file within a filesystem.
 The elements are the
-\grammarterm{root-name}\opt{},
-\grammarterm{root-directory}\opt{},
+\opt{\grammarterm{root-name}},
+\opt{\grammarterm{root-directory}},
 and an optional sequence of filenames.
 The maximum number of elements in the sequence is operating system dependent.
 
@@ -10647,7 +10647,7 @@ any operating system dependent \grammarterm{root-name}{s}}}
 
 \begin{ncbnf}
 \nontermdef{pathname}\br
-    root-name\opt{} root-directory\opt{} relative-path
+    \opt{root-name} \opt{root-directory} relative-path
 \end{ncbnf}
 
 \begin{ncbnf}
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -8649,12 +8649,12 @@ initializes the
 \tcode{filebuf}
 as required.
 It then opens a file, if possible, whose name is the
-\ntbs \tcode{s}
+\ntbs{} \tcode{s}
 (as if by calling
 \tcode{fopen(s, modstr)}).
 \indextext{NTBS}%
 \indexlibrary{\idxcode{fopen}}%
-The \ntbs \tcode{modstr} is determined from
+The \ntbs{} \tcode{modstr} is determined from
 \tcode{mode \& \~{}ios_base::ate}
 as indicated in Table~\ref{tab:iostreams.file.open.modes}.
 If \tcode{mode} is not some combination of flags shown in the table then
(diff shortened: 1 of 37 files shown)


source/macros.tex
```diff
@@ -172,11 +172,11 @@
 
 %%--------------------------------------------------
 %% Macros for funky text
-\newcommand{\Cpp}{\texorpdfstring{C\kern-0.05em\protect\raisebox{.35ex}{\textsmaller[2]{+\kern-0.05em+}}}{C++}\xspace}
-\newcommand{\CppIII}{\Cpp 2003\xspace}
-\newcommand{\CppXI}{\Cpp 2011\xspace}
-\newcommand{\CppXIV}{\Cpp 2014\xspace}
-\newcommand{\opt}{{\ensuremath{_\mathit{opt}}}\xspace}
+\newcommand{\Cpp}{\texorpdfstring{C\kern-0.05em\protect\raisebox{.35ex}{\textsmaller[2]{+\kern-0.05em+}}}{C++}}
+\newcommand{\CppIII}{\Cpp{} 2003}
```
godbyk: You might also consider `\Cpp~2003` if you want to keep 'C++ 2013' from splitting across two lines.

jensmaurer:
I have run "diffpdf" in "Words" mode (to show any egregious accidents), and other than pagination changes, only substantial whitespace improvements showed up. We sometimes had inter-word space between an italics term and a subsequent comma, for example.

source/macros.tex
jensmaurer: Good idea. Applied.

jensmaurer:
Rebased.

Eelis:
This is really great! (Needs rebase though)

jensmaurer:
Rebased.

tkoeppe:
Note that if there are genuine errors in the current text that you discovered as part of the change, I am sympathetic towards fixing those separately in advance of this change. No need to ship a standard with known warts.

jensmaurer:
@tkoeppe : Do you consider my pictures above "errors"?

tkoeppe:
@jensmaurer: Yes, definitely, things like the space in "*linkage* ," is definitely something I'd consider borderline broken.

jensmaurer:
Rebased.

source/iostreams.tex
```diff
@@ -10036,8 +10036,8 @@
 A sequence of elements that identify
 the location of a file within a filesystem.
 The elements are the
-\grammarterm{root-name\opt},
-\grammarterm{root-directory\opt},
+\grammarterm{\opt{root-name}},
+\grammarterm{\opt{root-directory}},
```
Eelis: Shouldn't the \opt go outside the \grammarterm?

tkoeppe:
In the meantime, I'm sending out #1544 to just address the appalling spacing issues in basic.link.

source/iostreams.tex
burblebee: Yes, I believe it should.  I see other places where grammarterm has problems as well, as can be seen in this grep:
```
$grep "grammarterm.*\<opt\>" *.tex
declarations.tex:\tcode{typename}\opt{} \grammarterm{nested-name-specifier}\opt{} \grammarterm{template-name}
declarations.tex:``\tcode{enum} \grammarterm{nested-name-specifier\opt} \grammarterm{identifier}''
declarations.tex:\grammarterm{attribute-specifier-seq\opt} \grammarterm{identifier}~(\ref{dcl.type.elab}), or
declarators.tex:is of the form \grammarterm{attribute-specifier-seq\opt}
declarators.tex:\grammarterm{attribute-specifier-seq\opt}
declarators.tex:\grammarterm{cv-qualifier-seq\opt} \grammarterm{ref-qualifier\opt}
declarators.tex:\grammarterm{cv-qualifier-seq}\opt \grammarterm{ref-qualifier}\opt
expressions.tex:parameter-type-list \cvqual{cv} \grammarterm{ref-qualifier\opt} returning \tcode{T}'', then
iostreams.tex:\grammarterm{root-name\opt},
iostreams.tex:\grammarterm{root-directory\opt}, 
```

Another editorial issue?

fwiw, I think the above should instead be coded as:
```
\grammarterm{root-name}\opt{},
\grammarterm{root-directory}\opt{},
```

jensmaurer:
Rebased (again).

source/iostreams.tex
jensmaurer: @burblebee: It's an express change in this patch series to switch from `blah\opt{}` to `\opt{blah}`, because it makes it less likely to get a spacing accident by forgetting the empty `{}`.

Eelis:
Another bad xspace space that might be worth fixing asap is the one resulting from the use of \term in

  https://github.com/cplusplus/draft/blob/master/source/expressions.tex#L4831

In C++98, the space was not there.

tkoeppe:
@Eelis: Fixed by 8d3cba2311bf49f744311fd695da4b23b92a86c6, thanks.

Eelis:
Perhaps now would be a good time to rebase/apply this?

jensmaurer:
Rebased.

tkoeppe:
Rebase, Dr Freeman.

Eelis:
It was mentioned on the mailing list that it would be desirable to have the new `\opt` reject empty arguments or arguments containing spaces. I'm a LaTeX noob, but the following seems to work:

```
\documentclass{article}
\usepackage{xstring}
\usepackage{ifthen}
\begin{document}
\newcommand{\opt}[1]{\IfSubStr{#1}{ }
  {\PackageError{main}{argument must not contain spaces}{}}
  {\ifthenelse{\equal{#1}{}}
    {\PackageError{main}{argument must not be empty}{}}
    {#1\ensuremath{_\mathit{opt}}}}}

hello \opt{world}
% \opt{hello world} % emits the first error
% hello world\opt{} % emits the second error
\end{document}
```

tkoeppe:
Nice. Can you also make it discover linebreaks?

Eelis:
The `IfSubStr` already recognizes newlines as spaces, it seems. :)

jensmaurer:
Do we want to apply this patch series as large as it is, or should I split it into separate patches along the lines of:
- change \Cpp and friends to \Cpp{} and remove \xspace use
- change NTBS macros to \blah{} and remove \xspace use
- morph blah\opt into \opt{blah} (with Eelis' macro above)

These could be reviewed separately, making smaller chunks of changes. Opinions?

tkoeppe:
Is this PR redundant in light of #1761?

jensmaurer:
@tkoeppe: Yes and no. #1761 removes \xspace everywhere except for \opt, which needs special treatment (to be done later). I'll close this pull request.

PR #1442, opened 2017-02-08
[container.requirements.general] Refer to table series a bit more smoothly

Eelis:
diff

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -69,9 +69,7 @@ aligned buffers and call \tcode{construct} to place the element into the buffer.
 \end{note}
 
 \pnum
-In Tables~\ref{tab:containers.container.requirements},
-\ref{tab:containers.reversible.requirements}, and
-\ref{tab:containers.optional.operations}
+In Tables~\ref{tab:containers.container.requirements}--\ref{tab:containers.optional.operations},
 \tcode{X} denotes a container class containing objects of type
 \tcode{T}, \tcode{a} and \tcode{b}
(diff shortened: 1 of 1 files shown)


jwakely:
This would be fine today, but if we added a new table before 85 then it would become"83-86", but we might not want it included in the range. What we have today would still be correct, referring to the individual tables.

I don't know if that concern is important.

Eelis:
Yeah, I don't know either. I started with just adding a comma after "and 85", but then I got inspired by 17.5.3.1p1, which has "Tables 20-27". :)

godbyk:
The [cleveref](http://ctan.org/pkg/cleveref) package handles these situations nicely:

```latex
\usepackage{cleveref}

In \Cref{tab:containers.container.requirements,tab:containers.reversible.requirements,tab:containers.optional.operations}
\tcode{X} denotes a container class containing objects of type...
```

is typeset as:

> In Tables 83–85 `X` denotes a container class containing objects of type...

`cleveref` sorts and compresses the references by default. It's smart enough to only compress it into a range (e.g., 83–85) when it is contiguous.

jwakely:
Well that's rather clever :)

Eelis:
Sounds nifty. But would that be an acceptable dependency?

tkoeppe:
I'm not wild about this change, I think the status quo is fine. @jwakely, @zygoloid, do you feel strongly either way?

jwakely:
Nope, I'm neutral.

PR #1445, opened 2017-02-08
[enum.copy_options] Rephrase restriction on copy_options values.

### LINKED ISSUE #1422: [enum.copy_options] Poor wording for requirement on elements set in copy_options
[enum.copy_options] says:

> Calling a library function with more than a single constant for an option group results in undefined behavior. 

I don't like how this is phrased. Firstly, which library? Does it apply to _any_ function in the standard library, e.g. `std::move`?

Secondly, I'd prefer if it used the term of art "is set" for bitmask elements set in a bitmask object.

This doesn't exactly roll off the tongue, but I think it's more precise:

"A value of type `copy_options` that is passed to a function described in subclause 27.10 shall have at most one element set from each option group."

jensmaurer:
Fixes #1422.

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -12416,9 +12416,7 @@ The \tcode{enum class} type \tcode{copy_options}
 is a bitmask type~(\ref{bitmask.types}) that specifies bitmask constants used to control the semantics of
 copy operations. The constants are specified in option groups with the meanings listed in Table~\ref{tab:fs.enum.copy_options}.
 Constant \tcode{none} is shown in each option group for purposes of exposition;
-implementations shall provide only a single definition. Calling a
-library function with more than a single constant for an option
-group results in undefined behavior.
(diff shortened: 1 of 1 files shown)


jensmaurer:
As discovered in #1422, it turns out that the requirements are already spelled out with the two function families that use copy_options. My new patch simply removes the general requirement and slightly adjusts the requires clauses with the individual functions.

jensmaurer:
@jwakely: What do you think?

timsong-cpp:
Hmm, one is "from each `copy_options` option group", the other is simply "from each option group". Might make them consistent as a drive-by fix.

jensmaurer:
Rebased.

jwakely:
Looks good to me.

jensmaurer:
Now also addressing @timsong-cpp's comment.

PR #1449, opened 2017-02-08
[istreambuf.iterator] Join subsections for operations descriptions.

### LINKED ISSUE #1429: [ostreambuf.iter.ops] not divided into subsections as [istreambuf.iter.ops]
there are several [istreambuf.iterator::op...] sections under [istreambuf.iterator],
but there is only a combined [ostreambuf.iter.ops] section under [ostreambuf.iterator].
This is inconsistent.

Consider split it into subsections.

Also: consider change `::` into `.` and `*` into `star` in section labels

jensmaurer:
There was one subsection for every operator, yet everything
fits on half a page.

Fixes #1429.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -3028,7 +3028,7 @@ namespace std {
 }
 \end{codeblock}
 
-\rSec3[istreambuf.iterator::proxy]{Class template \tcode{istreambuf_iterator::proxy}}
+\rSec3[istreambuf.iterator.proxy]{Class template \tcode{istreambuf_iterator::proxy}}
 
 \indexlibrary{\idxcode{proxy}!\idxcode{istreambuf_iterator}}%
 \begin{codeblock}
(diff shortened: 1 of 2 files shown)


jensmaurer:
I'll inform Daniel Krügler about the changes for the stable names.

jensmaurer:
This is a larger issue in all of [predef.iterators]; see #1429.

tkoeppe:
@zygoloid, @jensmaurer: Is this for C++17? It looks like a nice and simple change; can we do this in isolation or would you prefer to only do it together with #1429 in one go?

jensmaurer:
While this is a larger issue, other situations are less clear-cut than this one. So, I'd be happy to merge this for C++17. Except that we need to augment the new "stable names migration" annex.

jensmaurer:
Updated for "stable names migration" annex. Note that "op!=" can't be mentioned, because the exclamation mark badly interacts with LaTeX "glossary" operations. Maybe there's some escape-foo to make it work? @godbyk?

zygoloid:
A glossary is a weird kind of index, so you probably need to escape the ! with ". I'll give that a go.

tkoeppe:
@zygoloid: That'd be great, this has been bugging me for a very long time. Please check the `.ilg` logs to be sure that you've resolved all cases.

zygoloid:
That works, looking at other makeindex errors now.

PR #1455, opened 2017-02-09
[alg.min.max] Make lfloor/rfloor delimiters stretchable.

Eelis:
Effect:
diff
Unfortunately, while this stretches the \lfloor/\rfloor nicely, it also seems to introduce some extra space before the comma. Anybody know a trick to avoid that?

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -5190,7 +5190,7 @@ in \range{first}{last} such that no iterator in the range refers to a larger ele
 \pnum
 \complexity
 At most
-$\max(\lfloor{\frac{3}{2}} (N-1)\rfloor, 0)$
+$\max(\bigl\lfloor{\frac{3}{2}} (N-1)\bigr\rfloor, 0)$
 applications of the corresponding predicate, where $N$ is \tcode{last - first}.
 \end{itemdescr}
 


Eelis:
Ah, found the trick: \mathclose. No more extra space now.

diff2

jensmaurer:

source/algorithms.tex
```diff
@@ -5190,7 +5190,7 @@
 \pnum
 \complexity
 At most
-$\max(\lfloor{\frac{3}{2}} (N-1)\rfloor, 0)$
+$\max(\left\lfloor{\frac{3}{2}} (N-1)\right\rfloor\mathclose{}, 0)$
```
tkoeppe: Why do we need `\mathclose{}`?

Eelis: The best explanation I could find is at https://tex.stackexchange.com/a/2610

tkoeppe: What happens if you just use `\bigl`, `\bigr`? The advantage of those is that they are guaranteed to fit into the line, whereas the stretchable versions may end up taller than a line.

Eelis: Ah nice, using \bigl and \bigr produces the same result without needing the \mathclose, thanks!

Patch amended.

tkoeppe: If you like you can also make the outer round parens big to match. See which one looks better.

Eelis: Hmm, personally, I like it with the outer parens as-is. Less intimidating.

PR #1463, opened 2017-02-14
Fix index entries for 'nested/local class'.

Eelis:
- Changes the peculiar redirection entries "class nested" and "class local" to the more conventional "class, nested" and "class, local".
- Adds main index entries for "local class" and "nested class" (fixing some abuse of \grammarterm in the process)

[diffs.pdf](http://eel.is/classindex.pdf)

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -1387,12 +1387,11 @@ void f() {
 
 \rSec2[class.nest]{Nested class declarations}%
 \indextext{definition!nested class}%
-\indextext{class local|see{local class}}%
-\indextext{class nested|see{nested class}}
+\indextext{class!nested|see{nested class}}
 
 \pnum
 A class can be declared within another class. A class declared within
-another is called a \grammarterm{nested} class. The name of a nested class
+another is called a \defnx{nested}{nested class} class. The name of a nested class
 is local to its enclosing class.
 \indextext{nested class!scope of}%
 The nested class is in the scope of its enclosing class.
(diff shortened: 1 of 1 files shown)


source/classes.tex
```diff
@@ -1387,12 +1387,11 @@
...
 \indextext{definition!nested class}%
-\indextext{class local|see{local class}}%
-\indextext{class nested|see{nested class}}
+\indextext{class!nested|see{nested class}}
 
 \pnum
 A class can be declared within another class. A class declared within
-another is called a \grammarterm{nested} class. The name of a nested class
+another is called a \defnx{nested}{nested class} class. The name of a nested class
 is local to its enclosing class.
```
jensmaurer: The defined phrase is actually "nested class", so I'd like to see both words in italics. (Same for "local class".)

source/classes.tex
```diff
@@ -1387,12 +1387,11 @@
 
 \rSec2[class.nest]{Nested class declarations}%
 \indextext{definition!nested class}%
-\indextext{class local|see{local class}}%
-\indextext{class nested|see{nested class}}
+\indextext{class!nested|see{nested class}}
```
jensmaurer: In general, it seems to me we gravitate towards "class!nested" as the main index entry (where all the page numbers go) and have "nested class" as the "see ..." entry.

Eelis: Both "local class" and "nested class" have a bunch of subentries:

subentries

Should these be moved under "class, local" and "class, nested", respectively, as well, or stay where they are?

PR #1466, opened 2017-02-14
Fix destructor declaration indentation

CaseyCarter:
..and apparently some trailing whitespace cleanup because my text editor hates it as much as I do.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -5639,7 +5639,7 @@ namespace std {
     vector(const vector&, const Allocator&);
     vector(vector&&, const Allocator&);
     vector(initializer_list<bool>, const Allocator& = Allocator()));
-   ~vector();
+    ~vector();
     vector<bool, Allocator>& operator=(const vector<bool, Allocator>& x);
     vector<bool, Allocator>& operator=(vector<bool, Allocator>&& x);
     vector& operator=(initializer_list<bool>);
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -1110,7 +1110,7 @@ namespace std {
   class ios_base::Init {
   public:
     Init();
-   ~Init();
+    ~Init();
   private:
     static int init_cnt; // \expos
   };
@@ -10420,7 +10420,7 @@ namespace std::filesystem {
       path(const Source& source, const locale& loc);
     template <class InputIterator>
       path(InputIterator first, InputIterator last, const locale& loc);
-   ~path();
+    ~path();
 
     // \ref{path.assign}, assignments
     path& operator=(const path& p);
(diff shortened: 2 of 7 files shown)


source/iostreams.tex
```diff
@@ -10420,7 +10420,7 @@
       path(const Source& source, const locale& loc);
     template <class InputIterator>
       path(InputIterator first, InputIterator last, const locale& loc);
-   ~path();
+    ~path();
```
CaseyCarter: Actual change is here.

jwakely:
I think this is intentional, so the destructor lines up with the constructor, and we do it in a few places:

containers.tex:   ~vector();                                                                                                                                                                                      
iostreams.tex:   ~Init();                                                                                                                                                                                         
iostreams.tex:   ~path();                                                                                                                                                                                         
iostreams.tex:   ~file_status();                                                                                                                                                                                  
iostreams.tex:   ~directory_entry();                                                                                                                                                                              
iostreams.tex:   ~directory_iterator();                                                                                                                                                                           
iostreams.tex:   ~recursive_directory_iterator();                                                                                                                                                                 
iterators.tex:   ~istream_iterator() = default;                                                                                                                                                                   
iterators.tex:   ~ostream_iterator();                                                                                                                                                                             
locales.tex:   ~numpunct_byname();                                                                                                                                                                                
numerics.tex:   ~mask_array();                                                                                                                                                                                    
numerics.tex:   ~indirect_array();                                                                                                                                                                                
strings.tex:   ~basic_string();                                                                                                                                                                                   
utilities.tex:   ~allocator();      

That doesn't mean we have to keep it though.

CaseyCarter:
>  we do it in a few places

Blech, I only looked at a couple which of course didn't use this device. It would be nice to pick a rule and be consistent throughout the standard. (See `template<` vs. `template <`)

I'll repush with them all fixed. (Yes, I said "fixed." I'm totally unbiased.)

CaseyCarter:
> containers.tex: ~vector();

The fact that `vector<bool>` uses this device clearly indicates that it's very, very wrong.

CaseyCarter:
There's a consistent fix throughout.

source/containers.tex
```diff
@@ -4155,7 +4155,7 @@
 \begin{itemdescr}
 \pnum
 \effects If \tcode{sz < distance(begin(), end())}, erases the last \tcode{distance(begin(),
-end()) - sz} elements from the list. Otherwise, inserts \tcode{sz - distance(begin(), end())}
```
tkoeppe: Could you please submit the trailing whitespace deletions as a separate pull request?

jensmaurer: @tkoeppe: See #1468.

tkoeppe: @CaseyCarter: Please rebase.

source/iostreams.tex
CaseyCarter: ...and everywhere else.

source/containers.tex
CaseyCarter: Done.

PR #1468, opened 2017-02-15
Remove trailing whitespace in LaTeX sources.

jensmaurer:

PR #1477, opened 2017-02-18
Add Travis CI check for whitespace at the ends of lines.

godbyk:
Per @tkoeppe's [suggestion](https://github.com/cplusplus/draft/pull/1468#issuecomment-280022177).

This adds a new build type to the Travis CI build which only checks for whitespace at the ends of lines. If it finds any, it will fail the build.

Having this test as a separate 'build type' makes it easy to determine whether the whitespace test failed or whether the PDF compilation failed.

Note: If we plan on adding more tests in the future (e.g., checks for other nonconformant usage)—which I'm in favor of—then I'd recommend we migrate the code from `.travis-ci.yml` to a separate shell script where we can write more succinct code. (Travis CI wraps the `.travis-ci.yml` code in its own function and hijacks the exit code `$?` variable, so we have to capture it to test it later. Within our own shell script, this isn't a problem.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,12 +38,8 @@ script:
   - popd
   # Fail if there is whitespace at the ends of any lines
   - if [ "$BUILD_TYPE" = "check-whitespace" ]; then
-      export TEST_STATUS=0;
-      grep '\s$' source/*.tex && export TEST_STATUS=1;
-    fi;
-    if [ "0$TEST_STATUS" != "00" ]; then
-      false;
-    fi;
+      ! grep '\s$' source/*.tex;
+    fi
   # Check to see if generated files are out-dated
   - pushd source
   - for FIGURE in *.dot; do
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ env:
   - BUILD_TYPE=make      # build using Makefile
   - BUILD_TYPE=manual    # build manually
   - BUILD_TYPE=complete  # build manually and regenerate figures, grammer, and cross-references
+  - BUILD_TYPE=check-whitespace  # check for whitespace at the ends of lines
 
 script:
   # Build std.pdf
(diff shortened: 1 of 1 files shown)


jensmaurer:
@godbyk: And right with pull request, it'S the first time I'm seeing a Travis CI failure. How's that?

tkoeppe:
@godbyk: wonderful, thanks!

godbyk:
@jensmaurer I just pushed a fix. Waiting to see if Travis CI likes it.  (During my later testing, I was canceling the non-test-whitespace builds as I wasn't testing the PDF builds.)

.travis.yml
```diff
@@ -35,6 +36,14 @@ script:
       pdflatex std;
     fi
   - popd
+  # Fail if there is whitespace at the ends of any lines
+  - export TEST_STATUS=0;
+    if [ "$BUILD_TYPE" = "check-whitespace" ]; then
+      grep '\s$' source/*.tex && export TEST_STATUS=1;
```
tkoeppe: Hm, I don't know Travis, but if this is just shell code, couldn't you just say `grep '\s$' source/*.tex` here and let `grep`'s return status do the job? It seems that just `false` has the desired effect of failing the test, so `grep` should be able to d the same?

godbyk: @tkoeppe I need to invert grep's exit status.

tkoeppe: Oh, OK, what about `! grep ...` then?

godbyk: @tkoeppe When I tried that previously it didn't work, but it appears to be working now. (I guess it was a combination of bugs previously—testing the `.travis-ci.yml` file is a bit of trial-and-error.) Thanks for prompting me to give it another shot; it greatly simplifies the code!

tkoeppe: Less is more :-)

tkoeppe:
Thanks a lot! I've always wanted this to be cleaned up, and now we can prevent people from ever breaking it again!

godbyk:
@tkoeppe No problem!

PR #1479, opened 2017-02-19
[meta.unary.prop] Omit unhelpful second argument in static_asserts.

Eelis:
Probably leftover from the days the argument was mandatory.


source/utilities.tex
```diff
@@ -16166,10 +16166,10 @@
 union U2 final { };
 
 // the following assertions hold:
-static_assert(!is_final_v<int>, "Error!");
-static_assert( is_final_v<P>,  "Error!");
-static_assert(!is_final_v<U1>, "Error!");
-static_assert( is_final_v<U2>, "Error!");
+static_assert(!is_final_v<int>);
+static_assert( is_final_v<P>  );
```
tkoeppe: Let's not have the cutepsaces. Just format it normally.

Eelis: Alrighty, done

PR #1482, opened 2017-02-21
[localization, diff] Omit superfluous 'return 0;' in main.

Eelis:
As far as I can tell, these are the only 4 occurrences (out of 33 `main` functions in the document).

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -1114,7 +1114,6 @@ int main() {
   delete i;                     // single-object delete
   int* a = new int[3];
   delete [] a;                  // array delete
-  return 0;
 }
 \end{codeblock}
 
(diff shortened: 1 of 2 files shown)


jensmaurer:
Looks good to me, but I'm not sure there is a need to fix these. (However, I note these are not in the core section, where showing different programming styles is the intent.)

Eelis:
Oh, I don't know about "need", I just figured it would be nice for the example code fragments not to have unnecessary distractions in them. :)

jwakely:
I don't think these statements serve any purpose in these clauses, so I'd agree with removing them.

tkoeppe:
Agree, not a relevant change, but we've probably spent enough time on it.

PR #1484, opened 2017-02-23
Do not use code font for constexpr constructor and constexpr function

johelegp:
Completes #1153, which missed entries with a newline in-between, and utilities.tex.


tkoeppe:
@jensmaurer: What do you make of this?

jensmaurer:
Looks good to me. (I love the utilities.tex inconsistencies where we have one constexpr constructor phrase with and one without \tcode.)

PR #1489, opened 2017-02-27
[basic] Use less monospace in index entries.

Eelis:
diff

The other two are similar (but harder to show due to reflow caused by this last one.)


jensmaurer:
Looks good to me.

PR #1493, opened 2017-02-27
[macros] Remove terms-and-definitions entries from TOC and bookmarks

tkoeppe:
Following the approach of http://stackoverflow.com/a/2785740.

### DIFF
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -508,12 +508,11 @@
 }
 
 %%--------------------------------------------------
-%% Definitions section
-% usage: \definition{name}{xref}
-%\newcommand{\definition}[2]{\rSec2[#2]{#1}}
-% for ISO format, use:
+%% Definitions section for "Terms and definitions"
+\newcommand{\nocontentsline}[3]{}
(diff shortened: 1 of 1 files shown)


jwakely:
It would be nice to update the preceding comment, "For ISO format use:"

jwakely:
Maybe remove the comment entirely. The commented-out `\newcommand` is old, and with this proposed change the comment "For ISO format use:" doesn't apply to the line that follows it.

tkoeppe:
@jwakely: Done. As a drive-by, I also made the line break starred so we won't get a page break there.

jwakely:
I've adopted this for the Networking TS (https://github.com/cplusplus/networking-ts/commit/eec1cc947748563e37c62d3fd7e519e66011e156)

tkoeppe:
This change has the unfortunate effect of changing references from "3.x" to just "3".

@zygoloid, @jwakely: I'll look into this and will try to find a better solution that restores the references. I'll be in touch.

tkoeppe:
@jwakely: I changed the way I suppress the TOC entry in 7542c70e0647923d54e8a4d47ea4c9bad4f0072e, which is a superior solution. The first solution did not only break reference labels, but also modified the vertical layout slightly.

PR #1512, opened 2017-03-04
Consistent comma after e.g. and i.e.

AlisdairM:
After grepping to determine the preferred idiom, it seems clear that
the standard prefers to follow both e.g. and i.e. with a comma, rather
than omit it.  This patch applies that rule to the few places that
were missing the comma, including in LaTeX comments only for the editors'
benefit.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -4057,7 +4057,7 @@ template <class ExecutionPolicy, class InputIterator, class Predicate>
 \pnum
 \returns \tcode{true} if
 \range{first}{last} is empty or if
-\range{first}{last} is partitioned by \tcode{pred}, i.e. if all elements that satisfy \tcode{pred} appear before those that do not.
+\range{first}{last} is partitioned by \tcode{pred}, i.e., if all elements that satisfy \tcode{pred} appear before those that do not.
 
 \pnum
 \complexity Linear. At most \tcode{last - first} applications of \tcode{pred}.
@@ -4194,7 +4194,7 @@ template<class ForwardIterator, class Predicate>
 
 \begin{itemdescr}
 \pnum
-\requires \tcode{ForwardIterator}'s value type shall be convertible to \tcode{Predicate}'s argument type. \range{first}{last} shall be partitioned by \tcode{pred}, i.e. all elements that satisfy \tcode{pred} shall appear before those that do not.
+\requires \tcode{ForwardIterator}'s value type shall be convertible to \tcode{Predicate}'s argument type. \range{first}{last} shall be partitioned by \tcode{pred}, i.e., all elements that satisfy \tcode{pred} shall appear before those that do not.
 
 \pnum
 \returns An iterator \tcode{mid} such that \tcode{all_of(first, mid, pred)} and \tcode{none_of(mid, last, pred)} are both \tcode{true}.
--- a/source/declarators.tex
+++ b/source/declarators.tex
@@ -2257,7 +2257,7 @@ sometype* q = new sometype[3];  // error, deleted class \tcode{operator new[]}
 \end{example}
 
 \begin{example}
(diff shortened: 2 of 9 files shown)


source/declarators.tex
```diff
@@ -2257,7 +2257,7 @@
 \end{example}
 
 \begin{example}
-One can make a class uncopyable, i.e. move-only, by using deleted
+One can make a class uncopyable, i.e., move-only, by using deleted
```
zygoloid: This seems like abuse of "i.e." -- uncopyable and move-only are not equivalent. More drastic rewording seems necessary here.

tkoeppe: Suggestion: "uncopyable, which implies move-only, by using [...]"

source/algorithms.tex
```diff
@@ -4057,7 +4057,7 @@
 \pnum
 \returns \tcode{true} if
 \range{first}{last} is empty or if
-\range{first}{last} is partitioned by \tcode{pred}, i.e. if all elements that satisfy \tcode{pred} appear before those that do not.
+\range{first}{last} is partitioned by \tcode{pred}, i.e., if all elements that satisfy \tcode{pred} appear before those that do not.
```
zygoloid: More damage to the English language perpretrated by Strunk and White, but OK, this is the dominant American English style, so let's put up with it.

source/threads.tex
```diff
@@ -147,7 +147,7 @@
 An \defn{execution agent} is an entity such as a thread that may perform work in parallel with
 other execution agents. \begin{note} Implementations or users may introduce other kinds of
 agents such as processes or thread-pool tasks. \end{note} The calling agent is determined by
-context, e.g. the calling thread that contains the call, and so on.
+context, e.g., the calling thread that contains the call, and so on.
```
zygoloid: I do not think we need both an 'e.g.' and a 'and so on' here.

source/algorithms.tex
jensmaurer: Personally, I believe the "comma after i.e." thing is totally misguided. Nobody knows what "i.e." and "e.g." means in the expanded form, plus it's Latin anyway, so the argument "oh, it's a complete sentence lookalike, so should have commas around it" is broken. (Have you ever seen the expanded form in English?). We don't put a comma after "but" or "where" either.

source/threads.tex
jensmaurer: I actually thing the entire "e.g." part needs a more drastic rewrite. This just doesn't read well.
And what "call" are we talking about? A "function call" in the syntactic sense of [expr.call]?
Suggestion: "is determined by the context, such as the thread of execution that contains the invocation of the function."  (This is consistent with [intro.multithread] p2.)

Oh, and does this note attempt to define the term "calling agent"? Then it would be totally misguided.

jensmaurer:
One argument why we want a comma after e.g.: LaTeX gets the spacing wrong otherwise.
https://en.wikibooks.org/wiki/LaTeX/Tips_and_Tricks

tkoeppe:
@jensmaurer *that* issue should instead be fixed by either using an appropriate dictionary or by escaping the space, i.e. `i.e.\ `.

jensmaurer:
Editorial meeting consensus: Add definition of "partitioned by" and actually use that in all seven situations. Review all uses of "i.e.": if you specify a rule twice, at least one of them is wrong. Remove them. "e.g." should not be written in normative text, attempt to replace with "for example". Should only be used in notes. Also consider "[ Example ... ]" if containing at least a full sentence.

tkoeppe:
I can take this over. I'll probably create several commits for the various separate issues here. Leave it with me.

tkoeppe:
Stolen and committed (finally) as deb9fb19a039dbcd943a5955064d8c0e3d55a45c.

Issue #1452, opened 2017-02-09
Index all mentions of a grammarterm?


jensmaurer:
As an experiment, I have equipped \grammarterm with adding an index entry. The BNF definitions are in bold. Example:

![grammar](https://cloud.githubusercontent.com/assets/23412755/22763533/d988db30-ee65-11e6-8403-b07431908e22.png)

Is that a direction to pursue?

AlisdairM:
I like it - if for no other reason than it highlights capitalization issues, like for Class-specifier above.

More seriously, not being a core expert, I have gone reaching for this kind of tool many times - it would be extremely helpful to find it one day.

jensmaurer:
Editorial committee consensus: This approach is good. Fix intro text. (Do not remove grammar names from main index.) This makes it easy to see where we screw up.

jensmaurer:
"Try to also add index entries for grammar terms used in the definition of other grammar terms." is mighty hard, because we just mention other non-terminals inside a `bnf` environment, without further markup which we could hook with creating index entries.

We could define a short markup such as \nt{blah} inside bnf environments for the non-terminals, but we would still need to manually iterate through all grammar snippets to add the markup. Postponed until #1821 has been resolved.

PR #1519, opened 2017-03-07
[dcl.dcl] Capitalize grammarterms that start a sentence.

Eelis:
See [dcl.dcl]/1, [dcl.fct.spec]/1, and [class]/1 for existing examples of capitalized grammarterms that start sentences.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -98,9 +98,9 @@ attribute-specifier-seq\opt decl-specifier-seq\opt init-declarator-list\opt{} \t
 
 is divided into three parts.
 Attributes are described in~\ref{dcl.attr}.
-\grammarterm{decl-specifier}{s}, the principal components of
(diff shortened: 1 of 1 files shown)


jensmaurer:
I disagree with this change. Once we start indexing all mentions of \grammarterm #1452 , capitalized \grammarterms will have bad effects on the index. We should instead try to reword the existing phrases so that the \grammarterm is not at the start of a sentence. Maybe going from plural to singular helps.

tkoeppe:
I'm strongly in favour of rewording, and general style guidance to follow the rest of technical writing and avoid starting sentences with symbols or symbol-like terms. (And also avoid  relying on a particular internal vocalization of a symbol, like "Open from 5&ndash;9".)

Eelis:
Ah, rewording would be even nicer! Closing.

PR #1520, opened 2017-03-07
Move plural 's' out of \grammarterm.

Eelis:
Found a bunch more, not sure why I missed these in #1462.


jensmaurer:
Looks good to me.

PR #1528, opened 2017-03-09
[pairs.pair] use code font for 'pair'

jwakely:


tkoeppe:
Looks good. @zygoloid: do it?

jwakely:
Actually don't bother, I'll do this as part of LWG motion 3 which touches the end of the same paragraph.

PR #1529, opened 2017-03-10
[expr.mptr.oper] Add missing indentation in example code.

Eelis:


tkoeppe:
The general idea is that in the core language, we deliberately do not follow any particular code style, so as to demonstrate (by leading by example) that whitespace is not relevant and that the Standard takes no position on formatting.

The library uses a consistent style for its *own* content, but that's a different matter.

Eelis:
Ah ok, I was familiar with the policy but didn't know it was taken to this extreme to deliberately encourage poor style.

PR #1532, opened 2017-03-13
[expr.delete] Properly italicize the term 'delete array', as done elsewhere

AaronBallman:
Elsewhere in this subclause, delete array and delete object are italicized on use; this makes this usage consistent. I spoke with Richard about it at the Kona meeting, and he mentioned this seemed like the correct way to fix it.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -4136,11 +4136,13 @@ object\iref{intro.object} or array created by a
     \terminal{::}\opt{} \terminal{delete [ ]} cast-expression
 \end{bnf}
 
-The first alternative is for non-array objects, and the second is for arrays. Whenever
-the \tcode{delete} keyword is immediately followed by empty square brackets, it shall be
-interpreted as the second alternative.\footnote{A lambda expression with a
-\grammarterm{lambda-introducer} that consists of
-empty square brackets can follow the \tcode{delete} keyword if the lambda expression is
+The first alternative is a
+\defnx{single-object delete expression}{delete!single-object}, and the
+second is an \defnx{array delete expression}{delete!array}.
+Whenever the \tcode{delete} keyword is immediately followed by empty square
+brackets, it shall be interpreted as the second alternative.\footnote{A lambda
+expression with a \grammarterm{lambda-introducer} that consists of empty square
+brackets can follow the \tcode{delete} keyword if the lambda expression is
 enclosed in parentheses.}
 The operand shall be of pointer to object type or of class type. If of
 class type, the operand is contextually implicitly converted\iref{conv}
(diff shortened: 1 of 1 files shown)


burblebee:
I don't think any of these "alternative"s should be in italics except for the 1st ones in [expr.delete]/2 that (I think?) is supposed to be defining what the terms mean (even though they don't read like a definition).

AaronBallman:
I can go either way.

tkoeppe:
How do we normally use `\term`s? For the definition only, or for each use?

jwakely:
We're inconsistent, but it's _meant_ to be only when we introduce the term.

tkoeppe:
Then we should do what Dawn says!

AaronBallman:
I've done what Dawn suggested.

burblebee:
lgtm!  But I think we might need to wait for the edits branch to be created before we can merge this, even though it's a minor fix.  @zygoloid, what do you think?

zygoloid:
Should we be italicizing the first occurrence, even? This is not some grandiose term that we're defining to be referenced later; it is literally used only twice, and neither use is a definition. It seems much more like we're just using these terms to clarify what we mean by "first alternative" and "second alternative". So either we should be defining these two terms separately, and unitalicizing both uses, or we should just be unitalicizing both uses.

AaronBallman:
I think the terms "delete array" and "delete object" are generally useful terms of art that we might want to reuse elsewhere (C.2.9 seems like it could perhaps use them in place of single-object delete and array delete, for instance), but I can see this not being important enough to provide formal definitions for them.

I'm happy to do whatever the editors think is most appropriate, but if we go with formal definitions, perhaps we should modify p1 with:

The first alternative is for non-array objects<ins> and is a _delete object_ expression</ins>, and the second is for arrays<ins>and is a _delete array_ expression</ins>.

Or something along those lines, and then fix up the places where we say "first alternative" or "second alternative" to just say delete object or delete array expression.

zygoloid:
I much prefer the idea of actually defining these terms rather than implying a definition from a parenthetical. Can we give them better names at the same time? "single-object delete expression" and "array delete expression" seem much better than "delete object" and "delete array".

AaronBallman:
I've updated the patch (sorry for the delay).

source/expressions.tex
```diff
@@ -3993,11 +3993,13 @@
...
-the \tcode{delete} keyword is immediately followed by empty square brackets, it shall be
-interpreted as the second alternative.\footnote{A lambda expression with a
-\grammarterm{lambda-introducer} that consists of
-empty square brackets can follow the \tcode{delete} keyword if the lambda expression is
+The first alternative is a
+\defnx{single-object delete expression}{delete!single-object}, and the
+second is an \defnx{array delete expression}{delete!array}.
+Whenever the \tcode{delete} keyword is immediately followed by empty square
+brackets, it shall be interpreted as the second alternative.\footnote{A lambda
+expression with a \grammarterm{lambda-introducer} that consists of empty square
```
Eelis: Trailing whitespace here.

AaronBallman: I made a stab at fixing the whitespace issue. I'll keep an eye on the CI builder.

AaronBallman:
I don't understand the CI build failure. "LaTeX Error: File `ulem.sty' not found" does not seem like something I broke with my patch?

tkoeppe:
No, that's odd. @godbyk: Would you happen to have any idea why the Latex run is failing?

tkoeppe:
Could you rebase this again?

AaronBallman:
I believe it's been properly rebased now.

PR #1533, opened 2017-03-13
Add instructions for helpless Ubuntu users like myself.

AaronBallman:
Getting started on Ubuntu turns out to be a bit more complex than just getting latexmk and running with it. Building the documentation fails without these other packages installed, so we might as well mention that.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -26,6 +26,22 @@ then run the following command to install the other packages that the draft requ
 
    sudo tlmgr install latexmk isodate substr relsize ulem fixme rsfs extract layouts enumitem
 
+---------------------------------------
+Getting Started on Debian-based Systems
+---------------------------------------
+
+Install the following packages:
+
+   sudo apt-get install latexmk texlive-latex-recommended texlive-latex-extra texlive-fonts-recommended
+
+-------------------------
+Getting Started on Fedora
+-------------------------
+
+Install the following packages:
+
+   dnf install texlive texlive-isodate texlive-relsize texlive-ulem texlive-fixme texlive-extract
+
 ------------
 Instructions
 ------------


README.rst
```diff
@@ -26,6 +26,14 @@ then run the following command to install the other packages that the draft requ
 
    sudo tlmgr install latexmk isodate substr relsize ulem fixme rsfs extract layouts enumitem
 
+-------------------------
+Getting Started on Ubuntu
```
tkoeppe: Should this say "Debian" instead? Or "Debianesque systems"?

AaronBallman: It could. I picked Ubuntu because that's what OS I experienced the issue on. I suppose it could say "Getting Started on Systems With apt-get"?

tkoeppe: @AaronBallman: Yes. We call those systems "Debian" :-) (Or Debian-based.)

AaronBallman: Happy to change it however you'd prefer. Would you like me to go with Debian-based or Debian?

tkoeppe: I think "Debian-based" is good.

AaronBallman:
Okay, this should be set with the new title now.

jwakely:
FWIW the TeX packages I install on Fedora are:
texlive texlive-isodate texlive-relsize texlive-ulem texlive-fixme texlive-extract

AaronBallman:
Would you like me to add a Fedora section? If so, let me know what you'd like the title to be and what command you use to install those packages, and I'll add another section to the patch.

burblebee:
@AaronBallman Yes, please add Fedora info as well

AaronBallman:
Am I correct in understand this would be:

sudo yum install <packages>

?

burblebee:
should be

AaronBallman:
I've updated the patch for Fedora as well.

jwakely:
No, yum is deprecated, `sudo dnf install blah blah`

jwakely:
No, yum is deprecated, `dnf install blah blah` but I don't think `sudo` is configured by default, so maybe use:

(as root) `dnf install texlive texlive-isodate texlive-relsize texlive-ulem texlive-fixme texlive-extract`

AaronBallman:
One more try, this time with dnf instead of yum.

jwakely:
Hrm, sorry to be a pain, now I look at it, it would probably be better to put the (as root) after "Fedora" not before the `dnf` command. Otherwise looks good to me. And I don't see any reason we can't merge this now, since it doesn't affect the actual draft.

AaronBallman:
I wasn't too keen on putting (as root) after Fedora, but I didn't mind putting it in the instructions. How does this look to you?

tkoeppe:
I wouldn't even necessarily say "as root" at all -- people should know how to use their system; if they're already in the business of installing packages, they probably know how to do it right.

AaronBallman:
I can certainly remove it, but we do show sudo in the other two examples, and it seems somewhat strange to leave it off for Fedora. As @jwakely pointed out, sudo is not configured by default, but perhaps we could still use it for this case as well (and assume people know how to configure sudo or drop into su manually)?

tkoeppe:
Yes, that's true, and we can keep it in, I don't mind. It was just an idea.

burblebee:
Why not do sudo here too?

README.rst
```diff
@@ -26,6 +26,22 @@ then run the following command to install the other packages that the draft requ
...
+
+   sudo apt-get install latexmk texlive-latex-recommended texlive-latex-extra texlive-fonts-recommended
+
+-------------------------
+Getting Started on Fedora
+-------------------------
+
+Install the following packages (as root):
+
+   dnf install texlive texlive-isodate texlive-relsize texlive-ulem texlive-fixme texlive-extract
```
burblebee: add sudo here to be consistent with Debian-based instructions

AaronBallman:
Added sudo and removed the (as root).

jwakely:
No, please don't add `sudo` to the Fedora instructions, because that doesn't work on Fedora. Fedora != Ubuntu. It's not set up so you can just bork your system by adding `sudo` (https://xkcd.com/149/)

jwakely:
I'm fine with just dropping the "(as root)" part and assuming Fedora users building the standard know how to install packages, but a command that doesn't work is not helpful. I'd rather not mention Fedora at all than have a command that won't work.

AaronBallman:
Updated to remove the sudo command for Fedora.

README.rst
```diff
@@ -26,6 +26,22 @@ then run the following command to install the other packages that the draft requ
...
+---------------------------------------
+Getting Started on Debian-based Systems
+---------------------------------------
+
+Install the following packages:
+
+   sudo apt-get install latexmk texlive-latex-recommended texlive-latex-extra texlive-fonts-recommended
+
+-------------------------
+Getting Started on Fedora
```
tkoeppe: bleh, maybe not, not unless someone can actually test this on a few systems.

PR #1545, opened 2017-03-16
[any.modifiers] fix emplace return type for 'any'

AlisdairM:
LWG issue 2857 was applied verbatim, without noticing
that 'ValueType' had been previously changed to 'T'.
This makes things consistent by choosing 'T' as the
simpler form, more consistent with 'optional' and
'variant'.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -5547,9 +5547,9 @@ public:
 
   // \ref{any.modifiers}, modifiers
   template <class T, class... Args>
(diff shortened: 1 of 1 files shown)


jwakely:
This matches the intention of the changes to `any`, which were done because the "value type" that gets store is the decayed type, so calling the undecayed type `ValueType` is misleading.

tkoeppe:
@jwakely: merge that business

jwakely:
Are we open for that kind of business?

PR #1548, opened 2017-03-16
[containers] Rephrase deduction guide constraints

### LINKED ISSUE #1524: Assorted wording issues with P0433R2
In https://github.com/cplusplus/draft/commit/f913f2137c823f16940f907f5a6676972d95a913#diff-0fb8725b6f959c00f8cef02d54e8c125R1021 (currently in the motions-2017-03-lwg-19 branch) we have:

> Likewise, the extent to which an implementation determines that a type cannot be
> an allocator is unspecified, except that as a minimum a type \tcode{A} not satisfying
> both of the following conditions shall not qualify as an allocator:
> 
> - [condition 1]
> - [condition 2]

It's not clear whether a type that satisfies only one of the two conditions is guaranteed to fail to qualify as an allocator or not. Perhaps the sentence can be editorially reworded to make the intent clearer?

@jwakely also pointed out that the paper pervasively uses wording of the form "if it has an `X` template parameter that is called with a type that _meows_", which is problematic because a) deduction guides are not "called", and b) "that is called ..." is actually modifying "template parameter", which also can't be "called".

Finally, the input iterator and allocator weasel-wording are placed in [sequence.reqmts] but now are used for all containers, not just sequence containers. Perhaps they should be moved to a better place?

jwakely:
Replace "is called with" wording that doesn't apply to deduction guides.

Move rules about qualifying as input iterators or allocators to
[container.requirements.general].

Fixes #1524

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -700,6 +700,31 @@ linear													\\ \rowsep
 
 \end{libreqtab4a}
 
+\pnum
+In this Clause the behavior of certain member functions and deduction guides
+depends on whether types qualify as input iterators or allocators.
+The extent to which an implementation determines that a type cannot be an input
+iterator is unspecified, except that as a minimum integral types shall not qualify
+as input iterators.
+Likewise, the extent to which an implementation determines that a type cannot be
+an allocator is unspecified, except that as a minimum a type \tcode{A} shall not qualify
+as an allocator unless it satisfies both of the following conditions:
+
+\begin{itemize}
+\item The \grammarterm{qualified-id} \tcode{A::value_type}
+is valid and denotes a type~(\ref{temp.deduct}).
+
+\item The expression \tcode{declval<A\&>().allocate(size_t\{\})}
+is well-formed when treated as an unevaluated operand.
+\end{itemize}
+
+\pnum
+\begin{note}
+Implementations may impose additional conditions which must be satisfied
+to qualify as an input iterator or an allocator.
+\end{note}
+
+
 \rSec2[container.requirements.dataraces]{Container data races}
 
 \pnum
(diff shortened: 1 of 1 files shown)


source/containers.tex
```diff
@@ -700,6 +700,31 @@
 
 \end{libreqtab4a}
 
+\pnum
+In this Clause the behavior of certain member functions and deduction guides
```
timsong-cpp: ...`and Clause~\ref{strings}`. Also need to update the cross reference in [string.cons] after the move.

source/containers.tex
```diff
@@ -1002,26 +1027,10 @@
...
-an allocator is unspecified, except that as a minimum a type \tcode{A} not satisfying
-both of the following conditions shall not qualify as an allocator:
-
-\begin{itemize}
-\item The \grammarterm{qualified-id} \tcode{A::value_type}
-is valid and denotes a type~(\ref{temp.deduct}).
-
-\item The expression \tcode{declval<A\&>().allocate(size_t\{\})}
-is well-formed when treated as an unevaluated operand.
+if it has an \tcode{InputIterator} template parameter and a type that does not
```
timsong-cpp: Hmm, does "has a template parameter named \tcode{InputIterator}" sound better?

source/containers.tex
timsong-cpp: While we are in the vicinity, `basic_string`'s [deduction guide specification](https://github.com/cplusplus/draft/blob/master/source/strings.tex#L1491) is missing a `\pnum`.

jwakely: Fixed.

source/containers.tex
```diff
@@ -700,6 +700,31 @@
...
+\begin{itemize}
+\item The \grammarterm{qualified-id} \tcode{A::value_type}
+is valid and denotes a type~(\ref{temp.deduct}).
+
+\item The expression \tcode{declval<A\&>().allocate(size_t\{\})}
+is well-formed when treated as an unevaluated operand.
+\end{itemize}
+
+\pnum
+\begin{note}
```
zygoloid: I don't like this note as-written: it suggests that implementations can impose unbounded constraints on the types in questions, rather than merely constraints that are derivable from the allocator or input iterator requirements.

I'm inclined to merge this without the new note.

zygoloid:
Addressed timsong's comments, removed note, merged as d2b6fb6e.

PR #1549, opened 2017-03-18
[swappable.requirements] Mark up the swappable requirement in code font

AlisdairM:
Requiements like CopyConstrible and MoveAssignable are usually marked
up in code font.  This PR applies the same pattern to the swappable
requirement, but not 'swappable with'.  The need for this change was
most obvious for types that combined the swappble requirement with
others, such as MoveConstructible.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -3188,7 +3188,7 @@ template<class ExecutionPolicy, class BidirectionalIterator>
 \begin{itemdescr}
 \pnum
 \requires
-\tcode{*first} shall be swappable~(\ref{swappable.requirements}).
+\tcode{*first} shall be \tcode{swappable}~(\ref{swappable.requirements}).
 
 \pnum
 \effects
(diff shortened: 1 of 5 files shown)


timsong-cpp:
`CopyConstructible` etc. are CamelCased and apply to types, while "swappable" applies to expressions (hence the "lvalues of type `T` shall be swappable" wording). Not sure these are exactly analogous.

zygoloid:
If this is a concept, we should change it to be an actual concept now they're part of the language. If not, code font is inappropriate. Either way, this is LWG territory.

geoffromer:
In addition to being camel-cased, IIUC pseudo-concepts like `CopyConstructible` aren't supposed to be used as adjectives; we're supposed to say things like "satisfies the `CopyConstructible` requirements" instead. I don't think that would be an improvement (and as Tim observes, applying that pattern to values rather than types would be an innovation).

I suggest instead just dropping the code-font from the [swappable.requirements] section heading.

tkoeppe:
LWG reflector discussion seemed to agree that the comparison with CopyConstructible is a false analogy and that code font is inappropriate for "swappable". Instead, we're removing the undesired code font from the heading; fixed by e664eb29e22c8c31d8255948813fd0ba78c3a85a.

PR #1554, opened 2017-03-18
Move \opt out of \grammarterm

timsong-cpp:


tkoeppe:
Thanks, this looks good. I think we also need to move `\opt` out of `\terminal`s.

PR #1555, opened 2017-03-18
Remove bad \grammarterm for "injected-class-name"

timsong-cpp:


tkoeppe:
Why so many commits -- squash and rebase?

timsong-cpp:
Figured separate commits might be easier to review, but OK, squashed.

source/statements.tex
```diff
@@ -672,8 +672,8 @@
 class type \tcode{C}, the \grammarterm{unqualified-id}{s}
 \tcode{begin} and \tcode{end} are looked up in the scope of \tcode{C}
 as if by class member access lookup~(\ref{basic.lookup.classref}), and if either
-(or both) finds at least one declaration, \grammarterm{begin-expr} and
-\grammarterm{end-expr} are \tcode{__range.begin()} and \tcode{__range.end()},
+(or both) finds at least one declaration, \textit{begin-expr} and
```
tkoeppe: We don't really want to introduce *new* `\textit`s. Consult with @jensmaurer for details.

timsong-cpp: Every other occurrence of *meow-expr* in this very section uses `\textit`, which is why I used them here.

tkoeppe: I understand that. We still need to talk about this.

jensmaurer: It seems to me _begin-expr_ should be \placeholder, actually.

timsong-cpp: So `\placeholder` it throughout? That was admittedly my first thought, before seeing all-`\textit`s in the surroundings.

jensmaurer: See #1558.

source/declarators.tex
```diff
@@ -2110,7 +2110,7 @@
     attribute-specifier-seq\opt decl-specifier-seq\opt declarator virt-specifier-seq\opt{} \terminal{ = default ;}
 \end{ncbnf}
 
-is called an \grammarterm{explicitly-defaulted} definition.
+is called an \term{explicitly-defaulted} definition.
 A function that is explicitly defaulted shall
```
jensmaurer: This should be `\defn{explicitly-defaulted definition}`, probably with "definition!explicitly-defaulted" as the main index entry and a "see" pointer for "explicitly-defaulted definition".

timsong-cpp: There's already a `\indextext{definition!function!explicitly-defaulted}` ~7 lines above, which is why I went with `\term`.

jensmaurer: In my opinion, \term should only (if at all) survive for stuff that is local enough not to ever be referenced from elsewhere, and thus not needing an index entry at all. For this particular case, it seems the \indextext from before should simply be folded into the \defnx to be used after "is called an".

I'm slightly concerned that we seem to define "explicitly-defaulted definition" as the technical term, although we mean "explicitly-defaulted function definition". That probably needs a bit of research what we actually want here.

jensmaurer:
This needs a rebase.

timsong-cpp:
I dropped the explicitly-defaulted change; the indexing issue is above my pay grade. The only remaining change is injected-class-name.

tkoeppe:
@zygoloid: What do you think? It looks like the PR in its present form improves consistency.

jensmaurer:
Yes, it removes a bad \grammarterm.

PR #1558, opened 2017-03-19
[stmt.ranged] Do not use \grammarterm for begin-expr and end-expr.

jensmaurer:


source/statements.tex
```diff
@@ -659,24 +659,24 @@
...
 \tcode{begin} and \tcode{end} are looked up in the scope of \tcode{C}
 as if by class member access lookup~(\ref{basic.lookup.classref}), and if either
-(or both) finds at least one declaration, \grammarterm{begin-expr} and
-\grammarterm{end-expr} are \tcode{__range.begin()} and \tcode{__range.end()},
+(or both) finds at least one declaration, \placeholder{begin-expr} and
+\placeholder{end-expr} are \tcode{__range.begin()} and \tcode{__range.end()},
 respectively;
 
-\item otherwise, \textit{begin-expr} and \textit{end-expr} are \tcode{begin(__range)}
+\item otherwise, \placeholder{begin-expr} and \placeholder{end-expr} are \tcode{begin(__range)}
```
tkoeppe: @jensmaurer: This is fine, but when do we use literal `__` and when do we use `\xname`?

jensmaurer: I'm sorry, I don't quite understand the question. What's \xname?

And yes, I'm not totally happy with __range instead of introducing some readable variable and saying its name is unique.

tkoeppe: `\xname` adds double underscores, so we don't have to spell those out literally, and it takes care to add some small amount of whitespace so the two characters don't run together. I'm not sure if that motivation is ultimately justified, it's just a thing we have.

jensmaurer: It seems we should use \xname and \mname everywhere we currently have two literal underscores in C++ identifiers.

PR #1560, opened 2017-03-19
[library] Add a noun 'subclause' or 'annex' to references that are used as nouns

tkoeppe:
It is awkward to treat symbols or references like complete nouns. This change adds the words "subclause" or "annex" when references are used.

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -1126,12 +1126,12 @@ macros in \Cpp standard library headers.\footnote{In particular, including the
 standard header \tcode{<iso646.h>} or \tcode{<ciso646>} has no effect.}
 
 \pnum
-\ref{depr.c.headers}, C standard library headers, describes the effects of using
-the \tcode{\placeholder{name}.h} (C header) form in a \Cpp program.\footnote{ The
+Annex~\ref{depr.c.headers}, C standard library headers, describes the effects of using
+the \tcode{\placeholder{name}.h} (C header) form in a \Cpp program.\footnote{The
 \tcode{".h"} headers dump all their names into the global namespace, whereas the
 newer forms keep their names in namespace \tcode{std}. Therefore, the newer
 forms are the preferred forms for all uses except for \Cpp programs which are
-intended to be strictly compatible with C. }
+intended to be strictly compatible with C.}
 
 \pnum
 Annex K of the C standard describes a large number of functions,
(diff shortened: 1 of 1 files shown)


tkoeppe:
I'm not 100% convinced that this change is the best way forward. Perhaps there's a different kind of rewording that avoids the awkwardness of references-as-nouns that's less heavy-handed. What do you think?

source/lib-intro.tex
```diff
@@ -1318,9 +1318,9 @@
 \rSec3[using.overview]{Overview}
 
 \pnum
-This section describes how a \Cpp program gains access to the facilities of the
-\Cpp standard library. \ref{using.headers} describes effects during translation
-phase 4, while~\ref{using.linkage} describes effects during phase
+This subclause describes how a \Cpp program gains access to the facilities of the
```
jensmaurer: Can we leave the general section -> subclause rename separate?

jensmaurer:
I think for the phrases in question here, these changes look good. If someone thinks we should rephrase this stuff in more detail (to avoid the issue entirely), let's do that some other time.

source/lib-intro.tex
```diff
@@ -1165,7 +1165,7 @@
 \pnum
 Table~\ref{tab:c.annex.k.names} lists the Annex K names
 that may be declared in some header.
-These names are also subject to the restrictions of~\ref{macro.names}.
+These names are also subject to the restrictions of subclause~\ref{macro.names}.
```
tkoeppe: That's better, indeed.

source/lib-intro.tex
```diff
@@ -1138,12 +1138,12 @@
 standard header \tcode{<iso646.h>} or \tcode{<ciso646>} has no effect.}
 
 \pnum
-\ref{depr.c.headers}, C standard library headers, describes the effects of using
-the \tcode{\placeholder{name}.h} (C header) form in a \Cpp program.\footnote{ The
+Annex~\ref{depr.c.headers}, C standard library headers, describes the effects of using
```
tkoeppe: I'm also not sure if we can actually call it "Annex" if it refers to a strict subsection. Similarly, we only ever say "Clause" for the top-level chapters.

zygoloid:
ISO drafting guidelines (7th edition) subclause 20.4's examples suggest that we can use "A.4" or perhaps (surprisingly!) "Clause A.4" but not "Annex A.4".

tkoeppe:
I'm retracting this PR, I don't think it makes sense anymore. The rest of the clause use the same style as status quo, and unless we can think of a systematic improvement, that's better than having small deviations.

PR #1563, opened 2017-03-20
Rephrase "possibly const value" in terms of types

### LINKED ISSUE #1497: "a possibly const value of type X"
Change to "a value of type `X` or `const X`."

(library clauses)

26.2.6

jwakely:
Fixes #1497

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -866,7 +866,7 @@ and given a container type \tcode{X} having a \tcode{value_type} identical to \t
 and an \tcode{allocator_type} identical to \tcode{allocator_traits<A>::rebind_alloc<T>}
 and given an lvalue \tcode{m} of type \tcode{A},
 a pointer \tcode{p} of type \tcode{T*},
-an expression \tcode{v} of type (possibly \keyword{const}) \tcode{T},
+an expression \tcode{v} of type \tcode{T} or \tcode{const T},
 and an rvalue \tcode{rv} of type \tcode{T},
 the following terms are defined. If \tcode{X}
 is not allocator-aware or is a specialization of \tcode{basic_string},
@@ -2615,7 +2615,7 @@ In this subclause,
 \tcode{a2} denotes a value of a type with nodes compatible with type
 \tcode{X} (\tref{container.node.compat}),
 \item
-\tcode{b} denotes a possibly \keyword{const} value of type \tcode{X},
+\tcode{b} denotes a value or type \tcode{X} or \tcode{const X},
 \item
 \tcode{u} denotes the name of a variable being declared,
 \item
@@ -2625,9 +2625,9 @@ when \tcode{X} supports unique keys,
 \tcode{a_eq} denotes a value of type \tcode{X}
 when \tcode{X} supports multiple keys,
 \item
-\tcode{a_tran} denotes a possibly \keyword{const} value of type \tcode{X}
+\tcode{a_tran} denotes a value of type \tcode{X} or \tcode{const X}
 when the \grammarterm{qualified-id}
-\tcode{X::key_compare::is_transpa\-rent} is valid
+\tcode{X::key_compare::is_transparent} is valid
 and denotes a type\iref{temp.deduct},
 \item
 \tcode{i} and \tcode{j}
@@ -2654,7 +2654,7 @@ that models \tcode{\exposconcept{container-compatible-range}<value_type>},
 \item
 \tcode{k} denotes a value of type \tcode{X::key_type}, and
 \item
-\tcode{c} denotes a possibly \keyword{const} value of type \tcode{X::key_compare};
+\tcode{c} denotes a value of type \tcode{X::key_compare} or \tcode{const X::key_compare};
 \item
 \tcode{kl} is a value such that \tcode{a} is partitioned\iref{alg.sorting}
 with respect to \tcode{c(x, kl)},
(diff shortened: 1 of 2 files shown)


source/containers.tex
```diff
@@ -76,8 +76,8 @@
 \tcode{T}, \tcode{a} and \tcode{b}
 denote values of type \tcode{X}, \tcode{u}
```
zygoloid: Should `b` be of type `X` or `const X`? If not, why do we have both `b`and `r` with different descriptions?

jwakely: `b` is used in `a.swap(b)` so must be non-const. I have no idea why we have `b` and `r` described differently. I'll do some archaeology.

jwakely: N3035 says:

> In Tables 91 and 92, `X` denotes a container class containing objects of type `T`, `a` and `b` denote values of type
`X`, `u` denotes an identifier, `r` denotes an lvalue or a const rvalue of type `X`, and `rv` denotes a non-const rvalue
of type `X`.

By N3090 the distinction had been lost.

jwakely: http://www.open-std.org/jtc1/sc22/wg21/docs/lwg-defects.html#704

jwakely: I think we could get rid of `r` and use `b` in its place.

tkoeppe:
@zygoloid: Should/can we adopt this change? Seems like a useful improvement.

source/containers.tex
```diff
@@ -72,12 +72,10 @@
 In Tables~\ref{tab:containers.container.requirements},
 \ref{tab:containers.reversible.requirements}, and
 \ref{tab:containers.optional.operations}
-\tcode{X} denotes a container class containing objects of type
-\tcode{T}, \tcode{a} and \tcode{b}
-denote values of type \tcode{X}, \tcode{u}
```
zygoloid: How confident are you that `a` and `b` here weren't supposed to be possibly-const? The distinction between them and `r` makes me worry that (eg) the `r = a` requirement below was supposed to apply for both the case where `a` is of type `X` and the case where `a` is of type `const X`.

zygoloid:
I'm definitely happy to merge all but the first change in containers.tex. For that one change, I'm a little concerned that we're papering over a (pre-existing) wording bug.

source/containers.tex
jwakely: The expressions `a = rv` and `a.swap(b)` require non-const, so `a` and `b` are definitely not allowed to be const in those rows. However, you have revealed an existing bug: most of the expressions _should_be using possibly-const values, e.g. `X(a)` and `X u(a); X u = a;` (yes I know the latter aren't expressions :smiley:) and `(&a)->~X()`, and `a.begin()` and `a.end()`, `a.cbegin()` and `a.cend()`, `a == b` and `a != b`, `a.size()` etc. etc. etc. etc.

So I think we should make `a` and `b` possibly const, and introduce new variables for use in the handful of rows that aren't valid for const objects.

jwakely: I'll create an LWG issue for that.

jwakely: https://wg21.link/lwg3028

jensmaurer:
@jwakely reports that Marshall Clow is opposed to changing "const value of type X". Editorial meeting consensus: This is technically wrong, though.

jwakely:
Sorry for the accidental close.

source/containers.tex
```diff
@@ -462,7 +460,7 @@
 
 \indexcont{operator=}%
 \begin{itemdecl}
-r = a
+b = a
```
jensmaurer, line 463: This seems to be a pre-existing wording bug, as mentioned by @zygoloid : We don't require that "b" is an lvalue, which means this must work for xvalues, too. And that, effectively, means that operator= cannot be ref-qualified. Is that the intent?

jensmaurer (approved):
Looks mostly good.

@tkoeppe , please have a look.

source/containers.tex
jwakely: I think the current proposed resolution for https://wg21.link/lwg3028 fixes that. The commit msg for c69af479dacad40d86a8d2429f69cd302fc77047 refers to that issue.

jensmaurer: So, we're now intentionally creating a merge conflict between LWG3028 and this editorial issue? Could we at least advance LWG3028 expeditiously?

tkoeppe: Any news on this?

tkoeppe: @jwakely Ping?

tkoeppe:
Oh OK, but in that case we probably won't need this PR, either, since it'll come back in a paper?

jwakely:
I don't think there's going to be a paper making these changes. I'll rebase it to remove the part that conflicts with 3028, so we can make the non-conflicting editorial changes.

tkoeppe:
Oh, I see, thanks! Let me know when it's ready.

source/containers.tex
jwakely: I'm resolving this thread, because I've force-pushed a new revision without the conflicts.

source/containers.tex
```diff
@@ -2625,9 +2625,9 @@
 \tcode{a_eq} denotes a value of type \tcode{X}
 when \tcode{X} supports multiple keys,
 \item
-\tcode{a_tran} denotes a possibly \keyword{const} value of type \tcode{X}
+\tcode{a_tran} denotes a value of type \tcode{X} or \tcode{const X}
 when the \grammarterm{qualified-id}
-\tcode{X::key_compare::is_transpa\-rent} is valid
+\tcode{X::key_compare::is_transparent} is valid
```
jwakely, line 2630: This fits on a line now, so we don't need the hyphenation.

tkoeppe: Beautiful.

tkoeppe:
So is this editorial now? I'll remove the "lwg" label.

tkoeppe:
@jensmaurer, @zygoloid: would you care to provide a corish opinion and assent?

jensmaurer:
"value of type X or const X" is certainly in-your-face obvious and looks good to me.

off-topic: I wish we had some shorthand along the lines of "value of type cv X" or so.

jensmaurer (approved):

PR #1564, opened 2017-03-20
Consistently use math mode for index notation and ellipses in INVOKE expressions

jwakely:
This changes pseudocode expressions of the form `INVOKE(f, t1, t2, ..., tN)` so that the index notation and ellipsis are in math mode, and so not mistaken for C++ variables or the `...` token.

[func.require]:
func require
and:
func require2
[func.bind] already formatted the indices this way, but used code font for the `...` between the metavariables:
func bind
[func.memfn]:
func memfn
[futures.task.members]:
futures task members
and:
futures task members2

### DIFF
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -5038,7 +5038,7 @@ template <class F>
 \begin{itemdescr}
 \pnum
 \requires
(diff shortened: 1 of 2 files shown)


tkoeppe:
Looks legit.

PR #1565, opened 2017-03-20
[language.support] Reserve \xref for references to the C standard.

### LINKED ISSUE #1561: Don't abuse \xref
The `\xref` "see also" annotation is only meant for references to The Other Standard, not for simple cross-references within our own document. Such cross-references should simply say "See also" in ordinary text.

I have not yet surveyed the entire text for occurrences of this problem, but I fixed it in [library] and it's still rampant in [language.support].

jensmaurer:
Fixes #1561.


tkoeppe:
This is great, thanks! Since you're removing trailing full stops, could you do a quick grep for other such occurrences? I think I moved some full stops to Annex D.

jensmaurer:
Done.

source/utilities.tex
```diff
@@ -815,7 +815,7 @@
 \throws Nothing.
 \end{itemdescr}
 
-\xref ISO C~7.22.1.3, ISO C~7.22.1.4.
+\xref ISO C 7.22.1.3, ISO C 7.22.1.4
```
zygoloid: Elsewhere we don't repeat ISO C for each reference if there is more than one. (See [c.mb.wcs], [c.math.abs], [c.math.fpclass].) Suggest removing the second ISO C here.

tkoeppe: +1, good point.

tkoeppe: @jensmaurer: Ping.

jensmaurer: Fixed.

jensmaurer:
Rebased and fixed duplicate "ISO C".

PR #1566, opened 2017-03-20
[numerics] Use absolute-value markers in math mode.

### LINKED ISSUE #1534: [fs.definitions] needs overhaul
[fs.definitions] thoroughly fails to follow the rules for a "Terms and definitions" clause in an ISO standard. The definitions are an inconsistent mix of half-sentences (with a trailing period but a leading lowercase word) and full sentences, or in some cases multiple sentences and even multiple paragraphs with nested lists.

Each definition should be rewritten to conform to this rule from the ISO directives: "The definition shall be written in such a form that it can replace the term in its context. It shall not start with an article ("the", "a") nor end with a full stop."

We should also merge these definitions into [definitions]. (Merging [intro.defs] and [definitions] is also probably required for ISO directive conformance, but I'd prefer to investigate whether it would be feasible to  address that issue by splitting the standard into two Parts instead.)

### LINKED ISSUE #1537: Absolute values need to be displayed as maths, not code.
We should never say `\tcode{|x|}`, but rather `$\lvert\tcode{x}\rvert$` or equivalently `$|\tcode{x}|$`.

jensmaurer:
Fixes #1537.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -9690,10 +9690,10 @@ template <class M, class N>
 \begin{itemdescr}
 \pnum
 \requires
-\tcode{|m|} and \tcode{|n|} shall
+$|\tcode{m}|$ and $|\tcode{n}|$ shall
 be representable as a value of \tcode{common_type_t<M, N>}.
 \begin{note} These requirements ensure, for example,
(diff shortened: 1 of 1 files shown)


tkoeppe:
Wrong issue reference?

jensmaurer:
Fixed.

PR #1577, opened 2017-03-25
[string.cons] Consolidate two functions into one description

tkoeppe:
image

Superficially, this change removes a fair amount of duplication in the specification.

More importantly, though, this may be a step towards getting rid of those string effect tables that seem awkward and presentationally wasteful.


jensmaurer:
Looks good to me.

PR #1581, opened 2017-03-28
[meta.unary.prop] Fix index entry for is_aggregate.

Eelis:
Looks like a copy-paste mistake, because the next one is is_signed.

Note that '!class' makes sense for is_signed's index entry because there is also an entry for 'is_signed!numeric_limits', but there is no other is_aggregate in the stdlib, so I removed the '!class' (making it consistent with the index entries for e.g. is_final and is_abstract).

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -15926,7 +15926,7 @@ notwithstanding the restrictions of~\ref{declval}.
  can be marked with \tcode{final}. \end{note}                                        &
  If \tcode{T} is a class type, \tcode{T} shall be a complete type.                          \\ \rowsep
 
-\indexlibrary{\idxcode{is_aggregate_signed}!class}%
+\indexlibrary{\idxcode{is_aggregate}}%
 \tcode{template <class T>}\br
   \tcode{struct is_aggregate;}           &
  \tcode{T} is an aggregate type~(\ref{dcl.init.aggr}) &


source/utilities.tex
```diff
@@ -15926,7 +15926,7 @@
  can be marked with \tcode{final}. \end{note}                                        &
  If \tcode{T} is a class type, \tcode{T} shall be a complete type.                          \\ \rowsep
 
-\indexlibrary{\idxcode{is_aggregate_signed}!class}%
```
tkoeppe: And why does is_signed have `!class` below? Confused.

tkoeppe: I see, makes sense.

jwakely:
Huh, I already fixed this during the motion editing.  My review comments for library motion 20b were:

> I changed “\cv \tcode{void}” to “\cv~\tcode{void}” in the new table row because it appeared as cvvoid. I also fixed the new index entry from is_aggregate_signed!class to is_aggregate.

tkoeppe:
@jwakely: Did you push that?

jwakely:
I thought so, but it doesn't really matter now.

jwakely:
I guess not, because I am shown as contributing to c92cb1c33c885ef9c49a87bd9fb677e7a46e1ef0 (the tip of the lwg-20b branch, which has the fix for the cv void spacing) but my local caf741d68c0854d2945062f833f841c779472e61 commit isn't on github. I must have fixed the first problem, pushed, fixed the second problem, and not pushed it. Oops.

tkoeppe:
@jwakely: And you had *one* job! :-)

PR #1584, opened 2017-03-28
Do not use \tcode for non-syntax.

### LINKED ISSUE #106: Inconsistent formatting of "public" and "inline" in sentences
There are a bunch of inconsistencies around whether to `\tcode` `public` and `inline`, e.g. in "public inline member function" throughout the standards text.

jensmaurer:
This applies to inline, public, protected, private, and friend.

Fixes #106.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -6746,12 +6746,6 @@ template is neither a
 matrix class nor a field class.
 However, it is a very useful building block for designing such classes.}
 
-\pnum
-An implementation is permitted to qualify any of the functions declared in
-\tcode{<valarray>}
-as
-\tcode{inline}.
-
 \rSec3[valarray.cons]{\tcode{valarray} constructors}
 
 \indexlibrary{\idxcode{valarray}!constructor}%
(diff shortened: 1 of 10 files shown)


source/numerics.tex
```diff
@@ -6752,12 +6752,6 @@
 matrix class nor a field class.
 However, it is a very useful building block for designing such classes.}
 
-\pnum
-An implementation is permitted to qualify any of the functions declared in
-\tcode{<valarray>}
-as
-\tcode{inline}.
-
```
Eelis: Hmm, this is erasing the whole paragraph rather than just the tcode bit. Was that intended?

jensmaurer: Yes, the statement is entirely redundant with [global.functions] p1

> It is unspecified whether any non-member functions in the C++ standard library are defined as inline (10.1.6).

and [member.functions] p1

> It is unspecified whether any member functions in the C++ standard library are defined as inline (10.1.6).

Eelis: Ah ok, was just checking because this edit doesn't match the commit description.

zygoloid:
Looks good, needs rebasing.

PR #1585, opened 2017-03-28
Use 'subclause', not 'section'.

### LINKED ISSUE #307: Inconsistent use of "in this section" vs "in this sub-clause"
"In this section" is far more common, but we have the following mentions of sub-clauses:

[expr.const]
[meta.unary]
[meta.unary.prop.query]
[meta.rel]
[meta.trans]
[meta.numerics]
[rand.eng]
[futures.state]

jensmaurer:
The ISO Directives, Part 2, prescribe the term 'subclause'.

Fixes #307.

### DIFF
--- a/source/back.tex
+++ b/source/back.tex
@@ -25,8 +25,8 @@ called \defn{LIA-1} in this document.
 
 \clearpage
 \renewcommand{\glossaryname}{Cross references}
-\renewcommand{\preglossaryhook}{This annex lists each section label and the
-corresponding section number and page number, in alphabetical order by label.\\}
+\renewcommand{\preglossaryhook}{This annex lists each clause or subclause label and the
+corresponding clause or subclause number and page number, in alphabetical order by label.\\}
 \twocolglossary
 \renewcommand{\leftmark}{\glossaryname}
 {
@@ -37,7 +37,7 @@ corresponding section number and page number, in alphabetical order by label.\\}
 \clearpage
 \input{xrefdelta}
 \renewcommand{\glossaryname}{Cross references from ISO \CppXVII}
-\renewcommand{\preglossaryhook}{All section labels from
+\renewcommand{\preglossaryhook}{All clause and subclause labels from
 ISO \CppXVII (ISO/IEC 14882:2017, \doccite{Programming Languages --- \Cpp})
 are present in this document, with the exceptions described below.\\}
 \renewcommand{\leftmark}{\glossaryname}
@@ -74,7 +74,7 @@ corresponding page in the Grammar summary\iref{gram}.\\}
 }
 
 \clearpage
-\renewcommand{\preindexhook}{The entries in this section are rough descriptions; exact
+\renewcommand{\preindexhook}{The entries in this index are rough descriptions; exact
 specifications are at the indicated page in the general text.\\}
 \renewcommand{\indexname}{Index of implementation-defined behavior}
 \renewcommand{\leftmark}{\indexname}
--- a/source/diagnostics.tex
+++ b/source/diagnostics.tex
@@ -939,7 +939,7 @@ to identify the source and encoding of a particular category of error code.
 Classes may be derived from \tcode{error_category} to support
 categories of errors in addition to those defined in this document.
 Such classes shall behave as specified in this
-subclause. \begin{note} \tcode{error_category} objects are
+subclause~\ref{syserr.errcat}. \begin{note} \tcode{error_category} objects are
 passed by reference, and two such objects
 are equal if they have the same address. This means that applications using
 custom \tcode{error_category} types should create a single object of each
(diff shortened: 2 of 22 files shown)


source/back.tex
```diff
@@ -22,8 +22,8 @@ \chapter{Bibliography}
 
 \clearpage
 \renewcommand{\glossaryname}{Cross references}
-\renewcommand{\preglossaryhook}{This annex lists each section label and the
-corresponding section number and page number, in alphabetical order by label.\\}
+\renewcommand{\preglossaryhook}{This annex lists each subclause label and the
+corresponding subclause number and page number, in alphabetical order by label.\\}
```
zygoloid: These ones mean "clause or subclause". We could use "[...] lists each clause or subclause label and [...]" but perhaps there's a better way to say that?

source/back.tex
```diff
@@ -34,7 +34,7 @@ \chapter{Bibliography}
 \clearpage
 \input{xrefdelta}
 \renewcommand{\glossaryname}{Cross references from ISO \CppXIV}
-\renewcommand{\preglossaryhook}{All section labels from
+\renewcommand{\preglossaryhook}{All subclause labels from
```
zygoloid: Likewise, even though we haven't removed any Clause labels yet.

source/back.tex
```diff
@@ -71,7 +71,7 @@ \chapter{Bibliography}
 }
 
 \clearpage
-\renewcommand{\preindexhook}{The entries in this section are rough descriptions; exact
+\renewcommand{\preindexhook}{The entries in this subclause are rough descriptions; exact
```
zygoloid: This is not a subclause, it's an index.

source/back.tex
jensmaurer: I'm going for "clause or subclause" for now.

source/back.tex
jensmaurer: Fixed.

jensmaurer:
Rebased and addressed review comments.

tkoeppe:
Just a random observation, because I was also working on this: in lib-intro.tex, the proposed change is problematic, because we currently say for section `X`:

> This section describes Foo. Subclause `X.1` describes meta-Foos, subclause `X.2` describes para-Foos.

If you change the first "section" to "subclause", then the hierarchy is lost.

jensmaurer:
@tkoeppe: lib-intro.tex is rather inconsistent here. My suggestion (implemented in the patch) is to remove the "overview" sub-headings in these cases so that the overview text is at the right level. Yes, it's a "hanging paragraph" in ISO speak, but strictly speaking, these introductory sentences should not be shown as normative text anyway. Do we want to find a typographical convention other than "[ Note: ... ]" that keeps such texts readable, but makes them non-normative? Maybe some indent? Or a non-serif font? Both? (Oh, we wanted a non-serif font for the grammar.)

source/diagnostics.tex
```diff
@@ -939,7 +939,7 @@
 Classes may be derived from \tcode{error_category} to support
 categories of errors in addition to those defined in this document.
 Such classes shall behave as specified in this
-subclause. \begin{note} \tcode{error_category} objects are
+subclause~\ref{syserr.errcat}. \begin{note} \tcode{error_category} objects are
```
zygoloid: Use `\iref` for this inline reference.

source/lib-intro.tex
```diff
@@ -1324,10 +1324,8 @@
 
 \rSec2[using]{Using the library}
 
-\rSec3[using.overview]{Overview}
-
 \pnum
-This section describes how a \Cpp program gains access to the facilities of the
+This subclause describes how a \Cpp program gains access to the facilities of the
```
zygoloid: How about: keep the `Overview` subclause, and use "Subclause \ref{using} describes" here.

source/macros.tex
```diff
@@ -62,7 +62,7 @@
 
 %%--------------------------------------------------
 %% Sectioning macros.
-% Each section has a depth, an automatically generated section
+% Each subclause has a depth, an automatically generated subclause
```
zygoloid: I don't think this change is right. :)

jensmaurer: There didn't show up in diffpdf.

source/lib-intro.tex
jensmaurer: Fixed.

source/xrefdelta.tex
```diff
@@ -36,5 +36,10 @@
   \secref{fs.path.generic},
   \secref{fs.race.behavior}}
 
+% Separate sections for introductory paragraphs were removed.
+\removedxref{using.overview}
+\removedxref{constraints.overview}
+\removedxref{conforming.overview}
```
zygoloid: I think this is now redundant.

jensmaurer: Indeed. Fixed.

jensmaurer:
Change to [xrefdelta] reverted, as advised.

PR #1602, opened 2017-04-06
Remove \tcode from 'const object'.

### LINKED ISSUE #469: pick formatting for const object and apply it consistently
The definition of the term _const object_ (and friends) does not use teletype font for the "const", and nor do nearby uses of the term. However, other uses of this term in the standard **do** use a teletype font. We should pick one or the other and do it consistently. I'm not sure we **can** use italic teletype, so perhaps that forces us to use the non-teletype font everywhere.

(This same thing happens in a bunch of other places, for instance the terms 'constexpr constructor' and 'return statement'.)

jensmaurer:
Fixes #469.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -486,7 +486,7 @@ class X {
 The \tcode{mutable} specifier on a class data member nullifies a
 \tcode{const} specifier applied to the containing class object and
 permits modification of the mutable class member even though the rest of
-the object is \tcode{const}\iref{dcl.type.cv}.
+the object is const\iref{dcl.type.cv}.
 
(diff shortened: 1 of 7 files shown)


Eelis:
There are also some occurrences here that you could fix: http://eel.is/c++draft/temp.deduct#3.example-2

source/declarations.tex
```diff
@@ -1154,9 +1154,9 @@
 \end{note}
 
 \pnum
-\indextext{const object@\tcode{const}-object!undefined change to}%
+\indextext{const object@const-object!undefined change to}%
```
zygoloid: How about just `\indextext{const object!undefined change to}%`?

PR #1608, opened 2017-04-08
[temp.deduct.type] Use \cv instead of \grammarterm{cv-list}.

Eelis:
See https://github.com/cplusplus/draft/blob/master/source/declarators.tex#L141


jensmaurer:
Fine with me.

Eelis:
Actually, I'm starting to suspect that maybe \cv should be used here

Eelis:
diff

PR #1610, opened 2017-04-11
Fix header synopses

dgrunwald:
Fix a few "compiler errors" in header synopses:
* missing semicolons in strings.tex
* parameter type in `long double hypot(...)`
* missing `class` in `template<class charT> T4 setfill(charT c);`


jwakely:
Looks good to me.

tkoeppe:
@dgrunwald: Thanks, that's nice. Could you please update the commit messages to start with the section reference? Have a look at what our commit messages usually look like.

PR #1612, opened 2017-04-12
[intro.races] Clarify 'synchronization operation'.

### LINKED ISSUE #1611: [intro.races] CWG 2297: Unclear specification of atomic operations
> It is not sufficiently clear that the only atomic operations are the ones defined in clause 32 [atomics] by the library. The intent is that no accesses are atomic unless the Standard describes them as such.
> 
> An additional problem is that, e.g., new and delete are defined to be synchronization operations, but they are not defined in Clauses 32 [atomics] and 33 [thread].

CWG telecon 2017-04-10 determined this issue was editorial.

jensmaurer:
Fixes #1611 and CWG 2297.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -4655,17 +4655,27 @@ location\iref{intro.memory} and the other one reads or modifies the same
 memory location.
 
 \pnum
-The library defines a number of atomic operations\iref{atomics} and
-operations on mutexes\iref{thread} that are specially identified as
-synchronization operations. These operations play a special role in making
-assignments in one thread visible to another. A synchronization operation on one
-or more memory locations is either a consume operation, an acquire operation, a
-release operation, or both an acquire and release operation. A synchronization
-operation without an associated memory location is a fence and can be either an
-acquire fence, a release fence, or both an acquire and release fence. In
-addition, there are relaxed atomic operations, which are not synchronization
-operations, and atomic read-modify-write operations, which have special
-characteristics. \begin{note} For example, a call that acquires a mutex will
+All modifications to a particular atomic object \placeholder{M} occur in some
+particular total order, called the \defn{modification order} of \placeholder{M}.
+\begin{note} There is a separate order for each
+atomic object. There is no requirement that these can be combined into a single
+total order for all objects. In general this will be impossible since different
+threads may observe modifications to different objects in inconsistent orders.
+\end{note}
+
+\pnum
+Certain library calls \defn{synchronize with} other library calls performed by
+another thread.
+Such calls are also known as
+\defnx{synchronization operations}{synchronization operation}.
+\begin{note} Except in the specified cases, reading a later value does not
+necessarily ensure visibility as described below. Such a requirement would
+sometimes interfere with efficient implementation. \end{note} \begin{note} The
+specifications of the synchronization operations
+(\ref{support.signal}, \ref{util.smartptr.atomic}, \ref{atomics.order},
+\ref{atomics.fences}, \ref{thread}) define when one reads the value
+written by another. For atomic objects, the definition is clear.
+A call that acquires a mutex will
 perform an acquire operation on the locations comprising the mutex.
 Correspondingly, a call that releases the same mutex will perform a release
 operation on those same locations. Informally, performing a release operation on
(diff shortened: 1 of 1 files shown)


source/intro.tex
```diff
@@ -1177,9 +1177,11 @@
 memory location.
 
 \pnum
-The library defines a number of atomic operations (Clause~\ref{atomics}) and
-operations on mutexes (Clause~\ref{thread}) that are specially identified as
-synchronization operations. These operations play a special role in making
+The library identifies certain operations as
+\defnx{synchronization operations}{synchronization operation}
+(\ref{support.signal}, \ref{util.smartptr.shared.atomic}, \ref{atomics.order},
+\ref{atomics.fences}, Clause~\ref{thread}).
```
zygoloid: We should probably mention the *Synchronization:* element used in library descriptions here.

[support.signal] doesn't seem to define any synchronization operations.

jensmaurer: regarding [support.signal]: "A call to the function signal synchronizes with any resulting invocation of the signal handler so installed."

jensmaurer:
Rebased and addressed the review comment, although we do say "synchronizes with" outside of that library element, too.

jensmaurer:
Rebased.

tkoeppe:
Needs rebasing.

jensmaurer:
LWG should decide how they tag synchronization operations: via a "Synchronizes" element and/or by uttering "synchronizes with" and/or via something else.

source/basic.tex
```diff
@@ -4519,9 +4519,12 @@
 memory location.
 
 \pnum
-The library defines a number of atomic operations\iref{atomics} and
-operations on mutexes\iref{thread} that are specially identified as
-synchronization operations. These operations play a special role in making
+The library identifies certain operations as
+\defnx{synchronization operations}{synchronization operation}, as specified
+in the \sync element
```
jwakely: I don't see a \sync element in most of the named subclauses.

source/basic.tex
```diff
@@ -4624,35 +4624,37 @@
...
+\begin{note} There is a separate order for each
+atomic object. There is no requirement that these can be combined into a single
+total order for all objects. In general this will be impossible since different
+threads may observe modifications to different objects in inconsistent orders.
+\end{note}
+
+\pnum
+Certain library calls \defn{synchronize with} other library calls performed by
+another thread
+(\ref{support.signal}, \ref{util.smartptr.atomic}, \ref{atomics.order},
```
geoffromer: I'd prefer the list of sections to be more clearly non-normative.

jensmaurer: Hm... I could move it down into the note after "The specifications of the synchronization operations", but I'd actually prefer to keep the list here, right at the top. A missed cross-reference is at most an editorial bug.

(Also, the library clauses should make sure they have a cross-reference to this section whenever they mention "synchronize with". This would then be a clear indicator that we mean the normative cross-thread meaning of "synchronize with". Similar to what we did with odr-use in the core section.)

geoffromer: Are there other places where the standard, at the point where it defines a term, provides cross-references to uses of the term? I can't recall any, and it doesn't seem like a good practice- it would be hard to maintain, and the existence of the list would create the impression that the list is exhaustive (and thereby creating confusion around any use of "synchronize with" that's omitted from the list).

I'm fine with having uses of this term cross-reference the definition, of course, just not vice-versa.

I'd also be OK with making this list explicitly non-exhaustive, e.g. by starting it with "such as", although I still wouldn't understand why it's there.

jensmaurer: @geoffromer: The SFINAE rule lists the cases where it applies in a non-normative note [temp.deduct] p11.
I've moved the list of cross-references to the note; it still seems useful to the reader to have some cross-references to give an impression what those synchronization operations might be.

geoffromer: Looks good.

jensmaurer:
@zygoloid: I think we can move forward here now.

zygoloid:
I'm sorry, after staring at this wording change for a while and trying to think through the consequences, I can't convince myself that it preserves the intention -- in particular, that the initial struck text does not need to be retained, and that identifying "synchronization operations" with operations that "synchronize with" is correct (also, what happens if there is nothing for an operation to synchronize with? is that a synchronization operation still, or not?).

So while I think this may well be correct, and certainly seems like an improvement, I don't see a path to making this change editorially. (Perhaps while you're there, you could also suggest a fix for the "For atomic objects, the definition is clear." sentence, which serves no purpose other than to infuriate the readers for whom the definition is /not/ clear?)

jensmaurer:
@zygoloid: Synchronization operations always "synchronize with" something else, otherwise they couldn't contribute to the happens-before relationship [intro.races] p9.
So, is this due for a paper for SG1 and CWG? Or just a core issue where SG1 gives a nod?

jensmaurer:
This is also covered by LWG 2506.

jensmaurer:
Discussion on LWG 2506 in SG1 asked for a paper, which is well beyond the scope of an editorial pull request.

PR #1613, opened 2017-04-15
[streambuf.virt.put] Fix logic in overflow() effects specification.

Eelis:
The problem was introduced by commit 0b640ed8.


tkoeppe:
Thanks!

PR #1616, opened 2017-04-18
[dcl.init] A declarator does not have an initializer.

### LINKED ISSUE #1615: [dcl.init] should this 'declarator' be the 'initializer'8
Hi, I got confused when reading [dcl.init]
> A declarator can specify an initial value for the identifier being declared.

Should it be like

An **initializer** can specify an initial value for the identifier being declared.

Noticed that the BNF reflecting the relationship between a **declarator** and an **initializer**, 
```
\begin{bnf}
\nontermdef{init-declarator}\br
    declarator initializer\opt
\end{bnf}
```

FYI, @sdutoit, thanks!

jensmaurer:
Fixes #1615.

### DIFF
--- a/source/declarators.tex
+++ b/source/declarators.tex
@@ -2378,15 +2378,11 @@ the type of the \grammarterm{id-expression} \tcode{y} is ``\tcode{const volatile
 \indextext{initialization|(}
 
 \pnum
-A declarator can specify an initial value for the
-identifier being declared.
-The identifier designates a variable being initialized.
-The process of initialization described in the
-remainder of~\ref{dcl.init}
-applies also to initializations
-specified by other syntactic contexts, such as the initialization
-of function parameters\iref{expr.call} or
-the initialization of return values\iref{stmt.return}.
+The process of initialization described in this subclause applies to
+all initializations regardless of syntactic context, including the
+initialization of a function parameter\iref{expr.call}, the
+initialization of a return value\iref{stmt.return}, or when an
+initializer follows a declarator.
 
 \begin{bnf}
 \nontermdef{initializer}\br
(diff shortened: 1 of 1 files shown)


source/declarators.tex
```diff
@@ -2388,11 +2388,11 @@
 \indextext{initialization|(}
 
 \pnum
-A declarator can specify an initial value for the
+An \grammarterm{init-declarator} can specify an initial value for the
```
zygoloid: If we're going to switch from the English term to the grammar term here, I think we should use "*init-declarator* or *member-declarator*". We would also need to do something about *condition*s, which don't explicitly say that the variable is initialized from its *brace-or-equal-initializer*.

Maybe it would be better to say something like:

> A *declarator* can be followed by an *initializer* that specifies an initial value for the entity being declared. [Note: In some contexts, only the *brace-or-equal-initializer* form of *initializer* is permitted by the grammar, but these rules still apply.]

zygoloid: Hmm. It might also be better to put some distance between the initialization rules, which generally have an initializer and some target entity, and the special case of initializing in a declaration. Perhaps this paragraph should start by talking about initialization in general, and then say that one of the places this happens is when a declarator is followed by an initializer (and other places include initialization of function parameters from arguments and initialization of return values).

jensmaurer: Agreed. [dcl.init] started out at explaining the syntax for initialization, and is only ever so slowly morphing into a general description of all kinds of initialization, where the invoking syntax is secondary. The introductory paragraph follows that anti-pattern quite nicely.

Heuristack: > A declarator can be followed by an initializer that specifies an initial value for the entity being declared.

This is description is correct, however I think we should emphasize on the 'initializer' which is the topic of this section [dcl.init]

If we look at the introductory description of [dcl.decl] p1 
> The init-declarator-list appearing in a declaration is a comma-separated sequence of 
declarators, each of which can have an initializer.

It is easy to understand [dcl.init] was intended for "initializer".

jensmaurer:
Here's a larger rewrite that moves into the "all initializations are equal" direction.

tkoeppe:
@jensmaurer: Please rebase. (In fact, please rebase all your pending PRs.)

jensmaurer:
Rebased.

PR #1617, opened 2017-04-19
[dcl.init] It is the 'initializer' who specifies an initial value for…

Heuristack:
… its 'preceding declarator' (#1615)

### DIFF
--- a/source/declarators.tex
+++ b/source/declarators.tex
@@ -2388,7 +2388,7 @@ the type of the \grammarterm{id-expression} \tcode{y} is ``\tcode{const volatile
 \indextext{initialization|(}
 
 \pnum
-A declarator can specify an initial value for the
+An initializer can specify an initial value for the
 identifier being declared.
 The identifier designates a variable being initialized.
 The process of initialization described in the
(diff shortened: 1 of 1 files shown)


source/declarators.tex
```diff
@@ -2388,7 +2388,7 @@
 \indextext{initialization|(}
 
 \pnum
-A declarator can specify an initial value for the
+An initializer can specify an initial value for the
```
tkoeppe: Why "can"? Is it possible for an initializer to *not* specify an initial value?

jensmaurer:
Note the comments from @zygoloid in my competing pull request #1616. We certainly should not merge the present pull request as-is.

source/declarators.tex
Heuristack: My understanding here is "can" means "can be used", but obviously it is not clear enough.

Heuristack:
Hi @jensmaurer, my proposal is from the perspective of respecting the original text and author, and minimize the change and maximize consistency. However, since you have done a larger rewrite, I will close it when your change is merged.

But without considering a rewrite, I believe the my proposed change is a proper one. Thanks!

zygoloid:
#1616 merged, which also addresses this.

PR #1628, opened 2017-05-16
[any.nonmembers] fix LWG2769 by merging with #1220

lichray:


jwakely:
I'm not sure what I'm looking at here. Are you proposing a change to the DR? This doesn't seem to be editorial.

lichray:
It is editorial, `ValueType` is already a dangling reference in this paragraph (has been changed to `T` in the signature.

jwakely:
FTR 71f89fde12a0a6973d28b66aa52967a79b60c063 caused this regression.

PR #1637, opened 2017-05-26
Support XeTeX.

lichray:
graphicx and hyperref can detect pdf engines by themselves.
microtype font expansion doesn't work with xetex yet.

### DIFF
--- a/source/std.tex
+++ b/source/std.tex
@@ -30,9 +30,9 @@
 \usepackage{xspace}
 \usepackage{lmodern}
 \usepackage[T1]{fontenc}
-\usepackage[pdftex, final]{graphicx}
-\usepackage[pdftex,
-            pdftitle={C++ International Standard},
+\usepackage{ifxetex}
+\usepackage[final]{graphicx}
+\usepackage[pdftitle={C++ International Standard},
             pdfsubject={C++ International Standard},
             pdfcreator={Richard Smith},
             bookmarks=true,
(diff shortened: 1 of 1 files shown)


tkoeppe:
Why do we need to support XeTeX?

lichray:
Just for convenience, for people who for some reason don't have pdfTeX installed.  XeTeX won't be used to build the official PDF because of the lack of microtype support (for now).

zygoloid:
Ideally, there should be only one canonical build process to get a PDF from our sources. If the standard editors do not consistently agree on one set of build tools, then the finer-grained editing work that we do (hyphenation hints and the like) will not be done with a consistent view of the resulting document.

If you would like to make a case for switching fully to XeTeX, that's something I would consider, but only if it either improves our resulting document or somehow improves our editorial workflow without regressing the resulting document. (I'm aware that XeTeX has less sophisticated microtypography support than pdfLaTeX right now, but I don't know if that makes a material impact on the beauty of the resulting document.)

tkoeppe:
Even though I love XeTeX personally for its font handling, I've always found that that feature isn't useful when you need to do maths typesetting. We have complex typographic needs, and we need a maths font, maths symbol set, and a monotype font that all work together with the body font. When you go switching the body font on a whim, you end up with a much worse result where nothing matches. So XeTeX is great for traditional pure-prose book typesetting, but for technical material with complex typography I'm less convinced. If the motivation for XeTeX was to allow better font technologies, then such a proposal should come together with a harmonizing collection of fonts with a suitable license and a maths symbol set.

If the motivation for XeTeX is something else, please let me know. We could also reasonably consider some other typesetters like Context if it buys us something mission-critical, such as improved vertical control. But there would need to be a real benefit.

lichray:
I merely found that it's convenient to generate a readable doc with less dependencies...

jensmaurer:
Closed as agreed.

PR #1640, opened 2017-06-01
For 'signal-safe', refer to [support.signal] (21.10.4), not [csignal.syn] (21.10.3).

Eelis:
(Because [support.signal] is where 'signal-safe' is defined.)


tkoeppe:
Looks good, thanks!

PR #1653, opened 2017-07-06
Fix spelling mistake in 'if constexpr' wording

lewissbaker:
instantation -> instantiation


tkoeppe:
Thanks! The typo was introduced in 188ec4ad389572f270fb65777d6db2bbf7ef8c46, pasted from the wording in http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0292r2.html.

PR #1656, opened 2017-07-10
[allocator.requirements] Fix pointer_to expression

### LINKED ISSUE #1592: LWG 2260 and 'r'
The P/R of [LWG2260](https://wg21.link/lwg2260) (adopted in Kona '17) was written when the descriptive variable definitions (Table 27 in N4140) included a variable called [`r`](https://timsong-cpp.github.io/cppwp/n4140/allocator.requirements#tab:desc.var.def).

Then https://github.com/cplusplus/draft/commit/d2afc9a9b06e8fc495d94506620e54c0b62b36fc editorially removed that variable as unused, but the P/R of the issue was never updated. So now the allocator requirements table contains a dangling reference.

jwakely:
Fixes #1592

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -1872,6 +1872,8 @@ and \tcode{std::forward}, respectively.
 by calling \tcode{a1.allocate}, where \tcode{a1 == a}   \\ \rowsep
 \tcode{q}       &   a value of type \tcode{XX::const_pointer}
 obtained by conversion from a value \tcode{p}.          \\ \rowsep
+\tcode{r}       &   a value of type \tcode{T\&}
+obtained by the expression \tcode{*p}.                  \\ \rowsep
 \tcode{w}       &   a value of type \tcode{XX::void_pointer} obtained by
   conversion from a value \tcode{p}  \\ \rowsep
 \tcode{x}       &   a value of type \tcode{XX::const_void_pointer} obtained by
@@ -1960,7 +1962,7 @@ type (possibly \tcode{const}) \tcode{std::nullptr_t}. \\ \rowsep
 
 \tcode{pointer_traits<\brk{}X::pointer\brk{}>::pointer_to(r)}  &
   \tcode{X::pointer}        &
-                            & \\ \rowsep
+  same as \tcode{p}         & \\ \rowsep
 
 \tcode{a.allocate(n)}   &   \tcode{X::pointer}  &
 Memory is allocated for \tcode{n} objects of type \tcode{T} but objects


jensmaurer:
Looks good to me.

tkoeppe:
Go for it.

source/lib-intro.tex
```diff
@@ -1891,9 +1891,9 @@
   \tcode{X::const_pointer}                &
   \tcode{static_cast<} \tcode{X::const_pointer\brk{}>(x) == q} & \\ \rowsep
 
-\tcode{pointer_traits<\brk{}X::pointer\brk{}>::pointer_to(r)}  &
+\tcode{pointer_traits<\brk{}X::pointer\brk{}>::pointer_to(*p)}  &
   \tcode{X::pointer}        &
-                            & \\ \rowsep
+  same as \tcode{p}         & \\ \rowsep
```
zygoloid: This looks correct, but is this a novel normative requirement? (Or does this follow from existing wording somehow?)

source/lib-intro.tex
```diff
@@ -1891,9 +1891,9 @@
   \tcode{X::const_pointer}                &
   \tcode{static_cast<} \tcode{X::const_pointer\brk{}>(x) == q} & \\ \rowsep
 
-\tcode{pointer_traits<\brk{}X::pointer\brk{}>::pointer_to(r)}  &
+\tcode{pointer_traits<\brk{}X::pointer\brk{}>::pointer_to(*p)}  &
```
zygoloid: I'm a little worried that this can be interpreted as *only* applying to a construct of this form (ie, not applying to the case of `pointer_traits<X::pointer>::pointer_to(r)` for an arbitrary lvalue `r` of the right type). Eg, when we place requirements on `*r++` in the input iterator requirements, the user is not permitted to draw conclusions from that about the expression `r++` (without the `*`).

source/lib-intro.tex
jwakely: `If `pointer_to` is not the inverse of `*p` then `pointer_to` is unusable for the purpose it was designed for. Formally, `pointer_to(*p)` returns a pointer to `*p`, and I can't imagine a sane pointer-like type where that pointer isn't equal to `p`. Maybe it should be stated in terms of equality, rather than "same as".

Or we could simply define `r` before the table.

source/lib-intro.tex
jwakely: OK, we could introduce `r` before the table, as being the object that `*p` yields, so we can keep the requirement that `pointer_to(r) == p`.

jwakely: Or I could create an LWG issue to require the `pointer_to(r) == p` equialty if you're worried it's novel.

jwakely: d2afc9a9b06e8fc495d94506620e54c0b62b36fc shows the original definition of `r` that was removed.

source/lib-intro.tex
zygoloid: It sounds like you're saying this is a novel normative requirement, but is one that was intended to be required and omitted, right? If you say this does not need LWG review, I'll take your word for that, but my inclination would be to run this past LWG.

jwakely: I'll take it to LWG.

jwakely: I raised this on the lib reflector and @CaseyCarter cleverly pointed out:

> Per [allocator.requirements]/5, an allocator's pointer types must meet the random access iterator requirements: "`X​::​pointer` and `X​::​const_­pointer` shall also satisfy the requirements for a random access iterator and of a contiguous iterator." Random access iterators must satisfy the forward iterator requirements. Per [forward.iterators]/6, two forward iterators that denote the same object must be equal: "If `a` and `b` are both dereferenceable, then `a == b` if and only if `*a` and `*b` are bound to the same object." Per [pointer.traits.functions]/2, `pointer_to` returns a pointer that denotes its argument: "The first member function returns a pointer to `r` obtained by calling `Ptr​::​pointer_­to(r)` through which indirection is valid; ... ".

> So: given a fancy pointer `p` of type `T` that denotes an object `o`, `std::pointer_traits<T>::pointer_to(*p) == p`.

> The trail is a bit subtle. Perhaps a note would help?

So it's not a novel requirement, it's already there if you look hard enough. The part I was missing was "`a == b` if and only if `*a` and `*b` are bound to the same object" which forbids fancy pointers from having other "salient attributes" that determine equality, only the identity of the pointee determines equality.

source/lib-intro.tex
zygoloid: I see. So even the old formulation would have carried the implication that only `*p` can be used here. Let's reintroduce `r` to retain the modicum of indirection between the syntax here and the `*p`.

source/lib-intro.tex
zygoloid: OK, if this is not a novel requirement, it sounds like explicitly listing it here would improve clarity without changing intent. Looks good.

jwakely:
New patch coming soon ...

jwakely:
Rebased and modified to address review comments from @zygoloid

PR #1657, opened 2017-07-10
[stringbuf.virtuals] Rephrase DR 453 resolution for seekoff

jwakely:
This avoids the requirement to compare newoff to zero when the value of
newoff hasn't been determined.


jensmaurer:
Looks good to me.

PR #1661, opened 2017-07-15
Move reference to LIA-1 into a new section 'Bibliography'

tkoeppe:


tkoeppe:
It looks a bit silly at the moment:

image


@hubert-reinterpretcast: What do you think?

PR #1665, opened 2017-07-16
Replace \term+\indextext pairs with \defn(x).

Eelis:
No visual difference.

Makes some progress on #329.

### DIFF
--- a/source/declarators.tex
+++ b/source/declarators.tex
@@ -673,12 +673,10 @@ is ill-formed.
 
 
 \pnum
-\indextext{lvalue reference}%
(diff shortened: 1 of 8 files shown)


tkoeppe:
@zygoloid: This would be nice to get into 17. What do you think?

zygoloid:
I don't want to take anything on the c++17 branch that doesn't fix an important problem at this point. This pull request results in no visible changes, so I don't see a reason to take it for C++17.

tkoeppe:
OK, sure. Milestoned.

PR #1668, opened 2017-07-16
[re.traits] Use \placeholdernc for a placeholder.

Eelis:
Slightly changes whitespace:
diff


tkoeppe:
Or use `\placeholdernc` to avoid the space change?

Eelis:
Ah, thanks, amended. No visible differences now.

PR #1678, opened 2017-07-21
[Editorial] Clarify UTF-8 character literal definition

dhouck:
In [lex.ccon], the definition of UTF-8 character literals almost parallels the definition of `char16_t` literals, but there are a few differences.

One of the differences was that the definition of UTF-8 character literals was missing a few words, making a later pronoun lack an antecedent which made sense.

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -1136,11 +1136,11 @@ begins with \tcode{u8}, such as \tcode{u8'w'},
 \indextext{prefix!\idxcode{u8}}%
 is a character literal of type \tcode{char},
 known as a \defn{UTF-8 character literal}.
-The value of a UTF-8 character literal
+The value of a UTF-8 character literal containing a single \grammarterm{c-char}
 is equal to its ISO 10646 code point value,
 provided that the code point value
-is representable with a single UTF-8 code unit
-(that is, provided it is in the C0 Controls and Basic Latin Unicode block).
+is representable with a single UTF-8 code unit.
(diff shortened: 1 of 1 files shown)


source/lex.tex
```diff
@@ -1136,11 +1136,11 @@
...
 is a character literal of type \tcode{char},
 known as a \defn{UTF-8 character literal}.
-The value of a UTF-8 character literal
+The value of a UTF-8 character literal containing a single \grammarterm{c-char}
 is equal to its ISO 10646 code point value,
 provided that the code point value
-is representable with a single UTF-8 code unit
-(that is, provided it is in the C0 Controls and Basic Latin Unicode block).
+is representable with a single UTF-8 code unit.
+(That is, provided it is in the C0 Controls and Basic Latin Unicode block.)
```
zygoloid: I would prefer to change the `char16_t` wording to incorporate the parenthetical in the prior sentence rather than changing this wording to split it into a separate sentence.

dhouck: I would have too, but I chose this way because the `char16_t` wording has been around for longer. I can change it to the other version for both if other people prefer it too.

I didn’t see in the [How to submit an editorial issue](https://github.com/cplusplus/draft/wiki/How-to-submit-an-editorial-issue) page whether you’d prefer an amended commit or a separate one; I would guess amended, and that I should also rebase while I’m at it?

zygoloid: Yes, please amend and rebase this PR.

tkoeppe:
Please reopen if you're still interested in this and have addressed the above comment.

PR #1683, opened 2017-07-24
[over.call.object] Reference postfix-expression in call syntax correctly

hubert-reinterpretcast:
The grammar production corresponding to the call syntax has _postfix-expression_ as the callee operand. [over.call.object] appears to refer to this operand as the _primary-expression_ in the syntax.

### DIFF
--- a/source/overloading.tex
+++ b/source/overloading.tex
@@ -795,7 +795,7 @@ the call is ill-formed.
 
 \pnum
 If the
-\grammarterm{primary-expression}
+\grammarterm{postfix-expression}
 \tcode{E}
 in the function call syntax evaluates
 to a class object of type ``\cv{}~\tcode{T}'',


tkoeppe:
Please fix to make tests pass.

hubert-reinterpretcast:
Only the check-whitespace test was failing, and it seemed that the fix was already pending. In any case, I incorporated the fix in my branch.

tkoeppe:
Yes, the fix is to rebase on master.

hubert-reinterpretcast:
Okay; I don't recall having tried force pushing to a branch with a pull request active before. First time for everything.

tkoeppe:
Oh boy, get used to force pushing to PR branches all the time as you start interacting with the OCD nitpick master editors over here :-)

hubert-reinterpretcast:
Is that (OCD nitpick)+(master) editors, (OCD)+(nitpick master) editors, ((OCD nitpick) master) editors, something else, all, or some? :-)

tkoeppe:
Absolutely all of the above!

PR #1687, opened 2017-07-28
[class.derived] Add cross-reference to [class.inhctor.init]

### LINKED ISSUE #1681: [class.derived] has confusing cross-reference for inheriting constructors
[class.derived] p2 says:

> Constructors of a base class can also be inherited as described in 10.3.3.

But 10.3.3 is only about the syntax, it says nothing much about inheriting constructors. Could we also have a cross-reference to 15.6.3 [class.inhctor.init], even if only in a note?
> Constructors of a base class can also be inherited as described in 10.3.3. _[Note:_ But don't look there, you're probably trying to find 15.6.3 instead. _--end note]_

jensmaurer:
Fixes #1681.

### DIFF
--- a/source/derived.tex
+++ b/source/derived.tex
@@ -82,7 +82,8 @@ considered to be members of the derived class.
 Members of a base class other than constructors are said to be
 \defnx{inherited}{inheritance}
 by the derived class. Constructors of a base class
-can also be inherited as described in~\ref{namespace.udecl}.
+can also be inherited as described in~\ref{namespace.udecl}
+and~\ref{class.inhctor.init}.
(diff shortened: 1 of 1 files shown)


zygoloid:
[class.inhctor.init] doesn't immediately seem relevant to me. This wording is saying "the members of a base class act like members of the derived class by default, except for constructors; for constructors, [namespace.udecl] says how to get that effect". [class.inhctor.init] says nothing about that; it talks about how initialization is performed when a base class constructor is selected, which is irrelevant to what this wording is describing.

jensmaurer:
So, let's cancel this.

PR #1688, opened 2017-07-28
[std] Allow linebreaks before \ref in select places

jensmaurer:
Avoids 'Overfull \hbox' warnings.

Partially reverts commit bf363db2cd687cac2efb5d4ed161cf0583b93f19.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -1091,7 +1091,7 @@ If a name is in scope and is not hidden it is said to be \defn{visible}.%
 \pnum
 The name lookup rules apply uniformly to all names (including
 \grammarterm{typedef-name}{s}~(\ref{dcl.typedef}),
-\grammarterm{namespace-name}{s}~(\ref{basic.namespace}), and
+\grammarterm{namespace-name}{s} (\ref{basic.namespace}), and
 \grammarterm{class-name}{s}~(\ref{class.name})) wherever the grammar allows
 such names in the context discussed by a particular rule. Name lookup
 associates the use of a name with a set of declarations~(\ref{basic.def}) of
@@ -2508,7 +2508,7 @@ namespaces. \end{example}
 \indextext{termination!program}%
 Terminating the program
 without leaving the current block (e.g., by calling the function
-\tcode{std::exit(int)}~(\ref{support.start.term})) does not destroy any
+\tcode{std::exit(int)} (\ref{support.start.term})) does not destroy any
 objects with automatic storage duration~(\ref{class.dtor}). If
 \tcode{std::exit} is called to end a program during the destruction of
 an object with static or thread storage duration, the program has undefined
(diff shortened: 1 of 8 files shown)


zygoloid:
If we're going to convert the ~s to spaces whenever they cause overfull hboxes, why do we even bother with them at all? If we want to instruct TeX that line breaks should be inserted here as a last resort (as a kind of hyphenation rule), we should do so directly rather than manually maintaining some ~s and some spaces.

Perhaps we can define ` (\ref` to expand to `\nolinebreak[2] (\ref` or similar to automate this? But defining space as a macro seems pretty scary to me.

jensmaurer:
First, we only want to allow a linebreak before `(\ref` if all else fails. Second, there's always the option to actually rephrase the source text so that the "overfull \hbox" vanishes because the \ref is in the middle of the line, not at the end. But that doesn't work for all / most cases we're discussing here.

Can we maybe redefine `~` to mean `\nolinebreak[2]` or something like that?

zygoloid:
I have a somewhat more drastic approach in mind: adding a `\iref` command for inline references, to be used as "frobs the grunkifier\iref{grunk.frob}", where `\iref` expands to ` (\ref{grunk.frob})`.

To make that work, we need to change the \ref label for clauses and annexes to include the leading "Clause" or "Annex", but that seems to be a net improvement.

zygoloid:
OK, I have an approach that prevents linebreaks before a parenthesized reference unless doing so would result in a hyphenation or overfull hbox.

zygoloid:
See pull request #1690, which addresses this in a more holistic fashion. Unless there are objections, I intend to rebase and merge that after the LWG motions.

tkoeppe:
Regardless of whether we'll change the way we do references, I would suggest waiting with fixes like the present one until we publish the IS. The status quo should be the "persistent baseline", and changes like the present one should be "temporary finalization fixes" for publication.

jensmaurer:
@tkoeppe: You comment confuses me. I thought C++17 IS was branched and is essentially immutable except for ISO Central Secretariat and NB "must fix" comments? I thought we're only discussing C++20 changes here.

tkoeppe:
@jensmaurer: I'm talking about the C++20 IS!

jensmaurer:
@tkoeppe: Sorry, but "I would suggest waiting with fixes like the present one until we publish the IS" is still three *years* away if you mean publication of the C++20 IS. We can certainly fix \ref vs. \iref in that timeframe, I hope?

jensmaurer:
@tkoeppe: Ah, I understand now: You're saying my "Overfull \hbox removal" patch should not go in until we're near publication. (And we live with the Overfull \hboxes for three years!?) However, something like Richard's #1690 fixes the issue at its root, so is very well on-topic for short-term merge.

tkoeppe:
Correct, we shouldn't make ad-hoc adjustments that would become undesirable once the ambient text changes so as to make the adjustment unnecessary. A principled change like Richard's is fine, but any deviations that we make from the general rule should probably wait until the bulk of the text is stabilized. Otherwise it'd be very difficult to remember to *undo* the deviations.

I don't mind living with occasional overfull boxes in the working paper. If there's an obvious improvement that fixes them, of course we should do that, but in a case like this I could probably just wait.

tkoeppe:
I believe this PR is now obsolete; we don't have any overfull hboxes at head.

PR #1690, opened 2017-07-28
[everywhere] Rework how we mark up inline references.

zygoloid:
Visual changes:

 * The "Clause" and "Annex" in "Clause N" and "Annex N" are now part of
   the hyperlink.
 * Fewer words are hyphenated (we prefer to move an inline reference
   across a line break rather than hyphenating a word)
 * Some spurious "Clause"s and "Annex"es are gone and some missing ones
   are now present.

LaTeX-level changes:

Instead of "some text~(\ref{blah})", we now use "some text\iref{blah}".
Instead of "Clause~\ref{blah}", we now use simply "\ref{blah}", and the
"Clause" prefix is automatically added (likewise for "Annex").

For multi-reference parentheticals, use "some text (\ref{a}, \ref{b})".
A longer orphaned parenthetical is not really a big deal.

~s should now generally not be used near \ref's.

### DIFF
--- a/source/access.tex
+++ b/source/access.tex
@@ -69,8 +69,8 @@ referred to from declarations or expressions.
 \begin{note}
 Access control applies to names nominated by
 \tcode{friend}
-declarations~(\ref{class.friend}) and
-\grammarterm{using-declaration}{s}~(\ref{namespace.udecl}).
+declarations\iref{class.friend} and
+\grammarterm{using-declaration}{s}\iref{namespace.udecl}.
 \end{note}
 In the case of overloaded function names, access control is applied to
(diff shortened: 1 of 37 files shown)


tkoeppe:
Do the ISO drafting directives allow the word "Annex" in a reference? I think I have a similar pull request open where this question needs to be resolved.

zygoloid:
"Annex" is expected prior to a reference to an annex, per Directives Part 2 (7th ed), 6.1, and examples throughout the directives.

tkoeppe:
This has already been applied as beb88157cc49f76677fc467db8ec4a523dbc41d0.

PR #1692, opened 2017-07-29
[dcl.decl, temp.deduct.type] Fix inappropriate uses of \grammarterm.

Eelis:
No visible change.

### DIFF
--- a/source/declarators.tex
+++ b/source/declarators.tex
@@ -2069,7 +2069,7 @@ void print(int a, int) {
 
 \pnum
 In the \grammarterm{function-body}, a
-\grammarterm{function-local predefined variable} denotes a block-scope object of static
+\defnx{function-local predefined variable}{variable!function-local predefined} denotes a block-scope object of static
 storage duration that is implicitly defined (see~\ref{basic.scope.block}).
 
 \pnum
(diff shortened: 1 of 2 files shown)


source/declarators.tex
```diff
@@ -2105,16 +2105,16 @@
 
 \pnum
 In the \grammarterm{function-body}, a
-\grammarterm{function-local predefined variable} denotes a block-scope object of static
+\techterm{function-local predefined variable} denotes a block-scope object of static
```
zygoloid: Seems like `\defn` would be a better choice here.

source/declarators.tex
```diff
@@ -2139,7 +2139,7 @@
 \grammarterm{function-body}
 is of the form
 \tcode{= default ;}
-is called an \grammarterm{explicitly-defaulted} definition.
+is called an \techterm{explicitly-defaulted} definition.
```
zygoloid: Change this to `\defnx{explicitly-defaulted}{definition!function!explicitly-defaulted}`, and remove the `\indextext` above.

source/declarators.tex
Eelis: How would you feel about `\defnx{function-local predefined variable}{variable!function-local predefined}`?

zygoloid: Sounds good to me.

Eelis:
Rebased and amended.

PR #1703, opened 2017-08-12
[diff.cpp14.library] Add Annex C entry for new headers in C++17

### LINKED ISSUE #1700: [diff.cpp14.library] missing Annex C note for new C++17 headers
[diff.cpp03.library] has a note for new headers added in C++11, and [diff.cpp11.library] has a note for the one new header added in C++14. We do not appear to have a similar note for the headers added in C++17 `<any>`, `<optional>`, `<variant>`, `<filesystem>`, `<execution>`, `<memory_resource>`, maybe `<charconv>` (or should that be handled as C++20)?

timsong-cpp:
Fixes #1700.

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -1690,6 +1690,22 @@ and might not perform stack unwinding prior to such a call.
 
 \rSec2[diff.cpp14.library]{\ref{library}: library introduction}
 
+\ref{headers}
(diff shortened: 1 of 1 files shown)


tkoeppe:
@zygoloid: Possibly for inclusion in C++17?

PR #1707, opened 2017-08-15
[futures.async] remove parens from DECAY_COPY()

jwakely:
We're referring to the function name, not trying to call it with no arguments.

### DIFF
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -4838,7 +4838,7 @@ If \tcode{launch::async} is set in \tcode{policy}, calls
 \tcode{\placeholdernc{DECAY_COPY}(std::forward<Args>(args))...)}
 (\ref{func.require}, \ref{thread.thread.constr})
 as if in a new thread of execution represented by a \tcode{thread} object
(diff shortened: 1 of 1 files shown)


source/threads.tex
```diff
@@ -4838,7 +4838,7 @@
 \tcode{\placeholdernc{DECAY_COPY}(std::forward<Args>(args))...)}
 (\ref{func.require}, \ref{thread.thread.constr})
 as if in a new thread of execution represented by a \tcode{thread} object
-with the calls to \tcode{\placeholdernc{DECAY_COPY}()} being evaluated in the thread that called \tcode{async}.
+with the calls to \tcode{\placeholdernc{DECAY_COPY}} being evaluated in the thread that called \tcode{async}.
```
tkoeppe: @jwakely: Then get rid of the "nc", too. That was only needed for the spacing relative to the parens.

tkoeppe:
I would welcome this change.

zygoloid:
Merged to c++17 in 1b93dccf9aa16772e385cf76681289be54a9095e.

PR #1708, opened 2017-08-16
[intro.races] Add index entries for notions of coherence.

Eelis:
I realize these notions are only mentioned in notes, but I think it would be useful if people could find the relevant paragraphs if they searched the index for "coherence".

screenshot


jensmaurer:
Looks good to me.

PR #1736, opened 2017-09-07
[alg.reverse] remove duplicated Requires: element

### LINKED ISSUE #1726: [C++17 DIS comment 014] [alg.reverse] reverse has two Requires: clauses
Figure out what happened here: how did we end up with two Requires: clauses that say the same thing in different ways? Then fix.

jwakely:
Fixes ISO 2 (C++17 DIS)

Fixes #1726

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -3201,7 +3201,8 @@ template<class ExecutionPolicy, class BidirectionalIterator>
 \begin{itemdescr}
 \pnum
 \requires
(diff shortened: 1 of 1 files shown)


jwakely:
I've re-pushed this as a change against master.

It will need to be cherry-picked to the `c++17` branch to fix #1726 (which will have a merge conflict due to the `\ref` -> `\iref` changes).

Issue #1714, opened 2017-09-06
[C++17 DIS comment 002] update [intro.defs] to use text that can be substituted for the use of the term


zygoloid:
Per ISO directives, definitions should be drafted to replace the term in context. We need to survey the existing definitions and rephrase to suit this requirement.

zygoloid:
First pass at tackling this is here:

https://github.com/cplusplus/draft/compare/master...zygoloid:iso-2?expand=1

I've split this into a multiple changes for easier review, and because I think we may want to adopt only some of them. Specifically, I would propose taking 886b293 and d9b5f40e for C++17, but not a38a1311.

Thoughts?

jwakely:
886b2932850b088cbd77a0baeebf4a2b32be45c7 addresses the comment nicely.

d9b5f40e67f703fcb5f5fb44ac5cbf4f628282cb is great. (On the same subject, are we even allowed to have a second set of defns in lib-intro.tex? Those should also be in Clause 3, or not use the **Terms and definitions** format, no? Not something to fix for C++17 though).

tkoeppe:
Both changes look very nice!

jensmaurer:
886b293 just highlights some of our more dodgy definitions, e.g. "argument of throw expression" (it's probably an operand these days that _throw-expression_ is a first-class expression in clause 8 [expr]).

jensmaurer:
@jwakely regarding the "on the same subject note": I agree. Further, maybe we should review clause 3 afresh: It's not clear to me which definitions should go there as opposed to being inline. Something like "access", which is essentially only used in the memory model subsection is in Clause 3, but other more global definitions aren't, it seems.

zygoloid:
I've been contemplating whether to suggest we start to issue the standard in two parts (language and library). If we do that, each part gets to have its own "Terms and Definitions". Seems like something to talk over with at least Herb before making changes to the Terms and Definitions in Clause 20.

jensmaurer:
@zygoloid: We do have interdependencies between language and library; in particular the language refers to the library for some of its semantics (cf. `operator new` or Gor's coroutines) and (of course) the library fully depends on the language. That essentially requires synchronized releases of library and language, without a clear hierarchy between the two, unlike some other multipart standards where part -1 is "common terms and defintions", part -2 specifies the geometry of a screw, and part -3 specifies measuring apparatus / approaches to determine whether a given screw satisfies part -2.

zygoloid:
[...]

PR #1738, opened 2017-09-07
[intro.defs, definitions, fs.definitions] change notes in Terms and Definitions to say "Note X to entry"

### LINKED ISSUE #1715: [C++17 DIS comment 003] update [intro.defs] to use "Note to entry"
Proposed: ACCEPT, but use "[ *Note N to entry:* blah blah -- *end note* ]" as formatting style for consistency

tkoeppe:
Fixes ISO 3 (C++17 DIS)

Fixes #1715.

### DIFF
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -137,9 +137,12 @@ it can continue execution past the blocking operation
 
 \indexdefn{behavior!conditionally-supported}%
 \definition{conditionally-supported}{defns.cond.supp}
-program construct that an implementation is not required to support\\
-\begin{note} Each implementation documents all conditionally-supported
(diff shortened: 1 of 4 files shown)


jwakely:
I wonder if we should actually stop formatting [fs.definitions] like [definitions], since we aren't meant to have two sets of definitions, and they're not written as proper definitions (or terms that should apply across any ISO standard).

zygoloid:
Regarding [fs.definitions], I agree; I've made that change for ISO comment 2 (issue #1714).

zygoloid:
Rebased and merged to master as 6b0c835 and to c++17 as 6341a28.

PR #1741, opened 2017-09-08
Reword [over.ics.rank]/3.2.3

BRevzin:
The wording here is awkward. The intent is, I think, to prefer binding an rvalue to an rvalue reference than to an lvalue reference, but the phrase "to an rvalue" is attached to the rvalue reference part (to which it is redundant). Would be clearer to attach it to the lvalue reference part.

### DIFF
--- a/source/overloading.tex
+++ b/source/overloading.tex
@@ -2516,8 +2516,8 @@ or, if not that,
 \item \tcode{S1} and \tcode{S2} are reference bindings\iref{dcl.init.ref} and
 neither refers to an implicit object parameter of a non-static member function
 declared without a \grammarterm{ref-qualifier},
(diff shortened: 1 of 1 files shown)


zygoloid:
An rvalue reference can also bind to a function lvalue, so the first part of the change has normative impact. And I'm not comfortable making the assumption editorially that if one conversion sequence binds to an rvalue then the other conversion sequence does (with the diverse ways in which overload resolution is used, I'm not confident that's true in general).

=> We need CWG feedback on this.

tkoeppe:
@BRevzin: While you're waiting, feel free to rebase and fix the failing tests.

tkoeppe:
@BRevzin: ping

tkoeppe:
@BRevzin: ping

PR #1745, opened 2017-09-11
[temp.over.link] Change leading letter in comments in example to lowercase.

zygoloid:
Fixes NB JP 6 (C++17 DIS)


tkoeppe:
Looks good.

PR #1746, opened 2017-09-12
Add Foreword.

zygoloid:
Fixes ISO 16 (C++17 DIS)

### DIFF
--- a/source/preface.tex
+++ b/source/preface.tex
@@ -1,0 +1,42 @@
+%!TEX root = std.tex
+\chapter{Foreword}
+
+ISO (the International Organization for Standardization) and IEC (the
+International Electrotechnical Commission) form the specialized system for
+worldwide standardization. National bodies that are members of ISO or IEC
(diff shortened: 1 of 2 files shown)


source/preface.tex
```diff
@@ -0,0 +1,42 @@
...
+ISO (the International Organization for Standardization) and IEC (the
+International Electrotechnical Commission) form the specialized system for
+worldwide standardization. National bodies that are members of ISO or IEC
+participate in the development of International Standards through technical
+committees established by the respective organization to deal with particular
+fields of technical activity. ISO and IEC technical committees collaborate in
+fields of mutual interest. Other international organizations, governmental and
+non-governmental, in liaison with ISO and IEC, also take part in the work. In
+the field of information technology, ISO and IEC have established a joint
+technical committee, ISO/IEC JTC 1.\\
```
tkoeppe: Why these double linebreaks rather than just normal paragraph breaks?

PR #1747, opened 2017-09-12
Convert "must"s in normative wording that do not indicate logical

zygoloid:
necessity to a different form.

Normative requirements use "shall". Advice to users uses "should".
Description of the behavior of the implementation that does not of
itself constitute a requirement uses the imperative mood.

Fixes ISO 21 (C++17 DIS)

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -3737,7 +3737,8 @@ parameter (i.e., has the same constant value or the same type
 as the template parameter) can be used in place of that
 template parameter in a reference to the current
 instantiation. In the case of a non-type template argument,
-the argument must have been given the value of the
+this only applies if
+the argument was given the value of the
 template parameter and not an expression in which the
 template parameter appears as a subexpression.
 \begin{example}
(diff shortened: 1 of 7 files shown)


tkoeppe:
Check that whitespace!

source/templates.tex
```diff
@@ -583,18 +583,17 @@
   id-expression
 \end{bnf}
 
-\begin{note}
```
tkoeppe: Is this change relevant to the subject at hand?

source/templates.tex
```diff
@@ -3736,10 +3735,13 @@
 parameter (i.e., has the same constant value or the same type
 as the template parameter) can be used in place of that
 template parameter in a reference to the current
-instantiation. In the case of a non-type template argument,
+instantiation.
+\begin{note}
+In the case of a non-type template argument,
```
jensmaurer: We've discussed this recently for core issue 2307. Why is it safe to turn the normative text into a note here?

source/templates.tex
zygoloid: I think so. The below "name must be known to refer to a template" is supposed to be what normatively states what is and is not a *template-name*, but is hopelessly vague, and changing the "must" to a "shall" is also wrong -- the right thing is to say that an *identifier* is a *template-name* if name lookup associates it with one, which is what the note prior to the broken sentence says.

That said, I'd be happy to take a less-invasive alternate path through here.

source/templates.tex
zygoloid: `N` is not equivalent to, say, `(N)` per [temp.over.link]p5.

I'm happy to make a different change if you'd prefer. Perhaps something more in this direction:

> A template argument that is equivalent to a template parameter (i.e., has the same type as the template parameter or, *for a non-type template argument, is an expression that names the template parameter but not as a subexpression*) can be used in place of that template parameter in a reference to the current instantiation.

... but that's pretty awkward. Or perhaps:

> A template argument that is equivalent to a template parameter (i.e., has the same constant value or the same type as the template parameter) can be used in place of that template parameter in a reference to the current instantiation. In the case of a non-type template argument, this *only applies if the argument was* given the value of the template parameter and not an expression in which the template parameter appears as a subexpression.

jensmaurer: Well, we don't really mean the "equivalent" from [temp.over.link] here: the example claims that my_I and I are "equivalent", but they are not the same per the ODR (i.e. token-by-token identical). So, we need the normative text to explain the different notion of "equivalent". [The very fact we're even discussing this makes me uneasy about such a large-ish last-minute change.]

That said, your second proposal above seems to be least invasive, so I'd prefer that.

source/templates.tex
zygoloid: Let's just leave this one alone for C++17. I'm now concerned that this sentence may be intended to have normative impact (eg, what else would make `sizeof &operator""_foo < 8 > 0` parse as a comparison rather than as having a *template-argument-list*?)

source/templates.tex
zygoloid: Done.

source/templates.tex
zygoloid: Done.

PR #1748, opened 2017-09-14
Replace "this International Standard" with "this document" when

zygoloid:
referring to the document as a body of text.

Cases where that phrasing is used to compare this standard to other
revisions of the C++ standard retain this phrasing for clarity.

Fixes ISO 20 (C++17 DIS)

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3273,8 +3273,8 @@ The lifetime of a reference ends as if it were a scalar object.
 describes the lifetime of base and member subobjects. \end{note}
 
 \pnum
-The properties ascribed to objects and references throughout this International
-Standard apply for a given object or reference only during its lifetime. \begin{note}
+The properties ascribed to objects and references throughout this document
+apply for a given object or reference only during its lifetime. \begin{note}
 In particular, before the lifetime of an object starts and after its
 lifetime ends there are significant restrictions on the use of the
 object, as described below, in~\ref{class.base.init} and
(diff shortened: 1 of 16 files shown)


source/basic.tex
```diff
@@ -3897,7 +3897,7 @@
 floating-point types}.
 \indextext{floating-point type!implementation-defined}%
 \begin{note}
-This International Standard imposes no requirements on the accuracy of
```
tkoeppe: I feel this change is a regression, but I can't quite say why. I think it's clear to me that a Standard can make impositions, but it sounds odd that a document would do so.

zygoloid: I think this is an improvement, for what it's worth. It clarifies that this document imposes no requirements, but other documents (for instance, other standards that your implementation might follow, such as ISO/IEC 60559, or vendor-specific compiler flags, may impose further requirements). The fact that this document is an International Standard is irrelevant here.

tkoeppe: I suppose that makes sense. It's probably just my lack of familiarity with the phrase, and I'll get used to it. I can see the consistency argument.

source/intro.tex
```diff
@@ -333,34 +333,34 @@
...
-Standard except for those rules containing an explicit notation that
+consists of all syntactic and semantic rules in this document
+except for those rules containing an explicit notation that
 ``no diagnostic is required'' or which are described as resulting in
 ``undefined behavior''.
 
 \pnum
 \indextext{conformance requirements!method of description}%
-Although this International Standard states only requirements on \Cpp
+Although this document states only requirements on \Cpp
```
jensmaurer: I must admit I find changing some of these places to "this document" a regression. (And I disagree with the ISO directives essentially banning "this International Standard".) I full agree with "this document" when it comes to internal organizations, chapters, etc. Once we get into defining requirements (such as above, or in "imposes no requirements on floating-point"), I feel "this Int. Std." makes it clearer that while it's the job of the Standard to define requirements, it opted not to for this particular case. (And other standards may fill the gap.) Just talking about "this document" seems too low-profile to me.
But ISO says we should use "this document", so my opinion is probably irrelevant.

Issue #609, opened 2016-01-28
Missing capture default in example 5.1.2.4


t-lutz:
The last two lines of the example read:

```
int j;
auto x3 = []()->auto&& { return j; }; // OK: return type is int&
```

However `j` is not captured and the lambda doesn't have a default capture. 

The correct code should be:

```
int j;
auto x3 = [&]()->auto&& { return j; }; // OK: return type is int&
           ^
```

cpplearner:
I think `j` should be interpreted as having namespace scope, so it needn't (and can't) be captured.

t-lutz:
My bad., I wrongly reproduced the example with `j` in a local scope.

PR #1749, opened 2017-09-15
8.1.5.4 Add capture to the j variant to make the example to be compiled

anatoly-spb:
Here is  a small fix to make the 8.1.5.4 example to be compiled.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -710,7 +710,7 @@ type specified by the
 auto x1 = [](int i){ return i; };     // OK: return type is \tcode{int}
 auto x2 = []{ return { 1, 2 }; };     // error: deducing return type from \grammarterm{braced-init-list}
 int j;
-auto x3 = []()->auto&& { return j; }; // OK: return type is \tcode{int\&}
+auto x3 = [&j]()->auto&& { return j; }; // OK: return type is \tcode{int\&}
 \end{codeblock}
 \end{example}
 


cpplearner:
See #609

tkoeppe:
@cpplearner: Thanks for the reference.

Closing, since there is nothing wrong with the example.

berkus:
Since you got two reports on this where people think the capture is erroneously missing - it makes sense to add a comment to this snippet explaining the expected `j` scope.

tkoeppe:
The standard is not a tutorial?

@zygoloid: what do you think?

timsong-cpp:
[expr.prim.lambda]'s examples are all over the place w/r/t scope. Many are scope-neutral, so to speak, but we do have stuff like [this](https://timsong-cpp.github.io/cppwp/expr#prim.lambda.capture-13.example-1) that only works at block scope, stuff like the one at issue here that only works at namespace scope, and then stuff like [this](https://timsong-cpp.github.io/cppwp/expr#prim.lambda.closure-7.example-1) that actually needs to be split up and put into two different scopes to compile. Meanwhile, we also have examples like [this](https://timsong-cpp.github.io/cppwp/expr#prim.lambda.capture-5.example-1) that introduces a function solely to get a block scope so that captures can be used. It might not be a bad idea to come up with some rule on scopes for examples and apply it uniformly.

In this particular case, I would argue that the example could use some improvement; there's no explicit marker that the declarations are at namespace scope, and I'd wager that local lambdas arise more frequently than global ones, so the "missing capture" gets distracting because it resembles a common programming error with local lambdas. But I dislike an explicit comment too, as that also distracts from the point of the example, which has nothing to do with scope or captures. 

How about just adding a `static` before `int j;`?

PR #1756, opened 2017-09-26
support LuaLaTeX

Patterner:
`LuaLaTeX `does not know `\pdfshellescape` and `\pdffilesize`
Using `ifluatex`, `pdftexcmds` and some `\defs` fixes that.

(similar reasoning as the XeTeX PR #1637 but smaller implementation)

### DIFF
--- a/source/std.tex
+++ b/source/std.tex
@@ -5,6 +5,15 @@
 %% basics
 \documentclass[a4paper,10pt,oneside,openany,final]{memoir}
 
+\usepackage{ifluatex}
+\ifluatex
+\RequirePackage{pdftexcmds}
+\makeatletter
+\let\pdfshellescape\pdf@shellescape
+\let\pdffilesize\pdf@filesize
+\makeatother
+\fi
+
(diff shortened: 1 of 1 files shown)


tkoeppe:
Are there plans to merge eTeX and LuaTeX, or is eTeX going away in any shape or form? Same question as for the other PR, why do we need to support this?

tkoeppe:
@cor3ntin We already have an issue for other tex engines.

tkoeppe:
I had a quick look around, and found this:

- pdftex is stable and well supported
- luatex is quite a departure in API and would require some adjustments (does this PR still work?)
- xetex might be a better alternative; it has essentially the same API as pdftex
- xetex and luatex's support for microtype isn't as good as pdftex's (this will cause the document to look different)
- to support other scripts we would have to change fronts from Latin Modern to New Computer Modern; a detailed analysis of the differences between and the quality of the two font sets would be welcome (e.g. kerning, diacritic placement, hinting), as well as a look at their glyph sets

PR #1758, opened 2017-09-28
[unique.ptr, util.smartptr.shared] Remove redundant repetitions of declarations that already appear in the synopsis

tkoeppe:


source/utilities.tex
```diff
@@ -9186,94 +9131,6 @@
 
   template<class T> shared_ptr(weak_ptr<T>) -> shared_ptr<T>;
   template<class T, class D> shared_ptr(unique_ptr<T, D>) -> shared_ptr<T>;
```
zygoloid: We should figure out what to do with these. My first inclination is to move them from here to the header synopsis, because they're separate top-level declarations (just like `make_shared` etc). But I think that is an independent cleanup.

jwakely: But arguably explicit deduction guides are closer to constructors, in that they tell you which forms of initialization would support CTAD. So having them next to the class synopsis (where the constructors appear) has some benefit.

But ... the standard is not a tutorial.

tkoeppe: At the moment, *all* deduction guides appear only with the class definitions, and *never* in the header synopses. As jwakely says, the deduction guide kind of only makes sense in combination with the constructor overload set, so I'd say that's not unreasonable.

PR #1761, opened 2017-10-06
Remove \xspace everywhere except for \opt

jensmaurer:
As a first step towards `\xspace` removal (see #1433), these changes remove `\xspace` everywhere except for `\opt`.

(The consensus for \opt was to change its use from `xxx\opt`  to `\opt{xxx}`.)

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -4,11 +4,11 @@
 \gramSec[gram.basic]{Basic concepts}
 
 \pnum
-\begin{note} This Clause presents the basic concepts of the \Cpp language.
+\begin{note} This Clause presents the basic concepts of the \Cpp{} language.
 It explains the difference between an object and a
 name and how they relate to the value categories for expressions.
 It introduces the concepts of a
-declaration and a definition and presents \Cpp's
+declaration and a definition and presents \Cpp{}'s
 notion of type, scope, linkage, and
 storage duration. The mechanisms for starting and
 terminating a program are discussed. Finally, this Clause presents the
@@ -215,7 +215,7 @@ using N::d;                     // declares \tcode{d}
 \pnum
 \begin{note}
 \indextext{implementation-generated}%
-In some circumstances, \Cpp implementations implicitly define the
+In some circumstances, \Cpp{} implementations implicitly define the
 default constructor\iref{class.ctor},
 copy constructor\iref{class.copy},
 move constructor\iref{class.copy},
@@ -2540,7 +2540,7 @@ a major array bound\iref{dcl.array}. A violation of this rule on type
 identity does not require a diagnostic.
 
 \pnum
-\begin{note} Linkage to non-\Cpp declarations can be achieved using a
+\begin{note} Linkage to non-\Cpp{} declarations can be achieved using a
 \grammarterm{linkage-specification}\iref{dcl.link}. \end{note}%
 \indextext{linkage|)}
 
@@ -2550,7 +2550,7 @@ identity does not require a diagnostic.
 
 \pnum
 \indextext{memory model|(}%
-The fundamental storage unit in the \Cpp memory model is the
+The fundamental storage unit in the \Cpp{} memory model is the
 \defn{byte}.
 A byte is at least large enough to contain any member of the basic
 \indextext{character set!basic execution}%
@@ -2562,7 +2562,7 @@ bits,\footnote{The number of bits in a byte is reported by the macro
 the number of which is \impldef{bits in a byte}. The least
 significant bit is called the \defn{low-order bit}; the most
 significant bit is called the \defn{high-order bit}. The memory
-available to a \Cpp program consists of one or more sequences of
+available to a \Cpp{} program consists of one or more sequences of
 contiguous bytes. Every byte has a unique address.
 
 \pnum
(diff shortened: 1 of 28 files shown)


source/basic.tex
```diff
@@ -4,11 +4,11 @@
 \gramSec[gram.basic]{Basic concepts}
 
 \pnum
-\begin{note} This Clause presents the basic concepts of the \Cpp language.
+\begin{note} This Clause presents the basic concepts of the \Cpp{} language.
 It explains the difference between an object and a
 name and how they relate to the value categories for expressions.
 It introduces the concepts of a
-declaration and a definition and presents \Cpp's
+declaration and a definition and presents \Cpp{}'s
```
godbyk: Just FYI: I don't think the braces are required here (unless you want to prevent kerning).

godbyk:
In general, if a macro is followed by a punctuation character (or anything other than a space you want to keep), you don't need to use terminating `{}` unless you want to prevent kerning.

source/basic.tex
jensmaurer: The idea is to have a general rule that all \Cpp macros are followed by {}. Whether that's necessary in a particular situation should not matter.

source/macros.tex
```diff
@@ -217,12 +217,12 @@
...
 \newcommand{\noteoutro}[1]{\textit{\,---\,end #1}\,]}
-\newenvironment{note}[1][Note]{\noteintro{#1}}{\noteoutro{note}\xspace}
-\newenvironment{example}[1][Example]{\noteintro{#1}}{\noteoutro{example}\xspace}
+\newenvironment{note}[1][Note]{\noteintro{#1}}{\noteoutro{note}\space}
+\newenvironment{example}[1][Example]{\noteintro{#1}}{\noteoutro{example}\space}
 
 %% Library function descriptions
-\newcommand{\Fundescx}[1]{\textit{#1}\xspace}
-\newcommand{\Fundesc}[1]{\Fundescx{#1:}}
+\newcommand{\Fundescx}[1]{\textit{#1}}
```
godbyk: This is a separate issue, but is there any reason to keep `\Fundescx` around?

jensmaurer: It doesn't seem to be used, except for \Fundesc.

source/macros.tex
```diff
@@ -317,13 +318,13 @@
 
 %% Miscellaneous
 \newcommand{\uniquens}{\placeholdernc{unique}}
-\newcommand{\stage}[1]{\item{\textbf{Stage #1:}}\xspace}
-\newcommand{\doccite}[1]{\textit{#1}\xspace}
+\newcommand{\stage}[1]{\item{\textbf{Stage #1:}}}
```
godbyk: Separate issue: Should use square brackets: `\item[Stage #1:]` and adjust formatting of items in `description` environments. (Note that the `description` environment is also used in `preprocessor.tex` in a weird way, too.)

source/macros.tex
```diff
@@ -317,13 +318,13 @@
...
+\newcommand{\doccite}[1]{\textit{#1}}
 \newcommand{\cvqual}[1]{\textit{#1}}
 \newcommand{\cv}{\cvqual{cv}}
-\renewcommand{\emph}[1]{\textit{#1}\xspace}
-\newcommand{\numconst}[1]{\textsl{#1}\xspace}
-\newcommand{\logop}[1]{{\footnotesize #1}\xspace}
+\renewcommand{\emph}[1]{\textit{#1}}
```
godbyk: Separate issue: If you just want `\emph` to have the same effect as `\textit` (_i.e._, the text is always italic regardless of the context), then you could just write `\let\emph\textit`.

source/support.tex
```diff
@@ -1879,9 +1879,9 @@
 the behavior is undefined.
 
 \newcommand{\replaceabledesc}[1]{%
```
godbyk: Separate issue: Should this definition be moved to `macros.tex`?

source/basic.tex
godbyk: The only downsides are the adverse impact on readability of the source and (potentially) kerning.

I doubt that kerning is much of a problem here. And giving up a little bit of readability to gain consistency (for those editors who are less familiar with TeX's parsing rules) sounds like a reasonable thing to do.

(We could also add a lint check to warn in suspicious uses of `\Cpp` without the terminating `{}`.)

jensmaurer:
Rebased; should be ready for merge now.

jensmaurer:
Rebased again, fixed overfull hbox.

source/basic.tex
tkoeppe: This change does in fact not cause any visual diff.

@godbyk: Yes, sure, that sounds useful!

source/macros.tex
tkoeppe: Isn't that the default definition of `\emph` anyway?

godbyk: `\emph` alternates between italics and roman type as its nested, while `\textit` does not.

For example, `\emph{Hello, \emph{world}!}` would be typeset as “<i>Hello,</i> world<i>!</i>” In contrast, `\textit{Hello, \textit{world}!}` would be typeset as “<i>Hello, world!</i>”

tkoeppe: Ah yes, of course. Luckily we aren't quite that nestedly empathetic in the standard. :-)

tkoeppe: In fact, it's shocking how little we \emphasize in the standard compared to the time we spend talking about it!

source/macros.tex
tkoeppe: Done, thanks!

PR #1765, opened 2017-10-13
[fs.filesystem.syn] fix ordering of exists and equivalent

jwakely:


jensmaurer:
(This makes the order in the header synopsis consistent with the order of specification later in the section.)

Looks good to me.

jwakely:
The specification used to be wrong too, but I fixed that in 1a9d0120502097c64660da7c20e40c9d5ee392c5

The reason for changing them is that every other one of the filesystem operations is in alphabetical order. There's no reason for this one to differ, and it makes it (slightly) harder to find if you're expecting them to be in alphabetical order.

PR #1773, opened 2017-10-17
[utilities, depr] Replace "member typedef" with "nested type" 

cpplearner:
[class.mem]/3:

> Nested types are classes (12.1, 12.2.5) and enumerations (10.2) declared in the class and arbitrary types declared as members by use of a typedef declaration (10.1.3) or _alias-declaration_.

In most cases there's no normative change because the member must name a pre-existing type.

`aligned_storage` and `aligned_union` are special-cased because their member `type` may name a new type. In particular, `aligned_storage::type` is defined as a nested class type in both libstdc++ and libc++.

### DIFF
--- a/source/future.tex
+++ b/source/future.tex
@@ -2102,7 +2102,7 @@ with a base characteristic of \tcode{true_type}
 if \tcode{T} is a literal type\iref{basic.types}, and
 \tcode{false_type} otherwise.
 The partial specialization \tcode{result_of<Fn(ArgTypes...)>} is a
-\tcode{TransformationTrait} whose member typedef \tcode{type} is defined
+\tcode{TransformationTrait} whose nested type \tcode{type} is defined
 if and only if \tcode{invoke_result<Fn, ArgTypes...>::type} is defined.
 If \tcode{type} is defined, it names the same type as \tcode{invoke_result_t<Fn, ArgTypes...>}.
 
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -16534,7 +16534,7 @@ assert((is_same_v<remove_all_extents_t<int[][3]>, int>));
  The value of \textit{default-alignment} shall be the most
  stringent alignment requirement for any \Cpp object type whose size
  is no greater than \tcode{Len}\iref{basic.types}.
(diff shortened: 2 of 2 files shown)


jensmaurer:
In a phrasing like "the member typedef X names ..." or "the member typedef is T", I think this change makes things worse: We know we can't have a fresh type here, so why use a phrasing that allows this?

I agree the change for aligned_storage, aligned_union is good (it's not talking about a pre-existing type), but it seems borderline editorial. (Although it's probably unobservable whether something is a typedef to something else that cannot otherwise be named or is a new type by itself.)

cpplearner:
"Member typedef" is not a defined term, and does not clearly indicate that the member can also be declared using an alias-declaration.

"Nested type" is already used in e.g. [\[container.insert.return\]/1](http://eel.is/c++draft/container.insert.return#1), [\[move.iterator\]/1](http://eel.is/c++draft/move.iterator#1), [\[depr.func.adaptor.typedefs\]/4](http://eel.is/c++draft/depr.func.adaptor.typedefs#4), and in these cases the types have to be synonyms for pre-existing types (or specializations of a pre-existing template).

jwakely:
I think "the nested type `type` names ..." makes things worse. A typedef names another type, a nested type doesn't necessarily.

> does not clearly indicate that the member can also be declared using an alias-declaration

I disagree, since there's no semantic difference between the two. [dcl.typedef] p2 "A _typedef-name_ can also be introduced by an _alias-declaration_."

"Member typedef" seems clear to me, and we do say "typedef member" in [class.mfct]

jensmaurer:
I'd rather replace "nested type", when talking about synonyms, from [container.insert.return]/1, [move.iterator]/1, [depr.func.adaptor.typedefs]/4 than go the other way.

So, in the interest of progress, this pull request could go forward if it would only propose change 7b2c0db.

For the rest, I agree that "member typedef" might be viewed as excluding alias-declarations (although the difference is unobservable, so it doesn't matter). What about "The member X is a synonym for the type T." as the uniform standard phrasing?

tkoeppe:
Feel free to reopen if and when you'd like to address the above issues.

PR #1774, opened 2017-10-19
[macros] Prevent unfortunate page break

jensmaurer:
by accepting clubs or widows inside itemdescr.

### DIFF
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -396,7 +396,10 @@
 
 \newenvironment{itemdescr}
 {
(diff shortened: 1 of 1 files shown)


jensmaurer:
The only diffpdf-visible difference is that the page break in [util.smartptr.shared.const] that Richard was unhappy about has now been resolved.

tkoeppe:
This PR pessimizes the status quo. I wonder if we should delay tweaks like this until the next IS publication is imminent.

PR #1775, opened 2017-10-19
[lib] Harmonize introductory comments in synopses

### LINKED ISSUE #1709: Minor styles inconsistency in single-line comments
Result of searching "// hash" shows...

```
[vector.syn]
// hash support
[unord.map.overview]
// hash policy:
[unord.set.overview]
// hash policy:
[unord.multimap.overview]
// hash policy
[unord.multiset.overview]
// hash policy:
[thread.thread.id]
// Hash support
```

Well, there are at least 2 problems.

- Which case of letter should be used after `//`?
- When to use trailing colon in a single-line comment?

Not sure if there has been already some rules about styles in comments, esp. in synopses. (Nonetheless, the examples above seem to be unintended.)

jensmaurer:
Even if no section reference is given, make them lowercase
and remove trailing colons.

Fixes #1709.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -3157,7 +3157,7 @@ for operations where there is additional semantic information.
 namespace std {
   template <class T, size_t N>
   struct array {
-    //  types:
+    // types
     using value_type             = T;
     using pointer                = T*;
     using const_pointer          = const T*;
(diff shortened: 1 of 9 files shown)


tkoeppe:
This is good. We should try to further harmonize the comment with the section headings to which it points, but that's a separate step.

PR #1780, opened 2017-10-22
[travis] Make overfull hboxes a presubmit error.

tkoeppe:
Also delete the "manual" check since "complete" subsumes "manual".

@zygoloid, @godbyk: Could you please take a look whether this looks right?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,9 @@
 #
 
 env:
-  - BUILD_TYPE=latexmk   # build using latexmk
+  - BUILD_TYPE=latexmk   # build using latexmk, also check for overfull hboxes
   - BUILD_TYPE=make      # build using Makefile
-  - BUILD_TYPE=manual    # build manually
-  - BUILD_TYPE=complete  # build manually and regenerate figures, grammer, and cross-references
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -3,17 +3,17 @@
...
-  - BUILD_TYPE=complete  # build manually and regenerate figures, grammer, and cross-references
+  - BUILD_TYPE=complete  # build manually and regenerate figures, grammar, and cross-references
   - BUILD_TYPE=check-whitespace  # check for whitespace at the ends of lines
 
 script:
   # Build std.pdf
   - pushd source
   - if [ "$BUILD_TYPE" = "latexmk" ]; then
       latexmk -pdf std;
+      ! grep -Fe "Overfull \hbox" std.log;
```
godbyk: Need to escape the backslash with another backslash.

godbyk:
The `manual` method was originally to just perform the build steps as specified in the README. But `complete` does subsume `manual` in that it performs the additional step of regenerating the images.

.travis.yml
tkoeppe: Done, thanks.

tkoeppe:
OK, thanks, in that case I'd prefer to not rerun redundant steps that we'd catch anyway. No need to waste compute resources and time.

PR #1781, opened 2017-10-24
C++17

Mussabaheen:
Your code should not consume too much memory it will also decrease the speed of your program


tkoeppe:
We don't accept pull requests in to the "c++17" branch. Mostly because that ship has sailed.

Issue #1401, opened 2017-01-23
[any] "contained value" vs "contained object"


lynnboy:
In [optiona] and [variant], "contained value" is the term refer to the object inside, but in [any], in general it uses "contained object", but still use "contained value" at some places.
We should unify the term as "contained value".

jensmaurer:
Also, [variant] doesn't italics-define "contained value", but should, to make the index consistent.

PR #1795, opened 2017-11-08
[any.class] Revert dot to comma typo from bdff8687c

johelegp:


jwakely:
Looks correct to me. This was introduced by the fix for #1401

tkoeppe:
Agreed. Looks like this was a simple copy-paste error in bdff8687c.

Issue #1367, opened 2017-01-16
[diff] Consider assigning paragraph numbers to changes


Eelis:
I think paragraph numbers in the margin would make it visually easier to spot where one change ends and the next begins, and would also make referring to changes a bit easier. Thoughts?

zygoloid:
Direction: add one paragraph number per complete change.

Eelis:
Fixed by #1837.

PR #1803, opened 2017-11-13
[diff] Give numbers to changes.

Eelis:
This follows the direction given in #1367.

Example:

a

### DIFF
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -309,7 +309,7 @@
 
 %% Change descriptions
 \newcommand{\diffdef}[1]{\hfill\break\textbf{#1:}\xspace}
-\newcommand{\change}{\diffdef{Change}}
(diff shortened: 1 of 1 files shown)


jensmaurer:
Well, this is probably as good as we get visually in this otherwise not-really-pleasing annex.

zygoloid:
It seems a bit odd that the section number (which is part of the change description) is before the paragraph number. Putting a paragraph number before the section number might look a bit strange, though. Let's try adding a "*Something:*" before each such section number list, and moving the paragraph numbers before those.

Some possible options for the "Something": "Modified rule", "Affected wording", "See", "Clause or Subclause".

jensmaurer:
I've created #1837 which uses "Affected subclause" and also cleans up a few oddities in section references throughout [diff].

tkoeppe:
We'll probably use #1837 instead.

PR #1805, opened 2017-11-15
Fix some more subscripts for T, also fix cv in a few places.

jensmaurer:


source/utilities.tex
```diff
@@ -936,10 +936,10 @@
 \begin{itemdescr}
 \pnum
 \returns \tcode{pair<V1, V2>(std::forward<T1>(x), std::forward<T2>(y))},
-where \tcode{V1} and \tcode{V2} are determined as follows: Let \tcode{Ui} be
-\tcode{decay_t<Ti>} for each \tcode{Ti}. If \tcode{Ui} is a specialization
-of \tcode{reference_wrapper}, then \tcode{Vi} is \tcode{Ui::type\&},
-otherwise \tcode{Vi} is \tcode{Ui}.
+where \tcode{V1} and \tcode{V2} are determined as follows: Let $\tcode{U}_i$ be
+\tcode{decay_t<$\tcode{T}_i$>} for each $\tcode{T}_i$. If $\tcode{U}_i$ is a specialization
```
tkoeppe: This is odd now, because it says `T1` above on line 938, but here you say `T_1`.

jensmaurer: Agreed, it's odd. However, T1, T2, V1, and V2 are actual template parameters, so I can't really use subscript on those. I guess I should drop that particular change and create a new editorial issue with tag "decision required" so that we can talk about it later.

jensmaurer:
Dropped odd change that needs further discussion.

source/utilities.tex
tkoeppe: Correct, that's why I hadn't made such changes before. If we need the C++ identifier, we should just use that, or otherwise we need to introduce wording like "where T_i is `Ti`" etc. For abstract type lists, subscripting is good.

PR #1806, opened 2017-11-15
[unord.req], [fs.path.io] Fix "Effects: Equivalent to" styles

johelegp:
Some other places that might require change (wording relative to N4700):
- [string.view.modifiers] p2: Maybe convert to code block, but there are others like this.
- [string.view.modifiers] p4: Convert to expression style. Unlike the above, this is unique.
- [mem.poly.allocator.mem] p1: Maybe this was meant to be an _Effects:_ element.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -2509,7 +2509,7 @@ and \tcode{CopyAssignable}.\br
 &   \requires\ If \tcode{t} is a non-const rvalue expression, \tcode{value_type} shall be
     \tcode{MoveInsertable} into \tcode{X}; otherwise, \tcode{value_type} shall be
     \tcode{CopyInsertable} into \tcode{X}.\br
-    \effects\ Equivalent to a.insert(t).  Return value is an iterator pointing
+    \effects Equivalent to \tcode{a.insert(t)}.  Return value is an iterator pointing
 to the element with the key equivalent to that of \tcode{t}.  The
 iterator \tcode{p} is a hint pointing to where the search should
 start.  Implementations are permitted to ignore the hint.%
@@ -2521,8 +2521,8 @@ start.  Implementations are permitted to ignore the hint.%
 \tcode{a.insert(i, j)}
 &   \tcode{void}
 &   \requires\ \tcode{value_type} shall be \tcode{EmplaceConstructible} into \tcode{X} from \tcode{*i}.\br
-    \requires \tcode{i} and \tcode{j} are not iterators in \tcode{a}.
-    Equivalent to \tcode{a.insert(t)} for each element in \tcode{[i,j)}.%
+    \requires \tcode{i} and \tcode{j} are not iterators in \tcode{a}.\br
+    \effects Equivalent to \tcode{a.insert(t)} for each element in \tcode{[i,j)}.%
     \indextext{unordered associative containers!\idxcode{insert}}%
     \indextext{\idxcode{insert}!unordered associative containers}%
 &   Average case \bigoh{N}, where $N$ is \tcode{distance(i, j)}.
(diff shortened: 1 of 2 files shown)


jensmaurer:
This change overlaps with my #1815. We should pick one, but not both.

johelegp:
Here are a few changed lines not included in #1815. Where both PR change a line, the changes are the same, so they don't conflict. If it's still undesirable to have two open PRs with the same goal, you could include those changes in #1815.

source/containers.tex
```diff
@@ -2522,7 +2522,7 @@
 &   \tcode{void}
 &   \requires\ \tcode{value_type} shall be \tcode{EmplaceConstructible} into \tcode{X} from \tcode{*i}.\br
     \requires \tcode{i} and \tcode{j} are not iterators in \tcode{a}.
-    Equivalent to \tcode{a.insert(t)} for each element in \tcode{[i,j)}.%
+    \effects\ Equivalent to \tcode{a.insert(t)} for each element in \tcode{[i,j)}.%
     \indextext{unordered associative containers!\idxcode{insert}}%
```
jensmaurer: I'd prefer leaving container requirements tables alone; they need a serious overhaul anyway, and it's unclear whether \effects and friends should or should not appear there.

jensmaurer:
Well, in other places, #1815 has more changes, because it also folds `\returns` into "`\effects` Equivalent to" where appropriate.

tkoeppe:
Please rebase.

johelegp:
Done.

source/containers.tex
```diff
@@ -2509,7 +2509,7 @@
 &   \requires\ If \tcode{t} is a non-const rvalue expression, \tcode{value_type} shall be
     \tcode{MoveInsertable} into \tcode{X}; otherwise, \tcode{value_type} shall be
     \tcode{CopyInsertable} into \tcode{X}.\br
-    \effects\ Equivalent to a.insert(t).  Return value is an iterator pointing
+    \effects\ Equivalent to \tcode{a.insert(t)}.  Return value is an iterator pointing
```
tkoeppe: Please remove the trailing backslashes after `\effects` etc., as they are not needed.

source/containers.tex
tkoeppe: That's true, but for the time being, I think this change improves local consistency, so I don't mind it.

source/containers.tex
johelegp: There are many occurrences of this for many elements in many sources. Is it okay to make a separate PR for that?

tkoeppe: No, leave the existing document alone. It's just that since you're already touching this part, you might as well clean it up, and I don't want you to introduce new cases of this (in the other change). But don't bother changing other existing instances.

johelegp: Done.

source/containers.tex
```diff
@@ -2522,7 +2522,7 @@
 &   \tcode{void}
 &   \requires\ \tcode{value_type} shall be \tcode{EmplaceConstructible} into \tcode{X} from \tcode{*i}.\br
     \requires \tcode{i} and \tcode{j} are not iterators in \tcode{a}.
-    Equivalent to \tcode{a.insert(t)} for each element in \tcode{[i,j)}.%
+    \effects Equivalent to \tcode{a.insert(t)} for each element in \tcode{[i,j)}.%
```
tkoeppe: Wait, did you test this? You'll probably need a manual  line break of sorts here.

johelegp: No, I didn't. Guess I'll setup a proper latex editing environment with live rendering.

Issue #1807, opened 2017-11-15
rationalize library header name indexing


zygoloid:
Right now, we index library headers something like this:

```
\indextext{\idxhdr{<blah>}}%
\indexlibrary{\idxhdr{<blah>>}}%
```

We should provide a dedicated macro for this (perhaps `\indexhdr{blah}`). Once that's done, we should consider removing these from the index of library names, since they're not library names in the normal sense. (Maybe a separate 1-page index of library headers would be useful, though.)

jensmaurer:
Now moving all library headers from the "index of library names" into a new index. Headers stay in the main index.

PR #1810, opened 2017-11-15
[macros] Introduce \indexhdr and apply it for all headers.

jensmaurer:
Partially addresses #1807.


jensmaurer:
This was sed-converted from the status quo. We can redefine \indexhdr to something else at a later time.

zygoloid:
This will need some more changes once we land the motions (we're adding two new headers), but this looks good, thanks!

jensmaurer:
I've rebased this, fixed mentions of the additional headers, and while doing so, fixed bad index entries <<syncstream>> and <<type_traits>>. (\idxhdr implicitly adds the <>.)

PR #1811, opened 2017-11-15
[over.over] Consistently use 'set of overloaded functions'

### LINKED ISSUE #1691: [over.over]/p4 what "set of selected functions"?
After the addition of concepts, we have the wording "All functions with associated constraints that are not satisfied (17.4.2) are eliminated from the set of selected functions.", but there's no mention of a "set of selected functions" up to this point.

From comments to branch at https://github.com/cplusplus/draft/commit/18b788bde093045d34dff93e2b80d7b7a8aedc76:

burblebee:
> what "set of selected functions"? up to this point, we've only discussed if/whether a function is selected - there is no set. At least the previous text (although still problematic) started with "If more than one function is selected" so that one could extrapolate that the set must be from the selected functions.
> 
> The whole section could use some work for that matter. For example, where did "the overload set" in p1 come from? And in the next sentence, "A function
> template name is considered to name a set of overloaded functions in such contexts." - did we really mean to restrict this to function templates? And, if not in one of these "certain contexts" where a function is "resolved", the name doesn't name a set of overloaded functions?

zygoloid:
> The start of p1 says: "A use of an overloaded function name without arguments is resolved in certain contexts to a function, a pointer to function or a pointer to member function for a specific function from the overload set."
> 
> So this set is the set of functions associated with the "use of an overloaded function name".
> 
> I agree this subsection would benefit from rework, but I don't think this particular change materially affects the clarity in this regard. Please provide a separate editorial issue or pull request if you feel strongly about improving this wording.

I don't feel that the "overload set" mentioned in p1 is sufficient.  I would like to see this "reworked".  Even giving the overload set a name (S?) and referring to it would be a huge improvement and clarify things significantly.

jensmaurer:
Fixes #1691.

### DIFF
--- a/source/overloading.tex
+++ b/source/overloading.tex
@@ -2849,7 +2849,7 @@ ignored\iref{expr.prim}.
 \end{note}
 
 \pnum
-If the name is a function template, template argument deduction is
+For each function template found by name lookup, template argument deduction is
 done\iref{temp.deduct.funcaddr}, and if the argument deduction succeeds,
 the
 resulting template argument list is
@@ -2883,7 +2883,7 @@ described in~\ref{expr.unary.op}.
 All functions with
 associated constraints
 that are not satisfied\iref{temp.constr.decl}
-are eliminated from the set of selected functions.
+are eliminated from the set of overloaded functions.
 If more than one function in the set remains,
 all function template specializations
 in the set


zygoloid:
This change is incorrect. Here's how this section works today:

* The "set of overloaded functions" is the set that we form through name lookup and deduction.
* Then we apply the "A function with type \tcode{F} is selected for the function type \tcode{FT} of the target type required in the context if" rule to select some subset.
* Then we apply the "set of selected functions" rule to pick between the selected functions.

This change effectively moves the second bullet after the third by making the third bullet apply to the "set of overloaded functions" instead of the "set of selected functions".

jensmaurer:
Ok, I was of the mistaken belief that "selected" in the overload resolution context means "single function". I'm now seeing the "selected" state is an intermediate set where additional partial ordering etc. is applied to find the "single chosen one". For example, p3 talks about "is selected" and then requiring an ampersand to form the pointer-to-member, which should probably be the very last check after you've actually selected a function in p4.

It would probably be helpful to move the phrasing for the second bullet, which is in p1, to below p2, where template argument deduction is discussed (which happens before

tkoeppe:
@jensmaurer: can we close this in favour of a CWG issue?

jensmaurer:
Let's discuss in Jacksonville how to proceed here.

jensmaurer:
Parent issue #1691 to become a core issue.

PR #1812, opened 2017-11-15
[unique.ptr] Remove definition of 'transfers ownership'.

### LINKED ISSUE #1643: [unique.ptr]  Consider removing "transfer ownership" definition
All of `unique_ptr::operator=(unique_ptr&&)`, `shared_ptr(shared_ptr&&)`, and `shared_ptr::operator=(shared_ptr&&)` explicitly leave their source with `.get()==nullptr`. `unique_ptr(unique_ptr&&)`'s wording delegates that to the "transfer ownership" definition in [unique.ptr], without explicitly referring to the section of the definition, which caused me to miss that it was defined at all.

As far as I can see, that term is only used twice in the standard, for the `unique_ptr` move constructor and templated move constructor, and I believe its other 2 bullets are repeated by the existing postconditions for those operations. By adding `u.get()==nullptr` to those two sets of postconditions, we might be able to remove the whole definition of "transfer ownership".

jensmaurer:
It is mostly subsumed by the detailed descriptions of the move
constructors and assignment operators.

Fixes #1643.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -7950,23 +7950,6 @@ deleter, but properly disposes of its owned object via the associated
 deleter before such replacement is considered completed.
 
 \pnum
-Additionally, \textit{u} can, upon request, \defn{transfer ownership} to another
-unique pointer \textit{u2}. Upon completion of such a transfer, the following
-postconditions hold:
-
-\begin{itemize}
-\item \textit{u2.p} is equal to the pre-transfer \textit{u.p},
-\item \textit{u.p} is equal to \tcode{nullptr}, and
-\item if the pre-transfer \textit{u.d} maintained state, such state has been
-transferred to \textit{u2.d}.
-\end{itemize}
-
-As in the case of a reset, \textit{u2} properly disposes of its pre-transfer
-owned object via the pre-transfer associated deleter before the ownership
-transfer is considered complete. \begin{note} A deleter's state need never be
-copied, only moved or swapped as ownership is transferred. \end{note}
-
-\pnum
 Each object of a type \tcode{U} instantiated from the \tcode{unique_ptr} template
 specified in this subclause has the strict ownership semantics, specified above,
 of a unique pointer. In partial satisfaction of these semantics, each such \tcode{U}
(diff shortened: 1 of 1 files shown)


zygoloid:
I would like us to consider an alternative: retain the definition of transfer of ownership, and replace the repeated postconditions with

> \postcondition As specified for transfer of ownership from \tcode{foo} to \tcode{bar}\iref{unique.ptr}.

I'm weakly in favor of this alternative, to reduce duplication and give us somewhere central to talk about `unique_ptr`'s ownership semantics. However, if there's some reason the duplicated-postconditions approach is superior, I expect it would not take much to change my mind.

jensmaurer:
So, we would substantially shorten the existing postconditions.

Hm.. The "transfer ownership" wording is imprecise regarding the deleter: It says "if the pre-transfer u.d maintained state, such state has been transferred to u2.d", but the move constructors actually say that we copy-constructor or move-construct the deleter. Oh, move-assignment actually shows code rather than weasely wording.

source/utilities.tex
```diff
@@ -8316,15 +8299,16 @@
 throw an exception.
 
 \pnum
-\effects Constructs a \tcode{unique_ptr} by transferring ownership from
-\tcode{u} to \tcode{*this}. If \tcode{D} is a reference type, this
+\effects Constructs a \tcode{unique_ptr} from
+\tcode{u}. If \tcode{D} is a reference type, this
 deleter is copy constructed from \tcode{u}'s deleter; otherwise, this
 deleter is move constructed from \tcode{u}'s deleter. \begin{note} The
 deleter constructor can be implemented with \tcode{std::forward<D>}. \end{note}
```
zygoloid: (Drive-by:) "The construction of the deleter" would be more correct here.

zygoloid:
OK. The idea of using "transfers ownership" to mean the relevant set of rules doesn't work out, because the rules are different each time (we transfer the deleter in different ways, and may or may not reset the destination first). Let's go with this patch or something like it.

source/utilities.tex
tkoeppe: @jensmaurer: ping

PR #1815, opened 2017-11-16
[lib] Harmonize punctuation of 'Effects: Equivalent to'

### LINKED ISSUE #1119: Style of "equivalent to"
In the library, we have a term of art "equivalent to" to specify effects. What we don't have is a consistent way of formatting it. Currently observed forms:

1. Equivalent to: [codeblock]
1. Equivalent to: `statement;`
1. Equivalent to: `expression;`
1. Equivalent to `expression`.
1. Equivalent to `statement;` where foo is bar.

I think we should definitely have the colon after "to", and we should probably have both a codeblock and an inline form.

jensmaurer:
Fixes #1119.

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -11821,7 +11821,7 @@ template <class InputIterator>
 
 \begin{itemdescr}
 \pnum
-\effects Equivalent to \tcode{return *this += path(first, last)}.
+\effects Equivalent to: \tcode{return *this += path(first, last);}
 \end{itemdescr}
 
 \rSec4[fs.path.modifiers]{\tcode{path} modifiers}
@@ -11982,7 +11982,7 @@ const value_type* c_str() const noexcept;
 
 \begin{itemdescr}
 \pnum
-\returns Equivalent to \tcode{native().c_str()}.
+\effects Equivalent to: \tcode{return native().c_str();}
 \end{itemdescr}
 
 \indexlibrarymember{operator string_type}{path}%
@@ -12590,7 +12590,7 @@ void swap(path& lhs, path& rhs) noexcept;
 
 \begin{itemdescr}
 \pnum
-\effects Equivalent to: \tcode{lhs.swap(rhs);}
+\effects Equivalent to \tcode{lhs.swap(rhs)}.
 \end{itemdescr}
 
 \indexlibrary{\idxcode{hash_value}!\idxcode{path}}%
--- a/source/strings.tex
+++ b/source/strings.tex
@@ -2110,11 +2110,7 @@ basic_string& append(size_type n, charT c);
 
 \begin{itemdescr}
 \pnum
-\effects Equivalent to \tcode{append(basic_string(n, c))}.
-
-\pnum
-\returns
-\tcode{*this}.
+\effects Equivalent to: \tcode{return append(basic_string(n, c));}
 \end{itemdescr}
 
 \indexlibrarymember{append}{basic_string}%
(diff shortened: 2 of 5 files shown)


tkoeppe:
Thanks -- but all editorial changes are frozen until the motions are in. I think we'll get new cases like this coming in, so I will ask you to recheck at that point.

jensmaurer:
See #1806 for a competing pull request.

source/iostreams.tex
```diff
@@ -12823,7 +12823,7 @@
 
 \begin{itemdescr}
 \pnum
-\effects Equivalent to \tcode{pathobject = p},
+\effects Equivalent to: \tcode{pathobject = p},
```
tkoeppe: Do we want colons here? What follows is an expression, not a statement, non?

source/strings.tex
```diff
@@ -2449,7 +2425,7 @@
 
 \begin{itemdescr}
 \pnum
-\effects Equivalent to \tcode{insert(pos, basic_string(n, c))}.
+\effects Equivalent to: \tcode{return insert(pos, basic_string(n, c));}
```
tkoeppe: Now the \returns below is redundant.

source/utilities.tex
```diff
@@ -9439,9 +9439,8 @@
 \end{itemdecl}
 
 \begin{itemdescr}
-\pnum\effects  Equivalent to \tcode{shared_ptr(r).swap(*this)}.
-
-\pnum\returns  \tcode{*this}.
+\pnum\effects  Equivalent to:
+\tcode{shared_ptr(r).swap(*this); return *this;}
```
tkoeppe: We lost the \returns here?

source/utilities.tex
```diff
@@ -9466,10 +9465,8 @@
 
 \begin{itemdescr}
 \pnum
-\effects Equivalent to \tcode{shared_ptr(std::move(r)).swap(*this)}.
-
-\pnum
-\returns \tcode{*this}.
+\effects Equivalent to:
+\tcode{shared_ptr(std::move(r)).swap(*this); return *this;}
```
tkoeppe: Here too.

source/iostreams.tex
jensmaurer: No, it's not just the expression: all of what follows is part of the "Equivalent to". It seemed prudent to add the colon to highlight that fact.

source/utilities.tex
jensmaurer: No. Look at the end of line 9443; there's a second statement, which happens to be a return statement.

source/utilities.tex
jensmaurer: No, see above.

source/iostreams.tex
tkoeppe: Hm, I thought it reads nicer without the colon, like a normal sentence. @zygoloid, a native speaker's opinion?

source/utilities.tex
tkoeppe: Oh eek. Can we make that a `codeblock` then?

source/iostreams.tex
jensmaurer: Why do we add the colon for the "statement" case, then? I thought because "something big" was coming along.

source/utilities.tex
jensmaurer: Or keep the \returns, maybe?

source/iostreams.tex
tkoeppe: I always thought it was because of the grammatical nature of what follows: the colon separates sentences, or "statements", if you will. If you just have an expression, it's naturally an object of the current sentence.

source/utilities.tex
tkoeppe: Yes. I'm not keen on getting overly creative in compressing stuff. Readability is important. The `return *this = that;` is borderline, too (but I was going to let it slide).

jensmaurer: I've undone the changes in this area.

source/strings.tex

source/iostreams.tex
tkoeppe: (I'm happy to merge everything except these two, which I'd like discussed further.)

jensmaurer: I've dropped these two contentious changes.

jensmaurer:
Rebased + contentious change dropped.

PR #1817, opened 2017-11-16
[basic.def] 'class name declaration' is undefined.

### LINKED ISSUE #1641: The term "class name declaration" used in [basic.def]/2 (2.5) is not defined in the Standard
The term "class name declaration" used in [[basic.def]/2 (2.5)](http://eel.is/c++draft/basic.def#2.5) is not defined in the Standard. As a matter of fact, it is not even mentioned a second time in the spec. The term is informally introduced in the second example of [[class.name]/2](http://eel.is/c++draft/class.name#2) as follows:
```
struct s { int a; };
void g() {
    struct s; // hide global struct s with a block-scope declaration
    s* p; // refer to local struct s
    struct s { char* p; }; // define local struct s
    struct s; // redeclaration, has no effect
}
```
where the two declarations `struct s;` above, as far as I can tell, are what the Standard refers to in [basic.def]/2 (2.5) as a "class name declaration". But these declarations are already defined in the Standard as an [_elaborated-type-specifier_](http://eel.is/c++draft/dcl.type.elab#nt:elaborated-type-specifier) , or more specifically, as an _elaborated-type-specifier_ that is the sole constituent of its declaration, as can be seen in the first form shown in paragraph [[dcl.type.elab]/1](http://eel.is/c++draft/dcl.type.elab#1).

jensmaurer:
Fixes #1641.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -112,7 +112,8 @@ A declaration may also have effects including:
 \pnum
 \indextext{declaration!function}%
 \indextext{definition}%
-A declaration is a \defn{definition} unless
+Each entity declared by a \grammarterm{declaration} is
+also \defnx{defined}{define} by that declaration unless:
 \begin{itemize}
 \item
 it declares a function
@@ -137,7 +138,7 @@ and the variable was defined within the class with the \tcode{constexpr}
 specifier (this usage is deprecated; see \ref{depr.static_constexpr}),
 \item
 \indextext{declaration!class name}%
-it is a class name declaration\iref{class.name},
+it is introduced by an \grammarterm{elaborated-type-specifier}\iref{class.name},
 \item
 it is an
 \indextext{declaration!opaque enum}%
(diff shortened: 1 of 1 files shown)


source/basic.tex
```diff
@@ -128,7 +128,7 @@
 specifier (this usage is deprecated; see \ref{depr.static_constexpr}),
 \item
 \indextext{declaration!class name}%
-it is a class name declaration\iref{class.name},
+it is an \grammarterm{elaborated-type-specifier}\iref{class.name},
```
zygoloid: There's a category error here; an *elaborated-type-specifier* is not a *declaration*.

[I think the existing wording was attempting to refer to this portion of [class.name]p2: "A *declaration* consisting of *class-key identifier*`;` [...] introduces the class name into the current scope." but generalizing this to apply to other cases in which *elaborated-type-specifier*s introduce *class-name*s seems like a good idea.]

To some extent the problem is that this whole paragraph is bogus: "A declaration is a definition unless" is one big category error. A declaration declares zero or more entities, defines zero or more entities, and introduces zero or more names. But it's wrong to say that a declaration "is" or "is not" a definition.

Perhaps a different structure would help. For instance:

> Each entity declared by a *declaration* is also *defined* by that declaration unless:
> * [...]
> * the entity is introduced by an *elaborated-type-specifier*,
> * [...]
> A declaration is said to be a *definition* of each entity that it defines.

Or perhaps we need to formally divorce the notion of "*declaration*" (the grammar element) and "declaration" (the portion of a *declaration* that declares a specific entity).

jensmaurer: I like the "different structure" approach, absent the major task of doing the divorce.

jensmaurer:
Update as per @zygoloid desires + rebased.

PR #1819, opened 2017-11-17
Use \tcode for return/if/switch statement.

### LINKED ISSUE #1305: Code font in "return statement"
I count 14 instances of "return statement" and 23 instances of "`return` statement". The is also some variation in "if statement", "switch statement" and maybe others.

jensmaurer:
Fixes #1305.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3984,7 +3984,7 @@ be used only as an expression statement\iref{stmt.expr}, as an operand
 of a comma expression\iref{expr.comma}, as a second or third operand
 of \tcode{?:}\iref{expr.cond}, as the operand of
 \tcode{typeid}, \tcode{noexcept}, or \tcode{decltype}, as
-the expression in a return statement\iref{stmt.return} for a function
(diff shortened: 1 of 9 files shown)


jensmaurer:
Rebased.

source/templates.tex
```diff
@@ -2103,7 +2103,7 @@
 \grammarterm{requires-clause}{s},
 or
 \grammarterm{noexcept-specifier}{s}.
-For the purpose of instantiation, the substatements of a constexpr if
+For the purpose of instantiation, the substatements of a constexpr \tcode{if}
```
zygoloid: This change is incorrect; we define the (non-teletype) term "constexpr if statement" in [stmt.if]p2.

jensmaurer: Reverted (changes to templates.tex are gone; they all dealt with constexpr if statement).

(Similar to "contextually converted to bool", it might be worthwhile to consider changing "constexpr if statement" to teletype, but italics teletype is ugly, and we never finished discussing that aspect. So ignore for now.)

PR #1825, opened 2017-11-19
[structure.summary] Remove obsolete paragraph about Note(s): and Example(s): elements that we no longer provide

tkoeppe:


tkoeppe:
Note(s) removed in 23d739974007c3f827ad2e55da60cb4358681773.

jensmaurer:
What about the "Note:" in table 91 (unordered associative containers), row for `b.cbegin()`? It's not using the macro, but it's pretty close.

tkoeppe:
@jensmaurer: That should just be converted to a normal [Note:.

source/containers.tex
```diff
@@ -2746,8 +2746,10 @@
 \tcode{b.cbegin(n)}
 &   \tcode{const_local_iterator}
 &   \requires \tcode{n} shall be in the range \tcode{[0, b.bucket_count())}.
-    Note: \tcode{[b.cbegin(n), b.cend(n))} is a valid range containing
-    all of the elements in the $\texttt{n}^{\textrm{ th}}$ bucket.%
+    \begin{note}
+    \tcode{[b.cbegin(n), b.cend(n))} is a valid range containing
+    all of the elements in the $\texttt{n}^{\textrm{ th}}$ bucket.
+    \end{note}%
```
zygoloid: Where (if anywhere) is the normative rule justifying this claim? Should we copy the corresponding wording from `b.begin(n)` and `b.end(n)` to `b.cbegin(n)` and `b.cend(n)`, or is there general wording saying how `begin` and `cbegin` relate somewhere?

tkoeppe: For ordinary iterators and begin/end, you could squint at Table 83 to conclude this relation. But we never have *local* iterators anywhere. We don't have equivalent "operational semantics" fields in this table.

It might be safest to repeat the wording from the non-const versions and delete the note. What do you think?

zygoloid: That approach sounds good to me.

tkoeppe: OK, done, PTAL.

PR #1827, opened 2017-11-20
[dcl.enum] Merge duplicate normative paragraphs on redeclarations.

### LINKED ISSUE #1826: [dcl.enum] Duplicate sentences about redeclarations
[\[dcl.enum\]/4](http://eel.is/c++draft/dcl.enum#4.sentence-1):

> If the enum-key is followed by a nested-name-specifier, the enum-specifier shall refer to an enumeration that was previously declared directly in the class or namespace to which the nested-name-specifier refers (i.e., neither inherited nor introduced by a using-declaration), and the enum-specifier shall appear in a namespace enclosing the previous declaration.

Is mostly repeated in [\[dcl.enum\]/12](http://eel.is/c++draft/dcl.enum#12.sentence-1):

> If an enum-head contains a nested-name-specifier, the enum-specifier shall refer to an enumeration that was previously declared directly in the class or namespace to which the nested-name-specifier refers, or in an element of the inline namespace set ([namespace.def]) of that namespace (i.e., not merely inherited or introduced by a using-declaration), and the enum-specifier shall appear in a namespace enclosing the previous declaration.

One of these was probably intended to replace the other. I don't know which.

jensmaurer:
Fixes #1826.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -6166,7 +6166,7 @@ constants. Its name becomes an \grammarterm{enum-name} within its scope.
 
 \begin{bnf}
 \nontermdef{opaque-enum-declaration}\br
-    enum-key \opt{attribute-specifier-seq} \opt{nested-name-specifier} identifier \opt{enum-base} \terminal{;}
+    enum-key \opt{attribute-specifier-seq} enum-head-name \opt{enum-base} \terminal{;}
 \end{bnf}
 
 \begin{bnf}
@@ -6219,7 +6219,7 @@ type. \begin{example}
 
 \end{example}
 \end{note}
-If an \grammarterm{opaque-enum-declaration} contains
+If the \grammarterm{enum-head-name} of an \grammarterm{opaque-enum-declaration} contains
 a \grammarterm{nested-name-specifier},
 the declaration shall be an explicit specialization\iref{temp.expl.spec}.
 
@@ -6232,7 +6232,7 @@ The \grammarterm{enum-key}{s} \tcode{enum class} and
 \tcode{enum struct} are semantically equivalent; an enumeration
 type declared with one of these is a \defnadj{scoped}{enumeration},
 and its \grammarterm{enumerator}{s} are \defnx{scoped enumerators}{enumerator!scoped}.
-The optional \grammarterm{identifier} shall not be omitted in the declaration of a scoped enumeration.
+The optional \grammarterm{enum-head-name} shall not be omitted in the declaration of a scoped enumeration.
 The \grammarterm{type-specifier-seq} of an \grammarterm{enum-base}
 shall name an integral type; any cv-qualification is ignored.
 An \grammarterm{opaque-enum-declaration} declaring an unscoped enumeration shall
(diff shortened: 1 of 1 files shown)


source/declarations.tex
```diff
@@ -2040,10 +2040,15 @@
 If the \grammarterm{enum-key} is followed by a
```
zygoloid: Please add "in an enum-head or opaque-enum-declaration" after "enum-key". I think we may have duplicated this previously as a misguided and unclear attempt to address both cases separately.

zygoloid: Hmm, could we refactor opaque-enum-declaration to use enum-head-name, and reference that here? That would avoid confusion over whether this applies if there is an intervening attribute-specifier-seq.

tkoeppe:
@jensmaurer: actionable comments here

jensmaurer:
Actions performed. @zygoloid is next.

zygoloid:
Looks good, let's apply this after the Jacksonville motions.

source/declarations.tex
```diff
@@ -2038,12 +2038,18 @@
 underlying type as in the original declaration.
 
 \pnum
-If the \grammarterm{enum-key} is followed by a
-\grammarterm{nested-name-specifier}, the \grammarterm{enum-specifier} shall
+If an \grammarterm{enum-head-name} contains a
+\grammarterm{nested-name-specifier}, the latter
+shall not begin with a \grammarterm{decltype-specifier} and
```
zygoloid: Was this `decltype` rule for *opaque-enum-declaration*s present before (maybe implied by pre-existing wording)? I see we had such a rule for *enum-head*s.

jensmaurer: Well, we have (unchanged from before):

If an \grammarterm{opaque-enum-declaration} contains
a \grammarterm{nested-name-specifier},
the declaration shall be an explicit specialization\iref{temp.expl.spec}.

and we have 

11.3 [dcl.meaning]
The nested-name-specifier of a qualified declarator-id shall not begin with a decltype-specifier.

That seems close enough to me. (it seems we don't want redeclarations to use decltype to nominate the entity being redeclared. Which is totally reasonable.)

jensmaurer:
Rebased.

source/declarations.tex
zygoloid: I agree that we don't want to allow `decltype` here, and indeed current implementations don't distinguish between the opaque enum declaration case and the regular enum case, but if this is a normative change (as it appears to be), we should fix that portion of this via the CWG issue process.

jensmaurer: Fixed by limited the prohibition to the enum-head case.

tkoeppe:
@zygoloid says:

> I think what I'm most worried about here is that making the editorial change first turns the wording from "wrong but somewhat ambiguous" (it looks like we just forgot the rule) to "unambiguously wrong" (the difference between the two cases is clearly intentional), which I would expect to make it more likely for a user or implementer to get confused or do the wrong thing.

Let's run this by CWG then.

jensmaurer:
Rebased.

jensmaurer:
CWG in San Diego: Fine with the change.

http://wiki.edg.com/pub/Wg21sandiego2018/CoreWorkingGroup/cwg-ed1827.html

PR #1829, opened 2017-11-20
[swappable.requirements] Merge example's comment to one line

johelegp:
This follows the style of the surrounding comments.

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -1681,8 +1681,8 @@ void value_swap(T&& t, U&& u) {
 template <class T>
 void lv_swap(T& t1, T& t2) {
   using std::swap;
-  swap(t1, t2);                                 // OK: uses swappable conditions for
-}                                               // lvalues of type \tcode{T}
+  swap(t1, t2);                                 // OK: uses swappable conditions for lvalues of type \tcode{T}
+}
 
 namespace N {
   struct A { int m; };


tkoeppe:
I think I'd rather just get rid of the pointless line break, non?

johelegp:
Of the surrounding comments, if that was done to the one below, it would merge the comments. Although it can be done for the one in the fix and the one above it, the current style is most general. What do you think? Do we keep the one below and remove them from the two above.

johelegp:
```C++
  swap(t1, t2);     // OK: uses swappable conditions for lvalues of type \tcode{T}
```
Oh, you mean like that?

tkoeppe:
Yes.

johelegp:
Done.

tkoeppe:
Thanks!

jensmaurer:
Looks good to me.

tkoeppe:
Oh, wait, this would overrun the line, right? Can you render it and see what it looks like? I didn't realize how far out the comment is. If it doesn't fit, let's go back to the original proposal.

johelegp:
I'll try rendering it. I think I've never done it before. However, I thought it was fine by looking at the code, as the comment's font should be mostly shorter than the code's monospaced font.

johelegp:
Actually, I just assumed it wouldn't fit, but it does. Now I know I should render in these cases.

tkoeppe:
No, let's not do that, because the overall consistency with the remainder of the code makes it preferable to keep all the comments at the same column. Your original proposal was good.

tkoeppe:
Oh, does it actually fit? Great.

jensmaurer:
I just tried it out locally; the change does not cause an overfull hbox.

tkoeppe:
@jensmaurer: That's not relevant, though: what matters is that there isn't a line break. Long comments will happily go on as many lines as they need, all neatly justified.

PR #1830, opened 2017-11-20
[layout] Feet at bottom

tkoeppe:


tkoeppe:
This is a test. Move on, nothing to see here.

tkoeppe:
@godbyk: Do you know if/how one can request a more recent Travis system?

godbyk:
@tkoeppe It looks like Ubuntu Trusty is the current version supported by Travis CI.

We have a couple options:

1. Install the required packages using the upstream TeX Live installer. This would get us the latest versions of every package (which change on a nearly daily basis).
2. Use the new Docker functionality provided by Travis CI + Trusty image to create our own Docker image with a newer version of Ubuntu (or other distribution) and use the newer TeX packages from that distro.

I use the upstream TeX Live installer on my own computers and could update the build script to install those.  I've also used Docker for my own CI server (though I'd need to read the docs to see how we go about using it on Travis CI).

I'll do a little experimenting this weekend and see how fast or slow each option is. I think with option (1), we could install the TeX Live files into a local directory and have Travis CI cache them for us so we wouldn't have to worry about constantly downloading them each time a new build is triggered (and not have to deal with newer versions of packages for each subsequent build).  I don't know if Travis CI caches any Docker images or not; I'll skim through the docs and see how things work.

tkoeppe:
@godbyk: Thanks! There's no rush. We'll only need this for the next publication, so we can wait until Travis updates to something more modern if there isn't a fast way to backport this.

tkoeppe:
I was trying to do something with `tlmgr update memoir`, but it doesn't quite work.

Travis and @zygoloid are now the only systems not supporting `\feetatbottom`.

tkoeppe:
@zygoloid: How is your tex distribution these days?

tkoeppe:
Grand. I believe then we only need to find a way to get a recent memoir onto Travis. Cheers.

godbyk:
Whoops! This completely fell off my radar.

I just tried the newer Ubuntu Trusty (14.04) image that Travis provides and it still has an older version of memoir (memoir 2013/05/30 v3.7b).

Let me read up on some Travis CI stuff this evening and see what the best way to do this is. We have a few options available.

tkoeppe:
@godbyk: Thanks a ton, your help is as always greatly appreciated!

godbyk:
@tkoeppe I've updated the Travis CI stuff, so it should work now. While I haven't actually tested the `\feetatbottom` macro, I did check to ensure that it exists in the `memoir` package installed in the new setup.

tkoeppe:
Wonderful, thank you! I'll test it with this PR.

PR #1831, opened 2017-11-20
Add Arch Linux getting started instructions

johelegp:


tkoeppe:
Have you tried this based on a completely blank installation? I don't have the means to test this myself.

johelegp:
No. I'll set up a VM and test it. `yaourt` simplifies a process much like `latexmk`, so I'd have to swap that out for the raw instructions.

johelegp:
I reworded the instructions to reflect the Arch Linux philosophy. I have also tested them on a bare-bones installation, and they work.

PR #1832, opened 2017-11-20
[reverse.iterators] Dissolve single-item subclauses.

jensmaurer:
Partially addresses #1429.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -1254,24 +1254,22 @@ Additionally,
 \tcode{Iterator}
 shall meet the requirements of a random access iterator\iref{random.access.iterators}
 if any of the members
-\tcode{operator+}~(\ref{reverse.iter.op+}),
-\tcode{operator-}~(\ref{reverse.iter.op-}),
-\tcode{operator+=}~(\ref{reverse.iter.op+=}),
-\tcode{operator-=}~(\ref{reverse.iter.op-=}),
-\tcode{operator[]}\iref{reverse.iter.opindex},
-or the non-member operators
-\tcode{operator<}~(\ref{reverse.iter.op<}),
-\tcode{operator>}~(\ref{reverse.iter.op>}),\\
-\tcode{operator<=}~(\ref{reverse.iter.op<=}),
-\tcode{operator>=}~(\ref{reverse.iter.op>=}),
-\tcode{operator-}\iref{reverse.iter.opdiff}
+\tcode{operator+},
+\tcode{operator-},
+\tcode{operator+=},
+\tcode{operator-=}\iref{reverse.iter.nav},
+\tcode{operator[]}\iref{reverse.iter.elem},
+or the non-member operators\iref{reverse.iter.cmp}
+\tcode{operator<},
+\tcode{operator>},
+\tcode{operator<=},
+\tcode{operator>=},
+\tcode{operator-},
 or
-\tcode{operator+}\iref{reverse.iter.opsum}
+\tcode{operator+}\iref{reverse.iter.nonmember}
 are referenced in a way that requires instantiation\iref{temp.inst}.
 
-\rSec3[reverse.iter.ops]{\tcode{reverse_iterator} operations}
-
-\rSec4[reverse.iter.cons]{\tcode{reverse_iterator} constructor}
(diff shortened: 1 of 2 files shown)


jensmaurer:
@zygoloid : This merge went the wrong way: You merged "master" into my branch, it seems, rather than the other way round.

jensmaurer:
Rebased.

zygoloid:
@jensmaurer This is github's "rebase someone else's PR" feature. Sadly it appears to be broken right now (it fails to actually re-check whether the PR is mergeable afterwards or something).

source/iterators.tex
```diff
@@ -1377,7 +1369,40 @@
...
+
+\begin{itemdescr}
+\pnum
+\returns
+\tcode{current[-n-1]}.
+\end{itemdescr}
+
+\indexlibrarymember{operator+}{reverse_iterator}%
+\begin{itemdecl}
+constexpr reverse_iterator operator+(difference_type n) const;
```
zygoloid: I find it a bit strange to call this an "observer", since it does not of itself provide any mechanism to actually observe the state of the iterator. (It seems a little odd to call `operator*`, `operator->`, and `operator[]` observers, rather than something like "element access", too).

I might be tempted to instead lump `operator++`, `operator--`, `operator+`, `operator+=`, `operator-`, `operator-=` under something like "`reverse_iterator` navigation", put these three under "`reverse_iterator` element access", and keep `base` as a single-item subclause (which I think makes sense, as it's the one and only part of the `reverse_iterator` interface that is not part of some Iterator concept).

jensmaurer: Sounds reasonable.

AlisdairM:
This is worth calling out to @mclow as it will affect a small number of issues in the LWG issues list.  I think the proposed change looks cleaner though.

mclow:
Thanks for the callout; I can fix the issues.  
Anytime you change a stable name, I'd appreciate a shout-out.

zygoloid:
Postponed to post-meeting mailing due to changes to stable names.

jensmaurer:
Rebased.

source/iterators.tex
```diff
@@ -1315,8 +1313,6 @@
 \tcode{u.current}.
 \end{itemdescr}
 
-\rSec4[reverse.iter.op=]{\tcode{reverse_iterator::operator=}}
```
zygoloid: We're missing an xrefdelta entry for this.

source/xrefdelta.tex
```diff
@@ -52,5 +52,31 @@
...
+\movedxref{reverse.iter.opref}{reverse.iter.elem}
+\movedxref{reverse.iter.opindex}{reverse.iter.elem}
+
+\movedxref{reverse.iter.op+}{reverse.iter.nav}
+\movedxref{reverse.iter.op-}{reverse.iter.nav}
+\movedxref{reverse.iter.op++}{reverse.iter.nav}
+\movedxref{reverse.iter.op+=}{reverse.iter.nav}
+\movedxref{reverse.iter.op\dcr}{reverse.iter.nav}
+\movedxref{reverse.iter.op-=}{reverse.iter.nav}
+\movedxref{reverse.iter.op++}{reverse.iter.nav}
```
zygoloid: We already did this one.

source/xrefdelta.tex
```diff
@@ -52,5 +52,31 @@
...
 
+% Single-item [reverse.iterators] subclauses were dissolved.
+\movedxref{reverse.iter.op==}{reverse.iter.cmp}
+\movedxref{reverse.iter.op<}{reverse.iter.cmp}
+\movedxref{reverse.iter.op!=}{reverse.iter.cmp}
+\movedxref{reverse.iter.op>}{reverse.iter.cmp}
+\movedxref{reverse.iter.op>=}{reverse.iter.cmp}
+\movedxref{reverse.iter.op<=}{reverse.iter.cmp}
+
+\movedxref{reverse.iter.star}{reverse.iter.elem}
```
zygoloid: This was called reverse.iter.op.star, not reverse.iter.star, for some reason.

jensmaurer:
@zygoloid, your comments are addressed in pull request #2015.

PR #1835, opened 2017-11-21
Create a new index for library headers

### LINKED ISSUE #1807: rationalize library header name indexing

jensmaurer:
and remove them from the index of library names.

Fixes #1807.

### DIFF
--- a/source/Makefile
+++ b/source/Makefile
@@ -38,6 +38,7 @@ reindex:
 	$(STDPDF)
 	$(STDPDF)
 	makeindex generalindex
+	makeindex headerindex
 	makeindex libraryindex
 	makeindex grammarindex
 	makeindex impldefindex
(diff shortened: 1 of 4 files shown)


jensmaurer:
(The index is actually two pages, not one, contrary to the estimation by @zygoloid in #1807.)

The separate index seems rather useful to get an overview of all headers in the standard library. You might also want to mention it in the editor's report.

PR #1837, opened 2017-11-22
[diff] Introduce numbered paragraphs

jensmaurer:
Alternative for #1803.

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -10,7 +10,7 @@ ISO C, by the chapters of this document.
 
 \rSec2[diff.lex]{\ref{lex}: lexical conventions}
 
-\ref{lex.key}
+\diffref{lex.key}
 \change New Keywords\\
 New keywords are added to \Cpp{};
 see \ref{lex.key}.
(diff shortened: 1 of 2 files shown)


zygoloid:
Hmm. "Affected subclause:" is probably not quite what we want here, because it could be more than one section and it could be a clause rather than a subclause that we name. But other than that detail, this looks like it could be good. Can you provide a screenshot?

jensmaurer:
@zygoloid: I've attached the screenshot. I haven't found a [diff] section applying to an entire clause; it's always subclauses. And "Affected subclause" doesn't look so bad even if multiple subclause numbers follow.

Eelis:
Looks great, much nicer than my simplistic #1803!

Personally I think it might read even better if the "Change:" line came before "Affected subclause:".

zygoloid:
@jensmaurer Look at the first change in [diff.cpp03.library] for a "whole clause" case. In fact, that's a "whole 14 clauses" case, but it looks like it's the only example of its kind.

jensmaurer:
[diff.cpp03.library] is broken in a way, anyway: Why don't we have a similar paragraph for the other C++xy subclauses? Certainly, the library has evolved with each iteration of the standard.

jensmaurer:
It would be nice to say "Affected clause: 14-30", but \ref always produces "Clause" in front.

I've gone for "Affected: Clause 14 - Clause 30" for now:
diff-clause

PR #1838, opened 2017-11-23
Convert blah\opt to \opt{blah}

### LINKED ISSUE #1433: Review use of \xspace in macro definitions

jensmaurer:
Fixes #1433.

### DIFF
--- a/source/access.tex
+++ b/source/access.tex
@@ -195,7 +195,7 @@ Member declarations can be labeled by an
 (\ref{class.derived}):
 
 \begin{ncbnftab}
-access-specifier \terminal{:} member-specification\opt{}
+access-specifier \terminal{:} \opt{member-specification}
 \end{ncbnftab}
 
 An
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -776,7 +776,7 @@ The point of declaration of a class first declared in an
 \item for a declaration of the form
 
 \begin{ncbnf}
-class-key attribute-specifier-seq\opt{} identifier \terminal{;}
+class-key \opt{attribute-specifier-seq} identifier \terminal{;}
 \end{ncbnf}
 
 the \grammarterm{identifier} is declared to be a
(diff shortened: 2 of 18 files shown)


jensmaurer:
No diffs in diffpdf "words" mode; two expected line-breaking diffs in "appearance mode" for the C++ 20nn non-breaking space.

I can squash some of these commits if desired.

tkoeppe:
I stole a few commits I liked from this PR. Please rebase.

jensmaurer:
Rebased.

After @tkoeppe's partial merge, the remaining change is to morph blah\opt into \opt{blah}.

tkoeppe:
Is the "C++ 2003" change worth it? Either we already don't break the line, or if we need to break, it doesn't look like a terrible place to break. Why prevent the line break at all costs?

jensmaurer:
@tkoeppe: That's why it is a separate commit.

In my view, it's somewhat ugly to have "C++" at the end of a line and then have the next line start with "2003". And in the two situations where it makes a difference, TeX find another linebreak opportunity. Maybe it's better to apply some extra penalty as opposed to "prevent at all costs", though.

tkoeppe:
Yes, if we could "discourage" rather than "forbid absolutely", that'd be nice. I.e. it's good if we can *improve* the situation without *breaking* it.

jensmaurer:
Dropped patch for avoiding line breaks for earlier versions of C++.

jensmaurer:
@zygoloid : This merge went the wrong way: You merged "master" into my branch, it seems, rather than the other way round.

tkoeppe:
@jensmaurer: No, I think he was trying to rebase your PR. Apparently that's a thing you can do now.

jensmaurer:
@tkoeppe: Ahem. I thought "rebases" don't want to do a "merge", but a force-push, to keep the history clean?

jensmaurer:
Ah, the \tilde changes. Let me fix those in a proper rebase and push again.

tkoeppe:
No visual diff, looks good. Will leave to @zygoloid to merge because of the wide impact.

PR #1842, opened 2017-11-23
[basic.life],[mem.res.private],[mem.res.pool.mem],[mem.res.monotonic.buffer] fix cross-references

### LINKED ISSUE #1841: Does [basic.life] have the wrong cross-reference?
> Otherwise, such a pointer refers to allocated storage (6.6.4.4.2), 

>  Otherwise, such a glvalue refers to allocated storage (6.6.4.4.2),

This links to [basic.stc.dynamic.deallocation] which only mentions "storage" once, in "the deallocation function shall deallocate the storage referenced by the pointer".

Is it supposed to be [basic.stc.dynamic.**allocation**] instead, which has lots to say about "allocated storage"?

This bad cross-reference was copied by [N3916](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n3916.pdf#page=16) (see the note to the editor on page 16) and so appears in three more places.

jwakely:
6.6.4.4.2 [basic.stc.dynamic.deallocation] only mentions "storage" once, in "the deallocation function shall deallocate the storage referenced by the pointer".

I think the references are supposed to be to 6.6.4.4.1 [basic.stc.dynamic.**allocation**] instead, which has lots to say about "allocated storage".

Fixes #1841 (assuming I'm right about the references being wrong).

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -2877,7 +2877,7 @@ used but only in limited ways.
 For an object under construction or destruction, see~\ref{class.cdtor}.
 Otherwise, such
 a pointer refers to allocated
-storage\iref{basic.stc.dynamic.deallocation}, and using the pointer as
+storage\iref{basic.stc.dynamic.allocation}, and using the pointer as
 if the pointer were of type \tcode{void*}, is
 well-defined. Indirection through such a pointer is permitted but the resulting lvalue may only be used in
 limited ways, as described below. The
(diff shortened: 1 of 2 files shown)


jensmaurer:
Looks good to me.
(Maybe we should actually define the term "allocated storage", but at least we're no longer pointing into the weeds now.)

PR #1843, opened 2017-11-23
[func.require] Clarify which assignment operators

johelegp:


jensmaurer:
CopyAssignable implies MoveAssignable [utility.arg.requirements], so talking about "assignment operators" (plural) makes sense here. Though I'm wondering if the existing text would also (inadvertently) constrain compound assignment operators, in particular if we now have the plural.

"... , copy assignment operator, and move assignment operator" would clarify that.

johelegp:
I didn't think about the compound assignment operators. I'll use your clear wording.

PR #1848, opened 2017-11-24
[fs.op.funcs] Separate effects from returns.

### LINKED ISSUE #815: filesystem: Bad mixing of effects in Returns element
During the Oulo meeting's LWG discussion of some filesystem issues it has been observed that some current _Returns_ elements of fs operations unfortunately mix effects and the actual return specification in a single paragraph. 

For example, in [fs.op.file_size] p1 we have

> Returns: If `!exists(p) || !is_regular_file(p)` an error is reported (27.10.7). Otherwise, the
> size in bytes of the file `p` resolves to, determined as if by the value of the POSIX `stat` structure
> member `st_size` obtained as if by POSIX `stat()`. The signature with argument `ec` returns `static_-
> cast<uintmax_t>(-1)` if an error occurs.

Here, the part

> If `!exists(p) || !is_regular_file(p)` an error is reported (27.10.7).

would better belong to a separate _Effects_ element (such as in `copy` or `copy_file`). Then, the current _Returns_ element could be rephrased as follows:

> Returns: The size in bytes of the file `p` resolves to, determined as if by the value of the POSIX `stat` structure member `st_size` obtained as if by POSIX `stat()`. The signature with argument `ec` returns `static_cast<uintmax_t>(-1)` if an error occurs.

A similar problem exists for [fs.op.temp_dir_path] p1, where we have

> Returns: An unspecifed directory path suitable for temporary files. An error shall be reported if
> `!exists(p) || !is_directory(p)`, where `p` is the path to be returned. The signature with argument
> `ec` returns `path()` if an error occurs.

which could be split as follows:

> Let `p` be the path to be returned.
> 
> Effects: If `!exists(p) || !is_directory(p)`  an error is reported (27.10.7).
> 
> Returns: An unspecifed directory path suitable for temporary files. The signature with argument
> `ec` returns `path()` if an error occurs.

Is it possible to improve this situation editorially?

jensmaurer:
Fixes #815.

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -14957,12 +14957,14 @@ uintmax_t file_size(const path& p, error_code& ec) noexcept;
 
 \begin{itemdescr}
 \pnum
+\effects
+If \tcode{exists(p)} is \tcode{false}, an error is reported\iref{fs.err.report}.
+
+\pnum
 \returns
 \begin{itemize}
 \item
-  If \tcode{exists(p)} is \tcode{false}, an error is reported\iref{fs.err.report}.
-\item
-  Otherwise, if \tcode{is_regular_file(p)}, the size in bytes of the file
+  If \tcode{is_regular_file(p)}, the size in bytes of the file
   \tcode{p} resolves to, determined as if by the value of the POSIX \tcode{stat}
   structure member \tcode{st_size} obtained as if by POSIX \tcode{stat()}.
 \item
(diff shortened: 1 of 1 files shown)


tkoeppe:
Please fix.

jensmaurer:
Fixed.

source/iostreams.tex
```diff
@@ -15751,8 +15753,14 @@
...
-\returns An unspecified directory path suitable for temporary files. An error shall be reported if
-\tcode{exists(p)} is \tcode{false} or \tcode{is_directory(p)} is \tcode{false}, where \tcode{p} is the path to be returned.
+Let \tcode{p} be the path to be returned.
+
+\pnum
+\effects If \tcode{exists(p)} is \tcode{false} or \tcode{is_directory(p)} is
+  \tcode{false}, an error is reported\iref{fs.err.report}.
+
+\pnum
+\returns An unspecified directory path suitable for temporary files.
```
tkoeppe: Hm, I think it would be nice if we could make reference to `p` in the Returns element. Maybe the "unspecified directory path..." could move up into the defining paragraph, and we could say "returns `p`" here?

jensmaurer: Sounds like a good idea.

jensmaurer:
Rebased + rephrased for temp_directory_path().

PR #1851, opened 2017-11-25
Remove \tilde in favour of \~.

Eelis:
This was a FIXME in macros.tex.

No visible difference.

### DIFF
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -199,8 +199,6 @@
 \newcommand{\bigoh}[1]{\ensuremath{\mathscr{O}(#1)}}
 
 % Make all tildes a little larger to avoid visual similarity with hyphens.
-% FIXME: Remove \tilde in favour of \~.
-\renewcommand{\tilde}{\textasciitilde}
 \renewcommand{\~}{\textasciitilde}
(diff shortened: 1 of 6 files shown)


source/macros.tex
```diff
@@ -199,8 +199,6 @@
 \newcommand{\bigoh}[1]{\ensuremath{\mathscr{O}(#1)}}
 
 % Make all tildes a little larger to avoid visual similarity with hyphens.
-% FIXME: Remove \tilde in favour of \~.
-\renewcommand{\tilde}{\textasciitilde}
 \renewcommand{\~}{\textasciitilde}
```
godbyk: Note that this overrides the default definition of `\~` which applies the tilde accent to the following character (e.g., `\~n` is typeset as ‘ñ’).

In addition, I would move the `\raisebox` etc. code to this macro and leave the `\textasciitilde` macro unmolested. (Modify as few original macros as necessary. Try to keep all your changes localized.) This also saves you from having to copy the definition to `\OldTextAsciiTilde`, removes redirection, and improves readability.

tkoeppe: @godbyk: Could you spell that out for me? I've tried a few things, but they end up visually different.

PR #1855, opened 2017-11-28
[basic.compound], [conv.ptr] Improve index for 'null pointer value'

### LINKED ISSUE #1487: Harmonize index entries for "null pointer value"
We currently have the following two unrelated entries:

- ["null pointer value"](https://github.com/cplusplus/draft/blob/master/source/basic.tex#L3972), pointing to 3.9.2 [basic.compound]
- ["value, null pointer"](https://github.com/cplusplus/draft/blob/master/source/conversions.tex#L493), pointing to 4.11 [conv.ptr]

Maybe both should be made to use the same form, along with a redirect entry ("see") from one form to the other?

jensmaurer:
Fixes #1487.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -4074,9 +4074,11 @@ a \defn{pointer to} an object or function (the pointer is said to \defn{point} t
 \item
 a \defn{pointer past the end of} an object\iref{expr.add}, or
 \item
-the \defn{null pointer value}\iref{conv.ptr} for that type, or
+\indextext{null pointer value|see{value, null pointer}}
+the \defnx{null pointer value}{value!null pointer}\iref{conv.ptr} for that type, or
 \item
-an \defn{invalid pointer value}.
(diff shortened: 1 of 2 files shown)


source/basic.tex
```diff
@@ -4074,9 +4074,11 @@
 \item
 a \defn{pointer past the end of} an object\iref{expr.add}, or
 \item
-the \defn{null pointer value}\iref{conv.ptr} for that type, or
+\indextext{null pointer value|see{value, null pointer}}
+the \defnx{null pointer value}{value!null pointer}\iref{conv.ptr} for that type, or
```
tkoeppe: Can you check *.ilg for warnings, just to be sure?

jensmaurer: $ grep warnings *.ilg | grep -v " 0 warnings"
<empty>

Can't find any warnings.

PR #1859, opened 2017-11-28
[intro.scope] C++ is not a strict superset of C

### LINKED ISSUE #1664: intro.scope implies that C++ is a strict superset of C
In [intro.scope], the statement

> In addition to the facilities provided by C, C++ provides [ ... ]

implies that C++ is a strict superset of C.

jensmaurer:
Fixes #1664.

### DIFF
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -14,8 +14,8 @@ requirement appear at various places within this document.
 \Cpp{} is a general purpose programming language based on the C
 programming language as described in ISO/IEC 9899:2011
 \doccite{Programming languages --- C} (hereinafter referred to as the
-\defnx{C standard}{C!standard}). In addition to
-the facilities provided by C, \Cpp{} provides additional data types,
+\defnx{C standard}{C!standard}). \Cpp{} provides many facilities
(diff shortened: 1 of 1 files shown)


jensmaurer:
This change is a bit lame, but minimal.

source/intro.tex
```diff
@@ -14,7 +14,7 @@
 \Cpp{} is a general purpose programming language based on the C
 programming language as described in ISO/IEC 9899:2011
 \doccite{Programming languages --- C} (hereinafter referred to as the
-\defnx{C standard}{C!standard}). In addition to
+\defnx{C standard}{C!standard}). In addition to some of
```
zygoloid: Hmm, how about:

> C++ provides many facilities beyond those provided by C, including [...]

This simultaneously minimizes the suggestion that C++ is a strict superset of C, and removes the suggestion that the following list of C++ features is a complete list of features in C++ but not C.

jensmaurer: Done.

jensmaurer:
Rebased + implemented @zygoloid suggestion. (I fail to get excited about [intro.scope].)

PR #1863, opened 2017-11-29
[filesystems] 'dot' and 'dot-dot' are not \grammarterms

### LINKED ISSUE #1527: P0492R2 makes dot and dot-dot no longer grammarterms
Following the changes to the generic path grammar ([path.generic]) the special filenames _dot_ and _dot-dot_ are no longer grammar terms, but are simply defined terms in [fs.def.filename]. (In terms of the grammar they are covered by the _filename_ production.)

Throughout the clause we use `\grammarterm{dot}` and `\grammarterm{dot-dot}` to give them a special appearance. Should we use `\term` instead? Should they only be italicized on their first definition? (IMHO that would make the specification unclear, even though it's consistent with how we present other defined terms).

jensmaurer:
Fixes #1527.

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -10543,8 +10543,8 @@ that contain information about
 other files, possibly including other directory files.
 The \defn{parent directory} of a directory is
 the directory that both contains a
-directory entry for the given directory and is represented by the filename
-\grammarterm{dot-dot} in the given directory.
+directory entry for the given directory and is represented by the dot-dot
+filename\iref{fs.path.generic} in the given directory.
 The \defn{parent directory}
 of other types of files is a directory containing a directory
 entry for the file under discussion.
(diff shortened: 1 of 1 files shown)


source/iostreams.tex
```diff
@@ -10544,7 +10544,7 @@
 The \defn{parent directory} of a directory is
 the directory that both contains a
 directory entry for the given directory and is represented by the filename
-\grammarterm{dot-dot} in the given directory.
+dot-dot in the given directory.
```
jwakely: Should we add a forward reference to the definition or "dot-dot", given that it occurs after this point?

jensmaurer: Good idea.

jensmaurer:
Rebased + updated as per @jwakely's suggestion.

source/iostreams.tex
```diff
@@ -11259,10 +11259,10 @@
 be the same as one \grammarterm{directory-separator} character.
 
 \pnum
-The filename \grammarterm{dot} is treated as a reference to the current directory.
-The filename \grammarterm{dot-dot} is treated as a reference to the parent directory.
-What the filename \grammarterm{dot-dot} refers to
-relative to \grammarterm{root-directory} is \impldef{meaning of \grammarterm{dot-dot} in \grammarterm{root-directory}}.
+The filename dot is treated as a reference to the current directory.
```
zygoloid: "The filename dot" sounds like a particular kind of dot, not a particular kind of filename. How about changing to "The dot filename" and "the dot-dot filename" throughout?

jensmaurer: It turns out only this place needs changing; all other places already say "dot filename" or similar.

jensmaurer:
Rebased + updated.

PR #1867, opened 2017-11-30
[language.support, utilities] Condense description of exception classes

### LINKED ISSUE #748: Exceptions classes deriving directly from std::exception should be presented consistently
I am opening an issue, as part of resolving this will be adding the results of the 'what' function for each class below to the index of implementation defined behavior.  However, before looking into that, I want to ensure a consistent presentation of these classes with essentially the same behavior.
   bad_alloc
   bad_array_new_length
   bad_cast
   bad_exception
   bad_function_call
   bad_typeid
   bad_weak_ptr

At one extreme, we have the specification for the original bad_\* classes:
class bad_alloc : public exception {
public:
  bad_alloc() noexcept;
  bad_alloc(const bad_alloc&) noexcept;
  bad_alloc& operator=(const bad_alloc&) noexcept;
  virtual const char\* what() const noexcept;
};

At the other, we have the specification for the exceptions adopted via TR1:
class bad_weak_ptr: public std::exception {
public:
  bad_weak_ptr() noexcept;
};

Note that an override of 'what' is assumed, but not declared, with the specification for 'what' buried in the contract of the default constructor.

My preferred formulation would be:

struct bad_exception : exception {
  const char\* what() const noexcept override;
};

This strips out the redundant exceptions and assignment operators, which will be implicitly declared/defined with 'noexcept' specifier.  It uses 'struct' as defaulting everything to public access simplifies further, and finally adds the 'override' qualifier to 'what', for completeness.  As this involves a change of normative wording, I believe it would require an LWG issue to resolve.

Failing that, I would prefer to consistently apply the 'bad_alloc' format above, where everything is called out explicitly.  In both cases, we get a 'what' function in the index that can be linked to the index of implementation defined behavior.

jensmaurer:
Fixes #748.

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -2731,9 +2731,7 @@ this order.
 namespace std {
   class bad_alloc : public exception {
   public:
-    bad_alloc() noexcept;
-    bad_alloc(const bad_alloc&) noexcept;
-    bad_alloc& operator=(const bad_alloc&) noexcept;
+    // see \ref{exception} for the specification of the special member functions
     const char* what() const noexcept override;
   };
 }
@@ -2754,12 +2752,6 @@ const char* what() const noexcept override;
 \pnum
 \returns
 An \impldef{return value of \tcode{bad_alloc::what}} \ntbs{}.
-
-\pnum
-\remarks
-The message may be a null-terminated multibyte string\iref{multibyte.strings},
-suitable for conversion and display as a
-\tcode{wstring}~(\ref{string.classes}, \ref{locale.codecvt}).
 \end{itemdescr}
 
 \indexlibrary{\idxcode{bad_array_new_length}}%
@@ -2770,7 +2762,7 @@ suitable for conversion and display as a
 namespace std {
   class bad_array_new_length : public bad_alloc {
   public:
-    bad_array_new_length() noexcept;
+    // see \ref{exception} for the specification of the special member functions
     const char* what() const noexcept override;
   };
 }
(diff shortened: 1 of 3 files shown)


jensmaurer:
Note that this currently has a lame sentence "The constructors are specified in \ref{exception}." to point readers to the paragraph that says "all standard library classes T deriving from std::exception have these constructors". I've also removed the NTBS blurb from the derived classes; it's now implied by std::exception::what(). Maybe we can get rid of the lame sentence and just add a comment in the class synopsis, saying something like "// see also \ref{exception}. Suggestions welcome.

Oh, and before I forget: This is borderline editorial; for @zygoloid to decide.

zygoloid:
I would prefer a comment inside the class definition pointing at [exception] rather than wording in a section that a casual reader may well not look at.

Just using the defaulted special members here seems very reasonable to me. I *think* it's fine that we now give these exception subclasses move operations, but a second opinion on that would be good (@jwakely?).

jensmaurer:
Updated.

The fun part is that `logic_error` and other classes from `<stdexcept>` that take a user-defined string now also have a default constructor per [exception]. That might be undesirable, and is definitely not editorial.

jensmaurer:
Fixed the "logic_error now has a default constructor" issue.

We might want to have LWG ack the changes.

jensmaurer:
Editorial meeting consensus: Looks like the right direction. @zygoloid asks for LWG expert review. @jwakely ?

tkoeppe:
@jwakely: Ping

jensmaurer:
Rebased.

@jwakely, ping?

jensmaurer:
@jwakely, ping^2.

PR #1868, opened 2017-12-01
Use 'anonymous union object' where appropriate.

### LINKED ISSUE #973: After applying CWG1860, replace "anonymous union" and "anonymous union member" with "anonymous union object" where relevant
CWG1860 introduces the term "anonymous union object" for the object (or class member) implicitly declared by an anonymous union declaration. We should update existing woolly references to that object to use the defined term.

jensmaurer:
The term was introduced by CWG 1860.

Fixes #973.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -861,7 +861,7 @@ shall be initialized;
 
 \item
 if the class is a union-like class, but is not a union, for each of its anonymous union
-members having variant members, exactly one of them shall be initialized;
+objects having variant members, exactly one of them shall be initialized;
 
 \item
 for a non-delegating constructor, every constructor selected to initialize non-static
--- a/source/declarators.tex
+++ b/source/declarators.tex
@@ -2596,7 +2596,7 @@ if \tcode{T} is a union with at least one non-static data member,
 exactly one variant member has a default member initializer,
 \item
 if \tcode{T} is not a union,
-for each anonymous union member with at least one non-static data member (if any),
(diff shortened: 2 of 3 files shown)


jensmaurer:
I'm slightly uneasy with this change, because it claims that objects have members, but we often discuss properties at the level of the type here, not at the level of a specific object.

zygoloid:
I think the uneasiness is ultimately due to the term being inappropriate -- "anonymous union object" makes sense for namespace-scope or block-scope anonymous unions, but at class scope we're not defining an object. I wonder if it would be better to define the term "anonymous union member" in [class.union.anon] for the class-scope case.

Maybe something like this:

> A union of the form<br>
>   `union {` *member-specification* `} ;`<br>
> is called an *anonymous union*; it defines an unnamed type and an unnamed variable or non-static data member of that type. A non-static data member of anonymous union type is called an *anonymous union member*. An object of anonymous union type (either an unnamed variable introduced by an anonymous union declaration or a member subobject corresponding to an anonymous union member) is called an *anonymous union object*.

... though we'll need to check through the other references to "anonymous union object" to see whether they need changing too.

What do you think?

jensmaurer:
Sounds better, although we tended to avoid "anonymous union member" in the past because of ambiguity: a member of an anonymous union, or a member that is an anonymous union?

jensmaurer:
Editorial meeting consensus: not a net improvement.

PR #1873, opened 2017-12-02
[cpp] Replace \term with \defn

jensmaurer:
and slightly adjust indexing.

Partially addresses #329.

### DIFF
--- a/source/preprocessor.tex
+++ b/source/preprocessor.tex
@@ -1,10 +1,10 @@
 %!TEX root = std.tex
 \rSec0[cpp]{Preprocessing directives}%
-\indextext{preprocessing directives|(}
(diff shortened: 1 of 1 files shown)


source/preprocessor.tex
```diff
@@ -1430,7 +1429,7 @@
 \terminal{_Pragma} \terminal{(} string-literal \terminal{)}
 \end{ncbnf}
 
-is processed as follows: The string literal is \term{destringized}
+is processed as follows: The string literal is \defn{destringized}
```
zygoloid: We should index this as "destringization" or similar.

jensmaurer: Sure.

jensmaurer:
Rebased + updated.

PR #1876, opened 2017-12-03
[string.cons] Replace postcondition tables by ordinary Postcondition elements

### LINKED ISSUE #1875: [string.cons] Editorial change for caption on table 56
LWG closed issue [3033](https://wg21.link/LWG3033) as "NAD editorial", with a suggestion that changing the caption on the table would be both editorial and sufficient.

The current caption is "`basic_string(const basic_string&)` effects"

My first suggestion is: "`basic_string(const basic_string&)` and `basic_string(const basic_string&&)` effects", but I'm not going to insist on that.

The discussion occurred on the -lib reflector in the thread titled "Issue 3033: basic_string move ctor is underspecified"

tkoeppe:
Fixes #1875.

### DIFF
--- a/source/strings.tex
+++ b/source/strings.tex
@@ -1198,15 +1198,12 @@ explicit basic_string(const Allocator& a) noexcept;
 \begin{itemdescr}
 \pnum
 \effects
-Constructs an object of class
-\tcode{basic_string}.
-The postconditions of this function are indicated in Table~\ref{tab:strings.ctr.1}.
+Constructs an object of class \tcode{basic_string}.
 
-\begin{libefftabvalue}{\tcode{basic_string(const Allocator\&)} effects}{tab:strings.ctr.1}
-\tcode{data()}      &   a non-null pointer that is copyable and can have 0 added to it  \\ \rowsep
-\tcode{size()}      &   0                                                               \\ \rowsep
-\tcode{capacity()}  &   an unspecified value                                            \\
-\end{libefftabvalue}
+\pnum
+\postconditions
+\tcode{data()} is a non-null pointer that is copyable and can have \tcode{0} added
+to it, \tcode{size()} is \tcode{0}, and \tcode{capacity()} is an unspecified value.
 \end{itemdescr}
 
 \indexlibrary{\idxcode{basic_string}!constructor}%
@@ -1218,21 +1215,16 @@ basic_string(basic_string&& str) noexcept;
 \begin{itemdescr}
 \pnum
 \effects
-Constructs an object of class
-\tcode{basic_string}
-as indicated in Table~\ref{tab:strings.ctr.cpy}.
-In the second form, \tcode{str}
-is left in a valid state with an unspecified value.
+Constructs an object of class \tcode{basic_string}.
 
-\begin{libefftabvalue}
-{\tcode{basic_string(const basic_string\&)} effects}
-{tab:strings.ctr.cpy}
-\tcode{data()}      & points at the first element of an allocated
-                      copy of the array whose first element is
-                      pointed at by \tcode{str.data()}            \\
-\tcode{size()}      & \tcode{str.size()}                          \\
-\tcode{capacity()}  & a value at least as large as \tcode{size()} \\
-\end{libefftabvalue}
+\pnum
+\postconditions
+\tcode{data()} points at the first element of an allocated copy
+of the array whose first element is pointed at by the original
+value \tcode{str.data()}, \tcode{size()} is equal to the
+original value of \tcode{str.size()}, and \tcode{capacity()} is a value
+at least as large as \tcode{size()}.
+In the second form, \tcode{str} is left in a valid state with an unspecified value.
 \end{itemdescr}
 
 \indexlibrary{\idxcode{basic_string}!constructor}%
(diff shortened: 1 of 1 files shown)


tkoeppe:
@mclow, @zygoloid, could you please take a look?

mclow:
I think you lost the former table 61. Shouldn't there be a "Postcondition:" after (what is now) 24.3.2.2/25?

mclow:
On the other hand, maybe table 61 was just in the wrong place before (in N4713).

tkoeppe:
@mclow: The old Table 61 is now paragrah 23.

source/strings.tex
```diff
@@ -1345,25 +1327,17 @@
...
+\tcode{charT} of length \tcode{traits::length(s)}
+whose first element is designated by \tcode{s}.
+
+\pnum
+\postconditions
+\tcode{data()} points at the first element of an allocated copy
+of the array whose first element is pointed at by \tcode{s},
+\tcode{size()} is equal to \tcode{traits::length(s)}, and
+\tcode{capacity()}  is a value at least as large as \tcode{size()}.
 \end{itemdescr}
```
jensmaurer: Duplicate space after capacity()? (Not that it matters.)

jensmaurer:
Looks good to me.

source/strings.tex
tkoeppe: Thanks, will fix.

source/strings.tex
```diff
@@ -1218,21 +1216,16 @@
...
-Constructs an object of class
-\tcode{basic_string}
-as indicated in Table~\ref{tab:strings.ctr.cpy}.
+Constructs an object of class \tcode{basic_string}.
+
+\pnum
+\postconditions
+\tcode{data()} points at the first element of an allocated copy of the
+array whose first element is pointed at by \tcode{str.data()},
+\tcode{size()} is equal to \tcode{str.size()}, and \tcode{capacity()} is
```
zygoloid: I'm not sure this is right. `size()` is equal to the value that `str.size()` had prior to the call, right? `str.size()` is unspecified. (Likewise for `data()`.) I think we're usually more explicit about whether we're talking about the before / after value.

tkoeppe: Would it help if I added "the original value of" in a few places?

source/strings.tex
```diff
@@ -1478,32 +1444,19 @@
...
-modifies
-\tcode{*this}
-as shown in Table~\ref{tab:strings.op=}.
-
-\pnum
-If
-\tcode{*this}
-and \tcode{str} are the same object,
-the member has no effect.
+If \tcode{*this} and \tcode{str} are not the same object, modifies \tcode{*this}.
```
zygoloid: I think we still need to say "as described below" or something here. Otherwise, a reader is left thinking "ok, but how?" Maybe this should be merged into the postconditions instead, though, since the "points at [...] an allocated copy" postcondition is not correct in the `this == &str` case.

source/strings.tex
tkoeppe: @zygoloid: "original value" is what the allocator versions already use. In fact, my attempt to add "original value" to the description here led me to independently invent *identical* wording!

source/strings.tex
tkoeppe: I turned the two conditions around and made the second one say "see below".

source/strings.tex
```diff
@@ -1433,22 +1401,19 @@
 
 \begin{itemdescr}
 \pnum
-\effects Constructs an object of class \tcode{basic_string} as indicated in
-Table~\ref{tab:strings.ctr.6}. The stored allocator is constructed from
-\tcode{alloc}. In the second form, \tcode{str} is left in a valid state with an
+\effects Constructs an object of class \tcode{basic_string}.
+The stored allocator is constructed from \tcode{alloc}.
+In the second form, \tcode{str} is left in a valid state with an
```
zygoloid: For the corresponding constructors without an allocator parameter, this sentence is a postcondition rather than an effect. We should be consistent.

tkoeppe: Done.

source/strings.tex
```diff
@@ -1478,32 +1443,19 @@
...
-element is pointed at by \tcode{str.data()}                                             \\
-\tcode{size()}      &   \tcode{str.size()}                                              \\
-\tcode{capacity()}  &   a value at least as large as \tcode{size()}                     \\
-\end{libefftabvalue}
+\pnum
+\postconditions
+\tcode{data()} points at the first element of an allocated copy
+of the array whose first element is pointed at by \tcode{str.data()},
+\tcode{size()} is equal to \tcode{str.size()}, and
+\tcode{capacity()} is a value at least as large as \tcode{size()}.
```
zygoloid: I think the "If `*this` and `str` are the same object" condition should be moved down here -- the postcondition should ideally be written such that it applies for all (valid) calls to the function.

tkoeppe: Moved or copied? Do you want to delete the Effects clause?

source/strings.tex
```diff
@@ -1477,33 +1442,19 @@
...
-{tab:strings.op=}
-\tcode{data()}      &
-points at the first element of an allocated copy of the array whose first
-element is pointed at by \tcode{str.data()}                                             \\
-\tcode{size()}      &   \tcode{str.size()}                                              \\
-\tcode{capacity()}  &   a value at least as large as \tcode{size()}                     \\
-\end{libefftabvalue}
+\pnum
+\postconditions
+\effects
```
zygoloid: Stray `\effects`?

tkoeppe: Yikes, also stray `pnum` above. (Which prints very strangely.) Fixed.

PR #1878, opened 2017-12-06
[diff.mods.to.headers] De-index non-existent headers

AlisdairM:
Three C11 headers are called out as not being part of C++.
They should not be listed in the index of headers, which
currently points to the two places that explicitly state
these headers do not exist.

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -1926,9 +1926,9 @@ in~\ref{depr.c.headers}, but their use is deprecated in \Cpp{}.
 
 \pnum
 There are no \Cpp{} headers for the C headers
-\tcode{<stdatomic.h>}\indexhdr{stdatomic.h},
-\tcode{<stdnoreturn.h>}\indexhdr{stdnoreturn.h},
-and \tcode{<threads.h>}\indexhdr{threads.h},
+\tcode{<stdatomic.h>},
+\tcode{<stdnoreturn.h>},
+and \tcode{<threads.h>},
 nor are the C headers themselves part of \Cpp{}.
 
 \pnum
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -1129,9 +1129,6 @@ The facilities of the C standard library are provided in the
 additional headers shown in Table~\ref{tab:cpp.c.headers}.%
 \footnote{It is intentional that there is no \Cpp{} header
 for any of these C headers:
-\indexhdr{stdatomic.h}%
-\indexhdr{stdnoreturn.h}%
-\indexhdr{threads.h}%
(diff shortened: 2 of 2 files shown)


jensmaurer:
Ahem. Shouldn't we leave those headers in the main index?

tkoeppe:
@jensmaurer: I'm sympathetic to the idea of *not* indexing things that aren't part of C++. The PR wanted to do that; it was I who renamed the commit message to erroneously refer only to the header index. Sorry for that, I'll force-push and amendment.

jensmaurer:
@tkoeppe: "not indexing things that aren't part of C++": Why? Someone coming from C might want to ask the question "where did stdatomic.h go in C++" and look in the main index for that. Well, even the header index would maybe benefit from a "see" reference.

tkoeppe:
@jensmaurer: ...meh, maybe. I'm not convinced. If someone comes up complaining that they can't figure out what happened to `<stdnoreturn.h>`, I'll gladly reconsider, but I'd rather be rid of this otherwise.

W-E-Brown:
> On Dec 6, 2017, at 8:54 AM, Thomas Köppe <notifications@github.com> wrote:
> 
> @jensmaurer: ...meh, maybe. I'm not convinced. If someone comes up complaining that they can't figure out what happened to <stdnoreturn.h>, I'll gladly reconsider, but I'd rather be rid of this otherwise.

I respectfully but strongly disagree with this position, as it runs counter to my understanding of the purpose of an index.

If I want to find something, I look in the appropriate index with the expectation that, if a term is mentioned in the document, I will find the matching page reference(s).  By selectively omitting terms from indexing, I am at best hindered from locating what I am seeking, and at worst am given the impression that the document does not at all mention the term.

Accordingly, I strongly object to the present decision, and strongly support Jens' position.

Please revert this commit and let us reconsider how best to handle such situations.

tkoeppe:
As discussed with @zygoloid, since these are not C++ headers, they have no place in the C++ header index. I will add re-add main index entries for the headers, though, to make them findable again.

tkoeppe:
I have altered the index.

W-E-Brown:
> On Dec 6, 2017, at 12:55 PM, Thomas Köppe <notifications@github.com> wrote:
> 
> As discussed with @zygoloid, since these are not C++ headers, they have no place in the C++ header index. I will add re-add main index entries for the headers, though, to make them findable again.

That seems optimal to me.  Thank you.

jensmaurer:
@tkoeppe: Thank you.

AlisdairM:
Thanks Thomas - that more accurately reflects my original intent as well.

PR #1879, opened 2017-12-06
[basic.string] State class invariants about [data(), data() + size()] in the introduction.

tkoeppe:


tkoeppe:
LWG, @mclow: This PR is a suggestion to modernize the string wording a little bit, following the removal of the string requirement tables in #1876. I'd like to know whether you're interested in this at all, and whether there are further changes you'd like to see in this area.

source/strings.tex
```diff
@@ -1202,8 +1204,7 @@
 
 \pnum
 \postconditions
-\tcode{data()} is a non-null pointer that is copyable and can have \tcode{0} added
-to it, \tcode{size()} is \tcode{0}, and \tcode{capacity()} is an unspecified value.
+\tcode{capacity()} is an unspecified value.
```
jensmaurer: From where do we know that `size() == 0` here?

jensmaurer:

source/strings.tex
tkoeppe: Oh, yes, that should not have been removed. Fixed.

tkoeppe:
I'm adding "is true" to "`size() <= capacity()` is true", because we don't usually leave boolean values "hanging" like that.

mclow:
Sent to -lib reflector for feedback.

PR #1880, opened 2017-12-06
[conv] Replace \term with \defn or \defnx

jensmaurer:
Partially addresses #329.


zygoloid:
Per comments on #1881, I think we should use "conversion sequence!standard" rather than "sequence!standard conversion".

jensmaurer:
Fixed.

PR #1881, opened 2017-12-06
[over] Replace \term with \defnx to add index entries

jensmaurer:
Partially addresses #329.

### DIFF
--- a/source/overloading.tex
+++ b/source/overloading.tex
@@ -23,9 +23,7 @@ declaration is being referenced is determined by comparing the types
 of the arguments at the point of use with the types of the parameters
 in the overloaded declarations that are visible at the point of use.
 This function selection process is called
-\term{overload resolution}
-and
-is defined in~\ref{over.match}.
+\defn{overload resolution} and is defined in~\ref{over.match}.
 \begin{example}
 
 \indextext{overloading!example of}%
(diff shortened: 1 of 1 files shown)


source/overloading.tex
```diff
@@ -1912,9 +1909,7 @@
 \indextext{overloading!resolution!implicit conversions and|(}
 
 \pnum
-An
-\term{implicit conversion sequence}
-\indextext{sequence!implicit conversion}%
+An \defnx{implicit conversion sequence}{sequence!implicit conversion}
```
zygoloid: Yuck. Can we index this as "conversion sequence!implicit" rather than "sequence!implicit conversion"? Also, a \see from "implicit conversion sequence" would be really useful.

source/overloading.tex
```diff
@@ -2484,18 +2475,18 @@
 \pnum
 This subclause defines a partial ordering of implicit conversion
 sequences based on the relationships
-\term{better conversion sequence}
+\defnx{better conversion sequence}{conversion!sequence!better}
```
zygoloid: I think we should keep "conversion sequence" separate from "conversion" in the index, since they have quite different sets of rules.

source/overloading.tex

source/overloading.tex

jensmaurer:
Addressed @zygoloid's review comments.

PR #1884, opened 2017-12-08
[dcl.struct.bind] Clarify that structured bindings are entities (as contrast to names)

cpplearner:
[basic]/3:

> An *entity* is a value, object, reference, structured binding, function, enumerator, type, class member, bit-field, template, template specialization, namespace, or parameter pack.

The term "structured binding" is consistently used to denote a kind of entity, e.g. [dcl.type.simple]/(4.1):

> if e is an unparenthesized id-expression naming a structured binding, ...

Saying that something names a name would be nonsense.


jensmaurer:
Looks like an improvement to me. Now that structured bindings are entities, we have names referring to these entities, but the concept of "name" is different from the concept of "entity". So we shouldn't mix those two.

PR #1885, opened 2017-12-08
[defns.undefined] Constant expressions can have "library UB"

geoffromer:


jwakely:
This looks good to me. I think it qualifies as editorial; it's the intended meaning of that wording, and without this change the standard is internally inconsistent because it's not actually possible to diagnose many types of "library UB" that arise from violating _Requires:_ preconditions.

jensmaurer:
[expr.const] p2.6 already limits the undefined behavior rule to the core language sections of the standard, so this seems editorial to me.

PR #1886, opened 2017-12-10
[string.view.template] Add reference to [string.view.find] …

K-ballo:
…in basic_string_view synopsis


cpplearner:
dup of #1883 ?

K-ballo:
> dup of #1883 ?

Indeed.

PR #1891, opened 2017-12-22
Use 'one of' formulation for binary-digit

AlisdairM:
The grammar for literals consistently prefers using the 'one of' a set
formulation for character sets, rather than listing each character as
an option.  This holds for other parts of this grammar, even when the
set of options is just two items, such as hexadecimal-prefix,
unsigned-suffix, or long-suffix.

As binary-digit is the odd one out, this simple update makes it
consistent with the rest of the grammar specifications in this
clause.

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -869,9 +869,8 @@ ISO C. }
 \end{bnf}
 
 \begin{bnf}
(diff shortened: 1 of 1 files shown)


lichray:
+1

PR #1892, opened 2017-12-27
[basic.scope.class,basic.lookup.unqual] Use the term "default member initializer"

cpplearner:


jensmaurer:
Looks good to me. As an add-on, I'd mildly prefer adding a cross-reference to [class.mem]. For terms referenced from far away, that seems to be a service to the reader, highlighting we're talking about a defined term here.

PR #1894, opened 2017-12-30
[expr.const] Fix no longer correct comment

cpplearner:


cpplearner:
Actually, I don't understand this item. Why is a expression in a lambda-expression not a constant expression if it contains "a reference to `this` or to a variable with automatic storage duration defined outside that lambda-expression, where the reference would be an odr-use"?

(when) is this item still necessary or useful today? Is there any implementation difficulty or something, other than this item, that makes it impossible to permit the initialization of `j`?

PR #1895, opened 2018-01-02
[variant.visit] qualify std::forward for consistency

CaseyCarter:
...because Library always qualifies `move` and `forward`.

Also, my editor hates extra newlines at end-of-file.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -4845,8 +4845,10 @@ The expression in the \effects element shall be a valid expression of the same
 
 \pnum
 \effects
-Let \tcode{is...} be \tcode{vars.index()...}. Returns \tcode{\placeholdernc{INVOKE}(forward<Visitor>(vis), get<is>(}\brk{}
-\tcode{forward<Variants>(vars))...);}\iref{func.require}.
+Let \tcode{is...} be \tcode{vars.index()...}. Returns:
+\begin{codeblock}
+@\placeholdernc{INVOKE}@(std::forward<Visitor>(vis), get<is>(std::forward<Variants>(vars))...) // see \ref{func.require}
+\end{codeblock}
 
 \pnum
 \remarks
@@ -19782,4 +19784,3 @@ closest to the value of the string matching the pattern.
 \end{itemdescr}
 
 \xref ISO C 7.22.1.3, 7.22.1.4
-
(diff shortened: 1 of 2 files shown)


tkoeppe:
Good in principle; have you messed around with the linebreaking at all to see if the manual hints can be removed?

tkoeppe:
Could you please add a Travis check for trailing newlines to prevent future regressions?

CaseyCarter:
> Good in principle; have you messed around with the linebreaking at all to see if the manual hints can be removed?

I had not. The `\tcode` is almost a full column width long now, I think it looks better as a code block:

image

> Could you please add a Travis check for trailing newlines to prevent future regressions?

Sure.

source/utilities.tex
```diff
@@ -4845,8 +4845,10 @@
 
 \pnum
 \effects
-Let \tcode{is...} be \tcode{vars.index()...}. Returns \tcode{\placeholdernc{INVOKE}(forward<Visitor>(vis), get<is>(}\brk{}
-\tcode{forward<Variants>(vars))...);}\iref{func.require}.
+Let \tcode{is...} be \tcode{vars.index()...}. Returns:
+\begin{codeblock}
+@\placeholdernc{INVOKE}@(std::forward<Visitor>(vis), get<is>(std::forward<Variants>(vars))...); // see \ref{func.require}
```
tkoeppe: Please remove the semicolon. It has always been a mistake; we return expressions, not statements. Other instances of Returns: plus codeblock also never use a semicolon.

I could further question whether this shouldn't be a Returns: element rather than an Effects: one, but let's leave that as is for now.

CaseyCarter: Done. I had similar thoughts about this being an odd use of `Effects`, and also decided it wasn't worth touching for now.

tkoeppe: That's fine, leave it with me. I'll discuss it with Richard.

jwakely: Isn't the semi-colon consistent with the third form at https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#writing-effects-in-a-function-description ?

tkoeppe: @jwakely: Only if we actually said `return`. But we don't. But I think we should change this to use a *Returns*: element anyway, non?

PR #1896, opened 2018-01-03
[dcl.enum] use indefinite article before "fixed underlying type"

jwakely:
Everywhere else we say that an enumeration type has _a_ fixed underlying type.


jensmaurer:
Looks good to me.

PR #1898, opened 2018-01-09
[namespace.udir] Ambiguous lookup results are covered in [basic.lookup]

jensmaurer:
Relegate a duplicate normative statement in [namespace.udir]
to a note and add a cross-reference to [basic.lookup].


jensmaurer:
The previous change 487e870347a4c76c9b6dd664c4e5671e709c81d8 cleaned up 6.4p1, to which we now refer.

PR #1904, opened 2018-01-29
Update the description for array.data

mclow:
Make it match the one for `vector`.
Compare [vector.data]/1 with the proposed [array.data]/1

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -3254,8 +3254,8 @@ constexpr const T* data() const noexcept;
 
 \begin{itemdescr}
 \pnum \returns
-A pointer such that \tcode{data() == addressof(front())},
-and \range{data()}{data() + size()} is a valid range.
+A pointer such that \range{data()}{data() + size()} is a valid range. For a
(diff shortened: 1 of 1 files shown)


jwakely:
LGTM. Note that [array.zero] covers the empty (N==0) case.

tkoeppe:
This seems legit. The current phrasing is unfortunate in that the single-pass reading requires UB for the `addressof(front())` case, and you don't find out until later that that case is actually different. The proposed change makes it rather clear where you can go if you care about size zero.

mclow:
> Note that [array.zero] covers the empty (N==0) case.

Almost. This section codifies that `[data(), data() + size())` is a valid range, for all sizes.

mclow:
I wish we could find a place to say this in `[string.view]`, but I think that would NOT be editorial.

jwakely:
Right, I should have said that [array.zero] gives the exception to the `data() == addressof(front())` rule. I think the proposed change is an improvement because we don't imply it's undefined (due to calling `front()`), then change our mind later in [array.zero].

PR #1905, opened 2018-01-30
Fix references to container requirements tables.

jdennett:
LWG issue 704 ("MoveAssignable requirement for container value type overly strict") added a reference from [associative.reqmts] to what was at the time table 93 ("Allocator-aware container requirements").

That table is now table 86, but the cross reference is now (incorrectly) to table 83 ("Container requirements").  It should still refer to the Allocator-ware container requirements table.  This patch changes it to do so, matching what the committee voted in (many years ago).


jensmaurer:
Looks good to me.

PR #1907, opened 2018-02-05
Harmonize wording about reallocation invalidating pointers in `vector`

Quuxplusone:
`reserve`, `shrink_to_fit`, and `push_back/emplace_back/insert`
all use versions of the same wording. This eliminates unintentional
differences between their wordings.

Prompted by Hubert Tong's mail "[isocpp-lib] Validity of vector
iterators after non-reallocating insertion".

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -3635,8 +3635,9 @@ of a non-\tcode{CopyInsertable} \tcode{T} there are no effects.
 \complexity Linear in the size of the sequence.
 
 \pnum
-\remarks \tcode{shrink_to_fit} invalidates all the references, pointers, and iterators
-referring to the elements in the sequence as well as the past-the-end iterator.
+\remarks
+\tcode{shrink_to_fit} invalidates all the references, pointers, and iterators
+referring to the elements in the sequence, as well as the past-the-end iterator.
 \end{itemdescr}
 
 \rSec3[deque.modifiers]{\tcode{deque} modifiers}
@@ -5413,7 +5414,8 @@ may throw an appropriate exception.}
 \pnum
 \remarks
 Reallocation invalidates all the references, pointers, and iterators
-referring to the elements in the sequence.
+referring to the elements in the sequence, as well as the past-the-end iterator.
+\begin{note} If no reallocation takes place, they remain valid. \end{note}
 No reallocation shall take place during insertions that happen
 after a call to
 \tcode{reserve()}
@@ -5445,9 +5447,10 @@ of a non-\tcode{CopyInsertable} \tcode{T} there are no effects.
 \complexity Linear in the size of the sequence.
 
 \pnum
-\remarks Reallocation invalidates all the references, pointers, and iterators
-referring to the elements in the sequence as well as the past-the-end iterator.
-If no reallocation happens, they remain valid.
+\remarks
+Reallocation invalidates all the references, pointers, and iterators
+referring to the elements in the sequence, as well as the past-the-end iterator.
+\begin{note} If no reallocation takes place, they remain valid. \end{note}
 \end{itemdescr}
 
 \indexlibrary{\idxcode{swap}!\idxcode{vector}}%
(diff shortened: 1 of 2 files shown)


jensmaurer:
I would prefer to unify all "If /reallocation/ occurs, this happens" sentences and put a merged general sentence in the front matter for `vector`. Then, the individual functions can simply guarantee extra properties for the "if no reallocation occurs..." cases. Maybe something like "Reallocation (xref) may occur; if no reallocation occurs..."

source/strings.tex
```diff
@@ -1749,9 +1749,10 @@
 \complexity Linear in the size of the sequence.
 
 \pnum
-\remarks Reallocation invalidates all the references, pointers, and iterators
-referring to the elements in the sequence as well as the past-the-end iterator.
-If no reallocation happens, they remain valid.
+\remarks
+Reallocation invalidates all the references, pointers, and iterators
+referring to the elements in the sequence, as well as the past-the-end iterator.
+If no reallocation takes place, they remain valid.
```
Quuxplusone: @jensmaurer: Your comment about "front matter for `vector` reminded me that `shrink_to_fit` is also a member function of `basic_string` (as well as `deque`, which I already hit, above).

source/containers.tex
```diff
@@ -5551,8 +5554,11 @@
 \remarks
 Causes reallocation if the new size is greater than the old capacity.
 Reallocation invalidates all the references, pointers, and iterators
-referring to the elements in the sequence.
-If no reallocation happens, all the iterators and references before the insertion point remain valid.
+referring to elements in the sequence, as well as the past-the-end iterator.
+If no reallocation happens, all references, pointers, and iterators
+referring to elements before the insertion point remain valid; all other
+references, pointers, and iterators referring to elements in the sequence,
+as well as the past-the-end iterator, are invalidated.
```
zygoloid: Are the changes here justified by existing normative wording, or merely by this being the obvious (but unstated) intent of the existing rules?

Quuxplusone: The addition of "as well as the past-the-end iterator" is justified by (1) harmonizing with the existing normative wording for "reallocation" given in `shrink_to_fit`, and (2) in practice, vendors do invalidate it.
The addition of "all other foos are invalidated" is justified by (1) assuming that this merely clarifies an implication-by-omission in the old wording, and (2) in practice, vendors do invalidate them.

IMHO this is good enough justification to count as "editorial," but obviously you'd know better than I would. :)

zygoloid:
I'm pretty sure this is right, but let's ask LWG to take a quick look just in case.

source/containers.tex
```diff
@@ -5551,8 +5554,11 @@
 \remarks
 Causes reallocation if the new size is greater than the old capacity.
 Reallocation invalidates all the references, pointers, and iterators
-referring to the elements in the sequence.
-If no reallocation happens, all the iterators and references before the insertion point remain valid.
+referring to elements in the sequence, as well as the past-the-end iterator.
+If no reallocation happens, all references, pointers, and iterators
```
CaseyCarter: This bit is normative duplication.

Quuxplusone: Yes, but it's part of a sentence where the second part is *not* duplication. Can you think of a way to re-word it?

CaseyCarter:
```tex
Invalidates all references, pointers, and iterators from the insertion point forward, including the
past-the-end iterator. If the new size is greater than the old capacity, causes reallocation which
additionally invalidates all the references, pointers, and iterators referring to elements in the
sequence which precede the insertion point. \begin{note} When no reallocation occurs,
references, pointers, and iterators before the insertion point remain valid. \end{note} If an
exception is thrown ...
```

Quuxplusone: I guess I'm concerned about the potential for misreading "from the insertion point" as exclusive rather than inclusive. (Also, "forward" might better be "onward"? I guess that doesn't matter as it's hard to misread.)
My legalistic eye doesn't like how Casey's wording drops the "referring to elements" part, or uses "including" instead of "as well as"... but nonetheless, I would be fine with Casey's wording making it into the standard.

CaseyCarter: By all means, suggestive alternatives. We do WG21 and ourselves a favor by banging out wording upon which we can agree before asking LWG to rubber stamp it.

Quuxplusone:
Abandoning in favor of the new P/R on LWG 3077.

PR #1908, opened 2018-02-07
Fix decay index redirects.

Eelis:
The mismatch currently causes broken links at [eel.is/c++draft/generalindex](http://eel.is/c++draft/generalindex#:decay).


jensmaurer:
Simple indexing fix. Looks good to me.

PR #1913, opened 2018-02-13
[contents,fs.req.namespace] Qualify only namespace-scope names LWG2818

### LINKED ISSUE #1793: I do not think [fs.req.namespaces] means what it thinks it means LWG2818
>  Unless otherwise specified, references to entities described in this subclause are assumed to be qualified with `::std::filesystem::`.

"this subclause" only refers to 30.10.4.1 [fs.req.namespaces], but it really means the whole of 30.10 [filesystems], or at least all the subclauses after 30.4.2.1, not just 30.10.4.1 itself.

Do we really mean _all_ entities? Variables are entities. Class members are entities, and we don't mean to qualify all class members with the namespace name, because e.g. `native()` means `::std::filesystem::path::native()` not `::std::filesystem::native()` (which doesn't exist). I think we mean types and non-member functions. And maybe some typedefs, which are not entities.

20.5.1.1 [contents] specifies a similar rule differently, talking about "a name `x`" not "entities".

We could qualify all uses of types and non-member functions with `filesystem::` so that the general rule in [contents] applies, which might hurt readability, or we could fix [fs.req.namespaces].

jensmaurer:
It does not make sense to attempt to qualify class members
with a nested-name-specifier consisting of namespace-names.

Fixes #1793.

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -927,7 +927,7 @@ support multiple configurations of the library.
 \end{footnote}
 
 \pnum
-Whenever an unqualified name is used
+Whenever an unqualified name other than \tcode{swap} is used
 in the specification of a declaration \tcode{D}
 in \ref{\firstlibchapter} through \ref{\lastlibchapter} or \ref{depr},
 its meaning is established
@@ -946,6 +946,9 @@ The reference to \tcode{is_array_v} in the specification of \tcode{std::to_array
 Operators in expressions\iref{over.match.oper} are not so constrained;
 see \ref{global.functions}.
 \end{note}
+The meaning of the unqualified name \tcode{swap} is established
+in an overload resolution context
+for swappable values\iref{swappable.requirements}.
 
 \rSec3[headers]{Headers}
 


jensmaurer:
Editorial meeting: What if there are nested namespaces? What if this is ambiguous (different namespaces)?

jensmaurer:
I have rebased the patch, but it does not yet implement the "editorial meeting" outcome. This patch is currently not ready to be merged.

zygoloid:
I would like to make sure that LWG has a chance to look at this and comment on it before it's applied.

zygoloid:
We appear to have quite a few places that fail to satisfy this "unless otherwise specified" rule (these are pre-existing errors, but still):

[optional.swap]'s Table 43 says "calls `swap(*(*this), *rhs)` without saying that ADL is performed
[variant.swap]/2.2 likewise says 'calls `swap(...)`' without saying that ADL is performed, and similarly in /3 and /4
[queue.defn]'s inline definition of `std::queue<T>::swap` is:
```
void swap(queue& q) noexcept(is_nothrow_swappable_v<Container>)
  { using std::swap; swap(c, q.c); }
```
... which is wrong, and likewise for [priqueue.overview] and [stack.defn].
[alg.swap]/2 likewise says `swap(*(first1 + n), *(first2 + n))`, and [alg.swap]/6 says `swap(*a, *b)` with no invocation of ADL.

[container.node.modifiers] likewise uses `swap` without a "use ADL" proviso, but I think that one is actually correct: we don't want user code interfering with `node_handle` swaps, but user types can absolutely be associated classes of that lookup

The definition of `is_swappable_with` deals with this by saying "in an overload-resolution context for swappable values (15.5.3.2)", and looks like the *only* place we get this right other than the definition of the "swappable" requirements and `Swappable` concepts.

All this makes me think that perhaps we should move the "swap is found by ADL" rule into the front matter: "Unless otherwise specified, argument-dependent name lookup is only performed within the standard library for *operator-function-id*s and for the *identifier* `swap`." -- and we should change [container.node.modifiers] to explicitly specify that ADL is not performed. Though in the short term, we could instead update the places listed above to say "in an overload-resolution context for swappable values".

jensmaurer:
I've added the front-matter talk about "swap", which results in fewer changes than talking about ADL everywhere "swap" appears.

@jwakely, could you have LWG look at this?

jensmaurer:
Editorial 2021-04-30: Remove iostreams.tex changes; they are covered by LWG2818.

source/iostreams.tex
```diff
@@ -11541,12 +11539,6 @@
 Template parameters named \tcode{Allocator} shall meet the
 \oldconcept{Allocator} requirements (\tref{cpp17.allocator}).
 
-\rSec3[fs.req.namespace]{Namespaces and headers}
```
jwakely, line 11544: This part is present in the lwg2818 resolution, so not needed in the editorial changes, right?

source/lib-intro.tex
```diff
@@ -900,14 +900,19 @@
...
 
 \pnum
-Whenever a name \tcode{x} defined in the standard library is mentioned,
-the name \tcode{x} is assumed to be fully qualified as
-\tcode{::std::x},
-unless explicitly described otherwise. For example, if the \effects element
-for library function \tcode{F} is described as calling library function \tcode{G},
-the function
-\tcode{::std::G}
-is meant.
```
jwakely, line 910: This part is already being removed by LWG 2818 so I think this editorial change would be clearer with just the addition below (possibly rebased on 2818).

source/lib-intro.tex
```diff
@@ -900,14 +900,19 @@
...
+within the standard library
+\begin{itemize}
+\item
+for \grammarterm{operator-function-id}s and
+\item
+for the \grammarterm{identifier} \tcode{swap},
+in an overload-resolution context for swappable values\iref{swappable.requirements}.
+\end{itemize}
+\begin{note}
+An implementation could use \grammarterm{qualified-id}{s} to achieve that.
```
jwakely, line 914: I think it's a bit unclear what "that" refers to here. Maybe "Using a _qualified-id_ avoids argument-dependent lookup." instead?

jwakely (changes requested):

source/iostreams.tex
jensmaurer: I think this entire pull request is mostly moot with LWG 2818 applied. I'll wait until that one has landed before reviewing for residual cleanup.

tkoeppe: LWG 2818 has landed!

jensmaurer:
The only residual open issue is the special treatment of `swap`. See the updated diff.

jwakely:
See also [LWG 3629: `make_error_code` and `make_error_condition` are customization points](https://cplusplus.github.io/LWG/issue3629).

source/lib-intro.tex
```diff
@@ -946,6 +946,9 @@
 Operators in expressions\iref{over.match.oper} are not so constrained;
 see \ref{global.functions}.
 \end{note}
+The meaning of the unqualified name \tcode{swap} is established
+in an overload resolution context
+for swappable values\iref{swappable.requirements}.
```
Quuxplusone, line 952: On line 943, there's an example for `is_array_v`, which gives help where none is needed (type and variable names never change their lookup results based on qualification anyway); but there is no example given for line 937, which is where help _is_ needed (because it is the function lookup rules that are complicated already, and which are being changed by this wording). I suggest adding a new note, e.g.
```
\begin{note}
The reference to \tcode{swap_ranges} in the specification of \tcode{std::swap}\iref{utility.swap}
refers to \tcode{::std::swap_ranges}.
\end{note}
```

jwakely: Given that `swap` is the special case called out in this wording, using the definition of `std::swap` for an example where the special case doesn't apply can only lead to confusion.

jwakely: And I don't think this is needed anyway. The rule is consistent: unless specified otherwise, we do qualified lookup. Function or not, you do qualified lookup. So the extra complication that applies to lookup for functions (ADL) isn't relevant, we're explicitly not doing that.

Quuxplusone: > can only lead to confusion...

...or clarification. ;)

The rule from the implementor's point of view is "Variable and type names are treated as written (e.g. when the paper standard says `is_same_v`, the implementor should write `is_same_v` in the code). Function names get special treatment (e.g. when the paper standard says `swap_ranges`, the implementor should write `std::swap_ranges` in the code)." The extra complication exists _only_ for function names, and yet that's the one case where no clarifying example is provided.

jensmaurer:
> when the paper standard says swap_ranges, the implementor should write std::swap_ranges in the code

This is not accurate:
```
namespace N {
  namespace std {
    void f();
  }
}

using namespace N;

namespace std {
  void f();
  void g() { std::f(); }
}
```
x.cc: In function ‘void std::g()’:
x.cc:12:14: error: reference to ‘std’ is ambiguous
   12 |   void g() { std::f(); }
      |              ^~~
x.cc:3:13: note: candidates are: ‘namespace N::std { }’
    3 |   namespace std {
      |             ^~~
<built-in>: note:                 ‘namespace std { }’
``` 

So, the implementer needs to use `::std::function_name()` to disambiguate properly.
(If the function is not overloaded or a template, you might also get away with something like `(+f)()` (convert to a function pointer first, which doesn't use ADL).)

source/lib-intro.tex
Quuxplusone: @jensmaurer wrote:
> [Actually] the implementer needs to use [`::std::swap_ranges`] to disambiguate properly.

True. To rephrase:

> The rule from the implementor's point of view is "Variable and type names are treated as written. Function names get special treatment." The extra complication exists _only_ for function names, and yet that's the one case where no clarifying example is provided.

Proving that the special treatment of functions is _even more special_ than previously indicated, merely strengthens the force of the top-level statement.

PR #1914, opened 2018-02-13
[expr.typeid, time.syn, class.access.base] Replace \term with \placeholder where appropriate.

Eelis:
No visible difference.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -2770,7 +2770,7 @@ The result of a \tcode{typeid} expression is an lvalue of static type
 \indextext{\idxcode{type_info}}%
 \indexlibrary{\idxcode{type_info}}%
 \tcode{const} \tcode{std::type_info}\iref{type.info} and dynamic type \tcode{const}
-\tcode{std::type_info} or \tcode{const} \term{name} where \term{name} is an
+\tcode{std::type_info} or \tcode{const} \placeholder{name} where \placeholder{name} is an
 \impldef{derived type for \tcode{typeid}} class publicly derived from
 \tcode{std::type_info} which preserves the behavior described
 in~\ref{type.info}.\footnote{The recommended name for such a class is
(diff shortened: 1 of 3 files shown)


source/expressions.tex
```diff
@@ -2770,7 +2770,7 @@
 \indextext{\idxcode{type_info}}%
 \indexlibrary{\idxcode{type_info}}%
 \tcode{const} \tcode{std::type_info}\iref{type.info} and dynamic type \tcode{const}
-\tcode{std::type_info} or \tcode{const} \term{name} where \term{name} is an
+\tcode{std::type_info} or \tcode{const} \placeholder{name} where \placeholder{name} is an
```
jensmaurer: In general, I think we should use "\placeholder" for math-style single-letter variables, so these should probably we switched to "N" or "C". (The latter preferred; "N" is a namespace.)

source/utilities.tex
```diff
@@ -17543,12 +17543,12 @@
...
-    using milliseconds = duration<@\term{signed integer type of at least 45 bits}@, milli>;
-    using seconds      = duration<@\term{signed integer type of at least 35 bits}@>;
-    using minutes      = duration<@\term{signed integer type of at least 29 bits}@, ratio<  60>>;
-    using hours        = duration<@\term{signed integer type of at least 23 bits}@, ratio<3600>>;
+    using nanoseconds  = duration<@\placeholder{signed integer type of at least 64 bits}@, nano>;
+    using microseconds = duration<@\placeholder{signed integer type of at least 55 bits}@, micro>;
+    using milliseconds = duration<@\placeholder{signed integer type of at least 45 bits}@, milli>;
+    using seconds      = duration<@\placeholder{signed integer type of at least 35 bits}@>;
+    using minutes      = duration<@\placeholder{signed integer type of at least 29 bits}@, ratio<  60>>;
+    using hours        = duration<@\placeholder{signed integer type of at least 23 bits}@, ratio<3600>>;
```
jensmaurer: Sigh. Yeah, until we find better phrasing.

jensmaurer:
@eelis: ping.

Issue #1771, opened 2017-10-16
Hanging paragraphs vs. warm-up introduction


jensmaurer:
The ISO Directives, Part 2, require us to get rid of hanging paragraphs.

One sub-issue is that we sometimes have introductory warm-up sentences at the start of a clause or subclause. Occasionally, these are "note"s, making clear their non-normative character. On other occasions (e.g. [constraints.overview]), they are not. Sometimes those warm-up sentences are in a separate subclause "overview", sometimes they form "hanging paragraphs" (often together with normative material).

This issue serves as an anchor to discuss a uniform approach to dealing with these warm-up sentences. Ideas:
- Always put warm-up descriptions into a separate "overview" section (helps avoiding hanging paragraphs).
- Always make warm-up sentences a note.
- Find a differing typographical convention.
- Add warm-up overviews to subclauses which don't have them (yet).

jensmaurer:
See also #574.

jensmaurer:
Editorial meeting consensus: There are several different situations that warrant different treatment.
- Hanging paragraphs containing substantial normative wording (e.g. Clause 8 [expr]): Reconsider the clause structure in its entirety, move the hanging paragraphs into fresh subclauses, properly structuring the contents. For [expr], have "Properties of expressions" (with more specific subclauses), "Kinds of expressions" (current grammar tree), "Constant expressions".
- Non-normative introductory text as well as (short) normative snippets (e.g. defining a term or similar) should go into a "Preamble" subclause. Use the usual "note" mechanism for non-normative content. (Note: @tkoeppe opposes having a third way of identifying non-normative content (e.g. bars on the side or indentation or similar).) Start a new paragraph once normative content begins. Never mark an entire subclause as "informative".
- Non-normative paraphrase of the contained subclause headings: Delete.
- Having a short paragraph of non-normative introduction before a grammar snippet or a class synopsis is fine. In particular, 23.2.1p1 should be above the synopsis; merge 23.2p1, which also goes before the synopsis; mark as non-normative.

jensmaurer:
Hanging paragraphs were summarily fixed by addressing an ISO/CS comment on DIS C++20. Keeping open to possibly reconsider the details of the new structure.

JohelEGP:
[...]

PR #1920, opened 2018-02-14
[utility] Make content in hanging paragraph non-normative.

jensmaurer:
Partially addresses #1771.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -35,12 +35,19 @@ These utilities are summarized in Table~\ref{tab:util.lib.summary}.
 \rSec1[utility]{Utility components}
 
 \pnum
+\begin{note}
 This subclause contains some basic function and class templates that are used
 throughout the rest of the library.
+\end{note}
 
 \indexhdr{utility}%
 \indexlibrary{\idxcode{rel_ops}}%
 \rSec2[utility.syn]{Header \tcode{<utility>} synopsis}
+
+\pnum
+The header \tcode{<utility>} defines several types and function templates
+that are described in this Clause. It also defines the template \tcode{pair}
+and various function templates that operate on \tcode{pair} objects.
 
 \begin{codeblock}
 #include <initializer_list>     // see \ref{initializer_list.syn}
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -35,13 +35,20 @@
...
+\end{note}
 
 \indexhdr{utility}%
 \indexlibrary{\idxcode{rel_ops}}%
 \rSec2[utility.syn]{Header \tcode{<utility>} synopsis}
 
+\pnum
+The header \tcode{<utility>} defines several types and function templates
+that are described in this Clause. It also defines the template \tcode{pair}
+and various function templates that operate on \tcode{pair} objects.
```
tkoeppe: What on earth is the special call-out to `pair` achieving here? The header also defines `in_place_index` and `index_sequence_for`. Maybe this sentence came from an older time when `pair` really was the only class template here?

jensmaurer: This is pre-existing text that just got moved (from below the synopsis). There are similar introductory notes for most of the header synopses. If you feel this particular intro text warrants improvement, feel free to suggest something in a separate pull request.

tkoeppe: OK, let me just keep this open so as to not forget.

@zygoloid, @jwakely: do you think we should list all the other class templates explicitly in this strange little sentence, or do something more radical (such as removing it altogether)?

jensmaurer:
Rebased.

source/utilities.tex
zygoloid: I'd prefer we just drop this entirely. If someone wants to know what's in <utility>, that's what the synopsis is for. I don't see any point in including a paragraph that says "The header <utility> includes the stuff we say it includes in the synopsis."

zygoloid: Maybe we could ditch this and move the hanging paragraph into this section, replacing "This subclause" with "The header \<utility>".

zygoloid:
Merged with @tkoeppe's suggestion and committed as 63dd5c67.

PR #1921, opened 2018-02-14
[rand.req.eng] Use \text, not \mbox, in math mode

### LINKED ISSUE #1344: Replace \mbox with \text where appropriate
E.g. in [numerics.tex#L1915](https://github.com/cplusplus/draft/blob/master/source/numerics.tex#L1915)

jensmaurer:
Fixes #1344.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -1903,21 +1903,21 @@ according to \ref{strings} and \ref{input.output}.
     with the same initial state
     as all other default-constructed engines
     of type \tcode{E}.
-  & \bigoh{$\mbox{size of state}$}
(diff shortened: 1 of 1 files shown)


tkoeppe:
Does this produce a visual difference?

jensmaurer:
No.

tkoeppe:
OK, thanks. For the future, beware that `\text` respects the ambient italicization state (e.g. from `\emph`), so there are situations were you'd prefer `\mathrm` over `\text`. But it's fine here.

Eelis:
Aww, this would've been a perfect opportunity to also fix #1340

PR #1924, opened 2018-02-14
[insert.iterators] Dissolve single-item subclauses.

jensmaurer:
Partially addresses #1429.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -1755,9 +1755,7 @@ namespace std {
 }
 \end{codeblock}
 
-\rSec3[back.insert.iter.ops]{\tcode{back_insert_iterator} operations}
-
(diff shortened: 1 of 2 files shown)


source/iterators.tex
```diff
@@ -1755,9 +1755,7 @@
 }
 \end{codeblock}
 
-\rSec3[back.insert.iter.ops]{\tcode{back_insert_iterator} operations}
-
-\rSec4[back.insert.iter.cons]{\tcode{back_insert_iterator} constructor}
+\rSec4[back.insert.iter.ops]{\tcode{back_insert_iterator} operations}
```
tkoeppe: @jensmaurer: This is weird: we're jumping from rSec2 straight to rSec4?

tkoeppe: Oh no, never mind, there's an rSec3.

PR #1927, opened 2018-02-15
[over.match.oper] Add a note for conversions on synthesized candidates.

### LINKED ISSUE #1926: [over.match.oper] p6.2 Add a note for the synthesized <=> candidate
```
struct X {
  /* something */ operator<=>(const X&);   // member, compares X
};
X x;

struct Y : X { };
Y y;
```

`x == x` (obviously) works.
`x == y` gets us  `(x <=> y) == 0`   which works (conversion on second param)

`y == x`:  We check whether `0 == (x <=> y)` works (yes, it does), so we add the
synthesized candidate

      op<=>(const X&, const X& implicit_obj_param)

 to the candidate set for `==`.  The total candidate set for `y==x` is

* `op<=>(const X& implicit_obj_param, const X&);`
* `op<=>(const X&, const X& implicit_obj_param);    // #2`

Then, we match `y==x` against this and we convert `y` to `X` (first param of #2).

What bothers me a bit is that the property "implicit object param" (which prevents conversions on that operand) is not explicitly mentioned as being swapped in 16.3.1.2p6.2.

jensmaurer:
Fixes #1926.

### DIFF
--- a/source/overloading.tex
+++ b/source/overloading.tex
@@ -497,7 +497,7 @@ The set of candidate functions can contain both member and non-member
 functions to be resolved against the same argument list.
 So that argument and parameter lists are comparable within this
 heterogeneous set, a member function is considered to have an
-extra parameter, called the
+extra first parameter, called the
 \defn{implicit object parameter},
 which represents the object for which the member function has been
 called.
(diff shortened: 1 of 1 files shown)


hsutter:
👍 And this seems editorial.

zygoloid:
[over.match.funcs]p3: "Since arguments and parameters are associated by position within their respective lists, the convention is that the implicit object parameter, if present, is always the first parameter and the implied object argument, if present, is always the first argument."

That's normative wording, and contradicts this note. The note reflects the obvious intent of the wording, though, and the lack of an update to the above wording is just an oversight, so I'm OK with changing this editorially if there's no CWG opposition. Here's what I suggest:

 * Update the wording to also change the above paragraph.
 * Add the CWG label to request that CWG check they're OK with us changing this editorially.

jensmaurer:
CWG in Jacksonville: approved.

jensmaurer:
Rebased.

PR #1930, opened 2018-02-17
[intro.abstract] Change example for unspecified behavior.

### LINKED ISSUE #1742: Wrong example of unspecified behavior in [intro.abstract]/3
> Certain other aspects and operations of the abstract machine are described in this International Standard as unspecified (for example, evaluation of expressions in a *new-initializer* if the allocation function fails to allocate memory ([expr.new])). 

The example is no longer correct after [P0145](https://wg21.link/P0145).

jensmaurer:
Fixes #1742.


jensmaurer:
CWG in Jacksonville: approved.

PR #1931, opened 2018-02-17
[stmt.while] Generalize the equivalence for a declaration as the condition.

### LINKED ISSUE #1485: [stmt.while] Equivalence given for "while (T t = x)" should also cover "while (T t{...})"?
6.5.1/2 states:

> A while statement of the form ``while (T t = x) statement``    is equivalent to [...]

These days a declaration condition can also have the form ``T t{...}``. Presumably the equivalence is also intended to apply when the condition has that form. In that case, I suppose this might meet the criterion for an editorial issue:

> Problems with the presentation of the specification, which, when fixed, would not alter the (intended) meaning of the specification.

But I'm not sure :)

jensmaurer:
Fixes #1485.

### DIFF
--- a/source/statements.tex
+++ b/source/statements.tex
@@ -507,18 +507,14 @@ substatement.
 When the condition of a \tcode{while} statement is a declaration, the scope of
 the variable that is declared extends from its point of
 declaration\iref{basic.scope.pdecl} to the end of the \tcode{while}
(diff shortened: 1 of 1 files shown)


jensmaurer:
CWG in Jacksonville: add comment to example "condition" saying "declares t".

jensmaurer:
Example updated as discussed by CWG. This is ready for merge.

Issue #1453, opened 2017-02-09
[expr.dynamic.cast] CWG 2365: Incorrect wording regarding dynamic_cast


introom:
In paragraph 4, http://eel.is/c++draft/expr.dynamic.cast#4

If the value of v is a null pointer value in the pointer case, the result is the null pointer value of type T.

And in paragraph6,

Otherwise, v shall be a pointer to or a glvalue of a polymorphic type.

So the *otherwise* means the *v* in p4 doesn't have to be polymorphic.

However, this is not possible because the compiler can only check if "v" is a null pointer at runtime.  If "v" is not a null pointer, and unluckily has no RTTI, then the runtime could do nothing meaningufl, ill-formed?

BTW, Clang and GCC requires the "v" in p4 to be of pointer to polymorphic type.

cpplearner:
Do you mean this:

    struct T1 {};
    struct T2 {};
    T1* p = nullptr;
    dynamic_cast<T2*>(p);

This appears to be valid according to p4, but will not be valid if `p` is not a null pointer.

jensmaurer:
This does not seem to be editorial; the normative wording is nonsensical here. Submitted as a CWG issue: http://lists.isocpp.org/core/2017/02/1733.php .

jensmaurer:
CWG in Jacksonville: treat as a core issue

jensmaurer:
CWG in Jacksonville: To be handled as a core issue.

jensmaurer:
Resolved with commit e81049e5b37fbb38317c3385c3f4742381080cd3.

PR #1936, opened 2018-02-19
[expr.dynamic.cast] Remove redundant statements on casting away constness

### LINKED ISSUE #1934: [expr.dynamic.cast] cv-qualification cleanups
[expr.dynamic.cast]p1 says "The `dynamic_cast` operator shall not cast away constness." That makes some of the later wording redundant. Particulary p5:

"In both the pointer and reference cases, the program is ill-formed if cv2 has greater cv-qualification than cv1 or if B is an inaccessible or ambiguous base class of D."

(The "greater cv-qualification" part is pointless, and suggests incorrectly that this condition does not apply in the other cases in scope in this section.)

Also, p3 need not repeat the "same or more cv-qualification" part (which again has incorrect implications, in this case that we'd perform a runtime check if the cv-qualifications are incompatible).

While here, the p4 wording does not fit the pattern of this section (which is that you bail out early once you hit a matching paragraph). It incorrectly suggests that:

```
struct A { virtual ~A(); };
struct B : A {};
struct C : A, B {};
C *p = 0;
A *q = dynamic_cast<A*>(p);
```

... would be valid and produce a null pointer, because we never reach the "inaccessible or ambiguous" wording in p5.

jensmaurer:
and fix the null pointer provision.

Fixes #1934.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -2671,9 +2671,8 @@ type referred to by \tcode{T}. If \tcode{T} is an rvalue reference type,
 result is an xvalue of the type referred to by \tcode{T}.
 
 \pnum
-If the type of \tcode{v} is the same as \tcode{T}, or it is
-the same as \tcode{T} except that the class object type in \tcode{T} is
-more cv-qualified than the class object type in \tcode{v}, the result is
+If the type of \tcode{v} is the same as \tcode{T} (ignoring cv-qualifications),
+the result is
 \tcode{v} (converted if necessary).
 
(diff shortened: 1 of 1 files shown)


zygoloid:
Per cwg discussion on #1453, the null pointer part of this is not editorial. Striking the redundant cv-qualification checks seems fine, though.

jensmaurer:
Removed edits for null pointer issue, but kept removing the redundant cv-qualification checks.

Issue #1263, opened 2016-12-16
[lib] Referring to standard library requirements.


jensmaurer:
First, we have a mixture of "shall satisfy" (89) and "shall meet" (61). What's the preferred phrase?

Second, for CopyConstructible etc., we sometimes cross-reference the table number and sometimes the section. What's the preference here?

Third, for CopyConstructible etc., we sometimes say "shall meet/satisfy the CopyConstructible requirements" and sometimes we say "shall meet/satisfy the requirements of CopyConstructible". What's the preferred phrase?

For iterator requirements, we're now fairly consistently saying "shall meet/satisfy the requirements of a blah iterator (xref to section)", which is good.

jensmaurer:
Editorial meeting consensus:
- First: make an arbitrary decision; go with "shall satisfy" (cf. concepts)
- Second: cross-reference the table number
- Third: first choice; leave non-camelcase (iterators) alone.

jensmaurer:
[time.clock.req] p4.2 is an example of a long list of table references. Would this case be better served with a section reference to [utility.arg.requirements]?

jensmaurer:
A lot of references to CamelCase requirements are missing cross-references. For example, CopyConstructible appears 54 times, only 14 of these appearances have a "ref" nearby. Do we want to have a cross-reference after all mentions of CamelCase requirements?

jensmaurer:
@zygoloid opined that we'll reference the subclause (at the end) if we have a large-ish list of CamelCase requirements. Otherwise, we reference the table.

jensmaurer:
See #2176 and the new section "Requirements expressed by concepts" on the wiki.

AlisdairM:
It looks like @CaseyCarter did a lot of work to resolve this issue in a consistent manner for C++20.  Is there anything left to do before closing this issue?

jensmaurer:
Let's close this; if you have remaining concerns, please open a new issue.

PR #1937, opened 2018-02-20
[lib] Harmonize references to standard library requirements.

jensmaurer:
Partially addresses #1263.

I haven't squashed the commits; I can do so if desired.


jensmaurer:
Rebased.

tkoeppe:
No need to squash, but could you move the [macros] change to the bottom so it comes first?

jensmaurer:
@tkoeppe: Done.

source/strings.tex
```diff
@@ -275,10 +275,10 @@
 \pnum
 \requires
 \tcode{state_type}
-shall satisfy the requirements of
+shall satisfy the
 \tcode{CopyAssignable} (\tref{copyassignable}),
```
tkoeppe: This is one of the few cases where the placement of the reference is acceptable, since you are listing multiple requirements. In all cases where we only say *one* requirement, I would much prefer the reference *after* the word "requirements". Otherwise it's really disrutpive: "shall satisfy the Foo (Table n) requirements." vs "shall satisfy the Foo requirements (Table n)."

We even have a lot of precedent for that placement in the many other cases where we already say " shall satisfy the Foo requirements". So it wouldn't just read better, but also be more consistent.

source/iostreams.tex
```diff
@@ -13031,7 +13031,7 @@
 \indexlibrary{\idxcode{file_type}}%
 \pnum
 This enum class specifies constants used to identify file types,
-with the meanings listed in Table~\ref{tab:fs.enum.file_type}.
+with the meanings listed in Table~\tref{fs.enum.file_type}.
```
tkoeppe: @jensmaurer: Extraneous "Table" here.

source/iostreams.tex
```diff
@@ -13074,7 +13074,7 @@
 \pnum
 The \tcode{enum class} type \tcode{copy_options}
 is a bitmask type\iref{bitmask.types} that specifies bitmask constants used to control the semantics of
-copy operations. The constants are specified in option groups with the meanings listed in Table~\ref{tab:fs.enum.copy_options}.
+copy operations. The constants are specified in option groups with the meanings listed in Table~\tref{fs.enum.copy_options}.
```
tkoeppe: Another extraneous "Table".

tkoeppe: And again.

tkoeppe:
Oh, I see that #1263 discusses the (lack of) cross references, so everything may already be as intended.

tkoeppe:
I fixed the duplicate "Table"s and pushed the results.

PR #1943, opened 2018-02-27
[fs.class.filesystem_error] Rephrase para 1 to avoid "class defines t…

CaseyCarter:
…ype"

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -12814,9 +12814,8 @@ namespace std::filesystem {
 }
 \end{codeblock}
 \pnum
-The class \tcode{filesystem_error} defines the type of
-objects thrown as exceptions to report file system errors from functions described in this
-subclause.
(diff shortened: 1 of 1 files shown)


jensmaurer:
I support this change; it's a net improvement. But [res.on.exception.handling] already discusses derived classes; it would be good to defer the exact set of types that could be thrown to that section. Also, [except.handle] explains that only unambiguous public base classes fit, so something like
`class D : private filesystem_error { };`
is not suitable (but fits the wording as proposed).

tkoeppe:
@jensmaurer: Do you want a more extensive change here, or is it OK as is?

jensmaurer:
@tkoeppe: The proposed wording is not ok; it incorrectly handles private and ambiguous base classes.

CaseyCarter:
> The proposed wording is not ok

+1, this is broken and I need to fix it. Actually, I'll just close the PR until I have new wording.

PR #1947, opened 2018-02-28
[dcl.spec.auto] Use of undeduced placeholder types

### LINKED ISSUE #1942: update [dcl.spec.auto]p9 to match the form of wording used by cwg2285
CWG2285 adds a rule for "cannot appear within its own initializer" to structured bindings. During CWG discussion, we felt that formulation was preferable to the "type of entity with deduced type shall not be used to determine the type of something else" wording in [dcl.spec.auto]p9.

We should come up with a wording change to [dcl.spec.auto]p9 to match cwg2285 and take it back to CWG to check.

jensmaurer:
As discussed for CWG 2285, variables declared with a
placeholder type should never reference itself in the
initializer.  Similarly, clarify the treatment of
deduced function return types.

Fixes #1942.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -1651,14 +1651,14 @@ auto* g() { }                   // error, cannot deduce \tcode{auto*} from \tcod
 \end{example}
 
 \pnum
-If the type of an entity with an undeduced placeholder type is needed to
-determine the type of an expression, the program is ill-formed. Once a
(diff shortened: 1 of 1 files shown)


jensmaurer:
CWG in Jacksonville: "If the name of an entity with an undeduced placeholder type appears in an expression, the program is ill-formed."

PR #1948, opened 2018-02-28
[std] Review cross-references to [expr.prim]

jensmaurer:
Cross-references to [expr.prim] should instead point to one
of its subclauses.


jensmaurer:
Rebased.

PR #1949, opened 2018-03-03
CWG2345 Jumping across initializers in init-statements and conditions

jensmaurer:


source/statements.tex
```diff
@@ -898,7 +898,9 @@
 \indextext{initialization!jump past}%
 \indextext{\idxcode{goto}!initialization and}%
 It is possible to transfer into a block, but not in a way that bypasses
-declarations with initialization. A program that jumps\footnote{The transfer from the condition of a \tcode{switch} statement to a
+declarations with initialization (including such in \grammarterm{condition}s
```
jwakely: I find this use of "such" to be a little awkward.

jensmaurer: Let's try "ones" instead; Google found a similar sentence "Dozens of Vegetable Products, Including Ones Sold at Trader Joe's and Walmart, Recalled Over Listeria Concerns".

jwakely: Yes, I like that more. Another option that doesn't sound wrong to me is "including such declarations in ..."

jensmaurer:
CWG in Jacksonville: add in 9p1: "A condition that is not an expression is a declaration ([dcl])." Proceed editorially.

jensmaurer:
Updated for CWG comment.

PR #1959, opened 2018-03-18
[associative.reqmts] Turn emphasis into a note.

### LINKED ISSUE #606: Words in italics
Italic words are only specified to be the terms or syntactic categories. What about other kind of usage, i.e. to emphasize?

For example, in [associative.reqmts]/4:

> The phrase “equivalence of keys” means the equivalence relation imposed by the comparison and _not_ the `operator==` on keys.

Here the word "not" is italicized. Is this intended?

BTW, for this particular half of the sentence, is it better to be a note or a footnote? Note that the `operator==` can mean equivalence relation on keys occasionally.

jensmaurer:
Fixes #606.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -1511,11 +1511,7 @@ of a container.
 
 \pnum
 The phrase ``equivalence of keys'' means the equivalence relation imposed by the
-comparison and
-\textit{not}
-the
-\tcode{operator==}
-on keys.
+comparison object.
 That is, two keys
 \tcode{k1}
 and
@@ -1524,6 +1520,9 @@ are considered to be equivalent if for the
 comparison object
 \tcode{comp},
 \tcode{comp(k1, k2) == false \&\& comp(k2, k1) == false}.
+\begin{note}
+This is not necessarily the same as the result of \tcode{k1 == k2}.
+\end{note}
 For any two keys
 \tcode{k1}
 and


source/containers.tex
```diff
@@ -1507,11 +1507,7 @@
 
 \pnum
 The phrase ``equivalence of keys'' means the equivalence relation imposed by the
-comparison and
-\textit{not}
-the
-\tcode{operator==}
-on keys.
+comparison.
```
jwakely: Would it be a further improvement to say "imposed by the comparison object" not just "imposed by the comparison"?

source/containers.tex
```diff
@@ -1520,6 +1516,10 @@
 comparison object
 \tcode{comp},
 \tcode{comp(k1, k2) == false \&\& comp(k2, k1) == false}.
+\begin{note}
+This is not necessarily the same as the result of the \tcode{operator==}
+on keys.
```
jwakely: Maybe "This is not necessarily the same as the result of `k1 == k2`." ?

source/containers.tex
jensmaurer: Fixed.

source/containers.tex
jensmaurer: Fixed.

jensmaurer:
Updates per @jwakely's comments.

PR #1961, opened 2018-03-18
[xrefdelta] Use \deprxref

jensmaurer:


jensmaurer:
Rebased.

tkoeppe:
Sorry, could you briefly explain this change? What does it fix or address?

jensmaurer:
It simply uses a macro intended for the purpose instead of open-coding the moved xref. There should be zero change in appearance of the resulting PDF.

tkoeppe:
Ah good, thanks!

PR #1963, opened 2018-03-19
[expr.reinterpret.cast] Define reinterpret_cast from nullptr_t in terms of reinterpret_cast<void*> rather than c-style cast

RedBeard0531:
It seems clearer to directly use the specific cast rather than relying on the equivalence of the c-style cast.


timsong-cpp:
This is also wrong. `reinterpret_cast<void*>(0)` is not guaranteed to give you a null pointer value.

RedBeard0531:
Thanks for correcting me. I thought `(void*)0` was doing a reinterpret_cast as it would for any int other than a literal 0. I forgot that literal 0 is magical and results in a "standard pointer conversion" (http://eel.is/c++draft/conv#ptr-1) which doesn't have explicit syntax to designate.

Closing.

PR #1964, opened 2018-03-21
[mem.res.pool.overview] add proper definitions of terms

### LINKED ISSUE #1258: [mem.res.pool.overview] Implicit definition of terms
The terms "pools", "chunks", upstream memory resource", "upstream allocator" are proclaimed to be defined here, but the phrasing doesn't actually say "A _pool_ is ..." etc.

jwakely:
Update cross-reference in [mem.res.pool.options] to refer to
[mem.res.pool.overview].

Fixes #1258

### DIFF
--- a/source/memory.tex
+++ b/source/memory.tex
@@ -5802,21 +5802,43 @@ The \tcode{synchronized_pool_resource} and
 are general-purpose memory resources having the following qualities:
 \begin{itemize}
 \item
-Each resource frees its allocated memory on destruction,
-even if \tcode{deallocate} has not been called for some of the allocated blocks.
+A pool resource has an \defn{upstream memory resource},
+supplied on construction, which is used to allocate memory that
+the pool resource subdivides to satisfy calls to its own
+\tcode{do_allocate} member.
+
 \item
-A pool resource consists of a collection of \defn{pools},
-serving requests for different block sizes.
-Each individual pool manages a collection of \defn{chunks}
-that are in turn divided into blocks of uniform size,
-returned via calls to \tcode{do_allocate}.
+A \defn{chunk} is a region of storage allocated
+from the upstream memory resource. A pool resource divides
+each chunk into one or more blocks of the same block size.
+Different chunks can have different block sizes, and different
+numbers of blocks.
+
+\item
+A \defn{pool} is a collection of chunks that all have the same
+block size. Each pool resource contains a collection of pools
+with different block sizes.
+
+\item
+On any call to \tcode{release()}, a pool resource frees all the memory
+it obtained from its upstream memory resource by calling
+the upstream resource's \tcode{deallocate} function.
+\begin{note}
+This frees all memory returned by the pool resource's \tcode{do_allocate}
+function, even if \tcode{do_deallocate} has not been called for some
+of the allocated blocks.
+\end{note}
+
+\item
 Each call to \tcode{do_allocate(size, alignment)} is dispatched
-to the pool serving the smallest blocks accommodating at least \tcode{size} bytes.
+to the pool serving the smallest blocks that accommodate
+\tcode{size} and \tcode{alignment}.
+
 \item
 When a particular pool is exhausted,
 allocating a block from that pool results in the allocation
-of an additional chunk of memory from the \defn{upstream allocator}
-(supplied at construction), thus replenishing the pool.
+of additional chunks from the upstream memory resource,
+thus replenishing the pool.
 With each successive replenishment,
 the chunk size obtained increases geometrically.
 \begin{note}
@@ -5824,9 +5846,11 @@ By allocating memory in chunks,
 the pooling strategy increases the chance that consecutive allocations
 will be close together in memory.
 \end{note}
+
 \item
-Allocation requests that exceed the largest block size of any pool
-are fulfilled directly from the upstream allocator.
+Allocation requests that exceed the largest block size and/or alignment
+of all pools are fulfilled directly from the upstream allocator.
+
 \item
 A \tcode{pool_options} struct may be passed to the pool resource constructors
 to tune the largest block size and the maximum chunk size.
@@ -5921,8 +5945,8 @@ size_t max_blocks_per_chunk;
 
 \begin{itemdescr}
 \pnum
-The maximum number of blocks that will be allocated at once
-from the upstream memory resource\iref{mem.res.monotonic.buffer}
+The maximum number of blocks in any single chunk that will be allocated
+from the upstream memory resource\iref{mem.res.pool.overview}
 to replenish a pool.
 If the value of \tcode{max_blocks_per_chunk} is zero or
 is greater than an \impldef{largest supported value to configure the maximum number of blocks to replenish a pool}
--- a/source/memory.tex
+++ b/source/memory.tex
@@ -5821,7 +5821,7 @@ with different block sizes.
 
 \item
 On any call to \tcode{release()}, a pool resource frees all the memory
-it obtained from its upstream memory resource by calling
+it has obtained from its upstream memory resource by calling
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -11806,30 +11806,54 @@
...
 \item
-A pool resource consists of a collection of \defn{pools},
-serving requests for different block sizes.
-Each individual pool manages a collection of \defn{chunks}
-that are in turn divided into blocks of uniform size,
-returned via calls to \tcode{do_allocate}.
 Each call to \tcode{do_allocate(size, alignment)} is dispatched
-to the pool serving the smallest blocks accommodating at least \tcode{size} bytes.
+to the pool serving the smallest blocks accommodating at \tcode{size} bytes
+and are aligned to a multiple of \tcode{alignment}.
```
CaseyCarter: This sentence has gone astray: "... accommodating at `size` bytes and are aligned to a multiple of `alignment`." is not Englishing goodly. How about "Each call to `do_allocate(size, alignment)` is dispatched to the pool serving the smallest blocks that accommodate `size` and `alignment`."? "multiple of `alignment` seems extraneous to me given that all alignments are powers of two (http://eel.is/c++draft/basic.align#4).

jwakely: Yes I'm not sure how that got mangled. I'll push an update.

source/utilities.tex
```diff
@@ -11824,30 +11824,54 @@
...
+Different chunks can have different block sizes, and different
+numbers of blocks.
+
+\item
+A \defn{pool} is a collection of chunks that all have the same
+block size. Each pool resource contains a collection of pools
+with different block sizes.
+
+\item
+On destruction, a pool resource frees all the memory it obtained from
```
Quuxplusone: "On any call to `release()`..."? And then just specify that the destructor always calls `release()`?
Otherwise we'll have either that `release()` is underspecified, or the specification of `release()` duplicates all this excellent wording.

source/utilities.tex
```diff
@@ -11824,30 +11824,54 @@
...
-to the pool serving the smallest blocks accommodating at least \tcode{size} bytes.
+to the pool serving the smallest blocks that accommodate
+\tcode{size} and \tcode{alignment}.
+
 \item
 When a particular pool is exhausted,
 allocating a block from that pool results in the allocation
-of an additional chunk of memory from the \defn{upstream allocator}
-(supplied at construction), thus replenishing the pool.
+of an additional chunk from the upstream memory resource,
```
Quuxplusone: Does "an" imply that only one such chunk must be allocated? Could it be "one or more" chunks?

source/utilities.tex
```diff
@@ -11824,30 +11824,54 @@
...
 \begin{note}
 By allocating memory in chunks,
 the pooling strategy increases the chance that consecutive allocations
 will be close together in memory.\end{note}
+
 \item
-Allocation requests that exceed the largest block size of any pool
-are fulfilled directly from the upstream allocator.
+Allocation requests that exceed the largest block size and/or alignment
+of any pool are fulfilled directly from the upstream allocator.
```
Quuxplusone: Preexisting grammar nit: presumably "any pool" should read "all pools"

source/utilities.tex
```diff
@@ -11943,7 +11967,7 @@
 \begin{itemdescr}
 \pnum
 The maximum number of blocks that will be allocated at once
-from the upstream memory resource\iref{mem.res.monotonic.buffer}
+from the upstream memory resource\iref{mem.res.pool.overview}
```
Quuxplusone: If you take my suggestion that the resource be permitted to allocate "one or more" chunks to replenish a pool, then this wording must be updated to replace the phrase "at once" with the phrase "in any single chunk".

source/utilities.tex
CaseyCarter: Replacing "an additional chunk" with "additional chunks" would IMO be a simpler way to express that behavior than "one or more additional chunks".

jwakely:
Rebased, all comments from @CaseyCarter and @Quuxplusone addressed.

jensmaurer:
@jwakely , this is a large-ish change, and @zygoloid added the "lwg" label four years ago. Do you think this should go via LWG, or should be just collect some endorsement here from @CaseyCarter and others?

jwakely:
I think it should go via LWG. I'll ask for some reflector reviews of this and #2550 and some others with the lwg label.

source/memory.tex
```diff
@@ -5802,31 +5802,55 @@
...
+
+\item
+A \defn{pool} is a collection of chunks that all have the same
+block size. Each pool resource contains a collection of pools
+with different block sizes.
+
+\item
+On any call to \tcode{release()}, a pool resource frees all the memory
+it obtained from its upstream memory resource by calling
+the upstream resource's \tcode{deallocate} function.
```
Quuxplusone, line 5825: `it has obtained`
Abbreviating as `the upstream resource's` here feels just a little weird to me, since the true name is `the upstream memory resource's` and if we're aiming for brevity we could just say `the upstream's`.

I admit that my eye caught on this paragraph originally because it refers to "the memory
it [has] obtained from its upstream memory resource by calling the upstream resource's deallocate function"; when in fact that memory was obtained by calling the upstream's _allocate_ function! But I have no suggested fix for that grammar quirk, so my brain found other nits.

source/memory.tex
```diff
@@ -5802,31 +5802,55 @@
...
 By allocating memory in chunks,
 the pooling strategy increases the chance that consecutive allocations
 will be close together in memory.
 \end{note}
+
 \item
-Allocation requests that exceed the largest block size of any pool
-are fulfilled directly from the upstream allocator.
+Allocation requests that exceed the largest block size and/or alignment
+of all pools are fulfilled directly from the upstream allocator.
```
Quuxplusone, line 5852: `from the upstream memory resource`?

Quuxplusone (commented):
Some nits, but LGTM FWIW!

source/memory.tex
```diff
@@ -5802,31 +5802,55 @@
...
-even if \tcode{deallocate} has not been called for some of the allocated blocks.
+A pool resource has an \defn{upstream memory resource},
+supplied on construction, which is used to allocate memory that
+the pool resource subdivides to satisfy calls to its own
+\tcode{do_allocate} member.
+
+\item
+A \defn{chunk} is a region of storage allocated
+from the upstream memory resource. A pool resource divides
+each chunk into one or more blocks of the same block size.
```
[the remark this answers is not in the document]
JohelEGP: No. `\defn` is for marking the definition, not its uses.

source/memory.tex
jwakely: How about:

> On a call to `release()`, a pool resource calls its upstream memory resource's `deallocate` function to return all the memory that was obtained from its upstream memory resource.

jwakely: Or "that was obtained from upstream".

Quuxplusone: Minor tweak:

> On a call to `release()`, a pool resource calls its upstream memory resource's `deallocate` function to return all the memory that _has been_ obtained from its upstream memory resource.

Or "all the _outstanding_ memory," but that's a little jargony. Could even say "all the _chunks_ that have been obtained," if that's correct. Can't say "_has ever been_ obtained," because we really do mean only the outstanding chunks, not ones that were allocated before the last call to `release()` nor ones that have already been deallocated via `deallocate`.

Wildly ambitious rewrite:

> On a call to `release()`, a pool resource calls its upstream memory resource's `deallocate` function on every chunk that has not yet been deallocated.

jwakely: But it doesn't say all memory that has *ever* been allocated, and we don't bother making that distinction for containers freeing memory with an allocator.

PR #1965, opened 2018-03-21
Use \placeholder for alias templates in container deduction guides

### LINKED ISSUE #1523: Exposition-only alias templates for associative container deduction guides
(This applies to the changes in https://github.com/cplusplus/draft/tree/motions-2017-03-lwg-19 which is not yet merged).

We have `iter_key_t` and `iter_val_t` for the `key_type` and `mapped_type`, which are only applicable to maps. We could rename `iter_val_t` to `iter_mapped_t` and then define another alias template with the name `iter_val_t` to be the iterator's `value_type`. That would allow it to be used in the deduction guides for sets, as well as in the `iter_key_t` and `iter_mapped_t` types.

jwakely:
Rename the alias templates and style as placeholders:

`iter_key_t` -> _iter-key-type_
`iter_val_t` -> _iter-mapped-type_
`iter_to_alloc_t` -> _iter-to-alloc-type_

Add _iter-value-type_ for use in sequence containers and sets.

Fixes #1523


tkoeppe:
That's very nice!

jwakely:
There are some new overfull hboxes, due to `iter-mapped-type` being longer than `iter_val_t`. I'll fix those.

source/containers.tex
```diff
@@ -2924,6 +2924,15 @@
 \tcode{<list>}, and \tcode{<vector>} define class templates that meet the
 requirements for sequence containers.
 
+\pnum
+The following exposition-only alias template may appear in deduction guides for sequence containers:
+
+\begin{codeblock}
+template<class InputIterator>
+  using @\placeholder{iter-value-type}@ =
+    typename iterator_traits<InputIterator>::value_type>; // \expos
```
timsong-cpp: Stray `>` at the end.

source/containers.tex
```diff
@@ -5797,13 +5797,16 @@
 
 \begin{codeblock}
 template<class InputIterator>
-  using iter_key_t = remove_const_t<
+  using @\placeholder{iter-value-type}@ =
+    typename iterator_traits<InputIterator>::value_type>; // \expos
```
timsong-cpp: Here too.

jwakely: fixed both of these, thanks.

PR #1972, opened 2018-03-22
[intro.refs] Add normative reference to ISO 8601

### LINKED ISSUE #1971: [intro.refs] add normative reference on ISO 8601
ISO 8601 is cited normatively from the <chrono> calendar extensions. We should add a matching normative reference.

jensmaurer:
Fixes #1971.


jensmaurer:
Addressed by  commit f434bb7.

PR #1976, opened 2018-03-23
[namespace.udecl] Demote normative duplication to notes

### LINKED ISSUE #1975: [namespace.udecl]p16 references nonexistent "implicit this parameter"
> In particular, the implicit `this` parameter shall be treated as if it were a pointer to the derived class rather than to the base class.

The name is "implicit object parameter", not "implicit `this` parameter", and it's a reference, not a pointer. See [over.match.funcs]p4. There also appears to be some normative duplication and contradiction between these two paragraphs as [namespace.udecl]p16 contains no exception for conversion functions.

jensmaurer:
The effects of using-declarations naming member functions
are discussed in other parts of the standard.
The effects of initializing with an inherited constructor are
discussed elsewhere, too.

Fixes #1975.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -3043,24 +3043,28 @@ D2 d2(0);           // calls \tcode{D2::D2(int)}
 
 \pnum
 \indextext{overloading!using-declaration and}%
+\begin{note}
 For the purpose of forming a set of candidates during overload resolution,
 the functions that are
 introduced by a \grammarterm{using-declaration} into a derived class
-are treated as though they were members of the derived class. In
-particular, the implicit \tcode{this} parameter shall be treated as if
-it were a pointer to the derived class rather than to the base class.
+are treated as though they were members of the derived class\iref{class.member.lookup}. In
+particular, the implicit object parameter is treated as if
+it were a reference to the derived class rather than to the base class\iref{over.match.funcs}.
 This has no effect on the type of the function, and in all other
 respects the function remains a member of the base class.
-Likewise, constructors that are introduced by a \grammarterm{using-declaration}
+\end{note}
+
+\pnum
+Constructors that are introduced by a \grammarterm{using-declaration}
 are treated as though they were constructors of the derived class
 when looking up the constructors of the derived class\iref{class.qual}
 or forming a set of overload candidates~(\ref{over.match.ctor}, \ref{over.match.copy}, \ref{over.match.list}).
+\begin{note}
 If such a constructor is selected to perform the initialization
 of an object of class type, all subobjects other than the base class
 from which the constructor originated
 are implicitly initialized\iref{class.inhctor.init}.
-\begin{note}
-A member of a derived class is sometimes preferred to a member of a base class
+A constructor of a derived class is sometimes preferred to a member of a base class
 if they would otherwise be ambiguous\iref{over.match.best}.
 \end{note}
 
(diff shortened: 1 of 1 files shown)


source/declarations.tex
```diff
@@ -3017,25 +3017,29 @@
...
-particular, the implicit \tcode{this} parameter shall be treated as if
-it were a pointer to the derived class rather than to the base class.
+are treated as though they were members of the derived class\iref{class.member.lookup}. In
+particular, the implicit object parameter is treated as if
+it were a reference to the derived class rather than to the base class\iref{over.match.funcs}.
 This has no effect on the type of the function, and in all other
 respects the function remains a member of the base class.
-Likewise, constructors that are introduced by a \grammarterm{using-declaration}
+A member of a derived class is sometimes preferred to a member of a base class
+if they would otherwise be ambiguous\iref{over.match.best}.
```
zygoloid: This actually only applies to the constructor case, right? (That is, what's now the next paragraph.)

zygoloid:
The mix of normative and non-normative rules here is really strange. I think I'd like it more if we added a normative rule to [over.match.funcs] for the inherited constructor case (copying the rule from here) and turned everything in this section into a note.

source/declarations.tex
jensmaurer: I'm not sure. With using-declarations and multiple inheritance etc., maybe member name lookup can yield a lookup set that contains both members of a derived class and of a base class (if name lookup could find the members of the base class via some other path). In that case, that note would also apply in that overload resolution gets to choose a unique result.

jensmaurer:
@zygoloid: If we want to do this properly, we'd need a half-sentence (or so) in each of the three 16.3.1.x subclauses where constructors are enumerated, also in [class.qual] (we could have a using-declaration pointing to a constructor introduced by a using-declaration, recursively). Maybe something like ", including constructors introduced by a using-declaration [namespace.udecl]" or ", including inherited constructors [namespace.udecl]" is good enough. What do you think?

source/declarations.tex
zygoloid: The above "In particular, the implicit object parameter is treated as if it were a reference to the derived class" suggests that we don't take derived-to-base conversions on the implied object parameter into account. And actually, I'm not seeing anywhere in [over.match.funcs] that makes this clear; we are just told "X is the class of which the function is a member". So I think that part here is not duplicating other normative wording either.

So I'm not sure there's really all that much normative duplication here.

jensmaurer: @zygoloid, we have in p4 in [over.match.funcs] "For non-conversion functions introduced by a using-declaration into a derived class, the function is considered to be a member of the derived class for the purpose of defining the type of the implicit object parameter."  What's missing in that sentence that would be added by a normative "In particular ..." sentence?

source/declarations.tex
```diff
@@ -3018,25 +3018,29 @@
...
 are treated as though they were constructors of the derived class
 when looking up the constructors of the derived class\iref{class.qual}
 or forming a set of overload candidates~(\ref{over.match.ctor}, \ref{over.match.copy}, \ref{over.match.list}).
+\begin{note}
 If such a constructor is selected to perform the initialization
 of an object of class type, all subobjects other than the base class
 from which the constructor originated
 are implicitly initialized\iref{class.inhctor.init}.
-\begin{note}
-A member of a derived class is sometimes preferred to a member of a base class
```
jensmaurer: Editorial meeting consensus: Do not move this note, change "member" to "constructor".

jensmaurer: Done.

PR #1977, opened 2018-03-23
[expr.rel] Clarify auxiliary partial ordering

### LINKED ISSUE #1938: [expr.rel] "compares greater" term is needlessly confusing
In [expr.rell]p3, we implicitly introduce a relation called "compares greater". However, it's confusing to use this term, since there are other cases that might /also/ compare greater in the colloquial sense, if they fall into the ~(p compares greater than q) /\ ~(p compares greater than p) case. It's also easy to misread p3 as actually specifying the result of the comparison operators rather than just defining an intermediate term.

How about:

1) Changing the introductory sentence to "The result of comparing unequal pointers to elements is specified in terms of a "definitely compares greater" relation, defined as follows:"
2) Replace "compares greater" with "definitely compares greater" throughout.

Likewise in [expr.eq]p2 and p3 we should make it clear that we're just defining a "compares equal" relation, not actually defining the result of the operators. This is confounded by p3.3 and p3.4 actually trying to define the result of the the operators ("the result is unspecified"). As per the [expr.rel] changes, we should change the introductory sentence to clarify this, rename the relations to "definitely compare equal" and "definitely compare unequal", and change the "the result is unspecified" to "neither definitely compare equal nor definitely compare unequal".

jensmaurer:
Fixes #1938.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -5073,27 +5073,27 @@ them to their composite pointer type\iref{expr.prop}.
 After conversions, the operands shall have the same type.
 
 \pnum
-Comparing unequal pointers to objects%
+The result of comparing unequal pointers to objects%
 \footnote{An object that is not an array element is considered to belong to a
 single-element array for this purpose; see~\ref{expr.unary.op}.
 A pointer past the last element of an array \tcode{x} of $n$ elements
 is considered to be equivalent to a pointer to a hypothetical element
 $\mathtt{x[}n\mathtt{]}$ for this purpose; see~\ref{basic.compound}.}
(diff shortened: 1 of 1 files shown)


timsong-cpp:
Do we also want to change the `if a pointer \tcode{p} compares greater than a pointer \tcode{q}` in the following paragraph?

jensmaurer:
I don't think so.  The idea here is that we have a partial ordering that is only partially specified. So, there might be other (implementation-specific) situations beyond the "is required to compare greater" situations where a pointer compares greater than another one.

PR #1978, opened 2018-03-23
[containers] Removed redundant specifications of non-member swap

### LINKED ISSUE #1935: [unord] "Specialized algorithms" vs. "Swap"
The original containers use the heading "Specialized algorithms" for the section whose only purpose is to overload `std::swap()`. The unordered containers (which are fairly new) use the heading "Swap". That's inconsistent.

My personal view is that we should remove that subclause entirely:
- The declaration is already shown in the header synopsis.
- The meaning should be established by some front-matter wording in the container requirements, which we already have. Maybe add "equivalent to:" in front of the `a.swap(b)` equivalence in the corresponding table row so that `noexcept` is the same.

jensmaurer:
Fixes #1935.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -235,7 +235,7 @@ denotes a non-const rvalue of type \tcode{X}.
 
 \tcode{swap(a, b)}          &
   \tcode{void}              &
-  \tcode{a.swap(b)}         &
+  Equivalent to \tcode{a.swap(b)}         &
                             &
   (Note A)                  \\ \rowsep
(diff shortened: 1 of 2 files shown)


tkoeppe:
Please rebase.

@zygoloid: Attention: this change requires announcement.

jensmaurer:
Rebased.

jensmaurer:
Rebased again.

jensmaurer:
... and again.

PR #1981, opened 2018-03-23
[temp.local] Fix example not to name the constructor.

### LINKED ISSUE #1900: [temp.local]/3 Wrong example?
```c++
template<class T, template<class> class U = T::template Base> struct Third { };
Third<Base<int> > t;                    // OK: default argument uses injected-class-name as a template
```
With `T::template Base`, lookup for `Base` in the scope of `T` aka `Base<int>` finds the injected-class-name, and this lookup does not ignore function names, so per [CWG 1310](https://wg21.link/CWG1310) it should be considered to name the constructor of `Base`, making the code ill-formed.

If this analysis is correct, perhaps the example can be changed to
```c++
template<class T, template<class> class U = T::template Base> struct Third { };
Third<Derived<int> > t;                    // OK: default argument uses injected-class-name as a template
Third<Base<int> > t;                       // Error: injected-class-name in default argument names the constructor of Base ([class.qual])
```

jensmaurer:
Fixes #1900.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -4392,7 +4392,7 @@ template <class T> struct Derived: public Base<T> {
 };
 
 template<class T, template<class> class U = T::template Base> struct Third { };
-Third<Base<int> > t;                    // OK: default argument uses injected-class-name as a template
+Third<Derived<int> > t;                 // OK: default argument uses injected-class-name as a template
 \end{codeblock}
 \end{example}
 


source/templates.tex
```diff
@@ -4392,7 +4392,7 @@
 };
 
 template<class T, template<class> class U = T::template Base> struct Third { };
-Third<Base<int> > t;                    // OK: default argument uses injected-class-name as a template
```
tkoeppe: I feel that we're breaking someone's in-joke here, but fine.

PR #1984, opened 2018-03-25
[re.regex] Avoid duplicate list of constants

### LINKED ISSUE #1163: [re.regex]/3 regex_constants presentation
Each constant takes up two lines. That's ugly.

We could squeeze each on a single line (with comfortable padding) if we used the flag_type alias for regex_constants::syntax_option_type.

Opinions?

jensmaurer:
Shorten the synopsis and remove [re.regex.const].

Fixes #1163.


tkoeppe:
@zygoloid: The linked Issue says to "call this out in the editor's report". Please let this be called out.

PR #1985, opened 2018-03-25
[time.duration.nonmember] Replace type designator CR with its definition

### LINKED ISSUE #916: [time.duration.nonmember] is missing formatting
More than half of the _Returns:_ clauses in this subclause are in Roman rather than Teletype.

We also have inconsistency between this subclause and the next one: in [time.duration.comparisons](and in [time.point.nonmember]), we use italic teletype for the placeholder _`CT`_, but in this subclause (and in [time.duration.cast]) we use non-italic teletype for the placeholders `CD` and `CR`.

jensmaurer:
Fixes #916.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -18180,7 +18180,7 @@ static constexpr duration max();
 
 \pnum
 In the function descriptions that follow, \tcode{CD} represents the return type
-of the function. \tcode{CR(A, B)} represents \tcode{common_type_t<A, B>}.
+of the function.
 
 \indexlibrary{\idxcode{common_type}}%
 \begin{itemdecl}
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -18216,7 +18216,7 @@
 \begin{itemdescr}
 \pnum
 \remarks This operator shall not participate in overload
-resolution unless \tcode{Rep2} is implicitly convertible to \tcode{CR(Rep1, Rep2)}.
+resolution unless \tcode{Rep2} is implicitly convertible to \tcode{common_type_t(Rep1, Rep2)}.
```
timsong-cpp: These need to be angle brackets.

jensmaurer: Right. Embarrassing. Thanks. Fixed.

PR #1986, opened 2018-03-25
[lib] Avoid 'shall' and 'should' in footnotes.

jensmaurer:
Partially addresses #1414.

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -728,19 +728,7 @@ is a synonym for one of the signed basic
 integral types.
 It is used to represent the number of characters transferred in an I/O
 operation, or the size of I/O buffers.\footnote{\tcode{streamsize}
-is used in most places where ISO C would use
-\tcode{size_t}.
-Most of the uses of
-\tcode{streamsize}
-could use
-\tcode{size_t},
-except for the
-\tcode{strstreambuf}
-constructors, which require negative values.
-It should probably be the signed type corresponding to
-\tcode{size_t}
-(which is what Posix.2 calls
-\tcode{ssize_t}).}
+is used in most places where ISO C would use \tcode{size_t}.}
 \end{itemdescr}
 
 \rSec2[ios.base]{Class \tcode{ios_base}}
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -6818,7 +6818,7 @@ The elements are initialized with the values of the corresponding
 elements of \tcode{v}.\footnote{This copy constructor creates
 a distinct array rather than an alias.
 Implementations in which arrays share storage are permitted, but they
-shall implement a copy-on-reference mechanism to ensure that arrays are
+would need to implement a copy-on-reference mechanism to ensure that arrays are
 conceptually distinct.}
 \end{itemdescr}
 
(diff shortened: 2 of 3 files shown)


source/preprocessor.tex
```diff
@@ -1292,11 +1292,14 @@
 
 \indextext{__CPLUSPLUS@\xname{cplusplus}}%
 \item \xname{cplusplus}\\
-The integer literal \tcode{\cppver}.\footnote{It is intended that future
+The integer literal \tcode{\cppver}.
+\begin{note}
+It is intended that future
 versions of this International Standard will
 replace the value of this macro with a greater value.
 Non-conforming compilers should use a value with at most
```
timsong-cpp: The Directives have a similar ban against "shall"/"should" in notes too...

jwakely: Yes, this is still not allowed by ISO. How about "Non-conforming compilers can use a value ..."?

Or we could drop that part of the note, since it no longer reflects reality. Before C++17 was close to publication GCC used `201500` to indicate "more than C++14 but less than C++1z", and Clang used `201406`.

jensmaurer: First, we can't really say something about non-conforming compilers; they're non-conforming anyway. Second, we now have feature-test macros.
So, the update removes the "should" sentence talking about non-conforming compilers.

tkoeppe:
@jwakely, @timsong-cpp: Could you please review this?

source/numerics.tex
```diff
@@ -6818,7 +6818,7 @@
 elements of \tcode{v}.\footnote{This copy constructor creates
 a distinct array rather than an alias.
 Implementations in which arrays share storage are permitted, but they
-shall implement a copy-on-reference mechanism to ensure that arrays are
+would need to implement a copy-on-reference mechanism to ensure that arrays are
```
zygoloid: I'm going to merge as-is, but I wonder if we should just drop this -- in modern C++ this would also require taking a lock to provide thread-safety for const operations, which seems like a silly thing to bother describing as "permitted".

source/iostreams.tex
```diff
@@ -728,19 +728,7 @@
...
-could use
-\tcode{size_t},
-except for the
-\tcode{strstreambuf}
-constructors, which require negative values.
-It should probably be the signed type corresponding to
-\tcode{size_t}
-(which is what Posix.2 calls
-\tcode{ssize_t}).}
+is used in most places where ISO C would use \tcode{size_t}.}
```
jwakely, line 731: Sorry I've only noticed this now, but removing the mention of `ssize_t` seems unfortunate. Also, what remains of the footnote is a lie. We don't use `streamsize` for the size of an array _new-expression_ or the size passed to `operator new` (corresponding to `malloc` in C), or the length passed to `char_traits::compare` (`memcmp`), or "most" other places. Maybe we should qualify this footnote to be specific to I/O in C, or just to places that actually use `streamsize` e.g.

> Most places where `streamsize` is used would use `size_t` in ISO C, or `ssize_t` in POSIX.

tkoeppe: Could you make a PR so we don't lose track of this? The suggestion sounds good.

jwakely: Incorporated into PR #4299

PR #1987, opened 2018-03-26
[over.best.ics] p6, [over.ics.ref] p1 Fix case

FrankHB:
To avoid confusion, occurrence of "derived-to-base conversion" should not be with capital letter "C", which implies the name of a conversion category. The rank is already specified at the end of the paragraph.


jensmaurer:
Looks good to me.

PR #1989, opened 2018-03-26
Update container advice

tvaneerd:
Still somewhat out of date, but better.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -766,6 +766,7 @@ middle of the sequence.
 is the data structure of choice
 when most insertions and deletions take place at the beginning or at the
 end of the sequence.
+When choosing a container, remember \tcode{vector} is best; leave a comment to explain if you choose from the rest!
 
 \pnum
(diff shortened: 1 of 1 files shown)


timsong-cpp:
This whole paragraph has no normative impact and should be removed outright or at best turned into a note.

tvaneerd:
Or remove the rest and just leave my Seuss-ism. :-)

jfbastien:
Seuss note sounds good IMO.

jensmaurer:
Please add [sequence.reqmts] (the subclause that is changed) to the start of your commit message. Please turn p2 into a note instead of adding semi-funny normative statements.

rmpowell77:
Sorry to be a pest, but if we remove the rest, then what is best?
Best leave the rest, to stress the best.
Should we add a test?  To invest and stress?
Maybe a random test?  A guess-fest test?
Open that test?  A guest guest-fest test?
Decorate that test?  A dressed guest guess-fest test?

This thread is vest(ed) and gone west.

PR #1994, opened 2018-03-28
[any.class] Rephrase small-object optimization

### LINKED ISSUE #1360: [any.class]/3 Rephrase example
[[any.class]/3](http://eel.is/c++draft/any.class#3) reads:

> Implementations should avoid the use of dynamically allocated memory for a small contained object. [ Example: where the object constructed is holding only an int.  — end example ] Such small-object optimization shall only be applied to types T for which is_­nothrow_­move_­constructible_­v<T> is true.

Grammar-wise the phrasing of the example is not ideal.

jensmaurer:
Fixes #1360.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -5103,11 +5103,12 @@ The non-member \tcode{any_cast} functions provide type-safe access to the contai
 
 \pnum
 Implementations should avoid the use of dynamically allocated memory for a small contained value.
-\begin{example}
-where the object constructed is holding only an \tcode{int}.
-\end{example}
-Such small-object optimization shall only be applied to types \tcode{T} for which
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -5111,12 +5111,13 @@
 The non-member \tcode{any_cast} functions provide type-safe access to the contained value.
 
 \pnum
-Implementations should avoid the use of dynamically allocated memory for a small contained value.
+Implementations should avoid the use of dynamically allocated memory for a small contained value;
+any such small-object optimization shall only be applied to types \tcode{T} for which
```
zygoloid: A connective such as "however" would make more sense than a semicolon here, as we are partially retracting the advice given in the previous clause.

PR #1996, opened 2018-03-28
[basic.def.odr] Adds note that x may be part of a qualified id

ryanhaining:
Following from https://stackoverflow.com/questions/47952024
I'm sure there's a better way to word this.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -367,9 +367,9 @@ an expression or conversion as follows:
 \end{itemize}
 
 \pnum
-A variable \tcode{x} whose name appears as a
-potentially-evaluated expression \tcode{e}
-is \defnx{odr-used}{odr-use} by \tcode{e} unless
+A potentially-evaluated \grammarterm{id-expression}
+\tcode{e} that denotes a variable \tcode{x}
+\defnx{odr-uses}{odr-use} \tcode{x} unless
 \begin{itemize}
 \item
   \tcode{x} is a reference that is


jensmaurer:
Please add the section label to the start of your commit message. Please squash your commits.

source/basic.tex
```diff
@@ -355,7 +355,8 @@
 \end{itemize}
 
 \pnum
-A variable \tcode{x} whose name appears as a
+A variable \tcode{x} whose name appears, possibly as part of a
+\grammarterm{qualified-id}, as a
 potentially-evaluated expression \tcode{ex} is \defn{odr-used} by \tcode{ex} unless
```
jensmaurer: Would this read better if it said "A variable x whose name appears as a potentially-evaluated expression ex (possibly as part of a qualified-id), is odr-used by ex unless..."?

In general, I'm not convinced this change is a net improvement to start with. In the immediately preceding paragraph, we talk about a "function named by an expression", and qualified-ids are also included there, but not explicitly called out. Calling out qualified-ids in one place, but not another, makes the appearance of a difference where there is none.

ryanhaining: If you have a static variable referenced by `Widget::x`, is that covered by the term *name*? A "function named by an expression" imo carries a better implication of optionally being qualified. 

I don't expect this pr's proposed wording to be accepted.

ryanhaining:
tried to squash the commits, only made it worse.

source/basic.tex
zygoloid: How about this:

> A variable `x` that is named by a potentially-evaluated expression `ex` ([expr.prim.id]) is *odr-used* [...]

This "whose name appears" wording is problematically vague. For example:

```
int a;
void f(int a) { a = 0 }; // odr-uses ::a?
```

jensmaurer: I like @zygoloid's suggestion.

ryanhaining: done

source/basic.tex
```diff
@@ -355,8 +355,8 @@
 \end{itemize}
 
 \pnum
-A variable \tcode{x} whose name appears as a
-potentially-evaluated expression \tcode{ex} is \defn{odr-used} by \tcode{ex} unless
+A variable \tcode{x} that is named by a potentially-evaluated expression
+\tcode{ex}~(\ref{expr.prim.id}) is \defn{odr-used} by \tcode{ex} unless
```
tkoeppe: Can you use the `\iref` macro here instead?

jensmaurer:
Looks good to me.

tkoeppe:
I think @zygoloid was still concerned by the potential loss of clarity that it's now no longer clear whether the expression is in its entirety the thing that names `x`, or whether you could now read it that, say, `y<x>` "names `x`".

zygoloid:
Hmm, perhaps if we switch from passive to active-voice:

> A potentially-evaluated *id-expression* `ex` that denotes a variable `x` *odr-uses* `x` unless

tkoeppe:
@ryanhaining Could you please rebase and apply Richard's suggestion?

ryanhaining:
Yes, sorry you caught me in the middle of two back-to-back trips so might take me a bit, I'm having a really hard time with hotspots so just doing a pull is taking a while.

ryanhaining:
Okay, ready for review.

tkoeppe:
I think the rewording of https://wg21.link/p1787r6 improved the text sufficiently to resolve the original issue. Please double-check, and feel free to reopen if you would still like to tweak the wording further.

PR #1998, opened 2018-03-28
P0969R0 editorial fix

zygoloid:
Proposed editorial improvement to [dcl.struct.bind]p4:

 * removes unusual grammar construction "[...], well-formed [...]"
 * defines the `name` placeholder used in the `e.name` expressions


tkoeppe:
Is this obsolete now, @zygoloid?

tkoeppe:
The current wording now seems very close to what's proposed here; closing.

PR #1999, opened 2018-03-29
[time.cal.month.nonmembers,time.cal.wd.nonmembers] Do not digress to …

### LINKED ISSUE #1974: improve `std::chrono::month` `operator+`'s description of arithmetic mod 12
Perhaps we could just say:

Returns: a `month` with `m_` equal to the sum of `unsigned{x}` and `y.count()`, reduced modulo 12 to an integer in `[1, 12]`.

or something like that, rather than defining an auxiliary `modulo(n, 12)` function and going on an excursion to a definition of Euclidean division.

One problem: this is not entirely editorial, as the new wording does not admit any undefined behavior, whereas the description in the existing wording results in UB when `long long` and `months`' representation type are the same size and the `months` value is nearly `LLONG_MAX`. The undefined behavior here does not appear to be especially useful; implementations can easily produce the correct result by reducing the `months` value modulo 12 first, if they chose a representation type for `months` that is the same size as `long long`.

jensmaurer:
…Eucledian algorithm.

Fixes #1974.

### DIFF
--- a/source/time.tex
+++ b/source/time.tex
@@ -4401,16 +4401,9 @@ constexpr month operator+(const month& x, const months& y) noexcept;
 \begin{itemdescr}
 \pnum
 \returns
-\begin{codeblock}
-month{modulo(static_cast<long long>(unsigned{x}) + (y.count() - 1), 12) + 1}
-\end{codeblock}
-where \tcode{modulo(n, 12)} computes the remainder of \tcode{n} divided by 12 using Euclidean division.
-\begin{note}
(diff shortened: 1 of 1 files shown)


jensmaurer:
This puts additional burden on the implementation, because it has to avoid overflow in all cases, likely requiring additional run-time branches in cases where sizeof(long long) == sizeof(int). Can we do this without LWG input?

jensmaurer:
Rebased.

PR #2011, opened 2018-04-02
empty xrefdelta.gls

alfmin:
using "make reindex" in a clean directory generates an empty xrefdelta.gls file and therefore no "Cross references from ISO C++ 2017". An additional pdflatex seems needed before "makeindex -s basic.gst -o xrefdelta.gls xrefdelta.glo". Also added to README.rst.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -76,7 +76,9 @@ If you can't use latexmk or make for some reason, you can run LaTeX manually ins
 #. run ``makeindex libraryindex``
 #. run ``makeindex grammarindex``
 #. run ``makeindex impldefindex``
+#. run ``pdflatex std`` once more.
 #. run ``makeindex -s basic.gst -o xrefindex.gls xrefindex.glo``
+#. run ``makeindex -s basic.gst -o xrefdelta.gls xrefdelta.glo``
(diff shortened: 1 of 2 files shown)


tkoeppe:
Thank you, let me try to reproduce this.

zygoloid:
Well, that's what happens when you have multiple build systems for the same project. Perhaps it's time to ditch the Makefile (or replace its commands with invocations of latexmk)?

jensmaurer:
I would be very much in favor of having latexmk as the only supported build method. Shortens Travis CI cycles, too.

tkoeppe:
Thank you! I've stolen your commits and also updated the Travis script, which was also missing the glossary rebuild.

tkoeppe:
@jensmaurer, @zygoloid: I'm not too bothered. We don't really "support" anyone and nobody is hurt by "breakages" like this, so if someone finds it useful, meh, why not. I'd certainly like to keep the set of manual instructions around, since it's important for development and maintenance to know what's going on. I have no opinions on the Makefile and wouldn't mind it not existing, but I also see no harm in having it.

Issue #1917, opened 2018-02-13
Grammar index broken by #1858


tkoeppe:
The changes from #1858 have made caused a large number of index key errors in the grammar index, which show up as warnings in the index log file.

This is bad; we rely on clean logs to spot actual indexing errors. I was very close to adding an index log presubmit check, which would have blocked the application of the pull request.

Can you please fix this? (Perhaps by stripping out the formatting from the index key?)

jensmaurer:
The warnings are like this:

```
## Warning (input = grammarindex.idx, line = 2206; output = grammarindex.ind, line = 8):
   -- Conflicting entries: multiple encaps for the same page under same key.
```
which is caused by index entries like this:
```
\indexentry{access-specifier@\textit  {access-specifier}|idxbfpage}{227}
\indexentry{access-specifier@\textit  {access-specifier}|hyperpage}{227}
```
which means the \grammarterm access-specifier was defined on page 227 and used on page 227. As you can see, the index keys are already the same. I have no idea how to fix this while keeping the (desired) boldness of defining mentions of \grammarterms.

godbyk:
One solution would be to switch from `makeindex` to `xindy`. `xindy` can be instructed to give preference to the bold-faced page numbers. You can read some more details about this option [here](https://tex.stackexchange.com/a/362293/80).  In general, `xindy` is a more capable indexer.

jensmaurer:
Editorial meeting consensus: Let's try xindy.

godbyk:
I can write the xindy script to do this.

When a grammar term is both defined and used on the same page, do we want two page locators (one bold and one roman)? Or do we just want one (bold)?

zygoloid:
Just one (in bold) seems preferable.

godbyk:
@zygoloid Have a look at the grammar terms index with [this configuration](https://github.com/cplusplus/draft/pull/2013) and see if it resolves the issue.

godbyk:
I ran into a potential issue that's holding up my work on this PR. I've described it in detail [elsewhere](https://tex.stackexchange.com/questions/424722/xindy-suppress-cross-reference-target-entry-does-not-exist-error-for-entr).

The short version is that xindy complains about a cross-reference that targets a top-level index entry when it doesn't have any page numbers itself (i.e., it only has page numbers for its subentries).

I'm still hoping to find a solution to that, but in the meantime we have a couple options:

[...]

PR #2013, opened 2018-04-02
[grammar index] Fix warnings; prefer listing bold page locators.

godbyk:
**WORK IN PROGRESS &mdash; DO NOT MERGE YET**

Fixes issue #1917.

For this PR I used xindy only for the grammar terms index. But we could use it for the other indices as well.

I think latexmk may still use makeindex for all the indices. I don't know if it can be configured to use xindy for just the grammar index or not. (I don't use latexmk much. I'll have to check the docs.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ script:
       docker exec -it texlive-basic bash -c "cd /$TRAVIS_REPO_SLUG/source && pdflatex std";
       docker exec -it texlive-basic bash -c "cd /$TRAVIS_REPO_SLUG/source && makeindex generalindex";
       docker exec -it texlive-basic bash -c "cd /$TRAVIS_REPO_SLUG/source && makeindex libraryindex";
-      docker exec -it texlive-basic bash -c "cd /$TRAVIS_REPO_SLUG/source && makeindex grammarindex";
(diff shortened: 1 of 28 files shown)


jensmaurer:
Well, we should definitely get latexmk going. Looks like the "latexmkrc" file needs a setting for "makeindex", pointing to xindy instead. And yes, that will be used for all indexes, not just the grammar index.

godbyk:
I've updated the `latexmkrc` config file so that it runs `xindy`. I also modified it so that it automatically builds with pdfLaTeX and only builds the `std.tex` file. So you can just run `latexmk` with no arguments now and it should do the right thing.

I'm currently fixing up the index cross-references. xindy treats cross-references as first-class objects and will report an error if you refer to a nonexistent index entry. As a consequence of this, the argument to the `see` and `seealso` entries must match the unformatted main index entry to which it refers.  In other words, I'm changing `\index{blah|see{something, else}}` to `\index{blah|see{something!else}}`.  (Alternatively, I can instruct xindy to not check the cross-references. Personally, I think it's probably worth the initial pain as it makes the index syntax more consistent [i.e., always use `entry!subentry`] and helps us find dangling references.)

There are a few other odds and ends I'm going to check up on, too. I'll post progress updates as I go and will tidy up the git commit history before the final merge.

jensmaurer:
I've added a "needs rebase" tag, repurposing it as a general "needs more work" thing.

AlisdairM:
Is this project still active?  It looks like the last update was 2018?

tkoeppe:
Using a better indexer would definitely be nice, yes!

PR #2016, opened 2018-04-02
[stmt.while] Simplify rewrite rule.

### LINKED ISSUE #2008: [stmt.while] no rewrite rule for non-variable condition case
In [stmt.while], we say:

> A `while` statement whose *condition* is an initialized declaration of some variable `t` is equivalent to
> ```
> label:
> {
>   condition ;
>   if (t) {
>     statement
>     goto label;
>   }
> }
> ```

But fail to give any rewrite rule for the other kind of `while` statement. The result is both redundant and underspecified.

We could replace this with:

> A `while` statement is equivalent to
> ```
> label:
> if (condition) {
>   statement
>   goto label;
> }
> ```

... and make the wording simpler, less redundant, and more complete.

(Open question: is it useful to retain the braces around the outside of the if-statement in order to make it obvious that an `else` cannot bind to a `while`?)

jensmaurer:
Fixes #2008.

### DIFF
--- a/source/statements.tex
+++ b/source/statements.tex
@@ -498,18 +498,15 @@ substatement.
 When the condition of a \tcode{while} statement is a declaration, the scope of
 the variable that is declared extends from its point of
 declaration\iref{basic.scope.pdecl} to the end of the \tcode{while}
-\grammarterm{statement}. A \tcode{while} statement whose \grammarterm{condition}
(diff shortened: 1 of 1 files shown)


jensmaurer:
Editorial meeting: The rewrite has a problem, because you can't "goto" in constexpr functions. Ignore that for here. Merge "after motions".

PR #2019, opened 2018-04-03
[dcl.array] Clarify that an array bound is deduced in an explicit typ…

### LINKED ISSUE #2012: [dcl.init.aggr] Does p9 apply to function-style casts?
[\[dcl.init.aggr\]/9](http://eel.is/c++draft/dcl.init.aggr#9):

> An array of unknown bound initialized with a brace-enclosed initializer-list containing n initializer-clauses, where n shall be greater than zero, is defined as having n elements ([dcl.array]).

[\[dcl.array\]/3](http://eel.is/c++draft/dcl.array#3):

> An array bound may also be omitted when the declarator is followed by an initializer or when a declarator for a static data member is followed by a brace-or-equal-initializer ([class.mem]). In both cases the bound is calculated from the number of initial elements (say, N) supplied ([dcl.init.aggr]), and the type of the identifier of D is “array of N T”.

[\[expr.type.conv\]/2](http://eel.is/c++draft/expr.type.conv#2):

> Otherwise, the expression is a prvalue of the specified type whose result object is direct-initialized with the initializer.

Consider

```c++
using T = int[];
auto&& r = T{1,2};
```

Should `r` have type `int(&&)[2]` or `int(&&)[]` (or be ill-formed, since `T` is an incomplete type)?

[dcl.init.aggr]/9 seems to imply that the array bound is deduced in this case, but [dcl.array]/3 and [expr.type.conv]/2 seem more ambiguous.

All compilers I tested say that `r` has type `int(&&)[2]`.

jensmaurer:
…e conversion.

Fixes #2012.

### DIFF
--- a/source/declarators.tex
+++ b/source/declarators.tex
@@ -1065,10 +1065,11 @@ an array bound may be omitted in some cases in the declaration of a function
 parameter\iref{dcl.fct}.
 An array bound may also be omitted
 when the declarator is followed by an
-\grammarterm{initializer}\iref{dcl.init} or
+\grammarterm{initializer}\iref{dcl.init},
 when a declarator for a static data member is followed by a
-\grammarterm{brace-or-equal-initializer}\iref{class.mem}.
-In both cases the bound is calculated from the number
+\grammarterm{brace-or-equal-initializer}\iref{class.mem},
+or in an explicit type conversion\iref{expr.type.conv}.
+In these cases, the bound is calculated from the number
 \indextext{array size!default}%
 of initial elements (say,
 \tcode{N})


jensmaurer:
@zygoloid, please decide whether that is editorial or not. Implementations seem to agree on the intent of the wording.

cpplearner:
I'm actually confused by [dcl.array]/3 (unrelated to #2012). Consider `int (*p)[] = {0};`. It has an array declarator (`(*p)[]`), which is followed by an initializer (`= {0}`). Does it mean that the array bound should be deduced?

jensmaurer:
@cpplearner: The point here is that we're not actually initializing an array object. I agree that could be clearer in the wording.

cpplearner:
I just found [CWG 1525](http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_closed.html#1525) which seems to suggest that [dcl.init.aggr] should apply in an explicit type cast.

zygoloid:
OK, that's good enough for me.

PR #2021, opened 2018-04-03
[move.iterators] Dissolve single-item subclauses.

### LINKED ISSUE #1429: [ostreambuf.iter.ops] not divided into subsections as [istreambuf.iter.ops]

jensmaurer:
Fixes #1429.


tkoeppe:
Looks good, but let's make sure we capture this in an editor's report, and have @mclow emailed.

PR #2022, opened 2018-04-04
[macros, lex, over] Dissolve bnfkeytab environments. Use a fixed-width (\fw{...}) code field instead.

tkoeppe:
Before/after:

image

image

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -792,18 +792,20 @@ The lexical representation of \Cpp{} programs includes a number of
 preprocessing tokens which are used in the syntax of the preprocessor or
 are converted into tokens for operators and punctuators:
 
-\begin{bnfkeywordtab}
+\begin{bnf}
+%% Ed. note: character protrusion would misalign various operators.
+\microtypesetup{protrusion=false}\obeyspaces
 \nontermdef{preprocessing-op-or-punc} \textnormal{one of}\br
-\>\{ \>\} \>[ \>] \>\# \>\#\# \>( \>)\br
-\><: \>:> \><\% \>\%> \>\%: \>\%:\%: \>; \>: \>.{..}\br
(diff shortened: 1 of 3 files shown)


source/lex.tex
```diff
@@ -792,18 +792,20 @@
...
-\><: \>:> \><\% \>\%> \>\%: \>\%:\%: \>; \>: \>.{..}\br
-\>new \>delete \>? \>:: \>. \>.* \>-> \>->* \>\~\br
-\>! \>+ \>- \>* \>/ \>\% \>\caret \>\& \>|\br
-\>= \>+= \>-= \>*= \>/= \>\%= \>\caret= \>\&= \>|=\br
-\>== \>!= \>< \>> \><= \>>= \><=> \>\&\& \>||\br
-\><< \>>> \><<= \>>>= \>++ \>-{-} \>,\br
-\>and \>or \>xor \>not \>bitand \>bitor \>compl\br
-\>and_eq \>or_eq \>xor_eq \>not_eq
-\end{bnfkeywordtab}
+\fw{\{}\fw{\}}\fw{[}\fw{]}\fw{\#}\fw{\#\#}\fw{(}\fw{)}\br
```
godbyk: What about using the `\terminal` macro like in [[lex.name] above](https://github.com/cplusplus/draft/pull/2022/files#diff-a803830b6534424eb14ecf959b859c07R531)? It'd certainly be easier to read the source code that way.

tkoeppe: The spaces collapse, so the output would look very crowded.

godbyk: Oh. Throw in an `\obeyspaces` and then the spaces won't collapse.

tkoeppe: Hah, you win this one. Thanks!

tkoeppe:
I went with the `\terminal` version suggested by @godbyk. Thank you!

PR #2024, opened 2018-04-08
[template.slice.array] Harmonize presentation with neighboring parall…

### LINKED ISSUE #2023: [numarray] Wildly inconsistent formatting for valarray helpers
There are a number of apparently unnecessary inconsistencies between [template.slice.array.overview], [template.gslice.array.overview], [template.mask.array.overview], and [template.indirect.array.overview].

[template.slice.array.overview] p1 says:

>  The `slice_array` template is a helper template used by the `slice` subscript operator

    slice_array<T> valarray<T>::operator[](slice);

And [template.gslice.array.overview] p1 says:

> This template is a helper template used by the `slice` subscript operator
 
    gslice_array<T> valarray<T>::operator[](const gslice&);

Either the latter should say "`gslice` subscript operator" or the word `slice` should not be in code font (which would be consistent with "mask subscript operator" and "indirect subscript operator" in  [template.mask.array.overview] and  [template.indirect.array.overview] respectively).

Also [template.slice.array.overview] p1 says "The `slice_array` template is a helper template" but the other three helpers say "This template is a helper template". Presumably they could be consistent.

The "It has reference semantics ..." wording is part of p1 in [template.slice.array.overview] but is a separate paragraph p2 in the other three places.

The explanation of the effects ("the expression ... has the effect of ...") is an _Example:_ in [template.slice.array.overview], its own paragraph in [template.gslice.array.overview], and part of p2 in [template.mask.array.overview] and [template.indirect.array.overview].

jensmaurer:
…el statements

Fixes #2023.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -7818,21 +7818,16 @@ namespace std {
 \end{codeblock}
 
 \pnum
-The
-\tcode{slice_array}
-template is a helper template used by the
-\tcode{slice}
-subscript operator
-
+This template is a helper template used by the
+\tcode{slice} subscript operator
 \begin{codeblock}
 slice_array<T> valarray<T>::operator[](slice);
 \end{codeblock}
 
+\pnum
 It has reference semantics to a subset of an array specified by a
 \tcode{slice}
 object.
-
-\pnum
 \begin{example}
 The expression
 \tcode{a[slice(1, 5, 3)] = b;}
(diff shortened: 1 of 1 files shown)


source/numerics.tex
```diff
@@ -7818,21 +7818,18 @@
 \end{codeblock}
 
 \pnum
-The
-\tcode{slice_array}
-template is a helper template used by the
+This template is a helper template used by the
 \tcode{slice}
 subscript operator
```
tkoeppe: This newline can go, too. Otherwise fine.

tkoeppe: (Although I wonder why need to restate the member declaration here rather than just have a "usual" reference to the corresponding section [valarray.sub].)

jensmaurer: We restate the member declaration all the time.

PR #2027, opened 2018-04-13
Use \commentellip macro for omitted program text.

### LINKED ISSUE #1390: Comment out ellipses when they mean "more code here" rather than an actual ellipsis token.

jensmaurer:
Fixes #1390.

### DIFF
--- a/source/declarators.tex
+++ b/source/declarators.tex
@@ -75,12 +75,12 @@ One exception is when a name introduced by one of the
 \grammarterm{decl-specifier}{s} are used in a subsequent declaration,
 they do not have the same meaning, as in
 \begin{codeblock}
-struct S { ... };
+struct S { @\commentellip@ };
 S S, T;                 // declare two instances of \tcode{struct S}
 \end{codeblock}
 which is not equivalent to
 \begin{codeblock}
-struct S { ... };
+struct S { @\commentellip@ };
 S S;
 S T;                    // error
 \end{codeblock}
@@ -2027,7 +2027,7 @@ is the
 \tcode{c)}
 is the
 \grammarterm{declarator};
-\tcode{\{ /* ...\ */ \}}
+\tcode{\{ \commentellip{} \}}
 is the
 \grammarterm{function-body}.
 \end{example}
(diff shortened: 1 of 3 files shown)


source/lib-intro.tex
```diff
@@ -606,13 +606,13 @@
 The enumerated type \tcode{\placeholder{enumerated}} can be written:
 
 \begin{codeblock}
-enum @\placeholder{enumerated}@ { @$\tcode{\placeholder{V}}_{0}$@, @$\tcode{\placeholder{V}}_{1}$@, @$\tcode{\placeholder{V}}_{2}$@, @$\tcode{\placeholder{V}}_{3}$@, ..... };
+enum @\placeholder{enumerated}@ { @$\tcode{\placeholder{V}}_{0}$@, @$\tcode{\placeholder{V}}_{1}$@, @$\tcode{\placeholder{V}}_{2}$@, @$\tcode{\placeholder{V}}_{3}$@, @$\cdots$@ };
```
tkoeppe: Are these "center dots"? Why not dots on the baseline (\ldots)? Center dots are for omitted operands of binary folds, like "1 + 2 + \cdots + n".

tkoeppe:
I take it all the \commentellip uses are in non-normative code?

source/lib-intro.tex
tkoeppe: That sounds like (it will have been) an oversight that we should fix. Sure, omit this file from the change and I'll commit the rest.

jensmaurer:
@tkoeppe: Reduced pull request to just the \commentellip changes; the remainder is covered by #2031.

PR #2028, opened 2018-04-13
[time.syn][time.zone] Various editorial fixes

timsong-cpp:
- Add missing semicolons.
- Write out `operator!=` for `zoned_time`. [operators] does not apply because this comparison is heterogeneous (can have different `Duration` types).
- Fixes `TimeZonePtr2`-related typos in `zoned_time`.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -18288,7 +18288,7 @@ namespace std {
       bool operator==(const zoned_time<Duration1, TimeZonePtr>& x,
                       const zoned_time<Duration2, TimeZonePtr>& y);
 
-    template<class Duration1, class Duration, class TimeZonePtr2>
+    template<class Duration1, class Duration2, class TimeZonePtr>
       bool operator!=(const zoned_time<Duration1, TimeZonePtr>& x,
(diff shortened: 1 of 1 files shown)


tkoeppe:
Looks good to me, thanks! I've asked Howard to take a look.

source/utilities.tex
```diff
@@ -27272,7 +27272,7 @@
 \remarks
 This constructor does not participate in overload resolution unless
 \begin{codeblock}
-declval<TimeZonePtr&>()->to_sys(local_time<Duration>{})
+decltype(declval<TimeZonePtr&>()->to_sys(local_time<Duration>{}))
```
zygoloid: Do we really need / want a `decltype` here? "is (implicitly) convertible to" is defined for an expression and a destination type, not for a source type and a destination type (this actually matters in some cases, such as when the source expression is a bit-field and the destination is a reference type -- but probably not in this case).

I've opened #2029 for the general problem with the library using the meaningless term "is convertible to".

timsong-cpp: I mostly wanted local consistency and to avoid suggesting that the expression is actually evaluated.

PR #2031, opened 2018-04-14
[enumerated.types,bitmask.types] Use \vdots and \ldots instead of ...…

jensmaurer:
….. mega-ellipses.

Partially addresses #1390.

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -606,13 +606,13 @@ values\iref{basic.fundamental}.}
 The enumerated type \tcode{\placeholder{enumerated}} can be written:
 
 \begin{codeblock}
-enum @\placeholder{enumerated}@ { @$\tcode{\placeholder{V}}_{0}$@, @$\tcode{\placeholder{V}}_{1}$@, @$\tcode{\placeholder{V}}_{2}$@, @$\tcode{\placeholder{V}}_{3}$@, ..... };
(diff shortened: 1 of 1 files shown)


tkoeppe:
@zygoloid, @jensmaurer: I think we really want AmsMath's \dotsc behaviour for comma-separated lists, not blanket \cdots, and the guidance in #1390 should be adjusted accordingly.

jensmaurer:
@zygoloid said in #1390 that we wanted "center dots", but in this case, \ldots look substantially better. I've updated the pull request to use \ldots, so this can be applied right away once @zygoloid has changed his opinion.
@tkoeppe: We use both \ldots (45 appearances) and \dotsc (33 appearances). We should decide on one form. I've created #2032 to follow up on this.

tkoeppe:
@jensmaurer: they might not quite the same, though. \ldots is kind of generic, "presentational" form. \dotsc is specifically for use in math mode, I think: $x_0, x_1, \dotsc, x_n$. We should look more carefully at how \ldots and \dotsc are currently used; the status quo may already be WAI.

jensmaurer:
The predominant usage of \ldots in our standard is within math mode with comma-separated lists of things. The same statement applies to \dotsc.
As a first step, we should probably move towards \dotsc for comma-separated lists of things and see what remains.
@zygoloid?

zygoloid:
Let's move the discussion to #1390. This looks good to merge as-is, as a start down this road.

PR #2034, opened 2018-04-15
[class.spaceship] Use math mode for the entire formula,

jensmaurer:
not just for the subscript.
Also use \dotsc, not center dots, for comma-separated lists.

### DIFF
--- a/source/special.tex
+++ b/source/special.tex
@@ -3338,16 +3338,16 @@ in the order of their declaration in the \grammarterm{member-specification} of \
 form a list of subobjects.
 In that list, any subobject of array type is recursively expanded
 to the sequence of its elements, in the order of increasing subscript.
-Let \tcode{x}$_i$ be an lvalue denoting the $i^\textrm{th}$ element
+Let $\tcode{x}_i$ be an lvalue denoting the $i^\text{th}$ element
 in the expanded list of subobjects for an object \tcode{x}
 (of length $n$),
-where \tcode{x}$_i$ is
+where $\tcode{x}_i$ is
 formed by a sequence of
 derived-to-base conversions\iref{over.best.ics},
 class member access expressions\iref{expr.ref}, and
 array subscript expressions\iref{expr.sub} applied to \tcode{x}.
-The type of the expression \tcode{x}$_i$ \tcode{<=>} \tcode{x}$_i$
-is denoted by \tcode{R}$_i$.
+The type of the expression $\tcode{x}_i$ \tcode{<=>} $\tcode{x}_i$
+is denoted by $\tcode{R}_i$.
 It is unspecified
 whether virtual base class subobjects are compared more than once.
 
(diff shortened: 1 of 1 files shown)


source/special.tex
```diff
@@ -3372,46 +3372,46 @@
...
-\tcode{V} is \tcode{v}$_i$ converted to \tcode{R}.
+\tcode{V} is $\tcode{v}_i$ converted to \tcode{R}.
 If no such index exists, \tcode{V} is
 \tcode{std::strong_ordering::equal} converted to \tcode{R}.
 
 \pnum
 The \defn{common comparison type} \tcode{U}
 of a possibly-empty list of $n$ types
-\tcode{T}$_0$, \tcode{T}$_1$, $\cdots$, \tcode{T}$_{n-1}$
+$\tcode{T}_0$, $\tcode{T}_1$, $\dotsc$, $\tcode{T}_{n-1}$
```
tkoeppe: I think I'd prefer a non-math \ldots here, since we don't have math commas.

source/special.tex
```diff
@@ -3372,46 +3372,46 @@
...
 with parameters \tcode{x} and \tcode{y} of the same type
 is determined by comparing corresponding elements
-\tcode{x}$_i$ and \tcode{y}$_i$
+$\tcode{x}_i$ and $\tcode{y}_i$
 in the expanded lists of subobjects for \tcode{x} and \tcode{y}
 until the first index $i$
-where \tcode{x}$_i$ \tcode{<=>} \tcode{y}$_i$
-yields a result value \tcode{v}$_i$ where \tcode{v}$_i$\tcode{ != 0},
+where $\tcode{x}_i$ \tcode{<=>} $\tcode{y}_i$
+yields a result value $\tcode{v}_i$ where $\tcode{v}_i \tcode{!= } 0$,
```
tkoeppe: I take it you've checked that the spacing after != looks right?

source/special.tex
jensmaurer: We're doing this style elsewhere, too: "$ grep dotsc *.tex"

source/special.tex
tkoeppe: eek. Add a `\,`-space before?

tkoeppe: Or see how it looks without the space in the tcode.

godbyk: I propose using `\mathrel`: `$\tcode{v}_i$ where $\tcode{v}_i \mathrel{\tcode{!=}} 0$`.  `\mathrel` provides the same spacing around its contents as TeX provides around other math relations (e.g., =, >).

jensmaurer: Just plain spaces works:
space1

jensmaurer: And, even better, with \mathrel:
space1

Update pushed.

tkoeppe: Agreed, \mathrel is the right thing. Thanks, Kevin!

source/special.tex
tkoeppe: Hm, that's unfortunate, but fine.

PR #2035, opened 2018-04-15
[cpp, lex, over] Make bnf indentation in source match presentation.

Eelis:
Same as #1362:
- makes the source look nicer and more consistent
- no visual difference in PDF
- makes life easier for cxxdraft-htmlgen


Eelis:
Thanks!

PR #2040, opened 2018-04-16
[over.oper] Remove incorrect and redundant sentence in a note.

### LINKED ISSUE #2038: The Note in [over.oper]/1 requires a change
[[over.oper]/1](http://eel.is/c++draft/over.oper#1)

[ Note: The **last two** operators are function call (8.5.1.2) and subscripting (8.5.1.1). The operators new[],
delete[], (), and [] are formed from more than one token. —end note ]

The words **last two** above are obviously wrong.

jensmaurer:
Fixes #2038.

### DIFF
--- a/source/overloading.tex
+++ b/source/overloading.tex
@@ -3072,8 +3072,6 @@ the operator named in its
     \terminal{<<       >>       <<=      >>=      ++       --       ,}\br
 \end{bnf}
 \begin{note}
-The last two operators are function call\iref{expr.call}
-and subscripting\iref{expr.sub}.
 The operators
 \tcode{new[]},
 \tcode{delete[]},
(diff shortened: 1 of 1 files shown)


source/overloading.tex
```diff
@@ -3084,8 +3084,6 @@
     \terminal{<<       >>       <<=      >>=      ++       --       ,}\br
 \end{bnf}
 \begin{note}
-The last two operators are function call\iref{expr.call}
-and subscripting\iref{expr.sub}.
```
zygoloid: Maybe move this to the end of the note, and change "the last" to "the latter"? I think it's useful to mention in passing here what `operator()` and `operator[]` mean.

PR #2043, opened 2018-04-19
[temp.arg.explicit,temp.mem] Clarify note about explicit template arg…

### LINKED ISSUE #2041: [temp.arg.explicit] Misleading note about explicit template arguments on constructors
The note in paragraph 7 says "Because ...  constructor member function templates are called without using a function name, there is no way to provide an explicit template argument list for these function templates."

This reasoning falls short for the other situations where a constructor template may be named, e.g. in an explicit instantiation or specialization.

I'd suggest to replace the "called" with "named" and add cross-references to [class.ctor] and [class.mem].

There is 12-year-old [core issue 581](http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_active.html#581) on the subject whether explicit instantiations or specializations of constructors should be allowed to use explicit template arguments, but the status quo normative situation in 17.9.1 [temp.arg.explicit] p1 seems pretty clear ("function template name").

jensmaurer:
…uments

for conversion function templates and constructor templates.

Fixes #2041.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -2472,11 +2472,9 @@ int main() {
 \end{codeblock}
 \end{example}
 \begin{note}
-Because the explicit template argument list follows the function template
-name, and because conversion member function templates and constructor
-member function templates are called without using a function name,
-there is no way to provide an explicit template argument list for these
-function templates.
+There is no syntax to form a \grammarterm{template-id}\iref{temp.names}
+by providing an explicit template argument list\iref{temp.arg.explicit}
+for a conversion function template\iref{class.conv.fct}.
 \end{note}
 
 \pnum
@@ -6978,10 +6976,10 @@ void g() {
 \pnum
 \begin{note}
 Because the explicit template argument list follows the function
-template name, and because conversion member function templates and
-constructor member function templates are called without using a
-function name, there is no way to provide an explicit template
-argument list for these function templates.
+template name, and because
+constructor templates\iref{class.ctor} are named without using a
+function name\iref{class.qual}, there is no way to provide an explicit
+template argument list for these function templates.
 \end{note}
 
 \pnum


source/templates.tex
```diff
@@ -2472,9 +2472,11 @@
 \end{codeblock}
 \end{example}
 \begin{note}
+%% Ed. note: Using \iref would cause an overflowing \hbox; changing
```
zygoloid: I thought we'd fixed `\iref` so that we *can* break before it, just with a high penalty? :(

source/templates.tex
```diff
@@ -2472,9 +2472,11 @@
...
+%% the sentence would lose the parallelism to [temp.mem] p5.
 Because the explicit template argument list follows the function template
-name, and because conversion member function templates and constructor
-member function templates are called without using a function name,
+name (\ref{temp.arg.explicit}), and because conversion member function templates\iref{class.conv.fct}
+are named without using a function name,
 there is no way to provide an explicit template argument list for these
 function templates.
```
zygoloid: The wording that remains here is wrong, isn't it? The reason you can't provide an explicit template argument list to a conversion function is simply because grammar productions for *template-id* do not include a production for a *conversion-function-id* followed by a *template-argument-list*. (The reason for *that* is presumably the visual ambiguity between template arguments of the conversion function and template arguments of its type, combined with there being no way to write `operator T </*T=*/int>`, since you can't name `T`.) But it seems simply wrong to say that conversion member function templates are called without using a function name; they can be.

source/templates.tex
jensmaurer: Except that, when I tried it, it didn't work in this case.

zygoloid: Perhaps we could add an optional parameter to `\iref` to specify its `\nolinebreak` penalty. I'm sad that it doesn't "just work" here; it definitely does work in other places, such as [class.dtor]p14, but IIRC turning down the line break penalty any further globally causes TeX to wrap inline references even in places where it's unnecessary to do so.

zygoloid: From https://tex.stackexchange.com/questions/94217/what-is-the-difference-between-nobreak-and-nolinebreak it looks like we're probably effectively using `\penalty 301` currently (that's for article, but I'd guess memoir is probably not very different). Maybe we could binary search to find whatever value works in this case and try changing `\iref` to use something like that.

source/templates.tex
jensmaurer: I don't like these "because" sentences in any case.

source/templates.tex
```diff
@@ -6978,8 +6977,8 @@
 \pnum
 \begin{note}
 Because the explicit template argument list follows the function
-template name, and because conversion member function templates and
-constructor member function templates are called without using a
+template name, and because
+constructor member function templates\iref{class.ctor} are named without using a
```
zygoloid: Is the change "are called" -> "are named" here intentional? I think it's clear that (for example) the expression `A()` calls a constructor without using a function name, and more generally, there is no syntax to call a constructor using a function name. However, it's not clear to me that `template<typename T> A::A(T t)` names the constructor without using a function name -- in that context, `A::A` names the constructor, and the constructor is a function, so arguably `A::A` is a function name.

source/templates.tex
```diff
@@ -2472,11 +2472,10 @@
...
 \begin{note}
-Because the explicit template argument list follows the function template
-name, and because conversion member function templates and constructor
-member function templates are called without using a function name,
-there is no way to provide an explicit template argument list for these
-function templates.
+There is no syntax to form a \grammarterm{template-id}\iref{temp.names}
+by providing an explicit template argument
+list\iref{temp.arg.explicit} for a conversion member function
+template\iref{class.conv.fct}.
```
zygoloid: "conversion member function template" is an awkward phrase, as it splits the term "conversion function". The "member" here seems redundant, since every conversion function [template] is already a member function [template]. So how about replacing that with just "conversion function template"?

source/templates.tex
zygoloid: Instead of "constructor member function templates", I think "constructor templates" would be clearer and easier to understand.

jensmaurer: [temp.arg.explicit] p2 lists the circumstances under which an explicit template argument list can be supplied for a function template. "A function is called" is just one of five options. I agree that "named" is bad term here, though, because we use "naming the constructor" when we want to weasel-word around the fact that constructors don't have names to start with. p1 talks about "function template name", and the note wants to highlight that a strict reading of "name" is required here (and thus constructors can't have template argument lists).

source/templates.tex
jensmaurer: It was there before...

source/templates.tex
jensmaurer: Again, pre-existing junk.

jensmaurer:
Updated.

PR #2047, opened 2018-04-20
[over.match,over.match.ref] Drop obsolete mention of class prvalues.

### LINKED ISSUE #2046: [over.match] Why "glvalue or class prvalue"?
[\[over.match\]/(2.7)](http://eel.is/c++draft/over.match#2.7):

> invocation of a conversion function for conversion to a glvalue or class prvalue to which a reference ([dcl.init.ref]) will be directly bound.

[\[over.match.ref\]/1](http://eel.is/c++draft/over.match.ref#1.sentence-1):

> Under the conditions specified in [dcl.init.ref], a reference can be bound directly to a glvalue or class prvalue that is the result of applying a conversion function to an initializer expression.

In C++17, reference initialization no longer distinguishes class prvalue and non-class prvalue. Should these two sentences be changed?

jensmaurer:
Fixes #2046.

### DIFF
--- a/source/overloading.tex
+++ b/source/overloading.tex
@@ -439,9 +439,8 @@ copy-initialization\iref{dcl.init} of a class object\iref{over.match.copy};
 invocation of a conversion function for initialization of an object of a
 non-class type from an expression of class type\iref{over.match.conv}; and
 \item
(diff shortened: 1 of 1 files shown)


jensmaurer:
Updated.

PR #2052, opened 2018-04-26
Order library comparisons canonically

CaseyCarter:
The order [`==`, `!=`, `<`, `>`, `<=`, `>=`] seems to be most common. Let's make it canonical and use it uniformly.


source/iostreams.tex
```diff
@@ -14939,7 +14939,7 @@
 
 \pnum
 \begin{note} The current path as returned by many operating systems is a dangerous
-  global variable. It may be changed unexpectedly by a third-party or system
+  global variable. It may be changed unexpectedly by third-party or system
```
zygoloid: For future reference, I'd generally prefer for unrelated fixes such as this and the "duration" -> "duration's" change below to be provided as a separate pull request.

PR #2054, opened 2018-04-27
[over.match.copy] Clarify candidate function selection for references.

### LINKED ISSUE #2053: [over.match.copy]: "return lvalues or xvalues" makes me uncomfortable
**UPD**:  [over.match.copy] and [over.match.conv]

The sections [over.match.copy] and [over.match.conv] have statements like: "Conversion functions that return “reference to X” return lvalues or xvalues ..."

[diff.expr]/4 is also broken: "C++ is an object-oriented language, placing relatively more emphasis on lvalues. For example, functions may return lvalues."

"return _value category_" sounds illiterate for me.

Proposed wording:

```
The result of calling a
conversion function that returns ``reference to
\tcode{X}''
is
lvalue or xvalue, depending on the type of reference, of type
\tcode{X}
and is therefore considered to yield
\tcode{X}
for this
process of selecting candidate functions.
```

jensmaurer:
Fixes #2053.

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -281,8 +281,8 @@ Seldom.
 \change The result of a conditional expression, an assignment expression, or a comma expression may be an lvalue.
 \rationale
 \Cpp{} is an object-oriented language, placing relatively
-more emphasis on lvalues.  For example, functions may
-return lvalues.
+more emphasis on lvalues.  For example, function calls may
+yield lvalues.
 \effect
 Change to semantics of well-defined feature.  Some C
 expressions that implicitly rely on lvalue-to-rvalue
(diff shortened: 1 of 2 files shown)


languagelawyer:
[over.match.conv] also needs the fix!

zygoloid:
I agree, we should update [over.match.conv] too (though I'm not sure what needs changes in [diff.expr]/4?). I also think it would be reasonable to convert the wording in question to a note.

languagelawyer:
> though I'm not sure what needs changes in [diff.expr]/4

I've cited in #2053:
> C++ is an object-oriented language, placing relatively more emphasis on lvalues.
[For example, **functions may return lvalues**.](http://eel.is/c++draft/diff.expr#4.sentence-3)

jensmaurer:
Updated.

zygoloid:
Looks good, thanks. Did you consider and reject the idea of turning this into a note?

jensmaurer:
@zygoloid: For the overload resolution candidates, I believe the text needs to be normative (the remaining text, taking verbatim, seems to exclude conversion to references).

PR #2061, opened 2018-05-08
[atomics.types.operations] Avoid inappropriate use of 'underlying type'

### LINKED ISSUE #2057: [atomics.types.operations] misuses "underlying type"
The note in [atomics.types.operations] p23 says "if the underlying type has ..." but this is not the same meaning of "underlying type" used elsewhere ([basic.fundamental] and [dcl.enum]).

[atomics.types.generic] uses the phrase "the template argument for `T`", and the nested type `value_type` is also defined. Either would be better than "the underlying type", but I think the note would be improved by referring to "object representation of the type", possibly even making use of `std::has_unique_object_representations`.

jensmaurer:
Fixes #2057.

### DIFF
--- a/source/atomics.tex
+++ b/source/atomics.tex
@@ -1661,7 +1661,7 @@ would not, the strong one is preferable.
 \begin{note}
 Under cases where the \tcode{memcpy} and \tcode{memcmp} semantics of the compare-and-exchange
 operations apply, the outcome might be failed comparisons for values that compare equal with
-\tcode{operator==} if the underlying type has trap bits or alternate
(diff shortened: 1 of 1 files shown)


jensmaurer:
Rebased.

zygoloid:
Given the recent changes to compare-and-swap semantics, would it now be more useful to say "value representation" instead of "object representation"? (The memcpy / memcmp semantics only apply when the value representation and object representation are the same, but it may be more obvious if we reinforce that here.)

jensmaurer:
@zygoloid, agreed. Fixed.

PR #2065, opened 2018-05-11
Improve indexing for "trivial class" et al

jwakely:


source/classes.tex
```diff
@@ -141,6 +141,8 @@
 \item that has a trivial, non-deleted destructor\iref{class.dtor}.
 \end{itemize}
 
+\indextext{trivial class|see{class, trivial}}%
```
tkoeppe: @jwakely: The `\indextext` usually goes *after* the `\pnum` -- please fix?

PR #2066, opened 2018-05-11
[class] Split paragraph defining standard-layout class

jwakely:
This paragraph seems excessively long:

spectacle tj2604

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -166,14 +166,15 @@ for all non-static data members,
 \item has all non-static data members and bit-fields in the class and
 its base classes first declared in the same class, and
 
-\item has no element of the set $M(\mathtt{S})$ of types (defined below)
-as a base class.\footnote{This ensures that two subobjects that have the
-same class type and that
+\item has no element of the set $M(\mathtt{S})$ of types
+as a base class,
+where for any type \tcode{X}, $M(\mathtt{X})$ is defined as follows.\footnote{
+This ensures that two subobjects that have the same class type and that
 belong to the same most derived object are not allocated at the same
 address\iref{expr.eq}.}
-\end{itemize}
+\begin{note} $M(\mathtt{X})$ is the set of the types of all non-base-class subobjects
+that may be at a zero offset in \tcode{X}. \end{note}
 
-$M(\mathtt{X})$ is defined as follows:
 \begin{itemize}
 \item If \tcode{X} is a non-union class type with no (possibly
 inherited\iref{class.derived}) non-static data members, the set
@@ -198,10 +199,9 @@ and the elements of $M(\mathtt{X}_e)$.
 
 \item If \tcode{X} is a non-class, non-array type, the set $M(\mathtt{X})$ is empty.
 \end{itemize}
+\end{itemize}
 
-\begin{note} $M(\mathtt{X})$ is the set of the types of all non-base-class subobjects
-that may be at a zero offset in \tcode{X}. \end{note}
-
+\pnum
 \begin{example}
 \begin{codeblock}
    struct B { int i; };         // standard-layout class


source/classes.tex
```diff
@@ -201,6 +202,7 @@
 \item If \tcode{X} is a non-class, non-array type, the set $M(\mathtt{X})$ is empty.
 \end{itemize}
 
+\pnum
 \begin{note} $M(\mathtt{X})$ is the set of the types of all non-base-class subobjects
 that may be at a zero offset in \tcode{X}. \end{note}
```
tkoeppe: @jwakely: Please remove this secret intra-paragraph paragraph break here.

tkoeppe:
Actually, I'm not so sure about this split: *M(X)* is only defined inside this single paragraph at the moment. With your change, we would leak this expository device into the whole-subclause scope, if you see what I mean.

jwakely:
I see what you mean, but the current formatting is ... let's just say it leaves a lot to be desired.

Why do we say ("defined below)" if it's _right there_ in the next sentence? We already say "defined as follows" in that next sentence, how many times do we need a teaser trailer for that definition?

Why do we have that "M(X) is defined as follows" floating in the middle of the list, unattached to any single bullet?

Why is (8.8) a continuation of the same list, if it only applies to (8.7)? Wouldn't the bullets defining M(X) it be better as a nested list, e.g. (8.7.1), (8.7.2) and so on? I know we can't do that with our current `itemize` env, but here's what it looks like using `enumerate` instead:

std-layout-list

jwakely:
(That doesn't change the fact it's excessively long, but I think it improves the structure)

jensmaurer:
@jwakely, the "enumerate" stuff is a regression, not an improvement. I'm ok with a nested "itemize" to define M(S). And when you do that, the "(defined below)" should really go. Asking @zygoloid for an opinion.

tkoeppe:
I like the last version, with the M(X) part of the same paragraph where it's used, and the example separate.

I'd phrase the relevant item as "has no element of the set M(S) of types as a base class, where M(X) is defined as follows:".

tkoeppe:
Or even:

> has no element of the set M(S) of types as a base class, where for any type X, the set M(X) is defined as follows. [Note: M(X) is the set ... zero offset in X -- end note.]".

jwakely:
I'm not suggesting we actually use `enumerate`, obviously we don't want numbered bullets. When I tried a nested `itemize` it didn't number the nested bullets properly so I just used `enumerate` for argument's sake. But I see what I did wrong now (I still had the new `\pnum` after bullet (8.7)). Here's what I meant all along (nested list, new paragraph for the example, and no "(defined below)":

std-layout-list3

We could even join the two sentences in bullet 7:

std-layout-list3a

tkoeppe:
I mean the last screenshot in https://github.com/cplusplus/draft/pull/2066#issuecomment-400653871. But what you have now is equivalent for the purpose of the stuff I was talking about.

jensmaurer:
This is a drive-by, but we really should say what X is in M(X): "For a type X, M(X) is defined as follows..."

tkoeppe:
@jensmaurer: I just suggested that above :-)

zygoloid:
+1 to something in the direction of @tkoeppe's suggestion:

> has no element of the set M(S) of types as a base class, where for any type X, the set M(X) is defined as follows. [Note: M(X) is the set ... zero offset in X -- end note.]".

tkoeppe:
Oh, can we fold the footnote into the note?

zygoloid:
The footnote is a somewhat imprecise rationale for the rule, so I think it would need quite a lot of massaging if we wanted to promote it to a note. Maybe we should consider removing it. But I think this PR is already a significant improvement, so I think we should merge it as-is and leave the footnote for future cleanup.

PR #2070, opened 2018-05-11
[rand] Use itemize envs instead of enumeratea

jwakely:


source/numerics.tex
```diff
@@ -4009,7 +4009,7 @@
  each indexing operator applied to \tcode{begin} is to be taken modulo $n$,
  and $T(x)$ is defined as $x \, \xor \, (x \, \rightshift \, 27)$:
```
godbyk: Unrelated to the existing patch, but the extra spacing here caught my eye. The `\xor` and `\rightshift` macros already include thin spaces around them. May I recommend removing all instances of `\,` both in the text here and in the `\xor` and `\rightshift` (and similar) macros and instead using `\mathbin` to handle the spacing?

source/numerics.tex
```diff
@@ -4068,7 +4068,7 @@
    update $\tcode{begin[}k+q\tcode{]}$ by xoring it with $r_4$,
    and
    set $\tcode{begin[}k\tcode{]}$ to $r_4$.
```
godbyk: This could be written a bit more clearly as `set \tcode{begin[$k$]} to $r_4$.` (i.e., don't set the whole `\tcode` bit in math mode). Similarly, a couple lines previous.

source/numerics.tex
godbyk: Update: I created PR #2073 to address this issue.

jensmaurer:
@jwakely: Ping.

source/numerics.tex
tkoeppe: Thanks, this even catches some missing maths formatting!

@jwakely: I'm stealing your PR.

tkoeppe:
Stolen as 615a00855ef32814be590ce8d88a3cd0440cb66a.

Issue #653, opened 2016-03-18
"ith" should be consistently formatted


zygoloid:
Sometimes we use "ith", sometimes we use "$i^{th}$", etc.

tkoeppe:
If you're going with the maths notation, then it should be `$i^\text{th}$`.

tkoeppe:
Following 48efdfe014549e6d116ce12ca427f7df0e0f8c70, there's still work left to be done in the later parts of  [tuple.tuple].

tkoeppe:
(I've been following my recommended style on and off when I happen across relevant text, though I haven't actively sought them out.)

godbyk:
@tkoeppe Would it be better to create an `\ith` macro to help ensure consistency?

tkoeppe:
@godbyk: I'd be strongly against that. Text and maths is the bread and butter of LaTeX, and `$i^\text{th}$` is perfectly readable. Every new macro carries a heavy mental cost and needs to be carefully justified: readers need to know what it means, writers need to remember to use it, and the more trivial the subject the more arbitrary the macros become. We also don't define macros for `the behavior is undefined`, say, even though that's a much more common phrase. Macros shouldn't just zlib-compress the text, they should factor out conceptual units.

godbyk:
@tkoeppe  My proposed macro isn't about compression or saving a few keystrokes; it's about ensuring consistency.

I propose that it's easier for an author less accustomed to LaTeX and the finer points of typesetting to remember `\ith` than `$i^\text{th}$`.  After all, if everyone naturally typed `$i^\text{th}$`, this bug report wouldn't exist.

Macros also offer more flexibility when it comes to changing styles. If you later decide that `$i\text{th}$` is better than `$i^\text{th}$`, for instance, you must only change it in one place.

A better counter-argument would be, “What about `$j^\text{th}$` and `10000^\text{th}` and `$\tcode{I}^\text{th}$`?”  And for that I'd still recommend a macro that appends <sup>th</sup>. For example:

```latex
\let\thorn\th% stealing \th for our own purposes
\def\th{\ensuremath{{}^\text{th}}}

$i\th$  \qquad  \tcode{N}\th  \qquad  7\th
```

PR #2071, opened 2018-05-11
Unify typesetting of ordinals.

godbyk:
This PR unifies the typesetting of ordinals per the guidance provided in issue #653.


W-E-Brown:
> On May 11, 2018, at 5:58 PM, Kevin Godby <notifications@github.com> wrote:
> 
> This PR unifies the typesetting of ordinals per the guidance provided in issue #653.
> 
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/cplusplus/draft/pull/2071
> 
> Commit Summary
> 
> 	• Unify typesetting of ordinals.
> File Changes
> 
> 	• M source/classes.tex (2)
> 	• M source/containers.tex (2)
> 	• M source/declarators.tex (12)
> 	• M source/numerics.tex (20)
> 	• M source/overloading.tex (4)
> 	• M source/templates.tex (10)
> 	• M source/utilities.tex (6)
> Patch Links:
> 
> 	• https://github.com/cplusplus/draft/pull/2071.patch
> 	• https://github.com/cplusplus/draft/pull/2071.diff

This proposed patch seems to me to go beyond unifying typesetting.

The typesetting changes seem fine, but changing "st" to "th" in some ordinal expressions seems not okay, especially without drawing attention to such.  I these adjustments as not at all a purely typesetting matter.

There's no English word "oneth"; we say "first" instead.  While I know that some (mostly non-mathematicians) disagree, the ordinal version of "m+1" is pronounced "m plus first" rather than "m + oneth"; therefore, please revert the corresponding parts of this proposed patch and restore "st" as the superior characters in these cases.  (See line 1578 of file source/overloading.tex for one example; there are several more in the proposed patch.)

And yes, if we had ordinal versions of "m+2" or "m+3" I would want the corresponding superior characters to be "nd" and "rd", respectively; there's no modern English "twoth" or "threeth" AFAIK.

godbyk:
@W-E-Brown I restored the two `st` ordinals in `overloading.tex`. As far as I noticed, they were the only two there. Did you notice any other instances?

W-E-Brown:
> On May 11, 2018, at 8:46 PM, Kevin Godby <notifications@github.com> wrote:
> 
> @W-E-Brown I restored the two st ordinals in overloading.tex. As far as I noticed, they were the only two there. Did you notice any other instances?

I had thought there were some in numerics, but I don't now see any there.  Thank you.

zygoloid:
For posterity: I was curious why we're changing from `\textrm` and `\mathrm` to `\text`, and what the differences between them are. That led me to this: https://tex.stackexchange.com/questions/70632/difference-between-various-methods-for-producing-text-in-math-mode -- from which I conclude that `\text` (+ the `amsmath` package, which we already use) is the best choice.

PR #2075, opened 2018-05-12
Improved typesetting of probability functions.

godbyk:
By using the new `\p` and `\P` macros, probability functions will be typeset
more consistently and cleanly.  The syntax of both macros is the same.
Examples:

    \p{z}
    \p{z | a, b}
    \p*{z | \frac{1}{2}}
    \p[\big]{\frac{A}{B}}

The pipe (`|`) will have the proper spacing automatically.
The `\p*` variant will autoscale the parentheses and vertical bar to fit the contents of the expressions.
You can specify a particular size using the `\p[<size>]` variant.
Possible sizes are `\big`, `\Big`, `\bigg`, and `\Bigg`.

The `p` or `P` is set in roman instead of italics, but this can be changed if you prefer _p_ and _P_.

### DIFF
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -220,6 +220,30 @@
 \newcommand{\rightshift}{\ensuremath{\mathbin{\mathsf{rshift}}}}
 \newcommand{\leftshift}[1]{\ensuremath{\mathbin{\mathsf{lshift}_{#1}}}}
 
+%% Probability functions
+\RequirePackage{mathtools}
+\makeatletter
+\newcommand{\std@makeactivechar}[2]{%
+  \begingroup
+  \lccode`\~=`#1%
+  \lowercase{\endgroup\def~}{#2}%
+  \mathcode`#1="8000%
+}
+\DeclarePairedDelimiterX\p@aux[1](){%
+  \begingroup
+    \std@makeactivechar{|}{\mathrel{\delimsize\vert}}%
+    #1%
+  \endgroup
+}
+\let\pilcrow\P% steal \P from pilcrow symbol
+\def\P#1{%
+  P\p@aux{#1}%
+}
+\def\p#1{%
+  p\p@aux{#1}%
+}
+\makeatother
+
 %% Notes and examples
 \newcommand{\noteintro}[1]{[\,\textit{#1:}\space}
 \newcommand{\noteoutro}[1]{\textit{\,---\,end #1}\,]}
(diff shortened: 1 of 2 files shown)


Eelis:
(Since implementing the functionality of that new mathtools LaTeX package is too much work for cxxdraft-htmlgen, this change will result in some divergence with the HTML version.)

godbyk:
@Eelis I'm afraid I'm unfamiliar with cxxdraft-htmlgen. Is there a mechanism to provide a simplified, equivalent form of `\p` and `\P` for cxxdraft-htmlgen to parse so it doesn't have to read the real definitions?

Eelis:
@godbyk Ah, it's the [tool](https://github.com/Eelis/cxxdraft-htmlgen) that generates [eel.is/c++draft](http://eel.is/c++draft), which is used a lot for citations in online discussions and documentation, because it's very link-friendly (unlike the PDF).

I could easily make cxxdraft-htmlgen ignore the new stuff in macros.tex, and I could define my own replacements for \p and \P that just do a dumb expansion, but I guess I'd end up with poor spacing/sizing.

tkoeppe:
Could you kindly provide some diff screenshots?

godbyk:
@tkoeppe I've rebased the branch on master and uploaded a [PDF showing the diffs](http://kevin.godby.org/temp/cplusplus-draft/std-diff-probability.pdf) in the numerics chapter (see pages 16–17 and 31–46).

You should only see slight adjustments in the horizontal space around the vertical bar.

(There are a number of other improvements that could be made to the mathematics typesetting in this chapter, but I'll leave those for separate issues/PRs.)

tkoeppe:
So... as much as I appreciate this effort, and as much as I enjoy proper operator spacing, I must say that a) the status quo is by no means bad here, b) the visual difference that this change makes is very minor, c) it requires a lot of new setup, and d) it hogs two extremely short macro names for something extremely niche that we never ever use again outside that one subclause.

At this point, I would like to ask if you would mind if we just drop this. If not, we can explore alternatives and modifications, but all told I think there are more interesting fires I'd like to put out at this point (e.g. indexing, and converting the LFTS to LaTeX).

godbyk:
@tkoeppe I don't mind. I just pulled code in from my own collection of macros, so there wasn't a ton of dev effort in this.

What's the LFTS? I'm happy to help with other LaTeX-related issues if you need a hand.

tkoeppe:
OK, let's close this then.

The LFTS is the Library Fundamentals TS: https://github.com/cplusplus/fundamentals-ts

It's currently in some obsolescent state of XML. My current plan is to divine an XSLT rule to transform it into our familiar LaTeX.

PR #2079, opened 2018-05-22
[conv.lval] Move a misplaced note

languagelawyer:
Not sure should the Note follow the Example or vice versa.

### DIFF
--- a/source/conversions.tex
+++ b/source/conversions.tex
@@ -141,6 +141,11 @@ is applied to an expression \tcode{e}, and either
       \tcode{ex}\iref{basic.def.odr},
 \end{itemize}
 the value contained in the referenced object is not accessed.
+\begin{note}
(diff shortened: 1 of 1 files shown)


cpplearner:
I think the note is in the right place. From [CWG 2140](http://wg21.link/CWG2140):

> The current rules in 4.1 [conv.lval] paragraph 2 do not require fetching the value in memory of an object of type `std::nullpt_t` in order to produce its prvalue. This choice has implications that may not have been considered for questions like whether use of a `std::nullptr_t` that is an inactive member of a union results in undefined behavior or whether a volatile `std::nullptr_t` variable in a discarded-value expression produces a side effect.

languagelawyer:
Ah, ok. Note really feels to belong to the previous paragraph.

Issue #2067, opened 2018-05-11
Adding index cross-references for definitions where index entry is for a different term


jwakely:
We have:

````
A \defnx{trivially copyable class}{class!trivially copyable} is a class:
````

This defines the term _trivially copyable class_ but adds an index subentry for "class, trivially copyable". That means "trivially copyable class" can't be found in the index!

We need to manually add that entry (see #2065 for that change):

````
\indextext{trivially copyable class|see{class, trivially copyable}}%
````

Is there any way to automate that through `\defnx` or some other macro, or at least make it easier? Maybe something like this (although it would need to replace `foo!bar` with `foo, bar` in `#2`):

````
\newcommand{\defnxx}[2]{\indexdefn{#2}\indextext{#1|see{#2}}\textit{#1}}
````

**Comment by godbyk** [comment]
If we redefine `\defnx` as follows:

```latex
\newcommand{\defnx}[2]{%
  \indexdefn{#2}%
  \ifthenelse{\NOT\equal{#1}{#2}}{%
    \indextext{#1|see{#2}}%
  }{}%
  \textit{#1}%
}
```

then `\defnx{trivially copyable class}{class!trivially copyable}` and `\defn{blah}` would both do what you want.

jensmaurer:
Issue #1033 is related; we did decide that the page numbers go with the decomposed index entry. (That doesn't mean we should not have "see" entries for the other form.)
It is not obvious that the "see" form is always non-decomposed. For example "class, trivially copyable" has the page numbers and "trivially copyable, class" has the "see". There might also be a "trivially copyable, type" with another "see" to "type, trivially copyable" that points to [basic.types] or so.
@jwakely, I believe such an approach would still make the index useful to find "trivially copyable class"; do you agree?

jwakely:
> @jwakely, I believe such an approach would still make the index useful to find "trivially copyable class"; do you agree?

Yes. As long as I can go to T in the index and find "trivially copyable class" in some form or another I think that's fine. Currently it's not there at all under T.

I tried using additional packages so I could replace the "|" with ", " in the index entry, but didn't get it working. Breaking it into two arguments seems like a better solution.

jensmaurer:
In #2080, there is a specific change to create a \defnc for compound defined terms with uniform indexing.

godbyk:
@jensmaurer Why are we avoiding `\ifthenelse`?

[...]

PR #2080, opened 2018-05-24
[macros,class] Define \defnc and use it experimentally

jensmaurer:
Partially addresses #2067.


jensmaurer:
This is the trial balloon; if we agree on this approach, a full-scale cleanup is required.

jensmaurer:
Rebased.

zygoloid:
What is the `c` short for? Maybe `\defnadj{foo}{bar}` (as in, this is a definition of the adjective "foo" when applied to "bar"s) would be a be better mnemonic device?

jensmaurer:
"c" stands for "see", because this form of \defn adds a "see" entry. We might use this form also for non-adjective cases, e.g. \defnc{horse}{doctor} (adds "horse doctor, see doctor, horse").

zygoloid:
Hmm, would it really be appropriate to use the same macro for both of those cases? I could imagine we might choose one day to italicize (or, say, hyperlink) only the "trivial" in "A \defnc{trivial}{class} is [...]", whereas we'd want to italicize both words in "A \defnc{horse}{doctor} is [...]". The two seem semantically different, so using different markup may be best.

In the ISO Terms and Definitions sense, the former case is

> **trivial**
> \<class\> trivially-copyable with a trivial default constructor

whereas the latter is

> **horse doctor**
> doctor specializing in horses

jensmaurer:
@zygoloid: Adjusted to \defnadj.

PR #2083, opened 2018-05-24
[output.iterators] Strike useless sentence from note

CaseyCarter:
There's no such thing as an "insert ~~iterator~~pointer," and the rest of this sentence isn't much better. Let's just strike it.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -481,10 +481,6 @@ They should be
 \term{single pass}
 algorithms.
 Equality and inequality might not be defined.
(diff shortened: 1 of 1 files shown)


timsong-cpp:
`insert_iterator` is a thing. "insert pointer", OTOH, not so much.

jwakely:
I think "the standard is not a tutorial" justifies striking this.

jensmaurer:
I'm ok with this change.

tkoeppe:
"Insert pointer to joke here."

PR #2085, opened 2018-05-24
Drop redundant 'expression'

### LINKED ISSUE #2069: Consider consistently omitting "expression" in "{l,r,x,pr,gl}value expression"
Most places already do this, and simply say e.g. "a prvalue of type ...", rather than "a prvalue expression of type ...". Since prvalues (etc) are expressions by definition, I suppose "prvalue expression" is a bit like saying "mammal animal". :)

Occurrences of `[prxlg]value expression`:

    expressions.tex:184:Whenever a glvalue expression appears as an operand of an operator that
    expressions.tex:196:a prvalue expression of type \tcode{int} is required.
    expressions.tex:205:Whenever a prvalue expression appears as an operand of an operator that
    expressions.tex:230:to modify an object through a nonmodifiable lvalue expression or through an rvalue expression
    expressions.tex:418:The glvalue expression is evaluated and its value is discarded.
    expressions.tex:2803:When \tcode{typeid} is applied to a glvalue expression whose type is a
    expressions.tex:2807:glvalue refers. If the glvalue expression is obtained by applying the
    expressions.tex:3242:A glvalue expression of type \tcode{T1},
    classes.tex:1048:keyword \tcode{this} is a prvalue expression whose value is the
    intro.tex:175:\defncontext{prvalue} static type of the prvalue expression
    conversions.tex:277:A prvalue expression of type $\tcode{T}_1$
    containers.tex:1770:  \requires\ If \tcode{t} is a non-const rvalue expression, \tcode{value_type} shall be
    containers.tex:1783:  \requires\ If \tcode{t} is a non-const rvalue expression, \tcode{value_type} shall be
    containers.tex:1795:  \requires\ If \tcode{t} is a non-const rvalue expression, \tcode{value_type} shall be
    containers.tex:2500:&   \requires\ If \tcode{t} is a non-const rvalue expression, \tcode{value_type} shall be
    containers.tex:2513:&   \requires\ If \tcode{t} is a non-const rvalue expression, \tcode{value_type} shall be
    containers.tex:2523:&   \requires\ If \tcode{t} is a non-const rvalue expression, \tcode{value_type} shall be
    numerics.tex:272:If \tcode{z} is an lvalue expression of type \cv{} \tcode{complex<T>} then:
    utilities.tex:16884: conversions applied when an lvalue expression is used as an rvalue, but also

jensmaurer:
when calling out a value category.

Fixes #2069.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -1051,7 +1051,7 @@ virtual\iref{class.virtual} or pure virtual\iref{class.abstract}.
 \pnum
 \indextext{this pointer@\tcode{this} pointer|see{\tcode{this}}}%
 In the body of a non-static\iref{class.mfct} member function, the
-keyword \tcode{this} is a prvalue expression whose value is the
+keyword \tcode{this} is a prvalue whose value is the
 address of the object for which the function is called.
 \indextext{\idxcode{this}!type of}%
 The type of \tcode{this} in a member function of a class \tcode{X} is
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -1776,7 +1776,7 @@ and \tcode{nh} denotes a non-const rvalue of type \tcode{X::node_type}.
 
 \tcode{a_uniq.\brk{}insert(\brk{}t)}      &
   \tcode{pair<\brk{}iterator, bool>}   &
-  \requires\ If \tcode{t} is a non-const rvalue expression, \tcode{value_type} shall be
+  \requires\ If \tcode{t} is a non-const rvalue, \tcode{value_type} shall be
   \oldconcept{MoveInsertable} into \tcode{X}; otherwise, \tcode{value_type} shall be
   \oldconcept{CopyInsertable} into \tcode{X}.\br
   \effects\ Inserts \tcode{t} if and only if there is no element in the container
@@ -1789,7 +1789,7 @@ and \tcode{nh} denotes a non-const rvalue of type \tcode{X::node_type}.
 
 \tcode{a_eq.\brk{}insert(\brk{}t)}        &
   \tcode{iterator}               &
-  \requires\ If \tcode{t} is a non-const rvalue expression, \tcode{value_type} shall be
+  \requires\ If \tcode{t} is a non-const rvalue, \tcode{value_type} shall be
   \oldconcept{MoveInsertable} into \tcode{X}; otherwise, \tcode{value_type} shall be
   \oldconcept{CopyInsertable} into \tcode{X}.\br
   \effects\ Inserts \tcode{t} and returns the iterator pointing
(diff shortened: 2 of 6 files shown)


jensmaurer:
I've left "Zvalue expression" in two places where it was talking about evaluating; talking about "evaluating a Zvalue" sounds strange (there's already a value, right?). Evaluating an expression sounds a lot better.

languagelawyer:
Wait, wait, wait... I think you also need to fix [value category definition](http://eel.is/c++draft/expr#def:value_category): "This property of an expression is called its value category."

If "Zvalue" is a _property_ of an expression, then then "Zvalue expression" makes sense: it is expression with property "Zvalue". It analogous to something like "tennis ball". You can't drop "ball" and just write "tennis".

jensmaurer:
@zygoloid, I believe my changes make sense; we talk about "prvalue" (without mentioning "expression") quite a lot.

jensmaurer:
Rebased.

languagelawyer:
@jensmaurer sorry, my bad, I was not clear. I'm not saying the changes are wrong. I'm just saying that Zvalue defined both as "is expression (with some particular properties)" **almost everywhere** and in at least one place it is defined as a property of an expression. BOTH "prvalue" and "prvalue expression" are **correct**, but the later form does not necessarily sound like "mammal animal", "prvalue" there acts as an adjective.

My point is: if the usage of "Zvalue" as an adjective is completely eliminated, I think it makes sense to rewrite this wording where "Zvalue" defined as a property of an expression and leave only as a subclass of expression.

jensmaurer:
@languagelawyer, agreed, but there are a few situations where we should probably keep "Xvalue expression". So we can't go all the way where you want to be.

zygoloid:
@languagelawyer To continue our tortured mammal/animal analogy, the analogue to [basic.lval]p2 would be:

> Every animal belongs to exactly one of the classes in this taxonomy: amphibian, arachnid, [...], mammal, [...] xenoturbidellan, or xiphosurans. This property of an animal is called its *class*.  [...]

I think it's reasonable to say that the value category is a property of an expression, despite also saying that each value category is a specific kind of expression. "xvalue" is not a property of an expression, but "value category" is -- we can't say "this expression is xvalue", but we can say "the value category of this expression is xvalue" (but there are better ways to say that), or "if these two expressions have the same value category, then [...]".

Keeping around a few utterances of "xvalue expression" doesn't seem terrible, especially if they aid comprehensibility, but it should be understood that that is parsed like "doctor horse" (that is, a doctor who is also a horse), not like "horse doctor" (with "horse" modifying "doctor").

languagelawyer:
> the analogue to [basic.lval]p2 would be

Your analogue is not bijective: you've used the same word "class" for both "fundamental classification" and "value category".

PR #2088, opened 2018-05-28
[intro.abstract] Move \indextext under \pnum

languagelawyer:
Currently index entries point to the end of the previous paragraph, see pic. image

### DIFF
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -477,8 +477,8 @@ value is not used and that no
 side effects affecting the
 observable behavior of the program are produced.}
 
-\indextext{behavior!implementation-defined}%
 \pnum
+\indextext{behavior!implementation-defined}%
 Certain aspects and operations of the abstract machine are described in this
 document as implementation-defined (for example,
 \tcode{sizeof(int)}). These constitute the parameters of the abstract machine.
@@ -489,8 +489,8 @@ See~\ref{intro.compliance}.} Such documentation shall define the instance of the
 abstract machine that corresponds to that implementation (referred to as the
 ``corresponding instance'' below).
 
-\indextext{behavior!unspecified}%
 \pnum
+\indextext{behavior!unspecified}%
 Certain other aspects and operations of the abstract machine are
 described in this document as unspecified (for example,
 order of evaluation of arguments in a function call\iref{expr.call}).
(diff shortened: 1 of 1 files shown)


BitonicEelis:
The issue probably only affects the HTML version generated by cxxdraft-htmlgen, so in fairness this should probably be considered a cxxdraft-htmlgen deficiency. Having said that, fixing it by changing the LaTeX means less work for cxxdraft-htmlgen, so I'm all for it. :P

languagelawyer:
> this should probably be considered a cxxdraft-htmlgen deficiency

Completely disagree. `\indextext` refers to the "current location". And until the next paragraph starts, the current location is the last defined paragraph.

languagelawyer:
> As far as LaTeX is concerned, the index entry is just a location you can jump to.

Yep. And the location where the index entry was defined — was the end of the paragraph. Not the beginning.

> With an index entry right before a pnum, the jump location coincides with the start of the paragraph

How index entry should coincide with the start of the paragraph which haven't even started yet?

> As far as I know this is ordinary LaTeX behavior.

Have you ever used "ordinary" LaTeX? For example, `article` documentclass? Where do you put `\label` if you want to refer to a section or to a figure caption? Before or after `\section`/`\caption`?

Anyway, just read https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#source-code-formatting

BitonicEelis:
I'm not sure why you think there is trouble in having a (zero-width) location coincide with the thing that follows it. In the string "ABC", the location between B and C coincides with the start of C. No?

As for \label, that's a different command. It doesn't tell us anything about index entries.

Eelis:
A closely related question that the style guidelines do not address, is whether \indextext enries for list items should be placed *after* the \item as well. For eel.is/c++draft I currently use a [patch](https://github.com/Eelis/draft/commit/c335f631a2dcaa7bae78a513bf16515fe179b093) that moves one such index entry that currently occurs *before* \item, for much the same reason as described in the description of this pull request.

tkoeppe:
I wouldn't mind such a style rule, though I'm not sure if anyone would spend significant effort on modifying existing uses.

jwakely:
> And until the next paragraph starts, the current location is the last defined paragraph.

The next paragraph starts after the blank line, not at the `\pnum`.

languagelawyer:
> I wouldn't mind such a style rule

I think this is not a style rule, because it affects which paragraph an index entry points to.
And the rule already exists https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#source-code-formatting

> The next paragraph starts after the blank line, not at the \pnum.

No. The blank line ends the current paragraph:
image
(D. Knuth "The TeXbook", Chapter 8: The Characters You Type)
and puts TeX back into vertical mode:
image
(D. Knuth "The TeXbook", Chapter 25:  Summary of Horizontal Mode)
The next paragraph starts only when some commands appear in vertical mode:
image
(D. Knuth "The TeXbook", Chapter 24: Summary of Vertical Mode)

I think the `\pnum` has some of these commands in its definition and therefore it starts the new paragraph.

tkoeppe:
>> I wouldn't mind such a style rule
>
> I think this is not a style rule, because it affects which paragraph an index entry points to.

Yes, you're right. I think what I meant was that I'd only have such a rule apply to when the index entry is in fact meant to appertain to the following paragraph. 

> And the rule already exists 

Wonderful!

PR #2148, opened 2018-06-10
P1042R1 __VA_OPT__ wording clarifications

### LINKED ISSUE #2116: [2018-06 CWG Motion 4] P1042R1 __VA_OPT__ wording clarifications

tkoeppe:
Fixes #2116.


source/preprocessor.tex
```diff
@@ -743,20 +743,31 @@
...
 \tcode{\#\#}
-preprocessing token or followed by a
+preprocessing token, or followed by a
 \tcode{\#\#}
-preprocessing token (see below),
+preprocessing token (see below) ---
 is replaced by the corresponding argument after all macros
-contained therein have been expanded.
-Before being substituted,
+contained therein have been expanded, in which case,
```
zygoloid: I don't like this "in which case"; I think it makes the wording more confusing, because it could be referring to the case where the "unless" part holds or the case where it doesn't hold. For example, the more natural construction "X does Y unless Z, in which case Q" would mean that X does Y if !Z and Q happens if Z. I also don't like the condition for this being hyphened into the middle of the sentence. How about this instead:

> For each parameter in the replacement list that is neither preceded by a `#` or `##` preprocessing token nor followed by a `##` preprocessing token, the preprocessing tokens naming the parameter are replaced by a token sequence determined as follows:
>  * If the parameter is of the form `__VA_OPT__(`*content*`)`, the replacement preprocessing tokens are the preprocessing token sequence for the corresponding argument.
>  * Otherwise, the replacement preprocessing tokens are the preprocessing tokens of corresponding argument after all macros contained therein have been expanded. The argument's preprocessing tokens are completely macro replaced before being substituted as if they formed the rest of the preprocessing file with no other preprocessing tokens being available.
>
> [ *Example*: ... ]

source/preprocessor.tex
```diff
@@ -767,8 +778,8 @@
 \pnum
 \indextext{__VA_OPT__@\mname{VA_OPT}}%
 The identifier \mname{VA_OPT}
-shall always occur as part of the token sequence
-\mname{VA_OPT}\tcode{(}\placeholder{content}\tcode{)},
+shall always occur as part of the preprocessing token sequence
+\mname{VA_OPT}\tcode{(}\placeholder{cont\-ent}\tcode{)},
```
zygoloid: We should describe `__VA_OPT__(`*content*`)` in the grammar and give it a production name like we do for *defined-macro-expression* and *has-include-expression*.

zygoloid: Let's leave this one for after the motions.

source/preprocessor.tex
tkoeppe: I like that. I've left the open sentence of the containing paragraph intact and then applied this change.

@hubert-reinterpretcast: any opinions?

source/preprocessor.tex
tkoeppe: Filed issue https://github.com/cplusplus/draft/issues/2185 to track this.

PR #2149, opened 2018-06-10
P0929R2 Checking for abstract class types

### LINKED ISSUE #2117: [2018-06 CWG Motion 5] P0929R2 Checking for abstract class types

jensmaurer:
Added cross-reference from the new use of 'abstract class
type' in [except.throw], pointing to [class.abstract].

Fixes #2117.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -262,8 +262,11 @@ A class name can also be implicitly declared by an
 
 \pnum
 \indextext{type!incomplete}%
-A program is ill-formed if the definition of any object gives the object
-an incomplete type\iref{basic.types}.
+In the definition of an object,
+the type of that object shall not be
+an incomplete type\iref{basic.types},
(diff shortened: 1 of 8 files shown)


zygoloid:
It looks like [temp.deduct]p11 bullet 2 also needs an update. Eg:

```
template<typename T> void f(T (*)[5]);
struct Abstract { virtual ~Abstract() = 0; };
void g() { f<Abstract>(nullptr); } // now ok, despite what bullet 2 says
```

zygoloid:
Also, from the wording paper: "*Editorial note: I suggest to move the statement about function definitions to 11.4 [dcl.fct.def].*" This seems like a good idea to me; please either do this or open an issue so we don't forget.

jensmaurer:
I've fixed the two issues you mentioned, @zygoloid.

PR #2151, opened 2018-06-10
P0542R5 Support for contract based programming in C++

### LINKED ISSUE #2126: [2018-06 CWG Motion 14] P0542R5 Support for contract based programming in C++
See also #2108.

jensmaurer:
Fixes #2126.


source/basic.tex
```diff
@@ -550,6 +550,15 @@
 requirement applies recursively)\footnote{\ref{dcl.fct.default}
 describes how default argument names are looked up.}; and
 
+\item if \tcode{D} is a function that
```
zygoloid: This "is a function" restriction is new, and appears to be incorrect for the "invokes a function with a precondition" part. Perhaps:

> if `D` invokes a function with a precondition, or is a function that contains an assertion or has a postcondition, [...]

source/basic.tex
```diff
@@ -550,6 +550,15 @@
...
+contains an assertion,
+has a postcondition,
+or invokes a function with a precondition,
+the conditions under which each definition of \tcode{D}
```
zygoloid: This "the conditions under which" phrasing is awkward. Presumably the intent is that an implementation can say that some combinations are allowed and some are not, so we can't just say:

> [...] it is implementation-defined whether all definitions of `D` shall be translated using the same build level and violation continuation mode.

?

source/declarations.tex
```diff
@@ -4239,3 +4240,414 @@
 could have the same address as \tcode{buckets}
 if their respective types are all empty.
 \end{example}
+
+\rSec2[dcl.attr.contract]{Contract attributes}%
```
zygoloid: We keep the subclauses of [dcl.attr] in alphabetical order. (So far that's been by attribute name, but the contract attributes throw a spanner in the works here.)

I'm not sure what the best answer is here.

source/declarations.tex
```diff
@@ -4239,3 +4240,414 @@
...
+\indextext{attribute!contracts|(}
+
+\pnum
+Contract attributes are used to specify
+preconditions, postconditions, and assertions for functions.
+
+\begin{bnf}
+\nontermdef{contract-attribute-specifier}\br
+  \terminal{[} \terminal{[} \terminal{expects} \opt{contract-level} \terminal{:} conditional-expression \terminal{]} \terminal{]}\br
+  \terminal{[} \terminal{[} \terminal{ensures} \opt{contract-level} \terminal{:} \opt{identifier} \terminal{:} conditional-expression \terminal{]} \terminal{]}\br
```
zygoloid: There's a stray colon here before the *identifier*.

source/declarations.tex
```diff
@@ -4239,3 +4240,414 @@
...
+}
+
+int (*pf)(int) = g;                             // OK
+int x = pf(5);                                  // contract conditions of \tcode{g} are checked
+\end{codeblock}
+\end{example}
+\end{note}
+
+\pnum
+The predicate of a contract is potentially evaluated\iref{basic.def.odr}.
```
zygoloid: Move this sentence to the end of paragraph 5. Also, we should convert it to a note, since this follows from the definition of "potentially evaluated".

source/declarations.tex
```diff
@@ -4239,3 +4240,414 @@
...
+  [[expects: *p == 0]]                          // \#2
+{
+  *p = 1;
+}
+\end{codeblock}
+\end{example}
+
+\pnum
+An assertion is checked by evaluating its predicate
+as part of the evaluation of the null statement\iref{stmt.expr}
```
zygoloid: We appear to be completely missing wording saying that an assertion may be applied to a null statement.

source/declarations.tex
```diff
@@ -4239,3 +4240,414 @@
...
+it applies to.
+
+\pnum
+The only side effects of a predicate
+that are allowed in a \grammarterm{contract-attribute-specifier}
+are modifications of non-volatile objects
+whose lifetime began and ended within the evaluation of the predicate
+in functions invoked from that predicate.
```
zygoloid: This "in functions invoked from that predicate" appears to be noise, perhaps left over from the earlier attempt at this rule in the wording paper. Delete it?

source/declarations.tex
```diff
@@ -4239,3 +4240,414 @@
...
+\end{codeblock}
+\end{example}
+
+\pnum
+An assertion is checked by evaluating its predicate
+as part of the evaluation of the null statement\iref{stmt.expr}
+it applies to.
+
+\pnum
+The only side effects of a predicate
```
zygoloid: This is more about predicates in the abstract; maybe this should go after p5, where we talk about contextual conversion to bool (and now, about being potentially-evaluated)?

source/declarations.tex
```diff
@@ -4239,3 +4240,414 @@
...
+The translation of a program consisting of translation units
+where the build level is not the same in all translation units
+is conditionally-supported.
+There should be no programmatic way of setting, modifying, or querying
+the build level of a translation unit.
+
+\pnum
+It is unspecified whether a contract
+that would not be checked under the current build level is evaluated.
+If it would evaluate to \tcode{false}, the behavior is undefined.
```
zygoloid: The "it" here is missing an antecedent. How about

> If a contract that is not checked under the current build level would evaluate to `false`, the behavior is undefined.

source/declarations.tex
```diff
@@ -4239,3 +4240,414 @@
...
+\impldef{mechanism for selecting the build level}.
+The translation of a program consisting of translation units
+where the build level is not the same in all translation units
+is conditionally-supported.
+There should be no programmatic way of setting, modifying, or querying
+the build level of a translation unit.
+
+\pnum
+It is unspecified whether a contract
+that would not be checked under the current build level is evaluated.
```
zygoloid: "would not be" -> "is not"

source/declarations.tex
```diff
@@ -4239,3 +4240,414 @@
...
+
+\pnum
+A translation may be performed
+with one of the following \defnx{build levels}{build level}:
+\term{off}, \term{default}, or \term{audit}.
+A translation with build level set to \term{default}
+performs checking for \tcode{default} contracts.
+A translation with build level set to \term{audit}
+performs checking for \tcode{default} and \tcode{audit} contracts.
+If no build level is explicitly selected, the build level is \term{default}.
```
zygoloid: We should explicitly say that the build level *off* means checking is not performed for any contracts, rather than leaving the reader wondering.

source/declarations.tex
```diff
@@ -4239,3 +4240,414 @@
...
+is conditionally-supported.
+There should be no programmatic way of setting, modifying, or querying
+the build level of a translation unit.
+
+\pnum
+It is unspecified whether a contract
+that would not be checked under the current build level is evaluated.
+If it would evaluate to \tcode{false}, the behavior is undefined.
+During constant expression evaluation\iref{expr.const},
+contract evaluation is performed only for those contracts that are checked.
```
zygoloid: This contradicts the first sentence of this paragraph. How about combining the two:

> It is unspecified whether a contract that is not checked under the current build level is evaluated outside of constant expression evaluation ([expr.const]); during constant evaluation ([expr.const]), only checked contracts are evaluated.

source/declarations.tex
```diff
@@ -4239,3 +4240,414 @@
...
+
+\pnum
+The \defn{violation handler} of a program is a function of type
+``\opt{\tcode{noexcept}} function of
+(lvalue reference to \tcode{const std::contract_violation})
+returning \tcode{void}'' specified in an
```
zygoloid: Add a comma before "specified", or add ", and is".

source/declarations.tex
```diff
@@ -4239,3 +4240,414 @@
...
+\impldef{specification of violation handler} manner.
+The violation handler is invoked
+when the predicate of a checked contract evaluates to \tcode{false}
+(called a \defn{contract violation}).
+There should be no programmatic way of
+setting or modifying the violation handler.
+It is \impldef{argument for violation handler}
+how the violation handler is established for a program and
+how the \tcode{std::contract_violation} argument value is set.
```
zygoloid: I think we need an "except as specified below" here, since we're about to give some rules that dictate how parts of the `contract_violation` argument value are set.

source/derived.tex
```diff
@@ -910,6 +910,45 @@
...
+  int x = 42;
+};
+
+int x = 42;
+struct B {
+  virtual void g() [[expects: x == 0]];
+}
+
+struct C : A, B {
+  virtual void g();             // ill-formed, no diagnostic required
```
zygoloid: Wait, why is this "no diagnostic required"? The overrider doesn't satisfy the conditions of the ODR, and won't always produce the same value. This is just ill-formed, isn't it?

source/templates.tex
```diff
@@ -6178,7 +6178,12 @@
 function\iref{special}, the program is ill-formed.
 
 \pnum
-The \grammarterm{declaration} in an \grammarterm{explicit-instantiation} and the \grammarterm{declaration} produced by the corresponding substitution into the templated function, variable, or class are two declarations of the same entity.
+The \grammarterm{declaration} in an \grammarterm{explicit-instantiation} and
+the \grammarterm{declaration} produced by the corresponding substitution
+into the templated function, variable, or class
+are two declarations of the same entity.
+An \grammarterm{explicit-instantiation} of a function template
+shall not specify a contract condition\iref{dcl.attr.contract}.
```
zygoloid: This is not an appropriate place for this rule (it's placed between a normative rule and its note and example). In fact, we often place such rules with the rules about the prohibited thing, not in the explicit instantiation rules. For example, [dcl.stc] disallows storage classes in explicit instantiations, and [except.spec] gives the rules for noexcept in explicit instantiations. But I would note that some other related rules are in [temp.explicit]p1: "An explicit instantiation of a function template, member function of a class template, or variable template shall not use the inline or constexpr specifiers." (And I'm not sure where we disallow an explicit instantiation from having default arguments...)

But this is also redundant. [dcl.attr.grammar]p5 says "No attribute-specifier-seq shall appertain to an explicit instantiation". So I think we should either drop this change or convert it to a note in p1.

source/exceptions.tex
```diff
@@ -1028,6 +1028,19 @@
 \tcode{std::atexit} or \tcode{std::at_quick_exit}
 exits via an exception\iref{support.start.term}, or
 
+\item%
+when evaluation of a predicate\iref{dcl.attr.contract}
```
zygoloid: "a predicate" -> "the predicate of a contract"

source/lib-intro.tex
```diff
@@ -1013,41 +1013,46 @@
 {llll}
 \topline
 \tcode{<algorithm>} &
+\tcode{<fstream>} &
```
zygoloid: This needs merging / rebasing.

source/basic.tex
jensmaurer: The "pure" model says preconditions are checked in the caller (so we need to account for that), but we have otherwise drafted wording that assumes (under as-if) that preconditions are actually evaluated in the callee. So, I think we should actually care for both places where precondition checking code could be emitted.

source/basic.tex
jensmaurer: Yes, it's my understanding that the implementation has to actually state the list of conditions, and the authors hope (in vain?) it's not "always", but something more modular and user-friendly.

jensmaurer: I've rephrased a bit more; it's now less awkward and still talks about "conditions".

source/declarations.tex
jensmaurer: "We keep the subclauses of [dcl.attr] in alphabetical order."
Ah, that wasn't clear to me. But yes, the current list looks alphabetical.

Sort according to "contract" maybe? We already sort "likely" and "unlikely" under "likelihood".
@zygoloid, awaiting your answer before actually doing the move.

source/declarations.tex
jensmaurer: Uh, right.
So, this is a "fixup" against the original paper transcription. I'd really like to move that commit as a "fixup" right after the original paper application, but how do I do that given that I'm not supposed to force-push?

source/declarations.tex
jensmaurer: Agreed.

source/declarations.tex
jensmaurer: Agreed; also added a cross-reference to [support.contract.cviol].

source/declarations.tex
jensmaurer: Yes, indeed. Added.

source/declarations.tex
jensmaurer: Yes, modifying temporaries within the predicate should be fine, too.

source/declarations.tex
jensmaurer: I disagree.
The previous paragraphs talk about syntactic restrictions, e.g. the funny access rules and where the attributes may be applied. When then define "checked" in terms of "evaluated". And then we discuss constraints on the specific evaluation, such as the paragraph in question and the next one, which talks about prohibited things in the evaluation of the function body (outside of contracts).

source/declarations.tex

source/declarations.tex
jensmaurer: I moved "outside..." to the start of the sentence to make it parallel with the stuff after the semicolon. Otherwise, looks good.

jensmaurer: Ah, one more problem: "The behavior is undefined if ... would evaluate to false" should only apply to the "outside of constant expressions" case; we don't ever want undefined behavior at compile-time. Rephrasing once more.

source/declarations.tex

source/derived.tex
jensmaurer: Right, fixed.

source/templates.tex
jensmaurer: Removed due to redundancy. I'm not adding a note; we should decide whether we want the "explicit instantiation" restriction in [temp.explicit] or where the other feature is defined, and consistently put notes in the "other" places. See #2188, for a later time.

source/lib-intro.tex
jensmaurer: Do you want me to rebase (with force-push)? That is my preference; i have no idea what "merge" would do to me.

source/exceptions.tex
jensmaurer: Fixed.

source/declarations.tex
zygoloid: Sure, let's sort by heading name.

source/lib-intro.tex
zygoloid: Yes, pushing with --force-with-lease seems fine here. (I think that will probably do bad things to github's comment tracking, though I suppose there's only one way to find out.)

source/basic.tex
zygoloid: I agree. However, this appears to be a normative change relative to the wording we were given, so can you send an email to the core reflector to make sure this won't surprise anyone?

source/declarations.tex
zygoloid: What you did is fine, I'll figure out which commit it's a fixup for when I do the merging. Or if you're going to force push anyway to rebase and pick up the new header names table, you could merge it into the original paper commit yourself.

source/declarations.tex
zygoloid: OK

zygoloid:
All looks good so far. Thanks!

source/declarations.tex

jensmaurer:
Ok, pushed using --force-with-lease.

PR #2171, opened 2018-06-12
P0732R2 Class types in non-type template parameters

### LINKED ISSUE #2118: [2018-06 CWG Motion 6] P0732R2 Class types in non-type template parameters

jensmaurer:
Fixes #2118.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -365,24 +365,21 @@ class Map {
 \end{note}
 
 \pnum
-A non-type
-\grammarterm{template-parameter}
-shall have one of the following (optionally
-cv-qualified)
-types:
+A non-type \grammarterm{template-parameter}
+shall have one of the following (optionally cv-qualified) types:
 
 \begin{itemize}
-\item integral or enumeration type,
+\item a type that is literal,
+has strong structural equality\iref{class.compare.default},
+has no \tcode{mutable} or \tcode{volatile} subobjects,
+and in which if there is a defaulted member \tcode{operator<=>},
+then it is declared public,
 
-\item pointer to object or pointer to function,
+\item an lvalue reference type,
 
-\item lvalue reference to object or lvalue reference to function,
+\item a type that contains a placeholder type\iref{dcl.spec.auto}, or
 
-\item pointer to member,
-
-\item \tcode{std::nullptr_t}, or
-
-\item a type that contains a placeholder type\iref{dcl.spec.auto}.
+\item a placeholder for a deduced class type\iref{dcl.type.class.deduct}.
 \end{itemize}
 
 \pnum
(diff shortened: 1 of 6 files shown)


source/templates.tex
```diff
@@ -1245,27 +1253,28 @@
...
-the deduced parameter type is determined
-from the type of the \grammarterm{template-argument}
-by placeholder type deduction\iref{dcl.type.auto.deduct}.
+If the type \tcode{T} of a \grammarterm{template-parameter}\iref{temp.param}
+contains a placeholder type\iref{dcl.spec.auto}
+or a placeholder for a deduced class type\iref{dcl.type.class.deduct},
+the type of the parameter is the type deduced
+for the variable \tcode{x} in the invented declaration:
+\begin{codeblock}
+  T x = @\grammarterm{template-argument}@ ;
```
zygoloid: Maybe try removing the italic corection after template-argument? There's more whitespace there than I'd prefer.

source/templates.tex
```diff
@@ -1245,27 +1253,28 @@
...
-If the type of a \grammarterm{template-parameter}
-contains a placeholder type~(\ref{dcl.spec.auto}, \ref{temp.param}),
-the deduced parameter type is determined
-from the type of the \grammarterm{template-argument}
-by placeholder type deduction\iref{dcl.type.auto.deduct}.
+If the type \tcode{T} of a \grammarterm{template-parameter}\iref{temp.param}
+contains a placeholder type\iref{dcl.spec.auto}
+or a placeholder for a deduced class type\iref{dcl.type.class.deduct},
+the type of the parameter is the type deduced
+for the variable \tcode{x} in the invented declaration:
```
zygoloid: Consider removing the colon here.

source/templates.tex
```diff
@@ -1328,6 +1337,10 @@
 
 const char p[] = "Vivisectionist";
 X<int,p> x2;                    // OK
+
+class A { constexpr A(const char*) {} };
```
zygoloid: This example is wrong. Add a defaulted `operator<=>` here to fix it.

source/templates.tex
jensmaurer: That's a hard one. Fiddling with the italics correction doesn't change anything (same before and after):
before

\kern to the rescue:
kern

source/templates.tex
jensmaurer: The example is even more wrong: The type of the second template-parameter of X is still "const char *" and does not involve "A" at all.

source/templates.tex
```diff
@@ -1338,9 +1338,16 @@
...
+  constexpr A(const char*) {}
+  auto operator<=>(A, A) = default;
+};
+
+template<A value> class Y {
+  @\commentellip@
+};
 
-X<A, "Pyrophoricity"> x3;       // OK, string literal is a constructor argument to A
+Y<"Pyrophoricity"> x3;       // OK, string literal is a constructor argument to A
```
zygoloid: What's the rationale for the change from `X` to `Y` here?

zygoloid: Never mind, I see it now. Looks like a change in P0732R2 was missing markup: I think we were supposed to change the rest of the example to:
```
template<class T, T p> class X {
/* ... */
};
X<const char*, "Studebaker"> x1; // error: string literal as template-argument
const char p[] = "Vivisectionist";
X<const char*,p> x2; // OK
```

source/templates.tex
zygoloid: `\itcorr[-1]` does help:

screenshot from 2018-06-18 17-00-27

That's:

```
  T x = @\grammarterm{template-argument}@ ;
  T x = @\grammarterm{template-argument}\itcorr[-1]@ ;
  T x = @\grammarterm{template-argument}\kern-0.3em@ ;
```

I'm going with the `\itcorr` approach for consistency with what we do elsewhere. I think it also makes the left/right whitespace look more balanced than the 0.3em negative kerning hint.

godbyk: @zygoloid Does the `\itcorr[-1]` give you the same result as using `\grammartermnc`?

zygoloid: @godbyk It didn't, but that's because `\grammartermnc` was broken. Fixed in 1478e7c3.

PR #2174, opened 2018-06-12
[macros,headers] Introduce 'multicoltable'

jensmaurer:
and use it for the tables of C++ headers.


zygoloid:
`multicols` forces all columns to the same width, which isn't what we want here (we want the natural text width of the column instead). fbd725cb93af1ec52753926253f9189ef33230ff addresses this in a different way (though it's inferior to this approach in that it needs manual column breaks to be specified).

PR #2176, opened 2018-06-13
P0898R3 Standard Library Concepts

### LINKED ISSUE #2145: [2018-06 LWG Motion 28] P0898R3 Standard library concepts
[P0898R3](http://wiki.edg.com/pub/Wg21rapperswil2018/StrawPolls/p0898r3.pdf)

CaseyCarter:
Fixes #2145.

Pushing from a branch in a remote repo rather than `motions-2018-06-lwg-28` as directed in the [new application procedure](https://github.com/cplusplus/draft/wiki/Applying-papers-to-the-Working-Draft) since I don't have commit access and cannot make new branches.

Reviewers: Please be brutal. Tiny errors I've made herein will be magnified by one to two orders of magnitude when the One Ranges Proposal is merged - I need the practice.

I'd appreciate if someone could assign this to the "post-2018-06" milestone as well.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -3108,7 +3108,7 @@ shall not overlap. The expression \tcode{*result = *first} shall be valid.
 \begin{note}
 For the overloads with an \tcode{ExecutionPolicy}, there may be a performance
 cost if \tcode{iterator_traits<ForwardIterator1>::value_type} is not
-\oldconcept{Move\-Constructible} (\tref{moveconstructible}).
+\oldconcept{\-Move\-Constructible} (\tref{moveconstructible}).
 \end{note}
 
 \pnum
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -1146,39 +1146,39 @@ If an algorithm's template parameter is named
 or
 \tcode{Input\-Iterator2},
 the template argument shall satisfy the
-requirements of an input iterator\iref{input.iterators}.
+\oldconcept{InputIterator} requirements\iref{input.iterators}.
 \item
 If an algorithm's template parameter is named
 \tcode{OutputIterator},
 \tcode{OutputIterator1},
 or
 \tcode{Output\-Iterator2},
-the template argument shall satisfy the requirements
-of an output iterator\iref{output.iterators}.
+the template argument shall satisfy the
+\oldconcept{OutputIterator} requirements\iref{output.iterators}.
 \item
 If an algorithm's template parameter is named
 \tcode{ForwardIterator},
 \tcode{ForwardIterator1},
 or
 \tcode{Forward\-Iterator2},
-the template argument shall satisfy the requirements
-of a forward iterator\iref{forward.iterators}.
+the template argument shall satisfy the
+\oldconcept{ForwardIterator} requirements\iref{forward.iterators}.
 \item
 If an algorithm's template parameter is named
 \tcode{BidirectionalIterator},
 \tcode{Bidirectional\-Iterator1},
 or
 \tcode{BidirectionalIterator2},
-the template argument shall satisfy the requirements
-of a bidirectional iterator\iref{bidirectional.iterators}.
+the template argument shall satisfy the
+\oldconcept{BidirectionalIterator} requirements\iref{bidirectional.iterators}.
 \item
 If an algorithm's template parameter is named
 \tcode{RandomAccessIterator},
 \tcode{Random\-AccessIterator1},
 or
 \tcode{RandomAccessIterator2},
-the template argument shall satisfy the requirements
-of a random-access iterator\iref{random.access.iterators}.
+the template argument shall satisfy the
+\oldconcept{RandomAccessIterator} requirements\iref{random.access.iterators}.
 \end{itemize}
 
 \pnum
@@ -1648,9 +1648,9 @@ template<class InputIterator, class Function>
 \begin{itemdescr}
 \pnum
 \requires \tcode{Function} shall satisfy the
-\tcode{MoveConstructible} requirements (\tref{moveconstructible}).
+\oldconcept{MoveConstructible} requirements (\tref{moveconstructible}).
 \begin{note} \tcode{Function} need not meet the requirements of
-\tcode{CopyConstructible} (\tref{copyconstructible}). \end{note}
+\oldconcept{CopyConstructible} (\tref{copyconstructible}). \end{note}
 
 \pnum
 \effects
@@ -1692,7 +1692,7 @@ template<class ExecutionPolicy, class ForwardIterator, class Function>
 \begin{itemdescr}
 \pnum
 \requires
-\tcode{Function} shall satisfy the \tcode{CopyConstructible} requirements.
+\tcode{Function} shall satisfy the \oldconcept{CopyConstructible} requirements.
 
 \pnum
 \effects
@@ -1730,9 +1730,9 @@ template<class InputIterator, class Size, class Function>
 \begin{itemdescr}
 \pnum
 \requires
-\tcode{Function} shall satisfy the \tcode{MoveConstructible} requirements
+\tcode{Function} shall satisfy the \oldconcept{MoveConstructible} requirements
 \begin{note} \tcode{Function} need not meet the requirements of
-\tcode{CopyConstructible}. \end{note}
+\oldconcept{CopyConstructible}. \end{note}
 
 \pnum
 \requires
@@ -1766,7 +1766,7 @@ template<class ExecutionPolicy, class ForwardIterator, class Size, class Functio
 \begin{itemdescr}
 \pnum
 \requires
-\tcode{Function} shall satisfy the \tcode{CopyConstructible} requirements.
+\tcode{Function} shall satisfy the \oldconcept{CopyConstructible} requirements.
 
 \pnum
 \requires
@@ -2190,7 +2190,7 @@ if
 \tcode{InputIterator1}
 and
 \tcode{InputIterator2}
-meet the requirements of random access iterators\iref{random.access.iterators}
+meet the \oldconcept{RandomAccessIterator} requirements\iref{random.access.iterators}
 and
 \tcode{last1 - first1 != last2 - first2},
 then
@@ -2210,7 +2210,7 @@ if
 \tcode{ForwardIterator1}
 and
 \tcode{ForwardIterator2}
-meet the requirements of random access iterators and
+meet the \oldconcept{RandomAccessIterator} requirements and
 \tcode{last1 - first1 != last2 - first2}, then
 no applications of the corresponding predicate; otherwise,
 
@@ -2400,7 +2400,7 @@ Equivalent to: \tcode{return searcher(first, last).first;}
 
 \pnum
 \remarks
-\tcode{Searcher} need not meet the \tcode{CopyConstructible} requirements.
+\tcode{Searcher} need not meet the \oldconcept{CopyConstructible} requirements.
 \end{itemdescr}
 
 
@@ -2496,7 +2496,7 @@ template<class ExecutionPolicy, class ForwardIterator1, class ForwardIterator2,
 \begin{note}
 For the overload with an \tcode{ExecutionPolicy}, there may be a performance
 cost if \tcode{iterator_traits<ForwardIterator1>::value_type} is not
-\tcode{Move\-Constructible} (\tref{moveconstructible}).
+\oldconcept{\-Move\-Constructible} (\tref{moveconstructible}).
 \end{note}
 
 \pnum
@@ -3038,7 +3038,7 @@ template<class ExecutionPolicy, class ForwardIterator, class Predicate>
 \requires
 The type of
 \tcode{*first}
-shall satisfy the \tcode{MoveAssignable}
+shall satisfy the \oldconcept{MoveAssignable}
 requirements (\tref{moveassignable}).
 
 \pnum
@@ -3108,7 +3108,7 @@ shall not overlap. The expression \tcode{*result = *first} shall be valid.
 \begin{note}
 For the overloads with an \tcode{ExecutionPolicy}, there may be a performance
 cost if \tcode{iterator_traits<ForwardIterator1>::value_type} is not
-\tcode{Move\-Constructible} (\tref{moveconstructible}).
+\oldconcept{Move\-Constructible} (\tref{moveconstructible}).
 \end{note}
 
 \pnum
@@ -3158,7 +3158,7 @@ template<class ExecutionPolicy, class ForwardIterator, class BinaryPredicate>
 \requires
 The comparison function shall be an equivalence relation.
 The type of \tcode{*first} shall satisfy the
-\tcode{MoveAssignable} requirements (\tref{moveassignable}).
+\oldconcept{MoveAssignable} requirements (\tref{moveassignable}).
 
 \pnum
 \effects
@@ -3231,17 +3231,17 @@ shall be valid.
 \item
 For the overloads with no \tcode{ExecutionPolicy},
 let \tcode{T} be the value type of \tcode{InputIterator}.
-If \tcode{InputIterator} meets the forward iterator requirements,
+If \tcode{InputIterator} meets the \oldconcept{ForwardIterator} requirements,
 then there are no additional requirements for \tcode{T}.
-Otherwise, if \tcode{OutputIterator} meets the forward iterator
+Otherwise, if \tcode{OutputIterator} meets the \oldconcept{ForwardIterator}
 requirements and its value type is the same as \tcode{T},
-then \tcode{T} shall be \tcode{CopyAssignable} (\tref{copyassignable}).
+then \tcode{T} shall be \oldconcept{CopyAssignable} (\tref{copyassignable}).
 Otherwise, \tcode{T} shall be both
-\tcode{CopyConstructible} (\tref{copyconstructible}) and \tcode{CopyAssignable}.
+\oldconcept{CopyConstructible} (\tref{copyconstructible}) and \oldconcept{CopyAssignable}.
 \begin{note}
 For the overloads with an \tcode{ExecutionPolicy}, there may be a performance
 cost if the value type of \tcode{ForwardIterator1} is not both
-\tcode{CopyConstructible} and \tcode{CopyAssignable}.
+\oldconcept{CopyConstructible} and \oldconcept{CopyAssignable}.
 \end{note}
 \end{itemize}
 
@@ -3283,7 +3283,7 @@ template<class ExecutionPolicy, class BidirectionalIterator>
 \pnum
 \requires
 \tcode{BidirectionalIterator} shall satisfy the
-\tcode{ValueSwappable} requirements\iref{swappable.requirements}.
+\oldconcept{ValueSwappable} requirements\iref{swappable.requirements}.
 
 \pnum
 \effects
@@ -3367,9 +3367,9 @@ and
 \range{middle}{last}
 shall be valid ranges.
 \tcode{ForwardIterator} shall satisfy the
-\tcode{ValueSwappable} requirements\iref{swappable.requirements}. The type of \tcode{*first} shall satisfy
-the \tcode{MoveConstructible} (\tref{moveconstructible}) and
-\tcode{MoveAssignable} (\tref{moveassignable}) requirements.
+\oldconcept{ValueSwappable} requirements\iref{swappable.requirements}. The type of \tcode{*first} shall satisfy
+the \oldconcept{MoveConstructible} (\tref{moveconstructible}) and
+\oldconcept{MoveAssignable} (\tref{moveassignable}) requirements.
 
 \pnum
 \effects
@@ -3455,12 +3455,12 @@ template<class PopulationIterator, class SampleIterator,
 \requires
 \begin{itemize}
 \item
-\tcode{PopulationIterator} shall satisfy the requirements of an input iterator\iref{input.iterators}.
+\tcode{PopulationIterator} shall satisfy the \oldconcept{InputIterator} requirements\iref{input.iterators}.
 \item
-\tcode{SampleIterator} shall satisfy the requirements of an output iterator\iref{output.iterators}.
+\tcode{SampleIterator} shall satisfy the \oldconcept{OutputIterator} requirements\iref{output.iterators}.
 \item
-\tcode{SampleIterator} shall satisfy the additional requirements of a random access iterator\iref{random.access.iterators}
-unless \tcode{PopulationIterator} satisfies the additional requirements of a forward iterator\iref{forward.iterators}.
+\tcode{SampleIterator} shall satisfy the \oldconcept{RandomAccessIterator} requirements\iref{random.access.iterators}
+unless \tcode{Pop\-ulat\-ion\-Iter\-ator} satisfies the \oldconcept{ForwardIterator} requirements\iref{forward.iterators}.
 \item
 \tcode{PopulationIterator}'s value type shall be writable\iref{iterator.requirements.general} to \tcode{out}.
 \item
@@ -3496,7 +3496,7 @@ The end of the resulting sample range.
 \begin{itemize}
 \item
 Stable if and only if \tcode{PopulationIterator} satisfies the
-requirements of a forward iterator.
+\oldconcept{ForwardIterator} requirements.
 \item
 To the extent that the implementation of this function makes use of
 random numbers, the object \tcode{g} shall serve as the
@@ -3518,7 +3518,7 @@ template<class RandomAccessIterator, class UniformRandomBitGenerator>
 \pnum
 \requires
 \tcode{RandomAccessIterator} shall satisfy the
-\tcode{ValueSwappable} requirements\iref{swappable.requirements}.
+\oldconcept{ValueSwappable} requirements\iref{swappable.requirements}.
 The type
 \tcode{remove_reference_t<UniformRandomBitGenerator>}
 shall satisfy the requirements of a
@@ -3778,10 +3778,10 @@ template<class ExecutionPolicy, class RandomAccessIterator, class Compare>
 \pnum
 \requires
 \tcode{RandomAccessIterator} shall satisfy the
-\tcode{ValueSwappable} requirements\iref{swappable.requirements}. The type
+\oldconcept{ValueSwappable} requirements\iref{swappable.requirements}. The type
 of \tcode{*first} shall satisfy the
-\tcode{MoveConstructible} (\tref{moveconstructible}) and
-\tcode{MoveAssignable} (\tref{moveassignable}) requirements.
+\oldconcept{MoveConstructible} (\tref{moveconstructible}) and
+\oldconcept{MoveAssignable} (\tref{moveassignable}) requirements.
 
 \pnum
 \effects
@@ -3816,10 +3816,10 @@ template<class ExecutionPolicy, class RandomAccessIterator, class Compare>
 \pnum
 \requires
 \tcode{RandomAccessIterator} shall satisfy the
-\tcode{ValueSwappable} requirements\iref{swappable.requirements}. The type
+\oldconcept{ValueSwappable} requirements\iref{swappable.requirements}. The type
 of \tcode{*first} shall satisfy the
-\tcode{MoveConstructible} (\tref{moveconstructible}) and
-\tcode{MoveAssignable} (\tref{moveassignable}) requirements.
+\oldconcept{MoveConstructible} (\tref{moveconstructible}) and
+\oldconcept{MoveAssignable} (\tref{moveassignable}) requirements.
 
 \pnum
 \effects
@@ -3867,10 +3867,10 @@ template<class ExecutionPolicy, class RandomAccessIterator, class Compare>
 \pnum
 \requires
 \tcode{RandomAccessIterator} shall satisfy the
-\tcode{ValueSwappable} requirements\iref{swappable.requirements}. The type
+\oldconcept{ValueSwappable} requirements\iref{swappable.requirements}. The type
 of \tcode{*first} shall satisfy the
-\tcode{MoveConstructible} (\tref{moveconstructible}) and
-\tcode{MoveAssignable} (\tref{moveassignable}) requirements.
+\oldconcept{MoveConstructible} (\tref{moveconstructible}) and
+\oldconcept{MoveAssignable} (\tref{moveassignable}) requirements.
 
 \pnum
 \effects
@@ -3929,10 +3929,10 @@ template<class ExecutionPolicy, class ForwardIterator, class RandomAccessIterato
 \pnum
 \requires
 \tcode{RandomAccessIterator} shall satisfy the
-\tcode{ValueSwappable} requirements\iref{swappable.requirements}. The type
+\oldconcept{ValueSwappable} requirements\iref{swappable.requirements}. The type
 of \tcode{*result_first} shall satisfy the
-\tcode{MoveConstructible} (\tref{moveconstructible}) and
-\tcode{Move\-Assignable} (\tref{moveassignable}) requirements.
+\oldconcept{MoveConstructible} (\tref{moveconstructible}) and
+\oldconcept{\-Move\-Assignable} (\tref{moveassignable}) requirements.
 
 \pnum
 \effects
@@ -4068,10 +4068,10 @@ template<class ExecutionPolicy, class RandomAccessIterator, class Compare>
 \pnum
 \requires
 \tcode{RandomAccessIterator} shall satisfy the
-\tcode{ValueSwappable} requirements\iref{swappable.requirements}. The type
+\oldconcept{ValueSwappable} requirements\iref{swappable.requirements}. The type
 of \tcode{*first} shall satisfy the
-\tcode{MoveConstructible} (\tref{moveconstructible}) and
-\tcode{MoveAssignable} (\tref{moveassignable}) requirements.
+\oldconcept{MoveConstructible} (\tref{moveconstructible}) and
+\oldconcept{MoveAssignable} (\tref{moveassignable}) requirements.
 
 \pnum
 \effects
@@ -4380,7 +4380,7 @@ template<class ExecutionPolicy, class ForwardIterator, class Predicate>
 \pnum
 \requires
 \tcode{ForwardIterator} shall satisfy the
-\tcode{ValueSwappable} requirements\iref{swappable.requirements}.
+\oldconcept{ValueSwappable} requirements\iref{swappable.requirements}.
 
 \pnum
 \effects Places all the elements in the range \range{first}{last} that satisfy \tcode{pred} before all the elements that do not satisfy it.
@@ -4395,7 +4395,7 @@ range \range{i}{last}, \tcode{pred(*k) == false}.
 \begin{itemize}
 \item For the overload with no \tcode{ExecutionPolicy}, exactly $N$ applications
 of the predicate.  At most $N / 2$ swaps if \tcode{ForwardIterator} meets the
-\tcode{BidirectionalIterator} requirements and at most $N$ swaps otherwise.
+\oldconcept{BidirectionalIterator} requirements and at most $N$ swaps otherwise.
 \item For the overload with an \tcode{ExecutionPolicy},
 \bigoh{N \log N} swaps and \bigoh{N} applications of the predicate.
 \end{itemize}
@@ -4417,10 +4417,10 @@ template<class ExecutionPolicy, class BidirectionalIterator, class Predicate>
 \pnum
 \requires
 \tcode{BidirectionalIterator} shall satisfy the
-\tcode{ValueSwappable} requirements\iref{swappable.requirements}. The type
+\oldconcept{ValueSwappable} requirements\iref{swappable.requirements}. The type
 of \tcode{*first} shall satisfy the
-\tcode{MoveConstructible} (\tref{moveconstructible}) and
-\tcode{MoveAssignable} (\tref{moveassignable}) requirements.
+\oldconcept{MoveConstructible} (\tref{moveconstructible}) and
+\oldconcept{MoveAssignable} (\tref{moveassignable}) requirements.
 
 \pnum
 \effects
@@ -4479,19 +4479,19 @@ template<class ExecutionPolicy, class ForwardIterator, class ForwardIterator1,
 \begin{itemize}
 \item
 For the overload with no \tcode{ExecutionPolicy}, \tcode{InputIterator}'s
-value type shall be \tcode{CopyAssignable} (\tref{copyassignable}),
+value type shall be \oldconcept{\-Copy\-Assign\-able} (\tref{copyassignable}),
 and shall be writable\iref{iterator.requirements.general} to the \tcode{out_true}
-and \tcode{out_false} \tcode{OutputIterator}s, and shall be convertible to
+and \tcode{out_false} \tcode{Output\-Iter\-ator}s, and shall be convertible to
 \tcode{Predicate}'s argument type.
 
 \item
 For the overload with an \tcode{ExecutionPolicy}, \tcode{ForwardIterator}'s
-value type shall be \tcode{Copy\-Assign\-able}, and shall be writable to the
+value type shall be \oldconcept{\-Copy\-Assign\-able}, and shall be writable to the
 \tcode{out_true} and \tcode{out_false} \tcode{ForwardIterator}s, and shall be
 convertible to \tcode{Predicate}'s argument type.
 \begin{note}
 There may be a performance cost if \tcode{ForwardIterator}'s value type is not
-\tcode{CopyConstructible}.
+\oldconcept{CopyConstructible}.
 \end{note}
 
 \item
@@ -4618,10 +4618,10 @@ template<class ExecutionPolicy, class BidirectionalIterator, class Compare>
 The ranges \range{first}{middle} and \range{middle}{last} shall be
 sorted with respect to \tcode{operator<} or \tcode{comp}.
 \tcode{BidirectionalIterator} shall satisfy the
-\tcode{ValueSwappable} requirements\iref{swappable.requirements}. The type
+\oldconcept{ValueSwappable} requirements\iref{swappable.requirements}. The type
 of \tcode{*first} shall satisfy the
-\tcode{MoveConstructible} (\tref{moveconstructible}) and
-\tcode{MoveAssignable} (\tref{moveassignable}) requirements.
+\oldconcept{MoveConstructible} (\tref{moveconstructible}) and
+\oldconcept{MoveAssignable} (\tref{moveassignable}) requirements.
 
 \pnum
 \effects
@@ -5032,9 +5032,9 @@ The range
 \range{first}{last - 1}
 shall be a valid heap.
 The type of \tcode{*first} shall satisfy
-the \tcode{MoveConstructible} requirements
+the \oldconcept{MoveConstructible} requirements
 (\tref{moveconstructible}) and the
-\tcode{MoveAssignable} requirements
+\oldconcept{MoveAssignable} requirements
 (\tref{moveassignable}).
 
 \pnum
@@ -5070,10 +5070,10 @@ The range
 \range{first}{last}
 shall be a valid non-empty heap.
 \tcode{RandomAccessIterator} shall satisfy the
-\tcode{ValueSwappable} requirements\iref{swappable.requirements}. The type
+\oldconcept{ValueSwappable} requirements\iref{swappable.requirements}. The type
 of \tcode{*first} shall satisfy the
-\tcode{MoveConstructible} (\tref{moveconstructible}) and
-\tcode{MoveAssignable} (\tref{moveassignable}) requirements.
+\oldconcept{MoveConstructible} (\tref{moveconstructible}) and
+\oldconcept{MoveAssignable} (\tref{moveassignable}) requirements.
 
 
 \pnum
@@ -5107,9 +5107,9 @@ template<class RandomAccessIterator, class Compare>
 \begin{itemdescr}
 \pnum
 \requires The type of \tcode{*first} shall satisfy
-the \tcode{MoveConstructible} requirements
+the \oldconcept{MoveConstructible} requirements
 (\tref{moveconstructible}) and the
-\tcode{MoveAssignable} requirements
+\oldconcept{MoveAssignable} requirements
 (\tref{moveassignable}).
 
 \pnum
@@ -5140,10 +5140,10 @@ template<class RandomAccessIterator, class Compare>
 \pnum
 \requires The range \range{first}{last} shall be a valid heap.
 \tcode{RandomAccessIterator} shall satisfy the
-\tcode{ValueSwappable} requirements\iref{swappable.requirements}. The type
+\oldconcept{ValueSwappable} requirements\iref{swappable.requirements}. The type
 of \tcode{*first} shall satisfy the
-\tcode{MoveConstructible} (\tref{moveconstructible}) and
-\tcode{MoveAssignable} (\tref{moveassignable}) requirements.
+\oldconcept{MoveConst\-ruct\-ible} (\tref{moveconstructible}) and
+\oldconcept{MoveAssignable} (\tref{moveassignable}) requirements.
 
 \pnum
 \effects
@@ -5261,7 +5261,7 @@ template<class T, class Compare>
 \pnum
 \requires
 For the first form, type \tcode{T} shall be
-\tcode{LessThanComparable} (\tref{lessthancomparable}).
+\oldconcept{LessThanComparable} (\tref{lessthancomparable}).
 
 \pnum
 \returns
@@ -5286,8 +5286,8 @@ template<class T, class Compare>
 
 \begin{itemdescr}
 \pnum
-\requires \tcode{T} shall be \tcode{CopyConstructible} and \tcode{t.size() > 0}.
-For the first form, type \tcode{T} shall be \tcode{LessThanComparable}.
+\requires \tcode{T} shall be \oldconcept{CopyConstructible} and \tcode{t.size() > 0}.
+For the first form, type \tcode{T} shall be \oldconcept{LessThanComparable}.
 
 \pnum
 \returns The smallest value in the initializer list.
@@ -5311,7 +5311,7 @@ template<class T, class Compare>
 \pnum
 \requires
 For the first form, type \tcode{T} shall be
-\tcode{LessThanComparable} (\tref{lessthancomparable}).
+\oldconcept{LessThanComparable} (\tref{lessthancomparable}).
 
 \pnum
 \returns
@@ -5336,8 +5336,8 @@ template<class T, class Compare>
 
 \begin{itemdescr}
 \pnum
-\requires \tcode{T} shall be \tcode{CopyConstructible} and \tcode{t.size() > 0}.
-For the first form, type \tcode{T} shall be \tcode{LessThanComparable}.
+\requires \tcode{T} shall be \oldconcept{CopyConstructible} and \tcode{t.size() > 0}.
+For the first form, type \tcode{T} shall be \oldconcept{LessThanComparable}.
 
 \pnum
 \returns The largest value in the initializer list.
@@ -5362,7 +5362,7 @@ template<class T, class Compare>
 \pnum
 \requires
 For the first form, type \tcode{T} shall be
-\tcode{LessThanComparable} (\tref{lessthancomparable}).
+\oldconcept{LessThanComparable} (\tref{lessthancomparable}).
 
 \pnum
 \returns
@@ -5389,8 +5389,8 @@ template<class T, class Compare>
 
 \begin{itemdescr}
 \pnum
-\requires \tcode{T} shall be \tcode{CopyConstructible} and \tcode{t.size() > 0}.
-For the first form, type \tcode{T} shall be \tcode{LessThanComparable}.
+\requires \tcode{T} shall be \oldconcept{CopyConstructible} and \tcode{t.size() > 0}.
+For the first form, type \tcode{T} shall be \oldconcept{LessThanComparable}.
 
 \pnum
 \returns \tcode{pair<T, T>(x, y)}, where \tcode{x} has the smallest and \tcode{y} has the
@@ -5546,7 +5546,7 @@ template<class T, class Compare>
 \requires
 The value of \tcode{lo} shall be no greater than \tcode{hi}.
 For the first form, type \tcode{T}
-shall be \tcode{LessThan\-Comparable} (\tref{lessthancomparable}).
+shall be \oldconcept{LessThan\-Comparable} (\tref{lessthancomparable}).
 
 \pnum
 \returns
@@ -5742,7 +5742,7 @@ template<class BidirectionalIterator, class Compare>
 \pnum
 \requires
 \tcode{BidirectionalIterator} shall satisfy the
-\tcode{ValueSwappable} requirements\iref{swappable.requirements}.
+\oldconcept{ValueSwappable} requirements\iref{swappable.requirements}.
 
 \pnum
 \effects
@@ -5784,7 +5784,7 @@ template<class BidirectionalIterator, class Compare>
 \pnum
 \requires
 \tcode{BidirectionalIterator} shall satisfy the
-\tcode{ValueSwappable} requirements\iref{swappable.requirements}.
+\oldconcept{ValueSwappable} requirements\iref{swappable.requirements}.
 
 \pnum
 \effects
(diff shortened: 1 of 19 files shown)


jensmaurer:
Hi @CaseyCarter , thanks for the pull request.

First of all, we really want to get rid of \textit in the standard text, because it does not convey any semantics. If we ever want to change the formatting of, say, C++98 concepts, \textit is not enough.
So, please add a new macro \oldconcept or something like that to macros.tex, defined to \textit, and use \oldconcept{blah} throughout the text proper. (If you have a better name for the macro, I'm all ears.)

source/algorithms.tex
```diff
@@ -1126,39 +1126,39 @@
 or
 \tcode{Input\-Iterator2},
 the template argument shall satisfy the
-requirements of an input iterator\iref{input.iterators}.
+\textit{Cpp98InputIterator} requirements\iref{input.iterators}.
```
jensmaurer: Could you point to the location in the incoming paper where authorization is given to make this change? The paper says "prepend the prefix 'Cpp98' to uses of the named requirement sets below", but "input iterator" is not a CamelCase fixed-width requirement, it's a prose plain-text requirement, and does not appear in the list in subclause 1.3 of the paper.

source/algorithms.tex
```diff
@@ -1126,39 +1126,39 @@
...
 \item
 If an algorithm's template parameter is named
 \tcode{ForwardIterator},
 \tcode{ForwardIterator1},
 or
 \tcode{Forward\-Iterator2},
-the template argument shall satisfy the requirements
-of a forward iterator\iref{forward.iterators}.
+the template argument shall satisfy the
+\textit{Cpp98ForwardIterator} requirements\iref{forward.iterators}.
```
jensmaurer: The issue here is that a reference to "forward iterator" (plain text) refers to the entire section [forward.iterators], including (for example) the multi-pass guarantee. In contrast, _Cpp98ForwardIterator_ appears to refer to the table only (which just gives a bit of syntax). That feels like a regression to me.

CaseyCarter:
@jensmaurer Would it make sense to similarly define a `\concept` or `\libconcept` macro for styling (and potential indexing) of new-style concepts?

source/algorithms.tex
```diff
@@ -4514,10 +4514,10 @@
 The ranges \range{first}{middle} and \range{middle}{last} shall be
 sorted with respect to \tcode{operator<} or \tcode{comp}.
 \tcode{BidirectionalIterator} shall satisfy the
-\tcode{ValueSwappable} requirements\iref{swappable.requirements}. The type
+\textit{Cpp98ValueSwappable} requirements\iref{swappable.requirements}. The type
```
jensmaurer: Going forward, we really need to find standardized wording to say "satisfies the syntax constraints" vs. "syntax + semantics", and apply such phrasing uniformly.

source/algorithms.tex
```diff
@@ -1126,39 +1126,39 @@
...
 \item
 If an algorithm's template parameter is named
 \tcode{OutputIterator},
 \tcode{OutputIterator1},
 or
 \tcode{Output\-Iterator2},
-the template argument shall satisfy the requirements
-of an output iterator\iref{output.iterators}.
+the template argument shall satisfy the
+\textit{Cpp98OutputIterator} requirements\iref{output.iterators}.
```
jensmaurer: I'm wondering whether we can find a slightly less awkward spelling than "Cpp98Blah" for these old-style concepts. Maybe even something innovative such as OutputIterator<sub>98</sub> would work.

source/lib-intro.tex
```diff
@@ -219,6 +225,20 @@
 direct-initialization\iref{dcl.init}
 that is not list-initialization\iref{dcl.init.list}
 
+\definition{expression-equivalent}{defns.expression-equivalent}
+\indexdefn{expression-equivalent}%
+relationship that exists between two expressions \tcode{E1} and \tcode{E2} such that
+\begin{itemize}
```
jensmaurer: ISO wants us to phrase definitions of such terms in a way so that you can macro-expand the definition into the point-of-use and the sentence stays intact (modulo semantic parentheses implied by the macro). I think this wouldn't work here.

source/concepts.tex
```diff
@@ -0,0 +1,1249 @@
+%!TEX root = std.tex
+\rSec0[concepts.lib]{Concepts library}
```
jensmaurer: How much larger do you expect this section to grow? It's currently just 13 pages, which is a bit slim for a top-level clause in the library part of the standard.

source/lib-intro.tex
```diff
@@ -430,7 +452,7 @@
 \pnum
 Requirements are stated in terms of well-defined expressions that define valid terms of
 the types that satisfy the requirements. For every set of well-defined expression
-requirements there is a table that specifies an initial set of the valid expressions and
+requirements there is either a named concept or a table that specifies an initial set of the valid expressions and
```
jensmaurer: As explained elsewhere, this seems to be a lie for Cpp98ForwardIterator (for instance); the concept is not so named in the subclause, and the table of that name just shows syntax, not semantics.

source/lib-intro.tex
```diff
@@ -447,6 +469,23 @@
...
+concept\iref{concepts.lib.compare.stricttotallyordered} does not meet the
+semantic requirements of that concept when operating on NaNs.
+\end{example}
+This does not affect whether a type satisfies the concept.
+
+\pnum
+A declaration may explicitly impose requirements through its associated
+constraints\iref{temp.constr.decl}. When the associated constraints refer to a
+concept\iref{temp.concept}, additional semantic requirements are imposed on the
+use of the declaration.
```
jensmaurer: "additional" is great, but which ones in particular? Maybe the semantic requirements explained in the subclause that defines the concept?

source/algorithms.tex
```diff
@@ -1126,39 +1126,39 @@
...
 \item
 If an algorithm's template parameter is named
 \tcode{ForwardIterator},
 \tcode{ForwardIterator1},
 or
 \tcode{Forward\-Iterator2},
-the template argument shall satisfy the requirements
-of a forward iterator\iref{forward.iterators}.
+the template argument shall satisfy the
+\oldconcept{ForwardIterator} requirements\iref{forward.iterators}.
```
jensmaurer: The [forward.iterators] section is not so named, and the table of that name doesn't show semantics.

source/algorithms.tex
```diff
@@ -1126,39 +1126,39 @@
 or
 \tcode{Input\-Iterator2},
 the template argument shall satisfy the
-requirements of an input iterator\iref{input.iterators}.
+\oldconcept{InputIterator} requirements\iref{input.iterators}.
```
jensmaurer: Where in the incoming paper is authorization given for this change? InputIterator is not in the list of CamelCase concepts in section 1.3 of the paper.

source/lib-intro.tex
```diff
@@ -820,6 +859,47 @@
...
+Many of the customization point objects in the library evaluate function call
+expressions with an unqualified name which results in a call to a user-defined
+function found by argument dependent name lookup\iref{basic.lookup.argdep}. To
+preclude such an expression resulting in a call to unconstrained functions with
+the same name in namespace \tcode{std}, customization point objects specify that
+lookup for these expressions is performed in a context that includes deleted
+overloads matching the signatures of overloads defined in namespace \tcode{std}.
+When the deleted overloads are viable, user-defined overloads must be more
+specialized\iref{temp.func.order} or more constrained\iref{temp.constr.order} to
+be used by a customization point object.
```
jensmaurer: "more constrained" is a special case of "more specialized".

source/lib-intro.tex
```diff
@@ -2652,7 +2733,7 @@
 paragraph.
 \item
 if an incomplete type\iref{basic.types} is used as a template
-argument when instantiating a template component, unless specifically
+argument when instantiating a template component or evaluating a concept, unless specifically
```
jensmaurer: (Not editorial) I find it odd that an incomplete type would cause undefined behavior; it's a compile-time failure and should be ill-formed (possibly no diagnostic required).

jensmaurer:
@CaseyCarter "Would it make sense to similarly define a \concept or \libconcept macro for styling (and potential indexing) of new-style concepts?" 
Yes, please.

source/utilities.tex
```diff
@@ -16908,6 +16939,27 @@
...
 
+\indexlibrary{\idxcode{basic_common_reference}}%
+\tcode{template <class, class,}
+ \hspace*{2ex}\tcode{template <class> class,}
+ \hspace*{2ex}\tcode{template <class> class>}
+ \tcode{struct}
+ \hspace*{2ex}\tcode{basic_common_reference;}
+ &
+ A program may specialize this trait if at least one
+ template parameter in the specialization depends on a user-defined
```
jensmaurer: LWG 2139 (to be applied in #2193) now uses "program-defined type" here.

source/utilities.tex
```diff
@@ -16908,6 +16939,27 @@
...
+ type. If there is a member \tcode{type}, it shall be a \grammarterm{typedef-name}.
```
jensmaurer: (not editorial) This seems an overspecification; certainly a user could define the member "type" as a nested class (which is not a typedef-name). The difference is, I believe, unobservable.

source/utilities.tex
```diff
@@ -16959,6 +17011,49 @@
 supported\iref{basic.align}.
 
 \indexlibrary{\idxcode{common_type}}%
+\pnum
+Let:
+\begin{itemize}
+\item \tcode{CREF(A)} be \tcode{add_lvalue_reference_t<const remove_reference_t<A>{}>},
```
jensmaurer: We usually do \placeholder here (cf. INVOKE), because CREF(A) is not valid C++ source code; it's a funny abstract-level meta-type-function.

source/utilities.tex
```diff
@@ -16959,6 +17011,49 @@
...
+  of \tcode{A}'s cv and reference qualifiers, for a type \tcode{U} such
+  that \tcode{is_same_v<U, remove_cvref_t<U>>} is \tcode{true},
+\item \tcode{COPYCV(FROM, TO)} be an alias for type \tcode{TO} with the addition of
+  \tcode{FROM}'s top-level cv-qualifiers.
+  \begin{example}
+    \tcode{COPYCV(const int, volatile short)} is an alias for
+    \tcode{const volatile short}.
+  \end{example}
+\item \tcode{COND_RES(X, Y)} be
+  \tcode{decltype(false ? declval<X(\&)()>()() : declval<Y(\&)()>()())}.
```
jensmaurer: Why are we creating two invented function types here, just to call them and look at their result values? What's the net difference vs. omitting the indirection via the function references? What's the net difference vs. `common_type_t<X,Y>`?

source/utilities.tex
```diff
@@ -16982,13 +17077,18 @@
...
+      \tcode{common_type<D1, D2>}.
+    \end{note}
+  \item Otherwise, if
 \begin{codeblock}
 decay_t<decltype(false ? declval<D1>() : declval<D2>())>
 \end{codeblock}
-     \begin{note}
-     This will not apply if there is a specialization \tcode{common_type<D1, D2>}.
-     \end{note}
+    denotes a valid type, let \tcode{C} denote that type.
```
jensmaurer: Good, you already changed "its type" to "that type" here.

source/utilities.tex
```diff
@@ -17022,6 +17122,69 @@
 the same type, if any, as does \tcode{common_type_t<T2, T1>}.
 No diagnostic is required for a violation of this Note's rules.
 
+\pnum
+For the \tcode{common_reference} trait applied to a parameter pack \tcode{T} of
```
jensmaurer: We should add "Note C:" here and refer to it from the table, to be at least locally consistent. (Although, at a larger scale, we should move those "note-y" things out of the table and fully explain them outside of the table instead of having funny lettered notes which, adding insult to injury, are actually supposed to be normative.)

source/utilities.tex
```diff
@@ -16959,6 +17011,49 @@
...
 
 \indexlibrary{\idxcode{common_type}}%
+\pnum
+Let:
+\begin{itemize}
+\item \tcode{CREF(A)} be \tcode{add_lvalue_reference_t<const remove_reference_t<A>{}>},
+\item \tcode{XREF(A)} denote a unary class template \tcode{T} such that
+  \tcode{T<U>} denotes the same type as \tcode{U} with the addition
+  of \tcode{A}'s cv and reference qualifiers, for a type \tcode{U} such
+  that \tcode{is_same_v<U, remove_cvref_t<U>>} is \tcode{true},
```
jensmaurer: Is it too simple to say "for a cv-unqualified type U"?

source/utilities.tex
```diff
@@ -16959,6 +17011,49 @@
 supported\iref{basic.align}.
 
 \indexlibrary{\idxcode{common_type}}%
+\pnum
+Let:
+\begin{itemize}
+\item \tcode{CREF(A)} be \tcode{add_lvalue_reference_t<const remove_reference_t<A>{}>},
+\item \tcode{XREF(A)} denote a unary class template \tcode{T} such that
+  \tcode{T<U>} denotes the same type as \tcode{U} with the addition
+  of \tcode{A}'s cv and reference qualifiers, for a type \tcode{U} such
```
jensmaurer: How is that going to work? Suppose `U=int`; a class template `T<int>` can never be an "int". Is there a "::type" missing?

source/utilities.tex
```diff
@@ -17022,6 +17122,69 @@
...
+  \begin{itemize}
+  \item If there is such a type \tcode{C}, the member typedef \tcode{type} shall
+    denote the same type, if any, as \tcode{common_reference_t<C, Rest...>}.
+
+  \item Otherwise, there shall be no member \tcode{type}.
+  \end{itemize}
+\end{itemize}
+
+\pnum
+Notwithstanding the provisions of\iref{meta.type.synop}, and
```
jensmaurer: This needs a "Note D:" in front, for local consistency.

source/utilities.tex
```diff
@@ -17022,6 +17122,69 @@
...
+pursuant to\iref{namespace.std}, a program may specialize
```
jensmaurer: Both \iref's should be \ref's (with a space in front); the \iref automatically adds parens that we don't want here (compare to the parallel phrasing for common_type two paragraphs above).

source/utilities.tex
```diff
@@ -17022,6 +17122,69 @@
...
+the template arguments.
+\end{note}
+Such a specialization need not have a member named \tcode{type}, but if it does,
+that member shall be a \grammarterm{typedef-name} for an accessible and
+unambiguous type \tcode{C} to which each of the types \tcode{TQual<T>} and
+\tcode{UQual<U>} is convertible.
+Moreover, \tcode{basic_common_reference<T, U, TQual, UQual>::type} shall denote
+the same type, if any, as does \tcode{basic_common_reference<U, T, UQual, TQual>::type}.
+A program shall not specialize \tcode{basic_common_reference} on the third or
+fourth parameters, \tcode{TQual} or \tcode{UQual}. No diagnostic is required for
```
jensmaurer: What does that mean? Are you talking about partial specialization here? Then you should say so at the start of this paragraph. (A specialization is a non-template; a partial specialization is a template: two fundamentally different things.)

source/algorithms.tex
CaseyCarter: Subclause 1.3 is woefully incomplete: I missed the Iterator requirements tables and some more obscure sets of named requirements. The design intent - as discussed in LEWG and I believe LWG as well - was to rename *all* of the old named requirements tables to make way for current and future concepts.

It's also important that all references to the requirements use the proper name. The colloquialism "input iterators", for example, can now be interpreted to mean either types that meet the Cpp98InputIterator requirements or types that model the `InputIterator` concept. We need to very clearly define for users exactly to which set of requirements each such usage is intended to refer.

source/numerics.tex
```diff
@@ -1786,6 +1786,30 @@
 \indextext{requirements!uniform random bit generator|)}%
 \indextext{uniform random bit generator!requirements|)}
```
jensmaurer: Not done: "[Editor’s note: Relocate "Header /tcode<random> synopsis" [rand.synopsis] before 29.6.1 "Requirements"
[rand.req]]"

source/algorithms.tex
CaseyCarter: I'm working on that. My current plan is to uniformly use "`T` meets the CppXXX requirements" for "old" concepts (which is the prevalent library usage, this occurrence is aberrant), "`T` satisfies `Concept`" to mean purely syntactic conformance (which is the proper core language meaning of :"satisfy"), and "`T` models `Concept`" to mean that `T` must both satisfy `Concept` and must further meet the semantic requirements or the program is ill-formed NDR.

I'm working on rewording the combined Ranges paper to use that style, and plan to submit future editorial changes to the rest of the library specification to make everything uniform.

source/numerics.tex
```diff
@@ -1786,6 +1786,30 @@
...
 
+\pnum
+The \tcode{UniformRandomBitGenerator} concept is a slight relaxation of the uniform random bit generator requirements, in that it does not require a nested \grammarterm{typedef-name} \tcode{result_type}.
```
jensmaurer: Please add line breaks after at most 79 chars, preferably "semantic line breaks", i.e. do not break in the middle of a phrase or term.

source/numerics.tex
```diff
@@ -1786,6 +1786,30 @@
...
+
```
jensmaurer: The presentation seems suboptimal and redundant here. It seems more compact to introduce \concept{UniformRandomBitGenerator} first (which is the more relaxed concept) and then define "uniform random bit generator" in terms of that, adding the result_type requirement.

source/algorithms.tex
CaseyCarter: "forward iterators" will be ambiguous once Ranges are merged. I understand your point about the distinction between the table and the additional requirements outside the table, though. Do you agree it would be clearer if I retitle the existing sections e.g. "Cpp98ForwardIterator requirements"?

jensmaurer:
I've reviewed all changes except the entirely-new top-level clause "concepts".

source/algorithms.tex
CaseyCarter: I'm certainly not married to "Cpp98Blah". It's better than LEWG's suggestion "STL1Blah", but that wasn't a high bar to meet. Now that the styling is macro-ized, we can change it with a one-liner.

source/concepts.tex
CaseyCarter: I suspect library will add more concepts here in the future, though I can't speculate how many. Do you think this would fit better as a subclause somewhere in lib-intro?

source/algorithms.tex
jensmaurer: Yes, I think so. I fear there are Cpp98 concepts definitions that don't have a section of their own, though, so this strategy might fail here and there. Before you subscribe for more work, @zygoloid should really chime in how he wants to proceed here.
In any case, make sure all those changes are a (one) separate commit so that they can be messed with easily.

source/lib-intro.tex
CaseyCarter: [temp.constr.order]/3 defines "more constrained" as an orthogonal notion to "more specialized." The two notions interact in overload resolution where "more constrained" is used to discriminate between overloads neither of which is more specialized, but "more constrained" is not otherwise a special case.

source/algorithms.tex
jensmaurer: I understand the desires, but the paper as moved doesn't say so, and there isn't even a trace of a hint in the paper that this might be the intent, given that only CamelCased requirements are quoted in 1.3. (I'd be happy to be shown otherwise.)

@zygoloid, are we overstepping our editorial leeway here by changing such phrases, in particular since the result could be incorrectly interpreted as referring to the table only, not the entire set of requirements? (I think we had an editorial issue a few months ago where exactly that point was a concern.)

source/algorithms.tex
CaseyCarter: Repeating my reply to your original comment (that Github will likely bury after the `\oldconcept` push):

Subclause 1.3 is woefully incomplete: I missed the Iterator requirements tables and some more obscure sets of named requirements. The design intent - as discussed in LEWG and I believe LWG as well - was to rename all of the old named requirements tables to make way for current and future concepts.

It's also important that all references to the requirements use the proper name. The colloquialism "input iterators", for example, can now be interpreted to mean either types that meet the Cpp98InputIterator requirements or types that model the InputIterator concept. We need to very clearly define for users exactly to which set of requirements each such usage is intended to refer.

source/lib-intro.tex
CaseyCarter: How about "When the associated constraints refer to a concept\iref{temp.concept}, the semantic constraints specified for that concept are additionally imposed on the use of the declaration."?

source/algorithms.tex
CaseyCarter: Repeating my reply to your original comment (that Github will likely bury after the `\oldconcept` push):

"forward iterators" will be ambiguous once Ranges are merged. I understand your point about the distinction between the table and the additional requirements outside the table, though. Do you agree it would be clearer if I retitle the existing sections e.g. "Cpp98ForwardIterator requirements"?

source/lib-intro.tex
CaseyCarter: The tables typically have both syntactic and semantic requirements, but fail to include additional semantic requirements specified in prose in the section that contains the table. Do you agree that retitling the sections will clarify the intent?

source/lib-intro.tex
CaseyCarter: LWG has been addressing a similar problem with the type traits to require implementations to diagnose precondition violations. There's currently no technology to do so with concepts, but this at least leaves the avenue open for future changes.

source/utilities.tex
CaseyCarter: I'll update uses of "user-defined type" accordingly.

source/lib-intro.tex
CaseyCarter: I'll fix this.

source/concepts.tex
jensmaurer: Well, lib-intro only has meta-level info and no specific header files, so that doesn't quite fit. On the other hand, concepts are meta-level stuff, so those might indeed fit. @zygoloid, what do you think?
(The rough math is 30 top-level clauses vs. 1500 pages = 50 pages per clause.)

source/lib-intro.tex
jensmaurer: Yes, seems to be an improvement. (Make sure to keep those rephrasings separate from the original commit with the as-moved paper.)

source/utilities.tex
CaseyCarter: This is a truly disgusting hack to avoid `declval`'s `add_rvalue_reference_t` on the return type. `decltype(declval<X>())` is `add_rvalue_reference_t<X>`, `decltype(declval<X(&)()>()()` is exactly `X`. We discussed at length in LWG that this is abominable, and that no one had an idea for an improvement.

source/utilities.tex
CaseyCarter: This is the problematic wording I noted in the issue: "unary class template" here should be "unary alias template" or "unary type template". Marshall requested that I fix this via a followup LWG issue.

source/lib-intro.tex
jensmaurer: I'm complaining about the fact that this situation is in the "undefined behavior" category, not in a (separate) category "ill-formed, no diagnostic required". The latter would leave avenues open, too, but would also permit implementations to diagnose and refuse e.g. incomplete types today. And, maybe there is even a conflict here if some preconditions says "shall be a complete type" (i.e. compile-time diagnostic), yet this section says "undefined behavior" (i.e. only bad if you actually evaluate it, which seems a misguided idea for a compile-time violation).

source/utilities.tex
CaseyCarter: I somewhat blindingly replaced `ref`s with `iref`s - I'll audit them all.

source/utilities.tex
jensmaurer: ... and `common_type_t<X,Y>` doesn't work because?

source/concepts.tex
```diff
@@ -0,0 +1,1249 @@
...
+template <class T>
+void g(T) requires is_c1_v<T>; // \#3
+template <C2 T> void g(T);     // \#4
+\end{codeblock}
+Since \tcode{C2} subsumes \tcode{C1}, the expression \tcode{f(42)} unambiguously
+calls \#2. \tcode{g(42)}, however, is potentially ambiguous if there need be no
+subsumption relationship between \tcode{C1} and \tcode{is_c1_v}.
+\end{example}
+
+\rSec2[concepts.lib.corelang.same]{Concept \tcode{Same}}
```
jensmaurer: In general, let's try to make the labels a little shorter. Just "concepts.same" or ... singular "concept.same" seems sufficient. (Yes, rplease ename all labels from concepts.... to concept...)

source/concepts.tex
```diff
@@ -0,0 +1,1249 @@
...
+  template <class R, class T, class U>
+  concept Relation = @\seebelow@;
+
+  // \ref{concepts.lib.callable.strictweakorder}, StrictWeakOrder:
+  template <class R, class T, class U>
+  concept StrictWeakOrder = @\seebelow@;
+}
+\end{codeblock}
+
+\rSec1[concepts.lib.corelang]{Core language concepts}
```
jensmaurer: I don't think we call this "core language" in the standard ever. We have a "language support library", so maybe these are just "language-related concepts"?

source/concepts.tex
CaseyCarter: The intent was to put `<concepts>` early so that concept therein could be defined before appearing in the rest of the library spec. I suppose that's not really critical: we use type traits heavily before they are defined.

source/concepts.tex
```diff
@@ -0,0 +1,1249 @@
...
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+Let \tcode{C} be \tcode{common_reference_t<T, U>}. Let \tcode{t} be a
+function whose return type is \tcode{T}, and let \tcode{u} be a function
+whose return type is \tcode{U}. \tcode{CommonReference<T, U>} is satisfied
+only if:
+\begin{itemize}
+\item \tcode{C(t())} equals \tcode{C(t())} if and only if \tcode{t()} is an
```
jensmaurer: The "if and only if" feels wrong here; it seems to say that C(t()) must have a value different from C(t()) if t() is not equality preserving. That seems hard to do, since t() could modify some unknown global variable.

source/concepts.tex
```diff
@@ -0,0 +1,1249 @@
...
+
+\pnum
+\begin{note}
+Users can customize the behavior of \tcode{Common} by specializing the
+\tcode{common_type} class template\iref{meta.trans.other}.
+\end{note}
+
+\end{itemdescr}
+
+\rSec2[concepts.lib.corelang.integral]{Concept \tcode{Integral}}
```
jensmaurer: In general, we're trying to avoid single-paragraph sections. This area is getting dangerously close to the cliff, with four subheadings on one page. Is there any plan at all to possibly have multiple (simple) concepts in one subclause? Maybe not. Net result: no change requested for now.

source/concepts.tex
```diff
@@ -0,0 +1,1249 @@
...
+concept Integral = is_integral_v<T>;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+There need be no subsumption relationship between \tcode{Integral<T>} and
+\tcode{is_integral_v<T>}.
+\end{itemdescr}
+
+\rSec2[concepts.lib.corelang.signedintegral]{Concept \tcode{SignedIntegral}}
```
jensmaurer: Label: concept.signed.int

source/concepts.tex
```diff
@@ -0,0 +1,1249 @@
...
+\tcode{is_signed_v<T>}.
+
+\pnum
+\begin{note}
+\tcode{SignedIntegral<T>} can be satisfied even by types that are not signed
+integral types\iref{basic.fundamental}; for example, \tcode{char}.
+\end{note}
+\end{itemdescr}
+
+\rSec2[concepts.lib.corelang.unsignedintegral]{Concept \tcode{UnsignedIntegral}}
```
jensmaurer: Label: concept.unsigned.int

source/concepts.tex
```diff
@@ -0,0 +1,1249 @@
...
+\item After the definition \tcode{T u = v;}, \tcode{u} is equal to \tcode{v}.
+
+\item \tcode{T(v)} is equal to \tcode{v}.
+\end{itemize}
+
+\end{itemdescr}
+
+\rSec1[concepts.lib.compare]{Comparison concepts}
+
+\rSec2[concepts.lib.compare.general]{General}
```
jensmaurer: While ISO says "no hanging paragraphs", I'm not enamoured with single-sentence sections; I think having the essentially non-normative intro in the parent section would work, too. @zygoloid?

source/concepts.tex
```diff
@@ -0,0 +1,1249 @@
...
+
+\pnum
+An rvalue or lvalue \tcode{t} is \defn{swappable} if and only if \tcode{t} is
+swappable with any rvalue or lvalue, respectively, of type \tcode{T}.
+
+\begin{example}
+User code can ensure that the evaluation of \tcode{swap} calls
+is performed in an appropriate context under the various conditions as follows:
+\begin{codeblock}
+#include <utility>
```
jensmaurer: Is there a `#include <concepts>` missing?

source/utilities.tex
CaseyCarter: Yes, the intent here is that users partially specialize `basic_common_reference`. I admit that I can't see a reason now for "A program shall not specialize `\tcode{basic_common_reference}` on the third or fourth parameters, `\tcode{TQual}` or `\tcode{UQual}`." I believe the intent is to make it harder for users to do something stupid and violate the other requirements and/or point them more directly down the intended path.

@ericniebler, can you clarify the intent of this particular requirement?

source/algorithms.tex
timsong-cpp: I think that was me, and the issue was an edit that changed a cross-reference to the `NullablePointer` requirements section to point to the table instead.

source/algorithms.tex
timsong-cpp: _Cpp98MoveConstructible_ certainly feels anachronistic.

source/utilities.tex
timsong-cpp: Technically, it's observable in some cases with an _elaborated-type-specifier_.

source/utilities.tex
timsong-cpp: For cppreference, I just invented a `declval` substitute that returns `T` instead and used that, FWIW.

(`decltype(declval<X(&)()>()())` is also not exactly `X`; it's subject to the usual prvalue cv-qualification adjustment rules and the rvalue-reference-to-function-is-lvalue rule.)

source/numerics.tex
CaseyCarter: You mean something like [urbg.pdf](https://github.com/cplusplus/draft/files/2117498/urbg.pdf)? That seems to go a bit past "editorial", but I'm game if you folks are.

source/concepts.tex
```diff
@@ -0,0 +1,1249 @@
...
+based on knowledge of the prior values as observed via equality preserving expressions.
+It effectively forbids spontaneous changes to an object, changes to an object from
+another thread of execution, changes to an object as side effects of non-modifying
+expressions, and changes to an object as side effects of modifying a distinct object
+if those changes could be observable to a library function via an equality preserving
+expression that is required to be valid for that object.
+\end{note}
+
+\pnum
+Expressions declared in a \grammarterm{requires-expression} in this document are
```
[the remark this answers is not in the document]
timsong-cpp: We always want explicit convertibility as well when we require implicit convertibility, which means that we'd use `ConvertibleTo` instead. (But perhaps `ConvertibleTo` is missing a requirement that the conversion is equality-preserving?)

source/concepts.tex
```diff
@@ -0,0 +1,1249 @@
...
+An rvalue or lvalue \tcode{t} is \defn{swappable} if and only if \tcode{t} is
+swappable with any rvalue or lvalue, respectively, of type \tcode{T}.
+
+\begin{example}
+User code can ensure that the evaluation of \tcode{swap} calls
+is performed in an appropriate context under the various conditions as follows:
+\begin{codeblock}
+#include <utility>
+
+template <class T, SwappableWith<T> U>
```
timsong-cpp: and a `std::`.

source/numerics.tex
timsong-cpp: It also relaxes the rules by only requiring the result type to model `UnsignedIntegral` instead of be an unsigned integer type. The former permits `bool` and character types that are unsigned.

W-E-Brown:
> On Jun 19, 2018, at 9:12 PM, timsong-cpp <notifications@github.com> wrote:
> 
> @timsong-cpp commented on this pull request.
> 
> In source/numerics.tex:
> 
> > @@ -1786,6 +1786,30 @@
>  \indextext{requirements!uniform random bit generator|)}%
>  \indextext{uniform random bit generator!requirements|)}
>  
> +\pnum
> +The \tcode{UniformRandomBitGenerator} concept is a slight relaxation of the uniform random bit generator requirements, in that it does not require a nested \grammarterm{typedef-name} \tcode{result_type}.
> +
> 
> It also relaxes the rules by only requiring the result type to model UnsignedIntegral instead of be an unsigned integer type. The former permits bool and character types that are unsigned.

I believe not.  Please see [rand.req.genl]/1f.

timsong-cpp:
We don't use `UIntType` anywhere in [rand.req.urng]. The "uniform random bit generator" requirements merely requires `result_type` to be "an unsigned integer type". The `UniformRandomBitGenerator` concept relaxes that and only requires the result type model `UnsignedIntegral`. 

The engine templates in the standard do use `UIntType`, so instantiating them with, say, `char32_t` or `unsigned char`, is undefined. But it appears to be possible to write a type yourself with a `result_type` of `unsigned char` and have it meet the uniform random bit generator requirements.

source/concepts.tex
```diff
@@ -0,0 +1,1249 @@
...
+
+\pnum
+Not all input values must be valid for a given expression; e.g., for integers \tcode{a}
+and \tcode{b}, the expression \tcode{a / b} is not well-defined when \tcode{b} is \tcode{0}. This
+does not preclude the expression \tcode{a / b} being equality preserving. The
+\term{domain} of an expression is the set of input values for which the
+expression is required to be well-defined.
+
+\pnum
+Expressions required by this specification to be equality preserving are further
```
jensmaurer: this specification -> this document

source/concepts.tex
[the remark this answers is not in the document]
jensmaurer: Hm... ConvertibleTo has a _simple-requirement_, which contains a `static_cast<>` expression, which we're instructed needs to be equality preserving. Ah, but... see below.

source/concepts.tex
```diff
@@ -0,0 +1,1249 @@
...
+object to which it refers is valid but unspecified\iref{lib.types.movedfrom}.
+
+\item Otherwise, if \tcode{rhs} is a glvalue, the object to which it refers is not
+modified.
+\end{itemize}
+\end{itemize}
+
+\pnum
+There need be no subsumption relationship between \tcode{Assignable<LHS, RHS>}
+and \tcode{is_\-lval\-ue_\-ref\-er\-ence_v<LHS>}.
```
jensmaurer: Could you please limit the hyphenation hint to exactly the one you need to avoid the overfull \hbox, if any?

source/algorithms.tex
CaseyCarter: > Cpp98MoveConstructible certainly feels anachronistic.

We specifically pointed out that example in both LEWG discussion of the Ranges proposal and LWG discussion of this proposal, and both subgroups found the confusion level acceptable.

source/lib-intro.tex
CaseyCarter: I agree. I've made a note to file an LWG issue to adjust this wording to IFNDR.

source/utilities.tex
CaseyCarter: I'd be happy to change this to "If there is a member `\tcode{type}`, it shall denote a type." if we're willing to call that change editorial-ish.

source/utilities.tex
CaseyCarter: I'd hate to define yet another exposition-only helper here that's only going to be used in this one place. Do we agree that `template <class T> T __val() noexcept; // \expos` ... `decltype(false ? __val<X>() : __val<Y>())` is an improvement?

source/numerics.tex
CaseyCarter: I must have ignored this directive upon finding no header named `/tcode<random>` in the library ;)

source/utilities.tex
ericniebler: Yes, the intent here is to disallow specifying the third or fourth template parameters in either a partial or explicit specialization. Those parameters are specified by the `common_reference` trait. The exact unary templates used as template arguments to the (possibly user-specialized) `basic_common_reference` instantiation are unspecified, and implementors are free to change them between releases. Any code that specializes on those parameters (either partial or explicit) is _de facto_ wrong.

CaseyCarter: Does a program with such a specialization *need* to be IFNDR? IMO it's sufficient that `common_reference` specifies how and when it works, and users shouldn't be surprised when their broken specializations are never selected.

source/numerics.tex
CaseyCarter: I'll reword to:
```tex
A class \tcode{G} meets the \term{uniform random bit generator} requirements if
\tcode{G} models \libconcept{UniformRandomBitGenerator},
\tcode{invoke_result_t<G\&>} is an unsigned integer type\iref{basic.fundamental},
and
\tcode{G} provides a nested \grammarterm{typedef-name} \tcode{result_type}
that denotes the same type as \tcode{invoke_result_t<G\&>}.
```
to fix the normative difference between "unsigned integer type" and `UnsignedIntegral` and add this in a separate commit.

source/utilities.tex
ericniebler: I suppose not. Striketh sentence if it offendeth thee.

An alternative which I hadn't considered until just now is to redesign the trait to make it impossible to misuse:

```c++
template <class A, class B>
struct basic_common_reference {
  template <template <class> class AQual, template <class> class BQual>
  using type = ...;
};
```

source/concepts.tex
```diff
@@ -0,0 +1,1249 @@
...
+
+\pnum
+Expressions declared in a \grammarterm{requires-expression} in this document are
+required to be equality preserving, except for those annotated with the comment
+``not required to be equality preserving.'' An expression so annotated
+may be equality preserving, but is not required to be so.
+
+\pnum
+An expression that may alter the value of one or more of its inputs in a manner
+observable to equality preserving expressions is said to modify those inputs.
```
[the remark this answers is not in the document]
CaseyCarter: I believe all uses of "equality preserving" (with the obvious exception of the definition) are adjective phrases and should be hyphenated.

source/concepts.tex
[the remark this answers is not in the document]
CaseyCarter: We use the term "expressions" in library to refer to what are more precisely "expression patterns" that are required to be valid when program-defined types/values are substituted into the expression pattern. For example, [structure.requirements]/4 states: "Requirements are stated in terms of well-defined **expressions** that define valid terms of the types that satisfy the requirements." I'm (ab)using the term similarly here to refer to the expression patterns declared in simple and compound requirements in a *requires-expression*.

I think our intent is (or should be) to include implicit conversion constraints introduced by a *compound-requirement*, although the question is largely moot since we avoid implicit conversion constraints like the plague and (as @timsong-cpp said) typically require deduction constraints a la `{ E } -> ConvertibleTo<T>;` in preference to `{ E } -> T;`. (I've been telling anyone who will listen since 2015 that implicit conversion constraints are largely useless because they result in expression patterns that aren't composable.)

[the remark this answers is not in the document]
ericniebler: I wonder if in the library preamble we can handwave that when the library uses an implicit conversion constrain we also require explicit conversion (with the same semantics) and equality preservation. And then stump for more help from the core language regarding conversion constraints.

source/concepts.tex
CaseyCarter: I'm re-stable-naming all of the specific concept subclauses to "concept.name", and leaving the general and collection stable names as "concepts.syn", "concepts.general", "concepts.compare".

source/concepts.tex
CaseyCarter: or two.

source/concepts.tex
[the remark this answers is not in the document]
CaseyCarter: I'd prefer to be explicit about conversions. I don't think writing `{ E } -> ConvertibleTo<T>;` is a significant burden over `{ E } -> T;` - not enough so to have the same syntax implicitly result in different *checked* constraints, anyway. I'll file an issue to deal with the interaction of equality-preservation and implicit conversion constraints. It doesn't really matter to Ranges, but I suspect someone will eventually want to specify a library component with such a constraint and it needs to have a well-defined meaning rather than us simply telling them to not do that bad thing.

source/concepts.tex
CaseyCarter: I'd be mildly concerned about confusion as to which concepts correspond with which prose requirements - despite that prose requirements are always introduced with "Let `T` be ... Concept<T> is satisfied only if"- but I don't think that's insurmountable.

source/concepts.tex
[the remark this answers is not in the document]
CaseyCarter: Clarification: today we have to write `E; requires ConvertibleTo<decltype((E)), T>;` to express that constraint properly in terms of the exact type and value category of `E`. Once P1084 is adopted, it can be simplified to `{ E } -> ConvertibleTo<T>;`.

source/utilities.tex
CaseyCarter: > I'd hate to define yet another exposition-only helper here

I thought this discussion was familiar: https://github.com/ericniebler/stl2/issues/338#issuecomment-309616140.

source/utilities.tex
CaseyCarter: I don't know that I buy "impossible to misuse", but I'll make a note to possibly file an issue to make this design change after we investigate it more thoroughly.

...and strike the requirement for now, since we think it adds no useful normative effect.

source/concepts.tex
```diff
@@ -0,0 +1,1249 @@
...
+Not all input values must be valid for a given expression; e.g., for integers \tcode{a}
+and \tcode{b}, the expression \tcode{a / b} is not well-defined when \tcode{b} is \tcode{0}. This
+does not preclude the expression \tcode{a / b} being equality preserving. The
+\term{domain} of an expression is the set of input values for which the
+expression is required to be well-defined.
+
+\pnum
+Expressions required by this specification to be equality preserving are further
+required to be stable: two evaluations of such an expression with the same input
+objects must have equal outputs absent any explicit intervening modification of
```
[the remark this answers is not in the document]
CaseyCarter: I sincerely hope that "the rest of the text" means the text added in this PR, and not the full text of the Standard. There are a great many "must"s that we must eradicate to please ISO.

source/concepts.tex
```diff
@@ -0,0 +1,1249 @@
...
+\pnum
+The \tcode{ConvertibleTo} concept requires an expression of a particular type
+and value category to be both implicitly and explicitly convertible to some other
+type. The implicit and explicit conversions must produce equal results.
+
+\indexlibrary{\idxcode{ConvertibleTo}}%
+\begin{itemdecl}
+template <class From, class To>
+concept ConvertibleTo = is_convertible_v<From, To> &&
+  requires(From (&f)()) { static_cast<To>(f()); };
```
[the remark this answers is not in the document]
CaseyCarter: The "operands"/"inputs"/"outputs" wording in `concepts.general.equality` is still somewhat soft and in need of clarification: it's incredibly hard to specify without exploding the text into full mathematical rigor that would be unacceptable for the IS. My mildly hand-wavy response is that I would consider both `f` and `f()` to be operands of this expression pattern.

The prose does try to clarify this:
```tex
and let `\tcode{f}` be a function with no arguments and return type \tcode{From}
such that \tcode{f()} is equality-preserving.
\tcode{ConvertibleTo<From, To>} is satisfied only if:
```
but there's no correspondence established between `f` in the concept definition and this `f`. I suppose we could clarify in the concept definition itself:
```c++
template <class From, class To>
concept ConvertibleTo = is_convertible_v<From, To> &&
  requires(From (&f)()) { // f is an arbitrary function such that f() is an equality-preserving expression
    static_cast<To>(f());
  };
```
Would you consider that an improvement?

CaseyCarter:
Pushing what I have so far. I still need to fix the definition of "expression-equivalent" and cleanup the weird use of "if and only if" in `Common` and `CommonReference`, and we haven't reached a decision about renaming the "old concept" subclauses to "*Cpp98Foo* requirements".

source/algorithms.tex
jensmaurer: I'm ok with that, although that doesn't really remove the ambiguity if the table continues to be labeled the same as the section. @zygoloid?

source/algorithms.tex
jensmaurer: @zygoloid, what's your guidance here?

source/lib-intro.tex
jensmaurer: Thank you.

source/utilities.tex
jensmaurer: @timsong-cpp is right, such a change would be observable so can't be done editorially. @CaseyCarter, could you please check whether there is already an LWG issue for that, and file one if not?

source/utilities.tex
jensmaurer: Maybe we shouldn't change anything here and go with the formulation in the paper. These nuances are important, but the code snippets for the meta-functions are way too obtuse to communicate that effectively. Can we maybe add a few notes or examples explaining what the intent is (in particular involving subtleties of prvalues vs. xvalues)? (As a matter of principle, I'm wondering whether some prose relying on the core language mechanics of the ?: operator or "common type" would be shorter and easier to understand. Not now.)

source/algorithms.tex
CaseyCarter: Yes, I'd thought about the ambiguity of both subclauses and tables titled "Foo requirements," although that's pre-existing, for example, [allocator.requirements] is titled "Allocator requirements" and contains Table 31 "Allocator requirements". We should retitle those tables to something generic like "Allocator valid expressions with semantics" - [structure.requirements]/4 says they are "a
table that specifies an initial set of the valid expressions and their semantics." Which is also a misnomer.

I'm waiting for Richard's nod to enact these changes. I have more work to do on the merged Ranges proposal than time before the mailing deadline to do it in, so I don't want to walk down any blind alleys.

source/concepts.tex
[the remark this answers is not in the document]
jensmaurer: @CaseyCarter: I don't buy the "expression pattern" theory for the case of new-style concepts. (I agree regarding old-style concepts.) With concepts in the core language, we no longer have to handwave which expressions we mean:  Concepts show actual code that can be compiled. There are no "patterns" going on beyond what the core language prescribes. To wit: the text is referring to the grammar non-terminal _requires-expression_.
My narrow (editorial) issue here is that "expressions declared in..." sounds broken to me; expressions are never "declared". Further (new concern), it's unclear whether "in" is also intended to cover any subexpression of the top-level expressions in the requirements.
I'd appreciate a clear syntax decomposition of requires-expression to define which expressions are in view here (maybe call them "required expressions" or so). Maybe that's not editorial to start with. @zygoloid?

source/utilities.tex
CaseyCarter: Yes I will. I have a set of notes on all the comments here so I can track which changes I've accomplished. There are six entries now for issues to file post-merge.

source/concepts.tex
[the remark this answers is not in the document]
jensmaurer: Then we should also hyphenate the definition (and maybe consider rephrasing it to "An equality-preserving expression is an expression that ...")

source/concepts.tex
jensmaurer: Agreed with an LWG issue.

source/concepts.tex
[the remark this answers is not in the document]
jensmaurer: "f is an arbitrary function" -> "f refers to an arbitrary function" (f is still a reference, not a function)

source/utilities.tex
CaseyCarter: I did change the first sentence of what is now "Note D" to read:
> Notwithstanding the provisions of \ref{meta.type.synop}, and pursuant to \ref{namespace.std}, a program may **partially** specialize \tcode{basic_common_reference<T, U, TQual, UQual>} for types \tcode{T} and \tcode{U} such that \tcode{is_same_v<T, decay_t<T>{>}} and \tcode{is_same_v<U, decay_t<U>{>}} are each \tcode{true}.

which I consider to be an editorial clarification. If you feel that change is insufficient or not editorial, I'll revert it and file an issue.

source/concepts.tex
jensmaurer: How so? We usually don't prefix with `std::` in the library sections of the standard, but maybe I'm overlooking something.

source/concepts.tex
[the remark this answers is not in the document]
jensmaurer: @CaseyCarter: Yes, the text of the PR. Our other "musts" should be in notes only.

jensmaurer:
Sorry, @CaseyCarter, github still tells me there are conflicts that need a rebase.

To clarify the rules on "Fixup" vs. regular commits: "Fixup" is stuff that would have been in another commit if the typist of the commit wouldn't have made a dumb mistake. That includes fixing typos (commas, misspellings) and incorrect transcriptions from the incoming paper. For example, the move of the "random" synopsis is a "fixup", as is the \oldconcept stuff and the \defnx and \ref / \iref stuff and \placeholder. (Note that the point-of-reference for "fixup" is the rendered text, not the situation of the LaTeX source.)

The change of the section labels should come right after that.

Adding "Note C" and "Note D" (or striking a superfluous sentence), in contrast, is not a fixup, but a separate editorial rephrasing that should follow our regular commit guidelines. For example, start the commit comment with [section.label].

As agreed with @zygoloid, the feature-test macro is not a fixup because the actual wording diff was not in the incoming paper (for lack of a reference in the working draft).

Yes, you'll need to git push --force-with-lease the rephrasing of the commit logs that I ask for here. Since you'll be doing that, you might as well fold the real fixups into the commits they refer to --- but some of your "fixup"s are actually editorial rephrasings.

source/concepts.tex
timsong-cpp: Also `#include <cassert>`, or change the assertions in `main` to the shiny `[[assert: ... ]];`

source/concepts.tex
timsong-cpp: We also don't usually include headers in library examples. Since this particular example is hypothetical user code, not library code, with a `main()` and all, I think it makes more sense to write it like a user would.

source/concepts.tex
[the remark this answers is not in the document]
CaseyCarter: Would replacing "Expressions declared in a *requires-expression* in this document..." with "Expressions that appear in a *simple-requirement* or *compound-requirement* of a *requires-expression* in the standard library..." be an improvement? That seems to address much of your concerns.

source/concepts.tex
timsong-cpp: I expect these sections to shrink further once we remove the unnecessary "need be no subsumption" wording.

source/concepts.tex
CaseyCarter: Elaborating @timsong-cpp's statement: I dropped this example in R2 of the proposal, and Daniel Krügler specifically asked me to replace it. This particular example is crucial to readers understanding how the ADL 2-step works in practice. It needs full qualification to make it clear how the namespaces and qualifications interact.

CaseyCarter:
> Sorry, @CaseyCarter, github still tells me there are conflicts that need a rebase.

I shouldn't have bothered to ask: I'm touching enough that virtual every commit causes a conflict.

> To clarify the rules on "Fixup" vs. regular commits: ...

Thank you for the clarification. I'll clean up the commit history.

CaseyCarter:
The commit history has been so rewritten - after two broken attempts. Apologies for the commit spam.

timsong-cpp:
We'll need to make sure that we switch to `\oldconcept` for the stuff added by the other motions too.

source/concepts.tex
```diff
@@ -0,0 +1,1310 @@
...
+\pnum
+This section contains the definition of concepts corresponding to language
+features. These concepts express relationships between types, type
+classifications, and fundamental type properties.
+
+\pnum
+\begin{example}
+Several concept specifications in this subclause are specified in terms of a
+type trait, and include an annotation that there need be no subsumption
+relationship between the concept and the type trait. Given the following set of
```
zygoloid: This doesn't make sense. There *cannot* be any subsumption relationship between a concept and an arbitrary non-concept expression. The core language doesn't allow it.

We should strike this paragraph and all subsequent "There need be no subsumption relationship" paragraphs.

zygoloid: If you like, you could convert this example to a note that there will be no such subsumption relationship.

source/macros.tex
```diff
@@ -316,6 +316,10 @@
 %% Produces 9 output characters.
 \newcommand{\commentellip}{\tcode{/* ...\ */}}
 
+%% Concepts
+\newcommand{\oldconcept}[1]{\textit{Cpp98#1}}
```
zygoloid: This seems wrong to me. For a start, many of the things we'll be giving these names to did not exist in C++98 (why would we call things *Cpp98MoveConstrutible* or *Cpp98CopyInsertible* when C++98 had no idea that such things existed?). Also, C++98 doesn't exist any more, it was replaced by C++03. Using Cpp17 as the prefix instead would help with that...

But I'm also generally very unhappy about having two different definitions for the same term that are distinguished, effectively, by one of then being labeled "Old". Do we have a plan for getting rid of the old names before we ship the C++20 IS? If not, I think we should consider at least adding some introductory verbiage somewhere cautioning the reader that we have (eg) two things called CopyConstructible and they are not necessarily the same. Actually, that's wrong, we have three -- *Cpp98CopyConstructible*, `CopyConstructible<T>`, and `is_copy_constructible_v<T>` are all different. :-(

source/concepts.tex
zygoloid: Yes, that wording is intended to convey that we mean the same utterance of that expression. Please feel free to file an editorial issue to clarify that.

source/macros.tex
CaseyCarter: I'm happy to change the prefix to `Cpp17`, or nearly anything else - no one in LWG or LEWG review has yet expressed concern about or preference for a different prefix.

Getting rid of the old names would necessitate jettisoning the majority of the Standard Library which is not likely to happen for C++20. We could make the names more different, but that would actually be counterproductive: the new names for the "old" terms must be similar to avoid confusion, and the names for the new concepts should be similar to the name of the old term it is intended to supplant.

zygoloid: Perhaps just some words in the library front matter about the formatting convention would help? (Analogous to [syntax] in the language front matter.)

I've also realized that we have four different things called copy constructible, since the `CopyConstructible` concept has both syntactic requirements and semantic ones. Speaking of which: have we concluded how we should be specifying the syntactic versus semantic requirements yet? I like using "models" for the latter, and I'd suggest we use "satisfies" for the former (matching the core language definition) or the `Concept<X>` is `true` formulation suggested on the lib reflector.

CaseyCarter: > I've also realized that we have four different things called copy constructible, since the `CopyConstructible` concept has both syntactic requirements and semantic ones.

As does `Cpp17CopyConstructible` - make it five. And yes, that usage of "models" and "satisfies" is exactly what I'm working toward in the merged Ranges proposal - it is *far* clearer than the current "satisfies the syntactic/semantic requirements" mismash. I'll submit an editorial cleanup for uses in the working draft once I'm sure LWG is happy with the formulation.

source/concepts.tex
CaseyCarter: I've made this change, and pushed another commit to merge `Integral`, `SignedIntegral`, and `UnsignedIntegral` into a single subclause - they are now too small for individual clauses.

source/macros.tex
CaseyCarter: I've pushed a commit to change the slug to *Cpp17*, and another adding a description of the typographic convention used for requirement names to [structure.requirements].

zygoloid:
Merging to motions branch. Will open a new pull request to merge into master.

source/lib-intro.tex
CaseyCarter: This is [LWG 3142](https://wg21.link/lwg3142).

source/algorithms.tex
jensmaurer: See also some earlier discussion in #1263. I understand some decisions are obsolete now.

PR #2187, opened 2018-06-15
[basic.start,except] Harmonize references to std::terminate

### LINKED ISSUE #2150: Calling std::terminate() vs. std::terminate (no parens)
[basic.start.term] uses the latter, [except] mostly uses the former. Decide and harmonize.

jensmaurer:
Fixes #2150.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -5512,9 +5512,10 @@ It is \impldef{threads and program points at which deferred dynamic initializati
 in which threads and at which points in the program such deferred dynamic initialization occurs.
 
 \pnum
-If the initialization of a non-local variable with static or thread storage duration
-exits via
-an exception, \tcode{std::terminate} is called\iref{except.terminate}.%
+If the initialization of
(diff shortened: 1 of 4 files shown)


source/basic.tex
```diff
@@ -5487,7 +5487,7 @@
 \pnum
 If the initialization of a non-local variable with static or thread storage duration
 exits via
-an exception, \tcode{std::terminate} is called\iref{except.terminate}.%
+an exception, \tcode{std::terminate()} is called\iref{except.terminate}.%
```
zygoloid: We have a mix of "`std::terminate()` is called" and "`std::terminate` is called" here, with some examples changing in one direction and some changing in the other direction. I think we have guidance that "the function `std::terminate` is called" should hot have parentheses, but these cases without a "the function" prefix are ambiguous between calling a function and "calling" (evaluating) a function call expression.

jensmaurer: Ok, going for "the function std::terminate is called" globally.

PR #2195, opened 2018-06-16
[re.general] Refer to table as done in the other clauses

johelegp:
This change refers to the table introducing the first-level subclauses as done in the summary of the other clauses.


tkoeppe:
Note for the future: Please describe commits in English. (We can fix things up at merge time, but I'd prefer commits to arrive in good shape.)

jwakely:
Yes, the sed-like command is just confusing with `\t` in there. No pseudo-sed in commit summaries please.

johelegp:
Thanks for the comments. I have rushed it.

PR #2200, opened 2018-06-20
[cpp.predefined] Feature-test macro for P0732R2

jensmaurer:
P0732R2 Class types in non-type template parameters


jensmaurer:
Implemented by commit de4ddf0a8fa164c7b451b6bb500c492681a7738c.

PR #2205, opened 2018-06-21
[cpp.predefined] Feature-test macro for P0892R2

jensmaurer:
P0892R2 explicit(bool)


jensmaurer:
Using __cpp_conditionally_explicit, which is better than the original suggestino __cpp_explicit_bool in the paper.

jensmaurer:
Implemented by commit de4ddf0a8fa164c7b451b6bb500c492681a7738c.

PR #2212, opened 2018-06-26
[alg.shift], [fpos.operations] Annotate "old" requirements with \oldc…

CaseyCarter:
…oncept

Fixes a few occurrences introduced by recently merged motions.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -3564,7 +3564,7 @@ template<class ExecutionPolicy, class ForwardIterator>
 \pnum
 \requires
 The type of \tcode{*first} shall satisfy
-the \tcode{MoveAssignable} requirements.
+the \oldconcept{MoveAssignable} requirements.
 
 \pnum
 \effects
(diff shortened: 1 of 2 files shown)


source/algorithms.tex
```diff
@@ -3603,10 +3603,10 @@
 \pnum
 \requires
 The type of \tcode{*first} shall satisfy
-the \tcode{MoveAssignable} requirements.
+the \oldconcept{MoveAssignable} requirements.
 \tcode{ForwardIterator} shall meet
 the requirements of a bidirectional iterator\iref{bidirectional.iterators} or
```
timsong-cpp: \oldconcept{BidirectionalIterator}?

CaseyCarter: It wouldn't hurt to change this to `shall meet the \oldconcept{BidirectionalIterator} requirements\iref{bidirectional.iterators}`, but it's certainly not the only such occurrence. We need to audit the entire library for colloquialisms and replace them with proper requirement names to avoid confusion over whether e.g. "bidirectional iterator" means something that meets the old requirements, or the new requirements, or both.

timsong-cpp: The "colloquialism" was in fact the proper way to state such requirements. "bidirectional iterator" is not a CamelCasedRequirement before you made it one :)

CaseyCarter: It was in [algorithms] =P

PR #2214, opened 2018-06-26
[macros] Make colon in Note: and Example: upright and reduce spacing around brackets

tkoeppe:
image


source/classes.tex
```diff
@@ -801,6 +801,7 @@
 \end{codeblock}
 \end{example}
 \begin{note}
+\setlength\emergencystretch{1em}
```
jensmaurer: I thought \setlength was a LaTeX macro which needs some decorations around the arguments, i.e. `\setlength{\lengthname}{value_in_specified_unit}`.

tkoeppe: "some decorators" aren't *needed* just for the sake of it -- only if the argument consists of more than one token. In this case it doesn't. But I agree that it'd be nicer to keep the braces even in this degenerate case. I'll change it.

tkoeppe: Done.

tkoeppe:
@godbyk: Hi - would you by any chance see a way how we could move the `\setlength{\emergencystretch}{1em}` into, say, the `note` environment, or even set it globally? All my attempts to move those length changes away from the lines that immediately need them have failed.

godbyk:
@tkoeppe I think the `\emergencystretch` would apply to the entire paragraph the note is a part of. We could apply it to the note (and its surrounding paragraph), put the note in its own paragraph and apply it, or apply it globally.

You can also try out the `\sloppy` and `\fussy` macros. The former relaxes TeX's typesetting standards a bit to allow for looser line breaking. The latter resets the values to TeX's default. Their default definitions are as follows:

```latex
\def\sloppy{%
  \tolerance 9999%
  \emergencystretch 3em%
  \hfuzz .5\p@
  \vfuzz\hfuzz}

\def\fussy{%
  \emergencystretch\z@
  \tolerance 200%
  \hfuzz .1\p@
  \vfuzz\hfuzz}
```

PR #2221, opened 2018-06-28
[variant.get] Consistently place comma after "otherwise"

johelegp:


johelegp:
Then there wouldn't be consistency within the sibling subclauses. Have the editors come to an agreement on when to use of the new elements when doing drive-by fixes?

CaseyCarter:
>Then there wouldn't be consistency within the sibling subclauses.

That's a good enough argument to convince me.

> Have the editors come to an agreement on when to use of the new elements when doing drive-by fixes?

Not to my knowledge.

jwakely:
> should be done as a separate paper 

That paper is on my TODO list.

johelegp:
@jwakely Will your paper's proposed changes be a 1:1 transformation, or will it include "fixes"? For example, https://wg21.link/time.duration.io#1 should have been a _Remarks:_ element as I suggest in https://wg21.link/LWG3125. The new calendar and timezone library has many of those, so I'm wondering if I should submit a DR on those and any other I find while browsing the standard.

PR #2231, opened 2018-06-29
[class.mem] Define complete-class context

### LINKED ISSUE #2107: Add cross-references to lists of delayed-parsed region
6.3.7 [basic.class.scope], p. 1:
6.4.1 [basic.lookup.unqual], p7 and p8
12.2 p6

CWG in Rapperswil: Add a defined term "complete-class context" in one place (with cross-reference) and use the term everywhere, instead of maintaining several separate lists.

jensmaurer:
and use it, instead of having separate redundant lists
in [basic.scope.class] and [basic.lookup.unqual].

Fixes #2107.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -991,12 +991,8 @@ A name with global namespace scope is said to be a
 \pnum
 The potential scope of a name declared in a class consists not
 only of the declarative region following the name's point of
-declaration, but also of all function bodies, default arguments,
-\grammarterm{noexcept-specifier}{s},
-default member initializers\iref{class.mem},
-and contract conditions\iref{dcl.attr.contract}
-in that class (including such
-things in nested classes).
+declaration, but also of all complete-class contexts\iref{class.mem}
+of that class.
 
 \pnum
 A name \tcode{N} used in a class \tcode{S} shall refer to the same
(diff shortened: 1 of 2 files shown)


source/basic.tex
```diff
@@ -1336,13 +1332,8 @@
 \end{example}
 
 \pnum
-A name used in the definition of a class \tcode{X} outside of a member
-function body, default argument, \grammarterm{noexcept-specifier},
-default member initializer\iref{class.mem},
-contract condition\iref{dcl.attr.contract},
-or nested class definition\footnote{This refers to unqualified names
-following the class name; such a name may be used in the
-\grammarterm{base-clause} or may be used in the class definition.}
```
zygoloid: Is the removal of this footnote intentional? Clarifying that this covers the base-specifiers seems at least moderately useful.

source/basic.tex
```diff
@@ -1404,11 +1395,7 @@
 
 \pnum
 For the members of a class \tcode{X}, a name used
-in a member function body,
```
zygoloid: Whoops, we meant "function body" here, not "member function body". This is technically a normative change, but it's obvious that the new wording is correct.

source/basic.tex
jensmaurer: It seemed so... obvious to me, but we seem to have several similar footnotes in the vicinity. Restored.

source/basic.tex
jensmaurer: So, we're good here, right?

zygoloid: Yes.

source/basic.tex
```diff
@@ -1336,13 +1332,11 @@
...
-default member initializer\iref{class.mem},
-contract condition\iref{dcl.attr.contract},
-or nested class definition\footnote{This refers to unqualified names
-following the class name; such a name may be used in the
-\grammarterm{base-clause} or may be used in the class definition.}
+A name used in the definition of a class \tcode{X} outside of a
+complete-class context\iref{class.mem} of \tcode{X}\footnote{This
+refers to unqualified names following the class name;
+such a name may be used in the \grammarterm{base-clause} or
+may be used in the class definition.}
```
zygoloid: This (pre-existing) footnote text is appalling :(

"A name used in the definition of a class [...] refers to [...] a name used in the *base-clause* or [...] in the class definition".

What? So "in the definition of a class" is a broader syntactic category than "in the class definition"? I'll fix that after merging.

zygoloid: 2fdaf88f

PR #2232, opened 2018-06-30
[expr.prim.id.unqual] Fix cross reference

cpplearner:


cpplearner:
BTW, since it's possible to have a qualified-id that names a bit-field, should a similar note be added to [expr.prim.id.qual]?

```c++
struct A {
    int x : 1;
    void f();
};
namespace N {
    auto [b] = A{};
}
void A::f() {
    A::x; // bit-field
    N::b; // bit-field
}
```

jensmaurer:
Regarding the original issue: I'd like to keep the cross-reference as-is to highlight that structured bindings could also refer to a bit-field. If you want to add a cross-reference to class.bit, do so on the first mention of bit-field in the sentence (a little earlier).

Regarding your second point: Agreed, but make that a second pull request, because it actually adds normative wording, which might cause more resistance.

cpplearner:
1. I think structured binding should be explicitly mentioned earlier in this sentence, and it feels redundant if the same cross reference appears twice in this sentence, i.e. 

> The expression is an lvalue if the entity is a function, variable, <ins>structured binding\iref{dcl.struct.bind}</ins>, data member, or template parameter object and a prvalue otherwise\iref{basic.lval}; it is a bit-field if the identifier designates a bit-field<del>\iref{dcl.struct.bind}</del>.

(I'll send another PR for this change.)

2. I'm not sure where to put the rule on bit-field. Maybe instead of adding it to [expr.prim.id.qual], it's better to move the rule to [expr.prim.id], to avoid repetition?

PR #2237, opened 2018-07-02
[concept.swappable] remove wrongly included text

timsong-cpp:
These paragraphs are in magenta in P0898R3 and are not meant to be included into the IS working draft.

### DIFF
--- a/source/concepts.tex
+++ b/source/concepts.tex
@@ -604,24 +604,6 @@ includes:
 \end{itemize}
 \end{itemdescr}
 
-\begin{itemdescr}
-\pnum
-This subclause provides definitions for swappable types and expressions. In
-these definitions, let \tcode{t} denote an expression of type \tcode{T}, and let
-\tcode{u} denote an expression of type \tcode{U}.
-
-\pnum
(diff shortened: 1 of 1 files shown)


CaseyCarter:
@timsong-cpp is right: this is my mistake.

tkoeppe:
Ah, that's because plain "magenta" isn't a thing: P0893R3 only defines "mangeta with strikethrough". So I suppose that's an error in the original paper already that translated into erroneous application?

CaseyCarter:
> Ah, that's because plain "magenta" isn't a thing: P0893R3 only defines "mangeta with strikethrough".

The magenta comes from the `\oldtxt` macro, which is basically `\removed` with a different color. It can't be applied to markup, and large blocks of strikethrough do nothing for clarity, so I usually just `{\color{oldclr} ...}` large blocks. The error in the paper is not describing the usage of "red/magenta without strikethrough" and "cyan/gold without underline" correctly.

Also, "Mangeta" sounds like one of Godzilla's opponents - or a Super Sayan.

tkoeppe:
"Mangeta" is clearly the spin-off following the adventures of Magneta's manservant.

_Attractive._ &mdash; Critics.

PR #2239, opened 2018-07-02
\diffref and \diffrefs now take a comma-separated list of references.

godbyk:
An improvement to #2236: The `\diffref` and `\diffrefs` macros now take a single argument, a comma-separated list of references. 

The only difference in output is that now the word _and_ is inserted prior to the penultimate reference if there are more than one. For example:

```latex
\diffref{sequence.reqmts,associative.reqmts}
```
> **Affected subclauses:** 27.2.3 and 27.2.6

```latex
\diffrefs{complex.numbers}
```
> **Affected subclause:** 30.5

```latex
\diffref{istream::sentry, ostream::sentry, iostate.flags}
```
> **Affected subclauses:** 31.7.4.1.3, 31.7.5.1.3, and 31.5.5.4

### Notes

* I can remove the word _and_ if you prefer and just emit a comma-separated list of references.
* The `\diffref` and `\diffrefs` macros are exactly equivalent and may be used interchangeably. We can also remove one if you prefer.
* Spaces around references are ignored so the comma-separated list may use spaces before and/or after the comma to improve readability.
* _subclause_ is automatically pluralized if more than one reference is provided.
* I used the `expl3` (LaTeX 3) macros for the comma-separated list processing. While there are other packages we could use, this one provides better functionality and more facilities for improving other macros in the future. I would love to be able to take advantage of `expl3`, but if you prefer to stick to LaTeX 2-based packages, I can rewrite the code to use one of them instead. (Note that using `expl3` doesn't require using a new `pdflatex` binary. It's just a set of newer macros built atop the existing engines.)


zygoloid:
How does one determine if they have LaTeX 3?

godbyk:
@zygoloid Well, if you have the `expl3.sty` file, then you should be all set. On Linux and macOS, you can run `kpsewhich expl3.sty` to do the search.

Or just try compiling the PDF. If it works, then you know. :-)

zygoloid:
Reviewers: please can you check whether this works for your LaTeX environment?

tkoeppe:
@godbyk: 1) please rebase, 2) please check readmes, travises, makefiles etc for whether lists of required packages need to be updated.

source/compatibility.tex
```diff
@@ -1151,7 +1151,7 @@
 that interacts with newer \Cpp{} code using regions declared reachable may
 have different runtime behavior.
 
-\diffrefs{refwrap}{arithmetic.operations}, \ref{comparisons},
+\diffref{refwrap,arithmetic.operations,comparisons},
 \ref{logical.operations}, \ref{bitwise.operations}
```
tkoeppe: There are additional `\ref`s here.

Please also check the other cases, I haven't looked thoroughly.

godbyk: Good catch! I didn't notice any others.

godbyk:
@tkoeppe Done and done.

I updated the packages lists in the README by checking the contents of the packages through each distro's website. I haven't actually tested the build on each distro.

source/macros.tex
```diff
@@ -337,8 +337,24 @@
 
 %% Change descriptions
 \newcommand{\diffdef}[1]{\hfill\break\textbf{#1:}\space}
-\newcommand{\diffref}[1]{\pnum\textbf{Affected subclause:} \ref{#1}}
-\newcommand{\diffrefs}[2]{\pnum\textbf{Affected subclauses:} \ref{#1}, \ref{#2}}
+\RequirePackage{expl3,xparse}
```
jensmaurer: I'd prefer this to be \usepackage items in std.tex so that the list of required packages is clearly visible where a (LaTeX 2) person would expect it.

godbyk: @jensmaurer Done!

Eelis:
Needs rebase again I'm afraid.

godbyk:
@Eelis Is it a good time to rebase or are there a bunch more merges imminent?

Eelis:
I'm not sure, sorry. @zygoloid probably knows.

tkoeppe:
Since the latest working paper candidate has just been created from these sources, now is probably a good time to rebase. There shouldn't be many large changes that are currently eligible for consideration.

jensmaurer:
@zygoloid, can we make progress here, and either take it or leave it? (I like the regularity with the change.)

PR #2240, opened 2018-07-02
[concepts.object] Dissolve single-item subclauses.

timsong-cpp:


source/concepts.tex
```diff
@@ -967,33 +960,23 @@
 
 \rSec1[concepts.object]{Object concepts}
 
-\rSec2[concepts.object.general]{General}
 \pnum
-This section describes concepts that specify the basis of the
+This subclause describes concepts that specify the basis of the
```
CaseyCarter: *sigh* I thought I caught all of these.

Issue #1511, opened 2017-03-04
Move <numeric> to the algorithms clause


AlisdairM:
Especially with the addition of parallel algorithms, it is clear that the <numeric> header has more in common, and more shared wording, with the algorithms clause than the numerics clause.  For C++20, it would be good to move the header and relevant wording into clause 28 (Algorithms Library), and rename clause 29 from Numerics Library to Numerical Types Library.

cubbimew:
+1 cppreference has always listed numeric algorithms on its [algorithms page](http://en.cppreference.com/w/cpp/algorithm)

W-E-Brown:
<Adding SG6 chair Lawrence Crowl>

> On Mar 4, 2017, at 6:15 AM, Alisdair Meredith <notifications@github.com> wrote:
> Especially with the addition of parallel algorithms, it is clear that the header has more in common, and more shared wording, with the algorithms clause than the numerics clause. For C++20, it would be good to move the header and relevant wording into clause 28 (Algorithms Library), and rename clause 29 from Numerics Library to Numerical Types Library.
> 
Let's please be sure to consult with at least SG6 when we're ready to consider such renaming.

W-E-Brown:
On 3/4/17, W Brown <webrown.cpp@gmail.com> wrote:
> <Adding SG6 chair Lawrence Crowl>
>
> On Mar 4, 2017, Alisdair Meredith <notifications@github.com> wrote:
>> Especially with the addition of parallel algorithms, it is clear that
>> the header has more in common, and more shared wording, with the
>> algorithms clause than the numerics clause. For C++20, it would be
>> good to move the header and relevant wording into clause 28
>> (Algorithms Library), and rename clause 29 from Numerics Library to
>> Numerical Types Library.
>
> Let's please be sure to consult with at least SG6 when we're ready to
> consider such renaming.

I agree with a separate clause for number types.  There are many clauses
that seems to me to be bags of stuff.  I would like to see some of them
split, but I'll leave that to folks more directly involved.

-- 
Lawrence Crowl

zygoloid:
We like the direction suggested: move <numeric> to the Algorithms clause, and rename the Numerics clause to something that better describes its contents. LWG feedback is requested on this direction and on what Clause 29 should be named after the move.

jensmaurer:
[...]

PR #2243, opened 2018-07-03
P1076R1: Editorial clause reorganization

### LINKED ISSUE #1919: [conv,class] Clause reorganization in core sections
Following up on #1771, here's a specific suggestion at reorganization:

- Move [conv] as a subclause into [expr], immediately before [expr.arith.conv].
- Move [class.derived], [class.access], and [special] as subclauses into [class]: [class.access] as a subclause of [class.mem]; the rest immediately nested into [class].
- After the above, move [class.init] immediately nested into [class]. Same for [class.temporary]. Neither have anything to do with "special member functions" per se.
- (avoid hanging paragraphs:) Rename [class.name] to "Class declarations" and move [class]p1-3,5,11 there, at the top. Introduce [class.prop] "Class properties" and move [class]p4,6-10 there.

jensmaurer:
Fixes #1919 and #1511.


zygoloid:
[class.copy] seems a bit lost and lonely. (Also, the final cross-reference in p1 seems like it should point to [class.copy.assign], not to [over.ass].) Does [class.copy] serve any purpose? Is there any normative weight in either of its paragraphs? (Can we just strike it?)

I think the resulting structure of [algorithms] could do with some more attention. Listing the <algorithm> synopsis at the top level but putting the <numeric> synopsis under "Generalized numeric operations" creates an asymmetry between them. Perhaps the <numeric> synopsis could be promoted to 23.8 or to 23.3, to match the <algorithm> synopsis in 23.2?

[numeric.ops.overview]p1 appears to be redundant now.

Otherwise, this looks great.

jensmaurer:
[numeric.ops] as been integrated better into [algorithms]: Synopsis is a separate subclause, [numeric.ops.overview]p1 is gone, [numeric.ops.overview]p2 has been moved to [algorithms.requirements] (because the existing [alg.transform] also uses template parameters named UnaryOperation and BinaryOperation). This way, it's also very obvious that these template parameters are underspecified.

zygoloid:
I've reordered the <algorithm> synopsis a bit later. We now have:

 * General
 * Algorithm requirements
 * Parallel algorithms
 * <algorithm> synopsis
 * (<algorithm> content subclauses)
 * (<algorithm> content subclauses)
 * <numeric> synopsis
 * <numeric> content subclause
 * <cstdlib> algorithms

PR #2244, opened 2018-07-04
Travis CI: Check for undefined references.

godbyk:
This adds a Travis CI check for undefined references (e.g., a `\ref` that points to a nonexistent `\label`).

This PR causes no change in the PDF output.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ env:
   - BUILD_TYPE=make      # build using Makefile
   - BUILD_TYPE=complete  # build manually and regenerate figures, grammar, and cross-references
   - BUILD_TYPE=check-whitespace  # check for whitespace at the ends of lines
-  - BUILD_TYPE=check-newlines  # check for blank lines at the ends of files
+  - BUILD_TYPE=check-newlines    # check for blank lines at the ends of files
   - BUILD_TYPE=check-macro-use   # check for proper macro use
 
 script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -62,6 +63,11 @@ script:
       ! grep '\\opt[^{}]' source/*.tex;
       ! grep 'opt{}' source/*.tex;
     fi
+  # Fail if there are undefined references
+  - if [ "$BUILD_TYPE" = "check-references" ]; then
+      docker exec -it texlive-basic bash -c "cd /$TRAVIS_REPO_SLUG/source && make -j2";
+      ! grep -Fqe 'LaTeX Warning: There were undefined references.' std.log;
```
tkoeppe: Can't we roll this into one of the existing tests? It seems awfully wasteful to run a whole new build for this.

godbyk: @tkoeppe Well, this check needs to have the PDF built first (so LaTeX has a chance to try to resolve all the references). We can add the check to one or more of the builds, but then Travis CI would report a build failure instead of just a missing reference error.

tkoeppe: I can totally live with that. *All* that matters, ever, is green-vs-notgreen.

How about putting this together with the overfull box test?

(We already build this document way too many times. You really think that once should suffice...)

godbyk: Fair enough.

godbyk: @tkoeppe Tested and appears to work okay. The only downside is that it doesn't tell you which check fails, so someone will have to do more digging to determine that. Maybe we should add a `make test` option that runs the checks locally?

tkoeppe: Couldn't a slightly cleverer combination of parentheses and `echo` solve this?

godbyk: Possibly. I'll experiment with it and see what I can come up with. I think when I tried something like that previously something (docker?) swallowed stdout. But I might be misremembering.

godbyk: On second glance, a simpler option should work. I removed the `-q` flag (which suppresses grep's output). So ostensibly the console log should show the `Overfull \hbox` messages and the `undefined references` message if they exist in `std.log`. (Assuming the output doesn't get swallowed by Travis CI.)

PR #2245, opened 2018-07-04
New placeholder macro that uses italics or slanted type depending on the context

godbyk:
This PR demonstrates an option outlined in issue #1060: a smarter `\placeholder` macro that typesets the placeholder either italics or slanted depending on whether it's code or roman text.

 * In roman text, it sets the placeholder in italics.
 * In monospaced text, it sets the placeholder in slanted type.
 * In sans serif text, it sets the placeholder in italics.

A diff of the PDFs [is available here](http://kevin.godby.org/temp/cplusplus-draft/std-diff-placeholder.pdf). A sample of pages that differ include 29, 30, 433–435, 880–890.

### DIFF
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -186,8 +186,36 @@
 \newcommand{\fakegrammarterm}[1]{\gterm{#1}}
 \newcommand{\grammarterm}[1]{\indexgram{\idxgram{#1}}\gterm{#1}}
 \newcommand{\grammartermnc}[1]{\indexgram{\idxgram{#1}}\gterm{#1\nocorr}}
-\newcommand{\placeholder}[1]{\textit{#1}}
-\newcommand{\placeholdernc}[1]{\textit{#1\nocorr}}
+\usepackage{xifthen}
+\usepackage{etoolbox}
+\makeatletter
+\let\std@phcorr\relax
+\newrobustcmd{\placeholder}[1]{%
+  \ifthenelse{\equal{\f@family}{\rmdefault}}{%
+    \textit{#1\std@phcorr}%
+  }{%
+    \ifthenelse{\equal{\f@family}{\ttdefault}}{%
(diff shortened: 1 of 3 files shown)


jensmaurer:
"All checks have failed".

godbyk:
@jensmaurer That's due to

```
Overfull \hbox (0.85309pt too wide) in paragraph at lines 3965--3971
Overfull \hbox (0.85309pt too wide) in paragraph at lines 4260--4267
Overfull \hbox (6.936pt too wide) in paragraph at lines 539--558
```

godbyk:
I rebased and quelled the overfull line warnings with the use of `\sloppy`.

zygoloid:
(In the diffs, the cyan is the new formatting and the red is the current formatting.)

Initial reaction: I think I somewhat prefer the current formatting for teletype placeholders. Use of a different typeface makes the placeholders stand out more clearly; the difference between slanted and unslanted teletype is easier to overlook, precisely because it's more visually consistent (which I think is the goal here).

Let's discuss this in person. We should also consider how to format things like `@\seebelow@` (which is unaffected by the current patch).

jensmaurer:
Editorial meeting:
- Slanted font makes the placeholders stand out less in the library; we don't want them to be confused with identifiers. No change for that.
- Consider changing "see below" etc. to roman italics (same as comments). -> new issue with samples attached
- Use slanted font for definition for "contextually converted to bool".

zygoloid:
Before:

before

After:

after

I think this confirms:

> Slanted font makes the placeholders stand out less in the library; we don't want them to be confused with identifiers. No change for that.

(The difference between the slanted and non-slanted text has become too subtle.)

AlisdairM:
Are we still planning to land some part of this PR?  Should we consider closing?  It has been (mostly) dormant for almost 5 years now.

PR #2246, opened 2018-07-04
[basic.types] Clarify that 'value representation' does not depend on …

### LINKED ISSUE #2105: [basic.types] p4: definition of "value representation" unnecessarily vague
[As discussed in CWG on 2018-06-07] Consider:

```
union U { int n; char c; };
U u = {.c = 123};
```

It is unclear in the current wording that the value representation of `u` includes all the bits that are part of the `n` member, not just the bits that are part of the `c` member, however that is the intent (and the upcoming `compare_exchange_*` wording depends on that). We should clarify that the set of bits in the value representation depends only on the type, and not on which value of that type we are considering.

jensmaurer:
…the value.

Fixes #2105.


source/basic.tex
```diff
@@ -3675,8 +3675,8 @@
 \tcode{sizeof(T)}. The
 \indextext{representation!value}%
 \defn{value representation}
-of an object is the set of bits that hold
-the value of type \tcode{T}.
+of a type is the set of bits that hold
+the value in an object of that type.
```
zygoloid: I don't like that we're losing the parallel between "object representation" and "value representation" here. Maybe we could retain both by adding something like

> The *value representation * of an object *X* of type `T` is the set of bits within the object representation of *X* corresponding to the value representation of `T`.

But I think that's still not quite capturing an important property: that the set of bits is the same for each object of the type. Perhaps we could relate the value representation of a type back to the sequence of *N* `unsigned char` objects somehow ("which bits in which of the *N* `unsigned char` objects taken up by an object of the type participate in representing the value of an object of the type" or something like that?).

jensmaurer: Ok, I've changed the wording along the suggestion. It seems safe to say that the bits of the value representation are always a subset of the bits comprising the object representation; it's unclear why that was previously highlighted for trivially-copyable types only.
@zygoloid, your love is next.

zygoloid: I believe the idea of the prior wording is that for non-trivially-copyable (and non-standard-layout) types, the implementation is allowed to allocate some storage "elsewhere" to hold members of the type -- there is no guarantee that the members of such a type are actually stored within it.

One way this can happen (theoretically) is that a virtual base class is dynamically allocated separately from the object, and the object merely holds pointers to it; another way is that some of the storage for an object can be held separately in a GC heap. However, in practice, I'm not sure either approach is actually conforming because of the [basic.life] rule permitting an object to be discarded by merely reusing its storage.

zygoloid:
+CWG for the "are non-trivially-copyable objects necessarily stored entirely within their object representation?" question.

source/basic.tex
jensmaurer: Ah, good point. Since waiting for CWG on this matter is probably in vain for the next meeting or so, I've restored the "value representation might be partially outside object representation" meaning.  This is now a pure "value representation" wording fix-up with no change in meaning, so should be fine to be applied editorially.

jensmaurer:
Removed CWG, since the question by @zygoloid is no longer asked in this pull request.

PR #2253, opened 2018-07-08
[std] Fix cross-references that should point to [class.prop]

jensmaurer:


jensmaurer:
@tkoeppe, any chance you could have a look and merge?

PR #2255, opened 2018-07-09
"cv-qualified versions of cv-unqualified scalar types" are just "scalar types"

Quuxplusone:
This seems to be cruft accidentally left over from the dueling resolutions of CWG 1746 (which added "cv-unqualified") and CWG 2094 (which added "and cv-qualified versions of these types").

Editorial.


Quuxplusone:
Rebased.

PR #2257, opened 2018-07-11
Use the Oxford comma in [class.copy.assign].

Quuxplusone:
And turn a bunch of `\tcode{const} \tcode{X}` into `\tcode{const X}` on the assumption that this is better. (I see the latter style being used elsewhere, and it's certainly more readable in the TeX source.)

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -1777,11 +1777,11 @@ The implicitly-defined copy/move constructor for a union
 \indextext{operator!move assignment|see{assignment operator, move}}%
 A user-declared \term{copy} assignment operator \tcode{X::operator=} is a
 non-static non-template member function of class \tcode{X} with exactly one
-parameter of type \tcode{X}, \tcode{X\&}, \tcode{const} \tcode{X\&},
-\tcode{volatile} \tcode{X\&} or \tcode{const} \tcode{volatile}
-\tcode{X\&}.\footnote{Because a template assignment operator or an assignment
-operator taking an rvalue reference parameter is never a copy assignment
-operator, the presence of such an assignment operator does not suppress the
+parameter of type \tcode{X}, \tcode{X\&}, \tcode{const X\&},
+\tcode{volatile X\&}, or \tcode{const volatile X\&}.\footnote{Because
+a template assignment operator or an assignment operator
+taking an rvalue reference parameter is never a copy assignment operator,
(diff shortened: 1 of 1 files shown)


source/classes.tex
```diff
@@ -1879,40 +1879,20 @@
...
-and
+\tcode{const B\&}, \tcode{const volatile B\&}, or \tcode{B}, and
 \item
-for all the non-static data members of
-\tcode{X}
-that are of a class type
-\tcode{M}
-(or array thereof),
+for all the non-static data members of \tcode{X} that are of a class type
+\tcode{M} (or array thereof),
```
jensmaurer: Since you're re-flowing the source here: We want [semantic linebreaks](http://www.sembr.org), but the best position to break seems to be before the "that" (which starts a relative clause).

Quuxplusone:
Addressed!

tkoeppe:
Looks good, but could you kindly make and attach diffpdf screenshot (in "appearance" comparison mode)?

Quuxplusone:
> could you kindly make and attach diffpdf screenshot

Not I; I haven't got diffpdf. Maybe someone not-me could.

zygoloid:
We currently have a mixture of spaces-in-`\tcode` and spaces-between-`\tcode`s formatting, and we should pick our preferred options. Example:

With `\tcode` split on spaces:

image

With spaces inside `\tcode`:

image

Clearly our full code examples are going to use the spaces-inside-`\tcode` approach, so perhaps that's better for consistency. But I think splitting `\tcode` on spaces is more beautiful.

jensmaurer:
I agree it's more beautiful to split the `\tcode`s, but it's also less grep-able and less readable in the text editor for the LaTeX source. Let's not split `\tcode`.

Quuxplusone:
FWIW, I *don't* agree that splitting tcodes is more beautiful; I think the
different width spaces are noticeable and technically distracting (is a
`const` `int&` different from a `const int&`?). So I'd prefer not to split,
for that technical reason. But I don't care very much.

jensmaurer:
@Quuxplusone, could you please rebase, thereby resolving the conflicts?

jensmaurer:
Editorial meeting: If the space is part of the technical content (such as the "const volatile X&" example above), the space is part of the \tcode. With that decision made, we merge this change.

PR #2258, opened 2018-07-12
[std] Use \defnadj where appropriate

jensmaurer:


source/basic.tex
```diff
@@ -5212,7 +5208,7 @@
 \end{note}
 
 \pnum
-A \defnx{visible side effect}{side effects!visible} \placeholder{A} on a scalar object or bit-field \placeholder{M}
+A \defnadj{visible}{side effect} \placeholder{A} on a scalar object or bit-field \placeholder{M}
```
zygoloid: We currently have many index entries for "side effects" (plural), and this one has now changed from plural to singular. I don't especially like the result, but let's address that after this is merged.

PR #2259, opened 2018-07-12
Simplify definition of "data race"

geoffromer:
... by moving details into the definitions of "conflict" and "potentially concurrent".

I believe this change is editorial (has no normative impact), but I'm going to solicit feedback from SG1 since it changes the meanings of "conflict" and "potentially concurrent" in a way that could affect usages outside the standard.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -5037,11 +5037,6 @@ and detailed visibility constraints. However, it also implicitly supports a
 simpler view for more restricted programs. \end{note}
 
 \pnum
-Two expression evaluations \defn{conflict} if one of them modifies a memory
-location\iref{intro.memory} and the other one reads or modifies the same
-memory location.
-
-\pnum
 The library defines a number of atomic operations\iref{atomics} and
 operations on mutexes\iref{thread} that are specially identified as
 synchronization operations. These operations play a special role in making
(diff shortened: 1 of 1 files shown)


source/basic.tex
```diff
@@ -5304,15 +5299,19 @@
 as described above, satisfy the resulting constraints as imposed here. \end{note}
 
 \pnum
-Two actions are \defn{potentially concurrent} if
+Two actions \defn{conflict} if one of them modifies a memory
+location\iref{intro.memory}, the other one reads or modifies the same
+memory location, and at least one of them is not atomic.
+
+\pnum
+Two actions are \defn{potentially concurrent} if they are unsequenced, and
```
timsong-cpp: [intro.execution]/8:

> If A is not sequenced before B and B is not sequenced before A, then A and B are unsequenced.

"happens before" and "unsequenced" are not mutually exclusive.

geoffromer: I know. Can you be more specific about what problem you see with the proposed wording?

timsong-cpp: If A inter-thread happens before B, then A and B are unsequenced, and therefore by this definition they are potentially concurrent. And since you removed "neither happens before the other" from the data race definition, it's now a race.

geoffromer: Ah, I see your point. Fixed.

tkoeppe:
Is this issue still current, and could you perhaps try to get it onto SG1's plate?

geoffromer:
Sorry, this was [discussed](https://lists.isocpp.org/parallel/2018/07/2093.php) in 2018, but I never circled back. To summarize for those who can't access the linked thread, the overall rationale for this proposal was to restructure the definition of "data race" so that we have one term ("conflict") that captures all the necessary conditions on the kinds of operations, a separate term ("potentially concurrent") that captures all the necessary conditions on the execution context, and the two together are sufficient conditions for a data race (other than the `sig_atomic_t` exception, which can't be split in this way).

However, some folks on SG1 felt that the definition of "conflict" should not be changed, because the existing definition matches longstanding usage outside the standard (in research papers, etc). There was some openness to using a different term such as "potentially racing" instead, so this idea still has a path forward, but I'm no longer planning to pursue it.

tkoeppe:
Thanks for that summary!

PR #2262, opened 2018-07-15
[library.general] Fixup the library categories table

CaseyCarter:
by ordering [localization] correctly, and adding a row for [time]


jensmaurer:
@tkoeppe, can you apply?

PR #2263, opened 2018-07-15
[iterators.general] Fixup Iterators library summary table

CaseyCarter:
by correcting the title for [predef.iterators] to "Iterator adaptors", and adding rows for [iterator.range] and [iterator.container].


jensmaurer:
@tkoeppe, one more.

PR #2265, opened 2018-07-17
[basic.lval,expr.delete] 'dynamic type' refers to lvalues, not objects

### LINKED ISSUE #2190: "dynamic type" is defined as a property of expressions but sometimes used as a property of objects
See the definition at [defns.dynamic.type](http://eel.is/c++draft/defns.dynamic.type).

There are 11 uses of "the dynamic type of the object".

jensmaurer:
Fixes #2190.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -306,7 +306,7 @@ The intent of this list is to specify those circumstances in which an
 object can or cannot be aliased.
 \end{footnote}
 \begin{itemize}
-\item the dynamic type of the object,
+\item \cv{} \tcode{T}
 
 \item a type that is the signed or unsigned type corresponding to the
 dynamic type of the object, or
@@ -3102,10 +3102,10 @@ called. Otherwise, its final overrider\iref{class.virtual} in the dynamic type
 of the object expression is called; such a call is referred to as a
 \defnx{virtual function call}{function!virtual function call}.
 \begin{note}
-The dynamic type is the type of the object referred to by the
-current value of the object expression. \ref{class.cdtor}~describes the
-behavior of virtual function calls when the object expression
-refers to
+The dynamic type is the type of the most derived object referred to by the
+current value of the object expression\iref{defns.dynamic.type}.
+\ref{class.cdtor}~describes the behavior of virtual function calls
+when the object expression refers to
 an object under construction or destruction.
 \end{note}
 
@@ -4028,10 +4028,10 @@ contains the original member, or is a base or derived class of the class
 containing the original member, the resulting pointer to member points
 to the original member. Otherwise, the behavior is undefined.
 \begin{note}
-Although class \tcode{B} need not contain the original member, the
-dynamic type of the object with which indirection through the pointer
-to member is performed must contain the original member;
-see~\ref{expr.mptr.oper}.
+Although class \tcode{B} need not contain the original member,
+the dynamic type of the object expression
+appearing in a pointer-to-member operation
+must contain the original member; see~\ref{expr.mptr.oper}.
 \end{note}
 
 \pnum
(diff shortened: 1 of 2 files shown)


jensmaurer:
I don't really like the edits to [expr.delete]. The issue here is that the operand of the _delete-expression_ is a pointer, but "dynamic type" is only defined for lvalues. We can either add a definition of "dynamic type" for pointers (deferring to the glvalue case), or we can locally construct an lvalue in [expr.delete]. I've done the latter, but I'm not totally happy with it. Comments welcome; in particular the [basic.lval] rewrites are rather extensive.

jensmaurer:
... and while we're at it, the definition of "dynamic type" seems to belong to [expr.prop], since it's a property of an expression.

languagelawyer (changes requested):

jensmaurer:
Editorial meeting: Maybe replace "dynamic type" simply with "most derived type of the object the glvalue / pointer refers to". Well, no, "dynamic type" is too ingrained in the history of C++. "... even if the pointer points to a base class subobject."

"assures" -> "ensures"

source/expressions.tex
```diff
@@ -281,31 +281,21 @@
 \end{note}
 
 \pnum
-If a program attempts to access the stored value of an object through a glvalue
-of other than one of the following types the behavior is
-undefined:\footnote{The intent of this list is to specify those circumstances in which an
-object may or may not be aliased.}
+If a program attempts to access the stored value of an object of type \tcode{T}
+through a glvalue whose dynamic type is not one of the following types,
```
jensmaurer: use "static type" here

source/expressions.tex
```diff
@@ -281,31 +281,21 @@
...
-\item a type that is the signed or unsigned type corresponding to the
-dynamic type of the object,
-
-\item a type that is the signed or unsigned type corresponding to a
-cv-qualified version of the dynamic type of the object,
+\item a type that is the signed or unsigned type corresponding to \cv{} \tcode{T},
 
 \item an aggregate or union type that includes one of the aforementioned types among its
 elements or non-static data members (including, recursively, an element or non-static data member of a
 subaggregate or contained union),
```
jensmaurer: This is useless in C++, except for the copy of a union. Oh, covered by the final bullet, which covers object representations = arrays of unsigned char.

source/expressions.tex
```diff
@@ -2875,10 +2865,10 @@
 of the object expression is called; such a call is referred to as a
 \defnx{virtual function call}{function!virtual function call}.
 \begin{note}
-The dynamic type is the type of the object referred to by the
-current value of the object expression. \ref{class.cdtor}~describes the
-behavior of virtual function calls when the object expression
```
jensmaurer: Simply strike the first sentence of the note instead of trying to fix it.

jensmaurer:
Editorial issue: Changes to [basic.lval] collide with CWG 2051.

tkoeppe:
@jensmaurer: Rebase?

jensmaurer:
@tkoeppe, this needs (a lot) more work than a simple rebase. Not right now.

tkoeppe:
@jensmaurer: Is there still something left to pursue here?

jensmaurer:
Yes, I think this needs a bit of love + care at some later point. There is a terminological haziness around "static type" and "dynamic type" that ought to be fixed, possibly by eradicating "static type" entirely (it's simply the type of the expression) and doing some goodness to "dynamic type". After all, "dynamic type" means "type of the (most derived) object" (unrelated to any specific expression), but "static type" means "type of the expression as written in the source and specified by [expr]".

Crazy talk such as "the static type of the object to be deleted" just needs to go; talking about the static type of an object is a category error as explained above.

PR #2271, opened 2018-07-19
[diff] Remove headings that refer to dissolved clauses

jensmaurer:


jensmaurer:
(I can't reproduce the Travis failure here. Presumably, 5 LaTeX runs were not enough.)

PR #2274, opened 2018-07-21
[dcl.array] Rework description.

### LINKED ISSUE #2175: [dcl.array]/1 This grammar does not always declare an array.
The issue is described in this [StackOverflow question](https://stackoverflow.com/q/50823601/5376789). 

[dcl.array]/1 says "then the type of the identifier of D is an array type", but that's not true. For example, `int (*p)[42];` is a pointer type but satisfies the grammar.

In addition, [dcl.array]/1 also says "T is called the array element type", which means, for example, `int a[42][42];`, the element type of `a` is `int`. I don't know if it is intended.

jensmaurer:
Group examples with the corresponding normative statements.
Clarify the meaning of 'array type'.
Use 'declarator-id' instead of 'identifier'.

Fixes #2175.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -2959,89 +2959,76 @@ There is no ``reference-to-member'' type in \Cpp{}.
 \indextext{declarator!array}
 
 \pnum
-In a declaration
-\tcode{T}
-\tcode{D}
-where
-\tcode{D}
-has the form
+In a declaration \tcode{T} \tcode{D} where \tcode{D} has the form
 
 \begin{ncsimplebnf}
 \terminal{D1} \terminal{[} \opt{constant-expression} \terminal{]} \opt{attribute-specifier-seq}
 \end{ncsimplebnf}
 
-and the type of the identifier in the declaration
-\tcode{T}
-\tcode{D1}
-is
-``\placeholder{derived-declarator-type-list}
-\tcode{T}'',
-then the type of the identifier of
-\tcode{D}
-is an array type; if the type of the identifier of \tcode{D}
-contains the \tcode{auto} \grammarterm{type-specifier},
-the program is ill-formed.
-\tcode{T}
-is called the array
-\term{element type};
-this type shall not be a reference type, \cv{}~\tcode{void},
-or a function type.
-\indextext{declaration!array}%
-If the
-\grammarterm{constant-expression}\iref{expr.const}
-is present, it shall be a converted constant
-expression of type \tcode{std::size_t} and
-its value shall be greater than zero.
-The constant expression specifies the
-\indextext{array!bound}%
+and the type of the contained \grammarterm{declarator-id}
+in the declaration \tcode{T} \tcode{D1}
+is ``\placeholder{derived-declarator-type-list} \tcode{T}'',
+the type of the \grammarterm{declarator-id} in \tcode{D} is
+``\placeholder{derived-declarator-type-list} array of \tcode{N} \tcode{T}''.
+The \grammarterm{constant-expression}
+shall be a converted constant expression of type \tcode{std::size_t}\iref{expr.const}.
 \indextext{bound, of array}%
-\term{bound}
-of (number of elements in) the array.
-If the value of the constant expression is
-\tcode{N},
-the array has
-\tcode{N}
-elements numbered
-\tcode{0}
-to
-\tcode{N-1},
-and the type of the identifier of
-\tcode{D}
-is ``\placeholder{derived-declarator-type-list} array of
-\tcode{N}
-\tcode{T}''.
-An object of array type contains a contiguously allocated non-empty set of
-\tcode{N}
-subobjects of type
-\tcode{T}.
-Except as noted below, if
-the constant expression is omitted, the type of the identifier of
-\tcode{D}
-is ``\placeholder{derived-declarator-type-list} array of unknown bound of
-\tcode{T}'',
-an incomplete object type.
-The type ``\placeholder{derived-declarator-type-list} array of
-\tcode{N}
-\tcode{T}''
-is a different type from the type
-``\placeholder{derived-declarator-type-list} array of unknown bound of
-\tcode{T}'',
-see~\ref{basic.types}.
-Any type of the form
-``\grammarterm{cv-qualifier-seq} array of
-\tcode{N}
-\tcode{T}''
-is adjusted to
-``array of
-\tcode{N}
-\grammarterm{cv-qualifier-seq}
-\tcode{T}'',
-and similarly for
-``array of unknown bound of
-\tcode{T}''.
-The optional \grammarterm{attribute-specifier-seq} appertains to the array.
-\begin{example}
+Its value \tcode{N} specifies the \defnx{array bound}{array!bound},
+i.e., the number of elements in the array;
+\tcode{N} shall be greater than zero.
 
+\pnum
+In a declaration \tcode{T} \tcode{D} where \tcode{D} has the form
+
+\begin{ncsimplebnf}
+\terminal{D1 [ ]} \opt{attribute-specifier-seq}
+\end{ncsimplebnf}
+
+and the type of the contained \grammarterm{declarator-id}
+in the declaration \tcode{T} \tcode{D1}
+is ``\placeholder{derived-declarator-type-list} \tcode{T}'',
+the type of the \grammarterm{declarator-id} in \tcode{D} is
+``\placeholder{derived-declarator-type-list} array of unknown bound of \tcode{T}'', except as specified below.
+
+\pnum
+\indextext{declaration!array}%
+A type of the form ``array of \tcode{N} \tcode{U}'' or
+``array of unknown bound of \tcode{U}'' is an \defn{array type}.
+The optional \grammarterm{attribute-specifier-seq}
+appertains to the array type.
+
+\pnum
+\tcode{U} is called the array \defn{element type};
+this type shall not be
+a placeholder type\iref{dcl.spec.auto},
+a reference type,
+a function type,
+an array of unknown bound, or
+\cv{}~\tcode{void}.
+\begin{note}
+An array can be constructed
+from one of the fundamental types (except \tcode{void}),
+from a pointer,
+from a pointer to member,
+from a class,
+from an enumeration type,
+or from an array of known bound.
+\end{note}
+\begin{example}
+\begin{codeblock}
+float fa[17], *afp[17];
+\end{codeblock}
+declares an array of \tcode{float} numbers and
+an array of pointers to \tcode{float} numbers.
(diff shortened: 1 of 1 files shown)


source/declarations.tex
```diff
@@ -2869,159 +2869,123 @@
...
-is present, it shall be a converted constant
-expression of type \tcode{std::size_t} and
-its value shall be greater than zero.
-The constant expression specifies the
-\indextext{array!bound}%
+A type ``array of \tcode{N} \tcode{T}'' or
+``array of unknown bound of \tcode{T}''
+is an \defn{array type}
+and \tcode{T} is called the array \defn{element type};
+this type shall not be a reference type, \cv{}~\tcode{void}, or a function type.
```
zygoloid: Should the `auto` restriction also be here? I don't think we intend to disallow
```
int a[3];
auto (*p)[3] = &a;
```

source/declarations.tex
```diff
@@ -2869,159 +2869,123 @@
...
+the type of the \grammarterm{declarator-id} in \tcode{D} is
+``\placeholder{derived-declarator-type-list} array of \tcode{N} \tcode{T}'',
+where \tcode{N} is the value of
+the \grammarterm{constant-expression}\iref{expr.const};
+if the \grammarterm{constant-expression} is absent,
+the type of the \grammarterm{declarator-id} in \tcode{D} is
+``\placeholder{derived-declarator-type-list} array of unknown bound of \tcode{T}'', except as specified below.
+In either case,
+the type shall not contain the \tcode{auto} \grammarterm{type-specifier}.
+The optional \grammarterm{attribute-specifier-seq} appertains to the array.
```
zygoloid: "to the array" -> "to the array type"

source/declarations.tex
```diff
@@ -2869,159 +2869,123 @@
...
+\begin{codeblock}
+float fa[17], *afp[17];
+\end{codeblock}
+declares an array of \tcode{float} numbers and
+an array of pointers to \tcode{float} numbers.
+\end{example}
+
+\pnum
+If the \grammarterm{constant-expression} is present,
+it shall be a converted constant expression of type \tcode{std::size_t}.
```
zygoloid: The relative sequencing of this and the above rule forming the type seem to suggest that the conversion to `size_t` is not part of type formation. Can we move this paragraph earlier?

source/declarations.tex
jensmaurer: The existing text says

"if the type of the identifier of \tcode{D} contains the \tcode{auto} \grammarterm{type-specifier}, the program is ill-formed."

If we move the auto restriction to the list, it seems that wouldn't help. Your example is "pointer to array of 3 T" with T = auto. This seems to be prohibited before and after.  And yes, moving the "auto" restriction to this list seems plausible.

source/declarations.tex

source/declarations.tex
jensmaurer: I'm splitting up the grammar parsing, otherwise I can't factor the constant-expression stuff properly into the sentence.

source/declarations.tex
```diff
@@ -2868,159 +2868,132 @@
...
+
+and the type of the contained \grammarterm{declarator-id}
+in the declaration \tcode{T} \tcode{D1}
+is ``\placeholder{derived-declarator-type-list} \tcode{T}'',
+the type of the \grammarterm{declarator-id} in \tcode{D} is
+``\placeholder{derived-declarator-type-list} array of unknown bound of \tcode{T}'', except as specified below.
+
+\pnum
+\indextext{declaration!array}%
+In either case, the type is an \defn{array type}.
```
zygoloid: This seems untrue. Given `int (*p)[3];`, we apply the above rule with `T` = `int`, `D` = `(*p)[3]`, and `D1` = `(*p)`, and we're correctly told that the type of `p` is "pointer to array of 3 int". Then this paragraph says that that type is an array type, which it isn't.

This bug is pre-existing, but I thought this rework was supposed to fix it. Maybe we should say that an array type is a type of the form "array of N T" or of the form "array of unknown bound of T", and the element type is T, here.

source/declarations.tex
```diff
@@ -2868,159 +2868,132 @@
...
+The optional \grammarterm{attribute-specifier-seq}
+appertains to the array type.
+
+\pnum
+\tcode{T} is called the array \defn{element type};
```
zygoloid: This is also wrong; `T` isn't the element type. Given `int *p[3];`, the element type is `int*`, but the only element type computed by this algorithm for that declaration is `int`.

source/declarations.tex
```diff
@@ -2868,159 +2868,132 @@
...
+this type shall not be
+a placeholder type\iref{dcl.spec.auto},
+a reference type,
+\cv{}~\tcode{void}, or
+a function type.
```
zygoloid: I would prefer to put *cv* `void` last in this list.

source/declarations.tex
```diff
@@ -2868,159 +2868,132 @@
...
 An array bound may also be omitted
-when the declarator is followed by an
-\grammarterm{initializer}\iref{dcl.init},
-when a declarator for a static data member is followed by a
-\grammarterm{brace-or-equal-initializer}\iref{class.mem},
-or in an explicit type conversion\iref{expr.type.conv}.
-In these cases, the bound is calculated from the number
+when an object (but not a non-static data member) of array type is initialized
+and the declarator is followed by an initializer
+(\ref{dcl.init}, \ref{class.mem}, \ref{expr.type.conv}).
```
zygoloid: Hmm. So this continues to exclude [expr.new] because in `new (int[]) {1, 2, 3}` the declarator is not (immediately) followed by the initializer, and because `new int[] {1, 2, 3}` does not match the *new-type-id* grammar? (This is getting revised soon, but we need to preserve the existing behavior for now.)

source/declarations.tex
```diff
@@ -2868,159 +2868,132 @@
...
+\cv{}~\tcode{void}, or
+a function type.
+\begin{note}
+An array can be constructed from one of the fundamental types
+(except \tcode{void}),
+from a pointer,
+from a pointer to member,
+from a class,
+from an enumeration type,
+or from another array.
```
zygoloid: Can we put the "array element must be of known bound" rule here, and change the "multidimensional array" wording below to a note?

source/declarations.tex
jensmaurer: That does not seem to be accurate. T=`int *`; this was already computed by [dcl.ptr] before we even got here. Note that the initial parsing stuff gets us "array of N T" (or "array of unknown bound of T", so assuming this is a correct decomposition (I believe it is), T is actually `int *`.

source/declarations.tex

source/declarations.tex
jensmaurer: Let's make that explicit.

source/declarations.tex
jensmaurer: This is a note here, so a normative constraint must be added further up. But yes.

source/declarations.tex

source/declarations.tex

source/declarations.tex
zygoloid: Oh right, I see it now. This is confusing because the `T` here is a different `T` from that used in paragraph 1 and 2 (and indeed throughout the rest of [dcl.meaning]). In those other places, `T` is the type determined from the *type-specifier*s in the *decl-specifier-seq*, but here it's the eventual type following the ''array of N'' -- that is, it's the `T` introduced in p3, which is just shadowing the `T` from p1 and p2 and is in general a different type.

I think this would be much clearer if you replaced `T` with (say) `U` in p3 onwards.

source/declarations.tex
```diff
@@ -2957,159 +2957,135 @@
...
-\tcode{N}
-\tcode{T}''
-is a different type from the type
-``\placeholder{derived-declarator-type-list} array of unknown bound of
-\tcode{T}'',
-see~\ref{basic.types}.
+Its value \tcode{N} specifies the \defnx{array bound}{array!bound},
+i.e., the number of elements in the array;
+\tcode{N} shall be greater than zero.
+An object of such a type contains
```
zygoloid: This sentence "An object of" should be moved to later on, and should refer to an object of type ''array of `N` `U`'' (see below for `U` versus `T`). (In this paragraph we're just performing the "compute a type string" algorithm and don't actually have a type that will be used as the type of an object yet.)

source/declarations.tex
```diff
@@ -2957,159 +2957,135 @@
...
-``array of
-\tcode{N}
-\grammarterm{cv-qualifier-seq}
-\tcode{T}'',
-and similarly for
-``array of unknown bound of
-\tcode{T}''.
-The optional \grammarterm{attribute-specifier-seq} appertains to the array.
+``array of \tcode{N} \grammarterm{cv-qualifier-seq} \tcode{T}'',
+and similarly for ``array of unknown bound of T''.
```
zygoloid: Missing code font for T here.

source/declarations.tex
jensmaurer: Sorry, I still don't get it. In p1 and p2, we compute "(mumble) array of N T", so the T from p1 and p2 is actually the array element type. I do agree that p1 and p2 may not actually describe an array type, but something like a pointer to an array type, but that doesn't change the fact that T is the element type of the array part of the type.
Consider p1: "derived-declarator-type-list array of N T”. 
and p2: "derived-declarator-type-list array of unknown bound of T".

So, re-use of the T seems to be totally fine to me.

source/declarations.tex
jensmaurer: Moved later, kept T for now.

source/declarations.tex
cpplearner: I think the problem is that p1 and p2 only computes the type in the declaration of the form `T D1[n]`.

For `int *p[3]`, the procedure works by first computing the type in `int p[3]`, then adjusting the type by replacing final "int" with "pointer to int". The latter is specified by the first sentence in [dcl.ptr]/1.

jensmaurer: Ah, I think I see it now.  p1+p2 would never see "T=pointer to int" for your example, but the statements in p3+ are valid for any kind of array, not just for the T's we've discovered in p1+p2.

jensmaurer: Renamed T to U in p3+.

source/declarations.tex
jensmaurer: Renamed T to U.

k-satoda:
@jensmaurer Could you please take another look on PR #2327?

#2327 was closed as its change was subsumed by this PR. But I can't see
the change in #2327 (for [expr.new]) in this PR (for [dcl.array] only).

jensmaurer:
@k-satoda, I've reopened #2327. It needs a rebase, though.

PR #2278, opened 2018-07-24
[containers,utilities] Mark exposition-only names

### LINKED ISSUE #2014: unify capitalization of node_handle and INSERT_RETURN_TYPE
We render the placeholder type *`INSERT_RETURN_TYPE`* in uppercase italics but render the placeholder type *`node_handle`* in lowercase italics. Likewise, we render the exposition-only name *`signal-handler`* in lowercase italics but *`INVOKE`* in uppercase italics.

We should establish a convention here, and if there is a reason for the difference in capitalization, it should be documented somewhere.

jensmaurer:
with italics teletype and use hyphens, not underscores,
to highlight that these are not standard-prescribed names.

Fixes #2014.

There is also a drive-by that adjusts "type" to "template".

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -1729,14 +1729,14 @@ template<class F, class Tuple>
 Given the exposition-only function:
 \begin{codeblock}
 template<class F, class Tuple, size_t... I>
-constexpr decltype(auto)
-    apply_impl(F&& f, Tuple&& t, index_sequence<I...>) {                // exposition only
+constexpr decltype(auto) @\placeholdernc{apply-impl}@(F&& f, Tuple&& t, index_sequence<I...>) {
+                                                                        // \expos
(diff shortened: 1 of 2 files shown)


tkoeppe:
@jwakely Could you kindly review this?

source/utilities.tex
```diff
@@ -1730,13 +1730,13 @@
 \begin{codeblock}
 template<class F, class Tuple, size_t... I>
 constexpr decltype(auto)
-    apply_impl(F&& f, Tuple&& t, index_sequence<I...>) {                // exposition only
+    @\placeholdernc{apply-impl}@(F&& f, Tuple&& t, index_sequence<I...>) {                // \expos
   return @\placeholdernc{INVOKE}@(std::forward<F>(f), std::get<I>(std::forward<Tuple>(t))...);  // see \ref{func.require}
```
jwakely: I find the indentation here visually confusing. That's pre-existing, but changing to italics seems to exacerbate it.

PR #2285, opened 2018-08-02
[fs.filesystem.syn] Fixes related to chrono::file_clock usage in <filesystem>

iaanus:
The synopsis of <filesystem> refers to the types chrono::time_point and chrono::file_clock, so it must include <chrono> for their declarations
The removed paragraph refers to trivial-clock that has now been replaced by chrono::file_clock. The paragraph is now unnecessary since chrono::file_clock is fully described in [time.clock.file].
See also http://cplusplus.github.io/LWG/lwg-active.html#3145

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -10797,6 +10797,8 @@ subclause are assumed to be qualified with \tcode{::std::filesystem::}.
 \indexhdr{filesystem}%
 
 \begin{codeblock}
+#include <chrono>  // see \ref{[time.syn}
+
 namespace std::filesystem {
   // \ref{fs.class.path}, paths
   class path;
@@ -11009,9 +11011,6 @@ namespace std::filesystem {
 \end{codeblock}
 
 \pnum
-\tcode{\textit{trivial-clock}} is an \impldef{type of filesystem trivial clock} type
-that satisfies the \oldconcept{TrivialClock} requirements\iref{time.clock.req}
-and that is capable of representing and measuring file time values.
 Implementations should ensure that the resolution and range of
 \tcode{file_time_type} reflect the operating system dependent resolution and range
 of file time values.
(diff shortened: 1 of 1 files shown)


jwakely:
Requiring `<chrono>` is a normative change, not editorial. It's not necessary (implementations would be required to ensure the necessary types are available, but that doesn't mean the whole of `<chrono>` has to be).

iaanus:
Would the rest of the proposed change qualify as editorial if I remove the inclusion of <chrono>?

jwakely:
I'm not sure if the rest is editorial even. Removing redundant wording that applies to an unspecified type that is no longer used is editorial, but I wonder if the wording about reflecting the OS-dependent resolution and range of file time values should be present on `file_clock` instead. If we remove that wording then nothing would be left to suggest that `file_clock` should be able to represent file timestamps.

jensmaurer:
In [time.syn], there's a

> template<class Duration>
>   using file_time = time_point<file_clock, Duration>;

which looks pretty close to `file_time_type` defined in [fs.filesystem.syn].
I also notice that the suggested change leaves resolution/range wording in, applying to file_time_type, not to file_clock. Which probably is appropriate, since we then take the clock's duration type.
In short, the suggestion change of removing discussion of _trivial-clock_ seems correct and editorial to me.

jensmaurer:
It would be good to add "that is capable of representing and measuring file time values." to the definition of file_clock in [time.clock.file.overview] p1, though.

zygoloid:
The old type was required to be "capable of representing and measuring file time values". The new type "should [have sufficient] resolution and range [to] reflect the operating system dependent resolution and range of file time values", which seems like a strictly weaker requirement -- and appears to be an intentional weakening.

(I would note that there are some Linux file systems that represent 128-bit file times, and we probably want to permit a 64-bit file time implementation regardless, so the existing "should" wording may well be the right thing. But whether it's right or not is up to LWG.)

PR #2291, opened 2018-08-08
[temp.param] fix spaceship example

timsong-cpp:
A two parameter spaceship can't be a member, and the parameter types for a defaulted operator can only be const C& ([class.compare.default]p1).


source/templates.tex
```diff
@@ -417,7 +417,7 @@
 \begin{example}
```
tkoeppe: @timsong-cpp: would you mind removing this stray newline while you're here?

PR #2298, opened 2018-08-12
[dcl.init] Rephrase "user-defined conversion sequence"

cpplearner:
1. The conversion sequence is "governed by" the rules defined here ([\[over.best.ics\]](http://eel.is/c++draft/over.best.ics#1.sentence-2)). Saying "conversion sequence" here will cause circular definition.
2. Overload resolution selects a function, not a conversion sequence.

Also perform a minor reflow to keep "overload" and "resolution" in the same line.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -4328,7 +4328,7 @@ If no constructor applies, or the overload resolution is
 ambiguous, the initialization is ill-formed.
 \item
 Otherwise (i.e., for the remaining copy-initialization cases),
-user-defined conversion functions and converting constructors that can convert from the
+user-defined conversions that can convert from the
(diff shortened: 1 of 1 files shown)


source/declarations.tex
```diff
@@ -4328,10 +4328,11 @@
 ambiguous, the initialization is ill-formed.
 \item
 Otherwise (i.e., for the remaining copy-initialization cases),
-user-defined conversion sequences that can convert from the
+user-defined conversion functions and converting constructors that can convert from the
```
zygoloid: The usual phrase to use here is simply "user-defined conversions", defined in [class.conv]. That's also what [over.match.copy] talks about.

PR #2312, opened 2018-08-30
[array.zero] paragraph 2 has confusing mathmatical notation using C++…

EzoeRyou:
… syntax.

[array.zero] paragraph 2 says

> In the case that N == 0, begin() == end() == unique value.

The usage of double equals indicates this is a C++ expression.

But in C++, "a == b == c" means "(a==b) == c" which is "bool value == c".
Besides, it doesn't say anything about the result of that expression at all.

I think the intent of this wording is that, the return value of begin() and end() are the same unique value.


jwakely:
This wording is being changed anyway for https://wg21.link/lwg2157

EzoeRyou:
Oh I didn't know that.

PR #2313, opened 2018-09-01
[basic.life,etc] Fix imbalanced spacing around parens

akinomyoga:


tkoeppe:
I believe stylistic variations in the core language specification are deliberate, since the standard does not want to suggest that whitespace is in any sense relevant or that a specific style is prescribed. (The library clauses are different and generally strive for consistency, though nobody really wants to touch locales.)

PR #2316, opened 2018-09-02
[basic.lval] Clarify result object for prvalues as operands

### LINKED ISSUE #2300: Contradiction on result objects of prvalues
I find this part of the C++ standard draft conflicting, [section [basic.lval], § 5](http://eel.is/c++draft/basic.lval#5):

> The result object of a prvalue is the object initialized by the prvalue; a prvalue that is used to compute the value of an operand of an operator or that has type cv void has no result object. [Note: Except when the prvalue is the operand of a decltype-specifier, a prvalue of class or array type always has a result object. For a discarded prvalue, a temporary object is materialized; see 7.2. —end note]

What about a prvalue of class or array type being used as the operand of an operator? The normative sentence of the paragraph seems to state that it will have no result object. But the note states the opposite. So there is a conflict, unless a class or array prvalue cannot be the operand of an operator (except of a decltype-specifier).

[Section [basic.lval], § 7](http://eel.is/c++draft/basic.lval#7):

> Whenever a prvalue appears as an operand of an operator that expects a glvalue for that operand, the temporary materialization conversion (7.3.4) is applied to convert the expression to an xvalue.

and [section [class.temporary], § 2](http://eel.is/c++draft/class.temporary#2):

> [ Note: Temporary objects are materialized:
> - when binding a reference to a prvalue ([dcl.init.ref], [expr.type.conv], [expr.dynamic.cast], [expr.static.cast], [expr.const.cast], [expr.cast]),
> - when performing member access on a class prvalue ([expr.ref], [expr.mptr.oper]),
> - when performing an array-to-pointer conversion or subscripting on an array prvalue ([conv.array], [expr.sub]),
> - when initializing an object of type std​::​initializer_­list<T> from a braced-init-list ([dcl.init.list]),
> - for certain unevaluated operands ([expr.typeid], [expr.sizeof]), and
> - when a prvalue appears as a discarded-value expression ([expr.prop]).
> — end note ]

seem to head in that direction by stating that a class or array prvalue can never be an operand of a _built-in_ operator because of prvalue-to-xlvalue conversion. But this is not true for an overloaded operator, as it can directly take a prvalue operand (no prior prvalue-to-xlvalue conversion thanks to guaranteed copy elision):

```
struct A {
  A operator+(A);
};

A a = A() + A();
```

So shouldn’t the part be reformulated as following? (This has been suggested to me by Richard Smith.)

> The result object of a prvalue is the object initialized by the prvalue; a prvalue that is used to compute the value of an operand of a **built-in** operator or that has type cv void has no result object. [Note: Except when the prvalue is the operand of a decltype-specifier, a prvalue of class or array type always has a result object. For a discarded prvalue, a temporary object is materialized; see 7.2. —end note]

jensmaurer:
Fixes #2300.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -169,14 +169,14 @@ A prvalue whose result is the value \placeholder{V}
 is sometimes said to have or name the value \placeholder{V}.
 The \defn{result object} of a prvalue
 is the object initialized by the prvalue;
-a prvalue
-that is used to compute the value of an operand of an operator or
+a non-discarded prvalue
+that is used to compute the value of an operand of a built-in operator or
 that has type \cv{}~\tcode{void}
 has no result object.
 \begin{note}
 Except when the prvalue is the operand of a \grammarterm{decltype-specifier},
 a prvalue of class or array type always has a result object.
-For a discarded prvalue, a temporary object is materialized; see \ref{expr.prop}.
+For a discarded prvalue, a temporary object is materialized; see \ref{expr.context}.
 \end{note}
 The \defnx{result}{result!glvalue} of a glvalue is the entity denoted by the expression.
 


jensmaurer:
For the built-in comma operator, a discarded prvalue could still be of class type and thus have a target object for the materialization.

maggyero:
@jensmaurer @zygoloid By the way, shouldn’t the current definition of prvalues in the C++ standard, [section [basic.lval], § 1.2](http://eel.is/c++draft/basic.lval#1.2):

> A prvalue is an expression whose evaluation initializes an object or a bit-field, or computes the value of the operand of an operator, as specified by the context in which it appears

be updated accordingly to:

> A prvalue is an expression whose evaluation initializes an object or a bit-field, or computes the **non-discarded** value of the operand of **a built-in** operator, as specified by the context in which it appears

since, as I understand it, initialization of an object or bit-field occurs for discarded prvalues or non built-in operators?

maggyero:
@jensmaurer @zygoloid 
I have created a pull request for the article correction (which is minor): https://github.com/cplusplus/draft/pull/2359
For the _cv_ `void` correction, I am waiting for your feedback.

PR #2318, opened 2018-09-02
[expr.prim.this] Do not mention 'declarator';

### LINKED ISSUE #2306: [expr.prim.this]/2 alludes to the grammar production `declarator` without necessity
[[expr.prim.this]/2](http://eel.is/c++draft/expr.prim#this-2):

> If a declaration declares a member function or member function template of a class X, the expression this  is a prvalue of type “pointer to cv-qualifier-seq X” between the optional cv-qualifier-seq and the end of the function-definition, member-declarator, or **declarator**.  . . .

A _member-declarator_ is already [defined](http://eel.is/c++draft/class.mem#nt:member-declarator) in terms of a _declarator_. Therefore, the declarator can be eliminated above without loss of generality.

jensmaurer:
'member-declarator' suffices.

Fixes #2306.


jensmaurer:
Explicit instantiations that appear out-of-class use _declarator_.

Issue #2310, opened 2018-08-29
[intro.object]: the "subobject" term should be transitive


languagelawyer:
**Also, the term "direct subobject" should be defined and/or [except.ctor]/3 should be fixed.**

Richard Smith's answer to 'Is "subobject" a transitive relation?' question:
> I think it depends on the context of the use, sadly. Strictly, subobject is
not transitive, and an object is only a subobject in at most one way (it
cannot be both an array element and a base class subobject, for instance),
notwithstanding that a virtual base class subobject can be a subobject of
multiple objects at once. Most uses of the term seem to use it in that way.
>
> However, there are some uses that clearly intend to take the transitive
closure:
> * [basic.stc.dynamic.safety]/2.2 intends to allow transitive subobjects
(but should probably allow all objects nested within the original object)
> * [basic.types]/3's example says "subobject" but presumably means
something more general
> * [basic.type.qualifier]/1.1, /1.2, /1.3 have ambiguous antecedents and
could be reworded to make it clearer that they are transitive ("of such an
object" -> "of a const object").
> * [basic.start.static]/2 intends to allow transitive subobjects
> * [conv.ptr]/3 intends to allow anything in the transitive closure of
"base class subobject of".
> ... and so on. (I'm sure you can find more.)
>
> Feel like filing an editorial issue? We should aim to use the term
consistently.

From me:
[except.ctor]/3:
If the initialization or destruction of an object other than by delegating constructor is terminated by an exception, the destructor is invoked for each of the object's direct subobjects…
(Implies that subobjects can be indirect) 

If the 'direct subobjects' here are '(immediate) base class subobjects' or 'member subobjects', what does it mean for members of an array type? How can a destructor be invoked for them? (**Should I create another issue for this?**)

jensmaurer:
Before we go on a voyage to fix the wording, we need a good phrasing for the various cases.
It seems that [intro.object] p2 is pretty unambiguous that its definition of "subobject" is non-transitive.
There is some low-hanging fruit here that just requires minor tweaks (e.g. [basic.type.qualifier] p1 bullets).

[...]

PR #2323, opened 2018-09-05
[basic.type.qualifier] Clarify the definitions of const/volatile object

jensmaurer:
Partially addresses #2310.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -4587,14 +4587,13 @@ specified in the \grammarterm{decl-specifier-seq}\iref{dcl.spec},
 \grammarterm{new-type-id}\iref{expr.new} when the object is created.
 \begin{itemize}
 \item A \defnadj{const}{object} is an object of type \tcode{const T} or a
-  non-mutable subobject of such an object.
+  non-mutable subobject of a const object.
 
 \item A \defnadj{volatile}{object} is an object of type
-  \tcode{volatile T}, a subobject of such an object, or a mutable
-  subobject of a const volatile object.
+  \tcode{volatile T} or a subobject of a volatile object.
 
 \item A \defnadj{const volatile}{object} is an object of type
-  \tcode{const volatile T}, a non-mutable subobject of such an object,
+  \tcode{const volatile T}, a non-mutable subobject of a const volatile object,
   a const subobject of a volatile object, or a non-mutable volatile
   subobject of a const object.
 \end{itemize}


source/basic.tex
```diff
@@ -4589,14 +4589,14 @@
 \grammarterm{new-type-id}\iref{expr.new} when the object is created.
 \begin{itemize}
 \item A \defnadj{const}{object} is an object of type \tcode{const T} or a
-  non-mutable subobject of such an object.
+  non-mutable subobject of a const object.
 
 \item A \defnadj{volatile}{object} is an object of type
-  \tcode{volatile T}, a subobject of such an object, or a mutable
+  \tcode{volatile T}, a subobject of a volatile object, or a mutable
   subobject of a const volatile object.
```
zygoloid: The "or a mutable subobject of a const volatile object" part is redundant, isn't it? A const volatile object is a volatile object, so such an object would already qualify as "a subobject of a volatile object".

zygoloid:
Rebasing in the github UI failed somehow, can you rebase manually?

source/basic.tex
jensmaurer: @zygoloid, this logic seems somehow flawed: You seem to argue that any subobject of a const volatile object is a volatile object (regardless of "mutable"), simply because a const volatile object is-a volatile object. This seems obviously wrong, because it drops const-ness. I think the issue is the premise: A const volatile object is not a volatile object, it's something else.

jensmaurer:
Rebased, but I don't know why we now claim to have conflicts with the editor's report.

source/basic.tex
zygoloid: I think what I described is the consequence of the current wording. An object of type `const volatile int` is a const object because it's of type `const T`, where `T` is `volatile int`, and it's a volatile object because it's of type `volatile T` where `T` is `const int`.

If we want to define "const volatile object" such that a const object is not a const volatile object and a volatile object is not a const volatile object, then, in addition to fixing this wording to clarify that `T` is cv-unqualified, we need to fix most mentions of "const object" in the standard to say "const object or const volatile object" (and similarly for the handful of mentions of "volatile object"), and should replace "non-volatile const object" with "(non-volatile) const object" since the non-volatile part is implied. I'd be OK with doing that, but it's not my preference.

My preference is to make the terms match normal English usage for an adjective: that is, a const volatile object is a kind of const object and is a kind of volatile object (and all of these are kinds of object, of course). That's also how almost all the rest of the standard currently uses the terms, as far as I can see. (In fact, "const volatile object" only appears twice, and in both times it's in a list of "const object, volatile object, or const volatile object", which could be rephrased as "const and/or volatile object" or similar.)

jensmaurer: @zygoloid: Ok, but that means more surgery to these words, beyond the antecedent clarification requested by Hubert.

PR #2326, opened 2018-09-06
[unique.ptr.single.ctor] Simplify description of unique_ptr constructors

### LINKED ISSUE #1530: [unique.ptr.single.ctor] Sloppy wording after application of LWG 2905
After LWG 2905 (part of Kona 2017 LWG motion 20) is applied, we will have something along the lines of 

> If `D` is an lvalue reference type `A&`, then the signatures are:
>
>     unique_ptr(pointer p, A& d) noexcept;
>     unique_ptr(pointer p, A&& d) = delete;

The use of _signature_ is incorrect, as whether a function is defined as deleted or not is not part of the signature. See [defns.signature.member].

The paragraph dealing with the case where `D` is `const A&` will also have the same problem.

jensmaurer:
Fixes #1530.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -8216,39 +8216,11 @@ then the program is ill-formed.
 
 \indexlibrary{\idxcode{unique_ptr}!constructor}%
 \begin{itemdecl}
-unique_ptr(pointer p, @\seebelow@ d1) noexcept;
-unique_ptr(pointer p, @\seebelow@ d2) noexcept;
+unique_ptr(pointer p, const D& d) noexcept;
+unique_ptr(pointer p, remove_reference_t<D>&& d) noexcept;
 \end{itemdecl}
 
 \begin{itemdescr}
-\pnum
-The signature of these constructors depends upon whether \tcode{D}
-is a reference type. If \tcode{D} is a non-reference type
-\tcode{A}, then the signatures are:
-
-\begin{codeblock}
-unique_ptr(pointer p, const A& d) noexcept;
-unique_ptr(pointer p, A&& d) noexcept;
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -8263,7 +8235,9 @@
 from \tcode{std::forward<decltype(d)>(d)}.
 
 \pnum
-\remarks These constructors shall not participate in overload resolution
+\remarks If \tcode{D} is a reference type,
+the second constructor is defined as deleted.
+These constructors shall not participate in overload resolution
 unless \tcode{is_constructible_v<D, decltype(d)>} is \tcode{true}.
```
jwakely: There's a pre-existing problem here. If `D` is a reference type `A&` then `is_constructible<D, decltype(d)>` is `is_constructible<A&, A&&>` which is false, so it's irrelevant that it's deleted.

Do we want to tweak the `is_constructible` constraint so that the deleted constructor participates in overload resolution for both `A&` and `const A&` cases, so that trying to initialize a deleter of reference type from an rvalue always calls the deleted function?

jwakely: I think we could just change the _Remarks:_ to only apply to non-reference deleters. The constraint never removes the first constructor from overload resolution for a reference deleter, because `is_constructible<A&, A&>` and `is_constructible<const A&, const A&>` are true. For the second constructor we don't want to remove it from overload resolution for references deleters, so that the deleted function gets chosen.

So for both constructors, the _Remarks:_ isn't useful.

jwakely: I'll take that question to LWG. This patch doesn't introduce the problem, and does resolve #1530

PR #2327, opened 2018-09-08
[expr.new] Harmonize rules of constant array bounds > 0

k-satoda:
While "shall evaluate to a strictly positive value" is not wrong,
"shall be greater than zero" is easier and is used in [dcl.array] and
[dcl.init.aggr] to describe the same rule.

As a side note, I was surprized that negative bounds are not actually
banned by these "shall be greater than zero". It is now banned by a
narrowing conversion involved in a "converted constant expression of
type std::size_t" since approval of N3323(https://github.com/cplusplus/draft/commit/390dd7193de3d03e58e46d56216ff095f910f758#diff-baa06f0d56d1de4d541e8b3fca69df74).
Now these "shall be greater than zero" actually ban only zero.
But I think changing them to "shall not be zero" is not good (can be
more confusing).

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -4791,7 +4791,7 @@ to the associated array type.
 Every \grammarterm{constant-expression} in a
 \grammarterm{noptr-new-declarator} shall be a converted constant
 expression\iref{expr.const} of type \tcode{std::size_t} and
-shall evaluate to a strictly positive value.
+its value shall be greater than zero.
 \begin{example}
 Given the definition \tcode{int n = 42},
 \tcode{new float[n][5]} is well-formed (because \tcode{n} is the


jensmaurer:
I agree with the change, but I disagree with the side note. The previous wording used "integral constant expression", which could have been of a signed integer type and therefore would have allowed for negative values. The "converted constant expression of type std::size_t" indeed relies on the absence of narrowing, and I believe the wording (before and after) correctly requires a diagnostic for "new int[-1]", for instance.

k-satoda:
I don't see in what point you disagree. To clarify what I said, let me
add more context how I came to these wordings.

My naive reading was "Huh? Evaluating -1 as std::size_t will yields a
value greater than zero, and the rule is broken?" I realized that was
wrong after testing the following:
```cpp
struct S { constexpr operator int () const { return -1; } };
int a[S()];
```
gcc 8.2.0 says "error: size of array 'a' is negative". Hmm.
https://wandbox.org/permlink/fjNFyWdK5kOJ9Yfw
gcc HEAD 9.0.0 201809 says "error: narrowing conversion of '-1' from 'int' to 'long unsigned int'".
https://wandbox.org/permlink/YbonMCznxVuqThV0
Then I found how "converted constant expression of type T" is defined
and realized that gcc HEAD is more precise.

jensmaurer:
The spelling of error messages is not in scope for the standard.
But I'm now seeing where you're coming from: Since application of [N3323](http://wg21.link/n3323), the effect of the phrasing "shall be greater than zero" (previously necessary due to "integral constant expression") is now equivalent to "not equal to zero", and the rest of the cases is covered by the narrowing prohibition in "converted constant expression of type size_t". That said, I do agree that "greater than zero" should stay.

zygoloid:
Putting this on hold: I'd like to either resolve the more-substantial #2274 first, or merge this into that.

jensmaurer:
This change is subsumed by #2274.

tkoeppe:
Let's close this and concentrate on #2274 then.

jensmaurer:
After all, this is not covered by #2274.

k-satoda:
Rebased.
Also I updated the log message since the occurrence of
"shall be greater than zero" in [dcl.init.aggr], which was mentioned in
the initial version of this PR, had been removed in commit
5ef7f9cbb248402f6c74f01acba87292efb0b561.

Issue #1498, opened 2017-03-02
Revise problematic stable labels


tkoeppe:
Some of our stable labels are obnoxious (`string.op+`), debatable (`reverse.iter.op!=`), downright ridiculous (`istreambuf.iterator::op*`), or simply comical (`move.iter.op.+`).

We should review all our stable labels and change those that use unnecessarily special characters. Changing labels requires informing @Dani-Hub of the changes, so we should try and do all the work in one go.

We ought to be able to get a list of all labels from the "Cross references" glossary, but beware that certain special characters may have broken the indexing and thus don't even show up; check the glossary log separately.

jensmaurer:
string.op+ is gone already, but needs fixing for string.op+=.
reverse.iter.* should have its sub-subsections removed (there are >5 subsections on a single page).
istreambuf.iterator::op* is already fixed.
move.iter.* should have its sub-subsections removed (there are >5 subsections on a single page).

godbyk:
Here is the list of labels containing characters other than letters, numbers, underscore, period, or hyphen:

```
back.insert.iter.op*
back.insert.iter.op++
back.insert.iter.op=
front.insert.iter.op*
front.insert.iter.op++
front.insert.iter.op=
insert.iter.op*
insert.iter.op++
insert.iter.op=
ios::Init
ios::failure
ios::fmtflags
ios::iostate
ios::openmode
ios::seekdir
istream::sentry
move.iter.op.+
move.iter.op.+=
move.iter.op.-=
move.iter.op=
ostream::sentry
reverse.iter.op!=
reverse.iter.op+
reverse.iter.op++
reverse.iter.op+=
reverse.iter.op-=
reverse.iter.op<
reverse.iter.op<=
reverse.iter.op=
reverse.iter.op==
reverse.iter.op>
reverse.iter.op>=
string.op!=
string.op+
string.op+=
string.op<
string.op<=
string.op>
string.op>=
string.operator==
```

Once we've fixed the labels, we could add a CI test to check for these issues in the future.

godbyk:
Here is the list of labels containing characters other than letters, numbers, underscore, period, or hyphen:

```
back.insert.iter.op*
back.insert.iter.op++
back.insert.iter.op=
front.insert.iter.op*
front.insert.iter.op++
front.insert.iter.op=
insert.iter.op*
insert.iter.op++
insert.iter.op=
ios::Init
ios::failure
ios::fmtflags
ios::iostate
ios::openmode
[...]

PR #2328, opened 2018-09-08
[string.classes] Avoid special characters in stable labels

jensmaurer:


jensmaurer:
Partially addresses #1498.

jensmaurer:
Rebased.

jensmaurer:
Rebased. Note that I've also renamed [string.comparison] to [string.cmp] (see "Style Guidelines" wiki page).

PR #2329, opened 2018-09-10
Non-type paramaters can have class type

Kannen:


zygoloid:
@jensmaurer I addressed your comments when merging this.

PR #2333, opened 2018-09-15
Fix lambda-declarator

kazatsuyu:
The following code is well-formed.
```
[] () constexpr [[using std:]] noexcept {};
```
But the following code is currently ill-formed.
```
[] () [[using std:]] noexcept {};
```
I think it should be well-formed.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -1458,7 +1458,7 @@ by the \grammarterm{nested-name-specifier}.
 
 \begin{bnf}
 \nontermdef{lambda-declarator}\br
-    \terminal{(} parameter-declaration-clause \terminal{)} \opt{decl-specifier-seq}\br
+    \terminal{(} parameter-declaration-clause \terminal{)} \opt{decl-specifier-seq} \opt{attribute-specifier-seq}\br
     \bnfindent\opt{noexcept-specifier} \opt{attribute-specifier-seq} \opt{trailing-return-type}
 \end{bnf}
 


zygoloid:
This is wrong: attributes go after the noexcept-specifier, not before.

However, there are two bugs here:

1) `[] () constexpr [[using std:]] noexcept {};` should be ill-formed, but we don't prevent the `constexpr` *decl-specifier* from having attributes.
2) `[] () constexpr [[]] {};` is ambiguous, because there are two different *attribute-specifier-seq*s that could consume the attributes.

The fix would be to stop reusing *decl-specifier-seq* here, but this doesn't seem editorial to me.

zygoloid:
More bugs caused by the reuse of *decl-specifier-seq*:

 * [dcl.type]/3 says "Except in a declaration of a constructor, destructor, or conversion function, at least one *defining-type-specifier* that is not a *cv-qualifier* shall appear in a complete *type-specifier-seq* or a complete *decl-specifier-seq*.", which renders all `mutable` or `constexpr` *lambda-expression*s ill-formed.
 * Throughout [dcl.spec], wording assumes that a *decl-specifier-seq* has a corresponding *init-declarator-list* or *member-declarator-list*.

jensmaurer:
I think we simply should stop using _decl-specifier_seq_ for lambdas. It was cute when constexpr was introduced, but the fall-out is scary plus we never actually refer to this as a _decl-specifier_seq_ in the text that follows.

jensmaurer:
I have created a clean-up patch; see #2338. I hope to get CWG to review the issue in San Diego.

tkoeppe:
@jensmaurer: does https://wg21.link/p1102r2 fix the underlying issue?

jensmaurer:
Yes, the original issue is fixed with P1102R2. The /decl-specifier-seq/ concerns will be addressed by #2338.

PR #2338, opened 2018-09-22
[expr.prim.lambda] Avoid decl-specifier-seq in lambda-declarator

jensmaurer:
Using a decl-specifier-seq pulls in a lot of undesirable
connotations, plus the following text only refers to
individual decl-specifiers (by name), never to the entire
decl-specifier-seq.

Therefore, introduce a new grammar non-terminal
'lambda-specifier-seq'.

Fixes #2333.


timsong-cpp:
We need something that says each _lambda-specifier_ can only appear at most once in a complete _lambda-specifier-seq_.

jensmaurer:
@timsong-cpp: Right, fixed.

jensmaurer:
CWG in San Diego: This will become a core issue.

timsong-cpp:
This now also needs something that disallows having both `constexpr` and `consteval`.

jensmaurer:
CWG telecon 2021-12-09: Moved CWG2509 to Ready.

PR #2339, opened 2018-09-23
[temp.arg.nontype] Use struct instead of class to make example valid

ldionne:
This is an editorial fix.

[P0732R2](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0732r2.pdf) was merged in Rapperswil, but one example is invalid because the constructor for A is private, so it can't be constructed from the string literal.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -1377,7 +1377,7 @@ X<const char*, "Studebaker"> x; // error: string literal as template-argument
 const char p[] = "Vivisectionist";
 X<const char*, p> y;            // OK
 
-class A {
+struct A {
   constexpr A(const char*) {}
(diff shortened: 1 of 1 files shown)


jensmaurer:
Looks good. Please adjust your commit comment to leave out "This is an editorial fix." which has no long-term value.

cpplearner:
Is `auto operator<=>(A, A) = default;` valid? I think it should be `friend auto operator<=>(A, A) = default;` or `auto operator<=>(A) = default;`.

ldionne:
I don't think it is, but funnily enough it is used in many many places in the document. A separate fix is probably better for that.

PR #2344, opened 2018-10-04
[class.copy.ctor] Add missing cross-references to Annex D

AlisdairM:
Add the missing cross-references to corresponding Annex D entry
for the deprecated implicit declaration of copy constructor and
copy assignment operator when either the destructor or other
copy operation is user-declared.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -1620,7 +1620,7 @@ constructor or move assignment operator, the implicitly declared copy
 constructor is defined as deleted; otherwise, it is defined as
 defaulted\iref{dcl.fct.def}.
 The latter case is deprecated if the class has a user-declared copy assignment
-operator or a user-declared destructor.
+operator or a user-declared destructor \iref{depr.impldec}.
 
 \pnum
 The implicitly-declared copy constructor for a class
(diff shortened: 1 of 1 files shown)


tkoeppe:
Oh, could you also add a cross reference for the lambda capture of `this` via `[=]`?

source/classes.tex
```diff
@@ -1866,7 +1866,7 @@
 assignment operator is defined as deleted; otherwise, it is defined as
 defaulted\iref{dcl.fct.def}.
 The latter case is deprecated if the class has a user-declared copy constructor
-or a user-declared destructor.
+or a user-declared destructor  \iref{depr.impldec}.
```
BitonicEelis: Two spaces?

AlisdairM: @tkoeppe I believe that cross reference is already there in [expr.prim.lambda.capture]p7, just before the example.  I did a search on the word 'deprecated' and these were the only two uses without a cross reference.

@BitonicEelis thanks, good catch.  Force-pushed the quick correction to avoid cluttering the history when merged.

tkoeppe: Oh, very good, thanks!

PR #2346, opened 2018-10-07
Move "plain ints" statement to a note

jfbastien:
Jens and Tony believe it isn't normative and doesn't need a DR to fix. Let's do it now.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -4243,7 +4243,7 @@ extended signed integer types are collectively called
 \begin{note}
 Plain
 \tcode{int}s have the natural size suggested by the architecture of the
-execution environment%;
+execution environment;
 the other signed integer types are provided to meet special needs.
 \end{note}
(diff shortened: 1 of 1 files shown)


source/basic.tex
```diff
@@ -4240,14 +4240,12 @@
...
 \indextext{integral type!implementation-defined \tcode{sizeof}}%
+\begin{note}
 Plain
 \tcode{int}s have the natural size suggested by the architecture of the
-execution environment%
-\footnote{\tcode{int} must also be large enough
-to contain any value in the range
-\crange{INT_MIN}{INT_MAX},
-as defined in the header \tcode{<climits>}.};
+execution environment%;
```
CaseyCarter: The `%` was here to avoid introducing whitespace before the footnote. Since you've removed the footnote, you can remove the `%` as well.

PR #2351, opened 2018-10-11
[decl.init]/10 Fix specified initialization.

stryku:
According to [basic.start.static]/2, for objects with static storage duration,
zero initialization performs only if constant initialization does not.
[decl.init]/10 can be generalized to static initialization.
This is an editorial note change.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -4121,10 +4121,9 @@ of an entity
 of reference type is ill-formed.
 
 \pnum
-\begin{note} Every
-object of static storage duration is
-zero-initialized at program startup before any other initialization
-takes place.
+\begin{note} For every object of static storage duration,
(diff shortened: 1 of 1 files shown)


source/declarations.tex
```diff
@@ -4121,10 +4121,9 @@
 of reference type is ill-formed.
 
 \pnum
-\begin{note} Every
-object of static storage duration is
-zero-initialized at program startup before any other initialization
-takes place.
+\begin{note} Every object of static storage duration is
+static-initialized (\ref{basic.start.static}) at program startup
```
zygoloid: Please use `\iref` here instead of a parenthesized `\ref`.

stryku:
I've somehow missed your comments, sorry. I hope I understood the `\iref` syntax right.

PR #2353, opened 2018-10-12
[class.mem] Add opaque-enum-declaration to member-declaration

cpplearner:
[class.mem]/5:

> [...] an enumeration can be introduced with an opaque-enum-declaration [...]


source/classes.tex
```diff
@@ -448,7 +448,8 @@
     template-declaration\br
     deduction-guide\br
     alias-declaration\br
-    empty-declaration
+    empty-declaration\br
+    opaque-enum-declaration
```
jensmaurer: Could you please move this up one line, so that empty-declaration stays last?

PR #2354, opened 2018-10-12
Remove "shall equal"s naming from conditional/decay/enable_if

mclow:
Instead of saying "shall equal", which implies a condition on the user, we should say "names" or "is".

This change uses 'names', since that's what `identity` uses: The member typedef `type` names the type `T`.
Suggestion from Alisdair.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -16925,10 +16925,10 @@ assert((is_same_v<remove_all_extents_t<int[][3]>, int>));
 \tcode{template<class T>\br struct decay;}
  &
  Let \tcode{U} be \tcode{remove_reference_t<T>}. If \tcode{is_array_v<U>} is
- \tcode{true}, the member typedef \tcode{type} shall equal
+ \tcode{true}, the member typedef \tcode{type} names
  \tcode{remove_extent_t<U>*}. If \tcode{is_function_v<U>} is \tcode{true},
- the member typedef \tcode{type} shall equal \tcode{add_pointer_t<U>}. Otherwise
- the member typedef \tcode{type} equals \tcode{remove_cv_t<U>}.
+ the member typedef \tcode{type} names \tcode{add_pointer_t<U>}. Otherwise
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -16940,15 +16940,15 @@
 \tcode{template<bool B, class T = void>} \tcode{struct enable_if;}
  &
  If \tcode{B} is \tcode{true}, the member typedef \tcode{type}
- shall equal \tcode{T}; otherwise, there shall be no member
+ names \tcode{T}; otherwise, there shall be no member
```
CaseyCarter: s/shall be/is/ since you're touching this line anyway?

mclow: Makes sense to me. But then, should we fix `common_type` (probably not, since users can specialize it), `basic_common_reference` (ditto), `invoke_result` (probably should).

CaseyCarter: Yes, we should fix those, but notably none of them are on this line. ;)

jensmaurer:
I agree with the desire to remove "shall equal", but I'm wondering whether "names the type..." would be better. Otherwise, we have a type-id without any introducer at all.

W-E-Brown:
> On Oct 13, 2018, at 1:13 PM, Jens Maurer <notifications@github.com> wrote:
> 
> I agree with the desire to remove "shall equal", but I'm wondering whether "names the type..." would be better. Otherwise, we have a type-id without any introducer at all.

Alternatively, we could rephrase as "denotes" or "denotes the type", as we say in a number of places.

There's also "alias" (as in "is an alias for" or, as a verb, "aliases").  We do already use similar phrasing, e.g., "provides an alias" and "be an alias for".

lichray:
Oh, the issue I wanted to open is already here.

jensmaurer:
Editorial meeting consensus:
- Uses "denotes" without "the type".

jwakely:
#2678 supersedes this PR, so I'm closing this one.

PR #2360, opened 2018-10-17
[output.iterators] Remove misleading italics from note

CaseyCarter:
Since LWG asked me to remove these from P0896's `OutputIterator`, they probably don't belong in *`Cpp17OutputIterator`* either.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -476,11 +476,9 @@ are valid and have the indicated semantics.
 The only valid use of an
 \tcode{operator*}
 is on the left side of the assignment statement.
-\textit{Assignment through the same value of the iterator happens only once.}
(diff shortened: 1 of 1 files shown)


source/iterators.tex
```diff
@@ -476,11 +476,9 @@
 The only valid use of an
 \tcode{operator*}
 is on the left side of the assignment statement.
-\textit{Assignment through the same value of the iterator happens only once.}
+Assignment through the same value of the iterator happens only once.
```
jwakely: If we can't use blinking text here instead then I agree with removing the emphasis.

CaseyCarter:
```html
<marquee>Assignment through the same value of the iterator happens only once.</marquee>
```

PR #2361, opened 2018-10-17
[iterator.operations] Harmonize bad wording with shiny new Ranges wording

CaseyCarter:


source/iterators.tex
```diff
@@ -1110,7 +1110,7 @@
 iterator reference
 \tcode{i}
 by
-\tcode{n}.
```
tkoeppe: This is still awkward. How about something like:

> The iterator `i` is incremented by $n$ if $n$ is non-negative, and decremented by $-n$ otherwise.

CaseyCarter: I'd like to keep the active voice. I'll propose:
> Increments `i` by `n` if `n` is non-negative, and decrements by `-n` otherwise.

CaseyCarter: Ick. And while I'm here, I should change the Requires:
```diff
-Requires:
+Expects:
 n shall be negative only for bidirectional
-and random access
 iterators.
```

tkoeppe: Both sound good, but maybe make the second bit a separate change? Both commits can go in the same pull request of course.

CaseyCarter:
This changes the precondition wording for `advance` and `distance` to agree with the wording for `ranges::advance` and `::distance` which LWG was happy with in Batavia. The absolute-value change for the effects of `advance` is novel, but certain to be non-controversial (and I will change P0896 to agree).

source/iterators.tex
```diff
@@ -1099,18 +1099,14 @@
 
 \begin{itemdescr}
 \pnum
-\requires
+\expects
 \tcode{n}
-shall be negative only for bidirectional and random access iterators.
+shall be negative only for bidirectional iterators.
```
zygoloid: Do not use "shall" within *Effects*: elements. We're describing a condition, not a requirement.

source/iterators.tex
```diff
@@ -1121,6 +1117,13 @@
 \end{itemdecl}
 
 \begin{itemdescr}
+\pnum
+\expects
+\tcode{last} shall be reachable from \tcode{first}, or
+\tcode{InputIterator} shall meet the
+\oldconcept{RandomAccess\-Iterator} requirements and
+\tcode{first} shall be reachable from \tcode{last}.
```
zygoloid: Likewise, shall -> is.

source/iterators.tex
CaseyCarter: Fixed.

source/iterators.tex
CaseyCarter: Likewise fixed.

Issue #1242, opened 2016-12-14
[lib] Subheadings in class descriptions


jensmaurer:
We have both:

- Class template blah
-- blah constructors
(etc)

and

- Class template blah
-- Construction and assignment [note: omitting "blah" in subordinate headings]
(etc)

[basic.string] is an example for the former, [string.view.template] is an example for the latter.

Which way is the "right" way?

jensmaurer:
Editorial meeting consensus: Do not repeat a parent heading item in subordinate headings.

tkoeppe:
@jensmaurer: What about [container.node]?

jensmaurer:
@tkoeppe: [container.node] is odd. It specifies an exposition-only class _node_handle_, which need not exist under this name. Everywhere else in the vicinity, we specify requirements on classes or class templates by specifying the valid operations on them, instead of showing a class synopsis.

tkoeppe:
@jensmaurer: Yeah, that's true. Not sure what to do here. @zygoloid: Opinions?

My primary goal here is to get rid of a bold-italic-teletype font requirement, which we cannot satisfy :-)

zygoloid:
I'm happy with removing *`node_handle`* from these headings, as they are not really describing members of some class named *`node_handle`*; rather, they're describing members of the exposition-only notion of a "node handle". Done in 2fcacf3f.

johelegp:
Do you have a stance on the repetition in the headers' synopsis? That repetition might provide some context that's implicitly available in the subclauses for being nested.

jensmaurer:
@johelegp, I don't understand the question. Header synopses usually say "Header &lt;blah&gt; synopsis", and that doesn't mention a class name.

johelegp:
I mean the line comments like this:
```C++
    // [time.duration], class template duration
    template<class Rep, class Period = ratio<1>> class duration;
```

jensmaurer:
@johelegp, those comments quote the subclause heading, so those seem fine.

jensmaurer:
This should be complete now.

PR #2366, opened 2018-10-19
[input.output] Remove class name repeated in subheadings

jensmaurer:
Partially addresses #1242.


jensmaurer:
@tkoeppe, could you please merge this? We're on the finishing line here.

PR #2369, opened 2018-10-20
[rand.dist.uni.real] Correct a typo

majnemer:
The uniform_­real_­distribution produces numbers such that a ≤ x < b.
However, the requirement in [rand.dist.uni.real]p2 as written would permit a and b to be equal.


timsong-cpp:
This is intentional and not a typo. See https://wg21.link/lwg2168 and https://wg21.link/N3926.

PR #2370, opened 2018-10-25
[basic.lookup,over.call.func] Clarify lookup rules for function calls.

### LINKED ISSUE #2365: Wording issue between [over.call.func] and [basic.lookup]
[over.call.func]/3 says:

> In unqualified function calls, the name is not qualified by an -> or . operator and has the more general form of a primary-expression. The name is looked up in the context of the function call following the normal rules for name lookup in function calls. 

[basic.lookup]/2 says:

> A name “looked up in the context of an expression” is looked up as an unqualified name in the scope where the expression is found.

We can have an *unqualified function call* that nevertheless uses a *qualified name* (e.g. `foo::bar(1)`), but we're saying we have to look that name up using unqualified lookup instead of qualified lookup. That doesn't seem right. (h/t [9room](https://stackoverflow.com/q/52889777/2069064))

jensmaurer:
Also add cross-references to [class.member.lookup] when
talking about 'looked up in the class of the object expression'
in [basic.lookup.classref].

Fixes #2365.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -1209,8 +1209,8 @@ checking have succeeded are the attributes introduced by the name's
 declaration used further in expression processing\iref{expr}.
 
 \pnum
-A name ``looked up in the context of an expression'' is looked up as an
-unqualified name in the scope where the expression is found.
+A name ``looked up in the context of an expression'' is looked up
+in the scope where the expression is found.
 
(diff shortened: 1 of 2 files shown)


jensmaurer:
Rebased.

PR #2372, opened 2018-10-27
[conv.lval] Improve the note about std::nullptr_t case

k-satoda:
The meaning of "fetched from memory" is vague, and the two uses of term
"access" [defns.access] are inappropreate here because
  - it causes undefined behavior to access an inactive member of a
    union, based on the rule of out-of-lifetime object [basic.life]:
    > The program has undefined behavior if:
    >   - the glvalue is used to access the object, or
  - it implies reading the referenced object of std::nullptr_t,
    which is defined to be a side effect for volatile-qualified case.
    [intro.execution]
    > Reading an object designated by a volatile glvalue, ... are all
    > side effects, ...

---

The change blindly assumes that "no value is fetched from memory" is
true and rephrases it as "the conversion does not read the object", but
I'm not sure how it is true. In fact, gcc 8.2 emits read instruction for
volatile-qualified case https://godbolt.org/z/uhn72l (clang 7.0.0
doesn't), and both gcc 8.2 and clang 7.0.0 reads a member subobject of
type std::nullptr_t https://godbolt.org/z/fqBU-H.

FYI: The note was added by CWG issue 2140 for C++17.
https://wg21.cmeerw.net/cwg/issue2140
commit 28c12a81642138f0f25e8942298b03b4516d2a9a

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -631,9 +631,9 @@ following rules:
 \item If \tcode{T} is \cv{}~\tcode{std::nullptr_t}, the result is a
 null pointer constant\iref{conv.ptr}.
 \begin{note}
-Since no value is fetched from memory,
-there is no side effect for a volatile access\iref{intro.execution}, and
-an inactive member of a union\iref{class.union} may be accessed.
+Since the conversion does not read the object,
+there is no side effect even if \tcode{T} is volatile-qualified\iref{intro.execution}, and
+the glvalue can refer to an inactive member of a union\iref{class.union}.
 \end{note}
 
 \item Otherwise, if \tcode{T} has a class
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -631,7 +631,7 @@ following rules:
 \item If \tcode{T} is \cv{}~\tcode{std::nullptr_t}, the result is a
 null pointer constant\iref{conv.ptr}.
 \begin{note}
-Since the conversion does not read the object,
+Since the conversion does not access the object to which the glvalue refers,
 there is no side effect even if \tcode{T} is volatile-qualified\iref{intro.execution}, and
 the glvalue can refer to an inactive member of a union\iref{class.union}.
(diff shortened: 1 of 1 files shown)


source/expressions.tex
```diff
@@ -631,9 +631,9 @@
 \item If \tcode{T} is \cv{}~\tcode{std::nullptr_t}, the result is a
 null pointer constant\iref{conv.ptr}.
 \begin{note}
-Since no value is fetched from memory,
-there is no side effect for a volatile access\iref{intro.execution}, and
-an inactive member of a union\iref{class.union} may be accessed.
+Since the conversion does not read the object,
```
jensmaurer: Maybe "read" -> "access"?

k-satoda:
I found a bug report against clang about access on inactive
std::nullptr_t member of a union.
https://bugs.llvm.org/show_bug.cgi?id=23833
The test case still fails with clang 7.0.0 which means an exisiting
disagreement between the note and an implementation. Hmm.
https://wandbox.org/permlink/8UWvvFWLC1pF0OyK

zygoloid:
> The test case still fails with clang 7.0.0 which means an exisiting
> disagreement between the note and an implementation. Hmm.

Implementations take time to implement DR resolutions. (FWIW, I just fixed this in Clang trunk.)

k-satoda:
Understood. Glad to see the fix.

As for how can we say "no value is fetched from memory" or
"the conversion does not access ..." from non-note rules, my best guess
so far is that the last bullet ("the value contained in the object
indicated by the glvalue is the prvalue result") determines an access as
it implies reading the value. That means the same exemption can be
applied for empty classes;
```cpp
struct S {};
union { int x; S y; } a = {37};
S s = a.y; // No undefined behavior?
```
because member-wise copy (defined at [class.copy.ctor]) for such a class
does no further lvalue-to-rvalue conversions.
Correct?

PR #2375, opened 2018-10-29
[iterator.requirements.general] operations satisfy requirements, not …

CaseyCarter:
…operations

Drive-by: change "satisfy .... requirements" to "meet ... requirements".


jensmaurer:
@tkoeppe, this seems obviously correct and an improvement. Do you feel like merging?

PR #2379, opened 2018-11-01
[alg.find.end,alg.search] Remove redundant 'Effects' paragraphs

### LINKED ISSUE #2377: [alg.find.end] and [alg.search] should not use "subsequence"
["Subsequence"](https://en.wikipedia.org/wiki/Subsequence) means something else mathematically; what we mean here is a substring. But it seems that the affected _Effects_ paragraphs have no normative effect anyway (the _Returns_ fully specifies what we are searching for) and can be removed outright.

Compare https://wg21.link/LWG3115.

jensmaurer:
which were using the term 'subsequence' incorrectly.

Fixes #2377.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -3234,10 +3234,6 @@ in \range{first1}{last1 - (last2 - first2)}.
 \end{itemize}
 
 \pnum
-\effects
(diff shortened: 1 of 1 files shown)


jensmaurer:
Rebased.

PR #2389, opened 2018-11-07
[over.ics.user] Mark "user-defined conversion sequence" as a term of art

CaseyCarter:


tkoeppe:
@zygoloid I'd be happy to have this merged now. We don't need "after-motion" for very localized changes like this, right?

PR #2393, opened 2018-11-10
Clarify the contexts where void expressions appear

maggyero:
This pull request will:
* remove void expressions from temporary object materialization;
* add void expressions to the prvalue definition and precise that they have no results/result objects.


maggyero:
@jwakely Thank you for the suggestion, title changed.
@zygoloid Pull request rebased.

PR #2438, opened 2018-11-11
P1236R1 Alternative Wording for P0907R4 Signed Integers are Two's Com…

### LINKED ISSUE #2402: [2018-11 CWG Motion 10] P1236R1 Alternative Wording for P0907R4 Signed Integers are Two's Complement
[P1236R1](http://wg21.link/p1236r1)

jensmaurer:
…plement

Fixes #2402.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -4165,51 +4165,7 @@ layout-compatible standard-layout class types\iref{class.mem}.
 \rSec2[basic.fundamental]{Fundamental types}
 
 \pnum
-\indextext{type!integral}%
-\indextext{floating-point type|see{type, floating-point}}%
 \indextext{type!implementation-defined \tcode{sizeof}}%
-\indextext{type!Boolean}%
-\indextext{type!\idxcode{char}}%
-\indextext{type!character}%
-Objects declared as characters (\tcode{char}) shall be large enough to
-store any member of the implementation's basic character set. If a
-character from this set is stored in a character object, the integral
-value of that character object is equal to the value of the single
-character literal form of that character. It is \impldef{signedness of \tcode{char}}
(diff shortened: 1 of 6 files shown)


source/basic.tex
```diff
@@ -4257,98 +4214,175 @@
...
+Type \tcode{bool} is a distinct type that has
+the same object representation,
+value representation, and
+alignment requirements as
+an \impldef{underlying type of \tcode{bool}} unsigned integer type.
+The values of type \tcode{bool} are
+\tcode{true} and \tcode{false}.\footnote{Using a \tcode{bool} value
+in ways described by this document as ``undefined'',
+such as by examining the value of an uninitialized automatic object,
+might cause it to behave as if it is neither \tcode{true} nor \tcode{false}.}
```
jfbastien: Nitpick on the note: it can behave as neither true nor false, but it can also behave as both true and false.

```suggestion
might cause it to behave as if it is neither \tcode{true} nor \tcode{false},
or might cause it to behave as if it is both \tcode{true} and \tcode{false}.}
```

burblebee: @jfbastien Please open a separate editorial issue on this if you'd like to see this clarified - even better if you can suggest new wording :)

jfbastien: I suggested wording right there in GitHub, you can click "accept" and we're done. I can't suggest new wording right now without causing a merge conflict with this PR.

burblebee: @jfbastien please open a new issue and we can merge it later?  This is existing wording so it should not be handled as part of this paper.  Thanks for understanding!

jfbastien: https://github.com/cplusplus/draft/issues/2483

PR #2450, opened 2018-11-12
P1084R2 Today’s return-type-requirements Are Insufficient

### LINKED ISSUE #2397: [2018-11 CWG Motion 4] P1084R2 Today's return-type-requirements Are Insufficient
[P1084R2](http://wg21.link/p1084r2)

burblebee:
[expr.prim.req.compound] Renamed C5 to C3 in remaining example.

Fixes #2397.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -2597,7 +2597,7 @@ does not require that type to be complete\iref{basic.types}.
 \begin{bnf}
 \nontermdef{return-type-requirement}\br
     trailing-return-type\br
-    \terminal{->} \opt{cv-qualifier-seq} constrained-parameter \opt{cv-qualifier-seq} \opt{abstract-declarator}
+    \terminal{->} qualified-concept-name
 \end{bnf}
 
 \pnum
@@ -2607,44 +2607,55 @@ of template arguments (if any) and verification of
 semantic properties proceed in the following order:
 
 \begin{itemize}
-\item Substitution of template arguments (if any)
+\item
+Substitution of template arguments (if any)
 into the \grammarterm{expression} is performed.
 
-\item If the \tcode{noexcept} specifier is present,
+\item
+If the \tcode{noexcept} specifier is present,
 \tcode{E} shall not be a potentially-throwing expression\iref{except.spec}.
 
-\item If the \grammarterm{return-type-requirement} is present, then:
+\item
+If the \grammarterm{return-type-requirement} is present, then:
 
 \begin{itemize}
-\item Substitution of template arguments (if any)
+\item
+Substitution of template arguments (if any)
 into the \grammarterm{return-type-requirement} is performed.
 
-\item If the \grammarterm{return-type-requirement} is a
-\grammarterm{trailing-return-type},
+\item
+If the \grammarterm{return-type-requirement} is a
+\grammarterm{trailing-return-type}\iref{dcl.decl},
 %%% FIXME: is -> shall be
 \tcode{E} is implicitly convertible to
 the type named by the \grammarterm{trailing-return-type}.
 If conversion fails, the enclosing \grammarterm{requires-expression}
 is \tcode{false}.
 
-\item If the \grammarterm{return-type-requirement}
-starts with a \grammarterm{constrained-parameter}\iref{temp.param},
-the \grammarterm{expression} is deduced against
-an invented function template \tcode{F}
-using the rules in \ref{temp.deduct.call}.
-\tcode{F} is a \tcode{void} function template
-with a single type template parameter \tcode{T}
-declared with the \grammarterm{constrained-parameter}.
-A \grammarterm{cv-qualifier-seq} \cv{} is formed
-as the union of \tcode{const} and \tcode{volatile} specifiers
-around the \grammarterm{constrained-parameter}.
-\tcode{F} has a single parameter
-whose \grammarterm{type-specifier} is \cv{}~\tcode{T}
-followed by the \grammarterm{abstract-declarator}.
-%%% FIXME: Remove this; if deduction fails, the construct should
-%%% be ill-formed.
-If deduction fails,
-the enclosing \grammarterm{requires-expression} is \tcode{false}.
+\item
+If the \grammarterm{return-type-requirement} is
+a \grammarterm{qualified-concept-name}\iref{temp.param} of the form
+\opt{\grammarterm{nested-name-specifier}} \grammarterm{concept-name},
+the concept it denotes shall be satisfied with
+\tcode{decltype((E))} as its sole argument.
+If the \grammarterm{return-type-requirement} is
+a \grammarterm{qualified-concept-name} of the form
+\opt{\grammarterm{nested-name-specifier}} \grammarterm{concept-name}
+\tcode{<} \opt{\grammarterm{template-argument-list}} \tcode{>},
+the concept it denotes shall be satisfied with
+\tcode{decltype((E))} as its first argument and with
+the elements, in the order listed, of
+the \grammarterm{template-argument-list}
+comprising the concept's subsequent arguments.
+\begin{note}
+Thus, requirements of the form
+\tcode{\{ E \} -> Concept;} or of the form
+\tcode{\{ E \} -> Concept<>;} are equivalent to
+\tcode{E; requires Concept<decltype((E))>;}
+while a requirements of the form
+\tcode{\{ E \} -> Concept<A$_1$, A$_2$, $\cdots$, A$_n$>;} is equivalent to
+\tcode{E; requires Concept<decltype((E)), A$_1$, A$_2$, $\cdots$, A$_n$>;}.
+\end{note}
 \end{itemize}
 \end{itemize}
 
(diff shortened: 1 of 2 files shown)


source/expressions.tex
```diff
@@ -2600,7 +2600,7 @@
 \begin{bnf}
 \nontermdef{return-type-requirement}\br
     trailing-return-type\br
-    \terminal{->} \opt{cv-qualifier-seq} constrained-parameter \opt{cv-qualifier-seq} \opt{abstract-declarator}
+    qualified-concept-name
```
hubert-reinterpretcast: The `->` in the grammar is intended.

source/expressions.tex
```diff
@@ -2610,45 +2610,51 @@
...
+  \item
+    If the \grammarterm{return-type-requirement} is a
+    \grammarterm{trailing-return-type}\iref{dcl.decl},
+    %%% FIXME: is -> shall be
+    \tcode{E} is implicitly convertible to
+    the type named by the \grammarterm{trailing-return-type}.
+    If conversion fails, the enclosing \grammarterm{requires-expression}
+    is \tcode{false}.
+  \item
+    If the \grammarterm{return-type-requirement} is
```
hubert-reinterpretcast: If the _return-type-requirement_ has a [ ... ].

source/expressions.tex
```diff
@@ -2610,45 +2610,51 @@
...
+    a \grammarterm{qualified-concept-name}\iref{temp.param} of the form
+    \opt{\grammarterm{nested-name-specifier}} \grammarterm{concept-name},
+    the concept it denotes shall be satisfied with
+    \tcode{decltype((E))} as its sole argument.
+    If the \grammarterm{return-type-requirement} is
```
hubert-reinterpretcast: If the _return-type-requirement_ has a [ ... ].

source/expressions.tex
burblebee: missed that - thanks!

source/expressions.tex
burblebee: thanks!

burblebee:
This branch is failing the travis checks because commit 0af79b8a (which resolves conflicts between cwg-4 and cwg-15) refers to the new grammar term /type-constraint/ which won't exist until after cwg-15 is merged.  I couldn't add this commit in cwg-15 because it doesn't have the wording that needed to be changed.  How to deal with this situation moving forward?

@zygoloid please ignore the failing checks.  To merge this PR, squash the one "Fixup:" commit 2a38e296, and apply commit 0af79b8a after merging motions-2018-11-cwg-15.

W-E-Brown:
Unless I'm missing something, the library parts of this paper (affecting Clause [concepts]) seem not yet to have been entered.  See P1084R2, §5.2, top of page 5.

CaseyCarter:
> Unless I'm missing something, the library parts of this paper (affecting Clause [concepts]) seem not yet to have been entered.

I can prepare a commit with these changes, if it would be helpful. I'd like to fix [`UniformRandomNumberGenerator`](http://eel.is/c++draft/rand.req.urng) as well, but it's not in a subclause of [concepts]. Is that too great a stretch to be termed editorial?

jensmaurer:
@burblebee, Walter is right: The 5.2 changes are missing. You can use the changes in #2488, but I'd suggest that you integrate them in this pull request so that the entire paper is applied in one place (and yes, that's a "fixup").
@CaseyCarter, if you're writing a cleanup paper anyway, maybe we can postpone that? Or maybe create a separate editorial issue so that we can weigh the "editorial" propertly independently?

CaseyCarter:
> @CaseyCarter, if you're writing a cleanup paper anyway, maybe we can postpone that?

Yes, I'll just put it in the followup.

source/expressions.tex
```diff
@@ -2610,45 +2610,43 @@
...
-around the \grammarterm{constrained-parameter}.
-\tcode{F} has a single parameter
-whose \grammarterm{type-specifier} is \cv{}~\tcode{T}
-followed by the \grammarterm{abstract-declarator}.
-%%% FIXME: Remove this; if deduction fails, the construct should
-%%% be ill-formed.
-If deduction fails,
-the enclosing \grammarterm{requires-expression} is \tcode{false}.
-\end{itemize}
+\item
```
zygoloid: Please can you undo the indentation of these items? It makes it very hard to see what's actually changing here, and generally we don't indent like this.

source/expressions.tex
```diff
@@ -2610,45 +2610,43 @@
...
+    The immediately-declared constraint\iref(temp) of \tcode{decltype((E))}
+    shall be satisfied.
+    \begin{note}
+    Thus, constraints of the form
+    \tcode{\{ E \} -> Concept;} or of the form
+    \tcode{\{ E \} -> Concept<>;} are equivalent to
+    \tcode{E; requires Concept<decltype((E))>;}
+    while a constraint of the form
+    \tcode{\{ E \} -> Concept<A$_1$, A$_2$, . . . , A$_n$>;} is equivalent to
+    \tcode{E; requires Concept<decltype((E)), A$_1$, A$_2$, ..., A$_n$>;}.
```
zygoloid: It's jarring that the ellipsis on this line and the line above are formatted differently. These are both mathematical ellipses rather than code ellipses, so they should be either `\ldots` or `\cdots`; we tend to use `\cdots` in contexts such as this where there could otherwise be confusion with actual ellipses and parameter packs.

source/expressions.tex
```diff
@@ -2610,45 +2610,43 @@
...
+    \end{note}
```
zygoloid: This note is overall a little hard to read due to the frequent switching to and from code font and the `E; requires ...;` construction appearing to be two separate clauses. Perhaps a different presentation would help -- maybe converting the `tcode` here into `codeblock`s, or more drastically something like:

> [*Example*: Given a concept named `X`,
> ```
> requires {
>   { E1 } -> X;
>   { E2 } -> X<A1, A2, ..., An>;
> }
> ```
> is equivalent to
> ```
> requires {
>   E1; requires X<decltype((E1))>;
>   E2; requires X<decltype((E2)), A1, A2, ..., An>;
> };
> ```
> -- *end example*]

@W-E-Brown What do you think?

source/expressions.tex
```diff
@@ -2610,45 +2610,43 @@
...
+    is \tcode{false}.
+  \item
+    If the \grammarterm{return-type-requirement}
+    is of the form \tcode{->} \grammarterm{type-constraint}, then
+    the contextually-determined type being constrained
+    is \tcode{decltype((E))}.
+    The immediately-declared constraint\iref(temp) of \tcode{decltype((E))}
+    shall be satisfied.
+    \begin{note}
+    Thus, constraints of the form
```
zygoloid: These are requirements, not constraints.

W-E-Brown:
> On Nov 20, 2018, at 11:46 PM, Richard Smith <notifications@github.com> wrote:
> 
> @zygoloid commented on this pull request.
> 
> In source/expressions.tex <https://github.com/cplusplus/draft/pull/2450#discussion_r235259927>:
> 
> > +  \item
> +    If the \grammarterm{return-type-requirement}
> +    is of the form \tcode{->} \grammarterm{type-constraint}, then
> +    the contextually-determined type being constrained
> +    is \tcode{decltype((E))}.
> +    The immediately-declared constraint\iref(temp) of \tcode{decltype((E))}
> +    shall be satisfied.
> +    \begin{note}
> +    Thus, constraints of the form
> +    \tcode{\{ E \} -> Concept;} or of the form
> +    \tcode{\{ E \} -> Concept<>;} are equivalent to
> +    \tcode{E; requires Concept<decltype((E))>;}
> +    while a constraint of the form
> +    \tcode{\{ E \} -> Concept<A$_1$, A$_2$, . . . , A$_n$>;} is equivalent to
> +    \tcode{E; requires Concept<decltype((E)), A$_1$, A$_2$, ..., A$_n$>;}.
> +    \end{note}
> This note is overall a little hard to read due to the frequent switching to and from code font and the E; requires ...; construction appearing to be two separate clauses. Perhaps a different presentation would help -- maybe converting the tcode here into codeblocks, or more drastically something like:
> 
> [Example: Given a concept named X,
> 
> requires {
>   { E1 } -> X;
>   { E2 } -> X<A1, A2, ..., An>;
> }
> is equivalent to
> 
> requires {
>   E1; requires X<decltype((E1))>;
>   E2; requires X<decltype((E2)), A1, A2, ..., An>;
> };
> -- end example]
> 
> @W-E-Brown <https://github.com/W-E-Brown> What do you think?
> 


The "drastic" reformulation seems okay to me, stylistically, although my preference would be to use C (instead of X) for the concept name.  Also, CWG seemed to consider A's subscripts important here, so let's please keep those.

Note that the reformulation omits the code version with the empty angle brackets.  That's probably worth keeping.

The earlier suggestion re use of \cdots, too, seems reasonable for these contexts.

source/expressions.tex
burblebee: @zygoloid The reformatting bothered me too for the same reason, but @tkoeppe  suggested in past reviews that I fix the formatting to follow the new [guidelines](https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines) as I come across it, and it would be odd to fix the modified \items without fixing them all.  What to do?  As for how we generally format, I'm trying to follow those guidelines - did I misunderstand them?

tkoeppe: Maybe don't introduce too much non-semantic change in important motion applications like this, but keep those changes for pure-editorial changes during regular editorial review work after motions have been merged?

burblebee: @tkoeppe Sounds good.

source/expressions.tex
burblebee: Huh, odd that it cut and pasted that way, but good to know about \cdots v. \ldots (now added to our [Q & A](https://github.com/cplusplus/draft/wiki/Q-&-A) :) ).  That said, there does appear to be some confusion over when to use what, so I opened #2529.

PR #2456, opened 2018-11-13
P0972R0 <chrono> zero(), min(), and max() should be noexcept

### LINKED ISSUE #2415: [2018-11 LWG Motion 9] P0972R0 <chrono> zero(), min(), and max() should be noexcept
[P0972R0](http://wg21.link/p0972r0)

burblebee:
Fixes #2415.

### DIFF
--- a/source/time.tex
+++ b/source/time.tex
@@ -1133,9 +1133,9 @@ if it behaved like an integral type for the purpose of these conversions.
 template<class Rep>
   struct duration_values {
   public:
-    static constexpr Rep zero();
-    static constexpr Rep min();
-    static constexpr Rep max();
+    static constexpr Rep zero() noexcept;
+    static constexpr Rep min() noexcept;
+    static constexpr Rep max() noexcept;
 };
 \end{itemdecl}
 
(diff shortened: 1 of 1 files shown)


source/time.tex
```diff
@@ -1133,9 +1133,9 @@
 template<class Rep>
   struct duration_values {
   public:
-    static constexpr Rep zero();
-    static constexpr Rep min();
-    static constexpr Rep max();
+    static constexpr Rep zero() noexcept;;
```
aburch: The extra semicolon should not be there?

(I have no experience with the C++ standardization process, so excuse me if this is the wrong way to point out a typo.)

burblebee: Thanks for catching this!!

burblebee:
Should I remove the "changes required" labels now that the typos have been fixed and pushed?

I assume an editor will assign themselves to this PR for a final review?

jensmaurer:
@burblebee: Yes, you can remove the label after the fix. The label is just to call everybody's (including the author's) attention that this is not ready for final review.

PR #2462, opened 2018-11-14
Add an index entry for "extern template"

Quuxplusone:
If you forget (or never knew) what this construct is called, then the existing index entries don't help you locate it.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -6137,6 +6137,7 @@ expression.
 
 \pnum
 \indextext{instantiation!explicit}%
+\indextext{\idxcode{extern template}|see{instantiation, explicit}}%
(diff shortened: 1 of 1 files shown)


jfbastien:
I feel like this index entry should refer to an external document... 🤡

tkoeppe:
Mefeels we need an editorial incubator, too :-)

tkoeppe:
@jensmaurer: does this need to wait for after motions, or can we cope?

jfbastien:
> @jensmaurer: does this need to wait for after motions, or can we cope?

Can you koeppe? 🤔

OK OK I stop now.

tkoeppe:
That took me an unnecessarily long time to get :-)

PR #2468, opened 2018-11-14
P1020R1 Smart pointer creation with default initialization

### LINKED ISSUE #2424: [2018-11 LWG Motion 18] P1020R1 Smart pointer creation with default initialization
[P1020R1](http://wg21.link/p1020r1)

burblebee:
[memory.syn] Move return types to before function names for consistency.

Fixes #2424.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -6676,12 +6676,19 @@ namespace std {
   template<class T, class D = default_delete<T>> class unique_ptr;
   template<class T, class D> class unique_ptr<T[], D>;
 
-  template<class T, class... Args> unique_ptr<T>
-    make_unique(Args&&... args);                                                // \tcode{T} is not array
-  template<class T> unique_ptr<T>
-    make_unique(size_t n);                                                      // \tcode{T} is \tcode{U[]}
+  template<class T, class... Args>
+    unique_ptr<T> make_unique(Args&&... args);                                  // \tcode{T} is not array
+  template<class T>
+    unique_ptr<T> make_unique(size_t n);                                        // \tcode{T} is \tcode{U[]}
   template<class T, class... Args>
     @\unspecnc@ make_unique(Args&&...) = delete;                                // \tcode{T} is \tcode{U[N]}
+
+  template<class T>
+    unique_ptr<T> make_unique_default_init();                                   // \tcode{T} is not array
+  template<class T>
+    unique_ptr<T> make_unique_default_init(size_t n);                           // \tcode{T} is \tcode{U[]}
+  template<class T, class... Args>
+    @\unspecnc@ make_unique_default_init(Args&&...) = delete;                   // \tcode{T} is \tcode{U[N]}
 
   template<class T, class D>
     void swap(unique_ptr<T, D>& x, unique_ptr<T, D>& y) noexcept;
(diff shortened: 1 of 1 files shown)


jensmaurer:
@burblebee, could you please fix the overfull \hbox here? Thanks.

burblebee:
@jensmaurer I think it's complaining about this text in utilities.tex around line 9811 because allocate_shared_default_init doesn't fit:

```
\item
  When a (sub)object of non-array type \tcode{U} is initialized by
  \tcode{make_shared_default_init} or \tcode{allocate_shared_default_init},
  it is initialized via the expression \tcode{::new(pv) U},
  where \tcode{pv} has type \tcode{void*} and
  points to storage suitable to hold an object of type \tcode{U}.
```

How to fix???

tkoeppe:
Feel free to add hyphenation hints (`\-`) into the code.

burblebee:
@tkoeppe I was able to fix motions-2018-11-lwg-18 by forcing a linebreak.  It's the cleanest solution wrt readability, and there is no resulting underflow.

source/utilities.tex
```diff
@@ -6705,6 +6712,16 @@
     shared_ptr<T> allocate_shared(const A& a, size_t N,
                                   const remove_extent_t<T>& u);                 // \tcode{T} is \tcode{U[]}
 
+  template<class T>
+    shared_ptr<T> make_shared_default_init();                                   // \tcode{T} is not \tcode{U[]}
```
jensmaurer: This entire block of changes needs to move further down, according to the ingress paper.

source/utilities.tex
```diff
@@ -6705,6 +6712,16 @@
...
+  template<class T>
+    shared_ptr<T> allocate_shared_default_init(const A& a);                     // \tcode{T} is not \tcode{U[]}
```
jensmaurer: "A" is not a declared identifier here. (This is a bug in the ingress paper.) The correct declaration is in [util.smartptr.shared.create].

source/utilities.tex
```diff
@@ -6705,6 +6712,16 @@
...
+
+  template<class T>
+    shared_ptr<T> make_shared_default_init(size_t N);                           // \tcode{T} is \tcode{U[]}
+  template<class T>
+    shared_ptr<T> allocate_shared_default_init(const A& a, size_t N);           // \tcode{T} is \tcode{U[]}
```
jensmaurer: Same here: "A" is not declared.

source/utilities.tex
```diff
@@ -9653,6 +9714,10 @@
   shared_ptr<T> make_shared(@\placeholdernc{args}@);
 template<class T, class A, ...>
   shared_ptr<T> allocate_shared(const A& a, @\placeholdernc{args}@);
+template<class T, ...>
+  shared_ptr<T> make_shared_default_init(@\placeholdernc{args}@);
+template<class T, class A, ...>
+  shared_ptr<T> allocate_shared_default_init(const A& a, @\placeholdernc{args}@);
```
jensmaurer: This seems broken (also in the ingress paper): The whole point of ..._default_init is to have no arguments, so why are we talking about "args" here? Also, the pack-like "..." in the template argument list should go.

jensmaurer: Ah, that's to cover "size_t N" for the array case. The not-really parallelism to make_shared(args) is highly misleading.

source/utilities.tex
burblebee: yup, thanks!

source/utilities.tex
burblebee: fixed

PR #2471, opened 2018-11-15
P0318R1 unwrap_ref_decay and unwrap_reference

### LINKED ISSUE #2419: [2018-11 LWG Motion 13] P0318R1 unwrap_ref_decay and unwrap_reference
[P0318R1](http://wg21.link/p0318r1)

jensmaurer:
Add the type trait as subclause [refwrap.unwrapref],
instead of as a sibling of [refwrap].
Rephrase the wording to avoid referencing X before it
is introduced.
Do not refer to 'equals' as referring to types.

Fixes #2419.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -852,14 +852,16 @@ This function shall not participate in overload resolution unless
 \indexlibrary{\idxcode{make_pair}}%
 \begin{itemdecl}
 template<class T1, class T2>
-  constexpr pair<V1, V2> make_pair(T1&& x, T2&& y);
+  constexpr pair<unwrap_ref_decay_t<T1>, unwrap_ref_decay_t<T2>> make_pair(T1&& x, T2&& y);
 \end{itemdecl}
 
 \begin{itemdescr}
-Let \tcode{Vi} be \tcode{unwrap_ref_decay_t<Ti>}.
-
 \pnum
-\returns \tcode{pair<V1, V2>(std::forward<T1>(x), std::forward<T2>(y))}.
+\returns
+\begin{codeblock}
+pair<unwrap_ref_decay_t<T1>,
+     unwrap_ref_decay_t<T2>>(std::forward<T1>(x), std::forward<T2>(y))
+\end{codeblock}
 \end{itemdescr}
 
 \pnum
@@ -997,7 +999,7 @@ namespace std {
   inline constexpr @\unspec@ ignore;
 
   template<class... TTypes>
-    constexpr tuple<VTypes...> make_tuple(TTypes&&...);
+    constexpr tuple<unwrap_ref_decay_t<TTypes>...> make_tuple(TTypes&&...);
 
   template<class... TTypes>
     constexpr tuple<TTypes&&...> forward_as_tuple(TTypes&&...) noexcept;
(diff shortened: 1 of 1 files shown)


jensmaurer:
Open issue: [refwrap.unwrapref] might need a cross-reference to the definition of "transformation type traits", called Cpp17TransformationTrait, in [meta.rqmts].

source/utilities.tex
```diff
@@ -854,12 +854,10 @@
...
 \begin{itemdescr}
+Let \tcode{Vi} be \tcode{unwrap_ref_decay_t<Ti>}.
+
 \pnum
-\returns \tcode{pair<V1, V2>(std::forward<T1>(x), std::forward<T2>(y))},
-where \tcode{V1} and \tcode{V2} are determined as follows: Let \tcode{Ui} be
-\tcode{decay_t<Ti>} for each \tcode{Ti}. If \tcode{Ui} is a specialization
-of \tcode{reference_wrapper}, then \tcode{Vi} is \tcode{Ui::type\&},
-otherwise \tcode{Vi} is \tcode{Ui}.
+\returns \tcode{pair<V1, V2>(std::forward<T1>(x), std::forward<T2>(y))}.
```
CaseyCarter: It would seem to simplify this spec if we use `unwrap_ref_decay_t<T1>` and `unwrap_ref_decay_t<T2>` directly here in the `\returns`, and drop both the previous "Let" paragraph and the following "where T_i is ..." paragraph.

source/utilities.tex
```diff
@@ -1601,10 +1599,7 @@
...
 \begin{itemdescr}
 \pnum
-The pack \tcode{VTypes} is defined as follows. Let \tcode{U}$_i$ be \tcode{decay_t<T$_i$>} for each
-\tcode{T}$_i$ in \tcode{TTypes}. If \tcode{U}$_i$ is a specialization of
-\tcode{reference_wrapper}, then \tcode{V}$_i$ in \tcode{VTypes} is \tcode{U$_i$::type\&},
-otherwise \tcode{V}$_i$ is \tcode{U}$_i$.
+Let \tcode{VTypes...} be \tcode{unwrap_ref_decay_t<TTypes>...}.
 
 \pnum
 \returns \tcode{tuple<VTypes...>(std::forward<TTypes>(t)...)}.
```
CaseyCarter: Similarly, the "detour" through the definition of `VTypes...` here only complicates this specification.

source/utilities.tex
jensmaurer: Done.

source/utilities.tex
jensmaurer: Done.

cpplearner:
These declarations use `V1`, `V2`, or `VTypes`. Should they be changed?

https://github.com/cplusplus/draft/blob/7e43fecde1e4a17b425ee2bfcdb22c22801a42a7/source/utilities.tex#L853
https://github.com/cplusplus/draft/blob/7e43fecde1e4a17b425ee2bfcdb22c22801a42a7/source/utilities.tex#L1000
https://github.com/cplusplus/draft/blob/7e43fecde1e4a17b425ee2bfcdb22c22801a42a7/source/utilities.tex#L1599

CaseyCarter:
> These declarations use V1, V2, or VTypes. Should they be changed?

Corrected in #2490.

jensmaurer:
@cpplearner: Yes, we should not use (now undefined) variables. Fixed.

PR #1933, opened 2018-02-18
[containers] Remove class name repeated in subheadings

jensmaurer:
Partially addresses #1242.

PR #2475, opened 2018-11-15
[array.syn] Add reference to [array.tuple]

johelegp:
I retain the heading removed in #1933, since fixes to #1242 don't change them in the synopses.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -3011,6 +3011,7 @@ namespace std {
   template<class T, size_t N>
     void swap(array<T, N>& x, array<T, N>& y) noexcept(noexcept(x.swap(y)));
 
+  // \ref{array.tuple}, tuple interface to class template \tcode{array}
   template<class T> class tuple_size;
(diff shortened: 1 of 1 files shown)


jensmaurer:
We don't have a cross-reference for "swap", either, so adding one for [array.tuple] makes the presentation inconsistent. (We do have a cross-reference to the superordinate subclause at the start of the header.)
@zygoloid, which way do you want to go here? Add detailed cross-references in synopses whenever we have subclause headings? Or don't bother for small headers?

johelegp:
You are right. I must have done this because pair does it, too.

jensmaurer:
Editorial meeting:
- Cross-references are useful; cases that are missing are bugs.

johelegp:
Does that include cross-references to the containing subclauses that group cross-references? For example, [concepts.syn] and [algorithms.syn] first cross-reference the containing subclauses of the described concepts/algorithms. This is applied inconsistently.

Issue #2486, opened 2018-11-16
[func.bind_front] use unwrap_ref_t for bind_front


jensmaurer:
instead of the meta-function DECAY_UNWRAP.

See P0318, applied via #2471.

zygoloid:
Also: check for other places where this can be applied. (None known as yet.)

jensmaurer:
Can't find UNWRAP anywhere else.

PR #2487, opened 2018-11-16
P0356R5 Simplified partial function application

### LINKED ISSUE #2432: [2018-11 LWG Motion 26] P0356R5 Simplified partial function application
[P0356R5](http://wg21.link/p0356r5)

jensmaurer:
[support.limits.general] Introduce feature-test macro even though it
  was not presented as a wording diff in the paper.
[functional.syn] Use proper subclause name in cross-reference.
[func.bind_front] Do not use 'equals' when discussing types identity.

Fixes #2432.

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -555,6 +555,8 @@ the values of these macros with greater values.
   \tcode{<atomic>} \\ \rowsep
 \defnlibxname{cpp_lib_bit_cast}                             & \tcode{201806L} &
   \tcode{<bit>} \\ \rowsep
+\defnlibxname{cpp_lib_bind_front}                           & \tcode{201811L} &
+  \tcode{<functional>} \\ \rowsep
 \defnlibxname{cpp_lib_bool_constant}                        & \tcode{201505L} &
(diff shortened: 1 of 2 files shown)


jensmaurer:
Created #2486 to apply unwrap_ref_decay from [P0318](http://wg21.link/p0318).

jensmaurer:
Even after the editorial repairs, the wording continues to talk about an "argument pack" where there is none; there are simply the usual arguments of a function call. Also, LWG has added a definition for "direct-non-list-initialized", so there is no motivation to go through the "initializer" route.
@zygoloid, you need to decide whether to attempt to rescue this editorially or whether to flat-out reject application of the paper. (This is also a cost/effort consideration, given that the "ranges" paper is still on the plate.)

zygoloid:
I think we should apply this and file library issues for the wording problems. Even if we interpret the spec as I believe we were supposed to, it's utterly unimplementable (perfect forwarding is not as perfect as this wording requires). However, I think both a library vendor and a user could read this specification and get the right idea as to what is being specified.

PR #2490, opened 2018-11-18
[utility.syn,pairs.spec,tuple.syn,tuple.creation] Use unwrap_ref_deca…

CaseyCarter:
…y consistently

...in declarations and definitions of make_pair and make_tuple.


jensmaurer:
Mooted by additional fix-up for #2471.

PR #2506, opened 2018-11-26
[depr.locale.category] Replace tcode environments with codeblock environments.

Eelis:
tcode is not used as an environment anywhere else, so I doubt it was really intended here.


Eelis:
I see the code was added by 548ce7948c6162381536dcbd3b882cc760dd7a33, so maybe @jensmaurer can comment on the intent.

PR #2517, opened 2018-11-26
[ranges] Shorten and adjust stable labels

### LINKED ISSUE #2513: revise section labels introduced by ranges
In particular, [range] should be [ranges] for consistency with the rest of the library. There are probably other opportunities for cleanups here.

jensmaurer:
Fixes #2513.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -8,7 +8,7 @@ This Clause describes components that \Cpp{} programs may use to perform
 iterations over containers\iref{containers},
 streams\iref{iostream.format},
 stream buffers\iref{stream.buffers},
-and other ranges\iref{range}.
+and other ranges\iref{ranges}.
 
 \pnum
 The following subclauses describe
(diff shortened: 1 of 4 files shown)


jensmaurer:
@CaseyCarter, I'll try [common.alg.req.ind.cmp] etc.

jensmaurer:
Regarding [iterator.concept.weaklyincrementable]: Regrettably, none of the other stable labels in the vicinity abbreviate "iterator" as "iter". I've abbreviate the stuff at the end a bit.

source/iterators.tex
```diff
@@ -93,19 +93,19 @@
...
+  // \ref{iterator.concept.inc}, concept \tcode{Incrementable}
   template<class I>
     concept Incrementable = @\seebelow@;
 
   // \ref{iterator.concept.iterator}, concept \tcode{Iterator}
   template<class I>
     concept Iterator = @\seebelow@;
 
-  // \ref{iterator.concept.incrementable}, concept \tcode{Sentinel}
+  // \ref{iterator.concept.sentinel}, concept \tcode{Sentinel}
```
CaseyCarter: Good catch.

CaseyCarter:
> Do we even need "common" in front? [alg.req...] seems good enough.

[alg.req] could be easily confused with [algorithms.requirements], but I'm not concerned that this will be a problem in practice since [alg.req] contains only subclauses and [algorithms.requirements] contains only paragraphs.

Thanks for the cleanup.

PR #2519, opened 2018-11-26
[structure.summary] Add missing templated entities

### LINKED ISSUE #2505: [structure.summary] Unaccounted templated entities
The list in [[structure.summary]/2](https://wg21.link/structure.summary#2) is missing alias templates and variable templates. The paragraph above it suggests that they need to be on the list since header synopses list those templated entities. There's also an alias template with "detailed specification" in [[intseq.make]](https://wg21.link/intseq.make). Should [[structure.specifications]](https://wg21.link/structure.specifications) take that into account?

jensmaurer:
Fixes #2505.

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -459,10 +459,10 @@ The contents of the summary and the detailed specifications include:
 \begin{itemize}
 \item macros
 \item values
-\item types
+\item types and alias templates
 \item classes and class templates
 \item functions and function templates
-\item objects
+\item objects and variable templates
 \item concepts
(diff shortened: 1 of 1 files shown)


source/lib-intro.tex
```diff
@@ -459,10 +459,10 @@
 \begin{itemize}
 \item macros
 \item values
-\item types
+\item types and alias templates
 \item classes and class templates
 \item functions and function templates
-\item objects
+\item objects and variable templates
```
zygoloid: Do we really not have any global references? :)

PR #2531, opened 2018-11-27
[algorithms] Use math \min when computing complexities.

### LINKED ISSUE #2385: \tcode{min} vs. $\min$
The algorithm specifications - and potentially the rest of the library - use both forms for min and max. For example, `\complexity At most \tcode{min(last1 - first1, last2 - first2)} applications...` ([mismatch]) and `\complexity ... \bigoh{\min(\tcode{last1 - first1}, \tcode{last2 - first2})} applications...` ([alg.equal]). To me, it seems we should be using the "mathematical function" form for all cases that aren't actual code.

jensmaurer:
\tcode{min} is only appropriate when referring to the C++
function std::min.

Fixes #2385.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -3390,7 +3390,7 @@ if no such iterator is found.
 \pnum
 \complexity
 For the overloads with no \tcode{ExecutionPolicy}, exactly
-\tcode{min((i - first) + 1, (last - first) - 1)}
+\[ \min(\tcode{(i - first) + 1}, \ \tcode{(last - first) - 1}) \]
 applications of the corresponding predicate, where \tcode{i} is
 \tcode{adjacent_find}'s
 return value.  For the overloads with an \tcode{ExecutionPolicy},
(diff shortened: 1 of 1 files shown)


source/algorithms.tex
```diff
@@ -3390,7 +3390,7 @@
 \pnum
 \complexity
 For the overloads with no \tcode{ExecutionPolicy}, exactly
-\tcode{min((i - first) + 1, (last - first) - 1)}
+\[ \min(\tcode{(i - first) + 1}, \ \tcode{(last - first) - 1}) \]
```
tkoeppe: Are you sure we need the extra space after the comma? We don't generally do this at the moment (except once?); is it necessary?

jensmaurer: It looks much better that way; otherwise we have a wide space between "last - first" and a very small space after the comma.

Before:
before

After:
after

tkoeppe: OK, fair enough, thanks!

PR #2533, opened 2018-11-27
[cmp.categories] Replace 'operator admits' phrasing.

### LINKED ISSUE #2357: Wonky wording: operators that "admit" other operators?
E.g. in [[cmp.weakord]/1](http://eel.is/c++draft/cmp.weakord#1.sentence-1):
> The `weak_­ordering` type is typically used as the result type of a three-way comparison operator that (a) admits all of the six two-way comparison operators ([expr.rel], [expr.eq]), and (b) does not imply substitutability.

What does it mean for an operator to "admit" another operator? I can't find a definition of it anywhere. Is there maybe a better way to phrase what is intended here? :)

jensmaurer:
Fixes #2357.

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -4277,10 +4277,10 @@ that is accessible via the argument's public const members.
 \pnum
 The \tcode{partial_ordering} type is typically used
 as the result type of a three-way comparison operator\iref{expr.spaceship}
-that (a) admits all of the six
-two-way comparison operators\iref{expr.rel,expr.eq},
-(b) does not imply substitutability,
-and (c) permits two values to be incomparable.
+for a type that admits
+all of the six two-way comparison operators\iref{expr.rel,expr.eq},
+for which equality need not imply substitutability,
+and that permits two values to be incomparable.%
 \begin{footnote}
 That is, \tcode{a < b}, \tcode{a == b}, and \tcode{a > b} might all be \tcode{false}.
 \end{footnote}
@@ -4395,9 +4395,9 @@ constexpr partial_ordering operator<=>(@\unspec@, partial_ordering v) noexcept;
 \pnum
 The \tcode{weak_ordering} type is typically used
 as the result type of a three-way comparison operator\iref{expr.spaceship}
-that (a) admits all of the six
-two-way comparison operators\iref{expr.rel,expr.eq},
-and (b) does not imply substitutability.
+for a type that admits
+all of the six two-way comparison operators\iref{expr.rel,expr.eq}
+and for which equality need not imply substitutability.
 
 \indexlibraryglobal{weak_ordering}%
 \indexlibrarymember{less}{weak_ordering}%
@@ -4520,9 +4520,9 @@ constexpr weak_ordering operator<=>(@\unspec@, weak_ordering v) noexcept;
 \pnum
 The \tcode{strong_ordering} type is typically used
 as the result type of a three-way comparison operator\iref{expr.spaceship}
-that (a) admits all of the six
-two-way comparison operators\iref{expr.rel,expr.eq},
-and (b) does imply substitutability.
+for a type that admits
+all of the six two-way comparison operators\iref{expr.rel,expr.eq}
+and for which equality does imply substitutability.
 
 \indexlibraryglobal{strong_ordering}%
 \indexlibrarymember{less}{strong_ordering}%


source/support.tex
```diff
@@ -4283,9 +4283,10 @@
 \pnum
 The \tcode{partial_ordering} type is typically used
 as the result type of a three-way comparison operator\iref{expr.spaceship}
-that (a) admits all of the six two-way comparison operators (\ref{expr.rel}, \ref{expr.eq}),
-(b) does not imply substitutability,
-and (c) permits two values to be incomparable.%
+for a type that is intended to support
+all of the six two-way comparison operators (\ref{expr.rel}, \ref{expr.eq}),
```
Eelis: I think the space before `(\ref` should normally be a `~`.

W-E-Brown:
Sorry, but I fail to see any issue here worth fixing.  My English dictionary defines "admit" as "To be capable <https://en.wiktionary.org/wiki/capable> of; to permit <https://en.wiktionary.org/wiki/permit>", and that makes perfect sense to me in the given context.  Why do we need further definition?

By the way, the proposed rewording introduces some notion of intention, for which I can find no justification in this context.

In short, I'm at a loss to understand what's wrong with the plain English reading of what's in the Draft.

jensmaurer:
@W-E-Brown, the concern here is that "an operator" admits something, but it seems we should talk about the type having certain properties here, not the operator.

source/support.tex

jensmaurer:
Editorial meeting: Postpone due to concurrent LWG work in this area.

source/support.tex
```diff
@@ -4238,9 +4238,10 @@
 \pnum
 The \tcode{partial_ordering} type is typically used
 as the result type of a three-way comparison operator\iref{expr.spaceship}
-that (a) admits all of the six two-way comparison operators (\ref{expr.rel}, \ref{expr.eq}),
-(b) does not imply substitutability,
-and (c) permits two values to be incomparable.
+that is intended to support
```
jensmaurer, line 4241: add "for a type", consistent with the rest

jensmaurer:
Also drop "intent" if possible.

source/support.tex
```diff
@@ -4395,9 +4395,9 @@
 \pnum
 The \tcode{weak_ordering} type is typically used
 as the result type of a three-way comparison operator\iref{expr.spaceship}
-that (a) admits all of the six
-two-way comparison operators\iref{expr.rel,expr.eq},
-and (b) does not imply substitutability.
+for a type that admits
+all of the six two-way comparison operators\iref{expr.rel,expr.eq}
+and for which equality need not imply substitutability,
```
CaseyCarter, line 4400:
```suggestion
and for which equality need not imply substitutability.
```

source/support.tex
```diff
@@ -4520,9 +4520,9 @@
 \pnum
 The \tcode{strong_ordering} type is typically used
 as the result type of a three-way comparison operator\iref{expr.spaceship}
-that (a) admits all of the six
-two-way comparison operators\iref{expr.rel,expr.eq},
-and (b) does imply substitutability.
+for a type that admits
+all of the six two-way comparison operators\iref{expr.rel,expr.eq}
+and for which equality does imply substitutability,
```
CaseyCarter, line 4525:
```suggestion
and for which equality does imply substitutability.
```

CaseyCarter (changes requested):

source/support.tex
jensmaurer: fixed

CaseyCarter (approved):

source/support.tex
CaseyCarter: (You need to resolve these comments; only PR authors and folks with commit authority can resolve. Yes, that's right, Github doesn't allow comment authors to resolve their own comments.)

jensmaurer: Yes, but it seems more friendly to the comment author to have him review the unresolved comment back-and-forth before hiding it from view.

jensmaurer:
@tkoeppe, what about this one?

source/support.tex
CaseyCarter: That does make sense given that a great many of our commentors are not full-time GitHub residents.

source/support.tex
tkoeppe: obsolete

PR #2540, opened 2018-11-27
[dcl.spec] Harmonize phrasing of restrictions on decl-specifiers

### LINKED ISSUE #2199: [dcl] Large variation in phrasing of restrictions for decl-specifiers
[dcl.stc] p3 "It shall be applied only to the names of variables of namespace or block scope and to the names of static data members."
(Note "shall", "and", "names")

[dcl.stc] p5 "The extern specifier can be applied only to the names of variables and functions."
(Note "can", "and", "names")

[dcl.constexpr] p1 "The constexpr specifier shall be applied only to the definition of a variable or variable template or the declaration of a function or function template."
(Note "shall", "or", "definition/declaration")

[dcl.inline] p1 "The inline specifier can be applied only to the declaration or definition of a variable or function."
(Note "can", "or", "declaration or definition")

We should clean up the mess. Suggestion: use "shall", "or" ,and "declaration" / "definition" as appropriate (but only one of them), plus use singular.

jensmaurer:
Fixes #2199.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -370,8 +370,8 @@ has automatic storage duration by default\iref{basic.stc.auto}.
 The \tcode{thread_local} specifier
 indicates that the named entity has thread storage duration\iref{basic.stc.thread}. It
 shall be applied only
-to the names of variables of namespace
-or block scope and to the names of static data members.
+to the declaration of a variable of namespace
+or block scope or to the declaration of a static data member.
 When \tcode{thread_local} is applied to a variable of block scope the
 \grammarterm{storage-class-specifier} \tcode{static} is implied if no other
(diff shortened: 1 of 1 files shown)


source/declarations.tex
```diff
@@ -993,7 +993,7 @@
 \indextext{specifier!\idxcode{inline}}
 
 \pnum
-The \tcode{inline} specifier can be applied only to the declaration
+The \tcode{inline} specifier shall be applied only to the declaration
 or definition of a variable or function.
```
zygoloid: For consistency with the above cases, and for simplicity, remove the "or definition" here?

jensmaurer: Fixed, also in [dcl.align].

Issue #2483, opened 2018-11-16
[basic.fundamental] boolean note nitpick


jfbastien:
Booleans can behave as neither true nor false, but can also behave as both true and false.
Suggested wording here: https://github.com/cplusplus/draft/pull/2438#discussion_r232839704

jensmaurer:
The suggestion is to add "or might cause it to behave as if it is both \tcode{true} and \tcode{false}" to the corresponding footnote.

Opinion: I don't think attempting to inject more precision in a non-normative comment on possible results of undefined behavior is worthwhile. Actually, removing that footnote would be fine, given that the Undefined Behavior dragon is already out of the box here.

jfbastien:
The Undefined Behavior Dragon might or might not be out of the box. It's up to the Dragon to decide whether it'll come out of the box, and whether you'll be allowed to observe it come out of the box.

Agreed we can just remove the note.

PR #2542, opened 2018-11-29
undefined boolean note clarification

jfbastien:
Addresses #2483.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -4364,11 +4364,7 @@ value representation, and
 alignment requirements as
 an \impldef{underlying type of \tcode{bool}} unsigned integer type.
 The values of type \tcode{bool} are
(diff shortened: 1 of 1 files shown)


jensmaurer:
The commit comment should have a [section.label] prefix.

Other than that, I'm not sure what this footnote delivers us in the first place: Reading uninitialized data is undefined behavior, and attempting to list the shades and variations of undefined behavior leads nowhere at best and is seriously misleading otherwise. So, let's remove that footnote altogether.

See [dcl.init] p12 for the detailed treatise. I'd actually be in favor of moving most of this paragraph into [expr.prop] as a new subclause "Indeterminate values". @zygoloid, what do you think?

jfbastien:
The note is now gone. 👋

PR #2546, opened 2018-12-02
[view.interface] Use "inherits"/"derives" consistently.

Eelis:


Eelis:
Hmm, actually it might be even better to say "that is derived from it".

jensmaurer:
"that is derived from it" sounds better, and is probably clearer, given the curious recursive template pattern here.

Issue #2512, opened 2018-11-26
[lib] always use sub-namespace in itemdecls


zygoloid:
For declarations in sub-namespaces, particularly within `std::ranges`, our itemdecls are often very unclear about which declaration they're referring to. For example, the only clue that the `advance` described in [iterator.operations] is in a different namespace from the `advance` described in [range.iterator.operations.advance] is the subclause name and how it's cross-referenced from the synopsis. That's not good enough.

We should consistently use a `ranges::` prefix on the *declarator-id* for entities in namespace `std::ranges` (and likewise for other sub-namespaces of `std`).

jensmaurer:
@zygoloid, this is not a small job. Are we targeting the post-San Diego mailing with this?

johelegp:
Should apply to some algorithms in `utilities.tex` within `ranges::`, like `uninitialized_copy`.

PR #2547, opened 2018-12-02
[algorithms] Compactify and reformat source with no visual change

### LINKED ISSUE #2509: clean up [algorithms] source code
From @jensmaurer in #2501:

> The "algorithms" clause still has a lot of troff-era formatting where we start a new line for each non-prose item, such as \tcode{first}. And we have a lot of these. This makes for very long source code. Suggestion: Between the mailings, have me reformat the source code in a no-visual-changes diff.

jensmaurer:
except:
Move \begin{note} and \end{note} to lines of its own.
[alg.unique] fix two English-level typos
[alg.transform] remove superfluous space before period
[alg.partitions] remove duplicate \pnum
[alg.is_permutation,alg.move,is.sorted] add missing \tcode
[algorithms.parallel.exec] move comment in example

Fixes #2509.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -25,18 +25,19 @@ as summarized in \tref{algorithms.summary}.
 
 \rSec1[algorithms.requirements]{Algorithms requirements}
 \pnum
-All of the algorithms are separated from the particular implementations of data structures and are
-parameterized by iterator types.
-Because of this, they can work with program-defined data structures, as long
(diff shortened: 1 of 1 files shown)


jensmaurer:
@tkoeppe, @zygoloid: It would be good if we could apply this soon-ish, because I also need to replace the "namespace ranges { ... }" in the itemdecls with qualified declarator-ids as per #2512, which has (of course) high conflict potential.

jensmaurer:
Rebased.

PR #2549, opened 2018-12-03
[meta.logical] use math mode subscripts for Bi, B1, BN etc.

jwakely:


source/utilities.tex
```diff
@@ -18337,32 +18337,32 @@
 forms the logical conjunction of its template type arguments.
 
 \pnum
-For a specialization \tcode{conjunction<B1, ..., BN>},
-if there is a template type argument \tcode{Bi} for which \tcode{bool(Bi::value)} is \tcode{false},
-then instantiating \tcode{conjunction<B1, ..., BN>::value}
-does not require the instantiation of \tcode{Bj::value} for \tcode{j > i}.
+For a specialization $\tcode{conjunction<B}_1\tcode{,} \cdots\tcode{, B}_N\tcode{>}$,
```
tkoeppe: Yikes, this nesting is a bit horrific. How about:

    \tcode{conjuntion<$\tcode{B}_{1}$, $\dotsc$, $\tcode{B}_{N}$>}

(I suppose code commas are OK here.)

jwakely:
diff

source/utilities.tex
jwakely: Yes, it's horrific. That looks like an improvement, I'll try it.

tkoeppe: I think baseline dots are OK here (ldots or dotsc). I don't think there's any confusion with pack expansion here. I don't feel strongly, but I don't think that we need to go out of our (typographic) way here to prevent a misunderstanding.

source/utilities.tex
```diff
@@ -18337,32 +18337,32 @@
...
+if there is a template type argument $\tcode{B}_i$ for which $\tcode{bool(B}_i\tcode{::value)}$ is \tcode{false},
+then instantiating $\tcode{conjunction<B}_1\tcode{,} \cdots\tcode{, B}_N\tcode{>::value}$
+does not require the instantiation of $\tcode{B}_j\tcode{::value}$ for $j > i$.
```
tkoeppe: This is somewhat unrelated, but I would feel cosier if we stated the latter range as `$i < j \leq N$`. It just leaves no room for someone saying that $j = 5000$ isn't defined.

jwakely:
I've pushed a new commit using @tkoeppe's suggested nesting of `tcode` and math mode. The diffpdf result is identical to the one shown above, except for using `dotsc` instead of `cdots` (which does look better).

Issue #2548, opened 2018-12-03
[meta.logical] unclear wording


FrankHB:
In [meta.logical]/2
> then instantiating `conjunction<B1, ..., BN>::value` does not require the instantiation of `Bj::value` for `j > i`

What is "the instantiation of `Bj::value`"?

(The font of `j` should also be fixed.)

The change in #1193 (LWG 2567) also introduce another (possibly non-editorial) problem: does the explicit cast to `bool` of `value` mandated? For example:

```
#include <type_traits>

struct B
{
  constexpr explicit operator bool() const noexcept {return true;}
};

struct C
{
  static constexpr const B value = B();
};

int main()
{
  static_assert(bool(std::conjunction<C, std::true_type>::value));
}

```
Is the above program well-formed?

It seems intentionally ill-formed as per "shall have a member value which is convertible to `bool`" wording in [meta.logical]/4. (Is this presumely "implicit", or like `std::is_convertible`?) Then `bool(Bj::value)` is not appropriate as it suggests the cast shall be supported, as [implemented in cppreference](https://en.cppreference.com/w/cpp/types/conjunction).

Otherwise, libstdc++, libc++ and MSVC currently all has the bug not supporting resoultion of LWG 2567/2568, and [meta.logical]/4 should be reworded to clarify the intention. I guess this is not intentional because a) it is not proposed at first; b) the context here is more like operands of `&&` but not `if`, which needs no contextually conversion, so omission of explicit cast should not be encouraged.

jwakely:
> What is "the instantiation of Bj::value"?

It seems clear to me.

Maybe it would be more precise to say "does not require the definition of the static data member `Bj::value` to exist" with a cross-reference to [temp.inst].

jwakely:
> (The font of j should also be fixed.)

I'm not sure what you mean by "fixed" but I've submitted #2549 to use math mode for the subscripts _i_, _j_, _1_ and _N_.

> The change in #1133 (LWG 2567) also introduces another (possibly non-editorial) problem: does the explicit cast to bool of value mandated?

This seems non-editorial to me.

> Is the program above well-formed?

It's undefined, because it fails to meet the _Requires:_ precondition.

> Then bool(Bj::value) is not appropriate as it suggests the cast shall be supported, as implemented in cppreference.

The cast is supported, but it's not required. If an implementation wants to rely on an implicit conversion instead that's fine (as long as it does so without a narrowing conversion, as per lwg 2587).

[...]

PR #2550, opened 2018-12-03
[meta.logical] clarify short-circuiting property

jwakely:
This implements the clarification I suggested in https://github.com/cplusplus/draft/issues/2548#issuecomment-443689234 (because if `Bj` is not a template then talking about "instantiation of `Bj::value" doesn't make sense) and the improvement to the inequality that Thomas suggested at https://github.com/cplusplus/draft/pull/2549#discussion_r238262056

It also adds an example showing a consequence of this property.

### DIFF
--- a/source/meta.tex
+++ b/source/meta.tex
@@ -2186,7 +2186,7 @@ The class template \tcode{conjunction}
 forms the logical conjunction of its template type arguments.
 
 \pnum
-For the specialization \tcode{disjunction<>},
+For the specialization \tcode{conjunction<>},
 let $\tcode{B}_{i}$ be \tcode{true_type}.
 For a specialization \tcode{conjunction<$\tcode{B}_{1}$, $\dotsc$, $\tcode{B}_{N}$>},
 let $\tcode{B}_{i}$ be
--- a/source/meta.tex
+++ b/source/meta.tex
@@ -2194,6 +2194,23 @@ to exist\iref{temp.inst} for $i < j \leq N$.
 This is analogous to the short-circuiting behavior of
 the built-in operator \tcode{\&\&}.
 \end{note}
+\begin{note}
+This allows types without a \tcode{value} static data member
+to be template arguments for \tcode{conjunction}, as long as an
+earlier argument evaluates to \tcode{false}, for example:
+\begin{codeblock}
+template<class T>
+  concept HasValue = requires { { T::value } -> std::convertible_to<bool>; };
+template<class T>
+  using maybe_value = std::conjunction<std::bool_constant<HasValue<T>>, T>;
+
+struct X { };
+
+static_assert( ! maybe_value<X>::value );       // OK, \tcode{X::value} not used
+static_assert( ! maybe_value<int>::value );     // OK, \tcode{int::value} not used
+static_assert( maybe_value<std::true_type>::value );
+\end{codeblock}
+\end{note}
 
 \pnum
 Every template type argument
--- a/source/meta.tex
+++ b/source/meta.tex
@@ -2188,7 +2188,7 @@ For a specialization \tcode{conjunction<$\tcode{B}_{1}$, $\dotsc$, $\tcode{B}_{N
 if there is a template type argument $\tcode{B}_{i}$
 for which \tcode{bool($\tcode{B}_{i}$::value)} is \tcode{false},
 then instantiating \tcode{conjunction<$\tcode{B}_{1}$, $\dotsc$, $\tcode{B}_{N}$>::value}
-does not require the instantiation of \tcode{$\tcode{B}_{j}$::value} for $j > i$.
+does not require the instantiation of \tcode{$\tcode{B}_{j}$::value} for $i < j \leq N$.
 \begin{note}
 This is analogous to the short-circuiting behavior of
 the built-in operator \tcode{\&\&}.
@@ -2240,7 +2240,7 @@ For a specialization \tcode{disjunction<$\tcode{B}_{1}$, $\dotsc$, $\tcode{B}_{N
 if there is a template type argument $\tcode{B}_{i}$
 for which \tcode{bool($\tcode{B}_{i}$::value)} is \tcode{true},
 then instantiating \tcode{disjunction<$\tcode{B}_{1}$, $\dotsc$, $\tcode{B}_{N}$>::value}
-does not require the instantiation of \tcode{$\tcode{B}_{j}$::value} for $j > i$.
+does not require the instantiation of \tcode{$\tcode{B}_{j}$::value} for $i < j \leq N$.
 \begin{note}
 This is analogous to the short-circuiting behavior of
 the built-in operator \tcode{||}.
--- a/source/meta.tex
+++ b/source/meta.tex
@@ -374,6 +374,8 @@ namespace std {
   template<class... B> struct conjunction;
   template<class... B> struct disjunction;
   template<class B> struct negation;
+  template<> struct conjunction : true_type { };
+  template<> struct disjunction : false_type { };
 
   // \ref{meta.unary.cat}, primary type categories
   template<class T>
@@ -2184,12 +2186,27 @@ The class template \tcode{conjunction}
 forms the logical conjunction of its template type arguments.
 
 \pnum
+For the specialization \tcode{disjunction<>},
+let $\tcode{B}_{i}$ be \tcode{true_type}.
 For a specialization \tcode{conjunction<$\tcode{B}_{1}$, $\dotsc$, $\tcode{B}_{N}$>},
-if there is a template type argument $\tcode{B}_{i}$
-for which \tcode{bool($\tcode{B}_{i}$::value)} is \tcode{false},
-then instantiating \tcode{conjunction<$\tcode{B}_{1}$, $\dotsc$, $\tcode{B}_{N}$>::value}
-does not require a definition of \tcode{$\tcode{B}_{j}$::value}
-to exist\iref{temp.inst} for $i < j \leq N$.
+let $\tcode{B}_{i}$ be
+the first type in \tcode{$\tcode{B}_{1}$, $\dotsc$, $\tcode{B}_{N}$}
+for which \tcode{$\tcode{B}_{i}$::value} is \tcode{false},
+or $\tcode{B}_{N}$ if there is no such type.
+
+\pnum
+The specialization
+\tcode{conjunction<$\tcode{B}_{1}$, $\dotsc$, $\tcode{B}_{N}$>}
+has a public and unambiguous base class of type $\tcode{B}_{i}$.
+\begin{note}
+This means a specialization of \tcode{conjunction}
+does not necessarily inherit from
+either \tcode{true_type} or \tcode{false_type}.
+\end{note}
+
+\pnum
+Instantiation of \tcode{$\tcode{B}_{j}$::value} is required for $1 \leq j \leq i$.
+Instantiation of \tcode{$\tcode{B}_{k}$::value} is not required for $i < k \leq N$.
 \begin{note}
 This is analogous to the short-circuiting behavior of
 the built-in operator \tcode{\&\&}.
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -18341,10 +18341,25 @@
...
 then instantiating \tcode{conjunction<$\tcode{B}_{1}$, $\dotsc$, $\tcode{B}_{N}$>::value}
-does not require the instantiation of \tcode{$\tcode{B}_{j}$::value} for $j > i$.
+does not require a definition of \tcode{$\tcode{B}_{j}$::value} to exist\iref{temp.inst} for $i < j \leq N$.
 \begin{note} This is analogous to the short-circuiting behavior of
 the built-in operator \tcode{\&\&}.
 \end{note}
+\begin{example}
+This allows types without a \tcode{value} static data member
+to be template arguments for \tcode{conjunction}, as long as an
+earlier argument evaluates to \tcode{false}, e.g.
```
tkoeppe: And this with ", for example:". Otherwise you have a dangling sentence.

tkoeppe: Or maybe even  turn the example into a note and turn the example-in-the-example into an example? And then merge the note with the previous note?

source/utilities.tex
```diff
@@ -18341,10 +18341,25 @@
...
+\begin{codeblock}
+template<typename T>
+  concept HasValue = requires { { T::value } -> bool; };
+template<typename T>
+  using maybe_value = std::conjunction<std::bool_constant<HasValue<T>>, T>;
+struct X { };
```
tkoeppe: newline before and after, to separate setup from execution

source/utilities.tex
jwakely: I'd prefer to keep it separate from the previous Note. The run-time short-circuiting of `&&` doesn't allow you to have invalid expressions on the RHS or functions that aren't defined, which is the property being demonstrated by the example. It's analogous, but not actually the same.

jwakely: Change "e.g." to "for example:" and changed the Example to a second Note.

source/utilities.tex
```diff
@@ -18341,10 +18341,27 @@
 if there is a template type argument $\tcode{B}_{i}$
 for which \tcode{bool($\tcode{B}_{i}$::value)} is \tcode{false},
 then instantiating \tcode{conjunction<$\tcode{B}_{1}$, $\dotsc$, $\tcode{B}_{N}$>::value}
-does not require the instantiation of \tcode{$\tcode{B}_{j}$::value} for $j > i$.
+does not require a definition of \tcode{$\tcode{B}_{j}$::value} to exist\iref{temp.inst} for $i < j \leq N$.
```
zygoloid: As a reader of this, I'm left wondering if a declaration is required to exist.

source/utilities.tex
```diff
@@ -18341,10 +18341,27 @@
...
+struct X { };
+
+static_assert( ! maybe_value<X>::value );      // OK, \tcode{X::value} not used
+static_assert( ! maybe_value<int>::value );    // OK, \tcode{int::value} not used
+static_assert( maybe_value<std::true_type>::value );
+\end{codeblock}
+\end{note}
 
 \pnum
 Every template type argument
```
zygoloid: GitHub isn't letting me comment on it, but there's a "for which instantiation is required" in here that should also be updated.

source/utilities.tex
zygoloid: ... and I think fixing that takes us out of editorial country. :(

CaseyCarter: > (For LWG: is an implementation allowed to require later B_js to be complete types?)

This sentence is partially redundant with the blanket wording in http://eel.is/c++draft/meta#rqmts-5 that allows the templates in [meta] to be instantiated with arguments that are incomplete types. It might be clearer to specify this as "If there is a template type argument `B_i` for which `B_i::value` is `false`, then `B_j` must be a complete type for `0 <= j <= i`. [_Note:_ `B_j` can be an incomplete type for `i < j <= N`.-_end note_]"

timsong-cpp: > If there is a template type argument `B_i` for which `B_i::value` is `false`, then `B_j` must be a complete type for `0 <= j <= i`.

That's wrong. We want `conjunction_v<falsy, incomplete-or-hard-error, falsy>` to be valid.

jwakely: Ah yes, we need to only talk about the first `Bi` for which `Bi::value` is `false`.

CaseyCarter: Something like:

> For a specialization `conjunction<B_1 , ..., B_N>`, let `B_i` be the first type in `B` for which `B_i::value` is `false`, or `N` if there is no such type in `B`.  `B_j` shall be a complete type for `0 <= j <= i`. [*Note: ...

jensmaurer:
@jwakely: Ping.

jensmaurer:
@jwakely: Ping^2.

source/utilities.tex
jwakely: Can we just say "the first type in `B`" like that, or does it need to be "the first type in `B_1` ... `B_N`"?

source/utilities.tex
jwakely: Sorry, I'm not sure what needs to be updated, or how.

source/utilities.tex
jwakely: How about this?

conjunction

jwakely: Or with the paragraph defining the base class moved earlier, and the bogus "shalls" in the last paragraph replaced (because that's a requirement on the implementation of `conjunction`):

conjunction

I'm not very happy with the normative wording in paragraph 5 though.

jwakely:
See the screengrab at https://github.com/cplusplus/draft/pull/2550#discussion_r287990057 for the proposed rewrite. I haven't pushed that as a commit yet, as I'd like feedback first.

source/utilities.tex
jensmaurer: Yes, the "so" is too much explanation as opposed to specification.

jwakely: Also it should be `1 <= j <= i` in both cases, not `0 <= j <= i`.

How about just replacing "so" with a period, making it two normative sentences?

And maybe reordering it as:

Bj shall be a complete type for 1 <= j <= i. Instantiation of Bj::value is required for 1 <= j <= i. Instantiation of Bj::value is not required for i < j <= N.

CaseyCarter: > Also it should be `1 <= j <= i` in both cases, not `0 <= j <= i`.

I think the `0` managed to sneak in because I was thinking about defining `B_0` to be `true_type`. We've notably lost the fact that `conjunction<>` derives from `true_type` in the recent formulations. I suggest we replace para 3 with:

> For a specialization `conjunction<B_1, ... , B_N>`, let `B_0` be `true_type` and `B_i` be the first type in `B_0, ... , B_N` for which `bool(B_i::value)` is `false` or `B_N` if there is no such type.

It then notably doesn't matter of we use `0` or `1` as the lower bound for the requirements in the following paragraphs since `true_type` certainly meets all such requirements.

> How about just replacing "so" with a period, making it two normative sentences?

"`foo` shall be a complete type" is a necessary condition when "Instantiation of `foo::value` is required", we could drop the "shall be a complete type" bit, or move it into a note if we want to be painfully clear. (I'm personally inclined to drop it: I don't think this is a case where adding more words improves clarity.)

In para 6, "shall have a member `value` which is convertible to `bool`" falls a bit short, e.g., `struct S { double& value; }` has a member `value` which is convertible to `bool`. I think we want to require `bool(B_i::value)` to be a constant expression.

jensmaurer:
Looks like this needs some decision to go with @jwakely's suggestion (or not).

source/utilities.tex
jensmaurer: Richard can't find those words in the current draft, so let's ignore this comment.

jensmaurer:
Editorial meeting: The rephrasing seems like a clarification, but it's normatively unclear whether "declaration" or "definition" is the right constraint here. This is not subsumed by general wording elsewhere, because this exactly delineates when instantiations happen and when they don't happen, so we need to be precise. Defer to LWG to decide.

tkoeppe:
@jwakely: could you perhaps schedule an LWG-bug-Monday email slot for tihs?

jwakely:
I've pushed a rewrite that addresses @CaseyCarter's comments and removes the problematic "a definition of `Bi::value` is required to exist" wording that @zygoloid didn't like. Now it simply says "Instantiation of `Bj ::value` is required".

source/meta.tex
```diff
@@ -2184,44 +2186,58 @@
...
+template<class T>
+  concept HasValue = requires { { T::value } -> std::convertible_to<bool>; };
+template<class T>
+  using maybe_value = std::conjunction<std::bool_constant<HasValue<T>>, T>;
+
+struct X { };
+
+static_assert( ! maybe_value<X>::value );       // OK, \tcode{X::value} not used
+static_assert( ! maybe_value<int>::value );     // OK, \tcode{int::value} not used
+static_assert( maybe_value<std::true_type>::value );
```
JohelEGP, line 2228:
```suggestion
static_assert(!maybe_value<X>::value);       // OK, \tcode{X::value} not used
static_assert(!maybe_value<int>::value);     // OK, \tcode{int::value} not used
static_assert(maybe_value<std::true_type>::value);
```

source/meta.tex
```diff
@@ -2184,44 +2186,58 @@
 forms the logical conjunction of its template type arguments.
 
 \pnum
+For the specialization \tcode{disjunction<>},
```
JohelEGP, line 2189:
```suggestion
For the specialization \tcode{conjunction<>},
```

source/meta.tex
```diff
@@ -374,6 +374,8 @@
   template<class... B> struct conjunction;
   template<class... B> struct disjunction;
   template<class B> struct negation;
+  template<> struct conjunction : true_type { };
+  template<> struct disjunction : false_type { };
```
JohelEGP, line 378: These need indexing, too. Consider contributing towards #5111.

source/meta.tex
```diff
@@ -2184,44 +2186,58 @@
 forms the logical conjunction of its template type arguments.
 
 \pnum
+For the specialization \tcode{disjunction<>},
+let $\tcode{B}_{i}$ be \tcode{true_type}.
```
JohelEGP, line 2190:
```suggestion
```
Based on the addition of the explicit specialization to the synopsis, and that only the primary template is present in the `itemdecl` environment, I believe this is not necessary.

source/meta.tex
```diff
@@ -2236,43 +2252,45 @@
 forms the logical disjunction of its template type arguments.
 
 \pnum
+For the specialization \tcode{disjunction<>},
+let $\tcode{B}_{i}$ be \tcode{false_type}.
```
JohelEGP, line 2256:
```suggestion
```
Based on the addition of the explicit specialization to the synopsis, and that only the primary template is present in the `itemdecl` environment, I believe this is not necessary.

source/meta.tex
```diff
@@ -2184,44 +2186,58 @@
...
+Instantiation of \tcode{$\tcode{B}_{j}$::value} is required for $1 \leq j \leq i$.
+Instantiation of \tcode{$\tcode{B}_{k}$::value} is not required for $i < k \leq N$.
+\begin{note}
+This is analogous to the short-circuiting behavior of
+the built-in operator \tcode{\&\&}.
+\end{note}
+\begin{note}
+This allows types without a \tcode{value} static data member
+to be template arguments for \tcode{conjunction}, as long as an
+earlier argument evaluates to \tcode{false}, for example:
```
JohelEGP, line 2217: Why is this not an `example` environment after the notes?

JohelEGP (changes requested):

source/meta.tex
jwakely: They're not really necessary since I added "For the specialization `conjunction<>`, let Bi be `true_type`." Maybe I should remove them again.

source/meta.tex
jwakely: I agree we only need one or the other. The problem with specifying this via the synopsis is that it's easy to miss among dozens of other declarations. Specifying it here doesn't have that problem.

source/meta.tex
jwakely: Because the prose is clarifying the normative text, which is the purpose of a note. The statement is not specific to the example in the code, it's generally true. The example is a specific example showing it. We have lots of notes that include "for example:" like this. But I don't feel strongly about it being a note.

PR #2557, opened 2018-12-04
Escape spaces following mid-sentence periods.

### LINKED ISSUE #2555: [atomics.types.operations]/20 Awkward "E.g." at start of sentence
Consider the "E.g. " below:
image
Some thoughts:
- The space between "E.g." and "list" seems a bit wide, is it possible LaTeX is generating a sentence-space? Should maybe "\ " be used? (Or maybe even "~", to prevent a line break.)
- In the vast majority of cases, the document actually has a comma after "e.g.". Maybe that's the way to go here, too?
- Starting a sentence with the "e.g." abbreviation could be considered poor style to begin with. Maybe nicer to just say "For example,"

Eelis:
Fixes #2555.

### DIFF
--- a/source/atomics.tex
+++ b/source/atomics.tex
@@ -1651,7 +1651,7 @@ do {
 \end{example}
 \begin{example} Because the expected value is updated only on failure,
 code releasing the memory containing the \tcode{expected} value on success will work.
-E.g. list head insertion will act atomically and would not introduce a
+E.g.\ list head insertion will act atomically and would not introduce a
 data race in the following code:
 \begin{codeblock}
 do {
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -635,9 +635,9 @@ Ritchie: \doccite{The C Programming Language} (Prentice-Hall, 1978, ISBN
 
 \pnum
 Portions of the library Clauses of this document are based
-on work by P.J. Plauger, which was published as \doccite{The Draft
+on work by P.J.\ Plauger, which was published as \doccite{The Draft
 Standard \Cpp{}  Library} (Prentice-Hall, ISBN 0-13-117003-1, copyright
-\copyright 1995 P.J. Plauger).
+\copyright 1995 P.J.\ Plauger).
 
 \pnum
 POSIX\textregistered\ is a registered trademark of the Institute of Electrical and
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -6416,7 +6416,7 @@ shall be in scope at the point at which the member is defined.
 The definition of an explicitly specialized class is unrelated to the
 definition of a generated specialization.
 That is, its members need
-not have the same names, types, etc. as the members of a generated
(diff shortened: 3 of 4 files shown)


tkoeppe:
Could you attach a diffshot, or at least a representative subset?

tkoeppe:
That is odd! I imagine there may be built-in heuristics that detect things like initials and common abbreviations.

Could you please reduce the change to the minimal effective set? (And do feel free to also remove existing, ineffective backslashes.) Thanks!

Eelis:
For names with initials, how would you feel about using "~", as in "P.J.~Plauger"? We currently don't happen to have a line break there, but if preceding content is changed one day, that could change without us noticing, and might look bad.

tkoeppe:
I feel pretty much entirely indifferent. It's not currently a problem, we don't have a ton of names with initials, nor do we expect to gain more, and the existing occurrences aren't likely to be touched. So I'd just keep it as is, and simple.

Is there a visual diff if you use `~`?

godbyk:
There is no visual difference between using `~` and `\ ` aside from `~` inserting a non-breaking space.

The basic algorithm is that if the character preceding a period is uppercase, then interword spacing is used. Otherwise, end-of-sentence spacing it used.

So `P.J. Plauger` would result in an interword spacing while `E.g. list` would result in end-of-sentence spacing.

godbyk:
Oh, and if you use a comma after `e.g.` and `i.e.` then an interword space will be used there as well.

tkoeppe:
Ah yes, makes sense, thanks.

I would retain the status quo in that case.

Eelis:
Ah great, thanks for the explanation! I'll implement that "character preceding period is uppercase" heuristic for sentence splitting in cxxdraft-htmlgen as well.

That will resolve my real issue (because my ulterior motive was to use the "\ " or "~" to have cxxdraft-htmlgen not split the sentences there), and given that LaTeX apparently already handles almost all of this just fine by itself after all, I agree it's not worth caring about. :)

Closing, cheers!

tkoeppe:
I mean, do feel free to fix that one "E.g." case if you like. Or better yet, say "For example,". Or don't.

godbyk:
Personally, I'd prefer `For example,` as I don't like using `E.g.` or `I.e.` to start a sentence.

Eelis:
Ok you convinced me... but then I discovered that if you force push to a branch used in a pull request that is currently closed, then you cannot reopen that pull request anymore (the "Reopen and comment" button is greyed out now). :)

tkoeppe:
Hah, how weird.

johelegp:
> Ok you convinced me... but then I discovered that if you force push to a branch used in a pull request that is currently closed, then you cannot reopen that pull request anymore (the "Reopen and comment" button is greyed out now). :)

That's solved by force-pushing back to the commit of the PR.

PR #2565, opened 2018-12-07
[iterator.synopsis] Add reference to [alg.req.sortable]

johelegp:


CaseyCarter:
Don't cross-references in synopses usually link to the specification of the entity being declared?

johelegp:
Yeah. Or defined in this case of a concept.

CaseyCarter:
Ah, nevermind - I had convinced myself that [alg.req.sortable] was [alg.sorting] which would make no sense at all. Carry on ;)

PR #2571, opened 2018-12-07
[temp.func.order] Adjust example to rules in [temp.deduct.partial].

### LINKED ISSUE #2559: Incompatible example with CWG 1395
From this [SO answer](https://stackoverflow.com/a/41069348/5376789), the example in [temp.func.order]/5:

    template<class T, class... U> void f(T, U...);          // #1
    template<class T            > void f(T);                // #2
    template<class T, class... U> void g(T*, U...);         // #3
    template<class T            > void g(T);                // #4

    void h(int i) {
      f(&i);                                                // error: ambiguous
      g(&i);                                                // OK: calls #3
    }

is incompatible with [temp.deduct.partial]/11. `f(&i)` should call #2.

jensmaurer:
Fixes #2559.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -3818,7 +3818,7 @@ template<class T, class... U> void g(T*, U...);         // \#3
 template<class T            > void g(T);                // \#4
 
 void h(int i) {
-  f(&i);                                                // error: ambiguous
+  f(&i);                                                // OK: calls \#2
   g(&i);                                                // OK: calls \#3
 }
 \end{codeblock}


xskxzr:
After this change, this example seems to make no sense for (or even conflict with) the wording in [temp.deduct.partial]/5:

> Note: Since partial ordering in a call context considers only parameters for which there are explicit call arguments, some parameters are ignored ...

What about deleting this paragraph and moving these examples into corresponding paragraphs in [temp.deduct.partial]?

PR #2572, opened 2018-12-07
[template.mask.array.overview, iterator.concept.iterator] Use non-eng…

johelegp:
…lish period

The change in [iterator.concept.iterator] is done by removing an astray closing parenthesis.


tkoeppe:
Thanks, addressing this via separate commits.

PR #2578, opened 2018-12-08
[reverse.iter.cons] Indent consistently with the parent subclause

johelegp:


johelegp:
Barring other changes to [iterators], this is ready.

jensmaurer:
@tkoeppe, seems this is ready to go.

source/iterators.tex
```diff
@@ -2986,9 +2986,9 @@
     constexpr reverse_iterator<Iterator> make_reverse_iterator(Iterator i);
 
   template<class Iterator1, class Iterator2>
-    requires (!SizedSentinel<Iterator1, Iterator2>)
-  inline constexpr bool disable_sized_sentinel<reverse_iterator<Iterator1>,
-                                               reverse_iterator<Iterator2>> = true;
+      requires (!SizedSentinel<Iterator1, Iterator2>)
+    inline constexpr bool disable_sized_sentinel<reverse_iterator<Iterator1>,
```
zygoloid: The existing indentation is correct:
 * the requires clause should be indented 2 spaces from the line it's continuing (the template-head)
 * the variable declaration is a definition, so does not need indentation.

This style would also be acceptable, but is generally used only for non-defining declarations:

```
template<class Iterator1, class Iterator2>
  requires (!SizedSentinel<Iterator1, Iterator2>)
  inline constexpr bool disable_sized_sentinel<reverse_iterator<Iterator1>,
                                               reverse_iterator<Iterator2>> = true;
```

PR #2581, opened 2018-12-08
[expr.and][expr.or][expr.xor] Add grouping sentence to the binary and/or/xor operators.

Rakete1111:
It seems like every binary expression has a sentence which says which way the operator groups, except for and, or and xor.

For consistency, add a sentence saying that they group left-to-right.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -5982,6 +5982,7 @@ false.
 \end{bnf}
 
 \pnum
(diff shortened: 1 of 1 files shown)


jensmaurer:
The commit comment should start with "[blah,blurb,blubb]" when mentioning several labels. Could you please rebase and force-push? Thanks.

PR #2582, opened 2018-12-08
[move.iterators] Various (hopefully) editorial fixes

johelegp:
Hopefully the intention is crystal clear and this can be handled editorially.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -4406,7 +4406,7 @@ template<class Iterator>
 \begin{itemdescr}
 \pnum
 \constraints
-The expression \tcode{x + n} shall be valid and have type \tcode{Iterator}.
+The expression \tcode{x.base() + n} shall be valid and have type \tcode{Iterator}.
 
 \pnum
 \returns \tcode{x + n}.
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -4543,6 +4543,16 @@ template<class S2>
 \effects Equivalent to: \tcode{last = s.last; return *this;}
 \end{itemdescr}
 
+\indexlibrarymember{base}{move_sentinel}%
+\begin{itemdecl}
+constexpr S base() const;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\returns \tcode{last}.
+\end{itemdescr}
+
 \rSec2[iterators.common]{Common iterators}
 
 \rSec3[common.iterator]{Class template \tcode{common_iterator}}


johelegp:
Increased the scope of the PR. My first attempt to the fix in the second commit was wrong, so there's some noise.

source/iterators.tex
```diff
@@ -4406,7 +4406,7 @@
 \begin{itemdescr}
 \pnum
 \constraints
-The expression \tcode{x + n} shall be valid and have type \tcode{Iterator}.
+The expression \tcode{x + n} shall be valid.
```
CaseyCarter: The intent here is "The expression `\tcode{x.base() + n}` shall be valid and have type `\tcode{Iterator}`." Admittedly it's weird to constrain `n + move_iterator` when `move_iterator + n` isn't constrained; I'll open an LWG issue to constrain the member operators similarly.

CaseyCarter: ...and while I'm here, thanks for the incredibly detailed review of and corrections to the Ranges wording.

johelegp: > The intent here is "The expression `\tcode{x.base() + n}` shall be valid and have type `\tcode{Iterator}`." Admittedly it's weird to constrain `n + move_iterator` when `move_iterator + n` isn't constrained; I'll open an LWG issue to constrain the member operators similarly.

Right. I hoped that [move.iter.requirements] did that job.

> ...and while I'm here, thanks for the incredibly detailed review of and corrections to the Ranges wording.

Sure :)

source/iterators.tex
```diff
@@ -348,11 +349,32 @@
...
+    constexpr bool operator!=(
+      const move_sentinel<S>& x, const move_iterator<Iterator>& y);
+
+  template<class Iterator, SizedSentinel<Iterator> S>
+    constexpr iter_difference_t<Iterator> operator-(
+      const move_sentinel<S>& x, const move_iterator<Iterator>& y);
+  template<class Iterator, SizedSentinel<Iterator> S>
+    constexpr iter_difference_t<Iterator> operator-(
+      const move_iterator<Iterator>& x, const move_sentinel<S>& y);
+
```
CaseyCarter: These operators are very much intentionally hidden friends for the effect on name lookup. This change is not editorial.

johelegp: Thank you. I guess that's the case for all the friends added?

CaseyCarter: Yes, that is the case.

source/iterators.tex
```diff
@@ -4406,7 +4406,7 @@
 \begin{itemdescr}
 \pnum
 \constraints
-The expression \tcode{x + n} shall be valid and have type \tcode{Iterator}.
+The expression \tcode{x.base() + n} shall be valid and have type \tcode{Iterator}.
 
 \pnum
 \returns \tcode{x + n}.
```
zygoloid: This change creates an inconsistency between the constraints and the returned value.

`move_iterator:::operator+(difference_type)` has no constraints, so the expression in question is always valid (even though instantiation of the operator might fail), so presumably we're missing a constraint around [move.iter.nav]p7? Whether or not this expression is SFINAEable is not editorial, and the current specification appears to be normatively wrong, so I think this part needs to go to LWG.

johelegp: http://wg21.link/LWG3293 attempts to solve this.

CaseyCarter: Thanks for the reminder that I need to fix this issue's P/R.

johelegp:
The former change is handled by https://wg21.link/LWG3293 (P3) and the latter is in the draft.

PR #2585, opened 2018-12-09
[counted.iter.cmp] Add missing colon to specification element

johelegp:


source/iterators.tex
```diff
@@ -5481,7 +5481,7 @@
 
 \pnum
 \begin{note}
-The argument order in the \textit{Effects} element is reversed
+The argument order in the \textit{Effects:} element is reversed
 because \tcode{length} counts down, not up.
```
jensmaurer: This should simply use `\effects{}`.

jensmaurer:
@tkoeppe, this is ready to go.

PR #2592, opened 2018-12-11
[range.view] Add complexity requirement normatively

johelegp:
And drive-by move description above the definitions.
Also, could a \pnum be missing after the list in the example?

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -867,6 +867,11 @@ template<class T>
 
 \begin{itemdescr}
 \pnum
+\tcode{T} models \libconcept{View} only if the complexity
+for each initialization and assignment required by \libconcept{Copyable<T>}
+is constant time.
+
+\pnum
 For a type \tcode{T}, the default value of \tcode{enable_view<T>} is:
 \begin{itemize}
 
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -850,6 +850,10 @@ copying the container copies the elements,
 which cannot be done in constant time.
 \end{example}
 
+\pnum
+Since the difference between \libconcept{Range} and \libconcept{View} is largely
+semantic, the two are differentiated with the help of \tcode{enable_view}.
+
 \indexlibrary{\idxcode{enable_view}}%
 \indexlibrary{\idxcode{View}}%
 \begin{itemdecl}
@@ -862,10 +866,6 @@ template<class T>
 \end{itemdecl}
 
 \begin{itemdescr}
-\pnum
-Since the difference between \libconcept{Range} and \libconcept{View} is largely
-semantic, the two are differentiated with the help of \tcode{enable_view}.
-
 \pnum
 For a type \tcode{T}, the default value of \tcode{enable_view<T>} is:
 \begin{itemize}


source/ranges.tex
```diff
@@ -866,6 +866,12 @@
 \end{itemdecl}
 
 \begin{itemdescr}
+\pnum
+\tcode{T} models \libconcept{SizedRange} only if its
+copy constructor, move constructor (if any),
+copy assignment operator, and move assignment operator (if any)
+are \bigoh{1}.
+
```
jensmaurer: What does O(1) refer to? The number of bytes taken up by the range object?
Where is the existing normative statement that says this? If there is none, this change does not seem editorial.

jensmaurer:
Please fold your fixups by "git rebase -i master" and force-push for easier review.

source/ranges.tex
jensmaurer: The [range.view] p1 sentence is much clearer than the O(1) we have here; it clearly says "not proportional to the number of elements in the View". Can we find a formulation that actually says this? Maybe "... only if the execution time required for the initializations and assignments required by Copyable<T> is independent of the number of elements in the View"? This has one "required" too many, and maybe the library says something else for "execution time" when it wants to talk about time complexity.

johelegp: Right. I used that "that is" explanation as part of your suggested fix.

source/ranges.tex
```diff
@@ -866,6 +864,11 @@
 \end{itemdecl}
 
 \begin{itemdescr}
+\pnum
+\tcode{T} models \libconcept{View} only if the execution time
+for the the initializations and assignments required by \libconcept{Copyable<T>}
```
jensmaurer: duplicate "the"
Also, we should go singular here to avoid the question whether we mean each or all (together).
"if the execution time for each initialization and assignment required by ... is independent of the number of elements in T"
(If it's O(n^2), it's also "not proportional"; we want to say something stronger.)

johelegp: Sure. I applied the use of "complexity of ... constant time" and reverted the first paragraph of the description.

jensmaurer: Thanks.

johelegp: Do you think the previous sibling clause, [range.sized], also needs to be modified? That's where I took the use of O(1) from:
https://github.com/cplusplus/draft/blob/3d9f80e990daf459a85cff539bd2ce64c0986886/source/ranges.tex#L797

source/ranges.tex
```diff
@@ -863,8 +867,9 @@
 
 \begin{itemdescr}
 \pnum
-Since the difference between \libconcept{Range} and \libconcept{View} is largely
-semantic, the two are differentiated with the help of \tcode{enable_view}.
+\tcode{T} models \libconcept{View} only if the complexity
+for each initialization and assignment required by \libconcept{Copyable<T>}
+is constant time.
```
zygoloid: This may well be intended, but I don't see how it's editorial. Is this somehow justified by the existing text?

CaseyCarter: Paragraph 1 above: "The `View` concept specifies the requirements of a `Range` type that has constant time copy, move, and assignment operators...".

Although I wouldn't mind sending this to LWG: we need to add the missing requirement that destruction is O(1) as well which is *certainly* not editorial.

zygoloid: Let's send the whole thing to LWG then. If they want us to apply the non-normative change editorially we can, but I think it's generally better for holistic changes with a normative component to be applied together.

johelegp:
If there's agreement that introductory paragraphs are normative, this is subsumed by #3453.

tkoeppe:
I'm not sure that we have such agreement, or that ISO even allows that. It would be nice if that were allowed, and if we could have "(informative)" subclauses for introductions, but we tried to find permission for that in the ISO drafting directives and did not succeed.

I'm noticing that we _are_ using informative subclauses in the LFTS, e.g. https://cplusplus.github.io/fundamentals-ts/v3.html#general.plans.

johelegp:
Subsumed by #3453.

PR #2594, opened 2018-12-11
[range.utility, ranges.syn] Various editorial fixes

johelegp:


source/ranges.tex
```diff
@@ -108,7 +108,7 @@
 
   // \ref{view.interface}, class template \tcode{view_interface}
   template<class D>
-    requires is_class_v<D>
+    requires is_class_v<D> && Same<D, remove_cv_t<D>>
```
jensmaurer: Where is the existing normative text that states the requirement added here, i.e. that D is cv-unqualified?

johelegp: It's on the class definition.
https://github.com/cplusplus/draft/blob/a88d02d8d77e730bbf2730b6988a8b63319496c3/source/ranges.tex#L1025-L1027

jensmaurer: Thanks.

PR #2596, opened 2018-12-12
[range.adaptors, ranges.syn] Various editorial fixes

johelegp:


source/ranges.tex
```diff
@@ -4438,7 +4438,7 @@
     constexpr outer_iterator(Parent& parent, iterator_t<Base> current)
       requires ForwardRange<Base>;
     constexpr outer_iterator(outer_iterator<!Const> i)
-      requires Const && ConvertibleTo<iterator_t<V>, iterator_t<const V>>;
+      requires Const && ConvertibleTo<iterator_t<V>, iterator_t<Base>>;
```
jensmaurer: This looks like a normative change. What's the argument why this is editorial, i.e. implied by normative text elsewhere?

johelegp: `const V` and `Base` are equivalent, because when `Const` is `true` (required earlier in this requires clause), `Base` is `const V`:
https://github.com/cplusplus/draft/blob/a88d02d8d77e730bbf2730b6988a8b63319496c3/source/ranges.tex#L4422-L4423
The intention is to be consistent with the other range adaptor iterators/sentinels:
https://github.com/cplusplus/draft/blob/a88d02d8d77e730bbf2730b6988a8b63319496c3/source/ranges.tex#L2309-L2310
https://github.com/cplusplus/draft/blob/a88d02d8d77e730bbf2730b6988a8b63319496c3/source/ranges.tex#L2654-L2655
https://github.com/cplusplus/draft/blob/a88d02d8d77e730bbf2730b6988a8b63319496c3/source/ranges.tex#L3518-L3519
https://github.com/cplusplus/draft/blob/a88d02d8d77e730bbf2730b6988a8b63319496c3/source/ranges.tex#L3732-L3736
https://github.com/cplusplus/draft/blob/a88d02d8d77e730bbf2730b6988a8b63319496c3/source/ranges.tex#L4027-L4028

CaseyCarter: While the result is equivalent, this change seems to obfuscate the spec by introducing a layer of indirection. I think it would be better to change the *other* occurrences to use `const V` instead of `Base`.

johelegp: I agree with you in this case, but am nevertheless hesitant to make those changes.

jensmaurer: @johelegp, since you're not willing to change "Base" to "const V" in the "other" occurrences, please remove the "[range.split.outer] Use Base for consistency with other adaptors" patch from this pull request (force pushing is fine) and consider opening a new editorial issue. This allows the other two fixes to go in. Since @CaseyCarter is deeply involved with ranges, we should respect his editorial guidance by default.

johelegp: Right. For future referece, @CaseyCarter, should both `Base` in the following requires clause be changed?
https://github.com/cplusplus/draft/blob/a88d02d8d77e730bbf2730b6988a8b63319496c3/source/ranges.tex#L3732-L3736

jensmaurer: I'd say yes.

tkoeppe:
I wasn't following the discussion; does everyone agree that adding `&& is_object_v<Pred>` is editorial?

johelegp:
Yes. Here's the definition which has that part of the requires clause.
https://github.com/cplusplus/draft/blob/a88d02d8d77e730bbf2730b6988a8b63319496c3/source/ranges.tex#L1681-L1687

tkoeppe:
Thank you!

PR #2600, opened 2018-12-17
[alg.min.max] Fix typo in returns element of ranges::minmax

CaseyCarter:


source/algorithms.tex
```diff
@@ -8042,7 +8042,7 @@
 
 \pnum
 \returns
-\tcode{\{(b, a)\}} if \tcode{b} is smaller than \tcode{a}, and
+\tcode{\{b, a\}} if \tcode{b} is smaller than \tcode{a}, and
```
tkoeppe: Can you also please add a linebreak on line 8024, after the template head?

tkoeppe: @CaseyCarter: ping

tkoeppe:
Thanks!

PR #2602, opened 2018-12-17
Update expressions.tex

JonCavesMSFT:
Fix a couple of typos (missing ';') and make a comment clearer

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -2053,7 +2053,7 @@ auto y = [&r = x, x = x+1]()->int {
             return x+2;
          }();  // Updates \tcode{::x} to 6, and initializes \tcode{y} to 7.
 
-auto z = [a = 42](int a) { return 1; }; // error: parameter and \grammarterm{init-capture} have the same name
+auto z = [a = 42](int a) { return 1; }; // error: parameter and local variable have the same name
 \end{codeblock}
 \end{example}
 
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -2053,7 +2053,7 @@ auto y = [&r = x, x = x+1]()->int {
             return x+2;
          }();  // Updates \tcode{::x} to 6, and initializes \tcode{y} to 7.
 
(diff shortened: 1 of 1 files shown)


source/expressions.tex
```diff
@@ -2053,7 +2053,7 @@
             return x+2;
          }();  // Updates \tcode{::x} to 6, and initializes \tcode{y} to 7.
 
-auto z = [a = 42](int a) { return 1; } // error: parameter and local variable have the same name
+auto z = [a = 42](int a) { return 1; }; // error: parameter and \grammarterm{init-capture} have the same name
```
jensmaurer: The "local variable" here wants to point to the fact that init-captures are described as creating a local variable in the appropriate scope (which happens to be the same scope as the lambda's parameter). Using "init-capture" here obscures the situation at bit, in my view.

JonCavesMSFT: Yes I realize that but I still think that using the term 'local variable' here makes it overly obtuse - someone reading this who is not well versed in CWG "speak" will look at this and think but the only local variable here is 'z' and it does not have the same name as 'a'. Note: the parallel example for simple captures (just above) does use the term 'simple-capture'  even though a simple-capture creates an unnamed non-static data member (maybe). But is just a comment so I am happy to revert this change. Note: no-one tests for this (or the parallel rule for simple-captures).

jensmaurer: The parallel example for simple captures, the normative rules is explicitly talking about parameter names and identifiers for simple-captures. In contrast, there is no direct expression of the prohibition of naming conflicts between parameter names and names of init-captures. Instead, the example servers to highlight that the prohibition is a consequence of other normative rules, including the one on regular parameter names and variables.

JonCavesMSFT: Ok … I reverted the change to the comment

tkoeppe:
Stolen and committed as 0cd2126ddab16740d38a302f0da7fee5658f5fbb -- thank you!

PR #2603, opened 2018-12-19
[reverse.iterators] Use the public accessor function,

### LINKED ISSUE #2579: [reverse.iter.nonmember] Replace data member access with accessor
New wording seems to avoid describing non-member functions by accessing would-be private/protected data members, so it seems reasonable to update [reverse.iter.nonmember] to use `.base()` instead of `.current`. In fact, you'll notice that the return type of the first function uses `.base()`.

jensmaurer:
not the exposition-only member 'current', from non-member functions.

Fixes #2579.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -3305,12 +3305,12 @@ template<class Iterator1, class Iterator2>
 \begin{itemdescr}
 \pnum
 \constraints
-The expression \tcode{x.current == y.current} shall be valid and
+The expression \tcode{x.base() == y.base()} shall be valid and
 convertible to \tcode{bool}.
 
 \pnum
 \returns
-\tcode{x.current == y.current}.
+\tcode{x.base() == y.base()}.
 \end{itemdescr}
 
 \indexlibrarymember{operator"!=}{reverse_iterator}%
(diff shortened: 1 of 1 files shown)


source/iterators.tex
```diff
@@ -3985,11 +3985,11 @@
         operator-(const move_iterator& x, const move_sentinel<S>& y);
     friend constexpr iter_rvalue_reference_t<Iterator>
       iter_move(const move_iterator& i)
-        noexcept(noexcept(ranges::iter_move(i.current)));
+        noexcept(noexcept(ranges::iter_move(i.base())));
```
CaseyCarter: This has normative impact: `i.base()` is potentially throwing, whereas `i.current` is not. The proposed change is equivalent to `noexcept(false)`.

source/iterators.tex
```diff
@@ -3985,11 +3985,11 @@
...
     template<IndirectlySwappable<Iterator> Iterator2>
       friend constexpr void
         iter_swap(const move_iterator& x, const move_iterator<Iterator2>& y)
-          noexcept(noexcept(ranges::iter_swap(x.current, y.current)));
+          noexcept(noexcept(ranges::iter_swap(x.base(), y.base())));
```
CaseyCarter: Ditto normative change equivalent to `noexcept(false)`.

source/iterators.tex
jensmaurer: Good catch; reverted. Plus, I didn't want to mess with move_iterator anyway.
However, I'm left wondering why `base()` is not noexcept, given that no precondition is apparent.

source/iterators.tex
jensmaurer: Good catch; reverted.

source/iterators.tex
CaseyCarter: `base` returns a copy of `current`, which could potentially throw.

PR #2605, opened 2018-12-19
[range.adaptors] Replace Base with const V for clarity

johelegp:
See discussion at https://github.com/cplusplus/draft/pull/2596#discussion_r241916798.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -2307,7 +2307,7 @@ namespace std::ranges {
     iterator() = default;
     constexpr iterator(Parent& parent, iterator_t<Base> current);
     constexpr iterator(iterator<!Const> i)
-      requires Const && ConvertibleTo<iterator_t<V>, iterator_t<Base>>;
+      requires Const && ConvertibleTo<iterator_t<V>, iterator_t<const V>>;
 
     constexpr iterator_t<Base> base() const;
     constexpr decltype(auto) operator*() const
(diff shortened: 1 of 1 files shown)


jensmaurer:
[LWG 3181](https://wg21.link/lwg3181) says we must use "Base" here, not "const V", to correctly handle the template instantiation of the enclosing class.

CaseyCarter:
> @CaseyCarter, have you changed your opinion since you approved these changes?

Yes, and I've since changed it back *again* after running a canonical example past the core reflector. I currently believe these changes are valid and the concept implementations are buggy. Sorry for the confusion - it can be challenging to determine which issues are actual problems with the spec and which are bugs in the implementations.

johelegp:
> > @CaseyCarter, have you changed your opinion since you approved these changes?
> 
> [...] I currently believe these changes are valid and the concept implementations are buggy. [...]

That's right. Using requires clauses doesn't require you to do the same workaround of introducing a dummy template parameter defaulted to some outer one for the sake of delayed instantiation to apply SFINAE.

source/ranges.tex
```diff
@@ -2307,7 +2307,7 @@
     iterator() = default;
     constexpr iterator(Parent& parent, iterator_t<Base> current);
     constexpr iterator(iterator<!Const> i)
```
zygoloid: It'd be clearer to use `iterator<false>` here. Oh, we can't do that because that'd be ill-formed when instantiating with `Const = false`. Yuck, what a language!

zygoloid:
I believe this is technically not editorial: the precise spelling of the `requires` clause is detectable by user code, based on partial ordering in derived classes that inherit the constructors of this base class and add overloads.

johelegp:
https://wg21.link/P1983 will take care of this.

PR #2608, opened 2018-12-20
[expr.new] A new-expression might not invoke a constructor

### LINKED ISSUE #2484: [expr.new] Is a copy constructor call elided in a new-expression?
[expr.new] p19.2 makes it clear that we perform direct-initialization, which means the copy constructor of T is not called in a case like this:
```
struct T { };
T f();
T * p = new T(f());
```
However, p21 then talks about "the constructor" in the class object case, which could lead to the misunderstanding that some constructor is always called. This is amplified because allocation/deallocation functions are mentioned in the same sentence, and accessibility to those is always checked, even if the allocation is elided.
Suggestion: add "(if any)" after "the constructor". (The "no constructor" case also arises for aggregate initialization, it seems.)
Do we have a good cross-reference target when we want to talk about "mandatory copy elision"?

jensmaurer:
Fixes #2484.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -4824,7 +4824,8 @@ value computation of the
 If the \grammarterm{new-expression} creates an object or an array of
 objects of class type, access and ambiguity control are done for the
 allocation function, the deallocation function\iref{class.free}, and
-the constructor\iref{class.ctor}. If the \grammarterm{new-expression}
+the constructor\iref{class.ctor} selected for the initialization (if any).
+If the \grammarterm{new-expression}
 creates an array of objects of class type, the destructor is potentially
 invoked\iref{class.dtor}.
 


source/expressions.tex
```diff
@@ -4824,7 +4824,8 @@
 If the \grammarterm{new-expression} creates an object or an array of
 objects of class type, access and ambiguity control are done for the
 allocation function, the deallocation function\iref{class.free}, and
-the constructor\iref{class.ctor}. If the \grammarterm{new-expression}
+the constructor\iref{class.ctor} selected for the initialization (if any).
+If the \grammarterm{new-expression}
 creates an array of objects of class type, the destructor is potentially
```
CaseyCarter: Does "destructor" here need an "if any" as well?

jensmaurer: No. The "constructor" part needs an "if any" because we occasionally initialize class objects without calling a constructor (cf. aggregates initialized with braced-init-lists), or at least not invoking a constructor at the syntactic level of /new-expression/ (cf. the example in #2484). In contrast, a class type always has a destructor (it might be trivial, though), and we'll call it (guaranteed) for the elements 0..n-1 if initializing the n-th array element throws an exception. (Note we say "potentially invoked" here, so we're just checking the constraints, but are not making any promises right here that we will actually runtime-call the destructor.)

PR #2610, opened 2018-12-21
[std] Replace underscores in stable labels with periods.

### LINKED ISSUE #2609: decide what to do with underscores in section labels
Continuing from #1498, we have 81 section labels containing underscores. These look extremely ugly: the font we use to display section labels does not have a reasonable underscore character.

We should either:

 * change the font used to render section labels (both in the body text and in xrefindex and xrefdelta) to have a less ugly underscore, or
 * revise the section labels to avoid using underscores

Generally when we form a section label from an identifier, we change underscores into periods (eg, [unord.map.syn] not [unordered_map.syn]); fixing these section labels to do the same may be the most consistent option.

jensmaurer:
Fixes #2609.


jensmaurer:
@zygoloid, this is the patch that removes underscores in stable labels, if you want to be semi-brave.

zygoloid:
Sorry, I left this too late for this mailing. Let's do it for the post-Kona draft.

jensmaurer:
Rebased.

PR #2612, opened 2018-12-21
[iterators] Qualify declarator-id with sub-namespace.

jensmaurer:
Partially addresses #2512.


jensmaurer:
@zygoloid, @tkoeppe: This is the first in a series of patches to qualify /declarator-id/s with a sub-namespace, in particular for `ranges::`, as decided in #2512. Please check whether that is the right direction.

PR #2613, opened 2018-12-21
[mismatch] LWG3178 std::mismatch is missing an upper bound

### LINKED ISSUE #2611: [mismatch] LWG 3178: std::mismatch is missing an upper bound
https://cplusplus.github.io/LWG/issue3178
This was opened by @geoffromer , but @Dani-Hub thought it might be editorial.

jensmaurer:
Fixes #2611.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -3655,16 +3655,17 @@ Let $E$ be:
 \end{itemize}
 
 \pnum
+Let $N$ be $\min(\tcode{last1 - first1}, \ \tcode{last2 - first2})$.
+
+\pnum
 \returns
 \tcode{\{ first1 + n, first2 + n \}},
-where \tcode{n} is the smallest integer such that $E$ holds,
-or $\min(\tcode{last1 - first1}, \ \tcode{last2 - first2})$
-if no such integer exists.
+where \tcode{n} is the smallest integer in \range{0}{$N$} such that $E$ holds,
+or $N$ if no such integer exists.
 
 \pnum
 \complexity
-At most $\min(\tcode{last1 - first1}, \ \tcode{last2 - first2})$
-applications of the corresponding predicate and any projections.
+At most $N$ applications of the corresponding predicate and any projections.
 \end{itemdescr}
 
 \rSec2[alg.equal]{Equal}


jensmaurer:
@zygoloid, can we merge this under the "pellucidly obvious" doctrine?

source/algorithms.tex
```diff
@@ -3540,17 +3540,18 @@
...
+\pnum
+Let $N$ be $\min(\tcode{last1 - first1}, \ \tcode{last2 - first2})$.
+
 \pnum
 \returns
 \tcode{\{ first1 + n, first2 + n \}},
 where \tcode{n} is the smallest integer such that $E$ holds,
-or $\min(\tcode{last1 - first1}, \ \tcode{last2 - first2})$
-if no such integer exists.
+or $N$ if no such integer less than $N$ exists.
```
tkoeppe: This wording is awkward. It's much more common in mathematics to say precisely what you want the first time round, rather than adding constraints later:

>    where `n` is the smallest integer less than $N$ such that $E$ holds, or ... if no such integer exists.

tkoeppe: It's also weird that $E$ uses `n` but doesn't formally depend on it; I would much prefer to call it $E(n)$. Then, if you move the definition of $N$ up, you can even say that "$E(N)$ is true" (or rather, "$E(n) is ... true when n is N").

tkoeppe:
Pellucid, perhaps, graceful, certainly not.

source/algorithms.tex
jensmaurer: Fixed the first issue. Left the second issue alone; the pattern that the index variable is not a parameter of E is pervasive (see e.g. [alg.any_of] and friends).

source/algorithms.tex
```diff
@@ -3540,17 +3540,18 @@
...
+\pnum
+Let $N$ be $\min(\tcode{last1 - first1}, \ \tcode{last2 - first2})$.
+
 \pnum
 \returns
 \tcode{\{ first1 + n, first2 + n \}},
-where \tcode{n} is the smallest integer such that $E$ holds,
-or $\min(\tcode{last1 - first1}, \ \tcode{last2 - first2})$
-if no such integer exists.
+where \tcode{n} is the smallest integer less than $N$ such that $E$ holds,
```
zygoloid: ... smallest integer in `[0, N)` ...

Otherwise what stops this from picking a negative value of `n`?

jensmaurer: Fixed.

PR #2619, opened 2018-12-29
[range.factories] New subclause, split off from [range.adaptors]

### LINKED ISSUE #2615: [range.adaptors] Generators bundled together with the adaptors
At http://ericniebler.com/2018/12/05/standard-ranges/, right before the "Range Adaptors" section is "Range Generators." [range.adaptors], which is supposed to describe _range adaptors_, also bundles these generators. You'll notice from the description of the `view::X` objects that those for the adaptors are described as denoting range adaptor objects, while those of the generators denote customization point objects. Shouldn't the generators live in their own subclause?

jensmaurer:
Range adaptors take a range and produce a new range;
range factories take something else (or nothing) and produce a range.

Fixes #2615.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -18,6 +18,7 @@ as summarized in \tref{range.summary}.
   \ref{range.prim}         & Range primitives  & \\
   \ref{range.req}          & Requirements      & \\
   \ref{range.utility}      & Range utilities   & \\
+  \ref{range.factories}    & Range factories   & \\
   \ref{range.adaptors}     & Range adaptors    & \\
 \end{libsumtab}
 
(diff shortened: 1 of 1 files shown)


johelegp:
`git log -1 -p --minimal` generates a better diff.

CaseyCarter:
Should `span` be relocated into [range.factories], which would necessitate striking the IMO useless "Range factories are declared in namespace `std::ranges::view`" paragraph?

morinmorin:
> striking the IMO useless "Range factories are declared in namespace std::ranges::view"

+1

> Should `span` be relocated into [range.factories]

and `subrange`?

PR #2620, opened 2018-12-29
[range.prim,iterator.container] merge into preceding subclause

### LINKED ISSUE #2614: [iterator.container,range.prim] "range" vs. "container and/or view"
Could we editorially merge [iterator.container] into [iterator.range], and likewise merge [range.prim] into [range.access]? I defy anyone to explain to me the difference between "range" and "container and/or view".

Should I simply prepare a PR to do so, and we can decide on the basis of the proposed wording changes whether or not the change needs LWG approval?

jensmaurer:
Fixes #2614.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -25,7 +25,6 @@ as summarized in \tref{iterators.lib.summary}.
 \ref{predef.iterators}      & Iterator adaptors         &                    \\
 \ref{stream.iterators}      & Stream iterators          &                    \\
 \ref{iterator.range}        & Range access              &                    \\
-\ref{iterator.container}    & Container and view access &                    \\
 \end{libsumtab}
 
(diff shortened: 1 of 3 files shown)


johelegp:
Why not rename the stable labels from "range.prim.*" to "range.access.*", too? They're range accessors too, after all. [ranges.general] would be left using "range primitives", so maybe update that too (and drive-by mention the adaptors, perhaps).

jensmaurer:
We're trying to keep stable labels, well... stable. Just because the parent subclause has a different label doesn't mean we need to rename the subordinate labels.

CaseyCarter:
> We're trying to keep stable labels, well... stable.

The stable names in question have only been present in a single revision of the WD. It seems reasonable to me to change them.

johelegp:
If you decide to change the stable labels, there are others to consider. The stable names of the non-adaptor CPOs under [range.factories] end in adaptor, rather than factory.

PR #2621, opened 2018-12-29
[special.mem.concepts] Use 'models ... only if' phrasing for semantic…

### LINKED ISSUE #2618: [special.mem.concepts] Unconventional description
The semantic requirements need to use "`T` models _`exposition-only-concept`_ only if" wording.

jensmaurer:
… constraints.

Fixes #2618.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -8125,7 +8125,8 @@ concept @\placeholdernc{no-throw-input-iterator}@ = // exposition only
 
 \begin{itemdescr}
 \pnum
-No exceptions are thrown from increment,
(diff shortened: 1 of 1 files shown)


jensmaurer:
The /no-throw-input-iterator/ and /no-throw-input-range/ seem obvious, but for /no-throw-sentinel/, it's unclear whether the "no exceptions on copy" constraint applies to S or to I or to both. @CaseyCarter, what's the intent here?

CaseyCarter:
> what's the intent here?

The intent - which I agree is not clear - is that the constraints on construction and assignment apply to both `I` and `S`.

PR #2627, opened 2018-12-31
[iterator.synopsis] Index fix: unreachable_sentinel is defined here.

### LINKED ISSUE #2626: [unreachable.sentinel] In "Index of library names", unreachable_sentinel should point to [iterators.general]
In "Index of library names", `unreachable_sentinel` should point to p. 854 of [iterators.general] rather than p. 890 of [unreachable.sentinel], like `default_sentinel` points to p. 849 of [iterators.general] (which I think p. 854 would be better).

jensmaurer:
Fixes #2626.


jensmaurer:
@tkoeppe, this is a small indexing fix. Please consider applying.

PR #2633, opened 2019-01-01
[expr.prim.id.dtor] Fix comment in example.

### LINKED ISSUE #2630: [expr.prim.id.dtor] Comment in example reflects lexing misunderstanding
The example in [[expr.prim.id.dtor]/3](http://eel.is/c++draft/expr.prim.id.dtor#3) shows the following line of code:

    0.T::~T(); // error: 0. is a floating-point literal ([lex.fcon])

However, contrary to what the comment suggests, I believe `0.` does not appear as a floating-point literal on this line. I believe that what really appears on this line is the [pp-number](http://eel.is/c++draft/lex.ppnumber#nt:pp-number) token `0.T` (which is not a valid floating literal token).

A very similar case is described in [[lex.pptoken]/4](http://eel.is/c++draft/lex.pptoken#4.sentence-1):
> The program fragment 0xe+foo is parsed as a preprocessing number token (one that is not a valid floating or integer literal token)"

Eelis:
In translation phase 3, the `0.T` in the line of code is parsed as a (single) preprocessing-token.

In phase 7, this preprocessing token is converted into a user-defined-floating-literal whose ud-suffix component is `T`.

At no point is the `0.` partial token a floating-point literal. At most, it is the fractional-constant component in a bigger user-defined-floating-literal token.

Fixes #2630.


source/expressions.tex
```diff
@@ -1509,7 +1509,7 @@
   C().C::~C();      // undefined behavior: temporary of type \tcode{C} destroyed twice
   using T = int;
   0 .T::~T();       // OK, no effect
-  0.T::~T();        // error: \tcode{0.} is a floating-point literal\iref{lex.fcon}
+  0.T::~T();        // error: \tcode{0.T} is a floating-point literal\iref{lex.fcon}
```
jensmaurer: No, "0.T" is not a floating-point literal; this is a literal with a _ud-suffix_.

Eelis: Do you agree it's a user-defined-floating-literal?

jensmaurer: Then, let's go with "error: 0.T is a user-defined-floating-literal". This seems to be the most precise of the choices.

PR #2634, opened 2019-01-02
Fix some `operator<=>` declarations with the wrong number of parameters

Quuxplusone:
I noticed one of these mistakes while reading [P0732R2](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0732r2.pdf); looks like when it was merged, the editor added the missing `operator<=>` (correctly) but left off the `friend` keyword (incorrectly).

The other I found by grepping the source for "` operator<=>`".

My understanding is that Herb Sutter's preferred spelling is `auto operator<=>(const T&) const = default;` — but that the alternative spelling `friend auto operator<=>(const T&, const T&) = default;` is also valid and has approximately the same semantics, and it's not a problem that some examples use one style and some the other.


source/overloading.tex
```diff
@@ -1794,10 +1794,10 @@
 \begin{example}
 \begin{codeblock}
 struct S {
-  auto operator<=>(const S&, const S&) = default; // \#1
-  bool operator<(const S&, const S&);             // \#2
+  std::strong_ordering operator<=>(const S&) const;  // \#1
+  bool operator<(const S&) const;                    // \#2
```
jensmaurer: Why did you choose this avenue, instead of adding "friend"?
In particular, why did you change "auto" to std::strong_ordering" here?

source/templates.tex
```diff
@@ -1280,7 +1280,7 @@
 
 struct A {
   constexpr A(const char*) {}
-  auto operator<=>(A, A) = default;
+  friend auto operator<=>(A, A) = default;
```
Rakete1111: Can you add a `const&` to both parameters here too? Thanks

jensmaurer: Good point.

Quuxplusone: Done, thanks. I had been unaware that the pass-by-value form was "valid, but not `=default`-able." Tim pointed me at [[class.compare.default]](http://eel.is/c++draft/class.compare.default) and set me straight.

I guess I'm still dubious of

    struct C;
    std::strong_ordering operator<=>(const C&, const C&);
    struct C {
        friend std::strong_ordering operator<=>(const C&, const C&);
    };
    std::strong_ordering operator<=>(const C&, const C&) = default;

IIUC, this `C` has a defaulted `operator<=>` but it is not structural (because it was not defined as defaulted *in the definition of `C`*) and incidentally is not `constexpr` either. I would feel better if this `C` were ill-formed. (Like, if `=default` were permitted only inside the definition of `C`, for these new operators.)

source/overloading.tex
Quuxplusone: 1: I was trying not to deviate from what I believe to be Herb's preferred syntax, `auto operator<=>(const S&) const = default;`.
2: But in this case, the fact that this operator is *defaulted* is not relevant to the example, so I removed the `= default`.
3: But without an `= default`, the use of `auto` is at least weird (but I guess not actually ill-formed, as long as the function definition also uses `auto`). So I replaced `auto` with `strong_ordering`.

Pulling on this thread probably wasn't worth it. I think I'll stop trying to use `operator<=>(const S&) const` and go back to my rule that all operators should be ADL friend functions. It makes life so much simpler, even if it is more verbose. :)

source/classes.tex
```diff
@@ -6796,8 +6796,8 @@
 \begin{codeblock}
 struct C {
   friend std::strong_equality operator<=>(const C&, const C&);
-  friend bool operator==(const C& x, const C& y) = default; // OK, returns \tcode{x <=> y == 0}
-  bool operator<(const C&) = default;                       // OK, function is deleted
+  friend bool operator==(const C&, const C&) = default; // OK, returns \tcode{x <=> y == 0}
+  friend bool operator<(const C&, const C&) = default;  // OK, function is deleted
```
Quuxplusone: The old version could also have been rescued by adding a trailing `const`; but I am being no longer apologetic for my love of ADL friends. :)

jensmaurer: As a general rule, we do want to show different coding styles in the core section of the C++ standard. (That's different for the library seciton). So, showing "the other syntax" for operator< seems perfectly fine (yes, it needs a "const").

source/overloading.tex
```diff
@@ -1810,10 +1810,10 @@
 \begin{example}
 \begin{codeblock}
 struct S {
-  std::weak_ordering operator<=>(const S&, int);  // \#1
-  std::weak_ordering operator<=>(int, const S&);  // \#2
+  friend auto operator<=>(const S&, int);  // \#1
+  friend auto operator<=>(int, const S&);  // \#2
```
jensmaurer: "auto" return types have extra restrictions (e.g. the function definition must be visible when calling the function). I'd suggest to revert this change.

source/templates.tex
```diff
@@ -1280,7 +1280,7 @@
 
 struct A {
   constexpr A(const char*) {}
-  auto operator<=>(A, A) = default;
+  friend auto operator<=>(const A&, const A&) = default;
```
jensmaurer: This change is good; the =default is the definition, so "auto" works.

source/overloading.tex
Quuxplusone: IIUC, the extra restrictions aren't relevant to this example. But I was just using `auto` to make the line shorter; so, if it's acceptable at this length, I don't object to putting back the explicit ordering. Done.

source/templates.tex
Quuxplusone: Yes, this example wouldn't work without the `=default`, because the operator needs to be structural.

Quuxplusone:
Ping.

PR #2638, opened 2019-01-05
[basic.fundamental] rename 'range exponent' to 'width'

jfbastien:
Jens Gustedt is working on adopting WG21's changes from wg21.link/p1236r1 into WG14's working paper, and noticed that what WG21 calls 'range exponent' has been called 'width' in the C standard. See 'integer types' in C17:

  The precision of an integer type is the number of bits it uses to represent values, excluding any sign and padding bits. The width of an integer type is the same but including any sign bit; thus for unsigned integer types the two values are the same, while for signed integer types the width is one greater than the precision.

C++ should use the same term where it makes sense and ones already exist, as is the case here. Jens wants to get C2x to match C++20, but it seems silly to make them change terms they've used for a long time.


jfbastien:
Something we could do to further reduce the tiny potential for confusion is define "width of an integer", and only use "width" when context makes it obvious. I'm happy to update to this as well.

jensmaurer:
CWG in Kona: approved renaming "range exponent" to "width".

hubert-reinterpretcast:
Cross referencing https://cplusplus.github.io/LWG/issue2820, which would refer to the term in question.

jensmaurer:
@jfbastien , this needs a rebase.

PR #2646, opened 2019-01-14
[depr.strstreambuf.virtuals] Hyphenate "implementation-defined"

Quuxplusone:


source/future.tex
```diff
@@ -1018,7 +1018,7 @@
 \begin{itemdescr}
 \pnum
 \effects
-Implementation defined, except that
+Implementation-defined, except that
```
jensmaurer: This should use \impldef so that we get an entry in the "index of implementation-defined behavior".

Quuxplusone: Probably true, but that's above my pay grade for this PR.

Consider whether you want a new macro `\Impldef`, or an optional argument to the existing `\impldef`, or simply to re-word this Effects clause so that it doesn't start with a capital `I`.

Quuxplusone: Updated. (Notice that this implementation-defined behavior is also deprecated; I don't know if that matters, but I assume not.)

PR #2647, opened 2019-01-14
Clear up a note about condition variables

kiroma:
I've been looking far and wide but nowhere in the papers about condition variables have I found note about efficiency of them being platform dependent. The implementation of `std::condition_variable_any` in libc++ and in libstdc++ is a generalization of `std::condition_variable`, which points towards the efficiency being bound more to the implementation than the platform.

### DIFF
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -2848,8 +2848,8 @@ public:
 Condition variables provide synchronization primitives used to block a thread until
 notified by some other thread that some condition is met or until a system time is
 reached. Class \tcode{condition_variable} provides a condition variable that can only
-wait on an object of type \tcode{unique_lock<mutex>}, allowing maximum efficiency on some
-platforms. Class \tcode{condition_variable_any} provides a general condition variable
+wait on an object of type \tcode{unique_lock<mutex>}, allowing better efficiency on some
+implementations. Class \tcode{condition_variable_any} provides a general condition variable
 that can wait on objects of user-supplied lock types.
 
 \pnum
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -2848,9 +2848,9 @@ public:
 Condition variables provide synchronization primitives used to block a thread until
 notified by some other thread that some condition is met or until a system time is
 reached. Class \tcode{condition_variable} provides a condition variable that can only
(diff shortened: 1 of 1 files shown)


jwakely:
The efficiency of `std::condition_variable` is of course platform-dependent, because it depends on the synchronization primitives provided by the OS and hardware. But this change seems to replace one vague, imprecise statement with another.

jensmaurer:
I'm mildly in favor of this change, because we usually talk about "implementation" in the C++ standard when we mean anything related to the compiler or OS platform. For example, [fs.conform.9945] consistently talks about "implementations", encompassing both the C++ library and the OS platform. I'm lukewarm about the "maximum" -> "better" change, although relative statements (as opposed to absolute ones) seem more appropriate for a standard document.

jwakely:
Yes, it looks like all the other uses of "platform" are at least in Notes (and although this preamble is only semi-normative, it's not actually in a Note).

kiroma:
With this PR I wanted to start off some discussion about the issue.
I should clear up that "better performance" in this case is relative to `std::condition_variable_any`.
Maybe instead of `cond_var` being faster it should be said that `gen_cond_var` can be slower?

jwakely:
Well the spec of `std::condition_variable` is intended to _allow_ the implementation to achieve maximum performance for a condition variable that supports waiting and notifying, i.e. it shouldn't add any abstraction penalty to the underlying OS primitives. If the implementation fails to achieve maximum performance, that's not the spec's fault :smile: 

`std::condition_variable_any` _does_ add additional abstraction, by supporting user-defined locks, and so generally won't be as fast as the OS primitives.

If we're going to change this, I'd also suggest changing "some implementations", as that is meaningless. Which ones? Does it tell a user whether _their_ implementation gets maximum performance?

How about:

> Class `condition_variable` provides a condition variable that can only wait on an object of type `unique_lock<mutex>`, **allowing the implementation to be more efficient**. Class `condition_variable_any` provides a general condition variable that can wait on objects of user-supplied lock types.

jensmaurer:
@jwakely, sounds good to me.

kiroma:
Yep, that sounds much better.

PR #2649, opened 2019-01-20
[tuple.helper] remove duplicate paragraph

jfbastien:


CaseyCarter:
This is normative damage: the first paragraph refers to the three partial specializations of `tuple_size`, and the second to the three partial specializations of `tuple_element`.

PR #2655, opened 2019-01-25
[basic.lval]/9 all prvalues can have cv-qualified types CWG2481

languagelawyer:
[dcl.init.ref]/5.3: "If the converted initializer is a prvalue, its type T4 is adjusted to type “cv1 T4” and the temporary materialization conversion is applied."
The cv-qualifiers of the reference being initialized are added to the initializer prvalue's type.

[expr.type]/2: "If a prvalue initially has the type “cv T”, where T is a cv-unqualified non-class, non-array type, the type of the expression is adjusted to T prior to any further analysis."
is not applicable here, because it says "initially", so it strips qualifiers only when a prvalue is formed, but [dcl.init.ref]/5.3 does not form a new prvalue, it adjusts the type of an existing one.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -269,8 +269,6 @@ that class shall not be an abstract class\iref{class.abstract}.
 A glvalue shall not have type \cv{}~\tcode{void}.
 \begin{note}
 A glvalue may have complete or incomplete non-\tcode{void} type.
-Class and array prvalues can have cv-qualified types; other prvalues
-always have cv-unqualified types. See \ref{expr.prop}.
 \end{note}
 
 \pnum


jensmaurer:
It would be nice to retain the statement that a non-class non-array prvalue is never cv-qualified. We can get there if we rephrase [dcl.init.ref] p5.3 to say "If the converted initializer is a prvalue, the temporary materialization conversion is applied to produce an xvalue of type cv1 T4." However, the current definition of "temporary materialization conversion" only changes from prvalue to xvalue and does not change the type.

languagelawyer:
> "If the converted initializer is a prvalue, the temporary materialization conversion is applied to produce an xvalue of type cv1 T4."

The problem is that "Temporary materialization conversion [conv.rval]" explicitly says that "A prvalue of type `T` can be converted to an xvalue of type `T`.", so saying that it is used to produce an xvalue of type `cv T` from a prvalue of type `T` would be kinda contradictory.
Before [conv.rval] was introduced, [dcl.init.ref] itself was responsible for creating objects, saying that a temporary object of type `cv T` is created.

cv-qualified prvalues of non-class, non-array types are still not "observable", but now, when the Standard not only strips cv-qualifiers from prvalue of such types, but also adds them, this argument became less convincing.

zygoloid:
If we rephrase [dcl.init.ref] as suggested, we would lose the cv-qualifiers in the type of the temporary object. That breaks
```
const int &r = 0;
constexpr int n = r;
```

jensmaurer:
Editorial meeting: say "is immediately applied" and add a note per #2657.

tkoeppe:
@languagelawyer: could you please take a look (and rebase)?

languagelawyer:
@tkoeppe CWG2481 changed [dcl.init.ref] to say
> Otherwise, the initializer expression is implicitly converted to a prvalue of type `T1`. The temporary materialization conversion is applied<b>, considering the type of the prvalue to be “_cv1_ `T1`”,</b> and the reference is bound to the result.

I suppose «considering the type to be» ≠ «has the type», so that this PR can be closed, but I'd like to have confirmation.

jensmaurer:
Yes, that's the intent.

tkoeppe:
Thanks, all!

PR #2656, opened 2019-01-25
[algorithms] Qualify declarator-id with sub-namespace.

jensmaurer:
Partially addresses #2512.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -2898,14 +2898,12 @@ template<class ExecutionPolicy, class ForwardIterator, class Predicate>
   bool all_of(ExecutionPolicy&& exec, ForwardIterator first, ForwardIterator last,
               Predicate pred);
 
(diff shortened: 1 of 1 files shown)


jensmaurer:
I've just prefixed the declarator-id, but that exhibits an issue: While the concept names are all in namespace std (good), the return types such as safe_iterator_t or blah_result are not; they are in namespace std::ranges. There are three approaches:
- transform to trailing return types
- qualify the return type with "ranges::", too
- ignore the issue and assume (contrary to compiler behavior) that names in the return type are resolved by starting lookup in the declarator-id's namespace.

jensmaurer:
Editorial meeting:
- If there is a possibility of confusion, use namespace markings. Qualify declarator-id and return type etc.
- Otherwise, no change.

jensmaurer:
Rebased and return types qualified.

PR #2657, opened 2019-01-27
[dcl.init.ref]/5 Prevent prvalue type adjustment CWG2481

languagelawyer:
Fresh prvalues could fall prey to [expr.type]/2: "If a prvalue initially has the type “cv T”, where T is a cv-unqualified non-class, non-array type, the type of the expression is adjusted to T prior to any further analysis."

I've also rewritten the sentence in the semantic line breaks style.

However, this would be a wrong fix if it is decided to restore the "no cv-qualified non-class non-array prvalues" status quo (see #2655).

See CWG2481

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -5319,8 +5319,8 @@ copy-initialization, is then used to direct-initialize the reference.
 For this direct-initialization, user-defined conversions are not considered.
 \item
 Otherwise,
-the initializer expression is implicitly converted to a prvalue
-of type ``\cvqual{cv1} \tcode{T1}''.
+the initializer expression is implicitly converted to a prvalue of type \tcode{T1},
+then the type of the resulting prvalue is adjusted to ``\cvqual{cv1} \tcode{T1}''.
 The temporary materialization conversion is applied and the reference is
 bound to the result.
 \end{itemize}


languagelawyer:
The issue was spotted in https://stackoverflow.com/q/54381791/9585016

source/declarations.tex
```diff
@@ -5319,8 +5319,8 @@
 For this direct-initialization, user-defined conversions are not considered.
 \item
 Otherwise,
-the initializer expression is implicitly converted to a prvalue
-of type ``\cvqual{cv1} \tcode{T1}''.
+the initializer expression is implicitly converted to a prvalue of type \tcode{T1},
+then the type of the resulting prvalue is adjusted to ``\cvqual{cv1} \tcode{T1}''.
```
zygoloid: Seems clunky to convert to one type then adjust to another. There's also a risk that cv-qualifiers in the destination type of the implicit conversion could matter in some contexts. How about:

> Otherwise, the initializer expression is implicitly converted to a prvalue of type ''*cv1* `T`'' and the temporary materialization conversion is immediately applied; the reference is bound to the result of this conversion.

or similar?

languagelawyer: Do you think that it is clear that "immediately" means [expr.type]/2 does not apply?

jensmaurer:
Editorial meeting: Use Richard's suggestion and add a note saying that the cv-removal for prvalues is not done.

tkoeppe:
@languagelawyer: could you please rebase this and apply the above suggestion?

languagelawyer:
This was fixed differently https://github.com/cplusplus/draft/pull/4693/commits/fc923d6c35df85da91c0b005e8fdfd6269337247

PR #2662, opened 2019-02-02
[meta.rel] Add missing noexcept cross-refs for invokable traits

AlisdairM:
All the other traits that use the phrase 'is known not
to throw exceptions' also cross-reference the core
clause for the noexcept operator, so add the missing
cross reference to the more recently added traits.


jensmaurer:
@tkoeppe , one for you.

PR #2666, opened 2019-02-12
[temp.spec]/2 dead wording

18:
The terms defined in [temp.spec]/2 are defined, but never used again, anywhere in the standard. This is possibly a case of dead wording. 

[Here](http://eel.is/c++draft/temp.spec#2) is a link to the section in question.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -5483,23 +5483,6 @@ a member template is referred to as
 \defn{template instantiation}.
 
 \pnum
-A function instantiated from a function template is called an instantiated
-function.
-A class instantiated from a class template is called an instantiated class.
-A member function, a member class, a member enumeration, or a static data member of a class template
-instantiated from the member definition of the class template is called,
-respectively, an instantiated member function, member class, member enumeration, or static data
-member.
-A member function instantiated from a member function template is called an
-instantiated member function.
-A member class instantiated from a member class template is called an
-instantiated member class.
-A variable instantiated from a variable template is called an
-instantiated variable.
-A static data member instantiated from a static data member template
-is called an instantiated static data member.
-
-\pnum
 An explicit specialization may be declared for a function template,
 a class template, a member of a class template or a member template.
 An explicit specialization declaration is introduced by


Rakete1111:
You having grep-ed long enough :) See for example. http://eel.is/c++draft/temp#inst-11.sentence-1

Also a lot of places use "instantiation", "instantiated" so it's not like those terms are no longer needed. Also there is a whole clause for instantations [[temp.spec]](http://eel.is/c++draft/temp.spec). I don't think the term is as dead as you think it is :)

18:
@Rakete1111 In your example you provide, the term "instantiated class" is not used. In that sentence, instantiated is part of the word "implicitly instantiated", not "instantiated class/function...". The sentence would be read like this: " Implicitly instantiated (class, function, and variable template specializations) are placed in the namespace where the template is defined. The sentence is referring to implicitly instantiated specializations, not instantiated classes or functions.

Rakete1111:
Ah right, that was a bad example, my bad. Here's one (http://eel.is/c++draft/temp.spec#4.sentence-2):

> A specialization is a class, function, or class member that is either instantiated or explicitly specialized

[unrelated: seems to be missing an enumeration and a variable here]

I don't know if we should drop the term however since we're not really using it, as you say.

18:
@Rakete1111 Again, in your example there is no instance of the terms "instantiated class, instantiated function, instantiated member class..." being used. I'm am not proposing to remove the use of the word "instantiated" or "instantiation", I am proposing to remove "instantiated class/function" and family, as they are never used anywhere outside of the definition.

Rakete1111:
I would argue that a "class that is instantiated" does indeed use the definition you are trying to remove. Where else is that property defined; did I miss something?

18:
@Rakete1111 The definition of specialization specifies that the class/function in question is in a state, either instantiated or explicitly specialized. The process of instantiation is described in http://eel.is/c++draft/temp.spec#1.

jensmaurer:
The comments in the example in [temp.explicit] p5 seem to use the term "instantiated variable".

Let's note make the proposed change.

18:
@jensmaurer These still is the issue of the paragraph cited here (more specifically the phrase "a function instantiated from a function template is called a instantiated function" being the the exact definition used for a function specialization http://eel.is/c++draft/temp#fct.spec-1.sentence-1

PR #2668, opened 2019-02-15
[dcl.attr.contract.cond] example fails to return

jfbastien:
The function in the example returns int but fails to actually return.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -8454,11 +8454,10 @@ int f(int x)
   return ++x;                   // undefined behavior
 }
 
-int g(int * p)
-  [[ensures r: p != nullptr]]
+void g(int * p)
+  [[ensures: p != nullptr]]
 {
   *p = 42;                      // OK, \tcode{p} is not modified
-  return *p;
 }
 
 int h(int x)
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -8458,7 +8458,7 @@ int g(int * p)
   [[ensures r: p != nullptr]]
 {
   *p = 42;                      // OK, \tcode{p} is not modified
-  return p;
+  return *p;
 }
 
 int h(int x)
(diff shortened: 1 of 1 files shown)


source/declarations.tex
```diff
@@ -8458,6 +8458,7 @@
   [[ensures r: p != nullptr]]
 {
   *p = 42;                      // OK, \tcode{p} is not modified
+  return p;
```
zygoloid: `p` is not an `int`. Maybe change the return type to `void` and remove the `r` from the contract?

jfbastien: Changed to `*p`.

jfbastien: I think this is trying to show something about `ensures` with returns?

CaseyCarter: The paragraph to which this example is attached:

> If a postcondition odr-uses a parameter in its predicate and the function body makes direct or indirect modifications of the value of that parameter, the behavior is undefined

suggests it's trying to show that it's ok to refer to a parameter in an `ensures` when the function body doesn't modify it. The `r` is extraneous here, as is the `int` return value.

(Mind you, I'm unsure how the example in [dcl.attr.contract.syn]/6:
```c++
void push(int x, queue & q)
  [[expects: !q.full()]]
  [[ensures: !q.empty()]]
{
  /* ... */
  [[assert: q.is_valid()]];
  /* ... */
}
```
is supposed to ensure `!q.empty()` without "directly or indirectly modifying the value" of `q` - but *this* particular example seems sensible.)

zygoloid: The intent here is that the "value of the parameter" for a reference parameter is just the lvalue to which the reference is bound (which obviously cannot change during the execution of the function). Put another way, evaluating the postcondition in the caller with the values that it passed in should do the same thing as evaluating it in the callee. Can you open a separate editorial issue for that?

CaseyCarter: > Can you open a separate editorial issue for that?

Filed as #2669.

jfbastien: Updated accordingly.

PR #2676, opened 2019-02-22
[container.node] Add 3 template keywords for dependent name

birbacher:
On:
[container.node.overview]/4
[container.node.cons]/3.1
[container.node.dtor]/1

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -1265,7 +1265,7 @@ private:
   using container_node_type = @\unspecnc@;
   using ator_traits = allocator_traits<allocator_type>;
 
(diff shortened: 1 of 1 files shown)


jwakely:
Since the template parameters of the class type are not specified, it's not actually clear that the allocator is a dependent type in a node-handle (but in practice it's supposed to be).

birbacher:
> Since the template parameters of the class type are not specified, it's not actually clear that the allocator is a dependent type in a node-handle (but in practice it's supposed to be).

The wording in table 70 makes it dependent on the arguments of the container which uses the node handle, so the type must differ in some way.

tkoeppe:
Thanks!

PR #2722, opened 2019-03-01
P1227R2 Signed ssize() functions, unsigned size() functions

### LINKED ISSUE #2706: [2019-02 LWG Motion 15] P1227R2 Signed ssize() functions, unsigned size() functions
[P1227R2](https://wg21.link/p1227r2)

jensmaurer:
Fixes #2706.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -438,6 +438,9 @@ namespace std {
 
   template<class C> constexpr auto size(const C& c) -> decltype(c.size());
   template<class T, size_t N> constexpr size_t size(const T (&array)[N]) noexcept;
+  template<class C> constexpr auto ssize(const C& c)
+    -> common_type_t<ptrdiff_t, make_signed_t<decltype(c.size())>>;
+  template<class T, ptrdiff_t N> constexpr ptrdiff_t ssize(const T (&array)[N]) noexcept;
   template<class C> [[nodiscard]] constexpr auto empty(const C& c) -> decltype(c.empty());
   template<class T, size_t N> [[nodiscard]] constexpr bool empty(const T (&array)[N]) noexcept;
   template<class E> [[nodiscard]] constexpr bool empty(initializer_list<E> il) noexcept;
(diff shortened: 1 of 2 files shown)


source/iterators.tex
```diff
@@ -6520,6 +6523,28 @@
...
+\indexlibrary{\idxcode{ssize(C\& c)}}%
+\begin{itemdecl}
+template <class C>
+constexpr auto ssize(const C& c)
+    -> common_type_t<ptrdiff_t, make_signed_t<decltype(c.size())>>;
+\end{itemdecl}
+\begin{itemdescr}
+\pnum \returns
+\begin{codeblock}
+static_cast<common_type_t<ptrdiff_t, make_signed_t<decltype(c.size())>>>(c.size())
```
burblebee: Append a period (I realize this is in a codeblock, but it's really a sentence).

source/iterators.tex
```diff
@@ -6520,6 +6523,28 @@
 \pnum \returns \tcode{N}.
 \end{itemdescr}
 
+\indexlibrary{\idxcode{ssize(C\& c)}}%
+\begin{itemdecl}
+template <class C>
+constexpr auto ssize(const C& c)
```
burblebee: Place on same line so it appears as it does in the declaration in the synopsis.

source/iterators.tex
```diff
@@ -6520,6 +6523,28 @@
...
+    -> common_type_t<ptrdiff_t, make_signed_t<decltype(c.size())>>;
```
burblebee: Indent only 2 spaces (as it appears in the declaration in the synopsis).

source/iterators.tex
```diff
@@ -6520,6 +6523,28 @@
...
+\pnum \returns
+\begin{codeblock}
+static_cast<common_type_t<ptrdiff_t, make_signed_t<decltype(c.size())>>>(c.size())
+\end{codeblock}
+\end{itemdescr}
+
+\indexlibrary{\idxcode{ssize(T (\&array)[N])}}%
+\begin{itemdecl}
+template <class T, ptrdiff_t N>
+constexpr ptrdiff_t ssize(const T (&array)[N]) noexcept;
```
burblebee: Place on same line so it appears as it does in the declaration in the synopsis.

source/iterators.tex
jensmaurer: Sorry, that wouldn't be adequate, because the period would be in code font inside a codeblock. I don't think we do that elsewhere.

source/iterators.tex
jensmaurer: Fixed.

source/iterators.tex
jensmaurer: Fixed.

source/iterators.tex
jensmaurer: Fixed.

source/iterators.tex
burblebee: Yes, you're right.  I assuming you're using codeblocks to prevent an Overfull?  What we tend to do in this case is split the expr over two lines with a \tcode{} for the part of the expr that should go on each line.  If you want the whole expression on one line, would it work to tcode the whole expr, prefixing it with a newline and an indent?  In any case, what follows \returns is supposed to be a sentence, while codeblocks should be used for complete code snipets.

johelegp: According to the [Specification Style Guidelines](https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#writing-returns-in-a-function-description), codeblocks are also used for long expressions.

burblebee: @johelegp Thanks for pointing that out!  I don't remember seeing that the last time I read that page.  Is there a way to "see what's new" there?

johelegp: If you're watching the repository, you should be able to see when a new there's a new [revision](https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines/_history). It should appear in your GitHub dashboard. For more visibility, I follow my GitHub feed through RSS.

PR #2729, opened 2019-03-04
P1185R2 <=> != ==

### LINKED ISSUE #2692: [2019-02 CWG Motion 12] P1185R2 <=> != ==
[P1185R2](https://wg21.link/p1185r2)

jensmaurer:
Fixes #2692.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -6750,12 +6750,9 @@ with parameters \tcode{x} and \tcode{y} is defined as deleted if
   results in an ambiguity or a function
   that is deleted or inaccessible from the operator function, or
 \item
-  \tcode{x == y} or \tcode{y == x}
-  cannot be contextually converted to \tcode{bool}.
+  \tcode{x == y} cannot be contextually converted to \tcode{bool}.
 \end{itemize}
-Otherwise, the operator function yields \tcode{(x == y) ? false : true}
-if an operator \tcode{==} with the original order of parameters was selected,
-or \tcode{(y == x) ? false : true} otherwise.
+Otherwise, the operator function yields \tcode{(x == y) ?\ false :\ true}.
 
 \pnum
 \begin{example}
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -6664,6 +6664,11 @@ Otherwise, the implicitly-declared \tcode{==} operator function is of the form
 \begin{codeblock}
 friend bool operator==(const X&, const X&);
 \end{codeblock}
+\begin{note}
+Such a friend function is visible
+to argument-dependent lookup\iref{basic.lookup.argdep}
+only\iref{namespace.memdef}.
+\end{note}
 The operator is a \tcode{constexpr} function if its definition
 would satisfy the requirements for a \tcode{constexpr} function.
 \begin{note}
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -6642,28 +6642,51 @@ shall be a non-template function
 declared in the \grammarterm{member-specification} of \tcode{C}
 that is
 \begin{itemize}
-\item a non-static member of \tcode{C} having one parameter of type \tcode{const C\&}, or
+\item a non-static const member of \tcode{C} having one parameter of type \tcode{const C\&}, or
 \item a friend of \tcode{C} having two parameters of type \tcode{const C\&}.
 \end{itemize}
 
 \pnum
-\indextext{structural comparison operator|see{operator, structural comparison}}%
-A three-way comparison operator for a class type \tcode{C}
-is a \defnadj{structural comparison}{operator}
-if it is defined as defaulted in the definition of \tcode{C},
-and all three-way comparison operators it invokes
-are structural comparison operators.
-\indextext{strong structural equality|see{equality, strong structural}}%
-A type \tcode{T}
-has \defnadj{strong structural}{equality}
-if, for a glvalue \tcode{x} of type \tcode{const T},
-\tcode{x <=> x} is a valid expression
-of type \tcode{std::strong_ordering} or \tcode{std::strong_equality}
-and either does not invoke a three-way comparison operator
-or invokes a structural comparison operator.
+If the class definition
+does not explicitly declare an \tcode{==} operator function,
+but declares a defaulted three-way comparison operator function,
+an \tcode{==} operator function is declared implicitly
+with the same access as the three-way comparison operator function.
+The implicitly-declared \tcode{==} operator for a class \tcode{X}
+is an inline member and is defined as defaulted in the definition of \tcode{X}.
+If the three-way comparison operator function
+is declared as a non-static const member,
+the implicitly-declared \tcode{==} operator function is a member of the form
+\begin{codeblock}
+bool X::operator==(const X&) const;
+\end{codeblock}
+Otherwise, the implicitly-declared \tcode{==} operator function is of the form
+\begin{codeblock}
+friend bool operator==(const X&, const X&);
+\end{codeblock}
+The operator is a \tcode{constexpr} function if its definition
+would satisfy the requirements for a \tcode{constexpr} function.
+\begin{note}
+The \tcode{==} operator function is declared implicitly even if
+the defaulted three-way comparison operator function is defined as deleted.
+\end{note}
 
-\rSec2[class.spaceship]{Three-way comparison}
-\indextext{operator!three-way comparison!defaulted}%
+\pnum
+\indextext{structural comparison operator|see{operator, structural comparison}}%
+A type \tcode{C} has \defnadj{strong structural}{equality} if,
+given a glvalue \tcode{x} of type \tcode{const C}, either:
+\begin{itemize}
+\item
+  \tcode{C} is a non-class type and \tcode{x <=> x} is a valid expression
+  of type \tcode{std::strong_ordering} or \tcode{std::strong_equality}, or
+\item
+  \tcode{C} is a class type with an \tcode{==} operator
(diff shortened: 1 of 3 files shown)


source/classes.tex
```diff
@@ -6642,28 +6642,51 @@
...
+with the same access as the three-way comparison operator function.
+The implicitly-declared \tcode{==} operator for a class \tcode{X}
+is an inline member and is defined as defaulted in the definition of \tcode{X}.
+If the three-way comparison operator function
+is declared as a non-static const member,
+the implicitly-declared \tcode{==} operator function is a member of the form
+\begin{codeblock}
+bool X::operator==(const X&) const;
+\end{codeblock}
+Otherwise, the implicitly-declared == operator function is of the form
```
CaseyCarter:
```suggestion
Otherwise, the implicitly-declared \tcode{==} operator function is of the form
```

BRevzin:
Walter suggested adding something like this to the end of [class.compare.default]/1 to indicate that in the case that the implementation provides a hidden friend declaration of `==` that it's actually _hidden_ (i.e. the implementation is not allowed to additionally provide a non-member declaration). Should be editorial:

> In the latter case, there shall be no declaration or redeclaration of the `==` operator function outside `C`'s _member-specification_ [*Note*: The intent is that a two-argument defaulted operator function is to be found via argument-dependent lookup only. *- end note*]

source/classes.tex

jensmaurer:
@BRevzin: Added a note; see the commit for details.

burblebee:
@CaseyCarter I assigned this to you thinking that you had done a complete review of the changes - was that a mistake?  Just trying to make sure a "member" does a complete review here... Please let me know if I should reassign this to someone else.  Thanks!

CaseyCarter:
> I assigned this to you thinking that you had done a complete review of the changes - was that a mistake?

I have *not* made a proper review, but I did click "Start review" instead of "Add single comment", presumably prompting you to assign the PR to me. I plan to do a proper review this evening as an object lesson to remind me to choose my buttons more carefully.

source/classes.tex
```diff
@@ -6681,12 +6707,72 @@
...
+unless, for each $\tcode{x}_i$ in the expanded list of subobjects
+for an object \tcode{x} of type \tcode{C},
+$\tcode{x}_i\tcode{ == }\tcode{x}_i$ is a valid expression and
+contextually convertible to \tcode{bool}.
 
 \pnum
+The return value \tcode{V} of a defaulted \tcode{==} operator function
+with parameters \tcode{x} and \tcode{y} is determined
+by comparing corresponding elements $\tcode{x}_i$ and $\tcode{y}_i$
+in the expanded lists of subobjects for \tcode{x} and \tcode{y}
```
CaseyCarter: This notably doesn't specify the order in which corresponding elements are compared. Without an ordering, this wording allows comparing the same pair of corresponding elements more than once. @BRevzin, is it intended that the comparison order is unspecified, or should we have something like "in order by increasing subscript" here?

source/classes.tex
```diff
@@ -6681,12 +6707,72 @@
...
+\end{codeblock}
+\end{example}
+
+\rSec2[class.spaceship]{Three-way comparison}
+\indextext{operator!three-way comparison!defaulted}%
+
+\pnum
+Given an expanded list of subobjects for an object \tcode{x} of type \tcode{C},
+the type of the expression $\tcode{x}_i$ \tcode{<=>} $\tcode{x}_i$
+is denoted by $\tcode{R}_i$.
```
CaseyCarter: Given:
```c++
struct X {};
struct Y {
  X x;
  auto operator<=>(const Y&) = default;
};
```
This tells me that R_0 denotes the type of an ill-formed expression, which is nonsense. I think this should read:
```tex
Given [...],
$\tcode{R}_i$ denotes the type of the expression $\tcode{x}_i$ \tcode{<=>} $\tcode{x}_i$ 
if it is well-formed, and otherwise denotes \tcode{void}.
```

source/templates.tex
```diff
@@ -1945,7 +1942,7 @@
 have the same type and value
 after conversion to the type of the \grammarterm{template-parameter},
 where they are considered to have the same value if they compare equal
-with \tcode{operator<=>}, and}
+with \tcode{operator==}, and}
```
CaseyCarter: Pre-existing: `\tcode{operator==}` seems to only apply to class types. Shouldn't this say "if they compare equal with the \tcode{==} operator"?

W-E-Brown:
> On Mar 6, 2019, at 11:51 AM, Casey Carter <notifications@github.com> wrote:
> @CaseyCarter requested changes on this pull request.
> 
> In source/classes.tex <https://github.com/cplusplus/draft/pull/2729#discussion_r263008082>:
> 
<snip>
> +\begin{note}
> +Such a friend function is not visible to unqualified lookup\iref{namespace.memdef}.
> ...or to qualified name lookup. Maybe it would be better to say can only be found via argument dependent lookup\iref{basic.lookup.argdep} rather than describing the complement?
> 
Such an affirmative approach would be my preference, too, but let's move "only" to the end of that sentence.  (Otherwise, we're saying that such a friend function can only be found [and not, for example, be eaten, be painted, or be shot ☺︎] via ADL.)

Suggested phrasing for the Note in question:

Such a friend function is found via argument-dependent lookup ([basic.lookup.argdep]) only.

<snip>

CaseyCarter:
> Otherwise, we're saying that such a friend function can only be found [and not, for example, be eaten, be painted, or be shot ☺︎] via ADL.

As a writer of generic libraries, I would be unsurprised if ADL were to eat, paint, or shoot my functions.

source/classes.tex
BRevzin: I agree, I added the same idea to my wording for [P1186](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1186r1.html#wording) but I like your wording better than what I proposed there (I also changed the <code>R<sub>i</sub></code>'s to <code>S<sub>i</sub></code>'s to avoid confusion with `R` which is totally unrelated - but we can cross that bridge in that paper.)

source/classes.tex
BRevzin: Is that not implied by "until the first index `i` where ..."? 

If not, we need to add "in order by increasing subscript" to [class.spaceship]/3 as well, since there's where I copied this wording from.

source/classes.tex
```diff
@@ -6642,28 +6642,54 @@
...
+the implicitly-declared \tcode{==} operator function is a member of the form
+\begin{codeblock}
+bool X::operator==(const X&) const;
+\end{codeblock}
+Otherwise, the implicitly-declared \tcode{==} operator function is of the form
+\begin{codeblock}
+friend bool operator==(const X&, const X&);
+\end{codeblock}
+\begin{note}
+Such a friend function is not visible to unqualified lookup\iref{namespace.memdef}.
```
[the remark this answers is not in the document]
jensmaurer: We should leave the cross-reference to [namespace.memdef] in, because that's curiously the only place where we clearly say how friend name injection works.
Otherwise addressed, including moving "only" to the end.

source/classes.tex
jensmaurer: I don't think we should address this (if it needs addressing at all) in the paper merge cycle.

source/classes.tex
jensmaurer: Is this a new bug introduced by this paper, or an existing issue? Seems like an existing issue to me. Is this change even editorial to start with?

source/templates.tex
jensmaurer: Sounds reasonable. I'll also put in a cross-reference to [expr.eq].

jensmaurer:
@burblebee, I think maybe anyone making a review comment gets assigned as "reviewer". In any case, I'd expect a "full" reviewer to approve after all the fixes have been applied.

burblebee:
> @burblebee, I think maybe anyone making a review comment gets assigned as "reviewer". In any case, I'd expect a "full" reviewer to approve after all the fixes have been applied.

Yes, I'm often thrown off by this, which is why I've resorted to assigning reviewers, so that I know if someone is going to review the PR, and to make sure that someone knows to do a complete review (and approve, suggest changes, etc.).  There are too many PRs for all of us (or Richard) to review each one; this spreads the load.

source/classes.tex
CaseyCarter: > Is that not implied by "until the first index `i` where ..."?

That tells me that the process terminates upon finding such an index, it doesn't specify the order in which indices are chosen.

> If not, we need to add "in order by increasing subscript" to [class.spaceship]/3 as well, since there's where I copied this wording from.

+1

> I don't think we should address this (if it needs addressing at all) in the paper merge cycle.

I'd be happy to merge this as is and follow up with a core issue to determine the ordering if you don't think this falls under "editorial clarification of the wording".

source/classes.tex
CaseyCarter: I agree, this is pre-existing - I hadn't paid enough attention to this wording to notice it before. I'll submit a core issue.

source/classes.tex
jensmaurer: Please create a separate editorial issue for this. Let's see whether Richard is willing to address this editorially.

source/classes.tex
jensmaurer: Thanks.

source/classes.tex
CaseyCarter: Filed as https://github.com/cplusplus/draft/issues/2748. Feel free to mark this discussion "resolved".

burblebee:
@CaseyCarter many thanks for the full review :)

PR #2731, opened 2019-03-04
P1009R2 Array size deduction in new-expressions

### LINKED ISSUE #2690: [2019-02 CWG Motion 10] P1009R2 Array size deduction in new-expressions
[P1009R2](https://wg21.link/p1009r2)

jensmaurer:
Fixes #2690.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -4433,7 +4433,7 @@ object created by the \grammarterm{new-expression} has a cv-qualified type.
 
 \begin{bnf}
 \nontermdef{noptr-new-declarator}\br
-    \terminal{[} expression \terminal{]} \opt{attribute-specifier-seq}\br
+    \terminal{[} \opt{expression} \terminal{]} \opt{attribute-specifier-seq}\br
     noptr-new-declarator \terminal{[} constant-expression \terminal{]} \opt{attribute-specifier-seq}
 \end{bnf}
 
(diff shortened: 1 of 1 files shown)


burblebee:
Review completed - see comments below:
* The paper directs that the new text at the end of paragraph 6 should be part of paragraph 6, while the changes in the PR adds it as a new paragraph.

source/expressions.tex
```diff
@@ -4532,6 +4532,15 @@
 constant expression).
 \end{example}
 
+\pnum
```
burblebee: Should this be part of paragraph 6 as directed by the paper?

jensmaurer: First, the editing instructions in the incoming paper do say "change p6", but the actual layout of the paper text suggests that the text should actually be in a new paragraph.
Second, p6 (actually p7, it seems) talks about (really-present) _constant-expression_s and _expression_s in the _new-type-id_, whereas the new sentence talks about the absence of an expression.
So, actually I think the right fix is to move the new text slightly earlier, before we start talking about _expression_.

jensmaurer: Add-on patch (totally optional) pushed; @burblebee, what do you think?

burblebee: I like it :).

source/expressions.tex
```diff
@@ -4542,9 +4539,11 @@
 the \grammarterm{new-initializer}~(\ref{dcl.init.aggr}, \ref{dcl.init.string}).
 
 \pnum
+\indextext{\idxcode{new}}%
+If the \grammarterm{expression} in a \grammarterm{noptr-new-declarator}
+is present, it is implicitly converted to \tcode{std::size_t}.
 \indextext{function!allocation}%
```
burblebee: Do you think this index should precede the new text?

jensmaurer: It's existing stuff, and I'm not too happy about the indexing here in general. I'd like to leave it as-is for now.

PR #2733, opened 2019-03-05
P1459R1 Mandating the Standard Library: Clause 18 - Diagnostics library

### LINKED ISSUE #2699: [2019-02 LWG Motion 8] P1459R1 Mandating the Standard Library: Clause 18 - Diagnostics library
[P1459R1](https://wg21.link/p1459r1)

jensmaurer:
Fixes #2699.

### DIFF
--- a/source/diagnostics.tex
+++ b/source/diagnostics.tex
@@ -101,11 +101,6 @@ logic_error(const string& what_arg);
 
 \begin{itemdescr}
 \pnum
-\effects
-Constructs an object of class
-\tcode{logic_error}.
-
-\pnum
 \ensures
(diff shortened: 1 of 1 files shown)


source/diagnostics.tex
```diff
@@ -939,6 +849,8 @@
 such type. \end{note}
 
 \indexlibrary{\idxcode{error_category}}%
+\indexlibrary{\idxcode{error_category}!constructor}%
+\indexlibrary{\idxcode{error_category}!destructor}%
```
burblebee: Moving the indexes to here seems a bit odd, but so does having decls in the synopsis with no corresponding itemdecls.

PR #2743, opened 2019-03-05
P0960r3 Allow initializing aggregates from a parenthesized list of va…

### LINKED ISSUE #2689: [2019-02 CWG Motion 9] P0960R3 Allow initializing aggregates from a parenthesized list of values
[P0960R3](https://wg21.link/p0960r3)

burblebee:
Fixes #2689.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -4406,15 +4406,17 @@ type is the same class as, or a derived class of, the class of the destination,
 constructors are considered.
 The applicable constructors
 are enumerated\iref{over.match.ctor}, and the best one is chosen
-through overload resolution\iref{over.match}.
-The constructor so selected
+through overload resolution\iref{over.match}. Then:
+\begin{itemize}
+\item
+If overload resolution is successful,
+the selected constructor
 is called to initialize the object, with the initializer
 expression or \grammarterm{expression-list} as its argument(s).
-If no constructor applies and the destination type is not an aggregate, or
-the overload resolution is ambiguous, the initialization is ill-formed.
 \item
-Otherwise, if the destination type is
-a (possibly cv-qualified) aggregate class \tcode{A} and
+Otherwise, if no constructor is viable,
+the destination type is
+a (possibly cv-qualified) aggregate class \tcode{A}, and
 the initializer is a parenthesized \grammarterm{expression-list},
 the object is initialized as follows.
 Let $e_1$, $\dotsc$, $e_n$ be the elements of the aggregate\iref{dcl.init.aggr}.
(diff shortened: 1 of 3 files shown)


jensmaurer:
Please use uppercase "r" in the paper number of the commit message.

source/declarations.tex
```diff
@@ -4402,6 +4417,48 @@
 The call is used
 to direct-initialize, according to the rules above, the object
 that is the destination of the copy-initialization.
+\item
```
jensmaurer: I think this should precede bullet 17.6.3, because the preceding bullet has an unqualified "otherwise" that already eats the remainder of the cases. Also, the preceding bullet claims "copy initialization cases", but what we have here is direct-initialization.

source/declarations.tex
```diff
@@ -4361,7 +4361,22 @@
...
+the object is initialized as follows.
+Let $x_1$, $\dotsc$, $x_k$ be
+the elements of the \grammarterm{expression-list}.
+If the destination type is an array of unknown bound,
+it is defined as having $k$ elements.
+If $k$ is greater than the size of the array,
+the program is ill-formed.
+Otherwise, the $i^\text{th}$ array element is copy-initialized with
+$x_i$ for each $1 \leq i \leq k$, and
+value-initialized for each $k < i \leq n$.
```
jensmaurer: We don't introduce "n" anywhere for this array case. We could say "If $k$ is greater than the size of the array $n$" a few lines prior to here.

tkoeppe: Maybe add:

> as having $k$ elements<ins>; either way, let $n$ be the array size of the destination type</ins>.

burblebee: I went with:
`... ; either way, let $n$ denote the array size.`
I'm not thrilled with "either way", but it's better than nothing.

tkoeppe: Would you prefer "in either case"?

source/declarations.tex
burblebee: Done. @jensmaurer, please have a look now to make sure it's what you intended.

source/declarations.tex
burblebee: I think that's slightly better, but what bothers me is, what are the cases?  this is in a sentence where we've only introduced one case.  Making it a separate sentence still has a similar problem.

tkoeppe: Oh, right: the two cases are known and unknown bound. So you could say, "regardless of whether the original type was of known or unknown bound, let $n$ denote the array size."

Or: "Let $n$ denote the array size after this potential adjustment."

source/declarations.tex
zygoloid: The new bullet was still unreachable. Massaged more to make it reachable.

PR #2745, opened 2019-03-05
Motions 2019 02 lwg 2

### LINKED ISSUE #2711: [2019-02 LWG Motion 2] P1457R0 C++ Standard Library Issues (Kona 2019)

burblebee:
Fixes #2711.

Drafting notes:
* LWG3180: I didn't indent "using" (as we had discussed in Kona), in order to be consistent with the existing coding style.
* LWG3112: I used `\hfill\break` to workaround an Overfull issue.  Is there a better way to do this?

### DIFF
--- a/source/atomics.tex
+++ b/source/atomics.tex
@@ -1350,8 +1350,18 @@ namespace std {
 
 \indexlibrary{\idxcode{atomic}}%
 \pnum
-The template argument for
-\tcode{T} shall be trivially copyable\iref{basic.types}. \begin{note} Type arguments that are
+The template argument for \tcode{T} shall meet the
+\oldconcept{CopyConstructible} and \oldconcept{CopyAssignable} requirements.
+The program is ill-formed if any of
+\begin{itemize}
+\item \tcode{is_trivially_copyable_v<T>},
+\item \tcode{is_copy_constructible_v<T>},
+\item \tcode{is_move_constructible_v<T>},
+\item \tcode{is_copy_assignable_v<T>}, or
+\item \tcode{is_move_assignable_v<T>}
+\end{itemize}
+is \tcode{false}.
+\begin{note} Type arguments that are
 not also statically initializable may be difficult to use. \end{note}
 
 \pnum
--- a/source/concepts.tex
+++ b/source/concepts.tex
@@ -273,13 +273,18 @@ classifications, and fundamental type properties.
 \indexlibrary{\idxcode{Same}}%
 \begin{itemdecl}
 template<class T, class U>
-  concept Same = is_same_v<T, U>;
+  concept @\placeholdernc{same-impl}@ = is_same_v<T, U>;    // \expos
+
+template<class T, class U>
+  concept Same = @\placeholdernc{same-impl}@<T, U> && @\placeholdernc{same-impl}@<U, T>;
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
+\begin{note}
 \tcode{\libconcept{Same}<T, U>} subsumes \tcode{\libconcept{Same}<U, T>} and
 vice versa.
+\end{note}
 \end{itemdescr}
 
 \rSec2[concept.derivedfrom]{Concept \libconcept{DerivedFrom}}
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -2387,6 +2387,8 @@ namespace std::ranges {
     friend constexpr sentinel_t<R> end(@\placeholder{ref-view}@ r)
     { return r.end(); }
   };
+  template<class R>
+    @\placeholder{ref-view}@(R&) -> @\placeholder{ref-view}@<R>;
 }
 \end{codeblock}
 
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -1254,6 +1254,9 @@ constexpr subrange(I i, S s) requires (!StoreSize);
 
 \begin{itemdescr}
 \pnum
+\expects \range{i}{s} is a valid range.
+
+\pnum
 \effects Initializes \tcode{begin_} with \tcode{i} and \tcode{end_} with
 \tcode{s}.
 \end{itemdescr}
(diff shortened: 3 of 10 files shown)


jensmaurer:
Maybe just \linebreak?

source/atomics.tex
```diff
@@ -1350,8 +1350,15 @@
 
 \indexlibrary{\idxcode{atomic}}%
 \pnum
-The template argument for
-\tcode{T} shall be trivially copyable\iref{basic.types}. \begin{note} Type arguments that are
+The template argument for \tcode{T} shall
+meet the \tcode{CopyConstructible} and \tcode{CopyAssignable} requirements.
```
CaseyCarter:
```suggestion
meet the \oldconcept{CopyConstructible} and \oldconcept{CopyAssignable} requirements.
```

source/ranges.tex
```diff
@@ -2387,6 +2391,8 @@
     friend constexpr sentinel_t<R> end(@\placeholder{ref-view}@ r)
     { return r.end(); }
   };
+  template<class R>
+    ref_view(R&) -> ref_view<R>;
```
CaseyCarter: These should technically be `\placeholder{ref-view}`s in this PR, but motions-2019-lwg-16 is renaming *`ref-view`* to `ref_view` so this is the proper final state for this line post-Kona. Maybe the rename here should be in a separate post-merge commit?

source/atomics.tex
burblebee: @CaseyCarter The resolution for LWG3012 doesn't use the Cpp17 prefix here but I think you're right, so I've made the change.

source/ranges.tex
burblebee: Since the motions will all be merged at the same time, I think it's OK to leave it as is.  Less headaches for Richard :).

BTW, are you reviewing all the applied issues in this motion's PR?  If so, I'd like to assign you as the reviewer (to prevent someone else from doing a full review and duplicating your work).

source/atomics.tex
CaseyCarter: Yes, I should have clarified. The P/R wording for LWG3012 predates incorporation of P0898 into the WD that added all of the Cpp17 prefixes. LWG should have made the change in San Diego when they approved the P/R, but folks aren't used to the prefixes yet and I was too busy drafting to show up for issue processing.

Thanks!

source/ranges.tex
CaseyCarter: I looked over all the changes to make sure there was nothing insane, but I haven't compared with the ingress paper.

burblebee:
> Maybe just \linebreak?

@jensmaurer \linebreak doesn't work, because it causes the line to expand to full the entire line between the margins (causing severe underfull).  The \hfill gets it to fill the rest of the line with spaces apparently (a trick I found from the \br macro which is only active in lib tables apparently).  Seems like this is a question for the Latex experts, @tkoeppe ?

source/ranges.tex
burblebee: Should this PR be made red given that LWG3173 (in it's current state) depends on LWG motion 16 which is also red?

godbyk:
@burblebee A couple quick options to consider (in order of preference):

1. Set `\emergencystretch`. This value is used by TeX only when typesetting paragraphs with overfull lines and allows TeX to add extra interword space to stretch a line out beyond what the normal line-breaking parameters permit. Note that an overfull line warning may still be logged but the line won't actually be overfull when typeset. This option could be applied globally to the document with minimal impact.
2. Use the `sloppypar` environment. This employs the `\emergencystretch` parameter above as well as a few other line-breaking parameters to relax the line-breaking algorithm a bit. Wrap troubleblesome paragraphs in this environment to avoid overfull line warnings altogether. This option should be used only where necessary. Don't use `\sloppy` globally as it will adversely impact the readability of the document generally.

Finally, if we're running into problems with line-breaking bits of C++ code, we might want to look into better options such as introducing more line break points (e.g., after colons). It might also be worth considering how line-breaking code impacts readability&mdash;namely, does the introduction of a hyphen ever cause ambiguity or readability problems?

jensmaurer:
@burblebee: Just \newline works nicely (it's \hfill + \linebreak)

burblebee:
@jensmaurer @godbyk  thanks for the tip guys!

tkoeppe:
@burblebee: The commit for LWG3077 only seems to modify string, but not deque and vector, as the issue says.

source/strings.tex
```diff
@@ -1653,12 +1653,14 @@
...
 
 \pnum
-\complexity Linear in the size of the sequence.
+\complexity If the size is not equal to the old capacity,
+linear in the size of the sequence;
+otherwise constant.
 
 \pnum
 \remarks Reallocation invalidates all the references, pointers, and iterators
 referring to the elements in the sequence as well as the past-the-end iterator.
```
tkoeppe: There's a missing edit here to add a comma.

source/concepts.tex
```diff
@@ -273,13 +273,19 @@
 \indexlibrary{\idxcode{Same}}%
 \begin{itemdecl}
 template<class T, class U>
-  concept Same = is_same_v<T, U>;
+  concept @\placeholdernc{same-impl}@ = // exposition only
+    is_same_v<T, U>;
```
tkoeppe: Could we please compact this onto a single line? I doubt the comment will be confusing; we have other places where a data member including initializer is exposition-only, and I think it's clear that the value of the initializer salient.

source/strings.tex
zygoloid: The remaining edits from the issue resolution (changes to deque, string) were also missing! Fixed.

source/ranges.tex
zygoloid: I've changed this to *`ref-view`* for now. I'll keep this in mind when applying LWG Motion 16.

source/strings.tex
tkoeppe: @zygoloid: I said that in my comment above!

PR #2750, opened 2019-03-07
[func.not_fn], [func.bind_front] fix phrasing of \mandates and \expects

jwakely:
The Mandates: element should just state its condition, not say that it
"shall be true". Cpp17 concept requirements should be phrased as "X meets the
Y requirements" not "X shall meet the requirements of Y".

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -15261,11 +15261,11 @@ In the text that follows:
 \pnum
 \mandates
 \tcode{is_constructible_v<FD, F> \&\& is_move_constructible_v<FD>}
-shall be true.
+is true.
 
 \pnum
 \expects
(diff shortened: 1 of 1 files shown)


W-E-Brown:
> On Mar 7, 2019, at 6:00 AM, Jonathan Wakely <notifications@github.com> wrote:
> 
> The Mandates: element should just state its condition, not say that it
> "shall be true".
> 

It is my understanding that we do need to say "is true"  (but not "shall be").

<snip>

jwakely:
> It is my understanding that we do need to say "is true" (but not "shall be"). <snip>

Yes, the commit message could have been clearer, but the actual commit does change it to "is true".

PR #2764, opened 2019-03-09
[time.clock.req] Simplify requirements for Cpp17TrivialClock.

### LINKED ISSUE #2734: Remove requirement duplication from Cpp17TrivialClock
With the acceptance of [LWG 3133](https://wg21.link/lwg3133) during the Kona 2019 meeting it would be now possible to simplify the current wording in [time.clock.req] p4 as follows:

A type TC meets the Cpp17TrivialClock requirements if:
(4.1) - [...]
(4.2) - the types TC::rep, TC::duration, and TC::time_point satisfy the Cpp17EqualityComparable (Table23), Cpp17LessThanComparable (Table 24)<del>, Cpp17DefaultConstructible (Table 25), Cpp17CopyConstructible (Table 27), Cpp17CopyAssignable (Table 29), and Cpp17Destructible (Table 30)</del> requirements and the requirements of numeric types (25.2). [Note: This means, in particular, that operations on these types will not throw exceptions. —end note]
(4.3) - [...]
(4.4) - [...]
(4.5) - [...]

This is possible, because the revised numeric types requirements ([numeric.requirements] p1) now explicitly call out the Cpp17DefaultConstructible, Cpp17CopyConstructible, Cpp17CopyAssignable, and Cpp17Destructible requirements.

Could we please perform this simplification editorially?

jensmaurer:
Fixes #2734.

### DIFF
--- a/source/time.tex
+++ b/source/time.tex
@@ -1085,12 +1085,9 @@ A type \tcode{TC} meets the \oldconcept{TrivialClock} requirements if:
 \item \tcode{TC} satisfies the \oldconcept{Clock} requirements\iref{time.clock.req},
 
 \item the types \tcode{TC::rep}, \tcode{TC::duration}, and \tcode{TC::time_point}
-satisfy the \oldconcept{EqualityComparable} (\tref{equalitycomparable}),
-\oldconcept{LessThanComparable} (\tref{lessthancomparable}),
-\oldconcept{DefaultConstructible} (\tref{defaultconstructible}),
-\oldconcept{\-Copy\-Con\-structible} (\tref{copyconstructible}),
-\oldconcept{CopyAssignable} (\tref{copyassignable}), and
-\oldconcept{Destructible} (\tref{destructible}) requirements and the requirements of
+satisfy the \oldconcept{EqualityComparable} (\tref{equalitycomparable}) and
+\oldconcept{LessThanComparable} (\tref{lessthancomparable})
+requirements and the requirements of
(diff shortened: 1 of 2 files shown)


zygoloid:
I would be much happier with this if [numeric.requirements]p1 actually defined those requirements as being the requirements on numeric types. Right now, it's very unclear what that term even means.

jensmaurer:
Here you are, be happy.

source/time.tex
```diff
@@ -1085,12 +1085,9 @@
 \item \tcode{TC} satisfies the \oldconcept{Clock} requirements\iref{time.clock.req},
 
 \item the types \tcode{TC::rep}, \tcode{TC::duration}, and \tcode{TC::time_point}
-satisfy the \oldconcept{EqualityComparable} (\tref{equalitycomparable}),
-\oldconcept{LessThanComparable} (\tref{lessthancomparable}),
-\oldconcept{DefaultConstructible} (\tref{defaultconstructible}),
-\oldconcept{\-Copy\-Con\-structible} (\tref{copyconstructible}),
-\oldconcept{CopyAssignable} (\tref{copyassignable}), and
-\oldconcept{Destructible} (\tref{destructible}) requirements and the requirements of
+satisfy the \oldconcept{EqualityComparable} (\tref{equalitycomparable}) and
```
CaseyCarter: Since we're here:
```suggestion
meet the \oldconcept{EqualityComparable} (\tref{equalitycomparable}) and
```

PR #2768, opened 2019-03-11
Lex.charset fixes

### LINKED ISSUE #2752: [lex.charset] p2 \00NNNNNN should use placeholder
The NNNNNN here and in the vicinity are placeholders, not literal characters, and thus should use \placeholder.

zygoloid:
Fixes #2752.

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -241,18 +241,17 @@ other characters.
     \terminal{\textbackslash U} hex-quad hex-quad
 \end{bnf}
 
-The character designated by the \grammarterm{universal-character-name} \tcode{\textbackslash
-U00NNNNNN} is that character
-that has \tcode{U+NNNNNN} as a code point short identifier;
-the character designated by the \grammarterm{universal-character-name}
-\tcode{\textbackslash uNNNN} is that character
-that has \tcode{U+NNNN} as a code point short identifier.
-If a \grammarterm{universal-character-name} does not correspond to
-a code point in ISO/IEC 10646 or
-if a \grammarterm{universal-character-name} corresponds to
-a surrogate code point,
-the program is ill-formed. Additionally, if
-a \grammarterm{universal-character-name} outside
+A \grammarterm{universal-character-name}
+designates the character in ISO/IEC 10646 (if any)
+whose code point is the hexadecimal number represented by
+the sequence of \grammarterm{hexadecimal-digit}s
+in the \grammarterm{universal-character-name}.
+The program is ill-formed if that number is not a code point
+or if it is a surrogate code point.
+Noncharacter code points and reserved code points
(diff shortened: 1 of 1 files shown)


zygoloid:
image

rmartinho:
Instead of "(hexadecimal)", what about `$[\mathrm{D800}, \mathrm{DFFF}_16]$` instead? (pardon if my LaTeX is too rusty, but `_` makes subscripts, right?)

jensmaurer:
We're not doing "16" as subscript anywhere else in the standard, as far as I know. Maybe this single location (and in a note) is not a good place to start.

source/lex.tex
```diff
@@ -207,18 +207,17 @@
...
-\tcode{\textbackslash uNNNN} is that character
-that has \tcode{U+NNNN} as a code point short identifier.
-If a \grammarterm{universal-character-name} does not correspond to
-a code point in ISO/IEC 10646 or
-if a \grammarterm{universal-character-name} corresponds to
-a surrogate code point,
-the program is ill-formed. Additionally, if
-a \grammarterm{universal-character-name} outside
+A \grammarterm{universal-character-name}
+designates the character in ISO/IEC 10646 (if any)
```
jensmaurer: Can we avoid first talking about "characters" and then retracting our statement later by saying "if any", and then expressly discussing the non-character case?
Maybe saying "... designates a code point given by the hexadecimal number ...", then giving the ill-formed dance, and then continuing with "A code point designates an ISO/IEC ... character, is a reserved code point, or a noncharacter code point. The latter are considered to designate separate characters ..." would help.

jensmaurer:
Oh, and please change the commit comment to use a "[lex.charset]" prefix.

source/lex.tex
zygoloid: We use the word "character" in lots of places when talking about sequences of elements from the source character set, which is what a UCN is notionally forming (eg, [lex.name]: "All characters are significant.", [basic]/9: "Two names are the same if they are identifiers composed of the same character sequence", etc.). I don't want to change those all to be talking about sequences of code points, which means we're left with our notion of "character" including things that Unicode doesn't consider to be a character. =(

tahonermann:
@zygoloid, I see this issue is marked as needing review from SG16.  I'm assuming you're just looking for any comments SG16 participants might have to offer.  If you would like something more formal, please let me know.

jensmaurer:
@tahonermann: Specifically, since you are chair of SG16, do you feel this change should be reviewed by SG16? If not, could you please review it and comment whether you're ok with the change?

tahonermann:
@jensmaurer @zygoloid, I see no need for this change to be further reviewed by SG16.  I have reviewed it and am happy with it.

jensmaurer:

PR #2772, opened 2019-03-12
Fix a bunch of faulty parallelism with "either".

Quuxplusone:
This is intended as a purely editorial changeset.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -100,7 +100,7 @@ a = ((a + b) + 32765);
 since if the values for \tcode{a} and \tcode{b} were, respectively,
 -32754 and -15, the sum \tcode{a + b} would produce an exception while
 the original expression would not; nor can the expression be rewritten
-either as
+as either
 \begin{codeblock}
 a = ((a + 32765) + b);
 \end{codeblock}
@@ -2882,7 +2882,7 @@ containing a possibly empty, comma-separated list of
 constitute the arguments to the function. The postfix expression shall
 have function type or function pointer type.
 For a call to a non-member function or to a static member function,
-the postfix expression shall be either an lvalue that refers to a
+either the postfix expression shall be an lvalue that refers to a
 function (in which case the function-to-pointer standard
 conversion\iref{conv.func} is suppressed on the postfix expression),
 or it shall have function pointer type.
@@ -7124,8 +7124,8 @@ satisfies the following constraints:
 An entity is a
 \defnx{permitted result of a constant expression}{constant expression!permitted result of}
 if it is an
-object with static storage duration that is either not a temporary object or is
-a temporary object whose value satisfies the above constraints, or
+object with static storage duration that either is not a temporary object or is
+a temporary object whose value satisfies the above constraints, or if
 it is a non-immediate function.
 \begin{example}
 \begin{codeblock}
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -2882,10 +2882,10 @@ containing a possibly empty, comma-separated list of
 constitute the arguments to the function. The postfix expression shall
 have function type or function pointer type.
 For a call to a non-member function or to a static member function,
-either the postfix expression shall be an lvalue that refers to a
+the postfix expression shall either be an lvalue that refers to a
 function (in which case the function-to-pointer standard
 conversion\iref{conv.func} is suppressed on the postfix expression),
-or it shall have function pointer type.
+or have function pointer type.
 
 \pnum
 For a call to a non-static member function,
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -3871,7 +3871,7 @@ namespace std {
 
 \pnum
 Any instance of \tcode{variant} at any given time either holds a value
-of one of its alternative types, or it holds no value.
+of one of its alternative types or holds no value.
 When an instance of \tcode{variant} holds a value of alternative type \tcode{T},
 it means that a value of type \tcode{T}, referred to as the \tcode{variant}
 object's \defnx{contained value}{contained value!\idxcode{variant}}, is allocated within the storage of the
@@ -5165,7 +5165,7 @@ namespace std {
 An object of class \tcode{any} stores an instance of any type that satisfies the constructor requirements or it has no value,
 and this is referred to as the \defn{state} of the class \tcode{any} object.
 The stored instance is called the \defnx{contained value}{contained value!\idxcode{any}}.
-Two states are equivalent if either they both have no value, or both have a value and the contained values are equivalent.
+Two states are equivalent if either they both have no value, or they both have a value and the contained values are equivalent.
 
 \pnum
 The non-member \tcode{any_cast} functions provide type-safe access to the contained value.
@@ -11851,7 +11851,7 @@ has synchronization semantics corresponding to the value of \tcode{failure}.
 \remarks
 Two \tcode{shared_ptr} objects are equivalent if
 they store the same pointer value and
-either share ownership, or both are empty.
+share ownership, or if both are empty.
 The weak form may fail spuriously. See \ref{atomics.types.operations}.
 
 \pnum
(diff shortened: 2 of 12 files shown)


source/expressions.tex
```diff
@@ -2883,7 +2883,7 @@
 constitute the arguments to the function. The postfix expression shall
 have function type or function pointer type.
 For a call to a non-member function or to a static member function,
-the postfix expression shall be either an lvalue that refers to a
+either the postfix expression shall be an lvalue that refers to a
 function (in which case the function-to-pointer standard
 conversion\iref{conv.func} is suppressed on the postfix expression),
 or it shall have function pointer type.
```
zygoloid: This change is wrong. The "it" in the latter case is the postfix-expression, not the call. Suggest changing "shall be either" to "shall either be".

source/utilities.tex
```diff
@@ -11849,7 +11849,7 @@
 \remarks
 Two \tcode{shared_ptr} objects are equivalent if
 they store the same pointer value and
-either share ownership, or both are empty.
+share ownership, or if both are empty.
```
zygoloid: Hmm. This reads like a normative change, but appears correct: "empty" is defined as "does not own a pointer", so the previous "store the same pointer value and both are empty" case was meaningless by definition.

CaseyCarter: This is a normative change. It's possible for a `shared_ptr` to store a pointer that it does not own, via the aliasing constructors specified in [util.smartptr.shared.const]/14 - /17. Consequently two equivalent `shared_ptr`s can:
* store the same pointer value and share ownership, or
* store the same pointer value and both be empty.

 The proposed change makes *all* empty `shared_ptr`'s equivalent, regardless of their stored pointer values.

source/expressions.tex
Quuxplusone: Old code:

> For a call to a non-member function or to a static member function, the postfix expression shall be either an lvalue that refers to a function (in which case the function-to-pointer standard conversion [conv.func] is suppressed on the postfix expression), or it shall have function pointer type.

My version:

> For a call to a non-member function or to a static member function, either the postfix expression shall be an lvalue that refers to a function (in which case the function-to-pointer standard conversion [conv.func] is suppressed on the postfix expression), or it shall have function pointer type.

Your suggested version (IIUC):

> For a call to a non-member function or to a static member function, the postfix expression shall either be an lvalue that refers to a function (in which case the function-to-pointer standard conversion [conv.func] is suppressed on the postfix expression), or have function pointer type.


I think my version and your version have the same normative meaning, so I'm okay switching to your version, but I'd like to understand better why you interpret them as meaning different things.

Perhaps related: I encountered several places — this being one of them — where the normative text says "X either shall be Y or shall be Z." I read this as _literally_ saying that "it is required either that X be Y or that X be Z — but we're not going to tell you which of these two possible requirements is the actual requirement!"  Really, every single one of these instances ought to be recast as "X shall be either Y or Z." But sometimes that leads to harder-to-parse sentences, so generally I didn't do it in this patch.

source/utilities.tex
CaseyCarter: This is a normative change. [util.smartptr.shared.const]/17:
> [Note: This constructor allows creation of an empty `shared_ptr` instance with a non-null stored pointer.—end note]

source/utilities.tex
```diff
@@ -12088,7 +12088,7 @@
 \remarks
 Two \tcode{weak_ptr} objects are equivalent if
 they store the same pointer value and
-either share ownership, or both are empty.
+share ownership, or if both are empty.
```
CaseyCarter: This change, however, I believe is valid: there's no way to construct a `weak_ptr` that both stores a pointer value and is empty.

source/utilities.tex
Quuxplusone: @CaseyCarter: I think you're right. I was on autopilot here and thinking about the conditions for two shared_ptrs to be *equal* (`operator==`-wise); but this section is talking about the proposed semantics for *equivalence* (`compare_exchange_strong`-wise).

No vendor has yet implemented `atomic<shared_ptr<T>>` as far as I can tell, but if anyone ever did, I think they'd naturally make `shared_ptr<T>(p1, nullptr)` "equal but non-equivalent to" `shared_ptr<T>(p2, nullptr)`. That is, `compare_exchange_strong` will look at the bitwise values of the control-block pointer and the value pointer and treat the shared_ptrs as equivalent only if *both* pointers are equal.

https://wandbox.org/permlink/xAepLm6DdFXG5oa9

I think I'm unqualified to propose a "correct" wording here, especially if @zygoloid is correct that the original wording is meaningless. I'll take suggestions.

jwakely: A `shared_ptr` can **store** a pointer without **owning** it, and can share one pointer and own a different pointer. The original wording is not meaningless, it's correct.

source/utilities.tex
jwakely: Doesn't this do so?

    static int i = 0;
    std::shared_ptr<int> empty;
    std::shared_ptr<int> empty_not_null(empty, &i);
    std::weak_ptr<int> weak(empty_not_null);

And you can get another empty-not-null `shared_ptr` by constructing it from `weak` (although not by calling `weak.lock()` because that returns `shared_ptr<int>()` if it's empty.

CaseyCarter: > Doesn't this do so?

I initially thought so as well, but [util.smartptr.weak.const]/4 says:

> *Effects:* If `r` is empty, constructs an empty `weak_ptr` object; otherwise, constructs a `weak_ptr` object
that shares ownership with `r` and stores a copy of the pointer stored in `r`.

and now I'm wondering if I implement that correctly.

CaseyCarter: > and now I'm wondering if I implement that correctly.

Nope: we get this wrong.

jwakely: I checked the ctor efects, but apparently I can't read.

> Nope: we get this wrong.

￼I think we do too.

jwakely: Even if you can't create such a weak ptr, is using subtly different wording here an improvement?

source/utilities.tex
CaseyCarter: > A `shared_ptr` can **store** a pointer without **owning** it, and can **store** one pointer and **own** a different pointer. The original wording is not meaningless, it's correct.

For completeness: a `shared_ptr` can also own an object of type `nullptr_t` via the `(nullptr_t, deleter)` and `(nullptr_t, deleter, allocator) constructors; `deleter(nullptr)` is invoked when there are no more owners. (There is more than a little weirdness in `shared_ptr`.)

source/utilities.tex
CaseyCarter: > Even if you can't create such a `weak ptr`, is using subtly different wording here an improvement?

Probably not, and the existing wording will be correct if/when we apply the resolution of the LWG issue I just submitted. It may be better to leave this wording alone.

jensmaurer:
This was updated.

source/expressions.tex
zygoloid: In your version, the "it" has a somewhat-ambiguous antecedent. A reader could parse it as either:

> For a call to a non-member function or to a static member function, either the postfix expression shall be an lvalue that refers to a function (in which case the function-to-pointer standard conversion [conv.func] is suppressed on the postfix expression), or [the postfix expression] shall have function pointer type.

or as

> For a call to a non-member function or to a static member function, either the postfix expression shall be an lvalue that refers to a function (in which case the function-to-pointer standard conversion [conv.func] is suppressed on the postfix expression), or [the call to a non-member function or to a static member function] shall have function pointer type.

depending on whether they think the scope of the "postfix expression" ends at the end of the clause of the "either" in which it was introduced. (Imagine that you skip from the "either" to the "or" when reading because you know you don't care about that clause.) My alternative avoids that problem.

source/utilities.tex
zygoloid: OK, so it sounds like we're settling on the original wording being correct (but maybe slightly unnaturally phrased). Thank you all for catching this! How about:

> Two `shared_ptr` objects are equivalent if they store the same pointer value, and either they share ownership or both are empty.

(With or without the comma before "and", but definitely with no comma before "or".)

CaseyCarter: I think "Two `shared_ptr` objects are equivalent if they store the same pointer value and either share ownership or are both empty." is a bit simpler, but I certainly wouldn't object to your proposed phrasing.

zygoloid: See #2793 for that change.

PR #2778, opened 2019-03-13
[std] Index all mentions of keywords in grammar productions.

### LINKED ISSUE #2769: add index entries for identifiers with special meaning
Add index entries for identifiers with special meaning, both for their entry in Table 4 and for each use of them in the grammar.

We should probably add a dedicated macro for these special non-keyword terminals to automate the second part of this.

jensmaurer:
Fixes #2769.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -1140,7 +1140,7 @@ this behavior is deprecated\iref{depr.arith.conv.enum}.
 \begin{bnf}
 \nontermdef{primary-expression}\br
     literal\br
-    \terminal{this}\br
+    \keyword{this}\br
     \terminal{(} expression \terminal{)}\br
     id-expression\br
     lambda-expression\br
@@ -1414,7 +1414,7 @@ void f() {
 %
 \begin{bnf}
 \nontermdef{qualified-id}\br
-    nested-name-specifier \opt{\terminal{template}} unqualified-id
+    nested-name-specifier \opt{\keyword{template}} unqualified-id
 \end{bnf}
 
 \indextext{operator!scope resolution}%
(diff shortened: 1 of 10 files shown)


source/expressions.tex
```diff
@@ -2819,12 +2819,12 @@
     postfix-expression \opt{\terminal{-> template}} id-expression\br
     postfix-expression \terminal{++}\br
     postfix-expression \terminal{-{-}}\br
-    \terminal{dynamic_cast <} type-id \terminal{> (} expression \terminal{)}\br
-    \terminal{static_cast <} type-id \terminal{> (} expression \terminal{)}\br
-    \terminal{reinterpret_cast <} type-id \terminal{> (} expression \terminal{)}\br
-    \terminal{const_cast <} type-id \terminal{> (} expression \terminal{)}\br
-    \terminal{typeid (} expression \terminal{)}\br
-    \terminal{typeid (} type-id \terminal{)}
+    \keyword{dynamic_cast} \terminal{<} type-id \terminal{> (} expression \terminal{)}\br
```
zygoloid: Hmm. Somewhat pre-existing, but... I believe this will cause us to use a roman italic (narrow) space between the `dynamic_cast` and the `<`, but a teletype space between the `>` and `(`. We should think about what we actually want here; I'm inclined to prefer the roman italic space.

I'll file this as a separate issue.

source/lex.tex
```diff
@@ -668,8 +674,8 @@
 \tcode{audit}           \\
 \tcode{axiom}           \\
 \columnbreak
-\tcode{import}          \\
 \tcode{final}           \\
+\tcode{import}          \\
 \columnbreak
 \tcode{module}          \\
 \tcode{override}        \\
```
zygoloid: Can we use \keyword in this table instead of adding explicit index entries?

zygoloid: (Likewise in the keyword table below; those should get index entries for the keyword too, I think.)

source/expressions.tex
zygoloid: Filed as #2788

source/lex.tex
jensmaurer: I've also added the alternative operator spellings such as "and" and "bitand". They are sufficiently keyword-like to warrant an index entry, I believe.

jensmaurer: Ah, the \keyword macro is only defined within "bnfbase" environments. We need it outside, too.

PR #2780, opened 2019-03-14
Simplifying [class.rel] wording.

BRevzin:
`x` and `y` have the same type (this is in the context of a defaulted relational operator, so they have to be per [class.compare]/1). So either `x <=> y` finds something or it doesn't - there's no situation where we'd select the reversed candidate `y <=> x`, so we can simplify this wording. Same thing Richard pointed out http://lists.isocpp.org/core/2019/03/5706.php about defaulted `!=`.

Also, we never changed that `x <=> y @ 0` is actually convertible to `bool`, so added that too.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -6863,7 +6863,8 @@ or a function that is deleted or inaccessible from the operator function,
 or
 
 \item
-\tcode{x <=> y @ 0} is ill-formed.
+the operator \tcode{@}
+cannot be applied to the return type of \tcode{x <=> y}.
 \end{itemize}
 
 Otherwise, the operator function yields
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -6863,7 +6863,7 @@ or a function that is deleted or inaccessible from the operator function,
 or
 
 \item
-\tcode{x <=> y @ 0} cannot be converted to \tcode{bool}.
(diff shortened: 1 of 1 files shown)


source/classes.tex
```diff
@@ -6858,22 +6858,16 @@
...
 as applied to \tcode{x <=> y}
-(also considering synthesized candidates with reversed order of parameters\iref{over.match.oper}),
 results in an ambiguity
 or a function that is deleted or inaccessible from the operator function,
 or
 
 \item
-the operator \tcode{@}
-cannot be applied to the return type of \tcode{x <=> y} or \tcode{y <=> x}.
+\tcode{x <=> y @ 0} is ill-formed.
```
zygoloid: This fully covers the check in the first bullet, but it also covers some other things (overload resolution can succeed and pick a function that is nonetheless not callable). As such, changing to this wording is also non-editorial. =(

I think we should just drop the "or `y <=> x`" from this bullet for now.

BRevzin: Core issue it is then 😃

jensmaurer:
Changes have been applied; commits need squashing.

BRevzin:
As in... you want me to force push? Why not squash merge?

zygoloid:
I imagine Jens' comment was a reminder to me to squash rather than just rebasing.

PR #2784, opened 2019-03-14
[meta.type.synop] Align type_identity_t's "=" with following declarat…

CaseyCarter:
…ions

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -5758,8 +5758,8 @@ namespace std {
     unsigned long to_ulong() const;
     unsigned long long to_ullong() const;
     template<class charT = char,
-              class traits = char_traits<charT>,
-              class Allocator = allocator<charT>>
+             class traits = char_traits<charT>,
+             class Allocator = allocator<charT>>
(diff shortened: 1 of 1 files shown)


tkoeppe:
gnah, this sort of cutespacing is so dubious. It's hard to maintain, and when you're aligning with something that's separated by an intervening line, is it really an improvement, or does it just look like an accident? Eventually you'll form arbitrary bubbles of local alignment separated by harsh domain walls, and it becomes distracting.

Several container class definitions do the same questionable thing.

I'd semi-mildly prefer to get rid of all this and just use regular spacing everywhere.

CaseyCarter:
I'm also not a fan, but it's obviously the intent that this `=` aligns with the others, so I submitted a fix.  Might I suggest we apply this for now, and I'll open an editorial issue to discuss under what circumstances we allow/require alignment in declarations so we can enforce some consistency?

tkoeppe:
OK, but I expect we'll have hundreds more small whitespace fixes in utilities... could you at least have a quick look around the neighbourhood if anything else needs fixing while we're here?

PR #2787, opened 2019-03-15
[lib] Remove space between 'template' and argument list.

### LINKED ISSUE #2785: prevent `template <` from sneaking into the library wording
We should fix all the instances of `template <` in library clauses, and consider adding a CI check to prevent more regressions.

jensmaurer:
Fixes #2785.


zygoloid:
Rebased/recreated and submitted as 8d238bd4.

I'm not marking #2785 as fixed because I'd still like to consider enforcing this via CI.

PR #2789, opened 2019-03-15
[intro.execution] make function evaluation actually not interleave CWG2466

RedBeard0531:
While the intent was clear and explicitly stated in the footnote, the prior wording didn't require B to be consistently sequenced before or after every A in F, only that they be individually pairwise sequenced.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -5323,11 +5323,11 @@ expression, or with the postfix expression designating the called
 function, is sequenced before execution of every expression or statement
 in the body of the called function.
 For each function invocation \placeholder{F},
-for every evaluation \placeholder{A} that occurs within \placeholder{F} and
+for the set \placeholder{A} of all evaluations that occur within \placeholder{F} and
 every evaluation \placeholder{B} that does not occur within \placeholder{F} but
 is evaluated on the same thread and as part of the same signal handler (if any),
-either \placeholder{A} is sequenced before \placeholder{B} or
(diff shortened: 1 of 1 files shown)


jensmaurer:
Since sequenced-before is a transitive order, isn't your reformulation just creating the transitive hull, which is (specification-wise) a no-op?

zygoloid:
Example:
```
int f() { A1(); A2(); return 0; };
int k = f() + B();
```
Prior to this change, what prevents `A1 < B < A2` (where `<` is "sequenced before")?

Regardless, these words are subtle, and we shouldn't change them without CWG (and possibly SG1) consultation.

languagelawyer:
> what prevents `A1 < B < A2` (where `<` is "sequenced before")?

I have more crazy question: what prevents `A2();` < `A1();` or even `return 0;` < `A1();`?

RedBeard0531:
> what prevents A2(); < A1(); or even return 0; < A1();?

http://eel.is/c++draft/intro.execution#9

> Every value computation and side effect associated with a full-expression is sequenced before every value computation and side effect associated with the next full-expression to be evaluated.

languagelawyer:
Well, this doesn't say which full-expression is "next". But prolly I'm nagging to much. Continuing this way, we may start asking "where is it specified that the program is parsed from left to right and from top to bottom".

jensmaurer:
Addressed by CWG2466.

jensmaurer:
Fixed by CWG2466 in commit da92321bec1a392c9ea67cc5e458b4d682bd0720.

PR #2792, opened 2019-03-15
Switch to using constant-width italics for exposition-only names in the core wording

### LINKED ISSUE #2783: use italic teletype font for exposition-only identifiers in core wording
We have exposition-only identifiers in various parts of core wording:

[stmt.while]'s label `label`
[stmt.ranged]'s `__range`, `__begin`, and `__end`
[stmt.contin]'s label `contin`
[stmt.return.coroutine] / [dcl.fct.def.coroutine]'s label `__final_suspend`
[dcl.fct.def.coroutine]'s `__promise`

These are currently all plain teletype, with some being explicitly called out as for exposition only and others not. We should use the library exposition-only convention for them: italic teletype names in *`lower-kebab-case`*.

zygoloid:
We already used this typeface for exposition-only identifiers in the
library wording.

Fixes #2783.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -6274,17 +6274,17 @@ the $i^\textrm{th}$ function parameter otherwise.
 A coroutine behaves as if its \grammarterm{function-body} were replaced by:
 \begin{ncsimplebnf}
 \terminal{\{}\br
-\bnfindent promise-type \terminal{__promise} promise-constructor-arguments \terminal{;}\br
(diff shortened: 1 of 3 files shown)


tkoeppe:
Can we make something like `\tcode{\exposid{promise}.unhandled_exception()` work, i.e. permit `exposid` to appear inside a `tcode`?

jensmaurer:
@tkoeppe, I think that already works (and is used in the patch).

PR #2793, opened 2019-03-15
Clarify grouping of "either".

zygoloid:


source/utilities.tex
```diff
@@ -11850,8 +11850,8 @@
 \pnum
 \remarks
 Two \tcode{shared_ptr} objects are equivalent if
-they store the same pointer value and
-either share ownership, or both are empty.
+they store the same pointer value, and
+either they share ownership or both are empty.
```
Quuxplusone: Richard's proposed wording:

> Two `shared_ptr` objects are equivalent if they store the same pointer value, and either they share ownership or both are empty.

Casey's suggested wording:

> Two `shared_ptr` objects are equivalent if they store the same pointer value and either share ownership or are both empty.

I prefer Casey's, FWIW.

zygoloid: I prefer Casey's too. Updated.

jwakely:
A bit late, but LGTM too.

PR #2796, opened 2019-03-17
Library-wide: Use "meet" for non-concept type requirements

CaseyCarter:
...instead of "satisfy".

Partially addresses #1263.

### DIFF
--- a/source/time.tex
+++ b/source/time.tex
@@ -1024,7 +1024,7 @@ namespace std {
 A clock is a bundle consisting of a \tcode{duration}, a
 \tcode{time_point}, and a function \tcode{now()} to get the current \tcode{time_point}.
 The origin of the clock's \tcode{time_point} is referred to as the clock's \defn{epoch}.
- A clock shall satisfy the requirements in \tref{time.clock}.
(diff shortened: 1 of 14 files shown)


zygoloid:
@CaseyCarter Any chance you could rebase this in the next day or so, for the pre-Cologne working draft?

CaseyCarter:
> @CaseyCarter Any chance you could rebase this in the next day or so, for the pre-Cologne working draft?

I'm taking a look now (I hadn't noticed the tag). It looks like mostly conflicts with the table label rework; this shouldn't be much trouble to fix up.

CaseyCarter:
Rebased.

source/time.tex
```diff
@@ -1335,7 +1335,7 @@
 The defaulted copy constructor of duration shall be a
 constexpr function if and only if the required initialization
 of the member \tcode{rep_} for copy and move, respectively, would
-satisfy the requirements for a constexpr function.
+meet the requirements for a constexpr function.
```
zygoloid: This one should be "satisfy". See [dcl.constexpr]/3,4,6 and [class.copy.ctor]/12.

PR #2797, opened 2019-03-17
Library-wide: Use "model" instead of "satisfy" with library concept r…

### LINKED ISSUE #2591: [concepts] Use "model" instead of "satisfy" for the semantic requirements
See https://github.com/cplusplus/draft/pull/2176#discussion_r196590990.

CaseyCarter:
…equirements

Updates: [structure.requirements],[customization.point.object],[res.on.requirements],[concepts], and [range.semi.wrap].

Fixes #2591.

### DIFF
--- a/source/concepts.tex
+++ b/source/concepts.tex
@@ -134,8 +134,9 @@ struct T {
 };
 \end{codeblock}
 
-\tcode{T} fails to meet the implicit requirements of \tcode{C}, so \tcode{C<T>}
-is not satisfied. Since implementations are not required to validate the syntax
+\tcode{T} fails to meet the implicit requirements of \tcode{C},
+so \tcode{T} satisfies but does not model \tcode{C}.
+Since implementations are not required to validate the syntax
 of implicit requirements, it is unspecified whether an implementation diagnoses
 as ill-formed a program that requires \tcode{C<T>}.
 \end{example}
@@ -300,7 +301,8 @@ template<class Derived, class Base>
 \begin{itemdescr}
 \pnum
 \begin{note}
-\tcode{\libconcept{DerivedFrom}<Derived, Base>} is satisfied if and only if
+Types \tcode{Derived} and \tcode{Base} model
+\tcode{\libconcept{DerivedFrom}<Derived, Base>} if and only if
 \tcode{Derived} is publicly and unambiguously derived from \tcode{Base}, or
 \tcode{Derived} and \tcode{Base} are the same class type ignoring cv-qualifiers.
 \end{note}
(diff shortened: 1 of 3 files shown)


source/concepts.tex
```diff
@@ -1110,7 +1120,7 @@
 
 \pnum
 \begin{example}
-A random number generator does not satisfy
+A random number generator does not model
 \libconcept{RegularInvocable}.
```
CaseyCarter: Unwrap this line?

source/lib-intro.tex
```diff
@@ -527,13 +527,13 @@
 \pnum
 Required operations of any concept defined in this document need not be
 total functions; that is, some arguments to a required operation may
-result in the required semantics failing to be satisfied.
+result in the required semantics failing to be met.
```
CaseyCarter: This change really should be part of #2796.

source/concepts.tex
```diff
@@ -300,7 +301,8 @@
 \begin{itemdescr}
 \pnum
 \begin{note}
-\tcode{\libconcept{DerivedFrom}<Derived, Base>} is satisfied if and only if
+Types \tcode{Derived} and \tcode{Base} model
+\tcode{\libconcept{DerivedFrom}<Derived, Base>} if and only if
```
jwakely: What are the semantic requirements here that mean we have to use modelling not satisfaction?

source/concepts.tex
```diff
@@ -478,13 +482,13 @@
 \begin{itemdescr}
 \pnum
 \begin{note}
-\tcode{\libconcept{SignedIntegral}<T>} can be satisfied even by types that are
+\tcode{\libconcept{SignedIntegral}<T>} can be modeled even by types that are
```
jwakely: Should the `<T>` be here if we're talking about modelling? Other unary concepts seem to be used without, e.g. `X` models `Y` (rather than `Y<X>`), and we only have a template argument list for binary concepts, e.g. `X` and `Y` model `Z<X,Y>`. Or have I misunderstood the convention?

source/concepts.tex
```diff
@@ -478,13 +482,13 @@
...
 not signed integral types\iref{basic.fundamental}; for example, \tcode{char}.
 \end{note}
 
 \pnum
 \begin{note}
-\tcode{\libconcept{UnsignedIntegral}<T>} can be satisfied even by types that are
+\tcode{\libconcept{UnsignedIntegral}<T>} can be modeled even by types that are
```
jwakely: Same question about `<T>` here.

source/concepts.tex
CaseyCarter: I agree the minimal change is too minimal here (and on 491). I'll remove the `<T>`.

source/concepts.tex
CaseyCarter: This change is for consistency of presentation; false consistency, as you've pointed out. I'll revert.

PR #2800, opened 2019-03-19
Add coroutine to headers

leni536:


tkoeppe:
Wait, why do we have this file?

And why is "5" a header?

leni536:
Umm, disregard this PR, I thought this file had anything to do with the header index generation.

tkoeppe:
I mean, I would still like to know why we have that file...

CaseyCarter:
> I mean, I would still like to know why we have that file...

It hasn't changed since Stefanus added it in 2012 (https://github.com/cplusplus/draft/commit/4e08c317690a6013a3daa0dbe95acb5273cd317b#diff-9cc7be2307e175136de8adbd1cafeba5). I suspect it was added by mistake, and we could remove it.

PR #2803, opened 2019-03-20
[cmp.categories] Use kebab-case for \expos enumerations

### LINKED ISSUE #2801: [cmp.categories] eq, ord, and ncmp should use kebab-case
The three exposition-only scoped enumeration types `eq`, `ord`, and `ncmp` are used in the subject subclause with monospace names, which is inconsistent with our policy of using italic monospace for exposition-only library types.

CaseyCarter:
Affects `eq`, `ord`, and `ncmp` - as well as their enumerators - defined in [cmp.categories.pre]/1 and used throughout [cmp.categories].

Fixes #2801.

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -3951,10 +3951,10 @@ whose value typically corresponds to that of an enumerator
 from one of the following exposition-only enumerations:
 
 \begin{codeblock}
-enum class @\placeholder{eq}@ { @\placeholder{equal}@ = 0, @\placeholder{equivalent}@ = @\placeholder{equal}@,
-                @\placeholder{nonequal}@ = 1, @\placeholder{nonequivalent}@ = @\placeholder{nonequal}@ };   // \expos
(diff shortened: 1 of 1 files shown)


source/support.tex
```diff
@@ -3951,10 +3951,10 @@
 from one of the following exposition-only enumerations:
 
 \begin{codeblock}
-enum class eq { equal = 0, equivalent = equal,
-                nonequal = 1, nonequivalent = nonequal };   // \expos
-enum class ord { less = -1, greater = 1 };                  // \expos
-enum class ncmp { unordered = -127 };                       // \expos
+enum class @\placeholder{eq}@ { @\placeholder{equal}@ = 0, @\placeholder{equivalent}@ = @\placeholder{equal}@,
```
tkoeppe: Should we use the non-correcting forms of placeholder here to maintain vertical alignment? Can you maybe attach a diff screenshot?

CaseyCarter: Good idea - the comments are currently misaligned. I'll `nc` these.

CaseyCarter: Fixed with a "fixup" commit.

PR #2804, opened 2019-03-20
[limits.syn,numeric.limits] declare partial specializations of numeri…

CaseyCarter:
…c_limits in the header synopsis

...not in the class template synopsis.

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -729,6 +729,9 @@ namespace std {
 
   // \ref{numeric.limits}, class template \tcode{numeric_limits}
   template<class T> class numeric_limits;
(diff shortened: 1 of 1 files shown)


source/support.tex
```diff
@@ -729,6 +729,9 @@
 
   // \ref{numeric.limits}, class template \tcode{numeric_limits}
   template<class T> class numeric_limits;
+  template<class T> class numeric_limits<const T>;
```
tkoeppe: newline between the primary and these partials?

CaseyCarter: Sure.

PR #2808, opened 2019-03-28
[dcl.type.elab] Mix struct/class in example

jwakely:
Extend the example to show that the choice of `class` or `struct` in an
_elaborated-type-name_ referring to a class is not significant.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -1495,6 +1495,8 @@ declared using the \tcode{class} or \tcode{struct}
 \begin{codeblock}
 enum class E { a, b };
 enum E x = E::a;                // OK
+struct S { } s;
+class S* p = &s;                // OK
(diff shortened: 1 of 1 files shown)


jwakely:
I've seen it suggested t hat the sentence before the example isn't entirely unambiguous:

> either the `class` or `struct` _class-key_ shall be used to refer to a class ([class]) declared using the `class` or `struct` _class-key_.

I think it's entirely unambiguous, but the example would reinforce the point, and seems to be in the spirit of core examples not following consistent styles of brace placement, ptr-declarator placement etc.

jwakely:
We could also consider switching the order in that wording:

> either the `class` or `struct` _class-key_ shall be used to refer to a class ([class]) declared using the `struct` or `class` _class-key_.

:smiling_imp:

PR #2813, opened 2019-03-30
[iterator.operations] Clarify the observable side-effects of `distance`.

### LINKED ISSUE #2807: [iterator.operations] It should be stated clearly that `std::distance(first, last)` might invalidate all copies of `first` if it is not a ForwardIterator.

umanwizard:
The standard claims that for non-Random-Access iterators, `distance` "uses ++". I guess the reader is supposed to infer from this that you can't continue using an iterator that doesn't have a multi-pass guarantee after passing it to `distance`, but it could be more clear. Especially since `distance(a, a)` *won't* actually invalidate anything (at any rate, not in libc++ and I'd be quite surprised if it did in any standard library implementation)

I believe that this lack of clarity in the standard is an editorial oversight, and that the committee's intended meaning is that `distance(first, last)` has the same observable side-effects as calling pre-increment `++` on first until it equals last.

Fixes #2807.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -2614,8 +2614,16 @@ the \oldconcept{RandomAccessIterator} requirements and
 \pnum
 \effects
 If \tcode{InputIterator} meets the \oldconcept{RandomAccessIterator} requirements,
-returns \tcode{(last - first)}; otherwise, returns
-the number of increments needed to get from
+returns \tcode{(last - first)}; otherwise, equivalent to:
+\begin{codeblock}
+typename iterator_traits<InputIterator>::difference_type diff = 0;
+InputIterator i = first;
+while (i != last) {
+  ++i;
+  ++diff;
+}
+return diff;
+\end{codeblock}
 \tcode{first}
 to
 \tcode{last}.
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -2624,9 +2624,6 @@ while (i != last) {
 }
 return diff;
 \end{codeblock}
-\tcode{first}
-to
-\tcode{last}.
 \end{itemdescr}
 
 \indexlibrary{\idxcode{next}}%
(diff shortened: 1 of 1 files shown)


umanwizard:
@jwakely Sure, no good reason to introduce a separate `i`. I can change it.
 
I don't see how the implication is clear? It seems to say that `distance` will just return the number of necessary increments to get from point a to point b, not that it will necessarily perform them.

Let's say I make an iterator that has some side-effect like printing hello world every time it's incremented. And let's say it does not expose the interface of a random-access iterator, but the compiler and/or stdlib are able to figure out that it does in fact point into a random-access structure, and optimize away the increments to return the correct value from distance but avoid printing hello world.

Is that conforming? I think according to the standard as currently written, it is: the standard just says `distance` returns the number of increments required, not that it actually performs them. But I think according to the clear design intent, it wasn't meant to be conforming.

jwakely:
But the standard doesn't say that magic, impossible optimisation happens, so you can't assume it does, and have to assume the increments happen, and so other copies of input iterators can be affected.

Saying "no impossible magic happens" doesn't seem necessary to me. A note stating that input iterators might become non-dereferenceable might be reasonable, but specifying the function using code isn't an improvement IMHO.

umanwizard:
> But the standard doesn't say that magic, impossible optimisation happens, so you can't assume it does

Right, but it also doesn't state that the increments *do* happen, as far as I can tell.

jensmaurer:
I'm not convinced the rewrite as code is a net improvement. I do sympathize with the sentiment that "needed to get" is sloppy wording of sorts; is there maybe a small prose adjustment that helps here? "increments first until last is reached and returns the number of increments" or so?

jensmaurer:
Editorial meeting: Go with Jens' suggestion.

CaseyCarter:
Please be sure to keep the wording style consistent in `std::advance`, `std::distance`, `ranges::advance`, and `ranges::distance`.

PR #2815, opened 2019-04-04
[range.filter.iterator] Add bounds checking to operator--

johelegp:
Missing bounds check. UB ensues if no element is filtered, for example.


johelegp:
This was a misunderstanding by my part. It's all good.

PR #2818, opened 2019-04-05
[istream.sentry] Remove unreferenced typedef-name traits_type

CaseyCarter:
I have no idea how this got here - it was in the initial git commit - or why it isn't `\\expos`, but in any case it's unused.


CaseyCarter:
> We actually make that member public in our `sentry`

We have no such member; I'm "standardizing existing practice".

PR #2819, opened 2019-04-08
change chi_squared_distribution's integer literal to floating literal

EzoeRyou:
chi_squared_distribution's default constructor has delegating constructor with argument of integer literal `1`. But its type supposed to be the floating point types. Other distributions use floating literals in similar cases. So the suggested change.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -5595,8 +5595,8 @@ template<class RealType = double>
     using param_type  = @\unspec@;
 
     // constructor and reset functions
-    fisher_f_distribution() : fisher_f_distribution(1) {}
-    explicit fisher_f_distribution(RealType m, RealType n = 1);
+    fisher_f_distribution() : fisher_f_distribution(1.0) {}
+    explicit fisher_f_distribution(RealType m, RealType n = 1.0);
     explicit fisher_f_distribution(const param_type& parm);
     void reset();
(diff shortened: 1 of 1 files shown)


EzoeRyou:
I found the same issues on `fisher_f_distribution` and `student_t_distribution`.

jensmaurer:
I believe this change is unnecessary. RealType can only be float, double, or long double (cf. [rang.req.genl] p1.4), and RealType is not a template parameter for the constructor (it's a template parameter for the surrounding class template), so there is no template type deduction in the delegating constructor.

(Personally, I dislike source code that has "double x = 0.0" and similar all over the place; "double x = 0" is also clear and saves reading two characters. But that's just me.)

Are there other examples in the standard text where we do the equivalent of "double x = 1.0" as opposed to "double x = 1", so that we could argue with harmonization?

zygoloid:
We should decide how we want to format floating-point literals throughout the library wording and document that decision.

jensmaurer:
Editorial meeting: Every floating-point literal in the library section should have a decimal point and at least one digit on each side of the decimal point (also 1.0e3). Only potentially apply to examples. Use digit separator for constants for random number generators in 26.6.5 (e.g. mersenne_twister). Also apply spaces after commas in that section.

PR #2820, opened 2019-04-09
"Template template arguments" contributing to ADL?

Quuxplusone:
Discussed on Slack with @k-ballo and we decided we have no idea what the parenthesized wording here is supposed to mean; therefore I wildly propose that we just strike it. @zygoloid, please help prove me wrong!

Furthermore, I claim with stronger confidence that the phrase _"the templates used as template template arguments"_ is redundant, given the following phrases _"the namespaces of which any template template arguments are members; and the classes of which any member templates used as template template arguments are members."_ I don't think it is meaningful for a _template itself_ to be an associated entity.

K-ballo did find this completely crazy example which is supported by GCC (but not Clang, ICC, or MSVC):

```
// https://godbolt.org/z/zw8WAs
namespace A {
  template <class>
  void otherfun() {}
}

namespace B {
  class T;
  void call(void(*)()) {}
}

int main() {
    call(A::otherfun<B::T>);
}
```

Here, the type of `A::otherfun<B::T>` is unknown (because it's an overload set); but GCC believes that `B` is an associated namespace (and `A` is *not* an associated namespace), presumably somewhat on the basis of the extremely confusing wording touched by this PR.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -1692,8 +1692,7 @@ and a set of zero or more
 to be considered.
 The sets of namespaces and entities
 are determined entirely by
-the types of the function arguments
-(and the namespace of any template template argument).
+the types of the function arguments.
 Typedef names and \grammarterm{using-declaration}{s}
 used to specify the types
 do not contribute to this set.
@@ -1712,14 +1711,9 @@ Its associated namespaces are
 the innermost enclosing namespaces of its associated entities.
 Furthermore, if \tcode{T} is a class template specialization,
 its associated namespaces and entities also include:
-the namespaces and entities
-associated with the types of the template arguments
-provided for template type parameters
-(excluding template template parameters);
-the templates used as template template arguments;
+the namespaces and entities associated with the types used as template type arguments;
 the namespaces of which any template template arguments are members; and the
-classes of which any member templates used as template template
-arguments are members.
+classes of which any member templates used as template template arguments are members.
 \begin{note}
 Non-type template arguments do not
 contribute to the set of associated namespaces.


Quuxplusone:
Update: The "completely crazy example" is actually 100% legit and GCC is right to accept it, according to http://eel.is/c++draft/basic.lookup.argdep#2.sentence-8 —

> In addition, if the argument is the name or address of a set of overloaded functions and/or function templates, [...] if the aforementioned set of overloaded functions is named with a _template-id_, its associated entities and namespaces also include those of its type _template-arguments_ and its template _template-arguments_.

So that example is probably *not* relevant to this PR at all.

CaseyCarter:
>  I don't think it is meaningful for a template itself to be an associated entity.

Does 4.4 not make it so? Certainly 4.3 would if "declared in classes" were corrected to "declared in classes or class templates".

source/basic.tex
```diff
@@ -1692,8 +1692,7 @@
 to be considered.
 The sets of namespaces and entities
 are determined entirely by
-the types of the function arguments
-(and the namespace of any template template argument).
+the types of the function arguments.
```
CaseyCarter: If we're killing words here, take the next sentence as well. The moons of Saturn also don't contribute to the sets of associated namespaces and entities, but we don't need to proclaim that fact normatively.

Quuxplusone: The parenthetical was nonsensical; the next sentence is merely redundant. I wouldn't object to removing the next sentence, but maybe I should open a completely orthogonal PR for that.

zygoloid:
This wording was added by P1103R3 and made necessary by changes in that same paper. See this wording in p4:

"Any declaration D in N that is in the interface of a named module M (10.2) is visible if there is an
associated entity attached to M with the same innermost enclosing non-inline namespace as D."

Passing a template exported from a module M as a template template argument makes the interface of module M visible to the lookup, and that only happens because we now consider template template arguments to introduce associated entities.

PR #2824, opened 2019-04-11
add forward declaration of atomic class for atomic<shared_ptr<T>> (P0718R2)

faithandbrave:
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0718r2.html

Or, I think the synopsis needs forward declaration of `atomic` class for specialization.


faithandbrave:
OK. I'll fix and rebase.

faithandbrave:
fixed

PR #2827, opened 2019-04-14
[range.counted] Introduce "Counted view"

### LINKED ISSUE #2825: [range.counted] Missing description
This is the only view without a description. It goes straight into normative wording.

CaseyCarter:
Fixes #2825.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -4868,6 +4868,11 @@ the expression \tcode{view::split(E, F)} is expression-equivalent to
 \rSec2[range.counted]{Counted view}
 
 \pnum
+A counted view presents a \libconcept{View} of the elements
+of the counted range\iref{iterator.requirements.general} \range{i}{n}
+for some iterator \tcode{i} and non-negative integer \tcode{n}.
+
+\pnum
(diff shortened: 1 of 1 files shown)


source/ranges.tex
```diff
@@ -4867,6 +4867,11 @@
 
 \rSec2[range.counted]{Counted view}
 
+\pnum
+A counted view presents a \libconcept{View} of the elements
+of the counted range\iref{iterator.requirements.general} \range{i}{n}
+for some iterator \tcode{i} and non-negative integer \tcode{n}.
```
zygoloid: If you're going to require `n` to be non-negative, should this be `\range{i}{i+n}`?

CaseyCarter: No, it should not. Outside of [algorithm], `i+n` only has well-defined meaning when `i` models `RandomAccessIterator` - which was much of the motivation for defining "counted range" in the first place.

zygoloid: Ah, right, this is that newfangled "counted range" thingy.

I don't like this abuse of notation. #2932 filed for that. But that shouldn't block this change.

PR #2830, opened 2019-04-16
[fs.rec.dir.itr.members] change \ensures to \remarks for pop()

jwakely:
As discussed on the LWG reflector, it doesn't make sense to state this as a postcondition of the function.

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -14314,13 +14314,13 @@ void pop(error_code& ec);
   iterated over, and continue iteration over the parent directory.
 
 \pnum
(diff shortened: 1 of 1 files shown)


source/iostreams.tex
```diff
@@ -14314,7 +14314,7 @@
   iterated over, and continue iteration over the parent directory.
 
 \pnum
-\ensures Any copies of the previous value of \tcode{*this}
+\remarks Any copies of the previous value of \tcode{*this}
```
CaseyCarter: `\remarks` should be aftern `\throws`.

PR #2833, opened 2019-04-18
[char.traits.specializations.char8.t] Make all members constexpr.

### LINKED ISSUE #2795: char_traits<char8_t> members not all constexpr
The paper P1032 "Misc constexpr bits" added constexpr to all char_traits members, at the same time as P0432 "char8_t: A type for UTF-8 characters and strings" added a new specialization for char_traits<char8_t> (without constexpr).

Depending on the order of applying the papers we could get two different results.

The intention of P1032 surely is that even the new members should be constexpr.

jensmaurer:
Fix missed application of P1032R1 to the result of applying P0482R6;
both papers were moved at the 2018-11 meeting.

Fixes #2795.

### DIFF
--- a/source/strings.tex
+++ b/source/strings.tex
@@ -333,9 +333,9 @@ namespace std {
     static constexpr size_t length(const char_type* s);
     static constexpr const char_type* find(const char_type* s, size_t n,
                                            const char_type& a);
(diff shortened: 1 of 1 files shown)


jensmaurer:
@zygoloid, here's your chance to get rid of this issue right away.

PR #2839, opened 2019-04-22
[temp.deduct.guide] CWG 2422: deduction-guide grammar fix

BRevzin:
A _deduction-guide_ should have a conditional _explicit-specifier_ rather than a conditional `explicit` - this is just oversight from P0892.


BRevzin:
Hopefully this counts as editorial? I think the design direction is clear here, and we even have wording in like [\[over.match.class.deduct\]](http://eel.is/c++draft/over.match.class.deduct#2.sentence-3) which says:

> If the function or function template was generated from a constructor or _deduction-guide_ that had an _explicit-specifier_, each such notional constructor is considered to have that same _explicit-specifier_.

jensmaurer:
This looks good to me, but changing grammar feels like a non-editorial change.

On the other hand, [over.match.class.deduct] already assumes that a _deduction-guide_ has an _explicit-specifier_:

> ... If the function or function template was generated from a constructor or _deduction-guide_ that had an _explicit-specifier_, each such notional constructor is considered to have that same _explicit-specifier_. ...

@zygoloid ?

zygoloid:
This definitely needs at least CWG review. I don't feel entirely comfortable making a grammar change such as this editorially, even though I think it would be within editorial remit to do so; unless CWG strongly wants this fixed editorially, I'd prefer that this go through the core issue process.

BRevzin:
I don't understand what the "core issue process" actually is. Is there a process? 

I sent an email to Core describing it. It got one response. I assume at this point the thread is just dead. This seems to be fairly typical. So I went ahead and just opened a PR since other parts of the wording already assume this to be the case.

jensmaurer:
Sent a reminder e-mail to Mike Miller.

BRevzin:
Just a friendly and encouraging ping, @jensmaurer @zygoloid. Is there an issue for this? How do I make one?

jensmaurer:
This is CWG 2422.

jensmaurer:
Fixed with cplusplus/draft@7c9d35da990bcd917ce6a25a08f6a629b346ccfe

PR #2841, opened 2019-04-23
[expos.only.func] Add library index entry for decay-copy

AlisdairM:
Following the same principle as locating GENERALIZED_SUM in
the library index, add an entry for the exposition-only function
decay-copy, which is used to specify several parts of the library
quite remote from where it is now defined.


jensmaurer:
We don't consistently index the use of defined terms. For example, we don't index all uses of "odr-use", either.

PR #2843, opened 2019-04-24
Add index entry for atomic union with padding

jfbastien:


jensmaurer:
I think our index is bad enough without deliberate attempts at making it worse.

PR #2853, opened 2019-04-30
[namespace.def]/1 Remove redundant statement

18:
[[namespace.def]/1](http://eel.is/c++draft/basic.namespace#namespace.def-1) states: 
> Every namespace-definition shall appear in the global scope or in a namespace scope

This is redundant, as global scope is a namespace. This could confuse readers to think that these are two separate things, when they in-fact are not. 

Proposed change is to remove "in the global scope".

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -7006,8 +7006,7 @@ see~\ref{basic.scope.namespace}.
 \end{bnf}
 
 \pnum
-Every \grammarterm{namespace-definition} shall appear in the global scope
-or in a namespace scope\iref{basic.scope.namespace}.
+Every \grammarterm{namespace-definition} shall appear at namespace scope\iref{basic.scope.namespace}.
 
 \pnum
 In a \grammarterm{named-namespace-definition},


source/declarations.tex
```diff
@@ -7006,8 +7006,10 @@
 \end{bnf}
 
 \pnum
-Every \grammarterm{namespace-definition} shall appear in the global scope
-or in a namespace scope\iref{basic.scope.namespace}.
+Every \grammarterm{namespace-definition} shall appear in namespace scope\iref{basic.scope.namespace}.
```
jensmaurer: Please say "at namespace scope" (which avoids the misunderstanding that someplace nested inside the namespace scope is fine, too (e.g. inside a class definition). Also, please drop the note that follows; in other situation, we never explicitly mention "global scope" when we say "at namespace scope".

18:
@jensmaurer Should be good now.

jensmaurer:
Please squash ("rebase -i") the commits and fix the commit comment (and the title of this issue) to reflect the current reality.

18:
@jensmaurer Should be good now.

jensmaurer:
No, you did some kind of merge. Please rebase and force-push instead.

tkoeppe:
@jensmaurer: You can also do that as part of merging: "squash and merge" creates a single, merged commit.

18:
Alright, now it should be good. Let me know if there is anything else I need to take care of.
(I've never used git command line :P)

jensmaurer:
@tkoeppe: Yes, but since nobody else chimed in, I need to leave this change for @zygoloid to apply. So I thought I should make it as ready as I can. Unless you're ok with the technical contents of the change (if so, please hit "Approve"), in which case I can merge myself.

PR #2856, opened 2019-04-30
[dcl.init.ref] References are bound to objects and functions, not expressions

### LINKED ISSUE #2855: [dcl.init.ref] References are bound to entities, not expressions
There are several occurrences in the wording saying that references are bound to expressions, which is incorrect:

[[dcl.init.ref]/5.1 sentence 1](http://eel.is/c++draft/dcl.init.ref#5.1.sentence-1)
> [...] the reference is bound to the initializer expression lvalue in the first case and to the lvalue result of the conversion in the second case

[[dcl.init.ref]/5.3 sentence 3](http://eel.is/c++draft/dcl.init.ref#5.3.sentence-3)
>[...] the reference is bound to the resulting glvalue

[[dcl.init.ref]/5.4.2](http://eel.is/c++draft/dcl.init.ref#5.4.2)
>The temporary materialization conversion is applied and the reference is bound to the result

Obviously, the reference should be bound to the object or function that the expression refers to, so this wording is technically incorrect.

# Proposed fix 
In each case where a reference is said to bind to an expression, it will be replaced with "the object or reference that the initializer refers to", or a variant thereof.

sdkrystian:
Proposed fix to close #2855

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -5256,6 +5256,9 @@ void f() {
 \end{example}
 
 \pnum
+A reference is said to be \defn{bound} to an object or function \tcode{E} when it is initialized to refer to \tcode{E}.
+
+\pnum
 A reference cannot be changed to refer to another object after initialization.
 \indextext{assignment!reference}%
 \begin{note}
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -5324,8 +5324,8 @@ function\iref{class.conv.fct} returning a reference type.}
 functions\iref{over.match.ref} and choosing the best one through overload
 resolution\iref{over.match}),
 \end{itemize}
-then the reference is bound to the initializer expression lvalue in the
-first case and to the lvalue result of the conversion
+then the reference is bound to the object or function that the initializer expression refers to in the
+first case and to the the object that the lvalue result of the conversion refers to
 in the second case (or, in either case, to the appropriate base class subobject of the object).
 \begin{note}
 The usual lvalue-to-rvalue\iref{conv.lval}, array-to-pointer\iref{conv.array},
@@ -5386,8 +5386,8 @@ If the converted initializer is a prvalue,
 its type \tcode{T4} is adjusted to type ``\cvqual{cv1} \tcode{T4}''\iref{conv.qual}
 and the temporary materialization conversion\iref{conv.rval} is applied.
 In any case,
-the reference is bound to the resulting glvalue
-(or to an appropriate base class subobject).
+the reference is bound to the object or function that the resulting glvalue
+refers to (or to an appropriate base class subobject).
 
 \begin{example}
 
@@ -5429,7 +5429,7 @@ Otherwise,
 the initializer expression is implicitly converted to a prvalue
 of type ``\cvqual{cv1} \tcode{T1}''.
 The temporary materialization conversion is applied and the reference is
-bound to the result.
+bound to the object that the resulting xvalue refers to.
 \end{itemize}
 
 If
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -5324,8 +5324,8 @@ function\iref{class.conv.fct} returning a reference type.}
 functions\iref{over.match.ref} and choosing the best one through overload
 resolution\iref{over.match}),
 \end{itemize}
-then the reference is bound to the object or function that the initializer expression refers to in the
-first case and to the the object that the lvalue result of the conversion refers to
+then the reference is bound to the object or function to which the initializer expression refers in the
+first case and to the object to which the lvalue result of the conversion refers
 in the second case (or, in either case, to the appropriate base class subobject of the object).
 \begin{note}
 The usual lvalue-to-rvalue\iref{conv.lval}, array-to-pointer\iref{conv.array},
@@ -5386,8 +5386,8 @@ If the converted initializer is a prvalue,
 its type \tcode{T4} is adjusted to type ``\cvqual{cv1} \tcode{T4}''\iref{conv.qual}
 and the temporary materialization conversion\iref{conv.rval} is applied.
 In any case,
-the reference is bound to the object or function that the resulting glvalue
-refers to (or to an appropriate base class subobject).
+the reference is bound to the object or function to which the resulting glvalue
+refers (or to an appropriate base class subobject).
 
 \begin{example}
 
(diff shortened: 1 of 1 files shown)


source/declarations.tex
```diff
@@ -5324,8 +5327,8 @@
 functions\iref{over.match.ref} and choosing the best one through overload
 resolution\iref{over.match}),
 \end{itemize}
-then the reference is bound to the initializer expression lvalue in the
-first case and to the lvalue result of the conversion
+then the reference is bound to the object or function to which the initializer expression refers in the
+first case and to the object to which the lvalue result of the conversion refers
```
zygoloid: The "lvalue result of the conversion" here was already correct; see the definition of "result" in [basic.lval]p5. I think we could use "result of" more broadly in this change. Eg:

"the reference is bound to the lvalue result of the initializer expression in the first case and to the lvalue result of the conversion in the second case".

source/declarations.tex
```diff
@@ -5386,8 +5389,8 @@
 its type \tcode{T4} is adjusted to type ``\cvqual{cv1} \tcode{T4}''\iref{conv.qual}
 and the temporary materialization conversion\iref{conv.rval} is applied.
 In any case,
-the reference is bound to the resulting glvalue
-(or to an appropriate base class subobject).
+the reference is bound to the object or function to which the resulting glvalue
+refers (or to an appropriate base class subobject).
```
zygoloid: How about just:

is bound to the resulting glvalue -> is bound to the glvalue result

source/declarations.tex
```diff
@@ -5429,7 +5432,7 @@
 the initializer expression is implicitly converted to a prvalue
 of type ``\cvqual{cv1} \tcode{T1}''.
 The temporary materialization conversion is applied and the reference is
-bound to the result.
+bound to the object to which the resulting xvalue refers.
```
zygoloid: I think this was already correct. The result of a conversion is the entity to which the evaluation of the conversion refers.

source/declarations.tex
```diff
@@ -5255,6 +5255,9 @@
 \end{codeblock}
 \end{example}
 
+\pnum
+A reference is said to be \defn{bound} to an object or function \tcode{E} when it is initialized to refer to \tcode{E}.
```
zygoloid: We're splitting the specification for references between here and [dcl.ref]p4. Over there, we find:

"A reference shall be initialized to refer to a valid object or function. [Note: In particular, a null reference cannot exist in a well-defined program, because the only way to create such a reference would be to bind it to the “object” obtained by indirection through a null pointer, which causes undefined behavior. As described in 11.3.9, a reference cannot be bound directly to a bit-field. — end note]"

I think neither place is the right place for these rules to live, though; [basic.compound] seems like a much better home for the above part of [dcl.ref]p4, for this new text, and for the next paragraph ("cannot be changed to refer to another object").

Perhaps add a new paragraph in [basic.compound] such as:

"""
A reference is *bound* to an entity determined by its initialization ([dcl.init.ref]). [Note: Argument passing ([expr.call]) and function value return ([stmt.return]) are initializations. — end note] If the entity to which a reference is bound is not a valid object or function, behavior is undefined. [Note: In particular, a null reference cannot exist in a well-defined program, because the only way to create such a reference would be to bind it to the “object” obtained by indirection through a null pointer, which causes undefined behavior. As described in 11.3.9, a reference cannot be bound directly to a bit-field. — end note] A reference cannot be changed to bind to another entity after initialization. [Note: Assignment to a reference assigns to the object referred to by the reference ([expr.ass]). — end note]
"""

... and remove the relevant wording from [dcl.init.ref] and [dcl.ref].

Hmm. We also don't appear to anywhere say that naming a reference results in the entity to which it is bound. Maybe the simplest way to deal with that would be to add a rule, perhaps just after [basic.lval]p5:

"If the result of a glvalue expression is a reference, the result is adjusted to be the entity to which the reference is bound."

zygoloid (commented):
More problems in this area: we sometimes say that references are bound to objects, and sometimes say that [temporary] objects are bound to references; in a couple of places we even say that temporary expressions are bound to references.

source/declarations.tex
18: Sounds good

source/declarations.tex
18: Ok, I'll make the change

source/declarations.tex
18: Naming a reference will result in the referred to object or function, see [expr.type] p1.

As for the wording you propose to add in [basic.compound], I believe entity is way too broad of a term to use. References only bind to objects and functions, and an entity is a whole laundry list of things, including types and templates. This could potentially be confusing to the reader.

source/declarations.tex
18: @zygoloid The result of a conversion is defined in terms of value categories, meaning that it's an expression as stated in [conv] p6.

source/declarations.tex
18: Looks good. One gripe I have with it is that when we say "glvalue" result, it could be interpreted in one of two ways, the resulting glvalue of the conversion (see my other comment as to why conversions result in expressions), or the result of the glvalue. It would be best to make it as clear as possible to avoid any confusion.

source/declarations.tex
zygoloid: When we talk about expressions, we often mean "expression or conversion". That is the case when we define "result" as well as when we defined value categories. I continue to think that this wording is correct as-is, though wording elsewhere may need updating.

RealLitb:
Does this need to modify http://eel.is/c++draft/over.ics.ref#1.sentence-1 aswell?

18:
@RealLitb No, "bind directly" is defined in terms of expressions http://eel.is/c++draft/dcl.init.ref#5.sentence-2

source/declarations.tex
languagelawyer: @zygoloid 
> A reference cannot be changed to bind to another entity after initialization.

When I read this, I always have very mixed feelings. On the one hand it is "obvious" that this wording means assignment. On the other hand, this wording is so vague that one might think about indirect change of which object a reference refers to. I mean rules in [basic.life] which list situations when a reference will refer to a new object when suchan object is created.

@18 
> I believe entity is way too broad of a term to use

There is also an opinion that *object* should not be *entity* (@opensdh)

languagelawyer: @zygoloid 
> A reference shall be initialized to refer to a valid object or function.

This wording is the concern of [CWG453](http://wg21.link/cwg453). Is it OK to rewrite it?

PR #2862, opened 2019-05-02
[expr.alignof,expr.unary.noexcept] Reorder to after [expr.sizeof].

### LINKED ISSUE #2773: reorder subclauses of [expr.unary] to match grammar
The subclauses of [expr.unary] should be in the same order as the grammar productions. They aren't. Reorder one or both so that they match. (I'm moving the preincrement operators after unary-operators in the grammar as part of merging the coroutines wording, but the relative order of the final four productions still needs to be fixed.)

jensmaurer:
Fixes #2773.


jensmaurer:
@tkoeppe, since this is just a subclause move, could you maybe have a look, even though this is in the core section? I'm prepared to commit this myself, but I'd prefer if someone else did a double-check.

PR #2864, opened 2019-05-03
[views.span] Move description of iterators to [span.iterators].

### LINKED ISSUE #2806: [span.overview] Iterator requirements should be appropriately relocated
The normative requirements on the iterator types of `span` [are in](http://eel.is/c++draft/span.overview#2) its description (before its class synopsis). I believe it would be more appropriate that they be relocated in a manner similar to that of [`string_view`'s](http://eel.is/c++draft/string.view.iterators#itemdecl:1).

jensmaurer:
Fixes #2806.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -10469,17 +10469,6 @@ A \tcode{span} is a view over a contiguous sequence of objects,
 the storage of which is owned by some other object.
 
 \pnum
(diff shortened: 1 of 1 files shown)


jensmaurer:
@jwakely or @CaseyCarter , could you please have a look?

PR #2865, opened 2019-05-03
[string.view.template][string.view.iterators] Move requirements to a more appropriate place

johelegp:
Moved down constexpr iterator requirement with the other iterator requirements, and moved up the invalidation requirements alongside the complexity requirements as both apply as generally.

### DIFF
--- a/source/strings.tex
+++ b/source/strings.tex
@@ -4081,8 +4081,11 @@ The program is ill-formed if \tcode{traits::char_type} is not the same type as \
 \end{note}
 
 \pnum
-The type \tcode{iterator} satisfies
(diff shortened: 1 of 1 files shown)


johelegp:
Right away.

> Where is it placed for `std::string`, for comparison?

http://wg21.link/string.require#4 and presumably subsumed from http://wg21.link/basic.string#2.

jensmaurer:
Uh, the automated checks fail. Maybe there's a surplus space at the end of a line, or an overfull hbox. Could you have a look, please?

johelegp:
Before:

After:

Seems like the `()` at the end. How do I solve this?

johelegp:
Now it looks alright.

PR #2868, opened 2019-05-03
[utilities,containers] Harmonize presentation of tuple_element.

### LINKED ISSUE #2786: fix presentation of `tuple_element` specializations
We have a variety of `tuple_element` specializations that use different presentation styles:

 * Some give a definition of the specialization
 * Others don't, and merely specify the `::type` member
 * Some use a *Value:* element for the type, others use a *Type:* element

We should do this properly: the specialization needs a specified definition, and a member typedef should be specified as actually being a type. (The current wording for several of these specializations completely lacks any suggestion that `...::type` is a *typedef-name*, for example.)

jensmaurer:
Fixes #2786.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -3574,16 +3574,16 @@ template<class T, size_t N>
 
 \indexlibrary{\idxcode{tuple_element}}%
 \begin{itemdecl}
-tuple_element<I, array<T, N>>::type
+template<size_t I, class T, size_t N>
+  struct tuple_element<I, array<T, N>> {
+    using type = T;
+  };
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
 \mandates
 \tcode{I < N} is \tcode{true}.
-
-\pnum
-\cvalue The type \tcode{T}.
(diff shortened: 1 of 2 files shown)


johelegp:
Per
https://github.com/cplusplus/draft/blob/c2e1df0bc599102038900a4adba7882d47cc57d9/source/utilities.tex#L1169-L1176
the pseudo-code as actual-code `TI` in
https://github.com/cplusplus/draft/blob/c2e1df0bc599102038900a4adba7882d47cc57d9/source/utilities.tex#L1800-L1817
could be replaced with `$\tcode{T}_i$`. Then the _Type:_ element could be removed. A back-reference to [tuple.cnstr]p1 would be useful then.

jensmaurer:
We would need towrite $\tcode{T}_\tcode{i}$ (note capital I).I'm not sure this is close enough to $\tcode{T}_i$ for users to make the connection to the template parameter without further explanations.

I'll apply the patch as-is; I think it's a net improvement. If the TI is still bothering someone, we can improve that in another iteration.

PR #2877, opened 2019-05-11
[dcl.dcl]/11 Storage from object definitions has proper alignment

random-internet-cat:
Require implementations to reserve storage of both proper size _and aligment_ in an object definition. Previously, this sentence requires only storage of proper size be reserved.

Additionally, specified that space is only reserved for object definitions, as it does not make sense to reserve space for a function definition.

Rationale for being editorial: if implementations are not required to reserve storage of proper alignment, accessing any variable is then possibly-UB due to lifetime failing to begin ([basic.life]/1). This is obviously not the intent of the Committee.


random-internet-cat:
Closed due to creating under wrong user account (sorry).

PR #2884, opened 2019-05-15
[range.filter.sentinel] Fix argument to ranges::end

cpplearner:
Here `end_` should be initialized with the sentinel of `parent`'s underlying view (instead of the sentinel of `parent`, which may lead to infinite recursion, if not a type mismatch).


CaseyCarter:
This looks suspiciously like the PR I just submitted to correct this issue. *sigh* Sorry for stomping on your fix.

PR #2887, opened 2019-05-21
[concept.swappable] Strike array poison pill swap overload

CaseyCarter:
Since lookup is only performed when at least one argument has class or enumeration type, template argument deduction can never succeed for the poisin pill overload with two parameters of reference-to-array type; it can be struck with no normative impact.

### DIFF
--- a/source/concepts.tex
+++ b/source/concepts.tex
@@ -578,12 +578,10 @@ and \tcode{E2} is expression-equivalent to an expression
   \tcode{S} is \tcode{(void)swap(E1, E2)}\footnote{The name \tcode{swap} is used
   here unqualified.} if \tcode{E1} or \tcode{E2}
   has class or enumeration type\iref{basic.compound} and that expression is valid, with
-  overload resolution performed in a context that includes the declarations
+  overload resolution performed in a context that includes the declaration
 \begin{codeblock}
   template<class T>
     void swap(T&, T&) = delete;
-  template<class T, size_t N>
-    void swap(T(&)[N], T(&)[N]) = delete;
 \end{codeblock}
   and does not include a declaration of \tcode{ranges::swap}.
   If the function selected by overload resolution does not


source/concepts.tex
```diff
@@ -578,12 +578,10 @@
   \tcode{S} is \tcode{(void)swap(E1, E2)}\footnote{The name \tcode{swap} is used
   here unqualified.} if \tcode{E1} or \tcode{E2}
   has class or enumeration type\iref{basic.compound} and that expression is valid, with
-  overload resolution performed in a context that includes the declarations
+  overload resolution performed in a context that includes the declaration
 \begin{codeblock}
   template<class T>
     void swap(T&, T&) = delete;
-  template<class T, size_t N>
-    void swap(T(&)[N], T(&)[N]) = delete;
```
jensmaurer: It appears to me that this case could be triggered by a conversion operator to reference-to-array type. Note that E1 and E2 are not required to be the same type. Or is the argument instead that the array case is subsumed by the plain "T" case? (It's more specialized, though.)

Example:
```
namespace N {
  using conv_type_id = int(&)[3];
  struct S {
    operator conv_type_id() { };
  };
  void swap(conv_type_id, conv_type_id) = delete;
}

int main()
{
  int E1[3];
  N::S E2;
  swap(E1, E2);  // attempts to call N::swap
}
```

CaseyCarter: > Or is the argument instead that the array case is subsumed by the plain "T" case?

Unfortunately, I think my argument is "I've completely overlooked `swap(array, convertible_to_array)` and I need to submit an LWG issue to fix `ranges::swap`." Thanks for catching this.

CaseyCarter: Crisis averted: template argument deduction doesn't see conversions, so the second parameter of the poison pill can't be deduced from `E2` in the call `std::ranges::swap(E1, E2)`; `N::swap` - if non-`delete`d - will be correctly called (https://godbolt.org/z/v4tgHG). Removing the poison pill won't affect this case either.

jensmaurer: I'm sorry, but I'm totally confused here. Of course, the second parameter of the poison pill can't be deduced from E2, but E1 will cause deduction of T and N used in the first parameter, then substitution will plug that into the second parameter, and finally the call will succeed using the conversion.

Your original statement said "the poison pill overload with two parameters of reference-to-array type will never be selected by overload resolution"; I showed one way how such an overload might be selected. What am I missing?

CaseyCarter: > I'm sorry, but I'm totally confused here. Of course, the second parameter of the poison pill can't be deduced from E2, but E1 will cause deduction of T and N used in the first parameter, then substitution will plug that into the second parameter, and finally the call will succeed using the conversion on E2.

At the risk of making a fool of myself arguing TAD with a Core member: that's not how template argument deduction works. In this case, both the first and second function parameters contain *template-parameter*s that participate in template argument deduction - `T` and `N` don't appear in non-deduced contexts - so deduction against both arguments must succeed. E.g.: https://godbolt.org/z/LYsQGW.

jensmaurer: You're right; sorry for the confusion. The mechanism I described only applies if the structure of the (dependent) parameter type uses template parameters in non-deducible ways.

jensmaurer:
So, after the detour, could you please fix the intro sentence for the pull request?
"will never be selected by overload resolution" is slightly misleading; we want to say "template argument deduction will never succeed for both parameters".

CaseyCarter:
> could you please fix the intro sentence for the pull request?

Updated in the first post and the commit message.

jensmaurer:
Thanks. That's as much preparation I can do for this right now, but this is certainly for @zygoloid to apply.

PR #2888, opened 2019-05-21
[concept.boolean] Rephrase first requirement

CaseyCarter:
`requires ConvertibleTo<const remove_reference_t<B>&, bool>;` is a simplification of `b1; requires ConvertibleTo<decltype(b1), bool>;` which would have been cleaned up to `{ b1 } -> ConvertibleTo<bool>;` (the intended requirement) post P1084, had it not been simplified in P0896.

~Drive-by: reorder the binary expression requirements consistently as b1 op b2, b1 op a, a op b2.~

### DIFF
--- a/source/concepts.tex
+++ b/source/concepts.tex
@@ -819,7 +819,7 @@ template<class B>
     Movable<remove_cvref_t<B>> && // (see \ref{concepts.object})
     requires(const remove_reference_t<B>& b1,
              const remove_reference_t<B>& b2, const bool a) {
-      requires ConvertibleTo<const remove_reference_t<B>&, bool>;
+      { b1 } -> ConvertibleTo<bool>;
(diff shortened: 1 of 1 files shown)


CaseyCarter:
> reorder the binary expression requirements consistently

I should point out that this technically has normative impact due to the short-circuiting nature of *requires-expression*s: there are sets of parameters which would have failed to satisfy the prior formulation that would result in an ill-formed program with the proposed formulation, and vice versa. I can confirm that the design intent is that the first two *compound-requirement*s precede the others and that the order of the others is not significant, but I'll of course understand if the editors would like LWG to give this change the nod.

CaseyCarter:
> In short, please remove the drive-by fixes from this pull request and create a new one with just the reordering.

Done.

PR #2889, opened 2019-05-21
[styles] Also allow 'listings' package v1.8b

tschuett:
MacTex 2019 comes with listings in version 1.8b. Without this change, it does not compile.

### DIFF
--- a/source/styles.tex
+++ b/source/styles.tex
@@ -156,13 +156,13 @@
 %% (copied verbatim from listings.sty version 1.6 except where commented)
 \makeatletter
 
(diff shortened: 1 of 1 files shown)


jwakely:
Does changing this file make _"copied verbatim from listings.sty version 1.6 except where commented"_ untrue? Was it already untrue?

jwakely:
The change itself looks fine, I just wonder if we should adjust the comment I quoted, or add a comment about the newer versions that are allowed.

tschuett:
I just followed the previous change: https://github.com/cplusplus/draft/pull/2473

jensmaurer:
@jwakely, the comment is still correct: What we have is copied, but if it also works with the subsequent versions, all is well.

PR #2890, opened 2019-05-21
[concept.booolean] Reorder expression requirements consistently

CaseyCarter:
...to improve readability.

I should point out that this technically has normative impact due to the short-circuiting nature of requires-expressions: there are sets of borderline-pathological parameters which would have failed to satisfy the prior formulation that would result in an ill-formed program with the proposed formulation, and vice versa. I can confirm that the design intent is that the first two compound-requirements precede the others and that the order of the others is not significant, but I'll of course understand if the editors would like LWG to give this change the nod.

### DIFF
--- a/source/concepts.tex
+++ b/source/concepts.tex
@@ -821,11 +821,11 @@ template<class B>
              const remove_reference_t<B>& b2, const bool a) {
       requires ConvertibleTo<const remove_reference_t<B>&, bool>;
       { !b1 } -> ConvertibleTo<bool>;
-      { b1 &&  a } -> Same<bool>;
-      { b1 ||  a } -> Same<bool>;
       { b1 && b2 } -> Same<bool>;
+      { b1 &&  a } -> Same<bool>;
       {  a && b2 } -> Same<bool>;
       { b1 || b2 } -> Same<bool>;
+      { b1 ||  a } -> Same<bool>;
       {  a || b2 } -> Same<bool>;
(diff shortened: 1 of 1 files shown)


jensmaurer:
@zygoloid, is this editorial?

jensmaurer:
Editorial meeting: We want a more general statement in the library front matter that substitution order in library concepts is not guaranteed, so that reorderings like the one we're discussing here are clearly editorial. @CaseyCarter , could you please make such an LWG issue happen?

CaseyCarter:
> Editorial meeting: We want a more general statement in the library front matter that substitution order in library concepts is not guaranteed, so that reorderings like the one we're discussing here are clearly editorial. @CaseyCarter , could you please make such an LWG issue happen?

Definitely not: there are cases in which the correct functioning of concepts depends on substitution order. I'm happy for such reorderings to remain non-editorial. 

(Should I close this PR and resubmit as an LWG issue, or assume that the addition of the "lwg" tag implies that LWG has already been informed of this PR and asked to handle it?)

jensmaurer:
@CaseyCarter , thanks.

Your statement means we can't editorially decide whether some order was intentional or not. So, please make this particular reordering an LWG issue. No, the label just means "we think it's an LWG issue" to prevent us from applying something in this area editorially.

Once you have an LWG issue number, feel free to put that in a comment here.

CaseyCarter:
The proposed change is now the P/R of LWG 3208.

PR #2892, opened 2019-05-22
[dcl.init.aggr]/1 An aggregate can have private or protected data members

languagelawyer:
if they are not direct, like in the code below:
```c++
class B { int i; };      // i is private
struct A : B { int j; }; // but A is still an aggregate

A a = { {}, 1 }; // OK
```

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -4620,7 +4620,7 @@ An \defn{aggregate} is an array or a class\iref{class} with
 \item
 no user-declared or inherited constructors\iref{class.ctor},
 \item
-no private or protected non-static data members\iref{class.access},
+no private or protected direct non-static data members\iref{class.access},
 \item
 no virtual functions\iref{class.virtual}, and
 \item


jensmaurer:
So, the superficial argument seems to be that this is just an editorial confusion between direct vs. base-class-recursive non-static data members. Or it could be considered an oversight when default member initializers were made to work with aggregates. @zygoloid, this seems borderline editorial.

languagelawyer:
> it could be considered an oversight when default member initializers were made to work with aggregates

I think default member initializers are not the key here, `class B { int i; };` is also fine. It looks like an oversight when aggregates were allowed to have base classes (http://wg21.link/p0017r1). Note that the definition of aggregate _elements_ uses "direct non-static data members" wording. (Also, the proposal for designated initializers (http://wg21.link/p0329r4) speaks about "direct non-static data members".)

zygoloid:
This seems editorial to me. When we allowed aggregates to have base classes, the design intent was that base classes are just regular aggregate elements like non-static data members, which I think makes it reasonably clear that this wording should be interpreted as applying only to the direct members.

jensmaurer:
So, let's apply this.

PR #2896, opened 2019-05-24
[std] Show stable labels for tables.

### LINKED ISSUE #2674: Table names should be present as stable names
The latex sources of the standard have "stable" table identifications as their \label{tablename}. However, in contrast to regular sections, these names are never rendered in the text. This leads to tedious updates of table numbers in any paper referring to any of the tables. It would be great if the table labels would be printed with the standard like the stable names are printed in the header sections.

This would ease the work of the paper authors as well as the editors (i guess).

jensmaurer:
Fixes #2674.

Please squash when applying.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -15,7 +15,7 @@ sorting and related operations,
 and algorithms from the ISO C library,
 as summarized in \tref{algorithms.summary}.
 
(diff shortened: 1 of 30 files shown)


jensmaurer:
Switch to "colon" separation, as opined in #2674.

PR #2899, opened 2019-05-27
[range.prim.size] Clarify by adding 'respectively'.

### LINKED ISSUE #2794: [range.prim.size] Missing "respectively"
> <li id="range.prim.size-1.3"><div class="marginalizedparent" style="left:-7em"><a class="marginalized" href="#range.prim.size-1.3">(1.3)</a></div><div id="range.prim.size-1.3.sentence-1" class="sentence">Otherwise, <span class="texttt"><span class="parenthesis">(</span>ranges<span class="operator">:</span><span class="operator">:</span>end<span class="parenthesis">(</span>E<span class="parenthesis">)</span> <span class="operator">-</span> ranges<span class="operator">:</span><span class="operator">:</span>begin<span class="parenthesis">(</span>E<span class="parenthesis">)</span><span class="parenthesis">)</span></span>
>   if it is a valid expression and
>   the types <span class="texttt">I</span> and <span class="texttt">S</span> of <span class="texttt">ranges<span class="operator">:</span><span class="operator">:</span>begin<span class="parenthesis">(</span>E<span class="parenthesis">)</span></span> and
>   <span class="texttt">ranges<span class="operator">:</span><span class="operator">:</span>end<span class="parenthesis">(</span>E<span class="parenthesis">)</span></span> model
>   <span class="texttt">SizedSentinel<span class="anglebracket">&lt;</span>S, I<span class="anglebracket">&gt;</span></span> (<a href="iterator.concept.sizedsentinel">[iterator.concept.sizedsentinel]</a>) and
>   <span class="texttt">ForwardIterator<span class="anglebracket">&lt;</span>I<span class="anglebracket">&gt;</span></span><a class="hidden_link" href="#range.prim.size-1.3.sentence-1">.</a></div> <div id="range.prim.size-1.3.sentence-2" class="sentence">However, <span class="texttt">E</span> is evaluated only once<a class="hidden_link" href="#range.prim.size-1.3.sentence-2">.</a></div></li>

I don't remember similar wording not including "respectively", so I'd guess this wording allows the interpretation of `S` denoting `ranges::begin(E)` and `I` denoting `ranges::end(E)`.

jensmaurer:
Fixes #2794.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -525,7 +525,7 @@ object\iref{customization.point.object}. The expression
   Otherwise, \tcode{(ranges::end(E) - ranges::begin(E))}
   if it is a valid expression and
   the types \tcode{I} and \tcode{S} of \tcode{ranges::begin(E)} and
-  \tcode{ranges::end(E)} model
+  \tcode{ranges::end(E)} (respectively) model both
   \tcode{SizedSentinel<S, I>}\iref{iterator.concept.sizedsentinel} and
   \tcode{Forward\-Iterator<I>}.
   However, \tcode{E} is evaluated only once.


jensmaurer:
@CaseyCarter, could you please have a look?

source/ranges.tex
```diff
@@ -525,7 +525,7 @@
   Otherwise, \tcode{(ranges::end(E) - ranges::begin(E))}
   if it is a valid expression and
   the types \tcode{I} and \tcode{S} of \tcode{ranges::begin(E)} and
-  \tcode{ranges::end(E)} model
+  \tcode{ranges::end(E)}, respectively, model
   \tcode{SizedSentinel<S, I>}\iref{iterator.concept.sizedsentinel} and
   \tcode{Forward\-Iterator<I>}.
```
zygoloid: I'm worried that this will now be misread as `I` needing to model `SizedSentinel<S, I>` and `S` needing to model `ForwardIterator<I>`. How about also adding "both" after "model"? I also think the commas are too much of a disruption to the flow of the sentence, and "(respectively)" would read better.

CaseyCarter: > How about also adding "both" after "model"?

Or combining the two constraint expressions into one:

Otherwise, `(ranges::end(E) - ranges::begin(E))` if it is a valid expression and the types `I` and `S` of `ranges::begin(E)` and `ranges::end(E)` (respectively) model `SizedSentinel<S, I> && ForwardIterator<I>`.

zygoloid: Is "model `Thing1 && Thing2`" meaningful? I thought "model" means that some English side-condition is satisfied, and we don't have a side-condition specified for `Thing1 && Thing2` (though I suspect people will infer the right meaning anyway).

jensmaurer: Fixed by adding "both". I don't like the `&&`.

PR #2904, opened 2019-05-28
[allocator.adaptor.syn] Avoid confusing term 'memory resource'.

### LINKED ISSUE #2858: [allocator.adaptor.syn] talks about "the memory resource"
Now that "memory resource" actually means something different (but still related to allocators) we should use different wording here.

jensmaurer:
Fixes #2858.


jensmaurer:
@jwakely, what do you think?

jwakely:
Oh I see we already say "allocator resource" later in the same paragraph. Great!

PR #2905, opened 2019-05-28
[std] Consistently use 'immediately-declared constraint'.

### LINKED ISSUE #2477: [dcl.type.auto.deduct] Use "immediately-declared constraint of"
https://github.com/cplusplus/draft/blob/65a4bacd2942e0e889bf713fe8db2321ab18fc29/source/declarations.tex#L1894-L1897
This is an instance where the antecedent of "its" is not entirely clear on first reading in relation to "immediately-declared constraint", which is a new term. An immediately-declared constraint is always one "of" a type; therefore, we are talking about the immediately-declared constraint of the type deduced for the placeholder.

Suggested wording:
> For a _placeholder-type-specifier_ with a _type-constraint_, the immediately-declared constraint is that of the type deduced for the placeholder. This constraint shall be satisfied.

jensmaurer:
Harmonize the phrasing in [expr.prim.req.compound],
[dcl.type.auto.deduct], [temp], and [temp.param].

Fixes #2477.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -2013,10 +2013,11 @@ decltype(auto)*x7d = &i;        // error, declared type is not plain \tcode{decl
 \end{example}
 
 \pnum
-For a \grammarterm{placeholder-type-specifier} with
-a \grammarterm{type-constraint},
-if the type deduced for the placeholder does not satisfy its
-immediately-declared constraint\iref{temp}, the program is ill-formed.
+For a \grammarterm{placeholder-type-specifier}
+with a \grammarterm{type-constraint},
+the immediately-declared constraint\iref{temp}
+of the \grammarterm{type-constraint} for the type deduced for the placeholder
+shall be satisfied.
 
 \rSec3[dcl.type.class.deduct]{Deduced class template specialization types}
 \indextext{deduction!class template arguments}%
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -2707,9 +2707,8 @@ Substitution of template arguments (if any)
 into the \grammarterm{return-type-requirement} is performed.
 
 \item
-The contextually-determined type being constrained
-by the \grammarterm{type-constraint} is \tcode{decltype((E))}.
-The immediately-declared constraint\iref{temp} of \tcode{decltype((E))}
+The immediately-declared constraint\iref{temp}
+of the \grammarterm{type-constraint} for \tcode{decltype((E))}
 shall be satisfied.
 \begin{example}
 Given concepts \tcode{C} and \tcode{D},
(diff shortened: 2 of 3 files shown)


jensmaurer:
@hubert-reinterpretcast, do you like this?

source/expressions.tex
```diff
@@ -2698,10 +2698,9 @@
 \item
 If the \grammarterm{return-type-requirement}
 is of the form \tcode{->} \grammarterm{type-constraint}, then
-the contextually-determined type being constrained
-is \tcode{decltype((E))}.
-The immediately-declared constraint\iref{temp} of \tcode{decltype((E))}
-shall be satisfied.
+the immediately-declared constraint\iref{temp}
+of the \grammarterm{type-constraint} is introduced for \tcode{decltype((E))};
```
hubert-reinterpretcast: > the immediately-declared constraint of the _type-constraint_ for `decltype((E))` is introduced; [ ... ]

jensmaurer:
Rebased.

source/expressions.tex
```diff
@@ -2708,10 +2708,8 @@
 into the \grammarterm{return-type-requirement} is performed.
 
 \item
-The contextually-determined type being constrained
-by the \grammarterm{type-constraint} is \tcode{decltype((E))}.
-The immediately-declared constraint\iref{temp} of \tcode{decltype((E))}
-shall be satisfied.
+The immediately-declared constraint\iref{temp}
+of the \grammarterm{type-constraint} for \tcode{decltype((E))} is introduced.
```
zygoloid: Don't we still need to say that the immediately-declared constraint shall be satisfied? As with the other comment, I don't think the "is introduced" here means anything; perhaps replacing "is introduced" with "shall be satisfied" in the revised wording would work?

source/declarations.tex
```diff
@@ -2025,10 +2025,10 @@
...
 
 \pnum
-For a \grammarterm{placeholder-type-specifier} with
-a \grammarterm{type-constraint},
-if the type deduced for the placeholder does not satisfy its
-immediately-declared constraint\iref{temp}, the program is ill-formed.
+A \grammarterm{placeholder-type-specifier} with a \grammarterm{type-constraint}
+introduces the immediately-declared constraint\iref{temp}
+of the \grammarterm{type-constraint} for the type deduced for the placeholder;
+this constraint shall be satisfied.
```
zygoloid: The "introduced" we have for template parameters is referencing the "introduces [...] constraint-expressions" in [temp.constr.decl]. That's not relevant in this case, I think, so I'm not sure it means anything to say that the immediately-declared constraint is introduced, and instead we should just be requiring the immediately-declared constraints to be satisfied. Maybe:

> For a *placeholder-type-specifier* with a *type-constraint*, the immediately-declared constraint ([temp]) of the *type-constraint* for the type deduced for the placeholder shall be satisfied.

source/expressions.tex
jensmaurer: Applied.

source/declarations.tex
jensmaurer: Applied.

PR #2906, opened 2019-05-29
[basic.compound] Replace four refs with a single one to [dcl.meaning]

jwakely:
Currently the cross references to 9.2.3.[1245] are not in numerical order, which just looks odd:

>  These methods of constructing types can be applied recursively; restrictions are mentioned in 9.2.3.1, 9.2.3.4, 9.2.3.5, and 9.2.3.2.

basic compound

It seems like [9.2.3.3 [dcl.mptr]](https://wg21.link/dcl.mptr#3) also mentions a restriction:

>  A pointer to member shall not point to a static member of a class (11.3.8), a member with reference type, or “_cv_ `void`”.

Should [dcl.mptr] be added to the list of refs in [basic.compound] p2?

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -4825,8 +4825,8 @@ called \term{pointer-to-member} types.
 
 \pnum
 These methods of constructing types can be applied recursively;
-restrictions are mentioned in~\ref{dcl.ptr}, \ref{dcl.array},
-\ref{dcl.fct}, and~\ref{dcl.ref}. Constructing a type such that the number of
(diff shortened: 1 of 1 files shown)


jensmaurer:
I'd much prefer to simply refer to [dcl.meaning] instead of enumerating each subclause individually.

jwakely:
Ok I'll push an update shortly.

jwakely:
Updated, so it now says:

basic compound-2

PR #2912, opened 2019-06-03
[over.call.object] Fix surrogate calls with regards to cv-qualifiers

random-internet-cat:
Standardizes existing practice. All of GCC, Clang, MSVC, ICC in pedantic mode already behave as if all of these changes were made.

Evidence for the above: 

no method cv-qualifier: https://godbolt.org/z/C0AcCi
both method cv-qualifiers: https://godbolt.org/z/-zy4GY
const pointer: https://godbolt.org/z/EVQ781
volatile pointer: https://godbolt.org/z/tePbKG
const volatile pointer: https://godbolt.org/z/f-g1Ge
reference to const pointer: https://godbolt.org/z/uXEp-J
reference to volatile pointer: https://godbolt.org/z/l2KEs9
reference to const volatile pointer: https://godbolt.org/z/5Mhijc

Other justification for the first two (in document order) changes: it is unlikely to be Committee intent to permit surrogate calls only on conversion functions with exactly one cv-qualifier.

### DIFF
--- a/source/overloading.tex
+++ b/source/overloading.tex
@@ -845,11 +845,11 @@ In addition, for each non-explicit conversion function declared in \tcode{T} of 
 form
 
 \begin{ncsimplebnf}
-\keyword{operator} conversion-type-id \terminal{(\,)} cv-qualifier \opt{ref-qualifier} \opt{noexcept-specifier} \opt{attribute-specifier-seq} \terminal{;}
+\keyword{operator} conversion-type-id \terminal{(\,)} \opt{cv-qualifier-seq} \opt{ref-qualifier} \opt{noexcept-specifier} \opt{attribute-specifier-seq} \terminal{;}
 \end{ncsimplebnf}
 
-where
-\grammarterm{cv-qualifier}
+where the optional
+\grammarterm{cv-qualifier-seq}
 is the same cv-qualification as, or a greater cv-qualification than,
 \cv{},
 and where


source/overloading.tex
```diff
@@ -845,18 +845,18 @@
...
-where
-\grammarterm{cv-qualifier}
+where the optional
+\grammarterm{cv-qualifier-seq}
 is the same cv-qualification as, or a greater cv-qualification than,
 \cv{},
 and where
 \grammarterm{conversion-type-id}
-denotes the type ``pointer to function
+denotes the type ``\cv{} pointer to function
```
jensmaurer: That seems wrong: The conversion-type-id has no cv-qualifiers exposed for this computation.

random-internet-cat: Why so? Would this be after adjustment of scalar types to non-cv qualified types? I didn't think so because the type in the declaration would still have cv-qualifiers, even if they are later logically removed.

jensmaurer: This checks whether the conversion-type-id denotes a certain type (pointer to function), and (on the face of it) relates it to the cv-qualifiers of the object type cv T. Why would we do that, given that we've already related cv to the cv-qualifier-seq in the preceding part of the sentence?

random-internet-cat: Ah, I see what happened. Those "cv"s that I added were meant to mean "possibly cv-qualified", not "the same cv-qualification as 'cv T'".

jensmaurer: Given that the half-sentence before actually refers to the "other" "cv", the phrasing seems most unfortunate. It seems you'd need cv_1 or so.

I also dislike the repetition of the parameter lists in the stuff that follows. Maybe "denotes a type "cv_1 pointer to F", "reference to F", or "reference to cv_1 pointer to F" with F "function (P1..Pn) returning R" works better.  Oh, that forgets the noexcept-specifier that is part of the function type.

As a first step, could you please restrict this pull request to the cv-qualifier -> cv-qualifier-seq change? This is easy to review, "obviously correct", etc.

random-internet-cat: Sure, will do. Sorry, I guess I don't know this, but would you prefer an amended commit and a force-push or two commits on the branch?

jensmaurer: Please fix the commit and force-push. Thanks.

random-internet-cat: Done

PR #2914, opened 2019-06-05
[range.semi.wrap] Rename "semiregular" to "semiregular-box"

CaseyCarter:
...to avoid confusion with the concept `Semiregular`. Update the uses in [range.single.view, range.filter.view,range.transform.view], and add a reference to [range.single.view] which now precedes [range.semi.wrap] since the "range factories" were pulled out into a separate subclause before the "range adaptors."

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -1549,7 +1549,7 @@ namespace std::ranges {
     requires is_object_v<T>
   class single_view : public view_interface<single_view<T>> {
   private:
-    @\placeholdernc{semiregular}@<T> value_;      // \expos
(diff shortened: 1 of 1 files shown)


source/ranges.tex
```diff
@@ -2328,28 +2328,28 @@
 \rSec2[range.semi.wrap]{Semiregular wrapper}
 
 \pnum
-Many of the types in this subclause are specified in terms of
```
CaseyCarter: NB: I removed "of the" from this line to fix an overfull hbox.

PR #2918, opened 2019-06-06
[expr.unary.op] Modernize wording for obtaining a pointer.

### LINKED ISSUE #2645: [expr.unary.op]/3: the value of a pointer should be "pointer to", not "address of"
http://eel.is/c++draft/expr.unary.op#3.sentence-4:
> Otherwise, if the type of the expression is T, **the result has type “pointer to T”** and is a prvalue that is **the address of the designated object** ([intro.memory]) or a pointer to the designated function.

1. «The _result_ of a prvalue is the value that the expression stores into its context.» ([basic.lval]). The lvalue can be of type `T`, but denote an object of type `U`, so the type of the result of applying the address-of operator to it will be `pointer to U`, not `pointer to T` (and I'm unsure if we even can talk about the type of the result. The result is "pointer to object", the type of the prvalue is `pointer to T`, but I don't think this makes sense to say that the type of the result is `pointer to U`). `pointer to T` is the type of the prvalue, not its result.
2. the prvalue has the value "pointer to the designated object or function", not "the address of the designated object or a pointer to the designated function."

The note following the sentence above (http://eel.is/c++draft/expr.unary.op#3.note-1) also needs a fix:
> In particular, the address of an object of type “cv T” is “pointer to cv T”

I don't think that addresses have types.

jensmaurer:
Fixes #2645.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -4157,11 +4157,13 @@ If the operand is a \grammarterm{qualified-id} naming a non-static or variant me
 of some class \tcode{C} with type \tcode{T}, the result has type ``pointer to member
 of class \tcode{C} of type \tcode{T}'' and is a prvalue designating \tcode{C::m}.
 \item
-Otherwise, if the operand is an lvalue of type \tcode{T}, the result has type ``pointer to
-\tcode{T}'' and is a prvalue that is the address of the designated object\iref{intro.memory}
-or a pointer to the designated function. \begin{note} In particular, the address of an
-object of type ``\cv{}~\tcode{T}'' is ``pointer to \cv{}~\tcode{T}'', with the same
-cv-qualification. \end{note}
+Otherwise, if the operand is an lvalue of type \tcode{T},
+the resulting expression is a prvalue of type ``pointer to \tcode{T}''
+whose result is a pointer to the designated object\iref{intro.memory} or function.
+\begin{note}
+In particular, taking the address of a variable of type ``\cv{}~\tcode{T}''
+yields a pointer of type ``pointer to \cv{}~\tcode{T}''.
+\end{note}
 For purposes of pointer arithmetic\iref{expr.add} and
 comparison~(\ref{expr.rel}, \ref{expr.eq}),
 an object that is not an array element whose


source/expressions.tex
```diff
@@ -4155,11 +4155,13 @@
 If the operand is a \grammarterm{qualified-id} naming a non-static or variant member \tcode{m}
 of some class \tcode{C} with type \tcode{T}, the result has type ``pointer to member
 of class \tcode{C} of type \tcode{T}'' and is a prvalue designating \tcode{C::m}.
-Otherwise, if the type of the expression is \tcode{T}, the result has type ``pointer to
-\tcode{T}'' and is a prvalue that is the address of the designated object\iref{intro.memory}
-or a pointer to the designated function. \begin{note} In particular, the address of an
-object of type ``\cv{}~\tcode{T}'' is ``pointer to \cv{}~\tcode{T}'', with the same
-cv-qualification. \end{note}
+Otherwise, if the type of the expression is \tcode{T},
```
CaseyCarter: It's unclear what expression "the expression" refers to here. I think it means "the operand when the operand is an lvalue", which leaves the behavior undefined when the operand is a *qualified-id* which does *not* name "a non-static or variant member `m` of some class `C` with type `T`."

I think we want to replace the sentence on lines 4155-4157 with: 
> If the operand is a *qualified-id*, it shall name a non-static or variant member `m` of some class `C` with type `T`; the result has type “pointer to member of class `C` of type `T`” and is a prvalue designating `C::m`.

And change later occurrences of "the expression" to "the operand", which must be an lvalue in those later occurrences.

source/expressions.tex
```diff
@@ -4155,11 +4155,13 @@
...
+the result is a prvalue of type ``pointer to \tcode{T}''
+whose value is a pointer to the designated object\iref{intro.memory} or function.
+\begin{note}
+In particular, taking the address of a variable of type ``\cv{}~\tcode{T}''
+yields a pointer of type ``pointer to \cv{}~\tcode{T}''.
+\end{note}
```
CaseyCarter: I think this note is extraneous, but I suspect the editorial telecon already considered eliminating it and chose not to do so.

source/expressions.tex
CaseyCarter: No, Casey: a *qualified-id* could be an lvalue *id-expression*. 4155-4157 is fine The *new* sentence needs to exclude the "*qualified-id* that is not an lvalue" case.

jensmaurer: I think that's an independent issue. The "Otherwise...expression" phrasing was already there and is used unchanged. I do agree that the preceding sentence is bad because it incorrectly mixes conditions and "shall" requirements. Can we fix that separately, please?

jensmaurer: Created #2919 to handle the bad if/otherwise ladder.

source/expressions.tex
```diff
@@ -4155,11 +4155,13 @@
 If the operand is a \grammarterm{qualified-id} naming a non-static or variant member \tcode{m}
 of some class \tcode{C} with type \tcode{T}, the result has type ``pointer to member
 of class \tcode{C} of type \tcode{T}'' and is a prvalue designating \tcode{C::m}.
-Otherwise, if the type of the expression is \tcode{T}, the result has type ``pointer to
-\tcode{T}'' and is a prvalue that is the address of the designated object\iref{intro.memory}
-or a pointer to the designated function. \begin{note} In particular, the address of an
-object of type ``\cv{}~\tcode{T}'' is ``pointer to \cv{}~\tcode{T}'', with the same
-cv-qualification. \end{note}
+Otherwise, if the type of the expression is \tcode{T},
+the result is a prvalue of type ``pointer to \tcode{T}''
```
zygoloid: The use of the word "result" here is unfortunate, since "result" is also the name we give to the value produced by the expression. How about this:

"Otherwise, if the type of the operand is `T`, the resulting expression is a prvalue of type ''pointer to `T`'' whose result is a pointer to the designated object or function."

zygoloid: Or maybe better (to avoid having two parallel-looking "if"s where one is restrictive and the other is not), how about:

"Otherwise, for an operand of type `T`, the resulting expression is a prvalue of type ''pointer to `T`'' whose result is a pointer to the designated object or function."

jensmaurer: Applied.

source/expressions.tex
```diff
@@ -4157,11 +4157,13 @@
 of some class \tcode{C} with type \tcode{T}, the result has type ``pointer to member
 of class \tcode{C} of type \tcode{T}'' and is a prvalue designating \tcode{C::m}.
 \item
-Otherwise, if the operand is an lvalue of type \tcode{T}, the result has type ``pointer to
-\tcode{T}'' and is a prvalue that is the address of the designated object\iref{intro.memory}
-or a pointer to the designated function. \begin{note} In particular, the address of an
-object of type ``\cv{}~\tcode{T}'' is ``pointer to \cv{}~\tcode{T}'', with the same
-cv-qualification. \end{note}
+Otherwise, for an lvalue operand of type \tcode{T},
```
zygoloid: Now that we've applied (as a separate PR) the change to use bullets here with a final "Otherwise", I think we want to go back to "Otherwise, if the operand is an lvalue of type `T`", because we really do mean "if", not "this is necessarily the case".

PR #2920, opened 2019-06-06
[expr.add] Avoid x[i] syntax when defining pointer arithmetic.

### LINKED ISSUE #2670: "Syntactic" circularity between [expr.add]/4 and [expr.sub]/1?
[expr.sub]/1 defines subscripting through pointer arithmetic
> The expression `E1[E2]` is identical (by definition) to `*((E1)+(E2))`

[expr.add]/4, defining pointer arithmetic, uses syntax looking like subscripting
> … if P points to element x[i] of an array object x with n elements,81 the expressions P + J and J + P (where J has the value j) point to the (possibly-hypothetical) element x[i+j]…

Not that `x[i]` necessarily means subscripting expression, but I think it is still better to replace such constructs with plain English wording like
> i-th element of an array object `x`

instead of
> element `x[i]` of an array object `x`

jensmaurer:
Fixes #2670.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -5708,20 +5708,20 @@ the result has the type of \tcode{P}.
 \begin{itemize}
 \item If \tcode{P} evaluates to a null pointer value and
 \tcode{J} evaluates to 0, the result is a null pointer value.
-\item Otherwise, if \tcode{P} points to element $\mathtt{x[}i\mathtt{]}$
-of an array object \tcode{x} with $n$ elements,%
+\item Otherwise, if \tcode{P} points to an array element $i$
+of an array object \tcode{x} with $n$ elements\iref{dcl.array},%
 \footnote{An object that is not an array element is considered to belong to a
 single-element array for this purpose; see~\ref{expr.unary.op}.
 A pointer past the last element of an array \tcode{x} of $n$ elements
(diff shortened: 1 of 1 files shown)


source/expressions.tex
```diff
@@ -5708,20 +5708,20 @@
...
 \item If \tcode{P} evaluates to a null pointer value and
 \tcode{J} evaluates to 0, the result is a null pointer value.
-\item Otherwise, if \tcode{P} points to element $\mathtt{x[}i\mathtt{]}$
-of an array object \tcode{x} with $n$ elements,%
+\item Otherwise, if \tcode{P} points to an array element $i$
+of an array object \tcode{x} with $n$ elements\iref{dcl.array},%
 \footnote{An object that is not an array element is considered to belong to a
 single-element array for this purpose; see~\ref{expr.unary.op}.
 A pointer past the last element of an array \tcode{x} of $n$ elements
 is considered to be equivalent to a pointer to a hypothetical element
```
tkoeppe: Maybe insert "hypothetical <ins>array</ins> element $n$"?

jensmaurer: We do say "element" (without "array") later on, too. Maybe those should all be fixed.

tkoeppe: Yes, indeed, that's what I was thinking, too. I think this is nice now.

PR #2925, opened 2019-06-11
[algorithms.parallel.exec],[alg.set.operations],[alg.min.max],[alg.clibrary] remove empty parens

jwakely:
When referring to a function, use its name, not a call expression.


jwakely:
Right. The last one is especially bad:

>  _Throws:_ Any exception thrown by `compar()`

... so it can't throw then, because `compar()` isn't even well-formed, right‽

This happens in other clauses too, but I'm just fixing one for now.

PR #2929, opened 2019-06-11
[range.split.outer] Convert trailing cross-reference to prefix style.

### LINKED ISSUE #2790: [range.split.outer] value_type to use conventional cross-reference
See https://github.com/cplusplus/draft/pull/2765#issuecomment-471356091.

jensmaurer:
Fixes #2790.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -4475,7 +4475,8 @@ namespace std::ranges {
     using iterator_concept  =
       conditional_t<ForwardRange<Base>, forward_iterator_tag, input_iterator_tag>;
     using iterator_category = input_iterator_tag;
-    struct value_type;                                  // see \ref{range.split.outer.value}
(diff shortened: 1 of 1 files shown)


source/ranges.tex
```diff
@@ -4475,7 +4475,8 @@
     using iterator_concept  =
       conditional_t<ForwardRange<Base>, forward_iterator_tag, input_iterator_tag>;
     using iterator_category = input_iterator_tag;
-    struct value_type;                                  // see \ref{range.split.outer.value}
+    // \ref{range.split.outer.value}, class \tcode{split_view::outer_iterator::value_type}
+    struct value_type;
```
CaseyCarter:
```suggestion
    struct value_type;
```

CaseyCarter: (FYI, This trailing whitespace is causing the Travis failure.)

jensmaurer: Fixed.

PR #2937, opened 2019-06-20
[range.filter.sentinel] Correct typo in constructor Effects

### LINKED ISSUE #2880: [range.filter.sentinel] filter_view::sentinel's single-parameter constructor seems wrong
[\[range.filter.sentinel\]](http://eel.is/c++draft/range.filter.sentinel):

>     constexpr explicit sentinel(filter_view& parent);
>
> Effects: Initializes `end_` with `ranges::end(parent)`.

1. `ranges::end(parent)` should probably be `ranges::end(parent.base_)`.

2. The design of sentinel's constructor seems to be unnecessarily inconsistent among range adaptors. 
  - `filter_view::sentinel` ([\[range.filter.sentinel\]](http://eel.is/c++draft/range.filter.sentinel#itemdecl:1)) and `join_view::sentinel` ([\[range.join.sentinel\]](http://eel.is/c++draft/range.join.sentinel#itemdecl:1)) have constructors that accept a reference to parent type, and call `ranges::end` inside that constructor. The parents simply pass `*this` as sentinel constructor argument in `end()`.
  - `transform_view::sentinel` ([\[range.transform.sentinel\]](http://eel.is/c++draft/range.transform.sentinel#itemdecl:1)) and `take_view::sentinel` ([\[range.take.sentinel\]](http://eel.is/c++draft/range.take.sentinel#itemdecl:1)) have constructors that accept a `sentinel_t<Base>`, the parents pass `ranges::end(base_)` as sentinel constructor argument in `end()`.

CaseyCarter:
Resolves #2880.

Note that I'm once again stretching "editorial" to avoid burning LWG time. The existing specification is obviously incorrect - it would result in either an ill-formed program or stack overflow - but the fix is not quite as obviously correct. As always, feel free to call this non-editorial and I'll file an LWG issue instead.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -2838,7 +2838,7 @@ constexpr explicit sentinel(filter_view& parent);
 
 \begin{itemdescr}
 \pnum
-\effects Initializes \tcode{end_} with \tcode{ranges::end(parent)}.
+\effects Initializes \tcode{end_} with \tcode{ranges::end(parent.base_)}.
 \end{itemdescr}
 
 \indexlibrary{\idxcode{base}!\idxcode{filter_view::sentinel}}%


jensmaurer:
@jwakely, are you ok with this change?

jwakely:
I agree it's currently broken, and I think the change represents what was always intended, so this qualifies as editorial for me.

PR #2941, opened 2019-06-24
[expr.prim.lambda] Capture j so that it compiles

b1sub:
The lambda captured **nothing**, but it returns the variable from the enclosing block scope. I assume this is editorial, and can easily be fixed by adding `&j`.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -1586,10 +1586,10 @@ type specified by the
 \tcode{return} statements as described in~\ref{dcl.spec.auto}.
 \begin{example}
 \begin{codeblock}
-auto x1 = [](int i){ return i; };     // OK: return type is \tcode{int}
-auto x2 = []{ return { 1, 2 }; };     // error: deducing return type from \grammarterm{braced-init-list}
+auto x1 = [](int i){ return i; };         // OK: return type is \tcode{int}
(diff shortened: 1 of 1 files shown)


source/expressions.tex
```diff
@@ -1589,7 +1589,7 @@
 auto x1 = [](int i){ return i; };     // OK: return type is \tcode{int}
 auto x2 = []{ return { 1, 2 }; };     // error: deducing return type from \grammarterm{braced-init-list}
 int j;
-auto x3 = []()->auto&& { return j; }; // OK: return type is \tcode{int\&}
+auto x3 = [&j]()->auto&& { return j; }; // OK: return type is \tcode{int\&}
```
tkoeppe: This seems to be moving comments around; can you please adjust everything according to our usual comment placement guidance?

b1sub: I'm sorry. Just fixed the indent.

tkoeppe: Thanks - [see wiki](https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#code-blocks), that preferred comment columns are 20, 32, 40, 48, etc. Can you snap to one of those?

zygoloid:
`j` is a global variable here, so capturing it is both unnecessary and ill-formed.

tkoeppe:
Hah, of course :-) I think we had this very issue before, where someone thought this is an excerpt from some function body.

PR #2954, opened 2019-07-04
[basic.life] Lifetime of class objects is treated uniformly

### LINKED ISSUE #2953: [basic.life]/5 was overlooked during CWG2256 resolution?
[basic.life]/5 still says that
> A program may end the lifetime of any object by reusing the storage which the object occupies or by explicitly calling the destructor for an object of a class type with a non-trivial destructor.

even though the requirement about non-triviality of a destructor was removed from [basic.life]/1 during CWG2256 resolution and now it says that
> The lifetime of an object o of type T ends when:
 — if T is a non-class type, the object is destroyed, or 
 — if T is a class type, the destructor call starts, or

jensmaurer:
under CWG2256, regardless of triviality of the destructor.

Fixes #2953.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3137,12 +3137,12 @@ construction and destruction phases. \end{note}
 \pnum
 A program may end the lifetime of any object by reusing the storage
 which the object occupies or by explicitly calling the destructor for an
-object of a class type with a non-trivial destructor. For an object of a
-class type with a non-trivial destructor, the program is not required to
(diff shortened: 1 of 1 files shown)


jensmaurer:
@zygoloid, we no longer differentiate object lifetime according to the triviality of the destructor in [basic.life]. (If the destructor is trivial, p5 is vacuous anyway, because there are no side effects of the destructor execution.)

PR #2961, opened 2019-07-08
[INDEX] Ensure index entry for all Cpp17Requirements

AlisdairM:
Ensure that each Cpp17Requirement that uses the \oldconcept markup
has one, and exactly one, usage close to its definition that uses
the \oldconceptdefn markup instead.  This puts a consistent set of
entries with the Cpp17 prefix into the main index.

Note that there are existing, inconsistent, attempts to index these
requirements, with the container 'into' requirements appearing at
the top of the index due to a \defnx (and the markup somehow changing
the sort order) and most of the lib-intro requirements appearing as
sub-entries under 'Requirements'.  This change-set updates the
\oldconceptdefn macro to use \defnx consistent with the containers
clause usage, producing a single consistent list that is out of place
w.r.t. the spelling, and does not refactor the mark-up for containers.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -1730,7 +1730,7 @@ denotes a value of some type that is writable to the output iterator.
 \begin{note} For an iterator type \tcode{X} there must be an instantiation
 of \tcode{iterator_traits<X>}\iref{iterator.traits}. \end{note}
 
-\rSec3[iterator.iterators]{\oldconcept{Iterator}}
+\rSec3[iterator.iterators]{\oldconceptdefn{Iterator}}
 
 \pnum
 The \oldconcept{Iterator} requirements form the basis of the iterator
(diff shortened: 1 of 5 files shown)


AlisdairM:
Confirmed at Cologne that this needs some reworking, so I am closing the PR to avoid clogging the list of PRs at a busy time, and ensuring no even minor risk of it accidentally landing.  It will return once the C++20 ballot document is safely out for review.

CaseyCarter:
> This puts a consistent set of
> entries with the Cpp17 prefix into the main index.

Be forewarned that I plan to submit an editorial PR to strip the *`Cpp17`* prefix from the names of the requirement sets post-CD. (We no longer need the prefix to avoid clashes with True Library Concepts now that the library concepts use snake_case.) If that change is approved, you'll need to modify your strategy if you want the new index entries all together.

AlisdairM:
Thanks Casey - I already have several ideas in hand for how I will handle that, but am in no rush until after the ballot CD is complete.

PR #2962, opened 2019-07-09
[unique.ptr] Clarify nullptr case in destructor and reset()

chradcliffe:
In `unique_ptr::~unique_ptr` and `unique_ptr::reset`, the Requires statements are ambiguous about whether `get_deleter()(nullptr)` should have well-defined behaviour.

From the Effects statements for both specifications, it's clear that `get_deleter(get())` is not called in the case where the `unique_ptr` contains `nullptr`. This fix makes it clear that `get_deleter(nullptr)` is not required to have well-defined behaviour.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -9288,11 +9288,10 @@ to the stored deleter that was constructed from
 
 \begin{itemdescr}
 \pnum
-\requires The expression \tcode{get_deleter()(get())} shall be well-formed,
-shall have well-defined behavior, and shall not throw exceptions.
-\begin{note}
-The
-use of \tcode{default_delete} requires \tcode{T} to be a complete type.
+\requires The expression \tcode{get_deleter()(get())} shall be well-formed, and shall
+not throw exceptions. If \tcode{get() != nullptr}, \tcode{get_deleter()(get())}
+shall have well-defined behavior. \begin{note} The use of \tcode{default_delete}
+requires \tcode{T} to be a complete type.
 \end{note}
 
 \pnum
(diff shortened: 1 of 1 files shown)


chradcliffe:
@jwakely in my proposed change, there's no change to the applicability of "well-defined". The change is to the applicability of "well-defined behavior" in the case where the `unique_ptr` contains a `nullptr`. If I understand the standardese correctly, "well-defined behavior" can be dependent on runtime values.

jwakely:
Doh, you're quite right, I misread the diff, sorry.

jwakely:
Now that I've remembered how to read, I think the "shall not throw  exceptions" requirement should only apply to the `get() != nullptr` case. The deleter is allowed to throw exceptions in the general case, just not when invoked by `unique_ptr`.

jwakely:
Thanks for the update.

It might be easier to use the new specification structure that the standard library clauses are migrating to:

> _Mandates:_ The expression \tcode{get_deleter()(get())} shall be well-formed.

> _Expects:_ If \tcode{get() != nullptr}, the expression \tcode{get_deleter()(get())} has well-defined behavior and does not throw exceptions.

But somebody else is already working on doing that.

Whichever structure we use, this is a requirement that says "if this expression isn't well-defined, then it's undefined", which is a tautology. The "shall not throw" part is still useful though.

chradcliffe:
I stumbled across this because I was curious as to whether the standard required a null check on the part of the deleter. I think it's still helpful to say that there is no requirement for `get_deleter()(nullptr)` to have well-defined behaviour. Is there maybe a better way of expressing that?

jensmaurer:
Well, you could add a note (as in "[Note: blah -- end note]") to say that the result of get_deleter() is never invoked with a null pointer value, and thus no requirements are imposed on the deleter for that argument value.

CaseyCarter:
Were I to specify these destructors in a green field, I'd write simply:
> *Effects*: Equivalent to: `if (get()) get_deleter()(get());`
and not bother with explicit requirements.

(Note that "no effects" is just as wrong here as virtually every place we use it in the Library: `pointer` can be a user-defined type, so the `null` test is a user-observable effect.)

jensmaurer:
@CaseyCarter, I fully agree. However, introducing "Equivalent to" in an editorial issue is usually ...  not what we do. @zygoloid?

jensmaurer:

jensmaurer:
Superseded by #4736

PR #2994, opened 2019-07-21
[expr.unary.noexcept] Delete irrelevant copy-paste from previous section

YehezkelShB:
This looks like a copy-paste from previous operators (`sizeof`, `alignof`), which actually return `size_t`.
(See 8eff9b95)


YehezkelShB:
This is editorial as it is about a note, and was introduced by another editorial change.

YehezkelShB:
@jensmaurer (the author of the referenced commit)

PR #3037, opened 2019-07-22
P1208R6 Adopt source_location for C++20

### LINKED ISSUE #3036: [2019-07 LWG Motion 32] P1208R6 Adopt source_location for C++20

jensmaurer:
- Changed stable name to [support.srcloc].
 - Added cross-references.
 - Changed "description" environment showing result values to a table

Fixes #3036.

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -30,6 +30,7 @@ as summarized in \tref{support.summary}.
 \ref{support.start.term}  & Start and termination     &   \tcode{<cstdlib>}   \\ \rowsep
 \ref{support.dynamic}     & Dynamic memory management &   \tcode{<new>}       \\ \rowsep
 \ref{support.rtti}        & Type identification       &   \tcode{<typeinfo>}  \\ \rowsep
+\ref{support.srcloc}      & Source location           &   \tcode{<source_location>} \\ \rowsep
 \ref{support.exception}   & Exception handling        &   \tcode{<exception>} \\ \rowsep
 \ref{support.initlist}    & Initializer lists    & \tcode{<initializer_list>} \\ \rowsep
 \ref{cmp}                 & Comparisons               &   \tcode{<compare>}   \\ \rowsep
@@ -712,6 +713,8 @@ the values of these macros with greater values.
   \tcode{<memory>} \\ \rowsep
 \defnlibxname{cpp_lib_shared_timed_mutex}                   & \tcode{201402L} &
   \tcode{<shared_mutex>} \\ \rowsep
+\defnlibxname{cpp_lib_source_location}                      & \tcode{201907L} &
+  \tcode{<source_location>} \\ \rowsep
 \defnlibxname{cpp_lib_spaceship}                            & \tcode{201907L} &
   \tcode{<compare>} \\ \rowsep
 \defnlibxname{cpp_lib_string_udls}                          & \tcode{201304L} &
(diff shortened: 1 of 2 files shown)


jensmaurer:
@zygoloid, this might need more work.
- Both the section headings "Creation" and "Field access" are novel.
- The subclause nesting of "synopsis" inside "class source_location" is novel.
- Showing values of exposition-only data members (as opposed to results of member function calls) differs from at least [tab:re.results.const].

Please advise.

jwakely:
> Both the section headings "Creation" and "Field access" are novel.

We have subclauses called simply "Creation" e.g. [util.smartptr.shared.create] and [uniq.ptr.create], and we have "Tuple creation functions" [tuple.creation]. I don't like "fields" much. We use "observers" in some other places.

> The subclause nesting of "synopsis" inside "class source_location" is novel.

And there are hanging paragraphs.

> Showing values of exposition-only data members (as opposed to results of member function calls) differs from at least [tab:re.results.const].

The regex clause is not considered a model of good library wording :-) Tables showing postconditions in terms of member functions are used in many places, but if we do that, the _Returns:_ elements in [reflection.src_loc.fields] probably need reworking.

The reason for the presentation in the R6 revision is that previous proposals had prose for the member functions saying "represented by this object" and it wasn't clear what the location represented by the object was (especially if not created by a call to `current()`). By introducing exposition-only data members and saying how they're initialized in p1 (1.1) that problem went away, and the _Returns:_ elements in [reflection.src_loc.fields] were trivial to specify in terms of the data members.

jwakely:
To be clear, the table seems fine to me. I'm just trying to explain why the effects are stated in terms of the exposition-only data members rather than the member functions.

source/support.tex
```diff
@@ -3166,6 +3168,214 @@
...
+}
+\end{codeblock}
+
+\pnum
+The type \tcode{source_location} meets the
+\oldconcept{DefaultConstructible},
+\oldconcept{CopyConstructible},
+\oldconcept{Copy\-Assignable}, and
+\oldconcept{Destructible}
+requirements\iref{utility.arg.requirements}.
```
burblebee: The paper wanted a new paragraph here.

source/support.tex
```diff
@@ -3166,6 +3168,214 @@
...
+All of the following conditions are true:
+\begin{itemize}
+\item \tcode{is_nothrow_move_constructible_v<source_location>}
+\item \tcode{is_nothrow_move_assignable_v<source_location>}
+\item \tcode{is_nothrow_swappable_v<source_location>}
+\end{itemize}
+\begin{note}
+The intent of \tcode{source_location} is
+to have a small size and efficient copying.
+\end{note}
```
burblebee: The paper wants this in a separate paragraph, and I agree - it has to do with the class, not the conditions mentioned in the paragraph it's currently in.

source/support.tex
```diff
@@ -3166,6 +3168,214 @@
...
+
+\pnum
+The data members \tcode{file_name_} and \tcode{function_name_}
+always each refer to an NTBS.
```
burblebee: NTBS -> \ntbs{}

source/support.tex
```diff
@@ -3166,6 +3168,214 @@
...
+  however, an implementation is encouraged to use 0
+  when the line number is unknown. \\ \rowsep
+\tcode{column_}  &
+  An implementation-defined value denoting
+  some offset from the start of the line denoted by \tcode{line_}.
+  Column numbers are presumed to be 1-indexed;
+  however, an implementation is encouraged to use 0
+  when the column number is unknown. \\ \rowsep
+\tcode{file_name_} &
+  A presumed name of the current source file\iref{cpp.predefined} as an NTBS.
```
burblebee: NTBS -> \ntbs{}

source/support.tex
```diff
@@ -3166,6 +3168,214 @@
...
+  \\ \rowsep
+\tcode{function_name_} &
+  A name of the current function
+  such as in \tcode{__func__}\iref{dcl.fct.def.general} if any,
+  an empty string otherwise. \\
+\end{libefftabvalue}
+
+\item
+  Otherwise, that is, when invoked in some other way, returns a
```
burblebee: remove "that is,"

source/support.tex
```diff
@@ -3166,6 +3168,214 @@
...
+struct s {
+  source_location member = source_location::current();
+  int other_member;
+  s(source_location loc = source_location::current())
+    : member(loc)               // values of \tcode{member} will be from call-site
+  {}
+  s(int blather) :              // values of \tcode{member} should be hereabouts
+    other_member(blather)
+  {}
+  s(double)                     // values of \tcode{member} should be hereabouts
```
burblebee: Can we reword this?  values can't be "hereabouts".

source/support.tex
```diff
@@ -3166,6 +3168,214 @@
...
+\end{itemdescr}
+
+\pnum
+\begin{example}
+\begin{codeblock}
+struct s {
+  source_location member = source_location::current();
+  int other_member;
+  s(source_location loc = source_location::current())
+    : member(loc)               // values of \tcode{member} will be from call-site
```
burblebee: Can we do better than "from call-site"?

source/support.tex
```diff
@@ -3166,6 +3168,214 @@
...
+  {}
+  s(int blather) :              // values of \tcode{member} should be hereabouts
+    other_member(blather)
+  {}
+  s(double)                     // values of \tcode{member} should be hereabouts
+  {}
+};
+void f(source_location a = source_location::current()) {
+  source_location b = source_location::current();       // values in \tcode{b} represent this line
```
burblebee: huh?  values can represent a line???

source/support.tex
```diff
@@ -3166,6 +3168,214 @@
...
+}
+
+void g() {
+  f();                          // \tcode{f}'s first argument corresponds to this line of code
```
burblebee: need better wording here too - what does it mean for an argument to correspond to a line?

source/support.tex
jensmaurer: Yes, but single-sentence paragraphs are bad. Both this paragraph and the next describe requirements for the implementation-provided source_location type, so not having a new paragraph seems reasonable.

source/support.tex
jensmaurer: I disagree. The note wants to say that those "nothrow-copy" rules are conducive to the goal in the note. Again, I want to avoid single-sentence paragraphs.

source/support.tex
jensmaurer: Right. Fixed.

source/support.tex
jensmaurer: We could even remove "when invoked in some other way", because that's redundant with "Otherwise".

source/support.tex
jensmaurer: Hopefully improved.

source/support.tex
jensmaurer: Updated.

source/support.tex
jensmaurer: Updated.

source/support.tex

source/support.tex
```diff
@@ -3166,6 +3169,214 @@
...
+\oldconcept{DefaultConstructible},
+\oldconcept{CopyConstructible},
+\oldconcept{Copy\-Assignable}, and
+\oldconcept{Destructible}
+requirements\iref{utility.arg.requirements}.
+Lvalues of type \tcode{source_location}
+are swappable\iref{swappable.requirements}.
+
+\pnum
+All of the following conditions are true:
```
```suggestion
All of the following conditions are \tcode{true}:
```

source/support.tex
johelegp: > Right. Fixed.

Not really.

source/support.tex

source/support.tex
burblebee: Then combine it with the next paragraph?  otherwise, please obey the paper and place in a separate paragraph - it does not belong with this paragraph.

burblebee:
@jensmaurer please fix the build

jensmaurer:
@burblebee: Looks like some timeout during the build; nothing I can fix:

47
15.36s$ docker pull godbyk/texlive-basic:latest
448
Error response from daemon: Get https://registry-1.docker.io/v2/godbyk/texlive-basic/manifests/latest: Get https://auth.docker.io/token?scope=repository%3Agodbyk%2Ftexlive-basic%3Apull&service=registry.docker.io: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
449
The command "docker pull godbyk/texlive-basic:latest" failed and exited with 1 during .

source/support.tex
jensmaurer: Yes, sounds reasonable. Done.

zygoloid:
Some of the organization is novel, but the CD is already late so I'm not going to block on changing that.

source/support.tex
jwakely: I believe "hereabouts" was intentional, because it's unspecified whether the location is the end of the constructor's _parameter-declaration-clause_, or the start/end of the _ctor-initializer_, or the opening brace of the _function-body_, or somewhere else thereabouts.

But I suppose "refer to this location" in the new wording is vague enough as well.

source/support.tex
jwakely: That's the whole point of this class!

PR #3039, opened 2019-07-22
P0553R4 Bit operations

### LINKED ISSUE #3007: [2019-07 LWG Motion 3] P0553R4 Bit operations

jensmaurer:
- Renamed stable labels.

Fixes #3007.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -1317,6 +1317,24 @@ namespace std {
     constexpr T floor2(T x) noexcept;
   template<class T>
     constexpr T log2p1(T x) noexcept;
+
+  // \ref{bit.rotate}, rotating
+  template<class T>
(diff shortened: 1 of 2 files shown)


source/numerics.tex
```diff
@@ -1437,6 +1455,154 @@
 unless \tcode{T} is an unsigned integer type\iref{basic.fundamental}.
 \end{itemdescr}
 
+\rSec2[bit.rotate]{Rotating}
+
+In the following descriptions,
+let \tcode{N} denote \tcode{std::numeric_limits<T>::digits}.
```
jwakely: I don't think we need the `std::` qualification on `numeric_limits`, both here and other places below (sorry for not noticing that before we voted on it).

jensmaurer: Fixed.

PR #3041, opened 2019-07-22
P1331R2 Permitting trivial default initialization in constexpr contexts

### LINKED ISSUE #2982: [2019-07 CWG Motion 3] P1331R2 Permitting trivial default initialization in constexpr contexts

zygoloid:
[expr.const] Add missing "an" before "indeterminate value".

Fixes #2982

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -6936,6 +6936,10 @@ that is applied to a glvalue
 that refers to a non-active member of a union or a subobject thereof;
 
 \item
+an lvalue-to-rvalue conversion that is applied to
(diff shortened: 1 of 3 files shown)


source/expressions.tex
```diff
@@ -6895,6 +6895,10 @@
 that is applied to a glvalue
 that refers to a non-active member of a union or a subobject thereof;
 
+\item
+an lvalue-to-rvalue conversion that is applied to
+an object with an indeterminate value\iref{basic.indet};
```
cpplearner:
```suggestion
an object with an indeterminate value\iref{basic.indet};
```

PR #3043, opened 2019-07-22
P0848R3 Conditionally Trivial Special Member Functions

### LINKED ISSUE #2984: [2019-07 CWG Motion 5] P0848R3 Conditionally Trivial Special Member Functions

zygoloid:
Fixes #2984

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -143,22 +143,40 @@ Aggregates of class type are described in~\ref{dcl.init.aggr}.
 \rSec1[class.prop]{Properties of classes}
 
 \pnum
+Two special member functions are of the same kind if:
+\begin{itemize}
+\item they are both default constructors,
+\item they are both copy or move constructors
+with the same first parameter type, or
+\item they are both copy or move assignment operators
+with the same first parameter type
+and the same \grammarterm{cv-qualifier}s and \grammarterm{ref-qualifier}, if any.
+\end{itemize}
+
+\pnum
+An \defnadj{eligible}{special member function} is a special member function\iref{special} for which:
+\begin{itemize}
+\item the function is not deleted,
+\item the associated constraints\iref{temp.constr}, if any, are satisfied, and
+\item no special member function of the same kind is more constrained\iref{temp.constr.order}.
+\end{itemize}
+
+\pnum
 A \defnadj{trivially copyable}{class} is a class:
 
 \begin{itemize}
-\item where each copy constructor, move constructor, copy assignment operator,
-and move assignment operator~(\ref{class.copy.ctor}, \ref{class.copy.assign})
-is either deleted or trivial,
-\item that has at least one non-deleted copy constructor, move constructor,
-copy assignment operator, or move assignment operator, and
+\item that has at least one eligible copy constructor, move constructor,
+copy assignment operator, or
+move assignment operator~(\ref{class.copy.ctor}, \ref{class.copy.assign}),
+\item where each eligible copy constructor, move constructor, copy assignment operator,
+and move assignment operator is trivial, and
 \item that has a trivial, non-deleted destructor\iref{class.dtor}.
 \end{itemize}
 
 \pnum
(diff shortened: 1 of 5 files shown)


source/classes.tex
```diff
@@ -2090,7 +2108,7 @@
 \indextext{special member function|see{destructor}}%
 
 \pnum
-In a declaration of a destructor, the \grammarterm{declarator} is a
+In a declaration of a prospective destructor, the \grammarterm{declarator} is a
```
burblebee: Add ref to [class.dtor] please.

source/classes.tex
```diff
@@ -2118,10 +2136,25 @@
 same class as the \grammarterm{nested-name-specifier}.
 \end{itemize}
 
-A destructor shall take no arguments\iref{dcl.fct}.
+A prospective destructor shall take no arguments\iref{dcl.fct}.
```
burblebee: What is a "prospective destructor"???  We never define it - can we please?

source/classes.tex
```diff
@@ -2118,10 +2136,25 @@
...
+At the end of the definition of a class,
+overload resolution is performed
+among the prospective destructors declared in that class
+with an empty argument list
+to select the \defn{destructor} for the class,
+also known as the \defnadj{selected}{destructor}.
+The program is ill-formed if overload resolution fails.
+Destructor selection does not constitute
+a reference to,
+or odr-use\iref{basic.def.odr} of,
```
burblebee: If we add the reference after "reference to" as suggested in the paper, it would read better if the reference here were moved after the "of".

source/classes.tex
```diff
@@ -2118,10 +2136,25 @@
...
+\pnum
+At the end of the definition of a class,
+overload resolution is performed
+among the prospective destructors declared in that class
+with an empty argument list
+to select the \defn{destructor} for the class,
+also known as the \defnadj{selected}{destructor}.
+The program is ill-formed if overload resolution fails.
+Destructor selection does not constitute
+a reference to,
```
burblebee: The paper asks for a reference to [dcl.fct.def.delete] here.

source/templates.tex
```diff
@@ -6426,6 +6426,10 @@
 and is an explicit instantiation definition of only those
 members that have been defined at the point of instantiation.
 
+\pnum
+An explicit instantiation of a prospective destructor
```
burblebee: Add reference to `\iref{class.dtor}`.

source/classes.tex
CaseyCarter: This is the first paragraph of [class.dtor].

source/classes.tex
CaseyCarter: Much like how the prior state of this subclause ([class.dtor]) was to define the term "destructor" by describing the properties of such a thing, this subclause now defines "prospective destructor" by describing its properties. Effectively this subclause  *is* the definition of "prospective destructor".

CaseyCarter: That said, I certainly wouldn't mind wrapping the first mention of "prospective destructor" in paragraph 1 with `\defn`.

source/classes.tex
burblebee: Oops - wrong spot.

source/classes.tex
burblebee: I would like to see a proper `\defnadj{prospective}{destructor}` definition.  I know it wasn't in the paper, but it should have been.  Should be easy enough to fix it editorially.

source/classes.tex
zygoloid: I intentionally moved the reference to [dcl.fct.def.delete] to the end of the sentence, so it more clearly attaches to its referent.

source/templates.tex
zygoloid: Done.

source/classes.tex
zygoloid: Done, and I added the missing definition for "constructor" too (which used the same phrasing as we have here).

PR #3044, opened 2019-07-22
 P1186R3 When do you actually use <=>?

### LINKED ISSUE #2985: [2019-07 CWG Motion 6] P1186R3 When do you actually use <=>?

zygoloid:
Fixes #2985

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -6907,12 +6907,6 @@ is \tcode{auto},
 then the return type is deduced as
 the common comparison type (see below) of
 $\tcode{R}_0$, $\tcode{R}_1$, $\dotsc$, $\tcode{R}_{n-1}$.
-\begin{note}
-Otherwise,
-the program will be ill-formed
-if the expression $\tcode{x}_i$ \tcode{<=>} $\tcode{x}_i$
-is not implicitly convertible to the declared return type for any $i$.
-\end{note}
 If the return type is deduced as \tcode{void},
 the operator function is defined as deleted.
 If the declared return type of
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -6832,9 +6832,75 @@ struct D {
 \indextext{operator!three-way comparison!defaulted}%
 
 \pnum
+The \defnadj{synthesized}{three-way comparison}
+for comparison category type \tcode{R}\iref{cmp.categories}
+of glvalues \tcode{a} and \tcode{b} of the same type
+is defined as follows:
+
+\begin{itemize}
+\item
+If overload resolution for \tcode{a <=> b}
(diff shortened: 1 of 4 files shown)


source/classes.tex
```diff
@@ -6785,10 +6785,76 @@
 \rSec2[class.spaceship]{Three-way comparison}
 \indextext{operator!three-way comparison!defaulted}%
 
+\pnum
+The synthesized three-way comparison
```
burblebee: Please add the index entry `\indextext{operator!three-way comparison!synthesized}`.  FWIW, I would like to see "synthesized three-way comparison" kept as a definition as suggested in the paper - I realize we don't tend to do that when the term is only used locally, but what happens when it starts getting used elsewhere?

source/classes.tex
```diff
@@ -6785,10 +6785,76 @@
...
+a == b ? partial_ordering::equivalent :
+a < b  ? partial_ordering::less :
+b < a  ? partial_ordering::greater :
+         partial_ordering::unordered
+\end{codeblock}
+
+\item
+Otherwise, if \tcode{R} is \tcode{strong_equality}, then
+\begin{codeblock}
+a == b ? strong_equality::equal : strong_equality::nonequal
```
burblebee: Suggestion: place `strong_equality::nonequal;` on new line for consistency with the previous items?

source/classes.tex
```diff
@@ -6785,10 +6785,76 @@
...
+\end{codeblock}
+
+\item
+Otherwise, if \tcode{R} is \tcode{weak_equality}, then
+\begin{codeblock}
+a == b ? weak_equality::equivalent : weak_equality::nonequivalent
```
burblebee: ditto?

source/classes.tex
```diff
@@ -6785,10 +6785,76 @@
...
+\pnum
+The synthesized three-way comparison
+for comparison category type \tcode{R}\iref{cmp.categories}
+of glvalues \tcode{a} and \tcode{b} of the same type
+is defined as follows:
+
+\begin{itemize}
+\item
+If overload resolution for \tcode{a <=> b}
+finds a usable function \iref{over.match},
```
burblebee: Nitpik: Please remove preceding space before `\iref` for consistency.

source/classes.tex
zygoloid: Tried it, it looked a bit weird. We often only flow things onto multiple lines if they don't fit on one line; I think this case is similar.

source/classes.tex
zygoloid: Done.

source/classes.tex
zygoloid: Done.

PR #3049, opened 2019-07-22
P1207R4 Movability of Single-pass Iterators

### LINKED ISSUE #3026: [2019-07 LWG Motion 22] P1207R4 Movability of Single-pass Iterators

burblebee:
Fixes #3026.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -1282,7 +1282,7 @@ nor is the type required to be \libconcept{EqualityComparable}.
 \begin{codeblock}
 template<class I>
   concept WeaklyIncrementable =
-    Semiregular<I> &&
+    DefaultConstructible<I> && Movable<I> &&
     requires(I i) {
       typename iter_difference_t<I>;
       requires SignedIntegral<iter_difference_t<I>>;
@@ -1378,6 +1378,12 @@ template<class I>
     } &&
     WeaklyIncrementable<I>;
 \end{codeblock}
+
+\pnum
+\begin{note}
+Unlike the \oldconcept{Iterator} requirements,
+the \libconcept{Iterator} concept does not require copyability.
+\end{note}
 
 \rSec3[iterator.concept.sentinel]{Concept \tcode{Sentinel}}
 
@@ -1548,7 +1554,8 @@ They should be single-pass algorithms.
 \rSec3[iterator.concept.forward]{Concept \tcode{ForwardIterator}}
 
 \pnum
-The \libconcept{ForwardIterator} concept adds equality comparison and
+The \libconcept{ForwardIterator} concept adds
+copyability, equality comparison, and
 the multi-pass guarantee, specified below.
 
 \indexlibrary{\idxcode{ForwardIterator}}%
@@ -3946,7 +3953,8 @@ namespace std {
     template<class U> constexpr move_iterator(const move_iterator<U>& u);
     template<class U> constexpr move_iterator& operator=(const move_iterator<U>& u);
 
-    constexpr iterator_type base() const;
+    constexpr iterator_type base() const &;
+    constexpr iterator_type base() &&;
     constexpr reference operator*() const;
 
     constexpr move_iterator& operator++();
@@ -4035,7 +4043,7 @@ constexpr explicit move_iterator(Iterator i);
 \begin{itemdescr}
 \pnum
 \effects Constructs a \tcode{move_iterator}, initializing
-\tcode{current} with \tcode{i}.
+\tcode{current} with \tcode{std::move(i)}.
 \end{itemdescr}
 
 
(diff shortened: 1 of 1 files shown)


burblebee:
To lwg: In change to [counted.iter.access], should "\returns" become "\effects Equivalent to:" to mimic change in P0896R4 (e85af9533)?

jwakely:
> To lwg: In change to [counted.iter.access], should "\returns" become "\effects Equivalent to:" to mimic change in P0896R4 (e85af95)?

I don't see any great need to do that here. Also I see that [is.sorted] p4 is wrong since that commit, it's missing a `return`.

jwakely:
To be clear: it's certainly not the case that all _Returns:_ elements should be replaced by "_Effects:_ Equivalent to:". They mean different things.

burblebee:
> These changes seem to be P1207R4 not P1638R1.

Yup.  I didn't know how to fix that without force-pushing, hopefully that was ok.

source/iterators.tex
```diff
@@ -4043,14 +4051,30 @@
 
 \indexlibrarymember{base}{move_iterator}%
 \begin{itemdecl}
-constexpr Iterator base() const;
+constexpr Iterator base() const &;
 \end{itemdecl}
 
 \begin{itemdescr}
+\pnum
+\constraints \tcode{Iterator} satisfies \tcode{CopyConstructible}.
```
jwakely: I think this should be `\libconcept{CopyConstructible}`.

source/iterators.tex
```diff
@@ -4043,14 +4051,30 @@
...
+
+\pnum
+\effects \tcode{Iterator} models \tcode{CopyConstructible}.
```
jwakely: Same here.

burblebee:
> ... I see that [is.sorted] p4 is wrong since that commit, it's missing a `return`.

Don't think we should address that as part of this commit - please open a new issue?

source/iterators.tex
```diff
@@ -4945,7 +4969,8 @@
       requires Assignable<I&, const I2&>
         constexpr counted_iterator& operator=(const counted_iterator<I2>& x);
 
-    constexpr I base() const;
+    constexpr I base() const & requires CopyConstructible<I>;;
```
jwakely: Two semi-colons at the end of this line.

jwakely:
> > ... I see that [is.sorted] p4 is wrong since that commit, it's missing a `return`.
> 
> Don't think we should address that as part of this commit - please open a new issue?

Sure, I'm fixing it separately.

source/iterators.tex
burblebee: Ok, changed to "the \oldconcept{CopyConstructible} requirements".

source/iterators.tex
burblebee: ditto.

source/iterators.tex
burblebee: thanks!

source/iterators.tex
```diff
@@ -4043,14 +4051,32 @@
...
 \indexlibrarymember{base}{move_iterator}%
 \begin{itemdecl}
-constexpr Iterator base() const;
+constexpr Iterator base() const &;
 \end{itemdecl}
 
 \begin{itemdescr}
+\pnum
+\constraints \tcode{Iterator} satisfies
+the \oldconcept{CopyConstructible} requirements.
```
jwakely: Should be `\libconcept` not `\oldconcept`, here and in the next paragraph.

jwakely: Use \oldconcept for "meets the _Cpp17Foo_ requirements" and use \libconcept for "models `Bar`" and "satisfies `Baz`".

CaseyCarter: More precisely, this should be `satisfies \libconcept{CopyConstructible}.` without "the" and "requirements" - we only use "the XXXX requirements" for the old named requirement sets, not with concepts.

source/iterators.tex
```diff
@@ -4043,14 +4051,32 @@
...
+
+\pnum
+\effects \tcode{Iterator} models
+the \oldconcept{CopyConstructible} requirements.
```
CaseyCarter: Ditto `models \libconcept{CopyConstructible}.`

CaseyCarter: Oh, and `\effects` should be `\expects` here.

source/iterators.tex
```diff
@@ -5088,14 +5115,24 @@
...
 \end{itemdescr}
 
+\indexlibrarymember{base}{counted_iterator}%
+\begin{itemdecl}
+constexpr I base() &&;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\returns \tcode{std::move(current)}.
```
CaseyCarter: Bah, the ingress paper had incorrect context for line 5123 so we didn't notice the "*Effects*: Equivalent to:  `return current;`" / "*Returns*: `std::move(current)`." inconsistency.  I don't suppose I could convince the Editors to reword this paragraph to "*Effects*: Equivalent to: `return std::move(current);`" for consistency to save the trouble of an LWG issue?

burblebee: I'm afraid that's a normative change, but it's clear that that was what was intended.  @zygoloid what do you think?

source/iterators.tex
burblebee: Yes, the oldconcept was wrong here (my bad), but thanks to both of you for explaining the difference in wording between the two - hope I'll remember it for next time!

Fixed.

source/iterators.tex
burblebee: thanks!

source/iterators.tex
burblebee: @CaseyCarter Please open a separate issue for this (might have to be a LWG issue); I'm not comfortable making that change as part of this motion.  Thanks!

PR #3050, opened 2019-07-22
[is.sorted] Add missing "return" and semi-colon

jwakely:
This was lost when changing Returns: to Effects: for P0896R4. The
paper included this change, but it was lost when applying it.


jwakely:
I've added the post-2019-07 milestone, because this is a regression from the current IS, so it would be good to fix for the CD.

PR #3053, opened 2019-07-22
P1650R0 Output std::chrono::days with 'd' suffix

### LINKED ISSUE #3023: [2019-07 LWG Motion 19] P1650R0 Output std::chrono::days with 'd' suffix

burblebee:
Fixes #3023.


johelegp:
Doesn't this kind of behaviour change gets an entry in [diff]? Or this is a defect, and thus gets no such entry?

PR #3056, opened 2019-07-23
P1523R1 Views and Size Types

### LINKED ISSUE #3034: [2019-07 LWG Motion 30] P1523R1 Views and Size Types

burblebee:
Fixes #3034.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -589,10 +589,10 @@ object\iref{customization.point.object}. The expression
   \end{itemize}
 
 \item
-  Otherwise, \tcode{(ranges::end(E) - ranges::begin(E))}
+  Otherwise, \tcode{\placeholdernc{make-unsigned-like}(ranges::end(E) - ranges::begin(E))}\iref{range.subrange}
   if it is a valid expression and
   the types \tcode{I} and \tcode{S} of \tcode{ranges::begin(E)} and
-  \tcode{ranges::end(E)} (respectively) model both
+  \tcode{ranges::end(E)} (respectively) model both\hfill\break %avoid overfull
   \tcode{SizedSentinel<S, I>}\iref{iterator.concept.sizedsentinel} and
   \tcode{Forward\-Iterator<I>}.
   However, \tcode{E} is evaluated only once.
@@ -1209,7 +1209,7 @@ iterator and a sentinel into a single object that models the
 \begin{codeblock}
 namespace std::ranges {
   template<class T>
-    concept @\placeholdernc{pair-like}@ =                                 // \expos
+    concept @\placeholdernc{pair-like}@ =                                    // \expos
       !is_reference_v<T> && requires(T t) {
         typename tuple_size<T>::type;   // ensures \tcode{tuple_size<T>} is complete
         requires DerivedFrom<tuple_size<T>, integral_constant<size_t, 2>>;
@@ -1220,7 +1220,7 @@ namespace std::ranges {
       };
 
   template<class T, class U, class V>
-    concept @\placeholdernc{pair-like-convertible-to}@ =                  // \expos
+    concept @\placeholdernc{pair-like-convertible-to}@ =                     // \expos
       !Range<T> && @\placeholder{pair-like}@<remove_reference_t<T>> &&
       requires(T&& t) {
         { get<0>(std::forward<T>(t)) } -> ConvertibleTo<U>;
@@ -1228,11 +1228,11 @@ namespace std::ranges {
       };
 
   template<class T, class U, class V>
-    concept @\placeholdernc{pair-like-convertible-from}@ =                // \expos
+    concept @\placeholdernc{pair-like-convertible-from}@ =                   // \expos
       !Range<T> && @\placeholdernc{pair-like}@<T> && Constructible<T, U, V>;
 
   template<class T>
-    concept @\placeholdernc{iterator-sentinel-pair}@ =                    // \expos
+    concept @\placeholdernc{iterator-sentinel-pair}@ =                       // \expos
       !Range<T> && @\placeholdernc{pair-like}@<T> &&
       Sentinel<tuple_element_t<1, T>, tuple_element_t<0, T>>;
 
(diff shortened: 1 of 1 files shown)


burblebee:
I used \hfull\break to work around an Overfull issue - what should I have used?

Also, shouldn't make-unsigned-like be introduced and explained somewhere?  It first appears out of the blue in [range.prim.size]p1.3 (same for make-unsigned-like-t).

jwakely:
They're introduced in P1522R1.

source/ranges.tex
```diff
@@ -522,10 +522,10 @@
   \end{itemize}
 
 \item
-  Otherwise, \tcode{(ranges::end(E) - ranges::begin(E))}
+  Otherwise, \tcode{\placeholder{make-unsigned-like}(ranges::end(E) - ranges::begin(E))}\iref{range.subrange}
```
jensmaurer: Should this be \placeholdernc instead?

source/ranges.tex
```diff
@@ -1151,18 +1151,18 @@
...
-    static constexpr bool StoreSize =                   // \expos
+    static constexpr bool StoreSize =                      // \expos
       K == subrange_kind::sized && !SizedSentinel<S, I>;
-    I begin_ = I();                                     // \expos
-    S end_ = S();                                       // \expos
-    iter_difference_t<I> size_ = 0;                     // \expos; present only
-                                                        // when \tcode{StoreSize} is \tcode{true}
+    I begin_ = I();                                        // \expos
+    S end_ = S();                                          // \expos
+    @\placeholder{make-unsigned-like-t}@(iter_difference_t<I>) size_ = 0;  // \expos; present only
```
jensmaurer: \placeholdernc ?

source/ranges.tex
burblebee: Should it?  I've seen both used within tcode contexts throughout the spec; it's not clear when to use which - can you/someone clarify?

johelegp: From what I've seen, one's used within `\tcode`s and the other within code blocks.

zygoloid: Use the nc form to turn off italic correction in places where it looks bad. I think we have some examples in the style guide.

zygoloid: Looks like we don't have anything in the style guide for this. We should :)

burblebee: So what should be used in this case?  And yes, an example in the style guide would be very helpful.

source/ranges.tex
burblebee: ... and what should be used in this case?

source/ranges.tex
jensmaurer: My personal rule-of-thumb is to use \placeholdernc if there's something immediately afterwards (no space), and use \placeholder if a space follows.

burblebee: ok, done.

source/ranges.tex
burblebee: changed also.

source/ranges.tex
```diff
@@ -1151,18 +1151,18 @@
...
+    static constexpr bool StoreSize =                      // \expos
       K == subrange_kind::sized && !SizedSentinel<S, I>;
-    I begin_ = I();                                     // \expos
-    S end_ = S();                                       // \expos
-    iter_difference_t<I> size_ = 0;                     // \expos; present only
-                                                        // when \tcode{StoreSize} is \tcode{true}
+    I begin_ = I();                                        // \expos
+    S end_ = S();                                          // \expos
+    @\placeholdernc{make-unsigned-like-t}@(iter_difference_t<I>) size_ = 0;  // \expos; present only
+                                                           // when \tcode{StoreSize} is \tcode{true}
```
jensmaurer: Can these go on a single line?

burblebee: done.

burblebee: @jensmaurer can you approve now?

burblebee:
> Most of your uses of \placeholder in this change are similar to the uses where you changed to \placeholdernc. Did you make up your mind what looks better? Then, please be consistent.

I changed all added \placeholder{}s with \placeholdernc{}s to make you happy, but the surrounding Latex is inconsistent in its use of \placeholder v. \placeholdernc.  We need to come up with some rules on which to use where, and do a global sweep to fix  it.  

Leave this unresolved so we'll know to create an issue for it.

zygoloid:
> > Most of your uses of \placeholder in this change are similar to the uses where you changed to \placeholdernc. Did you make up your mind what looks better? Then, please be consistent.
> 
> I changed all added \placeholder{}s with \placeholdernc{}s to make you happy, but the surrounding Latex is inconsistent in its use of \placeholder v. \placeholdernc. We need to come up with some rules on which to use where, and do a global sweep to fix it.

Please don't blindly change `\placeholder` to `\placeholdernc`. The latter is a surgical tool for fixing cases of bad kerning due to italics corrections, and should be used sparingly only where there is actual bad kerning. Usually the italics correction is what we want; the bounding box for italicized text is wider due to the right overhang, and it so we want to create a little extra space.

I've found that `\placeholdernc` improves the presentation immediately before a period or a comma (where there is no higher text for the italics to get uncomfortably close to). So let's take that as rough guidance for now: use `\placeholdernc` before a teletype period or comma, where there would otherwise be an unsightly amount of horizontal space.

zygoloid:
Looking through a few more examples, `\placeholdernc` before a left paren also seems more pleasant to my eye, but before a left angle bracket I prefer including the italic correction.

burblebee:
> > Most of your uses of \placeholder in this change are similar to the uses where you changed to \placeholdernc. Did you make up your mind what looks better? Then, please be consistent.
> 
> I changed all added \placeholder{}s with \placeholdernc{}s to make you happy, but the surrounding Latex is inconsistent in its use of \placeholder v. \placeholdernc. We need to come up with some rules on which to use where, and do a global sweep to fix it.
> 
> Leave this unresolved so we'll know to create an issue for it.

opened #3135

burblebee:
> Looking through a few more examples, `\placeholdernc` before a left paren also seems more pleasant to my eye, but before a left angle bracket I prefer including the italic correction.

All the \placeholder{}s added in this motion occur before left parens, so it appears that the changes from \placeholder -> \placeholdernc in the Fixups are correct.

source/ranges.tex
johelegp: > From what I've seen, one's used within `\tcode`s and the other within code blocks.

I was confused. The difference I actually noticed was that one had to be wrapped in `@`s and the other not when in `\tcode`s or not, not whether it had to be `\placeholdernc` or the other.

PR #3066, opened 2019-07-23
P1452R2 On the non-uniform semantics of return-type-requirements

### LINKED ISSUE #2998: [2019-07 CWG Motion 18 P1452R2 On the non-uniform semantics of return-type-requirements

jensmaurer:
Fixes #2998.


burblebee:
@zygoloid this motion is not complete - must be completed after LWG motion 12.

PR #3072, opened 2019-07-24
P0660R10 Stop Token and Joining Thread

### LINKED ISSUE #3032: [2019-07 LWG Motion 28] P0660R10 Stop Token and Joining Thread, Rev 10

jwakely:
Editorial changes:

- Capitalization in subclause headings.
- Changed "determination" to "evaluated" in [thread.condvarany.interruptwait] p9.

Fixes #3032

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -5923,11 +5923,12 @@ this will happen in an unspecified but finite amount of time.
 
 \pnum
 It is \impldef{whether the thread that executes \tcode{main} and the threads created
-by \tcode{std::thread} provide concurrent forward progress guarantees} whether the
+by \tcode{std::thread} or \tcode{std::jthread} provide concurrent forward progress guarantees} whether the
 implementation-created thread of execution that executes
 \tcode{main}\iref{basic.start.main} and the threads of execution created by
-\tcode{std::thread}\iref{thread.thread.class} provide concurrent forward progress
-guarantees.
+\tcode{std::thread}\iref{thread.thread.class}
+or \tcode{std::jthread}\iref{thread.jthread.class}
+provide concurrent forward progress guarantees.
 \begin{note}
 General-purpose implementations should provide these guarantees.
 \end{note}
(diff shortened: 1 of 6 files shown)


source/basic.tex
```diff
@@ -5856,11 +5856,12 @@
...
 \pnum
 It is \impldef{whether the thread that executes \tcode{main} and the threads created
-by \tcode{std::thread} provide concurrent forward progress guarantees} whether the
+by \tcode{std::thread} or \tcode{std::jthread} provide concurrent forward progress guarantees} whether the
 implementation-created thread of execution that executes
 \tcode{main}\iref{basic.start.main} and the threads of execution created by
-\tcode{std::thread}\iref{thread.thread.class} provide concurrent forward progress
-guarantees.
+\tcode{std::thread}\iref{thread.thread.class}
+or \tcode{std::jthread}\iref{thread.jthread.class}
```
jwakely: Should this be "**and** `std::jthread`", not "or `std::jthread`"?

The rule applies to threads created by `std::thread` and threads created by `std::jthread`, not one or the other.

If "and" is used, it should be changed in both places in `support.tex` and also `algorithms.tex` which mirrors this wording and cross-references it.

source/threads.tex
```diff
@@ -482,175 +1164,515 @@
...
+
+\rSec2[thread.jthread.class]{Class \tcode{jthread}}
+
+\pnum
+The class \tcode{jthread} provides a mechanism
+to create a new thread of execution.
+The functionality is the same as for
+class \tcode{thread}\iref{thread.thread.class}
+with the additional ability to request that the thread stops
+and then joins the started thread.
```
jwakely: On the edit@ reflector I suggested changing this to:

> The class `jthread` provides a mechanism to create a new thread of execution. The functionality is the same as for class `thread` (32.4.2) with the additional abilities to provide a `stop_token` (32.3) to the new thread of execution, make stop requests, and automatically join.

source/basic.tex
burblebee: @jwakely As I read it, it is saying exactly what you want, that is, the rule applies to any thread created by either std::thread or std::jthread.  If this were changed to "and", it would only apply to threads created by both std::thread and std::jthread (i.e. the intersection of the threads created by each), which I presume can't happen.  Right?  If this is still wrong, please file an issue with LWG.  Thanks!

source/threads.tex
burblebee: @jwakely Is this still an issue?  If so, can you please create an editorial/lwg issue for it?

PR #3073, opened 2019-07-24
[algorithms] Split list items conventionally

johelegp:
All other _Returns:_ elements with the same form use "or".

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -2919,8 +2919,14 @@ template<input_range R, class Proj = identity,
 
 \begin{itemdescr}
 \pnum
-Let $E$ be \tcode{pred(*i)} and \tcode{invoke(pred, invoke(proj, *i))}
-for the overloads in namespace \tcode{std} and \tcode{std::ranges}, respectively.
+Let $E$ be:
+\begin{itemize}
+\item
+  \tcode{pred(*i)} for the overloads in namespace \tcode{std};
+\item
+  \tcode{invoke(pred, invoke(proj, *i))}
+  for the overloads in namespace \tcode{ranges}.
+\end{itemize}
 
 \pnum
 \returns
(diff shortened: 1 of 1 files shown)


CaseyCarter:
FWIW, most opinions are that the "or"s (which I added) are incorrect and should be changed to "and"s. I'm not an editor - so I can't click "merge" in any case - but I'd prefer to see a PR with the opposite change.

johelegp:
Me too. I noticed a motion PR using "or" and checked if it should be "and" instead. The majority's preference would have to wait until after its merge.

jensmaurer:
@zygoloid, this probably needs your opinion.

zygoloid:
Maybe take neither option, and replace ", and" with a semicolon and no conjunction.

jensmaurer:
Editorial meeting: [alg.all.of] p1 construction is no good. The "or" is definitely wrong (who gets to choose?); some sympathy to avoid the "and". Maybe have a period for "returns", semicolon in the other cases. No "or" or "and".

johelegp:
I'm currently preparing this PR to apply those changes. What do you recommend doing in light of [ P1718](wg21.link/P1718 "Mandating the Standard Library: Clause 25 - Algorithms library ")?

jensmaurer:
> I'm currently preparing this PR to apply those changes.

Thank you.

> What do you recommend doing in light of [ P1718](wg21.link/P1718)?

What do you mean? The paper will go forward as-is. The editors likely need to do the merge by hand. So be it.

johelegp:
IIRC, @mclow provides you the sources to ease the merge. With this PR, you'd definitely need to merge by hand.

source/algorithms.tex
```diff
@@ -8088,9 +8132,11 @@
 For the first form, type \tcode{T} shall be \oldconcept{LessThanComparable}.
 
 \pnum
-\returns
 Let \tcode{X} be the return type.
-Returns \tcode{X\{x, y\}},
+
+\pnum
+\returns
+\tcode{X\{x, y\}},
```
johelegp: Note the missing braces in the code.
Before:

After:

johelegp:
Questions:
- Should [alg.adjacent.find]'s _Complexity:_ be made into a list with one item per "overloads", like other lists touched in this PR, like in [alg.equal]'s _Complexity:_? Maybe the converse?
- Should [mismatch]'s 3 **Let**'s be joined into one **Let:** with 3 items like in [alg.find.end], [alg.equal], [alg.transform]? How about [alg.copy]p14, [alg.move]p11, [alg.remove]p8-9 (there's inconsistency in the ammount of \pnums), [alg.unique] p1 and p6? Maybe the converse?

johelegp:
Limiting the changes to what was approved in the editorial meeting and strictly grammar and whitespace fixes. Expect me bring up the other changes when this is merged.

Issue #3016, opened 2019-07-21
[2019-07 LWG Motion 12] P1614R2 The Mothership has Landed


jensmaurer:

jensmaurer:
@Dani-Hub notes:

The newly added [cmp.concept] says:

> "partially-ordered-with<T, U> is satisfied only if:" [bullet list of semantic requirements]

The usage of "satisfied" is inappropriate here and should be

> "T and U model partially-ordered-with<T, U> only if:" [bullet list of semantic requirements]

jensmaurer:
From P1452R2 On the non-uniform semantics of return-type-requirements:

"[If] P1614, “The Mothership Has Landed: Adding <=> to the Library”), replace said production with -> ConvertibleTo<Type> where Type is the type named by the trailing-return-type:"

Dani-Hub:
I couldn't find the resolution expressed in the second comment of this issue referring to [cmp.concept]'s notion of "partially-ordered-with<T, U> is satisfied only if:". Could you please help me understanding the decided resolution?

johelegp:
It is here:
https://github.com/cplusplus/draft/blob/4060676e3989167a4e83c08ae0550509b574f1db/source/support.tex#L4737-L4738
The resolution was to fix the wording to follow the convention.

Dani-Hub:
Thanks for the pointer - looks good!

PR #3077, opened 2019-07-25
P0980R1 Making std::string constexpr

### LINKED ISSUE #3031: [2019-07 LWG Motion 27] P0980R1 Making std::string constexpr

burblebee:
[string.find]: Added \indexlibrarymember references to the variations of find* described via the placeholder F.

Fixes #3031.

### DIFF
--- a/source/strings.tex
+++ b/source/strings.tex
@@ -539,73 +539,75 @@ namespace std {
     class basic_string;
 
   template<class charT, class traits, class Allocator>
-    basic_string<charT, traits, Allocator>
+    constexpr basic_string<charT, traits, Allocator>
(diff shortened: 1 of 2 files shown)


source/strings.tex
```diff
@@ -1672,7 +1703,7 @@
 
 \indexlibrarymember{empty}{basic_string}%
 \begin{itemdecl}
-[[nodiscard]] bool empty() const noexcept;
+[[nodiscard]] bool constexpr empty() const noexcept;
```
jwakely: Should be `constexpr bool` not `bool constexpr`

jwakely:
P1614R2 (#3016) adds new non-member functions which should also be marked `constexpr`.

PR #3078, opened 2019-07-25
P1004R2 Making std::vector constexpr

### LINKED ISSUE #3030: [2019-07 LWG Motion 26] P1004R2 Making std::vector constexpr

burblebee:
[vector.special] Could not apply the change to this section;
the section no longer exists, and its itemdecl for swap has been removed.

Fixes #3030.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -3212,19 +3212,19 @@ namespace std {
   template<class T, class Allocator = allocator<T>> class vector;
 
   template<class T, class Allocator>
-    bool operator==(const vector<T, Allocator>& x, const vector<T, Allocator>& y);
+    constexpr bool operator==(const vector<T, Allocator>& x, const vector<T, Allocator>& y);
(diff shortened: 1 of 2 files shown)


jwakely:
[vector.special] was removed by 8738c6b38db45d06724e4a75afd2134d3f6e6f3f because [tab:container.req] specifies `swap(a, b)` for all containers, so re-specifying it for each container was redundant.

jwakely:
P1614R2 (#3016) adds new non-member functions which should also be marked constexpr.

burblebee:
> P1614R2 (#3016) adds new non-member functions which should also be marked constexpr.

@jwakely How should we deal with that?

jwakely:
I suggest adding a commit to #3098 to add `constexpr` to the relevant functions (that pull request didn't exist when I made the earlier comment, that's why I commented here).

PR #3086, opened 2019-07-26
P1766R1 Mitigating minor modules maladies

### LINKED ISSUE #2992: [2019-07 CWG Motion 13] P1766R1 Mitigating minor modules maladies

jensmaurer:
Fixes #2992.

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -1922,6 +1922,34 @@ that do not result in an odr-use.
 
 \rSec2[diff.cpp17.dcl.dcl]{\ref{dcl.dcl}: declarations}
 
+\diffref{dcl.typedef}
+\change Unnamed classes with a typedef name for linkage purposes
+can only contain C-compatible constructs.
+\rationale Necessary for implementability.
+\effect Valid C++ 2017 code may be ill-formed in this International Standard.
+\begin{codeblock}
+typedef struct {
+  void f() {}           // ill-formed; previously well-formed
+} S;
+\end{codeblock}
+
+\diffref{dcl.fct.default}
+\change A function cannot have different default arguments
+in different translation units.
+\rationale Required for modules support.
+\effect Valid C++ 2017 code may be ill-formed in this International Standard,
+with no diagnostic required.
+\begin{codeblock}
+// Translation unit 1
+int f(int a = 42);
+int g() { return f(); }
+
+// Translation unit 2
+int f(int a = 76) { return a; }         // ill-formed (no diagnostic required); previously well-formed
+int g();
+int main() { return g(); }              // used to return 42
+\end{codeblock}
+
 \diffref{dcl.init.aggr}
 \change A class that has user-declared constructors is never an aggregate.
 \rationale Remove potentially error-prone aggregate initialization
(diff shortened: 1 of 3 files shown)


source/declarations.tex
```diff
@@ -735,6 +735,23 @@
 typedef decltype([]{}) C;       // the closure type has no name for linkage purposes
 \end{codeblock}
 \end{example}
```
burblebee: I think if you add a newline after this example, the next sentence will start out on a new line and look better.

source/declarations.tex
```diff
@@ -735,6 +735,23 @@
...
+An unnamed class with a typedef name for linkage purposes shall not
+\begin{itemize}
+\item
+  declare any members
+  other than non-static data members, member enumerations, or member classes,
+\item
+  have any base classes or default member initializers, or
+  contain a lambda-expression,
```
burblebee: Missing `\item` before "contain a lambda-expression" and, lambda-expression should be \grammarterm'ed.

source/compatibility.tex
```diff
@@ -1922,6 +1922,34 @@
...
 \rSec2[diff.cpp17.dcl.dcl]{\ref{dcl.dcl}: declarations}
 
+\diffref{dcl.typedef}
+\change Unnamed classes with a typedef name for linkage purposes
+can only contain C-compatible constructs.
+\rationale Necessary for implementability.
+\effect Valid C++ 2017 code may be ill-formed in this International Standard.
+\begin{codeblock}
+typedef struct {
+  void f() {}           // ill-formed; previously well-formed
```
burblebee: FYI - our coding style guidelines say that the starting columns for comments should be multiples of 4, but other examples' comments in [diff.cpp17.dcl.dcl] start on column 25 so I wouldn't change this without changing the others as well.

source/compatibility.tex
```diff
@@ -1922,6 +1922,34 @@
...
+with no diagnostic required.
+\begin{codeblock}
+// Translation unit 1
+int f(int a = 42);
+int g() { return f(); }
+
+// Translation unit 2
+int f(int a = 76) { return a; } // ill-formed (no diagnostic required); previously well-formed
+int g();
+int main() { return g(); }      // used to return 42
```
burblebee: ... however, these comments should be moved to multiples of 4 to follow our  guidelines.

source/declarations.tex

source/declarations.tex

source/compatibility.tex
jensmaurer: We seem to be measuring differently. This is column 24 in my book.

source/compatibility.tex
jensmaurer: These are at col 32. I've moved them to col 40 to satisfy the "at least two spaces" rule.

source/compatibility.tex
burblebee: Eek!  You're right!  Vim has been lying to me!  Which means all of *my* comments are off!!!
```
  void f() {}           // ill-formed; previously well-formed
0         1         2
0123456780912345678091234
...
compatibility.tex [+]                                         1932,25        77%
```
The "1932,25" above is what vim thinks is the line and column.

jensmaurer: I'd guess vim starts counting at 1, but everybody else starts counting at 0. I'm not sure who's right. What would a traditional typewriter do if you hit the "tab" key?

burblebee: Opened #3131 for this.

PR #3087, opened 2019-07-26
P1811R0 Relaxing redefinition restrictions for re-exportation robustness

### LINKED ISSUE #2993: [2019-07 CWG Motion 14] P1811R0 Relaxing redefinition restrictions for re-exportation robustness

jensmaurer:
Fixes #2993.

### DIFF
--- a/source/preprocessor.tex
+++ b/source/preprocessor.tex
@@ -624,7 +624,10 @@ up to an \impldef{nesting limit for \tcode{\#include} directives} nesting limit.
 \pnum
 If the header identified by the \grammarterm{header-name}
 denotes an importable header\iref{module.import},
-the preprocessing directive
+it is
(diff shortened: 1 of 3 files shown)


source/preprocessor.tex
```diff
@@ -621,6 +621,7 @@
 \pnum
 If the header identified by the \grammarterm{header-name}
 denotes an importable header\iref{module.import},
+it is implementation-defined whether
```
burblebee: `implementation-defined` should be replaced with an `\impldef`.  Perhaps `\impldef{replacing preprocessing directives with \grammarterm{preprocessing-token}{s} for an importable header}`?

jensmaurer: Good catch.

jensmaurer: Fixed.

PR #3089, opened 2019-07-26
[temp.deduct] Fix Ambiguous "this" in paragraph 3

hawkinsw:
This is an editorial change. There is an ambiguous "this"
in paragraph 3 of [temp.deduct]. Upon first reading this paragraph
it is not obvious that "this substitution" refers to the
substitution process described in the subsequent paragraphs.
This change specifies that the substitution process refers
to the one described in the subsequent paragraphs of the section.


hawkinsw:
I am abandoning this PR for something a little bit more comprehensive. Sorry for the noise!

PR #3092, opened 2019-07-27
P1466R3 Miscellaneous minor fixes for chrono

### LINKED ISSUE #3014: [2019-07 LWG Motion 10] P1466R3 Miscellaneous minor fixes for chrono

jensmaurer:
- Also adjusted the class synopsis in [time.cal.wd.overview].
 - Replaced "else" with "otherwise".

Fixes #3014.

### DIFF
--- a/source/time.tex
+++ b/source/time.tex
@@ -8037,12 +8037,12 @@ operator<<(basic_ostream<charT, traits>& os, hh_mm_ss<Duration> const& hms);
 \begin{itemdescr}
 \pnum
 \effects
-Outputs to \tcode{os} according to the format \tcode{"\%T"}\iref{time.format}.
-If \tcode{hms.is_negative()}, the output is preceded with \tcode{'-'}.
-
-\pnum
-\returns
-\tcode{os}.
+Equivalent to:
+\begin{codeblock}
+if (hms.is_negative())
+  os << @\placeholder{STATICALLY-WIDEN}@<charT>("-");
+return os << format(os.getloc(), @\placeholder{STATICALLY-WIDEN}@<charT>("{:%T}"), hms);
+\end{codeblock}
 
 \pnum
 \begin{example}
--- a/source/time.tex
+++ b/source/time.tex
@@ -747,8 +747,7 @@ namespace std {
 
     template<class charT, class traits, class Duration>
       basic_ostream<charT, traits>&
-        operator<<(basic_ostream<charT, traits>& os,
-                   const hh_mm_ss<Duration>& t);
+        operator<<(basic_ostream<charT, traits>& os, const hh_mm_ss<Duration>& t);
 
     // \ref{time.12}, 12/24 hour functions
     constexpr bool is_am(const hours& h) noexcept;
@@ -7797,15 +7796,7 @@ constexpr year_month_weekday_last
 
 \begin{codeblock}
 namespace std::chrono {
-  template <class Duration>
-  class hh_mm_ss
-  {
-    bool            is_neg;     // \expos
-    chrono::hours   h;          // \expos
-    chrono::minutes m;          // \expos
-    chrono::seconds s;          // \expos
-    precision       ss;         // \expos
-
+  template <class Duration> class hh_mm_ss {
   public:
     static unsigned constexpr fractional_width = @\seebelow@;
     using precision                            = @\seebelow@;
@@ -7821,11 +7812,14 @@ namespace std::chrono {
 
     constexpr explicit operator precision() const noexcept;
     constexpr precision to_duration() const noexcept;
-  };
 
-  template <class charT, class traits, class Duration>
-    basic_ostream<charT, traits>&
-      operator<<(basic_ostream<charT, traits>& os, hh_mm_ss<Duration> const& hms);
+  private:
+    bool            is_neg;     // \expos
+    chrono::hours   h;          // \expos
+    chrono::minutes m;          // \expos
+    chrono::seconds s;          // \expos
+    precision       ss;         // \expos
+  };
 }
 \end{codeblock}
 
@@ -7942,8 +7936,9 @@ which represents the \tcode{Duration d} with precision \tcode{precision}.
   with \tcode{duration_cast<precision>(abs(d) - hours() - minutes() - seconds())}.
 \end{itemize}
 \begin{note}
-When \tcode{precision} is \tcode{seconds} with integral representation,
-\tcode{subseconds()} always returns \tcode{0s}.
+When \tcode{precision::rep} is integral and
+\tcode{precision::period} is \tcode{ratio<1>},
+\tcode{subseconds()} always returns a value equal to \tcode{0s}.
 \end{note}
 
 \pnum
--- a/source/time.tex
+++ b/source/time.tex
@@ -16,7 +16,8 @@ utilities, as summarized in \tref{time.summary}.
 \ref{time.point}            & Class template \tcode{time_point}  & \\
 \ref{time.clock}            & Clocks                             & \\
 \ref{time.cal}              & Civil calendar                     & \\
-\ref{time.tod}              & Class template \tcode{time_of_day} & \\
+\ref{time.hms}              & Class template \tcode{hh_mm_ss}    & \\
+\ref{time.12}               & 12/24 hour functions               & \\
 \ref{time.zone}             & Time zones                         & \\
 \ref{time.format}           & Formatting                         & \\
 \ref{time.parse}            & Parsing                            & \\ \rowsep
@@ -263,6 +264,11 @@ namespace std {
                     utc_time<Duration>& tp,
                     basic_string<charT, traits, Alloc>* abbrev = nullptr,
                     minutes* offset = nullptr);
+
+    struct leap_second_info;
+
+    template <class Duration>
+      leap_second_info get_leap_second_info(utc_time<Duration> const& ut);
 
     // \ref{time.clock.tai}, class \tcode{tai_clock}
     class tai_clock;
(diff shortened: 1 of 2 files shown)


source/time.tex
```diff
@@ -8039,713 +8077,340 @@
...
-\indexlibrarymember{hours}{time_of_day<hours>}%
-\begin{itemdecl}
-constexpr chrono::hours hours() const noexcept;
-\end{itemdecl}
+The \tcode{hh_mm_ss} class template
+splits a \tcode{duration}
+into a multi-field time structure
+\placeholdernc{hours}:\placeholdernc{minutes}:\placeholder{seconds} and
+possibly \placeholder{subseconds},
+where \placeholder{subseconds} will be a duration uni
```
johelegp: s/uni/unit/

source/time.tex
```diff
@@ -8039,713 +8077,340 @@
...
+  with \tcode{duration_cast<chrono::seconds>(abs(d) - hours() - minutes())}.
+\item
+  If \tcode{treat_as_floating_point_v<precision::rep>} is \tcode{true},
+  initializes \tcode{ss} with \tcode{abs(d) - hours() - minutes() - seconds()}.
+  Otherwise, initializes \tcode{ss}
+  with \tcode{duration_cast<precision>(abs(d) - hours() - minutes() - seconds())}.
+\end{itemize}
 \begin{note}
-This specialization handles hours and minutes since midnight.
+When \tcode{precision} is \tcode{seconds} with integral representation,
```
johelegp: I think this note shouldn't use `seconds` in code font, since its referring to the concept of seconds rather than the specific implementation of `seconds` which already has an integral representation. I suggest replacing "is `seconds`" with "represents seconds", or to be more precise (perhaps unnecessarily for a note), "has a tick period of a second".

jensmaurer: "precision" is a typedef, and "seconds" is some chrono type, so a C++-level entity. So this part seems correct. Maybe the "integral representation" is not a restriction, but falls out from the rest, and should thus be phrased differently?

johelegp: Considering the nature of notes, it seems better to just do away with "with integral representation". The statement would still hold true when `precision` is `duration<int128_t>`, which also represents "`seconds`", but is not `seconds`.

jwakely: > Maybe the "integral representation" is not a restriction, but falls out from the rest, and should thus be phrased differently?

The statement is only true when `precision::rep` is an integral type, not for floating-point types. But it's also true for types other than `chrono::seconds`, e.g. `chrono::minutes` and `chrono::hours` and `chrono::duration<short, ratio<60>>`.

> Considering the nature of notes, it seems better to just do away with "with integral representation". The statement would still hold true when precision is `duration<int128_t>`, which also represents "seconds", but is not `seconds`.

But it would not hold true for `duration<double>` which represents seconds, but can have subsecond precision.

A more accurate (but less clear) wording would be "when `treat_as_floating_point_v<precision::rep>` is `false` and `is_same_v<precision::period, ratio<1>>` is `true` ...".

@HowardHinnant ?

jwakely: Or instead of using `is_same`, use `ratio_equal_v<precision::period, ration<1>>`.

But arguably the verbosity makes the note less useful.

source/time.tex
```diff
@@ -8039,713 +8077,340 @@
...
-whether it should be output
-as a 12-hour time format or a 24-hour time format.
+  template <class Duration>
+  class hh_mm_ss
+  {
+    bool            is_neg;     // \expos
+    chrono::hours   h;          // \expos
+    chrono::minutes m;          // \expos
+    chrono::seconds s;          // \expos
+    precision       ss;         // \expos
```
jwakely: This exposition-only member uses the typedef `precision` before it has been defined. Can the exposition-only members be moved to the end of the class?

source/time.tex
HowardHinnant: Jonathan has a complete understanding of this issue.

@johelegp 's suggestion above including "has a tick period of a second" seemed pretty on-the-money to me.  The _non-normative_  note applies when `rep` is integral and `period` is `ratio<1>`.

HowardHinnant: Also, technically speaking, `0s` isn't always returned, but instead `precision{0}`, the difference being that `0s` has type `std::chrono::seconds`.

I don't have strong feelings about how this is resolved.  The note doesn't specify anything that isn't specified elsewhere and is meant to be informative on a point that otherwise might be confusing.  Thus the more relaxed language.  The less confusing the better, whatever the editors deem less confusing.

source/time.tex
HowardHinnant: Or you could copy `public` and move to the top:
```
public:
    static unsigned constexpr fractional_width = _see below_;
    using precision                                              = _see below_;
```

source/time.tex
HowardHinnant: Here's a shot at it:

```
\begin{note}
When \tcode{precision::rep} is integral and \tcode{precision::period} is \tcode{ratio<1>},
\tcode{subseconds()} always returns a value equal to \tcode{0s}.
\end{note}
```

source/time.tex
HowardHinnant: Fwiw, the reason I have come to appreciate the "data members at the top" style is that it informs the reader about the specification/behavior of the immediately following but invisible compiler-supplied special members:
```
~hh_mm_ss() = default;
hh_mm_ss(hh_mm_ss const&) = default;
hh_mm_ss& operator=(hh_mm_ss const&) = default;
hh_mm_ss(hh_mm_ss&&) = default;
hh_mm_ss& operator=(hh_mm_ss&&) = default;
```

source/time.tex
```diff
@@ -8039,713 +8077,340 @@
...
-
-    constexpr void make24() noexcept;
-    constexpr void make12() noexcept;
+    constexpr explicit operator precision() const noexcept;
+    constexpr precision to_duration() const noexcept;
   };
+
+  template <class charT, class traits, class Duration>
+    basic_ostream<charT, traits>&
+      operator<<(basic_ostream<charT, traits>& os, hh_mm_ss<Duration> const& hms);
```
johelegp: Repetition of this non-member function already specified in the header synopsis should be removed.

source/time.tex
```diff
@@ -8039,713 +8077,340 @@
...
-\pnum
-\ensures
-\tcode{hours()} returns the integral number of hours \tcode{since_midnight} is after 00:00:00.
-\end{itemdescr}
-
-\indexlibrarymember{hours}{time_of_day<hours>}%
-\begin{itemdecl}
-constexpr chrono::hours hours() const noexcept;
-\end{itemdecl}
+The \tcode{hh_mm_ss} class template
```
johelegp: All this sounds like description and non-normative wording, and thus should be moved before the class synopsis.

source/time.tex

source/time.tex

source/time.tex
jensmaurer: Having the description after the class synopsis is locally consistent.

I won't change this.

source/time.tex
jensmaurer: "Immediately following"? I don't follow. These implicit special members are produced at the closing brace of the class definition, not before.

jensmaurer: I've moved the exposition-only members to the bottom.

source/time.tex
```diff
@@ -8039,713 +8076,338 @@
...
+  with \tcode{duration_cast<chrono::seconds>(abs(d) - hours() - minutes())}.
+\item
+  If \tcode{treat_as_floating_point_v<precision::rep>} is \tcode{true},
+  initializes \tcode{ss} with \tcode{abs(d) - hours() - minutes() - seconds()}.
+  Otherwise, initializes \tcode{ss}
+  with \tcode{duration_cast<precision>(abs(d) - hours() - minutes() - seconds())}.
+\end{itemize}
 \begin{note}
-This specialization handles hours and minutes since midnight.
+When \tcode{precision::rep} is integral an
```
```suggestion
When \tcode{precision::rep} is integral and
```

source/time.tex
```diff
@@ -7752,713 +7789,339 @@
...
-
-    constexpr explicit operator  precision()   const noexcept;
-    constexpr          precision to_duration() const noexcept;
-
-    constexpr void make24() noexcept;
-    constexpr void make12() noexcept;
-  };
-}
+if (hms.is_negative())
+  os << @\placeholder{STATICALLY-WIDEN}@<charT>("-");
```
timsong-cpp: This needs to be part of the format string; otherwise we'd be applying the the stream width/fill to this single character instead of the whole output.

jensmaurer: Agreed, but this was already merged (and a fix doesn't seem editorial to me). @timsong-cpp, please post to the LWG reflector and ask for a new library issue.

timsong-cpp: Well, the paper itself doesn't use this specification (it's not fmt-based), and I think you can at least arguably read the original spec as doing the right thing.

The translation to fmt that made the spec unambiguously incorrect is the result of an editorial change (1a37c22).

jensmaurer: @zygoloid, there is a problem here.

zygoloid: Hmm. The original wording was ambiguous in this regard. Does "the output is preceded by" mean a single stream operation or two?

"Outputs to `os` according to the format `%T`" doesn't mean anything any more after the `std::format` for \<chrono> changes, so reverting this is also wrong.

I was planning to mail LWG for feedback on this change anyway.

zygoloid: Fixed in 584a87ec1d48862b9e68a269d0a5eb7b05d6999d after consultation with LWG.

PR #3106, opened 2019-07-30
P1035R7 Input range adaptors

jensmaurer:
Fixes #3027.

PR #3107, opened 2019-07-30
P1638R1 basic_istream_view::iterator should not be copyable

### LINKED ISSUE #3028: [2019-07 LWG Motion 24] P1638R1 basic_istream_view::iterator should not be copyable

jensmaurer:
Fixes #3028.

I wanted this pull request to just contain the patch against P1035R7, but that apparently failed.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -5560,6 +5560,12 @@ namespace std::ranges {
     iterator() = default;
     constexpr explicit iterator(basic_istream_view& parent) noexcept;
 
+    iterator(const iterator&) = delete;
+    iterator(iterator&&) = default;
+
+    iterator& operator=(const iterator&) = delete;
+    iterator& operator=(iterator&&) = default;
+
     iterator& operator++();
     void operator++(int);
 


johelegp:
That's because you want to merge into `master`, which doesn't have P1035R7.

source/ranges.tex
```diff
@@ -3802,1398 +3855,2476 @@
...
+\begin{itemdecl}
+constexpr explicit inner_iterator(outer_iterator<Const> i);
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\effects Initializes \tcode{i_} with \tcode{i}.
+\end{itemdescr}
+
+\indexlibr
```
johelegp: Remove space between `iterator` and `&`.
```suggestion
    iterator& operator=(const iterator&) = delete;
```

johelegp: Same as above.
```suggestion
    iterator& operator=(iterator&&) = default;
```

jensmaurer: Fixed.

jensmaurer: Fixed.

CaseyCarter:
```suggestion
    iterator& operator=(const iterator&) = delete;
```

CaseyCarter:
```suggestion
    iterator& operator=(iterator&&) = default;
```

jensmaurer: Fixed.

burblebee:
@CaseyCarter can you approve this?

burblebee:
> The changes this PR applies from P1638R1 - the last three commits - are correct, but the PR also includes P1035 changes and **should not be merged without rebasing onto the result of merging #3106.**
> 
> @burblebee, Could we label this rebase-required to ensure no one merges it without performing that rebase?

@CaseyCarter, @zygoloid will do this when he merges this motion.

johelegp:
Or do as @zygoloid did with Text Formatting, and add it to the same PR.

PR #3109, opened 2019-07-31
P0408R7 Efficient Access to basic_stringbuf ’s Buffer

### LINKED ISSUE #3018: [2019-07 LWG Motion 14] P0408R7 Efficient Access to basic_stringbuf ’s Buffer

burblebee:
Fixes #3018.

Notes:
* Wording says:
    "Add a conditional noexcept specification to swap with see below:"
    but "see below" is not used in the wording for this swap member.

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -7379,7 +7379,7 @@ namespace std {
     basic_stringbuf() : basic_stringbuf(ios_base::in | ios_base::out) {}
     explicit basic_stringbuf(ios_base::openmode which);
     explicit basic_stringbuf(
-      const basic_string<charT, traits, Allocator>& str,
+      const basic_string<charT, traits, Allocator>& s,
       ios_base::openmode which = ios_base::in | ios_base::out);
     explicit basic_stringbuf(const Allocator& a)
       : basic_stringbuf(ios_base::in | ios_base::out, a) {}
@@ -8170,7 +8170,7 @@ namespace std {
     basic_istringstream() : basic_istringstream(ios_base::in) {}
     explicit basic_istringstream(ios_base::openmode which);
     explicit basic_istringstream(
-      const basic_string<charT, traits, Allocator>& str,
+      const basic_string<charT, traits, Allocator>& s,
       ios_base::openmode which = ios_base::in);
     basic_istringstream(ios_base::openmode which, const Allocator& a);
     explicit basic_istringstream(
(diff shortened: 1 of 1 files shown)


source/iostreams.tex
```diff
@@ -7566,32 +7689,149 @@
...
 
+\pnum
+The member functions getting the underlying character sequence
+all refer to a \tcode{high_mark} value,
+where \tcode{high_mark} represents the position
+one past the highest initialized character in the buffer.
+Characters can be initialized by writing to the stream,
+by constructing the \tcode{basic_stringbuf}
+passing a \tcode{basic_string} argument, or
+by calling one of the \tcode{str()} member functions
```
jensmaurer: Regarding str(): We don't want the parens unless we're actually talking about an expression. Here, we're just referring to a member function called "str", so no parens, please.

source/iostreams.tex
```diff
@@ -7566,32 +7689,149 @@
...
+  \end{itemize}
+\item If \tcode{ios_base::in} is set in \tcode{mode},
+  \tcode{eback()} points to \tcode{buf.front()}, and
+  \tcode{(gptr() == eback() \&\& egptr() == eback() + buf.size())}
+  is \tcode{true}.
+\end{itemize}
+
+\pnum
+\begin{note}
+For efficiency reasons
```
jensmaurer: Comma at the end of the line.

source/iostreams.tex
```diff
@@ -7566,32 +7689,149 @@
...
+stream buffer operations might violate invariants of \tcode{buf}
+while it is held encapsulated in the \tcode{basic_stringbuf},
+i.e., by writing to characters in the range
```
jensmaurer: i.e. -> e.g. (this is certainly not the only case where we could conceivably violate invariants of "buf")

source/iostreams.tex
jwakely: I've been trying to make this point in LWG but some people think the parents help them know it's a function :confounded:

Should I add it to the specification style guide?

jensmaurer: We already have it:
https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#describing-function-calls

burblebee: Totally: `str()` is wrong - it implies the `str` member function that takes no parameters, and there can only be one of those in this context :).  `str` is clarified by "member functions" in this case, so there should be no confusion.

Fixed.

source/iostreams.tex
burblebee: fixed

source/iostreams.tex
burblebee: done

burblebee:
@jensmaurer @jwakely Can one of you re-review and approve please?

jensmaurer:
There's a button next to the name under "Reviewers" to request a re-review.

Issue #2999, opened 2019-07-21
[2019-07 CWG Motion 19] P1152R4 Deprecating volatile


jensmaurer:

jfbastien:
I signed up to send an editorial PR with Annex D examples illustrating each of the Core features deprecated with the paper. I'll do this after the change is merged.

zygoloid:
@jfbastien The motion is merged; you're up.

PR #3123, opened 2019-08-05
Add examples to [depr.volatile]

jfbastien:
As promised in #2999.

### DIFF
--- a/source/future.tex
+++ b/source/future.tex
@@ -115,7 +115,7 @@ are deprecated; see~\ref{expr.ass}.
 \begin{codeblock}
 int neck, tail;
 volatile int brachiosaur;
-tail = brachiosaur = neck;      // deprecated: is this one or two loads from brachiosaur?
+tail = brachiosaur = neck;      // deprecated: does this touch brachiosaur once or twice?
 \end{codeblock}
 \end{example}
 
@@ -143,7 +143,7 @@ is deprecated.
 \begin{codeblock}
 struct linhenykus { volatile short forelimb; };
 void (linhenykus alvarezsauroid) {
-  auto& [what_is_this] = alvarezsauroid; // deprecated
+  auto& [what_is_this] = alvarezsauroid; // deprecated: very confusing
   // ...
 }
 \end{codeblock}
--- a/source/future.tex
+++ b/source/future.tex
@@ -131,12 +131,12 @@ with a \tcode{volatile}-qualified return type is deprecated.
 
 \begin{example}
 \begin{codeblock}
-volatile struct amber jurassic();  // deprecated: a volatile return type has no meaning
+volatile struct amber jurassic();   // deprecated: a volatile return type has no meaning
 
 // deprecated: volatile parameters aren't meaningful to the caller, volatile only applies within the function
 void trex(volatile short left_arm, volatile short right_arm);
 
-void fly(volatile struct pterosaur* pteranodon);   // OK, the pointer isn't volatile, the data it points to is
+void fly(volatile struct pterosaur* pteranodon);        // OK, the pointer isn't volatile, the data it points to is
 \end{codeblock}
 \end{example}
 
@@ -149,7 +149,7 @@ is deprecated.
 \begin{codeblock}
 struct linhenykus { short forelimb; };
 void park(linhenykus alvarezsauroid) {
-  volatile auto [what_is_this] = alvarezsauroid;   // deprecated: does the binding copy the \tcode{forelimb}s?
+  volatile auto [what_is_this] = alvarezsauroid;        // deprecated: does the binding copy the \tcode{forelimb}s?
   // ...
 }
 \end{codeblock}
--- a/source/future.tex
+++ b/source/future.tex
@@ -98,19 +98,57 @@ Postfix \tcode{++} and \tcode{\dcr} expressions\iref{expr.post.incr} and
 prefix \tcode{++} and \tcode{\dcr} expressions\iref{expr.pre.incr}
 of \tcode{volatile}-qualified arithmetic and pointer types are deprecated.
 
+\begin{example}
+\begin{codeblock}
+volatile int velociraptor;
+++velociraptor;                 // deprecated: is this a single memory operation, two, or a six-foot turkey?
+\end{codeblock}
+\end{example}
+
+
 \pnum
 Certain assignments
 where the left operand is a \tcode{volatile}-qualified non-class type
 are deprecated; see~\ref{expr.ass}.
+
+\begin{example}
+\begin{codeblock}
+int neck, tail;
+volatile int brachiosaur;
+tail = brachiosaur = neck;      // deprecated: is this one or two loads from brachiosaur?
+\end{codeblock}
+\end{example}
+
 
 \pnum
 A function type\iref{dcl.fct}
 with a parameter with \tcode{volatile}-qualified type or
 with a \tcode{volatile}-qualified return type is deprecated.
 
+\begin{example}
+\begin{codeblock}
+// Deprecated: a volatile return type has no meaning to the caller.
+volatile amber jurassic();
+// The volatile left_arm and right_arm parameter are useless to the caller.
+void trex(volatile short left_arm, volatile short right_arm);
+\end{codeblock}
+\end{example}
+
+
 \pnum
 A structured binding\iref{dcl.struct.bind} of a \tcode{volatile}-qualified type
 is deprecated.
+
+\begin{example}
+\begin{codeblock}
+struct linhenykus { volatile short forelimb; };
+void (linhenykus alvarezsauroid) {
+  auto& [what_is_this] = alvarezsauroid; // deprecated
+  // ...
+}
+\end{codeblock}
+\end{example}
+
 
 \rSec1[depr.static.constexpr]{Redeclaration of \tcode{static constexpr} data members}
 
--- a/source/future.tex
+++ b/source/future.tex
@@ -127,7 +127,7 @@ with a \tcode{volatile}-qualified return type is deprecated.
 
 \begin{example}
 \begin{codeblock}
-// Deprecated: a volatile return type has no meaning to the caller.
+// deprecated: a volatile return type has no meaning to the caller.
 volatile amber jurassic();
 // The volatile left_arm and right_arm parameter are useless to the caller.
 void trex(volatile short left_arm, volatile short right_arm);
--- a/source/future.tex
+++ b/source/future.tex
@@ -149,7 +149,7 @@ is deprecated.
 \begin{codeblock}
 struct linhenykus { volatile short forelimb; };
 void (linhenykus alvarezsauroid) {
-  auto& [what_is_this] = alvarezsauroid; // deprecated: does the binding copy the forelimbs?
+  auto& [what_is_this] = alvarezsauroid; // deprecated: does the binding copy the \tcode{forelimb}s?
   // ...
 }
 \end{codeblock}
--- a/source/future.tex
+++ b/source/future.tex
@@ -115,7 +115,7 @@ are deprecated; see~\ref{expr.ass}.
 \begin{codeblock}
 int neck, tail;
 volatile int brachiosaur;
-tail = brachiosaur = neck;      // deprecated: does this touch brachiosaur once or twice?
+tail = brachiosaur = neck;      // deprecated: does this access brachiosaur once or twice?
 \end{codeblock}
 \end{example}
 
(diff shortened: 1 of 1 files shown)


source/future.tex
```diff
@@ -98,20 +98,58 @@
...
 \pnum
 Certain assignments
 where the left operand is a \tcode{volatile}-qualified non-class type
 are deprecated; see~\ref{expr.ass}.
 
+\begin{example}
+\begin{codeblock}
+int neck, tail;
+volatile int brachiosaur;
+tail = brachiosaur = neck;      // deprecated: does this touch brachiosaur once or twice?
```
CaseyCarter:
```suggestion
tail = brachiosaur = neck;      // deprecated: does this access brachiosaur once or twice?
```

CaseyCarter: It's not clear from this example which of the two assignment expressions in this statement is deprecated.

CaseyCarter: Could we have a `volatile_lhs += rhs` example in here as well?

source/future.tex
```diff
@@ -98,20 +98,58 @@
...
+
+
 \pnum
 A function type\iref{dcl.fct}
 with a parameter with \tcode{volatile}-qualified type or
 with a \tcode{volatile}-qualified return type is deprecated.
 
+\begin{example}
+\begin{codeblock}
+// Deprecated: a volatile return type has no meaning to the caller.
```
CaseyCarter: Capitalization inconsistent with the other examples.
```suggestion
// deprecated: a volatile return type has no meaning to the caller.
```

source/future.tex
```diff
@@ -98,20 +98,58 @@
...
+volatile amber jurassic();
+// The volatile left_arm and right_arm parameter are useless to the caller.
```
CaseyCarter: All by-value parameters are "useless to the caller". This doesn't tell me what's special about `volatile` by-value parameters that they should be deprecated.

source/future.tex
```diff
@@ -98,20 +98,58 @@
...
+
 \pnum
 A structured binding\iref{dcl.struct.bind} of a \tcode{volatile}-qualified type
 is deprecated.
 
+\begin{example}
+\begin{codeblock}
+struct linhenykus { volatile short forelimb; };
+void (linhenykus alvarezsauroid) {
+  auto& [what_is_this] = alvarezsauroid; // deprecated: very confusing
```
CaseyCarter: This should maybe have a more specific description than "very confusing". If we're going to start deprecating C++ constructs that are "very confusing", we're going to have a *ton* of work to do.

source/future.tex
```diff
@@ -115,7 +115,11 @@
 \begin{codeblock}
 int neck, tail;
 volatile int brachiosaur;
+brachiosaur = neck;             // OK, a volatile store
```
CaseyCarter: "Ok" is an abbreviation of "Okay", not an acronym, so the "k" shouldn't be capitalized unless you intend the postal abbreviation for Oklahoma. Should these be "ok:" for consistency with "deprecated:"?

source/future.tex
```diff
@@ -127,10 +131,12 @@
 
 \begin{example}
 \begin{codeblock}
-// deprecated: a volatile return type has no meaning to the caller.
-volatile amber jurassic();
-// The volatile left_arm and right_arm parameter are useless to the caller.
+volatile amber jurassic(); // deprecated: a volatile return type has no meaning.
+
+// deprecated: volatile parameter aren't meaningful to the caller, volatile only applies within the function.
```
CaseyCarter: parameters

source/future.tex
```diff
@@ -98,20 +98,64 @@
...
 
+\begin{example}
+\begin{codeblock}
+int neck, tail;
+volatile int brachiosaur;
+brachiosaur = neck;             // OK, a volatile store
+tail = brachiosaur;             // OK, a volatile load
+tail = brachiosaur = neck;      // deprecated: does this access brachiosaur once or twice?
+brachiosaur += neck;            // deprecated: does this access brachiosaur once or twice?
+brachiosaur = brachiosaur + neck; // OK, a volatile load, an addition, a volatile store
```
johelegp: These comments need to be aligned at a multiple of 4, among other things. See https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#code-blocks.

source/future.tex
```diff
@@ -98,20 +98,64 @@
...
 where the left operand is a \tcode{volatile}-qualified non-class type
 are deprecated; see~\ref{expr.ass}.
 
+\begin{example}
+\begin{codeblock}
+int neck, tail;
+volatile int brachiosaur;
+brachiosaur = neck;             // OK, a volatile store
+tail = brachiosaur;             // OK, a volatile load
+tail = brachiosaur = neck;      // deprecated: does this access brachiosaur once or twice?
```
johelegp:
```suggestion
tail = brachiosaur = neck;      // deprecated: does this access \tcode{brachiosaur} once or twice?
```

source/future.tex
```diff
@@ -98,20 +98,64 @@
...
+brachiosaur += neck;            // deprecated: does this access brachiosaur once or twice?
```
johelegp:
```suggestion
brachiosaur += neck;            // deprecated: does this access \tcode{brachiosaur} once or twice?
```

source/future.tex
```diff
@@ -98,20 +98,64 @@
...
+
 \pnum
 A structured binding\iref{dcl.struct.bind} of a \tcode{volatile}-qualified type
 is deprecated.
 
+\begin{example}
+\begin{codeblock}
+struct linhenykus { volatile short forelimb; };
+void (linhenykus alvarezsauroid) {
+  auto& [what_is_this] = alvarezsauroid; // deprecated: does the binding copy the forelimbs?
```
johelegp:
```suggestion
  auto& [what_is_this] = alvarezsauroid; // deprecated: does the binding copy the \tcode{forelimb}s?
```

source/future.tex
```diff
@@ -98,20 +98,64 @@
...
+
+
 \pnum
 A function type\iref{dcl.fct}
 with a parameter with \tcode{volatile}-qualified type or
 with a \tcode{volatile}-qualified return type is deprecated.
 
+\begin{example}
+\begin{codeblock}
+volatile amber jurassic(); // deprecated: a volatile return type has no meaning.
```
johelegp: Comments above this one don't end in a full stop. I suggest looking around where this is added and following their ways.

source/future.tex
jfbastien: "OK" is used 597 in the standard, and "Ok" 0 times :)

jensmaurer: Deviating from the "OK" convention is not in scope for this particular pull request. Please create a new editorial issue if you feel discussion is warranted.

source/future.tex
```diff
@@ -98,20 +98,64 @@
...
+
+
 \pnum
 A function type\iref{dcl.fct}
 with a parameter with \tcode{volatile}-qualified type or
 with a \tcode{volatile}-qualified return type is deprecated.
 
+\begin{example}
+\begin{codeblock}
+volatile amber jurassic();  // deprecated: a volatile return type has no meaning
```
jensmaurer: The identifier "amber" does not appear to be defined.

source/future.tex
```diff
@@ -98,20 +98,64 @@
...
+
+// deprecated: volatile parameters aren't meaningful to the caller, volatile only applies within the function
+void trex(volatile short left_arm, volatile short right_arm);
+
+void fly(volatile pterosaur* pteranodon);   // OK: the pointer isn't volatile, the data it points to is
```
jensmaurer: Please add at least a declaration of "pterosaur", otherwise there's an undefined identifier used in this example.

source/future.tex
```diff
@@ -98,20 +98,64 @@
...
+
 \pnum
 A structured binding\iref{dcl.struct.bind} of a \tcode{volatile}-qualified type
 is deprecated.
 
+\begin{example}
+\begin{codeblock}
+struct linhenykus { volatile short forelimb; };
+void (linhenykus alvarezsauroid) {
+  auto& [what_is_this] = alvarezsauroid;    // deprecated: does the binding copy the \tcode{forelimb}s?
```
jensmaurer: This is using "auto&", why would it copy anything at all?

source/future.tex
```diff
@@ -98,20 +98,64 @@
...
+
+
 \pnum
 A structured binding\iref{dcl.struct.bind} of a \tcode{volatile}-qualified type
 is deprecated.
 
+\begin{example}
+\begin{codeblock}
+struct linhenykus { volatile short forelimb; };
+void (linhenykus alvarezsauroid) {
```
jensmaurer: Function name is missing.

source/future.tex
jfbastien: OK.

jwakely: "OK" is correct, it's not an abbreviation of okay. The "okay" spelling is a bacronym. "Ok" would be wrong.

CaseyCarter: I am willing to allow that "OK" is as correct as "Ok" - style guides vary, and the word has no clear etymology. Let's consider this point resolved. I'm Ok with "OK".

CaseyCarter: ...although I do find the colon vs. comma inconsistency in "deprecated: " and "OK, " disturbing.

source/future.tex
```diff
@@ -98,20 +98,64 @@
...
 with a \tcode{volatile}-qualified return type is deprecated.
 
+\begin{example}
+\begin{codeblock}
+volatile struct amber jurassic();  // deprecated: a volatile return type has no meaning
+
+// deprecated: volatile parameters aren't meaningful to the caller, volatile only applies within the function
+void trex(volatile short left_arm, volatile short right_arm);
+
+void fly(volatile struct pterosaur* pteranodon);   // OK: the pointer isn't volatile, the data it points to is
```
CaseyCarter: Why is this "OK:" when elsewhere you use "Ok, "?

jfbastien: For consistency with the rest of the standard, where we do `OK:` 257 times and `OK,` 131 times ;-)

I've updated to be consistent here.

source/future.tex
```diff
@@ -98,20 +98,64 @@
 prefix \tcode{++} and \tcode{\dcr} expressions\iref{expr.pre.incr}
 of \tcode{volatile}-qualified arithmetic and pointer types are deprecated.
 
+\begin{example}
+\begin{codeblock}
+volatile int velociraptor;
+++velociraptor;                 // deprecated: is this a single memory operation, two, or a six-foot turkey?
```
zygoloid: The tone of this comment is too flippant, and doesn't really work as a rationale for deprecation (since the same question applies for all non-deprecated volatile operations, which are all just as implementation-defined as each other).

Just saying "// deprecated" seems fine (here and below). We don't need to justify why things are deprecated.

jfbastien: It's a [very Cultured reference](https://www.youtube.com/watch?v=TbrsQNeEyIg), but I can drop it if you want.

jensmaurer: Quote from https://isocpp.org/std/standing-documents/sd-4-wg21-practices-and-procedures#code-of-conduct:
"Avoid metaphors, irony & be aware that jokes and humor may not translate"

jfbastien: I dropped the reference, but I don't think we want to drop the explanations. I was explicitly asked for them in CWG.

source/future.tex
```diff
@@ -98,20 +98,64 @@
 prefix \tcode{++} and \tcode{\dcr} expressions\iref{expr.pre.incr}
 of \tcode{volatile}-qualified arithmetic and pointer types are deprecated.
 
+\begin{example}
+\begin{codeblock}
+volatile int velociraptor;
+++velociraptor;                 // deprecated: is this a single memory operation, or two
```
zygoloid: Please remove the attempt to give implicit rationale here. Just say that this case is deprecated, nothing more.

source/future.tex
```diff
@@ -98,20 +98,64 @@
...
+
 \pnum
 A structured binding\iref{dcl.struct.bind} of a \tcode{volatile}-qualified type
 is deprecated.
 
+\begin{example}
+\begin{codeblock}
+struct linhenykus { volatile short forelimb; };
+void park(linhenykus alvarezsauroid) {
+  auto [what_is_this] = alvarezsauroid;    // deprecated: does the binding copy the \tcode{forelimb}s?
```
zygoloid: This example is not deprecated. The deprecated case is
```
struct whatever { int n; };
volatile auto [what_is_this] = whatever{1};
```

source/future.tex
jfbastien: Done.

source/future.tex
jfbastien: Done.

jfbastien:
@zygoloid @jensmaurer what's the status of this change? I don't think anything remains to be changed.

jensmaurer:
@zygoloid, where are we here?

PR #3138, opened 2019-08-06
[lib] Harmonize italics correction after \placeholder

### LINKED ISSUE #3135: \placeholder or \placeholdernc?
I've seen both \placeholder and \placeholdernc used within tcode/codeblock contexts throughout the spec.  The nc form should be used to turn off italic correction in places where it "looks bad", but it's not clear what constitutes such a place.  This should be clarified in our guidelines with examples of when to use each, and we should a do sweep of the Latex sources to fix any incorrect uses.

In #3056, @zygoloid provides some hints as to when to use which:

Please don't blindly change \placeholder to \placeholdernc. The latter is a surgical tool for fixing cases of bad kerning due to italics corrections, and should be used sparingly only where there is actual bad kerning. Usually the italics correction is what we want; the bounding box for italicized text is wider due to the right overhang, and it so we want to create a little extra space.

I've found that \placeholdernc improves the presentation immediately before a period or a comma (where there is no higher text for the italics to get uncomfortably close to). So let's take that as rough guidance for now: use \placeholdernc before a teletype period or comma, where there would otherwise be an unsightly amount of horizontal space.

Looking through a few more examples, \placeholdernc before a left paren also seems more pleasant to my eye, but before a left angle bracket I prefer including the italic correction.

jensmaurer:
Fixes #3135.

### DIFF
--- a/papers/n4827.html
+++ b/papers/n4827.html
@@ -1,0 +1,965 @@
+<!DOCTYPE html><html><head><meta charset="utf-8"><style>html { font-size: 100%; overflow-y: scroll; -webkit-text-size-adjust: 100%; -ms-text-size-adjust: 100%; }
+
+body{
+  color:#444;
+  font-family:Georgia, Palatino, 'Palatino Linotype', Times, 'Times New Roman',
+              "Hiragino Sans GB", "STXihei", "微软雅黑", serif;
+  font-size:12px;
+  line-height:1.5em;
+  background:#fefefe;
+  width: 45em;
+  margin: 10px auto;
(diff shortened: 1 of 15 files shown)


jensmaurer:
Rebased.

jensmaurer:
#3159 will (as fall-out) address some of these cases by using semantic markup for exposition-only concept names. I will refrain from rebasing until the fate of #3159 is decided.

AlisdairM:
#3159 landed a while back --- is this PR still of interest?

PR #3145, opened 2019-08-06
[char.traits.specializations.char8.t] Align member typedefs.

### LINKED ISSUE #3108: [char.traits.specializations.char8.t] Inconsistent formatting with sibling specializations
To be consistent with the formatting of the other specializations of `char_traits`, the member aliases should be aligned and there should be a line break after the `assign` member function.

jensmaurer:
Fixes #3108.

### DIFF
--- a/source/strings.tex
+++ b/source/strings.tex
@@ -317,10 +317,10 @@ returns
 \begin{codeblock}
 namespace std {
   template<> struct char_traits<char8_t> {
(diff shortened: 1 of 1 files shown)


johelegp:
The stable name should be (char)8, not 16.

jensmaurer:
@johelegp Oops, fixed.

PR #3153, opened 2019-08-08
Lost changes to P1754R1 (LWG Motion 11)

burblebee:
These fixes from branch motions-2019-07-lwg-11 were lost in the new branch motions-2019-07-lwg-11-rs.  The 2nd commit in this PR encodes additional but unconventional uses of concept names with \libconcept, but more work may still be needed - see related inline comments in PR #3149.

Update 2/25/21: Most of the changes in the original patch had already been fixed in master since this PR was created.  The rebased branch has been updated to include the few changes that remain in master.

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -4675,7 +4675,7 @@ This is \tcode{std::strong_ordering} if the expansion is empty.
 \end{note}
 \end{itemdescr}
 
-\rSec2[cmp.concept]{Concept \tcode{three_way_comparable}}
+\rSec2[cmp.concept]{Concept \libconcept{three_way_comparable}}
 
 \begin{codeblock}
 template<class T, class Cat>
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -14120,7 +14120,7 @@ resolves to a built-in operator comparing pointers.
 \begin{codeblock}
 struct compare_three_way {
   template<class T, class U>
-    requires three_way_comparable_with<T, U> || @\placeholdernc{BUILTIN-PTR-THREE-WAY}@(T, U)
+    requires @\libconcept{three_way_comparable_with}@<T, U> || @\placeholdernc{BUILTIN-PTR-THREE-WAY}@(T, U)
   constexpr auto operator()(T&& t, U&& u) const;
 
   using is_transparent = @\unspec@;
@@ -14129,7 +14129,7 @@ struct compare_three_way {
 
 \begin{itemdecl}
 template<class T, class U>
-  requires three_way_comparable_with<T, U> || @\placeholdernc{BUILTIN-PTR-THREE-WAY}@(T, U)
+  requires @\libconcept{three_way_comparable_with}@<T, U> || @\placeholdernc{BUILTIN-PTR-THREE-WAY}@(T, U)
 constexpr auto operator()(T&& t, U&& u) const;
 \end{itemdecl}
 
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -14121,7 +14121,7 @@ resolves to a built-in operator comparing pointers.
 struct compare_three_way {
   template<class T, class U>
     requires @\libconcept{three_way_comparable_with}@<T, U> || @\placeholdernc{BUILTIN-PTR-THREE-WAY}@(T, U)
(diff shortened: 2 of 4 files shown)


burblebee:
The build is timing out with "Timeout waiting for network availability."

jensmaurer:

tkoeppe:
I think this is obsolete now, @burblebee?

burblebee:
@tkoeppe, no these changes were absolutely needed. I didn't realize they hadn't been applied yet.
I still have the branch thankfully - will see what still needs to be applied after we get thru the current motions.

tkoeppe:
Aha, thank you, sorry about that! Could you please open a regular pull request for those changes then? Thanks!

burblebee:
> Aha, thank you, sorry about that! Could you please open a regular pull request for those changes then? (Or indeed reopen and update this one?) Thanks!

Is this not a regular PR?  Yes, will update and rebase after the current motions are applied ( I suspect many of these changes have been found and fixed already).

tkoeppe:
Sorry, yes, this is of course a regular PR, I failed to say that I was implicitly thinking "from a branch on your own fork", as opposed to adding PR branches to this repo. But let's leave this as is now, once it's rebased and dealt with, we'll delete the branch. Thanks!

source/support.tex
```diff
@@ -4188,9 +4188,9 @@
 
   // \ref{cmp.concept}, concept \libconcept{three_way_comparable}
   template<class T, class Cat = partial_ordering>
-    concept three_way_comparable = @\seebelow@;
+    concept @\deflibconcept{three_way_comparable}@ = @\seebelow@;
   template<class T, class U, class Cat = partial_ordering>
-    concept three_way_comparable_with = @\seebelow@;
+    concept @\deflibconcept{three_way_comparable_with}@ = @\seebelow@;
```
jensmaurer, line 4193: This should not be `\deflibconcept` here; the `\deflibconept` for these is further down, with the actual (semantic) definition. Make it a `\libconcept`, please.

source/utilities.tex
```diff
@@ -7093,9 +7093,9 @@
   template<class T1, class D1, class T2, class D2>
     requires @\libconcept{three_way_comparable_with}@<typename unique_ptr<T1, D1>::pointer,
                                        typename unique_ptr<T2, D2>::pointer>
-    compare_three_way_result_t<typename unique_ptr<T1, D1>::pointer,
-                               typename unique_ptr<T2, D2>::pointer>
-      operator<=>(const unique_ptr<T1, D1>& x, const unique_ptr<T2, D2>& y);
+      compare_three_way_result_t<typename unique_ptr<T1, D1>::pointer,
+                                 typename unique_ptr<T2, D2>::pointer>
+        operator<=>(const unique_ptr<T1, D1>& x, const unique_ptr<T2, D2>& y);
```
jensmaurer, line 7098: This should be dropped, I believe. "requires" does not cause additional two-space indentation.
(same for the rest here)

jensmaurer (changes requested):

source/support.tex
burblebee: right - I caught that also and fixed.  You're too fast for me! :)

source/utilities.tex
burblebee: We're inconsistent about this.  If we don't want extra indents after requires, then we should fix the places that do indent.  I'm fine either way, although I have a slight preference to not indent also.  I've removed the extra indents from this patch for now.

jensmaurer (approved):

source/support.tex
burblebee: Oh, and as far as making it a \libconcept, we don't seem to do that in general for the declaration.  Should we? In the synopsis, we (typically) only use \libconcept in requires clauses and parameters. We do use \exposconcept for the \expos case however (which makes sense).  See:
`grep "^[ ]*concept .*=" *.tex |grep -v deflibcon |grep -v defexpos |grep -v seebelow`

jensmaurer: It's not really useful to add "use" index entries pointing to the synopsis of a definition if we have a proper `\itemdecl` definition in the vicinity.

burblebee: agreed

PR #3159, opened 2019-08-08
[index] Add index of concept names.

### LINKED ISSUE #2959: An index of library concepts would be extremely useful
The standard library specification is filling up with concepts, especially "exposition only" concepts, and Cpp17Requirements.  It is often necessary to locate the definition of these concepts to understand the clause you are reading, and there is no easy collected reference to find them.  We could look into the index of library names, but it is not clear that the exposition-only names belong there.

It would be helpful to add a new index for all library concepts, including the exposition-only ones denoted by appearing in italics.  It might also be useful to index the Cpp17Requirements, but most of those are reasonably easy to find today, and with a consistent prefix naming convention, perhaps better served by consistently adding them to the main index (where some, but not all, already reside).

jensmaurer:
Fixes #2959.

### DIFF
--- a/source/Makefile
+++ b/source/Makefile
@@ -31,6 +31,7 @@ reindex:
 	makeindex -s libraryindex.ist libraryindex
 	makeindex grammarindex
 	makeindex impldefindex
+	makeindex conceptindex
 	$(STDPDF)
 	makeindex -s basic.gst -o xrefindex.gls xrefindex.glo
 	makeindex -s basic.gst -o xrefdelta.gls xrefdelta.glo
--- a/source/back.tex
+++ b/source/back.tex
@@ -84,6 +84,17 @@ the page where the synopsis of the header is shown.\\}
 }
 
 \clearpage
+\renewcommand{\preindexhook}{The bold page number for each entry is the page
+where the concept is defined.
+Other page numbers refer to pages where the concept is mentioned in the general text.\\}
+\renewcommand{\indexname}{Index of library concepts}
+\renewcommand{\leftmark}{\indexname}
+{
+\raggedright
+\printindex[conceptindex]
+}
+
+\clearpage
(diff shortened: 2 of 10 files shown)


tkoeppe:
I haven't looked at the output, but as long as it works with `latexmk` and looks as expected, the approach looks reasonable. Feel free to merge!

jensmaurer:
Sample index page:

index

Since this is replacing \placeholder/nc with semantic macros (without the "nc"), there are some slight whitespace changes in the main text.

Before:
before2

After:
after2

If this is undesirable, I fear I have to introduce "nc" variants of the macros. Please advise.

jensmaurer:
(Uh, same-as shouldn't be italics in the "after" picture. Will fix right now.)

jensmaurer:
Editorial teleconference: introduce \libconceptx to fix \cname use in ranges.tex. Otherwise good to merge.

jwakely:
This seems to have borked the bookmarks in the PDF:
Screenshot_20191010_102740

jensmaurer:
Rats.

jensmaurer:
Is there any way to do different things in an \rSec vs. outside?

jensmaurer:
Repaired by #3286.

Eelis:
For the `same_as` example, the concept index now also has an entry for _`same-as-impl`_, which is an exposition-only name only used in the implementation of `same_as`.
Was this intentional? :) Looks a bit weird, because why would a reader ever look up _`same-as-impl`_ in the index?

jensmaurer:
Well, in general we also add exposition-only concepts to the index, because they might be used far away from their definition. That's certainly not the case for same-as-impl, but it feels slightly inconsistent to decide this on a case-by-case basis.

Eelis:
Ah fair enough, thanks!

PR #3162, opened 2019-08-08
[thread.jthread.class] Rephrase introductory sentence.

### LINKED ISSUE #3157: [thread.jthread.class] suggests the thread joins itself
As I said at https://github.com/cplusplus/draft/pull/3072#pullrequestreview-266507401 I think this wording is unclear and inaccurate:

> The functionality is the same as for class `thread` (32.4.2) with the additional ability to request that the thread stops and then joins the started thread.

The threads referred to as "the thread" and "the started thread" are actually the same thread, so this says "the ability to request that the thread stops and then joins itself" which is not what the class does.

I suggest:

> The functionality is the same as for class `thread` (32.4.2) with the additional abilities to provide a `stop_token` (32.3) to the new thread of execution, make stop requests, and automatically join.

jensmaurer:
Fixes #3157

### DIFF
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -1360,9 +1360,9 @@ The class \tcode{jthread} provides a mechanism
 to create a new thread of execution.
 The functionality is the same as for
 class \tcode{thread}\iref{thread.thread.class}
-with the additional ability to request that the thread stops
-and then joins the started thread.
-
+with the additional abilities to provide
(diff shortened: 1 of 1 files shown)


jwakely:
Obviously I like this, as it's my wording :smile: but @lewissbaker also said he was OK with this wording.

PR #3163, opened 2019-08-08
[version.syn] Add synopsis for <version> header.

### LINKED ISSUE #3147: missing synopsis for <version> header
We have synopses for every header in [tab:headers.cpp] and [tab:headers.cpp.c] other than \<version>, but we lack one for \<version>. Yes, it'd be empty other than a note saying it includes all the feature-test macros, but we should include a synopsis anyway for consistency.

jensmaurer:
Fixes #3147.

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -530,6 +530,8 @@ supply characteristics of
 imple\-mentation-dependent
 arithmetic types\iref{basic.fundamental}.
 
+\rSec2[version.syn]{Header \tcode{<version>} synopsis}
+
 \pnum
 The header \libheaderdef{version}
 supplies implementation-dependent information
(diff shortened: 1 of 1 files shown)


jensmaurer:
@zygoloid , did you have something like this in mind?

CaseyCarter:
Alternately, we could eliminate the table and make the definitions explicit in the synopsis:
```c++
#define __cpp_lib_concepts 201806L // Also in: <concepts>
```
It seems silly to have a synopsis with simply `// what the table says`, and equally silly to have a table to depict things that we could as easily declare directly.

zygoloid:
I like @CaseyCarter's suggestion. Let's try treating these like any other macro the library provides.

jensmaurer:
Here's how a proper synopsis with #define would look like:

version

jensmaurer:
If desired, I can add section cross-references after each mention of a header. Please advise.

jwakely:
IMO the cross-references would look cluttered (especially if given on every mention of a header, not just the first mention).

zygoloid:
I like this. Has anyone already done a pass over a before and after document to check that the macros and versions all match? This seems like an easy change for a typo to slip into unnoticed.

source/support.tex
```diff
@@ -530,251 +530,141 @@
...
+#define @\defnlibxname{cpp_lib_make_reverse_iterator}@             201402L // also in \libheader{iterator}
+#define @\defnlibxname{cpp_lib_make_unique}@                       201304L // also in \libheader{memory}
+#define @\defnlibxname{cpp_lib_map_try_emplace}@                   201411L // also in \libheader{map}
+#define @\defnlibxname{cpp_lib_math_constants}@                    201907L // also in \libheader{numbers}
+#define @\defnlibxname{cpp_lib_math_special_functions}@            201603L // also in \libheader{cmath}
+#define @\defnlibxname{cpp_lib_memory_resource}@                   201603L // also in \libheader{memory_resource}
+#define @\defnlibxname{cpp_lib_node_extract}@                      201606L
+  // also in \libheader{map}, \libheader{set}, \libheader{unordered_map}, \libheader{unordered_set}
+#define @\defnlibxname{cpp_lib_nonmember_container_access}@        201411L
+  // also in \libheader{iterator}, \libheader{array}, \libheader{deque}, \libheader{forward_list}, \libheader{list}, \libheader{map}, \libheader{regex}, \libheader{set}, \libheader{string},
```
CaseyCarter: Pre-existing: `<iterator>` is out of order here.

source/support.tex
```diff
@@ -530,251 +530,141 @@
...
+#define @\defnlibxname{cpp_lib_allocator_traits_is_always_equal}@  201411L
+  // also in \libheader{memory}, \libheader{scoped_allocator}, \libheader{string}, \libheader{deque}, \libheader{forward_list}, \libheader{list}, \libheader{vector},
+  // \libheader{map}, \libheader{set}, \libheader{unordered_map}, \libheader{unordered_set}
+#define @\defnlibxname{cpp_lib_any}@                               201606L // also in \libheader{any}
+#define @\defnlibxname{cpp_lib_apply}@                             201603L // also in \libheader{tuple}
+#define @\defnlibxname{cpp_lib_array_constexpr}@                   201603L // also in \libheader{iterator}, \libheader{array}
+#define @\defnlibxname{cpp_lib_as_const}@                          201510L // also in \libheader{utility}
+#define @\defnlibxname{cpp_lib_atomic_flag_test}@                  201907L // also in \libheader{atomic}
+#define @\defnlibxname{cpp_lib_atomic_is_always_lock_free}@        201603L // also in \libheader{atomic}
+#define @\defnlibxname{cpp_lib_atomic_lock_free_type_aliases}@     201907L // also in\libheader{atomic}
```
CaseyCarter: `// also in\libheader{atomic}` is missing a space between "in" and `\libheader`.

source/support.tex
```diff
@@ -530,251 +530,141 @@
...
+
 \pnum
 The header \libheaderdef{version}
 supplies implementation-dependent information
 about the C++ standard library
 (e.g., version number and release date).
 
 \pnum
-The macros in \tref{support.ft} are defined
+The macros are defined
```
CaseyCarter: It's not clear what macros "The macros" refers to at this point. I suggest rephrasing this sentence: "Each of the macros defined in `<version>` is also defined after inclusion of any member of the set of library headers indicated in the corresponding comment in this synopsis."

CaseyCarter:
> Has anyone already done a pass over a before and after document to check that the macros and versions all match?

I have just performed such a pass; it's all good.

source/support.tex
jensmaurer: Are we usually alphabetical here? In any case, let's not try to fix this in this round.

source/support.tex

source/support.tex
jensmaurer: Sounds good. Applied.

source/support.tex
jensmaurer: Fixed anyway.

source/support.tex
```diff
@@ -530,251 +530,141 @@
...
+#define @\defnlibxname{cpp_lib_shared_timed_mutex}@                201402L // also in \libheader{shared_mutex}
+#define @\defnlibxname{cpp_lib_source_location}@                   201907L // also in \libheader{source_location}
+#define @\defnlibxname{cpp_lib_spaceship}@                         201907L // also in \libheader{compare}
+#define @\defnlibxname{cpp_lib_string_udls}@                       201304L // also in \libheader{string}
+#define @\defnlibxname{cpp_lib_string_view}@                       201606L // also in \libheader{string}, \libheader{string_view}
+#define @\defnlibxname{cpp_lib_three_way_comparison}@              201711L // also in \libheader{compare}
+#define @\defnlibxname{cpp_lib_to_array}@                          201907L // also in \libheader{array}
+#define @\defnlibxname{cpp_lib_to_chars}@                          201611L // also in \libheader{charconv}
+#define @\defnlibxname{cpp_lib_transformation_trait_aliases}@      201304L // also in \libheader{type_traits}
+#define @\defnlibxname{cpp_lib_transparent_operators}@             201510L // also in \libheader{memory} ,\libheader{functional}
```
zygoloid: Looks like the comma is the wrong side of the space here.

PR #3168, opened 2019-08-10
[std] Rename 'floating literal' to 'floating-point literal'.

### LINKED ISSUE #3165: consider renaming "floating literals" to "floating-point literals"
"Floating literals" do not float. They are literals of floating-point type, so we should call them "floating-point literals" instead. Survey of usage:

 * "floating literal" appears 17 times, including in two other defined terms (decimal floating literal and hexadecimal floating literal), plus one occurrence of "floating or integer literal", and we have three grammar productions floating-literal, decimal-floating-literal, and hexadecimal-floating-literal.
  * "floating-point literal" appears only in [time.duration.literals]p2 (and in the index).
  * "floating operands" and "floating expressions" appear in [expr.pre]p6 (those are meaningless and should just be fixed).
  * "floating type" does not appear anywhere.
  * "floating point type" appears 4 times (those should just be fixed; no need for more discussion on that).
  * an "entity is characterized [...] as floating" by [rand]/3.3; this simply seems wrong.
  * all other occurrences of "floating" are as part of the term "floating-point".

So other than bugs, "floating literal" is the only context in which we use "floating" not followed by "-point". We have a strong consistency argument to fix this.

jensmaurer:
Fixes #3165.

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -1618,7 +1618,7 @@ the basic source character set.
 \grammarterm{pp-number} can contain \tcode{p} \grammarterm{sign} and
 \tcode{P} \grammarterm{sign}.
 \rationale
-Necessary to enable hexadecimal floating literals.
+Necessary to enable hexadecimal floating-point literals.
 \effect
 Valid \CppXIV{} code may fail to compile or produce different results in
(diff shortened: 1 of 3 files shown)


jensmaurer:
Approved by today's CWG teleconference.

PR #3171, opened 2019-08-13
Editing committee editing

opensdh:
Mostly trivial changes made while reviewing.  No attempt is made to distinguish errors in motion application from errors in the motions (or merely nearby).

If the [expr.const]/5 changes are too much, the change to the bullets can certainly be omitted; the change before them is more important as it removes a redundancy.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -968,8 +968,8 @@ into that scope.
 \pnum
 \begin{note}
 For point of instantiation of a template, see~\ref{temp.point}.
-\end{note}
-\indextext{scope!declarations and|)}
+\end{note}%
+\indextext{scope!declarations and|)}%
 \indextext{declaration!point of|)}
 
 \rSec2[basic.scope.block]{Block scope}
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -668,7 +668,7 @@ These requirements also apply to corresponding entities
 defined within each definition of \tcode{D}
 (including the closure types of \grammarterm{lambda-expression}{s},
 but excluding entities defined within default arguments or
-defualt template arguments of either \tcode{D} or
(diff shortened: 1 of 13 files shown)


source/basic.tex
```diff
@@ -546,17 +546,17 @@
 \begin{itemize}
 \item class type\iref{class},
 \item enumeration type\iref{dcl.enum},
-\item inline function with external linkage\iref{dcl.inline},
-\item inline variable with external linkage\iref{dcl.inline},
+\item inline function or variable with external linkage\iref{dcl.inline},
```
jwakely: This _could_ be read as "either an inline function (with external/internal/no linkage), or a variable (inline or not) with external linkage". Separate bullets avoids that ambiguity.

source/overloading.tex
```diff
@@ -1548,7 +1548,8 @@
 of the \grammarterm{deduction-guide}.
 \end{itemize}
 \end{itemize}
-In addition, if \tcode{C} satisfies the conditions for an aggregate class
+In addition, if \tcode{C} satisfies the conditions for an
+aggregate class\iref{dcl.init.aggr}
```
jwakely: This paragraph doesn't follow the semantic line breaks convention, but since you're inserting a newline here anyway, I think the newline should be before "an" so that it is grouped with "aggregate class".

opensdh: It was a novice attempt at introducing a semantic line break.  (I've always used either regular filling or line-per-sentence.)  I'll fix it in the next push.

source/basic.tex
opensdh: @zygoloid said the same thing and approved of just relocating the cross-reference (as has now been done).

PR #3174, opened 2019-08-15
[locale.numpunct,locale.moneypunct] Canonicalize local grammar presen…

### LINKED ISSUE #3100: fix [locale.numpunct] to use the same presentation for its grammar as the rest of the standard
> The syntax for number formats is as follows, where `digit` represents the radix set specified by the `fmtflags` argument value, and `thousands-sep` and `decimal-point` are the results of corresponding `numpunct<charT>` members. Integer values have the format:
> ```
>     integer ::= [sign] units
>     sign ::= plusminus
>     plusminus ::= ’+’ | ’-’
>     units ::= digits [thousands-sep units]
>     digits ::= digit [digits]
> ```
> [...]

This notation is defined nowhere and is different from how we represent all our other grammars. It also got copied into P0645R10 (2019-07 LWG Motion 5), but I'm fixing that as part of applying the motion.

jensmaurer:
…tation.

Fixes #3100.

### DIFF
--- a/source/locales.tex
+++ b/source/locales.tex
@@ -3108,40 +3108,61 @@ locale or their wide character counterparts as if obtained by
 a call to
 \tcode{widen}.
 
+% FIXME: For now, keep the locale grammar productions out of the index;
+% they are conceptually unrelated to the main C++ grammar.
+% Consider renaming these en masse (to locale-* ?) to avoid this problem.
+\newcommand{\locnontermdef}[1]{{\BnfNontermshape#1\itcorr}\textnormal{:}}
+\newcommand{\locgrammarterm}[1]{\gterm{#1}}
+
 \pnum
-The syntax for number formats is as follows, where
-\tcode{digit}
-represents the radix set specified by the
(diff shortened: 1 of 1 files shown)


jwakely:
The change looks good to me, but if you're changing this anyway how about addressing two pre-existing issues.

Why do we have "integer" and "floatval" rather than "intval" and "floatval" or something like that?

Why do we need `plusminus`? Can't we just define `sign` as one of `+` and `-`?

Does it matter that _sign_ and _units_ are defined as part of the _integer_ grammar, but used in the _floatval_ one?

jensmaurer:
Fixed the two issues mentioned in the bullets. Leaving the "sign" and "units" issue as-is; the two grammar snippets are right next to each other.

jensmaurer:
Leaving for @zygoloid to apply.

zygoloid:
After rebasing, could you attach an image of the "after" presentation? Thanks!

jensmaurer:
Rebased. Here are the pictures:
numpunct
money

PR #3175, opened 2019-08-15
Use correct punctuation to terminate non-final list items.

Eelis:


Eelis:
Don't pull yet, I've spotted at least one more. Will amend shortly.

Eelis:
Done amending, these are all I can find.

PR #3182, opened 2019-08-18
Correct indentation for operator<<(basic_ostream<char, traits>&, const char32_t*) = delete.

tahonermann:


jensmaurer:
The commit comment should be something like

> [ostream] Correct indentation for operator<<(..., const char32_t*) = delete.

This shows the section label and shortens the comment.

PR #3187, opened 2019-08-20
[std] Use consistent margins for codeblocks.

### LINKED ISSUE #3186: [everywhere] Use consistent margins for codeblocks
Compare:

https://github.com/cplusplus/draft/blob/deb086c434dcaf864221f51e887477c0842aad2b/source/classes.tex#L221-L222

https://github.com/cplusplus/draft/blob/deb086c434dcaf864221f51e887477c0842aad2b/source/classes.tex#L249-L250

This difference is visible in the PDF:

image

Eelis:
Fixes #3186.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3461,12 +3461,12 @@ the behavior is undefined except in the following cases:
 \end{itemize}
 \begin{example}
 \begin{codeblock}
-  int f(bool b) {
-    unsigned char c;
-    unsigned char d = c;        // OK, \tcode{d} has an indeterminate value
-    int e = d;                  // undefined behavior
-    return b ? d : 0;           // undefined behavior if \tcode{b} is \tcode{true}
-  }
+int f(bool b) {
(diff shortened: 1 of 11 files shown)


tkoeppe:
I think this is good, but would like @zygoloid to confirm that this isn't undoing some of the "deliberate diversity of core style".

jensmaurer:
I certainly hope that random full-codeblock indentation isn't part of the "core diversity".

godbyk:
If you wish to keep the indentation in the LaTeX source code but have it removed in the PDF output, you can add <tt>gobble=<i>n</i></tt> to [`\lstset`](https://github.com/cplusplus/draft/blob/master/source/macros.tex#L410) (where _`n`_ is the number of characters to outdent the code block).

jensmaurer:
Thanks, @godbyk. Though I doubt we want to allow inconsistent codeblock indentation across the LaTeX source to start with.

PR #3189, opened 2019-08-21
[diff.library] Consistency for wide char types

danra:
[diff.char16] says `char16_t` and `char_32t` 
> ...do not appear as *macro* names...

[diff.wchar.t] says `wchar_t` 
> ...does not appear as a *type* name...

It would be nice to make these consistent if there is no reason for the difference. I thought there might not be because I saw at least one implementation in which `char16_t` and `char_32t` are `typedef`s rather than macros.

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -2616,7 +2616,7 @@ The types \tcode{char16_t} and \tcode{char32_t}
 are distinct types rather than typedefs to existing integral types.
 The tokens \tcode{char16_t} and \tcode{char32_t}
 are keywords in this International Standard\iref{lex.key}.
-They do not appear as macro names defined in
+They do not appear as macro or type names defined in
 \libheaderref{cuchar}.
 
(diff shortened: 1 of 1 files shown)


jensmaurer:
ISO C11 section 7.19 says wchar_t is a declared type (and goes on to talk about other things as macros, so this is hardly an oversight).

ISO C11 section 7.28 says char16_t and char32_t are "declared" types.

However, ISO C11 section 7.1.3 clearly says that all global identifiers from header files are reserved for macro use.

jensmaurer:
I think we want "macro or type name" in our updated wording, highlighting both facts. For C++, both the facts that this is not a macro and that it's not a typedef (to some existing type, hampering differentiation in overload resolution) are important.

danra:
Ok! I'll force push that fix to the pull request.

jensmaurer:

danra:
Rebased.

danra:
Hmm, it looks like in the rebase I accidentally pushed just "type names", not "macro or type names. Rebasing and force-pushing.

PR #3190, opened 2019-08-21
[lex.key],[diff.header.iso646.h] Consistent tokens order

danra:
Order the tokens in [diff.header.iso646.h] in the same way they are ordered in Table 6 in [lex.key]. Easier to verify the two lists are identical.


jensmaurer:
This change looks good to me. If I merge this now, the author name will stay as "danra" in the commit history. Is that your intent, or do you want to use a real name? (Force-pushing a changed commit is fine.)

danra:
Force pushed with my name, thanks.

PR #3191, opened 2019-08-21
[intro.refs] ISO/IEC 2382 missing date

danra:
Undated references are mentioned to indicate the latest edition. However, the latest edition of [ISO/IEC 2382](https://www.iso.org/standard/63598.html) is from 2015, and the only place in the standard which mentions ISO/IEC 2382 explicitly is [intro.defs] with the explicit 1993 edition.

So it seems like either the 1993 edition should be explicitly listed in [intro.refs] (this is the suggested fix in this pull request) or the document (or at least the explicit reference in [intro.defs]) should be updated to mention the 2015 edition (assuming there were no breaking changes since 1993, which I definitely did *not* verify).

### DIFF
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -43,7 +43,7 @@ Edited by E. Lear, P. Eggert.
 February 2012 [viewed 2018-03-26].
 Available at
 \url{https://www.ietf.org/rfc/rfc6557.txt}
-\item ISO/IEC 2382 (all parts), \doccite{Information technology ---
+\item ISO/IEC 2382:1993 (all parts), \doccite{Information technology ---
 Vocabulary}
 \item ISO 8601:2004, \doccite{Data elements and interchange formats ---
 Information interchange --- Representation of dates and times}


jensmaurer:
This might be a normative change, more suited for an NB comment.

zygoloid:
ISO/IEC 2382:1993 has been withdrawn (https://www.iso.org/standard/7229.html). If at all possible, we should instead reference the 2015 version (https://www.iso.org/standard/63598.html). We'll need to check which terms were added in the 2015 version and whether we're using those terms for conflicting purposes. That seems like a substantial amount of checking, and should probably be done via a paper (including a presentation of the findings -- what are the new and potentially-conflicting terms, how do they conflict, and what resolutions are being taken for each such case, if any).

jensmaurer:
ISO 2382:2015 is available publicly here: https://www.iso.org/obp/ui/#iso:std:iso-iec:2382:ed-1:v1:en
However, the HTML page is really slow. Here is a text rendition: 
[iso2382.txt.gz](https://github.com/cplusplus/draft/files/4463016/iso2382.txt.gz)

jensmaurer:
Obvious conflicts (just examples): "access" and "object". We need to discuss how to deal with such conflicts in general.

jensmaurer:
Resolved by cplusplus/nbballot#400 and #4187

PR #3193, opened 2019-08-22
[re.grammar] Improve grammar typesetting.

Eelis:
Currently, the regular expression grammar rules in [re.grammar] are formatted with crude codeblocks:

image

In the ECMA standard [itself](https://web.archive.org/web/20111103184035/http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-262%20edition%205.1%2C%20June%202011.pdf), rules are typeset like this:

image

So as an experiment, I tried mimicking that typesetting. Result:

image

Personally, I think it looks nicer than the crude codeblocks, and matching the upstream ECMAScript re grammar typesetting more closely is a nice bonus. What do you think? :)

(Incidentally, this also benefits cxxdraft-htmlgen, because it reduces the amount of non-C++ code in codeblocks.)


zygoloid:
Yes, as this is effectively a patch applied on top of another standard, following that standard's conventions here seems best.

PR #3199, opened 2019-08-24
[temp.param] Remove unused class template from example.

rogerorr:


source/templates.tex
```diff
@@ -475,7 +475,7 @@
 
 template<C1 T> struct s1;              // associates \tcode{C1<T>}
 template<C1... T> struct s2;           // associates \tcode{(C1<T> \&\& ...)}
-template<C2... T> struct s3;           // associates \tcode{(C2<T> \&\& ...)}
+template<C2... T> struct s3;           // associates \tcode{C2<T...>}
```
jensmaurer: I don't believe this change is correct.  I can see how the algorithm in [temp] p6 yields the original outcome, but not how it would yield your suggested outcome.

source/templates.tex
```diff
@@ -623,7 +623,6 @@
 \grammarterm{template-parameter}.
 \begin{example}
 \begin{codeblock}
-template <class T = float> struct B {};
```
jensmaurer: This just seems to be the removal of an unused identifier.

source/templates.tex
rogerorr: Hmm, sorry - this behaviour must have changed since the TS and I didn't notice.
Apologies.
I'll remove this change from this P/R.

jensmaurer: I think I dimly remember we made a change in this area.

rogerorr: Now removed.

source/templates.tex
rogerorr: Yes - it simply distracts from the example.

jensmaurer:
Thanks.

While the title of this pull request is now beautiful, you've actually queued up two commits, both of which lack a bit of splendor in their commit messages.

Please do a "git rebase -i master", merge the two commits, and clean up the commit message. Then do a "git push -u rogerorr -f template-editorial" to force-push and replace everything that was upstream.

(We can do that when we merge, but it's always easier if we have less work. Also, I can't merge without a second reviewer anyway, so we might forget eventually.)

rogerorr:
I have now rebased and force-pushed to get the right commit message and history.

rogerorr:
Jens, can the "change requested" label now be removed?

PR #3206, opened 2019-08-31
[range.range] replace hyphens with dashes

morinmorin:


source/ranges.tex
```diff
@@ -839,7 +839,7 @@
 model \tcode{\placeholdernc{forwarding-range}}. \tcode{subrange} provides
 non-member rvalue overloads of \tcode{begin} and \tcode{end} with the same
 semantics as its member lvalue overloads, and \tcode{subrange}'s iterators
-- since they are ``borrowed'' from some other range -
+-- since they are ``borrowed'' from some other range --
```
jensmaurer: Is an em-dash (three ---) more appropriate for this? Ah, https://en.wikipedia.org/wiki/Dash says we wouldn't have spaces around it then.

morinmorin: I also prefer spaced em dashes. If em dashes were consistently used in the document, I would have used em dahses.

> Ah, https://en.wikipedia.org/wiki/Dash says we wouldn't have spaces around it then.

It's **definitely OK** to put spaces around em dashes — there are two styles (one with spaces and one without spaces).

> Is an em-dash (three ---) more appropriate for this?

I just thought that the em dash just after the end of this sentence becomes interfering on the eye, if we use em dashes. But either works for me.

PR #3209, opened 2019-09-02
[iterators] s/a == b/bool(a == b)/ and s/++a == ++b/bool(++a == ++b)/

cjdb:
The boolean-expressions in [iterators] should be explicitly converted to
`bool`. This commit fixes a few strays that weren't boolified.


jwakely:
Is this correction already implied by a paper approved in Cologne? Otherwise, this is arguably not editorial and so would be better handled as an LWG issue.

jensmaurer:
References to prior art (papers, LWG issues) would be good. Otherwise, this sounds non-editorial.

PR #3214, opened 2019-09-05
[thread.latch] Subordinate [latch.syn] and [thread.latch.class]

### LINKED ISSUE #3169: move [latch.syn], [thread.latch.class] under [thread.latch]
The intended subclauses of [thread.latch] are at the same level as it rather than nested under it.

Note that fixing this will create a hanging paragraph under [thread.latch]. This should be fixed in some way. While fixing that, check the other subclauses of [thread] for hanging paragraphs; there are quite a few of them.

jensmaurer:
Fixes #3169.


jensmaurer:
Rebased.

PR #3216, opened 2019-09-07
Use consistent wording to indicate ADL-only lookup.

Quuxplusone:
Related to "[isocpp-lib] ADL wording formula" and [LWG3247](https://wg21.link/lwg3247); attn @CaseyCarter and @W-E-Brown among others.

This is intended to be basically editorial. The intent is:
- to use the _same_ wording in each affected place,
- to make sure the words "ADL" are used with a crossreference to the section on ADL,
- to eliminate references to "poison-pill" deleted overloads which generally are not needed, not even as an implementation detail, post-C++17.

(For example, `ranges::swap` doesn't need a poison pill because `std::swap` is properly constrained ever since C++17.)

Posting to gather wording feedback, not because I would expect this to get merged right away.
In particular, there's a whole paragraph in "lib-intro.tex" that is basically obsolete at this point, because all it's saying now is "when we say to look up a name by ADL only, we mean to look up a name by ADL only" — it doesn't have to explain the trick because there's no longer any trick. Maybe that paragraph could just be removed.

### DIFF
--- a/source/concepts.tex
+++ b/source/concepts.tex
@@ -583,10 +583,10 @@ and \tcode{E2} is expression-equivalent to an expression
 
 \begin{itemize}
 \item
-  \tcode{S} is \tcode{(void)swap(E1, E2)}\footnote{The name \tcode{swap} is used
-  here unqualified.} if \tcode{E1} or \tcode{E2}
-  has class or enumeration type\iref{basic.compound} and that expression is valid,
-  when `swap` is looked up using only argument-dependent lookup\iref{basic.lookup.argdep}.
+  \tcode{S} is \tcode{(void)swap(E1, E2)},
+  except that \tcode{swap} is looked up using only argument-dependent lookup\iref{basic.lookup.argdep},
+  if that expression is well-formed when treated as an unevaluated operand and
+  if \tcode{E1} or \tcode{E2} has class or enumeration type\iref{basic.compound}.
   If the function selected by overload resolution does not
   exchange the values denoted by
   \tcode{E1} and \tcode{E2},
--- a/source/concepts.tex
+++ b/source/concepts.tex
@@ -585,13 +585,8 @@ and \tcode{E2} is expression-equivalent to an expression
 \item
   \tcode{S} is \tcode{(void)swap(E1, E2)}\footnote{The name \tcode{swap} is used
   here unqualified.} if \tcode{E1} or \tcode{E2}
-  has class or enumeration type\iref{basic.compound} and that expression is valid, with
-  overload resolution performed in a context that includes the declaration
-\begin{codeblock}
-  template<class T>
-    void swap(T&, T&) = delete;
-\end{codeblock}
-  and does not include a declaration of \tcode{ranges::swap}.
+  has class or enumeration type\iref{basic.compound} and that expression is valid,
+  when `swap` is looked up using only argument-dependent lookup\iref{basic.lookup.argdep}.
   If the function selected by overload resolution does not
   exchange the values denoted by
   \tcode{E1} and \tcode{E2},
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -322,15 +322,10 @@ expression-equivalent to:
 
 \item
   Otherwise, \tcode{\placeholdernc{decay-copy}(begin(E))} if it is a
-  valid expression and its type \tcode{I} models
-  \libconcept{input_or_output_iterator} with overload
-  resolution performed in a context that includes the declarations:
-  \begin{codeblock}
-  template<class T> void begin(T&&) = delete;
-  template<class T> void begin(initializer_list<T>&&) = delete;
-  \end{codeblock}
-
-  and does not include a declaration of \tcode{ranges::begin}.
+  valid expression
+  when `begin` is looked up using only argument-dependent lookup\iref{basic.lookup.argdep}
+  and its type \tcode{I} models
+  \libconcept{input_or_output_iterator}.
 
 \item
   Otherwise, \tcode{ranges::begin(E)} is ill-formed.
@@ -367,19 +362,12 @@ expression-equivalent to:
   \end{codeblock}
 
 \item
-  Otherwise, \tcode{\placeholdernc{decay-copy}(end(E))} if it is a valid
-  expression and its type \tcode{S} models
+  Otherwise, \tcode{\placeholdernc{decay-copy}(end(E))} if it is a valid expression
+  when `end` is looked up using only argument-dependent lookup\iref{basic.lookup.argdep},
+  and its type \tcode{S} models
   \begin{codeblock}
   sentinel_for<decltype(ranges::begin(E))>
   \end{codeblock}
-  with overload
-  resolution performed in a context that includes the declarations:
-  \begin{codeblock}
-  template<class T> void end(T&&) = delete;
-  template<class T> void end(initializer_list<T>&&) = delete;
-  \end{codeblock}
-
-  and does not include a declaration of \tcode{ranges::end}.
 
 \item
   Otherwise, \tcode{ranges::end(E)} is ill-formed.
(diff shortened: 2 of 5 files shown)


source/concepts.tex
```diff
@@ -585,13 +585,8 @@
...
   here unqualified.} if \tcode{E1} or \tcode{E2}
-  has class or enumeration type\iref{basic.compound} and that expression is valid, with
-  overload resolution performed in a context that includes the declaration
-\begin{codeblock}
-  template<class T>
-    void swap(T&, T&) = delete;
-\end{codeblock}
-  and does not include a declaration of \tcode{ranges::swap}.
+  has class or enumeration type\iref{basic.compound} and that expression is valid,
+  when `swap` is looked up using only argument-dependent lookup\iref{basic.lookup.argdep}.
```
Quuxplusone: The `\note` is probably unnecessary; we have no such notes in other places.

I'm not happy grammatically with the phrase "X is equivalent to Y, if Y is valid when Z." What we mean is more like "X is equivalent to Y-when-Z, if Y is valid when Z."

source/ranges.tex
```diff
@@ -322,15 +322,10 @@
...
-  \begin{codeblock}
-  template<class T> void begin(T&&) = delete;
-  template<class T> void begin(initializer_list<T>&&) = delete;
-  \end{codeblock}
-
-  and does not include a declaration of \tcode{ranges::begin}.
+  valid expression
+  when `begin` is looked up using only argument-dependent lookup\iref{basic.lookup.argdep}
+  and its type \tcode{I} models
+  \libconcept{input_or_output_iterator}.
```
Quuxplusone: Many of these ranges customization points end with the formula "...and its type `I` models `some-concept`." I don't understand why we need to introduce the name `I`, given that we never use the name `I` again in this scope. However, I didn't change these formulas in this patch.

johelegp: I believe that's because [[customization.point.object]p5](http://eel.is/c++draft/library#customization.point.object-5) requires it, although some CPOs don't do that, IIRC.

jensmaurer:
It's not at all clear to me that this change is editorial (i.e. not changing semantics for corner cases), so (personally), I'd prefer a paper approved by LWG. Specifically, could you elaborate why the deleted overloads were necessary before, but no longer so?

Further, I'm not sure that "only ADL" is implementable in plain C++ (which, of course, does not necessarily constrain the standard library). For example, given
```
int strong_order(X,Y);
#include <stdlib_header>
```
it seems that strong_order may have been found previously (since unqualified lookup inside stdlib_header may eventually reach the global scope), but now you're not supposed to find that unless X or Y are declared in global scope.

Quuxplusone:
> Specifically, could you elaborate why the deleted overloads were necessary before, but no longer so?

My understanding (and maybe @ericniebler and/or @CaseyCarter could weigh in with better understanding) is that the "poison-pill" overloads in range-v3 are needed because range-v3 is a C++14-compatible library. In C++14, `std::swap` (for example) was an unconstrained template.[*] Adding a new deleted overload was needed in order to stop `std::mutex m1, m2; std::ranges::swap(m1, m2);` from accidentally dispatching to `std::swap<std::mutex>` and producing a hard error instead of SFINAE-friendliness. In C++17, `std::swap<std::mutex>` does not participate in overload resolution, so we do not worry that the ADL-only lookup of `swap(m1, m2)` might choose it.

> Further, I'm not sure that "only ADL" is implementable in plain C++

This seems very worrisome if true! But I don't understand your example. Here's my attempt to reproduce it with `swap` as an example: https://ranges.godbolt.org/z/zRUvCb
The implementation strategy for "ADL-only lookup" remains the same: the library implementor puts some untenable overload of `swap` (or `strong_order`) somewhere in the detail namespace that actually makes the call, so that unqualified lookup finds that untenable thing and stops before finding `std::swap`. It's just that in C++17, the untenable declaration doesn't need to have precisely the specified poison-pill form; it might be as simple as `int swap();`.
Example of what I'm thinking: https://godbolt.org/z/AEdYPd


[*] — That is, according to cppreference. According to libc++, [it's been constrained since 2011](https://github.com/llvm-mirror/libcxx/commit/1694d23e23d41c4037111d96324c0c16c744c09d#diff-48f5ee43879b5ad38888f0a6ead10113R3002-R3006). According to N4140 (C++14), it was constrained but only via SFINAE in its `noexcept` clause — admittedly too subtle.

jensmaurer:
Thanks for the explanation. It seems the "int swap()" poison-pill is actually better than the previous form, because it stops non-ADL name lookup, but otherwise is non-viable and doesn't interfere with overload resolution. Consider https://ranges.godbolt.org/z/2VMmyq :
Obviously, my template `swap` in namespace N is found via ADL (status quo of the example), however, `ranges::swap` doesn't find it (remove comments to see it). At least this implementation seems to violate the "nothing but ADL" rule.

source/ranges.tex
Quuxplusone: I didn't understand this answer, btw. My guess is that you were answering the unasked question "Why does the formula mention that the type models `some-concept`?", but not "Why does the formula introduce the name `I` to refer to a type that is never mentioned again?"

I mean, it seems like we should just say "...and its type models `some-concept`." Interjecting the unused identifier `I` seems unnecessary.

Quuxplusone:
> Consider https://ranges.godbolt.org/z/2VMmyq : Obviously, my [completely unconstrained] template `swap` in namespace `N` is found via ADL (status quo of the example), however, `ranges::swap` doesn't find it [...]

Hmm, good point. This change is far from editorial (but still, IMO, good! ;)) and needs a defect/paper. At the very least, your example shows that the current explanatory wording in [customization.point.object] is incorrect:

> To preclude such an expression resulting in a call to unconstrained functions [sic]
> with the same name in namespace \tcode{std}, [...]

The currently specified poison-pill also blocks unconstrained function templates with the same name in _associated_ namespaces! Which, as you say, means that the current poison-pill approach is _not_ the same thing as "ADL-only lookup."

Also, [customization.point.object] means "unconstrained" in the literal C++2a-lawyer sense. For example, [this template](https://ranges.godbolt.org/z/2IoltA) is considered "unconstrained" by C++2a and by Ranges-V3, even though someone coming from C++17 would probably have said colloquially that it was "constrained" to accept only `T`s derived from `N::A`. One might plausibly argue that users who write completely unconstrained, C++03-swap-style, templates deserve what they get — but I don't think we should ship a `ranges::swap` that claims to do ADL-only lookup yet quietly ignores this kind of "colloquially constrained" ADL template.

    template<class T>
    std::enable_if_t<std::is_base_of<N::A, T>::value>
    swap(T&, T&) { /* something */ }

jensmaurer:
We seem to agree the changes are not editorial. Could I kindly ask you to take the matter to LWG via their reflector and/or via their issues list?

source/ranges.tex
```diff
@@ -590,16 +569,10 @@
...
-    if it is a valid expression and its type \tcode{I}
-    is integer-like
-    with overload resolution performed in a context that includes
-    the declaration:
-    \begin{codeblock}
-    template<class T> void size(T&&) = delete;
-    \end{codeblock}
-
-    and does not include a declaration of \tcode{ranges::size}.
+    Otherwise, \tcode{\placeholdernc{decay-copy}(size(E))},
```
CaseyCarter: Removing the poison pill is not editorial, but probably a good idea.

CaseyCarter:
> Which, as you say, means that the current poison-pill approach is not the same thing as "ADL-only lookup."

We're trying to express that overload resolution is performed on a set of functions and function templates that includes the poison pill(s) as well as the results of ADL-only name lookup.

source/ranges.tex
```diff
@@ -321,16 +321,11 @@
   \libconcept{input_or_output_iterator}.
 
 \item
-  Otherwise, \tcode{\placeholdernc{decay-copy}(begin(E))} if it is a
-  valid expression and its type \tcode{I} models
-  \libconcept{input_or_output_iterator} with overload
-  resolution performed in a context that includes the declarations:
-  \begin{codeblock}
-  template<class T> void begin(T&&) = delete;
-  template<class T> void begin(initializer_list<T>&&) = delete;
```
BRevzin: I want to elaborate on @CaseyCarter's comments here in case there's some ambiguity. Removing this poison pill is not editorial and also a bad idea. Consider the following type:

```cpp
namespace N {
    struct my_vector { /* ... */ };
    int const* begin(my_vector const&);
    int const* end(my_vector const&);
}
```

It's the poison pills that prevent `N::my_vector` from satisfying _`forwarding-range`_, removing them would open up some unknowable number of containers to accidentally opt-in to this concept when they definitely should not.

jensmaurer:
@Quuxplusone, is there some editorial cleanup/harmonization left that you wish to have applied given @CaseyCarter's comments, or is this all LWG material?

jensmaurer:
@Quuxplusone, ping?

Quuxplusone:
> @Quuxplusone, is there some editorial cleanup/harmonization left that you wish to have applied given @CaseyCarter's comments, or is this all LWG material?

I don't think I'm likely to do anything else with this PR, so I think it should just be closed at this point.

PR #3219, opened 2019-09-10
[std] Move \pnum and library element markers to a line of their own.

### LINKED ISSUE #3128: Inconsistent Latex coding style used in clauses of \itemdescrs
We're inconsistent in the Latex coding style used for clauses of \itemdescrs.  Some clauses use 2 space indentation for the text (similar to \items), some place the text on the same line as \<clause_macro>, some place the \<clause_macro> and text on the same line as \pnum, etc..  Our [guidelines](https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines?) don't specify how they should be formatted.

Can we come up with some guidelines, document them, and do a global whitespace change to fix the old formatting style so that we can follow the guidelines when applying edits?

jensmaurer:
Fixes #3128.

### DIFF
--- a/source/statements.tex
+++ b/source/statements.tex
@@ -245,7 +245,6 @@ a block scope\iref{basic.scope}. If the substatement in a
 \grammarterm{compound-statement}, it is as if it was rewritten to be a
 \grammarterm{compound-statement} containing the original substatement.
 \begin{example}
-
 \begin{codeblock}
 if (x)
   int i;
@@ -478,7 +477,6 @@ a single statement and not a \grammarterm{compound-statement},
 it is as if it was rewritten to be
 a \grammarterm{compound-statement} containing the original statement.
 \begin{example}
-
 \begin{codeblock}
 while (--x >= 0)
   int i;
@@ -615,7 +613,6 @@ equivalent to \tcode{while(true)}.
 If the \grammarterm{init-statement} is a declaration, the scope of the
 name(s) declared extends to the end of the \tcode{for} statement.
 \begin{example}
-
 \begin{codeblock}
 int i = 42;
 int a[10];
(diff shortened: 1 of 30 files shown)


jensmaurer:
No "diffpdf" differences.

source/statements.tex
```diff
@@ -1108,5 +1107,6 @@
                                 // since it depends on \tcode{T2} being a type-name
 }
 \end{codeblock}
-\end{example}%
+\end{example}
+%
```
tkoeppe: This looks broken: if there really was need to avoid a whitespace, then this newline introduces one.

My suspicion is that this wasn't necessary in the first place (in which case we should remove the `%`), but as far as this "cleanup change" goes, this seems undesirable.

Could you take a look if a) this happens anywhere else, and b) if the `%` was necessary in the first place?

jensmaurer: Yes, this happens a few times. Frequently at the end of a section, just before the closing index marker. So, the % is totally bogus and should / will be removed.

tkoeppe: Could you maybe send a zero-visual-diff commit first that removes the bogus `%` everywhere?

jensmaurer:
Removed the stray "%".
Established a "%" after `\end{note}` where space would have been added before punctuation (e.g. for a note-within-a-bullet; [expr.const]).

source/statements.tex
jensmaurer: Uh, that will produce conflicts of all those removals with the rest of the patch.

I've just browsed through all pages with visual ("Appearance") diffs and there's nothing bad going on, except for the well-documented "space before _-- end note]_ becomes more consistent".

jensmaurer: Ok, I've done an isolated patch that removes trailing % from `note}` and `example}`.

diffpdf in "Appearance" mode shows a single difference in 13.9.2.1, with additional space after an "-- end example}". Which actually looks better now, in my view.

I think the "%" removal is safe.

tkoeppe:
OK, thanks! I would very mildly like to see the `%`-removal commit first, and the `%`-linebreak-reverting commit squashed into whatever commit necessitated the reverting, but only if you can work out how to do the rebasing. Otherwise good to go!

PR #3221, opened 2019-09-11
[diff] Move markers to lines of their own.

### LINKED ISSUE #3129: [diff] Inconsistent LaTeX source code style
We're inconsistent in the Latex coding style used for clauses in [diff.iso].  Some place the text on the same line as the <clause_macro>, others don't.  Our [guidelines](https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines?) don't specify how they should be formatted. I'd like to see us format them similarly to \items, where a single line is used if the text is short, and 2-space indentation is used for clauses which span multiple lines.

Can we come up with some guidelines, document them, and do a global whitespace change to fix the old formatting style so that we can follow the guidelines when applying edits?

jensmaurer:
Fixes #3129.

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -11,7 +11,8 @@ ISO C, by the chapters of this document.
 \rSec2[diff.lex]{\ref{lex}: lexical conventions}
 
 \diffref{lex.key}
-\change New Keywords\\
+\change
+New Keywords\\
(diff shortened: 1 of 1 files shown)


jensmaurer:
No visual changes from "diffpdf" appearance mode.

tkoeppe:
This is great, thanks!

PR #3224, opened 2019-09-13
[over.match.funcs] Remove bullet for single-item bulleted list.

### LINKED ISSUE #3176: [over.match.funcs, ...?] Consider shunning single-item lists
For example:
- https://github.com/cplusplus/draft/blob/173905005c2c419548418239518db72bfda9dd9a/source/overloading.tex#L1381
- https://github.com/cplusplus/draft/blob/173905005c2c419548418239518db72bfda9dd9a/source/overloading.tex#L1431

These look pretty weird. I think a "let's avoid single-item lists" rule would probably make sense for the Specification Style Guidelines wiki page. :)

jensmaurer:
Fixes #3176.


jensmaurer:
Rebased.

PR #3225, opened 2019-09-13
[basic.stc.dynamic.safety] Avoid undefined term 'dynamic object'.

### LINKED ISSUE #3185: [basic.stc.dynamic.safety] "dynamic object" used but not defined
[basic.stc.dynamic.safety]/2:

> A pointer value is a safely-derived pointer to a dynamic object only if it has an object pointer type and it is one of the following: [...]

This is the sole occurrence of "dynamic object" in the document.

jensmaurer:
Fixes #3185.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3830,7 +3830,7 @@ match those of some object pointer type.
 \end{itemize}
 
 \pnum
-A pointer value is a \defn{safely-derived pointer} to a dynamic object only if it
+A pointer value is a \defn{safely-derived pointer} to an object with dynamic storage duration only if it
 has an object pointer type and it is one of the following:
 \begin{itemize}
 \item the value returned by a call to the \Cpp{} standard library implementation of


W-E-Brown:
Sorry, but this merging seems a bit hasty — only 23 minutes (5:03 to 5:26 CDT) from issue start to issue finish, leaving only little time for reviewer comments.

FWIW:  IMO, the amended sentence (bold emphasis added)
> A pointer value is a safely-derived pointer to an object with dynamic storage duration only if it has an object pointer type and it is one of the following:
seems open to unintended multiple interpretations:
We now have additional nouns that might serve as the antecedent of "it":  does "it" stand in for the "value" or the "pointer" or the "object" or the "duration"?
Moreover, whether the two occurrences of "it" are even intended to share the same antecedent is also now unclear.

I hope we can clarify the revised phrasing so as to address these concerns.  For example:

The first "it" ("only if it has ...") could perhaps be replaced by the intended antecedent (e.g., "the value").
The second occurrence of "it" ("and it is one of ...") seems superfluous and could perhaps be simply removed, if both are intended to have the same antecedent.

Together, these would address the above concerns.  There may be other approaches, too.

tkoeppe:
Thank you, those seem like reasonable improvements.

jensmaurer:
@W-E-Brown, thanks for your comments. They are (hopefully) addressed in #3228.

PR #3229, opened 2019-09-16
Replace "shall not throw" with "does not throw" when it describes library behavior

Quuxplusone:
In these places we aren't saying "it's UB if it throws"; we're literally specifying the behavior of a library function as "it does not throw, we promise."

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -513,13 +513,13 @@ namespace std {
 \end{codeblock}
 
 \pnum
-Constructors and member functions of \tcode{pair} shall not throw exceptions unless one of
+Constructors and member functions of \tcode{pair} do not throw exceptions unless one of
 the element-wise operations specified to be called for that operation
 throws an exception.
 
 \pnum
-The defaulted move and copy constructor, respectively, of \tcode{pair} shall
-be a constexpr function if and only if all required element-wise
+The defaulted move and copy constructor, respectively, of \tcode{pair}
+is a constexpr function if and only if all required element-wise
 initializations for copy and move, respectively, would satisfy the
 requirements for a constexpr function.
 
@@ -16174,7 +16174,7 @@ function(const function& f);
 
 \pnum
 \throws
-Shall not throw exceptions if \tcode{f}'s target is
+Nothing if \tcode{f}'s target is
 a specialization of \tcode{reference_wrapper} or
 a function pointer. Otherwise, may throw \tcode{bad_alloc}
 or any exception thrown by the copy constructor of the stored callable object.
@@ -16245,10 +16245,10 @@ reference to an object and a member function pointer.
 
 \pnum
 \throws
-Shall not throw exceptions when \tcode{f} is a function pointer
-or a \tcode{reference_wrapper<T>} for some \tcode{T}. Otherwise,
-may throw \tcode{bad_alloc} or any exception thrown by \tcode{F}'s copy
-or move constructor.
+Nothing if \tcode{f} is
+a specialization of \tcode{reference_wrapper} or
+a function pointer. Otherwise, may throw \tcode{bad_alloc}
+or any exception thrown by \tcode{F}'s copy or move constructor.
 \end{itemdescr}
 
 
--- a/source/time.tex
+++ b/source/time.tex
@@ -1254,7 +1254,7 @@ If \tcode{Period} is not a specialization of \tcode{ratio}, the program is ill-f
 If \tcode{Period::num} is not positive, the program is ill-formed.
 
 \pnum
-Members of \tcode{duration} shall not throw exceptions other than
+Members of \tcode{duration} do not throw exceptions other than
 those thrown by the indicated operations on their representations.
 
 \pnum


jensmaurer:

jwakely:
The existing wording also applies to program-defined specializations of `duration` and `pair`. The proposed change makes that less clear IMO.

source/utilities.tex
```diff
@@ -16245,7 +16245,7 @@
 
 \pnum
 \throws
-Shall not throw exceptions when \tcode{f} is a function pointer
+Does not throw exceptions when \tcode{f} is a function pointer
```
Quuxplusone: @jwakely wrote:

> The existing wording also applies to program-defined specializations of `duration` and `pair`. The proposed change makes that less clear IMO.

...And program-defined specializations of `function`, is that right? So essentially your view is that any library wording that describes the intended behavior of a function template or class template must always use "shall", not "does," because the user can always produce UB by mis-specializing a template? (A "mis-specialization" is a program-defined specialization that does anything different from what the library template is actually documented to do. For example, if the library template is documented to have a copy constructor, and the program-defined specialization does something other than copy; or if the library template is documented as "does not throw exceptions" and the program-defined specialization does throw exceptions.)

I thought SD-8 would have something to say about the inadvisability of mis-specializing library templates, but [the actual text of SD-8](https://isocpp.org/std/standing-documents/sd-8-standard-library-compatibility) seems ridiculously conservative on that score.

jwakely: > So essentially your view is that any library wording that describes the intended behavior of a function template or class template must always use "shall", not "does,"

No, I didn't say that. I'm just pointing out that this may not be as obvious as it appears. However, I think [namespace.std] p2 means that even if we remove these "shalls", program-defined specializations still have to meet that same requirements.

jensmaurer: So, @jwakely , is this good to go as presented? If so, please hit "approve".

jwakely: Same here, "_Throws:_ Nothing if ..."

Quuxplusone: Updated. And caught some more instances of "shall not participate unless X" that can by the same logic be simplified to "participates only if X".

jwakely: > And caught some more instances of "shall not participate unless X" that can by the same logic be simplified to "participates only if X".

Please don't change that. The current phrasing is a consistent form that is used in hundreds of places, please don't change two of them. All occurrences will get replaced with _Constraints:_ elements by papers currently being reviewed by LWG. Please don't complicate that by making piecemeal editorial changes to them.

Quuxplusone: Okay, updated again.

source/utilities.tex
```diff
@@ -16259,7 +16259,7 @@
 \begin{itemdescr}
 \pnum
 \remarks
-This deduction guide participates in overload resolution only if
+This deduction guide shall not participate in overload resolution unless
```
jwakely: Please revert this part.

source/utilities.tex
```diff
@@ -16342,8 +16342,8 @@
 
 \pnum
 \remarks
-This assignment operator shall not participate in overload
-resolution unless \tcode{decay_t<F>} is
+This assignment operator shall not participate in overload resolution unless
+\tcode{decay_t<F>} is
 Lvalue-Callable\iref{func.wrap.func} for argument types \tcode{ArgTypes...} and
```
jwakely: This is needless churn, please revert.

jensmaurer:
@Quuxplusone, there are a few minor changes requested. Please apply them, rebase, and force-push. Thank you.

source/utilities.tex
```diff
@@ -504,13 +504,13 @@
...
-Constructors and member functions of \tcode{pair} shall not throw exceptions unless one of
+Constructors and member functions of \tcode{pair} do not throw exceptions unless one of
 the element-wise operations specified to be called for that operation
 throws an exception.
 
 \pnum
-The defaulted move and copy constructor, respectively, of \tcode{pair} shall
-be a constexpr function if and only if all required element-wise
+The defaulted move and copy constructor, respectively, of \tcode{pair}
+is a constexpr function if and only if all required element-wise
```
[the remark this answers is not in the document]
Quuxplusone: Right now the grammar is: "The X and Y, respectively, is Z if A and B, respectively, satisfies some condition."
I guess you're suggesting: "The X and Y, respectively, are Z if A and B, respectively, satisfy some condition."
Another alternative: "The X or Y, respectively, is Z if A or B, respectively, satisfies some condition."

FWIW, I think this grammar is just too far from proper to have a "right" answer, but my guess was that the singular is more correct — because it's possible for X to be Z independently of Y. Each of _X and Y is Z_ independently of the other.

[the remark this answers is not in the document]
CaseyCarter: > FWIW, I think this grammar is just too far from proper to have a "right" answer

Amen to that: this sentence is a painfully awkward demonstration of the lengths we'll go to in order to avoid repetition. We should probably simply strike it, since
* "shall be a constexpr function" doesn't mean what the Library thinks it means,
* the core language tells us when defaulted constructors are `constexpr`, there's no need to repeat it here, and
* [functions.within.classes] says we "do not describe copy/move constructors, assignment operators, or (non-virtual) destructors with the same apparent semantics as those that can be generated by default".

but that may be a bit of a stretch for an editorial change.

Feel free to ignore this comment, accept it, and/or file an LWG issue to strike the paragraph altogether.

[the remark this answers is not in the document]
zygoloid: This is awful. But it's pre-existing, so I'm not blocking this pull request on it. If someone wants to take some additional action here, please do! (I think I'll change the "and"s to something else as a short-term "fix".)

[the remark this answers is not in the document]
zygoloid: Also, hold on just a minute, this is saying that the defaulted move constructor is constexpr if all copies are valid and the defaulted copy constructor is constexpr if all moves are valid. What the ...?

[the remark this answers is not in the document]
CaseyCarter: > If someone wants to take some additional action here

Ideally, additional action involving cleansing flame.

PR #3230, opened 2019-09-16
Several editorial fixes in [re.regex].

mordante:
Fixes the consistency of several arguments in operator= and assign
and the constructor:
basic_regex from unnamed, that, e -> e
basic_string from p -> s
const charT* from ptr, p -> p
flag_type from unnamed -> f
initializer_list<charT> from unnamed -> il

These changes are in both [re.regex] and [re.regex.assign].

Moves the [re.regex.assign] LaTex reference in [re.regex] before the
operator= overloads to match the [re.regex.assign] section.

There is one more inconsistency [re.regex] declares:
basic_regex& assign(const charT* p, size_t len, flag_type f);

[re.regex.assign] declares:
basic_regex& assign(const charT* p, size_t len,
  flag_type f = regex_constants::ECMAScript);

Since this is not an editorial issue I'll file an LWG issue.

### DIFF
--- a/source/regex.tex
+++ b/source/regex.tex
@@ -1280,26 +1280,26 @@ namespace std {
       basic_regex(const basic_regex&);
       basic_regex(basic_regex&&) noexcept;
       template<class ST, class SA>
-        explicit basic_regex(const basic_string<charT, ST, SA>& p,
+        explicit basic_regex(const basic_string<charT, ST, SA>& s,
                              flag_type f = regex_constants::ECMAScript);
       template<class ForwardIterator>
         basic_regex(ForwardIterator first, ForwardIterator last,
                     flag_type f = regex_constants::ECMAScript);
-      basic_regex(initializer_list<charT>, flag_type = regex_constants::ECMAScript);
+      basic_regex(initializer_list<charT> il, flag_type f = regex_constants::ECMAScript);
 
       ~basic_regex();
 
-      basic_regex& operator=(const basic_regex&);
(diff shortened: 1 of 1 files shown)


source/regex.tex
```diff
@@ -1240,29 +1240,29 @@
       basic_regex();
       explicit basic_regex(const charT* p, flag_type f = regex_constants::ECMAScript);
       basic_regex(const charT* p, size_t len, flag_type f = regex_constants::ECMAScript);
-      basic_regex(const basic_regex&);
-      basic_regex(basic_regex&&) noexcept;
+      basic_regex(const basic_regex& e);
+      basic_regex(basic_regex&& e) noexcept;
```
jensmaurer: We're fairly inconsistent with naming parameters in class synopses; see e.g. [vector.overview]. I'm ok with the rest of the renames, but why do we need a parameter name here?

mordante: I didn't notice before the standards wasn't really consistent in the naming. Do you want me to remove this hunk of the diff?

jensmaurer: Yes, please. (Force-pushing is fine.) I actually believe that constructors and assignment operators should not have parameter names in synopses; there is no additional information conveyed in those parameter names and the signatures should otherwise speak for themselves.

mordante:
I updated the two constructors. Not sure whether you also wanted the assignment operator. If so let me know and I send another update.

jensmaurer:

mordante:
Rebased as requested

Issue #3096, opened 2019-07-28
[lib] Improve macros for library index entries


jensmaurer:
- Make a macro so that every mention of a library header is auto-indexed, with a bold page number for the header synopsis.
- Make a macro to replace `\indexlibrary{\idxcode{blah}}`, e.g. `\indexlibraryglobal{blah}`. This complements `\indexlibrarymember{member]{class}`.
- Make a macro so that an index entry may be performed from inside the `itemdecl` environment, while also showing the name of the item. Idea: `@\libglobal{blah}@` and `@\libmember{blah}{class}@` (only shows "blah", of course).

jensmaurer:
@zygoloid, any opinions on the above?

jensmaurer:
Editorial teleconference. Also want a macro for constructors and similar funny ways. Otherwise, go for it.

jensmaurer:
See also #2223 with the same thrust.

PR #3233, opened 2019-09-18
[lib] Use more abstract library indexing macros

jensmaurer:
Partially addresses #3096.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -2901,7 +2901,7 @@ namespace std {
 
 \rSec2[alg.all.of]{All of}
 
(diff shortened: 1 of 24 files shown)


tkoeppe:
Could you post a diffpdf screenshot? I think this is generally fine; I expect this causes no visible changes outside the index?

jensmaurer:
"diffpdf" in "Appearance" mode shows now changes outside the index.

Sample diffpdf on index (in "Words" mode to remove fluff):

Screenshot_2019-09-25_18-12-42

tkoeppe:
Looks good, ship it!

PR #3239, opened 2019-09-23
[std] Improve indexing of standard library headers.

jensmaurer:
Partially addresses #3096.


jensmaurer:
This should be judiciously squashed before application. I kept the individual commits for easier force-pushing.

Result:

PR #3242, opened 2019-09-24
[locale] Fix example.

### LINKED ISSUE #3238: [locale] Example calls time_put::put with ios_base::iostate& but there is no such overload
[\[locale\] paragraph 3](http://wg21.link/locale#3) is this example: https://github.com/cplusplus/draft/blob/e287e8227303672b78e01c44020087e644411d4e/source/locales.tex#L180-L203

In the `put(s, s, s.fill(), err, &tmbuf, 'x')` call, the fourth argument seems intended to be reference to `ios_base::iostate err`. However, in both `put` member functions in [\[locale.time.put.members\]](http://wg21.link/locale.time.put.members), the fourth parameter is `const tm* t`:
https://github.com/cplusplus/draft/blob/e287e8227303672b78e01c44020087e644411d4e/source/locales.tex#L3977-L3980

I suppose the definition and uses of `ios_base::iostate err` should be deleted from the example, and the code should instead call `ostreambuf_­iterator<charT, traits>::failed()`, like in [\[ostream.inserters.arithmetic\]](http://wg21.link/ostream.inserters.arithmetic).

If I understand correctly, the example is non-normative text, making this an [editorial issue](https://github.com/cplusplus/draft/wiki/How-to-tell-if-an-issue-is-editorial).

jensmaurer:
Fixes #3238.

### DIFF
--- a/source/locales.tex
+++ b/source/locales.tex
@@ -178,12 +178,10 @@ and
 
 \pnum
 \begin{example}
-An iostream
-\tcode{operator<<}
-might be implemented as:\footnote{Note that in the call to
-\tcode{put}
-the stream is implicitly converted to an
-\tcode{ostreambuf_iterator<charT, traits>}.}
+An iostream \tcode{operator<<} might be implemented as:%
+\footnote{Note that in the call to \tcode{put},
+the stream is implicitly converted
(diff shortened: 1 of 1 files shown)


jensmaurer:
@KalleOlaviNiemitalo, does this look right?

PR #3244, opened 2019-09-25
[expr.prim.id] Fix immediate function id-expression requirement

random-internet-cat:
Reason for being editorial: It is probably not the committee's intent to require all programs to use "an id-expression that denotes an immediate function".

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -1270,8 +1270,11 @@ int j = sizeof(S::m + 42);      // OK
 \pnum
 An \grammarterm{id-expression}
 that denotes an immediate function\iref{dcl.constexpr}
-shall appear as a subexpression of an immediate invocation or
-in an immediate function context\iref{expr.const}.
+shall appear only
+\begin{itemize}
+\item as a subexpression of an immediate invocation, or
+\item in an immediate function context\iref{expr.const}.
+\end{itemize}
 
 \pnum
 An \grammarterm{id-expression}


W-E-Brown:
Without expressing any opinion re whether we want some such rewording, the proposed placement of "only" is suboptimal.  To say that something "shall only appear ..." implies, for example, that it mustn't cook, clean, play, tap-dance, or snore -- the only thing it can do is appear.

If we decide that we do want a wording adjustment hereabouts, please consider injecting "only" in two places within this sentence:  "... shall appear only as a subexpression of an immediate invocation or only in an immediate function context\iref{expr.const}."  I'm less unhappy with that, but the sentence is too long, with too many qualifying phrases, to move "only" to the sentence end as I'd truly prefer.

It would be easiest if we had a short bullet list here:  "... shall appear only:  (bullet a) as a subexpression of an immediate invocation, or (bullet b) in an immediate function context\iref{expr.const}."

source/expressions.tex
```diff
@@ -1270,7 +1270,7 @@
 \pnum
 An \grammarterm{id-expression}
 that denotes an immediate function\iref{dcl.constexpr}
-shall appear as a subexpression of an immediate invocation or
+shall only appear as a subexpression of an immediate invocation or
```
jensmaurer: I like Walter's suggestion of adding bullets and moving the only a little farther towards the end.

random-internet-cat:
Made the requested changes.

PR #3246, opened 2019-09-25
[index] Create first-letter subheadings for the general index.

jensmaurer:


jensmaurer:
Here's the new style:
idx2

jensmaurer:
This is rather a basis for discussion.

jensmaurer:
And here is a crazy "table of contents" added:
idx

jwakely:
The table of contents is great, I've wanted that for years.

jensmaurer:
Instead of the funny mini-ToC, I've now added the index letters to the left-hand PDF navigation column, as-if they were subsections. This looks substantially nicer. But maybe people want the mini-ToC anyway.

PR #3247, opened 2019-09-25
Change type-id to defining-type-id [temp.alias] p4

18:
An alias declaration contains no type-id, its a defining-type-id.


jensmaurer:
p2 has another appearance of _type-id_ that should be fixed.

jensmaurer:
(This needs a squash + force-push. Can be done when applied.)

PR #3248, opened 2019-09-25
[temp.spec] Make "specialization" a proper definition & adjust definition.

18:
*Specialization* is the second most common term that is defined and used in [temp]. The definition in [temp.spec] p4 should be italicized to make it a proper definition. Also, the phrase " function, or class member that is [...] explicitly specialized" is incorrect: templates are explicitly specialized, not classes, functions etc (in some cases non-templates too, but rarely).

Additionally, a *specialization* of a alias template (such as the use here: http://eel.is/c++draft/temp.alias#2) is meaningless for non-class types. Consider the following:
```
template<typename T>
using A = int;
```

A specialization is a class, function, variable, or class member. The type denoted (if you could even say that, since its replaced) by the simple-template-id `A<void>` is none of these.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -5699,8 +5699,9 @@ An instantiated template specialization can be either implicitly
 instantiated\iref{temp.inst} for a given argument list or be explicitly
 instantiated\iref{temp.explicit}.
 A \defn{specialization} is a class, variable, function, or class member that is either
-instantiated\iref{temp.inst} from a template or an
-explicit specialization\iref{temp.expl.spec} of a template.
+instantiated\iref{temp.inst} from a
+templated entity\iref{temp.pre} or an
+explicit specialization\iref{temp.expl.spec} of a templated entity.
 
 \pnum
 For a given template and a given set of
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -5698,8 +5698,9 @@ template<> int B<>::x = 1;              // specialize for \tcode{T == int}
 An instantiated template specialization can be either implicitly
 instantiated\iref{temp.inst} for a given argument list or be explicitly
 instantiated\iref{temp.explicit}.
-A \defn{specialization} is a type, variable, function, or class member that is either
-instantiated from a template or specialized for by an explicit specialization\iref{temp.expl.spec}.
+A \defn{specialization} is a class, variable, function, or class member that is either
+instantiated\iref{temp.inst} from a template or produced by an 
+explicit specialization\iref{temp.expl.spec} of a template.
 
 \pnum
 For a given template and a given set of
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -5699,8 +5699,7 @@ An instantiated template specialization can be either implicitly
 instantiated\iref{temp.inst} for a given argument list or be explicitly
 instantiated\iref{temp.explicit}.
 A \defn{specialization} is a class, variable, function, or class member that is either
-instantiated\iref{temp.inst} from a
-templated entity\iref{temp.pre} or an
+instantiated\iref{temp.inst} from a templated entity or an
 explicit specialization\iref{temp.expl.spec} of a templated entity.
 
 \pnum
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -5699,7 +5699,7 @@ An instantiated template specialization can be either implicitly
 instantiated\iref{temp.inst} for a given argument list or be explicitly
 instantiated\iref{temp.explicit}.
 A \defn{specialization} is a class, variable, function, or class member that is either
-instantiated\iref{temp.inst} from a template or produced by an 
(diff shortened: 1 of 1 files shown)


W-E-Brown:
1) I agree that we should consider italicizing "specialization" in this sentence.

2) I doubt we need the 2 consecutive prepositions "for by" -- how about "via" instead?

3) Most importantly, do we really consider "int" to be a specialization of "template<typename T> using A = int;"?  Yes, we obtain a type when an alias template is instantiated, but is that type in any sense a specialization?  Alias templates seem rather unlike function, class, and variable templates in this regard.

18:
Yeah, I had trouble thinking of a way to describe aliases, since instantiation produces no entity, rather, replacement occurs. Another option would be to stomp out all references to "alias specializations", and replace them with something to the effect of "template-id where the template-name denotes an alias template", or something similar.

source/templates.tex
```diff
@@ -5698,8 +5698,8 @@
 An instantiated template specialization can be either implicitly
 instantiated\iref{temp.inst} for a given argument list or be explicitly
 instantiated\iref{temp.explicit}.
-A specialization is a class, variable, function, or class member that is either
-instantiated or explicitly specialized\iref{temp.expl.spec}.
+A \defn{specialization} is a type, variable, function, or class member that is either
+instantiated from a template or specialized for by an explicit specialization\iref{temp.expl.spec}.
```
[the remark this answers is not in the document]
18: How about the change to how we specify explicit specializations? I think that is sufficiently editorial, since we know that templates are what we explicitly specialize.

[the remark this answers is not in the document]
jensmaurer: I think "explicitly specialized" (verb form) is used in [temp.expl.spec] sufficiently often that we don't need to work around the term.
Maybe "that is either instantiated from a template (xref) or produced by an explicitly specialization of a template (xref)."

jensmaurer:
The Travis CI check failed. Maybe there's a space at the end of a line or so. Please fix that.

source/templates.tex
```diff
@@ -5698,8 +5698,9 @@
 An instantiated template specialization can be either implicitly
 instantiated\iref{temp.inst} for a given argument list or be explicitly
 instantiated\iref{temp.explicit}.
-A specialization is a class, variable, function, or class member that is either
-instantiated or explicitly specialized\iref{temp.expl.spec}.
+A \defn{specialization} is a class, variable, function, or class member that is either
+instantiated\iref{temp.inst} from a template or produced by an 
+explicit specialization\iref{temp.expl.spec} of a template.
```
jensmaurer: I am having second thoughts about "produced by ...". In fact, an explicit specialization *is* a class etc. right there; it doesn't obviously "produce" one. (Oh, and "produce" sounds awkward anyway.)
maybe "... or an explicit specialization of a template."
@zygoloid ?

jensmaurer: Looks to me now.

zygoloid: Hmm, are the "from a template" additions correct? For cases like:

```
template<typename T> struct X {
  void f() {}
};
```

... is `X<int>::f()` instantiated from a template, or from a member of a template?

zygoloid: Also, "is either instantiated from a template or an explicit specialization of a template" is missing a verb. Or *what* an explicit specialization? (If this is "or is an explicit specialization", what does that mean for members of templates:

```
template<typename T> struct Y {
  struct Z {};
};
template<> struct Y<int>::Z {};
template<> struct Y<float> { struct Z {}; };
```

... which of `Y<int>::Z` and `Y<float>::Z` are specializations? Do we really mean that the former is a specialization and the latter is not?

18: @zygoloid I think both of these could be addressed by saying:

> A *specialization* is a class, variable, function, or class member that is either instantiated from a template, or a member thereof, or is an explicit specialization of a template, or member thereof.

This seems a little clunky though.

jensmaurer: "templated entity" to the rescue, it seems. (This is actually a defined term.)

18: Templated entity would work, good idea. Looks like the note http://eel.is/c++draft/temp#pre-8.note-1 is wrong though, since a variable is not an entity.

jensmaurer: A variable declares an object or reference, and both of these things are entities. [basic.pre].

18: Correct, but the variable itself is not

jensmaurer: Well, it was certainly intended that variable templates are also templated entities. Maybe we should simply say that a variable is an entity, then.

18: A variable template is a template though, so it is a templates entity anyways.

jensmaurer: So, we're good with "templated entity", and all we need is an updated pull request.

18: Updated

Issue #1899, opened 2018-01-13
[everywhere] Remove paragraph breaks in the middle of sentences


Eelis:
For example, consider this sentence that spans multiple paragraphs: https://github.com/cplusplus/draft/blob/master/source/basic.tex#L631

The reason I care about this is that cxxdraft-htmlgen supports linking to individual sentences (which has proved to be a popular feature), but it can't do this properly for sentences that span multiple paragraphs.

So to get links like http://eel.is/c++draft/basic.scope.declarative#2.sentence-1 to correctly select the whole sentence, I currently use a [sloppy incomplete patch](https://github.com/Eelis/draft/commit/5842bfa4399f786c731ab4cd69cb007ce56e262e) that removes tons of these inappropriate paragraph breaks. I could clean that patch up, make it more complete, and submit it, if this issue is deemed worthy of fixing.

I vaguely recall that @tkoeppe was already planning something along these lines a while ago, but I can't find an existing ticket.

tkoeppe:
Well, my goal was to make it possible to remove those breaks, but to actually make edits to remove the breaks would be a lot of work, and there are edge cases where the additional break improves line breaking, weirdly. (Though perhaps this has changed with @jensmaurer's improved widow/club code.)

I'd certainly entertain patches that do nothing but break removal and cause no visual diffs.

jensmaurer:
Just to clarify, the issue is the empty line for cases like the following:
```
In
(empty)
\begin{codeblock}
int j = 24;
[...]
```
Because from a LaTeX perspective, there is no difference with and without the empty line (a codeblock is a vertical environment, so the previous LaTeX paragraph ends anyway).

There might actually be an argument for source code readability to keep the empty line.
@zygoloid , any opinion here?

tkoeppe:
As a long-term user of LaTeX, it's precisely the readability aspect that makes me want to *get rid* of the double line breaks. Double line breaks read as "paragraph break", so they're very confusing.

Eelis:
I noticed that removing an empty line between text and a codeblock changes the layout of the text. For example, removing [this empty line](https://github.com/cplusplus/draft/blob/616405fb021f33841280407e95a7a44e1183a719/source/lex.tex#L1858) results in the following change:

![diff](https://user-images.githubusercontent.com/22357/35414384-9c2a6ad4-0222-11e8-9cd3-57cb2b71978a.png)

(If you don't see it, look at the right side.)

[...]

PR #3253, opened 2019-09-30
[everywhere] Remove inappropriate paragraph breaks.

Eelis:
Addresses #1899.

Maybe not quite all of them yet, but a LOT.

[PDF diffs](http://eel.is/parabreakdiffs.pdf)

Most of these inappropriate paragraph breaks occur in the middle of sentences, which breaks sentence linking with cxxdraft-htmlgen, which is why I use this patch for eel.is/c++draft.


tkoeppe:
Agreed. These newlines are a holdover from before we got our vertical spacing sorted out, back when they were necessary (to add negative space (*sic*!)). Ever since, we have been wanting to clean them up, but I never embarked on a complete revision, only doing this as part of other edits.

Thanks a lot!

PR #3261, opened 2019-10-02
[alg.search] search_n. Use consistent wording with search

andreyfe1:
Hi,
For the `search_n` algorithm the Standard says:
> Returns: The first iterator i in the range [first, last-count) such that for every non-negative integer n less than count the following corresponding conditions hold: (i + n) == value, pred((i+ n),value) != false. Returns last if no such iterator is found.

It seems that if `n>=0` and `count <=0` we don't have a chance to apply a predicate. So, we won't find "such iterator". According to this logic `last` should be returned. However, [libcxx ](https://github.com/llvm-mirror/libcxx/blob/master/include/algorithm) and [Microsoft's C++ Standard Library](https://github.com/microsoft/STL/blob/master/stl/inc/algorithm) returns `first`. According to [the issue](https://github.com/microsoft/STL/issues/148) there was a different interpretation of `search_n` wording.

Moreover, `search` returns `first` in the same case. I hope it's just an editorial issue.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -3998,7 +3998,8 @@ The first iterator \tcode{i} in the range \range{first}{last-count}
 such that for every non-negative integer \tcode{n} less than \tcode{count}
 the following corresponding conditions hold:
 \tcode{*(i + n) == value, pred(*(i + n),value) != false}.
-Returns \tcode{last} if no such iterator is found.
+Returns \tcode{first} if \tcode{count <= 0},
+otherwise returns \tcode{last} if no such iterator is found.
 
 \pnum
 \complexity


jensmaurer:
Please squash and force-push your commits. Also, prefix your commit comment with "[alg.search]".

That said, there is explicit normative wording for "search" that we get "first1" for an empty second range. This wording is missing from "search_n", although the rest of the description has the same structure. I can't see how this issue would be editorial.

capatober:
The patch is only for clarification what should be returned from "search_n". I see "first" as a return value is obvious for standard libraries, so, let's document it anyhow to avoid misinterpretation

jensmaurer:
"obvious for standard libraries": Apparently not, otherwise https://github.com/microsoft/STL/issues/148 wouldn't be tagged "LWG issue needed".

source/algorithms.tex
```diff
@@ -3998,7 +3998,8 @@
 such that for every non-negative integer \tcode{n} less than \tcode{count}
 the following corresponding conditions hold:
 \tcode{*(i + n) == value, pred(*(i + n),value) != false}.
-Returns \tcode{last} if no such iterator is found.
+Returns \tcode{first} if \tcode{count <= 0},
```
CaseyCarter: This normatively duplicates the stated behavior of the first sentence. If we want to clarify the meaning of the first sentence for this case, then this should be a note.

jensmaurer: And why is this extra provision present for `search` (without `_n`)? Given the similar structure, I'd expect the "normative duplication" would applie to both.

CaseyCarter: I consider it a defect that the redundant provision is present for `std::search`. Notably none of `std::ranges::search`, `std::search_n`, nor `std::ranges::search_n` have this defect. Spreading the provision from one member of this set to a second does not harmonize the wording of all four.

jensmaurer: Thanks. So, we should have an LWG issue to remove the duplication from `search` and maybe add a note everywhere talking about the "empty range" case. Unless @zygoloid this approach would be editorial.

PR #3270, opened 2019-10-03
[std] Align all comments to multiples of four

### LINKED ISSUE #3131: In Latex sources, do columns start counting at 0 or 1?
Our coding style [guidelines](https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines?) say that the starting columns for comments in examples should be multiples of 4, but starting from where?  If we start counting at 1, the examples' comments in [diff.cpp17.dcl.dcl] start on column 25, suggesting that at least those comments start counting from 0, while other comments (including the ones I've edits) assume we start counting at 1.  Which is correct?  Example from compatibility.tex:
```
  void f() {}           // ill-formed; previously well-formed
0         1         2
0123456780912345678091234 does the above comment start on column 24?
1234567809123456780912345 or on column 25?
```
FWIW, the comments in the guideline's own Latex example text start on column 48 when counting from 1, suggesting that whoever wrote the guidelines was thinking they should start at 1:
`\begin{note}                                   %% own lines if possible
`

The guidelines should be fixed to specify, and the examples fixed to adhere to them.

jensmaurer:
Fixes #3131.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -8578,182 +8578,164 @@ namespace std {
            class BinaryOperation1, class BinaryOperation2>
     T inner_product(InputIterator1 first1, InputIterator1 last1,
                     InputIterator2 first2, T init,
-                    BinaryOperation1 binary_op1,
-                    BinaryOperation2 binary_op2);
+                    BinaryOperation1 binary_op1, BinaryOperation2 binary_op2);
 
   // \ref{transform.reduce}, transform reduce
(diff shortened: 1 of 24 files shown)


tkoeppe:
Does this change correctly take into account that the things left of the comment may be macros?

jensmaurer:
The script-based checking ignores all lines with @ signs left of the comment, as a measure of caution.

PR #3285, opened 2019-10-09
[std] Consistently use 'comparison function'

### LINKED ISSUE #3284: missing definition for "comparison operator"
In [dcl.fct.def.default]/1.1, we find that an explicitly-defaulted function shall

> be a special member function or a comparison operator (7.6.8, 7.6.9, 7.6.10)

but what is a "comparison operator"? The cross-references point to some [expr] subsections that have pretty much nothing to do with it. (Those are the places where we *don't* go when there is a comparison operator to use!)

Similarly [class.compare.default] talks about "A defaulted comparison operator function" with the same list of irrelevant cross-references, and likewise in [except.spec]p11. 

We should add a definition of "comparison operator" as one of the seven intended overloaded operator functions (`operator<`, `operator<=`, `operator>`, `operator<=`, `operator==`, `operator!=`, `operator<=>`), probably in [over.binary], and change all the above cross-references to point at that instead. I think the existing list of cross-references is sufficiently suggestive that this can be done editorially.

While we're here, we should consider renaming the term from "comparison operator" to "comparison function", matching "conversion function" and clarifying that we mean the function case and not the built-in operator case.

jensmaurer:
instead of 'comparison operator function', and
define the term in [over.binary].

Fixes #3284.

### DIFF
--- a/source/overloading.tex
+++ b/source/overloading.tex
@@ -3506,6 +3506,18 @@ If both forms of the operator function have been declared,
 the rules in~\ref{over.match.oper} determine which, if any, interpretation is
 used.
 
+\pnum
+An \defnadj{equality}{operator function} is an operator function
+for an equality operator\iref{expr.eq}.
+A \defnadj{relational}{operator function} is an operator function
+for a relational operator\iref{expr.rel}.
+A \defnadj{three-way comparison}{operator function} is an operator function
+for the three-way comparison operator\iref{expr.spaceship}.
+A \defnadj{comparison}{operator function} is
+an equality operator function,
+a relational operator function, or
+a three-way comparison operator function.
+
 \rSec2[over.ass]{Assignment}
 \indextext{assignment operator!overloaded}%
 \indextext{overloading!assignment operator}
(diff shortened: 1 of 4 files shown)


jensmaurer:
We're using "relational operator function", too, and quite a few places had to be fixed to avoid the term "comparison operator function". I'd mildly favor keeping the term "comparison operator function", since we're actually talking about a certain subset of operator functions here.
Adding a definition in [over.binary] is a good idea; maybe we should also define "relational operator function" and "equality operator function" there.

source/overloading.tex
```diff
@@ -3506,6 +3506,17 @@
 the rules in~\ref{over.match.oper} determine which, if any, interpretation is
 used.
 
+\pnum
+A \defn{comparison function} is an operator function
+for one of the following binary operators:
+\tcode{<},
+\tcode{<=},
+\tcode{>},
+\tcode{<=},
```
Eelis: This is the same one as 2 lines up :)

CaseyCarter: FWIW, the canonical ordering used for these operators in the library is:
* `==`
* `!=`
* `<`
* `>`
* `<=`
* `>=`

jensmaurer:
I think the term should be "comparison operator function"; it's a special kind of operator function. "comparison function" sounds a bit like it would include `memcmp`.

PR #3291, opened 2019-10-11
No indefinite articles before grammatical productions

18:
Pretty simple grammatical fix

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -4337,7 +4337,7 @@ parenthesized initializer shall be a single expression.
 \pnum
 \indextext{initialization!copy}%
 \indextext{initialization!direct}%
-The initialization that occurs in the \tcode{=} form of a
+The initialization that occurs in the \tcode{=} form of
 \grammarterm{brace-or-equal-initializer} or
 \grammarterm{condition}\iref{stmt.select},
 as well as in argument passing, function return,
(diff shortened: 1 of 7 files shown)


tkoeppe:
I don't think the status quo is wrong here. "A form of if statement" seems like a correct way to refer to the different forms.

18:
I feel that an article here is needed, since we are referring to a construct, and it reads a lot smoother, 

However, if the consensus if that its fine, then we should make the use consistent.

jensmaurer:
It seems to me the status quo is fine here. "In the first form of mutiny, all participants are incarcerated. In the second form of mutiny, all participants are beheaded." sounds reasonable.

18:
@jensmaurer Alright. Should we atleast make the use consistent? For example, http://eel.is/c++draft/dcl.dcl#dcl.init-16.5 uses an indefinite article before the noun, but http://eel.is/c++draft/dcl.dcl#dcl.init-1.sentence-2 doesn't.

jensmaurer:
So, this turned into a consistency question, and the suggested pull request here actually does not address the consistency issue to the extend we know about it.

18:
@jensmaurer I'll change the PR accordingly.

zygoloid:
"the second form of `if` statement" is talking about the general category of `if` statements, which have multiple forms, so omitting the indefinite article there seems correct (we're not talking about a single `if` statement that has multiple forms, so "an" is inappropriate). The cases affected by the revised pull request are talking about a specific instance of a grammar production, and for those cases, the indefinite article is appropriate.

I don't think there's a systematic problem here, and I'm unconvinced that any of the changes so far are improvements.

PR #3299, opened 2019-10-15
[range.subrange.access] fix typo

jwakely:


source/ranges.tex
```diff
@@ -1446,7 +1446,7 @@
 \effects
 \begin{itemize}
 \item If \tcode{StoreSize} is \tcode{true}, equivalent to: \tcode{return size_;}
-\item Otherwise, equivalent to: \tcode{return \placeholdernc{make-unsigned-like-t}(end_ - begin_);}
+\item Otherwise, equivalent to: \tcode{return \placeholdernc{make-unsigned-like}(end_ - begin_);}
```
CaseyCarter: This is correct in [P1523R1](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1523r1.pdf), but it's hard for me to blame the Editors when LWG moves a paper that adds both `make-unsigned-like(X)` and  `make-unsigned-like-t(X)`. I wonder if it would help to distinguish the two if `make-unsigned-like-t` used angle brackets, i.e., if it was an exposition-only alias template instead of an exposition-only macro metafunction?

CaseyCarter: (To be clear: I'm not requesting a change in the PR, so much as observing a possible issue and wondering if others agree with my proposed fix before I go to the trouble to submit a PR with such a fix.)

jwakely: Yes, the difference is too subtle.

We could also (or instead) rename _make-unsigned-like_ to _as-unsigned-like_ or _to-unsigned-like_.

CaseyCarter: Filed #3300.

CaseyCarter: (Feel free to resolve this conversation, I can't.)

PR #3312, opened 2019-10-21
[basic.scope.class] Reinstate a qualification that was lost in 0e26279b88c3b8b0a09babdeec8418d383f07419.

tkoeppe:
Without the introductory sentence that was deleted by that commit, we need to say explicitly that we are talking about a declaration _in a class_.

See also discussion on https://github.com/cplusplus/draft/pull/1137.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -1176,7 +1176,7 @@ the same name whose scope extends to or past the end of the member
 function's class.
 
 \pnum
-The potential scope of a declaration that extends to or past the
+The potential scope of a declaration in a class that extends to or past the
 end of a class definition also extends to the regions defined by its
 member definitions, even if the members are defined lexically outside
 the class (this includes static data member definitions, nested class


languagelawyer:
Why we don't need similar fix in the previous paragraph?

tkoeppe:
The next paragraph is just an example, isn't it?

languagelawyer:
The previous, not the next. The third paragraph also speaks about declarations whose scope extends to or past the end of a class definition. (Also the word "definition" is missing at the end).

tkoeppe:
Ah, sorry, I see. (I had somehow misread the original message as "fourth and fifth", sorry!) So for the previous paragraph, we can say "hides a declaration of the same name <ins>in that class</ins>"?

The "definition" was also absent from the original introductory paragraph, do you think it matters?

languagelawyer:
> I had somehow misread the original message as "fourth and fifth", sorry!

It **was** originally saying "fourht and fifth", this was my mistake, sorry.

> So for the previous paragraph, we can say "hides a declaration of the same name in that class"?
> The "definition" was also absent from the original introductory paragraph, do you think it matters?

Let me redirect this to @zygoloid

zygoloid:
@languagelawyer For the third paragraph, do you have an example in mind? I think the slightly broader wording we have here is equivalent to the more-restrictive version. In fact, paragraph 3 seems to be entirely redundant, since it follows from [basic.lookup.unqual]p1 and p8.

It's also unlikely to be worth spending much time on editorial improvements here, since http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1787r2.html is replacing all this wording and is likely to land next meeting (or maybe the one after, depending on how fast CWG gets through the NB comments).

languagelawyer:
@zygoloid No, I don't have an example. I doubt there is an example even for the fourth paragraph.
I've just pointed to the fact that normative wording has been changed without CWG review/approval.

PR #3327, opened 2019-10-25
[syncstream] Add default template parameters for basic_syncbuf and basic_osyncstream

jensmaurer:
Fixes cplusplus/nbballot#344.
Fixes cplusplus/nbballot#345.
Fixes cplusplus/nbballot#346.

The default template arguments were shown for the declarations in &lt;iosfwd>, thus this seems to be an editorial fix.
While we can't repeat default template arguments in C++, be consistent with the existing specification practice for the other iostreams classes.

Do not squash the commits; each addresses an individual NB comment.

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -10859,13 +10859,13 @@ calls
 #include <ostream>  // see \ref{ostream.syn}
 
 namespace std {
-  template<class charT, class traits, class Allocator>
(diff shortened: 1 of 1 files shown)


jensmaurer:
Editorial meeting: Accepted.

Also file an LWG issue to make clear which headers make &lt;ios> available. Once that's done, remove the default template arguments, because they're implied and redundant from &lt;iosfwd>.

PR #3335, opened 2019-10-26
[atomics.ref.int,atomics.types.int] Clarify notes on atomic/_ref<bool…

jensmaurer:
…> specializations.

Fixes NB US 357 (C++20 CD)

Fixes cplusplus/nbballot#353.

### DIFF
--- a/source/atomics.tex
+++ b/source/atomics.tex
@@ -1140,7 +1140,8 @@ For each such type \tcode{\placeholder{integral}},
 the specialization \tcode{atomic_ref<\placeholder{integral}>} provides
 additional atomic operations appropriate to integral types.
 \begin{note}
-For the specialization \tcode{atomic_ref<bool>}, see \ref{atomics.ref.generic}.
+The specialization \tcode{atomic_ref<bool>}
+uses the primary template\iref{atomics.ref.generic}.
 \end{note}
 
 \begin{codeblock}
@@ -2102,7 +2103,8 @@ and any other types needed by the typedefs in the header \libheaderref{cstdint}.
 For each such type \tcode{\placeholder{integral}}, the specialization
 \tcode{atomic<\placeholder{integral}>} provides additional atomic operations appropriate to integral types.
 \begin{note}
-For the specialization \tcode{atomic<bool>}, see \ref{atomics.types.generic}.
+The specialization \tcode{atomic<bool>}
+uses the primary template\iref{atomics.types.generic}.
 \end{note}
 
 \begin{codeblock}


source/atomics.tex
```diff
@@ -1140,7 +1140,8 @@
 the specialization \tcode{atomic_ref<\placeholder{integral}>} provides
 additional atomic operations appropriate to integral types.
 \begin{note}
-For the specialization \tcode{atomic_ref<bool>}, see \ref{atomics.ref.generic}.
+The specialization \tcode{atomic_ref<bool>}
+is based on the primary template\iref{atomics.ref.generic}.
```
CaseyCarter: I don't know what it means for a specialization to be "based on" a template. Do we mean "`atomic_ref<bool>` is an instantiation of the primary template"?

jensmaurer: Well, claiming that atomic_ref<bool> be an instantiation of the primary template would be overspecification. After all, the library could choose to provide an explicit specialization that is maybe faster / smaller than when instantiating the primary template. So, the vagueness of "based" in the note seems justified.

CaseyCarter: That would fall under "as-if": a user can't observe whether or not a particular specialization is instantiated from a particular template. The important thing is that it behaves as if instantiated from the primary template, which is what I think this note intends to communicate.

CaseyCarter: Similarly, implementations will likely implement the many explicit specializations specified in this subclause as a single partial specialization (as we do with `atomic`) since the difference is unobservable.

jensmaurer: Right, so why should the note claim more by saying "instantiation of the primary template" when no such requirement actually exists?

CaseyCarter: I don't see how the distinction you're trying to draw is useful when describing the behavior of the Standard Library. We don't specify how the library is implemented, we specify how the library behaves. `atomic_ref<bool>` must behave as if it is an instantiation of the primary template because we don't specify any different behavior for it.

This note exists for two reasons: (1) to call attention to the fact that `bool` isn't in the list of types here despite the fact that `bool` is an integral type and the subclause is titled "Specializations for integral types", and (2) to clarify that `atomic_ref<bool>` is nevertheless a valid specialization that behaves as if instantiated from the primary template. The NB comment is presumably the result of us not expressing (2) clearly, it seems to me the best fix is to be more explicit rather than try to invent euphemisms like "specialization is based on template".

jensmaurer: The editorial meeting resolved to "uses the primary template". Does that sound good enough?

CaseyCarter: I suppose "uses" is less likely to be misinterpreted to mean something different from "is an instantiation of" than is "is based on", but I believe it would be better to say what we mean rather than hint at it. Having said that, I will agree that we've invested plenty of time discussing this note and yield to the consensus of the editorial meeting.

PR #3336, opened 2019-10-26
[iterator.requirements.general] Define 'reachable from' to avoid conf…

jensmaurer:
…usion

with 'reachable' used finding declarations in modules.

Fixes NB US 258 (C++20 CD)

Fixes #cplusplus/nbballot#254.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -613,7 +613,7 @@ and up to but not including the element, if any, pointed to by
 the first iterator \tcode{j} such that \tcode{j == s}.
 
 \pnum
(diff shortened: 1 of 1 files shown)


W-E-Brown:
As a drive-by, could we replace, in this definition, "called" by "termed" or "known as"?  Or perhaps simply strike "called"?

jensmaurer:
@W-E-Brown , since this is the resolution of an NB comment, I'd prefer to stick to the topic of the comment as closely as possible for this change.

PR #3337, opened 2019-10-26
[iterators] Improve indexing of defined terms.

jensmaurer:


W-E-Brown:
As a drive-by, could we replace, in these revised definitions, "called" by "termed" or "known as"?

zygoloid:
Filed #3515 for the change that @W-E-Brown requested; we should make this change more broadly rather than as a drive-by for just these definitions.

PR #3340, opened 2019-10-26
[temp.constr.decl] Missing case when constraints are associated with …

jensmaurer:
…a declaration.

Fixes NB US 108 (C++20 CD)

Fixes cplusplus/nbballot#107.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -1618,8 +1618,8 @@ an expression:
 
 \pnum
 Constraints can also be associated with a declaration through the use of
-\grammarterm{type-constraint}{s} in a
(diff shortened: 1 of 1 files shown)


jensmaurer:
Editorial meeting: Remove "in a function declaration" (cf. lambda-declaration). Have Hubert and/or Andrew Sutton double-check. Accepted.

tkoeppe:
@hubert-reinterpretcast: Could you kindly take a look?

hubert-reinterpretcast:
Reading the "can" as "could", I think the change is okay. Note that the removal of "in a function declaration" (which I believe could only help) has not happened yet.

PR #3346, opened 2019-10-28
[dcl.inline] Function parameters shouldn't be inline

18:
It is nonsensical for function parameters to be declared `inline`, as the wording in [[dcl.inline]](http://eel.is/c++draft/dcl.inline) would be meaningless for function parameter. A function parameter does is not declared in block scope, so http://eel.is/c++draft/dcl.inline#5 doesn't apply.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -1044,7 +1044,8 @@ constinit const char * d = f(false);    // ill-formed
 
 \pnum
 The \tcode{inline} specifier shall be applied only to the declaration
-of a variable or function.
+of a variable or function. The \tcode{inline} specifier shall not be used in the declaration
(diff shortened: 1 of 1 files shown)


jensmaurer:
Maybe changing p5 to say "on a function parameter or block scope declaration" is better? Is this editorial to start with?

18:
Function parameters have no linkage, so it being an inline variable would have no effect whatsoever. 

If it is decided that this is not editorial, I can submit a DR.

jensmaurer:
Editorial meeting: Please submit a core issue. Other uses of "block scope" should probably also include to "function parameter scope".

18:
Submitted as issue 2441

PR #3355, opened 2019-10-30
[range.filter.overview] Avoid double-negative

CaseyCarter:


CaseyCarter:
> Or "`filter_view` presents a view of an underlying sequence, containing only the elements that satisfy a predicate."

I don't hate it, although it seems unfortunate to burn a whole clause ("presents a view of an underlying sequence") to redundantly describe what all range adaptors do. I'm happy for someone to pick either wording, push into the PR as needed, and merge.

PR #3361, opened 2019-11-01
[class.copy.assign] Remove semicolon in 'of the form' phrase.

jensmaurer:
This improves consistency with other such phrases.

Fixes late comment CH 02.

Fixes cplusplus/nbballot#376

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -1919,7 +1919,7 @@ struct S {
 \pnum
 The implicitly-declared move assignment operator for a class \tcode{X} will have the form
 \begin{codeblock}
-X& X::operator=(X&&);
+X& X::operator=(X&&)
 \end{codeblock}
 
 \pnum


source/classes.tex
```diff
@@ -1919,7 +1919,7 @@
 \pnum
 The implicitly-declared move assignment operator for a class \tcode{X} will have the form
 \begin{codeblock}
-X& X::operator=(X&&);
+X& X::operator=(X&&)
```
tkoeppe: This is awkward, as it now leaves the sentence unfinished.

Looking at the next paragraph, I see that there we use a _text_semicolon to structure the sentence. Do we have similar options here? Otherwise add a colon after "have the form"?

zygoloid: This is also the case for [class.copy.ctor]p7 and p9, [class.copy.assign]p2, and [class.dtor]p3.

The idea of this change (and the request of the NB comment) is just to get to somewhere consistent. We should tackle any other changes / improvements here as separate editorial issues.

PR #3363, opened 2019-11-02
[thread.jthread.class] Namespace not closed

seb-mtl:
Minor issue, my parser failed to parse the jthread codeblock because the std namespace is not closed.

### DIFF
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -1523,6 +1523,7 @@ namespace std {
   private:
     stop_source ssource;        // \expos
   };
+}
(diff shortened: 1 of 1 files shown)


johelegp:
Nice find.
According to https://github.com/cplusplus/draft/wiki/How-to-submit-an-editorial-issue, you need to prefix the PR title and commit message with [thread.jthread.class].

seb-mtl:
Hi, any updates on this? It's a minor fix

PR #3366, opened 2019-11-03
Small improvements

oss-assist:
Removed the silly message at the beginning of the specification, leaving only the sentence "This is an early draft." which transmits a lot more confidence to the ones who would read it.
Since even compilers use this document to implement early features of C++20, it's not that much incorrect.
Actually, the only noticeable incorrect message in the whole specification is this beginning notice.

### DIFF
--- a/source/cover-wd.tex
+++ b/source/cover-wd.tex
@@ -25,7 +25,5 @@
 Working Draft, Standard for Programming Language \Cpp{}}
 \end{center}
 \vfill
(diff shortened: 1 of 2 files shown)


jwakely:
This does no harm, I object to removing it.

jwakely:
Your new commit is wrong, `import` and `module` are in Table 4 [tab:lex.name.special] and are not keywords.

We don't want this so I'm closing it.

PR #3370, opened 2019-11-04
[std] Switch grammar to  \sffamily \itshape

### LINKED ISSUE #323: grammar terms are indistinguishable from defined terms
This issues was raised by Gaby - that you can't tell definitions from grammar terms, because both are _italics_.  Maybe use **_bold italics**_ for one of those - like the definition?

jensmaurer:
Fixes #323.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -5740,7 +5740,6 @@ D d(10);
 \end{codeblock}
 \end{example}
 \begin{note}
-\setlength{\emergencystretch}{1em}
 The initialization
 performed by each \grammarterm{mem-initializer}
 constitutes a full-expres\-sion\iref{intro.execution}.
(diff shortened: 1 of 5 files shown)


jensmaurer:
Also do a visual diffpdf.

jensmaurer:
Fix 9.5.4 promise-comstructor-arguments formatting, rename to $A$.

Eelis:
Does this patch preserve the regex grammar typesetting in [re.grammar]? Per https://github.com/cplusplus/draft/pull/3193, the grammar typesetting in that subclause should be consistent with the regex grammar typesetting in the upstream ECMAScript standard.

jensmaurer:
@Eelis: Yes, I've tried hard.

jensmaurer:
Editorial meeting: @tkoeppe to investigate other font choices in the next days.

jensmaurer:
@tkoeppe , any results from your investigation?

tkoeppe:
Yes, the various alternatives I had considered all don't really look any better. Give me another day or two to be sure, but we'll probably keep the initial proposal.

tkoeppe:
Let's proceed with what we have. I haven't found any alternatives that are clearly superior. DejaVu Sans is quite nice and has a condensed version (and needs to be scaled to 80%), but the difference with Latin Modern's version is rather minor. The ease of setup of the current solution is compelling, and we can iterate on minor variations later on.

PR #3435, opened 2019-11-11
[expr.sizeof] Replace inappropriate \term{n} with $n$.

Eelis:
I can't find other uses of `\term` for single-letter variables/placeholders; most places appear to use math-mode variables instead.


Eelis:
Yeah, https://github.com/cplusplus/draft/issues/329 tracks the more general problem of bad `\term`s. This pull request just fixes one particular kind of inappropriate use of `\term`.

Issue #3127, opened 2019-08-05
Indentation used in Latex for \items


burblebee:
Our guidelines[https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines?]()  say to use 2 space indentation for \items, which I find to be very helpful: it's clear to see where the bullets start and end, and when the bullets are nested.  When we edit old text which doesn't use this indentation however, we tend to stick with the old style for consistency.

Can we do a global whitespace change to fix the old formatting style so that we can follow the guidelines when applying edits?

Eelis:
(#1899 also concerns a global whitespace policy change involving newlines.)

jensmaurer:
Editorial teleconference: No consensus.

jensmaurer:
See also #2110.

jensmaurer:
For long \item, add an empty line in front and after.
Do not indent \item. Break after \item unless all of the bullets are very short.
Keep it as rough guidance, don't mass-update the source code.
Update the formatting wiki write-up accordingly.

johelegp:
> Update the formatting wiki write-up accordingly.

Please, do.

burblebee:
> > Update the formatting wiki write-up accordingly.
> > 
> Please, do.

Yes please!  It's not clear to me what is expected for the long items - fixing the wiki would make this clear.  From "For long \item, add an empty line in front and after", we'll have empty lines after the `\begin{itemize}` and before the `\end{itemize}`, often resulting in a `\begin{itemize}` or `\end{itemize}` statement having an empty line both before and after it which seems odd, is that what we want?  Perhaps we should instead have an empty line before the `\begin{itemize}` (except when starting a paragraph) and after the `\end{itemize}`?  Example:
```
\pnum
Blah blah blah...
Blah blah:

\begin{itemize}
\item
A multi-line text.

\item
Another multi-line text.
\end{itemize}

\pnum
Blah blah blah...
```

Eelis:
> Perhaps we should instead have an empty line before the \begin{itemize} (except when starting a paragraph) and after the \end{itemize}

When the list is part of a sentence, inserting empty lines around the list breaks the sentence, which breaks sentence linking on eel.is/c++draft. See https://github.com/cplusplus/draft/issues/1899

PR #3441, opened 2019-11-13
 P1907R1 Inconsistencies with non-type template parameters

### LINKED ISSUE #3401: [2019-11 CWG Motion 8] P1907R1 Inconsistencies with non-type template parameters

zygoloid:
Fixes #3401.

Also fixes NB US 092, US 093, US 100, US 102, US 114 (C++20 CD).

Fixes cplusplus/nbballot#91
Fixes cplusplus/nbballot#92
Fixes cplusplus/nbballot#99
Fixes cplusplus/nbballot#101
Fixes cplusplus/nbballot#113

Fixes cplusplus/papers#653

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -413,9 +413,9 @@ it is an lvalue and has type \tcode{const T}\iref{expr.prim.id.unqual}.
 \begin{example}
 \begin{codeblock}
 using X = int;
-struct A { friend bool operator==(const A&, const A&) = default; };
+struct A {};
 template<const X& x, int i, A a> void f() {
-  i++;                          // error: change of template-parameter value
+  i++;                          // error: change of \grammarterm{template-parameter} value
 
   &x;                           // OK
   &i;                           // error: address of non-reference template-parameter
@@ -428,16 +428,17 @@ template<const X& x, int i, A a> void f() {
 \end{example}
 
 \pnum
+\begin{note}
 A non-type
 \grammarterm{template-parameter}
-shall not be declared to have floating-point or void type.
+cannot be declared to have type \cv{} \tcode{void}.
 \begin{example}
 \begin{codeblock}
-template<double d> class X;     // error
-template<double* pd> class Y;   // OK
-template<double& rd> class Z;   // OK
+template<void v> class X;       // error
+template<void* pv> class Y;     // OK
 \end{codeblock}
 \end{example}
+\end{note}
 
 \pnum
 A non-type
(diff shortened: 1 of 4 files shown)


jensmaurer:
Augmented initial comment of pull request (not the commit message itself) to link and auto-close the respective NB comment issues.

source/templates.tex
```diff
@@ -367,28 +367,31 @@
...
+\item an lvalue reference type, or
+\item a literal class type with the following properties:
+\begin{itemize}
+\item
+all base classes and non-static data members are public and non-mutable and
+\item
+the types of all bases classes and non-static data members are
+structural types or (possibly multi-dimensional) array thereof.
+\end{itemize}
+\end{itemize}
```
johelegp: I think this is missing some indentation according to https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#latex-source-code-formatting.

source/templates.tex
```diff
@@ -1898,25 +1892,68 @@
...
+
+\item
+their corresponding non-type \grammarterm{template-argument}{s}
+are template-argument-equivalent (see below)
+after conversion to the type of the \grammarterm{template-parameter}, and
+
+\item
+their corresponding template \grammarterm{template-argument}{s}
+refer to the same template.
+\end{itemize}
```
johelegp: Items should be indented two spaces.

source/templates.tex
```diff
@@ -1898,25 +1892,68 @@
...
+
+\item
+they are of union type and either
+they both have no active member or
+they have the same active member and their active members are template-argument-equivalent, or
+
+\item
+they are of class type and
+their corresponding direct subobjects and reference members are template-argument-equivalent.
 \end{itemize}
```
johelegp: Same here. The long items call for two-space indentation.

That said, the separation of `\item`s by blank lines looks pretty nice, so maybe it's okay. Maybe a new entry to https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#latex-source-code-formatting is needed. I think this style is novel, as I don't remember seeing it before.

source/templates.tex
jensmaurer: We're not yet sure what to do with `\item`; see #3127. Thus, the formatting used here is as good as (most) any.

source/templates.tex
johelegp: The format is actually in accordance to the [new guidelines](https://github.com/cplusplus/draft/issues/3127#issuecomment-549577948), they're just not in the wiki yet. Disregard my review.

zygoloid:
> Please increase the feature-test macro `__cpp_nontype_template_args`.

We absolutely should bump some feature-test macro here, but that is neither requested by the wording paper nor an editorial change. (Also, is that the right one? Should we bump `__cpp_nontype_template_parameter_class` instead? Or maybe both?) If SG10 and CWG are happy, I think I could be OK with making the change editorially anyway, but it seems like a stretch.

I've started a thread on the relevant reflectors.

jensmaurer:
We're also editorially bumping __cplusplus, so bumping a feature-test macro seems similarly editorial to me. Having SG10 and CWG buy-in is good, though.

jensmaurer:
So, the actual changes are good, but I continue to believe we should bump a feature-test macro. I agree that removing one (as suggested on the reflector) is a bit beyond editorial.

zygoloid:
@BRevzin Can you take a look and confirm you consider the feature test macro changes to be acceptable as an editorial change?

BRevzin:
@zygoloid Yep. There were no objections on either list. Looks good to me.

Issue #3271, opened 2019-10-03
Upper-case vs Lower-case for expressions


BRevzin:
Throughout the language part of the standard, arbitrary expressions are introduced in lower-case, .e.g:

> The set of _potential results_ of an expression `e` is defined as follows: 

And arbitrary types are introduced in upper-case, e.g.:

> A class type `T` must be complete if:

Both of those are from [basic.def.odr], but it's consistent throughout the language section. 

<hr />

Much of the new library wording from Ranges breaks this convention by introducing expression in upper-case in all the definitions of the CPOs. For instance, from [range.access.begin]:

> The expression `ranges​::​​begin(E)` for some subexpression `E` is expression-equivalent to: 

Is there a particular reason why the Ranges wording uses capital letters for expressions instead of lower-case?

jensmaurer:
[algorithms] uses "Let $E$ ..." (using E as a meta-variable for an expression) pervasively.

@CaseyCarter, any opinion here?

CaseyCarter:
IIRC [algorithms] inconsistently used both `\tcode{E}` and `$E$` as expression placeholders before applying P0896. We chose to use the former as precedent for the placeholder in the CPO specifications. The change to use `$E$` consistently throughout [algorithms] happened in San Diego when I was integrating all of the range algorithm specifications into the `std` flavors.

I have a mild preference for `$E$` - whose slanted font makes it "clearly" a meta-syntactic variable - over `e` - which is itself valid syntax for a C++ expression - as a placeholder, but I'll agree that any consistent style would be an improvement over what's in the working draft right now.

jensmaurer:
The general idea for `\tcode` vs. (e.g.) math formatting is that the former could be valid source code, given the right environment, whereas the latter is used for meta-variables in other situations, such as "atomic operations" or similar. So, using `$E$` in the [algorithms] section for stuff that looks like valid code snippets is questionable. However, using lowercase "e" for the CPO specification (in harmony with the core language section) might make sense.

jensmaurer:
Editorial meeting:
- Do we want a consistent capitalization convention for the core language part, for types vs. expressions.
- Do we want to apply this to the library part as well?
- What should the convention be?

tkoeppe:
![image](https://user-images.githubusercontent.com/6378233/68156400-aa814100-ff43-11e9-9052-47f9dfbbbbbc.png)

jensmaurer:
[...]

Issue #3139, opened 2019-08-06
[std] Use of \placeholder for atomic operations meta-variables


jensmaurer:
Our current (undocumented) guidance is
- `\tcode{foo}` is used for stuff that may occur in source code, given a suitable environment.
- `\placeholder` is used for code meta-variables, e.g. exposition-only types and concepts
- Other meta-variables and indexing use math mode: `$\tcode{T}_i$`

First, we should document this policy. Second, we should also apply it to the atomics meta-variables when constructing all sorts of partial orders throughout [basic] and [atomics]. That is, these should change from `\placeholder` to `$math mode$`.

jensmaurer:
Editorial teleconference: \placeholder is a variable to be replaced by source code. Add examples. Agreed with the changes in [basic] and [atomics] for "atomic operation A" and "memory location M" etc.

jensmaurer:
Rules document in wiki.

jensmaurer:
Uh, using `$M$` for the memory model placeholders has an unfortunate side-effect: We lose the italics correction:
Before:
![comma-before](https://user-images.githubusercontent.com/23412755/64644939-121f8200-d414-11e9-9d20-c49cd4968344.png)
After:
![comma-after](https://user-images.githubusercontent.com/23412755/64644946-1a77bd00-d414-11e9-9397-fd5fdb3e54bd.png)

Not good.

jensmaurer:
Editorial teleconference: Wait for more LaTeX expertise to be present.

jensmaurer:
Editorial meeting: This change is actually intended; math-mode should stand out and not run into the surrounding text. We're fine with the extra space.

Issue #326, opened 2014-06-18
placeholders should be formatted with math fonts


zygoloid:
As discussed at Rapperswil, we'd like to experiment with this.

tkoeppe:
If we are talking about *code* placeholders here, then I'm worried that the math font won't harmonize with the teletype font. They have different contrast and different cap height. This may work in certain situations, but it may come to bite us in the future.

I find italic teletype appropriate for the job. Just like "normal" body font italics are appropriate for mathematical variables in the context of body text, italic teletype seems appropriate for "code variables".

jensmaurer:
See also #1060.

tkoeppe:
Indeed. I think the need for italics in this context is by and large the same as for any code placeholder, so I'd be perfectly happy treating the two the same way.

(Note that for capital letters, the distinction between italic and slanted isn't very big, so either is fine here.)

jensmaurer:
See #1139 for indexed (type) variables.

jensmaurer:
We should decide how to proceed and update the wiki accordingly.

jensmaurer:
Editorial meeting discussion: Variables of different kinds might be formatted differently. Possible kinds: type, variable, object, value, expression. It seems the convention is to use italics for all but the first and last; use code font for those. For meta-variables that expand to code, use code font. For other meta-variables, use math. For string interpolation, use \placeholder. Also for "op" in fold-expression.
Idea: Introduce \tvar for type placeholders and make them italics. Not pursued.

eisenwave:
@tkoeppe can we close this? We use `\placeholder` for all sorts of things like `character-type` at this point, and we probably don't want that to be in math font.

I think we could put what Jens said into the specification style guidelines; that seems to be the established convention at this point. Other than that, we're done with this issue, right?

PR #3442, opened 2019-11-13
P1979R0 Resolution to US086

### LINKED ISSUE #3402: [2019-11 CWG Motion 9]  P1979R0 Resolution to US086
http://wiki.edg.com/pub/Wg21belfast/StrawPolls/p1979r0.h

jensmaurer:
Also fixes NB US 086 (C++20 CD)

Fixes #3402 
Fixes cplusplus/nbballot#85

### DIFF
--- a/source/modules.tex
+++ b/source/modules.tex
@@ -495,11 +495,12 @@ all translation units imported by
 exported \grammarterm{module-import-declaration}{s}
 in $T$; such translation units are
 said to be \defnx{exported}{module!exported} by $T$.
-When a \grammarterm{module-import-declaration} in a module unit imports
-another module unit of the same module, it also imports
-all translation units imported by
-all \grammarterm{module-import-declaration}{s}
-in that module unit.
+Additionally, when a \grammarterm{module-import-declaration}
+in a module unit of some module $M$ imports another module unit $U$ of $M$,
+it also imports all translation units imported by
+non-exported \grammarterm{module-import-declaration}{s}
+in the module unit purview of $U$.\footnote{This is consistent
+with the rules for visibility of imported names\iref{basic.scope.namespace}.}
 These rules may in turn lead to the importation of yet more
 translation units.
 


source/modules.tex
```diff
@@ -495,11 +495,12 @@
...
-another module unit of the same module, it also imports
-all translation units imported by
-all \grammarterm{module-import-declaration}{s}
-in that module unit.
+Additionally, when a \grammarterm{module-import-declaration}
+in a module unit of some module $M$ imports another module unit $U$ of $M$,
+it also imports all translation units imported by
+non-exported \grammarterm{module-import-declaration}{s}
+in the module unit purview of $U$.\footnote{This is consistent
+with the rules for visibility of imported names ([basic.scope.namespace]).}
```
burblebee: Fix the reference to use \iref{}.

source/modules.tex
```diff
@@ -495,11 +495,12 @@
 exported \grammarterm{module-import-declaration}{s}
 in $T$; such translation units are
 said to be \defnx{exported}{module!exported} by $T$.
-When a \grammarterm{module-import-declaration} in a module unit imports
-another module unit of the same module, it also imports
-all translation units imported by
-all \grammarterm{module-import-declaration}{s}
-in that module unit.
+Additionally, when a \grammarterm{module-import-declaration}
+in a module unit of some module $M$ imports another module unit $U$ of $M$,
```
burblebee: Should we be using \placeholder here instead of math font?

jensmaurer: We have a healthy variety of opinions from the history of editorial meetings here: #326, #3139, #3271.
The latest and greatest is to use math variables for everything, so that's what I'm doing here.

source/modules.tex
jensmaurer: Fixed.

source/modules.tex
burblebee: Would be good to get some examples of when to use what added to our guidelines.

PR #3451, opened 2019-11-14
P1878R1 Constraining Readable Types

### LINKED ISSUE #3412: [2019-11 LWG Motion 9]  P1878R1 Constraining Readable Types
This resolves NB comments US263, US264, US268 and LWG issue 3279

zygoloid:
Instructions to change section names were ignored; the shorter section
names are preferable, and the inclusion of 'iterator.' seems sufficient
to indicate that we're talking about indirect reading and writing.

Fixes #3412.
Fixes cplusplus/nbballot#259.
Fixes cplusplus/nbballot#260.
Fixes cplusplus/nbballot#264.
Fixes cplusplus/papers#628

Also fixes NB US 263, US 264, US 268 (C++20 CD), and LWG3279.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -50,8 +50,8 @@ namespace std {
   template<class T>
     using iter_difference_t = @\seebelow@;
 
-  // \ref{readable.traits}, readable traits
-  template<class> struct readable_traits;
+  // \ref{readable.traits}, indirectly readable traits
+  template<class> struct indirectly_readable_traits;
   template<class T>
     using iter_value_t = @\seebelow@;
 
@@ -59,7 +59,7 @@ namespace std {
   template<class I> struct iterator_traits;
   template<class T> requires is_object_v<T> struct iterator_traits<T*>;
 
-  template<@\placeholder{dereferenceable}@ T>
+  template<@\exposconcept{dereferenceable}@ T>
     using iter_reference_t = decltype(*declval<T&>());
 
   namespace ranges {
@@ -73,7 +73,7 @@ namespace std {
     }
   }
 
-  template<@\placeholder{dereferenceable}@ T>
+  template<@\exposconcept{dereferenceable}@ T>
     requires requires(T& t) {
       { ranges::iter_move(t) } -> @\placeholder{can-reference}@;
     }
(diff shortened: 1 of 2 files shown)


source/iterators.tex
```diff
@@ -719,31 +719,34 @@
...
-if \tcode{iterator_traits<I>} names a specialization
+\tcode{incrementable_traits<J>::difference_type}
+if \tcode{iterator_traits<J>} names a specialization
 generated from the primary template, and
 
 \item
-\tcode{iterator_traits<I>::\brk{}difference_type} otherwise.
+\tcode{iterator_traits<J>::difference_type} otherwise,
 \end{itemize}
+where \tcode{J} is \tcode{remove_cvref_t<I>}.
```
jensmaurer: It's a bit unfortunate that we introduce J so late, after all the bullets. Can we introduce it before the bulleted list?

source/iterators.tex
```diff
@@ -1191,43 +1197,51 @@
...
+\end{codeblock}
+
 \pnum
-Given a value \tcode{i} of type \tcode{I}, \tcode{I} models \libconcept{readable}
+Given a value \tcode{i} of type \tcode{I}, \tcode{I} models \libconcept{indirectly_readable}
 only if the expression \tcode{*i} is equality-preserving.
 \begin{note}
 The expression \tcode{*i} is indirectly required to be valid via the
-exposition-only \placeholder{dereferenceable} concept\iref{iterator.synopsis}.
+exposition-only \tcode{\placeholder{dereferenceable}} concept\iref{iterator.synopsis}.
```
jensmaurer: This should be `\exposconcept`

source/iterators.tex
zygoloid: Do you have a concrete suggestion? I don't want to break up the flow of this paragraph or make it too different from the phrasing used in parallel sentences elsewhere.

One possibility:

> The type `iter_difference_t<I>` denotes
> * `iter_difference_t<remove_cvref<I>>` if `remove_cvref<I>` is not the same type as `I`, or
> * `incrementable_traits<I>::difference_type` if `iterator_traits<I>` names a specialization generated from the primary template, or
> * `iterator_traits<I>::difference_type` otherwise.

What do you think?

source/iterators.tex
zygoloid: We use `\placeholder` throughout this clause for this purpose:

```
\pnum
The definitions in this subclause make use of the following
exposition-only concepts:

\begin{codeblock}
template<class I>
concept @\placeholder{cpp17-iterator}@ =
  copyable<I> && requires(I i) {
...
```

```
Otherwise, if \tcode{I} satisfies the exposition-only concept
\tcode{\placeholder{cpp17-input-iterator}},
```

(etc). I think this cleanup should be done as a single pass across all the exposition-only concepts, if we really think this level of detail in our markup is worthwhile. (Why are we singling out concepts for this treatment?)

source/iterators.tex
jensmaurer: We don't seem to constraint the template parameter "I" at all, and we seem to use "class T" in the synopsis instead of I.
If these assertions hold, maybe something like "For a type I, where J denotes remove_cvref&lt;I>, the type iter_difference_t&lt;I> denotes... (text as shown in ingress paper)."

source/iterators.tex
jensmaurer: > Why are we singling out concepts for this treatment?

Because we have a special "index of concept names" that, ideally, should point to every mention of a concept. That's why I made `\deflibconcept` and `\libconcept` in the first place.
I've created #3494 for the global cleanup.

Issue #3321, opened 2019-10-23
The standard and the DRY principle: header descriptions


gennaroprota:
Several header descriptions in the standard say things like "the header <header> declares three functions, two classes and defines five macros".

These descriptions add really nothing useful, and can only become wrong under maintenance. Even if numbers are not provided, it may happen e.g. that a description only mention classes and functions but, later, function templates are added.

Here's a list of what are IMHO the problematic ones. I suggest removing them, but, for some of them, I also suggest a replacement in case the removal is not possible:

 -  In [char.traits.specializations]:

      "The header <string> defines five specializations of the class template  
      char_traits: char_traits<char>, char_traits<char8_t>,      
      char_traits<char16_t>, char_traits<char32_t>, and char_traits<wchar_t>."

    Suggested replacement:

      "The header <string> defines the following specializations:      
      char_traits<char>, char_traits<char8_t>, char_traits<char16_t>,      
      char_traits<char32_t>, and char_traits<wchar_t>."

    (Also, should it say "declares" rather than "defines"? NB: this question    
    applies to every description, but I will not repeat it.)

  - In [string.classes]:

      "The header <string> defines the basic_string class template for
      manipulating varying-length sequences of char-like objects and five
      typedef-names, string, u8string, u16string, u32string, and wstring, that
      name the specializations basic_string<char>, basic_string<char8_t>,
      basic_string<char16_t>, basic_string<char32_t>, and basic_string<wchar_t>,
      respectively."

    Suggested replacement:

      "The header <string> defines facilities for manipulating varying-length  
      sequences of char-like objects."

  - In [associative.general]:

      "The header <map> defines the class templates map and multimap; the header
      <set> defines the class templates set and multiset."

  - In [unord.general]:

      "The header <unordered_map> defines the class templates unordered_map and
      unordered_multimap; the header <unordered_set> defines the class templates
      unordered_set and unordered_multiset."

  - In [container.adaptors.general]:

      "The headers <queue> and <stack> define the container adaptors queue,    
      priority_queue, and stack."

  - In [views.general]:

      "The header <span> defines the view span."

  - In [complex.numbers]:

      "The header <complex> defines a class template, and numerous functions for
      representing and manipulating complex numbers."

    Suggested replacement:

      "The header <complex> defines facilities for representing and manipulating
      complex numbers."

  - In [valarray.syn]:

      "The header <valarray> defines five class templates (valarray,      
      slice_array, gslice_array, mask_array, and indirect_array), two classes      
[...]

PR #3453, opened 2019-11-14
P1456R1 Move-only views

### LINKED ISSUE #3414: [2019-11 LWG Motion 11] P1456R1 Move-only views

zygoloid:
Fixes #3414.

Fixes cplusplus/nbballot#273
Partially addresses cplusplus/nbballot#249
Partially addresses cplusplus/nbballot#266
Fixes cplusplus/nbballot#277

### DIFF
--- a/papers/nxxxx.md
+++ b/papers/nxxxx.md
@@ -43,6 +43,26 @@ CWG motion 2: [P1234R5 "Paper name"](http://wg21.link/p1234r5), resolving 3 NB c
 
 ...
 
+### LWG motion 11
+
+The description of this paper specifies that:
+
+> each such `base()` member [of a range adaptor, that returns a copy of the underlying view]
+> be replaced to by two overloads:
+> a `const`-qualified overload that requires the type of the underlying view to model CopyConstructible, and
+> a `&&`-qualified overload that extracts the underlying view from the adaptor
+
+but the wording changes omitted explicit editing instructions
+to make these changes to the
+`take_while_view`, `drop_view`, `drop_while_view` and `elements_view`
+range adaptors, which were added by [P1035R7](http://wg21.link/p1035r7)
+(2019-07 LWG Motion 23), after R0 of this paper was authored.
+
+Consistent with the proposal in the paper,
+and after consulting the paper authors and the LWG chair,
+the corresponding changes were also applied to
+the additional range adaptors listed above.
+
 ## Feature test macros
 
 The feature test macro `__cpp_nontype_template_parameter_class` has been removed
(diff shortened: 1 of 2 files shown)


jensmaurer:
Please augment the commit message to add references to the NB ballot comments mentioned above.

johelegp:
Missing changes to `base`: `take_while_view`, `drop_view`, `drop_while_view` and `elements_view`.

cor3ntin:
@johelegp No, you are right, all views with  a `base` member function should provide a rvale qualified overload - P1862 does that to iterators ( and iterators only ) @CaseyCarter

CaseyCarter:
> Missing changes to base: `take_while_view`, `drop_view`, `drop_while_view` and `elements_view`.

Yes, "Change [range.transform.view], [range.take.view], [range.common.view], and [range.reverse.view] similarly" in P1456 is missing the range adaptors added by P1035. Are we willing to construe P1456 as providing editorial instructions to make similar changes to all such range adaptors, or shall I submit an LWG issue to make this same change to these four that were missed?

zygoloid:
@CaseyCarter I think the "Proposal" section gives us sufficient motivation to consider this an editorial fix to the wording as moved:

> we propose that each such `base()` member be replaced to by two overloads: a `const`-qualified overload that requires the type of the underlying view to model CopyConstructible, and a `&&`-qualified overload that extracts the underlying view from the adaptor

(especially with you as a paper author telling us this was indeed the intent!).

papers/nxxxx.md
```diff
@@ -43,6 +43,25 @@ CWG motion 2: [P1234R5 "Paper name"](http://wg21.link/p1234r5), resolving 3 NB c
...
+
+but the wording changes omitted explicit editing instructions
+to make these changes to the
+`take_while_view`, `drop_view`, `drop_while_view` and `elements_view`
+range adaptors, which were added by [P1035R7](http://wg21.link/p1035r7)
+(2019-07 LWG Motion 23), after R0 of this paper was authored.
+
+Consistent with the proposal in the paper,
+and after consulting the paper authors and the LWG chair,
+the corresponding changes were also applied to the five additional range adaptors.
```
johelegp:
```suggestion
the corresponding changes were also applied to the four additional range adaptors.
```

johelegp: Hmm... This can be technically correct and not at the same time.

CaseyCarter: This is ironically reminiscent of #3321: Never list things, and never count them ;)

zygoloid: Oops. You'd think I'd be able to count to four. That's embarrassing. =)

Fixed by removing the number entirely. Thanks for catching that.

PR #3454, opened 2019-11-14
P1391R4 Range constructor for std::string_view

### LINKED ISSUE #3415: [2019-11 LWG Motion 12] P1391R4 Range constructor for std::string_view

zygoloid:
Renamed constructor parameters in description to match those in class
definition.

Fixes #3415.
Fixes cplusplus/nbballot#229
Partially addresses cplusplus/nbballot#268
Partially addresses cplusplus/nbballot#284
Fixes cplusplus/papers#184

### DIFF
--- a/source/strings.tex
+++ b/source/strings.tex
@@ -4109,7 +4109,7 @@ private:
   size_type size_;              // \expos
 };
 
-// \ref{string.view.deduction}, deduction guide
+// \ref{string.view.deduct}, deduction guide
 template<class It, class End>
   basic_string_view(It, End) -> basic_string_view<iter_value_t<It>>;
 \end{codeblock}
@@ -4190,17 +4190,17 @@ template<class It, class End>
 \pnum
 \constraints
 \begin{itemize}
-\item \tcode{It} satisfies \libconcept{contiguous_iterator},
-\item \tcode{End} satisfies \tcode{\libconcept{sized_sentinel_for}<It>},
-\item \tcode{is_same_v<iter_value_t<It>, charT>} is \tcode{true}, and
+\item \tcode{It} satisfies \libconcept{contiguous_iterator}.
+\item \tcode{End} satisfies \tcode{\libconcept{sized_sentinel_for}<It>}.
+\item \tcode{is_same_v<iter_value_t<It>, charT>} is \tcode{true}.
 \item \tcode{is_convertible_v<End, size_type>} is \tcode{false}.
 \end{itemize}
 
 \pnum
 \expects
 \begin{itemize}
-\item \range{begin}{end} is a valid range,
-\item \tcode{It} models \libconcept{contiguous_iterator}, and
+\item \range{begin}{end} is a valid range.
+\item \tcode{It} models \libconcept{contiguous_iterator}.
 \item \tcode{End} models \tcode{\libconcept{sized_sentinel_for}<It>}.
 \end{itemize}
 
@@ -4872,7 +4872,7 @@ Determines \tcode{xpos}.
 Otherwise, returns \tcode{npos}.
 \end{itemdescr}
 
-\rSec2[string.view.deduction]{Deduction guide}
+\rSec2[string.view.deduct]{Deduction guide}
 
 \begin{itemdecl}
 template<class It, class End>
@@ -4883,7 +4883,7 @@ template<class It, class End>
 \pnum
 \constraints
 \begin{itemize}
-\item \tcode{It} satisfies \libconcept{contiguous_iterator}, and
+\item \tcode{It} satisfies \libconcept{contiguous_iterator}.
 \item \tcode{End} satisfies \tcode{\libconcept{sized_sentinel_for}<It>}.
 \end{itemize}
 \end{itemdescr}
--- a/source/strings.tex
+++ b/source/strings.tex
@@ -4020,6 +4020,8 @@ public:
   constexpr basic_string_view& operator=(const basic_string_view&) noexcept = default;
   constexpr basic_string_view(const charT* str);
   constexpr basic_string_view(const charT* str, size_type len);
+  template<class It, class End>
+    constexpr basic_string_view(It begin, End end);
 
   // \ref{string.view.iterators}, iterator support
   constexpr const_iterator begin() const noexcept;
@@ -4106,6 +4108,10 @@ private:
   const_pointer data_;          // \expos
   size_type size_;              // \expos
 };
+
+// \ref{string.view.deduction}, deduction guide
+template<class It, class End>
+  basic_string_view(It, End) -> basic_string_view<iter_value_t<It>>;
 \end{codeblock}
 
 \pnum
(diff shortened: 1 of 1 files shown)


jensmaurer:
Please add "fixes" to the commit for the NB comments linked in this pull request.

source/strings.tex
```diff
@@ -4174,6 +4180,36 @@
...
+\begin{itemdecl}
+template<class It, class End>
+  constexpr basic_string_view(It begin, End end);
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\constraints
+\begin{itemize}
+\item \tcode{It} satisfies \tcode{contiguous_iterator},
```
johelegp:
```suggestion
\item \tcode{It} satisfies \libconcept{contiguous_iterator},
```
[As done in the similar change to `span`.](https://github.com/cplusplus/draft/pull/3456/files#diff-0fb8725b6f959c00f8cef02d54e8c125R10736)

source/strings.tex
```diff
@@ -4174,6 +4180,36 @@
...
+\item \tcode{End} satisfies \tcode{sized_sentinel_for<It>},
```
johelegp:
```suggestion
\item \tcode{End} satisfies \tcode{\libconcept{sized_sentinel_for}<It>},
```

source/strings.tex
```diff
@@ -4174,6 +4180,36 @@
...
+\item \tcode{is_same_v<iter_value_t<It>, charT>} is \tcode{true}, and
```
johelegp: The similar change to `span` ends the `\item`s with periods.

source/strings.tex
```diff
@@ -4174,6 +4180,36 @@
...
+\item \tcode{is_convertible_v<End, size_type>} is \tcode{false}.
+\end{itemize}
+
+\pnum
+\expects
+\begin{itemize}
+\item \range{begin}{end} is a valid range,
+\item \tcode{It} models \tcode{contiguous_iterator}, and
+\item \tcode{End} models \tcode{sized_sentinel_for<It>}.
```
johelegp: As above, regarding the use of `\libconcept` and periods.

source/strings.tex
```diff
@@ -4836,6 +4872,22 @@
...
+template<class It, class End>
+  basic_string_view(It, End) -> basic_string_view<iter_value_t<It>>;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\constraints
+\begin{itemize}
+\item \tcode{It} satisfies \tcode{contiguous_iterator}, and
+\item \tcode{End} satisfies \tcode{sized_sentinel_for<It>}.
```
johelegp: As above.

source/strings.tex
jensmaurer: Yes, please. This adds nice entries to the "index of concept names".

source/strings.tex
jensmaurer: Ditto.

source/strings.tex
jensmaurer: That was my change vs. the incoming paper; it seems a lot more readable with periods and avoids an awkward note-within-a-sentence for the very last bullet.

source/strings.tex
jensmaurer: sized_sentinel_for takes two arguments, not just one.

@zygoloid, it seems novel to rely on the "prototype parameter" rule for a description like this. Can we fix this editorially or do we reject this for LWG cleanup or something else?

source/strings.tex
```diff
@@ -4836,6 +4872,22 @@
 Otherwise, returns \tcode{npos}.
 \end{itemdescr}
 
+\rSec2[string.view.deduction]{Deduction guide}
```
jensmaurer: This should be [string.view.deduct] to keep it short. I've already added an item to our style guide.

source/strings.tex
jensmaurer: Uh, the "novel" part is not actually true. We have in [ranges]:

> if C models derived_from&lt;forward_iterator_tag>

jensmaurer: Let revert this.

johelegp: Right. As you mention in #3456, there are quite a few such uses. It is actually in the concept definition and not their uses that the full spelling is the way to go when saying that the concept's parameters model it when A, B, and C.

zygoloid: Similarly in `vformat_to`:

> `Out` satisfies `output_iterator<const charT\&>`.

Resolving this; I think this is consistent with what we do elsewhere. (Commas will be changed to periods as suggested.)

source/strings.tex
```diff
@@ -4174,6 +4180,36 @@
...
+\begin{itemize}
+\item \range{begin}{end} is a valid range.
+\item \tcode{It} models \libconcept{contiguous_iterator}.
+\item \tcode{End} models \tcode{\libconcept{sized_sentinel_for}<It>}.
+\end{itemize}
+
+\pnum
+\effects
+Initializes \tcode{data_} with \tcode{to_address(begin)} and
+initializes \tcode{size_} with \tcode{end - begin}.
```
jensmaurer: `sIze_t` is of type size_type, but `end-begin` is of type `ptrdiff_t`. The conversion will "just work" even for the overflow case, but it's a bit odd.

CaseyCarter: Bug: `end - begin` has some _`signed-integer-like`_ type (that isn't necessarily `ptrdiff_t`) but must be explicitly convertible to `size_t` per [iterator.concept.winc]/6. This needs to initialize `size_` with `static_cast<size_t>(end - begin)`.

As always, let-me-know-if-this-isn't-editorial-and-I'll-file-an-LWG-issue.

jensmaurer: We could simply say "direct-initializes `size_t` with `end-begin`", but that's a bit subtle.
@zygoloid , your call. (This also affects the similar text in the corresponding `span` changes.)

CaseyCarter: Sorry, my "bug" cry here was a bit of an overreaction - I (and apparently all of LWG) am still not used to non-integral difference types. Either "direct-initializes" or `static_cast<size_t>` would be great here, or we could clean it up post-merge via LWG issue. This certainly doesn't rise to the level of refusing to apply the proposal and sending it back to LWG, which "bug" would seem to imply.

zygoloid: It seems to me that this is just one instance of a systemic problem in the library wording. This phrasing "initializes x with y" is common, but underspecified (and formally meaningless) -- the library wording either needs to say what kind of initialization is performed, or specify an *initializer* (not an *expression*) with which to initialize. We should ask LWG to think about this; for each "initializes x with y" utterance, the reader should know what kind of initialization we mean.

Looking at random through the library wording, the first case I found: [queue.cons]/1:

> *Effects:* Initializes `c` with `cont`.

The meaning of this depends on whether this is direct- or copy-initialization. (It's obscure, but if `T` is not *Cpp17CopyInsertable* into the container, it could be the case that one form of initialization works and the other does not, or that they both work and do different things.)

Another random sample: [partial.sum]/2:

>  *Effects:* For a non-empty range, the function creates an accumulator `acc` whose type is `InputIterator`'s value type, initializes it with `*first`, and assigns the result to `*result`.

Again the difference between direct- and copy-initialization is observable here.

Perhaps the library should have blanket wording that when it says "initializes", it means by direct- or copy-initialization, and that it's unspecified which one you get (or something like that) -- and someone should go through all the instances and check if any of them mean something else (I doubt this is the only case that does). I'm not convinced that this particular instance of the problem needs to be dealt with while applying the motions.

@CaseyCarter Can you file an LWG issue, please?

zygoloid: I've mailed lwgchair to ask for an issue.

PR #3464, opened 2019-11-15
[basic.def.odr] Split long-winded bulleted sentence.

### LINKED ISSUE #3462: [basic.def.odr] Sentence has two full stops (incorrect list item punctuation)
image

jensmaurer:
Fixes #3462.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -577,14 +577,14 @@ no diagnostic is required unless a prior definition
 is reachable at a point where a later definition appears.
 Given such
 an entity named \tcode{D} defined in more than one translation unit,
-then
+all of the following requirements shall be satisfied.
 \begin{itemize}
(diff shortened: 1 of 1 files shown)


source/basic.tex
```diff
@@ -664,7 +664,8 @@
 D d2;                           // \tcode{X(int, int, int)} called by \tcode{D()};
                                 // \tcode{D()}'s implicit definition violates the ODR
 \end{codeblock}
-\end{example}
+\end{example}%
+; and
```
zygoloid: Oh no, not another one of these!

Can we convert all the bullets into separate sentences to avoid this? This is a page-long sentence as-is, and doesn't seem to need to be...

jensmaurer: Yes, we can.

PR #3465, opened 2019-11-15
[format.arg] Fix parameter type for basic_format_arg constructor.

### LINKED ISSUE #3461: [format.arg] Declaration mismatch
An exposition-only constructor of `basic_format_arg` is declared as

https://github.com/cplusplus/draft/blob/6deb1811d9c96ad0efe52ee18140b7998b16ba1b/source/utilities.tex#L21583-L21584

in the class template definition, but `const` is missing in the description:

https://github.com/cplusplus/draft/blob/6deb1811d9c96ad0efe52ee18140b7998b16ba1b/source/utilities.tex#L21739

jensmaurer:
Fixes #3461.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -21090,7 +21090,7 @@ Initializes \tcode{value} with
 \end{itemdescr}
 
 \begin{itemdecl}
-template<class T> explicit basic_format_arg(T* p) noexcept;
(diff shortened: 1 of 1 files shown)


jensmaurer:
This was fixed the wrong way around; see LWG3540.

PR #3470, opened 2019-11-16
[defns.access] Clarify definition of "access".

zygoloid:
Add cross-linking between the places that introduce accesses and the
definition of the term, and add a note explaining that we only ever
access objects of scalar type.

### DIFF
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -117,7 +117,19 @@ defined.
 
 \indexdefn{access}%
 \definition{access}{defns.access}
-\defncontext{execution-time action} read or modify the value of an object
+\defncontext{execution-time action}
+read\iref{conv.lval} or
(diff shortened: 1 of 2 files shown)


languagelawyer:
What was really missing is [conv.lval] saying that it may introduce a read.
But crossreferences would be helpful.

source/intro.tex
```diff
@@ -117,7 +117,19 @@
...
+read\iref{conv.lval} or
+modify (\ref{expr.ass}, \ref{expr.post.incr}, \ref{expr.pre.incr})
+the value of an object
+
+\begin{defnote}
+Only objects of scalar type can be accessed.
+Attempts to read or modify an object of class type
+typically invoke a constructor\iref{class.ctor}
+or assignment operator\iref{class.copy.assign};
+such invocations do not themselves constitue accesses,
```
jensmaurer: typo "constitue"

PR #3463, opened 2019-11-15
P1719R2 Mandating the Standard Library: Clause 26 - Numerics library

zygoloid:
Fixes #3427.
Fixes cplusplus/papers#485.

PR #3471, opened 2019-11-16
P1722R2 Mandating the Standard Library: Clause 30 - Regular Expression library

### LINKED ISSUE #3431: [2019-11 LWG Motion 29]  P1722R2 Mandating the Standard Library: Clause 30 - Regular Expression library

zygoloid:
Fixes #3431.
Fixes cplusplus/papers#488.

### DIFF
--- a/source/regex.tex
+++ b/source/regex.tex
@@ -891,10 +891,6 @@ regex_error(regex_constants::error_type ecode);
 
 \begin{itemdescr}
 \pnum
-\effects
-Constructs an object of class \tcode{regex_error}.
-
-\pnum
 \ensures
 \tcode{ecode == code()}.
 \end{itemdescr}
@@ -950,7 +946,7 @@ namespace std {
 \indextext{requirements!regular expression traits}%
 \indextext{\idxcode{regex_traits}!specializations}%
 The specializations \tcode{regex_traits<char>} and
-\tcode{regex_traits<wchar_t>} shall be valid and shall meet the
+\tcode{regex_traits<wchar_t>} meet the
 requirements for a regular expression traits class\iref{re.req}.
 
 \indexlibrarymember{regex_traits}{char_class_type}%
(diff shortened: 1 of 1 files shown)


source/regex.tex
```diff
@@ -2427,7 +2382,8 @@
 
 \begin{itemdescr}
 \pnum
-\requires \tcode{ready() == true} and \tcode{OutputIter} shall meet the requirements for a
+\expects
+\tcode{ready() == true} and \tcode{OutputIter} meets the requirements for a
```
johelegp:
```suggestion
\tcode{ready() == true} and \tcode{OutputIter} meets the requirements for an
```

source/regex.tex
```diff
@@ -1650,13 +1628,8 @@
 
 \begin{itemdescr}
 \pnum
-\requires
-\tcode{InputIterator} shall meet the
-\oldconcept{InputIterator} requirements\iref{input.iterators}.
```
johelegp: I see that this was dropped, whereas R1 does update it. #3463 keeps them and updates them, presumably because, unlike [algorithms], there's no blanket wording to this effect. It seems like [re] doesn't have such blanket wording, but I can see a lot of *`Iterator` uses and very few touches to those in this PR.

timsong-cpp: It's brought in via the `string_type` construction in the "_Effects:_ Equivalent to".

source/regex.tex
timsong-cpp: `a` is right since `\oldconcept` adds the _Cpp17_ prefix.

source/regex.tex
```diff
@@ -1336,11 +1332,11 @@
 basic_regex();
 \end{itemdecl}
 
+\ednote{All through this sub-clause, please move the "Throws:" elements to be after the "Effects:" and "Ensures:" elements.}
```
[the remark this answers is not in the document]
zygoloid: Thanks, fixed.

PR #3472, opened 2019-11-16
Motions 2019 11 lwg 1: P1917R0 C++ Standard Library Issues to be moved in Belfast

### LINKED ISSUE #3404: [2019-11 LWG Motion 1] P1917R0 C++ Standard Library Issues to be moved in Belfast
These issues resolve NB comments: GB166, US297, and US261

burblebee:
Fixes #3404.

Issues:
* From LWG3149:
** "immediate context" is used but never defined. It is used in reference to variables, expressions, and stand-alone as ``immediate context'', yet there is only a definition for "immediate function context".
** What is an unrelated context? E.g. LWG3149 adds this in [concept.defaultconstructible]:
      "Access checking is performed as if in a context unrelated to \tcode{T}."
* LWG3272 adds this wording to [time.parse]/p1:
      "In the following paragraphs, let is denote an object of type basic_istream<charT, traits>, where charT and traits are template parameters in that context.".
      What context?

Fixes cplusplus/nbballot#164
Fixes cplusplus/nbballot#257
Fixes cplusplus/nbballot#293

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -655,11 +655,7 @@ is undefined.
 \indextext{iterator!constexpr}%
 Iterators are called \defn{constexpr iterators}
 if all operations provided to meet iterator category requirements
-are constexpr functions, except for
-\begin{itemize}
-\item a pseudo-destructor call\iref{expr.prim.id.dtor}, and
-\item the construction of an iterator with a singular value.
-\end{itemize}
+are constexpr functions.
 \begin{note}
 For example, the types ``pointer to \tcode{int}'' and
 \tcode{reverse_iterator<int*>} are constexpr iterators.
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -1779,7 +1779,7 @@ and let \tcode{n} denote a value of type \tcode{D}.
   \tcode{(a + D(x + y))} is equal to \tcode{((a + x) + y)}.
 \item \tcode{(a + D(0))} is equal to \tcode{a}.
 \item If \tcode{(a + D(n - 1))} is valid, then
-  \tcode{(a + n)} is equal to \tcode{++(a + D(n - 1))}.
+  \tcode{(a + n)} is equal to \tcode{[](I c)\{ return ++c; \}(a + D(n - 1))}.
 \item \tcode{(b += -n)} is equal to \tcode{a}.
 \item \tcode{(b -= n)} is equal to \tcode{a}.
 \item \tcode{addressof(b -= n)} is equal to \tcode{addressof(b)}.
--- a/source/time.tex
+++ b/source/time.tex
@@ -5241,9 +5241,8 @@ constexpr weekday_indexed(const chrono::weekday& wd, unsigned index) noexcept;
 \begin{itemdescr}
 \pnum
 \effects
-Constructs an object of type \tcode{weekday_indexed} by
-initializing \tcode{wd_} with \tcode{wd} and \tcode{index_} with \tcode{index}.
-The values held are unspecified if \tcode{!wd.ok()} or \tcode{index} is not in the range \crange{1}{5}.
+Initializes \tcode{wd_} with \tcode{wd} and \tcode{index_} with \tcode{index}.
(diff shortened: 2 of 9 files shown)


source/iterators.tex
```diff
@@ -1783,7 +1779,7 @@
   \tcode{(a + D(x + y))} is equal to \tcode{((a + x) + y)}.
 \item \tcode{(a + D(0))} is equal to \tcode{a}.
 \item If \tcode{(a + D(n - 1))} is valid, then
-  \tcode{(a + n)} is equal to \tcode{++(a + D(n - 1))}.
+  \tcode{(a + n)} is equal to \tcode{[](I c)\{ return ++c; \}}.
```
johelegp:
```suggestion
  \tcode{(a + n)} is equal to \tcode{[](I c)\{ return ++c; \}(a + D(n - 1))}.
```
The addition should only replace `++`.

source/time.tex
```diff
@@ -9505,8 +9508,8 @@
 \pnum
 \effects
 Constructs a \tcode{zoned_time} by
-initializing \tcode{zone_} with \tcode{std::move(z)}.
-% FIXME: What about tp_?
+initializing \tcode{zone_} with \tcode{std::move(z)} and
+default constructing tcode{tp_}.
```
```suggestion
default constructing \tcode{tp_}.
```
Following the first change in this file, the first line of the `\effects` should be dropped and "initializing" changed to "Initializes".

source/time.tex
```diff
@@ -10795,6 +10827,18 @@
 except for white space, are parsed unchanged from the stream.
 A white space character matches zero or more white space characters in the input stream.
 
+\pnum
+If the type being parsed can not represent
+the information that the format flag refers to,
+\tcode{is.setstate(ios_base::failbit)} is called.
+\begin{example}
+A \tcode{duration} cannot represent a weekday.
```
```suggestion
A \tcode{duration} cannot represent a \tcode{weekday}.
```

source/utilities.tex
```diff
@@ -19093,7 +19094,7 @@
   };
 
   to_chars_result to_chars(char* first, char* last, @\seebelow@ value, int base = 10);
-
+  to_chars_result to_chars(char* first, char* last, bool value, int base = 10) = delete;
```
```suggestion
  to_chars_result to_chars(char* first, char* last, bool value, int base = 10) = delete;

```

source/time.tex
timsong-cpp: I think this is English "weekday", not the type.

johelegp: I considered it might be so, but while checking something else in the paper, I noticed it's actually code.

CaseyCarter:
> From LWG3149:
> * "immediate context" is used but never defined. It is used in reference to variables, expressions, and stand-alone as ``immediate context'', yet there is only a definition for "immediate function context".

This is [CWG 1844](https://wg21.link/cwg1844).

> * What is an unrelated context? E.g. LWG3149 adds this in [concept.defaultconstructible]:
"Access checking is performed as if in a context unrelated to \tcode{T}."

This is library-speak for "metaprogramming machinery cannot access protected or private class members regardless of the context in which they are evaluated". For example, this program:
```c++
class S { 
  S() {}
  static void f() {
    S s; // Fine
    static_assert(std::default_constructible<S>); // ill-formed
  }
};
```
is ill-formed despite that `std::default_constructible<S>` is evaluated where `S`'s default constructor *is* accessible.

jensmaurer:
The individual commits need "Also fixes NB ..." cross-reference to the NB comments they address, where applicable.
For example, LWG 3277 fixes US 266.

source/time.tex
zygoloid: It should be, but let's address that separately and more holistically.

zygoloid: Filed #3506 to track.

source/time.tex
zygoloid: I'm going with the formatting in the LWG issues list (`\tcode` for both instances of this), though I share the concern that this might actually mean English weekday.

PR #3474, opened 2019-11-16
P1892R1 Extended locale-specific presentation specifiers for std::format

### LINKED ISSUE #3424: [2019-11 LWG Motion 22] 1892R1 Extended locale-specific presentation specifiers for std::format
http://wiki.edg.com/pub/Wg21belfast/StrawPolls/P1892R1.pdf

This resolves NB comment GB226

burblebee:
Fixes #3424.

Fixes cplusplus/nbballot#223
Fixes cplusplus/papers#641

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -19884,17 +19884,25 @@ For string types it specifies how many characters will be used from the string.
 The \tcode{L} option causes
 the \defnx{locale-specific form}{locale-specific form!format string}
 to be used for the conversion.
-This option is only valid for arithmetic types.
+This option is only valid for arithmetic types, and
+its effect depends upon the type.
+\begin{itemize}
+\item
 For integral types, the locale-specific form
 causes the context's locale to be used
 to insert the appropriate digit group separator characters.
+
+\item
 For floating-point types, the locale-specific form
 causes the context's locale to be used
 to insert the appropriate digit group and decimal radix separator characters.
+
+\item
 For the textual representation of \tcode{bool}, the locale-specific form
 causes the context's locale to be used
 to insert the appropriate string as if obtained
 with \tcode{numpunct::truename} or \tcode{numpunct::falsename}.
+\end{itemize}
 
 \pnum
 The \fmtgrammarterm{type} determines how the data should be presented.
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -19682,7 +19682,7 @@ The syntax of format specifications is as follows:
 
 \begin{ncbnf}
 \fmtnontermdef{std-format-spec}\br
-    \opt{fill-and-align} \opt{sign} \opt{\terminal{\#}} \opt{\terminal{0}} \opt{width} \opt{precision} \opt{type}
+    \opt{fill-and-align} \opt{sign} \opt{\terminal{\#}} \opt{\terminal{0}} \opt{width} \opt{precision} \opt{\terminal{L}} \opt{type}
 \end{ncbnf}
 
 \begin{ncbnf}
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -19886,6 +19886,30 @@
...
+\pnum
+When the \tcode{L} option is used, the form used for the conversion
+is called the \defn{locale-specific} form.
+The \tcode{L} option is only valid for arithmetic types, and
+its effect depends upon the type.
+\begin{itemize}
+
+\item
+For integral types, the locale-specific form
+causes the context’s locale to be used
```
```suggestion
causes the context's locale to be used
```

source/utilities.tex
```diff
@@ -19886,6 +19886,30 @@
...
+to insert the appropriate digit group separator characters.
+
+\item
+For floating-point types, the locale-specific form
+causes the context’s locale to be used
```
```suggestion
causes the context's locale to be used
```

source/utilities.tex
```diff
@@ -19886,6 +19886,30 @@
...
+to insert the appropriate digit group and decimal radix separator characters.
+
+\item
+For the textual representation of \tcode{bool}, the locale-specific form
+causes the context’s locale to be used
```
```suggestion
causes the context's locale to be used
```

source/utilities.tex
```diff
@@ -19886,6 +19886,30 @@
 For floating-point types this field specifies the formatting precision.
 For string types it specifies how many characters will be used from the string.
 
+\pnum
+When the \tcode{L} option is used, the form used for the conversion
+is called the \defn{locale-specific} form.
```
jensmaurer: This just puts "local-specific" into the index. I think we want to define "locale-specific form" and use `\defnadj{locale-specific}{form}`

source/utilities.tex
```diff
@@ -19886,6 +19886,30 @@
...
+
+\item
+For integral types, the locale-specific form
+causes the context’s locale to be used
+to insert the appropriate digit group separator characters.
+
+\item
+For floating-point types, the locale-specific form
+causes the context’s locale to be used
+to insert the appropriate digit group and decimal radix separator characters.
```
jensmaurer: The examples in the paper show the "decimal" radix separator character to be used even with hexfloats. Something doesn't add up here.

jensmaurer:
... and please revert the lines in the commit comment: Pxxxx first, then "Fixes NB US 123" in the body.

source/utilities.tex
peter-b: @jensmaurer This was a drafting error on my part and I'm sorry it didn't get picked up.

The examples are correct and the wording is is not correct.  I suggest `s/decimal//`.

The intention is to give access the features currently available when using iostreams, e.g.

```c++
#include <iostream>
#include <locale>

int main() {
    const std::locale gb{"en_GB"};
    std::cout.imbue(gb);
    std::cout << std::hex << 123456 << "\n";
    return 0;
}
```
which produces `1e,240` in implementations I have access to right now.

source/utilities.tex
burblebee: I guess this a Unicode issue?  can't see any difference on a mac, so thanks for catching these!

source/utilities.tex
burblebee: yes, much better.  Thanks!

source/utilities.tex
burblebee: @zygoloid, are you ok with me making this normative change editorially?

peter-b: @burblebee @zygoloid If necessary, I will raise an LWG issue.

zygoloid: I think the intent is sufficiently clear, especially given the description given in the paper.

PR #3485, opened 2019-11-18
Motions 2019 11 cwg 2: P1968R0 Core Language Working Group "tentatively ready" Issues

### LINKED ISSUE #2664: [intro.execution] CWG2431: Destruction of a temp. object bound to a reference is not a full-expression?
http://eel.is/c++draft/intro.execution#def:full-expression:
> A *full-expression* is 
> — an invocation of a destructor generated at the end of the lifetime of an object other than a temporary object ([class.temporary]), or

Here is an example
```c++
void foo()
{
    const auto& rk = Klass{};
    Klass k;
    // ...
    // generated invocation `k.~Klass()` is a full-expression
    // generated invocation `rk.~Klass()` is not a full-expression?
}
```
I think it is intended to be a full-expression, so the Standard should say something like
> other than a temporary object **whose lifetime was not extended**

### LINKED ISSUE #3395: [2019-11 CWG Motion 2] P1968R0 Core Language Working Group "tentatively ready" Issues for the November, 2019 (Belfast) meeting

burblebee:
Fixes #3395.
Fixes #2664.

Issues:
* CWG2419:
    In [expr.add]/bullet 4.2, and, after applying CWG2419, in [numeric.ops.midpoint]/p5 and [expr.rel]/p4, we have the wording:
        "a pointer to a hypothetical array element n"
    This reads like n is the hypothetical array element, but n is the size of the array.
    Can we reword this?
* CWG2433: FYI, wording changed after Davis Herring's change "[basic.def.odr] Clean up new bullets"

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -9775,12 +9775,12 @@ template<class T>
 \tcode{a} and \tcode{b} point to, respectively,
 elements $\tcode{x}[i]$ and $\tcode{x}[j]$ of the same array object \tcode{x}.
 \begin{note}
-An object that is not an array element is considered to belong
-to a single-element array for this purpose; see \ref{expr.unary.op}.
-A pointer past the last element of an array \tcode{x} of $n$ elements
+As specified in \ref{basic.compound},
+an object that is not an array element
+is considered to belong to a single-element array for this purpose and
+a pointer past the last element of an array \tcode{x} of $n$ elements
 is considered to be equivalent to a pointer
-to a hypothetical element $\tcode{x}[n]$ for this purpose;
-see \ref{basic.compound}.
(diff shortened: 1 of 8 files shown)


jensmaurer:
CWG2419: What about "hypothetical n-th array element", except that gets us into the 0-th war. Maybe.

burblebee:
@jensmaurer 
> CWG2419: What about "hypothetical n-th array element", except that gets us into the 0-th war. Maybe.

It's an improvement, and I think the 0-th war isn't worsened by it - will apply.  Thanks!

zygoloid:
The newest wording here is that in [expr.add], which uses "array element *i* of `x`"; I've updated all the other places to use that same formulation (example: "hypothetical array element *n* of `x`").

PR #3486, opened 2019-11-18
Motions 2019 11 lwg 16: P1870R1 forwarding-range<T> is too subtle

### LINKED ISSUE #3418: [2019-11 LWG Motion 16]  P1870R1 forwarding-range<T> is too subtle
This paper resolves NB comments US279 and US279 and GB280.

burblebee:
Fixes #3418. 

Issues with paper:
* [range.access.begin]/p1 adds "Given a subexpression E and an lvalue t that denotes the same object as E", but E is a subexpression, not an object.
        Same for [range.access.end]/p1, [range.access.rbegin]/p1, and [range.access.rend]/p1.
* [range.range] range-impl is removed but is still referenced in p2 and p3.

Other editorial issues:
* cv-unqualified or \cv-unqualified?
      The spec is inconsistent in its representation of cv-unqualified v. /cv/-unqualified.
* When to use \deflibconcept?
      Some concept definitions in the library use \deflibconcept, others don't.
        Also fix inconsistencies in the of \defexposconcept for \expos concepts.    
* The library is inconsistent in its use of \exposconcept v. \tcode{\placeholder

Fixes cplusplus/papers#620
Fixes cplusplus/nbballot#275
Fixes cplusplus/nbballot#276

### DIFF
--- a/papers/nxxxx.md
+++ b/papers/nxxxx.md
@@ -63,6 +63,20 @@ and after consulting the paper authors and the LWG chair,
 the corresponding changes were also applied to
 the additional range adaptors listed above.
 
+### LWG motion 16
+
+This paper removed the exposition-only concept *`range-impl`*,
+inlining it into its only remaining user, the `range` concept.
+However, two uses of *`range-impl`* were left behind.
+These have been updated and suitably adjusted
+to refer to `range` instead.
+
+LWG motion 13 ([P1394R4](http://wg21.link/p1394r4))
+added a couple of new uses of
+the exposition-only concept *`forwarding-range`*,
+which was removed by this paper.
+These uses have been replaced with `safe_range`.
+
 ## Feature test macros
 
 The feature test macro `__cpp_nontype_template_parameter_class` has been removed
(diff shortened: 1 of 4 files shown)


CaseyCarter:
> `E` is a subexpression, not an object

I think we mean to say that `t` is an lvalue that denotes the result of `E`; my core-speak is failing me here.

> [range.range] _`range-impl`_ is removed but is still referenced in p2 and p3.

The intent is that these references to _`range-impl`_ be replaced by references to `range`. P2 is still a mess, however, since it refers to "The required expressions `ranges::begin(std::forward<T>(t))` and `ranges::end(std::forward<T>(t))` ..." which need to be changed to "The required expressions `ranges::begin(t)` and `ranges::end(t)` ...".

I won't even try to rationalize this rewording as "Editorial".

jensmaurer:
> cv-unqualified or \cv-unqualified?

I think we want "cv-qualified" and "cv-unqualified". We only use \cv{} if it is a placeholder for a (specific) set of cv-qualifiers.

> When to use \deflibcWhen to use \deflibconcept?oncept?

The intent is to use \deflibconcept where we actually show the definition of a library concept (i.e. not in the header synopsis if we have an itemdecl).

> The library is inconsistent in its use of \exposconcept v. \tcode{\placeholder

Yes, there are probably still a few issues like this left over.

burblebee:
@CaseyCarter 
> I think we mean  ...
> ...
> I won't even try to rationalize this rewording as "Editorial".

I've emailed @BRevzin and lwgchair for direction on what to do...  Would have cc'ed you but don't have your email address :)

PR #3487, opened 2019-11-18
Motions 2019 11 lwg 20: P0883R2 Fixing Atomic Initialization

### LINKED ISSUE #3422: [2019-11 LWG Motion 20] P0883R2 Fixing Atomic Initialization
This resolves NB comments CA353, US351, DE018, and RU006 and LWG issue 2334

burblebee:
Fixes #3422.

Issues:
* [depr.atomics] Check my use of sections - it wasn't clear to me which and when to repeat sections prefixed with "depr.".  Please check my use of sections.

Fixes cplusplus/papers#276

Fixes cplusplus/nbballot#6
Fixes cplusplus/nbballot#18
Fixes cplusplus/nbballot#347
Fixes cplusplus/nbballot#349

### DIFF
--- a/source/future.tex
+++ b/source/future.tex
@@ -2487,3 +2487,84 @@ For Windows-based operating systems a conversion from UTF-8 to
     UTF-16 occurs.
 \end{example}
 \end{itemdescr}
+
+\rSec1[depr.atomics]{Deprecated atomic initialization}
+
+\pnum
+The header \libheaderref{atomics} has the following additions.
+
+\begin{codeblock}
+namespace std {
+  template<class T>
+    void atomic_init(volatile atomic<T>*, typename atomic<T>::value_type) noexcept;
+  template<class T>
+    void atomic_init(atomic<T>*, typename atomic<T>::value_type) noexcept;
+
+  #define ATOMIC_VAR_INIT(value) @\seebelow@
+
+  #define ATOMIC_FLAG_INIT @\seebelow@
+}
+\end{codeblock}
+
+\rSec2[depr.atomics.nonmembers]{Non-member functions}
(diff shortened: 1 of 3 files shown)


source/future.tex
```diff
@@ -2487,3 +2487,84 @@
     UTF-16 occurs.
 \end{example}
 \end{itemdescr}
+
+\rSec1[depr.atomics]{Deprecated atomic initialization facets}
```
jensmaurer: "facet" sounds like a locale facet. Just remove the word.

source/future.tex
```diff
@@ -2487,3 +2487,84 @@
...
+\indexlibraryglobal{ATOMIC_VAR_INIT}%
+\begin{itemdecl}
+#define ATOMIC_VAR_INIT(value) @\seebelow@
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+The macro expands to a token sequence suitable for constant initialization of
+an atomic variable of static storage duration of a type that
+is initialization-compatible with \tcode{value}.
```
jensmaurer: What does "initialization-compatible" mean? That's not a defined term in C++, I believe.

jensmaurer:
The section labels in [depr] look fine to me.

source/future.tex
burblebee: Good catch.  Will email the author.

CaseyCarter: I find it very likely that the author has no idea what this wording - which we've been shipping since C++11 - is supposed to mean. We could file an LWG issue to have it cleaned up, but given that this now exists only in Annex D it's likely such an issue will never be addressed.

PR #3490, opened 2019-11-19
Strike mentions of undefined term "automatic object"

sdkrystian:
"Automatic object" is not a term defined by the standard. Replacing all mentions of it with "object with automatic storage duration" resolves the issue.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -5584,7 +5584,7 @@ is explicitly specified (see~\ref{class.init} and~\ref{dcl.init}).
 
 \pnum
 \begin{note}
-\indextext{order of execution!constructor and \tcode{static} objects}%
+\indextext{order of execution!constructor and static data members}%
 The order in which objects with static or thread storage duration
 are initialized is described in~\ref{basic.start.dynamic} and~\ref{stmt.dcl}.
 \end{note}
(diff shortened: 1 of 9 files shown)


zygoloid:
This term is fairly widely used, and is shorter than the alternative; maybe we should just define it? [basic.stc.auto]p1:

> Block-scope variables not explicitly declared `static`, `thread_local`, or `extern` have <i>automatic storage duration</i><ins>; such objects are called <i>automatic objects</i></ins>.

18:
@zygoloid I feel that it's not used enough to warrant it. We already have a bunch of similar definitions like "local variable" and "local entity".

jensmaurer:
I'd like to point out that @zygoloid's suggested change equates variables with objects, which we want to move away from. (There is a separate automatic object for each invocation of a function containing a non-static, non-thread-local, non-extern variable declaration.)

jensmaurer:
Editorial meeting: use "with X storage duration", also cover static object.

source/classes.tex
```diff
@@ -6555,7 +6555,7 @@
 \end{codeblock}
 Here the criteria for elision can
 eliminate
-the copying of the local automatic object
+the copying of the object with automatic storage duration
 \tcode{t}
```
zygoloid: This makes it sound like `t` is a name for the storage duration, not for the object.

sdkrystian: @zygoloid Changing it to "the object `t` with automatic storage duration" would resolve this

PR #3501, opened 2019-11-20
[module.private.frag] Fix indentation in example

johelegp:


johelegp:
Maybe the change was intentional, given that the language part doesn't follow the guidelines of the library (2 space indentation, in this case).

PR #3507, opened 2019-11-21
[lib] Remove 'Constructs an object of type ...' phrases

### LINKED ISSUE #3506: Remove all "Constructs a Foo by" from Effects: clauses

jensmaurer:
for constructors; this effect is implied by the
core language.  Only simple phrases are removed;
more complex sentence structures are left unchanged.

Partially addresses #3506.

### DIFF
--- a/source/future.tex
+++ b/source/future.tex
@@ -543,10 +543,7 @@ explicit strstreambuf(streamsize alsize_arg);
 \begin{itemdescr}
 \pnum
 \effects
-Constructs an object of class
-\tcode{strstreambuf},
-initializing the base class with
-\tcode{streambuf()}.
+Initializes the base class with \tcode{streambuf()}.
 The postconditions of this function are indicated in \tref{depr.strstreambuf.cons.sz}.
 \end{itemdescr}
 
@@ -567,10 +564,7 @@ strstreambuf(void* (*palloc_arg)(size_t), void (*pfree_arg)(void*));
 \begin{itemdescr}
 \pnum
 \effects
-Constructs an object of class
-\tcode{strstreambuf},
-initializing the base class with
-\tcode{streambuf()}.
+Initializes the base class with \tcode{streambuf()}.
 The postconditions of this function are indicated in \tref{depr.strstreambuf.cons.alloc}.
 
 \begin{libtab2}{\tcode{strstreambuf(void* (*)(size_t), void (*)(void*))} effects}
@@ -597,10 +591,7 @@ strstreambuf(unsigned char* gnext_arg, streamsize n,
 \begin{itemdescr}
 \pnum
 \effects
-Constructs an object of class
-\tcode{strstreambuf},
-initializing the base class with
-\tcode{streambuf()}.
+Initializes the base class with \tcode{streambuf()}.
 The postconditions of this function are indicated in \tref{depr.strstreambuf.cons.ptr}.
 
 \begin{libtab2}{\tcode{strstreambuf(charT*, streamsize, charT*)} effects}
@@ -1096,12 +1087,8 @@ explicit istrstream(char* s);
 \begin{itemdescr}
 \pnum
 \effects
-Constructs an object of class
-\tcode{istrstream},
-initializing the base class with
-\tcode{istream(\&sb)}
-and initializing \tcode{sb} with
-\tcode{strstreambuf(s,0)}.
+Initializes the base class with \tcode{istream(\&sb)} and
+\tcode{sb} with \tcode{strstreambuf(s, 0)}.
 \tcode{s} shall designate the first element of an \ntbs{}.%
 \indextext{NTBS}
 \end{itemdescr}
@@ -1115,12 +1102,8 @@ istrstream(char* s, streamsize n);
 \begin{itemdescr}
 \pnum
 \effects
-Constructs an object of class
-\tcode{istrstream},
-initializing the base class with
-\tcode{istream(\&sb)}
-and initializing \tcode{sb} with
-\tcode{strstreambuf(s,n)}.
+Initializes the base class with \tcode{istream(\&sb)}
+and \tcode{sb} with \tcode{strstreambuf(s, n)}.
 \tcode{s} shall designate the first element of an array whose length is
 \tcode{n} elements, and \tcode{n} shall be greater than zero.
 \end{itemdescr}
@@ -1195,12 +1178,8 @@ ostrstream();
 \begin{itemdescr}
 \pnum
 \effects
-Constructs an object of class
-\tcode{ostrstream},
-initializing the base class with
-\tcode{ostream(\&sb)}
-and initializing \tcode{sb} with
-\tcode{strstreambuf()}.
+Initializes the base class with \tcode{ostream(\&sb)} and
+\tcode{sb} with \tcode{strstreambuf()}.
 \end{itemdescr}
 
 \indexlibraryctor{ostrstream}%
@@ -1211,11 +1190,8 @@ ostrstream(char* s, int n, ios_base::openmode mode = ios_base::out);
 \begin{itemdescr}
 \pnum
 \effects
-Constructs an object of class
-\tcode{ostrstream},
-initializing the base class with
-\tcode{ostream(\&sb)},
-and initializing \tcode{sb} with one of two constructors:
+Initializes the base class with \tcode{ostream(\&sb)},
+and \tcode{sb} with one of two constructors:
 
 \begin{itemize}
 \item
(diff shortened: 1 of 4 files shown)


source/time.tex
```diff
@@ -9507,8 +9486,8 @@
 
 \pnum
 \effects
-Constructs a \tcode{zoned_time} by
-initializing \tcode{zone_} with \tcode{std::move(z)} and
+Initializes
+\tcode{zone_} with \tcode{std::move(z)} and
 default constructing \tcode{tp_}.
```
```suggestion
default constructs \tcode{tp_}.
```

source/time.tex
```diff
@@ -9631,8 +9612,9 @@
 
 \pnum
 \effects
-Constructs a \tcode{zoned_time} by
-initializing \tcode{zone_} with \tcode{std::move(z)} and \tcode{tp_} with \tcode{zone_->to_sys(tp, c)}.
+Initializing
```
```suggestion
Initializes
```

source/utilities.tex
```diff
@@ -5896,7 +5892,6 @@
 \begin{itemdescr}
 \pnum
 \effects
-Constructs an object of class \tcode{bitset<N>} as if by:
```
johelegp: The "as if by:" should be kept (properly capitalized).

source/utilities.tex
```diff
@@ -16056,8 +16054,10 @@
 
 \pnum
 \effects
-Constructs a \tcode{boyer_moore_horspool_searcher} object, initializing \tcode{pat_first_} with \tcode{pat_first},
-\tcode{pat_last_} with \tcode{pat_last}, \tcode{hash_} with \tcode{hf}, and \tcode{pred_} with \tcode{pred}.
+Initializes \tcode{pat_first_} with \tcode{pat_first},
```
johelegp: Should "Initializes" be on its own line? Many changes above do not follow the local convention of not doing so.

source/utilities.tex

source/time.tex

source/time.tex

source/utilities.tex

source/time.tex
```diff
@@ -9507,9 +9486,9 @@
 
 \pnum
 \effects
-Constructs a \tcode{zoned_time} by
-initializing \tcode{zone_} with \tcode{std::move(z)} and
-default constructing \tcode{tp_}.
+Initializes
+\tcode{zone_} with \tcode{std::move(z)} and
+default-constructs \tcode{tp_}.
```
johelegp: I see that above, a occurrence in numerics.tex uses a dash. Just below, you don't. I also remember one of the "Mandating the Library" papers not using a dash for such a change.

CaseyCarter: We use "default-constructed" as a hypenated adjective phrase, but verb usages like this one should be unhyphenated "default constructs" or (IMO better) "default-initializes" which is a hyphenated word of power.

source/numerics.tex
```diff
@@ -4241,8 +4228,7 @@
 \begin{itemdescr}
 \pnum
 \effects
-Constructs a \tcode{seed_seq} object
- as if by default-constructing its member \tcode{v}.
+Default-constructs its member \tcode{v}.
```
CaseyCarter: "Default-initializes the member \tcode{v}." would be better here, but we should instead `noexcept = default` the declaration in the class synopsis and remove the detailed specification completely:
* defaulted constructors default-initialize non-static data members with no default member initializers, so we needn't spell out that `v` is default-initialized
* `vector`'s default constructor is conditionally `noexcept(noexcept(Allocator()))`, which is `true` for `std::allocator` and therefore for `v`'s type `vector<result_type>`, so this constructor *should* be `noexcept` rather than "_Throws_: Nothing." per LWG policy. (I assume no one has noticed that we could strengthen `noexcept` here since the conditional `noexcept` was added to `vector`'s default constructor.)

source/time.tex
```diff
@@ -9476,9 +9455,8 @@
 
 \pnum
 \effects
-Constructs a \tcode{zoned_time} by
-initializing \tcode{zone_} with \tcode{traits::default_zone()} and
-default constructing \tcode{tp_}.
+Initializes \tcode{zone_} with \tcode{traits::default_zone()} and
+default constructs \tcode{tp_}.
```
CaseyCarter: default-initializes? (There are several occurrences of "default constructs" herein that I would prefer to be "default-initializes", I won't mark the remainder under the expectation that we'll consistently change all or none of them.)

source/utilities.tex
```diff
@@ -5896,7 +5892,7 @@
 \begin{itemdescr}
 \pnum
 \effects
-Constructs an object of class \tcode{bitset<N>} as if by:
+As if by:
```
CaseyCarter: LWG has been replacing "as if by:" with "Equivalent to:" in newer wording, but for cases like this (constructor is equivalent to other constructor) we simply use constructor delegation in the class synopsis and avoid prose specification.

source/utilities.tex
```diff
@@ -5805,9 +5805,7 @@
 \begin{itemdescr}
 \pnum
 \effects
-Constructs an object of class
-\tcode{bitset<N>},
-initializing all bits to zero.
+Initializes all bits to zero.
```
CaseyCarter: Could we say "all bits in `*this`" here? Without the implied context from "Constructs a meow of class woof," this suggests that it initializes ALL BITS EVERYWHERE IN THE UNIVERSE.

source/iostreams.tex
```diff
@@ -5926,8 +5914,7 @@
 \begin{itemdescr}
 \pnum
 \effects
-Constructs an object of class
-\tcode{basic_ostream}, initializing the base class subobject with
+Initializing the base class subobject with
```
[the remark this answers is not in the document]
CaseyCarter: This is my serious "request changes" comment - the others are simply suggestions.

source/time.tex
jensmaurer: Thanks, fixed.

source/iostreams.tex
```diff
@@ -1830,9 +1824,7 @@
 \begin{itemdescr}
 \pnum
 \effects
-Constructs an object of class
-\tcode{basic_ios}\iref{ios.base.cons}
-leaving its member objects uninitialized. The object
+Leaves its member objects uninitialized. The object
```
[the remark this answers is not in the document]
jensmaurer: Replaced its -> the.
("member objects" should be "subobjects", but that's for a different day.)

source/numerics.tex
jensmaurer: Please open a separate editorial issue for that. Let's not accumulate too much burden on this one.

source/time.tex
jensmaurer: Make it a separate editorial issue, please. I'm trying hard to limit this pull request to the "blatantly obvious" changes, ignoring all the other wording nightmares in this section.

source/utilities.tex
jensmaurer: -> separate editorial issue, please.

source/utilities.tex
jensmaurer: Why does my computer still work, then?

Fixed.

source/iostreams.tex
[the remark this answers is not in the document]
jensmaurer: Uh, thanks. Fixed.

source/iostreams.tex
```diff
@@ -3044,9 +3036,7 @@
 \begin{itemdescr}
 \pnum
 \effects
-Constructs an object of class
-\tcode{basic_streambuf<charT, traits>}
-and initializes:\footnote{The default constructor is protected for class
+Initializes:\footnote{The default constructor is protected for class
```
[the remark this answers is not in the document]
jensmaurer: Fixed.

source/future.tex
```diff
@@ -1115,12 +1102,8 @@
...
 \pnum
 \effects
-Constructs an object of class
-\tcode{istrstream},
-initializing the base class with
-\tcode{istream(\&sb)}
-and initializing \tcode{sb} with
-\tcode{strstreambuf(s,n)}.
+Initializes the base class with \tcode{istream(\&sb)}
+and \tcode{sb} with \tcode{strstreambuf(s,n)}.
```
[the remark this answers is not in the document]
jensmaurer: Fixed.

source/future.tex
```diff
@@ -1096,12 +1087,8 @@
...
 \pnum
 \effects
-Constructs an object of class
-\tcode{istrstream},
-initializing the base class with
-\tcode{istream(\&sb)}
-and initializing \tcode{sb} with
-\tcode{strstreambuf(s,0)}.
+Initializes the base class with \tcode{istream(\&sb)} and
+\tcode{sb} with \tcode{strstreambuf(s,0)}.
```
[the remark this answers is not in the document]
jensmaurer: Fixed.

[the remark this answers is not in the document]
johelegp: I didn't suggest this in case it was locally consistent.

jensmaurer:
Rebased.

source/utilities.tex
zygoloid: We have historically considered changing "As if by" to "Equivalent to" to not be editorial, because "Equivalent to" has implications for the other descriptive elements as described in [structure.specifications]p4 which are not (clearly) implied by "As if by" wording.

PR #3508, opened 2019-11-21
[lib] Replace 'this subclause' with numbered subclause references.

### LINKED ISSUE #3354: [lib] Refer to subclause being introduced rather than "this subclause"
#3350 just changed an occurrence of "This subclause" to "Subclause \ref{parent subclause}". There are many other such occurrences of "This subclause" in introductory paragraphs such as .general and .pre[amble] that actually mean the parent subclause. Should those be changed accordingly?

jensmaurer:
Fixes #3354.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -960,7 +960,7 @@ for piecewise construction of the elements of the \tcode{pair} object.
 
 \pnum
 \indexlibraryglobal{tuple}%
-This subclause describes the tuple library that provides a tuple type as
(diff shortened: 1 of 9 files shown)


source/utilities.tex
```diff
@@ -16233,7 +16233,7 @@
 The behavior of a program is undefined if:
 \begin{itemize}
 \item
-  an instantiation of a template specified in this subclause
+  an instantiation of a template specified in this subclause~\ref{meta}
```
```suggestion
  an instantiation of a template specified in subclause~\ref{meta}
```

PR #3513, opened 2019-11-24
[over.match.list] Acknowledge [over.ics.list] and clarify text.

### LINKED ISSUE #3372: [over.match.list] Acknowledge entry from 12.4.3.1.5 and fix presentation
CWG in Belfast:
In 12.3.1.7 [over.match.list], acknowledge entry from 12.4.3.1.5. Then, move the (to-be-struck, but we actually aren't striking) sentence into the first bullet avoid awkward ordering in presentation.

(based on CWG 2194 = NAD)

jensmaurer:
Fixes #3372

### DIFF
--- a/source/overloading.tex
+++ b/source/overloading.tex
@@ -1479,24 +1479,26 @@ the implicit object parameter of the conversion functions.
 \pnum
 When objects of non-aggregate class type \tcode{T} are
 list-initialized such that \ref{dcl.init.list} specifies that overload resolution
-is performed according to the rules in this subclause, overload resolution selects the
-constructor in two phases:
+is performed according to the rules in this subclause
+or when forming a list-initialization sequence according to \ref{over.ics.list},
+overload resolution selects the constructor in two phases:
 
 \begin{itemize}
 \item
-Initially, the candidate functions are the initializer-list constructors\iref{dcl.init.list}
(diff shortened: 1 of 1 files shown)


source/overloading.tex
```diff
@@ -1479,8 +1479,11 @@
 \pnum
 When objects of non-aggregate class type \tcode{T} are
 list-initialized such that \ref{dcl.init.list} specifies that overload resolution
-is performed according to the rules in this subclause, overload resolution selects the
-constructor in two phases:
+is performed according to the rules in this subclause
+or when forming a list-initialization sequence according to \ref{over.ics.list},
+overload resolution selects the constructor in two phases,
+except that the first phase is omitted
+if the initializer list has no elements and \tcode{T} has a default constructor.
```
zygoloid: I think it would be clearer to inline this into the first bullet:

> * If the initializer list is not empty or `T` has no default constructor, overload resolution is first performed where the candidate functions are the initializer-list constructors ([dcl.init.list]) of the class `T` and the argument list consists of the initializer list as a single argument.
> * Otherwise, or if no viable initializer-list constructor is found, overload resolution is performed again, where the candidate functions are all the constructors of the class `T` and the argument list consists of the elements of the initializer list.

jensmaurer: Yes, that reads better.

Issue #3457, opened 2019-11-14
[ranges] Parameter of range refinement should be named R


johelegp:
Following the practice of iterator concepts, `range` refinements should name their to-be-`range` parameter `R` instead of `T`. A handful of exposition-only concepts and `output_range<R, T>` (inevitably) already do.

jensmaurer:
Could you please be a little more specific? Sometimes, it's actually a "view" that is called V.

johelegp:
Right. I mean that the definition of concepts that refine `range` should name its parameter that will eventually be checked as `range<T>` so that it looks like `range<R>` instead.

The following already use `R` that way:
```C++
  template<class R, class T>
    concept output_range =
  template<class R>
    concept @\exposconcept{tiny-range}@ =
```
The following should change `T` to be `R`, similarly changing their specification to use `R` models `C` rather than `T` models `C`.
```C++
  template<class T>
    concept range =
  template<class T>
    concept sized_range =
  template<class T>
    concept view =
  template<class T>
    concept input_range =
  template<class T>
    concept forward_range =
  template<class T>
    concept bidirectional_range =
  template<class T>
    concept random_access_range =
  template<class T>
    concept contiguous_range =
  template<class T>
    concept common_range =
  template<class T>
    concept viewable_range =
template<class T>
  concept @\defexposconcept{forwarding-range}@ =
template<class R>
  concept @\defexposconcept{simple-view}@ =
```

Right now, the iterator concepts mostly use `I` (sometimes `In` or `Out`), whereas the range concepts mostly use `T`.

johelegp:
> Could you please be a little more specific? Sometimes, it's actually a "view" that is called V.

Indeed, the range adaptors use `V` for their `view` parameter. Except for the most recent additions that used `R` instead, creating inconsistency.

jensmaurer:
@CaseyCarter , are you ok with the suggested T -> R rename?

CaseyCarter:
I'm ambivalent. Pre-concepts, we used the names of template parameters to indicate requirements that the template argument must fulfill. Post-concepts, we generally pick a name that's short but still suggestive of the requirements that are explicit in a _type-constraint_ or _requires-clause_ (`Range R`, `forward_iterator I`). Neither of those motivations apply to the template parameters for concepts, so I'm not sure it makes to use something other than the generic `T` for "type" to name a type parameter for a concept, for example.

[...]

PR #3514, opened 2019-11-25
[ranges.syn, range.adaptors] Name view template parameter V for consistency

johelegp:
P1035 used `R` for such template parameters, introducing inconsistency. We could either restore consistency this way, or rename the `V`s to `R` once the converting constuctors of the other adaptors are removed, as I understand will happen.


jensmaurer:
See #3457

PR #3528, opened 2019-12-06
[library.headers],[cstdint.syn.2], [cstdint.syn.2] is not in line with intention. Also fixes #3521

### LINKED ISSUE #3521: [cstdint.syn.2] is not in line with intention
I find 

> The header defines all types and macros the same as the C standard library header <stdint.h>.

to not fully reflect the intention as the 

> defines... same as the C standard library header

would mean, that cstdint also needs to define non std:: version of aliases. This is also against: 

[library.5] 
> the contents of each header cname is the same as that of the corresponding header name.h as specified in the C standard library. In the C++ standard library, however, the declarations (except for names which are defined as macros in C) are within namespace scope of the namespace std. It is unspecified whether these names (including any overloads added in [language.support] through [thread] and [depr]) are first declared within the global namespace scope and are then injected into namespace std by explicit using-declarations.

I propose to reword the [cstdint.syn.2] or remove it (since library.5 already handles this case).

In case of rewording it could sound like:

> The header defines all types and macros the same as the C standard library header <stdint.h>, except for the types, which are defined in the namespace std.

dawidpilarski:
this is a fix for issue https://github.com/cplusplus/draft/issues/3521

- removed cstdint.syn bullet telling, that cheader provide the definitions same as header.h
- added bullet to the [library.headers] saying, that if there is corresponding typedef in the header.g for the cheader the typedefs must alias the same type.

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -1328,6 +1328,11 @@ It is unspecified whether these names (including any overloads added in
 are first declared within the global namespace scope
 and are then injected into namespace \tcode{std} by explicit
 \grammarterm{using-declaration}{s}\iref{namespace.udecl}.
+\begin{note}
+Every typedef declared in header \tcode{<c\placeholder{name}>}
+denotes the same type as the existing corresponding typedef
+in the C standard library header \tcode{<\placeholder{name}.h>}
+\end{note}
 
 \pnum
 Names which are defined as macros in C shall be defined as macros in the \Cpp{}
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -1330,7 +1330,7 @@ and are then injected into namespace \tcode{std} by explicit
 \grammarterm{using-declaration}{s}\iref{namespace.udecl}.
 \begin{note}
 Every typedef declared in header \tcode{<c\placeholder{name}>}
-denotes the same type as the existing corresponding typedef
+denotes the same type as the corresponding typedef
 in the C standard library header \tcode{<\placeholder{name}.h>}
 \end{note}
 
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -1328,11 +1328,11 @@ It is unspecified whether these names (including any overloads added in
 are first declared within the global namespace scope
 and are then injected into namespace \tcode{std} by explicit
 \grammarterm{using-declaration}{s}\iref{namespace.udecl}.
-\begin{note}
+
+\pnum
 Every typedef declared in header \tcode{<c\placeholder{name}>}
 denotes the same type as the corresponding typedef
 in the C standard library header \tcode{<\placeholder{name}.h>}
-\end{note}
 
 \pnum
 Names which are defined as macros in C shall be defined as macros in the \Cpp{}
--- a/source/support.tex
+++ b/source/support.tex
@@ -1776,10 +1776,6 @@ plus function macros of the form:
   [U]INT{8 16 32 64 MAX}_C
 \end{codeblock}
 
-\pnum
-The header defines all types and macros the same as
-the C standard library header \libheader{stdint.h}.
-
 \xrefc{7.20}
 
 \rSec1[support.start.term]{Start and termination}


jensmaurer:
See discussion in #3521. We might want a general statement instead of a point fix.

dawidpilarski:
@jensmaurer 

I believe what I did is general statement.

I added general note saying, that if there are corresponding typedefs in the header.h they need to be the same. This is regarding all cname vs name.h headers. The normative text is already in place for that:

http://eel.is/c++draft/headers#5
> Except as noted in [library] through [thread] and [depr], the contents of each header cname is the same as that of the corresponding header name.h as specified in the C standard library. In the C++ standard library, however, the declarations (except for names which are defined as macros in C) are within namespace scope of the namespace std. It is unspecified whether these names (including any overloads added in [language.support] through [thread] and [depr]) are first declared within the global namespace scope and are then injected into namespace std by explicit using-declarations.

plus I removed the potentially misleading bullet from cstdint.

source/lib-intro.tex
```diff
@@ -1328,6 +1328,11 @@
 are first declared within the global namespace scope
 and are then injected into namespace \tcode{std} by explicit
 \grammarterm{using-declaration}{s}\iref{namespace.udecl}.
+\begin{note}
+Every typedef declared in header \tcode{<c\placeholder{name}>}
+denotes the same type as the existing corresponding typedef
```
jensmaurer: "existing" should go.

jensmaurer:
Ok, then please change the commit message accordingly. Currently, it reads as-if you'd only change [cstdint.syn].
Instructions: https://github.com/cplusplus/draft/wiki/Commit-message-format

jensmaurer:
For actual types, we have [extern.types].

I'm not seeing in the text that you quoted that types designated by typedefs must be the same. Consider:
```
typedef int my_int;
typedef __my_int uint32_t;

namespace std {
  typedef long __my_int;
  typedef __my_int uint32_t ;
}
```

Arguably.  the presence or absence of implementation-reserved symbols should not make a difference for "same contents", and the two typedefs for uint32_t are token-for-token identical, yet we want such an implementation to be non-conforming. Where do we currently say that, normatively?

dawidpilarski:
As to my understanding

>  the contents of each header cname is the same as that of the corresponding header name.h as specified in the C standard library

doesn't mean, that only specified typedefs definitions needs to be the same. Whole header files content needs to be the same.

So I think, that 

```
typedef int my_int;
typedef __my_int uint32_t;

namespace std {
  typedef long __my_int;
  typedef __my_int uint32_t;
}
```

will break the 

> the contents of each header cname is the same as that of the corresponding header name.h as specified in the C standard library. 

because, the content will be different with how __my_int is defined.

Standard allows differences only in where the names are defined (in namespace `std`).

Is my understanding wrong?

jensmaurer:
So, my devil implementation has:
```
// __impl.h
typedef int __my_int;
namespace std { typedef long __my_int; }
```

cname header:
```
#include "__impl.h"

namespace std {
  typedef __my_int uint32_t;
}
```
name.h header:
```
#include "__impl.h"
typedef __my_int uint32_t;
```

It seems plausible to read "contents is the same" as being satisfied here. After all, we have [basic.def.odr] if we want to say "same tokens and same meaning", and we're obviously not doing this here.

dawidpilarski:
:) You are right. Now I see your point. The note should be normative text in fact. I will change that!

jensmaurer:
... and adding normative text is usually not editorial, but (here) LWG matter. Let's ask around whether we can squint the right way in this case.

dawidpilarski:
According to the discussion in #3521 it's editorial :
https://github.com/cplusplus/draft/issues/3521#issuecomment-561662304

should I ask more people about that? Maybe on the reflectors?

jwakely:
@dawidpilarski 

> Whole header files content needs to be the same.

I don't know what t hat means. The whole problem with the current wording is that "the same" is imprecise, and effectively meaningless.

And if I've interpreted you correctly, you're wrong. This should be a conforming implementation:

```:cxx
// <stdint.h>
typedef int __int32_t;
typedef __int32_t int32_t;

// <cstdint>
namespace std {
  using int32_t = signed int;
}
```

I don't think anybody would reasonably argue that the "whole header file content" is "the same". That's why I want to change the wording to be precise, and say that the typedefs specified in the standard **denote the same types**. That doesn't prevent implementations adding other declarations to the `<name.h>` header and/or the `<cname>` header (using reserved names as appropriate) and doesn't require identical tokens.

@jensmaurer 

> I'm not seeing in the text that you quoted that types designated by typedefs must be the same.

That's certainly the intention, but if it's not clear then let's open an LWG issue.

dawidpilarski:
@jwakely can you confirm then, that the introduced change fixes that issue and it's editorial?
Should whole headers#5 disappear, because of it's impreciseness?

jwakely:
I don't get to decide what is and isn't editorial.

Removing the whole paragraph also removes the wording about the macros. The new paragraph you're adding doesn't say anything about macros.

dawidpilarski:
@jwakely @jensmaurer 

can I please ask for further guidance? Shall I:

- create LWG issue,
- ask on reflectors, whether this kind of change is editorial or not,
- write a paper,
- something else?

jwakely:
I suggest an LWG issue. https://cplusplus.github.io/LWG/lwg-active.html#submit_issue

jensmaurer:
The editorial group will get together in Prague. We can decide there.

jwakely:
It's just come to my attention that http://wg21.link/extern.types already requires `::int32_t` and `std::int32_t` to be the same type.

jensmaurer:
Editorial meeting: C standard library headers are a different thing than C++ library headers, even if they have notionally the same name. The specification is incorporating the definition of the C std library headers into the definition of the C++ &lt;cblah> header.

PR #3531, opened 2019-12-09
[expr.comma] Remove "temporary expression"

sdkrystian:
[[expr.comma] p1 sentence 5](http://eel.is/c++draft/expr.comma#1.sentence-5) says:
> If the right operand is a temporary expression, the result is a temporary expression.

This is incorrect as no such grammar term exists, and if interpreted as meaning "xvalue", redundant.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -6785,8 +6785,6 @@ The type and value of the
 result are the type and value of the right operand; the result is of the same
 value category as its right operand, and is a bit-field if its
 right operand is a bit-field.
-If the right operand is a temporary expression\iref{class.temporary},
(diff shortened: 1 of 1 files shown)


zygoloid:
I believe this wording is now redundant; the relevant normative rule is [class.temporary]/6.8.

@jensmaurer "temporary expressions" came from your temporaries wording cleanup, can you check whether this is correct?

jensmaurer:
Yes, I think that's correct. My "temporary expression" cleanup was supposed to transport object identity through comma and ternary operators (and possibly more), but I think that's now totally redundant with the "mandatory copy elision" we have.

PR #3532, opened 2019-12-09
qualifies indirectly-readable-impl's use of iter_move

cjdb:
The exposition-only concept _`indirectly-readable-impl`_ relies on
`ranges::iter_move`, but the text was missing the `ranges::` qualifier.
This commit adds it, as the intention is to use `ranges::iter_move`, not
`std::iter_move`.

cc @jwakely @CaseyCarter @ericniebler @timsong-cpp

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -1208,7 +1208,7 @@ template<class In>
       typename iter_reference_t<In>;
       typename iter_rvalue_reference_t<In>;
       { *in } -> same_as<iter_reference_t<In>>;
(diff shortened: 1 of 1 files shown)


jwakely:
Thanks, Chris. Please add the stable tag [iterator.concept.readable] to the start of your commit message (and re-push).

CaseyCarter:
I find it incredibly ironic that the commit message doesn't say "...use of `ranges::iter_move`" (No change requested) =)

PR #3536, opened 2019-12-11
[functional.syn] Move is_bind_expression_v & is_placeholder_v to be right after is_bind_expression & is_placeholder in the <functional> synopsis

brycelelbach:


jensmaurer:
Commit message shortened as suggested.

PR #3538, opened 2019-12-11
[dcl.decl] Change "templated function" to instead use "templated entity"

sdkrystian:
The [resolution for CA378](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1971r0.html) used the term "templated function", but it is unclear what exactly this means (the intent is known, so I believe this is sufficiently editorial). This should be changed to "templated entity that is a function or function template" as is used in [[dcl.dcl]](http://eel.is/c++draft/dcl.dcl#dcl.spec.auto-12).

I didn't change the comment in the example to use this wording since it seems to be a little "extra" for a comment, but that can be changed too if deemed sufficiently important.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -166,7 +166,8 @@ in that scope.
 
 \pnum
 \indextext{entity!templated}%
-A \defn{templated entity} is
+An entity is \defn{templated}
+if it is
 \begin{itemize}
 \item a template,
 \item an entity defined\iref{basic.def} or created\iref{class.temporary}


source/declarations.tex
```diff
@@ -2172,7 +2172,8 @@
 The optional \grammarterm{requires-clause}\iref{temp.pre} in an
 \grammarterm{init-declarator} or \grammarterm{member-declarator}
 shall be present only if the declarator declares a
-templated function\iref{dcl.fct}.
+a templated entity that is a function
```
johelegp:
```suggestion
templated entity that is a function
```

zygoloid:
The intent when introducing the term "templated entity" was that we could then use it to talk about "templated functions", "templated variables", and so on. We should define those terms rather than removing uses of them.

18:
Perhaps we add this below the definition of templated entity:
> A *templated function* is a templated entity that is a function or a function template. A *templated variable* is a templated entity that is a variable or a variable template. A *templated class* is a templated entity that is a class or a class template.

zygoloid:
> Perhaps we add this below the definition of templated entity:

Yes, perhaps. Another option to consider is to define only the adjective "templated", not the noun phrase "templated entity". ("An entity is *templated* if it is [...]")

jensmaurer:
Editorial meeting: Go with "An entity is template if it is..."

sdkrystian:
@jensmaurer is a *template* or is *templated*?

tkoeppe:
@sdkrystian: That was a typo, already fixed.

sdkrystian:
Whoops, looks like deleting commits closes the PR. Changes applied.

PR #3542, opened 2019-12-13
[cmp.categories.pre] Remove unused enumerators

### LINKED ISSUE #3541: [cmp.categories.pre] nonequal and nonequivalent are non-used
After [P1959R0](https://wg21.link/P1959R0) "Remove `std::weak_equality` and `std::strong_equality`" was adopted, [N4842](https://wg21.link/N4842) [17.11.2.1 [cmp.categories.pre]/1](https://eel.is/c++draft/cmp.categories.pre#1) says:

> Each is specified in terms of an exposition-only data member named `value` whose value typically corresponds to that of an enumerator from one of the following exposition-only enumerations:
> `enum class` *`eq`* `{` *`equal`* `= 0,` *`equivalent`* `=` *`equal`*`,` *`nonequal`* `= 1,` *`nonequivalent`* `=` *`nonequal`* `};` *// exposition only*

However, *`nonequal`* and *`nonequivalent`* are now unused. As they're exposition-only enumerators, they should be editorially removed.

CaseyCarter:
Fixes #3541.

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -4056,9 +4056,8 @@ whose value typically corresponds to that of an enumerator
 from one of the following exposition-only enumerations:
 
 \begin{codeblock}
-enum class @\placeholdernc{eq}@ { @\placeholdernc{equal}@ = 0, @\placeholdernc{equivalent}@ = @\placeholdernc{equal}@ };            // \expos
-enum class @\placeholdernc{ord}@ { @\placeholdernc{less}@ = -1, @\placeholdernc{greater}@ = 1 };                  // \expos
-enum class @\placeholdernc{ncmp}@ { @\placeholdernc{unordered}@ = -127 };                       // \expos
+enum class @\placeholdernc{ord}@ { @\placeholdernc{equal}@ = 0, @\placeholdernc{equivalent}@ = @\placeholdernc{equal}@, @\placeholdernc{less}@ = -1, @\placeholdernc{greater}@ = 1 }; // \expos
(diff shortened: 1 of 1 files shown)


zygoloid:
Do we still need the `eq` enum at all? Perhaps we could merge it into `ord`?

CaseyCarter:
> Do we still need the `eq` enum at all? Perhaps we could merge it into `ord`?

Yes, this is feasible: `eq` is only used as a constructor parameter type, and the classes with such a constructor already have a constructor that takes `ord`.

tkoeppe:
@jensmaurer: sorry, I missed that ping; I agree that a more aggressive cleanup is a nice idea.

CaseyCarter:
I've made the requested changes.

jwakely:
We don't really need the `equal` enumerator either, it's used in exactly one place and we could just use `equivalent` there. It might even make it clearer that equivalence is the same as equality for `strong_ordering`.

CaseyCarter:
> We don't really need the `equal` enumerator either, it's used in exactly one place and we could just use `equivalent` there. It might even make it clearer that equivalence is the same as equality for `strong_ordering`.

I'd vaguely prefer to keep the `equal` vs. `equivalent` distinction to reflect "substitutability", but it's not a hill I'm willing to die on.

jwakely:
OK, I don't feel strongly about it either.

CaseyCarter:
Ping. Could someone please remove the changes-requested label so this is visible again?

jensmaurer:
Done.

PR #3545, opened 2019-12-14
[cpp.replace] Distribute examples from [cpp.scope]

### LINKED ISSUE #3544: Examples in [cpp.scope]p3 onwards are misplaced
The examples in [cpp.scope]p3-p9 (most of [cpp.scope]) have very little to do with macro definition scope, and are in fact examples for all of [cpp.replace]. They should be relocated next to the wording that they are exemplifying.

jensmaurer:
where they fit more naturally, omitting some of the
now-redundant introductory phrases.

Fixes #3544

### DIFF
--- a/source/preprocessor.tex
+++ b/source/preprocessor.tex
@@ -869,6 +869,26 @@ and the two replacement lists are identical,
 otherwise the program is ill-formed.
 
 \pnum
+\begin{example}
+The following sequence is valid:
+\begin{codeblock}
+#define OBJ_LIKE      (1-1)
+#define OBJ_LIKE      @\tcode{/* white space */ (1-1) /* other */}@
+#define FUNC_LIKE(a)   ( a )
+#define FUNC_LIKE( a )(     @\tcode{/* note the white space */ \textbackslash}@
+                a @\tcode{/* other stuff on this line}@
+                  @\tcode{*/}@ )
+\end{codeblock}
+But the following redefinitions are invalid:
+\begin{codeblock}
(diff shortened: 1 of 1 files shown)


source/preprocessor.tex
```diff
@@ -1194,6 +1267,51 @@
 \tcode{\#\#}
 operators is unspecified.
 
+\pnum
+\begin{example}
+To illustrate the rules for creating character string literals
```
opensdh: We can probably do without the "To illustrate" here and below now that they're in the right places.

source/preprocessor.tex
```diff
@@ -983,6 +1013,29 @@
 arguments that would otherwise act as preprocessing directives,\footnote{A \grammarterm{conditionally-supported-directive} is a preprocessing directive regardless of whether the implementation supports it.}
 the behavior is undefined.
 
+\pnum
+\begin{example}
+The following defines a function-like
+macro whose value is the maximum of its arguments.
+It has the advantages of working for any compatible types of the arguments
+and of generating in-line code without the overhead of function calling.
```
zygoloid: It's not clear what these are advantages compared to. Maybe we could just strike this sentence, and describe the macro form of `max` has having only disadvantages?

source/preprocessor.tex
jensmaurer: Fixed.

source/preprocessor.tex
jensmaurer: Agreed. Removed.

PR #3554, opened 2019-12-15
[std] Use $E$ as expression placeholder.

### LINKED ISSUE #3271: Upper-case vs Lower-case for expressions

jensmaurer:
Fixes #3271.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -300,38 +300,38 @@ An expression or conversion is \defn{potentially evaluated} unless it is
 an unevaluated operand\iref{expr.prop},
 a subexpression thereof, or
 a conversion in an initialization or conversion sequence in such a context.
-The set of \defn{potential results} of an expression \tcode{e} is
+The set of \defn{potential results} of an expression $E$ is
 defined as follows:
 \begin{itemize}
-\item If \tcode{e} is an
+\item If $E$ is an
 \grammarterm{id-expression}\iref{expr.prim.id}, the set
-contains only \tcode{e}.
-\item If \tcode{e} is a subscripting operation\iref{expr.sub} with
+contains only $E$.
+\item If $E$ is a subscripting operation\iref{expr.sub} with
 an array operand, the set contains the potential results of that operand.
-\item If \tcode{e} is a class member access
+\item If $E$ is a class member access
 expression\iref{expr.ref} of the form
-\tcode{e1 . \opt{template} e2}
+\tcode{$E_1$ . \opt{template} $E_2$}
 naming a non-static data member,
-the set contains the potential results of \tcode{e1}.
-\item If \tcode{e} is a class member access expression
+the set contains the potential results of $E_1$.
+\item If $E$ is a class member access expression
 naming a static data member,
 the set contains the \grammarterm{id-expression} designating the data member.
-\item If \tcode{e} is a pointer-to-member
+\item If $E$ is a pointer-to-member
 expression\iref{expr.mptr.oper} of the form
-\tcode{e1 .* e2},
-the set contains the potential results of \tcode{e1}.
-\item If \tcode{e} has the form \tcode{(e1)}, the set contains the
-potential results of \tcode{e1}.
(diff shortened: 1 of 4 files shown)


source/basic.tex
```diff
@@ -300,38 +300,38 @@
...
 \grammarterm{id-expression}\iref{expr.prim.id}, the set
-contains only \tcode{e}.
-\item If \tcode{e} is a subscripting operation\iref{expr.sub} with
+contains only $E$.
+\item If $E$ is a subscripting operation\iref{expr.sub} with
 an array operand, the set contains the potential results of that operand.
-\item If \tcode{e} is a class member access
+\item If $E$ is a class member access
 expression\iref{expr.ref} of the form
 \tcode{e1 . \opt{template} e2}
```
zygoloid: Can we also convert the `e1` and `e2` here and below?

source/expressions.tex
```diff
@@ -600,11 +600,11 @@
...
 When an lvalue-to-rvalue conversion
-is applied to an expression \tcode{e}, and either
+is applied to an expression $E$, and either
 \begin{itemize}
-\item \tcode{e} is not potentially evaluated, or
-\item the evaluation of \tcode{e} results in the evaluation of a member
-      \tcode{ex} of the set of potential results of \tcode{e}, and \tcode{ex}
+\item $E$ is not potentially evaluated, or
+\item the evaluation of $E$ results in the evaluation of a member
+      \tcode{ex} of the set of potential results of $E$, and \tcode{ex}
```
zygoloid: `ex` is an expression too.

source/basic.tex
jensmaurer: Fixed.

source/expressions.tex

source/basic.tex
```diff
@@ -300,38 +300,38 @@
...
+\item If $E$ is a pointer-to-member
 expression\iref{expr.mptr.oper} of the form
-\tcode{e1 .* e2},
-the set contains the potential results of \tcode{e1}.
-\item If \tcode{e} has the form \tcode{(e1)}, the set contains the
-potential results of \tcode{e1}.
-\item If \tcode{e} is a glvalue conditional
+\tcode{$E_1$ .* $E_2$},
+the set contains the potential results of $E_1$.
+\item If $E$ has the form ($E_1$), the set contains the
```
zygoloid: We lost the `\tcode` for the parens here.

source/expressions.tex
```diff
@@ -600,13 +600,13 @@
...
 \begin{itemize}
-\item \tcode{e} is not potentially evaluated, or
-\item the evaluation of \tcode{e} results in the evaluation of a member
-      \tcode{ex} of the set of potential results of \tcode{e}, and \tcode{ex}
+\item $E$ is not potentially evaluated, or
+\item the evaluation of $E$ results in the evaluation of a member
+      $E_x$ of the set of potential results of $E$, and $E_x$
       names a variable \tcode{x} that is not odr-used by
-      \tcode{ex}\iref{basic.def.odr},
+      $E_x$\iref{basic.def.odr},
```
zygoloid: The subscript x should probably be in code font since that's the same `x` as the variable name.

source/basic.tex
jensmaurer: Fixed.

PR #3557, opened 2019-12-15
[std] Harmonize comments indicating errors.

### LINKED ISSUE #3126: Inconsistencies in comments for ill-formed/undefined behavior examples
We're inconsistent in how we format/word comments in examples for ill-formed and undefined behavior code.  The following cases show a few of the variations used:
```
utilities.tex:any_cast<string&>(y);                       // error; cannot
g<int>(0);        // error, substituting parameter type instantiates \tcode{A<int>}
S<char> s1;                     // error: no matching constructor
template<> enum A<char>::E : char { echar };        // ill-formed, \tcode{A<char>::E} was instantiated
struct Inner2;                // ill-formed, no diagnostic required
// ill-formed, no diagnostic required: the two expressions are functionally equivalent but not equivalent
void A<Y>::B<double>::mf2() { }       // ill-formed; \tcode{B<double>} is specialized but
template<class T> void f(T::R);         // ill-formed (no diagnostic required), attempt to declare
p->x.j = 99;                            // undefined: modifies a const subobject
int m = g(false);   // undefined behavior due to access of \tcode{x.n} outside its lifetime
C().C::~C();      // undefined behavior: temporary of type \tcode{C} destroyed twice
pb->f();          // undefined behavior, lifetime of \tcode{*pb} has ended
```

Suggestion: Change all to either of the following as appropriate:
```
  // ill-formed, no diagnostic required: <explanation>
  // error: <explanation>
  // undefined behavior: <explanation>
```
where `: <explanation>` can be omitted if it is clear from the surrounding text.

This does not apply in compatibility.tex, where we should stick with:
`       f(u8"text");                    // ill-formed; previously well-formed
`

jensmaurer:
Fixes #3126.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -318,7 +318,7 @@ declare an overloaded\iref{over} function \tcode{f()} and not
 simply a single function \tcode{f()} twice. For the same reason,
 \begin{codeblock}
 struct S { int a; };
-struct S { int a; };            // error, double definition
+struct S { int a; };            // error: double definition
 \end{codeblock}
 is ill-formed because it defines \tcode{S} twice.
 \end{example}
@@ -2477,7 +2477,7 @@ struct Y {
 };
 
 Y a;
-int b = a;          // error, \tcode{a.operator X().operator int()} not tried
+int b = a;          // error: no viable conversion (\tcode{a.operator X().operator int()} not considered)
 int c = X(a);       // OK: \tcode{a.operator X().operator int()}
 \end{codeblock}
 \end{example}
@@ -2500,7 +2500,7 @@ struct Y : X {
 };
 
 void f(Y& a) {
-  if (a) {          // ill-formed: \tcode{X::operator int()} or \tcode{Y::operator char()}
+  if (a) {          // error: ambiguous between \tcode{X::operator int()} and \tcode{Y::operator char()}
   }
 }
 \end{codeblock}
@@ -2643,7 +2643,7 @@ struct Z {
 
 void h(Z z) {
   Y y1(z);          // OK: direct-initialization
-  Y y2 = z;         // ill-formed: copy-initialization
+  Y y2 = z;         // error: no conversion function candidate for copy-initialization
   Y y3 = (Y)z;      // OK: cast notation
 }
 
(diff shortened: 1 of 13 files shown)


source/classes.tex
```diff
@@ -2477,7 +2477,7 @@
 };
 
 Y a;
-int b = a;          // error, \tcode{a.operator X().operator int()} not tried
+int b = a;          // error: \tcode{a.operator X().operator int()} not tried
```
zygoloid: The text after the colon here is not a description of the error. Maybe

// error: no viable conversion (`a.operator X().operator int()` not considered)

source/classes.tex
```diff
@@ -2500,7 +2500,7 @@
 };
 
 void f(Y& a) {
-  if (a) {          // ill-formed: \tcode{X::operator int()} or \tcode{Y::operator char()}
+  if (a) {          // error: \tcode{X::operator int()} or \tcode{Y::operator char()}
```
zygoloid: // error: ambiguous between `X::operator int()` and `Y::operator char()`

maybe?

source/classes.tex
```diff
@@ -2643,7 +2643,7 @@
 
 void h(Z z) {
   Y y1(z);          // OK: direct-initialization
-  Y y2 = z;         // ill-formed: copy-initialization
+  Y y2 = z;         // error: copy-initialization
```
zygoloid: // error: no candidate conversion functions for copy-initialization

or something like that?

source/overloading.tex
```diff
@@ -624,7 +623,7 @@
 public:
   C(int);
 };
-T a = 1;            // ill-formed: \tcode{T(C(1))} not tried
+T a = 1;            // error: \tcode{T(C(1))} not tried
```
zygoloid: // error: no viable conversion (`T(C(1))` not considered)

source/overloading.tex
```diff
@@ -983,7 +982,7 @@
 String operator + (const String&, const String&);
 
 void f() {
- const char* p= "one" + "two";  // ill-formed because neither operand has class or enumeration type
+ const char* p= "one" + "two";  // error: neither operand has class or enumeration type
```
zygoloid: // error: cannot add two pointers; overloaded `operator+` not considered
// because neither operand has class or enumeration type

source/classes.tex

source/classes.tex

source/classes.tex

source/overloading.tex

source/overloading.tex
jensmaurer: Fixed. (Also fixed code indentation.)

PR #3563, opened 2019-12-16
[range.refinements] Simplify viewable_range

CaseyCarter:
Since pointer types cannot model `view`, `view<decay_t<T>>` and `view<remove_cvref_t<T>>` are equivalent for a type `T`. LWG prefers `remove_cvref_t` to `decay_t` in cases where the two are equivalent.


zygoloid:
This is a normative change: while it might not affect the cases under which `viewable_range` is satisfied, it does affect the subsumption relationship between `viewable_range<T>` and `view<decay_t<T>>`. Please file an LWG issue if you want to see this changed.

PR #3565, opened 2019-12-18
[range.prim.size] rewords p1.3 to make it easier to parse

cjdb:
I expressed a desire to reword [range.prim.size] p1.3 on the LWG reflector, as I felt it is currently difficult to grasp the exact requirements (from an implementer's point-of-view). Hopefully this wording conveys the necessary requirements in a more concise manner.

Little feedback was given, so I'm unsure if this is something that should first go through a full review in Prague.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -638,7 +638,7 @@ template<class T> void size(T&&) = delete;
   if it is a valid expression, and
   \tcode{T} models \tcode{forward_range}\iref{range.refinements}, and
   \tcode{iterator_t<T>} and \tcode{sentinel_t<T>}\iref{ranges.syn} model
-  \tcode{\libconcept{sized_sentinel_for}<S, I>}\iref{iterator.concept.sizedsentinel}.
+  \tcode{\libconcept{sized_sentinel_for}<sentinel_t<T>, iterator_t<T>>}\iref{iterator.concept.sizedsentinel}.
   However, \tcode{E} is evaluated only once.
 
 \item
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -634,12 +634,11 @@ template<class T> void size(T&&) = delete;
   \end{itemize}
 
 \item
-  Otherwise, \tcode{\placeholdernc{make-unsigned-like}(ranges::end(E) - ranges::begin(E))}\iref{range.subrange}
-  if it is a valid expression and
-  the types \tcode{I} and \tcode{S} of \tcode{ranges::begin(E)} and
-  \tcode{ranges::end(E)} (respectively) model both
-  \tcode{\libconcept{sized_sentinel_for}<S, I>}\iref{iterator.concept.sizedsentinel} and
-  \tcode{\libconcept{forward_iterator}<I>}.
+  Otherwise, \tcode{\placeholdernc{make-unsigned-like}(ranges::end(E) - ranges::begin(E))}\iref{ranges.syn}
+  if it is a valid expression, and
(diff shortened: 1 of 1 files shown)


source/ranges.tex
```diff
@@ -634,12 +634,11 @@
...
-  if it is a valid expression and
-  the types \tcode{I} and \tcode{S} of \tcode{ranges::begin(E)} and
-  \tcode{ranges::end(E)} (respectively) model both
-  \tcode{\libconcept{sized_sentinel_for}<S, I>}\iref{iterator.concept.sizedsentinel} and
-  \tcode{\libconcept{forward_iterator}<I>}.
+  Otherwise, \tcode{\placeholdernc{make-unsigned-like}(ranges::end(E) - ranges::begin(E))}\iref{ranges.syn}
+  if it is a valid expression, and
+  \tcode{T} models \tcode{forward_range}\iref{range.refinements}, and
+  \tcode{iterator_t<T>} and \tcode{sentinel_t<T>}\iref{ranges.syn} model
+  \tcode{\libconcept{sized_sentinel_for}<S, I>}\iref{iterator.concept.sizedsentinel}.
```
johelegp:
```suggestion
  \tcode{\libconcept{sized_sentinel_for}<sentinel_t<T>, iterator_t<T>>}\iref{iterator.concept.sizedsentinel}.
```
You're getting rid of what `S` and `I` are.

source/ranges.tex
```diff
@@ -634,12 +634,11 @@
...
-  Otherwise, \tcode{\placeholdernc{make-unsigned-like}(ranges::end(E) - ranges::begin(E))}\iref{range.subrange}
-  if it is a valid expression and
-  the types \tcode{I} and \tcode{S} of \tcode{ranges::begin(E)} and
-  \tcode{ranges::end(E)} (respectively) model both
-  \tcode{\libconcept{sized_sentinel_for}<S, I>}\iref{iterator.concept.sizedsentinel} and
-  \tcode{\libconcept{forward_iterator}<I>}.
+  Otherwise, \tcode{\placeholdernc{make-unsigned-like}(ranges::end(E) - ranges::begin(E))}\iref{ranges.syn}
+  if it is a valid expression, and
+  \tcode{T} models \tcode{forward_range}\iref{range.refinements}, and
+  \tcode{iterator_t<T>} and \tcode{sentinel_t<T>}\iref{ranges.syn} model
```
jensmaurer: No two cross-references to the same place within the same paragraph, please.

jensmaurer: How editorially-obvious is it that `iterator_t&lt;T>` and the type of `ranges::begin(E)` are actually the same type? For example, could `ranges::begin(E)` return an lvalue, which would make `iterator_t&lt;T>` a reference? (i think the answer is "no", because weakly_incrementable requires default-constructible. And the answer is also "no" for the similar question regarding `ranges::end(E)`, because semiregular also requires default-constructible. But it takes a few hops to get there.)

jensmaurer:
Please rebase and force-push.

tkoeppe:
@cjdb: Ping, are you still interested in this PR?

source/ranges.tex
jwakely: No, they cannot return references. They either return a pointer, or _decay-copy_`(something).

tkoeppe:
I'm not sure why this got closed; it wasn't meant to be.

tkoeppe:
@jensmaurer, @jwakely, @CaseyCarter: are you interested in taking this further? You can check out the last commit via `git fetch origin pull/3565/head` and create a new PR if you like.

PR #3570, opened 2019-12-19
[ranges] Do not indent codeblocks

### LINKED ISSUE #3568: [ranges] Indented \end{codeblock} causes trailing empty line
Notice the trailing empty line after the codeblock:
image
This is caused by the fact that the `\end{codeblock}` is indented:

https://github.com/cplusplus/draft/blob/8166369c5d5ce508d92ec35e72c14c0aa707e486/source/ranges.tex#L485-L487

[ranges] is the only chapter that has indented `\end{codeblock}`s:

    ranges.tex:354:  \end{codeblock}
    ranges.tex:394:  \end{codeblock}
    ranges.tex:401:  \end{codeblock}
    ranges.tex:407:  \end{codeblock}
    ranges.tex:487:  \end{codeblock}
    ranges.tex:528:  \end{codeblock}
    ranges.tex:535:  \end{codeblock}
    ranges.tex:540:  \end{codeblock}
    ranges.tex:632:    \end{codeblock}

It might be a good idea to include a rule about this in the specification guidelines wiki.

CaseyCarter:
Fixes #3568.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -348,10 +348,10 @@ expression-equivalent to:
   valid expression and its type \tcode{I} models
   \libconcept{input_or_output_iterator} with overload
   resolution performed in a context that includes the declarations:
-  \begin{codeblock}
+\begin{codeblock}
 template<class T> void begin(T&&) = delete;
 template<class T> void begin(initializer_list<T>&&) = delete;
-  \end{codeblock}
+\end{codeblock}
   and does not include a declaration of \tcode{ranges::begin}.
 
 \item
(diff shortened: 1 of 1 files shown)


jensmaurer:
Casey, I've merged my alternative pull request #3572 which also augments the check.sh script so that such atrocities never again find their way into the source code.

CaseyCarter:
> Casey, I've merged my alternative pull request #3572 which also augments the check.sh script so that such atrocities never again find their way into the source code.

Good idea - it should have occurred to me to add CI coverage.

PR #3574, opened 2019-12-20
Replace #include by import-declaration

daveedvdv:
This example was previously valid, but we've removed the implicit use of header units through #include directives.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -5691,7 +5691,7 @@ int dog(bar, bar);
 
 \begin{codeblocktu}{Module interface unit of \tcode{C2}}
 module;
-#include "bar.h" // imports header unit \tcode{"bar.h"}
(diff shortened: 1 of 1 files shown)


source/templates.tex
```diff
@@ -5691,7 +5691,7 @@
 
 \begin{codeblocktu}{Module interface unit of \tcode{C2}}
 module;
-#include "bar.h" // imports header unit \tcode{"bar.h"}
+import "bar.h" // imports header unit \tcode{"bar.h"}
```
zygoloid: This is missing the terminating semicolon. (Please also check that the comment starts in a suitable column per https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#code-blocks -- looks like we might be missing enforcement of that for `codeblocktu`.)

jensmaurer:
Please rebase and force-push.

PR #3575, opened 2019-12-20
Connect pp-import to the preprocessor grammar

daveedvdv:
Clarify how the grammar pp-import relates to control-line.

### DIFF
--- a/source/preprocessor.tex
+++ b/source/preprocessor.tex
@@ -56,7 +56,7 @@ within what would otherwise be an invocation of a function-like macro.
 \begin{bnf}\obeyspaces
 \nontermdef{control-line}\br
     \terminal{\# include} pp-tokens new-line\br
-    \opt{\terminal{export}} \terminal{import} pp-tokens new-line\br
+    import-line\br
     \terminal{\# define } identifier replacement-list new-line\br
     \terminal{\# define } identifier lparen \opt{identifier-list} \terminal{)} replacement-list new-line\br
     \terminal{\# define } identifier lparen \terminal{... )} replacement-list new-line\br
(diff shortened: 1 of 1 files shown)


source/preprocessor.tex
```diff
@@ -671,6 +676,9 @@
 \indextext{header unit!preprocessing}%
 \indextext{macro!import|(}%
 
+\pnum
+An \grammarterm{import-line} shall be of the form
```
jensmaurer: For "of the form" constraints, we usually don't show another non-terminal, but just the grammar sequence that needs to be satisfied. That said, I think the overall change is a net improvement.

daveedvdv: I noticed that, but the subsequent text refers to things like "all three forms of pp-import" etc.  So I thought it nice to keep that name.  I suppose we could change it to "all three forms above" and "the second form", etc.  Not sure if that would be better though.

zygoloid: We do the same thing in [class.conv.fct], which is much more egregious than this (it lists three productions and only means the first one). Though I do wonder if we can do better here; perhaps we could list *pp-import* as a *control-line* production (perhaps renaming it to *import-line*), and change [cpp.pre]/2 to say that a *text-line* is only matched if the line is not a preprocessor directive. (That way, an import directive that's not a *pp-import* would simply not match the preprocessor grammar, just like an unmatched `#if` or `#endif`.)

jensmaurer:
Please rebase and force-push.

tkoeppe:
@daveedvdv: Could you please take @zygoloid's suggestion into consideration? And rebase?

tkoeppe:
We're not sure if this is editorial. We agree that there's a bug in the wording, but the detailed resolution needs oversight, too, so could you please raise this proposal as a CWG issue?

jensmaurer:
Fixed by 862fe4b8bd312203dd91e2d8be232f829a2f791a (Feb 2020)
P1857R3 Modules Dependency Discovery

PR #3580, opened 2019-12-26
[basic.scope.pdecl] Fix example of self-referential initialization.

### LINKED ISSUE #3578: [basic.scope.pdecl]/1 Example is outdated
From https://stackoverflow.com/questions/59442521/use-of-variable-in-own-initializer  
After [the resolution](https://github.com/cplusplus/draft/commit/f3a6b2e671efdee5bacb74a5ea01d4bfb1900d3e) of CWG2256¹ changed the rules of an object lifetime start, `unsigned char x = x;` no longer initializes `x` with its own indeterminate value as the Example in [basic.scope.pdecl]/1 says, but triggers undefined behavior, because the initializer accesses the object whose lifetime has not started yet (because variable initialization hasn't completed).

---
1) Which, if ignore references, kinda mixes variables with objects…

jensmaurer:
Fixes #3578.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -836,8 +836,9 @@ complete declarator\iref{dcl.decl} and before its
 unsigned char x = 12;
 { unsigned char x = x; }
 \end{codeblock}
-Here the second \tcode{x} is initialized with its own (indeterminate)
-value.
+Here, the initialization of the second \tcode{x} has undefined behavior,
+because the initializer accesses the second \tcode{x}
+outside its lifetime\iref{basic.life}.
 \end{example}
 
 \pnum


source/basic.tex
```diff
@@ -836,8 +836,9 @@
 unsigned char x = 12;
 { unsigned char x = x; }
 \end{codeblock}
-Here the second \tcode{x} is initialized with its own (indeterminate)
-value.
+Here, the initialization of the second \tcode{x} is undefined behavior,
```
zygoloid: "is" -> "has"

jensmaurer: Fixed.

PR #3581, opened 2019-12-26
[class.this] Member functions are not cv-qualified. 

### LINKED ISSUE #3579: [dcl.fct]/7 [class.this]/3 cv-qualified function types
[The Note in [dcl.fct]/7](http://eel.is/c++draft/dcl.fct#7.note-1) says:
> A function type that has a cv-qualifier-seq is not a cv-qualified type; there are no cv-qualified function types.

[[class.this]/3](http://eel.is/c++draft/class.this#3) says:
> A cv-qualified member function can be called on an object-expression only if the object-expression is as cv-qualified or less-cv-qualified than the member function.

I assume «as cv-qualified or less-cv-qualified» means «not [more cv-qualified](http://eel.is/c++draft/basic.type.qualifier#def:more_cv-qualified)», as defined in [[basic.type.qualifier]/4](http://eel.is/c++draft/basic.type.qualifier#4):
> There is a partial ordering on cv-qualifiers, so that a type can be said to be _more cv-qualified_ than another. Table 11 shows the relations that constitute this ordering.

---
If a function type can't be cv-qualified, how could one compare whether a non-stastic member function (type(?)) more or less cv-qualified than an object-expression (type(?))?

jensmaurer:
Fixes #3579.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -1057,14 +1057,10 @@ In the body of a non-static\iref{class.mfct} member function, the
 keyword \tcode{this} is a prvalue whose value is
 a pointer to the object for which the function is called.
 \indextext{\idxcode{this}!type of}%
-The type of \tcode{this} in a member function of a class \tcode{X} is
-\tcode{X*}.
-If the member function is declared \tcode{const}, the type of
-\tcode{this} is \tcode{const} \tcode{X*},
-if the member function is declared \tcode{volatile}, the type of
-\tcode{this} is \tcode{volatile} \tcode{X*}, and if the member function
-is declared \tcode{const} \tcode{volatile}, the type of \tcode{this} is
-\tcode{const} \tcode{volatile} \tcode{X*}.
+The type of \tcode{this} in a member function
+whose type has a \grammarterm{cv-qualifier-seq} \cv{} and
+whose class is \tcode{X}
+is ``pointer to \cv{} \tcode{X}''.
 \begin{note}
 Thus in a const member function, the object for which the function is
 called is accessed through a const access path.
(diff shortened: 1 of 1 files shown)


jensmaurer:
We should probably move the restriction on cv-qualifiers of the member function vs. the object expression to [expr.ref], where the expression is actually formed. @zygoloid , any opinion here?

source/classes.tex
```diff
@@ -1089,14 +1085,18 @@
...
 
 \pnum
-A cv-qualified member function can be called on an
-object-expression\iref{expr.ref} only if the object-expression is as
-cv-qualified or less-cv-qualified than the member function.
+A member function whose type has a \grammarterm{cv-qualifier-seq} \cvqual{cv1}
+can be called on an
+object expression\iref{expr.ref} of type \cvqual{cv2} \tcode{T} only
+if \cvqual{cv2} is not more cv-qualified than
+\cvqual{cv1}\iref{basic.type.qualifier}.
```
timsong-cpp: "not more" is not right. You can't call a `volatile` member function on a `const` object, even though "volatile" is not more cv-qualified than "const" (or vice-versa - they are unordered).

zygoloid:
The restriction on cv-qualifiers of the member function vs the object expression are already enforced in two other places: in overload resolution and in [expr.call]/7 (which says "the `this` parameter of the function is initialized with a pointer to the object of the call, converted as if by an explicit type conversion").

I wouldn't want to move this wording to [expr.ref] unless we can remove it from [expr.call], which I don't think we can, since we do still need to define the initialization of the implicit object parameter in [expr.call].

Curiously, the pointer-to-member case currently says it has the same restrictions on cv-qualification that are described in [expr.ref], which doesn't describe any such restrictions...

PR #3588, opened 2019-12-28
[std] Consistently use 'overload set'.

### LINKED ISSUE #3297: fix residual suggestions that overload resolution might not always happen
CWG has moved wording changes to make overload resolution always happen, but there are some remaining cleanups. As proposed by CWG, we should assemble a pull request for these and run them past CWG before applying.

Specifically: [over.match.call] talks about only applying to "a set of overloaded functions" (also search for that phrase elsewhere), and [expr.call]p7 suggests we could have selected a function that lacks a default argument for a parameter (that can't happen after applying the wording of editorial issue #3235 as a core issue).

jensmaurer:
Define the term in [basic.lookup] and use it throughout.
Avoid the term 'set of overloaded functions', because it
is ambiguous with the declaration view in [over.load].
(An overload set might contain functions from different
scopes that cannot be overloaded per [over.load].)

Fixes #3297.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -351,8 +351,8 @@ an expression or conversion as follows:
 \begin{itemize}
 \item
   A function is named by an expression or conversion
-  if it is the unique result of a name lookup or the selected member
-  of a set of overloaded functions~(\ref{basic.lookup}, \ref{over.match}, \ref{over.over})
+  if it is the selected member
+  of an overload set~(\ref{basic.lookup}, \ref{over.match}, \ref{over.over})
   in an overload resolution performed
   as part of forming that expression or conversion,
   unless it is a pure virtual function and either
@@ -1364,11 +1364,12 @@ The name lookup rules apply uniformly to all names (including
 such names in the context discussed by a particular rule. Name lookup
 associates the use of a name with a set of declarations\iref{basic.def} of
 that name.
-The declarations found by name lookup shall either all denote the same entity or
-shall all denote functions or function templates;
-in the latter case,
-the declarations are said to form a set of overloaded
-functions\iref{over.load}. Overload resolution\iref{over.match}
+If the declarations found by name lookup
+all denote functions or function templates,
+the declarations are said to form an \defn{overload set}.
+The declarations found by name lookup shall either
+all denote the same entity or form an overload set.
+Overload resolution~(\ref{over.match}, \ref{over.over})
 takes place after name lookup has succeeded. The access rules\iref{class.access}
 are considered only once name lookup and
 function overload resolution (if applicable) have succeeded. Only after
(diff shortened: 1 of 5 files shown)


jensmaurer:
@zygoloid, do you want a CWG review for this rather simple renaming exercise?

source/overloading.tex
```diff
@@ -701,14 +701,12 @@
 \begin{ncsimplebnf}
 postfix-expression \terminal{(} \opt{expression-list} \terminal{)}
 \end{ncsimplebnf}
-if the \grammarterm{postfix-expression} denotes a set of overloaded functions and/or
-function templates, overload resolution is applied as specified in \ref{over.call.func}.
+overload resolution is applied as specified in \ref{over.call.func}.
```
zygoloid: I don't think this change is correct; as discussed in Belfast, we don't want to perform overload resolution when the *postfix-expression* is a function pointer or similar. (This could be observably different, for example if there were a case where an implicit conversion sequence couldn't be formed despite implicit conversion being possible.)

source/templates.tex
```diff
@@ -1204,8 +1204,7 @@
 of the type of the \grammarterm{template-parameter}.
 \begin{note}
 If the \grammarterm{template-argument}
-represents a set of overloaded functions
-(or a pointer or member pointer to such),
+represents an overload set (or a pointer or member pointer to such),
```
zygoloid: What is a pointer to an overload set? That doesn't seem to fit the new (or old) terminology very well.

source/basic.tex
```diff
@@ -1367,8 +1367,8 @@
 The declarations found by name lookup shall either all denote the same entity or
 shall all denote functions or function templates;
 in the latter case,
-the declarations are said to form a set of overloaded
-functions\iref{over.load}. Overload resolution\iref{over.match}
+the declarations are said to form an \defn{overload set}.
```
zygoloid: I think this definition is not completely clear: if name lookup finds exactly one function, is that an overload set or not? (Presumably these days the answer should be yes.) How about something more like:

> If the declarations found by name lookup all denote functions or function templates, the declarations are said to form an *overload set*. The declarations found by name lookup shall either all denote the same entity or form an overload set.

source/templates.tex
zygoloid: I'd be slightly more comfortable with "the address of an overload set", as it suggests use of the `&` operator rather than a specific pointer value. (And maybe "is" rather than "represents", since "represents the address of" is again a specifically defined property of real pointer values, rather than a syntactic property of an expression.)

source/overloading.tex
jensmaurer: Fixed: "names at least one function or function template"

source/templates.tex
jensmaurer: Fixed.

source/basic.tex
jensmaurer: Sounds good. Fixed.

PR #3594, opened 2019-12-30
[range.req.general, range.sized] Fix and improve description

johelegp:


source/ranges.tex
```diff
@@ -892,8 +892,8 @@
 
 \pnum
 The \libconcept{sized_range} concept specifies the requirements
-of a \libconcept{range} type that knows its size in constant time with the
-\tcode{size} function.
+of a \libconcept{range} type that knows its size in constant time with
+\tcode{ranges::size}.
```
jensmaurer: Can we maybe fix this a bit more? "knows its size" doesn't sound like a specification, and "with" is a bit non-specific. Maybe "of a range type whose size can be determined in constant time using ranges::size." or so.

johelegp: Sure.

source/ranges.tex
```diff
@@ -892,8 +892,8 @@
 
 \pnum
 The \libconcept{sized_range} concept specifies the requirements
-of a \libconcept{range} type that knows its size in constant time with the
-\tcode{size} function.
+of a \libconcept{range} type whose size can be determined in constant time
```
zygoloid: This sounds like it's talking about determining the size of the type (`sizeof(T)`). In [range.req.general]p2 we phrase this as:

> The `sized_range` concept refines range with the requirement that the number of elements in the
range can be determined in constant time using the `ranges::size` function.

... which seems like better phrasing. Do we really need to say this in two places?

johelegp: > Also, can we avoid saying this same thing in two places?

Any suggestions?

johelegp: How does it look like now?

johelegp:
I guess it should say _amortized_ now, right?

zygoloid:
I think this is definitely an improvement, but the addition of "amortized" seems like a normative change.

@jwakely What do you think? LWG issue? Or is this not a normative change in your view?

CaseyCarter:
> I think this is definitely an improvement, but the addition of "amortized" seems like a normative change.

It's reflecting the normative change we made to the specification of `sized_range` in [LWG-3363](https://cplusplus.github.io/LWG/issue3363). Maybe it would be better to avoid the duplication in the first place with a less precise but equally tutorial statement like `...refines \libconcept{range} with
the requirement that the number of elements in the range can be determined efficiently using \tcode{ranges::size}`?

zygoloid:
OK, I agree that both wording changes are are in not-really-normative introductory sentences, and the change reflects the actual normative wording. On that basis this seems fine. (Would be nice if at least the first changed paragraph here were converted to a note, though that's a separate issue.)

jwakely:
I like Casey's suggestion, but this seems fine either way.

PR #3595, opened 2019-12-30
[iterator.concept.sizedsentinel] Improve wording

johelegp:


source/iterators.tex
```diff
@@ -1543,7 +1543,7 @@
 \pnum
 The \libconcept{sized_sentinel_for} concept specifies
 requirements on an \libconcept{input_or_output_iterator} and
-a corresponding \libconcept{sentinel_for}
+a corresponding \libconcept{sentinel}
```
jensmaurer: This doesn't look right; there is no concept "sentinel", it seems.

johelegp: Ah, right.

source/iterators.tex
```diff
@@ -1543,7 +1543,7 @@
 \pnum
 The \libconcept{sized_sentinel_for} concept specifies
 requirements on an \libconcept{input_or_output_iterator} and
-a corresponding \libconcept{sentinel_for}
+a corresponding sentinel
```
jensmaurer: It would be good to keep the "is-a sentinel_for" here. Maybe use a partial-concept-id: "a corresponding `sentinel_for<I>`" after introducing "I" for the input_or_output_iterator.

johelegp: Sounds like the best solution, if I'm able to introduce `I` without too much noise.

source/iterators.tex
```diff
@@ -1542,8 +1542,8 @@
 
 \pnum
 The \libconcept{sized_sentinel_for} concept specifies
-requirements on an \libconcept{input_or_output_iterator} and
-a corresponding \libconcept{sentinel_for}
+requirements on an \libconcept{input_or_output_iterator} type \tcode{I} and
```
johelegp: Do you think I should've used math mode _I_ instead?

jensmaurer: Not right now in this isolated instance.

PR #3603, opened 2020-01-02
[iterator.requirements.general,range.counted] Rework notation for cou…

### LINKED ISSUE #2932: Counted ranges abuse mathematical notation
Counted ranges are presented as

  [`A`, `N`)

but don't have the normal mathematical meaning; instead, they mean [`A`, `A`+`N`) (where the "+" means "the value that would be obtained by `N` applications of ++ to `A`" rather than `operator+`).

We should not abuse established mathematical notation in this way, and should find a different and unambiguous way to present counted ranges.

jensmaurer:
…nted ranges.

Fixes #2932.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -620,15 +620,15 @@ and only if there is a finite sequence of applications of the expression
 \range{i}{s} denotes a valid range.
 
 \pnum
-A \defnadj{counted}{range} \range{i}{n} is empty if \tcode{n == 0};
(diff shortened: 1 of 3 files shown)


jensmaurer:
@CaseyCarter , agreed. Fixed.

CaseyCarter:
> @CaseyCarter , agreed. Fixed.

Thank you, that is *far* more readable in the source.

PR #3618, opened 2020-01-10
[expr.prim.lambda.closure]/3 Fix note: a generic lambda has a function call operator template, not a function call operator.

danra:


jensmaurer:
@zygoloid , I'd suggest to shorten the commit message a bit.

PR #3623, opened 2020-01-10
[ptr.launder] Fix note and example for std::launder.

### LINKED ISSUE #3613: [ptr.launder] Note and Example became incorrect after RU007/US042 NB comments resolution
http://eel.is/c++draft/ptr.launder#5 and http://eel.is/c++draft/ptr.launder#6 became incorrect after http://wg21.link/p1971r0#RU007 (https://github.com/cplusplus/draft/pull/3481/commits/fd8ff6441f93024bd0ee6e03a03c08be8e1b5ce0)

jensmaurer:
The applicable rules have changed in response to
NB RU 007, US 042 (C++20 CD).

Fixes #3613.

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -2877,7 +2877,7 @@ If a new object is created
 in storage occupied by an existing object of the same type,
 a pointer to the original object
 can be used to refer to the new object
-unless the type contains \tcode{const} or reference members;
+unless its complete object is a const object or it is a base class subobject;
 in the latter cases,
 this function can be used to obtain a usable pointer to the new object.
 See~\ref{basic.life}.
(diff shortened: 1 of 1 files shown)


source/support.tex
```diff
@@ -2877,19 +2877,19 @@
...
 \pnum
 \begin{example}
 \begin{codeblock}
-struct X { const int n; };
-X *p = new X{3};
+struct X { int n; };
+const X *p = new const X{3};
 const int a = p->n;
-new (p) X{5};                       // \tcode{p} does not point to new object\iref{basic.life} because \tcode{X::n} is \tcode{const}
+new (const_cast<X*>(p)) const X{5};  // \tcode{p} does not point to new object\iref{basic.life} because its type is \tcode{const}
```
```suggestion
new (const_cast<X*>(p)) const X{5};    // \tcode{p} does not point to new object\iref{basic.life} because its type is \tcode{const}
```

source/support.tex
```diff
@@ -2877,19 +2877,19 @@
 in storage occupied by an existing object of the same type,
 a pointer to the original object
 can be used to refer to the new object
-unless the type contains \tcode{const} or reference members;
-in the latter cases,
+unless its complete object is a const object;
```
zygoloid: Another remaining case that we might want to call out here is if the object is a base class subobject.

PR #3624, opened 2020-01-10
[dcl.fct,expr.ref] Fix description of class member access expressions

### LINKED ISSUE #3607: [expr.ref]/(6.3.2) function type spelling doesn't match [dcl.fct]/1
[[expr.ref]/(6.3.2)](http://eel.is/c++draft/expr.ref#6.3.2) says:
> Otherwise, if `E1.E2` refers to a non-static member function and the type of `E2` is “function of parameter-type-list _cv_ ref-qualifier<sub>opt</sub> returning `T`”… The type of `E1.E2` is “function of parameter-type-list _cv_ returning `T`”.

However, this doesn't match how function types are spelled in [[dcl.fct]/1](http://eel.is/c++draft/dcl.fct#1):
> … the type of the declarator-id in `D` is “derived-declarator-type-list noexcept<sub>opt</sub> function of (parameter-declaration-clause) cv-qualifier-seq<sub>opt</sub> ref-qualifier<sub>opt</sub> returning `T`” …

[expr.ref]/(6.3.2):
* uses _cv_ instead of _cv-qualifier-seq<sub>opt</sub>_
* uses "parameter-type-list" instead of "(_parameter-declaration-clause_)"
* misses `noexcept`_<sub>opt</sub>_

---

Also, why _ref-qualifier_<sub>opt</sub> of`E2`'s type is missing in the type of `E1.E2`?

jensmaurer:
involving non-static member functions.

Fixes #3607.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -3276,30 +3276,23 @@ is ``\cvqual{vq12} \tcode{T}''. If \tcode{E2} is not declared to be a
 \tcode{mutable} member, then the type of \tcode{E1.E2} is
 ``\cvqual{cq12} \cvqual{vq12} \tcode{T}''.
 
-\item If \tcode{E2} is a (possibly overloaded) member function, function
-overload resolution\iref{over.match} is used to determine whether
-\tcode{E1.E2} refers to a static or a non-static member function.
+\item If \tcode{E2} is a (possibly overloaded) member function,
+function overload resolution\iref{over.match}
+is used to select the function to which \tcode{E2} refers.
+The type of \tcode{E1.E2} is the type of \tcode{E2}
+and \tcode{E1.E2} refers to the function referred to by \tcode{E2}.
 
 \begin{itemize}
-\item If it refers to a static member function and the type of
-\tcode{E2} is ``function of parameter-type-list returning \tcode{T}'',
-then \tcode{E1.E2} is an lvalue; the expression designates the static
-member function. The type of \tcode{E1.E2} is the same type as that of
-\tcode{E2}, namely ``function of parameter-type-list returning
-\tcode{T}''.
+\item If \tcode{E2} refers to a static member function,
+\tcode{E1.E2} is an lvalue.
 
-\item Otherwise, if \tcode{E1.E2} refers to a non-static member
-function and the type of \tcode{E2} is ``function of
-parameter-type-list \cv{} \opt{\grammarterm{ref-qualifier}} returning \tcode{T}'', then
-\tcode{E1.E2} is a prvalue. The expression designates a
-non-static member function. The expression can be used only as the
+\item Otherwise (when \tcode{E2} refers to a non-static member function),
+\tcode{E1.E2} is a prvalue. The expression can be used only as the
 left-hand operand of a member function call\iref{class.mfct}.
 \begin{note}
 Any redundant set of parentheses surrounding the expression
 is ignored\iref{expr.prim.paren}.
 \end{note}
-The type of \tcode{E1.E2} is
-``function of parameter-type-list \cv{} returning \tcode{T}''.
 \end{itemize}
 
 \item If \tcode{E2} is a nested type, the expression \tcode{E1.E2} is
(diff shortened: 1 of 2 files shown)


source/expressions.tex
```diff
@@ -3289,7 +3289,7 @@
 \tcode{T}''.
 
 \item Otherwise, if \tcode{E1.E2} refers to a non-static member
```
zygoloid: How about:

> Otherwise, if `E1.E2` refers to a non-static member function of type `T`, then `E1.E2` is a prvalue of type `T` designating that non-static member function. The expression can be used only as the left-hand operand of a member function call ([class.mfct]). [Note: ... ]

(And similarly in the static member function case.)

Do we really need to say "if `E1.E2` refers to a non-static member function then `E1.E2` designates that non-static member function"? What distinction between "refers to" and "designates" are we trying to create here?

After simplifying this wording, the bullets for static and non-static member functions are the same other than the "can only be used" sentence and its note; perhaps the common wording could be extracted?

jensmaurer: Maybe what we want to say is that E2 names a non-static member function of type T, in which case E1.E2 is a prvalue of type T designating that non-static member function. We make a connection from the individual parts to the compound "E1.E2" expression.

jensmaurer: static member functions are lvalues, non-static member functions are prvalues, it seems

jensmaurer: Updated; should be better now.

source/expressions.tex
```diff
@@ -3277,29 +3277,22 @@
 ``\cvqual{cq12} \cvqual{vq12} \tcode{T}''.
 
 \item If \tcode{E2} is a (possibly overloaded) member function, function
-overload resolution\iref{over.match} is used to determine whether
-\tcode{E1.E2} refers to a static or a non-static member function.
+overload resolution\iref{over.match} is used to determine to which
+function \tcode{E1.E2} refers.
+If \tcode{E2} refers to a member function of type \tcode{T},
+\tcode{E1.E2} has type \tcode{T} and refers to that member function.
```
zygoloid: The "If" here sounds like it's introducing a condition, but it's not. Perhaps:

> If `E2` is a (possibly overloaded) member function, function overload resolution ([over.match]) is used to determine to which function `E2` refers. The type of `E1.E2` is the type of `E2` and `E1.E2` refers to the function referred to by `E2`."

source/expressions.tex
```diff
@@ -3277,29 +3277,22 @@
...
 
 \begin{itemize}
-\item If it refers to a static member function and the type of
-\tcode{E2} is ``function of parameter-type-list returning \tcode{T}'',
-then \tcode{E1.E2} is an lvalue; the expression designates the static
-member function. The type of \tcode{E1.E2} is the same type as that of
-\tcode{E2}, namely ``function of parameter-type-list returning
-\tcode{T}''.
+\item If \tcode{E1.E2} refers to a static member function,
```
zygoloid: I think "If `E2` refers to a static member function" would be a little simpler.

source/expressions.tex
```diff
@@ -3277,29 +3277,22 @@
...
+then \tcode{E1.E2} is an lvalue.
 
-\item Otherwise, if \tcode{E1.E2} refers to a non-static member
-function and the type of \tcode{E2} is ``function of
-parameter-type-list \cv{} \opt{\grammarterm{ref-qualifier}} returning \tcode{T}'', then
-\tcode{E1.E2} is a prvalue. The expression designates a
-non-static member function. The expression can be used only as the
+\item Otherwise, if \tcode{E1.E2} refers to a non-static member function,
+then \tcode{E1.E2} is a prvalue. The expression can be used only as the
```
zygoloid: Maybe:

> Otherwise (when `E2` refers to a non-static member function), `E1.E2` is a prvalue.

source/expressions.tex
jensmaurer: Fixed with a slightly bigger hammer.

source/expressions.tex

PR #3625, opened 2020-01-11
[over.oper] Clarify handling of individual operators.

### LINKED ISSUE #1642: [over.call] wording makes little sense
This is a purely editorial issue. The wording currently reads:

> Thus, a call `x(arg1,...)` is interpreted as `x.operator()(arg1, ...)` for a class object `x` of type `T` if `T​::​operator()(T1, T2, T3)` exists and if the operator is selected as the best match function by the overload resolution mechanism ([over.match.best]).

What does it mean for `T::operator()(T1, T2, T3)` to exist? What are `T1`, `T2`, and `T3`? The idea here, I think, is to clarify that `operator()` is called if it exists, as opposed to some function pointer conversion that could happen and be a better match... I can't propose a better wording since I'm not entirely sure what the wording is currently intended to convey.

jensmaurer:
Fixes #1642.

### DIFF
--- a/source/overloading.tex
+++ b/source/overloading.tex
@@ -851,22 +851,14 @@ the call is ill-formed.
 \rSec4[over.call.object]{Call to object of class type}
 
 \pnum
-If the
-\grammarterm{postfix-expression}
-\tcode{E}
+If the \grammarterm{postfix-expression} \tcode{E}
 in the function call syntax evaluates
 to a class object of type ``\cv{}~\tcode{T}'',
-then the set of candidate
-functions includes at least the function call operators of
-\tcode{T}.
-The
-function call operators of
-\tcode{T}
-are obtained by ordinary lookup of
-the name
-\tcode{operator()}
-in the context of
-\tcode{(E).operator()}.
+then the set of candidate functions
+includes at least the function call operators of \tcode{T}.
+The function call operators of \tcode{T}
+are obtained by ordinary lookup of the name \tcode{\keyword{operator}()}
+in the context of \tcode{(E).\keyword{operator}()}.
 
 \pnum
 In addition, for each non-explicit conversion function declared in \tcode{T} of the
@@ -1005,25 +997,20 @@ for the built-in operator\iref{expr.compound}.
 {l|m|m|m}
 \topline
 \hdstyle{Subclause} &   \hdstyle{Expression} &   \hdstyle{As member function} &   \hdstyle{As non-member function} \\ \capsep
-\ref{over.unary}    &   \tcode{@a}   &   \tcode{(a).operator@ (\,)}  &   \tcode{operator@(a)}    \\
-\ref{over.binary}   &   \tcode{a@b}  &   \tcode{(a).operator@ (b)}   &   \tcode{operator@(a, b)} \\
-\ref{over.ass}      &   \tcode{a=b}  &   \tcode{(a).operator= (b)}   &                           \\
-\ref{over.sub}      &   \tcode{a[b]} &   \tcode{(a).operator[](b)}   &                           \\
-\ref{over.ref}      &   \tcode{a->}  &   \tcode{(a).operator->(\,)}  &                           \\
-\ref{over.inc}      &   \tcode{a@}   &   \tcode{(a).operator@ (0)}   &   \tcode{operator@(a, 0)} \\
+\ref{over.unary}    &   \tcode{@a}   &   \tcode{(a).\keyword{operator}@ (\,)}  &   \tcode{\keyword{operator}@(a)}    \\
+\ref{over.binary}   &   \tcode{a@b}  &   \tcode{(a).\keyword{operator}@ (b)}   &   \tcode{\keyword{operator}@(a, b)} \\
+\ref{over.ass}      &   \tcode{a=b}  &   \tcode{(a).\keyword{operator}= (b)}   &                           \\
+\ref{over.sub}      &   \tcode{a[b]} &   \tcode{(a).\keyword{operator}[](b)}   &                           \\
+\ref{over.ref}      &   \tcode{a->}  &   \tcode{(a).\keyword{operator}->(\,)}  &                           \\
+\ref{over.inc}      &   \tcode{a@}   &   \tcode{(a).\keyword{operator}@ (0)}   &   \tcode{\keyword{operator}@(a, 0)} \\
 \end{floattable}
 
 \pnum
-For a unary operator
-\tcode{@}
-with an operand of a type whose cv-unqualified version is
-\tcode{T1},
-and for a binary operator
-\tcode{@}
-with a left operand of a type whose cv-unqualified version is
-\tcode{T1}
-and a right operand of a type whose cv-unqualified version is
-\tcode{T2},
+For a unary operator \tcode{@}
+with an operand of type \cvqual{cv1} \tcode{T1},
+and for a binary operator \tcode{@}
+with a left operand of type \cvqual{cv1} \tcode{T1}
+and a right operand of type \cvqual{cv2} \tcode{T2},
 four sets of candidate functions, designated
 \defnx{member candidates}{member candidate},
 \defnx{non-member candidates}{non-member candidate},
@@ -1033,15 +1020,14 @@ and
 are constructed as follows:
 \begin{itemize}
 \item
-If
-\tcode{T1}
+If \tcode{T1}
 is a complete class type or a class currently being defined, the set of member candidates is the
 result of the qualified lookup of
-\tcode{T1::operator@}\iref{over.call.func}; otherwise, the set of member
+\tcode{T1::\keyword{operator}@}\iref{over.call.func}; otherwise, the set of member
 candidates is empty.
 \item
 The set of non-member candidates is the result of the unqualified lookup of
-\tcode{operator@}
+\tcode{\keyword{operator}@}
 in the context of
 the expression according to the usual rules for name
 lookup in unqualified function calls\iref{basic.lookup.argdep} except
@@ -1164,13 +1150,13 @@ struct A {
 A operator+(const A&, const A&);
 void m() {
   A a, b;
-  a + b;                        // \tcode{operator+(a, b)} chosen over \tcode{int(a) + int(b)}
+  a + b;                        // \tcode{\keyword{operator}+(a, b)} chosen over \tcode{int(a) + int(b)}
 }
 \end{codeblock}
 \end{example}
 
 \pnum
-If a rewritten \tcode{operator<=>} candidate
+If a rewritten \tcode{\keyword{operator}<=>} candidate
 is selected by overload resolution
 for an operator \tcode{@},
 \tcode{x @ y}
@@ -1179,12 +1165,12 @@ is interpreted as
 if the selected candidate is a synthesized candidate
 with reversed order of parameters,
 or \tcode{(x <=> y) @ 0} otherwise,
-using the selected rewritten \tcode{operator<=>} candidate.
+using the selected rewritten \tcode{\keyword{operator}<=>} candidate.
 Rewritten candidates for the operator \tcode{@}
 are not considered in the context of the resulting expression.
 
 \pnum
-If a rewritten \tcode{operator==} candidate
+If a rewritten \tcode{\keyword{operator}==} candidate
 is selected by overload resolution
 for an operator \tcode{@},
 its return type shall be \cv{} \tcode{bool}, and
@@ -1202,7 +1188,7 @@ otherwise, if \tcode{@} is \tcode{!=},
 otherwise (when \tcode{@} is \tcode{==}),
 \tcode{y == x},
 \end{itemize}
-in each case using the selected rewritten \tcode{operator==} candidate.
+in each case using the selected rewritten \tcode{\keyword{operator}==} candidate.
 
 \pnum
 If a built-in candidate is selected by overload resolution, the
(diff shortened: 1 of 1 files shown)


source/overloading.tex
```diff
@@ -3415,15 +3401,10 @@
...
-\tcode{new},
-\tcode{operator}
-\tcode{new[]},
-\tcode{operator}
-\tcode{delete}
-and
-\tcode{operator}
-\tcode{de\-lete\brk[]},
+\keyword{operator} \keyword{new},
+\keyword{operator} \tcode{keyword{new}[]},
```
zygoloid: Missing a `\` here.

source/overloading.tex
```diff
@@ -3415,15 +3401,10 @@
...
+\keyword{operator} \keyword{delete}, and
+\keyword{operator} \tcode{de\-lete\brk[]},
```
zygoloid: Is the lack of `\keyword` markup for this intentional? (There should be some way to combine line breaking hints with `\keyword`, but I'm not sure what.)

source/overloading.tex
```diff
@@ -1010,25 +1002,20 @@
...
-\tcode{@}
-with a left operand of a type whose cv-unqualified version is
-\tcode{T1}
-and a right operand of a type whose cv-unqualified version is
-\tcode{T2},
+For a unary operator \tcode{@}
+with an operand of a type whose cv-unqualified version is \tcode{T1},
+and for a binary operator \tcode{@}
+with a left operand of a type whose cv-unqualified version is \tcode{T1}
+and a right operand of a type whose cv-unqualified version is \tcode{T2},
```
zygoloid: While we're here: "with an operand of type cv1 T1"?

source/overloading.tex
```diff
@@ -3489,24 +3470,23 @@
 \indextext{overloading!unary operator}
 
 \pnum
-A prefix unary operator shall be implemented by a
-non-static member function\iref{class.mfct} with no parameters or a
-non-member function with one parameter.
+For a \grammarterm{unary-operator} \tcode{@},
```
zygoloid: Should we use `\placeholder` for these `@`s?

source/overloading.tex
```diff
@@ -3489,24 +3470,23 @@
...
+an operator function \tcode{\keyword{operator}@} shall be
+a non-static member function\iref{class.mfct} with no parameters or
+a non-member function with one parameter.
```
zygoloid: I think this rephrasing in terms of *unary-operator* is problematic: it suggests that (eg) `operator+` shall not have more than one parameter (including the implicit object parameter), which is incorrect.

The prior phrasing ("a prefix unary operator" here, "a binary operator" below) seems to avoid this problem if we add another rule that "an operator function shall declare a prefix unary operator, a binary operator, ...[other cases]...").

source/overloading.tex

source/overloading.tex
jensmaurer: \keyword puts the thing into an index. Line-breaking hints are a good way to screw up the index.

source/overloading.tex
jensmaurer: I think we have more @s for the three-way comparisons; if we do something in this area, we should do it in a separate effort.

source/overloading.tex
jensmaurer: Fixed. There isn't a linebreak where the hints are, to start with.

source/overloading.tex

source/overloading.tex

I just noticed that [over.ass] syntactically appears to talk about copy-assignment only. It should probably talk about @= instead.

source/overloading.tex
```diff
@@ -3486,24 +3480,23 @@
 \indextext{overloading!unary operator}
 
 \pnum
-A prefix unary operator shall be implemented by a
-non-static member function\iref{class.mfct} with no parameters or a
-non-member function with one parameter.
+For a prefix unary operator \tcode{@},
+an operator function \tcode{\keyword{operator}@} shall be
+a non-static member function\iref{class.mfct} with no parameters or
+a non-member function with one parameter.
```
zygoloid: I wonder if phrasing these as "is" rather than "shall" would work better:

> A prefix unary operator function for a *unary-operator* `@` is an operator function `operator@` that is either a non-static member function with no parameters or a non-member function with one parameter.

(Then the new "shall" requirement gives us the requirement that `operator+` effectively has either one or two parameters, and so on.)

source/overloading.tex
zygoloid: I think it's fine and probably appropriate that we treat `@=` as binary operators and `=` as its own thing, because `operator@=` can be a member or non-member whereas `operator=` can only be a member. Though maybe "Simple assignment" would be a better subclause title than "Assignment"?

source/overloading.tex
jensmaurer: Good idea.

source/overloading.tex
jensmaurer: Yes, definitely. Fixed.

source/overloading.tex
```diff
@@ -3453,8 +3434,20 @@
...
+\pnum
+An operator function shall declare
+a prefix unary operator,
+a binary operator,
+a simple assignment,
+a function call,
+a subscripting,
+a class member access using \tcode{->},
+an increment or decrement,
+or a literal operator.
```
zygoloid: Looks like, per [over.oper]/1, literal operators are not operator functions. Maybe we should move the description of them out of [over.oper] too?

source/overloading.tex
```diff
@@ -3486,24 +3480,23 @@
 \indextext{overloading!unary operator}
 
 \pnum
-A prefix unary operator shall be implemented by a
-non-static member function\iref{class.mfct} with no parameters or a
-non-member function with one parameter.
+A \defnadj{prefix unary}{operator function} \tcode{\keyword{operator}@}
+for a prefix unary operator \tcode{@}\iref{expr.unary.op} is
+a non-static member function\iref{class.mfct} with no parameters or
+a non-member function with one parameter.
```
zygoloid: I find this definition a bit confusing. I want to look at this and determine whether a particular function is a prefix unary operator function, and it doesn't seem to quite get us there (it reads like the choice of `@` is part of the input, not part of the definition). Maybe reorder a little:

> A *prefix unary operator function* is a function named `operator@` for a prefix unary operator `@` that is either a non-static member function with no parameters or a non-member function with one parameter.

or

> An operator function `operator@` is a *prefix unary operator function* if `@` is a prefix unary operator and the function is either a non-static member function with no parameters or a non-member function with one parameter.

or something.

source/overloading.tex
jensmaurer: That sounds like a good idea. In fact, [over.literal] has nothing to do with [over.oper].

source/overloading.tex
jensmaurer: The first alternative sounds promising.

source/overloading.tex

source/overloading.tex
```diff
@@ -3487,24 +3480,24 @@
...
-\tcode{++}
-and
-\tcode{\dcr}.
+For a \grammarterm{unary-expression}
+of the form \tcode{@ \grammarterm{cast-expression}},
+the operator function is selected by overload resolution\iref{over.match.oper}.
+If a member function is selected,
+the expression is interpreted as
+\tcode{\grammarterm{cast-expression} . \keyword{operator}@()};
+otherwise, the expression is interpreted as
```
zygoloid: We also need to handle the case where a built-in candidate is selected.

"otherwise, if a non-member function is selected, the expression is interpreted as"

maybe?

jensmaurer: Sounds good.

I've also fixed the similar occurence for binary operators a few lines down.

source/overloading.tex
```diff
@@ -3540,24 +3528,25 @@
...
+If a member function is selected,
+the expression is interpreted as
+\tcode{\grammarterm{cast-expression} . \keyword{operator}@()};
+otherwise, if a non-member function is selected,
+the expression is interpreted as
+\tcode{\keyword{operator}@( \grammarterm{cast-expression} )}.
+\begin{note}
+The operators \tcode{++} and \tcode{\dcr}\iref{expr.pre.incr}
+are described in~\ref{over.inc}.
+\end{note}
```
zygoloid: Hmm. They're actually described in this section too, and the rules are (fortunately) identical.
We should make these orthogonal: either `operator--` and `operator++` are not prefix unary operator functions, or the section on increment / decrement should instead describe postfix unary operator functions and not repeat the description of prefix `--` and `++`. What do you think?

source/overloading.tex
```diff
@@ -3599,34 +3588,37 @@
 a relational operator function, or
 a three-way comparison operator function.
 
-\rSec2[over.ass]{Assignment}
+\rSec2[over.ass]{Simple assignment}
```
zygoloid: I think we should indent this to `\rSec3` since this is a kind of binary operator function.

source/overloading.tex
```diff
@@ -3657,60 +3649,42 @@
...
-\tcode{T3)}
-exists and if the operator is selected as the best match function by
-the overload resolution mechanism\iref{over.match.best}.
+where the \grammarterm{postfix-expression} evaluates to a class object,
+the operator function
+is selected by overload resolution\iref{over.match.oper},
+and the expression is interpreted as
+\begin{ncsimplebnf}
+postfix-expression \terminal{.} \keyword{operator} \terminal{(}\terminal{)} \terminal{(} \opt{expression-list} \terminal{)}
+\end{ncsimplebnf}
```
zygoloid: This is wrong for the case where we select a surrogate candidate. The old wording was correct but incomplete.

source/overloading.tex
```diff
@@ -3657,60 +3649,42 @@
...
+For an expression of the form
 \begin{ncsimplebnf}
 postfix-expression \terminal{[} expr-or-braced-init-list \terminal{]}
 \end{ncsimplebnf}
+the operator function
+is selected by overload resolution\iref{over.match.oper},
+and the expression is interpreted as
+\begin{ncsimplebnf}
+postfix-expression . \keyword{operator} \terminal{[}\terminal{]} \terminal{(} expr-or-braced-init-list \terminal{)}
+\end{ncsimplebnf}
```
zygoloid: This is wrong if we select a built-in candidate (the old wording was correct).

source/overloading.tex
jensmaurer: A hanging paragraph! Yay!

source/overloading.tex
jensmaurer: I think we should only describe postfix ++ and -- as special. I should also apply the grammar rewriting to that section so that we can show an actual call to `operator++(0)`.

source/overloading.tex
jensmaurer: Added "If a member function is selected,", in line with the rest.

source/overloading.tex
jensmaurer: We use over.call.object, not over.match.oper for the function call operator. Fixed xref.

jensmaurer: Fixed, also removing the bad-placed "how do we call this in the end" paragraph from over.call.object.

source/overloading.tex
jensmaurer: Ah, no, that's wrong. Increment/decrement is not a _unary-operator_ per the grammar; it's something special. I am now highlighting that by using the \grammarterm.

Pushing for further review.

zygoloid: This approach seems fine to me.

source/overloading.tex
```diff
@@ -3657,60 +3641,49 @@
...
-\tcode{x}
-of type
-\tcode{T}
-if
-\tcode{T::operator()(T1,}
-\tcode{T2,}
-\tcode{T3)}
-exists and if the operator is selected as the best match function by
-the overload resolution mechanism\iref{over.match.best}.
+where the \grammarterm{postfix-expression} evaluates to a class object,
```
zygoloid: "evaluates to a class object" is a runtime property. I think we want "is of class type", right?

source/overloading.tex
```diff
@@ -3757,11 +3722,9 @@
...
 
 \pnum
-The user-defined function called
-\tcode{operator++}
-implements the prefix and postfix
-\tcode{++}
-operator.
+An \defnadj{increment}{operator function}
+is a function named \tcode{\keyword{operator}++}
+that implements the prefix and postfix \tcode{++} operator.
```
zygoloid: This last line is not part of the definition of "increment operator function". As written, this would suggest that an `operator++` that (somehow) doesn't implement a prefix and [or?] postfix `++` operator is not an "increment operator function".

source/overloading.tex
```diff
@@ -3816,114 +3779,6 @@
 The prefix and postfix decrement operators
 \tcode{-{-}}
 are handled analogously.
```
zygoloid: We never define what a "decrement operator function" is. Perhaps:

> A *decrement operator function* is handled analogously to an increment operator function, except for the operator `--` instead of `++`.

or something like that.

source/overloading.tex
```diff
@@ -3801,8 +3764,8 @@
 void f(X a, Y b) {
   ++a;                          // \tcode{a.operator++();}
   a++;                          // \tcode{a.operator++(0);}
```
zygoloid: `\keyword` here too?

source/overloading.tex
```diff
@@ -2954,7 +2931,7 @@
 };
 A& operator<<(A&&, char);
 A() << 1;                       // calls \tcode{A::operator<<(int)}
-A() << 'c';                     // calls \tcode{operator<<(A\&\&, char)}
+A() << 'c';                     // calls \tcode{\keyword{operator}<<(A\&\&, char)}
 A a;
 a << 1;                         // calls \tcode{A::operator<<(int)}
 a << 'c';                       // calls \tcode{A::operator<<(int)}
```
zygoloid: `\keyword` for all of these comments. (I guess your search missed `operator` after `::`?)

source/overloading.tex
zygoloid: Maybe convert "that implements the prefix and postfix `++` operator" to "[Note: An increment operator function implements either the prefix or postfix `++` operator.]", or just remove it -- we say the same thing but better in the next couple of sentences.

source/overloading.tex
jensmaurer: Copied from over.call.object. Fixed.

source/overloading.tex
jensmaurer: Removed.

source/overloading.tex
jensmaurer: I did:

A \defnadj{decrement}{operator function}
is a function named \tcode{\keyword{operator}\dcr}
and is handled analogously to an increment operator function.

source/overloading.tex
jensmaurer: There are still a lot of \keywords missing.

jensmaurer: Fixed the obvious ones.

PR #3628, opened 2020-01-13
[dcl.attr] Introduce 'Recommended practice' paragraphs.

### LINKED ISSUE #2978: [dcl.attr.*] and Recommended Practice
One of the tools the C standard uses for its attribute wording is a Recommended Practice section (http://www.open-std.org/jtc1/sc22/wg14/www/docs/n2385.pdf 6.7.11). These sections correspond to the same wording we put in informative notes currently.

From what I understand, a Recommended Practice is a special kind of section (like Constraints or Semantics) that places normative encouragement on implementations without being a normative requirement. Would it make sense to follow suit in C++? I think it may allow us to speak more authoritatively than a note while still serving the purpose of not being a normative requirement.

jensmaurer:
Fixes #2978.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -8766,13 +8766,12 @@ Redeclarations using different forms of the attribute (with or without the
 \grammarterm{attribute-argument-clause}{s}) are allowed.
 
 \pnum
-\begin{note}
-Implementations may use the \tcode{deprecated} attribute to produce a diagnostic
+\recommended
+Implementations should use the \tcode{deprecated} attribute to produce a diagnostic
 message in case the program refers to a name or entity other than to declare it, after a
-declaration that specifies the attribute. The diagnostic message may include the text provided
+declaration that specifies the attribute. The diagnostic message should include the text provided
 within the \grammarterm{attribute-argument-clause} of any \tcode{deprecated} attribute applied
 to the name or entity.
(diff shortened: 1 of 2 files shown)


zygoloid:
@jensmaurer Please can you mail the CWG reflector just to draw people's attention to this change so that this style can be used going forward?

zygoloid:
Actually... let's get CWG to look over this before applying it. Converting "may" to "should", while not introducing a requirement, does introduce a recommendation, and that seems like something for which we should ask for oversight.

jensmaurer:
Reference to CWG e-mail: http://lists.isocpp.org/core/2020/01/8153.php

source/intro.tex
```diff
@@ -600,6 +600,10 @@
 \Cpp{} and C\@. Certain features of \Cpp{} exist solely for compatibility
 purposes; \ref{depr} describes those features.
 
+\pnum
+Throughout this document,
+non-normative encouragement for implementations is introduced by \recommended.
```
tvaneerd: That sounds pretty definitive.  Are you implying *all* instances of non-normative encouragement? What about all the places we missed, and are still in notes or wherever?

zygoloid:
Approved by CWG 2020-02-14.

PR #3629, opened 2020-01-14
[func.wrap.badcall] Add missing itemdecl for what()

timsong-cpp:


jensmaurer:
The additional constraint here seems to be that it's not an NTMBS. My guess is that's not intentional, but ... oh well.

PR #3630, opened 2020-01-14
[basic.lookup] Grammar

languagelawyer:
Now this doesn't look like a question


jensmaurer:
I think the existing text might be a bit poetic, but it's still good English. NAD in my view, but I'll defer to a native speaker.

zygoloid:
As Jens notes, "Only after <X happens> <does Y happen>." is correct English, but if it's less accessible than a different phrasing that's equally good English, I think we should be open to changing it. However, the suggested alternative sounds incorrect to me. (So: not exactly this change, but some change here may be justified.)

languagelawyer:
If this is emphasis, then ok.

PR #3643, opened 2020-01-20
Fix example in [temp.constr.op]/p4

akrzemi1:


jensmaurer:
As far as I can see, the parentheses have no semantic effect in this case.
This is in a core language section, where we usually claim that coding-style harmonization is usually not applied "to show the diversity of the core language" or something to that effect.

CaseyCarter:
> As far as I can see, the parentheses have no semantic effect in this case.

The example is ill-formed: `(sizeof(T) > 1) && get_value<T>()` is not a valid *constraint-logical-or-expression* since `get_value<T>()` is not a *primary-expression*. `(get_value<T>())` is a primary expression, so the example is well-formed after the change.

jensmaurer:
Thanks, @CaseyCarter, for explaining that. You're right, it seems.

PR #3645, opened 2020-01-22
[lex.operators] Add `co_await` to the list

wreien:
`co_await` is missing from the list of _preprocessing-op-or-punc_. Since it is an operator like `new` and `delete` (see [over.oper]), it should be here, otherwise it would be an identifier in phases 3 through 6. This is a clear inconsistency.

(I'm not certain about how to layout the table, as `co_await` is almost too big to fit, but the existing patterns should be maintained.)

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -873,9 +873,9 @@ are converted into tokens for operators and punctuators:
 %% Ed. note: character protrusion would misalign various operators.
 \microtypesetup{protrusion=false}\obeyspaces
 \nontermdef{preprocessing-op-or-punc} \textnormal{one of}\br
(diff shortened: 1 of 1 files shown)


jensmaurer:
I note that the table also does not contain "throw", which appears in a _throw-expression_.

What exactly breaks if co_await and throw are not present in _preprocessing-op-or-punc_?
Just some #if conditionals? If so, it seems perfectly reasonable to omit co_await and throw from the list, because those have no reason to appear in an #if.

jensmaurer:
Editorial meeting: Not editorial. Covered by CWG 189 and CWG 1901.

PR #3648, opened 2020-01-23
Use "requirement" in grammar font in expr.prim.req

akrzemi1:
[expr.prim.req] uses grammarterm <em><code>requirement</code></em> in paragraph 5, but normal font in paragraph 6. Term "requirement" is never defined in this clause, but it has a different meaning in the library.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -2512,7 +2512,7 @@ are unevaluated operands\iref{expr.prop}.
 
 \pnum
 \begin{example}
-A common use of \grammarterm{requires-expression}{s} is to define
+A common use of \grammarterm{requires-expression}s is to define
 requirements in concepts such as the one below:
 \begin{codeblock}
 template<typename T>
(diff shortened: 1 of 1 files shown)


source/expressions.tex
```diff
@@ -2567,7 +2567,7 @@
 \pnum
 The substitution of template arguments into a \grammarterm{requires-expression}
 may result in the formation of invalid types or expressions in its
-requirements or the violation of the semantic constraints of those requirements.
+\grammarterm{requirement}{}s or the violation of the semantic constraints of those \grammarterm{requirement}{}s.
```
jensmaurer: Please move the "s" inside the empty brace immediately preceding it.

source/expressions.tex
```diff
@@ -2577,7 +2577,7 @@
 the \grammarterm{requires-expression} evaluates to \tcode{true}.
 \begin{note}
 If a \grammarterm{requires-expression} contains invalid types or expressions in
-its requirements, and it does not appear within the declaration of a templated
+its \grammarterm{requirement}{}s, and it does not appear within the declaration of a templated
```
jensmaurer: Also here.

jensmaurer:
You change some "s" in places that weren't previously in scope for this pull request (ok), but you didn't change the "s" things for the stuff that you originally changed here. That's not really progress.

akrzemi1:
Sorry. Learning git pull requests and tex at the same time...

zygoloid:
There's no need for braces around or near the `s`s at all. Will remove those now...

jensmaurer:
(Braces were required in former times due to LaTeX-isms.)

Issue #3636, opened 2020-01-16
[lex] Replace "character literal" with references to grammar term character-literal


zygoloid:
From CWG discussion of P2029R0.

Remove [lex.ccon] paragraph 1, which defines "character literal" as exactly the same thing as a (grammar term) *character-literal*, and replace all references to "character literal" with references to the grammar term *character-literal*.

Also remove [lex.string] paragraph 1, which defines in plain English the grammar term *string-literal* to be the same thing it already is in the grammar.

jensmaurer:
[lex.icon] and [lex.fcon] have the same problem, fixed by #3650 

Since [lex.string] and [lex.ccon] is being worked on by P2029R0, I'll leave those alone for the time being.

PR #3650, opened 2020-01-25
[lex.icon,lex.fcon] Rework description to avoid redundancies.

jensmaurer:
Partially addresses #3636.

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -1598,7 +1598,7 @@ the basic source character set.
 \grammarterm{pp-number} can contain \tcode{p} \grammarterm{sign} and
 \tcode{P} \grammarterm{sign}.
 \rationale
-Necessary to enable hexadecimal floating-point literals.
+Necessary to enable \grammarterm{hexadecimal-floating-point-literal}s.
 \effect
 Valid \CppXIV{} code may fail to compile or produce different results in
 this International Standard. Specifically, character sequences like \tcode{0p+0}
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -397,13 +397,14 @@ None have any observable spelling.
 \pnum
 \begin{example}
 The program fragment \tcode{0xe+foo} is parsed as a
-preprocessing number token (one that is not a valid integer or floating-point
-literal token), even though a parse as three preprocessing tokens
+preprocessing number token (one that is not a valid
+\grammarterm{integer-literal} or \grammarterm{floating-point-literal} token),
+even though a parse as three preprocessing tokens
 \tcode{0xe}, \tcode{+}, and \tcode{foo} might produce a valid expression (for example,
 if \tcode{foo} were a macro defined as \tcode{1}). Similarly, the
 program fragment \tcode{1E1} is parsed as a preprocessing number (one
-that is a valid floating-point literal token), whether or not \tcode{E} is a
-macro name.
+that is a valid \grammarterm{floating-point-literal} token),
+whether or not \tcode{E} is a macro name.
 \end{example}
 
 \pnum
@@ -583,14 +584,15 @@ depending on the implementation.}%
 \end{bnf}
 
 \pnum
-Preprocessing number tokens lexically include all integer literal
-tokens\iref{lex.icon} and all floating-point literal
-tokens\iref{lex.fcon}.
+Preprocessing number tokens lexically include
+all \grammarterm{integer-literal} tokens\iref{lex.icon} and
+all \grammarterm{floating-point-literal} tokens\iref{lex.fcon}.
 
 \pnum
 A preprocessing number does not have a type or a value; it acquires both
-after a successful conversion to an integer literal token or a floating-point literal
-token.%
+after a successful conversion to
+an \grammarterm{integer-literal} token or
+a \grammarterm{floating-point-literal} token.%
 \indextext{number!preprocessing|)}
 
 \rSec1[lex.name]{Identifiers}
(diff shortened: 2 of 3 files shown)


source/compatibility.tex
```diff
@@ -1597,7 +1597,7 @@
 \grammarterm{pp-number} can contain \tcode{p} \grammarterm{sign} and
 \tcode{P} \grammarterm{sign}.
 \rationale
-Necessary to enable hexadecimal floating-point literals.
+Necessary to enable \grammarterm{hexadecimal-floating-point-literal}{s}.
```
zygoloid: No braces around `s` please.

source/lex.tex
```diff
@@ -1007,28 +1009,29 @@
...
-(base ten) begins with a digit other than \tcode{0} and
-consists of a sequence of decimal digits.
-A \defnx{hexadecimal integer literal}{literal!hexadecimal}
-(base sixteen) begins with
-\tcode{0x} or \tcode{0X} and consists of a sequence of hexadecimal
-digits, which include the decimal digits and the letters \tcode{a}
-through \tcode{f} and \tcode{A} through \tcode{F} with decimal values
-ten through fifteen.
+In an \grammarterm{integer-literal},
+the sequence of digits is interpreted as a base-$N$ number
```
zygoloid: We've lost the connection between the word "digits" and the grammar. Perhaps

> the sequence of *binary-digit*s, *octal-digit*s, *digit*s, or *hexadecimal-digit*s is interpreted as a base-$N$ number
> as shown in table [tab:lex.icon.base];
> the lexically first digit of the sequence of digits is the most significant.
> [Note: The prefix and any optional separating single quotes are ignored when determining the value.]

source/lex.tex
```diff
@@ -1007,28 +1009,29 @@
...
+{lex.icon.base}
+{lr}
+\topline
+\lhdr{Kind of \grammarterm{integer-literal}} & \rhdr{base $N$} \\ \capsep
+\grammarterm{binary-literal} & 2 \\
+\grammarterm{octal-literal} & 8 \\
+\grammarterm{decimal-literal} & 10 \\
+\grammarterm{hexadecimal-literal} & 16 \\
+\end{simpletypetable}
+
```
zygoloid: The above is a floating table, so we shouldn't have an unnumbered paragraph break here. The table might be rendered somewhere else.

source/lex.tex
```diff
@@ -1007,28 +1009,29 @@
...
+In a \grammarterm{hexadecimal-literal},
+the letters \tcode{a} through \tcode{f} and \tcode{A} through \tcode{F}
```
zygoloid: > The *hexadecimal-digit*s \tcode{a} through \tcode{f} and \tcode{A} through \tcode{F}

source/lex.tex
```diff
@@ -1396,7 +1405,7 @@
 \begin{bnf}
 \nontermdef{hexadecimal-fractional-constant}\br
     \opt{hexadecimal-digit-sequence} \terminal{.} hexadecimal-digit-sequence\br
-    hexadecimal-digit-sequence \terminal{.}
+    hexadecimal-digit-sequence \opt{\terminal{.}}
```
zygoloid: Hmm. I can see why this was done the way it was (to make the decimal and hexadecimal cases more parallel). Do you think this is worthwhile?

source/compatibility.tex
jensmaurer: Fixed. (We should probably fix that globally, later.)

source/lex.tex
jensmaurer: Sounds good.

source/lex.tex
jensmaurer: Then, let's add a number.

source/lex.tex

source/lex.tex
jensmaurer: It's a bit unfortunate that we don't have a good way to talk about the significand in the decimal case, because it's two grammar productions. Other than that, the old grammar just seemed needlessly verbose.  If you feel the axe was too large for that particular part, I can undo the change.

zygoloid: I would prefer to keep the grammar productions parallel between the decimal and hexadecimal case. This also keeps our lexical grammar matching C's in this regard, which seems valuable. So yes, please revert the grammar change.

source/lex.tex
```diff
@@ -1033,28 +1035,39 @@
...
+\topline
+\lhdr{Kind of \grammarterm{integer-literal}} & \rhdr{base $N$} \\ \capsep
+\grammarterm{binary-literal} & 2 \\
+\grammarterm{octal-literal} & 8 \\
+\grammarterm{decimal-literal} & 10 \\
+\grammarterm{hexadecimal-literal} & 16 \\
+\end{simpletypetable}
+
+\pnum
+In a \grammarterm{hexadecimal-literal},
```
zygoloid: Not only in a *hexadecimal-literal*, but in general (this applies to *hexadecimal-floating-point-literal*s too).

zygoloid: (Maybe just delete this line?)

source/lex.tex
```diff
@@ -1068,18 +1081,19 @@
...
 \indextext{suffix!\idxcode{l}}%
 \indextext{suffix!\idxcode{u}}%
-The type of an integer literal is the first of the corresponding list
-in \tref{lex.icon.type} in which its value can be
-represented.
+Using the optional \grammarterm{integer-suffix},
+the type of an \grammarterm{integer-literal} is
+the first of the corresponding list in \tref{lex.icon.type}
+in which its value can be represented.
+An \grammarterm{integer-literal} is a prvalue.
```
zygoloid: I find this phrasing a little awkward. How about:

> The type of an *integer-literal* is the first type in the list in \tref{lex.icon.type} corresponding to its optional *integer-suffix* in which its value can be represented.

source/lex.tex
jensmaurer: Grammar change reverted.

source/lex.tex
jensmaurer: Deleted.

source/lex.tex
jensmaurer: Yes, that's better.

source/lex.tex
jensmaurer: I've also augmented the definition of "significand" to account for the two different grammar options.

source/lex.tex
```diff
@@ -1454,64 +1473,64 @@
...
+\indextext{literal!floating-point}%
+The \defn{significand} of a \grammarterm{floating-point-literal}
+is the \grammarterm{fractional-constant} or \grammarterm{digit-sequence}
+of a \grammarterm{decimal-floating-point-literal}
+or the \grammarterm{hexadecimal-fractional-constant}
+or \grammarterm{hexadecimal-digit-sequence}
+of a \grammarterm{hexadecimal-floating-point-literal}.
+A \grammarterm{digit-sequence} is interpreted as a decimal (base ten) number.
+A \grammarterm{hexadecimal-digit-sequence} is interpreted
+as a hexadecimal (base sixteen) number.
```
zygoloid: I find this to be underspecified in the fractional case; the number 0123.0123 is not "one hundred and twenty three point one hundred and twenty three". Similar to what we did for the integer case, can we replace all of this with:

> In the significand, the sequence of *digit*s or *hexadecimal-digit*s and optional period are interpreted as a base $N$ real number, where $N$ is 10 for a *decimal-floating-point-literal* or 16 for a *hexadecimal-floating-point-literal*.

(maybe with a table) and get rid of the "optional separating single quotes [...] are ignored" wording below?

source/lex.tex
```diff
@@ -1454,64 +1473,64 @@
...
+In a \grammarterm{decimal-floating-point-literal},
+the \grammarterm{exponent-part}, if present,
+indicates the power of 10 by which the significand is to be scaled.
```
zygoloid: We don't say how to interpret the *exponent-part*. Maybe

> If an *exponent-part* or *binary-exponent-part* is present, the *exponent* of the *floating-point-literal* is the result of interpreting the sequence of an optional *sign* and the *digit*s as a base 10 integer. Otherwise, the exponent is 0. The *scaled value* is *s* * 10^*e* for a *decimal-floating-point-literal* and *s* * 2^*e* for a *hexadecimal-floating-point-literal*, where *s* is the significand and *e* is the exponent.

source/lex.tex
jensmaurer: A table with two lines? Seems overkill.
Used "and" not "or" for the 10 vs. 16 part (both parts apply, under their respective conditions).
I'd prefer to turn the single quotes sentence into a note, similar to lex.icon.

zygoloid: Using a note sounds good to me. I'm fine with either a table or non-table approach (slight preference for non-table), but was leaving the option open in case you would favor consistency with the integer literals section.

source/lex.tex
jensmaurer: Hyphenated base-10, actually introduced $s$ and used math-mode \times. Also replaced "the exponent is 0" with "the exponent $e$ is 0".

jensmaurer: Hyphenated base-10, actually introduced $s$ and used math-mode \times. Also replaced "the exponent is 0" with "the exponent $e$ is 0".

source/lex.tex
```diff
@@ -1033,28 +1035,38 @@
...
-digits, which include the decimal digits and the letters \tcode{a}
-through \tcode{f} and \tcode{A} through \tcode{F} with decimal values
-ten through fifteen.
+In an \grammarterm{integer-literal},
+the sequence of
+\grammarterm{binary-digit}s,
+\grammarterm{octal-digit}s,
+\grammarterm{digit}s, or
+\grammarterm{hexadecimal-digit}s
+is interpreted as a base-$N$ number as shown in table \tref{lex.icon.base};
```
zygoloid: Remove the hyphen here to avoid confusion with base -N, which is very different from what we mean :) Also maybe "integer" rather than "number"?

jensmaurer: It's a compound adjective. Plus the "minus" is not in math font, so we're good. (Is there a difference to start with?) Maybe "as an integer of base $N$" to avoid the adjective? Or "as an integer represented with base $N$"?

jensmaurer:
"base $N$ integer" and "base 10 integer" is it.

source/lex.tex
```diff
@@ -1878,8 +1901,7 @@
 \nontermdef{user-defined-floating-point-literal}\br
     fractional-constant \opt{exponent-part} ud-suffix\br
     digit-sequence exponent-part ud-suffix\br
-    hexadecimal-prefix hexadecimal-fractional-constant binary-exponent-part ud-suffix\br
-    hexadecimal-prefix hexadecimal-digit-sequence binary-exponent-part ud-suffix
+    hexadecimal-prefix hexadecimal-fractional-constant binary-exponent-part ud-suffix
```
zygoloid: This change also needs to be reverted.

jensmaurer: Done. Sorry, I don't know how I missed it earlier.

PR #3656, opened 2020-01-31
[range.join.sentinel], [range.split.view] fix typos

pppalka:
In [range.join.sentinel], we were referring to the nonexistent data member `parent_` instead of `parent`.

In [range.split.view], there was an invalid `const` qualifier in the signature for `split_view::inner-iterator::operator++()`.


pppalka:
(Sorry about those accidental force-pushes, was trying to prepare another PR.)

PR #3659, opened 2020-02-04
[class.ctor/dtor] consteval, constructors, and destructors

sdkrystian:
It seems that the specification as to where `consteval` can be applied to regarding constructors and destructors has been... swapped. Intent in [the paper](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1073r3.html) seems pretty clear to me.


jensmaurer:
I'm not seeing a reference to [class.ctor] in the referenced paper. 
It seems http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0784r7.html did that, though. And that paper very clearly edits [class.dtor], but not [class.ctor].

Please be more specific why you believe the omission from the list in [class.ctor] is editorial.

sdkrystian:
Ah yeah, this is definitely not editorial. I'll open an issue.

PR #3660, opened 2020-02-05
Consistent cv-qualification notation

sdkrystian:
There are a lot of inconsistencies in how we specify cv-qualifiers in the text, so I went through and fixed most of the more offending examples. Most of the changes involved replacing "possibly cv-qualified" where *cv* could be used, however there were a few other things I went ahead and fixed:

- The "house style" is to say "possibly cv-qualified", but there are two occurrences of "optionally cv-qualified". One was changed to use *cv*, the other was changed to the correct phrasing.

- [[dcl.init] p 17 sub 6.2.2](http://eel.is/c++draft/dcl.init#17.6.2.2) Introduces the named type `A`, but doesn't reference it again.

- [[over.match.copy]](http://eel.is/c++draft/over.match.copy) had a type with described as "reference to possibly cv-qualified `T`". Quite obvious what's wrong there.

A couple other more unrelated things I noticed going through:

[[basic.type.qualifier] p2](http://eel.is/c++draft/basic.type.qualifier#2.sentence-2) is made redundant by [[basic.type.qualifier] p6](http://eel.is/c++draft/basic.type.qualifier#6), and [[basic.type.qualifier] p3](http://eel.is/c++draft/basic.type.qualifier#3) can probably be turned into a note.

In [[dcl.init.list] p2](http://eel.is/c++draft/dcl.init.list#2) we say that constructors take `std::initializer_list<E>` and references to possibly cv-qualified `std::initializer_list<E>` are initializer list constructors. I think *cv* is missing from the first one, but that is definitely not editorial.

In [[conv.ptr] p3](http://eel.is/c++draft/conv.ptr#3) we say "The null pointer value is converted to the null pointer value of the destination type.". The definite article does not seem to fit there, as a specific null pointer value was not established (presumably its referring to a derived to base pointer conversion where the value of the derived pointer is a null pointer value)

[[temp.deduct.partial] p5](http://eel.is/c++draft/temp.deduct.partial#5) and [[temp.deduct.partial] p7](http://eel.is/c++draft/temp.deduct.partial#7) are a little excessive for what they do. All other instances of this is done in a much less verbose fashion e.x. [[temp.deduct.call]](http://eel.is/c++draft/temp.deduct.call)

In [[over.built]](http://eel.is/c++draft/over.built), every occurance of "every cv-qualified or cv-unqualified [...] `T`" can simply be replaced with "every [...] *cv* `T`".

This is a little outside the scope of what I was looking for here, but the 3 paragraphs in [[class.init]](http://eel.is/c++draft/class.init) are just watered down versions of the wording they link to. These paragraphs could just be removed, or turned into notes, as they don't serve any normative purpose.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -4588,7 +4588,7 @@ are collectively called \defn{standard-layout types}.
 \pnum
 A type is a \defn{literal type} if it is:
 \begin{itemize}
-\item possibly cv-qualified \tcode{void}; or
+\item \cv{}~\tcode{void}; or
 \item a scalar type; or
 \item a reference type; or
 \item an array of literal type; or
@@ -5073,8 +5073,7 @@ even though they have the same address.
 \pnum
 \indextext{pointer|seealso{\tcode{void*}}}%
 \indextext{\idxcode{void*}!type}%
-A pointer to \cv-qualified\iref{basic.type.qualifier} or \cv-unqualified
-\tcode{void}
+A pointer to \cv{}~\tcode{void}
 can be used to point to objects of
 unknown type. Such a pointer shall be able to hold any object pointer.
 An object of type \cv{}~\tcode{void*}
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -1538,9 +1538,7 @@ or
 \tcode{volatile}
 \tcode{X}
 cannot initialize an object of type
-(possibly
-cv-qualified)
-\tcode{X}.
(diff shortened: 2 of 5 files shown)


jensmaurer:
Please create separate github issues for the other concerns.

sdkrystian:
@jensmaurer I committed one of the commits to the wrong branch; should be fixed now.

PR #3661, opened 2020-02-05
[range.istream.iterator] Fix typo

pppalka:
The description of basic_istream_view::iterator::operator*() refers to a
nonexistent data member value_ of basic_istream_view instead of the data
member object_.


pppalka:
Thanks, the subclause stable name in the commit message has been adjusted via force-push.

PR #3667, opened 2020-02-10
[class.access.base] Add missing formatting to an access specifier

maggyero:


source/classes.tex
```diff
@@ -4748,7 +4748,7 @@
 public and protected members of the base class are accessible as
 protected members of the derived class.
 If a class is declared to be a base class for another class using the
-private
+\tcode{private}
```
jensmaurer: That should have been \keyword. Oh well.

maggyero: `\keyword` seems to only be used in BNF contexts.

tkoeppe: The sibling uses of this pattern use `tcode` as well, so the change is at least locally consistent. A global cleanup would of course be welcome.

PR #3671, opened 2020-02-11
Fix definition of has-tuple-element

akrzemi1:
It is the only example in the WD that uses a removed feature of specifying a type of the expression.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -6103,7 +6103,7 @@ namespace std::ranges {
       typename tuple_size<T>::type;
       requires N < tuple_size_v<T>;
       typename tuple_element_t<N, T>;
-      { get<N>(t) } -> const tuple_element_t<N, T>&;
+      { get<N>(t) } -> convertible_to<const tuple_element_t<N, T>&>;
     };
 
 


jwakely:
Why `convertible_to` instead of `same_as`?

akrzemi1:
My reasoning was, it used to be *trailing-return-type* which used to mean almost the same as `convertible_to<T>` prior to [P1452r2](http://wiki.edg.com/pub/Wg21cologne2019/StrawPolls/p1452r2.htm). `convertible_to` preserves this intent. 

Unless `-> const tuple_element_t<N, T>&` was a bug from the outset?

jwakely:
I _think_ it probably should have always required exactly that type, it was just more awkward to specify that prior to P1452R2. But I'm not sure about the original intent.

I agree that as an editorial change, using `convertible_to` is closer to the earlier meaning. But I wonder if we actually have a non-editorial defect and should require exactly that type.

@CaseyCarter ?

CaseyCarter:
The conversion was intended in the original design: `std::get<X>(t)` may return a reference to `const`, or reference to non-`const`, or even a prvalue for `subrange`. The exact type and value category are largely immaterial - this is just a "sanity check" requirement - since the only use to which the implementation puts `get` is to return `std::get<X>(...)` from `operator*`.

CaseyCarter:
I should point out that [LWG 3323](https://cplusplus.github.io/LWG/issue3323) resolves this, which we will be voting to move into the WD this week in Prague.

akrzemi1:
Ok, so where do we go from here? Should I close this PR given that it is reported as a LWG issue?

zygoloid:
Fixed by 74487d32b52ea64f5e6fb63908b584d2a8568c30.

PR #3672, opened 2020-02-12
[class.access] Eliminate the friend case for protected member access from derived class

geryogam:
This PR will fix an inconsistency in the specification described here:
https://stackoverflow.com/q/60178872/2326961

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -4472,8 +4472,8 @@ of the class in which it is declared.
 \indextext{access control!\idxcode{protected}}%
 protected;
 that is, its name can be used only by members and friends
-of the class in which it is declared, by classes derived from that class, and by their
-friends (see~\ref{class.protected}).
+of the class in which it is declared, and members of classes derived from that class
+(see~\ref{class.protected}).
 \item
 \indextext{access control!\idxcode{public}}%
 public;


zygoloid:
Hmm. It's unclear whether [class.access]p1 is abstractly talking about a direct member of a class, or about a member of a class when named as a member of some (possibly) different class. We should clarify. I think it probably does mean the latter, which would make this change correct.

Maybe adding a note:

> [Note: A member of a base class is also a member of a derived class, and the same entity may have different access in each class in which it is a member.]

would help?

source/classes.tex
```diff
@@ -4472,8 +4472,8 @@
 \indextext{access control!\idxcode{protected}}%
 protected;
 that is, its name can be used only by members and friends
-of the class in which it is declared, by classes derived from that class, and by their
-friends (see~\ref{class.protected}).
+of the class in which it is declared, and members of classes derived from that class
```
jensmaurer: "... and by members of classes derived..." (add "by" to keep the sentence structure manageable)

jensmaurer: One more concern here: We change "by classes" to "by members of classes". This change seems to exclude the base-specifier, which is not in a member. Example:

struct B { };
struct D1 : protected B { };
struct D2 : D1  {};

I think the injected-class-name B is protected in D1, thus accessible from D2. But the name doesn't appear in a member.

That said, this paragraph is just one of those intro sentences where the actual rules appear later. (Note the funny punctuation, btw.) So, we shouldn't say wrong things, but we don't need to be ultra-precise either.

jensmaurer:
@zygoloid , I'm not in favor of adding a note here. This is an intro sentence, so let's keep it as vague as we possibly can.

zygoloid:
This is too far into the technical details to address without CWG oversight.

tkoeppe:
Could you please rebase?

PR #3675, opened 2020-02-13
[conv.rank] Change "size" to "width" in conversion rank relation

tkoeppe:
The rules about integer range limits imply (in a roundabout fashion)
that increasing rank implies increasing width, so this is a
clarification that makes this implication explicit.


zygoloid:
Approved by CWG 2020-02-13.

zygoloid:
The other two occurrences of 'size' in [conv.rank] and the occurrence in [basic.fundamental]p1 should also be fixed.

tkoeppe:
Thanks! A squashable commit has been pushed.

PR #3678, opened 2020-02-14
[stmt.return.coroutine] Use notion of "enclosing" for "return"

birbacher:


birbacher:
We use “enclosing” (defined in [stmt.pre]) for defining a coroutine in [dcl.fct.def.coroutine], so also use it here.  Jens suggested to move the sentence I changed from [stmt.return.coroutine] to [stmt.return] somewhere.

PR #3730, opened 2020-02-17
[atomic] Minor corrections to atomic index

AlisdairM:
The index for the primary atomic template was split across
two entries, atomic and atomic<T>.  Reconciled into one.

A couple of operations for atomic smart pointers are
specificed in the primary template, but were missing the
cross-reference into that specification, which is supplied
in the same way for all the other atomic specializations.


CaseyCarter:
> The index for the primary atomic template was split across two entries, atomic and atomic.

I can't tell if there's a markup problem here or if this is British humor.

AlisdairM:
Sorry, somewhere along the lines the angle brackets for atomic[T] got swallowed as a bad HTML tag - trying this comment with squared brackets to be safe...

PR #3735, opened 2020-02-18
[dcl.meaning] General wording cleanup

sdkrystian:
The conventions used to specify the meaning of declarators are a little inconsistent, and the sub-clause itself has a lot of single sentence paragraphs that look messy.

For example, [dcl.meaning] has a mix of specifying the "type of the identifier" and "type of the _declarator-id_", [dcl.ptr], [dcl.ref] and [dcl.mptr] determine the "type of the identifier", and [dcl.array] and [dcl.fct] determine the "type of the _declarator-id_". The mentions of "determining the type of the identifier" have been changed to determining the type of the _declarator-id_, as this is a better defined term.

Further, [[dcl.meaning] p3](http://eel.is/c++draft/dcl.meaning#3) correctly acknowledged that `T` can contain an  _attribute-specifier-seq_, but in the very next paragraph it assumes that `T` will just be a _decl-specifier-seq_. [dcl.meaning] p3 is rewritten to account for that, and it additionally specified that the _attribute-specifier-seq_ appertains to the declarator (this is also specified in [[dcl.pre] p2](http://eel.is/c++draft/dcl.pre#2.sentence-5), perhaps that mention could be removed, as it makes more sense to have it where the meaning of declarators is determined).

More stylistically, [dcl.meaning] uses `codeblock` when specifying a declaration of the for `T D`, which needlessly uses whitespace, and is inconsistent with the other sub-clauses. These were changed to use `\tcode`.

Lastly, [the last sentence of [dcl.meaning] p6](http://eel.is/c++draft/dcl.meaning#6.sentence-2) is completely redundant, and should be just turned into a note.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -2519,25 +2519,15 @@ the type specified for each \grammarterm{declarator-id} depends on
 both the \grammarterm{decl-specifier-seq} and its \grammarterm{declarator}.
 
 \pnum
-Thus, a declaration of a particular identifier has the form
-\tcode{T} \tcode{D} where \tcode{T}
-is of the form \opt{\grammarterm{attribute-specifier-seq}}
-\grammarterm{decl-specifier-seq}
-and \tcode{D} is a declarator.
-Following is a recursive procedure for determining
-the type specified for the contained
-\grammarterm{declarator-id}
-by such a declaration.
+Thus, a declaration of a particular \grammarterm{declarator-id} has the form
+\opt{\grammarterm{attribute-specifier-seq}} \grammarterm{decl-specifier-seq}
+\grammarterm{declarator}. The optional \grammarterm{attribute-specifier-seq}
+appertains to the declared entity. Following is a recursive procedure for
+determining the type of the \grammarterm{declarator-id} and the entity which it declares.
 
 \pnum
-First, the type \tcode{T} is determined.
-In a declaration of the form
-\begin{ncsimplebnf}
-\opt{attribute-specifier-seq} decl-specifier-seq \terminal{D}
-\end{ncsimplebnf}
-the \grammarterm{decl-specifier-seq} determines the type \tcode{T}.
-The optional \grammarterm{attribute-specifier-seq} appertains
-to the entity declared by \tcode{D}.
+First, a type \tcode{T} is determined from
+the \grammarterm{decl-specifier-seq}.
 \begin{example}
 In the declaration
 \begin{codeblock}
@@ -2551,27 +2541,20 @@ determine the type
 \end{example}
 
 \pnum
-In a declaration \tcode{T} \tcode{D}
-where \tcode{D} has the form
+A \grammarterm{declarator} of the form
 \begin{ncsimplebnf}
 declarator-id \opt{attribute-specifier-seq}
 \end{ncsimplebnf}
-the type of the \grammarterm{declarator-id}
-in \tcode{D} is ``\tcode{T}''.
+determines the type of its
+\grammarterm{declarator-id} to be ``\tcode{T}''.
 The optional \grammarterm{attribute-specifier-seq}
-appertains to the entity declared by \tcode{D}.
+appertains to the declared entity.
 
 \pnum
-In a declaration
-\tcode{T} \tcode{D} where
-\tcode{D} has the form
-\begin{ncsimplebnf}
-\terminal{(} \terminal{D1} \terminal{)}
-\end{ncsimplebnf}
-the type of the \grammarterm{declarator-id}
-in \tcode{D} is the same as that of the
-contained \grammarterm{declarator-id}
-in the declaration \tcode{T} \tcode{D1}.
+A \grammarterm{declarator} of the form \tcode{(D)}
+determines the type of its \grammarterm{declarator-id}
+to be that of the \grammarterm{declarator-id}
+in the declaration \tcode{T} \tcode{D}.
 \indextext{declaration!parentheses in}%
 \begin{note}
 Parentheses do not alter the type of the embedded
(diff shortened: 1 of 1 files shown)


source/declarations.tex
```diff
@@ -2563,13 +2565,23 @@
...
-is an unadorned identifier the type of this identifier is
-``\tcode{T}''.
+has the form
+\begin{ncsimplebnf}
+declarator-id \opt{attribute-specifier-seq}
+\end{ncsimplebnf}
+the type of the
+\grammarterm{declarator-id}
+in
+\tcode{D}
```
jensmaurer: I think  "in D" can go here; there is only one declarator-id we talk about in this context.

source/declarations.tex
```diff
@@ -2563,13 +2565,23 @@
...
+is ``\tcode{T}''.
+The optional
+\grammarterm{attribute-specifier-seq}
```
jensmaurer: Since this seems to be new text, please follow the "semantic linebreaks" conventions. In particular, don't start a new line when a \xxx LaTeX thing occurs. (These are historic accidents from when we did the standard in troff.)

source/declarations.tex
jensmaurer: Ah, no, that's the pattern we use elsewhere, too.

jensmaurer (approved):
Looks fine (and is an improvement). It would be good if the line breaks wouldn't follow a troff tradition.

source/declarations.tex
sdkrystian: Should I change this for all the modified lines? Or just for new text.

jensmaurer: As a first step, the usual green blob

then the type of the
\grammarterm{declarator-id}
in

should be on one line, and also the corresponding blob above (maybe that needs two lines).
Also
\tcode{T}
\tcode{D}
should be on one line for the new next.

Please don't wholesale-change the unchanged stuff; that should be part of a different change.

source/declarations.tex
```diff
@@ -2521,34 +2520,24 @@
...
-In a declaration
-\begin{codeblock}
-T D
-\end{codeblock}
-the
-\grammarterm{decl-specifier-seq}
-\tcode{T}
-determines the type
-\tcode{T}.
+First, the type \tcode{T} is determined.
```
zygoloid: We have two different things we're calling `T` here: a *decl-specifier-seq* and a type. This rewrite makes it less clear that that's happening by removing the "the *decl-specifier-seq* `T` determines the type `T`" part. I think we should take this a bit further. How about replacing the previous paragraph ("Thus[...]") with:

> Thus, a declaration of a particular identifier has the form *attribute-specifier-seq*[opt] *decl-specifier-seq* *declarator*. The optional *attribute-specifier-seq* appertains to the declared entity. Following is a recursive procedure for determining the type of the declared entity.

Then we only have one `T`, and it's a type, not a *decl-specifier-seq*. So we can start this second paragraph with something like:

> First, a type `T` is determined from the *decl-specifier-seq*. [Example: ...

source/declarations.tex
```diff
@@ -2562,37 +2551,33 @@
...
-``\tcode{T}''.
+In a declaration \tcode{T} \tcode{D}
+where \tcode{D} has the form
+\begin{ncsimplebnf}
+declarator-id \opt{attribute-specifier-seq}
+\end{ncsimplebnf}
+the type of the \grammarterm{declarator-id}
+in \tcode{D} is ``\tcode{T}''.
+The optional \grammarterm{attribute-specifier-seq}
+appertains to the entity declared by \tcode{D}.
```
zygoloid: We don't need the "In a declaration" part here. How about reducing this to:

> A *declarator* of the form *declarator-id* *attribute-specifier-seq*[opt] determines the type of its declared entity to be `T`. The optional *attribute-specifier-seq* appertains to the declared entity.

source/declarations.tex
```diff
@@ -2562,37 +2551,33 @@
...
+the type of the \grammarterm{declarator-id}
+in \tcode{D} is the same as that of the
+contained \grammarterm{declarator-id}
+in the declaration \tcode{T} \tcode{D1}.
 \indextext{declaration!parentheses in}%
+\begin{note}
 Parentheses do not alter the type of the embedded
 \grammarterm{declarator-id},
 but they can alter the binding of complex declarators.
+\end{note}
```
zygoloid: > A *declarator* of the form `(` *D* `)` determines the type of its declared entity to be the same as the type determined by the *declarator* *D*. [Note: ...]

source/declarations.tex
```diff
@@ -2607,12 +2592,16 @@
...
 \terminal{*} \opt{attribute-specifier-seq} \opt{cv-qualifier-seq} \terminal{D1}
 \end{ncsimplebnf}
-and the type of the identifier in the declaration
+and the type of the contained
+\grammarterm{declarator-id}
+in the declaration
 \tcode{T}
 \tcode{D1}
 is ``\placeholder{derived-declarator-type-list}
 \tcode{T}'',
```
zygoloid: These, again, are two completely different `T`s. =(

source/declarations.tex
```diff
@@ -2607,12 +2592,16 @@
...
-then the type of the identifier of
+then the type of the
+\grammarterm{declarator-id}
+in
 \tcode{D}
 is ``\placeholder{derived-declarator-type-list} \grammarterm{cv-qualifier-seq} pointer to
 \tcode{T}''.
```
zygoloid: > A *declarator* of the form `*` *attribute-specifier-seq*[opt] *cv-qualifier-seq* *D* determines the type of its declared entity to be the type of the entity declared by the declaration *U* *D*, where *U* is a *type-specifier* that denotes the type ''*cv-qualifier-seq* pointer to `T`''.

source/declarations.tex
```diff
@@ -2713,12 +2702,16 @@
...
 \tcode{D1}
 is ``\placeholder{derived-declarator-type-list}
 \tcode{T}'',
-then the type of the identifier of
+then the type of the
+\grammarterm{declarator-id}
+in
 \tcode{D}
 is ``\placeholder{derived-declarator-type-list} reference to
 \tcode{T}''.
```
zygoloid: > A *declarator* of the form `&` *attribute-specifier-seq*[opt] *D* or `&&` *attribute-specifier-seq*[opt] *D* determines the type of its declared entity to be the type of the entity declared by the declaration *U* *D*, where *U* is a *type-specifier* that denotes the type ''reference to `T`''.

(And so on for later subclauses.)

zygoloid (changes requested):
I think we can clean this up quite a bit more. Suggestion follows.

source/declarations.tex
jensmaurer: The problem is that a declarator (alone) does not determine any type at all. Example: `(*p)` is a declarator, but doesn't actually say what the type is: "pointer to whatever" is not a type. The rephrasing here departs from the careful constructions involved derived-declarator-type-list.

jensmaurer:
While I very much appreciate a general cleanup in this area, I'm mildly favoring a core issue for this, or at least CWG review of the editorial change. @zygoloid , are you reasonably sure we're not destroying `int (*p)()` vs. `int *p()` with the reformulation? (The reformulation looks reasonable to me, but this is not the first time in history that CWG was bitten by this wording. Maybe that's an argument for refomulating.) Note that we've lived with the status quo since C++98, so I'm slightly uneasy about a last-minute fix for C++20.
Please don't read this as a veto, just as a "github diff format isn't actually the right tool for this kind of change".

source/declarations.tex
sdkrystian: Its true that its not quite needed, however it makes it consistent with the rest of [dcl.meaning]

source/declarations.tex
sdkrystian: I think that there is no issue here - the point is to use the same decl-specifier-seq in invented declaration `T D1`.

source/declarations.tex
sdkrystian: @zygoloid The removal of _derived-declarator-type-list_ is definitely a good idea, however, moving the form of the declarator to be inline does degrade readability, especially for longer declarator forms (i.e. functions). Perhaps we keep the form of the declarators indented as is done currently, and apply the other changes to the paragraphs?

sdkrystian: Also, maintaining `T` throughout subclauses is a little awkward, and could lead to confusion (typically, when "variables" are established, we carry them no further than then end of a paragraph, let alone a subclause)

source/declarations.tex
sdkrystian: See my comments on maintaining `T` through subclauses

source/declarations.tex
sdkrystian: Perhaps we want to talk about the type of the *declarator-id* rather than the declared entity? Declarations introduce names (which denote entities), after all.

sdkrystian:
@zygoloid I applied the changes you suggested, but rather than referring to "the declared entity", I changed it to refer to the _declarator-id_ itself.

sdkrystian:
@jensmaurer I walked through a few complex declarators manually following the new wording and it all checks out.

tkoeppe:
@sdkrystian: could you please rebase this (maybe after the mailing goes out), and let me know, and then I'll ask CWG to review this. The cleanup looks laudable, but I'd like some committee oversight on it.

sdkrystian:
@jensmaurer @tkoeppe I apologize for going dark for the last few years... I will be going through my open PRs addressing comments/rebasing over the coming weeks.

PR #3736, opened 2020-02-18
[basic.type.qualifier] Remove redundant wording

sdkrystian:
[[basic.type.qualifier] p2](http://eel.is/c++draft/basic.type.qualifier#2.sentence-2) and [[basic.type.qualifier] p6](http://eel.is/c++draft/basic.type.qualifier#6.sentence-1) are made reduntand by [[dcl.array] p5](http://eel.is/c++draft/dcl.array#5). 

Removing [basic.type.qualifier] p2 sentence 2, adjusting the first sentence to acknowledge what is stated in p6, and turning [basic.type.qualifier] p6 sentence 1 into a note resolves this.

In addition to this, [[basic.type.qualifier] p3](http://eel.is/c++draft/basic.type.qualifier#3) should be turned into a note, as it has no normative effects.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -5120,15 +5120,38 @@ interchangeability as arguments to functions, return values from
 functions, and non-static data members of unions.}
 
 \pnum
-\indextext{array!\idxcode{const}}%
-A compound type\iref{basic.compound} is not cv-qualified by the
-cv-qualifiers (if any) of the types from which it is compounded. Any
-cv-qualifiers applied to an array type
-affect the array element type\iref{dcl.array}.
+Except for array types, a compound type\iref{basic.compound} is not cv-qualified by the
+cv-qualifiers (if any) of the types from which it is compounded.
 
 \pnum
+\indextext{array!\idxcode{const}}%
+An array type whose elements are cv-qualified
+is also considered to have the same cv-qualifications
+as its elements.
+\begin{note}
+Cv-qualifiers applied to an array
+type attach to the underlying element type, so the notation
+``\cv{}~\tcode{T}'', where \tcode{T} is an array type, refers to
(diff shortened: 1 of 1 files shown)


source/basic.tex
```diff
@@ -5168,12 +5167,16 @@
 \end{example}
 
 \pnum
+\indextext{array!\idxcode{const}}%
+An array type whose elements are cv-qualified
+is also considered to have the same cv-qualifications
+as its elements.
+\begin{note}
 Cv-qualifiers applied to an array
 type attach to the underlying element type, so the notation
```
zygoloid: We need two rules:

 1) `identity_t<const T>[N]` is a const-qualified type because its element type is const-qualified.
 2) `const identity_t<int[3]>` has a const-qualified element type.

Before this change, we say (2) in two places (unless I missed one). This change removes one of them (on line 5126) and converts the other (here) to a note, removing any normative specification of this rule. The new normative text (lines 5170-5174) says only (1), so we're left with nowhere that provides rule (2).

source/basic.tex
```diff
@@ -5168,12 +5167,16 @@
...
 ``\cv{}~\tcode{T}'', where \tcode{T} is an array type, refers to
-an array whose elements are so-qualified. An array type whose elements
-are cv-qualified is also considered to have the same cv-qualifications
-as its elements.
+an array whose elements are so-qualified.
+\end{note}
```
zygoloid: Please add a cross-reference to [dcl.array] for this.

zygoloid: Let's reorder this to right after the "from which it is compounded" rule, so the reader isn't left wondering what happens for array types for several paragraphs while we talk about other things.

sdkrystian: @zygoloid Good idea. Both changes applied.

PR #3747, opened 2020-02-19
CWG Poll 15: P2095R0 Resolve lambda init-capture pack grammar (CWG2378)

### LINKED ISSUE #3699: [2020-02 CWG Poll 15] P2095R0 Resolve lambda init-capture pack grammar (CWG2378)

burblebee:
Fixes #3699.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -1947,23 +1947,22 @@ named in a friend declaration\iref{class.friend}.
 
 \begin{bnf}
 \nontermdef{capture}\br
-    simple-capture \opt{\terminal{...}}\br
-    % FIXME: This is wrong. The ... should go after the &.
-    \opt{\terminal{...}} init-capture
+    simple-capture\br
+    init-capture
 \end{bnf}
 
 \begin{bnf}
 \nontermdef{simple-capture}\br
-    identifier\br
-    \terminal{\&} identifier\br
+    identifier \opt{\terminal{...}}\br
+    \terminal{\&} identifier \opt{\terminal{...}}\br
     \keyword{this}\br
     \terminal{*} \terminal{this}
 \end{bnf}
 
 \begin{bnf}
 \nontermdef{init-capture}\br
-    identifier initializer\br
-    \terminal{\&} identifier initializer
+    \opt{\terminal{...}} identifier initializer\br
+    \terminal{\&} \opt{\terminal{...}} identifier initializer
 \end{bnf}
 
 \pnum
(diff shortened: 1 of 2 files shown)


source/ranges.tex
```diff
@@ -3022,6 +3022,8 @@
     constexpr V base() const& requires copy_constructible<V> { return @\exposid{base_}@; }
     constexpr V base() && { return std::move(@\exposid{base_}@); }
 
+    constexpr const Pred& pred() const;
```
johelegp: This is unrelated.

source/expressions.tex
```diff
@@ -1980,7 +1979,7 @@
 \grammarterm{capture-default} that is \tcode{=}, each
 \grammarterm{simple-capture} of that \grammarterm{lambda-capture} shall
 be of the form
-``\tcode{\&} \grammarterm{identifier}'',
+``\tcode{\&} \opt{\tcode{...}} \grammarterm{identifier}'',
```
cpplearner: `...` on the wrong side?

source/expressions.tex
```diff
@@ -1947,23 +1947,22 @@
...
+    simple-capture\br
+    init-capture
 \end{bnf}
 
 \begin{bnf}
 \nontermdef{simple-capture}\br
-    identifier\br
-    \terminal{\&} identifier\br
+    identifier \opt{\terminal{...}}\br
+    \terminal{\&} identifier \opt{\terminal{...}}\br
```
BRevzin: Wow, I am... just... absolutely the worst at this. I am so sorry.

This PR matches the paper. The paper is incorrect. The `...` should be between the `&` and the `identifier`, as they are in the init-capture case. Can we just Do The Right Thing ™️ here?

@zygoloid @jensmaurer

source/expressions.tex
BRevzin: This one is on the correct (and right 😉 ) side. But the grammar is wrong, sigh.

BRevzin: God I am just the very worst at this. And then I confused myself enough to leave the totally wrong comment about this. 

This PR matches the paper, but the paper is wrong... the `...` should definitely be on the right side of the _`identifier`_. This makes me very sad. I'm going to go sit in a corner and think about what I did.

In the meantime, can we just Do The Right Thing ™️  and move the `...`?

burblebee: While technically a normative change, I think this was an obvious mistake; fixed editorially.

PR #3753, opened 2020-02-19
LWG Poll 14: P1868R2 🦄 width: clarifying units of width and precision in std::format

### LINKED ISSUE #3716: [2020-02 LWG Poll 14] P1868R2 🦄 width: clarifying units of width and precision in std::format
resolves US228 and LWG issue 3290

burblebee:
Fixes #3716.

Fixes cplusplus/papers#618.
Fixes cplusplus/nbballot#225.

### DIFF
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -35,10 +35,6 @@ For undated references, the latest edition of the referenced document
 \begin{itemize}
 \item Ecma International, \doccite{ECMAScript Language Specification},
 Standard Ecma-262, third edition, 1999.
-% FIXME: Is the following comment refering to the entry above or
-%        the entry removed in LWG3319?  I.e. is it still valid?
-%%% Format for this entry is based on that specified at
-%%% http://www.iec.ch/standardsdev/resources/draftingpublications/directives/principles/referencing.htm
 \item ISO/IEC 2382 (all parts), \doccite{Information technology ---
 Vocabulary}
 \item ISO 8601:2004, \doccite{Data elements and interchange formats ---
@@ -56,6 +52,11 @@ Microprocessor Systems --- Floating-Point arithmetic}
 \item ISO 80000-2:2009, \doccite{Quantities and units ---
 Part 2: Mathematical signs and symbols
 to be used in the natural sciences and technology}
+%%% Format for the following entry is based on that specified at
+%%% http://www.iec.ch/standardsdev/resources/draftingpublications/directives/principles/referencing.htm
+\item The Unicode Consortium. Unicode Standard Annex, UAX \#29, \doccite{Unicode Text Segmentation} [online].
+Edited by Mark Davis. Revision 35; issued for Unicode 12.0.0. 2019-02-15 [viewed 2020-02-23].
+Available at \url{http://www.unicode.org/reports/tr29/tr29-35.html}
 \end{itemize}
 
 \pnum
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -683,4 +683,7 @@ Electronic Engineers, Inc.
 ECMAScript\textregistered\ is a registered trademark of Ecma International.
 
 \pnum
+Unicode\textregistered\ is a registered trademark of Unicode, Inc.
+
+\pnum
 All rights in these originals are reserved.
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -19399,8 +19399,7 @@ as defined by UAX \#29.
 The estimated width of the following code points is 2:
 \begin{itemize}
 \item \tcode{U+1100-U+115F}
-\item \tcode{U+2329}
-\item \tcode{U+232A}
+\item \tcode{U+2329-U+232A}
 \item \tcode{U+2E80-U+303E}
 \item \tcode{U+3040-U+A4CF}
 \item \tcode{U+AC00-U+D7A3}
@@ -19409,10 +19408,10 @@ The estimated width of the following code points is 2:
 \item \tcode{U+FE30-U+FE6F}
 \item \tcode{U+FF00-U+FF60}
 \item \tcode{U+FFE0-U+FFE6}
-\item \tcode{U+20000-U+2FFFD}
-\item \tcode{U+30000-U+3FFFD}
 \item \tcode{U+1F300-U+1F64F}
 \item \tcode{U+1F900-U+1F9FF}
+\item \tcode{U+20000-U+2FFFD}
+\item \tcode{U+30000-U+3FFFD}
 \end{itemize}
 The estimated width of other code points is 1.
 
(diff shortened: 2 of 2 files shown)


source/intro.tex
```diff
@@ -60,6 +60,7 @@
 \item ISO 80000-2:2009, \doccite{Quantities and units ---
 Part 2: Mathematical signs and symbols
 to be used in the natural sciences and technology}
+\item Unicode 12.1.0 Standard, Annex \#29, Unicode Text Segmentation
```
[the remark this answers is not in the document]
burblebee: In trying to follow how this information is formatted in other entries, I went with the following.  Is it correct?

```
\item Unicode Standard Annex, UAX \#29, \doccite{Unicode Text Segmentations},
version 12.1.0.
Available at \url{http://www.unicode.org/reports/tr29/tr29-35.html}
```

source/utilities.tex
```diff
@@ -19909,6 +19909,58 @@
...
+This is the case for Windows-based and many POSIX-based operating systems.
+\end{note}
+
+\pnum
+For a string in a Unicode encoding,
+implementations should estimate the width of a string
+as the sum of estimated widths of
+the first code points in its extended grapheme clusters
+as defined by
+Unicode Standard Annex \#29 Unicode Text Segmentation.
```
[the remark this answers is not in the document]
burblebee: done

source/utilities.tex
```diff
@@ -19909,6 +19909,58 @@
...
+
+\begin{example}
+The estimated width of the following code points is 2:
+\begin{codeblock}
+U+1100 .. U+115F
```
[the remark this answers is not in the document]
burblebee: Ok, replaced " .. " with "-".

source/intro.tex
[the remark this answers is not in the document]
hubert-reinterpretcast: Noting that the normative reference to RFC 6557 is going away (so we do not have to match too closely), I think the requirements and recommendations in ISO 690 would give:
The Unicode Consortium. Unicode Standard Annex, UAX #29: _Unicode Text Segmentation_ [online]. Edited by Mark Davis. Revision 35. 2019-02-15 [viewed 2020-02-23]. Available from: http://www.unicode.org/reports/tr29/tr29-35.html

In particular, the recommended form is "Available from" and not "Available at".

Note that the cited document does not self-identify as version 12.1.0 but it is the document indicated by The Unicode Standard, Version 12.1.0.

source/utilities.tex
```diff
@@ -19909,6 +19909,58 @@
...
+For a string in a Unicode encoding,
+implementations should estimate the width of a string
+as the sum of estimated widths of
+the first code points in its extended grapheme clusters
+as defined by
+Unicode Standard Annex \#29 Unicode Text Segmentation.
+
+\begin{example}
+The estimated width of the following code points is 2:
+\begin{codeblock}
```
hubert-reinterpretcast: Should this be a list instead of a code region?

source/intro.tex
```diff
@@ -60,6 +60,9 @@
 \item ISO 80000-2:2009, \doccite{Quantities and units ---
 Part 2: Mathematical signs and symbols
 to be used in the natural sciences and technology}
+\item Unicode Standard Annex, UAX \#29, \doccite{Unicode Text Segmentations},
```
hubert-reinterpretcast: I've made a further suggestion (attached to the discussion on the previous version of this): https://github.com/cplusplus/draft/pull/3753#discussion_r383037073. Also, the title should have "Segmentation" (that is, without the "s").

source/utilities.tex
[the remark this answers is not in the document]
hubert-reinterpretcast: Was the change done and not pushed? I did not notice a change here.

source/utilities.tex
burblebee: I have no idea!  I don't even know if it's supposed to be an example.

source/intro.tex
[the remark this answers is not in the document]
burblebee: I'm going to assume you're right and reword it as you suggested :)

source/intro.tex
burblebee: thanks!

source/utilities.tex
hubert-reinterpretcast: I believe this is meant to be normative text. Presentation as a list or table would work.

timsong-cpp: Yes, this is supposed to be normative, not example.

vitaut: The list of code point ranges is part of the normative text. It is rendered as a code block in P1868 for simplicity but something like a bullet list might be a better presentation format.

burblebee: Thanks for the clarification - it's now a list.  Please have a look.

source/intro.tex
```diff
@@ -60,6 +60,9 @@
 \item ISO 80000-2:2009, \doccite{Quantities and units ---
 Part 2: Mathematical signs and symbols
 to be used in the natural sciences and technology}
+\item The Unicode Consortium. Unicode Standard Annex, UAX \#29, \doccite{Unicode Text Segmentation} [online].
+Edited by Mark Davis. Revision 35. 2019-02-15 [viewed 2020-02-23].
```
jensmaurer: Maybe say "Revision 35 for Unicode 12.0.0" to bring in the Unicode version number as well?

burblebee: I had the version in there 2 tries before and Hubert had me to take it out.  It doesn't help that the "Unicode Text Segmentation" section says it's version 12.0.0 while the document itself says it's 12.1.0.

@hubert-reinterpretcast do you agree to putting the version back in as Jens suggests?

hubert-reinterpretcast: "Revision 35; issued for Unicode 12.0.0"?

burblebee: applied Hubert's suggestion above

source/utilities.tex
```diff
@@ -19935,6 +19935,8 @@
 the first code points in its extended grapheme clusters
 as defined by
 Unicode Standard Annex \#29 Unicode Text Segmentation.
+
+\begin{example}
```
zygoloid: I believe this is the definition of "estimated width", not an example.

The "as defined by" is referring only to "extended grapheme clusters" (which are defined by Annex \#29), not to "estimated width" (which is not defined by Annex \#29).

hubert-reinterpretcast: @zygoloid, a later commit corrects this.

zygoloid: ? No, the example markers were present at the tip of the branch. I've removed them now.

source/intro.tex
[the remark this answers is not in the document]
zygoloid: From what I can discover, ISO 690 suggests what information should be present, not how to format it. All the examples in https://www.iso.org/sites/directives/current/part2/index.xhtml#_idTextAnchor119 and http://www.iec.ch/standardsdev/resources/draftingpublications/directives/principles/referencing.htm use "Available at" not "Available from", so I've changed that back.

PR #3750, opened 2020-02-19
Fix #3666, plus some drive-by edits

CaseyCarter:
See the three commit logs.

Note that the change to `basic_istream_view::iterator` uses the name `default_initializable` instead of `default_constructible` in anticipation of LWG motion 2.

Issue #3561, opened 2019-12-16
[range.adaptors] Base vs base-t


johelegp:
P1035 "Input range adaptors" used `base_t` (in `elements_view` and `take_while_view`) where the other views used `Base`. Which way to go?

jensmaurer:
Asked LWG which way to go: http://lists.isocpp.org/lib/2019/12/14918.php

jensmaurer:
Editorial meeting: It's like a template parameter. Use "Base" (italics).

PR #3754, opened 2020-02-19
LWG Poll 13: P1994R1 elements_view needs its own sentinel

### LINKED ISSUE #3715: [2020-02 LWG Poll 13] P1994R1 elements_view needs its own sentinel
resolves LWG3386

burblebee:
Also fixes LWG3386.

Fixes #3715.
Fixes cplusplus/papers#717.

Notes/Issues:
* [range.elements.sentinel] Should \exposid{Base} be \exposid{base-t} as it is in [range.elements.iterator]?  We're inconsistent wrt Base v. base-t elsewhere as well.  There are also other differences between sentinel and iterator, like "other" and "base() const" (no ref) - I assume these are intentional?

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -6420,7 +6420,9 @@ Equivalent to: \tcode{return x.\exposid{current_} - y.\exposid{current_};}
 \indexlibraryglobal{elements_view::sentinel}%
 \begin{codeblock}
 namespace std::ranges {
-  template<class V, size_t N>
+  template<input_range V, size_t N>
+    requires view<V> && @\placeholder{has-tuple-element}@<range_value_t<V>, N> &&
+             @\placeholder{has-tuple-element}@<remove_reference_t<range_reference_t<V>>, N>
   template<bool Const>
   class elements_view<V, N>::@\exposid{sentinel}@ {                 // \expos
   private:
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -6036,11 +6036,17 @@ namespace std::ranges {
     constexpr auto begin() const requires @\placeholder{simple-view}@<V>
     { return @\exposid{iterator}@<true>(ranges::begin(@\exposid{base_}@)); }
 
-    constexpr auto end() requires (!@\placeholder{simple-view}@<V>)
-    { return ranges::end(@\exposid{base_}@); }
+    constexpr auto end()
+    { return @\exposid{sentinel}@<false>{ranges::end(@\exposid{base_}@)}; }
 
-    constexpr auto end() const requires @\placeholder{simple-view}@<V>
-    { return ranges::end(@\exposid{base_}@); }
+    constexpr auto end() requires common_range<V>
+    { return @\exposid{iterator}@<false>{ranges::end(@\exposid{base_}@)}; }
+
+    constexpr auto end() const requires range<const V>
+    { return @\exposid{sentinel}@<true>{ranges::end(@\exposid{base_}@)}; }
+
+    constexpr auto end() const requires common_range<const V>
+    { return @\exposid{iterator}@<true>{ranges::end(@\exposid{base_}@)}; }
 
     constexpr auto size() requires sized_range<V>
     { return ranges::size(@\exposid{base_}@); }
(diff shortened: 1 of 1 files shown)


CaseyCarter:
> * [range.elements.sentinel] Should \exposid{Base} be \exposid{base-t} as it is in [range.elements.iterator]?  We're inconsistent wrt Base v. base-t elsewhere as well.  

Per #3561, it should be `\exposid{Base}`.

> There are also other differences between sentinel and iterator, like "other" and "base() const" (no ref) - I assume these are intentional?

Yes they are: `iterator` may adapt a move-only iterator, but sentinels are always semiregular and therefore copyable.

source/ranges.tex
```diff
@@ -6517,28 +6507,105 @@
 Equivalent to: \tcode{return x.\exposid{current_} - y.\exposid{current_};}
 \end{itemdescr}
 
-\indexlibrarymember{operator-}{elements_view::iterator}%
+\rSec3[range.elements.sentinel]{Class template \tcode{elements_view::\exposid{sentinel}}}
+
+\indexlibraryglobal{elements_view::sentinel}%
+\begin{codeblock}
+namespace std::ranges {
+  template<class V, size_t N>
```
CaseyCarter: This should be:
```c++
  template<input_range V, size_t N>
    requires view<V> && @\placeholder{has-tuple-element}@<range_value_t<V>, N> &&
             @\placeholder{has-tuple-element}@<remove_reference_t<range_reference_t<V>>, N>
```
to match the *template-head* on the `elements_view` template. (Yes, we've consistently been getting this wrong for quite some time, #3750 fixes all pre-existing instances I'm aware of.)

source/ranges.tex
```diff
@@ -6138,6 +6146,7 @@
 
   private:
     template<bool> struct @\exposid{iterator}@;                     // \expos
+    template<bool> struct @\exposid{sentinel}@;                     // \expos
```
CaseyCarter: We've been adding cross-links to the definitions of these nested types in other ranges:
```suggestion
    // \ref{range.elements.sentinel}, class template \tcode{elements_view::\exposid{sentinel}}
    template<bool> struct @\exposid{sentinel}@;                     // \expos
```
(#3750 adds the similar cross-reference for `iterator`.)

source/ranges.tex
```diff
@@ -6124,11 +6124,19 @@
...
+    constexpr auto end()
+    { return @\exposid{sentinel}@<false>{ranges::end(@\exposid{base_}@)}; }
 
-    constexpr auto end() const requires @\placeholder{simple-view}@<V>
-    { return ranges::end(@\exposid{base_}@); }
+    constexpr auto end() requires common_range<V>
+    { return @\exposid{iterator}@<false>{ranges::end(@\exposid{base_}@)}; }
+
+    constexpr auto end() const
+      requires range<const V>
```
CaseyCarter: There seems to be no particular reason not to put this *requires-clause* on the previous line, which would be more consistent with the style of the other declarations in this class. (Ditto line 6138.)

source/ranges.tex
burblebee: This is a normative change - please open an issue with LWG (for this and all of #3750).

jensmaurer: I believe this fix is editorial: We attempt a redeclaration of the nested class `elements_view::sentinel`, and that only works if the template-head is the same. The code as presented won't compile, because we can't add members to a class later on. The correct requires-clause is shown with the declaration of elements_view, so it seems obvious what is meant here.

source/ranges.tex
```diff
@@ -6146,7 +6144,7 @@
 
   private:
     template<bool> struct @\exposid{iterator}@;                     // \expos
-    template<bool> struct @\exposid{sentinel}@;                     // \expos
+    template<bool> struct @\exposid{sentinel}@;                     // \expos{} (see \ref{range.elements.sentinel})
```
johelegp: @CaseyCarter's review shows the correct way to cross-reference.

source/ranges.tex
burblebee: I see, fixed.

source/ranges.tex
burblebee: We're (sadly) inconsistent in how we add references for non methods and one-liners, but I see that we've used the suggested referencing for \exposid{sentinel}s elsewhere, so I'll follow that convention here.

PR #3767, opened 2020-02-20
CWG Poll 16: P2002R1 Defaulted comparison specification cleanups

### LINKED ISSUE #3300: make-unsigned-like and make-unsigned-like-t are too similar
[ranges.syn]/1 defines an exposition only function _`make-unsigned-like`_ and an exposition-only metafunction _`make-unsigned-like-t`_. These two names are hard to distinguish, doubly so with an argument list:

>  _`make-unsigned-like(meow)`_ vs. _`make-unsigned-like-t(woof)`_. 

It would help somewhat if _`make-unsigned-like-t`_ were treated as an exposition-only alias template:

>  _`make-unsigned-like(meow)`_ vs. _`make-unsigned-like-t<woof>`_ 

but ideally the names themselves should be more distinct as well. [@jwakely suggests](https://github.com/cplusplus/draft/pull/3299/files/ad046bfccbb3d34dac34ba73ab5645fd73ba614b#r335210466) renaming _`make-unsigned-like`_ to _`as-unsigned-like`_ or _`to-unsigned-like`_, which preserves the intended similarity between _`make-unsigned-like-t`_ and `make_unsigned_t`.

(As discussed in https://github.com/cplusplus/draft/pull/3299#discussion_r335187689).

### LINKED ISSUE #3700: [2020-02 CWG Poll 16] P2002R1 Defaulted comparison specification cleanups

burblebee:
[class.compare.default] Added "was" to "no overload resolution performed".

Fixes #3700. 
Fixes cplusplus/papers#739.

Notes/Issues:
* [everywhere] Please check my \defns and \indextexts - I wasn't sure how I should specify them.
* [class.compare.default] We just say "operator@" not "operator@ function".  We're also inconsistent in whether we encode is as "operator \tcode{@}" v. "\tcode{operator@}".
* [class.compare.default] We usually say "defaulted comparison operator function" instead of "defaulted comparison function".
* [class.compare.default] In the last 2 bullets of the new paragraph after p2, the "for" at the end of each bullet is hard to parse correctly: "
    ** for an operator<=> whose return type is auto or for an operator==, for a comparison between an element of the expanded list of subobjects and itself, or
    ** for a secondary comparison operator @, for the expression x @ y."

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -1219,6 +1219,13 @@ For a class, its non-static data members, its non-virtual direct base classes,
 and, if the class is not abstract\iref{class.abstract}, its virtual base
 classes are called its \term{potentially constructed subobjects}.
 
+\pnum
+A defaulted special member function is
+\indextext{member function!constexpr-compatible}%
+\defnx{constexpr-compatible}{constexpr-compatible!defaulted special member function}
+if the corresponding implicitly-declared special member function
+would be a constexpr function.
+
 \rSec2[class.ctor]{Constructors}%
 \indextext{constructor}%
 \indextext{special member function|see{constructor}}%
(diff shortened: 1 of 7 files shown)


source/classes.tex
```diff
@@ -1212,6 +1212,13 @@
 and, if the class is not abstract\iref{class.abstract}, its virtual base
 classes are called its \term{potentially constructed subobjects}.
 
+\pnum
+A defaulted special member function is
+\indextext{function!member!special!defaulted!constexpr-compatible}%
```
jensmaurer: This gives a five-level deep index entry. We don't want that.
This "function index" area needs a cleanup, but for the time being, please just use "member function!constexpr-compatible".

source/classes.tex
```diff
@@ -6660,53 +6667,142 @@
...
 a friend of \tcode{C} having either
 two parameters of type \tcode{const C\&} or
 two parameters of type \tcode{C}.
 \end{itemize}
-
-\pnum
-A defaulted comparison operator function for class \tcode{C}
+A defaulted comparison operator function for class \tcode{C} that
+is not defined as deleted is
+\indextext{operator!comparison!defaulted!implicitly defined}%
```
jensmaurer: Please check the existing index entries and find a suitable place. This one isn't.
Maybe "operator function!comparison!implicitly-defined".

source/classes.tex
```diff
@@ -6660,53 +6667,142 @@
...
+\indextext{operator!\idxcode{<=>}!deleted}%
+\indextext{operator!\idxcode{==}!deleted}%
 is defined as deleted if
 any non-static data member of \tcode{C} is of reference type or
-\tcode{C} is a union-like class\iref{class.union.anon}.
+\tcode{C} has variant members\iref{class.union.anon}.
+
+\pnum
+A binary operator expression \tcode{a @ b} is
+\indextext{operator!\idxcode{@}!usable}%
```
jensmaurer: drop the \idxcode{@}; it doesn't help in the index entry.

source/classes.tex
```diff
@@ -6660,53 +6667,142 @@
...
+neither \tcode{a} nor \tcode{b} is of class or enumeration type and
+\tcode{a @ b} is a valid expression.
+\end{itemize}
 
 \pnum
-If the class definition
-does not explicitly declare an \tcode{==} operator function,
-but declares a defaulted three-way comparison operator function,
+A defaulted comparison function is
+\indextext{operator!comparison!defaulted!constexpr-compatible}%
```
jensmaurer: Just "operator function!constexpr-compatible" for now

source/classes.tex
```diff
@@ -6817,43 +6903,45 @@
...
+let $\tcode{x}_i$ be the elements of
+the expanded list of subobjects for
+an object \tcode{x} of type \tcode{C}.
+
 \begin{itemize}
 \item
-If \tcode{R} is \tcode{auto},
-then the return type is deduced as
+If \tcode{R} is \tcode{auto}, then
+let $\tcode{\cv{}}_i$ $\tcode{R}_i$ be
```
jensmaurer: This should probably go into a single math env. (Avoiding a line break in between sounds like a good idea.)

source/classes.tex
```diff
@@ -6817,43 +6903,45 @@
...
+the type of the expression $\tcode{x}_i$ \tcode{<=>} $\tcode{x}_i$.
```
jensmaurer: same here: one math env

source/classes.tex
```diff
@@ -6901,12 +6984,15 @@
...
 
-\rSec2[class.rel]{Relational operators}
+\rSec2[class.compare.secondary]{Secondary comparison operators}
 \indextext{operator!relational!defaulted}%
 
 \pnum
-A defaulted relational operator function\iref{over.binary}
-for some operator \tcode{@}
+\indextext{operator!comparison!secondary}%
+A \defn{secondary comparison operator} is
```
jensmaurer: drop the \indextext and just say \defnadj{secondary}{comparison operator}

source/classes.tex
burblebee: For all my index entries, I do scan the existing ones and to try and find a suitable place.  As I said in my "Notes/issues", it isn't clear how we should write index entries.  I don't like having so many levels either, but which part to leave out, and which parts to combine?

For this case,  we don't have "operator function" but we do have "operator", so I'll just remove the "defaulted" level.  Thanks!

source/classes.tex
burblebee: This is how we format operators in the index, e.g.:
```
\indextext{operator!\idxcode{+=}}%
\indextext{operator!\idxcode{<=>}!deleted}%
\indextext{\idxcode{++}|see{operator, increment}}%
```

source/classes.tex
burblebee: Went with `\indextext{operator!constexpr-compatible}%` (there is no entry for ""operator function").

source/classes.tex
burblebee: I went with `\defnadj{secondary}{operator!comparison}` since we don't have a top level index entry for "comparison operator".

jensmaurer: I don't think "!" renders correctly in the \defnadj. You need to do a manual \defnx if you want to get clever.

PR #3769, opened 2020-02-20
LWG Poll 12: P2102R0 Make 'implicit expression variations' more explicit

### LINKED ISSUE #3714: [2020-02 LWG Poll 12] P2102R0 Make 'implicit expression variations' more explicit (Wording for US185)
resolves US185

http://wiki.edg.com/pub/Wg21prague/StrawPolls/p2102r0.html

jensmaurer:
Also fixes NB US 185 (C++20 CD).

Fixes #3714.
Fixes cplusplus/nbballot#183.

### DIFF
--- a/source/concepts.tex
+++ b/source/concepts.tex
@@ -29,6 +29,21 @@ An expression is \defnx{equality-preserving}{expression!equality-preserving} if,
 given equal inputs, the expression results in equal outputs. The inputs to an
 expression are the set of the expression's operands. The output of an expression
 is the expression's result and all operands modified by the expression.
+For the purposes of this subclause,
+the operands of an expression are the largest subexpressions that include only:
+\begin{itemize}
+\item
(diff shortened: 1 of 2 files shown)


source/concepts.tex
```diff
@@ -29,6 +29,21 @@
 given equal inputs, the expression results in equal outputs. The inputs to an
 expression are the set of the expression's operands. The output of an expression
 is the expression's result and all operands modified by the expression.
+For the purposes of this subclause,
+the operands of an expression are the largest subexpressions that include only:
```
jwakely: Was it intentional that _operands_ is not formatted as a definition?

jensmaurer: Thinking about it now, we shouldn't make "operand" an italics definition here, because it means something different than the core-language "operand" of expressions. And since it's a local definition only, we're probably ok as-is.

PR #3752, opened 2020-02-19
LWG Poll 18: P2091R0 Issues with range access CPOs

burblebee:
Fixes #3720.
Fixes cplusplus/nbballot#271.

PR #3771, opened 2020-02-20
LWG Poll 26: P1970R2 Consistency for size() functions: Add ranges::ssize

### LINKED ISSUE #3728: [2020-02 LWG Poll 26] P1970R2 Consistency for size() functions: Add ranges::ssize
resolves DE269

burblebee:
Also fixes NB DE 269 (C++20 CD).

Fixes #3728.
Fixes cplusplus/papers#703.
Fixes cplusplus/nbballot#265.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -43,6 +43,7 @@ namespace std::ranges {
     inline constexpr @\unspec@ crend = @\unspec@;
 
     inline constexpr @\unspec@ size = @\unspec@;
+    inline constexpr @\unspec@ ssize = @\unspec@;
     inline constexpr @\unspec@ empty = @\unspec@;
     inline constexpr @\unspec@ data = @\unspec@;
     inline constexpr @\unspec@ cdata = @\unspec@;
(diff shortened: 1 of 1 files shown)


source/ranges.tex
```diff
@@ -656,6 +657,29 @@
 type is integer-like.
 \end{note}
 
+\rSec2[range.prim.ssize]{\tcode{ranges::ssize}}
+
+\pnum
+The name \tcode{ssize} denotes
```
jwakely: There should be an index entry for ranges::ssize (it looks like nothing in this subclause is indexed though, so maybe fix them all together).

CaseyCarter: I think this should be `The name \tcode{ranges::ssize} denotes` to be consistent with other CPO wording.

source/ranges.tex
```diff
@@ -656,6 +657,29 @@
...
+a customization point object\iref{customization.point.object}.
+The expression \tcode{ranges::ssize(E)}
+for a subexpression \tcode{E} of type \tcode{T}
+is expression-equivalent to:
+
+\begin{itemize}
+\item
+If \tcode{range_difference_t<T>} has width less than \tcode{ptrdiff_t}:
+\begin{codeblock}
+static_cast<ptrdiff_t>(ranges::size(E)); // \ref{range.prim.size}
```
CaseyCarter: Semicolons seem odd here given that this is intended to be an expression rather than a statement. I also think the cross-reference to the immediately preceding subclause adds no value. `If \tcode{range_difference_t<T>} has width less than \tcode{ptrdiff_t}, \tcode{static_cast<ptrdiff_t>(ranges::size(E))}.` would seem to be an improvement. (Ditto for the following case.)

johelegp:
Seems like this wording didn't get the same treatment as the sibling CPOs in https://github.com/cplusplus/draft/pull/3752/files. Is that ok?

source/ranges.tex
burblebee: @jwakely I added one for ranges::ssize, please open an edit issue for the rest.

burblebee: I fixed this one, but there are several other cases which should be fixed as well - please open an edit issue for them:
```
The name \tcode{size} denotes a customization point
The name \tcode{empty} denotes a customization point
The name \tcode{data} denotes a customization point
The name \tcode{cdata} denotes a customization point
```

burblebee:
> Seems like this wording didn't get the same treatment as the sibling CPOs in https://github.com/cplusplus/draft/pull/3752/files. Is that ok?

@johelegp please send a note to lwgchair about this.

source/ranges.tex
burblebee: done as a separate commit

source/ranges.tex
CaseyCarter: #3752 fixes `size`, `empty`, and `data`; I've submitted #3793 to fix `cdata`.

PR #3772, opened 2020-02-20
LWG Poll 16: P1976R2 Fixed-size span construction from dynamic range

### LINKED ISSUE #3718: [2020-02 LWG Poll 16] P1976R2 Fixed-size span construction from dynamic range
resolves PL250

jwakely:
Also fixes NB PL 250  (C++20 CD).

[span.sub] p3 and p6: Add explicit return types to account for
constructor being explicit now.

[span.objectrep] p1 and p3: Move extent of return type into a remark.
Add explicit return type to account for explicit constructor.

Fixes #3718
Fixes cplusplus/papers#707
Fixes cplusplus/nbballot#246

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -11024,7 +11024,8 @@ template<size_t Count> constexpr span<element_type, Count> first() const;
 
 \pnum
 \effects
-Equivalent to: \tcode{return \{data(), Count\};}
+Equivalent to: \tcode{return R\{data(), Count\};}
+where \tcode{R} is the return type.
 \end{itemdescr}
 
 \indexlibrarymember{span}{last}%
@@ -11043,7 +11044,8 @@ template<size_t Count> constexpr span<element_type, Count> last() const;
 
 \pnum
 \effects
-Equivalent to: \tcode{return \{data() + (size() - Count), Count\};}
+Equivalent to: \tcode{return R\{data() + (size() - Count), Count\};}
+where \tcode{R} is the return type.
 \end{itemdescr}
 
 \indexlibrarymember{span}{subspan}%
@@ -11309,7 +11311,8 @@ template<class ElementType, size_t Extent>
 \begin{itemdescr}
 \pnum
 \effects
-Equivalent to: \tcode{return \{reinterpret_cast<const byte*>(s.data()), s.size_bytes()\};}
+Equivalent to: \tcode{return R\{reinterpret_cast<const byte*>(s.data()), s.size_bytes()\};}
+where \tcode{R} is the return type.
 \end{itemdescr}
 
 \indexlibraryglobal{as_writable_bytes}%
@@ -11326,7 +11329,8 @@ template<class ElementType, size_t Extent>
 
 \pnum
 \effects
-Equivalent to: \tcode{return \{reinterpret_cast<byte*>(s.data()), s.size_bytes()\};}
+Equivalent to: \tcode{return R\{reinterpret_cast<byte*>(s.data()), s.size_bytes()\};}
+where \tcode{R} is the return type.
 \end{itemdescr}
 
 \rSec3[span.tuple]{Tuple interface}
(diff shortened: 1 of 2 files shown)


source/containers.tex
```diff
@@ -11236,20 +11260,29 @@
...
+Equivalent to:
+\begin{codeblock}
+return span<const byte, @\seebelow@>{reinterpret_cast<const byte*>(s.data()), s.size_bytes()};
+\end{codeblock}
+
+\pnum
+\remarks The second template argument of the returned \tcode{span} type is:
+\begin{codeblock}
+Extent == dynamic_extent ? dynamic_extent : sizeof(ElementType) * Extent
+\end{codeblock}
```
CaseyCarter: This feels awkward to me given that we have an actual mechanism in the language for "I don't want to write this messy return type here". Did you consider:
* Change the return type of `as_bytes` to `auto`, 
* "Let `U` be `span<const byte, Extent == dynamic_extent ? dynamic_extent : sizeof(ElementType) * Extent>`.", and
* `return U{/*...*/};`?
(Ditto `as_writable_bytes`.)

jwakely: I did consider introducing a type (or the extent) with "Let" but for some reason I didn't consider using `auto`. The form I used is what Tomasz (paper author) proposed in a pending-LWG issue.

I'll use your suggestion and push a new commit.

jwakely: After trying a few ways to say it, I think the clearest is:

> _Effects:_ Equivalent to:
````
auto p = reinterpret_cast<const byte*>(s.data());
auto n = s.size_bytes();
if constexpr (Extent == dynamic_extent)
  return span<const byte>{p, n};
else
  return span<const byte, sizeof(ElementType) * Extent>{p, n};
````

**Comment by tomaszkam** [comment]
I used the wording to be consistient with [span.sub]p9. However, I do not have any preference for one or another.

File: source/containers.tex [hunk 2]
```diff
@@ -11236,20 +11260,29 @@
 \indexlibraryglobal{as_bytes}%
 \begin{itemdecl}
 template<class ElementType, size_t Extent>
-  span<const byte, Extent == dynamic_extent ? dynamic_extent : sizeof(ElementType) * Extent>
+  span<const byte, @\seebelow@>
```
johelegp: These changes are missing in the synopsis.

source/containers.tex
CaseyCarter: Works for me.

CaseyCarter: I think we could factor out some more commonality with:

> Let `E` be `dynamic_extent` if `Extent == dynamic_extent`, and `sizeof(ElementType) * Extent` otherwise.
>
> _Effects_: Equivalent to: `return span<const byte, E>{reinterpret_cast<const byte*>(s.data()), s.size_bytes()};`

but this is much better in any case - feel free to ignore if you have better things to do than twaddle with this bit of wording over and over.

source/containers.tex
jwakely: Not really, since the synopsis just spells out `Extent == dynamic_extent ? dynamic_extent : sizeof(ElementType) * Extent` in full, which is exactly what the _see below_ expands to.

I pushed a new commit that does it differently anyway.

jensmaurer:
Please address the Travis failures. You can run "../tools/check.sh" in a Unix-y environment to see the errors right there.

jwakely:
I don't see any errors when I build the PDF and run the check.sh script locally. :man_shrugging:

tkoeppe:
@jwakely: Did you check the return value? I think you're falling afoul of https://github.com/cplusplus/draft/blob/master/tools/check.sh#L35, but the check doesn't emit a terribly useful diagnostic.

jensmaurer:
@jwakely , @tkoeppe is right. \expects and friends should be alone on a line.

jwakely:
Thanks, should be fixed now.

zygoloid:
Whether a function has a deduced return type is an observable part of its semantics; I'm not convinced we can change that editorially.

PR #3775, opened 2020-02-20
Simplify build process for the standard draft

zygoloid:
Remove all build methods other than `latexmk`. Make checking for out-of-date `.pdf` figures actually work and move it to `check.sh`. Simplify Travis CI to only run the single remaining sensible build configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 #
-# Builds the C++ standard document on Travis CI <https://travis-ci.org/cplusplus/draft>
+# Builds the C++ standard document on Travis CI <https://travis-ci.com/cplusplus/draft>
 #
 
 dist: trusty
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -14,43 +14,8 @@ before_install:
   - docker run -itd -v $TRAVIS_BUILD_DIR:/$TRAVIS_REPO_SLUG --name texlive-basic godbyk/texlive-basic
 
 env:
-  - BUILD_TYPE=latexmk   # build using latexmk, also apply all checks
-  - BUILD_TYPE=make      # build using Makefile
-  - BUILD_TYPE=complete  # build manually and regenerate figures, grammar, and cross-references
+  - BUILD_TYPE=normal    # build using latexmk, also apply all checks
```
tkoeppe: I think you can omit the entire `env` block if you have nothing useful to set.

PR #3776, opened 2020-02-20
LWG Poll 20: P1831R1 Deprecating volatile: library

### LINKED ISSUE #3722: [2020-02 LWG Poll 20] P1831R1 Deprecating volatile: library
resolves CZ004, CA210, US211

http://wiki.edg.com/pub/Wg21prague/StrawPolls/P1831R1.html

jensmaurer:
Also fixes NB CZ 004, CA 210, and US 211 (C++20 CD).

- atomic_init was already deprecated, with and without volatile

Fixes #3722 .
Fixes cplusplus/papers#589.
Fixes cplusplus/nbballot#4
Fixes cplusplus/nbballot#208.

### DIFF
--- a/source/atomics.tex
+++ b/source/atomics.tex
@@ -1719,7 +1719,7 @@ The \tcode{order} argument is neither \tcode{memory_order::consume},
 \pnum
 \constraints
 For the \tcode{volatile} overload of this function,
-\tcode{atomic<T>::is_always_lock_free} is \tcode{true}.
+\tcode{is_always_lock_free} is \tcode{true}.
 
 \pnum
 \effects
@@ -1741,7 +1741,7 @@ T operator=(T desired) noexcept;
 \pnum
 \constraints
 For the \tcode{volatile} overload of this function,
-\tcode{atomic<T>::is_always_lock_free} is \tcode{true}.
+\tcode{is_always_lock_free} is \tcode{true}.
 
 \pnum
 \effects
@@ -1771,7 +1771,7 @@ The \tcode{order} argument is neither \tcode{memory_order::release} nor \tcode{m
 \pnum
 \constraints
 For the \tcode{volatile} overload of this function,
-\tcode{atomic<T>::is_always_lock_free} is \tcode{true}.
+\tcode{is_always_lock_free} is \tcode{true}.
 
 \pnum
 \effects
@@ -1795,7 +1795,7 @@ operator T() const noexcept;
 \pnum
 \constraints
 For the \tcode{volatile} overload of this function,
-\tcode{atomic<T>::is_always_lock_free} is \tcode{true}.
+\tcode{is_always_lock_free} is \tcode{true}.
 
 \pnum
 \effects
@@ -1818,7 +1818,7 @@ T exchange(T desired, memory_order order = memory_order::seq_cst) noexcept;
 \pnum
 \constraints
 For the \tcode{volatile} overload of this function,
-\tcode{atomic<T>::is_always_lock_free} is \tcode{true}.
+\tcode{is_always_lock_free} is \tcode{true}.
 
 \pnum
 \effects
@@ -1872,7 +1872,7 @@ The \tcode{failure} argument is neither \tcode{memory_order::release} nor
 \pnum
 \constraints
 For the \tcode{volatile} overload of this function,
-\tcode{atomic<T>::is_always_lock_free} is \tcode{true}.
+\tcode{is_always_lock_free} is \tcode{true}.
 
 \pnum
 \effects
@@ -2260,7 +2260,7 @@ T fetch_@\placeholdernc{key}@(T operand, memory_order order = memory_order::seq_
 \pnum
 \constraints
 For the \tcode{volatile} overload of this function,
-\tcode{atomic<T>::is_always_lock_free} is \tcode{true}.
+\tcode{is_always_lock_free} is \tcode{true}.
 
 \pnum
 \effects
@@ -2304,7 +2304,7 @@ T operator @\placeholder{op}@=(T operand) noexcept;
 \pnum
 \constraints
 For the \tcode{volatile} overload of this function,
-\tcode{atomic<T>::is_always_lock_free} is \tcode{true}.
+\tcode{is_always_lock_free} is \tcode{true}.
 
 \pnum
 \effects
@@ -2423,7 +2423,7 @@ T A::fetch_@\placeholdernc{key}@(T operand, memory_order order = memory_order_se
 \pnum
 \constraints
 For the \tcode{volatile} overload of this function,
-\tcode{atomic<T>::is_always_lock_free} is \tcode{true}.
+\tcode{is_always_lock_free} is \tcode{true}.
 
 \pnum
 \effects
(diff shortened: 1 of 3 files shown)


source/future.tex
```diff
@@ -1459,6 +1459,123 @@
...
+\begin{itemdescr}
+\pnum
+Let \tcode{TE} denote \tcode{tuple_element_t<I, T>}
+of the cv-unqualified type \tcode{T}.
+Then each of the two templates shall satisfy
+the \tcode{TransformationTrait} requirements
+with a member typedef type that names the following type:
+\begin{itemize}
+\item for the first specialization, add_volatile_t<TE>, and
+\item for the second specialization, add_cv_t<TE>.
```
burblebee: \tcode{}s needed

source/future.tex
```diff
@@ -1459,6 +1459,123 @@
...
+template<size_t I, class T> class tuple_element<I, const volatile T>;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+Let \tcode{TE} denote \tcode{tuple_element_t<I, T>}
+of the cv-unqualified type \tcode{T}.
+Then each of the two templates shall satisfy
+the \tcode{TransformationTrait} requirements
+with a member typedef type that names the following type:
```
burblebee: Bad wording - how can a member typedef satisfy requirements for a template?

source/future.tex
```diff
@@ -1459,6 +1459,123 @@
 \end{note}
 \end{itemdescr}
 
+\rSec1[depr.tuple]{Tuple}
+
+Header \libheaderref{tuple}:
+
+\begin{codeblock}
+namespace std {
+  // \ref{tuple.helper}, tuple helper classes
```
burblebee: This reference should be to the deprecated \itemdecls which are now in this section.  Either add a [depr.tuple.helper] as the paper suggests and use that reference here or remove the reference.

I'd suggest just removing the reference since the section is so small.

source/future.tex
```diff
@@ -1459,6 +1459,123 @@
...
+\libheaderref{ranges}, or
+\libheaderref{utility}
+are included.
+\end{itemdescr}
+
+\rSec1[depr.variant]{Variant}
+
+\begin{codeblock}
+namespace std {
+  // \ref{variant.helper}, variant helper classes
```
burblebee: Fix/remove reference

source/future.tex
```diff
@@ -1459,6 +1459,123 @@
...
+template<class T> class variant_size<const volatile T>;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+Let \tcode{VS} denote \tcode{variant_size<T>}
+of the cv-unqualified type \tcode{T}.
+Then each of the two templates shall satisfy
+the \tcode{UnaryTypeTrait} requirements
+with a base characteristic of \tcode{integral_constant<size_t, VS::value>}.
```
burblebee: Bad wording

source/future.tex
```diff
@@ -1459,6 +1459,123 @@
...
+template<size_t I, class T> class variant_alternative<I, const volatile T>;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+Let \tcode{VA} denote \tcode{variant_alternative<I, T>}
+of the cv-unqualified type \tcode{T}.
+Then each of the two templates shall meet
+the \tcode{TransformationTrait} requirements
+with a member typedef type that names the following type:
```
burblebee: Bad wording

source/future.tex
```diff
@@ -2477,6 +2594,30 @@
...
+bool compare_exchange_strong(T& expected, T desired,
+                             memory_order order = memory_order::seq_cst) volatile noexcept;
+T fetch_@\placeholdernc{key}@(T operand, memory_order order = memory_order::seq_cst) volatile noexcept;
+T operator @\placeholdernc{op}@=(T operand) volatile noexcept;
+T* fetch_@\placeholdernc{key}@(ptrdiff_t operand, memory_order order = memory_order::seq_cst) volatile noexcept;
+\end{codeblock}
+
 \rSec2[depr.atomics.nonmembers]{Non-member functions}
 
 \indexlibraryglobal{atomic_init}%
```
burblebee: The second part of this change to [depr.atomics] is missing:
"The following non-member function is available when atomic<T>::is_always_lock_free is false:"

I realize atomic_init was already deprecated, so maybe change the wording to say "For the volatile overload of this function, ..."

source/atomics.tex
```diff
@@ -1716,6 +1716,11 @@
 The \tcode{order} argument is neither \tcode{memory_order::consume},
 \tcode{memory_order::acquire}, nor \tcode{memory_order::acq_rel}.
 
+\pnum
+\constraints
+For the \tcode{volatile} overload of this function,
+\tcode{atomic<T>::is_always_lock_free} is true.
```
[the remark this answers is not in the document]
jensmaurer: Fixed.

source/future.tex
jensmaurer: Fixed.

source/future.tex
jensmaurer: Fixed.

source/future.tex
jensmaurer: Fixed.

source/future.tex
jensmaurer: see above

source/future.tex
jensmaurer: see above

source/future.tex
jensmaurer: Here's the problem: The entirety of atomic_init is deprecated and the "volatile" part is double-deprecated. So the "If" reduces to "If true".

jensmaurer:
Regarding atomic_init, I think this is correct, because it's now all in  [depr.atomics] where this paper also says "The following non-member function is available when atomic<T>::is_always_lock_free is false:" So, this seems to fold away to "if true" or "constraint: true".

atomics.types.pointer is fixed.

source/future.tex
jensmaurer: This is a usual pattern when using these type traits. The idea is that a "Transformation Trait" requires a member typedef, and that member typedef is then required to name "the following type".

jensmaurer:
@zygoloid, [atomics.types.pointer] says now "Constraints: For the `volatile` overload of this function,
`atomic<T>::is_always_lock_free` is true." But this is in the partial specialization for "pointer to T", and we likely don't want to talk about T (the pointee) here, but about "pointer to T".
Fixing this doesn't look editorial to me, but maybe you can squint the right way.

source/future.tex
```diff
@@ -1459,6 +1459,123 @@
...
+template<class T> class tuple_size<const volatile T>;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+Let \tcode{TS} denote \tcode{tuple_size<T>} of the cv-unqualified type \tcode{T}.
+If the expression \tcode{TS::value} is well-formed
+when treated as an unevaluated operand,
+then each of the two templates shall satisfy
+the \tcode{TransformationTrait} requirements with a base characteristic of
```
CaseyCarter:
```suggestion
the \oldconcept{TransformationTrait} requirements with a base characteristic of
```
(Throughout.)

jensmaurer: Fixed.

PR #3777, opened 2020-02-21
LWG Poll 24: P1739R4 Avoid template bloat for safe_ranges in combination with "sub…

### LINKED ISSUE #3726: [2020-02 LWG Poll 24] P1739R4 Avoid template bloat for safe_ranges in combination with "subrange-y" view adaptors
resolves DE288, US272

burblebee:
…range-y" view adaptors

Also fixes NB DE 288 and US 272 (C++20 CD).

[range.counted] Moved note to work with new wording.

Fixes #3726.
Fixes cplusplus/papers#503.
Fixes cplusplus/nbballot#284.
Fixes cplusplus/nbballot#268.

Notes:
* [ranges] FYI - fixed capitalization.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -2038,6 +2038,7 @@ namespace std::ranges {
     constexpr explicit iota_view(W value);
     constexpr iota_view(type_identity_t<W> value,
                         type_identity_t<Bound> bound);
+    constexpr iota_view(@\exposid{iterator}@ first, @\exposid{sentinel}@ last) : iota_view(*first, last.@\exposid{bound_}@) {}
 
     constexpr @\exposid{iterator}@ begin() const;
     constexpr auto end() const;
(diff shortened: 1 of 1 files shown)


source/ranges.tex
```diff
@@ -4091,9 +4092,37 @@
...
+is expression-equivalent to:
+
+\begin{itemize}
+\item
+If \tcode{T} is a specialization
+of \tcode{ranges::empty_view}\iref{range.empty.view},
+then \tcode{((void) F, \placeholdernc{decay-copy}(E))};
+
+\item
+otherwise, if \tcode{T} models
```
jensmaurer: another "models"; see http://lists.isocpp.org/lib/2020/02/15452.php

(similarly for the other occurrences below)

This seems to make the entire paper unimplementable, and it thus should be rejected.

jensmaurer:
@zygoloid , this paper incorrectly uses "models" pervasively, which makes it essentially unimplementable. I've emailed LWG, but I'm not sure if anything here can be resurrected editorially.

jensmaurer:
P2101R0 (applied with LWG poll 10) adds "If the validity or meaning of a program depends on whether a sequence of template arguments models a concept, and the concept is satisfied but not modeled, the program is ill-formed, no diagnostic required."
Thus the phrasing "If X does not model Y, the program is ill-formed" sort-of makes sense, although I find the outcome to be bad in several respects.

zygoloid:
> Thus the phrasing "If X does not model Y, the program is ill-formed" sort-of makes sense, although I find the outcome to be bad in several respects.

I encountered this when editing in a different paper. I find the current approach quite problematic from a comprehensibility standpoint, and I said this in person at Prague too, but it's too late in the cycle to fix this for C++20 (would require someone going through all the library wording and fixing all the places that say "models" when they mean to apply a machine-checkable set of requirements).

PR #3779, opened 2020-02-21
LWG Poll 19: P0586R2 Safe integral comparisons

### LINKED ISSUE #3721: [2020-02 LWG Poll 19] P0586R2 Safe integral comparisons
with the change of the name of the feature test macro from __cpp_lib_cmp_equal to __cpp_lib_integer_comparison_functions. 

resolves DE208

burblebee:
Also fixes NB DE 208 (C++20 CD).

Fixes #3721. 
Fixes cplusplus/papers#259.
Fixes cplusplus/nbballot#206.

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -618,6 +618,7 @@ the values of these macros with greater values.
 #define @\defnlibxname{cpp_lib_incomplete_container_elements}@     201505L
   // also in \libheader{forward_list}, \libheader{list}, \libheader{vector}
 #define @\defnlibxname{cpp_lib_int_pow2}@                          202002L // also in \libheader{bit}
+#define @\defnlibxname{cpp_lib_integer_comparison_functions}@      202002L // also in \libheader{utility}
 #define @\defnlibxname{cpp_lib_integer_sequence}@                  201304L // also in \libheader{utility}
 #define @\defnlibxname{cpp_lib_integral_constant_callable}@        201304L // also in \libheader{type_traits}
 #define @\defnlibxname{cpp_lib_interpolate}@                       201902L // also in \libheader{cmath}, \libheader{numeric}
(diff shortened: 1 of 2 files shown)


source/utilities.tex
```diff
@@ -75,6 +75,25 @@
   // \ref{declval}, declval
   template<class T>
     add_rvalue_reference_t<T> declval() noexcept;   // as unevaluated operand
+
+  // \ref{utility.safeintcmp}, Safe integral comparisons
```
```suggestion
  // \ref{utility.safeintcmp}, safe integral comparisons
```

source/support.tex
```diff
@@ -583,6 +583,7 @@
 #define @\defnlibxname{cpp_lib_chrono}@                            201907L // also in \libheader{chrono}
 #define @\defnlibxname{cpp_lib_chrono_udls}@                       201304L // also in \libheader{chrono}
 #define @\defnlibxname{cpp_lib_clamp}@                             201603L // also in \libheader{algorithm}
+#define @\defnlibxname{cpp_lib_cmp_equal}@                         202002L // also in \libheader{utility}
```
timsong-cpp: Per the motion, the feature-test macro should be `__cpp_lib_integer_comparison_functions`.

source/utilities.tex
timsong-cpp: We should not use "safe".

jwakely: And we should not use "integral" either.

I suggest "integer comparison functions" instead.

burblebee: changed to "integer comparison functions"

source/support.tex
burblebee: done

source/utilities.tex
```diff
@@ -75,6 +75,25 @@
   // \ref{declval}, declval
   template<class T>
     add_rvalue_reference_t<T> declval() noexcept;   // as unevaluated operand
+
+  // \ref{utility.safeintcmp}, integer comparison functions
```
timsong-cpp: Change the stable name too?

burblebee: renamed to utility.intcmp

PR #3782, opened 2020-02-21
CWG Poll 1: P2103R0 (Core Language Changes for NB Comments at the February, 2020 (Prague) meeting

### LINKED ISSUE #3685: [2020-02 CWG Poll 1] P2103R0 (Core Language Changes for NB Comments at the February, 2020 (Prague) meeting
addressing US028, US033, US041, CA104, CA107, US115, and US117

http://wiki.edg.com/pub/Wg21prague/StrawPolls/p2103r0.html

zygoloid:
Fixes NB US028, US033, US041, CA104, CA107, US115, and US117 (C++20 CD).

Fixes #3685.
Also fixes cplusplus/nbballot#27.
Also fixes cplusplus/nbballot#32.
Also fixes cplusplus/nbballot#40.
Also fixes cplusplus/nbballot#103.
Also fixes cplusplus/nbballot#106.
Also fixes cplusplus/nbballot#114.
Also fixes cplusplus/nbballot#116.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3310,7 +3310,11 @@ at the storage location which the original object occupied, a pointer
 that pointed to the original object, a reference that referred to the
 original object, or the name of the original object will automatically
 refer to the new object and, once the lifetime of the new object has
-started, can be used to manipulate the new object, if:
+started, can be used to manipulate the new object, if
+the original object is transparently replaceable (see below)
+by the new object.
+An original object is \defn{transparently replaceable}
+by a new object if:
 \begin{itemize}
 \item the storage for the new object exactly overlays the storage
 location which the original object occupied, and
@@ -3323,7 +3327,14 @@ a complete object that is const-qualified nor
 a subobject of such an object, and
 
 \item neither the original object nor the new object
-is a potentially-overlapping subobject\iref{intro.object}.
+is a potentially-overlapping subobject\iref{intro.object}, and
+
+\item neither the original object nor the new object are subobjects
+or both are subobjects, and
+
+\item the (unique) object (if any) that contains the original object
+as a subobject\iref{intro.object} is transparently replaceable by
+the (unique) object (if any) that contains the new object.
 \end{itemize}
 \begin{example}
 \begin{codeblock}
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -2603,28 +2603,15 @@ of a sequence of declarations.
 
 \begin{bnf}
 \nontermdef{translation-unit}\br
-    \opt{top-level-declaration-seq}\br
-    \opt{global-module-fragment} module-declaration \opt{top-level-declaration-seq} \opt{private-module-fragment}
-\end{bnf}
-
-\begin{bnf}
-\nontermdef{top-level-declaration-seq}\br
-    top-level-declaration\br
-    top-level-declaration-seq top-level-declaration
-\end{bnf}
-
-\begin{bnf}
-\nontermdef{top-level-declaration}\br
-    module-import-declaration\br
-    declaration
+    \opt{declaration-seq}\br
+    \opt{global-module-fragment} module-declaration \opt{declaration-seq} \opt{private-module-fragment}
 \end{bnf}
 
 \pnum
 A token sequence beginning with
 \opt{\tcode{export}} \tcode{module}
 and not immediately followed by \tcode{::}
-is never interpreted as the \grammarterm{declaration}
-of a \grammarterm{top-level-declaration}.
+is never interpreted as a \grammarterm{declaration}.
 
 \pnum
 \indextext{linkage}%
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -261,12 +261,32 @@ name mangling and linking.
 \end{defnote}
 
 \indexdefn{signature}%
+\definition{signature}{defns.signature.friend}
+\defncontext{non-template friend function with trailing \grammarterm{requires-clause}}
+name,
+parameter-type-list\iref{dcl.fct},
+enclosing class,
+and
+trailing \grammarterm{requires-clause}\iref{dcl.decl}
+
+\indexdefn{signature}%
 \definition{signature}{defns.signature.templ}
 \defncontext{function template}
 name,
 parameter-type-list\iref{dcl.fct},
 enclosing namespace (if any),
 return type,
+\grammarterm{template-head},
+and
+trailing \grammarterm{requires-clause}\iref{dcl.decl} (if any)
+
+\indexdefn{signature}%
+\definition{signature}{defns.signature.templ.friend}
+\defncontext{friend function template with constraint involving enclosing template parameters}
+name,
+parameter-type-list\iref{dcl.fct},
+return type,
+enclosing class,
 \grammarterm{template-head},
 and
 trailing \grammarterm{requires-clause}\iref{dcl.decl} (if any)
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -1746,6 +1746,38 @@ and those of
 \tcode{f7} are \tcode{C2<T> $\land$ C1<T>}.
 \end{example}
 
+\pnum
+When determining whether a given introduced
+\grammarterm{constraint-expression} $C_1$ of a declaration
+in an instantiated specialization of a templated class
+is equivalent\iref{temp.over.link} to the corresponding
+\grammarterm{constraint-expression} $C_2$ of a declaration
+outside the class body,
+$C_1$ is instantiated.
+If the instantiation results in an invalid expression,
+the \grammarterm{constraint-expression}s are not equivalent.
+\begin{note}
+This can happen when determining which member template is specialized
+by an explicit specialization declaration.
+\end{note}
+\begin{example}
+\begin{codeblock}
+template <class T> concept C = true;
+template <class T> struct A {
+  template <class U> U f(U) requires C<typename T::type>;   // \#1
+  template <class U> U f(U) requires C<T>;                  // \#2
+};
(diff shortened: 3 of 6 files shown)


source/templates.tex
```diff
@@ -1612,13 +1613,68 @@
...
+template <unsigned M> void g(AddOneTy<2 * M> *);
+
+void h() {
+  g<0>(nullptr);    // OK, there is only one \tcode{g}
+}
+\end{codeblock}
+\end{example}
+This similarity includes the situation where a program is ill-formed, no diagnostic required,
+when the meaning of the program depends on whether two constructs are equivalent,
+and they are functionally equivalent but not equivalent.
```
burblebee: What is "they"?  The "two constructs"?  How can they be "equivalent" and "not
equivalent" at the same time?

hubert-reinterpretcast: That portion is a quote from [temp.over.link]/7. I am having a hard time finding another interpretation of "they" that would necessitate deviating from the formulation in [temp.over.link]. As for the latter question, they can (at the same time) be "functionally equivalent" and "not equivalent".

source/basic.tex
```diff
@@ -3336,7 +3327,14 @@
...
 \item neither the original object nor the new object
-is a potentially-overlapping subobject\iref{intro.object}.
+is a potentially-overlapping subobject\iref{intro.object}, and
+
+\item neither the original object nor the new object are subobjects
+or both are subobjects, and
+
+\item the (unique) object (if any) that contains the original object
+as a subobject\iref{intro.object} is transparently replaceable by
+the (unique) object (if any) that contains the new object.
```
hubert-reinterpretcast: The last should probably say "contains the new object as a subobject".

source/intro.tex
```diff
@@ -271,6 +280,17 @@
 and
 trailing \grammarterm{requires-clause}\iref{dcl.decl} (if any)
 
+\indexdefn{signature}%
+\definition{signature}{defns.signature.templ.friend}
+\defncontext{friend function template with constraint involving enclosing template parameters}
```
hubert-reinterpretcast: I believe this should say "one or more constraints".

source/basic.tex
zygoloid: Please consider #3796.

source/templates.tex
burblebee: This doesn't help me - I read "they" as the "two constraints", and this makes no sense.  Yes, entities can (at the same time) be "functionally equivalent" and "not equivalent". but here we're saying that they can be  "equivalent" and "not
equivalent" at the same time: "two constructs are equivalent,
and they are functionally equivalent but not equivalent."

jwakely: No, that's not what it says. It is talking about cases where the **program depends on whether they are equivalent**, but **they are not equivalent**. That's not "they are equivalent and not equivalent", it's "they need to be equivalent, but they are not".

burblebee: Ok, thanks to @jwakely, I finally get what this wording is trying to say, in which case the "and" should be changed to "but" at the very least, or preferably reworded so that it's not possible to parse it as I did.

hubert-reinterpretcast:
```suggestion
+but they are merely functionally equivalent and not equivalent.
```

burblebee: @hubert-reinterpretcast Much better, thank you!  Sadly, Richard has already merged this motion, so we'll have to create a separate PR for this and your suggestion below (for "one or more constraints").  Would you mind doing that since it's your wording?

hubert-reinterpretcast: On reading this again, I am having trouble with the change I suggested above. I put in a more extensive change in #3807.

source/templates.tex
```diff
@@ -3160,7 +3160,14 @@
...
 \pnum
-A non-template friend declaration shall not have a \grammarterm{requires-clause}.
+A non-template friend declaration
+with a \grammarterm{requires-clause}
+shall be a definition.
+A friend function template
+with a constraint that depends on a template parameter from an enclosing template
+shall be a definition.
+Such a constrained friend function or function template declaration
+does not declare the same function or function template as a declaration in any other scope.
```
dfrib, line 3170: How is "in any _other_ scope" to be interpreted here? [[class.friend]/7](https://eel.is/c++draft/class.friend#7) mentions lexical scope (of the class) for hidden friend definitions, but namespace scope, as per [[class.friend]/6](https://eel.is/c++draft/class.friend#6), for the function itself.

As per [[over.dcl]/1](https://eel.is/c++draft/over.dcl#1) two function declarations refer to the same function only if they have the same trailing requires-clauses (if any). This should arguably also apply for non-template hidden friend definitions which differs only in their trailing requires-clauses. The non-normative text of [over.dcl]/1 contains an example for this particular case but for non-template member functions as compared to hidden friends.

This seems to have a few compilers confused, at least, as GCC, Clang and MSVC all (using current HEAD) implement different behaviours for overloading hidden friends only by different trailing requires-clauses:

```
struct Base {};

template<int N>
struct S : public Base {
    friend int f(Base&) requires (N == 1) { return 1; }
    friend int f(Base&) requires (N == 2) { return 3; }
    // Separate functions (which overload resolution will
    // have a hard time to disambiguate)?
};

int main() {
    S<1> s1{};
    S<2> s2{};    // GCC error: redefinition 
    (void)f(s1);  // MSVC error: ambiguous call (correct?)
    (void)f(s2);
    // Clang error: definition with same mangled name
    //              '_Z3fooR4Base' as another definition
}
```

Is MSVC correct here?

Should I be filing compiler bug reports for this, or should the normative text possibly be updated to be more explicit regarding this special case?

hubert-reinterpretcast: I think MSVC is correct here. With respect to the Clang behaviour, the description of the signature indicates that the mangling should be unique. With respect to the GCC behaviour, the "does not declare" wording is meant to say that the friends declared by each specialization is unique.

dfrib: Hubert, thanks for the answer and the clarification; it's clear that Clang is wrong by violation of [[defns.signature.friend]](https://eel.is/c++draft/defns.signature.friend), and that GCC is wrong as per the intended meaning of [temp.friend]/9.

However, I'm wondering what is the actual correct behaviour is here; is MSVC really correct, or is "correct" even well-specified?

I'm having a hard time finding the normative reference for constraints checking when finding _viable_ candidates in overload resolution for the case of overloaded, constrained hidden friends, where the signatures solely differ by means of the their (mutually exclusive) trailing requires-clauses. Particularly, whether and how constraints checking will be performed during overload resolution on a free (friend) non-template function.

Starting with lookup, [[basic.lookup.argdep]/4.3](https://eel.is/c++draft/basic.lookup.argdep#4.3) states that any (namespace-scope) friend function
> [...] declared in classes with reachable definitions in the set of associated entities are visible [...]

can be found via ADL on associated entities. Thus, as per [[basic.lookup.argdep]/2.2](https://eel.is/c++draft/basic.lookup.argdep#2.2), for the calls `f(s1)` and the `f(s2)` in the example above the associated entity is the class template itself, even if the arguments are objects of particular specializations of the class template. This would arguably mean that ADL, for both calls `f(s1)` and `f(s2)`, can find both overloads as candidate functions? So far, this is the interpretation MSVC has done, as it proceeds (and subsequently fails) with overload resolution.

For finding viable functions during overload resolution, [[over.match.viable]/3](https://eel.is/c++draft/over.match.viable#3) applies:
> Second, for a function to be viable, if it has associated constraints ([temp.constr.decl]), those constraints shall be satisfied ([temp.constr.constr]).

but the case of hidden friends with associated constraints is special in this regard. In say the call `f(s1)`, how would constraint satisfaction be checked? At this point, given that ADL has actually found both candidates (as is the interpretation of MSVC), the constraint is in no apparent way connected to the particular argument (which is an object of the specialization which uniquely defined one of the overloads).

Intuitively, the only plausible variations are that

 - (A) constraints satisfaction cannot evaluated, and both candidate functions should be considered non-viable, or that
 - (B) ADL should never find a constrained hidden friend as a candidate function for lookup on a specialization for which the friend's constraint is not satisfied.

(A) would result in the program above being ill-formed (no viable function as constraints cannot be checked for satisfaction), but this would generalize also to non-overloaded hidden friends and thus possibly not be applicable.

This leads us to (B), which would make the program above well-formed by only finding candidate functions whose constraints are actually satisfied, but afaict constraint satisfaction is not part of finding candidates, but only of the later step of rejecting candidates when create the set of viable candidates.

None of (A) or (B) are implemented by MSVC, which seemingly accepts both overloads as viable and fails due to ambiguity when attempting to find the best viable function (no constraints checking?).

hubert-reinterpretcast: > Starting with lookup, [[basic.lookup.argdep]/4.3](https://eel.is/c++draft/basic.lookup.argdep#4.3) states that any (namespace-scope) friend function
> 
> > [...] declared in classes with reachable definitions in the set of associated entities are visible [...]
> 
> can be found via ADL on associated entities. Thus, as per [[basic.lookup.argdep]/2.2](https://eel.is/c++draft/basic.lookup.argdep#2.2), for the calls `f(s1)` and the `f(s2)` in the example above the associated entity is the class template itself, even if the arguments are objects of particular specializations of the class template. This would arguably mean that ADL, for both calls `f(s1)` and `f(s2)`, can find both overloads as candidate functions? So far, this is the interpretation MSVC has done, as it proceeds (and subsequently fails) with overload resolution.

Yes, it seems it should have succeeded because only one of the candidate functions is viable in each lookup result.

> In say the call `f(s1)`, how would constraint satisfaction be checked?

The constraint satisfaction can be checked in the context of the template instantiation.

PR #3774, opened 2020-02-20
LWG Poll 23: P2106R0 Alternative wording for GB315 and GB316

jensmaurer:
Also fixes NB GB 315 and NB GB 316.

Fixes #3725 .
Fixes cplusplus/nbballot#311.
Fixes cplusplus/nbballot#312.

Open issues:
- Changes contingent on D1243R4 not applied (for_each_n_result).

PR #3783, opened 2020-02-21
LWG Poll 22: P1243R4 Rangify New Algorithms

### LINKED ISSUE #3724: [2020-02 LWG Poll 22] P1243R4 Rangify New Algorithms
resolves FR305, US318, US307

burblebee:
Also fixes NB FR 305, US 318, and US 307 (C++20 CD).

Fixes #3724. 
Fixes cplusplus/papers#152
Fixes cplusplus/nbballot#301.
Fixes cplusplus/nbballot#314.
Fixes cplusplus/nbballot#303.

Notes:
* [alg.foreach] FYI: the referenced "Remarks:" is paragraph p27 (after the 2nd `for_each_n` declaration).

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -675,6 +675,11 @@ namespace std {
                   ForwardIterator first, ForwardIterator last, Function f);
 
   namespace ranges {
+    template<input_iterator I, class Proj = identity,
+             indirectly_unary_invocable<projected<I, Proj>> Fun>
+      constexpr for_each_n_result<I, Fun>
+        for_each_n(I first, iter_difference_t<I> n, Fun f, Proj proj = {});
+
     template<class I, class F>
     struct for_each_result {
       [[no_unique_address]] I in;
@@ -1775,6 +1780,20 @@ namespace std {
                           SampleIterator out, Distance n,
                           UniformRandomBitGenerator&& g);
 
+  namespace ranges {
+    template<input_iterator I, sentinel_for<I> S,
+             weakly_incrementable O, class Gen>
+      requires (forward_iterator<I> || random_access_iterator<O>) &&
+               indirectly_copyable<I, O> &&
+               uniform_random_bit_generator<remove_reference_t<Gen>>
+      O sample(I first, S last, O out, iter_difference_t<I> n, Gen&& g);
+    template<input_range R, weakly_incrementable O, class Gen>
+      requires (forward_range<R> || random_access_iterator<O>) &&
+               indirectly_copyable<iterator_t<R>, O> &&
+               uniform_random_bit_generator<remove_reference_t<Gen>>
+      O sample(R&& r, O out, range_difference_t<R> n, Gen&& g);
+  }
+
   // \ref{alg.random.shuffle}, shuffle
   template<class RandomAccessIterator, class UniformRandomBitGenerator>
     void shuffle(RandomAccessIterator first,
@@ -2809,6 +2828,13 @@ namespace std {
   template<class T, class Compare>
     constexpr const T& clamp(const T& v, const T& lo, const T& hi, Compare comp);
 
+  namespace ranges {
+    template<class T, class Proj = identity,
+             indirect_strict_weak_order<projected<const T*, Proj>> Comp = ranges::less>
+      constexpr const T&
+        clamp(const T& v, const T& lo, const T& hi, Comp comp = {}, Proj proj = {});
+  }
+
   // \ref{alg.lex.comparison}, lexicographical comparison
   template<class InputIterator1, class InputIterator2>
     constexpr bool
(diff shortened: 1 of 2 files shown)


source/algorithms.tex
```diff
@@ -667,6 +667,11 @@
                   ForwardIterator first, ForwardIterator last, Function f);
 
   namespace ranges {
+    template<input_iterator I, class Proj = identity,
```
timsong-cpp: Someone will need to apply P2106R0's addition of `for_each_n_result`. Not sure if that should happen here, in #3774, or separately.

source/algorithms.tex
```diff
@@ -8345,7 +8433,7 @@
```
CaseyCarter: Pre-existing: I've never been really happy with the handwavy "is no greater than" and "is less than" in the spec of this algorithm; adding projections to the mix exacerbates the problem. Consider rewording:
```
\pnum
Let tcode{proj} be \tcode{identity\{\}} 
for the overloads with no parameter \tcode{proj}, 
and let \tcode{comp} be \tcode{less\{\}} 
for the overloads with no parameter \tcode{comp}.

\pnum
\expects
\tcode{bool(comp(proj(hi), proj(lo)))} is \tcode{false}.
For the first form, type \tcode{T}
meets the \oldconcept{LessThan\-Comparable}
requirements (\tref{cpp17.lessthancomparable}).

\pnum
\returns
\tcode{lo} if \tcode{bool(comp(proj(v), proj(lo)))} is \tcode{true},
\tcode{hi} if \tcode{bool(comp(proj(hi), proj(v)))} is \tcode{true},
otherwise \tcode{v}.
```

source/algorithms.tex
jensmaurer: @burblebee , please do that in a separate commit in this pull request. Thanks.

jensmaurer:
@CaseyCarter's cleanup suggestions should go into a separate pull request, because the review committee wants to see the "original" applications first.

source/algorithms.tex
CaseyCarter: (I've pulled this out into #3785, please resolve this comment since I can't.)

source/algorithms.tex
burblebee: done, though I wasn't sure how the commit message should be worded

source/algorithms.tex
```diff
@@ -3233,6 +3259,44 @@
...
+\begin{itemdescr}
+\pnum
+\expects
+\tcode{n >= 0} is \tcode{true}.
+
+\pnum
+\effects
+Calls \tcode{invoke(f, invoke(proj, *i))}
+for every iterator \tcode{i} in the range
+\range{\tcode{first}}{\tcode{first + n}} in order.
```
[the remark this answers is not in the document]
burblebee: Ah, so it is.  Thanks!

burblebee:
> @CaseyCarter's cleanup suggestions should go into a separate pull request, because the review committee wants to see the "original" applications first.

Oops, didn't see this until after I pushed.  That said, Casey's suggestions don't appear to rise to the level that they warrant a separate commit IMHO.  FWIW I've made similar FIXUPs to other papers for similar cosmetic changes.

source/algorithms.tex
```diff
@@ -666,7 +666,15 @@
     void for_each(ExecutionPolicy&& exec,                       // see \ref{algorithms.parallel.overloads}
                   ForwardIterator first, ForwardIterator last, Function f);
 
+  template<class I, class F>
+  using for_each_n_result = in_fun_result<I, F>;
```
timsong-cpp: This needs to be in `namespace ranges`.

source/algorithms.tex
```diff
@@ -667,6 +667,11 @@
                   ForwardIterator first, ForwardIterator last, Function f);
 
   namespace ranges {
+    template<input_iterator I, class Proj = identity,
+             indirectly_unary_invocable<projected<I, Proj>> Fun>
+      constexpr for_each_n_result<I, Fun>
+       for_each_n(I first, iter_difference_t<I> n, Fun f, Proj proj = {});
```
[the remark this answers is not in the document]
burblebee: Thanks!

source/algorithms.tex
burblebee: Doh!  Thanks.

tkoeppe:
I agree, the fixups in this branch are not observable as far as the moved wording goes; they only exercise presentational discretion.

source/algorithms.tex
zygoloid: Thanks for doing this. I'm going to move these edits to the LWG 23 branch as part of merging; I think they make more sense there since that was the motion that directed us to make the edits. Putting them here prior to merge is fine, though -- I expect it saved you from needing to branch LWG23 off LWG22.

FYI: in future, it'd be useful to leave a comment on the pull request when things like this are necessary, so I understand what's going on when I come to merge. (I read through the "Conversation" tab on the pull request, but not through resolved comments on the patch itself.)

PR #3796, opened 2020-02-25
Fixes for 2020-02 CWG Poll 1

zygoloid:
Improvements inspired by a comment from @hubert-reinterpretcast on #3782.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3313,28 +3313,23 @@ refer to the new object and, once the lifetime of the new object has
 started, can be used to manipulate the new object, if
 the original object is transparently replaceable (see below)
 by the new object.
-An original object is \defn{transparently replaceable}
-by a new object if:
+An object $o_1$ is \defn{transparently replaceable}
+by an object $o_2$ if:
 \begin{itemize}
-\item the storage for the new object exactly overlays the storage
-location which the original object occupied, and
+\item the storage that $o_2$ occupies exactly overlays
+the storage that $o_1$ occupied, and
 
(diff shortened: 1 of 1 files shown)


source/basic.tex
```diff
@@ -3313,28 +3313,23 @@
...
+the storage that $o_1$ occupied, and
 
-\item the new object is of the same type as the original object
+\item $o_1$ and $o_2$ are of the same type
 (ignoring the top-level cv-qualifiers), and
 
-\item the original object is neither
-a complete object that is const-qualified nor
-a subobject of such an object, and
+\item $o_1$ is not a complete const object, and
```
hubert-reinterpretcast: I believe the following is almost as brief and a more direct adaptation of the existing bullet:
```suggestion
\item the complete object of $o_1$ is not a const object, and
```

zygoloid: Yes. The question is, do we want to check for a const complete object on every level of the upward recursion, or only when we reach the complete object? The latter seems simpler to me, but I don't feel at all strongly.

hubert-reinterpretcast: My thinking on this is that it is easier to get this bullet out of the way by checking it first and then ignoring it during the recursion.

jensmaurer: Assuming "the complete object of x" is x if x is a complete object, it feels that Hubert's formulation is a little clearer.

tkoeppe: It's just a trade: with the HST wording, we have to do the "complete object of" search in every recursion round.

PR #3797, opened 2020-02-25
[atomics.nonmembers,diff.cpp17.dcl.dcl,library,version.syn] Replace non-code "C++" with "\Cpp{}"

### LINKED ISSUE #3794: Replace non-code "C++" with "\Cpp{}"
I guess that `\Cpp{}` was introduced to avoid the hideous looking "++" with the PDF's font. There are plain "C++" that would look better if they were replaced with "\Cpp{}". `check.sh` should be enhanced to detect these. This may be obvious to pro LaTeX users, but pay attention to the index entries containing "C++". Some come from macro definitions like `\defnx{\Cpp{} standard library}{library!C++ standard}`, which should also replace the plain "C++" with "\Cpp{}" so that the index looks nice too.

johelegp:
Resolves #3794.
Index entries before/after:

 
---
 
---

### DIFF
--- a/source/atomics.tex
+++ b/source/atomics.tex
@@ -3296,7 +3296,7 @@ If no such member function exists, the program is ill-formed.
 \pnum
 \begin{note}
 The non-member functions enable programmers to write code that can be
-compiled as either C or C++, for example in a shared header file.
+compiled as either C or \Cpp{}, for example in a shared header file.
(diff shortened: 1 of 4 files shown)


johelegp:
I guess I should fixup the commit messages to include the stable names.

PR #3806, opened 2020-02-27
Fix for US 115 to avoid use of nouns with missing articles

hubert-reinterpretcast:


zygoloid:
I don't know a better name for this terse writing style than https://en.wikipedia.org/wiki/Headlinese but it's the same style we use for (eg) index entries (eg, "class name, scope of" not "class name, scope of a"), and I think it's appropriate both here and in the index entries.

hubert-reinterpretcast:
> I think it's appropriate both here and in the index entries.

Sure.

PR #3807, opened 2020-02-27
Restructure CA 107 functionally equivalent IFNDR note

hubert-reinterpretcast:
Adjusts a note identified in cplusplus/draft#3782 as being difficult to parse.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -1662,8 +1662,7 @@ void h() {
 \end{codeblock}
 \end{example}
 This similarity includes the situation where a program is ill-formed, no diagnostic required,
-when the meaning of the program depends on whether two constructs are equivalent,
-and they are functionally equivalent but not equivalent.
+when the meaning of the program would be changed by considering constructs that are merely functionally equivalent to be equivalent.
 \begin{example}
 \begin{codeblock}
 template <unsigned N> void f2()


source/templates.tex
```diff
@@ -1662,8 +1662,7 @@
 \end{codeblock}
 \end{example}
 This similarity includes the situation where a program is ill-formed, no diagnostic required,
-when the meaning of the program depends on whether two constructs are equivalent,
-and they are functionally equivalent but not equivalent.
+when the meaning of the program would be changed by considering constructs that are merely functionally equivalent to be equivalent.
```
jwakely: Would saying "because the meaning of the program depends ..." instead of "when the meaning of the program depends ..." help? Or "due to the meaning of the program depending on ..."

hubert-reinterpretcast: Although I originally argued against having this be different from the temp.over.link text, I think the new formulation here points at the rationale better. I also found the number of commas awkward before (but had no alternative ready).

zygoloid: I agree that it's formally equivalent to ask "are they functionally equivalent but not equivalent and does the meaning depend on whether they're equivalent?" or to ask "would the meaning be changed if we considered functionally-equivalent constructs to be equivalent?" However, given that this is referring to normative wording specified elsewhere, I'd like to keep the descriptive approach broadly the same between the two.

How about replacing this sentence with:

"As specified in [temp.over.link], if the validity or meaning of the program depends on whether two constructs are equivalent, and they are functionally equivalent but not equivalent, the program is ill-formed, no diagnostic required."

If we're not happy with that formulation, we should change it everywhere.

hubert-reinterpretcast: > How about replacing this sentence with:
> 
> "As specified in [temp.over.link], if the validity or meaning of the program depends on whether two constructs are equivalent, and they are functionally equivalent but not equivalent, the program is ill-formed, no diagnostic required."

I'm fine with that. @burblebee, as the person who raised the readability concern with respect to the note, do you have an opinion on this?

zygoloid:
@burblebee Do you have an opinion on this?

source/templates.tex
tkoeppe: I too find @zygoloid's suggestion clear and straight-forward, and easier to parse than the current proposal.

@burblebee: What do you think? Could you please update the PR if you agree?

jensmaurer: @burblebee , ping?

tkoeppe: @hubert-reinterpretcast: maybe just go ahead and make the update?

hubert-reinterpretcast:
Done by Jens and merged: https://github.com/cplusplus/draft/pull/5071

PR #3809, opened 2020-02-27
[lib] Format true/false in code font conventionally

johelegp:
These are conventionally wrapped in `\tcode{}`


jensmaurer:
Please rebase and force-push.

PR #3814, opened 2020-03-03
[utility.syn] Index in-place construction types

AlisdairM:
Add index entries for the inplace constructuction tag types
and values, as they exist only in the header synopsis.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -162,9 +162,7 @@ namespace std {
   template<class T2, class T1>
     constexpr const T2&& get(const pair<T1, T2>&& p) noexcept;
 
-  // \ref{pair.piecewise}, pair piecewise construction%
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -162,23 +162,33 @@
   template<class T2, class T1>
     constexpr const T2&& get(const pair<T1, T2>&& p) noexcept;
 
-  // \ref{pair.piecewise}, pair piecewise construction
+  // \ref{pair.piecewise}, pair piecewise construction%
+\indexlibraryglobal{piecewise_construct_t}%
+\indexlibraryglobal{piecewise_construct}
```
jwakely: These are already indexed, pointing to [pair.piecewise]. Do we usually index them in the header synopsis too?

AlisdairM: Good point, fixing now.

PR #3820, opened 2020-03-05
[rand.util.canonical] Convert normative duplication into a note.

zygoloid:


tkoeppe:
This seems more economical and less ad-hoc.

@opensdh: would you care to opine?

source/numerics.tex
```diff
@@ -4419,37 +4419,10 @@
...
- maps the result of one or more invocations
- of a supplied uniform random bit generator \tcode{g}
- to one member
- of the specified \tcode{RealType}
- such that,
- if the values $g_i$
- produced by \tcode{g}
- are uniformly distributed,
- the instantiation's results
- $t_j$, $0 \leq t_j < 1$,
```
opensdh: This normative text claims something desirable that the algorithm does not satisfy (because the algorithm is buggy: [LWG2524](https://cplusplus.github.io/LWG/issue2524)).  Should we wait to remove it until we have corrected the algorithm?

opensdh:
> This seems more economical and less ad-hoc.
> 
> @opensdh: would you care to opine?

I agree that in general this is a good idea.

source/numerics.tex
zygoloid: To my reading, the old wording only said that results in [0, 1) are uniformly distributed, and didn't say the results were necessarily in that range. But it does seem ambiguous. I think the new note below is correct, though: S/R^k is in [0, 1), because it's a real number, prior to the conversion to `RealType`. (Though what "using arithmetic of type `RealType`" means is anyone's guess.)

opensdh: It reads to me as if the inequality is an aside that's meant to apply to all values, and it's certainly not the case (in reality) that some are more uniformly distributed than others.  You're right that it seems to be conditioned on the uniform distribution from `g`.  The new note is plainly correct; I'm just a little uneasy about making the standard unambiguously say something bad (before fixing it to say something good anywhere).  I don't really think that anyone is going to reject P0952 because of this pull request; it's just a feeling about proper ordering.

zygoloid: OK, so how do we proceed? This existing paragraph smells bad, and I want to do *something* to it. I don't want to promote the aside the 0 <= t_j < 1 into an actual specification over the produced value of type `RealType`, because I don't think that is required by the current wording; this whole paragraph reads as a non-normative description of the intent of the math specified below. But we are in a pickle, because the intent is correct and the math is wrong.

Perhaps I should just pretend I didn't see this and wait for your paper to fix it :)

opensdh: I am generically in favor of the editors changing the standard as and only as I see fit, yes.

Anyway, you have an approval from the first author of that paper.  If you want to be rid of this, go ahead; I have already said I don't predict any actual harm (and implementations already do the bad thing that the rest of the wording specifies).  But I'd be marginally happier waiting (the paper won't even conflict with this).

tkoeppe:
I've also voiced similar concerns to Richard offline, but it seems like
this change is creating a mildly more tolerable situation, and as you said,
it shouldn't interfere with our paper (which will fix this once and for
all).

PR #3821, opened 2020-03-05
[time.format] Add/fix region-index entry for "format"

tkoeppe:
I can't tell what the perceived problem behind the `FIXME` was; I'd welcome feedback; @AlisdairM?

### DIFF
--- a/source/time.tex
+++ b/source/time.tex
@@ -10418,7 +10418,6 @@ strong_ordering operator<=>(const time_zone_link& x, const time_zone_link& y) no
 
 \rSec1[time.format]{Formatting}
 \indexlibrary{\idxcode{format}|)}%
-%% See also FIXME below
 
 \pnum
 Each \tcode{formatter}\iref{format.formatter} specialization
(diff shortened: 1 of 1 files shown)


source/time.tex
```diff
@@ -10417,6 +10417,8 @@
 \end{itemdescr}
 
 \rSec1[time.format]{Formatting}
+\indexlibrary{\idxcode{format}|)}%
+%% See also FIXME below
```
jensmaurer: Please drop the FIXME; if there is an issue, a github issue is the right procedure to track it.

tkoeppe: Done. It was partially to serve as a review hint to solicit suggestions for a better indexing range.

source/time.tex
```diff
@@ -10938,7 +10939,6 @@
 };
 \end{codeblock}
 
-% FIXME: Decide how to index this. This template-id is a lie.
```
jensmaurer: Now you've removed this totally unrelated FIXME.

Could I ask you limit this change to just the indexing snafu? Thanks.

jensmaurer:
Ah, that was where the "see below" pointed to. Anyway, please create separate github issues for any follow-up tasks.

source/time.tex
tkoeppe: This _is_ the snafu. It's a previously known one, just not one that was previously resolved...

PR #3823, opened 2020-03-05
[conv.ptr] Fix unclear specification of derived-to-base conversions for null pointers

sdkrystian:
[[conv.ptr] p3](http://eel.is/c++draft/conv.ptr#3) states:
> The result of the conversion is a pointer to the base class subobject of the derived class object. The null pointer value is converted to the null pointer value of the destination type.

While null pointer values are established as a singular value a pointer can have in [basic.compound], it is unique to the pointer type. Saying "The null pointer value" is rather unclear in what it refers to, as there is no single null pointer value.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -1021,14 +1021,15 @@ class of \tcode{D}, or a base class of a virtual base class of
 \tcode{D}, a program that necessitates this conversion is ill-formed.
 The result of the conversion refers to the same member as the pointer to
 member before the conversion took place, but it refers to the base class
-member as if it were a member of the derived class. The result refers to
-the member in \tcode{D}'s instance of \tcode{B}. Since the result has
+member as if it were a member of the derived class. If the source prvalue
+is a null member pointer value, the result is a null member pointer value;
+otherwise, the result refers to the member in the \tcode{B}
+base class subobject of \tcode{D}. Since the result has
 type ``pointer to member of \tcode{D} of type \cv{} \tcode{T}'',
 indirection through it with a \tcode{D} object is valid. The result is the same
 as if indirecting through the pointer to member of \tcode{B} with the
-\tcode{B} subobject of \tcode{D}. The null member pointer value is
-converted to the null member pointer value of the destination
-type.\footnote{The rule for conversion of pointers to members (from pointer to member
+\tcode{B} subobject of \tcode{D}.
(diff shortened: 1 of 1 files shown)


source/expressions.tex
```diff
@@ -986,10 +986,10 @@
 of \tcode{D}. If \tcode{B} is an
 inaccessible\iref{class.access} or
 ambiguous\iref{class.member.lookup} base class of \tcode{D}, a program
-that necessitates this conversion is ill-formed. The result of the
-conversion is a pointer to the base class subobject of the derived class
-object. The null pointer value is converted to the null pointer value of
-the destination type.
+that necessitates this conversion is ill-formed.
+If the source value is a null pointer value, the result is a
```
jensmaurer: What's the "source value"? Can we reformulate without using this novel term?

sdkrystian: @jensmaurer In [conv], there's a mixture of "source value", "source expression" and "source prvalue". I suppose that the best replacement here would be "source prvalue".

jensmaurer: Sounds better, because it aligns with the phrasing in the "If" at the start of the paragraph.

sdkrystian: @jensmaurer Applied

jensmaurer (approved):

sdkrystian:
@jensmaurer I applied the changes in [conv.ptr] to [conv.mem], [expr.static.cast], [expr.reinterpret.cast], and [expr.const.cast]. Another small error I fixed was that in [conv.mem] we say "The result refers to the member in `D`'s instance of `B`"; "instance" here was replaced with "base class subobject"

source/expressions.tex
zygoloid: I really don't like using "source prvalue" here. A prvalue is a kind of expression, so this sounds like it's talking about a (static) property of the expression, which it's not.

I think something like "A null pointer value of the source type is converted to a null pointer value of the destination type. A pointer to an object *o* of type `D` is converted to a pointer to the `B` base class subobject of *o*." would work, and fits much better into our pointer value taxonomy in [basic.compound]. Arguably this would need CWG review since it makes it much more clear that such a conversion applied to any other pointer value is undefined.

jensmaurer:
@sdkrystian, could you please rebase and implement @zygoloid's suggestion, which sounds like an improvement?

tkoeppe:
@sdkrystian: friendly ping.

PR #3824, opened 2020-03-05
Consistent null pointer value usage

sdkrystian:
Continuing the crusade for consistent conventions, this pull request addresses the inconsistent (and incorrect) usage of constructs such as "null" and "null pointer" when referring to null pointer values. 

A few related changes are included:
- The incorrect use of "null pointer"/"null" in [support.exceptions]; replaced with "the null value" to account for fancy pointers.
- More of a nitpick, but the specifying the "result" of a new-expression sounds more correct than specifying the "value" of the new-expression. While these are synonymous, "result" sounds more correct as a new-expression performs an operation, as opposed to representing some static value. This was applied in http://eel.is/c++draft/expr.new#9.6.1 and http://eel.is/c++draft/expr.new#20.sentence-4

While on the topic, crosslinks linking [conv.ptr] at mentions of "null pointer value" should be changed to [basic.compound[.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3954,7 +3954,7 @@ to an allocation function are unspecified.
 Even if the size of the space
 requested is zero, the request can fail. If the request succeeds, the
 value returned by a replaceable allocation function
-is a non-null pointer value\iref{conv.ptr}
+is a non-null pointer value\iref{basic.compound}
 \tcode{p0} different from any previously returned value \tcode{p1},
 unless that value \tcode{p1} was subsequently passed to a
 replaceable deallocation function.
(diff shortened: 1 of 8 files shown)


zygoloid:
I've pushed a minor fixup to make some of the new wording a little terser. I think this is going further than I'd like; it's useful to be able to say that a pointer is or is not null, and that's part of the very-well-established C++ vernacular, so I think we should define somewhere (maybe [basic.compound]) that a pointer is null if its value is a null pointer value. That would at least avoid a lot of the churn in the library wording.

jensmaurer:
This needs a rebase + force-push, as well as some updates.

PR #3825, opened 2020-03-05
[front.insert.iterator] Correct mis-incorporation of P0896R4

CaseyCarter:
What should be a defaulted default constructor is instead an ill-formed defaulted non-default constructor. This appears to be copy pasta from the original merge of P0896R4 in e85af9533ab6ab0f0b101ac61fc01ba47c406503.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -3825,7 +3825,7 @@ namespace std {
     using reference         = void;
     using container_type    = Container;
 
-    constexpr front_insert_iterator(Container& x) noexcept = default;
+    constexpr front_insert_iterator() noexcept = default;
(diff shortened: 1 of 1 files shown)


johelegp:
The `constexpr` is usually omitted. `counted`- and `common`- `_iterator`s also deviate from the norm.

CaseyCarter:
> The `constexpr` is usually omitted.

I don't think we're consistent enough about `constexpr` being explicit/implicit on defaulted default constructors to have a "usually." That said, I don't particularly care - my primary concern here is to make the smallest possible change to the DIS to unbreak it. If the editors ask me to remove the extraneous `constexpr` or `noexcept` I'm happy to do so.

PR #3826, opened 2020-03-06
[expr.ref] clarify access of non-static data members

Kerdek:
According to the definitions, objects don't have members, and member subobjects don't have names. The standard is somewhat unclear on how the member subobjects of an object of class type relate to the non-static data members of the class. This change helps clarify that there is a member subobject associated with each non-static data member.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -3274,8 +3274,8 @@ the named member of the class. The type of \tcode{E1.E2} is \tcode{T}.
 
 \item If \tcode{E2} is a non-static data member and the type of
 \tcode{E1} is ``\cvqual{cq1 vq1} \tcode{X}'', and the type of \tcode{E2}
-is ``\cvqual{cq2 vq2} \tcode{T}'', the expression designates the named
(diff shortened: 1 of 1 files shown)


source/expressions.tex
```diff
@@ -3274,8 +3274,8 @@
 
 \item If \tcode{E2} is a non-static data member and the type of
 \tcode{E1} is ``\cvqual{cq1 vq1} \tcode{X}'', and the type of \tcode{E2}
-is ``\cvqual{cq2 vq2} \tcode{T}'', the expression designates the named
-member of the object designated by the first expression. If \tcode{E1}
+is ``\cvqual{cq2 vq2} \tcode{T}'', the expression designates the associated
```
zygoloid: "corresponding" would be a better word than "associated" here, to indicate we mean the member subobject corresponding to `E2`.

Kerdek: I agree. The relationship is a bijection, commonly called a "one-to-one correspondence". "Corresponding" is better. Changed.

PR #3827, opened 2020-03-06
[basic], [expr] Fix incorrect cross-references to [conv.ptr] for null pointer values

sdkrystian:
The definition of *null pointer value" was moved to [basic.compound], however, multiple cross-references still refer to [conv.ptr].


tkoeppe:
There is also a circular refence in [basic.compound]p(3.3) now.

sdkrystian:
@tkoeppe Removing the reference to [conv.ptr] would resolve this.

tkoeppe:
I agree, that reference is no longer needed.

PR #3828, opened 2020-03-06
Use consistent terminology when referring to what a pointer to member points to

sdkrystian:
Currently, a mixture of "points", "refers", "selects", "designates", and "denotes" is used when referring to what a pointer to member points to. While pointer to members aren't the same as *pointers*, they certainly are "pointers" in the sense that an object or function can be obtained through them via indirection, hence the replacement with "points to" in this PR. Additionally, we already use "refers to", "designates", and "denotes" for glvalue expressions, so this clears up any ambiguity.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -4469,7 +4469,7 @@ was obtained through one of the following:
 \item
   a pointer-to-member operation\iref{expr.mptr.oper} using the \tcode{.*} operator
   where the left operand is one of these expressions and
-  the right operand is a pointer to data member of non-reference type,
+  the right operand is of pointer to data member type,
 \item
   a
   \begin{itemize}
@@ -5196,7 +5196,7 @@ Each distinct enumeration constitutes a different
 \defnx{pointers to non-static class members}{pointer to member},%
 \footnote{Static class members are objects or functions, and pointers to them are
 ordinary pointers to objects or functions.}
-which identify members of a given
+which \defn{designate} members of a given
 type within objects of a given class, \ref{dcl.mptr}.
 Pointers to data members and pointers to member functions are collectively
 called \term{pointer-to-member} types.
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -3417,7 +3417,7 @@ shall appear only as:
 \begin{itemize}
 \item the function type for a non-static member function,
 
-\item the function type to which a pointer to member refers,
+\item the function type to which a pointer to member points,
 
 \item the top-level function type of a function typedef declaration
 or \grammarterm{alias-declaration},
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -2921,8 +2921,9 @@ is ``\placeholder{derived-declarator-type-list}
 then the type of the identifier of
 \tcode{D}
 is ``\placeholder{derived-declarator-type-list} \grammarterm{cv-qualifier-seq} pointer to member of class
-\grammarterm{nested-name-specifier} of type
-\tcode{T}''.
+\grammarterm{nested-name-specifier} of type \tcode{T}''.
+\tcode{T} is known as the \defn{member type};
+it shall not be a reference type or ``\cv{}~\tcode{void}''.
 The optional \grammarterm{attribute-specifier-seq}\iref{dcl.attr.grammar} appertains to the
 pointer-to-member.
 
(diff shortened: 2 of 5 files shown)


zygoloid:
We need to decide which word we want to use here. I've asked CWG for opinions.

zygoloid:
Conclusion from committee discussion: use "designates" to describe the relationship between a pointer-to-member type `T M::*` and the type `T`, and also to describe the relationship between a pointer-to-member value `&M::x` and the member `M::x`.

zygoloid:
Some specific examples for the case of referring to the member type:

 * Add a sentence "The type `T` is the *member type* of the pointer-to-member type." To near the end of [dcl.mptr]p1.

 * "the function type to which a pointer to member refers" in [dcl.fct]p6 would become "the member type of a pointer to member type"

 * "For a call to a non-static member function, the postfix expression shall be [...] a pointer-to-member expression selecting a function member" could be just "For a call to a non-static member function, the postfix expression shall be [...] a pointer-to-member expression of function type".

sdkrystian:
@zygoloid Alright, I went ahead and applied the changes. In addition to this, I restructured [expr.mptr.oper] to more closely follow the layout of [expr.ref], and condensed the wording (we had requirements placed on operands all over the place, and it appears that there was an attempt to make the subclause more like [expr.ref], but it was never fully completed)

sdkrystian:
Just noticed that we have no normative wording that says the `mutable` specifier is ignored by pointer to member operators. Should we add that to this PR?

jensmaurer:
"mutable" is not part of the pointer-to-member type, so why would the pointer-to-member operators care?

sdkrystian:
`mutable` isn't part of the members type either, it's a property of the declaration. Since the cv-qualification of the pointer to member expression is the same as that of an equivalent class member access expression, the "mutability" of the designed member is still considered.

jensmaurer:
For clarity, could you please show an example?

source/basic.tex
```diff
@@ -4469,7 +4469,7 @@
 \item
   a pointer-to-member operation\iref{expr.mptr.oper} using the \tcode{.*} operator
   where the left operand is one of these expressions and
-  the right operand is a pointer to data member of non-reference type,
+  the right operand is of pointer to data member type,
```
jensmaurer, line 4472: Why is "non-reference" removed here?

jensmaurer: Ah, I see, the restriction is below.

source/declarations.tex
```diff
@@ -2987,11 +2988,8 @@
 \end{example}
 
 \pnum
-A pointer to member shall not point to a static member
-of a class\iref{class.static},
-a member with reference type,
-or
-``\cv{}~\tcode{void}''.
+A pointer to member shall not designate a static member
+of a class\iref{class.static}.
```
jensmaurer, line 2992: This restriction seems redundant; how could the "static member" situation ever arise?
&C::static_member yields a regular pointer, not a pointer to member.

source/expressions.tex
```diff
@@ -5582,39 +5582,35 @@
...
-type ``pointer to \tcode{U}''
-where \tcode{U} is either \tcode{T} or
-a class of which \tcode{T}
-is an unambiguous and accessible base class.
-The expression \tcode{E1->*E2} is converted into the equivalent form
-\tcode{(*(E1)).*E2}.
+In a pointer-to-member expression of the form \tcode{E1->*E2}, \tcode{E2} shall
+be of type ``pointer to member of \tcode{T}'' and \tcode{E1} shall be of type
+``pointer to \cv{}~\tcode{U}'' where \tcode{U} is \tcode{T} or a type of which
+\tcode{T} is an accessible and unambiguous base class. The expression is converted
```
jensmaurer, line 5588: This seems to be redundant with the checks after the rewrite.

source/declarations.tex
sdkrystian: True; I didn't want to drastically change it _too_ much, be this can certainly be turned into a note saying something along the lines of: "A pointer to member designating a static member cannot be formed as there is no syntax to do so."

source/expressions.tex
sdkrystian: My intent was to allow for the reader to clearly reason the requirements on the operands, but this could be removed in favor of simply saying:
> In a pointer-to-member expression of the form `E1->*E2`,  `E1` shall be a prvalue of pointer to object type. The expression is converted to the equivalent form `(*(E1)).*E2`; the remainder of this subclause will
address only the operator `.*`.

sdkrystian:
> For clarity, could you please show an example?

```
struct S
{
    mutable int a;
};

void f()
{
    const S obj;
    auto mem = &S::a;
    obj.*mem = 10;
    // type of obj is const S
    // type of mem is int S::*
    // given a hypothetical class member access expression obj.id where id names the same member mem designates (a), then the const qualifier would be removed from the point-to-member expression because it would also be removed in the corresponding class member access expression as a is declared mutable ([expr.ref] p6.2)
}
```

PR #3832, opened 2020-03-07
[atomics] Consistent use of enum class memory_order

AlisdairM:
Since the application of P0439R0, the library wording has been partially
updated to use the new scoped enumaration values whenever memory order
constants are required, and partially retains use of the inline
constants supplied for back compatibility.

This patch conistently uses the memory_order::enumerators, retaining
the old name only for their declaration, the index, and for Annex C
wording on C++17 compatibility, which /should/ use the old spelling.

### DIFF
--- a/source/atomics.tex
+++ b/source/atomics.tex
@@ -714,20 +714,20 @@ namespace std {
     atomic_ref(const atomic_ref&) noexcept;
     atomic_ref& operator=(const atomic_ref&) = delete;
 
-    void store(T, memory_order = memory_order_seq_cst) const noexcept;
+    void store(T, memory_order = memory_order::seq_cst) const noexcept;
     T operator=(T) const noexcept;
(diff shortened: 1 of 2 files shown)


AlisdairM:
I need help understanding the Travis failure - my local builds all succeed.  Which part of the message above indicates the nature of the failure that I need to address?

timsong-cpp:
`future.tex:Overfull \hbox (8.65416pt too wide) in paragraph at lines 1891--1892`.

Looks like the change caused the code block to run into the right margin.

AlisdairM:
Reverted the changes to Annex D, so at least the text of the main standard can benefit from the update.  Need someone more skilled in the art of LaTeX, typography, and the margins, to figure a better solution for the long expressions for a couple of the Returns: clauses in the Annex D deprecated atomic access to shared_ptr API.

PR #3833, opened 2020-03-07
[memory.syn] Index the pointer_safety enum class

AlisdairM:
As the enum class 'pointer_safety' is defined only in the header synopsis
for <memory>, it had escaped indexing.  This adds the missing entries.


AlisdairM:
Note that this creates a second "relaxed" entry in the library index, as there is a 'relaxed' enumerator for two different scoped enums, memory_order and pointer_safety, and I don't see how to coalesce those into a single parent with two children.

PR #3836, opened 2020-03-08
Fix incorrect references to past the end pointers

sdkrystian:
The proper terminology is used and defined in [[basic.compound] p3](http://eel.is/c++draft/basic.compound#3):
> [...] a pointer past the end of the last element of an array `x` of `n` elements [...]

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -5764,8 +5764,8 @@ the result has the type of \tcode{P}.
 of an array object \tcode{x} with $n$ elements\iref{dcl.array},%
 \footnote{As specified in \ref{basic.compound},
 an object that is not an array element
-is considered to belong to a single-element array for this purpose and
-a pointer past the last element of an array of $n$ elements
+is considered to belong to a single-element array for this purpose, and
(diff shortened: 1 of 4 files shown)


source/expressions.tex
```diff
@@ -6118,7 +6118,7 @@
 \begin{itemize}
 \item
 If one pointer represents the address of a complete object, and another
-pointer represents the address one past the last element of a different
+pointer represents the address past the end\iref{basic.compound} of a different
```
zygoloid: This is about the represented address, not the pointer value, so I think this change isn't what we want. Something like this might work, but it's a bit clumsy: "and another pointer represents the same address as a pointer past the end of a different complete object".

sdkrystian: How about "pointer represents the address of a pointer value past the end of a different complete object"

jensmaurer: I think we need "same address" somewhere in the mix.

tkoeppe:
I'm inclined to close this suggestion, given that it doesn't work as is, and it's not clear that we know of an actual improvement.

PR #3843, opened 2020-03-09
[class.mfct] consteval member functions are also inline.

### LINKED ISSUE #3663: [class.mfct]p2's list of keywords that make a function inline forgets consteval
[class.mfct]p2 says:

> An inline member function (whether static or non-static) may also be defined outside of its class definition provided either its declaration in the class definition or its definition outside of the class definition declares the function as `inline` or `constexpr`.

Obviously this should also permit `consteval`.

jensmaurer:
Fixes #3663.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -854,30 +854,29 @@ in which case it is an inline\iref{dcl.inline} member function
 if it is attached to the global module, or
 it may be defined outside of its class
 definition if it has already been declared but not defined in its class
-definition. A member function definition that appears outside of the
+definition.
+\begin{note}
+A member function is also inline if it is declared
+\tcode{inline}, \tcode{constexpr}, or \tcode{consteval}.
+\end{note}
+
+\pnum
+A member function definition that appears outside of the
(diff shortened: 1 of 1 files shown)


source/classes.tex
```diff
@@ -866,7 +866,8 @@
 An inline member function (whether static or non-static) may
 also be defined outside of its class definition provided either its
 declaration in the class definition or its definition outside of the
-class definition declares the function as \tcode{inline} or \tcode{constexpr}.
+class definition declares the function as
+\tcode{inline}, \tcode{constexpr}, or \tcode{consteval}.
```
zygoloid: I was going to say: maybe just "declares the function as an inline function".

But then this sentence says: "an inline member function may be defined outside the class if it's an inline function". Which reduces to "an inline member function may be defined outside the class". But that's also true for non-inline member functions and we said it in the previous paragraph. So what point are we trying to make here exactly?

Maybe this should be a note attached to the previous paragraph? Or removed entirely?

jensmaurer: I've fixed this with a slightly larger hammer.

PR #3844, opened 2020-03-09
[lex.ccon,lex.string] Remove redundant repetition of the grammar.

### LINKED ISSUE #3636: [lex] Replace "character literal" with references to grammar term character-literal

jensmaurer:
Also use the grammar non-terminals character-literal and
string-literal throughout the standard.

Fixes #3636.

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -98,7 +98,7 @@ void f(char*) {
 
 \howwide
 Programs that have a legitimate reason to treat string literal objects
-as pointers to potentially modifiable memory are probably rare.
+as potentially modifiable memory are probably rare.
 
 \rSec2[diff.basic]{\ref{basic}: basics}
 
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -32,7 +32,7 @@ Common.
 
 \diffref{lex.ccon}
 \change
-Type of character literal is changed from \tcode{int} to \tcode{char}.
+Type of \grammarterm{character-literal} is changed from \tcode{int} to \tcode{char}.
 \rationale
 This is needed for improved overloaded function argument type
 matching.
@@ -64,7 +64,7 @@ Programs which depend upon \tcode{sizeof('x')} are probably rare.
 \diffref{lex.string}
 \change
 String literals made const.\\
-The type of a string literal is changed
+The type of a \grammarterm{string-literal} is changed
 from ``array of \tcode{char}''
 to ``array of \tcode{const char}''.
 The type of a UTF-8 string literal is changed
@@ -97,7 +97,7 @@ void f(char*) {
 \end{codeblock}
 
 \howwide
-Programs that have a legitimate reason to treat string literals
+Programs that have a legitimate reason to treat string literal objects
 as pointers to potentially modifiable memory are probably rare.
 
 \rSec2[diff.basic]{\ref{basic}: basics}
@@ -889,15 +889,15 @@ by the chapters of this document.
 
 \diffref{lex.pptoken}
 \change
-New kinds of string literals.
+New kinds of \grammarterm{string-literal}s.
 \rationale
 Required for new features.
 \effect
 Valid \CppIII{} code may fail to compile or produce different results in
 this International Standard. Specifically, macros named \tcode{R}, \tcode{u8},
 \tcode{u8R}, \tcode{u}, \tcode{uR}, \tcode{U}, \tcode{UR}, or \tcode{LR} will
-not be expanded when adjacent to a string literal but will be interpreted as
-part of the string literal. For example:
+not be expanded when adjacent to a \grammarterm{string-literal} but will be interpreted as
+part of the \grammarterm{string-literal}. For example:
 \begin{codeblock}
 #define u8 "abc"
 const char* s = u8"def";        // Previously \tcode{"abcdef"}, now \tcode{"def"}
@@ -1435,7 +1435,7 @@ Necessary to enable single quotes as digit separators.
 Valid \CppXI{} code may fail to compile or may change meaning in this
 International Standard. For example, the following code is valid both in \CppXI{} and in
 this International Standard, but the macro invocation produces different outcomes
-because the single quotes delimit a character literal in \CppXI{}, whereas they are digit
+because the single quotes delimit a \grammarterm{character-literal} in \CppXI{}, whereas they are digit
 separators in this International Standard:
 
 \begin{codeblock}
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -1152,13 +1152,15 @@ this behavior is deprecated\iref{depr.arith.conv.enum}.
 \rSec2[expr.prim.literal]{Literals}
 
 \pnum
-A
 \indextext{literal}%
 \indextext{constant}%
-\grammarterm{literal}
-is a primary expression.
-Its type depends on its form\iref{lex.literal}.
-A string literal is an lvalue; all other literals are prvalues.
+A \grammarterm{literal} is a primary expression,
+whose type is determined based on its form as specified in \ref{lex.literal}.
+A \grammarterm{string-literal} is an lvalue,
+a \grammarterm{user-defined-literal}
+has the same value category
+as the corresponding operator call expression described in \ref{lex.ext},
+and any other \grammarterm{literal} is a prvalue.
 
 \rSec2[expr.prim.this]{This}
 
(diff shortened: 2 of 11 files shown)


source/compatibility.tex
```diff
@@ -97,7 +97,7 @@
 \end{codeblock}
 
 \howwide
-Programs that have a legitimate reason to treat string literals
+Programs that have a legitimate reason to treat \grammarterm{string-literal}s
```
zygoloid: I think this means "string literal objects" not *string-literal*s.

source/compatibility.tex
```diff
@@ -64,7 +64,7 @@
 \diffref{lex.string}
 \change
 String literals made const.\\
-The type of a string literal is changed
+The type of a \grammarterm{string-literal} is changed
```
zygoloid: (Digression: I think we are presenting *string-literal*s very strangely. We should have something like:

> *string-literal-token*:
> &nbsp; &nbsp; \<current *string-literal* productions>

> *string-literal*:
> &nbsp; &nbsp; *string-literal-token*
> &nbsp; &nbsp; *string-literal* *string-literal-token*

And we should delete phase 6 of translation and move the string literal concatenation and string literal objects wording out of [lex] into [expr]. Then we could clearly distinguish between the cases where we're talking about *string-literal*s, such as this, and the cases where we're talking about *string-literal-token*s, which covers a bunch of the other changes in this PR.)

source/lex.tex
```diff
@@ -319,8 +319,8 @@
 phases 3 through 6. The categories of preprocessing token are: header names,
 placeholder tokens produced by preprocessing \tcode{import} and \tcode{module} directives
 (\grammarterm{import-keyword}, \grammarterm{module-keyword}, and \grammarterm{export-keyword}),
-identifiers, preprocessing numbers, character literals (including user-defined character
-literals), string literals (including user-defined string literals), preprocessing
+identifiers, preprocessing numbers, \grammarterm{character-literal}s (including user-defined character
+literals), \grammarterm{string-literal}s (including user-defined string literals), preprocessing
```
zygoloid: This change is incorrect. User-defined {character,string} literals are not *character-literal* or *string-literal* tokens.

source/lex.tex
```diff
@@ -334,8 +334,8 @@
 circumstances during translation phase 4, white space (or the absence
 thereof) serves as more than preprocessing token separation. White space
 can appear within a preprocessing token only as part of a header name or
-between the quotation characters in a character literal or string
-literal.
+between the quotation characters in a \grammarterm{character-literal} or
+\grammarterm{string-literal}.
```
zygoloid: This is also unclear in the presence of UDLs.

source/lex.tex
```diff
@@ -256,8 +256,8 @@
 If a \grammarterm{universal-character-name} outside
 the \grammarterm{c-char-sequence}, \grammarterm{s-char-sequence}, or
 \grammarterm{r-char-sequence} of
-a character or
-string literal corresponds to a control character or
+a \grammarterm{character-literal} or
+\grammarterm{string-literal} corresponds to a control character or
```
zygoloid: This change is unclear in the presence of UDLs. (*string-literal* is both a syntactic production and a morphological production, and it's unclear what this means.)

source/expressions.tex
```diff
@@ -1158,7 +1158,7 @@
 \grammarterm{literal}
 is a primary expression.
 Its type depends on its form\iref{lex.literal}.
-A string literal is an lvalue; all other literals are prvalues.
+A \grammarterm{string-literal} is an lvalue; all other literals are prvalues.
```
zygoloid: Hmm. The existing wording here is just plain wrong. UDLs can have any value category.

source/compatibility.tex
jensmaurer: "string literal objects" it is.

source/compatibility.tex
jensmaurer: Uh, is introducing new grammar-terms and deleting a phase of translation considered editorial these days?

source/lex.tex
jensmaurer: Reverted.

source/lex.tex
jensmaurer: Reverted.

source/lex.tex
jensmaurer: We are in [lex], so grammar is assumed to refer to the lexical grammar. And UDLs do use the productions string-literal and character-literal.

source/expressions.tex
jensmaurer: Applied a larger hammer.

source/compatibility.tex
zygoloid: No, hence the "digression" part. :) I think that is something we should take to CWG for the C++23 cycle.

source/expressions.tex
```diff
@@ -1152,13 +1152,16 @@
...
-Its type depends on its form\iref{lex.literal}.
-A string literal is an lvalue; all other literals are prvalues.
+A \grammarterm{literal} is a primary expression.
+A \grammarterm{user-defined-literal} is interpreted as a function call
+as described in \ref{lex.ext};
+the type and value category of such a literal is determined by the
+rules for function calls\iref{expr.call}.
+For other literals,
+the type is determined by its form as specified in \ref{lex.literal};
+a \grammarterm{string-literal} is an lvalue; all other literals are prvalues.
```
zygoloid: This is a bit too semicolon-heavy for my tastes. How about something like:

> A *literal* is a primary expression, whose type is determined based on its form as specified in [lex.literal]. A *string-literal* is an lvalue, a *user-defined-literal* has the same value category as the corresponding operator call expression described in [lex.ext], and any other *literal* is a prvalue.

source/lex.tex
```diff
@@ -127,16 +127,16 @@
...
+\grammarterm{string-literal}, as well as each escape sequence and \grammarterm{universal-character-name} in a
+\grammarterm{character-literal} or a non-raw string literal, is converted to the corresponding
 member of the execution character set~(\ref{lex.ccon}, \ref{lex.string}); if
 there is no corresponding member, it is converted to an \impldef{converting
 characters from source character set to execution character set} member other
 than the null (wide) character.\footnote{An implementation need not convert all
 non-corresponding source characters to the same execution character.}
 
-\item Adjacent string literal tokens are concatenated.
+\item Adjacent \grammarterm{string-literal} tokens are concatenated.
```
zygoloid: This is not correct; as specified in [lex.ext]p8, this also includes *user-defined-string-literal* tokens, and we're definitely talking about complete tokens and not lexer rules here.

source/lex.tex
zygoloid: I don't think this is clear. (Similar questions: do the reservations in [lex.name]/3 apply to *ud-suffix*es, or only to identifier tokens? Is "orbit" as an identifier ill-formed because it contains the identifier "or" which is reserved and "shall not be used otherwise"?)

Adding an "(including within a *user-defined-literal*)" or similar would make this completely clear. If you think this is already clear normatively, maybe use a footnote "This includes a *string-literal* or *character-literal* within a *user-defined-string-literal* or *user-defined-character-literal* token." rather than a parenthetical?

source/templates.tex
```diff
@@ -1221,7 +1221,7 @@
 or be the address of (respectively):
 \begin{itemize}
 \item a temporary object\iref{class.temporary},
-\item a string literal\iref{lex.string},
+\item a \grammarterm{string-literal}\iref{lex.string},
```
zygoloid: string literal object

source/lex.tex
jensmaurer: I don't think that [lex] is clear normatively, at all, in particular the little lexer sub-grammar used here with unclear relation to the "big" grammar.

source/templates.tex
```diff
@@ -1269,8 +1269,8 @@
   @\commentellip@
 };
 
-X<const char*, "Studebaker"> x; // error: string literal as \grammarterm{template-argument}
-X<const char*, "Knope" + 1> x2; // error: subobject of string literal as \grammarterm{template-argument}
+X<const char*, "Studebaker"> x; // error: \grammarterm{string-literal} as \grammarterm{template-argument}
+X<const char*, "Knope" + 1> x2; // error: subobject of \grammarterm{string-literal} as \grammarterm{template-argument}
```
zygoloid: "string literal object" (x2)

source/templates.tex
jensmaurer: Fixed.

source/templates.tex
jensmaurer: Fixed.

source/lex.tex
jensmaurer: Reverted.

source/lex.tex

source/expressions.tex

PR #3845, opened 2020-03-09
[support.dynamic] Improve cross-referencing for dynamic memory subcla…

AlisdairM:
…uses

Update several cross-references to subclause 17.6 with more precise
refernces to subclauses nested one level deeper.  For the synopsis
of the header, added a couple of banner comments with cross-references
too.

In addition, looked into ordering the global namespace declarations
in front of opening namespace std to improve clarity, but that fails
due to use of types declared in namespace std by this header!

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -2003,14 +2003,9 @@ functions that manage the allocation of dynamic storage in a program.
 It also defines components for reporting storage management errors.
 
 \rSec2[new.syn]{Header \tcode{<new>} synopsis}
-
-\indexlibraryglobal{align_val_t}%
-\indexlibraryglobal{destroying_delete_t}%
-\indexlibraryglobal{destroying_delete}%
-\indexlibraryglobal{nothrow_t}%
-\indexlibraryglobal{nothrow}%
 \begin{codeblock}
 namespace std {
+  // \ref{alloc.errors}, memory allocation exceptions
   class bad_alloc;
   class bad_array_new_length;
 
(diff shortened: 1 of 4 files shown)


source/support.tex
```diff
@@ -2019,6 +2014,12 @@
   };
   inline constexpr destroying_delete_t destroying_delete{};
 
+  // global operator new control%
```
```suggestion
  // global \tcode{operator new} control%
```

source/support.tex
```diff
@@ -2036,6 +2037,7 @@
   inline constexpr size_t hardware_constructive_interference_size = @\impdef{}@;
 }
 
+// \ref{new.delete}, global operator new
```
```suggestion
// \ref{new.delete}, storage allocation and deallocation 
```

source/support.tex
```diff
@@ -2003,14 +2003,9 @@
...
 \rSec2[new.syn]{Header \tcode{<new>} synopsis}
-
-\indexlibraryglobal{align_val_t}%
-\indexlibraryglobal{destroying_delete_t}%
-\indexlibraryglobal{destroying_delete}%
-\indexlibraryglobal{nothrow_t}%
-\indexlibraryglobal{nothrow}%
 \begin{codeblock}
 namespace std {
+  // \ref{alloc.errors}, memory allocation exceptions
```
```suggestion
  // \ref{alloc.errors}, storage allocation errors 
```

AlisdairM:
I had managed to convince myself otherwise, but the counterexamples are few and far between, and mostly dropping a word or two, not rephrasing.  Thanks for supplying the fixes too!

PR #3849, opened 2020-03-10
[range.split.outer], [range.split.inner] Fix misuses of current_

### LINKED ISSUE #3848: [range.split] confuses current and current_
At http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0896r4.pdf#page=121 we can see that the _Effects:_ of `split_view::outer_iterator::operator++` is supposed to use _current_ but in the working draft [range.split.outer] p6 uses `current_` which is not the same thing (see [range.split.outer] p1 and #3847).

The same problem exists in [range.split.outer] p8 (comparison with default sentinel), and [range.split.inner] in the synopsis of _`inner-iterator`_ (definition of `operator*`).

The original application of P0896R4 (e85af9533ab6ab0f0b101ac61fc01ba47c406503) was consistent with P0896R4. Then 7989bb445478b41a9fee3539bebffcaa46ba5270 got rid of _current_ and made _`current_`_ stand in for both `current_` and `parent_->current_`. Then 89abe001f5cba2775ec1f908eadf765572413fd5 (fixing #3562) reintroduced _current_ but didn't restore uses of _current_, leaving them incorrectly referring to _`current_`_.

I think this must be fixed for C++20.

jwakely:
This restores references to the 'current' placeholder.

Fixes #3848

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -5590,17 +5590,17 @@ constexpr @\exposid{outer-iterator}@& operator++();
 Equivalent to:
 \begin{codeblock}
 const auto end = ranges::end(@\exposid{parent_}@->@\exposid{base_}@);
-if (@\exposid{current_}@ == end) return *this;
+if (@\placeholder{current}@ == end) return *this;
 const auto [pbegin, pend] = subrange{@\exposid{parent_}@->@\exposid{pattern_}@};
-if (pbegin == pend) ++@\exposid{current_}@;
+if (pbegin == pend) ++@\placeholder{current}@;
 else {
   do {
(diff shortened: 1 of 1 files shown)


jwakely:
> What a mess.

Those were my words exactly when I discussed this on IRC :smiley:

CaseyCarter:
While this is certainly correct, I think we are just asking to screw this up again if we don't rename one of _`current`_ or _`current_`_ to something more clearly distinct (no, not _`_current`_!). I don't have any great ideas, or I wouldn't have used the names _`current`_ and `current_` in the first place.

jensmaurer:
Let's merge this for now, and discuss long-term remedies in #3851.

PR #3861, opened 2020-03-11
[module.unit] module-declarations no longer contain \tcode{export}

timsong-cpp:


source/modules.tex
```diff
@@ -49,7 +49,7 @@
 
 \pnum
 A \defn{module interface unit} is a module unit whose
-\grammarterm{module-declaration} contains the \tcode{export} keyword;
+\grammarterm{module-declaration} contains an \grammarterm{export-keyword};
```
jensmaurer: This should say "whose /module-declaration/ starts with /export-keyword/." for a bit more precision. "contains" is usually danger-zone wording (it might transitively contain). Except with the special case of /export-keyword/, we're probably safe.

source/modules.tex
```diff
@@ -153,7 +153,7 @@
 
 \pnum
 A \grammarterm{module-declaration}
-that contains neither \tcode{export}
+that contains neither an \grammarterm{export-keyword}
```
jensmaurer: This one probably needs to stay as-is.

source/modules.tex
timsong-cpp:
```suggestion
\grammarterm{module-declaration} starts with \grammarterm{export-keyword};
```

source/modules.tex
timsong-cpp: By "this one" do you mean "contains" or `export`?

jensmaurer: I mean "contains".

PR #3867, opened 2020-03-12
[alg.replace] Clarify which value gets assigned depending on E.

### LINKED ISSUE #3317: [alg.replace]p8 Ambiguous assignment
https://github.com/cplusplus/draft/blob/9af3fc206fb538760af373b78d49f9658a0eeec8/source/algorithms.tex#L4823-L4826
The spec isn't clear on which expression is assigned when `E` has which value.

jensmaurer:
Fixes #3317

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -5006,10 +5006,13 @@ do not overlap.
 
 \pnum
 \effects
-Assigns to every iterator \tcode{i}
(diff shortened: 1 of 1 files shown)


zygoloid:
@jwakely Please can you double-check the new bullets are the correct way around, per the intent here?

PR #3872, opened 2020-03-13
[temp.explicit] Fix unclear reference to elaborated-type-specifier

sdkrystian:
It is unclear what elaborated-type-specifier we are referring to, and the current wording does not do a good job specifying what form the explicit instantiation should take.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -127,13 +127,29 @@ the declaration declares a class template partial specialization\iref{temp.class
 \end{note}
 
 \pnum
-In a
-\grammarterm{template-declaration},
-explicit specialization, or explicit instantiation the
-\grammarterm{init-declarator-list}
-in the declaration shall contain at most one declarator.
-When such a declaration is used to declare a class template,
-no declarator is permitted.
+A \grammarterm{template-declaration} or \grammarterm{explicit-specialization}
+declares a class or a class template if its \grammarterm{declaration} contains no
+\grammarterm{declarator} and it does not declare an alias template\iref{temp.alias} or
+deduction guide\iref{temp.deduct.guide}; the \grammarterm{decl-specifier-seq} of such a
+declaration shall contain a single \grammarterm{defining-type-specifier} that is either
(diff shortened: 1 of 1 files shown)


source/templates.tex
```diff
@@ -6572,10 +6572,9 @@
 shall appertain to an explicit instantiation.
 
 \pnum
-If the explicit instantiation is for a class or member class, the
-\grammarterm{elaborated-type-specifier} in the \grammarterm{declaration}
-shall include a \grammarterm{simple-template-id};
-otherwise, the \grammarterm{declaration}
+If the explicit instantiation is for a class or member class, the \grammarterm{declaration}
+shall consist of a single \grammarterm{elaborated-type-specifier} that includes a 
+\grammarterm{simple-template-id}; otherwise, the \grammarterm{declaration}
```
zygoloid: I think the "shall consist of a single" part of this rule belongs in [temp.pre]p5:

> When such a declaration is used to declare a class template <ins>or class template specialization</ins>, no declarator is permitted.

As written, this doesn't seem quite right: requiring the *declaration* to consist of only an *elaborated-type-specifier* would disallow attributes.

But there's something really fishy about the rules here in general: we're starting (in both [temp.pre] and here) by assuming we know whether we're explicitly instantiating a class or something else, and there's no rules to say when we are doing so.

What we *really* mean is something like: if there is no *declarator*, then the template, template specialization, or explicit instantiation declaration is declaring a class or class template, and the decl-specifier-seq shall contain a single defining-type-specifier that is either a class-specifier, an enum-specifier, or an elaborated-type-specifier. And for an explicit instantiation, it shall be an elaborated-type-specifier that either includes a simple-template-id or whose nested-name-specifier includes a simple-template-id.

jensmaurer:
And please fix the Travis failure.

sdkrystian:
@zygoloid I added a commit which specifies what a template actually declares in [temp.pre]... however, this is a little difficult without being very wordy.

zygoloid:
The wording changes here are substantial; let's have CWG give them a look over before we apply them.

tkoeppe:
Could you please rebase?

sdkrystian:
I'll have it rebased tomorrow

PR #3873, opened 2020-03-14
[index] Use LaTeX package 'imakeidx' to allow for more indices.

jensmaurer:
This doesn't consume a \write stream per index.
The disadvantage is that it needs the `-shell-escape` option for pdflatex to split up the single index file.

We can now easily add a separate index of exposition-only library names.

"diffpdf" says the before/after PDFs are the same.

### DIFF
--- a/source/back.tex
+++ b/source/back.tex
@@ -77,11 +77,9 @@ are present in this document, with the exceptions described below.\\}
 }
 
 \clearpage
-\renewcommand{\indexname}{Index of grammar productions}
 \renewcommand{\preindexhook}{The first bold page number for each entry is the page in the
 general text where the grammar production is defined. The second bold page number is the
 corresponding page in the Grammar summary\iref{gram}. Other page numbers refer to pages where the grammar production is mentioned in the general text.\\}
-\renewcommand{\leftmark}{\indexname}
 {
 \raggedright
 \printindex[grammarindex]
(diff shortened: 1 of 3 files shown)


tkoeppe:
Instead of this, could we try again to get xindy to work? What was the blocker there, @godbyk?

godbyk:
@tkoeppe If I remember correctly it was an issue the cross-references. Ah, here are [my notes](https://github.com/cplusplus/draft/issues/1917#issuecomment-392093234).

jensmaurer:
@tkoeppe, my understanding is that xindy and this change are complementary. In fact, imakeidx can use xindy instead of makeindex by virtue of a configuration option. However, without imakeidx, our LaTeX index production would still be hampered by the 16 streams `\write` limit, regardless of whether we use makeindex or xindy for the postprocessing.

godbyk:
If the underlying issue is the 16 `\write` limit, does `\usepackage{morewrites}` do the trick? (See the [`morewrites` docs](https://ctan.org/pkg/morewrites) for more details.)

jensmaurer:
"morewrites" is extremely slow. For comparison: "make" after "make clean" with the status quo takes

real    2m56,070s
user    2m50,371s
sys     0m1,175s

Simply adding "morewrites" into std.tex results in this (after a "make clean"):
real    15m32,945s
user    15m2,865s
sys     0m14,112s

That a factor of five. No, thanks.

PR #3876, opened 2020-03-16
[index] Add index of exposition-only library names.

### LINKED ISSUE #3392: 'voidify' is not indexed
It's defined and used in [specialized.algorithms] but not in any index.

The similar exposition-only function _`decay-copy`_ is in the Index of Library Names, although I'm not sure if that's a good idea (most things in that index are reserved by the implementation, so cannot be used as macros by users, but that isn't true of _`voidify`_.

jensmaurer:
and index 'voidify'.

Fixes #3392.

This is based on #3873 and probably needs a rebase after merging the latter.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -10252,11 +10252,11 @@ result in undefined behavior.
 \end{note}
 
 \pnum
-Some algorithms specified in \ref{specialized.algorithms} make use of the exposition-only function
-\tcode{\placeholdernc{voidify}}:
+Some algorithms specified in \ref{specialized.algorithms} make use of
+the exposition-only function \exposidnc{voidify}:
 \begin{codeblock}
 template<class T>
-  constexpr void* @\placeholdernc{voidify}@(T& obj) noexcept {
+  constexpr void* @\libexposidnc{voidify}@(T& obj) noexcept {
     return const_cast<void*>(static_cast<const volatile void*>(addressof(obj)));
   }
 \end{codeblock}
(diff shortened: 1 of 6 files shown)


jensmaurer:
the index of exposition-only names is a bit weird. We have quite a few exposition-only data members that appear in the index, but whose name is not very interesting. And then, there are names such as _STATICALLY-WIDEN_ or _voidify_ or _is-integer-like_ that are of more global interest. Maybe we should add those (few) names that are of global interest to the general library index (probably with an explicit marker "exposition-only" and in italics) and not add a separate index.

zygoloid:
I agree that many of these names don't seem useful to index. I suppose the question is which index they should go in: the general index or the index of library names. They're not library names, but people might look in that index anyway.

jensmaurer:
Editorial meeting: Have exposition-only concepts in the concept index (status quo). Have other exposition-only names (e.g. voidify) in the general library names index. Add intro text to say that italics names are exposition-only.

tkoeppe:
@jensmaurer: Will you retool this PR accordingly (no new index, just integrate names into existing indexes)?

jensmaurer:
@tkoeppe: Sure, this is my pull requst, and I got my marching orders.

jensmaurer:
Fixed.

jensmaurer:
@tkoeppe, I think this is now ready.

tkoeppe:
Thanks, let me just check this.

tkoeppe:
The meeting decision included, "Add intro text to say that italics names are exposition-only." Did you add any such text, or did it ultimately turn out not to be necessary?

tkoeppe:
@jensmaurer: sorry, I lost sight of this; could you take another look to see if this is still ready, and also see my last comment?

tkoeppe:
Oh, we probably need to add the couple of new expos-only names the last meeting added, as well as document this process in the wiki.

jensmaurer:
Note to self: Add intro text saying that italics identifiers are exposition-only.

jensmaurer:
The intro text was already there (intro for the library names index): "Names in italics are exposition-only library names."

@tkoeppe, this seems to be ready.

jensmaurer:
@tkoeppe, ping.

PR #3877, opened 2020-03-16
[dcl.dcl] Improve note regarding nodeclspec-function-declarations

sdkrystian:
The [current note](http://eel.is/c++draft/dcl.dcl#dcl.pre-12.note-1) on *nodeclspec-function-declarations* is a little lacking when it comes to providing useful information as to why the outlined restriction exists. This clarifies it.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -216,8 +216,9 @@ any appropriate initialization\iref{dcl.init} to be done.
 A \grammarterm{nodeclspec-function-declaration} shall declare a
 constructor, destructor, or conversion function.
 \begin{note}
-A \grammarterm{nodeclspec-function-declaration} can only be used in a
-\grammarterm{template-declaration}\iref{temp.pre},
+Because a member function cannot be subject to a non-defining declaration
+outside of a class definition\iref{class.mfct}, a \grammarterm{nodeclspec-function-declaration}
+can only be used in a \grammarterm{template-declaration}\iref{temp.pre},
 \grammarterm{explicit-instantiation}\iref{temp.explicit}, or
 \grammarterm{explicit-specialization}\iref{temp.expl.spec}.
 \end{note}


source/declarations.tex
```diff
@@ -237,8 +237,9 @@
 A \grammarterm{nodeclspec-function-declaration} shall declare a
 constructor, destructor, or conversion function.
 \begin{note}
-A \grammarterm{nodeclspec-function-declaration} can only be used in a
-\grammarterm{template-declaration}\iref{temp.pre},
+Since a member function cannot be subject to a non-defining declaration
```
zygoloid, line 240: Please use "Because" rather than "Since" for relationships that are causal rather than temporal (since can mean either, whereas because is unambiguous).

tkoeppe: @zygoloid: I made the requested change, PTAL?

source/declarations.tex
```diff
@@ -216,8 +216,9 @@
 A \grammarterm{nodeclspec-function-declaration} shall declare a
 constructor, destructor, or conversion function.
 \begin{note}
-A \grammarterm{nodeclspec-function-declaration} can only be used in a
-\grammarterm{template-declaration}\iref{temp.pre},
+Because a member function cannot be subject to a non-defining declaration
+outside of its class definition\iref{class.mfct}, a \grammarterm{nodeclspec-function-declaration}
```
zygoloid, line 220: Hm, the first part of this is not strictly true: a member function can also be subject to a non-defining friend declaration. But a nodeclspec-function-declaration can't appear at class scope and in any case `friend` is a decl-specifier.  Perhaps:

> Because a non-defining declaration of a member function can only appear at class scope, a nodeclspec-function-declaration can only be used in a ...

zygoloid: Or replace "its class definition" with "a class definition"?

tkoeppe: That's simpler, PTAL?

tkoeppe: Going with the simpler solution.

PR #3880, opened 2020-03-17
\defnc and \defna

opensdh:
New macros to handle the common case where the displayed defined term is a child or parent of another in the index.  Each has about the same number of uses; I found them to make the actual prose in the source more readable (since the other `{foo}` can be read as a kind of gloss), but it's up to edit@ to decide whether it's worth having more macro names.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -833,7 +833,7 @@ namespace; these restrictions apply to both regions.
 \pnum
 For a given declarative region \placeholder{R}
 and a point \placeholder{P} outside \placeholder{R},
-the set of \defnx{intervening}{region!declarative!intervening} declarative regions
+the set of \defna{intervening}{region!declarative} declarative regions
 between \placeholder{P} and \placeholder{R}
 comprises all declarative regions
 that are or enclose \placeholder{R} and do not enclose \placeholder{P}.
@@ -4246,8 +4246,8 @@ set of values, which may be empty.
 Every alignment value shall be a non-negative integral power of two.
 
 \pnum
-Alignments have an order from \defnx{weaker}{alignment!weaker} to
-\defnx{stronger}{alignment!stronger} or \defnx{stricter}{alignment!stricter} alignments. Stricter
+Alignments have an order from \defna{weaker}{alignment} to
+\defna{stronger}{alignment} or \defna{stricter}{alignment} alignments. Stricter
 alignments have larger alignment values. An address that satisfies an alignment
 requirement also satisfies any weaker valid alignment requirement.
 
@@ -5107,7 +5107,7 @@ This document imposes no requirements on the accuracy of
 floating-point operations; see also~\ref{support.limits}.
 \end{note}
 Integral and floating-point types are collectively
-called \defnx{arithmetic}{type!arithmetic} types.
+called \defna{arithmetic}{type} types.
 \indextext{\idxcode{numeric_limits}!specializations for arithmetic types}%
 Specializations of the standard library template
 \tcode{std::numeric_limits}\iref{support.limits} shall specify the
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -2863,7 +2863,7 @@ int k();            // error: matches \#4
 \end{example}
 As a consequence of these rules,
 all declarations of an entity are attached to the same module;
-the entity is said to be \defnx{attached}{attached!entity} to that module.
+the entity is said to be \defnc{attached}{entity} to that module.
 
 \pnum
 \indextext{consistency!type declaration}%
@@ -2934,7 +2934,7 @@ explicit specializations of it might be usable in other translation units.
 \end{note}
 
 \pnum
-An entity is \defnx{TU-local}{TU-local!entity} if it is
+An entity is \defnc{TU-local}{entity} if it is
 \begin{itemize}
 \item
 a type, function, variable, or template that
@@ -2966,7 +2966,7 @@ The specialization might have been implicitly or explicitly instantiated.
 \end{itemize}
 
 \pnum
-A value or object is \defnx{TU-local}{TU-local!value or object} if either
+A value or object is \defnc{TU-local}{value or object} if either
 \begin{itemize}
 \item
 it is, or is a pointer to,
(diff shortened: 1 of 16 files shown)


opensdh:
@zygoloid, @jensmaurer: Please review/judge.

tkoeppe:
I'd only mildly wonder about the name "defnc", which sounds like "def but no italics correction" (given how we use "nc" elsewhere (or maybe a "non-copying def", if you look elsewhere yet). But that's fairly minor.

opensdh:
@tkoeppe: I used the short names mostly to parallel the `\defnx`; we could obviously use `\defnctx` and still have the usage be much shorter than the status quo.

tkoeppe:
I understand why you did it; I just wanted to point out that this is increasinlgy crowding the space of "nc" in names.

We can also consider renaming "defnadj" first if that name is better suited to the proposed facility. But that one seems to definitlely adjoin some precendent for multiletter "defn"-suffixes!

Note also that we _also_ already have "defncontext" :-)

opensdh:
More ideas: `\defnc` and `\defna` could be `\defnpar` and `\defnsub`?

tkoeppe:
That sounds good!

source/basic.tex
```diff
@@ -833,7 +833,7 @@
 \pnum
 For a given declarative region \placeholder{R}
 and a point \placeholder{P} outside \placeholder{R},
-the set of \defnx{intervening}{region!declarative!intervening} declarative regions
+the set of \defna{intervening}{region!declarative} declarative regions
```
zygoloid: Can this be `\defnadjx{intervening}{declarative region}{region!declarative}`? Similar question for various other uses of `\defna` below.

source/basic.tex
```diff
@@ -2863,7 +2863,7 @@
 \end{example}
 As a consequence of these rules,
 all declarations of an entity are attached to the same module;
-the entity is said to be \defnx{attached}{attached!entity} to that module.
+the entity is said to be \defnc{attached}{entity} to that module.
```
zygoloid: Maybe `\defnctx`; this is really the same thing as `\defncontext` but for an inline definition rather than a Terms and Definitions definition, so using a similar macro seems reasonable.

source/basic.tex
opensdh: Can you use `!` in the third argument to `\defnadjx` (_i.e._, in the `see{}` in `\indextext`)?

source/basic.tex
opensdh: Well, the syntax is a bit different, but you're right about the parallel.  Do you have a suggestion for `\defna`?

source/basic.tex
zygoloid: I think we'd need some magic to turn `!`s into `, `s. I don't know if `\see` does that for us but it seems relatively straightforward to build.

zygoloid: That said, should we really be indexing "declarative region" under "region" rather than under "declarative region" in the first place?

source/basic.tex
zygoloid: Well... should `\defna{x}{y}` be an alias for `\defnadjx{x}{}{y}`? (I think we probably always want the *see* entry, so I'm inclined to say yes.) If so, perhaps we could use `\defnadjx{x}{}{y}` directly rather than introducing a new macro?

opensdh: `\defnadjx{x}{}{y}` renders as "_x_ ", which would be problematic at least if punctuation follows.  I'm sure we could teach it to not do so; maybe `#1 #2\unskip` would be enough.

source/basic.tex
opensdh: We provide "a b, _see_ b, a" because it's standard to look up "noun, adjective", so maybe people do look for "region, declarative" even though the generic word "region" is not useful here?  (I also haven't checked to see whether the other uses of `\defna` are "misuses" like this.)

zygoloid: Well, we don't have index entries for "binding, structured" or "operand, unevaluated". I think that's somewhat appropriate: these aren't "adjective noun" so much as an atomic noun phrase.

I think the right thing here is:

 * This should be `\defnadj{intervening}{declarative region}`
 * The "region, declarative" index entry should be changed to say "*see* declarative region" so that it points to this index sub-entry

source/basic.tex
zygoloid: Another idea for `\defnctx`: use an optional argument: `\defn[entity]{attached}`. That seems pretty clean. Although now I come to think of it, that might instead be a nice way to get rid of the `\...x` macros instead.

Maybe:
* `\defnx{bodytext}{indextext}` -> `\defn[indextext]{bodytext}`
* `\defnc{bodytext}{indexgroup}` -> `\defnctx[indexgroup]{bodytext}`
* `\defnadjx{bodyadj}{bodynoun}{indexnoun}` -> `\defnadj{bodyadj}[indexnoun]{bodynoun}`
* `\defna{bodyadj}{indexnoun}` -> `\defnadj{bodyadj}[indexnoun]{}`
? (Leaving room for `\defnadj[indexadj]{bodyadj}{bodynoun}` if we ever want that.)

What do you think?

opensdh: In general, that seems like an improvement: fewer magic names and less in the way of argument-order guesswork.  `\newcommand` supports only _leading_ optional arguments, but we could certainly use a raw `\def` for your first `\defnadj`.  We still have to deal with emitting a space only if the `bodynoun` is non-empty.  I don't immediately know how to support optional arguments in two different places, but I'm sure something like `\expandafter` can be used to do it if it ever matters.

source/basic.tex
opensdh: You mean `\defnadjx{intervening}{declarative regions}{declarative region}` for this case, right?  And are you saying that (`\defnadj` and) `\defnadjx` should be changed to _always_ index _adjective noun_ and generate a _see_ for _noun, adjective_?

source/basic.tex
zygoloid: There's a standard trick for an optional argument after a mandatory argument: `\defnadj` would be a two-parameter command with an optional argument that saves its arguments in some scratch space then expands to another two-parameter command with an optional argument.

jensmaurer:
Editorial meeting: Introduce `\defnadj{arithmetic}[type]{types}` and apply it for the cases here.

tkoeppe:
@opensdh I think we're still interested in this!

PR #3881, opened 2020-03-17
Nicer diff support

opensdh:
Add two files that are special to git&mdash;if you **opt into them** with the configuration commands in the commit messages&mdash;and cause `git diff` to provide `[stable.names]` and present clauses in order.

Note that I've seen a case where a command run from inside `.git` complained about failing to find `.gitorder`, but now I can't reproduce it (with a different Git version)?

@zygoloid: Is this sufficiently useful to promulgate despite each user having to configure for it?


zygoloid:
Seems like a couple of nice ideas.

Would it be reasonable to also check in a `.gitconfig` file containing the two relevant git settings (and any others we come up with later) or a shell script that configures them or similar?

opensdh:
> Would it be reasonable to also check in a `.gitconfig` file containing the two relevant git settings (and any others we come up with later) or a shell script that configures them or similar?

What would the user do with such a `.gitconfig` file?  Append it to some other file?  By design, Git doesn't consult any working copy file for configuration variables.  The shell script could certainly work; where would you want it?

zygoloid:
> What would the user do with such a `.gitconfig` file?

Either append it to `.git/config` or -- if they trust us -- `git config --local include.path ../.gitconfig`.

> The shell script could certainly work; where would you want it?

`tools/gitconfig.sh` maybe?

opensdh:
Presumably the shell script is better: it explicitly depends on trust (and doesn't ask for permission for silent changes in the future) and can easily be idempotent (even though certain _changes_ to it may require user intervention).

PR #3882, opened 2020-03-17
[temp.names] Do not use "template-name" for "name referring to template"

hubert-reinterpretcast:
From core/2020/03/8668.

The structure of [temp.names]/2 – 3 now (following P0846R0, "ADL and Function Templates that are not Visible"):

- Uses _template-name_ basically to handle only the class template cases, and
- introduces the notion that names can be considered to refer to templates.

However, paragraph 3 is prefaced by "[when] a name is **considered to be a _template-name_** [emphasis mine]", which differs from "when a name is **considered to refer to a template**". In particular, it is a stretch to take a name not having the syntactic form of a _template-name_ (such as an _operator-function-id_) as being considered to be a _template-name_ merely because the name is considered to refer to a template.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -722,8 +722,7 @@ and name lookup either finds one or more functions or finds nothing.
 
 \pnum
 \indextext{\idxcode{<}!template and}%
-When a name is considered to be a
-\grammarterm{template-name},
+When a name is considered to refer to a template,
 and it is followed by a \tcode{<},
 the \tcode{<}
 is always taken as the delimiter of a


zygoloid:
The wording in this area for template names that are not *template-name*s is ambiguous and there is implementation variance; see core/2019/05/6251 and replies. Note also that the previous paragraph talks only about the *template-name* case and not the other forms of template names. I don't think it's clear whether the committee intended for P0846R0 to apply to *operator-function-id*s and *literal-operator-id*s.

jensmaurer:
This has a stray commit adding the foreword. Please rebase + force-push.

hubert-reinterpretcast:
> This has a stray commit adding the foreword. Please rebase + force-push.

Done.

zygoloid:
We addressed this in a different way in 84087b3d3. We now define whether an *identifier* is grammatically a *template-name*, rather than whether "a name is considered to refer to a template", so this wording has become meaningful.

PR #3886, opened 2020-03-17
Non-normative cleanup

opensdh:
Changes to notes, examples, and LaTeX comments resulting from DIS review.  Note that just the first line of the big note that moved is also changed.

@zygoloid: please review and/or delegate.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -4008,6 +4008,59 @@ void m() {
 \end{example}
 
 \pnum
+\begin{note}
+Since, in a call context, such type deduction considers only parameters
+for which there are explicit call arguments, some parameters are ignored (namely,
+function parameter packs, parameters with default arguments, and ellipsis
+parameters).
+\begin{example}
+\begin{codeblock}
+template<class T> void f(T);                            // \#1
+template<class T> void f(T*, int=1);                    // \#2
+template<class T> void g(T);                            // \#3
+template<class T> void g(T*, ...);                      // \#4
+
+\end{codeblock}
+\begin{codeblock}
+int main() {
+  int* ip;
+  f(ip);                                                // calls \#2
+  g(ip);                                                // calls \#4
+}
+\end{codeblock}
+\end{example}
+\begin{example}
+\begin{codeblock}
+template<class T, class U> struct A { };
+
+template<class T, class U> void f(U, A<U, T>* p = 0);   // \#1
+template<         class U> void f(U, A<U, U>* p = 0);   // \#2
+template<class T         > void g(T, T = T());          // \#3
+template<class T, class... U> void g(T, U ...);         // \#4
+
+void h() {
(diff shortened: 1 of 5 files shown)


source/templates.tex
```diff
@@ -6492,7 +6492,7 @@
 when determining whether the constraints are satisfied
 or as specified in \ref{temp.constr.decl} when comparing declarations.
 \begin{note}
-The satisfaction of constraints is determined during name lookup or overload
+The satisfaction of constraints is determined during overload
```
zygoloid: I think technically the satisfaction of constraints *is* determined during name lookup for an *conversion-function-id*.

opensdh: I don't think I follow (it doesn't help that we don't really describe name lookup for a *conversion-function-id* beyond hints like [basic.lookup.classref]/7).  Are you saying that
```cpp
struct A {operator int();};
template<int=0> struct B : A {operator int() requires false;};
int f(B<> b) {return b;}
```
should work because name lookup discovers that `B<0>::operator int` is non-viable and proceeds to the base class?

zygoloid: No, but
```
template<typename> struct A {
  template<typename T> requires (sizeof(T) == 1) operator T();
};
auto p = &A<int>::operator char;
```
... checks that the constraints of `A<int>::operator T<char>` are satisfied during name lookup.

opensdh: I'm sorry, I'm just still not getting it.  (Maybe P1787 has destroyed my ability to understand name lookup?)

My understanding there is that `A` being a template is irrelevant, that name lookup performs deduction and identifies `A<int>::operator T<char>` as the correct specialization to add to the (otherwise empty) overload set, and that overload resolution (of the [new kind](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1972r0.html) but without any contribution from [temp.deduct.funcaddr] in this case) _then_ checks the constraints ([over.over]/5) and either accepts or discards the one candidate (the latter, typically, in the situation where `A<int>::operator T<double>` was chosen by name lookup).

zygoloid: Yeah, changing to "deduction and overload resolution" would be an improvement.

I think the difference between doing it during name lookup and deferring to overload resolution is probably only visible in a variant of your example such as

```
struct A { operator int(); };
template<bool B> struct B : A { template<same<int> T> operator T() requires B; };
```

... for which (unlike in your example) I think `B<false>::operator int` *would* select `A::operator int`.

zygoloid: I'll merge this manually with that one change.

opensdh: Thanks for the fix (and the explanation).

zygoloid:
Landed as caffde5ccfcc558d9e6c12c9b64009aa62d22d55 ... 36ac6c50b197b5d1c1fea8d0a2faf15f07bbfab2.

PR #3887, opened 2020-03-17
Normative cleanup

opensdh:
Changes to normative text (that are editorial unless I've misunderstood the wording), mostly resulting from DIS review.

@zygoloid: please review and/or delegate.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -580,7 +580,7 @@ the definitions satisfy the following requirements.
 Given such an entity \tcode{D} defined in more than one translation unit,
 for all definitions of \tcode{D}, or,
 if \tcode{D} is an unnamed enumeration,
-for all definitions of \tcode{D} that are reachable at each program point,
+for all definitions of \tcode{D} that are reachable at any given program point,
 the following requirements shall be satisfied.
 \begin{itemize}
 \item Each such definition of \tcode{D}
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3327,7 +3327,7 @@ that operation implicitly creates and starts the lifetime of
 zero or more objects of implicit-lifetime types\iref{basic.types}
 in its specified region of storage
 if doing so would result in the program having defined behavior.
-If no such sets of objects would give the program defined behavior,
+If no such set of objects would give the program defined behavior,
 the behavior of the program is undefined.
 If multiple such sets of objects would give the program defined behavior,
 it is unspecified which such set of objects is created.
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -575,7 +575,7 @@ There can be more than one definition of a
 \item default template argument\iref{temp.param}
 \end{itemize}
 in a program provided that
-each definition appears in a different translation unit, and
+each definition appears in a different translation unit and
 the definitions satisfy the following requirements.
 Given such an entity \tcode{D} defined in more than one translation unit,
 for all definitions of \tcode{D}, or,
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -583,14 +583,14 @@ if \tcode{D} is an unnamed enumeration,
 for all definitions of \tcode{D} that are reachable at any given program point,
 the following requirements shall be satisfied.
 \begin{itemize}
-\item Each such definition of \tcode{D}
+\item Each such definition
 shall not be attached to a named module\iref{module.unit}.
-\item Each such definition of \tcode{D} shall consist of
+\item Each such definition shall consist of
 the same sequence of tokens,
 where the definition of a closure type
 is considered to consist of the sequence of tokens of
 the corresponding \grammarterm{lambda-expression}.
-\item In each such definition of \tcode{D}, corresponding names, looked up
+\item In each such definition, corresponding names, looked up
 according to~\ref{basic.lookup}, shall refer to the same entity, after
 overload resolution\iref{over.match} and after matching of partial
 template specialization\iref{temp.over}, except that a name can refer to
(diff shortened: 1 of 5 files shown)


source/basic.tex
```diff
@@ -575,22 +575,22 @@
...
 if \tcode{D} is an unnamed enumeration,
-for all definitions of \tcode{D} that are reachable at each program point,
+for all definitions of \tcode{D} that are reachable at any given program point,
 the following requirements shall be satisfied.
 \begin{itemize}
-\item Each such definition of \tcode{D}
-shall not be attached to a named module\iref{module.unit}.
-\item Each such definition of \tcode{D} shall consist of
+\item Each such definition
+shall be attached to the global module\iref{module.unit}.
```
zygoloid: Default [template] arguments aren't attached to any module, because they aren't entities, despite the introductory lie "Given such an entity `D`". To the extent that this is a normative change (which, given said lie, is debatable), I think it disallows repeating default [template] arguments in any context, which seems undesirable.

source/modules.tex
```diff
@@ -554,8 +554,7 @@
 
 \pnum
 A translation unit has an \defn{interface dependency} on a translation unit \tcode{U}
-if it contains a \grammarterm{module-declaration} or
-\grammarterm{module-import-declaration} that imports \tcode{U} or if it has
+if it imports \tcode{U} or if it has
```
zygoloid: I don't think we define what it means for a translation unit to import a translation unit, do we?

source/preprocessor.tex
```diff
@@ -222,9 +222,8 @@
 \pnum
 At the start of phase 4 of translation,
 the \grammarterm{group} of a \grammarterm{pp-global-module-fragment} shall
-neither contain a \grammarterm{control-line}
-not starting with a \tcode{\#} preprocessing token,
-nor contain a \grammarterm{text-line}.
+contain neither a \grammarterm{text-line} nor a \grammarterm{control-line}
+not starting with a \tcode{\#} preprocessing token.
```
zygoloid: The new wording has ambiguous grouping: to what does the "not starting with" attach? (Now, I know that it doesn't matter, because a *text-line* can't start with a `#`, but we should still try to avoid ambiguities even if they don't matter.)

source/basic.tex
opensdh: Maybe we can rectify (or at least reduce) the lie and clarify this bullet at the same time?  Something like "Given such an entity or default (template) argument _D_" and then change the bullet in question here to "If _D_ is an entity, each such definition&hellip;".

source/modules.tex
opensdh: [basic.scope.namespace]/2 and the note in [module.import]/2 seem to think it's meaningful, and I took (say) "a translation unit defines" as being the natural converse of common phrases like "defined in a translation unit".  If you prefer, I'll add at least a parenthetical definition to keep this phrasing in general&mdash;it gave me the strong impression that there was some other way to import a translation unit that didn't constitute an interface dependency.

source/preprocessor.tex
opensdh: Wouldn't it have to be "neither X, nor Y, not starting with Z" to be ambiguous?

source/basic.tex
zygoloid: This takes us into problematic grounds: an editorial change that moves from "wording is ambiguous and intent is unclear" to "wording is unambiguous and we have no idea if it matches the intent". In particular: should default (template) arguments be repeatable in multiple TUs if the entity to whom they are providing a default is attached to a named module?

I think we should move this to a core issue. (Obviously the removing the unnecessary repetitive "of `D`" and other cleanups are fine to do editorially.)

source/modules.tex
zygoloid: I think [basic.scope.namespace]/2 could do with a little cleanup:

> If a translation unit *Q* is imported <del>into</del> <ins>by a declaration in</ins> a translation unit *R*, ...

but the rest of the sentence already makes it contextually clear what it's referring to ("the first *module-import-declaration* or *module-declaration* in *R* that imports *Q*").

[module.import]/2's note is imprecise, but it's non-normative and seems clear enough in that context ("the importing translation unit after the import declaration" suggests that "the importing translation unit" is the one that contains the import declaration). I would prefer that we keep this normative wording precise.

I'd be OK with adding a definition for what it means for a translation unit to import another translation unit and taking this change. I'm not sure we'd use the definition very much, since we typically do care about the point of import.

source/preprocessor.tex
zygoloid: I don't think the lack of commas disambiguates it sufficiently. I find the revised sentence substantially harder to read than the original. How about this:

> At the start of phase 4 of translation, the *group* of a *pp-global-module-fragment* shall not contain a *text-line* nor a *pp-import*.

source/basic.tex
opensdh: I think the "current intent" (_i.e._, old status quo plus what has been _discussed_ and voted on) is clear that such default arguments can be repeated, but I'm glad to leave it to a Core issue to decide whether that intent is bad.  The next revision here will do that.

source/modules.tex
opensdh: How about

> A translation unit has an _interface dependency_ on a translation unit _U_ if it contains a <ins>declaration (possibly a</ins> *module-declaration*<del> or *module-import-declaration*</del><ins>)</ins> that imports _U_ or if &hellip;

?  I just want to avoid the suggestion that there are other ways of importing that don't count for this.

source/preprocessor.tex
opensdh: Good; I've done that.  If we ever add more non-`#` directives, we can introduce a collective grammar term for them.

source/modules.tex
zygoloid: Sounds fine to me.

opensdh: Done.

opensdh: I finally found the other place I remembered seeing this: [module.reach]/1, which _does_ use it with a point of import.

PR #3894, opened 2020-03-24
[dcl.init, dcl.stc] Move specification as to where extern is allowed into [dcl.stc]

sdkrystian:
The current specification lives in [dcl.init], but it fits better in the specification of `extern`, since the only way a local variable can acquire external or internal linkage is through the use of `extern`, and it's better to specify such broad restrictions in one place.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -406,7 +406,8 @@ see~\ref{basic.link}.
 \indextext{restriction!\idxcode{extern}}%
 The \tcode{extern} specifier shall be applied only to the declaration of a variable
 or function. The \tcode{extern} specifier shall not be used in the
-declaration of a class member or function parameter.
+declaration of a class member, function parameter, or block-scope variable with
+an \grammarterm{initializer}.
 \indextext{\idxcode{extern}!linkage of}%
 \indextext{consistency!linkage}%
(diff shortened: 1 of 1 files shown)


source/declarations.tex
```diff
@@ -406,7 +406,8 @@
 \indextext{restriction!\idxcode{extern}}%
 The \tcode{extern} specifier shall be applied only to the declaration of a variable
 or function. The \tcode{extern} specifier shall not be used in the
-declaration of a class member or function parameter.
+declaration of a class member, function parameter, or local variable with
```
jensmaurer: local -> block-scope
is probably clearer

source/declarations.tex
```diff
@@ -406,7 +406,8 @@
 \indextext{restriction!\idxcode{extern}}%
 The \tcode{extern} specifier shall be applied only to the declaration of a variable
 or function. The \tcode{extern} specifier shall not be used in the
-declaration of a class member or function parameter.
+declaration of a class member, function parameter, or block-scope variable with
+an \grammarterm{initializer}.
```
zygoloid, line 410: This change makes it sound like it's OK to use `extern` with a class member or function parameter so long as there's no initializer. Moving the "block-scope variable with an initializer" case to the start of the list would help.

PR #3895, opened 2020-03-24
Add Foreword

### LINKED ISSUE #3668: Produce foreword with list of changes for C++20
@tkoeppe to produce a list of changes for C++20.

Then, Richard will produce the foreword.

zygoloid:
The general text was provided by the ISO CS. A list of incorporated
technical specifications was derived from N4824. A list of technical
changes since C++17 was derived from P2131R0.

Fixes #3668.

### DIFF
--- a/source/preface.tex
+++ b/source/preface.tex
@@ -86,8 +86,8 @@ that are not reached during constant evaluation
 \item addition of \keyword{using} \keyword{enum} declaration
 \item addition of \keyword{char8_t} type
 \item guarantee that \keyword{char16_t} and \keyword{char32_t} literals
-are encoded as UTF-16 and UTF-32 respectively
-\item support for an initializer statement in range-based for loops
+are encoded as UTF-16 and UTF-32, respectively
+\item support for an initializer statement in range-based \tcode{for} statements
 \item support for default member initializers for bit-fields
 \item support for parenthesized aggregate initialization
 \item extended support for lambda expressions, including
@@ -133,7 +133,7 @@ waiting, notifications, semaphores, latches, and barriers, and
 \item addition of \tcode{unsequenced} execution policy
 \item new utility functions, types, and templates in the standard library, including
 \begin{itemize}
-  \item \raggedright new type traits
+  \item \raggedright type traits
     \tcode{is_bounded_array},
     \tcode{is_corresponding_member},
     \tcode{is_layout_compatible},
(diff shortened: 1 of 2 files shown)


source/preface.tex
```diff
@@ -0,0 +1,184 @@
...
+
+This sixth edition cancels and replaces the fifth edition
+(ISO/IEC 14882:2017), which has been technically revised.
+It also incorporates the Technical Specifications
+\begin{itemize}
+\item ISO/IEC TS 19217:2015 \textit{Programming Languages --- \Cpp{} Extensions for concepts},
+\item ISO/IEC TS 21425:2017 \textit{Programming Languages --- \Cpp{} Extensions for ranges},
+\item ISO/IEC TS 22277:2017 \textit{Programming Languages --- \Cpp{} Extensions for Coroutines},
+\item ISO/IEC TS 21544:2018 \textit{Programming Languages --- Extensions to \Cpp{} for Modules},
+\item portions of ISO/IEC TS 19571:2016 \textit{Programming Languages --- Technical specification for \Cpp{} extensions for concurrency}, and % latch and barrier
```
tkoeppe: Why does this entry get an explanatory comment, and the next one does not?

source/preface.tex
```diff
@@ -0,0 +1,184 @@
...
+  \item uninitialized variables
+\end{itemize}
+\item support for constexpr functions to contain
+\keyword{catch} handlers and \grammarterm{asm-declaration}s
+that are not reached during constant evaluation
+\item support for controlling destruction in a class-specific operator delete function
+\item addition of \keyword{using} \keyword{enum} declaration
+\item addition of \keyword{char8_t} type
+\item guarantee that \keyword{char16_t} and \keyword{char32_t} literals
+are encoded as UTF-16 and UTF-32 respectively
```
tkoeppe: Should there be a comma before "respectively"?

source/preface.tex
```diff
@@ -0,0 +1,184 @@
...
+\item support for an initializer statement in range-based for loops
```
tkoeppe: Missing code font formatting for for formulation

source/preface.tex
```diff
@@ -0,0 +1,184 @@
...
+    \tcode{is_nothrow_convertible},
+    \tcode{is_pointer_interconvertible_base_of},
+    \tcode{is_pointer_interconvertible_with_class}, and
+    \tcode{is_unbounded_array}
+  \item new transformation traits
+    \tcode{remove_cvref},
+    \tcode{type_identity},
+    \tcode{unwrap_ref_decay}, and
+    \tcode{unwrap_reference}
+  \item new standard library primitive functions \tcode{assume_aligned} and
```
tkoeppe: (Is "primitive" a good word for this?)

zygoloid:
> I thought we already had this earlier?

Some preliminary review was done (but those were comments left on a random commit, which github is not well-suited to preserving). Now I'm asking for signoff on the result of addressing those comments.

source/preface.tex
zygoloid: Only because I found it non-obvious whether we took anything from this one, and easier to determine we took things from Lib Fundamentals.

source/preface.tex
zygoloid: OK. Also loops -> statements, per [stmt.ranged].

source/preface.tex
zygoloid: Ehh. I was thinking "things that can't be implemented in the language", but perhaps that's just not an interesting distinction to be drawing here. Simplified.

source/preface.tex
zygoloid: Added.

source/preface.tex
tkoeppe: OK :-)

PR #3899, opened 2020-03-25
[expr.sub] Really deprecate comma in braces

AlisdairM:
The core convention is to deprecate in core wording, and then
cross-reference from Annex D.  If all references to deprecating
a feature are notes, then there is no normative deprecation.
Resolved by promoting the two notes in core text to straight
normative text proclaiming the deprecation.


tkoeppe:
I applied the (fixed) change via 5a9f575feb08fe2784ad0df2af648051a882831d.

PR #3900, opened 2020-03-25
[temp.deduct.call] Clear up wording regarding function parameter packs appearing in non-deduced contexts

sdkrystian:
[CWG 1388](http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#1388) and [CWG 1399](http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#1399) were both resolved by the [resolution for CWG 1388](http://wiki.edg.com/pub/Wg21kona2012/CoreWorkingGroup/proposed_resolution_core-1388.html), however the wording doesn't account for patterns that don't solely consist of a template parameter pack, or include other template parameters that are otherwise deductible. This PR clears this up, and explicitly specifies that such template parameter pack are never deduced, even if they appear in a deduced context elsewhere.

This is borderline editorial, but I believe the defects and accompanying notes on the wiki sufficiently express the intended meaning here.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -7881,8 +7881,8 @@ as the corresponding function template parameter type.
 Each deduction deduces template arguments for subsequent positions in
 the template parameter packs expanded by the function parameter pack.
 When a function parameter pack appears in a non-deduced
-context\iref{temp.deduct.type}, the type of that pack is
-never deduced.
+context\iref{temp.deduct.type}, the template parameter pack(s)
+expanded by that pack are never deduced, even if they appear in a deduced context elsewhere.
 \begin{example}
 \begin{codeblock}
 template<class ... Types> void f(Types& ...);


sdkrystian:
Here's an example: https://godbolt.org/z/vMgAZL

source/templates.tex
```diff
@@ -7881,8 +7881,8 @@
 Each deduction deduces template arguments for subsequent positions in
 the template parameter packs expanded by the function parameter pack.
 When a function parameter pack appears in a non-deduced
-context\iref{temp.deduct.type}, the type of that pack is
-never deduced.
+context\iref{temp.deduct.type}, the template parameter pack(s)
+expanded by that pack are never deduced, even if they appear in a deduced context elsewhere.
```
zygoloid, line 7885: The new wording seems confusing: packs are expanded by pack expansions, not by packs. Replacing "that pack" with "that function parameter pack" would help. Also, "(s)" -> "s", please.

Is it clear that the "even if" part is the intent? This seems ambiguous.

zygoloid: ... not only ambiguous, but inconsistent with unanimous implementation practice: https://godbolt.org/z/78TjoY

xmh0511: IMO, it seems the current wording is suitable for either case. 
> If a template parameter is used only in non-deduced contexts and is not explicitly specified, template argument deduction fails.  
>
> When a function parameter pack appears in a non-deduced context ([temp.deduct.type]), the type of that pack is never deduced.  

1. Template parameter that appears in non-deduced context does not participate in template argument deduction 
2. These template parameters that **only** appear in non-deduced context should be explicitly specified their template arguments, otherwise, the deduction fails. 

In addition, except that [temp.deduct.call#1] explicitly says that
> When a function parameter pack appears in a non-deduced context ([temp.deduct.type]), the type of that pack is **never deduced**.

It seems that the other kinds of argument deduction do not strong this concept, that is  
> any template parameter that appears in a non-deduced context is never deduced

PR #3901, opened 2020-03-25
[dcl.stc] Define mutable subobject and make use of keyword consistent

sdkrystian:
We use the term "mutable subobject" to refer to objects corresponding to members declared `mutable`, but we don't define this anywhere. This defines it in [dcl.stc], and additionally changes adds \tcode to "mutable" when referring to a member declared `mutable`, since it's a keyword in those contexts (this is used for the majority of references to members declared `mutable`).

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -483,7 +483,7 @@ The \keyword{mutable} specifier shall appear only in the declaration of
 a non-static data member\iref{class.mem}
 whose type is neither const-qualified nor a reference type.
 A member subobject\iref{intro.object} corresponding to a non-static
-data member declared with the \keyword{mutable} specifier is a \term{mutable subobject}.
+data member declared with the \keyword{mutable} specifier is a \defnadj{mutable}{subobject}.
 \begin{example}
 \begin{codeblock}
 class X {
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -482,6 +482,8 @@ void h() {
 The \tcode{mutable} specifier shall appear only in the declaration of
 a non-static data member\iref{class.mem}
 whose type is neither const-qualified nor a reference type.
+A member subobject\iref{intro.object} corresponding to a non-static
+data member declared with the \tcode{mutable} specifier is a \term{mutable subobject}.
 \begin{example}
 \begin{codeblock}
 class X {
@@ -495,7 +497,7 @@ class X {
 \begin{note}
 The \tcode{mutable} specifier on a class data member nullifies a
 \tcode{const} specifier applied to the containing class object and
-permits modification of the mutable class member even though the rest of
+permits modification of the \tcode{mutable} class member even though the rest of
 the object is const~(\ref{basic.type.qualifier}, \ref{dcl.type.cv}).
 \end{note}
 
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -479,11 +479,11 @@ void h() {
 \end{example}
 
 \pnum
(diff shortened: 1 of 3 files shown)


jensmaurer:
Please use \keyword, not \tcode, when referring to keywords. That also helps the index.

source/declarations.tex
```diff
@@ -479,9 +479,11 @@
 \end{example}
 
 \pnum
-The \tcode{mutable} specifier shall appear only in the declaration of
+The \keyword{mutable} specifier shall appear only in the declaration of
 a non-static data member\iref{class.mem}
 whose type is neither const-qualified nor a reference type.
+A member subobject\iref{intro.object} corresponding to a non-static
+data member declared with the \keyword{mutable} specifier is a \term{mutable subobject}.
```
jensmaurer: \term is bad. Use \defn and friends. I think we want \defnadj{mutable}{subobject}.

source/declarations.tex
```diff
@@ -495,9 +495,9 @@
 
 \pnum
 \begin{note}
-The \tcode{mutable} specifier on a class data member nullifies a
+The \keyword{mutable} specifier on a class data member nullifies a
```
sdkrystian: Hm, I don't think I should be using \keyword in non-normative text, since it adds to the index. @jensmaurer can you confirm?

jensmaurer: What's wrong with using it in non-normative text (and thus getting an index entry)? If we ever color the standard, we definitely want keywords to show up like keywords everywhere. If we feel the index becomes less usable, we should find a way to highlight the "definition" of a keyword.

source/expressions.tex
```diff
@@ -6938,7 +6938,7 @@
 \item
 an invocation of an implicitly-defined copy/move constructor or
 copy/move assignment operator
-for a union whose active member (if any) is mutable,
+for a union whose active member (if any) is \keyword{mutable},
```
jensmaurer: We say "const" (no special decoration), so I think we shouldn't make this case a \keyword.

jensmaurer (commented):
My view: Take the first change (\keyword + definition of "mutable subobject") and leave the rest alone.

source/expressions.tex
sdkrystian: The motivation for this was to keep the distinction between objects and members (i,e. a `mutable` member corresponds to a *mutable subobject*).

source/declarations.tex
sdkrystian: Good point.

source/declarations.tex
```diff
@@ -479,9 +479,11 @@
 \end{example}
 
 \pnum
-The \tcode{mutable} specifier shall appear only in the declaration of
+The \keyword{mutable} specifier shall appear only in the declaration of
 a non-static data member\iref{class.mem}
 whose type is neither const-qualified nor a reference type.
+A member subobject\iref{intro.object} corresponding to a non-static
+data member declared with the \keyword{mutable} specifier is a \defnadj{mutable}{subobject}.
```
zygoloid, line 486: Should "mutable subobject" be transitive? That is, should a (non-const) subobject of a mutable subobject also be a mutable subobject? The use of "mutable subobject" in [basic.type.qualifier]/1.1 in the definition of "const object" appears to assume transitivity -- but maybe it means "or a non-mutable <ins>direct</ins> subobject of a const object"

zygoloid:
I would prefer that we define what it means for a non-static data member to be "mutable" (no `\keyword`) rather than talking about `\keyword{mutable}` members.

PR #3907, opened 2020-03-25
[class.union.anon] Turn redundant wording into a note

sdkrystian:
[[class.union.anon] p1](http://eel.is/c++draft/class.union.anon#1) already specifies that a declaration of a union with declarators isn't an anonymous union, so p3 can be turned into a note. Additionally, objects of pointer type are objects, so "pointers" in p3 can be removed.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -3317,6 +3317,7 @@ members\iref{class.access}. An anonymous union shall not have
 member functions.
 
 \pnum
+\begin{note}
 A union for which objects, pointers, or references are declared is not an anonymous union.
 \begin{example}
 \begin{codeblock}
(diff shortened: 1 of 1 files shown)


source/classes.tex
```diff
@@ -3317,7 +3317,8 @@
 member functions.
 
 \pnum
-A union for which objects, pointers, or references are declared is not an anonymous union.
+\begin{note}
```
jensmaurer: Regarding objects vs. pointers: I think we should keep this as-is, because "for which ... are declared" is a bit hazy. What does "for which" mean, exactly? So, let's keep the hand-waving. In particular since this is a note now, anyway.

PR #3908, opened 2020-03-25
Correctly annotate keywords

sdkrystian:
Fixed a few instances of keywords not being properly annotated.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -824,7 +824,7 @@ which the declaration occurs. In particular,
 \grammarterm{elaborated-type-specifier}{s}\iref{dcl.type.elab} and
 friend declarations\iref{class.friend} may introduce a (possibly not
 visible) name into an enclosing namespace; these restrictions apply to
-that region. Local extern declarations\iref{basic.link} may introduce
+that region. Local \tcode{extern} declarations\iref{basic.link} may introduce
 a name into the declarative region where the declaration appears and
 also introduce a (possibly not visible) name into an enclosing
 namespace; these restrictions apply to both regions.
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -824,9 +824,10 @@ which the declaration occurs. In particular,
 \grammarterm{elaborated-type-specifier}{s}\iref{dcl.type.elab} and
 friend declarations\iref{class.friend} may introduce a (possibly not
 visible) name into an enclosing namespace; these restrictions apply to
-that region. Local \tcode{extern} declarations\iref{basic.link} may introduce
-a name into the declarative region where the declaration appears and
-also introduce a (possibly not visible) name into an enclosing
+that region.
+Function and \tcode{extern} variable declarations at block scope
+may introduce names into the declarative region where the declaration appears
+and also introduce a (possibly hidden) name into an enclosing
 namespace; these restrictions apply to both regions.
 \end{note}
 
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -1969,9 +1969,9 @@ void foo() {
 }
 \end{codeblock}
 
-ISO C will accept this usage of pointer to \tcode{void} being assigned
+ISO C accepts this usage of pointer to \tcode{void} being assigned
 to a pointer to object type.
-\Cpp{} will not.
+\Cpp{} does not.
 \rationale
 \Cpp{} tries harder than C to enforce compile-time type safety.
 \effect
@@ -2104,7 +2104,7 @@ Besides,
 promising to return a value of a given type, and then not returning
 such a value, has always been recognized to be a questionable
 practice, tolerated only because very-old C had no distinction between
-\tcode{void} functions and \tcode{int} functions.
+functions with \tcode{void} and \tcode{int} return types.
 \effect
 Deletion of semantically well-defined feature.
 \difficulty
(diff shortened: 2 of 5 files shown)


source/compatibility.tex
```diff
@@ -2104,7 +2104,7 @@
 promising to return a value of a given type, and then not returning
 such a value, has always been recognized to be a questionable
 practice, tolerated only because very-old C had no distinction between
-void  functions and  int  functions.
+\tcode{void} functions and \tcode{int} functions.
```
zygoloid: Let's fix this properly while we're here: "functions with \tcode{void} and \tcode{int} return types"

source/basic.tex
```diff
@@ -824,7 +824,7 @@
 \grammarterm{elaborated-type-specifier}{s}\iref{dcl.type.elab} and
 friend declarations\iref{class.friend} may introduce a (possibly not
 visible) name into an enclosing namespace; these restrictions apply to
-that region. Local extern declarations\iref{basic.link} may introduce
+that region. Local \tcode{extern} declarations\iref{basic.link} may introduce
```
zygoloid: The intent is to also cover block-scope function declarations, and "local" by definition excludes non-automatic storage duration entities, so this is still wrong :)

How about "Function and \tcode{extern} variable declarations at block scope\iref{basic.link} may introduce [...]".

source/compatibility.tex
```diff
@@ -1969,7 +1969,7 @@
 }
 \end{codeblock}
 
-ISO C will accept this usage of pointer to void being assigned
+ISO C will accept this usage of pointer to \tcode{void} being assigned
 to a pointer to object type.
 \Cpp{} will not.
```
zygoloid: "will accept" -> "accepts", "will not" -> "does not", while we're fixing things here.

source/compatibility.tex
```diff
@@ -2225,13 +2225,13 @@
 \end{codeblock}
 
 \rationale
-In \Cpp{}, implicit int creates several opportunities for
+In \Cpp{}, implicit \tcode{int} creates several opportunities for
```
zygoloid: "creates" -> "would create"

source/compatibility.tex
```diff
@@ -2225,13 +2225,13 @@
...
 ambiguity between expressions involving function-like
 casts and declarations.
 Explicit declaration is increasingly considered
 to be proper style.
 Liaison with WG14 (C) indicated support for (at least)
-deprecating implicit int in the next revision of C.
+deprecating implicit \tcode{int} in the next revision of C.
```
zygoloid: I think we can say something more current here; it's been a long time since C was considering deprecation.

@AaronBallman What's the status quo with C17 and C2x? Can you suggest something suitable?

source/locales.tex
```diff
@@ -444,7 +444,7 @@
 components can be instantiated.
 A template parameter with name
 \tcode{International}
-represents the set of all possible specializations on a bool parameter.
+represents the set of all possible specializations on a \tcode{bool} parameter.
```
zygoloid: This sentence (and the ones before) appear to be wrong and confusing. I think what they mean are that the name `C` when appearing as a template *argument* (not parameter) in [tab:locale.spec] means each type in a particular set, and the name `International` means both `true` and `false`. But this is an awful way to express that.

Let's fix this more properly:

1) Use `\exposid` for these placeholder identifiers `C` and `International`
2) Remove paragraph 6 and add to paragraph 4:

> An implementation is required to provide those specializations for facet templates identified as members of a category, and for <del>those</del> <ins>those specializations of the forms</ins> shown in [tab:locale.spec]<ins>, where:</ins>
>   * <ins>*`InputIterator`* denotes any type that meets the Cpp17InputIterator requirements,</ins>
>   * <ins>*`OutputIterator`* denotes any type that meets the Cpp17OutputIterator requirements,</ins>
>   * <ins>*`C`* denotes any type from an implementation-defined set of character types, including at least `char` and `wchar_t`, that meet the requirements for a character on which any of the iostream components can be instantiated, and</ins>
>   * <ins>*`International`* denotes any value of type `bool`</ins>.

source/locales.tex
```diff
@@ -1371,7 +1371,7 @@
 \tcode{char}
 values to the corresponding
 \tcode{charT}
-value or values.\footnote{The char argument of
+value or values.\footnote{The \tcode{char} argument of
```
zygoloid: "char" -> "\tcode{c}" would be clearer I think.

source/basic.tex
sdkrystian: Well... local variable doesn't :P, but that still doesn't cover functions. Applied!

source/compatibility.tex
AaronBallman: Implicit `int` was removed in C99 (WG14 N635 N661 N692 N722), so at this point, we can just say it's incompatible with C.

jensmaurer: So, we should just drop this entire entry: There is (no longer) any incompatibility, because implicit int is not a thing in (current) C and not in C++. We refer to ISO 9899:2018 in the front matter, btw.

jensmaurer:
Please use \keyword for keywords, which also helps with index entries.

Also, some of the drive-by fixes should be in separate commits.

sdkrystian:
@jensmaurer Sorry! I'll make sure to use \keyword for all future changes (apologies for you having to repeat yourself on 3 different PRs!). 

I'll split the drive-by edits into separate commits.

jensmaurer:
This needs a rebase + force-push, plus some attention to the latest comments.

jensmaurer:
@sdkrystian , ping

source/compatibility.tex
jensmaurer: "implicit int" will be removed by #5088

PR #3909, opened 2020-03-25
Clean up of discarded-value expressions

sdkrystian:
This fixes a few issues with discarded-value expressions:

1. The current wording in [expr.prop] states that the temporary materialization conversion is applied to prvalues of `void` type, which makes discarded-value expressions of `void` type ill-formed (temporary materialization conversion requires a complete type). This fixes that, and ensures that all discarded-value expressions are evaluated (with the exception of unevaluated operands, of course)

2. In [expr.static.cast], grammatical term *expression* is used where "operand" is much clearer, and more consistent with the rest of the wording. Additionally, specifying that the operand of the cast becomes a discarded-value expression is much clearer than saying "the expression", since it could be interpreted as saying that the whole expression is a discarded value expression (it shouldn't, since the operand itself needs to be temporarily materialized and evaluated). A note is also added to clarify that the conversion will be an expression of type `void` with no result.

3. A `return` statement can sometimes have an operand of type `void`, in which case it *should* be a discarded-value expression, but this is not specified. 

As for the editorial nature of this, the intent here is pretty clear :)

Will rename commit after edits :)

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -456,15 +456,15 @@ following:
 Using an overloaded operator causes a function call; the
 above covers only operators with built-in meaning.
 \end{note}
-If the (possibly converted) expression is a prvalue,
-the temporary materialization conversion\iref{conv.rval} is applied.
+The temporary materialization conversion\iref{conv.rval} is applied
+if the (possibly converted) expression is a prvalue of object type.
+The expression is evaluated and its result (if any) is discarded.
 \begin{note}
 If the expression is an lvalue of
 class type, it must have a volatile copy constructor to initialize the
 temporary object that is the result object of the lvalue-to-rvalue
 conversion.
 \end{note}
-The glvalue expression is evaluated and its value is discarded.
 
 \rSec1[conv]{Standard conversions}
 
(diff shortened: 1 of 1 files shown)


source/statements.tex
```diff
@@ -820,7 +820,8 @@
 \indextext{\idxcode{return}!constructor and}%
 \indextext{\idxcode{return}!constructor and}%
 A \tcode{return} statement with an operand of type \tcode{void} shall be used only
-in a function whose return type is \cv{}~\tcode{void}.
+in a function whose return type is \cv{}~\tcode{void}, in which case the operand is
+a discarded-value expression\iref{expr.context}.
```
jensmaurer, line 824: I don't think we need/want to say this here. If you have an operand of void type, nothing needs to be discarded to start with.

source/expressions.tex
```diff
@@ -456,15 +456,15 @@
...
-the temporary materialization conversion\iref{conv.rval} is applied.
+The temporary materialization conversion\iref{conv.rval} is applied
+if the (possibly converted) expression is a prvalue of object type.
+The expression is evaluated and its result (if any) is discarded.
 \begin{note}
 If the expression is an lvalue of
 class type, it must have a volatile copy constructor to initialize the
 temporary object that is the result object of the lvalue-to-rvalue
 conversion.
 \end{note}
```
zygoloid, line 467: This note is quite distant from "the expression" that it's talking about, which is the original discarded-value expression prior to any conversions. Can we get it a bit closer to the wording above that talks about introducing an lvalue-to-rvalue conversion?

jensmaurer: But the temporary materialization conversion is the one that actually creates the object (and needs the volatile copy constructor).

Suggestion: Say "If the original expression" and replace "lvalue-to-rvalue conversion" with "temp. mat. conv." and move "The expression is evaluated and its result (if any) is discarded." to after the note.

PR #3917, opened 2020-04-04
[temp.param] Remove definitions in footnote and fix wording referring to non-type template parameters

sdkrystian:
Removes some more of those pesky definitions in non-normative text, and has a small drive-by edit to [[temp.param] p2](http://eel.is/c++draft/temp.param#2) (a *parameter-declaration* does not declare a type except for certain cases with *elaborated-type-specifier*s where it kind of does (but not really), but that does not apply in a *template-parameter* either way).

Also, what is stated in the footnote should probably have corresponding normative wording, since we don't currently have any.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -277,17 +277,16 @@ names a template type parameter.
 \tcode{typename}
 followed by a
 \grammarterm{qualified-id}
-denotes the type in a non-type%
+denotes the type in a \grammarterm{parameter-declaration}.%
 \footnote{Since template
 \grammarterm{template-parameter}{s}
 and template
 \grammarterm{template-argument}{s}
 are treated as types for descriptive purposes, the terms
-\term{non-type parameter}
-and
-\term{non-type argument}
-are used to refer to non-type, non-template parameters and arguments.}
-\grammarterm{parameter-declaration}.
+non-type \grammarterm{template-parameter} and
+non-type \grammarterm{template-argument}
+are used to refer to non-type, non-template \grammarterm{template-parameter}{s}
+and \grammarterm{template-argument}{s}.}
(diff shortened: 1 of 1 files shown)


source/templates.tex
```diff
@@ -277,17 +277,16 @@
...
 and template
 \grammarterm{template-argument}{s}
 are treated as types for descriptive purposes, the terms
-\term{non-type parameter}
-and
-\term{non-type argument}
-are used to refer to non-type, non-template parameters and arguments.}
-\grammarterm{parameter-declaration}.
+non-type \grammarterm{template-parameter} and
+non-type \grammarterm{template-argument}
```
jensmaurer, line 287: I actually think those definitions should be normative and indexed; they are used quite a lot and we shouldn't rely on the "since" rationale offered here for readers to understand a fairly basic term.

sdkrystian: Should we do _non-type template-parameter_ like we do for _trailing requires-clause_, or _non-type template parameter_? The former seems to be used more often than the latter.

source/templates.tex
```diff
@@ -277,17 +277,16 @@
 \tcode{typename}
 followed by a
 \grammarterm{qualified-id}
-denotes the type in a non-type%
+denotes the type in a \grammarterm{parameter-declaration}.%
```
zygoloid, line 280: We now have a footnote explaining what we mean by "non-type" attached to wording that doesn't say "non-type". That doesn't seem ideal. But in any case the disambiguation rule for `template<class T>` seems like a very strange place to be defining "non-type".

AlisdairM:
Is this PR still relevant following the reworking of non-type template parameters in P2841R7?

PR #3922, opened 2020-04-07
[temp.class.spec.match] Strengthen wording for matching partial specializations

sdkrystian:
The current wording for [[temp.class.spec.match] p2](http://eel.is/c++draft/temp.class.spec.match#2) leaves much to be desired in terms of precision:
> A partial specialization matches a given actual template argument list if the template arguments of the partial specialization can be deduced from the actual template argument list, and the deduced template arguments satisfy the associated constraints of the partial specialization, if any.

Template argument deduction uses a concrete type "reference" to deduce the values of template parameters of a "target". Presumably when we say "template arguments of the partial specialization" we mean the *template-argument-list*, however, the "`A` can be deduced from `B`" relation doesn't make the most sense here since it implies that `A` is the resulting argument for a template parameter produced by the deduction, rather than the "target" of the deduction.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -3393,11 +3393,13 @@ primary template.
 \end{itemize}
 
 \pnum
-A partial specialization matches a given actual template argument
-list if the template arguments of the partial specialization can be
-deduced from the actual template argument list\iref{temp.deduct},
-and the deduced template arguments satisfy the associated constraints
+A class template partial specialization \defnx{matches}{specialization!matches}
+the template argument list of a class template specialization \tcode{A} if, given the type \tcode{P}
+formed by the \grammarterm{simple-template-id} of the partial specialization,
+template argument deduction\iref{temp.deduct.type} for \tcode{P} from \tcode{A} would succeed
+and the deduced template arguments would satisfy the associated constraints
 of the partial specialization, if any\iref{temp.constr.decl}.
+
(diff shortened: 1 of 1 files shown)


source/templates.tex
```diff
@@ -3393,11 +3393,13 @@
...
 \pnum
-A partial specialization matches a given actual template argument
-list if the template arguments of the partial specialization can be
-deduced from the actual template argument list\iref{temp.deduct},
-and the deduced template arguments satisfy the associated constraints
+A class template partial specialization \defnx{matches}{specialization!matches}
+the template argument list of a class template specialization \tcode{A} if, given the type \tcode{P}
+formed by the \grammarterm{simple-template-id} of the partial specialization,
+template argument deduction\iref{temp.deduct.type} for \tcode{P} from \tcode{A} would succeed
+and the deduced template arguments would satisfy the associated constraints
```
zygoloid, line 3400:
```suggestion
template argument deduction\iref{temp.deduct.type} for \tcode{P} from \tcode{A} succeeds
and the deduced template arguments satisfy the associated constraints
```
No need for the hypothetical here.

zygoloid:
I don't find the existing wording especially unclear, but this does seem more precise.

jensmaurer:
Editorial meeting: No appetite to apply.

PR #3924, opened 2020-04-08
[dcl.fct.spec] Remove "class" from "non-static class member function" and cleanup sentence

sdkrystian:


source/declarations.tex
```diff
@@ -522,9 +522,8 @@
 
 \pnum
 \indextext{specifier!\idxcode{virtual}}%
-The \tcode{virtual} specifier shall be used only in the initial
-declaration of a non-static class member function;
-see~\ref{class.virtual}.
+The \tcode{virtual} specifier shall only be used in a \grammarterm{member-declaration}
```
jensmaurer, line 525: This change alters semantics: If a friend declaration (which is a _member-declartion_) declares a virtual member function of another class, the rephrasing makes "virtual" allowed. It wasn't allowed before. Please keep that part of the sentence unaltered.

sdkrystian: Hah! I guess I forgot to stage my changes before I force pushed 2. I did indeed fix this, and also fixed it for *member-declarations* that had multiple *member-declarators*. I'll push correctly this time -- let me know if we want to keep the current wording or my altered version :)

jensmaurer: The rewording doesn't seem much of an improvement in clarity. Is "initial" troublesome because of modules + "reachable"? Or is there any other reason why "initial" needs to go?

Two ways forward (not necessarily exclusive): Strike "class" (I can apply that right away) or have the full rewrite (needs to wait for Richard).

sdkrystian: I think I was a tad overzealous here. The "every *member-declarator*" wording is unnecessary due to [dcl.decl] p3. AFAICT "initial" would have no conflicts here and is unambiguous.

PR #3925, opened 2020-04-08
[dcl.fct.def] Fix incorrect cross-ref referring to "this"

sdkrystian:
On a somewhat related note, it might be time to just remove [[class.this]](http://eel.is/c++draft/class.this). It only has 3 normative sentences, the first 2 being redundant by [expr.prim.this] and the third by [over.match.funcs].


jensmaurer:
Please create a separate issue for the "related" note.

PR #3932, opened 2020-04-11
[class.mem] Deduction guides do not declare new members

sdkrystian:
The intent in [P0091R3](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0091r3.html) is pretty clear (emphasis mine):
> Deduction guides must name a class template and must be introduced within the same semantic scope of the class template, but **they do not becomes members of that scope (i.e if that scope is a class)**.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -513,6 +513,8 @@ A \grammarterm{member-declaration} does not declare new members of the class
 if it is
 \begin{itemize}
 \item a friend declaration\iref{class.friend},
+\item a \grammarterm{deduction-guide}\iref{temp.deduct.guide},
+\item a \grammarterm{template-declaration} whose \grammarterm{declaration} is one of the above,
 \item a \grammarterm{static_assert-declaration},
 \item a \grammarterm{using-declaration}\iref{namespace.udecl}, or
 \item an \grammarterm{empty-declaration}.


source/classes.tex
```diff
@@ -513,6 +513,8 @@
 if it is
 \begin{itemize}
 \item a friend declaration\iref{class.friend},
+\item a \grammarterm{deduction-guide}\iref{temp.deduct.guide},
+\item a \grammarterm{template-declaration} whose \grammarterm{declaration} is one of the above,
```
jensmaurer, line 517: I agree that a _deduction-guide_ does not declare a new member (as explained in the commit comment), but here materializes another rule that isn't mentioned. (Plus, it seems to me a template friend is still a friend declaration of sorts.)

sdkrystian: > Please don't mix unrelated changes in the same commit.

I wouldn't say that it's entirely unrelated, so should I just change the commit message & PR name?

> It seems to me a template friend is still a friend declaration of sorts.

I consider a friend template to be a friend declaration as much as I consider a class template to be a class declaration.

> here materializes another rule that isn't mentioned

I'm assuming that you're referring to *deduction-guides* declared outside of a *template-declaration*. With respect to such declarations, I don't see any wording (normative or informative) that explicitly states such *deduction-guides* can be declared, however we can deduce that it's allowed from: 
- grammar (a *deduction-guide* (without a *template-head*) is a *declaration* as well as a *member-declaration*) and,
- normative wording ([[over.match.class.deduct] p1.4](http://eel.is/c++draft/over.match.class.deduct#1.4) says "if any" when referring to the template parameters; you can't explicitly specialize a deduction guide so the only context where this would occur is when there is no *template-head*)

PR #3934, opened 2020-04-12
[basic.lval] Dynamic type is not a property of objects

sdkrystian:
This is not the only place where such wording exists, but fixing other occurrences would require more substantial rewrites.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -289,15 +289,15 @@ is ill-formed~(\ref{expr.ass}, \ref{expr.post.incr}, \ref{expr.pre.incr}).
 \pnum
 If a program attempts to access\iref{defns.access}
 the stored value of an object through a glvalue
-whose type is not similar\iref{conv.qual} to
+whose static type is not similar\iref{conv.qual} to
 one of the following types the behavior is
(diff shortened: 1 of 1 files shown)


jensmaurer (changes requested):
I agree that the terminology "dynamic type" vs. "static type" is less than ideal, but we need a wholesale fix for that instead of changing a single occurrence (that isn't even the worst).

zygoloid:
"static" versus "dynamic" isn't at all relevant here, since we have no accesses of class type. I think I'd prefer to strike the "dynamic" and not add a "static".

jensmaurer:
Good point; an "access" in C++ is always on a built-in type.

tkoeppe:
@jensmaurer Should we take over this PR? It looks like useful cleanup.

jensmaurer:
Yes, I think there is some cleanup necessary in this area..

tkoeppe:
@jensmaurer: Shall we start by just deleting the two instances of "dynamic" here and leaving wider cleanup for a separate effort?

jensmaurer:
I think I'd prefer a more wholesale approach with CWG review.

PR #3935, opened 2020-04-12
[temp.over] Fix example

languagelawyer:
[[temp.pre]/10](http://eel.is/c++draft/temp.pre#10):
"""
A definition of a function template ... shall be reachable from the end of every definition domain ([basic.def.odr]) in which it is implicitly instantiated ([temp.inst]) unless the corresponding specialization is **explicitly instantiated** ([temp.explicit]) in some translation unit; no diagnostic is required.
"""

So, implicit instantiation (or explicit specialization) won't make the example well-formed.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -9267,9 +9267,8 @@ is well-formed even if the template
 \tcode{f}
 is only declared and not defined at the point of the call.
 The program will be ill-formed unless a specialization for
-\tcode{f<const char*>},
-either implicitly or explicitly generated,
-is present in some translation unit.
+\tcode{f<const char*>}
+is explicitly instantiated in some translation unit\iref{temp.pre}.
 \end{example}
 
 \indextext{template|)}


languagelawyer:
Some historic context: the requirement for template explicit instantiation existence was added in the resolution of CWG323 in 2002 whilst the example was added in Sep 1994 WP and has its current shape since Jun 1996 WP.

source/templates.tex
```diff
@@ -9267,9 +9267,8 @@
 \tcode{f}
 is only declared and not defined at the point of the call.
 The program will be ill-formed unless a specialization for
-\tcode{f<const char*>},
-either implicitly or explicitly generated,
-is present in some translation unit.
+\tcode{f<const char*>}
+is explicitly instantiated in some translation unit.
```
jensmaurer, line 9271: Please append a cross-reference to [temp.pre] in case some reader is interested in the normative rule.

languagelawyer: Done

languagelawyer:
@jensmaurer could you re-review plz?

PR #3937, opened 2020-04-13
[expr.static.cast] Remove unused cv-qualifier notation

sdkrystian:


W-E-Brown:
This excision seems wrong, as it would disallow conversions that are now permitted.  At the least, it seems non-editorial.

sdkrystian:
@W-E-Brown How so? Given a type `const int`, it can be represented as `T`, where `T` is `const int`, or `cv T`, where `cv` is `const` and `T` is `int`. 

The only context where this would present any kind of questionable difference is:
> If `T2` is an inaccessible or ambiguous base class of `T1`, a program that necessitates such a cast is ill-formed.

However, we ignore cv-qualifiers in the determination of if a type `B` is a base class of `D`, so there would be no difference in effect.

W-E-Brown:
That argument is beyond me.  Fewer words does not always produce a better specification.

We routinely go out of our way to indicate when a rule applies to a possibly cv-qualified type.  The proposed change seems to go in the other direction and thus obscures, rather than illuminates.

sdkrystian:
We use the cv notation when we need to establish constraints based what they are, (e.g. the cv-qualification must be the same, or `cv1` must be less cv-qualified than `cv2`). If anything, using the cv notation and then not using the resulting "variables" we introduce is _more_ confusing, since we generally only use the notation when we care about the cv-qualification. Personally, when I see `cv` I automatically assume that it will be used to enforce a restriction.

source/expressions.tex
```diff
@@ -3644,10 +3644,9 @@
 
 \pnum
 An lvalue
-of type ``\cvqual{cv1} \tcode{T1}'' can be cast to type ``rvalue
-reference to \cvqual{cv2} \tcode{T2}'' if ``\cvqual{cv2} \tcode{T2}'' is
-reference-compatible with ``\cvqual{cv1}
-\tcode{T1}''\iref{dcl.init.ref}. If the value is not a bit-field,
+of type \tcode{T1} can be cast to type ``rvalue
```
jensmaurer, line 3647: We could say "An lvalue of (possibly cv-qualified) type T1 ... " to address Walter's concerns. But it's harder to fit in the "possibly cv-qualified" for T2. I do agree with the intent here that we only write "cv T" if we want to talk about "cv" separately.

PR #3938, opened 2020-04-13
[expr.reinterpret.cast] Strike definition in footnote

sdkrystian:


source/expressions.tex
```diff
@@ -3995,7 +3995,7 @@
 No temporary is created, no copy is made, and
 no constructors\iref{class.ctor} or conversion
 functions\iref{class.conv} are called.%
-\footnote{This is sometimes referred to as a \defn{type pun}
+\footnote{This is sometimes referred to as a type pun
```
jensmaurer, line 3998: We're losing "type pun" from the index. That's an undesirable regression.

sdkrystian: Should we just add an index entry above the paragraph?

jensmaurer: Yes, that would address my concerns.

jensmaurer (approved):

PR #3939, opened 2020-04-13
[expr.comma] Remove mid-sentence example

sdkrystian:
Having an example after a comma mid-sentence is _horrible_ (I had to read this several times to realize that it wasn't a sentence fragment).

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -6771,14 +6771,11 @@ value category as its right operand, and is a bit-field if its
 right operand is a bit-field.
 
 \pnum
-In contexts where comma is given a special meaning,
-\begin{example}
-in
-lists of arguments to functions\iref{expr.call} and lists of
-initializers\iref{dcl.init}
-\end{example}
-the comma operator as
-described in this subclause can appear only in parentheses.
+\begin{note}
+In contexts where the comma token is given special meaning
+(e.g. function calls\iref{expr.call}, lists of initializers\iref{dcl.init},
+or \grammarterm{template-argument-list}{s}\iref{temp.names}),
+the comma operator as described in this subclause can appear only in parentheses.
 \begin{example}
 \begin{codeblock}
 f(a, (t=3, t+2), c);
@@ -6786,6 +6783,7 @@ f(a, (t=3, t+2), c);
 has three arguments, the second of which has the value
 \tcode{5}.
 \end{example}
+\end{note}
 
 \pnum
 \begin{note}


W-E-Brown:
Seems like this correctly identifies a problem, but also seems to address it in quite an extreme way.

All that's needed is (as is done elsewhere) to turn the mid-sentence example into a parenthesized phrase that's introduced by 'e.g.' -- let's please not rewrite more than necessary.

sdkrystian:
I considered this, but I don't think we need to make this wording normative. If we were to make it normative, a rephrasing (as done here) would certainly be necessary (it's quite dated and imprecise). I'd argue that doing so would be the more "extreme" approach, as opposed to moving it to the end where we already have informative wording.

timsong-cpp:
Do we even need a normative statement here? Does this not fall out of the grammar productions for the relevant constructs?

sdkrystian:
@timsong-cpp This is indeed redundant (as older wording tends to be)

source/expressions.tex
```diff
@@ -6771,21 +6771,20 @@
...
-in
-lists of arguments to functions\iref{expr.call} and lists of
-initializers\iref{dcl.init}
-\end{example}
+In contexts where comma is given special meaning,
 the comma operator as
 described in this subclause can appear only in parentheses.
+\begin{note}
+Such contexts include function calls\iref{expr.call},
+declarators\iref{dcl.decl} and initializer lists\iref{dcl.init.list}.
```
jensmaurer, line 6779: A declarator is not a context where an expression can appear naturally, so this addition seems not helpful.
Also, an "initializer list" is not the only place where a lits of initializers can appear, e.g. "T(1,2,3)" is another one.
If you want to add another context, it could be a _template-argument-list_.
Also, I think we can make this entire paragraph a note, because the grammar uses _assignment-expression_ for as the constituents of an _expression-list_; same for _template-argument-list_ (via _constant-expression_). So, this paragraph has zero normative content.
Suggestion: "[ Note: In contexts such as function calls (xref), lists of initializers (xref), or template-argument-lists (xref), the comma operator as
described in this subclause can appear only in parentheses. [ Example: ... ] ]"

sdkrystian: Good idea

source/expressions.tex
```diff
@@ -6771,21 +6771,18 @@
...
-in
-lists of arguments to functions\iref{expr.call} and lists of
-initializers\iref{dcl.init}
-\end{example}
-the comma operator as
-described in this subclause can appear only in parentheses.
+\begin{note}
+In contexts such as function calls\iref{expr.call}, lists of initializers\iref{dcl.init},
+or \grammarterm{template-argument-list}{s}\iref{temp.names},
+the comma operator as described in this subclause can appear only in parentheses.
```
zygoloid, line 6777: The old wording only applied to places where comma has special meaning; the new (non-normative, but still) wording appears to also apply even in subexpressions where the special meaning is removed. For example:

```
int a = true ? 1, 2 : 3;  // disallowed by this note, because comma is not in parentheses
int arr[3];
int b = arr[1, 2];  // ok but deprecated, disallowed by this note
int c = []{ foo(), bar(); return 0; }(); // comma operator disallowed by this note
```

Can we preserve the "where comma is given a special meaning" part, to avoid giving a false impression of the rules here?

jensmaurer (approved):

PR #3940, opened 2020-04-13
[basic.type.qualifier] Incomplete and complete object types are just object types

sdkrystian:


source/basic.tex
```diff
@@ -5323,9 +5323,8 @@
 \indextext{\idxcode{const}}%
 \indextext{\idxcode{volatile}}%
 A type mentioned in~\ref{basic.fundamental} and~\ref{basic.compound} is
-a \defnadj{cv-unqualified}{type}. Each type which is a
-cv-unqualified complete or incomplete object type or is
-\tcode{void}\iref{basic.types} has three corresponding cv-qualified
+a \defnadj{cv-unqualified}{type}. Every cv-unqualified object type
+and \tcode{void}\iref{basic.types} has three corresponding cv-qualified
```
timsong-cpp, line 5327: I don't think the grammar is right. "Every...type and `void` has"?

sdkrystian: It honestly reads fine to me (but I am certainly not an unbiased source). If this is a problem, how about:
> Cv-unqualified object types and `void` have three corresponding [...]

jensmaurer: Can we maybe apply the minimal change here and remove "complete or incomplete" and be done with it?

sdkrystian: Sure, I think that would be best

PR #3941, opened 2020-04-15
[string.conversions] Replaced comparisons of a pointer against 0 with comparision against nullptr

iaanus:
While technically correct, comparing pointers against nullptr is usually preferred to comparing them against the null pointer costant 0. Moreover, it seems more consistent to compare against nullptr, since the default value of the pointer idx is actually declared as nullptr in the synopsis.


jwakely:
Please add the [stable.name] for the affected subclause(s) to the commit summary.

iaanus:
Sorry. Done it.

PR #3943, opened 2020-04-17
[basic.lval] Amend note

danra:


source/expressions.tex
```diff
@@ -244,7 +244,7 @@
 \end{note}
 \begin{note}
 Because cv-qualifiers are removed from the type of an expression of
-non-class type when the expression is converted to a prvalue, an lvalue
+non-class, non-array type when the expression is converted to a prvalue, an lvalue
```
jensmaurer, line 247: [conv.lval] p1 says array glvalues can't be converted to prvalues. HIghlighting arrays here gives the false impression that they can be converted to prvalues.

PR #3945, opened 2020-04-18
[expr.unary.op] Fix usage of "result"

languagelawyer:
Do paragraphs 7-10 need similar fix or we fine with «the type of the result is `X`» when it means the type of the resulting prvalue?

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -4163,13 +4163,12 @@ Expressions with unary operators group right-to-left.
 \pnum
 \indextext{expression!unary operator}%
 \indextext{operator!unary}%
-The unary \tcode{*} operator performs \defn{indirection}:
+The unary \tcode{*} operator performs \defn{indirection}.
 \indextext{dereferencing|see{indirection}}%
-the expression to which it is applied shall be a pointer to an object
-type, or a pointer to a function type and the result is an lvalue
-referring to the object or function to which the expression points. If
-the type of the expression is ``pointer to \tcode{T}'', the type of the
-result is ``\tcode{T}''.
+Its operand shall be a prvalue of type ``pointer to \tcode{T}'',
+where \tcode{T} is an object or function type.
+The operator yields an lvalue of type \tcode{T}
+denoting the object or function to which the operand points.
 \begin{note}
 \indextext{type!incomplete}%
 Indirection through a pointer to an incomplete type (other than
@@ -4179,7 +4178,7 @@ lvalue must not be converted to a prvalue, see~\ref{conv.lval}.
 \end{note}
 
 \pnum
-The result of each of the following unary operators is a prvalue.
+Each of the following unary operators yields a prvalue.
 
 \pnum
 \indextext{name!address of cv-qualified}%


jensmaurer:
Why exactly is a fix necessary here? [basic.lval] defines "result" of a glvalue and result of a prvalue, but the usage here seems to be "result of an operator".

source/expressions.tex
```diff
@@ -4166,10 +4166,10 @@
 The unary \tcode{*} operator performs \defn{indirection}:
 \indextext{dereferencing|see{indirection}}%
 the expression to which it is applied shall be a pointer to an object
-type, or a pointer to a function type and the result is an lvalue
+type, or a pointer to a function type and the resulting expression is an lvalue
```
jensmaurer, line 4169: "resulting expression" sounds like a static property, but we're actually talking about the runtime result value (which is an lvalue) here.

source/expressions.tex
```diff
@@ -4166,10 +4166,10 @@
...
 referring to the object or function to which the expression points. If
```
jensmaurer, line 4170: If some renovation is desired here, replacing "expression" with "operand" would probably help. Also, integrating the "pointer to T" phrasing with the preceding "shall be a pointer to object type" rule would probably help.

Maybe something like "The unary * operator performs indirection. Its operand shall be a prvalue of type "pointer to T", where T is an object or function type. The lvalue result of type T refers to the object or function to which the operand points."

languagelawyer:
> Why exactly is a fix necessary here?

I find [this](https://timsong-cpp.github.io/cppwp/n4861/expr.unary.op#1.sentence-2) confusing. Result is the object to which the lvalue refers. And the type of the result is the type of the object, which is not necessary the same as the type of the lvalue.

jensmaurer:
Ah, dynamic type vs. static type.

It seems we need a clearer dictionary to talk about the static type of an operand / the result of the operation vs. the (dynamic) type of the result. Talking about "static" and "dynamic" type has its own set of ugliness, so what words should be in that dictionary? Once we have that clarified, we should fix all of [expr] to do the right thing.

What do you think about my reformulation?

source/expressions.tex
```diff
@@ -4163,13 +4163,12 @@
...
+The unary \tcode{*} operator performs \defn{indirection}.
 \indextext{dereferencing|see{indirection}}%
-the expression to which it is applied shall be a pointer to an object
-type, or a pointer to a function type and the result is an lvalue
-referring to the object or function to which the expression points. If
-the type of the expression is ``pointer to \tcode{T}'', the type of the
-result is ``\tcode{T}''.
+Its operand shall be a prvalue of type ``pointer to \tcode{T}'',
+where \tcode{T} is an object or function type.
+Application of the operator is an lvalue of type \tcode{T}
```
zygoloid, line 4170: Describing the expression as "Application of the operator" seems a bit unusual. We could make this parallel with the wording change below:
```suggestion
The operator yields an lvalue of type \tcode{T}
```

jensmaurer (approved):

tkoeppe:
Will take a look.

source/expressions.tex
```diff
@@ -4163,13 +4163,12 @@
...
 \indextext{operator!unary}%
-The unary \tcode{*} operator performs \defn{indirection}:
+The unary \tcode{*} operator performs \defn{indirection}.
 \indextext{dereferencing|see{indirection}}%
-the expression to which it is applied shall be a pointer to an object
-type, or a pointer to a function type and the result is an lvalue
-referring to the object or function to which the expression points. If
-the type of the expression is ``pointer to \tcode{T}'', the type of the
-result is ``\tcode{T}''.
+Its operand shall be a prvalue of type ``pointer to \tcode{T}'',
```
tkoeppe, line 4168: @jensmaurer: is it a normative change to require that the operand be a prvalue, or was that previously implied by other wording?

jensmaurer: The descriptions of most operators are unclear whether their operands are prvalues or lvalues.
Anyway, for the dereferencing operator, it's pelucidly clear that it takes a prvalue, always. Maybe the prior text saying "shall be a pointer to an object" could be interpreted as "shall be a pointer value...:", meaning a prvalue of pointer type.

tkoeppe: I suppose this is largely inherited from C, where things aren't lvalues unless they're explicitly called out as lvalues.

The normative impact/question is whether there's an lvalue-to-rvalue conversion in something like `*p`. I agree that this can't really be anything else (after all, you need to "load" the pointer value before you can dereference it), but it just feels a tiny bit bold to establish such clarity editorially. Let's proceed optimistically, but let CWG know via email.

languagelawyer: @tkoeppe In general, it is CWG1642.

tkoeppe: @languagelawyer: thank you!

tkoeppe:
@jensmaurer I think we can take this, but I would appreciate if you took a final fresh core look at this and confirm that this is both correct and an improvement.

jensmaurer:
I think I want to have a CWG look at this, just to be on the safe side.

jensmaurer:
Approved by CWG telecon 2022-08-26.

tkoeppe:
@jensmaurer Just to confirm, we can remove the label:cwg label?

jensmaurer:
Removed "cwg" label.

PR #3951, opened 2020-04-20
[basic.lval] Xvalue don't always denote entities whose resources can be reused

sdkrystian:
Entities denoted by an xvalue can't always have their resources reused (i.e. inaccessible members, cv-qualification, etc.), and such entities generally have no guarantee as to when their lifetime will end. It's best we make this wording non-normative, and instead just say they are glvalue that denote objects/bit-fields and are not lvalues. 

While this definition might sound circular, the value category of an expression is determined from the context in which it appears, not from the definition of what the value categories are.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -143,7 +143,10 @@ Expressions are categorized according to the taxonomy in \fref{basic.lval}.
 or computes the value of an operand of an operator,
 as specified by the context in which it appears,
 or an expression that has type \cv{}~\tcode{void}.
-\item An \defn{xvalue} is a glvalue that denotes an object or bit-field whose resources can be reused (usually because it is near the end of its lifetime).
+\item An \defn{xvalue} is a glvalue that denotes an object or bit-field and is not an lvalue.
+\begin{note}
+Such expressions typically denote entities near the end of their lifetime whose resources can be reused.
+\end{note}
 \item An \defn{lvalue} is a glvalue that is not an xvalue.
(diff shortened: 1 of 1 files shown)


source/expressions.tex
```diff
@@ -143,7 +143,10 @@
 or computes the value of an operand of an operator,
 as specified by the context in which it appears,
 or an expression that has type \cv{}~\tcode{void}.
-\item An \defn{xvalue} is a glvalue that denotes an object or bit-field whose resources can be reused (usually because it is near the end of its lifetime).
+\item An \defn{xvalue} is a glvalue that denotes an object or bit-field and is not an lvalue.
+\begin{note}
+Such expressions typically denote entities near the end of their lifetime whose resources can be reused.
+\end{note}
 \item An \defn{lvalue} is a glvalue that is not an xvalue.
```
zygoloid, line 150: I don't think this is really a reasonable pair of definitions any more, due to the circularity. I agree the old definition of "xvalue" was also not really a definition, but I'm not convinced this is an improvement. Perhaps we could arrange this more like "A *glvalue* is either an *lvalue* or an *xvalue*; lvalues typically refer to this kind of thing and xvalues refer to that kind of thing", but we should make sure we don't lose the "no function xvalues" property. But I don't have a good suggestion, so I'm leaning towards leaving this as-is, even though the existing wording is muddying semantics with intent.

tkoeppe:
Please reopen this if you'd like to continue the discussion or have new suggestions.

PR #3955, opened 2020-04-23
[expr] Clarify performing / applying standard conversions.

zygoloid:
We previously had a conflict between wording demanding that certain
conversions be applied, and the definitions of the conversions saying
they could not be applied. We now distinguish two cases:

 * When we say that a conversion is applied, we're talking about a
   change to the resulting model of the program -- we mean that the
   conversion actually happens (and it had better be applicable).
 * When we say that a conversion is performed, we're talking about
   performing a process, which applies the conversion if possible and
   otherwise has no effect.

In addition, wording saying to perform the three kinds of conversion
from glvalue to prvalue have been changed to say that a prvalue operand
is expected.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -235,9 +235,9 @@ a temporary object is materialized; see \ref{expr.context}.
 
 \pnum
 Whenever a glvalue appears as an operand of an operator that
-expects a prvalue for that operand, the
+expects a prvalue for that operand, an
 lvalue-to-rvalue\iref{conv.lval}, array-to-pointer\iref{conv.array},
-or function-to-pointer\iref{conv.func} standard conversions are
+or function-to-pointer\iref{conv.func} standard conversion is
 applied to convert the expression to a prvalue.
 \begin{note}
 An attempt to bind an rvalue reference to an lvalue is not such a context; see~\ref{dcl.init.ref}.
@@ -433,7 +433,7 @@ In some contexts, an expression only appears for its side effects. Such an
 expression is called a \defn{discarded-value expression}.
 The array-to-pointer\iref{conv.array}
 and function-to-pointer\iref{conv.func} standard conversions are not
-applied. The lvalue-to-rvalue conversion\iref{conv.lval} is applied
+applied. The lvalue-to-rvalue conversion\iref{conv.lval} is performed
 if and only if
 the expression is a glvalue of volatile-qualified type and it is one of the
 following:
@@ -531,6 +531,15 @@ An expression $E$ can be
 \defnx{implicitly converted}{conversion!implicit} to a type \tcode{T} if and only if the
 declaration \tcode{T t=$E$;} is well-formed, for some invented temporary
 variable \tcode{t}\iref{dcl.init}.
+
+\pnum
+Certain language constructs require that a conversion be performed
+on an operand. In such cases, if the conversion is applicable to that operand,
+it is applied, otherwise performing the conversion has no effect.
+\begin{example}
+Performing the lvalue-to-rvalue conversion on an array lvalue has no effect,
+because the lvalue-to-rvalue conversion cannot be applied to arrays.
+\end{example}
 
 \pnum
 Certain language constructs require that an expression be converted to a Boolean
@@ -665,8 +674,10 @@ See also~\ref{basic.lval}.
 An lvalue or rvalue of type ``array of \tcode{N} \tcode{T}'' or ``array
 of unknown bound of \tcode{T}'' can be converted to a prvalue of type
 ``pointer to \tcode{T}''.
-The temporary materialization conversion\iref{conv.rval} is applied.
-The result is a pointer to the first element of the array.
+The temporary materialization conversion\iref{conv.rval} is first applied
+if the expression is a prvalue.
+The result is a pointer to the first element of
+the possibly-converted array glvalue.
 
 \rSec2[conv.func]{Function-to-pointer conversion}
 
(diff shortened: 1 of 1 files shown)


source/expressions.tex
```diff
@@ -433,7 +433,7 @@
 expression is called a \defn{discarded-value expression}.
 The array-to-pointer\iref{conv.array}
 and function-to-pointer\iref{conv.func} standard conversions are not
-applied. The lvalue-to-rvalue conversion\iref{conv.lval} is applied
+applied. The lvalue-to-rvalue conversion\iref{conv.lval} is performed
```
jensmaurer, line 436: I thought I had understand "applied", but why is this "performed", not "applied"? We do convert the lvalue to an rvalue and fail if that's not possible.

source/expressions.tex
```diff
@@ -3111,10 +3122,7 @@
 thus each such argument has a corresponding parameter when a function template
 specialization is actually called.
 \end{note}
-The
-lvalue-to-rvalue\iref{conv.lval}, array-to-pointer\iref{conv.array},
-and function-to-pointer\iref{conv.func} standard conversions are
-performed on the argument expression.
+In this case, argument expression is expected to be a prvalue\iref{basic.lval}.
```
jensmaurer, line 3125: missing "the"?

source/expressions.tex
```diff
@@ -6092,17 +6092,13 @@
...
-lvalue-to-rvalue\iref{conv.lval},
-array-to-pointer\iref{conv.array},
-and function-to-pointer\iref{conv.func}
-standard conversions are performed on the operands.
+The operands are expected to be prvalues\iref{basic.lval}.
 The comparison is deprecated if
-both operands were of array type
-prior to these conversions\iref{depr.array.comp}.
+an array-to-pointer\iref{conv.array} conversion
+is applied to both operands\iref{depr.array.comp}.
```
jensmaurer, line 6098: The "operand" here means the original operand, but...

source/expressions.tex
```diff
@@ -6092,17 +6092,13 @@
...
 
 \pnum
-The converted operands shall have arithmetic, enumeration, pointer,
+The operands shall have arithmetic, enumeration, pointer,
```
jensmaurer, line 6101: The "operand" here means the converted prvalue operand. That's non-obvious.

source/expressions.tex
```diff
@@ -235,9 +235,9 @@
 
 \pnum
 Whenever a glvalue appears as an operand of an operator that
-expects a prvalue for that operand, the
+expects a prvalue for that operand, an
```
jensmaurer, line 238: Reading all the "expected to be" phrasings below makes me want to ask for "requires" here.
Example: "that requires a prvalue for that operand..." or "The operand is required to be a prvalue."
The "required" conveys better that, if you can't make it so, you're toast. (Just a slight wording preference, not a strong feeling.)

jensmaurer (changes requested):
We have lots more places where we expect a prvalue, but currently don't say so.
Example: expr.or, expr.bit.and, expr.xor.
Maybe we should say that "usual arithmetic conversions" yield prvalues? Currently, they only deal with types, not with value category.

source/expressions.tex
zygoloid: No, we don't, per discussion on the core reflector; clarifying this was the original motivation of the change.

```
void f() {
  volatile int arr[3];
  arr;
}
```

is valid and does not apply an lvalue-to-rvalue conversion to `arr`.

jensmaurer: Ok.

source/expressions.tex
```diff
@@ -665,8 +674,10 @@
 An lvalue or rvalue of type ``array of \tcode{N} \tcode{T}'' or ``array
 of unknown bound of \tcode{T}'' can be converted to a prvalue of type
 ``pointer to \tcode{T}''.
-The temporary materialization conversion\iref{conv.rval} is applied.
-The result is a pointer to the first element of the array.
+The temporary materialization conversion\iref{conv.rval} is first applied
+if the expression is a prvalue.
+The result is a pointer to the first element of
+the possibly-converted array glvalue.
```
languagelawyer, line 680: glvalue is an expression, it doesn't have elements.

languagelawyer:
> We have lots more places where we expect a prvalue, but currently don't say so.

Probably you remember, but just in case: this is http://wg21.link/CWG1642

source/expressions.tex
```diff
@@ -6481,10 +6477,9 @@
 subclause.
 
 \pnum
-Lvalue-to-rvalue\iref{conv.lval}, array-to-pointer\iref{conv.array},
-and function-to-pointer\iref{conv.func} standard conversions are
-performed on the second and third operands. After those conversions, one
-of the following shall hold:
+The resulting second and third operands are expected to be
```
sdkrystian, line 6480: Perhaps we say "possibly-converted" instead of "resulting"

source/expressions.tex
```diff
@@ -6011,17 +6015,13 @@
     relational-expression \terminal{>=} compare-expression
 \end{bnf}
 %
-The
-lvalue-to-rvalue\iref{conv.lval},
-array-to-pointer\iref{conv.array},
-and function-to-pointer\iref{conv.func}
-standard conversions are performed on the operands.
+The operands are expected to be prvalues\iref{basic.lval}.
 The comparison is deprecated if
```
sdkrystian, line 6019: Ending the previous sentence with a semicolon and continuing with "the comparison is deprecated if [...]" flows better.

source/expressions.tex
```diff
@@ -6011,17 +6015,13 @@
...
-both operands were of array type
-prior to these conversions\iref{depr.array.comp}.
+an array-to-pointer\iref{conv.array} conversion
```
sdkrystian, line 6020: The array-to-pointer conversion is a single conversion -- I think "an" should be "the" here.

source/expressions.tex
```diff
@@ -6092,17 +6092,13 @@
...
```
sdkrystian, line 6097: The array-to-pointer conversion is a single conversion -- I think "an" should be "the" here.

PR #3958, opened 2020-04-25
[expr.spaceship] Clarify treatment of pointers.

### LINKED ISSUE #3956: [expr.spaceship]: Wording should be clearer that function pointers don't do <=>
[expr.spaceship] p7+8 use a convoluted and partially misleading way to exclude function pointers from three-way comparison:

> 7 If the composite pointer type is an object pointer type, p <=> q is of type std::strong_ordering. If
> two pointer operands p and q compare equal (7.6.10), p <=> q yields std::strong_ordering::equal; if p
> and q compare unequal, p <=> q yields std::strong_ordering::less if q compares greater than p and
> std::strong_ordering::greater if p compares greater than q (7.6.9). Otherwise, the result is unspecified.
> 8 Otherwise, the program is ill-formed.

The first problem is that to me the "arm-length" of the first if in p7's first sentence is unclear and the second problem is that the end of p7 and the start of p8 are confusing. My initial reading lead to the understanding that the first p7 sentence is a very special sentence about object pointers and the following ones of that paragraph would apply to potentially other pointers as well. 

I would therefore suggest to follow the style used in [expr.rel] p4 and reformulate p7 by converting it into a bullet list as follows:

7 If the composite pointer type is an object pointer type, p <=> q is of type std::strong_ordering <ins>and the result is defined in terms of the following rules:</ins><del>.</del> 
- If two pointer operands p and q compare equal (7.6.10), p <=> q yields std::strong_ordering::equal; 
- <ins>otherwise,</ins> if p and q compare unequal, p <=> q yields std::strong_ordering::less if q compares greater than p and std::strong_ordering::greater if p compares greater than q (7.6.9). 
- Otherwise, the result is unspecified.

8 Otherwise, the program is ill-formed.

During a corresponding discussion with Richard on the CWG reflector he generally agreed with the suggested change direction and he suggested to additionally improve the previous p6 wording as well as follows:

6 If at least one of the operands is of <ins>object</ins> pointer type and the other operand is of <ins>object</ins> pointer or array type, array-to pointer conversions (7.3.2), pointer conversions (7.3.11), and qualification conversions (7.3.5) are performed on both operands to bring them to their composite pointer type (7.2.2). [..]

jensmaurer:
Fixes #3956.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -5955,8 +5955,8 @@ converting the operands to the underlying type of \tcode{E}
 and applying \tcode{<=>} to the converted operands.
 
 \pnum
-If at least one of the operands is of pointer type and
-the other operand is of pointer or array type,
+If at least one of the operands is of object pointer type and
+the other operand is of object pointer or array type,
 array-to-pointer conversions\iref{conv.array},
 pointer conversions\iref{conv.ptr},
 and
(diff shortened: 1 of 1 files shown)


source/expressions.tex
```diff
@@ -5954,17 +5954,23 @@
 
 \pnum
 If the composite pointer type is an object pointer type,
-\tcode{p <=> q} is of type \tcode{std::strong_ordering}.
+\tcode{p <=> q} is of type \tcode{std::strong_ordering} and
+the result is defined in terms of the following rules:
```
zygoloid, line 5958: "in terms of" -> "by".

source/expressions.tex
```diff
@@ -5954,17 +5954,23 @@
 
 \pnum
 If the composite pointer type is an object pointer type,
```
zygoloid, line 5956: Hm. We now only "bring them to their composite pointer type" (and thereby introduce the "composite pointer type") when one of the operands is of object pointer type. And in that case, the composite pointer type is *always* an object pointer type. So what we really mean here now is "If we did that stuff in the previous paragraph". So maybe:

```suggestion
In this case,
```

jensmaurer: Agreed. Fixed.

source/expressions.tex
```diff
@@ -5953,18 +5953,24 @@
 \end{note}
 
 \pnum
```
zygoloid, line 5955: Maybe drop the paragraph number too, as we're continuing the same case?

jensmaurer:
Rebased again to fix Travis failure.

PR #3960, opened 2020-04-29
[class.abstract] Clean up redundant wording

sdkrystian:
[class.abstract] p2 sentence 3 is entirely redundant by [class.abstract] p4, [class.abstract] sentence 5 is redundant by [basic.def.odr] p3.1, the normative specification for [class.abstract] p6 makes more sense in [class.cdtor] p4, [class.abstract] p2 note 2 and [class.abstract] p3 can be merged into a single note, and the definition of abstract class in [class.abstract] p4 doesn't need to care about final overriders, since members of a base class that aren't redeclared are members of the derived class.

The definition of _pure virtual function_ should probably use `\defnadj`, but please let me know what makes the most sense :)

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -4114,25 +4114,17 @@ provide a variety of implementations.
 \end{note}
 
 \pnum
-A virtual function is specified as
-a \defnx{pure virtual function}{function!virtual!pure} by using a
-\grammarterm{pure-specifier}\iref{class.mem} in the function declaration
-in the class definition.
+A virtual function is a \defnx{pure virtual function}{function!virtual!pure}
+if its \grammarterm{member-declarator} has a \grammarterm{pure-specifier}\iref{class.mem}.
 \begin{note}
 Such a function might be inherited: see below.
 \end{note}
-A class is an \defnadj{abstract}{class}
-if it has at least one pure virtual function.
-\begin{note}
-An abstract class can be used only as a base class of some other class;
-no objects of an abstract class can be created
-except as subobjects of a class
-derived from it~(\ref{basic.def}, \ref{class.mem}).
-\end{note}
 \indextext{definition!pure virtual function}%
+\begin{note}
 A pure virtual function need be defined only if called with, or as if
 with\iref{class.dtor}, the \grammarterm{qualified-id}
-syntax\iref{expr.prim.id.qual}.
+syntax\iref{expr.prim.id.qual}; see~\ref{basic.def.odr}.
+\end{note}
 \begin{example}
 \begin{codeblock}
 class point { @\commentellip@ };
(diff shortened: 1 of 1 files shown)


source/classes.tex
```diff
@@ -4114,25 +4114,16 @@
...
-no objects of an abstract class can be created
-except as subobjects of a class
-derived from it~(\ref{basic.def}, \ref{class.mem}).
-\end{note}
 \indextext{definition!pure virtual function}%
+\begin{note}
 A pure virtual function need be defined only if called with, or as if
 with\iref{class.dtor}, the \grammarterm{qualified-id}
-syntax\iref{expr.prim.id.qual}.
+syntax\iref{expr.prim.id.qual}, see~\ref{basic.def.odr}.
```
jensmaurer, line 4126: Semicolon before "see", not comma.

source/classes.tex
```diff
@@ -4146,6 +4137,7 @@
 };
 \end{codeblock}
 \end{example}
+\end{note}
```
jensmaurer, line 4140: I think the example is for the entire facility, not just for "need be defined only". So, please move the example out of the note.

source/classes.tex
```diff
@@ -4158,31 +4150,17 @@
...
-because the resulting prvalue would be of abstract class type\iref{basic.lval}.
-However, pointers and references to abstract class types
-can appear in such contexts.
-\end{note}
-
 \pnum
 \indextext{function!virtual!pure}%
-A class is abstract if it contains or inherits at least one pure virtual
-function for which the final overrider is pure virtual.
+A class is an \defnadj{abstract}{class} if it has at least one member that is
```
jensmaurer, line 4155: Semantic line breaks, please: the stuff after (and including) the "if" should be on a single separate line.

source/classes.tex
```diff
@@ -4201,6 +4179,27 @@
 \tcode{circle::draw()} must be provided.
 \end{example}
 
+\pnum
+\begin{note}
+In general, an abstract class type can only be used
+\begin{itemize}
+\item
+as the referred to type in a reference,
```
jensmaurer, line 4187: No. That's awkward and needs at least a hyphen, but even better, needs a rephrasing. Maybe "referenced type" and "pointee type" or so?

source/classes.tex
```diff
@@ -4201,6 +4179,27 @@
 \tcode{circle::draw()} must be provided.
 \end{example}
 
+\pnum
+\begin{note}
+In general, an abstract class type can only be used
```
jensmaurer, line 4184: Are there exceptions to "in general"? I don't think so, so please strike that.

source/classes.tex
```diff
@@ -4201,6 +4179,27 @@
...
+\begin{itemize}
+\item
+as the referred to type in a reference,
+\item
+as the pointed to type in a pointer,
+\item
+in an explicit type conversion
+to the above types~(\ref{expr.static.cast}, \ref{expr.reinterpret.cast},
+\ref{expr.const.cast}, \ref{expr.dynamic.cast}), or
```
jensmaurer, line 4193: That's an odd point to make, given that you already highlighted that you can form pointers and references to abstract class types.

source/classes.tex
```diff
@@ -4210,13 +4209,15 @@
...
 \pnum
 \indextext{class!constructor and abstract}%
+\begin{note}
 Member functions can be called from a constructor (or destructor) of an
 abstract class;
 \indextext{virtual function call!undefined pure}%
 the effect of making a virtual call\iref{class.virtual} to a pure
 virtual function directly or indirectly for the object being created (or
-destroyed) from such a constructor (or destructor) is undefined.%
+destroyed) from such a constructor (or destructor) is undefined, see~\ref{class.cdtor}.%
```
jensmaurer, line 4218: Semicolon before "see".

source/classes.tex
```diff
@@ -4210,13 +4209,15 @@
...
 \indextext{derived class|)}
+\end{note}
```
jensmaurer, line 4220: The closing indexing bracket should be after the "end note".

source/classes.tex
```diff
@@ -6335,7 +6336,9 @@
 destruction,
 the function called is the
 final overrider in the constructor's or destructor's class and not one
-overriding it in a more-derived class.
+overriding it in a more-derived class;
+if the final overrider is a pure virtual function,
```
jensmaurer, line 6340: make that a new sentence and change: the final overrider -> that final overrider
(We're not talking about the final overrider in the entire hierarchy, but the one of the ctor's class.)

jensmaurer: Also add a cross-reference to the definition of "pure virtual function".

jensmaurer (changes requested):
The motivation is good, but the note highlighting the allowed / disallowed situations for abstract classes isn't stellar yet.

PR #3961, opened 2020-04-29
[temp.local] Remove space between angle brackets in example

sdkrystian:
We don't do this very much, if at all.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -4747,7 +4747,7 @@ template <class T> struct Derived: public Base<T> {
 };
 
 template<class T, template<class> class U = T::template Base> struct Third { };
-Third<Derived<int> > t;                 // OK: default argument uses injected-class-name as a template
+Third<Derived<int>> t;                  // OK: default argument uses injected-class-name as a template
 \end{codeblock}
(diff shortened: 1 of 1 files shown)


source/templates.tex
```diff
@@ -4747,7 +4747,7 @@
 };
 
 template<class T, template<class> class U = T::template Base> struct Third { };
-Third<Derived<int> > t;                 // OK: default argument uses injected-class-name as a template
+Third<Derived<int>> t;                  // OK: default argument uses injected-class-name as a template
```
jensmaurer, line 4750: I think we don't want to change this under the "core sections show a variety of syntactically valid styles" doctrine.

sdkrystian: @jensmaurer Could we at least make the spacing symmetric? :)

sdkrystian: Ah, the one place I didn't look when checking for more instances was... [temp]. I can still wait for @zygoloid to look this over if we want, but I agree with you that we shouldn't make the change

jensmaurer:
Author agreed not to apply the change.

PR #3962, opened 2020-04-29
[class.dtor] Remove incorrect uses of virtual as a keyword

sdkrystian:
These shouldn't be `\tcode`, as these are defined terms (a function can be a _virtual function_ ora  _pure virtual function_)

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -2228,7 +2228,7 @@ A defaulted destructor for a class
 \pnum
 A destructor is trivial if it is not user-provided and if:
 \begin{itemize}
-\item the destructor is not \tcode{virtual},
+\item the destructor is not virtual,
 
 \item all of the direct base classes of its class have trivial destructors, and
 
@@ -2260,10 +2260,9 @@ implicitly defined.
 \pnum
 \indextext{destructor!virtual}%
 \indextext{destructor!pure virtual}%
-A prospective destructor can be declared
-\tcode{virtual}\iref{class.virtual}
-or pure
-\tcode{virtual}\iref{class.abstract}.
+A prospective destructor can be
+declared \keyword{virtual}\iref{class.virtual}
+and with a \grammarterm{pure-specifier}\iref{class.abstract}.
 If the destructor of a class is virtual and
 any objects of that class or any derived class are created in the program,
 the destructor shall be defined.


source/classes.tex
```diff
@@ -2260,10 +2260,9 @@
...
-A prospective destructor can be declared
-\tcode{virtual}\iref{class.virtual}
-or pure
-\tcode{virtual}\iref{class.abstract}.
+A prospective destructor can be
+a virtual\iref{class.virtual}
+or pure virtual function\iref{class.abstract}.
 If the destructor of a class is virtual and
 any objects of that class or any derived class are created in the program,
 the destructor shall be defined.
```
jensmaurer, line 2268: Side note (separate pull request, please): I think this "shall be defined" business should be / is covered by [basic.def.odr], so this can become a note.

source/classes.tex
```diff
@@ -2260,10 +2260,9 @@
 \pnum
 \indextext{destructor!virtual}%
 \indextext{destructor!pure virtual}%
-A prospective destructor can be declared
-\tcode{virtual}\iref{class.virtual}
-or pure
-\tcode{virtual}\iref{class.abstract}.
+A prospective destructor can be
+a virtual\iref{class.virtual}
+or pure virtual function\iref{class.abstract}.
```
jensmaurer, line 2265: Certainly, a pure virtual function is also virtual, so this seems a bit redundant. But that's a pre-existing defect.

source/classes.tex
```diff
@@ -2260,10 +2260,9 @@
 \pnum
 \indextext{destructor!virtual}%
 \indextext{destructor!pure virtual}%
-A prospective destructor can be declared
-\tcode{virtual}\iref{class.virtual}
-or pure
-\tcode{virtual}\iref{class.abstract}.
+A prospective destructor can be
+a virtual\iref{class.virtual}
```
jensmaurer, line 2264: We're slightly changing the semantics here. Previously, we said we can declare those things with the virtual keyword, now we're saying they're virtual functions (which might be endowed due to inheritance).

Maybe: "A prospective destructor can be declared `virtual` (class.virtual), including as a pure virtual function (class.abstract)."

source/classes.tex
```diff
@@ -2228,7 +2228,7 @@
 \pnum
 A destructor is trivial if it is not user-provided and if:
 \begin{itemize}
-\item the destructor is not \tcode{virtual},
+\item the destructor is not virtual,
```
jensmaurer, line 2231: This change is good and should survive.

source/classes.tex
sdkrystian: _Pure virtual function_ is _technically_ the correct terminology, albeit a tad redundant.

source/classes.tex
sdkrystian: Will do.

source/classes.tex
sdkrystian: I don't think we are changing semantics here -- prospective destructors are still non-static member functions, thus [class.virtual] p1 applies (they can be virtual either though the use of the specifier, or through overriding), and so does [class.virtual] p2 (a destructor can override one declared in a base class). This paragraph and [class.virtual[ p7 reinforce that virtual prospective destructors are allowed, but aren't what permit this behavior.

sdkrystian: @jensmaurer (while looking at [class.virtual] p2 it looks like ref-qualifier should be a grammar term and "cv-qualification" should be _cv-qualifier-seq_. I'll open a seperate PR)

source/classes.tex
sdkrystian: #3966

source/classes.tex
sdkrystian: This sentence could probably just be turned into a note in #3966

jensmaurer: I think a prospective destructor never overrides anything (it's still a templated entity). Only at instantiation time do you choose the destructor, and that might be a virtual function via inheritance.

sdkrystian: A prospective destructor doesn't necessarily have to be a templated entity:
```
struct T
{
  virtual ~T(); // prospective destructor
}; // selected destructor
```

Similarly, to permit the use of `override`, a prospective destructor would _need_ to be a virtual function by inheritance.

source/classes.tex
jensmaurer: I'm arguing on the semantic level. "pure virtual function" implies "virtual function".

source/classes.tex
jensmaurer: A virtual functions can't have a trailing-requires-clause [class.virtual] p6, so my point is moot, I believe.

jensmaurer: I've thought about this some more. I think we should clearly distinguish between "declared `virtual`" and "virtual function". The former refers to the token appearing in the class definition, the latter refers to the resulting attribute of the member function. The original sentence referred to the token (make that `\keyword`), so we should retain that.
In order not to mix the levels here, maybe saying "A prospective destructor can be declared `virtual` (class.virtual) and with a _pure-specifier_ (class.abstract)." limts the statement to the syntax. The rest follows from the referenced places.

sdkrystian: Limiting it to syntax is reasonable -- perhaps we add a note saying that it may be virtual by inheritance?

jensmaurer (approved):

PR #3963, opened 2020-04-29
[conv.integral, conv.double] Fix promotion conversions

sdkrystian:
- Instead of delegating integral conversions to `bool` to [conv.bool], we can just exclude them from the set of integral conversions (no normative effects as it delegates). 
- [conv.integral] p2 & p3 can be merged as the work better as a single paragraph (due to "Otherwise [...]").
- There is only one floating-point promotion conversion (from `float` to `double`); we need not say "conversions".

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -928,8 +928,7 @@ a prvalue of an integer type.
 
 \pnum
 \indextext{conversion!bool@\tcode{bool}}%
-If the destination type is \keyword{bool}, see~\ref{conv.bool}. If the
-source type is \keyword{bool}, the value \keyword{false} is converted to
+If the source type is \keyword{bool}, the value \keyword{false} is converted to
 zero and the value \keyword{true} is converted to one.
 
 \pnum
@@ -940,8 +939,8 @@ that is congruent to the source integer modulo $2^N$,
 where $N$ is the width of the destination type.
 
 \pnum
-The conversions allowed as integral promotions are excluded from the set
-of integral conversions.
+The conversions performed by integral promotions and boolean conversions are
+excluded from the set of integral conversions.
 
 \rSec2[conv.double]{Floating-point conversions}
 
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -939,7 +939,7 @@ that is congruent to the source integer modulo $2^N$,
 where $N$ is the width of the destination type.
 
 \pnum
-The conversions performed by integral promotions and boolean conversions are
+The conversions allowed as integral promotions and boolean conversions are
 excluded from the set of integral conversions.
 
 \rSec2[conv.double]{Floating-point conversions}
@@ -961,7 +961,7 @@ destination values, the result of the conversion is an
 Otherwise, the behavior is undefined.
 
 \pnum
-The conversions allowed as floating-point promotions are excluded from
+Floating-point promotion is excluded from
 the set of floating-point conversions.
 
 \rSec2[conv.fpint]{Floating-integral conversions}


jensmaurer (changes requested):
These seem to be three independent changes. Please make that three commits.

sdkrystian:
@jensmaurer Should be good now.

source/expressions.tex
```diff
@@ -884,8 +884,7 @@
 
 \pnum
 \indextext{conversion!bool@\tcode{bool}}%
-If the destination type is \tcode{bool}, see~\ref{conv.bool}. If the
```
jensmaurer, line 887: What exactly is the rationale for removing this? There seems to be unresolved overlap/conflict between [conv.integral] and [conv.bool].

source/expressions.tex
```diff
@@ -896,8 +895,8 @@
 where $N$ is the width of the destination type.
 
 \pnum
-The conversions allowed as integral promotions are excluded from the set
-of integral conversions.
+The conversions performed by integral promotions and boolean conversions are
```
jensmaurer, line 896: We say "can", so "allowed" seems the appropriate phrasing for that.

source/expressions.tex
jensmaurer: Ah, you exclude boolean conversions below. Ok.

source/expressions.tex
```diff
@@ -884,12 +884,10 @@
 
 \pnum
 \indextext{conversion!bool@\tcode{bool}}%
-If the source type is \tcode{bool}, the value \tcode{false} is converted to
-zero and the value \tcode{true} is converted to one.
-
-\pnum
 \indextext{conversion!to unsigned}%
 \indextext{conversion!to signed}%
+If the source type is \tcode{bool}, the value \tcode{false} is converted to
```
jensmaurer, line 889: I like having two separate paragraphs here; bool is indeed handled totally differently from any other integer.

source/expressions.tex
```diff
@@ -909,7 +909,7 @@
 Otherwise, the behavior is undefined.
 
 \pnum
-The conversions allowed as floating-point promotions are excluded from
+The conversion performed by floating-point promotion is excluded from
```
jensmaurer, line 912: Again, we say "can", so "allowed" seem fine.

Going plural -> singular is, of course, good.

jensmaurer (approved):

zygoloid:
This appears to be a normative change: it affects which conversions are permitted in converted constant expressions. It might be a good change, and it's certainly clearer to have each conversion have only one kind, but this should probably be handled by a core issue.

tkoeppe:
@sdkrystian Could you please rebase?

sdkrystian:
@tkoeppe Yup, will do it Monday

source/expressions.tex
```diff
@@ -912,7 +911,7 @@
 Otherwise, the behavior is undefined.
 
 \pnum
-The conversions allowed as floating-point promotions are excluded from
+The conversion allowed as floating-point promotion is excluded from
```
opensdh, line 914:
```suggestion
Floating-point promotion is excluded from
```

sdkrystian:
@tkoeppe Rebased

tkoeppe:
Thanks! We'll have to look at it in Core, but we wanted a clean diff.

tkoeppe:
@jensmaurer I think we'll still want a CWG check here?

jensmaurer:
@zygoloid said "This appears to be a normative change:"  so we can't do this editorially.

tkoeppe:
OK. Shall we close this then? Does it need an EWG paper?

jensmaurer:
The last change is clearly editorial (fixing plural -> singular). The other two changes could be handled by a core issue, but since it's supposed to be a no-op beautifying change, it doesn't feel urgent.

tkoeppe:
> The last change is clearly editorial (fixing plural -> singular). 

Hm, I suppose it's not even _wrong_ per se. It's a set of one element. And given that we keep doing things with floating point types, it may well not remain so.

PR #3964, opened 2020-04-29
[dcl.name] Change informative style normative wording into note

sdkrystian:


sdkrystian:
Instead of `typedef` we can also do "typedef declaration" since it's a defined term.

source/declarations.tex
```diff
@@ -2364,8 +2364,10 @@
 \end{example}
 
 \pnum
-A type can also be named (often more easily) by using a
-\tcode{typedef}\iref{dcl.typedef}.
+\begin{note}
+Oftentimes, a complex type be more easily named by using
+\tcode{typedef}\iref{dcl.typedef} or an \grammarterm{alias-declaration}.
+\end{note}
```
jensmaurer, line 2370: The xref should be after the two parts typedef and alias-decl. "Oftentimes" seems ... novel for the standard.

Suggestion: "A (possibly complex) type can also be named by a _typedef-name_, which is introduced with a `typedef`  declaration or _alias-declaration_ ([dcl.typedef])."

sdkrystian: I agree with the removal of "Oftentimes", but I think "A (possibly complex) type [...]" sounds a little too formal for a note. Maybe something more like "Types can also be named with *typedef-names*, which are introduced by typedef declarations and _alias-declarations_ ([dcl.tyledef]). In situations where a complex declarator ([dcl.decl]) is used to specify a type, it is often easier to use a _typedef-name_."

jensmaurer: I'd drop the "In situations..." sentence, otherwise looks good. (The standard is not a tutorial; getting the cross-reference in is good enough.)

source/declarations.tex
```diff
@@ -2364,8 +2364,11 @@
 \end{example}
 
 \pnum
-A type can also be named (often more easily) by using a
-\tcode{typedef}\iref{dcl.typedef}.
+\begin{note}
+Types can also be named with \grammarterm{typedef-name}{s},
+which are introduced by typedef declarations
+and \grammarterm{alias-declaration}{s}\iref{dcl.typedef}.
```
jensmaurer, line 2370: Could you make the sentence singular throughout, please?

jensmaurer (changes requested):

source/declarations.tex
sdkrystian: Updated

jensmaurer (approved):

PR #3965, opened 2020-04-29
[class.static.mfct] Strike redundant normative wording

sdkrystian:
All of the normative wording here is redundant: the overloading rule is specified in [over.load], the _cv-qualifier-seq_ rule in [dcl.fct], and the virtual rule in [dcl.fct.spec]. 

I added the trailing _requires-clause_ rule for overloading, and also changed "shall" to "cannot" since it shouldn't be used within informative wording.

It seems that we ran out of normative wording in this sub clause with this change, so perhaps more drastic measures should be taken (eventually).

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -2797,12 +2797,12 @@ functions.
 \begin{note}
 A static member function does not have a \tcode{this}
 pointer\iref{class.this}.
+A static member function cannot be declared \tcode{virtual}\iref{dcl.fct.spec}.
+There cannot be a static and a non-static member function with
+the same name, parameter-type-list, and trailing \grammarterm{requires-clause}\iref{over.load}.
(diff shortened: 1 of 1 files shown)


source/classes.tex
```diff
@@ -2797,12 +2797,12 @@
 \begin{note}
 A static member function does not have a \tcode{this}
 pointer\iref{class.this}.
+A static member function cannot be declared \tcode{virtual}\iref{dcl.fct.spec}.
+There cannot be a static and a non-static member function with
+the same name, parameter-type-list, and trailing \grammarterm{requires-clause}\iref{over.load}.
+A static member function cannot not be declared \tcode{const},
```
jensmaurer, line 2803: Uh, this has a "cannot not" bad phrasing. This has been merged already; could you post a follow-up pull request, please?

PR #3966, opened 2020-04-29
[class.dtor] Change redundant wording into note

sdkrystian:
First sentence was recommended by @jensmaurer (non-pure virtual functions are always odr-used), the second is redundant because destructors are non-static data members and inheritance doesn't play into whether a virtual function is overridden (i.e. this is covered by [class.virtual] p1 & p2).

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -2098,9 +2098,10 @@ implicitly defined.
 A prospective destructor can be
 declared \keyword{virtual}\iref{class.virtual}
 and with a \grammarterm{pure-specifier}\iref{class.abstract}.
-If the destructor of a class is virtual and
-any objects of that class or any derived class are created in the program,
-the destructor shall be defined.
+\begin{note}
+Since a pure virtual destructor will be invoked by the destructor of
+a derived class, it must still be defined in most cases; see~\ref{basic.def.odr}.
+\end{note}
 
 \pnum
 \begin{note}


source/classes.tex
```diff
@@ -2264,11 +2264,15 @@
...
+\begin{note}
 If the destructor of a class is virtual and
 any objects of that class or any derived class are created in the program,
-the destructor shall be defined.
-If a class has a base class with a virtual destructor, its  destructor
-(whether user- or implicitly-declared) is virtual.
+the destructor must be defined; see~\ref{basic.def.odr}.
+\end{note}
+\begin{note}
+If a class has a base class with a virtual destructor, its destructor
```
jensmaurer, line 2273: While we're in that area, the "its" has an ambiguous antecedent (it obviously refers to "a virtual destructor").
Suggestion: "If a class C has a base class with a virtual destructor, the destructor of C ... is virtual (xref)."

jensmaurer (approved):

source/classes.tex
```diff
@@ -2264,11 +2264,15 @@
...
 any objects of that class or any derived class are created in the program,
-the destructor shall be defined.
-If a class has a base class with a virtual destructor, its  destructor
-(whether user- or implicitly-declared) is virtual.
+the destructor must be defined; see~\ref{basic.def.odr}.
+\end{note}
+\begin{note}
+If a class \tcode{C} has a base class with a virtual destructor,
+the destructor of \tcode{C} (whether user- or implicitly-declared) is virtual\iref{class.virtual}.
+\end{note}
```
zygoloid, line 2275: This seems entirely wrong, both before and after these changes. If the destructor is virtual and not pure virtual, it must be defined regardless of whether any objects exist. I think the intent is to talk about *pure* virtual destructors here, in which case "any objects of that class or any derived class are created" is not the right constraint either! Merely deriving from the class with a class that has a non-pure virtual destructor is enough to (indirectly) require the pure virtual base class destructor to exist, because the derived class destructor is virtual and therefore odr-used, and it in turn odr-uses the base class destructor.

So I think instead we should say something like:

> [*Note*: A pure virtual destructor must still be defined in most cases, because it is invoked by the destructor of any derived class.]

source/classes.tex
```diff
@@ -2098,9 +2098,10 @@
 A prospective destructor can be
 declared \keyword{virtual}\iref{class.virtual}
 and with a \grammarterm{pure-specifier}\iref{class.abstract}.
-If the destructor of a class is virtual and
-any objects of that class or any derived class are created in the program,
-the destructor shall be defined.
+\begin{note}
+Since a pure virtual destructor will be invoked by the destructor of
+a derived class, it must still be defined in most cases; see~\ref{basic.def.odr}.
```
tkoeppe, line 2103: We shouldn't use "must" in notes, and certainly not in this way: we're not talking about an existing, external requirement here, but we're explaining the reason for the rule in [basic.def.odr].

@jensmaurer, @opensdh, any ideas for how to phrase an "explanation of a rule"? How about:

"Since a pure virtual destructor will be invoked by the destructor of a derived class, it has a definition in most cases; see [basic.def.odr]."

sdkrystian: "Since a pure virtual destructor will be invoked by the destructor of a derived class, a definition is generally required; see [basic.def.odr]." 

Another alternative would be "usually required".

tkoeppe: @zygoloid suggested:

"Because a pure virtual destructor will be invoked by the destructor of a derived class, [basic.def.odr] requires a definition in most cases."

I'm a bit hesitant about having "requirements in notes", which we aren't allowed to, and maybe this latter form is a bit further from the forms ISO considers "requirements" than "is required" (https://www.iso.org/sites/directives/current/part2/index.xhtml#_idTextAnchor082).

opensdh: How about
> [_Note_: Even a pure virtual destructor is odr-used by any derived concrete class.  &mdash;&nbsp;_end note_]
to avoid talking about any requirement at all?  Otherwise, you can use statements like "… is necessary" that more clearly appear to be (mere) statements of fact.

sdkrystian: I don't think that's quite correct, since an abstract derived class will call the destructor of a direct non-virtual abstract base class. I'm also not a fan of starting the sentence with "Even", as it implies elaboration on a previous broader statement. How about:
> [_Note:_ The destructor of a derived class will odr-use ([basic.def.odr]) the selected destructor of a direct base class -- even if that destructor is pure virtual. -- _end note_]

Or
> [_Note:_ A pure virtual selected destructor will be odr-used ([basic.def.odr]) by any derived classes. -- _end note_]

opensdh: The note doesn't have to be exhaustive; having a concrete class is a sufficient but not necessary condition for the odr-use.  The first version seems correct with minor edits:

> [_Note:_ A definition of a destructor of a derived class odr-uses ([basic.def.odr]) the selected destructor of a direct base class, even if that destructor is pure virtual.  &mdash;&nbsp;_end note_]

I don't think the second is right, since those derived classes could all have pure virtual destructors with no definition.

sdkrystian: The second version is intended to be more concise but less precise... I'll apply your suggested changes to the first version next week :)

tkoeppe: @sdkrystian Friendly ping.

PR #3974, opened 2020-05-03
[basic.fundamental] Provide more information about which types are integral

sdkrystian:
Saying "standard and extended integer types" provides more information to the reader than "signed and unsigned integer types" (integer type implies signed and unsigned) and doesn't alter meaning.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -5074,7 +5074,7 @@ Types
 \tcode{bool},
 \tcode{char}, \tcode{wchar_t},
 \tcode{char8_t}, \tcode{char16_t}, \tcode{char32_t},
-and the signed and unsigned integer types are
+and the standard and extended integer types are
 collectively called
 \defnx{integral types}{integral type}.
 A synonym for integral type is \defn{integer type}.


sdkrystian:
@jensmaurer It's certainly a trivial change, but I see the question of "are extended integer types integral types?"/"Do the standard conversions apply to extended integer types?" more often than "do the standard integer types include unsigned and signed standard integer types" (not seen this one yet, luckly :P). 

Specifying standard/extended subtly provides more information to the reader in the context of that single paragraph, which I believe is useful as paragraphs in the standard are typically not read sequentially.

zygoloid:
My first reaction was that this change is incorrect: we should just define that "integer types" are "standard and extended integer types". But it appears that `bool`, `char`, `wchar_t`, and `charN_t` aren't actually "standard integer types". That seems like a normative error to me. Consider the (very few) uses of "standard integer type":

[conv.rank]/1.5: "The rank of any standard integer type shall be greater than the rank of any extended integer type with the same width."
[conv.rank]/1.7: "The rank of bool shall be less than the rank of all other standard integer types."

These both obviously meant to include all non-extended integer types.

[utility.intcmp]/1, /5, /9: "Mandates: Both T and U are standard integer types or extended integer types (6.8.2)."

I have no idea what the intent was here.

This needs CWG and LWG attention.

jensmaurer:
I think we can remove this from CWG scope, because we have

conv.rank p1.8: "The ranks of char8_t, char16_t, char32_t, and wchar_t shall equal the ranks of their underlying types"
and we have transitivity via p1.10, so those are covered.

We should fix "[conv.rank]/1.7: "The rank of bool shall be less than the rank of all <del>other</del> standard integer types."
to avoid the hint that "bool" is a standard integer type.

jwakely:
We chose the wording in [utility.intcmp] *specifically* because it doesn't include `bool`, `char` etc.

The wording matches the intent.

PR #3976, opened 2020-05-03
[dcl.enum] Turn redundant and informal wording into a note

sdkrystian:


source/declarations.tex
```diff
@@ -6921,6 +6921,7 @@
 \end{codeblock}
 \end{example}
 \indextext{member!enumerator}%
+\begin{note}
 An enumerator declared in class scope can be referred to using the class
 member access operators (\tcode{::}, \tcode{.} (dot) and \tcode{->}
```
jensmaurer, line 6926: I appreciate we might want a change here, but maybe we should start with the obvious stuff here: `::` is not a class member access operator and is not specified in [expr.ref]. Maybe [class.qual] is a better target for that case.

source/declarations.tex
```diff
@@ -6921,6 +6921,7 @@
...
 (arrow)), see~\ref{expr.ref}.
```
jensmaurer, line 6927: There should be a semicolon before "see" (instead of a comma).

source/declarations.tex
sdkrystian: How about:
> An enumerator declared in class scope can be referred to using the class
member access operators `.` (dot) and `->` (arrow) ([expr.ref]), or with the scope resolution operator `::`([class.qual]).

PR #3977, opened 2020-05-03
[basic, class, expr] Replace enumerated type with enumeration type

sdkrystian:
We already define _enumerated type_ in [enumerated.types], and we only use it a handful of times in core. We should strike it to prevent any confusion and to improve consistency.


zygoloid (approved):

jensmaurer:
Please rebase + force-push.

PR #3983, opened 2020-05-06
[vector.capacity] Remove duplicate \pnum

### LINKED ISSUE #3982: [vector.capacity]p17/18 has two paragraph numbers
[vector.capacity]/17 has two paragraph numbers, that are rendered superimposed on top of each other.

We should fix this for the C++20 IS.

jensmaurer:
Also prevent future occurrences by updating check.sh

Fixes #3982

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -5947,11 +5947,11 @@ constexpr void resize(size_type sz, const T& c);
 
 \begin{itemdescr}
 \pnum
-\pnum
 \expects
 \tcode{T} is
 \oldconcept{CopyInsertable} into \tcode{*this}.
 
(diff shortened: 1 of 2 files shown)


source/containers.tex
```diff
@@ -5947,7 +5947,6 @@
 
 \begin{itemdescr}
 \pnum
-\pnum
```
CaseyCarter, line 5950: As https://github.com/cplusplus/draft/issues/3982#issuecomment-625452674 points out, this should be moved to the following `\expects` paragraph.

jensmaurer: Done.

PR #3984, opened 2020-05-06
[concepts.equality] Replace spurious 'this document'

### LINKED ISSUE #3980: [concepts.equality] Spurious "this document" wording
In [concepts.equality] there are several normative phrases referring to "this document" such as in p3, p4, and p5. This phrase is used to define the locality of the term equality preservation which again has impact on how _requires-expression_'s are supposed to be interpreted. In theory "this document" can be interpreted in the widest possible form and could cover the complete International Standard (including core language). After a talk with Casey about the reason for this wording, he explained to me that this is a spurious left-over from the original Ranges-TS and is intended to cover the complete Standard Library specification. This issue is a modest attempt trying to ask whether "this document" in this [concepts.equality] subclause can be editorially changed to "throughout the library", or whether such a request would be non-editorially and would instead require an LWG issue.

jensmaurer:
with 'the library Clauses'.

Fixes #3980.

### DIFF
--- a/source/concepts.tex
+++ b/source/concepts.tex
@@ -53,7 +53,7 @@ being equality-preserving. The \term{domain} of an expression is the set of
 input values for which the expression is required to be well-defined.
 
 \pnum
-Expressions required by this document to be equality-preserving are further
+Expressions required to be equality-preserving are further
 required to be stable: two evaluations of such an expression with the same input
 objects are required to have equal outputs absent any explicit intervening
 modification of those input objects.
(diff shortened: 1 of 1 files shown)


source/concepts.tex
```diff
@@ -53,7 +53,7 @@
 input values for which the expression is required to be well-defined.
 
 \pnum
-Expressions required by this document to be equality-preserving are further
+Expressions required by the library Clauses to be equality-preserving are further
```
CaseyCarter, line 56: This sentence really doesn't need a qualifier at all: "Expressions required to be equality-preserving are further required to..." seems just fine.

jensmaurer: Fixed.

source/concepts.tex
```diff
@@ -69,15 +69,15 @@
 \end{note}
 
 \pnum
-Expressions declared in a \grammarterm{requires-expression} in this document are
+Expressions declared in a \grammarterm{requires-expression} in the library Clauses are
```
zygoloid, line 72: We define "library clauses" with a lowercase C in [intro.structure]/2.

PR #3987, opened 2020-05-07
Cleanup of [class.virtual]

sdkrystian:
This is an attempt to make our definition of "override" and "final overrider" more consistent with its use, and includes some small cleanups along the way.

In [[class.cdtor] p4](http://eel.is/c++draft/class.cdtor#4) we have the following wording:
> [...] the function called is the final overrider in the constructor's or destructor's class [...]

We also have the following wording in [[expr.call] p3](http://eel.is/c++draft/expr.call#3):
> Otherwise, its final overrider in the dynamic type of the object expression is called

In both of these cases, we refer to the final overrider relative to another class. Changing the definition to match the use here seems reasonable. The approach taken here is to define the final overrider of a function relative to a class (as far as I can tell this wording shouldn't change semantics; it accounts for redeclared members not being members of the derived class).

There also was a missing \grammarterms around ref-qualifier, and cv-qualification was changed to be cv-qualifier-seq to match function declarators. Since we no longer need to consider virtual functions to override themselves, it allows us to simplify the wording a bit in other places.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -3748,7 +3748,7 @@ by \tcode{X} and \tcode{Y}, as shown in \fref{class.virtnonvirt}.
 
 \pnum
 A non-static member function is a \defnadj{virtual}{function}
-if it is first declared with the keyword \tcode{virtual} or
+if it is declared with the keyword \tcode{virtual} or
 if it overrides a virtual member function declared in a base class
 (see below).\footnote{The use of the \tcode{virtual} specifier in the
 declaration of an overriding function is valid but redundant (has empty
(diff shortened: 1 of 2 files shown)


AlisdairM:
This looks like a nice cleanup, but has no feedback.  Would we be interested in seeing a rebase to land for C++26 FDIS?

source/classes.tex
```diff
@@ -3748,7 +3748,7 @@
 
 \pnum
 A non-static member function is a \defnadj{virtual}{function}
-if it is first declared with the keyword \tcode{virtual} or
+if it is declared with the keyword \tcode{virtual} or
```
jensmaurer, line 3751: I think the "first" wants to say that adding "virtual" on the out-of-class definition does nothing.

jensmaurer (changes requested):

jensmaurer:
This is more than moving words around and thus needs CWG review.

PR #3990, opened 2020-05-09
Reference the normative text for data member order

TBBle:
[CWG2404](bb641cba40211374d8b2dd3472f42274de662761) demoted this paragraph to a note, but didn't include a reference to _why_ the ordering of non-static data member addresses was still defined.

Credit to [aschepler on StackOverflow](https://stackoverflow.com/a/61590855/166389) for the reference.


TBBle:
I haven't actually tested that this does exactly the right thing, format-wise, as I don't have local pipeline set up for rendering the draft at this time. So I have no concerns if the same change needs to be redone in a different manner.

PR #4005, opened 2020-05-21
[std] Harmonize phrasing 'terminate is invoked'

### LINKED ISSUE #3079: "terminate is called" inconsistency
Clause [except] consistently says "the function `std::terminate` is called".

[support.start.term] says "the function `std::terminate` shall be called" (which is an incorrect use of "shall", and redundant `std`-qualification).

[terminate.handler] says "The type of a _handler function_ to be called by `std::terminate()` when terminating exception processing" (redundant qualification, completely bogus to use `()` when referring to a function by name, rather than describing a function call expression).

[execpol.seq], [execpol.par], [execpol.parunseq], and [execpol.unseq] say "`terminate()` shall be called" (incorrect "shall", unclear if it's trying to name the function or show a call expression, but arguably either is correct).

[stopsource] and [stopcallback.constr] (added by P0660R10, #3072) say "`terminate()` is called".

[thread.thread.constr] and [thread.jthread.constr] (the latter added by P0660R10 #3072) say "`terminate` shall be called" (incorrect "shall").

[thread.thread.destr] and [thread.thread.assign] say "calls `terminate()`".

[thread.condition.condvar] says "`terminate()` shall be called" many times (incorrect "shall").

This is a mess.

jensmaurer:
instead of saying 'is called'.
Also add cross-references to [except.terminate].
Also use 'exits via an exception' consistently.

Fixes #3079.

### DIFF
--- a/source/exceptions.tex
+++ b/source/exceptions.tex
@@ -334,7 +334,7 @@ it is rethrown.
 If the exception handling mechanism
 handling an uncaught exception\iref{except.uncaught}
 directly invokes a function that exits via an
-exception, the function \tcode{std::terminate} is called\iref{except.terminate}.
+exception, the function \tcode{std::terminate} is invoked\iref{except.terminate}.
 \begin{example}
 \begin{codeblock}
 struct C {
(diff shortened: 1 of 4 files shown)


jensmaurer:
@tkoeppe, could you review this, please? If you agree, I would apply without consulting @zygoloid .

tkoeppe:
Agreed, this looks very nice -- esp. the cleanup around the exit handlers looks like an improvement. Ship it!

PR #4008, opened 2020-05-22
Fix index entries for CPOs

### LINKED ISSUE #4003: [concept.swappable] Missing an index entry for ranges::swap
The `ranges::swap` CPO is in the `<concepts>` header, quite removed from the original `swap` function in the `<utility>` header.  It should have its own index entry so that it is easy to find.

jensmaurer:
Fixes #4003


CaseyCarter (approved):
What's here is good, but I see some of the range adaptor objects (the ones that live in `std::ranges::views`) are unindexed. I see an entry for `all`, for example, but not `empty` or `filter` (I didn't audit all of them). Is that omission intentional? I imagine the objects that share the name of an algorithm. e.g., `transform`, will be problematic.

PR #4013, opened 2020-05-24
[thread.mutex.requirements] Harmonize wording for try_lock.

### LINKED ISSUE #3996: Inconsistent wording for mutex types
In Clause [thread] we have:

- Returns: true if the lock was acquired, false otherwise.
- Returns: true if the lock was acquired, false otherwise.
- Returns: true if the lock was acquired, false otherwise.  
- Returns: true if ownership of the mutex was obtained for the calling thread, otherwise false.
- Returns: true if ownership was obtained, otherwise false.
- Returns: true if ownership was obtained, otherwise false.  
- Returns: true if the shared ownership lock was acquired, false otherwise.
- Returns: true if the shared lock was acquired, false otherwise.
- Returns: true if the shared lock was acquired, false otherwise.

I don't see any reason why:

- We use "lock was acquired" in some places and "ownership was obtained" in others.
- We say "false otherwise" sometimes and "otherwise false".
- In eight places we say "ownership" but then in one place we say "ownership of the mutex".
- In two places we say "shared lock" but in one place we say "shared ownership lock".

jensmaurer:
Fixes #3996

### DIFF
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -249,7 +249,7 @@ exception is thrown then a lock shall not have been acquired for the current exe
 
 \pnum
 \returns
-\tcode{true} if the lock was acquired, \tcode{false} otherwise.
+\tcode{true} if the lock was acquired, otherwise \tcode{false}.
 \end{itemdescr}
 
 \rSec3[thread.req.lockable.timed]{\oldconcept{TimedLockable} requirements}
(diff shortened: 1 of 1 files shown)


zygoloid:
Merged to master, but I'm not convinced this is important enough to put into C++20 right before publication.

jensmaurer:
Removed C++20 milestone.

PR #4018, opened 2020-05-29
[res.on.expects] Replace obsolete term 'expects' in the title.

### LINKED ISSUE #4016: [res.on.expects] is still titled "Expects paragraph"
Even though we've replaced _Expects:_ with _Preconditions:_.

jensmaurer:
Fixes #4016

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -2976,7 +2976,7 @@ the behavior is undefined unless otherwise specified.
 This applies even to objects such as mutexes intended for thread synchronization.
 \end{note}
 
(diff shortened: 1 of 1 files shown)


jensmaurer:
Looks plausible. The status quo label was introduced with commit 829b921bdc8be954fc3680088ffc38dc1d4d106a just a few months ago.

jensmaurer:
(The Travis failure is due to a missing xrefdelta entry. But we don't need one if we rename this in C++20 as well.)

jensmaurer:
Obsoleted by #4042.

PR #4026, opened 2020-06-06
[class.default.ctor] Fix sloppy wording.

### LINKED ISSUE #4019: [class.default.ctor]/4 /5 Constructors do not create objects
/4: A default constructor that is defaulted and not defined as deleted is implicitly defined when it is odr-used ([basic.def.odr]) to **create** an object of its class type

/5: Default constructors are called implicitly to **create** class objects of static, thread, or automatic storage duration ([basic.stc.static], [basic.stc.thread], [basic.stc.auto]) defined without an initializer ([dcl.init]), are called to **create** class objects of dynamic storage duration

jensmaurer:
Initialization is fully specified in [dcl.init],
so turn the list of default-initialization cases
into a note and shorten it appropriately.

Fixes #4019

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -1304,7 +1304,7 @@ that is defaulted and not defined as deleted
 is
 \defnx{implicitly defined}{constructor!implicitly defined}
 when it is odr-used\iref{term.odr.use}
-to create an object of its class type\iref{intro.object},
+to initialize an object of its class type\iref{intro.object},
 when it is needed for constant evaluation\iref{expr.const}, or
 when it is explicitly defaulted after its first declaration.
 The implicitly-defined default constructor performs the set of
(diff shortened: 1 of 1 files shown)


source/classes.tex
```diff
@@ -1445,17 +1445,20 @@
...
+\item
+to initialize a class object of static, thread,
+or automatic storage duration~(\ref{basic.stc.static}, \ref{basic.stc.thread}, \ref{basic.stc.auto})
+defined without an initializer\iref{dcl.init},
+\item
+to initialize a class object of dynamic storage duration\iref{basic.stc.dynamic}
+created by a \grammarterm{new-expression}
+in which the \grammarterm{new-initializer} is omitted\iref{expr.new}, or
+\item
+when the explicit type conversion syntax\iref{expr.type.conv} is used.
```
zygoloid, line 1460: I'm not sure this last case is an "invoked implicitly" case. If we want to list this here, then perhaps we should also cover `new T()` and the various list-initialization cases?

source/classes.tex
```diff
@@ -1445,17 +1445,20 @@
...
+\end{itemize}
 A program is ill-formed if the default constructor for an object
 is implicitly used and the constructor is not accessible\iref{class.access}.
```
zygoloid, line 1463: Maybe replace "used" with "invoked" to make it clear that this is talking about the same set of things we just mentioned.

Please consider reordering the note after the normative sentence. As it stands, it sounds like the normative rule uses a term defined in the note; reordering the two would give the (presumably intended) impression that the note is just a summary of things said elsewhere.

source/classes.tex
jensmaurer: Well, I just bulletized the pre-existing text.

If you believe that "explicit type conversion" is already properly handled elsewhere, then we can strike it here.

Speaking of which, it seems the entire "implicitly invoked" paragraph talking about omitted initializers ought to be covered by [dcl.init] p11 and p7.  In particular, p11 talks about overload resolution, so we get access checking for free.

It feels the entire paragraph can become a note.

source/classes.tex
jensmaurer: used -> invoke done

I don't understand the second part of the comment; the notes in the vicinity are totally unrelated.

source/classes.tex
jensmaurer: Turned into a note and shortened substantially.

zygoloid (commented):
LGTM

tkoeppe:
@jensmaurer: Could you try and come up with a more descriptive commit message? I'm not sure how to best say it, but maybe something slightly more specific? Then feel free to merge immediately, e.g. just edit the message as part of a "squash". Thanks!

PR #4031, opened 2020-06-09
[predef.iterators] Singularize heading of singular sentinels.

### LINKED ISSUE #4028: [default.sentinels, unreachable.sentinels] Inconsistent grammatical numbering and hanging paragraph
I think [[default.sentinels]](http://wg21.link/default.sentinels) should be titled in singular like [[unreachable.sentinels]](http://wg21.link/unreachable.sentinels), because there's only one type, unlike the sibling iterators that are class templates. Consider using singular in the stable name, too. But for that, [[unreachable.sentinel]](http://wg21.link/unreachable.sentinel) [would have to be dissolved](https://github.com/cplusplus/draft/issues/3525) (note that this one is in singular).

jensmaurer:
Also adjust stable labels.

Fixes #4028.


jensmaurer:
Updated commit message.

source/iterators.tex
```diff
@@ -377,7 +377,7 @@
   template<@\libconcept{input_iterator}@ I, class S>
     struct iterator_traits<common_iterator<I, S>>;
 
-  // \ref{default.sentinels}, default sentinels
+  // \ref{default.sentinel}, default sentinels
```
johelegp, line 380:
```suggestion
  // \ref{default.sentinel}, default sentinel
```

jensmaurer:
@zygoloid , this has a failed Travis check, because it renames some stable labels that were introduced post-C++17. We should fix these for C++20. Please advise how to avoid the Travis failure.

zygoloid:
> @zygoloid , this has a failed Travis check, because it renames some stable labels that were introduced post-C++17. We should fix these for C++20. Please advise how to avoid the Travis failure.

Build from the C++20 branch, grab the produced `xrefindex.glo` file, and copy that file over `source/xrefprev` on the master branch.

jensmaurer:
Rebased + copied xrefprev as advised.

"xrefprev" also contains page numbers, which is immaterial for its purpose and makes the "diff" larger than it should be. We should only have a preprocessed representation of the old cross-reference in git.

jensmaurer:
This change (but not the xrefprev repair) should also be applied to C++20, in which case the Travis error "missing default.sentinels" will (indirectly) vanish, after updating xrefprev.

PR #4037, opened 2020-06-11
[util.smartptr.weak] Remove declarations already in [memory.syn]

johelegp:
Partially addresses #4041.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -11007,10 +11007,6 @@ namespace std {
 
   template<class T>
     weak_ptr(shared_ptr<T>) -> weak_ptr<T>;
-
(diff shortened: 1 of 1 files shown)


jensmaurer:
@jwakely , can you confirm?

jensmaurer:
@jwakely , ping?

jwakely (commented):
It might be consistent, but I'm not sure it's actually an improvement. The declaration of `swap` in [memory.syn] is very far away. This declaration puts [util.smartptr.weak.spec] in context.

jwakely:
I won't object if the editor (or editorial team) wants to do this though.

jensmaurer:
Editorial meeting: Remove redundancy = apply.

PR #4045, opened 2020-06-15
[structure.specifications] Integrate [res.on.expects].

jensmaurer:
The latter consisted of a single sentence that is
best integrated into the place where the Preconditions:
element is introduced.

Also fixes LWG3168.

Fixes #4042 for C++20.

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -620,7 +620,8 @@ and also define the function as deleted.
 \item
 \expects
 the conditions
-that the function assumes to hold whenever it is called.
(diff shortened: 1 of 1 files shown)


johelegp:
Is it intentional that there's no redirecting xref from [res.on.expects] to [structure.specifications]?

jwakely:
Yes, because [res.on.expects] was never in any published standard.

jwakely:
It was called [res.on.required] in C++17, and that is listed in Annex F.

johelegp:
Ah, that makes sense. Thank you.

PR #4051, opened 2020-06-20
[container.requirements] Improve index entries for contiguous and reversible containers.

Eelis:
image

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -366,10 +366,11 @@ with value \tcode{a.end()} before the swap will have value \tcode{b.end()} after
 swap.
 
 \pnum
+\indextext{reversible container|see{container, reversible}}%
 If the iterator type of a container belongs to the bidirectional or
 random access iterator categories\iref{iterator.requirements},
 the container is called
-\term{reversible}
+\defnx{reversible}{container!reversible}
 and meets the additional requirements
 in \tref{container.rev.req}.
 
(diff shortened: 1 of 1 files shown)


source/containers.tex
```diff
@@ -366,10 +366,11 @@
 swap.
 
 \pnum
+\indextext{reversible container|see{container, reversible}}%
 If the iterator type of a container belongs to the bidirectional or
 random access iterator categories\iref{iterator.requirements},
 the container is called
-\term{reversible}
+\defnx{reversible}{container!reversible}
```
jensmaurer, line 373: Please use \defnadj{reversible}{container} and drop the "see" above (which comes for free now).

source/containers.tex
```diff
@@ -465,8 +466,8 @@
 within that container.
 
 \pnum
-\indextext{container!contiguous}%
-A \defn{contiguous container}
+\indextext{contiguous container|see{container, contiguous}}%
+A \defnx{contiguous container}{container!contiguous}
```
jensmaurer, line 470: Same here: \defnadj, please

source/containers.tex
Eelis: Hmm, using \defnadj would result in "reversible container" in the running text, but here we only want the word "reversible" to appear in the running text. \defnadjx also seems unsuitable for the same reason.

Maybe there ought to be a macro for defining an adjective while only making the adjective itself (and not the noun) appear in the running text?

jensmaurer: Ah, you're right. But that doesn't apply to the second change, below.

source/containers.tex
Eelis: Done, thanks!

Eelis:
Amended to use \defnadj.

PR #4056, opened 2020-06-24
[expr.const] Add missing paragraph break

johelegp:


jensmaurer:
I'm not convinced we should have a \pnum here. We use "permitted result of a constant expression" right before the bulleted list here.

johelegp:
I agree, because there's other places in [expr.const] that could similarly have a \pnum added.

My intention was to make grammatical paragraphs map to paragraph number. I vaguely remember seeing this kind of inconsistency elsewhere, but if the editors agree to this mapping, I could make a more general PR for this subclause. Otherwise, this should be closed.

zygoloid:
I think this is intentionally structured as it is: this is all part of the same definition of "constant expression", which uses an auxiliary term recursively defined in terms of itself. Keeping this all in the same paragraph seems important -- if nothing else -- to identify the referent of "the above constraints".

PR #4057, opened 2020-06-25
[basic.fundamental] Use canonical types in Table 12

sdkrystian:
In all other places in this sub-clause we use the canonical forms of type names as described in [dcl.type.simple]. We should do so here as well.


jensmaurer:
Let's call this a typo fix.

PR #4058, opened 2020-06-25
[class.derived] Remove second definition of base class subobject

sdkrystian:
This is already defined in [intro.object].

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -3528,8 +3528,8 @@ class\iref{dcl.init.ref}.
 \end{note}
 
 \pnum
-The \grammarterm{base-specifier-list} specifies the type of the
-\term{base class subobjects} contained in an
+The \grammarterm{base-specifier-list} specifies the types of the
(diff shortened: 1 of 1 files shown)


jensmaurer (changes requested):
I can't find an italics definition of "base class subobject" in [intro.object]. Can you point me to one?

sdkrystian:
@jensmaurer http://eel.is/c++draft/intro.object#def:base_class_subobject

jensmaurer:
Thanks. Unfortunately, the perceived definition in [intro.object] doesn't seem to be one. The mention in [class.derived] seems to be more like a definition, although there is certainly room for improvement.

In short, I think the proposed change is not an improvement.

jensmaurer:
Editorial meeting: We need to talk about "base class object" during name lookup, where we have no concrete object in sight. class.derived: Talk about base classes, not base class subobjects. Turn into a core issue.

tkoeppe:
@jensmaurer Could you please create this CWG issue?

PR #4061, opened 2020-06-26
[dcl.array] Clarify that arrays do not have extra padding.

### LINKED ISSUE #3791: [dcl.array] Potentially ambiguous specification of array contents
[[dcl.array] p6](http://eel.is/c++draft/dcl.array#6) states:
> An object of type “array of _N_ `U`” contains a contiguously allocated non-empty set of _N_ subobjects of type `U`, known as the elements of the array, and numbered `0` to `N -1`.

The use of contains here does not seem to be exclusive, and could potentially be interpreted to say that padding is permitted before the first element, allowing for the address of an array and its first element to differ (further, we don't normatively state that an array and it first element have the same address)

jensmaurer:
Fixes #3791

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -3084,7 +3084,7 @@ has cv-qualified type; see~\ref{basic.type.qualifier}.
 \end{note}
 
 \pnum
-An object of type ``array of \tcode{N} \tcode{U}'' contains
+An object of type ``array of \tcode{N} \tcode{U}'' consists of
 a contiguously allocated non-empty set
 of \tcode{N} subobjects of type \tcode{U},
 known as the \defnx{elements}{array!element} of the array,


source/declarations.tex
```diff
@@ -3084,7 +3084,7 @@
 \end{note}
 
 \pnum
-An object of type ``array of \tcode{N} \tcode{U}'' contains
+An object of type ``array of \tcode{N} \tcode{U}'' consists of
 a contiguously allocated non-empty set
```
languagelawyer, line 3088: Is «contiguously allocated» normative? If not, could it be fixed editorially in this PR or it should be a separate PR?

jensmaurer: What's wrong with "contiguously allocated"?

languagelawyer: I've thought we agreed in core reflector that it is not normative, [like the wording about how member subobjects are allocated](https://github.com/cplusplus/draft/commit/bb641cba40211374d8b2dd3472f42274de662761).

jensmaurer: I don't remember. Do you have a pointer?

languagelawyer: https://lists.isocpp.org/core/2020/01/8112.php

jensmaurer: The participation on the core reflector was low for this question. I'd like to restrict this pull request to the obvious fix "contains" -> "consists of".

PR #4062, opened 2020-06-27
[span.iterators] Specify aliases, like [string.view.iterators] does

johelegp:


source/containers.tex
```diff
@@ -11236,7 +11236,10 @@
 requirements\iref{random.access.iterators},
 and
 meets the requirements for
-constexpr iterators\iref{iterator.requirements.general}.
+constexpr iterators\iref{iterator.requirements.general},
+whose \tcode{value_type} is the template parameter \tcode{ElementType}.
```
CaseyCarter, line 11240: This should be `remove_cv_t<ElementType>`; value types are not cv-qualified.

johelegp: Thank you.

source/containers.tex
```diff
@@ -11237,7 +11237,8 @@
 and
 meets the requirements for
 constexpr iterators\iref{iterator.requirements.general},
-whose \tcode{value_type} is \tcode{remove_cv_t<ElementType>}.
+whose \tcode{value_type} is \tcode{remove_cv_t<ElementType>} and
+whose \tcode{reference_type} is \tcode{ElementType&}.
```
CaseyCarter, line 11241: Since `span` has nested types `value_type` and `reference`, why not simply use those? (i.e., `whose value type is \tcode{value_type} and whose reference type is \tcode{reference}`)

johelegp: Excellent!

johelegp:
I was trying to specify what seems underspecified, which I noticed when looking at `string_view::const_iterator`. Perhaps the problem lies elsewhere and an LWG issue is required.

Now I'm left wondering if the `reference` of `string_view::const_iterator` is underspecified. They specify that "All requirements on container iterators ([container.requirements]) apply to _the-view-type_::​iterator as well", and [container.requirements]p6 says that "`begin()` returns an iterator referring to the first element in the container", but neither view is a container.

jensmaurer:
The commit description seems broken. "specify aliases" feels misleading.
Maybe "Specify iterator value_type and reference"

And please squash all commits and force-push.

source/containers.tex
```diff
@@ -11236,7 +11236,11 @@
 requirements\iref{random.access.iterators},
 and
 meets the requirements for
-constexpr iterators\iref{iterator.requirements.general}.
+constexpr iterators\iref{iterator.requirements.general},
+whose \tcode{value_type} is \tcode{value_type} and
+whose \tcode{reference} type is \tcode{reference}.
```
CaseyCarter, line 11241: It's not clear what entities `\tcode{value_type}` and `\tcode{reference}` refer to here. Presumably we mean member type aliases of `iterator_traits<iterator>`, so we could say `such that \tcode{iterator_traits<iterator>::value_type} is \tcode{value_type}` (and similarly for `reference`) but I think it would be clearer to simply use prose "value type" and "reference type" which name those properties of iterators:
```suggestion
whose value type is \tcode{value_type} and
whose reference type is \tcode{reference}.
```

johelegp: Looks good, but deviates from [string.view.iterators], which uses `value_type`. If the editors prefer this, I'd also like to extend the PR to [string.view.iterators] so that it also reads "value type".

jensmaurer: Yes, I think that would be an improvement. Having `\tcode{value_type}` mean different things two words apart is not good.

jensmaurer (changes requested):
Please follow Casey's suggestion.

johelegp:
Decided against my suggested change to [string.view.iterators].

jensmaurer (approved):

jensmaurer:
@CaseyCarter , are you ok with this end result?

CaseyCarter (approved):
LGTM!

source/containers.tex
CaseyCarter: Feel free to resolve these comments, @johelegp. I can't since I'm a happily powerless contributor here =)

PR #4064, opened 2020-06-29
[special] Remove index entry for 'X(X&)'.

Eelis:
This index entry seems useless because how would one even know to look for "X(X&)" in the index, given that the name X is arbitrary? :)


jensmaurer:
In particular since `X(const X&)` is not mentioned.

Issue #3677, opened 2020-02-14
Check element order programmatically and presubmit


tkoeppe:
The specification elements (*Remarks":, "Preconditions*:, etc.) should appear in a specific, fixed order; we could try and check that this is the case programmatically and run this check as part of the presubmit checks.

jwakely:
My script found the following (at commit b8e16d5bd8f6b4c8f214a79900da4d0f4ea197c5):

```
algorithms.tex:9945: \expects after \effects
containers.tex:848: \expects after \effects
containers.tex:855: \effects after \ensures
containers.tex:894: \expects after \effects
containers.tex:901: \expects after \effects
containers.tex:908: \expects after \effects
containers.tex:928: \expects after \effects
containers.tex:1103: \expects after \effects
containers.tex:1113: \expects after \effects
containers.tex:1125: \expects after \effects
containers.tex:1136: \expects after \effects
containers.tex:1147: \expects after \effects
containers.tex:1159: \expects after \effects
containers.tex:1171: \expects after \effects
containers.tex:1180: \expects after \effects
containers.tex:3773: \expects after \effects
containers.tex:3795: \expects after \effects
iostreams.tex:2208: \effects after \ensures
iostreams.tex:2318: \effects after \ensures
iostreams.tex:3961: \expects after \effects
iostreams.tex:10799: \ensures after \throws
lib-intro.tex:595: \constraints after \expects
lib-intro.tex:607: \mandates after \expects
lib-intro.tex:2193: \effects after \ensures
lib-intro.tex:3001: \expects after \throws
numerics.tex:1414: \expects after \throws
strings.tex:2282: \effects after \throws
strings.tex:2423: \effects after \throws
strings.tex:2449: \effects after \throws
strings.tex:2477: \effects after \throws
strings.tex:2505: \effects after \throws
strings.tex:2612: \effects after \throws
strings.tex:2652: \effects after \throws
strings.tex:3007: \effects after \throws
strings.tex:4484: \expects after \throws
strings.tex:4488: \effects after \throws
strings.tex:4514: \effects after \throws
support.tex:2229: \expects after \effects
support.tex:2241: \expects after \effects
support.tex:2246: \expects after \effects
support.tex:2322: \expects after \effects
support.tex:2334: \expects after \effects
support.tex:2339: \expects after \effects
support.tex:2482: \expects after \effects
support.tex:2494: \expects after \effects
support.tex:2499: \expects after \effects
support.tex:2555: \expects after \effects
[...]

Issue #2209, opened 2018-06-22
[time] Inconsistent ordering of elements of description


johelegp:
Throughout [time], the ordering of _Remarks:_ (mostly), _Throws:_ and probably some other elements is inconsistent with their ordering in the rest of the library, which follow the order as described in https://wg21.link/structure.specifications#3. There are also local inconsistencies in the ordering.

jensmaurer:
As far as I can see (from sampling the first few occurrences of "Remarks"), it's inconsistently ordered in exactly those cases where it talks about participation in overload resolution. This will be replaced by a \constraints item in the near future, I'd guess, which is sorted correctly near the top.

PR #4067, opened 2020-07-03
[lib] Canonicalize order of library descriptive elements.

jensmaurer:
Partially addresses #3677

(There are 6 remaining mis-orderings that need more attention.)

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -3265,15 +3265,12 @@ to an integral type~(\ref{conv.integral}, \ref{class.conv}).
 
 \pnum
 \expects
+\tcode{n >= 0} is \tcode{true}.
 \tcode{Function} meets the \oldconcept{MoveConstructible} requirements.
 \begin{note}
 \tcode{Function} need not meet
 the requirements of \oldconcept{CopyConstructible}.
 \end{note}
-
-\pnum
-\expects
-\tcode{n >= 0} is \tcode{true}.
 
 \pnum
 \effects
@@ -3308,11 +3305,8 @@ to an integral type~(\ref{conv.integral}, \ref{class.conv}).
 
 \pnum
 \expects
-\tcode{Function} meets the \oldconcept{CopyConstructible} requirements.
-
-\pnum
-\expects
 \tcode{n >= 0} is \tcode{true}.
+\tcode{Function} meets the \oldconcept{CopyConstructible} requirements.
 
 \pnum
 \effects
@@ -5203,13 +5197,13 @@ Let $j$ be the end of the resulting range. Returns:
 \end{itemize}
 
 \pnum
-\remarks
-Stable\iref{algorithm.stable}.
-
-\pnum
 \complexity
 Exactly \tcode{last - first} applications
 of the corresponding predicate and any projection.
+
+\pnum
+\remarks
+Stable\iref{algorithm.stable}.
 
 \pnum
 \begin{note}
@@ -6319,7 +6313,6 @@ the \oldconcept{MoveConstructible} (\tref{cpp17.moveconstructible}) and
 \oldconcept{\-Move\-Assignable} (\tref{cpp17.moveassignable}) requirements.
 
 \pnum
-\expects
 For iterators \tcode{a1} and \tcode{b1} in \range{first}{last}, and
 iterators \tcode{x2} and \tcode{y2} in \range{result_first}{result_last},
 after evaluating the assignment \tcode{*y2 = *b1}, let $E$ be the value of
@@ -8111,17 +8104,17 @@ For the first form, \tcode{T} meets the
 \pnum
 \returns
 The smaller value.
-
-\pnum
-\remarks
 Returns the first argument when the arguments are equivalent.
-An invocation may explicitly specify
-an argument for the template parameter \tcode{T}
-of the overloads in namespace \tcode{std}.
 
 \pnum
 \complexity
 Exactly one comparison and two applications of the projection, if any.
+
+\pnum
+\remarks
+An invocation may explicitly specify
+an argument for the template parameter \tcode{T}
+of the overloads in namespace \tcode{std}.
 \end{itemdescr}
 
 \indexlibraryglobal{min}%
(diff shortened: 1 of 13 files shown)


johelegp:
What are those 6 remaining cases? I remember @zygoloid mentioning some that needed LWG attention because their order may have been intended and reordering the elements could change the interpretation of the content being described.

jensmaurer:
iostreams.tex:3703-3794: bad element ordering:  returns remarks effects
iostreams.tex:3836-3898: bad element ordering:  remarks returns ensures
iostreams.tex:3932-4032: bad element ordering:  effects remarks expects returns
locales.tex:1976-2088: bad element ordering:  expects effects remarks returns
numerics.tex:4427-4477: bad element ordering:  complexity effects returns throws
regex.tex:1600-1620: bad element ordering:  returns effects ensures throws

johelegp:
> I remember @zygoloid mentioning some that needed LWG attention because their order may have been intended and reordering the elements could change the interpretation of the content being described.

Maybe I actually meant this comment by @jensmaurer: https://github.com/cplusplus/draft/issues/2209#issuecomment-402953484. I'm still looking for it, just in case.

jensmaurer:
This particular concern has been addressed by the "mandating" papers.

timsong-cpp:
See thread starting at https://lists.isocpp.org/lib/2020/02/15447.php.

source/algorithms.tex
```diff
@@ -6322,8 +6314,6 @@
 the \oldconcept{MoveConstructible} (\tref{cpp17.moveconstructible}) and
 \oldconcept{\-Move\-Assignable} (\tref{cpp17.moveassignable}) requirements.
 
-\pnum
-\expects
```
johelegp, line 6326: Not really an order correction, but a merge. Shouldn't it become a list? This is how it looks like (single paragraph number, multiple grammatical paragraphs, treating the two different matters being merged):

source/strings.tex
```diff
@@ -2623,6 +2614,16 @@
...
+
+\pnum
+\throws
+\begin{itemize}
+\item \tcode{out_of_range} if \tcode{pos1 > size()},
+\item \tcode{length_error} if the length of the resulting string
+would exceed \tcode{max_size()} (see below), or
+\item any exceptions thrown by \tcode{allocator_traits<Allocator>::allocate}.
+\end{itemize}
+
```
johelegp, line 2626:
```suggestion
```

source/strings.tex
```diff
@@ -2623,6 +2614,16 @@
 \pnum
 \returns
 \tcode{*this}.
+
+\pnum
+\throws
+\begin{itemize}
+\item \tcode{out_of_range} if \tcode{pos1 > size()},
+\item \tcode{length_error} if the length of the resulting string
+would exceed \tcode{max_size()} (see below), or
```
johelegp, line 2623: `max_size()` is defined above, so maybe the "see below" refers to the now above _Effects:_ for when the length would exceed `max_size()`, which should be redundant given [string.require]p1.
```suggestion
would exceed \tcode{max_size()} (see above), or
```
```suggestion
would exceed \tcode{max_size()}, or
```

source/strings.tex
```diff
@@ -2663,6 +2655,15 @@
...
```
johelegp, line 2664: Same concern as https://github.com/cplusplus/draft/pull/4067/files#r449724740.

source/support.tex
```diff
@@ -2237,12 +2215,10 @@
 \tcode{operator delete}.
 
 \pnum
-\expects
 If an implementation has strict pointer safety\iref{basic.stc.dynamic.safety}
 then \tcode{ptr} is a safely-derived pointer.
 
 \pnum
-\expects
```
johelegp, line 2245: Wouldn't it be a good idea to deindent the elements a little? Although it might not be worth it considering there aren't many examples like these:

source/algorithms.tex
johelegp: Similar changes in this PR didn't dropped the `\expects` and not the `\pnum`, so perhaps that would do here.

source/utilities.tex
```diff
@@ -1957,6 +1957,15 @@
...
 
+\pnum
+\returns
+A \tcode{tuple} object constructed by initializing the ${k_i}^\text{th}$
+type element $\tcode{e}_{ik}$ in \tcode{$\tcode{e}_i$...} with
+\begin{codeblock}
+get<@$k_i$@>(std::forward<@$\tcode{T}_i$@>(@$\tcode{tp}_i$@))
+\end{codeblock}
+for each valid $k_i$ and each group $\tcode{e}_i$ in order.
+
```
johelegp, line 1968: This uses the _`e`<sub>i</sub>_ defined in the _Remarks:_ below. Should it be at the top of the description, like elsewhere?

source/algorithms.tex
jensmaurer: Fixed.

source/strings.tex
jensmaurer: Fixed.

source/support.tex
jensmaurer: I think we do both styles. @jwakely , any opinion here?

source/utilities.tex
jensmaurer: If it would only define e_i has a tool, then yes. But it's also specifying requirements on the result type CTypes.

Reverted.

source/strings.tex
jensmaurer: Fixed (second option)

source/strings.tex

jensmaurer:
@timsong-cpp , I've removed the reordering of the two mandates/constraints pairs that are the subject of the references reflector thread.

johelegp:
> > I remember @zygoloid mentioning some that needed LWG attention because their order may have been intended and reordering the elements could change the interpretation of the content being described.
> 
> Maybe I actually meant this comment by @jensmaurer: [#2209 (comment)](https://github.com/cplusplus/draft/issues/2209#issuecomment-402953484). I'm still looking for it, just in case.

Ah, that reminds me where it was: https://github.com/cplusplus/draft/pull/3788#discussion_r382932110.

source/utilities.tex
```diff
@@ -9819,6 +9815,10 @@
 the expression \tcode{delete p} is well-formed and
 \tcode{Y*} is convertible to \tcode{T*}.
 
+\pnum
+\mandates
+\tcode{Y} is a complete type.
+
```
johelegp, line 9821: Is this OK in light of https://github.com/cplusplus/draft/pull/3788#discussion_r383605890?

jensmaurer: Reverted.

jensmaurer:
After the reversions, the remaining ordering violations are:

iostreams.tex:3703-3794: bad element ordering:  returns remarks effects
iostreams.tex:3836-3898: bad element ordering:  remarks returns ensures
iostreams.tex:3932-4032: bad element ordering:  effects remarks expects returns
locales.tex:1976-2088: bad element ordering:  expects effects remarks returns
numerics.tex:4427-4477: bad element ordering:  complexity effects returns throws
regex.tex:1600-1620: bad element ordering:  returns effects ensures throws
utilities.tex:1940-1985: bad element ordering:  expects remarks returns
utilities.tex:4631-4664: bad element ordering:  mandates constraints effects ensures returns throws remarks
utilities.tex:4672-4705: bad element ordering:  mandates constraints effects ensures returns throws remarks
utilities.tex:9807-9851: bad element ordering:  mandates constraints expects effects ensures throws

PR #4068, opened 2020-07-04
[basic.lval] bit-fields are objects

languagelawyer:
no need to mention them separately

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3069,8 +3069,8 @@ in~\ref{basic.types}.
 \end{note}
 
 \pnum
-A \defn{memory location} is either an object of scalar type or a maximal
-sequence of adjacent bit-fields all having nonzero width.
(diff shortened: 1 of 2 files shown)


zygoloid:
Looks like a little more cleanup may be desirable: [intro.memory]p3: "A memory location is either an object of scalar type <ins>that is not a bit-field</ins> or a maximal sequence of adjacent bit-fields all having nonzero width." The note in p4 makes it completely obvious that this is the intent, so I think it's fine to fix this editorially.

While we're cleaning things up in this area, I think we can clarify [intro.object]/8.4: "has subobjects of nonzero size or <ins>unnamed</ins> bit-fields of nonzero length."

languagelawyer:
@zygoloid Do I understand correctly: you want to stress that unnamed bit-fields are not members and [are not|do not have corresponding] subobjects, yet they "contribute to" a memory location and affect whether a class object is of nonzero size?

PR #4069, opened 2020-07-04
[expr.prim.lambda] Move operator() return type description

languagelawyer:


source/expressions.tex
```diff
@@ -1636,12 +1624,26 @@
 The closure type for a \grammarterm{lambda-expression} has a public
 inline function call operator (for a non-generic lambda) or
 function call operator template (for a generic lambda)\iref{over.call}
-whose parameters and return type
-are described by the \grammarterm{lambda-expression}'s
-\grammarterm{parameter-declaration-clause} and \grammarterm{trailing-return-type}
-respectively, and whose
-\grammarterm{template-parameter-list} consists of
+whose parameters are described by the \grammarterm{lambda-expression}'s
```
jensmaurer, line 1627: I agree with the gist of the change, but having four "whose" in a row is not optimal.

Maybe say 

"The closure type for a \grammarterm{lambda-expression} has a public
inline function call operator (for a non-generic lambda) or
function call operator template (for a generic lambda)\iref{over.call} with the following properties:
(bulleted list)
its parameters...
its return type is...
etc.
"

languagelawyer: Done, but maybe we should write the `operator()`'s declaration syntactically? I.e.
"""
The closure type for a \grammarterm{lambda-expression} has a public
function call operator as if declared by

   `template` `<` _template-parameter-list_ `>` _requires-clause_
   `auto` `operator()` `(` _parameter-declaration-clause_ `)` ....

where:
— _parameter-declaration-clause_ is the _lambda-declarator_'s _parameter-declaration-clause_,
...
"""

source/expressions.tex
```diff
@@ -1636,18 +1624,44 @@
...
+\begin{codeblock}
+auto x1 = [](int i){ return i; };       // OK: return type is \tcode{int}
+auto x2 = []{ return { 1, 2 }; };       // error: deducing return type from \grammarterm{braced-init-list}
+int j;
+auto x3 = []()->auto&& { return j; };   // OK: return type is \tcode{int\&}
+\end{codeblock}
+\end{example}
+\end{note}
+
+\item \grammarterm{requires-clause} in the \grammarterm{declaration}
```
jensmaurer, line 1651: Please don't avoid good terms such as "trailing requires-clause". Phrases like "requires-clause in (huge thing)" are not helpful.

languagelawyer: Missed that trailing requires-clause is an official term...

source/expressions.tex
jensmaurer: We're trying not to be syntactic in this area.

source/expressions.tex
```diff
@@ -1636,18 +1624,43 @@
...
+auto x3 = []()->auto&& { return j; };   // OK: return type is \tcode{int\&}
+\end{codeblock}
+\end{example}
+\end{note}
+
+\item trailing \grammarterm{requires-clause} is the \grammarterm{lambda-declarator}'s
+\grammarterm{requires-clause}, if any,
+
+\item \grammarterm{template-parameter-list} is the \grammarterm{lambda-expression}'s
+\grammarterm{template-parameter-list}, if any,
```
jensmaurer, line 1655: this needs "and" at the end.

source/expressions.tex
```diff
@@ -1636,18 +1624,43 @@
...
+\tcode{return} statements as described in~\ref{dcl.spec.auto}.
+\begin{example}
+\begin{codeblock}
+auto x1 = [](int i){ return i; };       // OK: return type is \tcode{int}
+auto x2 = []{ return { 1, 2 }; };       // error: deducing return type from \grammarterm{braced-init-list}
+int j;
+auto x3 = []()->auto&& { return j; };   // OK: return type is \tcode{int\&}
+\end{codeblock}
+\end{example}
+\end{note}
```
jensmaurer, line 1649: Sorry for the piecemeal feedback, but we shouldn't have full-sentence (long) notes inside a normative sentence (which, in this case, spans multiple bullets) if we can help it. Please move the note to after the bulleted list.

languagelawyer: The most difficult decision: Example after or inside the Note

jensmaurer: The example is explaining the note, so goes inside the note.

jensmaurer (approved):

source/expressions.tex
languagelawyer: > We're trying not to be syntactic in this area.

Why? There are places where the standard does syntactic transformation, like for `for` loops or a coroutine _funciton-body_. Why not do it to define a closure's `operator()`?

source/expressions.tex
```diff
@@ -1636,18 +1624,43 @@
...
-respectively, and whose
-\grammarterm{template-parameter-list} consists of
-the specified \grammarterm{template-parameter-list}, if any.
-The \grammarterm{requires-clause} of the function call operator template
+whose:
+\begin{itemize}
+\item \grammarterm{parameter-declaration-clause} is the \grammarterm{lambda-declarator}'s
+\grammarterm{parameter-declaration-clause},
+
+\item return type is \tcode{auto},
```
zygoloid, line 1632: I think that this isn't the return type, so much as it is the type in the *decl-specifier-seq*. (The *declared* return type is determined by the *trailing-return-type* if any, and the *actual* return type is what we deduce. Per [dcl.spec.auto.general]p3, the `auto` is just something that can appear in the *decl-specifier-seq*; it's not really a return type at all.) So:

```suggestion
\item \grammarterm{decl-specifier-seq} is the \grammarterm{lambda-declarator}'s \grammarterm{decl-specifier-seq} followed by \keyword{auto}, except that \keyword{constexpr} is also included if [...]
```

Also... should we say something about the *attribute-specifier-seq*?

jensmaurer:
@languagelawyer , this needs a bit more work.

tkoeppe:
Can you please rebase, and address the open comments?

tkoeppe:
Please reopen if you would like to continue with this.

PR #4070, opened 2020-07-04
[stmt.pre, re.grammar, gram] Use a separate bnf environment for each nonterminal.

Eelis:
This is already done everywhere else in the document as far as I can tell.

No visible difference in the PDF.

(Incidentally, this change makes life easier for cxxdraft-htmlgen.)

### DIFF
--- a/source/grammar.tex
+++ b/source/grammar.tex
@@ -33,7 +33,9 @@ typedef-name:\br
 namespace-name:\br
 	identifier\br
 	namespace-alias
+\end{ncbnf}
 
+\begin{ncbnf}
 namespace-alias:\br
 	identifier
 \end{ncbnf}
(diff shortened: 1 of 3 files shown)


jensmaurer:
Ideally, we would script-check this in the future, to avoid introducing regressions.
Also, if most every "bnf" environment starts with a `\nontermdef`, we might want to consider having a `bnfnontermdef` environment or so. Later, maybe.

jensmaurer:
@ŧkoeppe, could you have a look, please? This seems an easy one.

tkoeppe:
Looks good, thanks! I agree that if we never want to have multiple definitions per environment, we could have a more structured environment -- that would probably also obviate the need for a presubmi check (assuming we could get rid of the permissive existing environment that allows multiple definitions).

PR #4071, opened 2020-07-04
[alg.is.permutation] Rephrase to simplify

CaseyCarter:
Defines `last2` and `pred` for the overloads in `std` with no parameters by those names so the remainder of the specification need not exhaustively describe each case.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -3968,7 +3968,7 @@ template<class ForwardIterator1, class ForwardIterator2,
 \pnum
 Let \tcode{last2} be \tcode{first2 + (last1 - first1)}
 for the overloads with no parameter named \tcode{last2},
-and let \tcode{pred} be \tcode{equal<>\{\}}
+and let \tcode{pred} be \tcode{equal_to\{\}}
 for the overloads with no parameter \tcode{pred}.
 
 \pnum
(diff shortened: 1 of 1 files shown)


source/algorithms.tex
```diff
@@ -3965,6 +3965,11 @@
 \end{itemdecl}
 
 \begin{itemdescr}
+\pnum
+Let \tcode{last2} be \tcode{first2 + (last1 - first1)}
+for the overloads with no parameter named \tcode{last2},
+and let \tcode{pred} be \tcode{equal<>\{\}}
```
johelegp, line 3971:
```suggestion
and let \tcode{pred} be \tcode{equal\{\}}
```

timsong-cpp: Surely this should be `equal_to`?

CaseyCarter:
```suggestion
and let \tcode{pred} be \tcode{equal_to\{\}}
```

CaseyCarter:
> I have a weak preference for `equal_to<>{}` rather than `equal_to{}` but that's probably my personal dislike of CTAD in general.

I have no personal stake in CTAD, but the usage is endemic since LWG instructed me to replace the occurrences of `equal_to<>{}` / `less<>{}` that P0896 added to the algorithms with `equal_to{}` and `less{}`. I believe this algorithm should be consistent with others and a uniform change to non-CTAD should be made separately by someone who cares ;)

PR #4074, opened 2020-07-07
[re.grammar] Don't use monospace for ordinary text.

Eelis:
image

### DIFF
--- a/source/regex.tex
+++ b/source/regex.tex
@@ -3703,17 +3703,17 @@ names shall be recognized:
 In addition the following expressions shall be equivalent:
 
 \begin{codeblock}
-\d and [[:digit:]]
+\d @\textnormal{and}@ [[:digit:]]
 
-\D and [^[:digit:]]
+\D @\textnormal{and}@ [^[:digit:]]
 
-\s and [[:space:]]
+\s @\textnormal{and}@ [[:space:]]
 
-\S and [^[:space:]]
+\S @\textnormal{and}@ [^[:space:]]
 
-\w and [_[:alnum:]]
+\w @\textnormal{and}@ [_[:alnum:]]
 
-\W and [^_[:alnum:]]
+\W @\textnormal{and}@ [^_[:alnum:]]
 \end{codeblock}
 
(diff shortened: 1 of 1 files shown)


source/regex.tex
```diff
@@ -3692,19 +3692,14 @@
...
-
-\S and [^[:space:]]
-
-\w and [_[:alnum:]]
-
-\W and [^_[:alnum:]]
-\end{codeblock}
+\begin{tabular}{@{\quad}l@{and}l}
+  \tcode{\textbackslash d } & \tcode{ [[:digit:]]} \\[3mm]
+  \tcode{\textbackslash D } & \tcode{ [}\tcode{\^}\tcode{[:digit:]]} \\[3mm]
```
jensmaurer, line 3697: Do we, in fact, have to make separate `\tcode`s here? The source looks ugly.

source/regex.tex
```diff
@@ -3692,19 +3692,14 @@
...
-\s and [[:space:]]
-
-\S and [^[:space:]]
-
-\w and [_[:alnum:]]
-
-\W and [^_[:alnum:]]
-\end{codeblock}
+\begin{tabular}{@{\quad}l@{and}l}
+  \tcode{\textbackslash d } & \tcode{ [[:digit:]]} \\[3mm]
```
jensmaurer, line 3696: The "3mm" extra spacing seems oddly arbitrary.

Would a smaller change be to escape from the "codeblock" environment and just write @and@ to get a prose-font "and"?

tkoeppe: I'd like to see that option explored, too. Generally, I tried to get the vertical whitespace be the same everwhere as much as possible, so I'd be reluctant to introduce an explicit special size here.

Eelis: Ah, after trying a bunch of different ways to escape the codeblock environment, I gave up and used a tabular instead, but @ seems to work like a charm, thanks! :) Updating patch.

Eelis:
Amended to use @-escaping instead of resorting to a tabular environment.

PR #4080, opened 2020-07-13
[expr.reinterpret.cast] Fix a note

languagelawyer:
```c++
struct S { int i; } s;

auto fp = reinterpret_cast<float*>(&s); // the value of `fp` is  "pointer to s"

reinterpret_cast<float*>(reinterpret_cast<int*>(fp));
// converting to `int*` and back to `float*` does NOT yield the original pointer value,
// because the value became "pointer to s.i" when casting to `int*`.
```
@hubert-reinterpretcast

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -3929,11 +3929,13 @@ restriction that a \tcode{reinterpret_cast} cannot cast away constness.}
 When a prvalue \tcode{v} of object pointer type is converted to
 the object pointer type ``pointer to \cv{}~\tcode{T}'', the result is \tcode{static_cast<\cv{} T*>(static_cast<\cv{}~void*>(v))}.
 \begin{note}
-Converting a prvalue of type ``pointer to \tcode{T1}'' to
-the type ``pointer to \tcode{T2}'' (where \tcode{T1} and \tcode{T2} are
-object types and where the alignment requirements of \tcode{T2} are no
-stricter than those of \tcode{T1}) and back to its original type yields
-the original pointer value.
+Converting a pointer of type ``pointer to \tcode{T1}''
+that points to an object of type \tcode{T1}
+to the type ``pointer to \tcode{T2}''
+(where \tcode{T2} is an object type
+and the alignment requirements of \tcode{T2}
+are no stricter than those of \tcode{T1})
+and back to its original type yields the original pointer value.
 \end{note}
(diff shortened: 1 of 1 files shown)


hubert-reinterpretcast:
Would it make sense to try to salvage the note by indicating that the resulting pointer could also be a pointer to an object that is pointer-interconvertible with the object (if any) to which the original pointer points?

jensmaurer:
It sounds reasonable to highlight the round-trip guarantee, but that needs a bit of wording surgery. Maybe just avoiding phrasing in terms of types would help. "If a pointer to a type T1 is converted to type "pointer to T2" and back to "pointer to T1" and (the stars are aligned correctly), you get the original pointer value."

hubert-reinterpretcast:
I think just making the original pointer value something more specific works:
Converting a prvalue of type "pointer to `T1`" that points to an object of type `T1` [ ... ]

languagelawyer:
I didn't like "a prvalue that points" and rewrote as "a pointer that points", also, since `T1` became known to be an object type, I had to change the "where" part. I also decided to remove parens and reflow the sentence.

source/expressions.tex
```diff
@@ -3929,11 +3929,13 @@
...
 the object pointer type ``pointer to \cv{}~\tcode{T}'', the result is \tcode{static_cast<\cv{} T*>(static_cast<\cv{}~void*>(v))}.
 \begin{note}
-Converting a prvalue of type ``pointer to \tcode{T1}'' to
-the type ``pointer to \tcode{T2}'' (where \tcode{T1} and \tcode{T2} are
-object types and where the alignment requirements of \tcode{T2} are no
-stricter than those of \tcode{T1}) and back to its original type yields
-the original pointer value.
+Converting a pointer of type ``pointer to \tcode{T1}''
+that points to an object of type \tcode{T1}
+to the type ``pointer to \tcode{T2}'',
```
hubert-reinterpretcast, line 3934: I still prefer the parentheses in place of the comma. I find ending up with a comma before "and" to be unfortunate when the instance of "and" is not being used to join independent clauses into a sentence.

hubert-reinterpretcast:
LGTM, but I have no commit access.

jensmaurer:
@hubert-reinterpretcast , thanks for the review.

PR #4087, opened 2020-07-21
[dcl.init] Remove redundant paragraph about initialization with () initializer

### LINKED ISSUE #4047: [dcl.init]/11 and [dcl.init]/(17.4) say the same thing?
[dcl.init]/11: An object whose initializer is an empty set of parentheses, i.e., `()`, shall be value-initialized.

[dcl.init]/17:
> The semantics of initializers are as follows. The destination type is the type of the object or reference being initialized and the source type is the type of the initializer expression. If the initializer is not a single (possibly parenthesized) expression, the source type is not defined.
...
(17.4) — If the initializer is `()`, the object is value-initialized.

languagelawyer:
I've left the Note, because I don't know where to move it.

Fixes #4047

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -4292,11 +4292,6 @@ In some cases, additional initialization is done later.
 \end{note}
 
 \pnum
-An object whose initializer is an empty set of parentheses, i.e.,
-\tcode{()},
-shall be
-value-initialized.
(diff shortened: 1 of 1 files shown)


zygoloid:
This was addressed by #3891.

PR #4089, opened 2020-07-22
Add additional example to CWG2303 example to clarify meaning.

erichkeane:
As discussed on the reflector, CWG2303 causes shadowing of a base class
when attempting deduction in the case where the type is a base of
another successful base class.  This patch adds an example that
clarifies that intent.


jensmaurer (approved):
The example seems to be in line with the wording, and does highlight a slightly surprising outcome.

PR #4092, opened 2020-07-25
[fs.class.path, fs.path.nonmember] Fix inline whitespace

johelegp:


jwakely:
I agree that using whitespace to align names in class synopses is not our usual convention. But if you're going to remove it here, why not also do it above where `void  clear() noexcept;` and `void  swap(path& rhs) noexcept;` are aligned with the declarations between them? And also for `string`, `wstring`, `u8string` etc. and `generic_string`, `generic_wstring`, generic_u8string`?

I think either we should remove all that whitespace used for alignment, or none of it.

source/iostreams.tex
```diff
@@ -11796,11 +11796,11 @@
     friend bool operator==(const path& lhs, const path& rhs) noexcept;
     friend strong_ordering operator<=>(const path& lhs, const path& rhs) noexcept;
 
-    friend path operator/ (const path& lhs, const path& rhs);
+    friend path operator/(const path& lhs, const path& rhs);
 
     // \ref{fs.path.native.obs}, native format observers
     const string_type& native() const noexcept;
-    const value_type*  c_str() const noexcept;
+    const value_type* c_str() const noexcept;
```
johelegp, line 11803: I didn't notice that this was actually aligned. The parentheses are what stood out to me, not the name. I will keep only the other change, whose alignment became outdated with the addition of `operator<=>`.

PR #4093, opened 2020-07-26
[conv.qual]  Fix example for cv-decomposition

languagelawyer:
Nothing in [conv.qual] says that `U` can't have top-level qualifiers.

Discovered by https://stackoverflow.com/users/775806/n-pronouns-m


languagelawyer:
Well, ok. [basic.type.qualifier]/6 says: "For a type cv T, the top-level cv-qualifiers of that type are those denoted by cv", so the form of cv-decomposition "... _cv_<sub>n</sub> `U`" suggests that top-level qualifiers of `U` must be in _cv_<sub>n</sub>.

languagelawyer:
Reopening just to know if something needs a fix

PR #4102, opened 2020-07-31
[std] Add missing \pnums.

Eelis:


source/utilities.tex
```diff
@@ -21014,6 +21014,7 @@
 \tcode{i < size_ ?\ data_[i] :\ basic_format_arg<Context>()}.
 \end{itemdescr}
 
+\pnum
```
jwakely, line 21017: This one is a non-normative note making normative encouragement, which needs to be fixed. I'll file an LWG issue.

PR #4105, opened 2020-08-01
[lib] Change 'expression inside noexcept' to 'exception specification'

johelegp:
https://wg21.link/[stringbuf.assign]#5 doesn't use a code block. Is that OK?

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -4182,7 +4182,7 @@ Any exception thrown by move-constructing any $\tcode{T}_i$ for all $i$.
 \pnum
 \remarks
 The exception specification is equivalent to the logical \logop{AND} of
-\tcode{is_nothrow_move_constructible_v<$\tcode{T}_i$>} for all $i$.
+\tcode{is_nothrow_move_con\-structible_v<$\tcode{T}_i$>} for all $i$.
 If \tcode{is_trivially_move_constructible_v<$\tcode{T}_i$>}
 is \tcode{true} for all $i$, this constructor is trivial.
 \end{itemdescr}
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -227,7 +227,7 @@ Exchanges values stored in two locations.
 \remarks
 This function
 is a designated customization point\iref{namespace.std}.
-The expression inside \tcode{noexcept} is equivalent to:
+The exception specification is equivalent to:
 
 \begin{codeblock}
 is_nothrow_move_constructible_v<T> && is_nothrow_move_assignable_v<T>
@@ -931,7 +931,7 @@ and to \tcode{second} with\\ \tcode{std::forward<second_type>(p.second)}.
 
 \pnum
 \remarks
-The expression inside \tcode{noexcept} is equivalent to:
+The exception specification is equivalent to:
 \begin{codeblock}
 is_nothrow_move_assignable_v<T1> && is_nothrow_move_assignable_v<T2>
 \end{codeblock}
@@ -979,7 +979,7 @@ Swaps
 
 \pnum
 \remarks
-The expression inside \tcode{noexcept} is equivalent to:
+The exception specification is equivalent to:
 \begin{codeblock}
 is_nothrow_swappable_v<first_type> && is_nothrow_swappable_v<second_type>
 \end{codeblock}
@@ -1719,7 +1719,7 @@ For all $i$, assigns \tcode{std::forward<$\tcode{T}_i$>(get<$i$>(u))} to
 
 \pnum
 \remarks
-The expression inside \tcode{noexcept} is equivalent to the logical \logop{AND} of the
+The exception specification is equivalent to the logical \logop{AND} of the
 following expressions:
 
 \begin{codeblock}
@@ -1849,7 +1849,7 @@ Nothing unless one of the element-wise \tcode{swap} calls throws an exception.
 
 \pnum
 \remarks
-The expression inside \tcode{noexcept} is equivalent to the logical
+The exception specification is equivalent to the logical
 \logop{AND} of the following expressions:
 
 \begin{codeblock}
@@ -2342,7 +2342,7 @@ As if by \tcode{x.swap(y)}.
 
 \pnum
 \remarks
-The expression inside \tcode{noexcept} is equivalent to:
+The exception specification is equivalent to:
 
 \begin{codeblock}
 noexcept(x.swap(y))
@@ -2601,7 +2601,7 @@ Any exception thrown by the selected constructor of \tcode{T}.
 
 \pnum
 \remarks
-The expression inside \tcode{noexcept} is equivalent to
+The exception specification is equivalent to
 \tcode{is_nothrow_move_constructible_v<T>}.
 If \tcode{is_trivially_move_constructible_v<T>} is \tcode{true},
 this constructor is trivial.
@@ -2908,7 +2908,7 @@ no effect \\
 
 \pnum
 \remarks
-The expression inside \tcode{noexcept} is equivalent to:
+The exception specification is equivalent to:
 \begin{codeblock}
 is_nothrow_move_assignable_v<T> && is_nothrow_move_constructible_v<T>
 \end{codeblock}
@@ -3189,7 +3189,7 @@ Any exceptions thrown by the operations in the relevant part of \tref{optional.s
 
 \pnum
 \remarks
-The expression inside \tcode{noexcept} is equivalent to:
+The exception specification is equivalent to:
 \begin{codeblock}
 is_nothrow_move_constructible_v<T> && is_nothrow_swappable_v<T>
 \end{codeblock}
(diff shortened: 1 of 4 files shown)


source/utilities.tex
```diff
@@ -3186,6 +3188,8 @@
 \begin{codeblock}
 is_nothrow_move_constructible_v<T> && is_nothrow_swappable_v<T>
 \end{codeblock}
+
+\pnum
```
johelegp, line 3192: Like above, in the assignment operator.

source/utilities.tex
```diff
@@ -4170,7 +4174,7 @@
 
 \pnum
 \remarks
-The expression inside \tcode{noexcept} is equivalent to the logical \logop{AND} of
+The expression inside \tcode{noexcept} is equivalent to the logical \textsc{and} of
```
johelegp, line 4177: Two other places above in this file use this. Changed these two accordingly.

source/utilities.tex
```diff
@@ -4489,7 +4493,7 @@
 \tcode{is_trivially_move_assignable_v<$\tcode{T}_i$> \&\&}
 \tcode{is_trivially_destructible_v<$\tcode{T}_i$>}
 is \tcode{true} for all $i$, this assignment operator is trivial.
-The expression inside \tcode{noexcept} is equivalent to:
+The expression inside \tcode{noexcept} is equivalent to
```
johelegp, line 4496: The other few inline cases don't use colon.

jwakely:
Personally, I don't like this language at all, neither "in" nor "inside".

[except.spec] says "that constant expression **is the exception specification** of the function type in which the noexcept-specifier appears" (emphasis mine).

So we could just say "the exception specification is equivalent to..."

Or we could define a new element which defines the _see below_ expression, and do away with "the expression inside `noexcept` is equivalent to" entirely.

Others might not agree though, so don't rewrite everything based on this comment alone.

source/utilities.tex
jwakely: I would expect consistency with https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#writing-effects-in-a-function-description

johelegp: You mean that it should be a code block because it's not a plain expression of type `void`?

jwakely: No, it doesn't say you must use a code block for non-void expressions, it says you must use `return expression` not just `expression`, but that part of the guideline is referring to return values anyway.

I mean this part:

>  When code is used to describe an effects element, a colon (:) should follow the introductory phrase if the code is a statement or a code block, whereas expressions should end with a period (.).

i.e. there should be a colon when the expression is in a code block, but not when it's inline. So I agree with removing the colon here.

jensmaurer:
There are several different changes here, and I'd like to see them separated a bit more.
- There is the question of phrasing for "the expression inside", which applies (in principle) to both noexcept and explicit. This needs a wider audience to decide.
- There are typo-level fixes (logop, colons) that could be applied right away.

@johelegp , please post a separate pull request that only contains separate commits for the "logop" fixes and the colon fixes.

zygoloid:
We should pick our preferred wording for this construct and apply it uniformly everywhere.

jensmaurer:
Editorial meeting: Say "the /constant-expression/ of the /noexcept-specifier/ is equivalent to" and see how it feels.

johelegp:
Will these two be enough to gauge how it feels?

johelegp:
For `requires `_`see below`_, we have:

https://timsong-cpp.github.io/cppwp/reverse.iterators
```C++
    constexpr pointer   operator->() const requires see below;
```
```C++
constexpr pointer operator->() const
  requires (is_pointer_v<Iterator> ||
            requires (const Iterator i) { i.operator->(); });
```
---
https://timsong-cpp.github.io/cppwp/iterators.common
```C++
    decltype(auto) operator->() const
      requires see below;
```
```C++
decltype(auto) operator->() const
  requires see below;
```
The expression in the requires-clause is equivalent to:
```C++
indirectly_­readable<const I> &&
(requires(const I& i) { i.operator->(); } ||
 is_reference_v<iter_reference_t<I>> ||
 constructible_­from<iter_value_t<I>, iter_reference_t<I>>)
```
> No _Remarks:_.
---
https://timsong-cpp.github.io/cppwp/#ranges
```C++
  template<movable Val, class CharT, class Traits = char_traits<CharT>>
    requires see below
  class basic_istream_view;

  template<movable Val, class CharT, class Traits>
    requires default_initializable<Val> &&
             stream-extractable<Val, CharT, Traits>
  class basic_istream_view : public view_interface<basic_istream_view<Val, CharT, Traits>> {

  template<input_range V, size_t N>
    requires see below;
  class elements_view;

  template<input_range V, size_t N>
    requires view<V> && has-tuple-element<range_value_t<V>, N> &&
             has-tuple-element<remove_reference_t<range_reference_t<V>>, N>
  class elements_view : public view_interface<elements_view<V, N>> {
```
> These use exposition-only concepts defined before the class template definition.
---
https://timsong-cpp.github.io/cppwp/range.iota
```C++
    constexpr auto size() const requires see below;
```
```C++
constexpr auto size() const requires see below;
```
Remarks: The expression in the requires-clause is equivalent to
```C++
(same_as<W, Bound> && advanceable<W>) || (integral<W> && integral<Bound>) ||
  sized_sentinel_for<Bound, W>
```
---
https://timsong-cpp.github.io/cppwp/range.ref.view
```C++
    template<not-same-as<ref_view> T>
      requires see below
    constexpr ref_view(T&& t);
```
```C++
template<not-same-as<ref_view> T>
  requires see below
constexpr ref_view(T&& t);
```
Remarks: Let FUN denote the exposition-only functions
```C++
void FUN(R&);
void FUN(R&&) = delete;
```
The expression in the requires-clause is equivalent to
```C++
convertible_to<T, R&> && requires { FUN(declval<T>()); }
```
---
https://timsong-cpp.github.io/cppwp/#time
```C++
    template<class Rep1, class Period1, class Rep2, class Period2>
      requires see below
      constexpr auto operator<=>(const duration<Rep1, Period1>& lhs,
                                 const duration<Rep2, Period2>& rhs);
```
```C++
template<class Rep1, class Period1, class Rep2, class Period2>
    requires three_­way_­comparable<typename CT::rep>
  constexpr auto operator<=>(const duration<Rep1, Period1>& lhs,
                             const duration<Rep2, Period2>& rhs);
```
> `CT` is defined in the subclause of the `itemdecl`.
---
Consider how the `iota_view` case would look like with the resolution of the editorial meeting, if applied for consistency:
```C++
constexpr auto size() const requires see below;
```
```diff
-Remarks: The expression in the requires-clause is equivalent to
+Remarks: The _constraint-logical-or-expression_ of the _requires-clause_
is equivalent to:
```
```C++
(same_as<W, Bound> && advanceable<W>) || (integral<W> && integral<Bound>) ||
  sized_sentinel_for<Bound, W>
```

jensmaurer:
Editorial meeting: We've reconsidered. Please try "The exception specification is equivalent to:"

johelegp:
Still gauging how it feels, or should I fix it generally now?

jensmaurer:
Still gauging how it feels; we don't want to generate a lot of work while we're still unsure.

jwakely (approved):
This works for me.

tkoeppe:
I'd also find this a nice improvement. Let's leave this for another week or two for feedback (@CaseyCarter, @brycelelbach?), but let's tentatively consider this ready.

brycelelbach:
Personally I'm happy if @jwakely is happy. I think "The exception specification is equivalent to" is a more understandable/approachable for new C++ proposal authors. I do wonder, though, do we have enough of these to merit their own paragraph type? E.g. "*Exception Specification:* Equivalent to:"? This would nicely mirror the pattern of "*Effects:* Equivalent to:".

tkoeppe:
Thanks! We had talked about a new specification element, too, and were generally sympathetic to that idea. I suggested as a first step to go with this proposed change for now, and if it turns out to come up often enough to warrant a new element, we can consider that as a second step.

johelegp:
There's 22 of these.

jwakely:
>  E.g. "_Exception Specification:_ Equivalent to:"? This would nicely mirror the pattern of "_Effects:_ Equivalent to:".

I don't think mirroring "_Effects:_ Equivalent to ..." is necessarily desirable. For the _Effects:_ paragraph that has very specific meaning with magic powers (it inherits all the constraints and other properties from the equivalent construct). We don't need that for the exception specification, we're just describing a boolean constant.

jensmaurer (approved):
The actual proposed changes look fine to me, and in line with what was decided in the editorial meeting.

tkoeppe:
OK, let's go with the current version for now, as agreed. I think @jwakely makes a good point, so this might be good enough and we just leave it like this, esp. if only 22 paragraphs would be affected by this anyway.

PR #4108, opened 2020-08-02
[lib] Fix typos in "The expression in...is equivalent to"

johelegp:


jensmaurer:
This looks harmless and the right thing to do. @jwakely , could you please have a quick look?

jwakely (approved):
I wish we had a better way to say "inside `explicit`" or "inside `noexcept`", but given our current wording style, this is a clear improvement. Thanks.

PR #4111, opened 2020-08-09
[range.prim.empty] Fix expression specification

onihusube:
Is `ranges​::​empty(E)` in the text here not `E`?

>Given a subexpression ranges​::​empty(E) with type T, let t be an lvalue that denotes the reified object for E.

- [24.3.11 ranges​::​empty [range.prim.empty] - eel.is](http://eel.is/c++draft/range.access#range.prim.empty-2.sentence-1)


CaseyCarter (approved):
Good catch! It looks like we mis-merged P2091R0 here.

PR #4113, opened 2020-08-17
[expr.call] [dcl.fct] Move return type requiremens

languagelawyer:
An array element type is constrained by [dcl.array]/4:
> U is called the array element type; this type shall not be a placeholder type ([dcl.spec.auto]), a reference type, a function type, an array of unknown bound, or cv void.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -3490,10 +3490,7 @@ parameter-type-lists, and \grammarterm{requires-clause}{s}\iref{temp.over.link}.
 \pnum
 \indextext{function return type|see{return type}}%
 \indextext{return type}%
-Functions shall not have a return type of type array or function,
-although they may have a return type of type pointer or reference to such things.
-There shall be no arrays of functions, although there can be arrays of pointers
-to functions.
+The return type shall be a non-array object type, a reference type, or \cv{}~\tcode{void}.
 
 \pnum
 A volatile-qualified return type is deprecated;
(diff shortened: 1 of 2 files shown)


languagelawyer:
It is funny that this removes prohibition on array return type and it was approved as an editorial change

jensmaurer (approved):

source/declarations.tex
```diff
@@ -3490,10 +3490,7 @@
 \pnum
 \indextext{function return type|see{return type}}%
 \indextext{return type}%
-Functions shall not have a return type of type array or function,
-although they may have a return type of type pointer or reference to such things.
-There shall be no arrays of functions, although there can be arrays of pointers
-to functions.
+The return type shall be a non-array object type, a reference type or \cv{}~\tcode{void}.
```
zygoloid, line 3493:
```suggestion
The return type shall be a non-array object type, a reference type, or \cv{}~\tcode{void}.
```
We generally use the Oxford comma.

tkoeppe:
@jensmaurer: do you think this is ready to go? I would perhaps like a more detailed commit message explaining why the big deletion is warranted.

jensmaurer:
@tkoeppe: I think this is ready to go. The "array of function" prohibition is in [dcl.array] p4. The rest is just fluff / wordiness.

I'm merging it and fixing the commit description.

tkoeppe:
Thanks, @jensmaurer!

PR #4129, opened 2020-08-27
[format.syn] Avoid forward references

jensmaurer:
by moving the declarations of w/format_context and
w/format_args to the front.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -19052,6 +19052,19 @@ Nothing.
 \indexlibrarymember{size}{format_to_n_result}%
 \begin{codeblock}
 namespace std {
+  // \ref{format.context}, class template \tcode{basic_format_context}
+  template<class Out, class charT> class basic_format_context;
+  using format_context = basic_format_context<@\unspec@, char>;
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -19052,6 +19052,16 @@
...
+  // \ref{format.context}, class template \tcode{basic_format_context}
+  template<class Out, class charT> class basic_format_context;
+  using format_context = basic_format_context<@\unspec@, char>;
+  using wformat_context = basic_format_context<@\unspec@, wchar_t>;
+
+  // \ref{format.args}, class template \tcode{basic_format_args}
+  template<class Context> class basic_format_args;
+  using format_args = basic_format_args<format_context>;
+  using wformat_args = basic_format_args<wformat_context>;
+
```
CaseyCarter, line 19064: `format_args_t` is used in the declaration of the `vformat_to` overloads before being declared. I think that's the only remaining occurrence.

jensmaurer: Fixed.

jwakely (approved):

PR #4145, opened 2020-09-06
[diff.mods.to.declarations] Also mention 'byte' and 'to_integer.

### LINKED ISSUE #4144: Inconsistent description of C++ extensions to header <cstddef>
In addition to importing the contents of <stddef.h> into namespace std, the header <cstddef> also adds the typedef nullptr_t, and type byte, and operations on type byte.  This additions are partially described in two different locations:
[cstddef.syn]p1 notes the addition of byte and its operations, but makes no mention of nullptr_t.
[diff.mods.to.declarations]p3 (in the C compatibility annex) notes the addition of nullptr_t, but says nothing about byte and its operations.

For consistency, we should either consolidate this descriptive material to either one place, or duplicate it completely in both.  Opening and issue, rather than filing an initial PR, as not sure which direction would be preferred.

tkoeppe:
These names, as well as operators for std::byte, are part of <cstddef>.

Fixes #4144.

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -2787,7 +2787,9 @@ Subclause \ref{cwchar.syn} describes the changes.
 
 \pnum
 Header \libheaderref{cstddef}
-declares the name \tcode{nullptr_t} in addition to the names declared in
+declares the names \tcode{nullptr_t}, \tcode{byte}, and \tcode{to_integer},
+and the operators and operator templates in \iref{support.types.byteops},
+in addition to the names declared in
 \libheaderrefx{stddef.h}{depr.c.headers} in the C standard library.
 
 \rSec2[diff.mods.to.behavior]{Modifications to behavior}


jwakely (changes requested):
NACK. I think this is a normative change, because currently `<stddef.h>` is required to declare `::nullptr_t`. That's how libstdc++ and libc++ interpret it.

C.6.3 [diffs.mods.to.declarations] is informative (not normative), and my interpretation is that it's saying "The C++ version of this header adds `nullptr_t`, which isn't in the C version of this header". It's not saying that `<cstddef>` declares it and C++'s `<stddef.h>` doesn't.

tkoeppe:
You're right. Should we create an issue for this then? (I still think this is a desirable change, but it is indeed normative.)

I will replace the wording changes here with those for [diffs.mods.to.declarations].

tkoeppe:
@jwakely: PTAL.

jwakely (approved):

PR #4154, opened 2020-09-07
[macros] Fix PDF links to clauses and annexes

jensmaurer:
Such links were pointing to immediately after the clause
or annex title, not to immediately before them. The issue
was introduced with commit beb88157cc49f76677fc467db8ec4a523dbc41d0.

Fixes NB JP 014 and JP 015 (C++20 DIS)

Fixes cplusplus/nbballot#390
Fixes cplusplus/nbballot#391

### DIFF
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -70,8 +70,7 @@
 % Set the xref label for a clause to be "Clause n", not just "n".
 \makeatletter
 \newcommand{\customlabel}[2]{%
-\protected@write \@auxout {}{\string \newlabel {#1}{{#2}{\thepage}{#2}{#1}{}} }%
-\hypertarget{#1}{}%
+\@bsphack \begingroup \protected@edef \@currentlabel {\protect \M@TitleReference{#2}{\M@currentTitle}}\MNR@label{#1}\endgroup \@esphack%
 }
 \makeatother
 \newcommand{\clauselabel}[1]{\customlabel{#1}{Clause \thechapter}}


tkoeppe:
Looks good! Would you like me to test it? Otherwise if it works for you, feel free to commit.

johelegp:
Please do, it doesn't work for me.

jensmaurer:
Well, there is a bit of implementation-dependence here, because the "chapter.x" thing is modeled after what is produced by the regular \label. If your particular memoir or hyperref package writes something else there, this won't work.

jensmaurer:
Ok, I've now analyzed this some more and I've copied a little bit of the original (memoir) \label definition. This means fewer assumptions about href naming are made.
@johelegp , please try again: This should work for you.
Test case: Click on any "Clause X" hyperlink in the PDF and you should be taken to just before the clause X heading.

tkoeppe:
I cherrypicked this onto the main branch, since the c++20 branch hasn't got the version update fix for one of the packages, and there it worked as intended for me. But it's possible that there are dependencies on unstable implementation details.

tkoeppe:
@jensmaurer, @zygoloid: Could we perhaps cherrypick all the non-visible changes onto the c++20 branch (e.g. the version requirements), so that the "build environment" is the same for both?

tkoeppe:
The new version of this change also works for me.

jensmaurer:
@tkoeppe, consider approving this pull request if it works for you.

tkoeppe (approved):

johelegp:
Yes, it works now.

PR #4157, opened 2020-09-08
[std] Remove superfluous final \rowsep or \hline in tables.

Eelis:
- Only a dozen or so tables are affected.
- Effect:
image
- The miniscule amount that the content below the affected tables is raised by is too small to cause reflow anywhere.


tkoeppe:
Nice, thanks!

tkoeppe (approved):

PR #4159, opened 2020-09-08
[conv.qual] Fix punctuation in bulleted list.

### LINKED ISSUE #4119: [conv.qual] Bad list punctuation
image

jensmaurer:
Fixes #4119

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -749,14 +749,14 @@ similar to \tcode{T1} whose cv-decomposition is such that:
 \begin{itemize}
 \item
 for every $i > 0$, $\cv{}^3_i$ is the union of
-$\cv{}^1_i$ and $\cv{}^2_i$;
+$\cv{}^1_i$ and $\cv{}^2_i$,
 \item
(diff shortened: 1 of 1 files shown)


source/expressions.tex
```diff
@@ -754,7 +754,7 @@
 \item
 if the resulting $\cv{}^3_i$ is different from $\cv{}^1_i$ or $\cv{}^2_i$,
 or the resulting $P^3_i$ is different from $P^1_i$ or $P^2_i$,
-then \tcode{const} is added to every $\cv{}^3_k$ for $0 < k < i$.
+then \tcode{const} is added to every $\cv{}^3_k$ for $0 < k < i$;
```
zygoloid, line 757: I think:

* This is connecting the bullets to the "where", so we don't need to use the same character as for the other bullets here. I'd expect a comma rather than a semicolon before a "where".
* We're missing an "and" at the end of the second bullet.
* It would be fine to use commas rather than semicolons for the first two bullets. The bullets are sufficient to indicate the structure here.

jensmaurer:
Fixed, rebased, and force-pushed.

PR #4162, opened 2020-09-08
[over.literal] Mark use of a reserved identifier with 'error'.

### LINKED ISSUE #4048: Example in [over.literal]/8 misses the word error in one of its declarations
The comment after the declaration below in [[over.literal]/8](http://eel.is/c++draft/over.literal#8) seems to imply an error, but it doesn't explicitly says so. Shouldn't the comment start with the word **error**, as some of the other declarations do in the example?

`double operator"" _Bq(long double);                 // uses the reserved identifier _­Bq ([lex.name]`

Just a reminder, the error comes from [[lex.name]/(3.1)](http://eel.is/c++draft/lex.name#3.1).

jensmaurer:
Fixes #4048

### DIFF
--- a/source/overloading.tex
+++ b/source/overloading.tex
@@ -4271,9 +4271,11 @@ void operator "" _km(long double);                  // OK
 string operator "" _i18n(const char*, std::size_t); // OK
 template <char...> double operator "" _\u03C0();    // OK: UCN for lowercase pi
 float operator ""_e(const char*);                   // OK
-float operator ""E(const char*);                    // error: reserved literal suffix~(\ref{usrlit.suffix}, \ref{lex.ext})
-double operator""_Bq(long double);                  // OK: does not use the reserved identifier \tcode{_Bq}\iref{lex.name}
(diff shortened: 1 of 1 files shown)


source/overloading.tex
```diff
@@ -4259,7 +4259,7 @@
 float operator ""_e(const char*);                   // OK
 float operator ""E(const char*);                    // error: reserved literal suffix~(\ref{usrlit.suffix}, \ref{lex.ext})
 double operator""_Bq(long double);                  // OK: does not use the reserved identifier \tcode{_Bq}\iref{lex.name}
-double operator"" _Bq(long double);                 // uses the reserved identifier \tcode{_Bq}\iref{lex.name}
+double operator"" _Bq(long double);                 // error: uses the reserved identifier \tcode{_Bq}\iref{lex.name}
```
zygoloid, line 4262: Do we say "error:" for "no diagnostic required" cases?

jensmaurer: At least, we do two lines earlier for `""E`.

In other cases, we say "ill-formed, no diagnostic required" in the example comment.

jensmaurer:
Fixed, rebased, and force-pushed.

Issue #3525, opened 2019-12-06
[std] There shall be no .1 subclause without a .2 sibling


johelegp:
> A subclause shall not be created unless there is at least one further subclause at the same level. For example, text in Clause 10 shall not be designated subclause “10.1” unless there is also a subclause “10.2”.
> - https://www.iso.org/sites/directives/current/part2/index.xhtml#_idTextAnchor302

Affected subclauses, partially related to #574. Note that these are the parent subclauses of .1 subclauses without a .2 sibling:
- [x] [intro.compliance]
- [x] [dcl.spec.auto]
- [x] [class.mfct.non-static]
- [x] [class.union]
- [x] [allocator.requirements]
- [x] [cstdint]
- [x] [errno]
- [x] [c.locales]
- [x] [associative.reqmts]
- [x] [unord.req]
- [x] [unreachable.sentinels]
- [x] [view.interface]
- [x] [range.all]
- [x] [cfenv]
- [x] [conversions]
- [x] [fs.req]
- [x] [fs.class.filesystem.error]
- [x] [stopcallback]
- [x] [thread.sharedmutex.requirements]
- [x] [thread.sharedtimedmutex.requirements]
- [x] [depr.iterator.primitives]

jwakely:
[class.union] looks ok

jensmaurer:
@jwakely, how so? There is [class.union.anon] nested inside.

jwakely:
Oh I misread the directive as only talking about a subclause with a single .1 paragraph. But if course ISO doesn't even number paragraphs, it's taking about subclause numbers. Sorry.

jwakely:
N.B. the problem here is really that we have hanging paragraphs. If the content of [class.union] was in [class.union.xxx] then that would be a sibling of [class.union.anon], and we wouldn't be breaking this rule.

In the ISO world, where hanging paragraphs don't exist, a .1 subclause with no .2 sibling would mean the parent has no text of it's own and has an "only child". That's obviously pointless and the content of the only child could be moved up into the empty parent.

With hanging paragraphs the parent is not actually empty, and an only child can make sense.

jensmaurer:
Editorial meeting: This is duplicative of the hanging paragraphs issue #1730.

opensdh:
cplusplus/papers#533 plans to remove [class.this], making [class.mfct.non-static]'s paragraphs non-hanging.

zygoloid:
I'm using this Perl script to find these:
```
[...]

Issue #1730, opened 2017-09-06
[C++17 DIS comment 018] Hanging paragraphs


zygoloid:
Perhaps we should see how hard ISO pushes for this one. It's probably not too hard to script up the addition of a bunch of "General" headings if we need to, but I would worry that such a heading would be inappropriate in many cases.

jensmaurer:
Agreed: Fixing these for real (e.g. clause [expr]) means more work than adding a bunch of trivial subsection headings. Postpone to C++20.

zygoloid:
Response: REJECTED. We will consider this change for the next revision of the document, but adding meaningful headings for these paragraphs would require significant reorganization of the text and potentially updating of cross-references, which introduces too much risk for this stage of the standardization process.

zygoloid:
Leaving this open for consideration post-C++17 IS.

jensmaurer:
See also #574 and #1771.

zygoloid:
We have addressed a lot of issues here for C++20, but there's more work to do. Retargeting at C++23.

PR #4169, opened 2020-09-09
Fix hanging paragraphs for C++20

zygoloid:
Mechanically fix all hanging paragraphs by adding "General" sections.

For C++23, we should go through these and rename them as appropriate, but for C++20 let's just mechanically call them all "General" and name them [foo.general].

Fixes cplusplus/nbballot#394.


source/declarations.tex
```diff
@@ -2084,6 +2090,8 @@
 \indextext{specifier|)}%
 
 \rSec1[dcl.decl]{Declarators}%
+
+\rSec2[dcl.decl.general]{General}%
 \indextext{declarator|(}
```
jensmaurer, line 2095: This is not nice; the "declarator" start of index range should probably be before the "General" section. But maybe this is not the right time to fiddle with this.

jensmaurer (approved):
I'm creating an extension for tools/check.sh to detect hanging paragraphs using sed and awk.

johelegp:
Fixes #3525 (I think).
Partially addresses #1771.
Related to #1730 and #574.

source/declarations.tex
zygoloid: We should fix this for the `master` branch. I think this is fine for C++20.

zygoloid:
> Fixes #3525 (I think).

Well, technically this PR doesn't fix #3525, but the merge to master will :)

zygoloid:
> I'm creating an extension for tools/check.sh to detect hanging paragraphs using sed and awk.

Ah, we were both working on this at the same time. I built a perl script to do this; do you prefer one approach or the other?

jensmaurer:
My check script is here: #4170

johelegp:
> > Fixes #3525 (I think).
> 
> Well, technically this PR doesn't fix #3525, but the merge to master will :)

Maybe not:
> # FIXME: We should also check for subclauses that have no siblings.

Does your script fixes this, @jensmaurer?

jensmaurer:
@zygoloid , your script introduces a dependency on perl, which is novel for the check.sh script. My variant just uses sed and perl.

@johelegp , no, my script does not check for lone siblings. I think those only currently appear together with hanging paragraphs.

johelegp:
I've marked as done those that will be fixed by this. Please, do check the others: #3525.

zygoloid:
> @zygoloid , your script introduces a dependency on perl, which is novel for the check.sh script. My variant just uses sed and awk. 

For what it's worth, using `perl` here is actually more portable than your uses of `awk` (which rely on GNU awk-isms, and consequently don't work out of the box on Mac OS). But I'm fine sticking with something more similar to what we already do here if there's a worry that `perl` won't be available. Dropping the checking part from this PR.

jensmaurer:
@zygoloid , I think I'm not using GNU awk specifics, at least not in the "hanging paragraph" check.
(I think you discovered a GNU sed dependency, though.)

zygoloid:
Sorry, you're right, it's `sed` not `awk` that's the problem in this script in general.

zygoloid:
See also follow-up fixes #4173 and #4174.

PR #4172, opened 2020-09-09
Fix all hanging paragraphs by adding "General" sections.

### LINKED ISSUE #1730: [C++17 DIS comment 018] Hanging paragraphs

zygoloid:
Fixes #1730.
Partially addresses #574, #1771, #3525.


jensmaurer:
Introduces

future.tex:Overfull \hbox (2.28091pt too wide) in paragraph at lines 355--379

PR #4174, opened 2020-09-09
Replace "this subclause" with an explicit reference in all formerly-hanging paragraphs

zygoloid:
These typically mean the parent subclause now. Mechanically:

"this subclause" became "\ref{parent}".

"This subclause" became "Subclause \ref{parent}", because it doesn't
look great to start a sentennce with a subclause number.

... with some manual fixups for cases where that didn't work out well.

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -442,7 +442,7 @@ and \tcode{x.front()} can be called only if \tcode{x.empty()} returns
 \rSec2[description.general]{General}
 
 \pnum
-This subclause describes the conventions used to specify the \Cpp{} standard
+Subclause \ref{description} describes the conventions used to specify the \Cpp{} standard
 library. \ref{structure} describes the structure of the normative
 \ref{\firstlibchapter} through \ref{\lastlibchapter} and
 \ref{depr}. \ref{conventions} describes other editorial conventions.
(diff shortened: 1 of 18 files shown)


source/lib-intro.tex
```diff
@@ -1119,7 +1119,7 @@
 programs having only a single thread of execution.
 
 \pnum
-Within this subclause, \ref{organization} describes the library's contents and
+\ref{organization} describes the library's contents and
```
jensmaurer, line 1122: Is this a case for "Subclause" at the start of the sentence? Maybe repeat "subclause" for the other parts of the enumeration.

source/locales.tex
```diff
@@ -1043,8 +1043,8 @@
 in case of a parse error.
 
 \pnum
-Within this clause it is unspecified whether one virtual function calls another
-virtual function.
+Within subclause \ref{locale.categories} it is unspecified whether one virtual
+function calls another virtual function.
```
jensmaurer, line 1047: This linebreak doesn't move us towards semantic line breaks.
Either leave the original (bad) linebreak (i.e. don't touch) or make it right.

jensmaurer (changes requested):
Two concerns.

source/lib-intro.tex
zygoloid: I decided to leave off the "Subclause" here for consistency with what we already do two paragraphs later. Happy to change it as suggested if you would prefer.

source/locales.tex
zygoloid: Sorry, that was the result of a `gq` in vim on autopilot :) Fixing.

source/lib-intro.tex
jensmaurer: Fine, no big deal either way.

PR #4175, opened 2020-09-09
[intro.ack] Dissolve subclause.

jensmaurer:
Integrate trademark acknowledgements into [intro.refs].
Add base works to the bibliography.

Fixes cplusplus/nbballot#399

### DIFF
--- a/source/back.tex
+++ b/source/back.tex
@@ -2,10 +2,20 @@
 
 \chapter{Bibliography}
 
-The following documents are cited informatively in this document.
-
 \begin{itemize}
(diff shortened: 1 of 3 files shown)


source/back.tex
```diff
@@ -6,6 +6,18 @@ \chapter{Bibliography}
```
zygoloid, line 6: The statement on the previous line isn't true; we don't cite some of these documents informatively. ISO/CS wanted that statement removed, so let's do that at the same time.

jensmaurer: Removed.

PR #4178, opened 2020-09-10
Fix solitary subclauses

zygoloid:
[ISO/CS comment 018 on the C++20 DIS](https://github.com/cplusplus/nbballot/issues/394) requests that we avoid subclauses with no siblings.

Merging this to master will address #3525.

Fixes cplusplus/nbballot#394


jensmaurer (approved):
Looks good.

Issue #392, opened 2014-11-06
[class.static.data] unusual wording


cpplearner:
9.4.2/2 says "The declaration of a static data member in its class definition is not a definition and may be of an incomplete type other than cv-qualified void."
Why not simply say "[...] and may be of an incomplete type other than cv void." or "[...] and may be of an incomplete object type."?

PR #4180, opened 2020-09-10
[definitions] Integrate into [intro.defs]

jensmaurer:

Fixes cplusplus/nbballot#396


jensmaurer:
For easier review, some follow-up fixes have been put into separate commits. Squash (or not) as you please.

tkoeppe:
See #392 for final disposition of the single ballot comment that contains this detail.

PR #4182, opened 2020-09-10
[intro.refs] Fix clause reference to ISO/IEC 9899.

jensmaurer:
Fixes cplusplus/nbballot#402


jensmaurer:
I haven't found any other place where we did the wrong thing in conjunction with "ISO/IEC" external references.
We do have a lot of "See also: ISO C x.y" cross-references; I'm assuming those are not in scope to be changed to "See also: ISO/IEC 9899:year, x.y".

PR #4186, opened 2020-09-10
Fix solitary subclauses

### LINKED ISSUE #3525: [std] There shall be no .1 subclause without a .2 sibling

zygoloid:
Merge of #4178 to master.

Fixes #3525.

### DIFF
--- a/source/future.tex
+++ b/source/future.tex
@@ -1660,9 +1660,7 @@ with a member typedef \tcode{type} that names the following type:
 \end{itemize}
 \end{itemdescr}
 
-\rSec1[depr.iterator.primitives]{Deprecated iterator primitives}
-
-\rSec2[depr.iterator.basic]{Basic iterator}
+\rSec1[depr.iterator]{Deprecated \tcode{iterator} class template}
 
 \pnum
 The header \libheaderrefx{iterator}{iterator.synopsis} has the following addition:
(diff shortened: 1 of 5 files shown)


johelegp:
This missed https://github.com/cplusplus/draft/pull/4178/files#diff-601602f584c2b79724118aad6ebef4ae (delta [unreachable.sentinels]).

jensmaurer:
I don't understand. When we introduce new section, we don't add an entry to xrefdelta.

johelegp:
Do you mean that the [unreachable.sentinels] entry shouldn't have been there? I think so.

PR #4188, opened 2020-09-10
[intro.refs,time.format] Fix normative references.

jensmaurer:
ISO/IEC/IEEE 60559:2011 and UAX#29 are not normative
requirements of C++ and thus were moved to the bibliography.
For ISO 8601:2004, highlighted its normative impact on
time formatting.
Move the footnote about the Unicode trademark to the
new first mention of the term in [intro.memory].

Fixes ISO/CS 002 (C++20 DIS)

Fixes cplusplus/nbballot#378

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3056,7 +3056,11 @@ The fundamental storage unit in the \Cpp{} memory model is the
 A byte is at least large enough to contain any member of the basic
 \indextext{character set!basic execution}%
 execution character set\iref{lex.charset}
-and the eight-bit code units of the Unicode UTF-8 encoding form
+and the eight-bit code units of the Unicode%
+  \footnote{Unicode\textregistered\ is a registered trademark of Unicode, Inc.
+  This information is given for the convenience of users of this document and
+  does not constitute an endorsement by ISO or IEC of this product.}
+UTF-8 encoding form
 and is composed of a contiguous sequence of
 bits,\footnote{The number of bits in a byte is reported by the macro
 \tcode{CHAR_BIT} in the header \libheaderref{climits}.}
--- a/source/back.tex
+++ b/source/back.tex
@@ -26,6 +26,16 @@
 \item
   ISO 4217:2015,
   \doccite{Codes for the representation of currencies}
+\item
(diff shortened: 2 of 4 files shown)


source/basic.tex
```diff
@@ -3056,7 +3056,11 @@
 A byte is at least large enough to contain any member of the basic
 \indextext{character set!basic execution}%
 execution character set\iref{lex.charset}
-and the eight-bit code units of the Unicode UTF-8 encoding form
+and the eight-bit code units of the Unicode%
+  \footnote{Unicode\textregistered\ is a registered trademark of Unicode, Inc.
+  This information is given for the convenience of users of this document and
+  does not constitute an endorsement by ISO or IEC of this product.}
```
johelegp, line 3062: I assumed from https://github.com/cplusplus/nbballot/issues/399 that this only needs to appear the first time, which already does with its resolving PR https://github.com/cplusplus/draft/pull/4175.

johelegp: I see that you moved this text from intro.tex to here, even though intro.tex appears first in the PDF. If their appearance in intro.tex doesn't count as the first time, shouldn't the other trademarks be moved accordingly? ECMA's to [tab:re.synopt] and POSIX's to, maybe, [namespace.posix]. Should I open an issue?

jensmaurer: We are moving the reference to UAX#29 from the front matter in [intro.refs] to the back of the standard in the bibliography.
The footnote explaining that "Unicode" is a trademark is now attached to the first appearance of "Unicode" in the standard, which is no longer the bibliographical reference, but the naming of Unicode in [basic].
In contrast, ECMA and POSIX references remain in [intro.refs] and thus the trademark footnote stays there.

source/basic.tex
```diff
@@ -3056,7 +3056,11 @@
...
+UTF-8 encoding form
```
zygoloid, line 3063: It seems a bit silly to talk about UTF-8 at all here, now that we guarantee 2's complement. "eight-bit" is all we need to say; anything else is irrelevant rationalization.

But let's go with what you have for C++20; we can reconsider this for C++23.

PR #4190, opened 2020-09-10
[everywhere] Follow ISO rules for the presentation of notes and examples

zygoloid:
Number notes and examples, make them one point smaller, and move them to (un-numbered) paragraphs of their own.

Partially addresses ISO/CS 016 (C++20 DIS).

Fixes https://github.com/cplusplus/nbballot/issues/398.

### DIFF
--- a/source/layout.tex
+++ b/source/layout.tex
@@ -32,14 +32,18 @@
 %%--------------------------------------------------
 %% Paragraph and bullet numbering
 
-\newcounter{Paras}
-\counterwithin{Paras}{chapter}
-\counterwithin{Paras}{section}
-\counterwithin{Paras}{subsection}
-\counterwithin{Paras}{subsubsection}
-\counterwithin{Paras}{paragraph}
-\counterwithin{Paras}{subparagraph}
+% create a new counter that resets for each new subclause
+\newcommand{\newsubclausecounter}[1]{
+\newcounter{#1}
+\counterwithin{#1}{chapter}
+\counterwithin{#1}{section}
+\counterwithin{#1}{subsection}
+\counterwithin{#1}{subsubsection}
+\counterwithin{#1}{paragraph}
+\counterwithin{#1}{subparagraph}
+}
 
+\newsubclausecounter{Paras}
 \newcounter{Bullets1}[Paras]
 \newcounter{Bullets2}[Bullets1]
 \newcounter{Bullets3}[Bullets2]
--- a/source/layout.tex
+++ b/source/layout.tex
@@ -65,8 +65,13 @@
 }}}
 \makeatother
 
-\def\pnum
-{\parabullnum{Paras}{0pt}}
+% Register our intent to number the next paragraph. Don't actually number it
+% yet, because we might have a paragraph break before we see its contents (for
+% example, if the paragraph begins with a note or example).
+\def\pnum{%
+\global\def\maybeaddpnum{\global\def\maybeaddpnum{}\parabullnum{Paras}{0pt}}%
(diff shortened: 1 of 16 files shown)


zygoloid:
Note that this adds an overfull vbox somewhere in Clause 23, but I've not been able to determine what it's referring to yet.

source/macros.tex
```diff
@@ -634,4 +641,6 @@
 \let\addcontentsline\oldcontentsline%
 }
 \newcommand{\defncontext}[1]{\textlangle#1\textrangle}
-\newenvironment{defnote}{\addtocounter{termnote}{1}\noteintro{Note \thetermnote{} to entry}}{\noteoutro{note}\space}
+\newenvironment{defnote}
```
jensmaurer, line 644: While we're at it, shouldn't this get a \par in front, too?

source/layout.tex
```diff
@@ -65,8 +65,13 @@
...
 \makeatother
 
-\def\pnum
-{\parabullnum{Paras}{0pt}}
+% Register our intent to number the next paragraph. Don't actually number it
+% yet, because we might have a paragraph break before we see its contents (for
+% example, if the paragraph beings with a note or example).
+\def\pnum{%
+\global\def\maybeaddpnum{\global\def\maybeaddpnum{}\parabullnum{Paras}{0pt}}%
+\everypar=\expandafter{\the\everypar\maybeaddpnum}%
```
jensmaurer, line 73: You're never resetting \everypar (which is probably fine, given that a large number of paragraphs should be numbered).

However, that means you can set \everypar once (outside of \pnum) and not reset it to the same value every time.

Otherwise, very nice solution.

jensmaurer (changes requested):
minor stuff

jensmaurer:
Overfull \vbox: Maybe it's the longtable in [bidirectional.iterators]. Its bottom part looks very close to the footer.

source/layout.tex
zygoloid: memoir sets `\everypar` itself at various points that are not necessarily easy to know about in advance. Setting it only once won't work. I'd like to reset it after invoking `\maybeaddpnum`, but it's not clear that there's a reliable way to do that without clobbering memoir's own uses of it.

This approach might end up with a few dozen `\maybeaddpnum` calls per `\par` for subclauses with lots of `\pnum`s (at least `\rSec` resets `\everypar`, so the number is bounded above by the number of `\pnum`s in a subclause). That doesn't seem to cause any real problems so far. If there's a reliable way to reset `\everypar` without clobbering other uses of it, I'd be very happy to use such a thing!

jensmaurer: Ok, fine-ish, I guess.

source/macros.tex
zygoloid: Done, and duplication between this and note/example eliminated too.

source/layout.tex
```diff
@@ -65,8 +65,13 @@
 }}}
 \makeatother
 
-\def\pnum
-{\parabullnum{Paras}{0pt}}
+% Register our intent to number the next paragraph. Don't actually number it
+% yet, because we might have a paragraph break before we see its contents (for
+% example, if the paragraph beings with a note or example).
```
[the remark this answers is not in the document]
zygoloid: Done.

zygoloid:
> Overfull \vbox: Maybe it's the longtable in [bidirectional.iterators]. Its bottom part looks very close to the footer.
> Plus it's two longtables on the page.

Yes, looks like that was it. Fixed.

PR #4191, opened 2020-09-10
[intro.defs] Remove cross-references to main part

jensmaurer:

Fixes cplusplus/nbballot#397

### DIFF
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -110,12 +110,13 @@ defined.
 \indexdefn{access}%
 \definition{access}{defns.access}
 \defncontext{execution-time action}
-read\iref{conv.lval} or
-modify (\ref{expr.ass}, \ref{expr.post.incr}, \ref{expr.pre.incr})
-the value of an object
+read or modify the value of an object
 
 \begin{defnote}
 Only objects of scalar type can be accessed.
+Reads of scalar objects are described in \ref{conv.lval} and
+modifications of scalar objects are describred in
+\ref{expr.ass}, \ref{expr.post.incr}, and \ref{expr.pre.incr}.
 Attempts to read or modify an object of class type
 typically invoke a constructor\iref{class.ctor}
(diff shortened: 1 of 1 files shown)


tkoeppe (approved):
Maybe extend the commit message to say "cross-references from the Terms and Definitions to the main part".

source/intro.tex
```diff
@@ -121,9 +121,7 @@
 \indexdefn{access}%
 \definition{access}{defns.access}
 \defncontext{execution-time action}
-read\iref{conv.lval} or
-modify (\ref{expr.ass}, \ref{expr.post.incr}, \ref{expr.pre.incr})
-the value of an object
```
zygoloid, line 126: Can we add some text to the note below to retain these useful xrefs? Perhaps add between the first and second sentence:

> Reads of scalar objects are described in \ref{conv.lval} and modifications of scalar objects are describred in \ref{expr.ass}, \ref{expr.post.incr}, and \ref{expr.pre.incr}.

I think it's important to the use of this definition that we know what these four cases are.

source/intro.tex
```diff
@@ -345,8 +343,8 @@
 
 \definition{iostream class templates}{defns.iostream.templates}
 \defncontext{library}
-templates, defined in \ref{input.output},
-that take two template arguments
+templates that take two template arguments
+and are declared in header \libheader{iosfwd}
```
zygoloid, line 347: Can we keep the two conditions the other way around? It makes more sense to me to start with all templates in `<iosfwd>` and then filter by "takes two arguments" than to start with the universe of all templates taking two arguments and then filter by "declared in `<iosfwd>`".

source/intro.tex
jensmaurer: Fixed.

source/intro.tex
jensmaurer: Fixed.

PR #4203, opened 2020-09-18
[everywhere] Reduce width of space after Note: and Example:.

zygoloid:
Use upright font for : in Preconditions: and friends.


jensmaurer (approved):
Yes, that's an improvement, I think.

PR #4209, opened 2020-09-20
[conv.rank] Avoid hinting that 'bool' be a standard integer type.

jensmaurer:
Partially addresses #3974

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -5316,7 +5316,7 @@ rank of any extended integer type with the same width.
 \item The rank of \keyword{char} equals the rank of \tcode{\keyword{signed} \keyword{char}}
 and \tcode{\keyword{unsigned} \keyword{char}}.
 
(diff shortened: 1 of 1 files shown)


jwakely (approved):

zygoloid:
I'm not convinced that this reflects the intent. As with #3974 I want CWG oversight before we try to normatively resolve whether "standard integer types" were intended to include all standard integer types or not.

tkoeppe:
@jensmaurer, @zygoloid: should we close this issue, since this is firmly in CWG's territory?

jensmaurer:
I think this should be presented to CWG on the "editorial oversight" track. I've added this issue to https://wiki.edg.com/bin/view/Wg21telecons2021/CoreWorkingGroup

jensmaurer:
Approved by CWG as an editorial fix.

jensmaurer:

PR #4215, opened 2020-09-20
[expr.unary.op] Use 'negative', not 'negation'.

### LINKED ISSUE #4131: [expr.unary.op]/8
[[expr.unary.op]/8](http://eel.is/c++draft/expr.unary#op-8):

> The operand of the unary `-` operator shall have arithmetic or unscoped enumeration type and the result is the **negation** of its operand. ...

The **negation** is obviously wrong.

jensmaurer:
Fixes #4131


tkoeppe (approved):

zygoloid (changes requested):
Let's hold off on merging this until we come to consensus on the issue as to what change we want here.

PR #4218, opened 2020-09-21
footnote environment

jensmaurer:
Create a "footnoteenv" environment for footnotes.
This allows to clearly separate them in the LaTeX source, enabling automatic checking for normative words such as "shall" or "may" inside them.

[temp.constr.order] has an example each inside two footnotes that now gets formatted differently.
We should get rid of that for unrelated reasons (bad rendering with the new separate-paragraph examples).

With this compensated, this patch has no visible differences in "appearance" mode except for footnote 12 (different hyphenation) and footnote 256 (slight differences in spacing/kerning, it seems).

### DIFF
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -289,6 +289,22 @@
 \newnoteenvironment{note}{Note \arabic{note}}{end note}
 \newnoteenvironment{example}{Example \arabic{example}}{end example}
 
+\makeatletter
+\let\footnote\@undefined
+\global\newsavebox{\@tempfootboxa}   % must be global, to escape tables/figures
+\newsavebox{\@templfootbox}
+\newenvironment{footnote}{%
+  \unskip\footnotemark%    no space before the mark
+  \normalfont%
+  \footnotesize%           text size for rendering the footnote text
+  \begin{lrbox}{\@templfootbox}%       temporarily save to local box
+}{%
+  \end{lrbox}%
+  \global\setbox\@tempfootboxa\hbox{\unhbox\@templfootbox}%  copy to global box
+  \footnotetext{\unhbox\@tempfootboxa}%
+}
+\makeatother
+
 %% Library function descriptions
 \newcommand{\Fundescx}[1]{\textit{#1}}
 \newcommand{\Fundesc}[1]{\Fundescx{#1}:\space}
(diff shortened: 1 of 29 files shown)


jensmaurer:
This is work-in-progress.

jensmaurer:
... and (drum roll) with a slight patch to detect "shall", "may", "should" also inside footnotes, we now find

lex.tex:672:characters may produce a long external identifier, but \Cpp{} does not <--- "shall", "should", or "may" inside a note
preprocessor.tex:558:a preprocessing token shall not follow a <--- "shall", "should", or "may" inside a note
preprocessor.tex:564:comments may appear anywhere in a source file, <--- "shall", "should", or "may" inside a note
support.tex:2490:expression, may, however, increase the \tcode{size} argument to <--- "shall", "should", or "may" inside a note
templates.tex:9077:may be deduced from an array bound, the resulting value will always be <--- "shall", "should", or "may" inside a note

jensmaurer:
@godbyk , I think I need your help here. Please have a look at this pull request and consider the definition of the "footnoteenv" macro in macros.tex. The goal is to have a proper environment for footnotes and not the inline-style `\footnote{...}` stuff. The proper environment helps our script-based sanity checking of the LaTeX source code.

I essentially got it working with some "lrbox" operations, and it looks as before (except for a hyphenation change that I can only explain by some difference in hyphenation regime in footnotes, but that's not the point here).

In order to make the "footnoteenv" work inside tables inside floats, I need to make sure I employ \global to have the footnote's contents "escape" properly. So far, so good. However, I am unable to achieve this with a named box; it only works with a numbered box. See the line that copies the local \box99 to the \global named box. Of course, using a specific numbered box risks conflicts. For example, \box0 seems to cause such a conflict; all page numbers in the cross-reference are off by the size of the table of contents. My attempts at using a named box have failed with "expected a <box> here" errors. Do you have any ideas?

jensmaurer:
Rebased due to conflict.

jensmaurer:
Rebased again. This is now ready for revuew/merge, from my perspective.

source/macros.tex
```diff
@@ -289,6 +289,24 @@
 \newnoteenvironment{note}{Note \arabic{note}}{end note}
 \newnoteenvironment{example}{Example \arabic{example}}{end example}
 
+% \show \lrbox
+% \show \endlrbox
```
zygoloid, line 293: Please remove these comments.

zygoloid (approved):

PR #4222, opened 2020-09-22
Fix modal verb "must"

burblebee:
Partially addresses cplusplus/nbballot#393

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -4079,8 +4079,9 @@ is additional semantic information.
 \begin{note}
 Modifying any list requires access to the element preceding the first element
 of interest, but in a \tcode{forward_list} there is no constant-time way to access a
-preceding element. For this reason, ranges that are modified, such as those supplied to
-\tcode{erase} and \tcode{splice}, must be open at the beginning.
+preceding element.
+For this reason, \tcode{erase_after} and \tcode{splice_after}
+take fully-open ranges, not semi-open ranges.
 \end{note}
 
 \begin{codeblock}
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -6434,13 +6434,16 @@ of the operand expression \tcode{E2} as follows:
 
 \begin{itemize}
 \item If \tcode{E2} is an lvalue, the target type is
-``lvalue reference to \tcode{T2}'', subject to the
-constraint that in the conversion the reference must bind
-directly\iref{dcl.init.ref} to a glvalue.
+``lvalue reference to \tcode{T2}'', and
+implicit conversion sequences for which
+the reference does not bind directly\iref{dcl.init.ref}
+to a glvalue are disregarded.
 
 \item If \tcode{E2} is an xvalue, the target type is
-``rvalue reference to \tcode{T2}'',
-subject to the constraint that the reference must bind directly.
+``rvalue reference to \tcode{T2}'', and
+implicit conversion sequences for which
+the reference does not bind directly
+are disregarded.
 
 \item If \tcode{E2} is a prvalue or if neither of the conversion sequences above can be
 formed and at least one of the operands has (possibly cv-qualified) class type:
@@ -6615,8 +6618,8 @@ The exception is reactivated with the existing exception object;
 no new exception object is created.
 The exception is no longer considered to be caught.
 \begin{example}
-Code that must be executed because of an exception, but cannot
-completely handle the exception itself, can be written like this:
+An exception handler that cannot completely handle the exception itself
+can be written like this:
 \begin{codeblock}
 try {
   // ...
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -6434,16 +6434,13 @@ of the operand expression \tcode{E2} as follows:
 
 \begin{itemize}
 \item If \tcode{E2} is an lvalue, the target type is
-``lvalue reference to \tcode{T2}'', and
-implicit conversion sequences for which
-the reference does not bind directly\iref{dcl.init.ref}
-to a glvalue are disregarded.
+``lvalue reference to \tcode{T2}'', subject to the
+constraint that in the conversion the reference binds
+directly\iref{dcl.init.ref} to a glvalue.
 
 \item If \tcode{E2} is an xvalue, the target type is
-``rvalue reference to \tcode{T2}'', and
-implicit conversion sequences for which
-the reference does not bind directly
-are disregarded.
+``rvalue reference to \tcode{T2}'',
+subject to the constraint that the reference binds directly.
 
 \item If \tcode{E2} is a prvalue or if neither of the conversion sequences above can be
 formed and at least one of the operands has (possibly cv-qualified) class type:
--- a/source/grammar.tex
+++ b/source/grammar.tex
@@ -10,8 +10,8 @@ It is not an exact statement of the language.
 In particular, the grammar described here accepts
 a superset of valid \Cpp{} constructs.
 Disambiguation rules~(\ref{stmt.ambig}, \ref{dcl.spec}, \ref{class.member.lookup})
-must be applied to distinguish expressions from declarations.
-Further, access control, ambiguity, and type rules must be used
+are applied to distinguish expressions from declarations.
+Further, access control, ambiguity, and type rules are used
 to weed out syntactically valid but meaningless constructs.
 
 \rSec1[gram.key]{Keywords}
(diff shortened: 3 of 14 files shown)


source/containers.tex
```diff
@@ -4079,10 +4079,12 @@
...
 Modifying any list requires access to the element preceding the first element
 of interest, but in a \tcode{forward_list} there is no constant-time way to access a
-preceding element. For this reason, ranges that are modified, such as those supplied to
-\tcode{erase} and \tcode{splice}, must be open at the beginning.
+preceding element.
+For this reason, \tcode{erase_after} and \tcode{splice_after}
+take fully-open ranges, not semi-open ranges.
 \end{note}
 
+%FIXME: Are these examples wrong?
```
zygoloid, line 4087: It looks like this FIXME is applying to the `forward_list` synopsis. Is the FIXME still necessary after the above change?
```suggestion
```

source/expressions.tex
```diff
@@ -6612,7 +6615,7 @@
 no new exception object is created.
 The exception is no longer considered to be caught.
 \begin{example}
-Code that must be executed because of an exception, but cannot
+Code that is executed because of an exception, but cannot
 completely handle the exception itself, can be written like this:
```
zygoloid, line 6619: I don't like this phrasing (before or after the "must" fix). I think this would be good enough for now:
```suggestion
An exception handler that cannot completely handle the exception itself
can be written like this:
```

source/grammar.tex
```diff
@@ -10,8 +10,8 @@
 In particular, the grammar described here accepts
 a superset of valid \Cpp{} constructs.
 Disambiguation rules~(\ref{stmt.ambig}, \ref{dcl.spec}, \ref{class.member.lookup})
-must be applied to distinguish expressions from declarations.
-Further, access control, ambiguity, and type rules must be used
+shall be applied to distinguish expressions from declarations.
+Further, access control, ambiguity, and type rules shall be used
```
zygoloid, line 14:
```suggestion
are applied to distinguish expressions from declarations.
Further, access control, ambiguity, and type rules are used
```

We can't "shall" from an informative Annex, and I think this is supposed to be descriptive rather than introducing a requirement.

source/iostreams.tex
```diff
@@ -14251,8 +14249,7 @@
 \begin{note}
 Implementations of \tcode{directory_iterator}\iref{fs.class.directory.iterator}
 are prohibited from directly or indirectly calling the \tcode{refresh} function
-since it must access the external file system,
-and the objective of caching is to avoid unnecessary file system accesses.
+as described in \ref{fs.class.directory.iterator.general}".
```
zygoloid, line 14252:
```suggestion
as described in \ref{fs.class.directory.iterator.general}.
```

source/overloading.tex
```diff
@@ -1604,7 +1604,7 @@
 \pnum
 When resolving a placeholder for a deduced class type\iref{dcl.type.simple}
 where the \grammarterm{template-name} names an alias template \tcode{A},
-the \grammarterm{defining-type-id} of \tcode{A} must be of the form
+the \grammarterm{defining-type-id} of \tcode{A} shall be of the form
```
zygoloid, line 1607: We have an "as specified in" below pointing to a normative rule elsewhere, so I think "must" here was OK following our "internal constraint" meaning.
```suggestion
the \grammarterm{defining-type-id} of \tcode{A} must be of the form
```

source/statements.tex
```diff
@@ -1001,8 +1001,9 @@
 declaration.
 If control enters the declaration concurrently while the variable is
 being initialized, the concurrent execution shall wait for completion
-of the initialization.\footnote{The implementation must not introduce
-any deadlock around execution of the initializer. Deadlocks might
+of the initialization. The implementation shall not introduce
+any deadlock around execution of the initializer.
+\footnote{Deadlocks might
 still be caused by the program logic; the implementation need only
 avoid deadlocks due to its own synchronization operations.} If control
```
zygoloid, line 1008: I'm nervous about this; the phrasing of this is vague, and "shall"ing it might add a requirement that's stronger than what we meant. I think rather than a "shall" here, we want to point out that introducing a deadlock would not satisfy "as if". How about this:

```suggestion
of the initialization.
\begin{note}
A conforming implementation cannot introduce
any deadlock around execution of the initializer.
Deadlocks might still be caused by the program logic;
the implementation need only avoid deadlocks
due to its own synchronization operations.
\end{note}
If control
```

source/templates.tex
```diff
@@ -694,18 +694,13 @@
...
-For a
-\grammarterm{template-name}
-to be explicitly qualified by the template arguments,
-the name must be considered to refer to a template.
+An identifier is a \grammarterm{template-name}
+if it is associated by name lookup with
+a template or an overload set that contains a function template, or,
+if the identifier is followed by \tcode{<},
+the \grammarterm{template-id} would form an \grammarterm{unqualified-id}, and
+name lookup either finds one or more functions or finds nothing.
```
zygoloid, line 703: Please also delete the text after the note that this partially duplicates.

source/templates.tex
```diff
@@ -694,18 +694,13 @@
...
-that is, to identify a name as a
-\grammarterm{template-name}.
-\end{note}
-
 \pnum
-For a
-\grammarterm{template-name}
-to be explicitly qualified by the template arguments,
-the name must be considered to refer to a template.
+An identifier is a \grammarterm{template-name}
```
zygoloid, line 698:
```suggestion
An \grammarterm{identifier} is a \grammarterm{template-name}
```

source/templates.tex
```diff
@@ -694,18 +694,13 @@
...
+if it is associated by name lookup with
+a template or an overload set that contains a function template, or,
+if the identifier is followed by \tcode{<},
```
zygoloid, line 701:
```suggestion
a template or an overload set that contains a function template, or
the \grammarterm{identifier} is followed by \tcode{<},
```

source/templates.tex
```diff
@@ -7136,12 +7131,13 @@
 otherwise, it is a declaration.
 \begin{note}
 The definition of a static data member of a template
-that requires default-initialization must use a \grammarterm{braced-init-list}:
+that requires default-initialization
```
zygoloid, line 7134:
```suggestion
for which default-initialization is desired
```
(Just a drive-by to avoid using "requires" where we're not specifying a requirement.)

source/threads.tex
```diff
@@ -94,7 +94,7 @@
 The functions whose names end in \tcode{_for} take an argument that
 specifies a duration. These functions produce relative timeouts. Implementations
 should use a steady clock to measure time for these functions.\footnote{All
```
zygoloid, line 96:
```suggestion
should use a steady clock to measure time for these functions.%
```

source/threads.tex
```diff
@@ -94,7 +94,7 @@
...
-implementations for which standard time units are meaningful must necessarily
+implementations for which standard time units are meaningful will typically
```
zygoloid, line 97:
```suggestion
\footnote{Implementations for which standard time units are meaningful will typically
```
("All implementations [...] will typically" sounds self-contradictory.)

zygoloid (commented):
Please include "Partially addresses ISO/CS 017 (C++20 DIS)." in the commit description.

source/containers.tex
burblebee: Removed FIXME

source/expressions.tex
burblebee: done

source/grammar.tex
burblebee: done

source/iostreams.tex
burblebee: doh!  thanks, done

source/overloading.tex
burblebee: done

source/statements.tex
burblebee: done

source/templates.tex
burblebee: I initially grammarterm'ed these identifiers, but we never talk about a grammarterm being another grammarterm; we speak of it as having the form of another grammarterm.

Suggestions applied nonetheless.

source/threads.tex
burblebee: done, but why the trailing %?

zygoloid: To avoid a space between the period and the footnote marker.

burblebee:
> Please include "Partially addresses ISO/CS 017 (C++20 DIS)." in the commit description.

Will do when I rebase and force push; should I do that now or wait for more feedback?

burblebee:
> > Please include "Partially addresses ISO/CS 017 (C++20 DIS)." in the commit description.
> 
> Will do when I rebase and force push; should I do that now or wait for more feedback?

Went ahead and did this now since c++20 was force-pushed.

source/templates.tex
```diff
@@ -715,10 +710,6 @@
 name lookup finds
 a \grammarterm{template-name}
 or an overload set that contains a function template.
```
zygoloid, line 712: Please remove this sentence too, this is also captured by the new text above.

Issue #4228, opened 2020-09-23
Process: page-by-page review tracking


tkoeppe:
See #4219 for detailed instructions.

Please edit this comment to claim a section of the document.
DP: See comment for questions about the sections I reviewed.

- {TK: good} Front matter + Clauses 1, 2, 3
- {TK: done} 4 General principles[intro]
- {TK: done} 5 Lexical conventions[lex]
- {TK: done} 6 Basics[basic]
- {DP: done} 7 Expressions[expr]
- {DP: done} 8 Statements[stmt.stmt]
- {DP: done} 9 Declarations[dcl.dcl]
- {DP: good} 10 Modules[module]
- {DP: done} 11 Classes[class]
- {DP: done} 12 Overloading[over]
- {DP: done, see note in comment below} 13 Templates[temp]
- {DP: good} 14 Exception handling[except]
- {DP: good} 15 Preprocessing directives[cpp]
- {DP: done} 16 Library introduction[library]
- {DP: done} 17 Language support library[support]
- {DP: done} 18 Concepts library[concepts]
- {DP: good} 19 Diagnostics library[diagnostics]
- {DP: done} 20 General utilities library[utilities]
- {DP: done} 21 Strings library[strings]
- {DP: done} 22 Containers library[containers]
- {TK: done} 23 Iterators library[iterators]
- {TK: done} 24 Ranges library[ranges]
- (JM: done) 25 Algorithms library[algorithms]
- (JM: good) 26 Numerics library[numerics]
- (JM: done; still done after "tailnote") 27 Time library[time]
- (JM: done) 28 Localization library[localization]
- (JM: re-done after #4249) 29 Input/output library[input.output]
- (JM: done) 30 Regular expressions library[re]
- (JM: done) 31 Atomic operations library[atomics]
- {JM: ~done~ ~please revisit~ still done} 32 Thread support library[thread]
- {TK: done} Annex A (informative) Grammar summary[gram]
- {TK: good} Annex B (~informative~ normative) Implementation quantities[implimits]
- {TK: done} Annex C (informative) Compatibility[diff]
- {TK: done} Annex D (normative) Compatibility features
- {TK: done} Back matter

Key:

* "good": Good as is, no changes needed.
* "done": Good after review, but changes need to be applied (and are ready to be merged).
* anything else: still needs attention

tkoeppe:
[...]

PR #4231, opened 2020-09-24
Page breaks (jmaurer)

jensmaurer:
Partially addresses #4228


zygoloid:
26.8: I wonder if we could keep the `foo`, `fool`, `foof` overloads together on the same page. Not a big deal, though, and I don't think we should do this for C++20.

32.5.4.4.2: The widowed class-head is unfortunate. But I think we can live with this if a newpage here causes problems. (Similar in 32.5.5.3, but I think that'd be too much to add a newpage before if fixing 32.5.4.4.2 doesn't affect it.)

32.9.10.2/17: I think we should move the "*Error conditions:*" line onto the next page with its bullets.

Other than that, Clause 25-32 look good. Thanks!

jensmaurer:
I've addressed the [thread.] stuff. Regrettably, there was quite a bit of fall-out; needs re-review Force-pushed.

zygoloid (approved):
Thanks, looks great!

PR #4232, opened 2020-09-24
[futures.async] Avoid note within a sentence.

jensmaurer:
Partially addresses #4228


source/threads.tex
```diff
@@ -7184,6 +7180,11 @@
 \end{note}
```
tkoeppe, line 7180: What about this other note -- would it also work better outside the list?

jensmaurer: Maybe. I'm trying to limit the size of the changes this late in the process.

tkoeppe (approved):

source/threads.tex
zygoloid: We could further limit the size of the changes by turning these bullets into proper sentences. (There's already a period before the first note.) Eg:

> *Synchronization:* The following apply regardless of the provided `policy` argument:
> - The invocation of `async` synchronizes with the invocation of `f`.
>    [Note:...]
> - The completion of the function `f` is sequenced before the shared state is made ready.
>    [Note:...]

For master we could be a bit more aggressive:

> *Synchronization:* The invocation of `async` synchronizes with the invocation of `f`. The completion of the function `f` is sequenced before the shared state is made ready.
> [Note: These apply regardless of the provided `policy` argument, and even if the corresponding `future` object is moved to another thread. However, `f` might not be called at all, so its completion might never happen.]

tkoeppe: I like both parts of this.

jensmaurer:
This is now the minimal fix against C++20. Please check.

tkoeppe (approved):

jensmaurer:
The larger fix for C++23 is in #4244.

PR #4233, opened 2020-09-24
Page breaks (tkoeppe)

tkoeppe:
Partially addresses #4228

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -235,6 +235,9 @@ copy constructor, move constructor\iref{class.copy.ctor},
 copy assignment operator, move assignment operator\iref{class.copy.assign},
 or destructor\iref{class.dtor} member functions.
 \end{note}
+
+\newpage
+
 \begin{example}
 Given
 \begin{codeblock}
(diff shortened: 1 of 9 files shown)


zygoloid:
Do you think we should insert a page break before [lex.literal]?

Other than that, looks good to the end of Clause 6.

[iterator.concept.sentinel]: wrapping of `sentinel_for` is unfortunate. But maybe there's nothing we can do about that without violating local consistency. It's not terrible.

Otherwise, Clause 23 and 24 look OK.

C.5.4 [diff.expr] paragraph 1 has too much vertical whitespace after the second codeblock. Same thing in C.5.2/3 and C.5.3/1 and ... looks like we get bonus vertical space everywhere that a codeblock is followed by one of our bold Annex C labels.

Otherwise, Annexes look OK.

Back matter: split_view::inner-iterator and ::outer-iterator are in the wrong typeface in the index of library names.

tkoeppe:
Page break added before [lex.literal].

Annex C.5: The extraneous whitespace is a problem. We should try to fix that and re-review.

Back matter split view: Fixing separately.

sentinel-for: see below.

tkoeppe:
The sentinel-for awkwardness was fixed with a single, local `\newpage`, no changes outside the two affected pages.

PTAL.

zygoloid (approved):
Thanks, looks great!

PR #4234, opened 2020-09-24
[lex.charset] Replace "is required" wording in footnote with simple statement.

tkoeppe:


source/lex.tex
```diff
@@ -216,10 +216,10 @@
 new-line, plus the following 91 graphical characters:\footnote{The glyphs for
 the members of the basic source character set are intended to
 identify characters from the subset of ISO/IEC 10646 which corresponds to the ASCII
-character set. However, because the mapping from source file characters to the source
+character set. However, the mapping from source file characters to the source
 character set (described in translation phase 1) is specified as
 \impldef{mapping from physical source file characters to basic source character set},
-an implementation is required to document how the basic source characters are
+and hence the implementation documents how the basic source characters are
```
jensmaurer, line 222: Would that be a case where we introduce "must", meaning a logical consequence of normative text? Or did we excise "must" entirely?
The current phrasing reads a bit awkward one way or the other.

tkoeppe: I was thinking that too, but doesn't ISO say that "must" is only for _external_ necessities?

tkoeppe: I applied "must".

burblebee (approved):

PR #4237, opened 2020-09-25
[macros] Avoid page break between example and codeblock introducers

jensmaurer:
if they are immediately following each other.

Partially addresses #4228

### DIFF
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -280,7 +280,7 @@
 \newcommand{\leftshift}[1]{\ensuremath{\mathbin{\mathsf{lshift}_{#1}}}}
 
 %% Notes and examples
-\newcommand{\noteintro}[1]{[\textit{#1}:}
+\newcommand{\noteintro}[1]{[\textit{#1}: }
 \newcommand{\noteoutro}[1]{\textit{\,---\,#1}\kern.5pt]}
 
 % \newnoteenvironment{ENVIRON}{BEGIN TEXT}{END TEXT}
@@ -292,7 +292,7 @@
 \newcommand{\newnoteenvironment}[3]{
 \newsubclausecounter{#1}
 \newenvironment{tail#1}
-{\par\small\stepcounter{#1}\noteintro{#2}}
+{\par\small\stepcounter{#1}\noteintro{#2}\hbox to 0pt{}\ignorespaces}
 {\noteoutro{#3}}
 \newenvironment{#1}
 {\begin{tail#1}}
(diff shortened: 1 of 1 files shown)


jensmaurer:
This destroys the penalty after a heading and thus needs more fine-tuning.

jensmaurer:
Let's not do this for C++20, but it seems viable for C++23.

tkoeppe:
I agree. Thanks a lot, though, this is a very interesting tweak to explore.

source/macros.tex
```diff
@@ -473,8 +478,27 @@
 
 \lstnewenvironment{codeblock}{\CodeBlockSetup}{}
 
-% Left-align listings titles
 \makeatletter
+
+\newsavebox\@signalbox
+
+% Redefine \begin{codeblock} to look for an immediately-preceding
+% \begin{example}.
```
zygoloid, line 486: Instead of redefining it, can we rename the listing environment and define our `codeblock` environment in terms of the listings environment? Or is that not possible due to the way that listings monkeys with the input stream?

jensmaurer: I don't know what you mean. We already have `\lstnewenvironment{codeblock}{\CodeBlockSetup}{}`, above.
The problem is that I can't put the "detect marker" code into \CodeBlockSetup, because that's too late and I can't get at the preceding box anymore. In order to get control earlier, I'm redefining the codeblock environment to get an earlier grab.

zygoloid: What I'm asking is whether we can instead use (say) `\lstnewenvironment{codeblockimpl}....` and then define a new `codeblock` environment in terms of `codeblockimpl`. Modifying an existing environment like this, rather than composing this environment out of the existing one, is a hack that we should avoid if we can. (Maybe we can't; so be it.)

jensmaurer: Tried this:
```
\newenvironment{codeblock}
{%
  \ifhmode%                             no last box in vertical mode
    \setbox\@signalbox=\lastbox%        grab and consume last box
    \copy\@signalbox%                   typeset the last box again
    \ifhbox\@signalbox%
      \ifdim\wd\@signalbox=0pt%         zero-width box is our marker
        \def\codeblockbeginpenalty{100}%
      \fi%
    \fi%
  \fi%
\begin{codeblockimpl}}
{\end{codeblockimpl}}
```

Result:

! Extra }, or forgotten \endgroup.
<recently read> \egroup 
                        
l.228 ...( ) < > % : ; . ? * + - / ^ & | ~ ! = , @
                                                  \textbackslash@ " '


So, the environment nesting appears to negatively impact @ recognition. No idea why.

zygoloid:
@godbyk Can you take a look at this?

I don't understand LaTeX well enough to know if this is reasonable.

godbyk:
@zygoloid Yeah, I can take a look. Do you have a couple examples of bad page breaks I can use for testing?

jensmaurer:
@tkoeppe, this does fix a few situations where we have a lone "[ Example 1:" at the bottom of the preceding page. Those introducers are moved to the next page. You could argue that the mechanics is too hacky, but we should decide whether we want to go forward here or not.

AlisdairM:
This sounds useful and hopefully something we can still work toward landing.  Is it something to consider for the editors' meeting at the coming Hagenberg meeting?

Issue #4236, opened 2020-09-25
[class.mem.general] Awkward punctuation around note in list item


Eelis:
![image](https://user-images.githubusercontent.com/22357/94317083-d6574c00-ff85-11ea-9177-da16d427139f.png)

tkoeppe:
Thank you, we'll look at this during the final review. Looks like something we need to rearrange.

godbyk:
I created PR #4241 to address this issue. It makes use of the new `note*` environment in three places (including the one mentioned above).

PR #4241, opened 2020-09-27
Add new note* and example* environments to suppress paragraph break.

godbyk:
The new `note*` and `example*` environments will suppress the trailing paragraph break. This is useful when the environment is followed by a punctuation mark.

This PR addresses issue #4236.

### DIFF
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -275,12 +275,17 @@
 % ending with END TEXT. A counter with name ENVIRON indicates the
 % number of this kind of note / example that has occurred in this
 % subclause.
-\newcommand{\newnoteenvironment}[3]{
-\newsubclausecounter{#1}
-\newenvironment{#1}
-{\def\noteend{#3}\par\small\stepcounter{#1}\noteintro{#2}}
-{\noteoutro{\noteend}\par}
+\ExplSyntaxOn
+\NewDocumentCommand \newnoteenvironment { m m m } {
+  \newsubclausecounter{#1}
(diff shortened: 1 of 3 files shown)


source/macros.tex
```diff
@@ -275,12 +275,17 @@
...
-{\noteoutro{\noteend}\par}
+\ExplSyntaxOn
+\NewDocumentCommand \newnoteenvironment { m m m } {
+  \newsubclausecounter{#1}
+  \NewDocumentEnvironment {#1*} { }
+    {\def\noteend{#3}\par\small\stepcounter{#1}\noteintro{#2}}
+    {\noteoutro{\noteend}}
+  \NewDocumentEnvironment {#1} { }
+    {\begin{note*}}
+    {\end{note*}\par}
```
timsong-cpp, line 286: Should these be something like `#1*`?

jensmaurer (commented):
In general, we should move notes to outside of (even long) sentences. Regardless of punctuation and paragraph breaks, notes don't belong inside normative sentences. We should really try hard to move the notes (if all else fails, we might put the material into a footnote).

source/macros.tex
godbyk: It absolutely should! I've fixed this. Thanks!

jensmaurer:
superseded by #4249

godbyk:
@jensmaurer You might evaluate the other two cases I found as well to see if they should be corrected in the same fashion.

jensmaurer:
We settled on "tailnote" and "tailexample".

PR #4249, opened 2020-09-29
[std] Introduce tailnote environment to avoid empty lines in tables.

jensmaurer:
Partially addresses #4228.

This should also be applied to master.


burblebee (approved):
Can we pick a more meaningful name than note*?  otherwise looks good

jensmaurer:
This is for c++20 and for "master".

jensmaurer:
Cherry-picked and merged to master.

PR #4252, opened 2020-09-30
Page breaks (burblebee)

zygoloid:
Partially addresses #4228

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -1771,6 +1771,9 @@ a function declaration\iref{dcl.fct}.
 The formation of the associated constraints
 establishes the order in which constraints are instantiated when checking
 for satisfaction\iref{temp.constr.constr}.
+
+\newpage
+
 \begin{example}
 \begin{codeblock}
 template<typename T> concept C = true;
@@ -5273,6 +5276,7 @@ template<class T> class A {
     A<T>::type i;               // OK: refers to a member of the current instantiation
     typename A<T>::other j;     // error: neither a member of the current instantiation nor
                                 // a member of an unknown specialization
+@\newpage@
   }
 };
 \end{codeblock}
@@ -5749,6 +5753,8 @@ external linkage introduced in those namespaces in all translation units,
 not just considering those declarations found in the template definition and
 template instantiation contexts, then the program has undefined behavior.
 
+\newpage
+
 \pnum
 \begin{example}
 \begin{codeblocktu}{Source file \tcode{"X.h"}}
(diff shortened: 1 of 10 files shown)


burblebee:
These changes should also be applied to master:
* 7130f45 [tab:meta.trans.other] Remove blank line before Preconditions.
* 4bc3589 [tab:container.hash.req] Add a line break before next itemdecl clause.
 * 4d1baf4 [expr.const] Replace notes in sentences with footnotes.

zygoloid:
Please remove the "partially addresses" note from the individual commit messages. We want that on the pull request, not in the commit messages.

burblebee:
> Please remove the "partially addresses" note from the individual commit messages. We want that on the pull request, not in the commit messages.

done.

zygoloid:
[temp.dep.type]p9 has a mid-sentence note.

Consider adding a page break before [cpp.replace.general]p10

Please see if you can move the declarations before [new.delete.single]p10 all onto the same page without reflowing that page -- it looks like there might be enough white space at the bottom for that. If it doesn't fit, try putting two of the declarations on each page instead.

[string.view.synop] would probably benefit from a page break before. The `#include` looks lonely.

There's a spurious line break in table [tab:container.seq.opt] in the description of `emplace_back`, before "*Returns*:".

Otherwise, looks good. Thanks!

burblebee:
I fixed all the issues you raised above.

> There's a spurious line break in table [tab:container.seq.opt] in the description of `emplace_back`, before "_Returns_:".

I can't explain this one - looks like a bug in Latex, triggered when a sentence ends at the column boundary and is followed by `\br`. It could trip us up elsewhere, so worth looking into.  I went ahead and removed the \br from the previous itemdecl clause to fix.

source/templates.tex
```diff
@@ -5331,11 +5335,9 @@
...
 template arguments is a dependent type or an expression that is type-dependent
-or value-dependent or is a pack expansion
-\begin{note}
-This includes an injected-class-name\iref{class.pre} of a class template
-used without a \grammarterm{template-argument-list}.
-\end{note}%
+or value-dependent or is a pack expansion%
+\footnote{This includes an injected-class-name\iref{class.pre} of a class template
+used without a \grammarterm{template-argument-list}.}
 , or
```
zygoloid, line 5341: Please move the footnote after the punctuation.

burblebee: interesting, didn't know that footnotes should follow punctuation like this, will try to remember for future.

done.

zygoloid:
Thanks, looks great! (I can't actually approve because I created this PR, but... approved!)

PR #4261, opened 2020-10-01
[time.clock.utc.overview] Move explicit line break in example.

### LINKED ISSUE #4257: [time.clock.utc.overview] example can probably be line-wrapped better
[time.clock.utc.overview]/1's example was manually line-broken before "is 946'684'822s". Now that notes use a smaller typeface, it looks like this would fit on a single line, wrapping instead before "which".

jensmaurer:
Fixes #4257


jensmaurer:
There is no change to pagination, so this can be readily applied to c++20 and master.

tkoeppe (approved):

PR #4262, opened 2020-10-01
Improved docs

sakshipatidar:


jwakely:
Why do you consider this an improvement?

The punctuation is incorrect (there should be no space before the colon). I don't know what "Standard draft sources" you're referring to, isn't that what the entire project contains? Why does the README for the Standard draft sources need to say that it mentions the project? Isn't that obvious?

jwakely (commented):
I don't think this is an improvement.

PR #4263, opened 2020-10-01
[move.iter.elem] Add missing 'return'

pppalka:


CaseyCarter (approved):

jwakely (approved):

jensmaurer (approved):

tkoeppe:
@jensmaurer: shouldn't we take this for C++20, too?

jwakely:
It's unlikely any implementation will do the wrong thing, or any users misunderstand the spec, but it would be nice to fix in C++20.

PR #4271, opened 2020-10-02
[tab:container.hash.req] Capitalize start of sentence.

### LINKED ISSUE #4269: [tab:container.hash.req] missing capitalization in description of node_type
[tab:container.hash.req]'s "Return type" for `X::node_type` is a full sentence but starts with a lowercase letter, unlike the other table cells nearby.

jensmaurer:
Fixes #4269

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -1728,7 +1728,7 @@ and \tcode{nh} denotes a non-const rvalue of type \tcode{X::node_type}.
 
 \indexordmem{node_type}%
 \tcode{X::node_type} &
- a specialization of a \exposid{node-handle}
+ A specialization of a \exposid{node-handle}
  class template, such that the public nested types are
  the same types as the corresponding types in \tcode{X}. &
  see~\ref{container.node} &


source/containers.tex
```diff
@@ -1728,7 +1728,7 @@
 
 \indexordmem{node_type}%
 \tcode{X::node_type} &
- a specialization of a \exposid{node-handle}
+ A specialization of a \exposid{node-handle}
```
CaseyCarter, line 1731: This is technically not a sentence - there's no verb. Ditto for the two cells above. To me, it would be more consistent with e.g. the `Hash` and `Pred` entries to remove the lowercase the two cells above this and remove the period at the end of all three.

zygoloid: This is also true for our "*Returns*:" elements, but we format those as sentences (with a leading capital letter and a trailing period).

CaseyCarter: It's fairly clear in that case that "*Returns*: Forty-two." is a stylized form of the sentence "Returns forty-two." Here, the words "[The] Return type [is]" needed to make this a sentence are much further away, and the majority of the cells don't include periods. The notion that the contents of the table cells are portions of larger sentences is a bit of a stretch.

That said, I don't find this usage particularly offensive. I'll let this one go so we can keep editorial focus on bigger issues.

PR #4272, opened 2020-10-02
[tab:container.assoc.req] Add missing 'Effects'

### LINKED ISSUE #4268: [tab:container.assoc.req] missing an "Effects:"
[tab:container.assoc.req]'s entry for `emplace_hint` appears to be missing an "*Effects:*" before "equivalent to ...". Likewise for `insert`.

jensmaurer:
Fixes #4268


jensmaurer:
Pagination is not changed by this fix.

zygoloid:
#4268 says "likewise for `insert`", referring to the `a.insert(il)` case later in the table that also has this problem.

jensmaurer:
Fixed + force-pushed.

PR #4275, opened 2020-10-03
C++20 tweaks

burblebee:
Build fix:
* 30353134 [tab:container.alloc.req] Indent clauses to fix the build.

Optional fixes:
* 727aff3e [locale.codecvt] Use 2 space indentation (but leave inconsistent indentation for parameters in synopsis as is).
* 6ab88976 [time.clock.cast.fn] Use 2 space indentation in the constraints for clock_cast.

### DIFF
--- a/source/locales.tex
+++ b/source/locales.tex
@@ -1802,18 +1802,18 @@ namespace std {
       explicit codecvt(size_t refs = 0);
 
       result out(
-          stateT& state,
-          const internT* from, const internT* from_end, const internT*& from_next,
-                externT*   to,       externT*   to_end,       externT*&   to_next) const;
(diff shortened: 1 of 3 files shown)


source/locales.tex
```diff
@@ -1802,18 +1802,18 @@
       explicit codecvt(size_t refs = 0);
 
       result out(
-          stateT& state,
-          const internT* from, const internT* from_end, const internT*& from_next,
-                externT*   to,       externT*   to_end,       externT*&   to_next) const;
```
burblebee, line 1807: @zygoloid Should we fix the parameters as well?  I would, but the current indentation does make the differences between these functions more obvious.

PR #4276, opened 2020-10-03
[dcl.fct.def.coroutine] Replace "variable" with "object"

languagelawyer:


jensmaurer (approved):

jensmaurer (changes requested):
What if T is a reference?

languagelawyer:
Hm, `T` indeed can be a reference type, so I think it is not possible to reword this paragraph separately

languagelawyer:
Hm. [basic.stc]/2: ... automatic storage durations are associated with objects introduced by declarations ([basic.def]) and implicitly created by the implementation (temporary objects).
[dcl.fct.def.coroutine]/13 says that the copies are the objects with automatic storage duration, but doesn't say that they are introduced by a declaration and I doubt that they are temporary objects. So [basic.stc]/2 misses this case.

Given that [it is described how a coroutine's _function-body_ is replaced](https://timsong-cpp.github.io/cppwp/n4861/dcl.fct.def.coroutine#5), maybe we copy parameters syntactically, by injecting variable declarations initialized from parameters before _promise-type `promise` promise-constructor-arguments `;`_ and replacing _id-expressions_ naming parameters with _id-expressions_ naming the corresponding variables in the _function-body_? This way:

- there will be no need to fix [basic.stc]/2, because objects will be introduced by declarations
- it will be possible to remove this wording from  [dcl.fct.def.coroutine]/13:
>The initialization and destruction of each parameter copy occurs in the context of the called coroutine. Initializations of parameter copies are sequenced before the call to the coroutine promise constructor and indeterminately sequenced with respect to each other. The lifetime of parameter copies ends immediately after the lifetime of the coroutine promise object ends.

zygoloid:
I think we should leave this alone for now, and wait for it to be handled by CWG's split of "object" and "variable".

PR #4287, opened 2020-10-06
[dcl.fct] Clarify which declarations we're talking about in example.

burblebee:
This changes "These declarations" to "The declarations above" inn example so it's clear which declarations we're talking about.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -3755,16 +3755,14 @@ void g2(C1 auto&...);
 void g3(C3 auto...);
 void g4(C3 auto);
 \end{codeblock}
-
-These declarations are functionally equivalent (but not equivalent) to
-the following declarations.
+The declarations above are functionally equivalent (but not equivalent) to
+their respective declarations below:
 \begin{codeblock}
 template<C1 T, C2 U> void g1(const T*, U&);
 template<C1... Ts>   void g2(Ts&...);
 template<C3... Ts>   void g3(Ts...);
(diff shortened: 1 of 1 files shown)


burblebee:
@tkoeppe I think this worked, yes?  The issue was I had to use https://github.com/burblebee/draft.git instead of git@github.com:burblebee/draft.git because my username locally is different from github's.

tkoeppe:
@burblebee: yes, looks like it worked!

source/declarations.tex
```diff
@@ -3661,8 +3661,8 @@
 void g4(C3 auto);
 \end{codeblock}
```
tkoeppe, line 3663: Maybe also delete these blank lines? We aren't really breaking any paragraphs here. (Also on line 3672.)

tkoeppe (changes requested):
OK!

source/declarations.tex
burblebee: fixed, thanks!

tkoeppe (approved):

jensmaurer:
Add-on improvement suggestion: We should somehow say that g1 is compared against g1 etc. Maybe add "corresponding" or "respective" somewhere. As it stands, it might be misread as-if all of the declarations are mutually functionally equivalent, but they're not.

jensmaurer:
Apparently, the Project Editor supports my change suggestion. Please augment this pull request accordingly.

jensmaurer (approved):

PR #4288, opened 2020-10-06
[intro.defs] Integrate [re.def].

### LINKED ISSUE #4283: merge [re.def] into Clause 3
We missed another "Terms and Definitions"-like subclause when merging together all such sections. We should merge [re.def] into Clause 3.

jensmaurer:
Also rephrase the regular expression definitions to fit
the ISO-mandate style.

Fixes #4283


jensmaurer:
Fixed + force-pushed.

tkoeppe (approved):

PR #4289, opened 2020-10-06
[stmt.ranged] Align font for begin-expr and end-expr.

### LINKED ISSUE #4285: [stmt.ranged] Inconsistent placeholder formatting
image

jensmaurer:
Those are placeholders for expressions.

Fixes #4285

### DIFF
--- a/source/statements.tex
+++ b/source/statements.tex
@@ -639,8 +639,8 @@ is equivalent to
 \terminal{\{}\br
 \bnfindent \opt{init-statement}\br
 \bnfindent \keyword{auto} \terminal{\&\&}\exposid{range} \terminal{=} for-range-initializer \terminal{;}\br
-\bnfindent \keyword{auto} \exposid{begin} \terminal{=} begin-expr \terminal{;}\br
-\bnfindent \keyword{auto} \exposid{end} \terminal{=} end-expr \terminal{;}\br
+\bnfindent \keyword{auto} \exposid{begin} \terminal{=} \exposid{begin-expr} \terminal{;}\br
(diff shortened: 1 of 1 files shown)


zygoloid:
@tkoeppe I'd like your opinion on the cacophony of fonts in use here. We have upright text, slanted grammar terms, italicized placeholders, teletype code, and slanted teletype exposition-only IDs, in a short space. And the teletype font have rather different font weight when used in the codeblock and when used in running text (eg, in bullets (1.2) and (1.3.2)).

The use of `grammarterm` for the `begin-expr` and `end-expr` is certainly wrong, but I wonder if there's more here that we could improve.

tkoeppe:
I'll take a look. I had similar thoughts when the issue was first raised.

tkoeppe:
(Tangent: I think the font weight of the slanted teletype is always quite big, even in the codeblock. There's just less roman font around for comparison. We can look into that separately.)

tkoeppe:
I wouldn't be opposed to more "slanted teletype" for these expression placeholders. Nothing says we can use that style only for placeholder variables, and indeed we areadly happily use it for placeholder types and concepts I believe.

tkoeppe (changes requested):
In my opinion we could use `\exposid` everywhere for the two expr-placeholders and it wouldn't look all that confusing. We already use italic-teletype for a variety of placeholders (concepts, namespaces, types), so I don't have a problem with also using it for expressions.

tkoeppe:
@jensmaurer: Looks like something went wrong with the push, could you check again?

jensmaurer:
@tkoeppe , should be better now.

tkoeppe (approved):

PR #4291, opened 2020-10-06
[macros] Prefer page break after note or example

jensmaurer:


zygoloid:
Can you provide a representative before/after screenshot to show what difference this makes?

jensmaurer:
Here are two diffpdf screenshots:

note-1
note-2

PR #4293, opened 2020-10-07
[layout] Make "small" line spacing equal to normal spacing.

tkoeppe:
Now "normal" is 10pt/12pt (as it has always been) and "small" is
9pt/12pt. This makes notes and examples use the same line spacing as
the ambient text.

image

Left: new, right: before.

### DIFF
--- a/source/layout.tex
+++ b/source/layout.tex
@@ -32,6 +32,12 @@
 \feetatbottom
 
 %%--------------------------------------------------
+%% Make the "small" size, used for notes and examples, have the same
+%% line height as the normal text.
+\renewcommand{\normalsize}{\fontsize{10pt}{12pt}\selectfont}
+\renewcommand{\small}{\fontsize{9pt}{12pt}\selectfont}
+
+%%--------------------------------------------------
 %% Paragraph and bullet numbering
 
 % create a new counter that resets for each new subclause


jensmaurer:
Hm... I think I like that notes are smaller in all directions. I think they should be separate from the next paragraph by full-size vertical space, but the internal spacing should be smaller so that the text sticks together naturally.

burblebee:
With 9/12 font for notes, the note by itself it doesn't look bad, but when seen in context with regular text, the proportionally greater vertical space stands out. I much prefer the 9/10.5 spacing.

tkoeppe:
Try to consider this in the wider context. With the smaller line height, the overall vertical rhythm is more uneven as you go into and out of notes, which makes the page look wobbly and tentative. By contrast, with the fixed spacing everything looks regular and calm and even.

That's certainly fairly subjective, but I'd like to encourage you to look at this again with that in mind. That is, don't just look at an individual note, but see how the whole document looks as you leaf through page after page.

tkoeppe:
For an example, try out the page with the [class.prop] heading. It has two notes just before the heading. I find the "evenly spaced" style quite nice here, and somehow "less surprising" than the narrower style of the status quo. Likewise for notes in the middle of lists, e.g. [over.match.oper]b(3.4).

tkoeppe:
Here's a nice mix of notes and normative text:

image

zygoloid:
This change makes notes and examples less visually distinct. We should decide whether or not that's a goal.

Our friends in WG14 use smaller inter-line spacing within notes and examples (and, presumably uncontroversially, both documents use less inter-line spacing in footnotes).

jensmaurer:
Personally, I think the smaller inter-line spacing is preferable, exactly because it sets notes and examples apart visually from the normative text, and a smaller inter-line spacing does fit the smaller font.

tkoeppe:
Would you perhaps be willing to do a little experiment and use this patch for a couple of days of "regular work" with the Standard, where you read it as a normal user, not an editorial reviewer?

I wonder if there's just the familiarity with the status quo, or the particular focus on reworking notes, that makes it so desirable to distinguish notes from non-notes in so many ways, and whether this is really something a regular reader would want. Also please consider whether you would have asked for the spacing to be made narrower if this even spacing had always been like this since the beginning.

PR #4294, opened 2020-10-08
[rand.eng.sub] drop no-op modulo operation

jwakely:
The `e()` mod 2<sup>32</sup> operation is a no-op, because `e.max()` < 2<sup>32</sup>.

When I raised this on the LWG reflector @CaseyCarter agreed that this is the case. We should simplify the spec by removing the noise.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -3419,7 +3419,7 @@ linear_congruential_engine<result_type,
  Then, to set each $X_k$,
  obtain new values $z_0, \dotsc, z_{n-1}$
  from $n = \lceil w/32 \rceil$ successive invocations
- of \tcode{e} taken modulo $2^{32}$.
+ of \tcode{e}.
  Set $X_k$ to $\left( \sum_{j=0}^{n-1} z_j \cdot 2^{32j}\right) \bmod m$.
 
 \pnum


tkoeppe:
This seems legit, though I would now start to complain that *k* is not defined in this paragraph. We should say something like "for *k* = 0, ..., 8" etc.

zygoloid:
The existing wording could be read as saying that we take ceil(w/32) modulo 2^32. I assume that's not the intent?

Regarding @tkoeppe's comment, we say we're "Set[ting] the values *X<sub>-r</sub>*, ..., *X<sub>-1</sub>*, in that order, as specified below" and then below say "To set the values *X<sub>k</sub>*, ...", so I think based on that it's sufficiently clear that *k* takes the values *-r*,...,-1, in that order.

While we're here, though, the wording talks about *r*, *s*, and *w*, which are not defined (though `r`, `s`, and `w` are). If we mean the values of the template parameters, we should use code font to refer to them, or we should define the math variables as the values of the template parameters.

tkoeppe:
Regarding the non-defined maths-variables, I raised that elsewhere and @jwakely has a proposal for blanket wording somewhere.

Regarding undefined indexes: I'm OK with a bound dummy index without introduction, e.g. `\sum_i x_i` would be fine without a range. But I'm a lot more nervous to introduce a _free_, dangling variable like that.

tkoeppe:
The blanket wording is the subject of #4295.

jensmaurer (approved):
Let's apply this as-is as a clear improvement / fix.

PR #4295, opened 2020-10-08
[rand.eng], [rand.adapt] Use maths to describe relations between constants

### LINKED ISSUE #4292: [rand.eng.mers] should use math font for (1u<<w)-1u
In [rand.eng.mers] p4 the expression `(1u<<w) - 1u` is used repeatedly. If `w==numeric_limits<unsigned>::digits` that is ill-formed.

In the synopsis above we use <code>{ return 2<sup>_w_</sup> - 1; }</code> despite that being actual C++ code, so why can't we do the same in the prose that follows it?

i.e. replace the code font expressions like  `a <= (1u<<w) - 1u` with `a` &le; 2<sup>_w_</sup> - 1 to be clear that the rules of C++ unsigned integer arithmetic do not apply.

jwakely:
Fixes #4292 

I know that "unless otherwise specified" is strictly unnecessary in the blanket wording, but I think it helps here. In [rand.eng.lcong] we have:

>  If the template parameter `m` is _0_, the modulus _m_ used throughout this subclause (...) is [...]

That specifies _m_ for the `m==0` case, but the blanket wording applies for all other values of `m`.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -3197,10 +3197,8 @@ namespace std {
 \end{codeblock}
 
 \pnum
-The following relations shall hold:
-  \tcode{0 < r}
-and
-  \tcode{r <= p}.
+The following relation shall hold:
+  $0 < r \leq p$.
 
 \pnum
 The textual representation
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -2987,11 +2987,9 @@ namespace std {
 
 \pnum
 The following relations shall hold:
-  \tcode{0u < s},
-  \tcode{s < r},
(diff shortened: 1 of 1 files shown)


tkoeppe:
@jwakely, @CaseyCarter, @W-E-Brown: I would love to land this change. It would leave a lot less to the imagination and make it clear how the code variables relate to the maths variables. Any final thoughts?

jwakely:
I still like this change. It might be even better if we used something like _D_ for `numeric_limits<result_type>::digits` so that we didn't have that C++ expression amongst the math relations, but that could be done separately.

PR #4300, opened 2020-10-20
[ostream.manip] Introduce SYNCBUF to detect basic_syncbuf

jwakely:
This attempts to address the problem of `Allocator` being mentioned in
the function descriptions without being defined. We cannot say that

> `os.rdbuf()` is a `basic_syncbuf<charT, traits, Allocator>*`

because firstly, that type is not defined, and secondly `os.rdbuf()` is
a `basic_streambuf<charT, traits>*` and not any other type.

By introducing SYNCBUF we can define the manipulators properly, by
talking about a base class subobject rather than "is a".

This introduces an apparently normative change that the syncbuf type
must not use a program-defined specialization. Without that additional
restriction the implementation suggested by the note doesn't work,
because program-defined specializations cannot derive from the
intermediate base class, and therefore cannot be detected by SYNCBUF.

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -6771,6 +6771,24 @@ Each instantiation of any of the function templates
 specified in this subclause
 is a designated addressable function\iref{namespace.std}.
 
+\pnum
+In this subclause,
+\exposid{SYNCBUF}\tcode{(p)} for a pointer \tcode{p} of type \tcode{B*}
+is determined as follows.
+If \tcode{*p} is a base class subobject of an object of type \tcode{S},
+where \tcode{S} is a specialization generated from
+the \tcode{basic_syncbuf} primary template,
+and \tcode{is_convertible_v<S*, B*>} is \tcode{true},
+then \exposid{SYNCBUF}\tcode{(p)} is \tcode{dynamic_cast<S*>(p)}.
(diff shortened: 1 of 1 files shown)


jwakely:
CI fails because of overfull hboxes

Screenshot_20201020_135518

jensmaurer:
Say "Let p be SYNCBUF(os.rdbuf())." before \effects, then rephrase \effects in terms of "p":
```
\effects
If p is not null, calls p->set_emit_on_sync(false);
```
This should take care of the overfull hboxes, and reduces redundancy in the exposition.

In any case, this seems not-obviously-correct, so I'd prefer a minuted LWG teleconference consensus that this editorial fix is desirable. (We probably don't need an LWG issue if there is no effective normative change and there are no objections in LWG.)

jwakely:
Yes, that works and looks good. Thanks for the suggestion.

> In any case, this seems not-obviously-correct, so I'd prefer a minuted LWG teleconference consensus that this editorial fix is desirable. (We probably don't need an LWG issue if there is no effective normative change and there are no objections in LWG.)

Yes, I'd prefer that too. That's why this is a draft and not submitted for review yet.

jwakely:
LWG has decided this needs a DR, so isn't editorial.

Dani-Hub:
A corresponding LWG issue has been created, this is [LWG 3501](https://cplusplus.github.io/LWG/issue3501).

PR #4313, opened 2020-11-03
Fix list item punctuation

### LINKED ISSUE #4044: [class.access.general] Sentence has three full stops (bad list item punctuation)
[[class.access]/1](http://eel.is/c++draft/class.access#1) reads:

> A member of a class can be
> - private; that is, its name can be used only by members and friends of the class in which it is declared.
> - protected; that is, its name can be used only by members and friends of the class in which it is declared, by classes derived from that class, and by their friends (see [class.protected]).
> - public; that is, its name can be used anywhere without access restriction.

jensmaurer:
Fixes #4044

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -4076,18 +4076,18 @@ A member of a class can be
 \begin{itemize}
 \item
 \indextext{access control!\idxcode{private}}%
-private;
+private,
 that is, it can be named only by members and friends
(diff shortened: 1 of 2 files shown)


tkoeppe (approved):

jensmaurer:
Rebased and force-pushed.

PR #4315, opened 2020-11-06
[dcl.type.auto.deduct] Clarify initializer for placeholder type deduction.

### LINKED ISSUE #4088: [dcl.type.auto.deduct] Incorrect description of deduction for a variable with copy-initialization?
[[dcl.type.auto.deduct]/2](http://eel.is/c++draft/dcl.type.auto.deduct#2) says:
"""
A type `T` containing a placeholder type, and a corresponding initializer _E_, are determined as follows: 
­— for a variable declared with a type that contains a placeholder type, `T` is the declared type of the variable and _E_ is the initializer. ...
"""

[[dcl.type.auto.deduct]/4](http://eel.is/c++draft/dcl.type.auto.deduct#4) says:
"""
If the _placeholder-type-specifier_ is of the form _type-constraint_<sub>opt</sub> `auto`, the deduced type `T` replacing `T` is determined using the rules for template argument deduction.
Obtain `P` from `T` by replacing the occurrences of _type-constraint_<sub>opt</sub> `auto` either with a new invented type template parameter `U`, or ... [doesn't matter].
Deduce **a value** (!!!) for `U` using the rules of template argument deduction from a function call, where `P` is a function template parameter type and the corresponding argument is _E_.
...
[Example:
```
const auto &i = expr;
```
The type of `i` is the deduced type of the parameter `u` in the call `f(expr)` of the following invented function template:
```
template <class U> void f(const U& u);
```
— end example]
"""

There are several issues here (both in normative wording and the Example): the major one is that for a variable declared as `const auto &i = expr;`, the initializer is `= expr`, not `expr`. And one can't call a function with this initializer as an argument, like `f(= expr)`. ~~Probably, the paragraph should say that the initializer _E_ is the **default** argument of an invented function template (and the type is deduces as if the function is called without argument)?~~ Nah, the grammar for _parameter-declaration_ doesn't allow _initializer_.

Also, is it ok that the Standard calls the deduced type "a **value** of a type parameter" and uses the instructing tone like _Obtain `P` from `T` ..._ _Deduce a value ..._.

jensmaurer:
Fixes #4088

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -1718,8 +1718,6 @@ shall be followed by one or more
 each of which shall
 be followed by a non-empty
 \grammarterm{initializer}.
-If the \grammarterm{initializer} is a parenthesized \grammarterm{expression-list},
-the \grammarterm{expression-list} shall be a single \grammarterm{assignment-expression}.
 \begin{example}
 \begin{codeblock}
 auto x = 5;                     // OK: \tcode{x} has type \tcode{int}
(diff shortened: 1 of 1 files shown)


tkoeppe:
Looks reasonable; could @zygoloid perhaps confirm that this is editorial?

source/declarations.tex
```diff
@@ -1923,25 +1923,28 @@
 are determined as follows:
```
zygoloid, line 1923: On the line above, do we want `E` to be a corresponding *initializer-clause* rather than a corresponding initializer?

source/declarations.tex
```diff
@@ -1923,25 +1923,28 @@
...
 \begin{itemize}
 \item
-for a non-discarded \tcode{return} statement that occurs
+For a non-discarded \tcode{return} statement that occurs
 in a function declared with a return type
 that contains a placeholder type,
 \tcode{T} is the declared return type
 and $E$ is the operand of the \tcode{return} statement.
```
zygoloid, line 1930: This doesn't work either, if we assume a syntactic translation: `return a, b;` doesn't produce an *initializer-clause* that can be the argument in a function call.

source/declarations.tex
jensmaurer: I guess so: An assignment-expression, for example, is an initializer-clause.

source/declarations.tex
jensmaurer: fixed

source/declarations.tex
```diff
@@ -1896,29 +1896,35 @@
...
-for a non-discarded \tcode{return} statement that occurs
+For a non-discarded \tcode{return} statement that occurs
 in a function declared with a return type
 that contains a placeholder type,
 \tcode{T} is the declared return type
-and $E$ is the operand of the \tcode{return} statement.
+and $E$ is the operand of the \tcode{return} statement;
+if the operand is an \grammarterm{expression} $X$
+that is not an \grammarterm{assignment-expression}\iref{expr.comma},
+$E$ is \tcode{($X$)}.
```
zygoloid, line 1910: Ugh, this is awkward. This does the wrong thing for:
```
decltype(auto) f(int n) { return n; }
```
... where we should deduce `int`, but if `E` is `(n)` then we'll deduce `int&` below.

source/declarations.tex
```diff
@@ -1896,29 +1896,35 @@
...
+\tcode{T} is the declared type of the variable.
+If the initializer of the variable is a \grammarterm{brace-or-equal-initializer}
+of the form \tcode{= \grammarterm{initializer-clause}},
+$E$ is the \grammarterm{initializer-clause}.
 If the initialization is direct-list-initialization,
 the initializer shall be a \grammarterm{braced-init-list}
 containing only a single \grammarterm{assignment-expression}
-and $E$ is the \grammarterm{assignment-expression};
+and $E$ is the \grammarterm{assignment-expression}.
+Otherwise, $E$ is the initializer.
```
zygoloid, line 1925: There's more `decltype(auto)` awkwardness here; I think this does the wrong thing for:
```
int a;
decltype(auto) b(a);
```
... which I think should deduce the type of `b` as `int`, but if we take `E` as being `(a)` then we presumably reinterpret it as a parenthesized expression, resulting in the type of `b` being `int&`.

zygoloid: Would it make sense to move [dcl.spec.auto.general]/4's "If the initializer is a parenthesized expression-list, the expression-list shall be a single assignment-expression." here, and then modify it to match the treatment of the braced-init-list case?

source/declarations.tex
jensmaurer: "n" is (indirectly) an assignment-expression.

source/declarations.tex
jensmaurer: Yes, that makes a lot of sense.

I've also integrated the two sentences about constraints on return type deduction that followed the bulleted list. It all looks clean and sane now, except for the wrinkle about forming a parenthesized expression when we start out with a comma expression. This has now been highlighted in a note.

jensmaurer:
@zygoloid, I've now applied a larger hammer, which rounded a few more of the edges here.

source/declarations.tex
tkoeppe: @zygoloid: does that address your concerns? Is this still too subtle?

source/declarations.tex
tkoeppe: @zygoloid: PTAL

zygoloid (approved):
Looks good to me, thanks!

tkoeppe (approved):

PR #4318, opened 2020-11-07
 [time.clock.gps.members]/2 return type typo

MattStephanson:
The returns specification for `gps_clock::to_utc` says `gps_time` instead of `utc_time`, probably a copy-paste typo from p3. The original proposal, [P0355R7](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0355r7.html), also says `gps_time`, but the intent seems obvious, and the status quo doesn't compile.

### DIFF
--- a/source/time.tex
+++ b/source/time.tex
@@ -3233,7 +3233,7 @@ template<class Duration>
 \pnum
 \returns
 \begin{codeblock}
-gps_time<common_type_t<Duration, seconds>>{t.time_since_epoch()} + 315964809s
+utc_time<common_type_t<Duration, seconds>>{t.time_since_epoch()} + 315964809s
 \end{codeblock}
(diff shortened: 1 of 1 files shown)


jensmaurer (approved):

StephanTLavavej (approved):

MattStephanson:
@tkoeppe I was just wondering, since this is my first PR here, if there's anything else I need to do to get it merged? Thanks.

jensmaurer:
@MattStephanson , there are no changes requested from you. However,
- we do require (informally) two approving reviews for pull requests from outsiders and
- we were busy with merging a very large paper.
Sorry for us taking so long.

tkoeppe:
Indeed -- thank you for the fix!

PR #4333, opened 2020-11-10
[syncstream.syncbuf.cons] Remove bogus rdbuf() calls

jwakely:
You don't call rdbuf() to get to the streambuf, this type is the streambuf.


tkoeppe:
@CaseyCarter: Could you kindly take a look?

jensmaurer (approved):

Issue #4319, opened 2020-11-09
could and might


zygoloid:
ISO wants us to avoid "could" and "might" appearing anywhere in the document. TODO list:

- [x] Front matter + Clauses 1, 2, 3 -- no changes
- [x] 4 General principles[intro] -- no changes
- [x] 5 Lexical conventions[lex] -- jmaurer
- [x] 6 Basics[basic] -- jmaurer
- [x] 7 Expressions[expr] -- jmaurer
- [x] 8 Statements[stmt.stmt] -- jmaurer
- [x] 9 Declarations[dcl.dcl] -- jmaurer
- [x] 10 Modules[module] -- no changes
- [x] 11 Classes[class] -- jmaurer
- [x] 12 Overloading[over] -- jmaurer
- [x] 13 Templates[temp] -- jmaurer
- [x] 14 Exception handling[except] -- jmaurer
- [x] 15 Preprocessing directives[cpp] -- jmaurer
- [x] 16 Library introduction[library] -- jmaurer
- [x] 17 Language support library[support] -- jmaurer
- [x] 18 Concepts library[concepts] -- jmaurer
- [x] 19 Diagnostics library[diagnostics] -- jmaurer
- [x] 20 General utilities library[utilities] -- jmaurer
- [x] 21 Strings library[strings] -- jmaurer
- [x] 22 Containers library[containers] -- jmaurer
- [x] 23 Iterators library[iterators] -- jmaurer
- [x] 24 Ranges library[ranges] -- jmaurer
- [x] 25 Algorithms library[algorithms] -- jmaurer
- [x] 26 Numerics library[numerics] -- jmaurer
- [x] 27 Time library[time] -- jmaurer
- [x] 28 Localization library[localization] -- jmaurer
- [x] 29 Input/output library[input.output] -- jmaurer
- [x] 30 Regular expressions library[re] -- jmaurer
- [x] 31 Atomic operations library[atomics] -- jmaurer
- [x] 32 Thread support library[thread] -- jmaurer
- [x] Annex A (informative) Grammar summary[gram] -- no changes
- [x] Annex B (informative normative) Implementation quantities[implimits] -- no changes
- [x] Annex C (informative) Compatibility[diff] -- jmaurer
- [x] Annex D (normative) Compatibility features -- jmaurer
- [x] Back matter -- no changes

Please add your name next to a section to claim it, and check the box when it's done.

johelegp:
[...]

PR #4334, opened 2020-11-10
[utilities] Replace 'could' and 'might'

jensmaurer:
as directed by ISO/CS.

Partially addresses #4319

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -2288,9 +2288,9 @@ is a tuple containing all but the first element of \tcode{r}.
 \pnum
 \begin{note}
 The above definition does not require \tcode{t$_{\mathrm{tail}}$}
-(or \tcode{u$_{\mathrm{tail}}$}) to be constructed. It might not
-even be possible, as \tcode{t} and \tcode{u} are not required to be copy
-constructible. Also, all comparison operator functions are short circuited;
+(or \tcode{u$_{\mathrm{tail}}$}) to be constructed;
+this is impossible if \tcode{t} or \tcode{u} is not copy-constructible.
+Also, all comparison operator functions are short circuited;
 they do not perform element accesses beyond what is required to determine the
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -15121,8 +15122,8 @@
 
 \pnum
 \begin{note}
-The types deduced by the deduction guides for \tcode{function}
-might change in future revisions of \Cpp{}.
+It is possible that the types deduced by the deduction guides for \tcode{function}
+change in future revisions of \Cpp{}.
```
zygoloid, line 15126: I don't think this has the same implications. @jwakely, what about: "The types deduced by the deduction guides for `function` have been identified as candidates for being changed in a future revision of \Cpp{}."? Is that accurate?

jwakely: Yes, I think so.

jensmaurer: fixed

source/utilities.tex
```diff
@@ -4724,7 +4724,8 @@
 
 \pnum
 \begin{note}
-A \tcode{variant} might not hold a value if an exception is thrown during a
+It is possible for a \tcode{variant} not to hold a value
+if an exception is thrown during a
```
jwakely, line 4728: Another awkward "possible not to ..." but I don't have a better suggestion.

jwakely (approved):

source/utilities.tex
jwakely: Ah , in [variant.variant.general] we already say "or holds no value", so if we don't mind using "hold no value" as the opposite of "hold a value", we could say "It is possible for a `variant` to hold no value if ...".

jensmaurer: fixed

jwakely (approved):
Great, thanks!

PR #4336, opened 2020-11-10
[time] Replace 'could' and 'might'

jensmaurer:
as directed by ISO/CS.

Partially addresses #4319

### DIFF
--- a/source/time.tex
+++ b/source/time.tex
@@ -1324,7 +1324,7 @@ No overflow is induced in the conversion and
 \begin{note}
 This
 requirement prevents implicit truncation error when converting between
-integral-based \tcode{duration} types. Such a construction could easily lead to
+integral-based \tcode{duration} types. Such a construction can lead to
 confusion about the value of the \tcode{duration}.
 \end{note}
 \begin{example}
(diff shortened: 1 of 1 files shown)


source/time.tex
```diff
@@ -9267,13 +9267,14 @@
...
 if it were off daylight saving time.
-However, this information should not be taken as authoritative.
-The only sure way to get such information
-is to query the \tcode{time_zone} with a \tcode{time_point}
+A correct offset for the \tcode{time_zone} off daylight saving time
+can be obtained by querying the \tcode{time_zone} with a \tcode{time_point}
 that returns a \tcode{sys_info} where \tcode{save == 0min}.
-There is no guarantee what \tcode{time_point} might return such a \tcode{sys_info}
+There is no guarantee
+what \tcode{time_point} (if any) returns such a \tcode{sys_info}
```
jwakely, line 9277: I find this last sentence hard to understand (before and after this change).

> For any given `time_zone`, it is possible that there is no `time_point` that returns a `sys_info` not on daylight saving time. If such a `time_point` does exist, it is guaranteed not to be in the range \[`begin`,end`) of any `sys_info` on daylight saving time.

I also think "querying the `time_zone` could be more precise, by referring to the `get_info` member.

jensmaurer: I think any such larger rework is for C++23. Added to #4340.

zygoloid (approved):

source/time.tex
zygoloid: @jwakely If you're happy with the suggested approach (this change for C++20, something else for C++23), please can you approve or merge this PR?

jwakely (approved):

PR #4341, opened 2020-11-11
[iterators] Replace 'could' and 'might'

jensmaurer:
as directed by ISO/CS.

Partially addresses #4319

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -2081,7 +2081,7 @@ such an algorithm should be a single-pass algorithm.
 The only valid use of an \tcode{operator*}
 is on the left side of the assignment statement.
 Assignment through the same value of the iterator happens only once.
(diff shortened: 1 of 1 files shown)


source/iterators.tex
```diff
@@ -2081,7 +2081,7 @@
 The only valid use of an \tcode{operator*}
 is on the left side of the assignment statement.
 Assignment through the same value of the iterator happens only once.
-Equality and inequality might not be defined.
+It is possible that equality and inequality are not defined.
```
jwakely, line 2084: I find this a bit clunky.  I think it would read better as "Equality and inequality are not necessarily defined."

jensmaurer: fixed + force-pushed

jwakely (approved):

PR #4346, opened 2020-11-11
[input.output] Replace 'could' and 'might'

jensmaurer:
as directed by ISO/CS.

Partially addresses #4319

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -3659,7 +3659,7 @@ or
 will fail.\footnote{\tcode{underflow}
 or
 \tcode{uflow}
-might fail by throwing an exception prematurely.
+can fail by throwing an exception prematurely.
 The intention is not only that the calls will not return
 \tcode{eof()}
 but that they will return ``immediately''.}
@@ -3895,8 +3895,8 @@ are the same as for
 \returns
 \tcode{traits::eof()}
 to indicate failure.
-Failure may occur because the input sequence could not be backed up, or if for some
-other reason the pointers could not be set consistent with the constraints.
+Failure may occur because the input sequence cannot be backed up, or if for some
+other reason the pointers cannot be set consistent with the constraints.
 \tcode{pbackfail()}
 is called only when put back has really failed.
 
@@ -7774,7 +7774,7 @@ according to \tcode{mode}.
 \pnum
 \begin{note}
 For efficiency reasons,
-stream buffer operations might violate invariants of \tcode{buf}
+stream buffer operations can violate invariants of \tcode{buf}
 while it is held encapsulated in the \tcode{basic_stringbuf},
 e.g., by writing to characters in the range
 \range{\tcode{buf.data() + buf.size()}}{\tcode{buf.data() + buf.capacity()}}.
@@ -11165,7 +11165,7 @@ Initializes the base class with \tcode{basic_ostream<charT, traits>(addressof(sb
 \begin{note}
 The member functions of the provided stream buffer
 can be called from \tcode{emit()} while a lock is held,
-which might result in a deadlock if used incautiously.
+which can result in a deadlock if used incautiously.
 \end{note}
 
 \pnum
@@ -12512,9 +12512,9 @@ Otherwise, modifies \tcode{*this} as if by these steps:
 \pnum
 \begin{example}
 Even if \tcode{//host} is interpreted as a \grammarterm{root-name},
+and even if \tcode{operator/} appends a backslash as the separator,
 both of the paths \tcode{path("//host")/"foo"} and \tcode{path("//host/")/"foo"}
-equal \tcode{"//host/foo"} (although the former might use backslash as the
-preferred separator).
+equal \tcode{"//host/foo"}.
 
 Expression examples:
 \begin{codeblock}
@@ -12586,8 +12586,8 @@ template<class Source>
 \effects
 Appends \tcode{path(x).native()} to the pathname in the native format.
 \begin{note}
-This directly manipulates the value of \tcode{native()}
-and might not be portable between operating systems.
+This directly manipulates the value of \tcode{native()},
+which is not necessarily portable between operating systems.
 \end{note}
 
 \pnum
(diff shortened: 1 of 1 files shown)


source/iostreams.tex
```diff
@@ -16277,8 +16278,8 @@
 
 \pnum
 \begin{note}
-A postcondition of \tcode{last_write_time(p) == new_time} is not specified since it might not hold for file systems
-  with coarse time granularity.
+A postcondition of \tcode{last_write_time(p) == new_time} is not specified
+since it is possible not to hold for file systems with coarse time granularity.
```
jwakely, line 16282: Another "possible not to" phrase, which I find awkward.

Maybe: "No postcondition of `last_write_time(p) == new_time` is specified, because a file system with coarse time granularity cannot represent all values of `file_time_type` and can store a value that is less than `new_time`."

What is a "coarse time granularity" though? It would be conforming for `file_time_type` to use picosecond resolution, in which case the postcondition wouldn't be possible even file systems with nanosecond resolution, which is not "coarse" by most definitions.

Maybe just:

> _Note:_ Sets the greatest value supported by the file system that is not greater than `new_time`. Some values cannot be represented if the file system's timestamps have coarser granularity than `file_time_type`.

This isn't a normative change, because we already say it's done as if by [`futimens`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/futimens.html) which has the "greatest value <= `new_time`" requirement.

source/iostreams.tex
```diff
@@ -12513,7 +12513,7 @@
 \begin{example}
 Even if \tcode{//host} is interpreted as a \grammarterm{root-name},
 both of the paths \tcode{path("//host")/"foo"} and \tcode{path("//host/")/"foo"}
-equal \tcode{"//host/foo"} (although the former might use backslash as the
+equal \tcode{"//host/foo"} (although the former can use backslash as the
```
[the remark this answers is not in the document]
jensmaurer: Yeah, that's better. Fixed.

source/iostreams.tex
```diff
@@ -12586,8 +12586,8 @@
 \effects
 Appends \tcode{path(x).native()} to the pathname in the native format.
 \begin{note}
-This directly manipulates the value of \tcode{native()}
-and might not be portable between operating systems.
+This directly manipulates the value of \tcode{native()};
+it is possibly not portable between operating systems.
```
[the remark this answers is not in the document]
jensmaurer: fixed

source/iostreams.tex
jensmaurer: less intrusive suggestion pushed:
"since it does not necessarily hold for file systems with coarse time granularity"

source/iostreams.tex
```diff
@@ -15201,7 +15202,7 @@
 For POSIX-based operating systems,
 \tcode{absolute(p)} is simply \tcode{current_path()/p}.
 For Windows-based operating systems,
-\tcode{absolute} might have the same semantics as \tcode{GetFullPathNameW}.
+\tcode{absolute} can have the same semantics as \tcode{GetFullPathNameW}.
```
zygoloid, line 15205: I don't think we're talking about ability or capability here; rather, we're talking about possibility. Given https://github.com/microsoft/STL/blob/master/stl/inc/filesystem#L3001, I think we can be so bold as to say:
```suggestion
\tcode{absolute} typically has the same semantics as \tcode{GetFullPathNameW}.
```

source/iostreams.tex
```diff
@@ -15762,7 +15763,7 @@
 \pnum
 \begin{note}
 The current path as returned by many operating systems is a dangerous
-  global variable. It might be changed unexpectedly by third-party or system
+  global variable. It can be changed unexpectedly by third-party or system
```
zygoloid, line 15766: The second sentence isn't true in isolation -- we don't know that it can be changed asynchronously. How about this:
```suggestion
  global variable, and can be changed unexpectedly by third-party or system
```

source/iostreams.tex
```diff
@@ -15789,7 +15790,7 @@
 \pnum
 \begin{note}
 The current path for many operating systems is a dangerous
-  global state. It might be changed unexpectedly by a third-party or system
+  global state. It can be changed unexpectedly by a third-party or system
```
zygoloid, line 15793:
```suggestion
  global state, and can be changed unexpectedly by a third-party or system
```

source/iostreams.tex
```diff
@@ -16277,8 +16278,8 @@
 
 \pnum
 \begin{note}
-A postcondition of \tcode{last_write_time(p) == new_time} is not specified since it might not hold for file systems
-  with coarse time granularity.
+A postcondition of \tcode{last_write_time(p) == new_time} is not specified
+since it does not necessarily hold for file systems with coarse time granularity.
```
zygoloid, line 16282:
```suggestion
because it does not necessarily hold for file systems with coarse time granularity.
```
(Drive-by: use "since" only when dealing with time, and "because" for causality.)

source/iostreams.tex
jensmaurer: fixed

source/iostreams.tex
jensmaurer: fixed

source/iostreams.tex
jensmaurer: fixed

source/iostreams.tex
jensmaurer: fixed

jensmaurer:
All fixed and force-pushed.

source/iostreams.tex
jwakely: Ahem, there are two implementations that support Windows. But [the other one](https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=libstdc%2B%2B-v3/src/c%2B%2B17/fs_ops.cc;h=2cb26e4605b89b9d5ef89dbcde1dcd7d4ec7ef42;hb=HEAD#l118) also uses `GetFullPathNameW` here.

PR #4347, opened 2020-11-11
[re] Replace 'could' and 'might'

jensmaurer:
as directed by ISO/CS.


source/regex.tex
```diff
@@ -869,7 +869,7 @@
 \tcode{error_stack}
 &
 There was insufficient memory to determine whether the regular
-expression could match the specified character sequence.  \\
+expression matched the specified character sequence.  \\
```
jwakely, line 872: Using the past tense here seems wrong. If an exception was thrown, it didn't match. Maybe "matches" or "would match"?

jensmaurer: It's conditional, not past tense.

jensmaurer: But "would match" sounds good.

jensmaurer: fixed + force-pushed

jwakely (approved):

source/regex.tex
```diff
@@ -869,7 +869,7 @@
 \tcode{error_stack}
 &
 There was insufficient memory to determine whether the regular
```
zygoloid, line 871: Please convert all the entries in this table into the present tense. (This is another one of ISO/CS's requests, and in this case I think it's reasonable.) That'd also let us replace "would match" with "matches" here without losing the hypothetical feeling.

jensmaurer: fixed

PR #4348, opened 2020-11-11
[atomics] Replace 'could' and 'might'

jensmaurer:
as directed by ISO/CS.

Partially addresses #4319

### DIFF
--- a/source/atomics.tex
+++ b/source/atomics.tex
@@ -769,7 +769,7 @@ through a referencing \tcode{atomic_ref} are atomic with respect to
 atomic operations applied through any other \tcode{atomic_ref}
 referencing the same object.
 \begin{note}
-Atomic operations or the \tcode{atomic_ref} constructor could acquire
+Atomic operations or the \tcode{atomic_ref} constructor can acquire
 a shared resource, such as a lock associated with the referenced object,
 to enable atomic operations to be applied to the referenced object.
 \end{note}
@@ -791,14 +791,15 @@ which is at least \tcode{alignof(T)}.
 
 \pnum
 \begin{note}
-Hardware could require an object
+Hardware can require an object
 referenced by an \tcode{atomic_ref}
 to have stricter alignment\iref{basic.align}
 than other objects of type \tcode{T}.
 Further, whether operations on an \tcode{atomic_ref}
-are lock-free could depend on the alignment of the referenced object.
-For example, lock-free operations on \tcode{std::complex<double>}
-could be supported only if aligned to \tcode{2*alignof(double)}.
+are lock-free can depend on the alignment of the referenced object.
+For example, it is possible that the hardware supports
+lock-free atomic operations on a \tcode{std::complex<double>} object
+only if the object is aligned to \tcode{2*alignof(double)}.
 \end{note}
 \end{itemdescr}
 
@@ -1607,7 +1608,7 @@ The program is ill-formed if any of
 is \tcode{false}.
 \begin{note}
 Type arguments that are
-not also statically initializable might be difficult to use.
+not also statically initializable can be difficult to use.
 \end{note}
 
 \pnum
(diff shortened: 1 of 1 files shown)


source/atomics.tex
```diff
@@ -1999,7 +2000,7 @@
 \end{note}
 \begin{note}
 For a union with bits that participate in the value representation
-of some members but not others, compare-and-exchange might always fail.
+of some members but not others, compare-and-exchange can always fail.
```
jwakely, line 2003: Maybe "... but not others, it is possible for compare-and-exchange to always fail."  ?

Or "it is possible that a compare-and-exchange operation will never succeed."

source/atomics.tex
```diff
@@ -791,14 +791,15 @@
...
 to have stricter alignment\iref{basic.align}
 than other objects of type \tcode{T}.
 Further, whether operations on an \tcode{atomic_ref}
-are lock-free could depend on the alignment of the referenced object.
-For example, lock-free operations on \tcode{std::complex<double>}
-could be supported only if aligned to \tcode{2*alignof(double)}.
+are lock-free can depend on the alignment of the referenced object.
+For example, it is possible for
+lock-free operations on \tcode{std::complex<double>}
+to be supported only if aligned to \tcode{2*alignof(double)}.
```
jwakely, line 802: Maybe "For example, operations referencing a `std::complex<double>` object that is not aligned to `2*alignof(double)`" are not guaranteed to be lock-free."

jensmaurer: That's a change in meaning, because it loses the twist "they are indeed lock-free if properly aligned".

source/atomics.tex
jensmaurer: fixed and force-pushed

source/atomics.tex
jensmaurer: I've rephrased slightly + force-pushed. Maybe you like this better.

source/atomics.tex
```diff
@@ -791,14 +791,15 @@
...
 to have stricter alignment\iref{basic.align}
 than other objects of type \tcode{T}.
 Further, whether operations on an \tcode{atomic_ref}
-are lock-free could depend on the alignment of the referenced object.
-For example, lock-free operations on \tcode{std::complex<double>}
-could be supported only if aligned to \tcode{2*alignof(double)}.
+are lock-free can depend on the alignment of the referenced object.
+For example, it is possible for
+operations on a \tcode{std::complex<double>} object
+to be lock-free only if the object is aligned to \tcode{2*alignof(double)}.
```
jwakely, line 802: This seems to be ambiguous now. "It is possible for X only if Y" can be read as "if and only if Y".

This ISO rule is stupid.

How about "For example, operations on `std::complex<double>` objects can be lock-free if aligned to `2*alignof(double)` but the hardware can require a lock when aligned to `alignof(double)`."

I really thing "might" is a better word for what we're trying to say.

zygoloid: Completely agree about the stupidity of the rule. But I have argued, my face is blue, and we got no closer to publishing the standard. :(

I don't like using "can" for possibility rather than ability or capability. How about this:

"For example, there might exist hardware that supports lock-free atomic operations on `std::complex<double>` objects only if those objects are aligned to `2*alignof(double)`."

johelegp: That uses "might", which are also being replaced.

jensmaurer: > I don't like using "can" for possibility rather than ability or capability.

@zygoloid : Sorry, now I'm confused. Your initial mail gave the guidance "Use "can" when referring to ability or possibility in all cases." This clearly says "possibility", it seems.

zygoloid: You're right, that's what ISO/CS wanted. Doesn't mean that I necessarily like the outcome though :)

@johelegp Curses. But I think that's fairly easy to fix: "For example, it is possible that the hardware supports lock-free atomic operations on `std::complex<double>` objects only if those objects are aligned to `2*alignof(double)`."

jensmaurer: fixed

jwakely: It now mixes plural and singular. It starts talking about objects (plural) then says "only if the object" (singular). Richard's "only if those objects are aligned" seems right to me.

Apart from that, this looks good now.

jensmaurer: Good point, but we should go singular instead. Fixed.

PR #4349, opened 2020-11-11
[thread] Replace 'could' and 'might'

jensmaurer:
as directed by ISO/CS.

Partially addresses #4319

### DIFF
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -134,7 +134,7 @@ should decrease the duration of the wait when the clock is adjusted forwards.
 \pnum
 \begin{note}
 If the clock is not synchronized with a steady clock, e.g., a CPU time clock,
-it is possible for these timeouts not to provide useful functionality.
+it is possible that these timeouts do not provide useful functionality.
 \end{note}
 
 \pnum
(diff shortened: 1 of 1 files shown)


source/threads.tex
```diff
@@ -133,8 +133,8 @@
 
 \pnum
 \begin{note}
-If the clock is not synchronized with a steady clock, e.g., a CPU time clock, these
-timeouts might not provide useful functionality.
+If the clock is not synchronized with a steady clock, e.g., a CPU time clock,
+it is possible for these timeouts not to provide useful functionality.
```
zygoloid, line 137:
```suggestion
it is possible that these timeouts do not provide useful functionality.
```

source/threads.tex
```diff
@@ -7243,10 +7243,12 @@
...
 
 \begin{note}
-Line \#1 might not result in concurrency because
-the \tcode{async} call uses the default policy, which might use
-\tcode{launch::deferred}, in which case the lambda might not be invoked until the
-\tcode{get()} call; in that case, \tcode{work1} and \tcode{work2} are called on the
+It is possible for line \#1 not to result in concurrency
+because the \tcode{async} call uses the default policy,
+which possibly uses \tcode{launch::deferred},
+in which case it is possible for the lambda not to be invoked until the
```
zygoloid, line 7249:
```suggestion
which can use \tcode{launch::deferred},
in which case it is possible that the lambda is not invoked until the
```
I think this reads a bit better. Still incredibly awkward, but maybe this is the best we can do.

PR #4351, opened 2020-11-11
[depr] Replace 'could' and 'might'

jensmaurer:
as directed by ISO/CS.

Partially addresses #4319

### DIFF
--- a/source/future.tex
+++ b/source/future.tex
@@ -209,8 +209,6 @@ The implicit definition of a copy assignment operator\iref{class.copy.assign}
 as defaulted is deprecated if the class has
 a user-declared copy constructor or
 a user-declared destructor.
-In a future revision of \Cpp{}, these implicit definitions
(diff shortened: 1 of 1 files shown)


source/future.tex
```diff
@@ -210,7 +210,7 @@
 a user-declared copy constructor or
 a user-declared destructor.
 In a future revision of \Cpp{}, these implicit definitions
-could become deleted\iref{dcl.fct.def.delete}.
+can become deleted\iref{dcl.fct.def.delete}.
```
zygoloid, line 213: I don't think this is a "can". In fact, given the current indecision on P2139 around whether to undeprecate this case, I think removing the sentence outright would best reflect the current committee's thinking, that is, ¯\_(ツ)_/¯.

jensmaurer: I can't parse the graphical symbols at the end of your utterance.

Sentence removed.

PR #4352, opened 2020-11-11
[diagnostics] Replace 'could' and 'might'

jensmaurer:
as directed by ISO/CS.

Partially addresses #4319

### DIFF
--- a/source/diagnostics.tex
+++ b/source/diagnostics.tex
@@ -1051,7 +1051,7 @@ system_category())}. What constitutes correspondence for any given operating
 system is unspecified.
 \begin{note}
 The number of potential system error codes is large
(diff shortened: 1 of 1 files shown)


jwakely (approved):

source/diagnostics.tex
```diff
@@ -1051,7 +1051,7 @@
 system is unspecified.
 \begin{note}
 The number of potential system error codes is large
-and unbounded, and some might not correspond to any POSIX \tcode{errno} value. Thus
+and unbounded, and some possibly do not correspond to any POSIX \tcode{errno} value. Thus
```
zygoloid, line 1054:
```suggestion
and unbounded, and it is possible that some do not correspond to any POSIX \tcode{errno} value. Thus
```

PR #4353, opened 2020-11-11
[concepts] Replace 'could' and 'might'

jensmaurer:
as directed by ISO/CS.

Partially addresses #4319

### DIFF
--- a/source/concepts.tex
+++ b/source/concepts.tex
@@ -63,7 +63,7 @@ objects based on knowledge of the prior values as observed via
 equality-preserving expressions. It effectively forbids spontaneous changes to
 an object, changes to an object from another thread of execution, changes to an
 object as side effects of non-modifying expressions, and changes to an object as
-side effects of modifying a distinct object if those changes could be observable
+side effects of modifying a distinct object if those changes can be observable
 to a library function via an equality-preserving expression that is required to
 be valid for that object.
 \end{note}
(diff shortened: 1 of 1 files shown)


source/concepts.tex
```diff
@@ -1138,8 +1138,8 @@
 \pnum
 \begin{note}
 The \libconcept{semiregular} concept is modeled by types that behave similarly
-to built-in types like \tcode{int}, except that they might not
-be comparable with \tcode{==}.
+to built-in types like \tcode{int}, except that they are not necessarily
```
CaseyCarter, line 1141: I think I'd prefer "they need not be comparable" to "they are not necessarily comparable" here, but it's not a huge issue.

CaseyCarter (approved):

source/concepts.tex
jensmaurer: fixed + force-pushed

CaseyCarter (approved):
Approved harder!

source/concepts.tex
```diff
@@ -394,7 +394,7 @@
 are modeled, then \tcode{T} and \tcode{U} share a
 \term{common reference type}, \tcode{C}.
 \begin{note}
-\tcode{C} could be the same as \tcode{T}, or \tcode{U}, or it could be a
+\tcode{C} can be the same as \tcode{T}, or \tcode{U}, or it can be a
```
zygoloid, line 397:
```suggestion
\tcode{C} can be the same as \tcode{T} or \tcode{U}, or can be a
```

source/concepts.tex
```diff
@@ -437,8 +437,8 @@
 \tcode{C}, then \tcode{T} and \tcode{U} share a \term{common type},
 \tcode{C}.
 \begin{note}
-\tcode{C} could be the same as \tcode{T}, or \tcode{U}, or it could be a
-different type. \tcode{C} might not be unique.
+\tcode{C} can be the same as \tcode{T}, or \tcode{U}, or it can be a
```
zygoloid, line 440:
```suggestion
\tcode{C} can be the same as \tcode{T} or \tcode{U}, or can be a
```

PR #4354, opened 2020-11-11
[support] Replace 'could' and 'might'

jensmaurer:
as directed by ISO/CS.

Partially addresses #4319

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -542,8 +542,9 @@ Each of the macros defined in \libheader{version} is also defined
 after inclusion of any member of the set of library headers
 indicated in the corresponding comment in this synopsis.
 \begin{note}
-Future revisions of \Cpp{} might replace
-the values of these macros with greater values.
+Future revisions of \Cpp{} will replace
+the values of these macros with greater values
+when the corresponding feature is modified.
 \end{note}
 
 \begin{codeblock}
@@ -1948,13 +1949,13 @@ The
 data race\iref{res.on.data.races}.
 \end{note}
 \begin{note}
-The order of registration could be indeterminate if \tcode{at_quick_exit} was called from more
-than one thread.
+The order of registration can be indeterminate
+if \tcode{at_quick_exit} is called from more than one thread.
 \end{note}
 \begin{note}
 The
 \tcode{at_quick_exit} registrations are distinct from the \tcode{atexit} registrations,
-and applications might need to call both registration functions with the same argument.
+and applications possibly need to call both registration functions with the same argument.
 \end{note}
 
 \pnum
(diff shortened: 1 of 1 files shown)


source/support.tex
```diff
@@ -542,7 +542,7 @@
 after inclusion of any member of the set of library headers
 indicated in the corresponding comment in this synopsis.
 \begin{note}
-Future revisions of \Cpp{} might replace
+Future revisions of \Cpp{} can replace
```
jwakely, line 545: I know it's necessary, but this does not seem like an improvement.

source/support.tex
```diff
@@ -1948,13 +1948,13 @@
...
-The order of registration could be indeterminate if \tcode{at_quick_exit} was called from more
-than one thread.
+The order of registration can be indeterminate
+if \tcode{at_quick_exit} is called from more than one thread.
 \end{note}
 \begin{note}
 The
 \tcode{at_quick_exit} registrations are distinct from the \tcode{atexit} registrations,
-and applications might need to call both registration functions with the same argument.
+and applications can call both registration functions with the same argument.
```
jwakely, line 1957: This seems to change the meaning. I know it's effectively the same, but it imparts different information to the reader.

source/support.tex
CaseyCarter: How about rephrasing completely: "The values of these macros will increase in future revisions of \Cpp{} which modify the corresponding feature."?

jensmaurer: Future revisions of \Cpp{} will replace
the values of these macros with greater values
when the corresponding feature is modified.

source/support.tex
jensmaurer: Tried a different approach.

jwakely (approved):

source/support.tex
```diff
@@ -2236,7 +2237,7 @@
 the program shall also define
 the corresponding version without the \tcode{size} parameter.
 \begin{note}
-The default behavior below might change in the future, which will require
+The default behavior below can change in the future, which will require
 replacing both deallocation functions when replacing the allocation function.
```
zygoloid, line 2241: I think this has lost the intent, which is that we've identified this as a desirable change. How about:
```suggestion
It is anticipated that the default behavior will change in the future and will instead require
replacing both deallocation functions when replacing the allocation function.
```

source/support.tex
```diff
@@ -2491,7 +2492,7 @@
 the program shall also define
 the corresponding version without the \tcode{size} parameter.
 \begin{note}
-The default behavior below might change in the future, which will require
+The default behavior below can change in the future, which will require
```
zygoloid, line 2495:
```suggestion
It is anticipated that the default behavior will change in the future and will instead require
```

jensmaurer:
applied, squashed + force-pushed

PR #4359, opened 2020-11-12
[temp] Replace 'could' and 'might'

jensmaurer:
as directed by ISO/CS.

Partially addresses #4319

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -732,8 +732,8 @@ as the end of the \grammarterm{template-argument-list} and completes
 the \grammarterm{template-id}.
 \begin{note}
 The second \tcode{>}
-token produced by this replacement rule could terminate an enclosing
-\grammarterm{template-id} construct or it could be part of a different
+token produced by this replacement rule can terminate an enclosing
+\grammarterm{template-id} construct or it can be part of a different
 construct (e.g., a cast).
 \end{note}
 \begin{example}
@@ -2142,7 +2142,7 @@ denotes a unique dependent type. Two such \grammarterm{decltype-specifier}{s}
 refer to the same type only if their \grammarterm{expression}{s} are
 equivalent\iref{temp.over.link}.
 \begin{note}
-However, such a type might be aliased,
+However, such a type can be aliased,
 e.g., by a \grammarterm{typedef-name}.
 \end{note}
 
@@ -2214,13 +2214,13 @@ for an unbounded set of related types.
 
 \pnum
 \begin{example}
-A single class template
+It is possible for a single class template
 \tcode{List}
-might provide an unbounded set of class definitions:
+to provide an unbounded set of class definitions:
 one class \tcode{List<T>} for every type \tcode{T},
 each describing a linked list of elements of type \tcode{T}.
 Similarly, a class template \tcode{Array} describing a contiguous,
-dynamic array might be defined like this:
+dynamic array can be defined like this:
 \begin{codeblock}
 template<class T> class Array {
   T* v;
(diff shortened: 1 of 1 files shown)


source/templates.tex
```diff
@@ -2216,11 +2216,11 @@
 \begin{example}
 A single class template
```
zygoloid, line 2217:
```suggestion
It is possible for a single class template
```

source/templates.tex
```diff
@@ -2216,11 +2216,11 @@
...
 \tcode{List}
-might provide an unbounded set of class definitions:
+provides an unbounded set of class definitions:
```
zygoloid, line 2219:
```suggestion
to provide an unbounded set of class definitions:
```

source/templates.tex
```diff
@@ -2216,11 +2216,11 @@
...
 one class \tcode{List<T>} for every type \tcode{T},
 each describing a linked list of elements of type \tcode{T}.
 Similarly, a class template \tcode{Array} describing a contiguous,
-dynamic array might be defined like this:
+dynamic array can be defined like this:
 \begin{codeblock}
 template<class T> class Array {
   T* v;
```
zygoloid, line 2226: (Can't comment on this directly because github is terrible.) There's a bogus "may" a few lines below that should be a "can". ISO/CS noticed and called out this one in our discussion so we need to fix it.

source/templates.tex
```diff
@@ -3344,7 +3344,7 @@
 
 \item
 The template parameter list of a specialization shall not contain default
-template argument values.\footnote{There is no way in which they could be used.}
+template argument values.\footnote{There is no way in which they can be used.}
```
zygoloid, line 3347: Ugh, ISO's rules are awful. I think this doesn't really make sense, because such default arguments can't even exist, so of course they can't be used. How about:
```suggestion
template argument values.\footnote{There is no context in which they would be used.}
```

source/templates.tex
```diff
@@ -8678,8 +8678,8 @@
 struct A { @\commentellip@ };
 struct B : A { @\commentellip@ };
 void g(A a, B b) {
-  f(a,b);           // error: \tcode{T} could be \tcode{A} or \tcode{B}
-  f(b,a);           // error: \tcode{T} could be \tcode{A} or \tcode{B}
+  f(a,b);           // error: \tcode{T} can be \tcode{A} or \tcode{B}
+  f(b,a);           // error: \tcode{T} can be \tcode{A} or \tcode{B}
```
zygoloid, line 8682: I don't think this is right; it can't be either because this is ill-formed. How about:

"error: `T` deduced as both `A` and `B`"

source/templates.tex
```diff
@@ -8699,8 +8699,8 @@
 
 void r() {
   f(g1);            // OK: \tcode{T} is \tcode{int} and \tcode{U} is \tcode{float}
-  f(g2);            // error: \tcode{T} could be \tcode{char} or \tcode{int}
-  f(g3);            // error: \tcode{U} could be \tcode{char} or \tcode{float}
+  f(g2);            // error: \tcode{T} can be \tcode{char} or \tcode{int}
+  f(g3);            // error: \tcode{U} can be \tcode{char} or \tcode{float}
```
zygoloid, line 8703: Similarly here.

source/templates.tex
jensmaurer: fixed

source/templates.tex
jensmaurer: fixed

source/templates.tex
jensmaurer: may -> can

jensmaurer:
applied, squashed + force-pushed

PR #4360, opened 2020-11-12
[over] Replace 'could' and 'might'

jensmaurer:
as directed by ISO/CS.

Partially addresses #4319

### DIFF
--- a/source/overloading.tex
+++ b/source/overloading.tex
@@ -2156,12 +2156,12 @@ Run a simple tournament to find a function
 that is not
 worse than any
 opponent it faced.
-Although another function
+Although it is possible that another function
 \tcode{F}
 that
 \tcode{W}
 did not face
-might be at least as good as
+is at least as good as
 \tcode{W},
 \tcode{F}
 cannot be the best function because at some point in the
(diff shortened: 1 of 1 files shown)


source/overloading.tex
```diff
@@ -2161,7 +2161,7 @@
 that
 \tcode{W}
 did not face
-might be at least as good as
+can be at least as good as
```
zygoloid, line 2164: I don't think "can" really works here. How about:

"Although it is possible that another function `F` that `W` did not face is at least as good as `W`, ..."

source/overloading.tex
```diff
@@ -4199,7 +4199,7 @@
 \pnum
 A declaration whose \grammarterm{declarator-id} is a
 \grammarterm{literal-operator-id} shall be a declaration of a namespace-scope
-function or function template (it could be a friend
+function or function template (it can be a friend
 function\iref{class.friend}), an explicit instantiation or specialization of a
 function template, or a \grammarterm{using-declaration}\iref{namespace.udecl}.
```
zygoloid, line 4204: "can" makes it sound like this is an exhaustive list, as if "it can be ..., and nothing else". This seems a little better:

```suggestion
function or function template (for example, it can be a friend
function\iref{class.friend}), an explicit instantiation or specialization of a
function template, or a \grammarterm{using-declaration}\iref{namespace.udecl}.
```

jensmaurer: fixed

source/overloading.tex
jensmaurer: fixed

PR #4361, opened 2020-11-12
[class] Replace 'could' and 'might'

jensmaurer:
as directed by ISO/CS.

Partially addresses #4319

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -614,7 +614,7 @@ struct S {
 In a \grammarterm{member-declarator} for a bit-field,
 the \grammarterm{constant-expression} is parsed as
 the longest sequence of tokens
-that could syntactically form a \grammarterm{constant-expression}.
+that matches the syntax of a \grammarterm{constant-expression}.
 \begin{example}
 \begin{codeblock}
 int a;
@@ -733,8 +733,8 @@ members have higher addresses within a class object.
 The order of allocation of non-static data members
 with different access control
 is unspecified.
-Implementation alignment requirements might cause two adjacent members
-not to be allocated immediately after each other; so might requirements
+Implementation alignment requirements can cause two adjacent members
+not to be allocated immediately after each other; so can requirements
 for space for managing virtual functions\iref{class.virtual} and
 virtual base classes\iref{class.mi}.
 \end{note}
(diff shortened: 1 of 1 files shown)


source/classes.tex
```diff
@@ -614,7 +614,7 @@
 In a \grammarterm{member-declarator} for a bit-field,
 the \grammarterm{constant-expression} is parsed as
 the longest sequence of tokens
-that could syntactically form a \grammarterm{constant-expression}.
+that would syntactically form a \grammarterm{constant-expression}.
```
zygoloid, line 617: I think we can avoid the hypothetical here (and in other cases where we describe tentative parses):
```suggestion
that matches the syntax of a \grammarterm{constant-expression}.
```

source/classes.tex
```diff
@@ -2303,9 +2303,9 @@
 \begin{note}
 A
 \tcode{return}
-statement\iref{stmt.return} in a destructor might not directly return to the
+statement\iref{stmt.return} in a destructor does not directly return to the
 caller; before transferring control to the caller, the destructors for the
-members and bases are called.
+members and bases (if any) are called.
```
zygoloid, line 2308: The pre-existing wording is unnecessarily verbose. I think the "does not" might be confusing in the case where there are no members or bases. How about replacing this with:

"A `return` statement ([stmt.return]) in a destructor calls the destructors for the members and bases (if any) before transferring control to the caller."

source/classes.tex
```diff
@@ -2692,7 +2692,7 @@
 in a
 \grammarterm{conversion-function-id}
 is the longest sequence of
-tokens that could possibly form a \grammarterm{conversion-type-id}.
+tokens that would possibly form a \grammarterm{conversion-type-id}.
```
zygoloid, line 2695:
```suggestion
tokens that match the syntax of a \grammarterm{conversion-type-id}.
```

source/classes.tex
```diff
@@ -6546,7 +6546,7 @@
 requiring a constant expression\iref{expr.const}
 and in constant initialization\iref{basic.start.static}.
 \begin{note}
-Copy elision might be performed
+It is possible for copy elision to be performed
```
zygoloid, line 6549:
```suggestion
It is possible that copy elision is performed
```

source/classes.tex
jensmaurer: fixed

source/classes.tex
jensmaurer: fixed

source/classes.tex
jensmaurer: "that matches" (singular); it's referring to the sequence

fixed

jensmaurer:

PR #4362, opened 2020-11-12
[dcl.dcl] Replace 'could' and 'might'

jensmaurer:
as directed by ISO/CS.

Partially addresses #4319

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -924,14 +924,11 @@ shall additionally satisfy the following requirement:
 \end{itemize}
 
 \pnum
-For a constexpr function or constexpr constructor
-that is neither defaulted nor a template,
-if no argument values exist such that
-an invocation of the function or constructor could be an evaluated subexpression of a core
-constant expression\iref{expr.const}, or,
-for a constructor, an evaluated subexpression of
-the initialization full-expression of some constant-initialized object\iref{basic.start.static},
-the program is ill-formed, no diagnostic required.
+A constexpr function that is neither defaulted nor a template
+is ill-formed, no diagnostic required,
+if it is not possible for an evaluation of an invocation of the function
+to be performed
+while evaluating any valid manifestly constant-evaluated expression.
 \begin{example}
 \begin{codeblock}
 constexpr int f(bool b)
@@ -1238,7 +1235,7 @@ function types.
 Redundant cv-qualifications are ignored.
 \begin{note}
 For example,
-these could be introduced by typedefs.
+these can be introduced by typedefs.
 \end{note}
 
 \pnum
@@ -1317,9 +1314,9 @@ the behavior is undefined.
 \indextext{\idxcode{volatile}!implementation-defined}%
 \begin{note}
 \tcode{volatile} is a hint to the implementation to avoid aggressive
-optimization involving the object because the value of the object might
-be changed by means undetectable by an implementation.
-Furthermore, for some implementations, \tcode{volatile} might indicate that
+optimization involving the object because it is possible for the value of the object
+to change by means undetectable by an implementation.
+Furthermore, for some implementations, \tcode{volatile} can indicate that
 special hardware instructions are required to access the object.
 See~\ref{intro.execution} for detailed semantics. In general, the
 semantics of \tcode{volatile} are intended to be the same in \Cpp{} as
@@ -2400,8 +2397,9 @@ In that context, the choice is between a function declaration with
 a redundant set of parentheses around a parameter name and an object declaration
 with a function-style cast as the initializer.
 Just as for the ambiguities mentioned in~\ref{stmt.ambig},
-the resolution is to consider any construct that could possibly
-be a declaration a declaration.
+the resolution is to consider any construct
+that matches the syntax of a declaration
+to be a declaration.
 \begin{note}
 A declaration can be explicitly disambiguated by adding parentheses
 around the argument.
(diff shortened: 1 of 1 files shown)


tkoeppe (approved):

source/declarations.tex
```diff
@@ -927,7 +927,7 @@
 For a constexpr function or constexpr constructor
 that is neither defaulted nor a template,
 if no argument values exist such that
-an invocation of the function or constructor could be an evaluated subexpression of a core
+an invocation of the function or constructor would be an evaluated subexpression of a core
```
zygoloid, line 930: I find this very hard to understand with "would" here. I think part of the problem is that the original wording is unclear. What we mean is:

"For a constexpr function or constexpr constructor that is neither defaulted nor a template, if there does not exist a valid core constant expression whose evaluation would evaluate an invocation of the function or constructor, and, for a constructor, there does not exist a valid declaration of a constant-initialized variable whose evaluation would evaluate an invocation of the function or constructor."

That is, this is really "for all possible core constant expressions", not "for all possible argument values".

I think we can simplify and clarify that:

"A constexpr function that is neither defaulted nor a template is ill-formed, no diagnostic required, if it is not possible for an evaluation of an invocation of the function to be performed while evaluating any valid manifestly constant-evaluated expression."

source/declarations.tex
```diff
@@ -2400,7 +2400,7 @@
 a redundant set of parentheses around a parameter name and an object declaration
 with a function-style cast as the initializer.
 Just as for the ambiguities mentioned in~\ref{stmt.ambig},
-the resolution is to consider any construct that could possibly
+the resolution is to consider any construct that can possibly
 be a declaration a declaration.
```
zygoloid, line 2404:
```suggestion
the resolution is to consider any construct
that matches the syntax of a declaration
to be a declaration.
```

source/declarations.tex
```diff
@@ -2428,7 +2428,7 @@
 An ambiguity can arise from the similarity between a function-style
 cast and a
 \grammarterm{type-id}.
-The resolution is that any construct that could possibly be a
+The resolution is that any construct that can possibly be a
 \grammarterm{type-id}
 in its syntactic context shall be considered a
 \grammarterm{type-id}.
```
zygoloid, line 2434:
```suggestion
The resolution is that any construct
that matches the syntax of a \grammarterm{type-id}
is interpreted as a \grammarterm{type-id}.
```

source/declarations.tex
```diff
@@ -4602,9 +4602,9 @@
 no constructor of the object's class
 is invoked for the initialization.
 \begin{note}
-Such an object might have been value-initialized
-or initialized by aggregate initialization\iref{dcl.init.aggr}
-or by an inherited constructor\iref{class.inhctor.init}.
+Such an object can use value-initializion,
+or aggregate initialization\iref{dcl.init.aggr}
```
zygoloid, line 4606:
```suggestion
aggregate initialization\iref{dcl.init.aggr},
```

source/declarations.tex
```diff
@@ -5868,8 +5868,8 @@
 int x = 999;                    // \tcode{x} is not a constant expression
 const int y = 999;
 const int z = 99;
-char c1 = x;                    // OK, though it might narrow (in this case, it does narrow)
-char c2{x};                     // error: might narrow
+char c1 = x;                    // OK, though it can narrow (in this case, it does narrow)
+char c2{x};                     // error: can narrow
```
zygoloid, line 5872: I don't think "can" makes sense here, because the code is ill-formed, so it can't do anything. How about "error: potentially narrows"?

source/declarations.tex
```diff
@@ -5878,7 +5878,7 @@
 signed int si1 =
   { (unsigned int)-1 };         // error: narrows
 int ii = {2.0};                 // error: narrows
-float f1 { x };                 // error: might narrow
+float f1 { x };                 // error: can narrow
```
zygoloid, line 5881: Likewise.

source/declarations.tex
```diff
@@ -8793,7 +8793,7 @@
 Function \tcode{g}'s second parameter has a \tcode{carries_dependency} attribute,
 but its first parameter does not. Therefore, function \tcode{h}'s first call to
 \tcode{g} carries a dependency into \tcode{g}, but its second call does not. The
-implementation might need to insert a fence prior to the second call to
+implementation can be required to insert a fence prior to the second call to
```
zygoloid, line 8796: "can be required" sounds like the programmer can cause the implementation to be required to do this, which isn't really right. This solves that problem but I'm still not overjoyed by it:
```suggestion
implementation can need to insert a fence prior to the second call to
```

source/declarations.tex
```diff
@@ -8876,7 +8876,7 @@
 \pnum
 \recommended
 The use of a fallthrough statement should suppress
-a warning that an implementation might otherwise issue
+a warning that an implementation would otherwise issue
```
zygoloid, line 8879:
```suggestion
any warning that an implementation would otherwise issue
```
(To avoid the suggestion that there would necessarily be such a warning.)

source/declarations.tex
jensmaurer: fixed

source/declarations.tex
jensmaurer: fixed

source/declarations.tex
jensmaurer: fixed

source/declarations.tex
jensmaurer: "can need"?

ugh

"It is possible that the implementation needs to insert a fence..."

jensmaurer:

PR #4364, opened 2020-11-12
[expr] Replace 'could' and 'might'

jensmaurer:
as directed by ISO/CS.

Partially addresses #4319

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -770,7 +770,7 @@ if the cv-combined type of \tcode{T1} and \tcode{T2} is \tcode{T2}.
 If a program were able to assign a pointer of type \tcode{T**} to a pointer of
 type \tcode{const} \tcode{T**} (that is, if line \#1 below were
 allowed), it would be possible for a const object to be inadvertently modified
-(as it is done on line \#2). For example,
+(as is done on line \#2). For example,
 \begin{codeblock}
 int main() {
   const char c = 'c';
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -109,7 +109,7 @@ or
 \begin{codeblock}
 a = (a + (b + 32765));
 \end{codeblock}
-since the values for \tcode{a} and \tcode{b} might have been,
+since the values for \tcode{a} and \tcode{b} can be,
 respectively, 4 and -8 or -17 and 12. However on a machine in which
 overflows do not produce an exception and in which the results of
 overflows are reversible, the above expression statement can be
@@ -165,7 +165,7 @@ values they expect and yield are determined by their parameter and return types.
 \pnum
 \begin{note}
 Historically, lvalues and rvalues were so-called
-because they could appear on the left- and right-hand side of an assignment
+because they can appear on the left- and right-hand side of an assignment
 (although this is no longer generally true);
 glvalues are ``generalized'' lvalues,
 prvalues are ``pure'' rvalues,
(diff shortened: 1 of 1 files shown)


source/expressions.tex
```diff
@@ -165,7 +165,7 @@
 \pnum
 \begin{note}
 Historically, lvalues and rvalues were so-called
-because they could appear on the left- and right-hand side of an assignment
+because they can appear on the left- and right-hand side of an assignment
```
tkoeppe, line 168: This is awkward, given the parenthetical. Happy to turn a blind eye, but not for 23.

source/expressions.tex
```diff
@@ -1927,7 +1927,7 @@
 operators otherwise\iref{class.copy.assign}.
 \begin{note}
 These special member functions are implicitly defined as
-usual, and might therefore be defined as deleted.
+usual, and can therefore be defined as deleted.
```
tkoeppe, line 1930: This sounds uncomfortably ambiguous with the meaning of "and therefore you the user can define it as deleted". The previous wording wasn't quite so readily interpretable that way. Maybe extra words like "and can therefor end up as being defined as deleted". (Maybe for 23.)

tkoeppe (approved):

source/expressions.tex
```diff
@@ -767,9 +767,9 @@
 can be converted to type \tcode{T2}
 if the cv-combined type of \tcode{T1} and \tcode{T2} is \tcode{T2}.
 \begin{note}
-If a program could assign a pointer of type \tcode{T**} to a pointer of
+If a program were able to assign a pointer of type \tcode{T**} to a pointer of
 type \tcode{const} \tcode{T**} (that is, if line \#1 below were
-allowed), a program could inadvertently modify a const object
+allowed), a program would be able to inadvertently modify a const object
```
zygoloid, line 772:
```suggestion
allowed), it would be possible for a const object to be inadvertently modified
```
I think we want to say that it's possible for this to happen, not that a program is able to do it. I still think this is a wording regression, since it creates the impression that this rule prevents inadvertent modification of const objects in general, which it does not.

source/expressions.tex
zygoloid: How about:
```suggestion
usual, which can result in them being defined as deleted.
```

source/expressions.tex
```diff
@@ -2135,20 +2135,20 @@
...
 }
 \end{codeblock}
-Within \tcode{g1}, an implementation might optimize away
+Within \tcode{g1}, an implementation can optimize away
 the capture of \tcode{x} as it is not odr-used.
 \end{example}
 \begin{note}
 The set of captured entities is determined syntactically,
-and entities might be implicitly captured
+and entities can be implicitly captured
```
zygoloid, line 2151:
```suggestion
and entities are implicitly captured
```

jensmaurer:

source/expressions.tex
tkoeppe: Great, another example to go on the "noteworthy examples" list!

source/expressions.tex
zygoloid: Looks like github lost my comment on this :(
```suggestion
because they appeared on the left- and right-hand side of an assignment
```

source/expressions.tex
```diff
@@ -767,9 +767,9 @@
 can be converted to type \tcode{T2}
 if the cv-combined type of \tcode{T1} and \tcode{T2} is \tcode{T2}.
 \begin{note}
-If a program could assign a pointer of type \tcode{T**} to a pointer of
+If a program were able to assign a pointer of type \tcode{T**} to a pointer of
 type \tcode{const} \tcode{T**} (that is, if line \#1 below were
-allowed), a program could inadvertently modify a const object
+allowed), it would be possible for a const object to be inadvertently modified
 (as it is done on line \#2). For example,
```
zygoloid, line 773:
```suggestion
(as is done on line \#2). For example,
```

PR #4365, opened 2020-11-12
[basic] Replace 'could' and 'might'

jensmaurer:
as directed by ISO/CS.

Partially addresses #4319

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -695,10 +695,10 @@ including in the application of these requirements to other entities.
 The entity is still declared in multiple translation units, and \ref{basic.link}
 still applies to these declarations. In particular,
 \grammarterm{lambda-expression}{s}\iref{expr.prim.lambda}
-appearing in the type of \tcode{D} might result
+appearing in the type of \tcode{D} can result
 in the different declarations having distinct types, and
 \grammarterm{lambda-expression}{s} appearing in a default argument of \tcode{D}
-might still denote different types in different translation units.
+can still denote different types in different translation units.
 \end{note}
 
 \pnum
@@ -1534,7 +1534,7 @@ void A::N::f() {
 A name used in the definition of a class \tcode{X}%
 \footnote{This
 refers to unqualified names following the class name;
-such a name might be used in a \grammarterm{base-specifier} or
+such a name can be used in a \grammarterm{base-specifier} or
 in the \grammarterm{member-specification} of the class definition.}
 outside of a complete-class context\iref{class.mem} of \tcode{X}
 shall be declared in one of the following ways:
@@ -2837,7 +2837,7 @@ Similarly,
 \grammarterm{using-directive}{s}
 do not declare entities.
 Enumerators do not have linkage,
-but might serve as the name of an enumeration with linkage\iref{dcl.enum}.
+but can serve as the name of an enumeration with linkage\iref{dcl.enum}.
 \end{note}
 
 \pnum
@@ -2934,7 +2934,7 @@ that is not an odr-use\iref{basic.def.odr},
 or defines a constexpr variable initialized to a TU-local value (defined below).
 \begin{note}
 An inline function template can be an exposure even though
-explicit specializations of it might be usable in other translation units.
+explicit specializations of it are possibly usable in other translation units.
 \end{note}
 
 \pnum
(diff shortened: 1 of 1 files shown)


source/basic.tex
```diff
@@ -2965,7 +2965,7 @@
 a specialization of a template
 whose (possibly instantiated) declaration is an exposure.
 \begin{note}
-The specialization might have been implicitly or explicitly instantiated.
+A specialization can be produced by implicitly or explicitly instantiation.
```
tkoeppe, line 2968: Should these adverbs become adjectives?

source/basic.tex
```diff
@@ -3442,7 +3442,7 @@
 lifetime ends there are significant restrictions on the use of the
 object, as described below, in~\ref{class.base.init} and
 in~\ref{class.cdtor}. Also, the behavior of an object under construction
-and destruction might not be the same as the behavior of an object whose
+and destruction is possibly not the same as the behavior of an object whose
```
tkoeppe, line 3445: This is very nice. We should give that as an example of possible rewordings.

source/basic.tex
```diff
@@ -5262,8 +5261,8 @@
 \begin{note}
 A pointer past the end of an object\iref{expr.add}
 is not considered to point to an unrelated object
-of the object's type
-that might be located at that address.
+of the object's type,
+even if the unrelated object would be located at that address.
```
tkoeppe, line 5265: Can we just say "is located" here? What does the extra qualification achieve here?

source/basic.tex
```diff
@@ -6351,7 +6350,7 @@
 guarantees, which in turn are stronger than weakly parallel forward progress
 guarantees.
 \begin{note}
-For example, some kinds of synchronization between threads of execution might only
+For example, some kinds of synchronization between threads of execution would only
 make progress if the respective threads of execution provide parallel forward progress
```
tkoeppe, line 6354: I feel like this should now say "if the respective threads ... _provided_ parallel forward progress guarantees." What do you think?

tkoeppe (changes requested):

source/basic.tex
jensmaurer: fixed

source/basic.tex
jensmaurer: fixed

source/basic.tex
jensmaurer: of course
fixed

jensmaurer:
fixed + force-pushed

source/basic.tex
```diff
@@ -6277,7 +6276,7 @@
 \begin{example}
 A library I/O function that blocks until the I/O operation is complete can
 be considered to continuously check whether the operation is complete. Each
-such check might consist of one or more execution steps, for example using
+such check possibly consists of one or more execution steps, for example using
```
zygoloid, line 6279: What is the alternative? (If it doesn't consist of one or more execution steps, then what does it do?) Can we drop the "possibly"?

source/basic.tex
```diff
@@ -6351,8 +6350,8 @@
 guarantees, which in turn are stronger than weakly parallel forward progress
 guarantees.
 \begin{note}
-For example, some kinds of synchronization between threads of execution might only
-make progress if the respective threads of execution provide parallel forward progress
+For example, some kinds of synchronization between threads of execution would only
+make progress if the respective threads of execution provided parallel forward progress
 guarantees, but will fail to make progress under weakly parallel guarantees.
```
zygoloid, line 6355: Is this really right? Should this instead say:
```suggestion
For example, some kinds of synchronization between threads of execution are only guaranteed to
make progress if the respective threads of execution provide parallel forward progress
guarantees, but will not necessarily make progress under weakly parallel guarantees.
```

source/basic.tex
```diff
@@ -4225,7 +4225,7 @@
 When \tcode{D} is the type of a complete object, it will have a subobject of
 type \tcode{B}, so it must be aligned appropriately for a \tcode{long double}.
 If \tcode{D} appears as a subobject of another object that also has \tcode{B}
-as a virtual base class, the \tcode{B} subobject might be part of a different
+as a virtual base class, it is possible for the \tcode{B} subobject to be part of a different
 subobject, reducing the alignment requirements on the \tcode{D} subobject.
```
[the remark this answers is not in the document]
jensmaurer: fixed

source/basic.tex
jensmaurer: dropped

jensmaurer:

PR #4366, opened 2020-11-12
[lex] Replace 'could' and 'might'

jensmaurer:
as directed by ISO/CS.

Partially addresses #4319

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -344,7 +344,7 @@ given character:
 \begin{itemize}
 \item
 \indextext{literal!string!raw}%
-If the next character begins a sequence of characters that could be the prefix
+If the next character begins a sequence of characters that can be the prefix
 and initial double quote of a raw string literal, such as \tcode{R"}, the next preprocessing
 token shall be a raw string literal. Between the initial and final
 double quote characters of the raw string, any transformations performed in phases
@@ -362,7 +362,7 @@ itself and not as the first character of the alternative token \tcode{<:}.
 
 \item Otherwise,
 the next preprocessing token is the longest sequence of
-characters that could constitute a preprocessing token, even if that
+characters that matches the syntax of a preprocessing token, even if that
 would cause further lexical analysis to fail,
 except that a \grammarterm{header-name}\iref{lex.header} is only formed
 \begin{itemize}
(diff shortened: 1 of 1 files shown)


source/lex.tex
```diff
@@ -401,7 +401,7 @@
 preprocessing number token (one that is not a valid
 \grammarterm{integer-literal} or \grammarterm{floating-point-literal} token),
 even though a parse as three preprocessing tokens
-\tcode{0xe}, \tcode{+}, and \tcode{foo} might produce a valid expression (for example,
+\tcode{0xe}, \tcode{+}, and \tcode{foo} can produce a valid expression (for example,
 if \tcode{foo} were a macro defined as \tcode{1}). Similarly, the
```
tkoeppe, line 405: If we keep "can" here, then I think the parenthetical needs to say "is", not "were".

Alternatively, could we say "would be able to produce"?

source/lex.tex
```diff
@@ -362,7 +362,7 @@
 
 \item Otherwise,
 the next preprocessing token is the longest sequence of
-characters that could constitute a preprocessing token, even if that
+characters that would constitute a preprocessing token, even if that
```
tkoeppe, line 365: I'm not sure "would" works here. Second options? @zygoloid?

source/lex.tex
```diff
@@ -1893,7 +1893,7 @@
 \end{example}
 The syntactic non-terminal preceding the \grammarterm{ud-suffix} in a
 \grammarterm{user-defined-literal} is taken to be the longest sequence of
-characters that could match that non-terminal.
+characters that would match that non-terminal.
```
tkoeppe, line 1896: (Same here as above. If the above is OK, so is this.)

source/lex.tex
jensmaurer: We have a bunch of max-munch rules that all follow the same pattern.

tkoeppe: OK, then maybe this is fine as is (likewise below).

zygoloid: What I've been suggesting in other PRs is:
"longest sequence of characters that matches the syntax of a *preprocessing-token*"
... which seems more precise than our prior wording in the various tentative parsing areas, and distinguishes between matching the syntax of X and actually forming a valid X, without getting into hypotheticals.

source/lex.tex
zygoloid: I'm happy with either. I'm a little worried that ISO will decide to declare "would" as verboten next (the upcoming revised Directives disallow use of any verbal forms not explicitly permitted, which would disallow "would"). But we can face that if and when it happens.

source/lex.tex
zygoloid:
```suggestion
characters that match the syntax of that non-terminal.
```

source/lex.tex
jensmaurer: fixed

source/lex.tex
jensmaurer: fixed to "is"

We need to be able to talk about hypotheticals in proper English phrasing, that includes "could" and "were" and "would" and "might". Everything else is just stupid.
(I agree with some of the -> can changes we have been doing; they seem to be clear improvements occasionally.)

PR #4374, opened 2020-11-14
[thread.sharedtimedmutex.requirements.general] Fix typo

timsong-cpp:


jensmaurer (approved):

jwakely (approved):

jwakely:
"an instantiation of" needs to change, but we should do that everywhere and it shouldn't block this fix.

PR #4385, opened 2020-11-22
[over.call.func, gram.key] Make colon in bnf non-italic.

Eelis:


jensmaurer (changes requested):
For "normal" bnf, we have `\nontermdef` for the initial thing.

"ncbnf" are the non-copied BNF parts (those that aren't copied to Annex A), but we should have a similar macro defined there, instead of hand-adding a colon or (gosh) hand-changing it to `\textnormal`.

godbyk:
We could make the colon an &ldquo;active character.&rdquo; This would allow you to just type an unadorned colon in the source (within the BNF environments) and LaTeX would interpret it as a macro that expands to `\textnormal{:}` (or similar).

tkoeppe:
@jensmaurer's suggestion sounds somewhat more structured -- what do you think? That way people don't have to remember to type colons at all.

godbyk:
As long as `\nontermdef` is only used for the symbol being defined, then that'd work just fine.

source/overloading.tex
```diff
@@ -745,7 +745,7 @@
 parentheses, has one of the following forms:
 
 \begin{ncbnf}
-postfix-expression:\br
+postfix-expression\textnormal{:}\br
```
cpplearner, line 748: Does this line add any value? Can it be removed?

Eelis:
I noticed there was already a non-indexed version of \nontermdef used for format nontermdefs: \fmtnontermdef.

So I've changed the patch to:
- move that \fmtnontermdef to macros.tex
- rename it to \ninontermdef for "Non-Indexed nontermdef" (similar to how "ncbnf" means "Non-Copied bnf")
- use \ninontermdef in the new places to fix the italic colons.

source/macros.tex
```diff
@@ -573,7 +573,8 @@
 
 \newenvironment{bnfbase}
  {
- \newcommand{\nontermdef}[1]{{\BnfNontermshape##1\itcorr}\indexgrammar{\idxgram{##1}}\textnormal{:}}
+ \newcommand{\nontermdef}[1]{{\BnfNontermshape##1\itcorr}\indexgrammar{\idxgram{##1}}\textnormal{:}}%
```
jensmaurer, line 576: Drop the trailing percent; it has not use inside macros.tex

source/utilities.tex
```diff
@@ -19238,7 +19238,6 @@
 % FIXME: For now, keep the format grammar productions out of the index, since
 % they conflict with the main grammar.
 % Consider renaming these en masse (to fmt-* ?) to avoid this problem.
-\newcommand{\fmtnontermdef}[1]{{\BnfNontermshape#1\itcorr}\textnormal{:}}
```
jensmaurer, line 19241: I would prefer if you would leave the uses of `\fmtnontermdef` as-is and just change its definition to be `\ninontermdef`.
(This keeps the semantic markup of the fmt-related grammar.)

jensmaurer (changes requested):

PR #4386, opened 2020-11-23
Replace "could" and "might", Clauses 16-32.

tkoeppe:
These changes editorially remove the words "could" and "might".

Please review this change carefully with the following goals:

Flag anything that seems even remotely non-editorial, e.g. even if the explanation provided by a Note becomes less useful or (more) misleading.
Also flag anything that could conceivably be worded better by a wider review.
I would like all remaining instances of "could" and "might" to be reviewed by CWG/LWG, so please don't hold back on rejecting anything at this point.

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -3926,7 +3926,7 @@ Whether the input sequence is backed up or modified in any other way is unspecif
 \returns
 \tcode{traits::eof()}
 to indicate failure.
-Failure may occur because the input sequence cannot be backed up, or if for some
+Failure may occur because the input sequence could not be backed up, or if for some
 other reason the pointers cannot be set consistent with the constraints.
 \tcode{pbackfail()}
 is called only when put back has really failed.
@@ -14942,7 +14942,7 @@ as a \tcode{directory_iterator} unless otherwise specified.
 \pnum
 \begin{note}
 If the directory structure being iterated over contains cycles
-then it is possible for the end iterator to be unreachable.
+then it is possible that the end iterator is unreachable.
 \end{note}
 
 \rSec3[fs.rec.dir.itr.members]{Members}
@@ -15865,7 +15865,7 @@ The \tcode{current_path()} name was chosen to emphasize that the returned value 
 \pnum
 \begin{note}
 The current path as returned by many operating systems is a dangerous
-  global variable, and can be changed unexpectedly by third-party or system
+  global variable and can be changed unexpectedly by third-party or system
   library functions, or by another thread.
 \end{note}
 \end{itemdescr}
@@ -15892,7 +15892,7 @@ As specified in~\ref{fs.err.report}.
 \pnum
 \begin{note}
 The current path for many operating systems is a dangerous
-  global state, and can be changed unexpectedly by a third-party or system
+  global state and can be changed unexpectedly by third-party or system
   library functions, or by another thread.
 \end{note}
 \end{itemdescr}
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -3696,7 +3696,7 @@ will fail.
 \tcode{underflow}
 or
 \tcode{uflow}
-might fail by throwing an exception prematurely.
+can fail by throwing an exception prematurely.
 The intention is not only that the calls will not return
 \tcode{eof()}
 but that they will return ``immediately''.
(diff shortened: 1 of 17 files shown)


tkoeppe:
@opensdh: Could you perhaps take a look at the filesystem changes?

@geoffromer: Could you perhaps take a look at the atomics and threads changes?

Thank you!

source/iostreams.tex
```diff
@@ -14942,7 +14942,7 @@
 \pnum
 \begin{note}
 If the directory structure being iterated over contains cycles
-then the end iterator might be unreachable.
+then it is possible for the end iterator to be unreachable.
```
opensdh, line 14945: I might say, slightly more directly, "then it is possible that the end iterator is unreachable".

source/iostreams.tex
```diff
@@ -15865,7 +15865,7 @@
 \pnum
 \begin{note}
 The current path as returned by many operating systems is a dangerous
-  global variable. It might be changed unexpectedly by third-party or system
+  global variable, and can be changed unexpectedly by third-party or system
```
opensdh, line 15868: No comma here.

source/iostreams.tex
```diff
@@ -15892,7 +15892,7 @@
 \pnum
 \begin{note}
 The current path for many operating systems is a dangerous
-  global state. It might be changed unexpectedly by a third-party or system
+  global state, and can be changed unexpectedly by a third-party or system
```
opensdh, line 15895: No comma or "a" here.

opensdh (commented):
The `filesystem` changes generally look fine, but I made a few minor comments (one of which is just caught up in the same file).

source/threads.tex
```diff
@@ -136,8 +136,8 @@
 
 \pnum
 \begin{note}
-If the clock is not synchronized with a steady clock, e.g., a CPU time clock, these
-timeouts might not provide useful functionality.
+If the clock is not synchronized with a steady clock, e.g., a CPU time clock,
+it is possible that these timeouts do not provide useful functionality.
```
geoffromer, line 140: This feels a little clearer and less awkward:

```suggestion
If the clock is not synchronized with a steady clock, e.g., a CPU time clock, these
timeouts can fail to provide useful functionality.
```

Something like "...these timeouts are not guaranteed to provide..." would be even better, but I'm not sure if that passes ISO muster.

source/threads.tex
```diff
@@ -7175,8 +7175,8 @@
 \begin{note}
 These apply regardless of the provided \tcode{policy} argument, and
 even if the corresponding \tcode{future} object is moved to another thread.
-However, \tcode{f} might not be called at all,
-so its completion might never happen.
+It is possible for \tcode{f} not to be called at all,
```
geoffromer, line 7178:
```suggestion
However, it is possible for \tcode{f} not to be called at all,
```

geoffromer (commented):
This generally looks good, but I should note that I haven't been active in SG1 since February (or in editorial work for several years).

source/iostreams.tex
```diff
@@ -3926,8 +3926,8 @@
 \returns
 \tcode{traits::eof()}
 to indicate failure.
-Failure may occur because the input sequence could not be backed up, or if for some
-other reason the pointers could not be set consistent with the constraints.
+Failure may occur because the input sequence cannot be backed up, or if for some
```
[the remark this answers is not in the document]
tkoeppe: I agree that the existing wording is already questionable, so let's keep this entire change out and have it reviewed by the wording group.

source/iostreams.tex
tkoeppe: That's nice, thanks. That's exactly the kind of comment I am looking for: we don't just want to change the bad words away mechanically, but ideally find nice ways to say something equivalent that avoids them.

source/iostreams.tex
tkoeppe: Done.

source/iostreams.tex
tkoeppe: Done.

source/threads.tex
tkoeppe: Thanks, the suggestion is a nice improvement. About "not guaranteed to provide useful functionality", I suspect that that additional strengthening wouldn't help much if we don't also define "useful functionality", so I think what we have now will do the job.

tkoeppe:
@CaseyCarter: could you perhaps take a look at the ranges and concepts, particularly with a view to whether "might->can" confuses risk with possibility?

PR #4390, opened 2020-11-25
Replace "could" and "might" in Annexes.

tkoeppe:


tkoeppe:
@zygoloid, @opensdh: Your feedback would be most welcome.

source/future.tex
```diff
@@ -204,8 +204,6 @@
 as defaulted is deprecated if the class has
 a user-declared copy constructor or
 a user-declared destructor.
-In a future revision of \Cpp{}, these implicit definitions
-could become deleted\iref{dcl.fct.def.delete}.
```
tkoeppe, line 208: @zygoloid: Do you think this deletion is sufficiently editorial, and we're not deleting WG21 intent?

opensdh: In constructs like this, it seems adequately equivalent (and apparently ISO-legitimate) to say "It is possible that future versions of C++ will specify that these implicit definitions are deleted ([dcl.fct.def.delete]).", avoiding the question of changing the intent.

source/compatibility.tex
```diff
@@ -1439,7 +1440,7 @@
 \rationale
 Overly constrained, simplify overload resolution rules.
 \effect
-A valid \CppIII{} program could get a different result than in this
+A valid \CppIII{} program can get a different result than in this
```
opensdh, line 1443: I'd omit "than" here.  Also, couldn't such a program become ambiguous?

source/compatibility.tex
```diff
@@ -1851,7 +1852,7 @@
 to ``array of \tcode{const wchar_t}''.
 \rationale
 This avoids calling an inappropriate overloaded function,
-which might expect to be able to modify its argument.
+which would possibly attempt to modify its argument.
```
opensdh, line 1855: This sounds like any such inappropriate function has a chance of making such a modification.  Maybe say "can be expected"?  It's a bit of a cheat, since we're not really saying that it's _possible_ for a person to expect the function to do something, but rather that it's a _reasonable_ expectation, but it's an allowed verb.

source/compatibility.tex
```diff
@@ -2097,7 +2098,7 @@
 block.
 Allowing jump past initializers would require
 complicated runtime determination of allocation.
-Furthermore, any use of the uninitialized object could be a
+Furthermore, any use of the uninitialized object can be a
```
opensdh, line 2101: This again seems to convey a risk: even taking the address of the uninitialized object can be a disaster!  Perhaps more formal language will help:

> Furthermore, many operations on such an uninitialized object have undefined behavior.

source/compatibility.tex
```diff
@@ -2155,7 +2156,7 @@
 In \Cpp{}, class members can be declared with the \tcode{static} storage
 class specifier.
 Allowing storage class specifiers on type
-declarations could render the code confusing for users.
+declarations can render the code confusing for users.
```
opensdh, line 2159: _Allowing_ them can't render the code confusing because we don't allow them.  Just strike "Allowing" (implicitly describing C instead).

source/compatibility.tex
```diff
@@ -2547,7 +2548,7 @@
 \indextext{bit-field!implementation-defined sign of}%
 Bit-fields of type plain \tcode{int} are signed.
 \rationale
-Leaving the choice of signedness to implementations could lead to
+Leaving the choice of signedness to implementations can lead to
```
opensdh, line 2551: Similarly, the leaving can't do anything since it doesn't happen.  Maybe

> The signedness needs to be consistent among template specializations.

source/future.tex
tkoeppe: Thanks, that sounds like a better solution!

source/compatibility.tex
```diff
@@ -387,9 +387,9 @@
 \effect
 Equality and inequality expressions between two objects of different types,
 where one is convertible to the other,
-could invoke a different operator.
+can invoke a different operator.
```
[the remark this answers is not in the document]
tkoeppe: Thank you, this too will go into a separate PR!

source/compatibility.tex
```diff
@@ -668,8 +668,8 @@
 \change
 Remove \tcode{raw_storage_iterator}.
 \rationale
-The iterator encouraged use of algorithms that might throw exceptions, but did
-not return the number of elements successfully constructed that might need to
+The iterator encouraged use of potentially-throwing algorithms, but did
+not return the number of elements successfully constructed, which need to
```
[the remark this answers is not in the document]
tkoeppe: Ditto.

source/compatibility.tex
```diff
@@ -705,7 +705,7 @@
 The traits had unreliable or awkward interfaces. The \tcode{is_literal_type}
 trait provided no way to detect which subset of constructors and member
 functions of a type were declared \tcode{constexpr}. The \tcode{result_of}
-trait had a surprising syntax that could not report the result of a regular
+trait had a surprising syntax that would not report the result of a regular
```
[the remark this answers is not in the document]
tkoeppe: Ditto.

source/compatibility.tex
```diff
@@ -766,7 +766,7 @@
 \effect
 A valid \CppXIV{} expression utilizing the increment operator on
 a \tcode{bool} lvalue is ill-formed in this revision of \Cpp{}.
-Note that this might occur when the lvalue has a type given by a template
+Note that this can occur when the lvalue has a type given by a template
```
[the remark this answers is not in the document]
tkoeppe: Yeah, that's another case of @zygoloid's risk/possiblity confusion. Reverting.

source/compatibility.tex
```diff
@@ -1111,15 +1111,15 @@
...
+Valid \CppXI{} code can declare a global placement allocation function and
 deallocation function as follows:
 \begin{codeblock}
 void* operator new(std::size_t, std::size_t);
 void operator delete(void*, std::size_t) noexcept;
 \end{codeblock}
 
 In this revision of \Cpp{}, however, the declaration of \tcode{operator delete}
-might match a predefined usual (non-placement)
+can match a predefined usual (non-placement)
```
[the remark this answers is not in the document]
tkoeppe: Oh, I'm not sure, is it? So, just "matches"?

source/compatibility.tex
tkoeppe: Separate PR.

source/compatibility.tex
tkoeppe: Ditto

source/compatibility.tex
tkoeppe: Ditto.

source/compatibility.tex
```diff
@@ -922,7 +922,7 @@
 Violating a non-throwing dynamic exception specification
 calls \tcode{terminate}
 rather than \tcode{unexpected}
-and might not perform stack unwinding prior to such a call.
+and it is unspecified whether stack unwinding is performed prior to such a call.
```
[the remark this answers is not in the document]
tkoeppe: Done

source/compatibility.tex
tkoeppe: Agreed, separate PR.

tkoeppe:
Thanks a lot, @opensdh!

source/compatibility.tex
```diff
@@ -1300,8 +1300,9 @@
 \rationale
 C99 compatibility.
 \effect
-Certain integer literals larger than can be represented by \tcode{long} could
-change from an unsigned integer type to \tcode{signed long long}.
+Certain integer literals
```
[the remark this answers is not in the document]
tkoeppe: I was thinking that, too, but couldn't the type also be something else, e.g. `unsigned long long`, or an extended integer type?

source/compatibility.tex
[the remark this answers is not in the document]
tkoeppe: Can we get rid of "to avoid leaks"? Destruction is important for so many more reasons.

source/compatibility.tex
tkoeppe: I'll leave this for the future.

source/compatibility.tex
tkoeppe: I mourn the loss of our last instance of "disaster".

source/compatibility.tex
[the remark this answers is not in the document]
opensdh: That's true: the `unsigned long long` is of course not a change "from an unsigned integer type", but a relevant extended integer type could be signed.  Presumably Core should review this for accuracy.

source/compatibility.tex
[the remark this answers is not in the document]
opensdh: In that case, we can just say "as would be necessary to destroy them".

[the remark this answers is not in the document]
tkoeppe: Done. Separate PR is coming; I'm also adding your 1-15 suggestions.

source/compatibility.tex
opensdh: We've had only that one instance since Git history began, but it is nice to finally be able to say that C++ isn't a disaster.

tkoeppe: At least we will always have major catastrophe.

PR #4393, opened 2020-11-25
Rewordings to avoid "might" and "could"

tkoeppe:
These rewordings were proposed by @opensdh during earlier review. They are more extensive than local replacements of "might" or "could", and we should review them on their own merit, i.e. whether they are an improvement (or at least not making things worse), and we should also consider other (possible even wider-ranging) rewordings.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -2870,7 +2870,7 @@ that is not an odr-use\iref{term.odr.use},
 or defines a constexpr variable initialized to a TU-local value (defined below).
 \begin{note}
 An inline function template can be an exposure even though
-explicit specializations of it might be usable in other translation units.
+certain explicit specializations of it would be usable in other translation units.
 \end{note}
 
 \pnum
@@ -2965,7 +2965,7 @@ namespace N {
 }
 void adl(double);
 
-inline void h(auto x) { adl(x); }   // OK, but a specialization might be an exposure
+inline void h(auto x) { adl(x); }   // OK, but certain specializations are exposures
 \end{codeblocktu}
 \begin{codeblocktu}{Translation unit \#2}
 module A;
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -631,11 +631,11 @@ reversed and rewritten candidates.
 Improve consistency of equality with three-way comparison
 and make it easier to write the full complement of equality operations.
 \effect
-Equality and inequality expressions between two objects of different types,
-where one is convertible to the other,
-could invoke a different operator.
-Equality and inequality expressions between two objects of the same type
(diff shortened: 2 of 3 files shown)


tkoeppe:
@zygoloid, @jensmaurer, @opensdh: thoughts on this PR would be most welcome: I'd be happy to apply as much or as little of this as we agree would make an improvement.

jensmaurer:
Let's not rush it for the mailing deadline.

tkoeppe:
@jensmaurer, @opensdh: I'd welcome feedback on whether any of these changes sound like clearn improvements.

source/basic.tex
```diff
@@ -2870,7 +2870,8 @@
 or defines a constexpr variable initialized to a TU-local value (defined below).
 \begin{note}
 An inline function template can be an exposure even though
-explicit specializations of it might be usable in other translation units.
+some
+explicit specializations of it would be usable in other translation units.
```
opensdh, line 2874: Maybe "certain" instead of "some"?  We don't want to suggest that such explicit specializations always exist, merely that they, er, _could_.

source/compatibility.tex
```diff
@@ -2411,8 +2409,8 @@
 with a type.
 In \Cpp{}, class members can be declared with the \keyword{static} storage
 class specifier.
-Allowing storage class specifiers on type
-declarations could render the code confusing for users.
+Storage class specifiers on type
+declarations can render the code confusing for users.
```
opensdh, line 2413: Maybe just "can be confusing", since we're now talking about the specifiers themselves rather than allowing them?

source/exceptions.tex
```diff
@@ -799,9 +799,7 @@
 \end{codeblock}
 The call to
 \tcode{f}
-is well-formed even though, when called,
-\tcode{f}
-might throw an exception.
+is well-formed even though it is not known whether \tcode{f} throws any exception.
```
opensdh, line 802: Since it's well-formed even if `f` definitely throws an exception (_cf._ the direct use of `throw`), we should say something like "is well-formed despite the possibility that, when called, &hellip;" or "is well-formed despite the possibility that the call to `f` throws an exception".

opensdh (approved):
I don't know how much an approval means from me who initially suggested (some of) these.

source/exceptions.tex
tkoeppe: How about "... despite the possibility for `f` to thrown an exception", or "... for the call to `f` to throw an exception"?

tkoeppe: Even better, "the call to `f` is already the subject, so just "it" suffices.

PR #4396, opened 2020-11-28
[temp.over.link] Remove (apparent) copy/paste extra lines from example code

cuzdav:
The first two lines in this example were an exact copy from the
previous example (about 40 lines up), and seem unrelated to the
topic being demonstrated, so look like a cut/paste error.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -3748,9 +3748,6 @@ same function template differ in the result of this name lookup, the
 result for the first declaration is used.
 \begin{example}
 \begin{codeblock}
-template <int I, int J> void f(A<I+J>);         // \#1
-template <int K, int L> void f(A<K+L>);         // same as \#1
-
 template <class T> decltype(g(T())) h();
 int g(int);
 template <class T> decltype(g(T())) h()         // redeclaration of \tcode{h()} uses the earlier lookup\ldots


cpplearner:
The example demonstrates that

> Two expressions involving template parameters are considered equivalent if two function definitions containing the expressions would satisfy the one-definition rule, except that the tokens used to name the template parameters may differ as long as a token used to name a template parameter in one expression is replaced by another token that names the same template parameter in the other expression.

I don't think it's an error.

cuzdav:
It seems to me you're taking the text describing example 2, while my proposed change is for example 3 (which is about dependent names).  _Both_ examples 2 and 3 contain these exact same lines, and it is correct in example 2, but appears to be copy-paste leftover in example 3.

> For determining whether two dependent names\iref{temp.dep} are
equivalent, only the name itself is considered, not the result of name
lookup in the context of the template. If multiple declarations of the
same function template differ in the result of this name lookup, the
result for the first declaration is used.

tkoeppe:
@jensmaurer, @opensdh: could you kindly take a look? The original wording seems to have been in place since the initial commit.

jensmaurer:
I agree this repeats parts of the earlier example, but it seems as relevant to the given paragraph ("equivalent") than to the previous one. Maybe the second line of the previous example should actually be removed, because the paragraph specifies that dependent expressions are part of the signature (difference between + and - in the return type), not that template parameter renames are benign.

cuzdav:
My apologies for imprecise phrasing.  It _looked_ enough like the previous
example that I thought it was a cut and paste issue, but you are correct
and my confusion was partially my own fault.  However, it does seem that
there's room for improvement, so thanks for taking time to look at this.

tkoeppe:
@cuzdav: could you make the suggested change, or would you like me to?

PR #4405, opened 2020-12-03
Use \keyword pervasively

jensmaurer:


source/basic.tex
```diff
@@ -5145,8 +5145,8 @@
 \indextext{type!integral}%
 Types
 \tcode{bool},
```
johelegp, line 5147: What's the intrinsic? Why not all `\tcode{`_`X`_`}`, where _`X`_ is a [_keyword_](http://wg21.link/lex.key#nt:keyword)?

source/exceptions.tex
```diff
@@ -210,8 +210,8 @@
 can be caught by a
 \grammarterm{handler}
 of
-\tcode{const}
-\tcode{char*}
+\keyword{const}
+\tcode{\keyword{char}*}
```
johelegp, line 214: Above `char` remains with `\tcode`. I agree with @jwakely with regards to `\libconcept` that this makes reviews harder. Have you considered simply using `\tcode` and `\codeblock` to delimit C++ code and use a custom tool that adds all the bells and whistles to the LaTeX?

source/basic.tex
jensmaurer: All keywords should be marked `\keyword`. This way, we can make them a different color sometime in the future. You also get an index entry.

source/exceptions.tex
jensmaurer: Huh? It does not. This change adds `\keyword` around `char`.

I do agree that adding \libconcept makes reading the source harder for codeblocks, but this is actually outside of a codeblock. (For listings = codeblocks, there is already keyword autodetection; maybe we can also have concept autodetection there and add \libconcept automatically.)

source/lex.tex
```diff
@@ -464,22 +464,22 @@
...
+\keyword{xor}           &   \tcode{\caret}     &
+\keyword{not}           &   \tcode{!}          \\ \rowsep
 \tcode{\%:}             &   \tcode{\#}         &
-\tcode{compl}           &   \tcode{\~}         &
-\tcode{not_eq}          &   \tcode{!=}         \\ \rowsep
+\keyword{compl}         &   \tcode{\~}         &
+\keyword{not_eq}        &   \tcode{!=}         \\ \rowsep
 \tcode{\%:\%:}          &   \tcode{\#\#}       &
-\tcode{bitand}          &   \tcode{\&}         &
+\keyword{bitand}        &   \tcode{\&}         &
```
johelegp, line 482: These are not keywords.

jensmaurer: They're close enough to being keywords for the purposes of indexing and highlighting, I think.

tkoeppe:
@jensmaurer If you rebase this, I think this can land -- it has no visual effect, right?

jensmaurer:
@tkoeppe: Rebased. Confirmed no visual effect, except one strange false positive / spacing diff shown by diffpdf.

jensmaurer:
@tkoeppe, we should consider squashing this.

source/basic.tex
```diff
@@ -6381,12 +6382,12 @@
...
 and it shall have a declared return type of type
-\tcode{int}, but otherwise its type is \impldef{parameters to \tcode{main}}.
+\keyword{int}, but otherwise its type is \impldef{parameters to \tcode{main}}.
 \indextext{\idxcode{main} function!implementation-defined parameters to}%
 An implementation shall allow both
 \begin{itemize}
-\item a function of \tcode{()} returning \tcode{int} and
-\item a function of \tcode{(int}, pointer to pointer to \tcode{char)} returning \tcode{int}
+\item a function of \tcode{()} returning \keyword{int} and
+\item a function of \tcode{(\keyword{int}}, pointer to pointer to \keyword{char}) returning \keyword{int}
```
tkoeppe, line 6390: Missing `\tcode` around the closing `)`.

tkoeppe (changes requested):

tkoeppe:
I extraced a nesting fix into b0a253cc6ba0e7f556f5afefab94f7cc1d1e5ced; please rebase.

tkoeppe:
Well, I can actually just fix this myself, so leave it with me.

source/basic.tex
jensmaurer: fixed

jensmaurer:
Hm... That situation is actually fine in my copy. But apparently, I need to rebase.

tkoeppe:
Hm, you lost the `tcode` around the closing paren. I'd reinstate `\tcode{\keyword{int})`.

jensmaurer:
Are we talking about the same thing? I have

`a function of \tcode{(\keyword{int}}, pointer to pointer to \tcode{\keyword{char})} returning \keyword{int}`

jensmaurer:
(These are parens around a parameter-type-list; the closing paren is at the end.)

tkoeppe:
The "function of" issue was a problem with your PR, which you have now fixed. The const_cast nesting issue was pre-existing, and your PR had fixed it, but I extraced it into a separate commit (since it has nothing to do with `\keyword`).

jensmaurer:
Yes, and I've rebased/force-pushed and fixed the "function of" issue. Are there any other issues?

tkoeppe:
No, that should be it.

tkoeppe (approved):

PR #4406, opened 2020-12-03
[cpp.import] Clarify header units as the source of macro definitions.

### LINKED ISSUE #4404: [cpp.import] Use "header unit" where appropriate
http://lists.isocpp.org/modules/2020/12/0885.php

Don't talk about "translation units" where "header unit" is adequate. Only header units can export macros.

jensmaurer:
Fixes #4404

### DIFF
--- a/source/preprocessor.tex
+++ b/source/preprocessor.tex
@@ -811,18 +811,22 @@ in the \tcode{import} \grammarterm{control-line}
 are processed just as in normal text
 (i.e., each identifier currently defined as a macro name
 is replaced by its replacement list of preprocessing tokens).
+\begin{note}
 An \tcode{import} directive
 matching the first two forms of a \grammarterm{pp-import}
 instructs the preprocessor to import macros
 from the header unit\iref{module.import}
-denoted by the \grammarterm{header-name}.
+denoted by the \grammarterm{header-name},
+as described below.
+\end{note}
 \indextext{point of!macro import|see{macro, point of import}}%
 The \defnx{point of macro import}{macro!point of import} for the
 first two forms of \grammarterm{pp-import} is
 immediately after the \grammarterm{new-line} terminating
 the \grammarterm{pp-import}.
 The last form of \grammarterm{pp-import} is only considered
-if the first two forms did not match.
+if the first two forms did not match, and
+does not have a point of macro import.
 
 \pnum
 If a \grammarterm{pp-import} is produced by source file inclusion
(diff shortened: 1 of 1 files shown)


tkoeppe:
@zygoloid: could you kindly review this?

source/preprocessor.tex
```diff
@@ -811,11 +811,14 @@
...
 matching the first two forms of a \grammarterm{pp-import}
 instructs the preprocessor to import macros
 from the header unit\iref{module.import}
-denoted by the \grammarterm{header-name}.
+denoted by the \grammarterm{header-name},
+as described below.
+\end{note}
 \indextext{point of!macro import|see{macro, point of import}}%
 The \defnx{point of macro import}{macro!point of import} for the
 first two forms of \grammarterm{pp-import} is
```
zygoloid, line 824: For completeness, should we say "The last form of *pp-import* is only considered if the first two forms did not match<ins>, and does not have a point of macro import</ins>." below?

source/preprocessor.tex
```diff
@@ -876,16 +877,18 @@
 if the macro name is not lexically identical to a keyword\iref{lex.key}
 or to the \grammarterm{identifier}{s} \tcode{module} or \tcode{import},
 the first point
-of macro import of a translation unit containing a point of definition for the
-macro definition, if any (in any other translation unit).
+of macro import of a header unit containing a point of definition for the
+macro definition, if any.
+In the latter case, the macro is said
+to be \defnx{imported}{macro!import} from the header unit.
```
zygoloid, line 883: I wonder if we can make this a little clearer by separating the two cases with bullets and moving the parenthesized text to the front:

> The *point of definition* of a macro definition within a translation unit <ins>T</ins> is:
>  * <ins>if the `#define` directive of the macro definition occurs within T,</ins> the point at which <del>its `#define`</del> <ins>that</ins> directive occurs <del>(in the translation unit containing the `#define` directive)</del>, or <ins>otherwise</ins>,
>  * if the macro name is not lexically identical to a keyword or to the *identifier*s `module` or `import`, the first point of macro import <ins>in T</ins> of a <del>translation</del> <ins>header</ins> unit containing a point of definition for the macro definition, if any.

zygoloid (approved):
I think this is correct and an improvement. Some possible further improvements suggested but I don't think they're essential.

source/preprocessor.tex
jensmaurer: fixed

source/preprocessor.tex
jensmaurer: fixed

zygoloid (approved):

jensmaurer:
@tkoeppe, please merge per http://lists.isocpp.org/core/2021/03/10570.php

tkoeppe:
Done, thanks!

PR #4408, opened 2020-12-05
Fix up the wording of one note, and add a second note, on implicit move.

Quuxplusone:
These notes probably should have been changed when P1825 was adopted.

I'm not sure if this is editorial.

Also, I disagree with the use of the word "can" in the existing note; it seems to me that this is *not* a choice the user gets to make, and so this was an incorrect search-and-replace in e5455e3bb — but maybe this is just ISO style now, I dunno.

### DIFF
--- a/source/statements.tex
+++ b/source/statements.tex
@@ -758,11 +758,10 @@ the \tcode{return} statement initializes the
 glvalue result or prvalue result object of the (explicit or implicit) function call
 by copy-initialization\iref{dcl.init} from the operand.
 \begin{note}
-A \tcode{return} statement can involve
-an invocation of a constructor to perform a copy or move of the operand
+A \tcode{return} statement can involve a copy or move of the operand
 if it is not a prvalue or if its type differs from the return type of the function.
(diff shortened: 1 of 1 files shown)


jensmaurer:
This is mostly fixed with #3998, I believe.

jensmaurer (changes requested):
Please make sure the commit description is properly formatted.
It should start with the [section.label] and then describe the change.

For this particular change, maybe it's less relevant that we changed two notes,
but more relevant that we now highlight implicitly movable entities.
Suggestion: "Fix and add notes about implicit moving in co_/return statements"

tkoeppe:
Please reopen (and rebase) if you'd like to continue working on this.

Quuxplusone:
I'm not actually sure how to reopen; but I have rebased now (and updated the commit summary line).

tkoeppe:
Oh, I see, the reopen button isn't working for me either. Odd. Did you "recreate" the branch? In that case maybe you could start a new pull request off that branch?

Quuxplusone:
Okay, opened #5783!

PR #4416, opened 2020-12-14
[class.mem.general] Allow non-defining declarations of nested classes.

### LINKED ISSUE #4409: [class.mem.general]/12 seems to conflict with [class.nest]/3
[class.mem.general]/12:

> The *member-declarator-list* can be omitted only after a *class-specifier* or an *enum-specifier* or in a friend declaration.

[class.nest]/3:

> If class `X` is defined in a namespace scope, a nested class `Y` may be declared in class `X` and later defined in the definition of class `X` or be later defined in a namespace scope enclosing the definition of class `X`.

Consider

```
class E {
  class I1;                     // forward declaration of nested class
};
```

AFAIK, `class I1;` is a *member-declaration* with an *elaborated-type-specifier* and no *member-declarator-list*. According to [class.mem.general]/12, this is not permitted because there's no *class-specifier* or *enum-specifier* or `friend`. But [class.nest]/3 seems to suggest that it's valid. (How else can one declare but not define a class name in a class scope?)

I believe it's obvious that the example in question is valid, and [class.mem.general]/12 needs to account for *elaborated-type-specifier*. (Maybe it should use similar wording as in [dcl.pre]/5.)

jensmaurer:
Fixes #4409

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -666,11 +666,19 @@ The \grammarterm{decl-specifier-seq} may be omitted in constructor, destructor,
 and conversion function declarations only;
 when declaring another kind of member the \grammarterm{decl-specifier-seq}
 shall contain a \grammarterm{type-specifier} that is not a \grammarterm{cv-qualifier}.
-The
-\grammarterm{member-declarator-list} can be omitted only after a
-\grammarterm{class-specifier} or an \grammarterm{enum-specifier} or in a
-friend declaration\iref{class.friend}. A
-\grammarterm{pure-specifier} shall be used only in the declaration of a
+The \grammarterm{member-declarator-list} can be omitted only
+after
+\begin{itemize}
+\item
+a \grammarterm{class-specifier},
+\item
+an \grammarterm{enum-specifier}, or
+\item
+an \grammarterm{elaborated-type-specifier}
+that is the sole constituent of a declaration\iref{dcl.type.elab}, or
+\end{itemize}
+in a friend declaration\iref{class.friend}.
+A \grammarterm{pure-specifier} shall be used only in the declaration of a
 virtual function\iref{class.virtual}
 that is not a friend declaration.
 


tkoeppe:
OK, but I'm not 100% sure this is editorial. Could we get CWG sign-off on this, or at least a second pair of core eyes?

jensmaurer:
@zygoloid , could you please take a look?

source/classes.tex
```diff
@@ -666,11 +666,19 @@
 and conversion function declarations only;
 when declaring another kind of member the \grammarterm{decl-specifier-seq}
 shall contain a \grammarterm{type-specifier} that is not a \grammarterm{cv-qualifier}.
-The
-\grammarterm{member-declarator-list} can be omitted only after a
-\grammarterm{class-specifier} or an \grammarterm{enum-specifier} or in a
-friend declaration\iref{class.friend}. A
-\grammarterm{pure-specifier} shall be used only in the declaration of a
+The \grammarterm{member-declarator-list} can be omitted only
+after
```
zygoloid, line 670: I think "can be omitted only after" is problematic phrasing -- it's unclear whether it's giving permission, stating a restriction, or both. (Can a *member-declarator-list* be omitted after *any* *elaborated-type-specifier*, or only some of them?) This is made a little worse by this change, because three of the four top-level grammar productions for *elaborated-type-specifier* aren't allowed in this context, per the end of [dcl.type.elab]p1. This might help:

> The \grammarterm{member-declarator-list} shall not be omitted unless
> the declaration is a friend declaration or its sole constituent is
> * ...

zygoloid (commented):
I have no doubt that this change reflects the CWG intent, and that the wording and example in [class.nest]p3 are correct.

As for whether that's enough to apply this editorially without explicit CWG oversight... perhaps it's worth sending an email past the core reflector to check if anyone would prefer this to be handled non-editorially.

tkoeppe:
Thanks a lot! That seems clear, let's put this on the CWG list and get sign-off.

AlisdairM:
Did this ever get resolution on the Core reflector?

PR #4418, opened 2020-12-14
[class.copy.assign] Fix phrasing in note

### LINKED ISSUE #4388: [class.copy.assign]p8 needs rewording
From a could/might review, a suggestion by @zygoloid regarding [class.copy.assign]p8:

---

>  assignment operator with a parameter type that **could**/**can** be that of a

I don't like the old or new words here. I think we should rephrase this whole sentence.

I also wonder if this is actually a normative rule at all, or if this is a consequence of the rule in [class.copy.assign] talking about "explicitly declar[ing] a copy assignment operator". If it's a consequence, then it should be a note, and if not, then it shouldn't be phrased as "this happens because the declaration is not considered an explicit declaration of the operator" but instead as something like:

> A *using-declaration* that brings in an assignment operator from a base class never suppresses the implicit declaration of an assignment operator of the derived class, even if the base class assignment operator, when considered as a member of the derived class, is a copy assignment operator or a move assignment operator ([class.copy.assign]).

I think this is repairable editorially, but not by a one-word change.

jensmaurer:
to avoid 'could'.

Fixes #4388


tkoeppe:
@opensdh: could you take a look? This is wording proposed by @zygoloid.

tkoeppe:
(I like it and will apply it if there are no objections.)

source/classes.tex
```diff
@@ -1873,11 +1873,13 @@
...
-copy/move assignment operator for a
-derived class does not suppress the implicit declaration of the derived class
-operator\iref{namespace.udecl}.
+A \grammarterm{using-declaration}\iref{namespace.udecl}
+that brings in an assignment operator from a base class
+never suppresses the implicit declaration of
+an assignment operator of the derived class,
+even if the base class assignment operator,
+when considered as a member of the derived class,
+is a copy assignment operator or a move assignment operator.
```
opensdh, line 1882: I don't like the "is" because the assignment operator never is one of those things.  I already provided my own [variation](https://github.com/cplusplus/draft/issues/4388#issuecomment-735898621) that avoids that phrasing, but maybe it's inferior in some other way.

tkoeppe: Yes, I see what you mean, and I agree. Something like "would" is really what we'd need here.

tkoeppe: Merging the two suggestions, how about:

> A using-declaration in a derived class `C` that brings in an assignment operator from a base class never suppresses the implicit declaration of an assignment operator of `C`, even if the base class assignment operator would be a copy or move assignment operator if declared as a member of `C`.

opensdh: Looks fine, except that "brings in" is a reversion of P1787R6's "names", which says that *using-declaration*s do not copy-paste declarations around.  (Admittedly, this is (now) a note, but I don't want to encourage the old way of thinking.)

tkoeppe: Ah, I see. How about this then:

> A using-declaration in a derived class `C` that names an assignment operator from a base class never suppresses the implicit declaration of an assignment operator of `C`, even if the base class assignment operator would be a copy or move assignment operator if declared as a member of `C`.

opensdh: You can't name an assignment operator other than from a base class, but maybe it helps readability to explain it in more words like this.  In any case, ship it.

tkoeppe: You mean that "from a base class" is redundant? If so, then yes, that's true, but in a note this seems helpful explanation, since it's exactly the puzzle solving that would be needed in the absence of this explanation that the note tries to spell out. (We could also say "an assignment operator (from a base class)", or "an assignment operator, which must be from a base class". But we shan't.)

jensmaurer: Fixed

jensmaurer:
@tkoeppe, this looks ready.

tkoeppe:
@jensmaurer: Yeah, I had still been ponderin Davis's remark about the redundancy of saying "from a base class". I think it's helpful to say that, though, and I can't think of a way to make it more parenthetical. Absent any better ideas we can apply this.

jensmaurer:
@tkoeppe, any final thoughts here?

PR #4419, opened 2020-12-14
[dcl.array] Arrays of unknown bound are allowed in function parameters.

### LINKED ISSUE #4387: [dcl.array] p.7 may be outdated
Paragraph 7 of [dcl.array] (https://eel.is/c++draft/dcl.array#7) states:

> In addition to declarations in which an incomplete object type is allowed, an array bound may be omitted **in some cases** in the declaration of a function parameter ([dcl.fct]). An array bound may also be omitted when an object (but not a non-static data member) of array type is initialized and the declarator is followed by an initializer ([dcl.init], [class.mem], [expr.type.conv], [expr.new]). In these cases, the array bound is calculated from the number of initial elements (say, N) supplied ([dcl.init.aggr]), and the type of the array is “array of N U”.

The wording "an array bound may be omitted **in some cases** in the declaration of a function parameter", among other changes, was added by the resolution to [CWG619](http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#619).

Actually, the highlighted part ("in some cases") is the only difference between the [proposed resolution in February of 2010](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2010/n3026.html#619) compared to the [proposed resolution in March of 2010](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2010/n3083.html#619).

I believe that the only cases in which it was not allowed (at that time) to omit array bounds in the declaration of a function parameter were specified by [dcl.fct] p.8, which stated (in N3035, published in 2010):

> **If the type of a parameter includes a type of the form “pointer to array of unknown bound of T” or “reference
to array of unknown bound of T,” the program is ill-formed**. Functions shall not have a return type of
type array or function, although they may have a return type of type pointer or reference to such things.
There shall be no arrays of functions, although there can be arrays of pointers to functions.
Types shall not be defined in return or parameter types. The type of a parameter or the return type for a
function definition shall not be an incomplete class type (possibly cv-qualified) unless the function definition
is nested within the member-specification for that class (including definitions in nested classes defined within
the class).

However, the highlighted part of [dcl.fct] p.8 (forbidding pointers or references to arrays of unknown bound in the type of a parameter) was removed by the resolution to [CWG393](http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#393).

I do not know any cases in which, by the current specification, one cannot use arrays of unknown bound in the type of a function parameter. Are there any such cases that I am overlooking?

jensmaurer:
Fixes #4387

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -3192,7 +3192,7 @@ and numbered \tcode{0} to \tcode{N-1}.
 
 \pnum
 In addition to declarations in which an incomplete object type is allowed,
-an array bound may be omitted in some cases in the declaration of a function
+an array bound may be omitted in the declaration of a function
 parameter\iref{dcl.fct}.
 An array bound may also be omitted
 when an object (but not a non-static data member) of array type is initialized


tkoeppe:
@jensmaurer: so you're sure that there are cases left now where arrays of unknown bound aren't allowed?

jensmaurer:
In a function parameter, I think all add-on restrictions are gone. Unless you feel the phrasing is still too vague to handle
```
void f(C<sizeof(T[])> x);
```
(This isn't allowed, because we're applying sizeof to an incomplete type, but the phrasing "in a function parameter" might suggestion it's giving permission to write something like that.)

tkoeppe:
Good find. OK, let's think about this some more.

PR #4425, opened 2020-12-23
[time.zone.leap.nonmembers] Fix requires-clause indentation

johelegp:


jensmaurer:
"requires" is aligned the same as the return type; see `<algorithm>`.

johelegp:
#4428 fixes the one I thought was right.

PR #4426, opened 2020-12-24
[range.iota.iterator,range.iota.sentinel] Complete iota_view's requires-clause

johelegp:
Completes https://github.com/cplusplus/draft/commit/ae15dd1b9116975f9f6d71955802536551d95d59.


jensmaurer (approved):

tkoeppe:
I appreciate this, but is this editorial? The LWG issue doesn't contain this change, what makes this an "obvious oversight"?

@jwakely: could you perhaps provide an LWG perspective on this PR?

jwakely (approved):
These changes should have been part of the LWG issue. The extra constraint was added to the primary template, and so should also be present on the out-of-class definitions of the nested members of that class.

tkoeppe:
Thanks!

PR #4436, opened 2020-12-27
[localization] Reformat LaTeX source code.

jensmaurer:
Also add two missing periods and remove one instance
of superfluous whitespace.

### DIFF
--- a/source/locales.tex
+++ b/source/locales.tex
@@ -7,14 +7,17 @@
 This Clause describes components that \Cpp{} programs may use to
 encapsulate (and therefore be more portable when confronting)
 cultural differences.
-The locale facility includes internationalization
-support for character classification and string collation, numeric,
-monetary, and date/time formatting and parsing, and message retrieval.
+The locale facility includes
+internationalization support for character classification and string collation,
+numeric, monetary, and date/time formatting and parsing, and
+message retrieval.
 
 \pnum
 The following subclauses describe components for
-locales themselves, the standard facets, and facilities
-from the ISO C library, as summarized in \tref{localization.summary}.
+locales themselves,
+the standard facets, and
+facilities from the ISO C library,
+as summarized in \tref{localization.summary}.
 
 \begin{libsumtab}{Localization library summary}{localization.summary}
 \ref{locales} & Locales                   &   \tcode{<locale>}    \\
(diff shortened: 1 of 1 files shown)


source/locales.tex
```diff
@@ -2573,64 +2266,62 @@
...
+\item
+For a signed integer value, the function \tcode{strtoll}.
+\item
+For an unsigned integer value, the function \tcode{strtoull}.
+\item
+For a \tcode{float} value, the function \tcode{strtof}.
+\item
+For a \tcode{double} value, the function \tcode{strtod}.
+\item
+For a \tcode{long double} value, the function \tcode{strtold}.
```
johelegp, line 2308: Why not indent each `\item` text by 2 spaces [as required](https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines), or just remove the original blank lines between items?

jensmaurer: I don't think we've made up our mind on `\item` source formatting just yet.

source/locales.tex
```diff
@@ -4605,82 +4067,41 @@
...
-\tcode{char},
-rather than an array of
-\tcode{part},
-is specified for
-\tcode{pattern::field}
-purely for efficiency.
+An array of \tcode{char},
+rather than an array of \tcode{part},
+is specified for \tcode{pattern::field} purely for efficiency.
 \end{footnote}
```
johelegp, line 4080:
```suggestion
\end{footnote}.
```

source/locales.tex
```diff
@@ -4605,82 +4067,41 @@
...
-such that the
-\tcode{part}
-value
-\tcode{static_cast<part>(p.field[i])}
-determines the
-$\tcode{i}^\text{th}$
-component of the format
+specified by a \tcode{pattern} value \tcode{p},
+such that the \tcode{part} value \tcode{static_cast<part>(p.field[i])}
+determines the $\tcode{i}^\text{th}$ component of the format
```
johelegp, line 4075:
```suggestion
\end{footnote}.
```

source/locales.tex
```diff
@@ -2573,64 +2266,62 @@
...
-\item the most positive representable value,
+\item
+the most positive representable value,
 if the field to be converted to an unsigned integer type represents a value
 that cannot be represented in \tcode{val}.
-
-\item the converted value, otherwise.
+\item
+the converted value, otherwise.
 \end{itemize}
```
johelegp, line 2325: The capitalization here needs fixing.

jensmaurer: Not in this patch.

source/locales.tex
tkoeppe: @zygoloid was ultimately against formatting that was difficult to relocate or regroup, so I think we may be moving away from that rule.

PR #4440, opened 2020-12-28
[system.error.syn] Specify the what: templates

johelegp:


source/diagnostics.tex
```diff
@@ -825,7 +825,7 @@
 exposes the \libheader{cerrno} macros is unspecified.
 
 \pnum
-The \tcode{is_error_code_enum} and \tcode{is_error_condition_enum} may be
+Templates \tcode{is_error_code_enum} and \tcode{is_error_condition_enum} may be
```
jensmaurer, line 828: The most straightforward fix is to assume a dropped word "templates" in the right spot:
```suggestion
The \tcode{is_error_code_enum} and \tcode{is_error_condition_enum} templates may be
```

jensmaurer (changes requested):

source/diagnostics.tex
johelegp: Fixed.

jensmaurer (approved):

PR #4448, opened 2021-01-03
[lib] Ensure non-member swap is declared in the header synopsis only

### LINKED ISSUE #4041: [lib] Non-member swap inconsistently specified in class synopsis
| In header synopsis only | In class synopsis |
| ------------------ | ----------------- |
| pair, tuple, optional, variant, any |
| {unique,shared}_ptr | weak_ptr |
| | function |
| array, stack | deque, forward_list, list, vector-not-of-bool, (multi){map,set}, unordered_(multi){map,set}, (priority_)queue |
| valarray |
| | basic_{stringbuf,(i,o)stringstream,filebuf,(i,o)fstream,syncbuf} **(only)** |
| fs::path |
| basic_regex, match_results |
| thread |
| | {unique,shared}_lock |
| | promise, packaged_task

The remaining are friends or depend on the generic `std::swap` via ADL.

jensmaurer:
Fixes #4041

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -3767,11 +3767,6 @@ namespace std {
   template<class InputIterator, class Allocator = allocator<@\placeholder{iter-value-type}@<InputIterator>>>
     deque(InputIterator, InputIterator, Allocator = Allocator())
       -> deque<@\placeholder{iter-value-type}@<InputIterator>, Allocator>;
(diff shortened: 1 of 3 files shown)


tkoeppe:
Good, finally something to _reduce_ the page count.

PR #4451, opened 2021-01-04
[basic.stc.dynamic] Rework presentation

### LINKED ISSUE #3315: [basic.stc.dynamic.deallocation] Use of incorrect grammar
[[basic.stc.dynamic.deallocation] p3](http://eel.is/c++draft/basic.stc.dynamic.deallocation#3) says:
> A deallocation function may be an instance of a function template.

and 

> A template instance is never a usual deallocation function, regardless of its signature.

"instance of a function template" here presumably should just say "function template" to match deallocation functions. Additionally, "template instance" is not a defined term in the standard, presumably it should be "function template specialization".

jensmaurer:
Clarify syntactic vs. semantic constraints.
Split off treatment of global (de)allocation functions.

Fixes #3315

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3809,71 +3809,12 @@ execution\iref{intro.execution}, using
 \indextext{\idxcode{new}}%
 \grammarterm{new-expression}{s}\iref{expr.new}, and destroyed using
 \indextext{\idxcode{delete}}%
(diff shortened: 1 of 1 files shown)


jensmaurer:
@tkoeppe, how would you like to proceed here?

tkoeppe:
@jensmaurer: interested, but maybe to be discussed in a meeting.

PR #4454, opened 2021-01-05
[container.requirements] Improve indexing of members described in requirements

### LINKED ISSUE #1425: Indexing issues in [containers]
1. There are many member declarations that are not indexed using `\indexlibrarymember` but simply `\indexlibrary`.
2. Only the requirement tables in [unord.req] are indexed for member names, while other requirement tables are not indexed for member names at all.

jensmaurer:
Fixes #1425

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -80,6 +80,24 @@ In Tables~\ref{tab:container.req},
 \tcode{r} denotes a non-const value of type \tcode{X}, and
 \tcode{rv} denotes a non-const rvalue of type \tcode{X}.
 
+% Local command to index names as members of all containers.
+\newcommand{\indexcont}[1]{%
+\indexlibrarymisc{#1}{containers}%
+\indexlibrarymemberx{array}{#1}%
+\indexlibrarymemberx{deque}{#1}%
+\indexlibrarymemberx{forward_list}{#1}%
+\indexlibrarymemberx{list}{#1}%
+\indexlibrarymemberx{vector}{#1}%
+\indexlibrarymemberx{map}{#1}%
+\indexlibrarymemberx{set}{#1}%
+\indexlibrarymemberx{multiset}{#1}%
+\indexlibrarymemberx{multimap}{#1}%
+\indexlibrarymemberx{unordered_map}{#1}%
+\indexlibrarymemberx{unordered_set}{#1}%
+\indexlibrarymemberx{unordered_multiset}{#1}%
+\indexlibrarymemberx{unordered_multimap}{#1}%
+}
+
 \begin{libreqtab5}
 {Container requirements}
 {container.req}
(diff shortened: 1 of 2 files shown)


source/containers.tex
```diff
@@ -824,6 +903,23 @@
...
+\indexlibrarymemberx{list}{#1}%
+\indexlibrarymemberx{vector}{#1}%
+\indexlibrarymemberx{map}{#1}%
+\indexlibrarymemberx{set}{#1}%
+\indexlibrarymemberx{multiset}{#1}%
+\indexlibrarymemberx{multimap}{#1}%
+\indexlibrarymemberx{unordered_map}{#1}%
+\indexlibrarymemberx{unordered_set}{#1}%
+\indexlibrarymemberx{unordered_multiset}{#1}%
+\indexlibrarymemberx{unordered_multimap}{#1}%
```
johelegp, line 920: Won't this index the sequence container members as members of these non-sequence containers?

source/containers.tex
```diff
@@ -617,6 +676,22 @@
 \tcode{t} denotes an lvalue or a const rvalue of type \tcode{X}, \tcode{rv} denotes a
 non-const rvalue of type \tcode{X}, and \tcode{m} is a value of type \tcode{A}.
 
+% Local command to index names as members of all containers.
+\renewcommand{\indexcont}[1]{%
+\indexlibrarymisc{#1}{allocator-aware containers}%
+\indexlibrarymemberx{deque}{#1}%
+\indexlibrarymemberx{list}{#1}%
```
johelegp, line 683:
```suggestion
\indexlibrarymemberx{forward_list}{#1}%
\indexlibrarymemberx{list}{#1}%
```

source/containers.tex
jensmaurer: Yes. Fixed.

source/containers.tex
jensmaurer: fixed

source/containers.tex
```diff
@@ -80,6 +80,24 @@
 \tcode{r} denotes a non-const value of type \tcode{X}, and
 \tcode{rv} denotes a non-const rvalue of type \tcode{X}.
 
+% Local command to index names as members of all containers.
+\newcommand{\indexcont}[1]{%
```
tkoeppe, line 84: It's a little bit brittle to have one "newcommand" and several "renewcommand"s; that could easily be missed. Maybe make them all "renewcommand", and add an intial (error-causing?) placeholder "newcommand"?

tkoeppe (changes requested):

source/containers.tex
jensmaurer: I understand the concerns about symmetry here, but what can possibly go wrong? If someone uses the macro before it's defined -> error. If someone uses the macro in a bad context -> tough luck.

Neither situation is helped with "renewcommand".

tkoeppe: Hm, I was concerned about silent changes of behaviour when we move blocks of text around, but you're right that renewcommand wouldn't help with that.

PR #4455, opened 2021-01-05
fix: small typo errors

0xflotus:


tkoeppe:
I appreciate the effort, but we cannot touch published N-numbered papers; that would cause far more confusion and uncertainty than the fix would be worth.

PR #4457, opened 2021-01-06
[expr.prim.req.type] Clarify example comment re: validity of name only

hubert-reinterpretcast:
The comment that a class template specialization is required can be taken as meaning that more is required than the validity of the _template-id_ (which is all that the requirement really checks).

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -2737,7 +2737,8 @@ template<typename T> using Ref = T&;
 
 template<typename T> concept C = requires {
   typename T::inner;    // required nested member name
-  typename S<T>;        // required class template specialization
+  typename S<T>;        // required valid~(\ref{temp.names}) \grammarterm{template-id};
+                        // fails if \tcode{T::type} does not exist as a type to which \tcode{0} can be implicitly converted
   typename Ref<T>;      // required alias template substitution, fails if \tcode{T} is void
 };
 \end{codeblock}


source/expressions.tex
```diff
@@ -2737,7 +2737,7 @@
 
 template<typename T> concept C = requires {
   typename T::inner;    // required nested member name
-  typename S<T>;        // required class template specialization
+  typename S<T>;        // required valid \grammarterm{template-id}
```
jensmaurer, line 2740: [temp.names] defines what it means for a _template-id_ to be valid. Do we mean this here? If so, we should add a cross-reference.

Further, this is presumably invalid for T==int, because `int::type` is nonsense. Can we say that, to parallel the next line?

jensmaurer (changes requested):

source/expressions.tex
```diff
@@ -2737,7 +2737,8 @@
 
 template<typename T> concept C = requires {
   typename T::inner;    // required nested member name
-  typename S<T>;        // required class template specialization
+  typename S<T>;        // required valid~(\ref{temp.names}) \grammarterm{template-id};
+                        // fails if substitution failure occurs for the second template parameter of \tcode{S}
```
jensmaurer, line 2741: Thanks for the cross-reference.
"substitution failure" feels too unspecific here, because it could mean substitution failure while instantiating all of `S<T, X>` ("substitution for" is not really "substitution into"). What we mean is substitution failure into the template-head. Suggestion, a little more tailored to the example:
"fails if \tocde{T} has no nested type \tcode{type}"

jensmaurer (changes requested):

source/expressions.tex
hubert-reinterpretcast: > "substitution failure" feels too unspecific here, because it could mean substitution failure while instantiating all of `S<T, X>` ("substitution for" is not really "substitution into").

Okay, I can change it to say "substitution into".

> What we mean is substitution failure into the template-head. Suggestion, a little more tailored to the example:
> "fails if \tocde{T} has no nested type \tcode{type}"

Well, it also fails if `0` cannot be implicitly converted to `T::type`.

jensmaurer (approved):

jensmaurer:
(needs squashing)

PR #4458, opened 2021-01-06
[specialized.algorithms] Qualify declarator-id with sub-namespace.

### LINKED ISSUE #4253: [special.mem.concepts] Adjust presentation to use ranges:blah
in \itemdecl code blocks.

jensmaurer:
As decided in #2656, return types from the sub-namespace were qualified as well.
"namespace ranges { ... }" was excised from \itemdecl.

Partially addresses #2512
Fixes #4253

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -10368,14 +10368,12 @@ for (; first != last; ++first)
 
 \indexlibraryglobal{uninitialized_default_construct}%
 \begin{itemdecl}
-namespace ranges {
-  template<@\placeholdernc{nothrow-forward-iterator}@ I, @\placeholder{nothrow-sentinel-for}@<I> S>
-    requires @\libconcept{default_initializable}@<iter_value_t<I>>
-    I uninitialized_default_construct(I first, S last);
-  template<@\placeholdernc{nothrow-forward-range}@ R>
-    requires @\libconcept{default_initializable}@<range_value_t<R>>
-    borrowed_iterator_t<R> uninitialized_default_construct(R&& r);
-}
+template<@\placeholdernc{nothrow-forward-iterator}@ I, @\placeholder{nothrow-sentinel-for}@<I> S>
+  requires @\libconcept{default_initializable}@<iter_value_t<I>>
+  I ranges::uninitialized_default_construct(I first, S last);
+template<@\placeholdernc{nothrow-forward-range}@ R>
+  requires @\libconcept{default_initializable}@<range_value_t<R>>
+  borrowed_iterator_t<R> ranges::uninitialized_default_construct(R&& r);
 \end{itemdecl}
 
 \begin{itemdescr}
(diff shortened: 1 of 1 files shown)


johelegp:
> As decided in #2656, return types from the sub-namespace were qualified as well.

That sounds as if you qualified all those actually in `ranges::`. But you only did it for the *`_result` templates. Granted, the editorial meeting decision in the linked issue says
> - If there is a possibility of confusion, qualify declarator-id and return type etc.

Are the \*`_result` more confusing than `borrowed_`*`_t`, which weren't qualified?

Although consistent with the second decision
> - Otherwise, no change.

I noticed that `range_`*`_t` remain unqualified too.

There's also the case for
> `ranges::uninitialized_copy_result<borrowed_iterator_t<IR>, borrowed_iterator_t<OR>>`

If you decide you qualify `borrowed_`\*`_t` in the return type too, would you do it here too? Even though they're template arguments to the return type? If so, then you'd wonder what'd make `range_`*`_t` so special to remain unqualified.

jensmaurer:
I think what I did is consistent with the changes done elsewhere in [algorithms] by the other pull request(s) linked in #2512.

The `\itemdecl`s are not exactly intended to be valid freestanding declarations (for example, we regularly show member functions referring to other class members). Instead, we want to uniquely identify the declaration (from the header synopsis) we're talking about.

Thus, I think the most important aspect is qualifying the _declarator-id_. I have qualified the `*_result` return types because we apparently did that for the other ranges algorithms as well, and it is at least conceivable that non-ranges algorithms would also yield similar `*_return` values (but not in namespace ranges, of course). The confusion potential seems much less for the "borrowed" and "ranges" parts, which only make sense in a ranges ecosystem.

I think the presentation would deteriorate if we attempted to qualify everything. Maybe a valid fallback position is to say that names in the return type are interpreted as-if they were in a trailing return type, and thus don't really need qualification at all. (Such an as-if approach breaks down for concept-ids in template-parameter-lists and requires-clauses, though, so maybe a better conceptual approach would be to say that the qualification of the declarator-id is instead as-if turned into a `namespace X { ... }` wrapper for the entire declaration.

@tkoeppe , do you think we need more discussion in a wider forum here, or does #2512 establish sufficient precedence?

tkoeppe:
This seems like an improvement. @jwakely, @CaseyCarter: any final thoughts?

jwakely:
I like it, but it does seem strange that `borrowed_iterator_t` is not qualified.

tkoeppe:
Well, but that seems to be how we do things, all the other uses of `borrowed_iterator_t` are also unqualified even if they appear as part of a declaration that uses `ranges::` elsewhere.

jwakely:
Yes, and that seems strange :slightly_smiling_face:

tkoeppe:
Yeah. I'd be happy to entertain suggestions!

jwakely (approved):
This is an improvement. Qualifying `borrowed_iterator_t` (here and elsewhere) can be done separately if desired.

tkoeppe:
> Yes, and that seems strange

You can't have std::range without strange!

tkoeppe:
So... I sense general approval, or at least lack of sustained opposition? @CaseyCarter, final chance to veto?

Issue #1571, opened 2017-03-22
figure out what's wrong with the "Allocator requirements" table and prevent it ever happening again


zygoloid:
See #1567 for pictures of the problem: under some circumstances we end up with parts of a longtable outside the typeblock, overlapping the bottom margin and sometimes flowing off the bottom of the page.

TeX is providing us with no warning or other indication that this is happening.

See this [reduced example](https://github.com/cplusplus/draft/files/860048/badtables.txt) of sources triggering the problem. My suspicion is that it's a bug in the `longtable` package.

The required setup for the problem appears to be a page that contains:

1) A table (in my case, at the top of the page, but I don't know whether that matters).
2) A longtable that entirely fits within the page.
3) Another longtable.

Under those circumstances, the permitted height of the second longtable before it first breaks to a new page appears to be miscomputed by an amount equal to the height of the floating table.

zygoloid:
We have confirmation that this is a `longtable` bug, but it's not clear that we have a reliable workaround. :(

For C++17, we'll be doing a manual page-by-page inspection of page breaks, so we don't need to block on this.

tkoeppe:
@zygoloid: Can you add a link to the SE discussion so we have an entry point for the follow-up?

jensmaurer:
https://www.latex-project.org/cgi-bin/ltxbugs2html.new?pr=tools/3512

jensmaurer:
This occurs with the current status of the draft.

PR #4460, opened 2021-01-07
Dismantle requirements table in [re.req] and [container.requirements]

### LINKED ISSUE #120: [sequence.reqmts] Table "Sequence container requirements" is bad, and it should feel bad
This table uses ugly shorthand for pre/postconditions, like so:

_Requires_: `T` shall be [...]
post: `distance(begin(), end()) == distance(i, j)`

It also misses \effects clauses, e.g.:

_Requires_ : `T` shall be [...]
pre: `i` and `j` are not iterators into `a`.
Inserts copies of elements in `[i, j)` before `p`

### LINKED ISSUE #4785: Don't reuse term "note" for normative requirements in tables
See https://github.com/cplusplus/draft/pull/4744#discussion_r667538507, the linked https://www.iso.org/sites/directives/current/part2/index.xhtml#_idTextAnchor333 and in particular its "For rules on notes to tables, see 29.5.1." reference. Using "note" suggest that these are not normative requirements.

jensmaurer:
We've wanted to get rid of the huge requirement tables since ages. Here's a specific suggestion, converting [re.req] as a guinea pig. Note that this is a fairly mechanical conversion, avoiding larger rephrasing.

Fixes #120.
Fixes #4785
Partially addresses #1571.

regex

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -456,10 +456,6 @@ swap(a, b)
 
 \begin{itemdescr}
 \pnum
-\result
-\keyword{void}
-
-\pnum
 \effects
 Equivalent to \tcode{a.swap(b)}.
 \end{itemdescr}
@@ -1530,20 +1526,8 @@ a.insert(p, il)
 
 \begin{itemdescr}
 \pnum
-\result
-\tcode{iterator}
-
-\pnum
 \effects
 Equivalent to \tcode{a.insert(p, il.begin(), il.end())}.
-
-\pnum
-\returns
-\begin{removedblock}
-An iterator
-that points to the copy of the first element inserted into \tcode{a}, or
-\tcode{p} if \tcode{il} is empty.
-\end{removedblock}
 \end{itemdescr}
 
 \indexcont{erase}%
@@ -1657,10 +1641,6 @@ a.assign(il)
 \end{itemdecl}
 
 \begin{itemdescr}
-\pnum
-\result
-\keyword{void}
-
 \pnum
 \effects
 Equivalent to \tcode{a.assign(il.begin(), il.end())}.
@@ -3046,10 +3026,6 @@ a.insert(il)
 
 \begin{itemdescr}
 \pnum
-\result
-\keyword{void}
-
-\pnum
 \effects
 Equivalent to \tcode{a.insert(il.begin(), il.end())}.
 \end{itemdescr}
@@ -3403,10 +3379,6 @@ a.clear()
 
 \begin{itemdescr}
 \pnum
-\result
-\keyword{void}
-
-\pnum
 \effects
 Equivalent to \tcode{a.erase(a.begin(), a.end())}.
 
@@ -3512,10 +3484,6 @@ b.contains(k)
 \pnum
 \effects
 Equivalent to: \tcode{return b.find(k) != b.end();}
-
-\pnum
-\complexity
-Logarithmic.
 \end{itemdescr}
 
 \indexordmem{contains}%
@@ -3531,10 +3499,6 @@ a_tran.contains(ke)
 \pnum
 \effects
 Equivalent to: \tcode{return a_tran.find(ke) != a_tran.end();}
-
-\pnum
-\complexity
-Logarithmic.
 \end{itemdescr}
 
 \indexordmem{lower_bound}%
@@ -4631,10 +4595,6 @@ a.insert(il)
 
 \begin{itemdescr}
 \pnum
-\result
-\keyword{void}
-
-\pnum
 \effects
 Equivalent to \tcode{a.insert(il.begin(), il.end())}.
 \end{itemdescr}
@@ -5075,10 +5035,6 @@ b.contains(k)
 
 \begin{itemdescr}
 \pnum
-\result
-\tcode{bool}
-
-\pnum
 \effects
 Equivalent to \tcode{b.find(k) != b.end()}.
 \end{itemdescr}
@@ -5089,10 +5045,6 @@ a_tran.contains(ke)
 \end{itemdecl}
 
 \begin{itemdescr}
-\pnum
-\result
-\tcode{bool}
-
 \pnum
 \effects
 Equivalent to \tcode{a_tran.find(ke) != a_tran.end()}.
@@ -5410,16 +5362,8 @@ a.reserve(n)
 
 \begin{itemdescr}
 \pnum
-\result
-\keyword{void}
-
-\pnum
 \effects
 Equivalent to \tcode{a.rehash(ceil(n / a.max_load_factor()))}.
-
-\pnum
-\complexity
-Average case linear in \tcode{a.size()}, worst case quadratic.
 \end{itemdescr}
 
 \pnum
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -31,10 +31,10 @@ as summarized in
 \end{libsumtab}
 
 
-\rSec1[container.requirements]{Container requirements}%
+\rSec1[container.requirements]{Requirements}%
 \indextext{requirements!container}
 
-\rSec2[container.requirements.general]{General container requirements}
+\rSec2[container.requirements.pre]{Preamble}
 
 \pnum
 Containers are objects that store other objects.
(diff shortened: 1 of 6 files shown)


Eelis:
Getting rid of the huge requirements tables would be great!

One small Q: is "return type" really appropriate for something like X::char_type? A quick search for "return type" in the spec suggests that so far, it's been used exclusively to refer to the return types of actual functions.

"Effects" also sounds kinda weird when describing compile-time type expressions that have no side-effects.

jensmaurer:
> One small Q: is "return type" really appropriate for something like X::char_type?

No, but it's used like that pervasively in the requirements tables. I'd be happy fold this into a generic "Requirements" element or so. In general, those requirement tables are not abundantly clear whether a member is a type or an expression, anyway. (For example, X::size_type could be a static data member.)

jensmaurer:
Updated with @tkoeppe; samples:

reqtab-type-stmt

reqtab-expr

tkoeppe:
LWG supports this. Next step: turn this into a paper.

jensmaurer:
Editorial meeting 2021-04-30: "is an expression with the following properties:" Otherwise, good to go. Make a paper with containers and regex tables converted.

jensmaurer:
use
```
X u = X();
X u;
```

drop presenting `X(a,b,c)`  (it does not introduce "u") otherwise

jensmaurer:
@tkoeppe The first pass over the container requirements is done; all cross-references are fixed.

jensmaurer:
See P2416R0.

JohelEGP:
> Fixes #4785

Did you confirm that this fixes it completely and not partially?

source/containers.tex
```diff
@@ -88,218 +91,371 @@
...
+\end{itemdescr}
+
+\begin{itemexpr}
+a == b
+\end{itemexpr}
+
+\begin{itemdescr}
+\pnum
+\expects
+\tcode{T} meets the \oldconcept{EqualityComparable} requirements
```
JohelEGP, line 318:
```suggestion
\tcode{T} meets the \oldconcept{EqualityComparable} requirements.
```

source/containers.tex
```diff
@@ -378,97 +534,143 @@
...
+
+\begin{itemexpr}
+a.rbegin()
+\end{itemexpr}
+
+\begin{itemdescr}
+\pnum
+\result
+\tcode{reverse_iterator};
+\tcode{const_reverse_iterator} for constant \tcode{a}
```
JohelEGP, line 566:
```suggestion
\tcode{const_reverse_iterator} for constant \tcode{a}.
```

jensmaurer:
No.

source/containers.tex
```diff
@@ -378,97 +534,143 @@
...
+\end{itemize}
+
+\begin{itemexpr}
+a.rbegin()
+\end{itemexpr}
+
+\begin{itemdescr}
+\pnum
+\result
+\tcode{reverse_iterator};
```
JohelEGP, line 565: Shouldn't the _Result:_\s of the reversible container requirements also specify that they are prvalues?

JohelEGP (changes requested):
Looks like properties of the form
> _Description-element:_ `code`

don't end in a period. This is unlike
> _Returns:_ `x`.

and all others.
Although some tables are really succinct and omit them, existing function specifications don't omit them.
It's also applied inconsistently. There are "_Postconditions:_ `code`" with and without a final period.

---

Some expressions assume we're in the body of a member function. This is clear when specifying functions. Is that as clear in this presentation style? Consider, for example:
> a.size() is an expression with the following properties:
> Value: distance(a.begin(), a.end()), i.e. the number of elements in the container.

> a.max_size() is an expression with the following properties:
> Returns: distance(begin(), end()) for the largest possible container.

and

> a.rbegin() is an expression with the following properties:
> Value: reverse_iterator(end())

> a.crbegin() is an expression with the following properties:
> Value: const_cast<X const&>(a).rbegin()

source/containers.tex
```diff
@@ -88,3030 +90,5181 @@
...
+\indexlibrary{\idxcode{#1}!ordered associative containers}%
+\indexlibrary{\idxcode{set}!\idxcode{#1}}%
+\indexlibrary{\idxcode{map}!\idxcode{#1}}%
+\indexlibrary{\idxcode{multiset}!\idxcode{#1}}%
+\indexlibrary{\idxcode{multimap}!\idxcode{#1}}%
+}
+
+\indexordmem{key_type}%
+\begin{itemdecl}
+typename {X::key_type
```
JohelEGP, line 2436:
```suggestion
typename X::key_type
```

JohelEGP (changes requested):

source/containers.tex
```diff
@@ -88,218 +91,371 @@
...
+\pnum
+A type \tcode{X} meets the \defn{container} requirements
+if the following types and expressions are well-formed and
+have the specified semantics.
+
+\begin{itemize}
+\item
+The name \tcode{X::value_type} denotes the type \tcode{T}.
+\tcode{T} is \oldconcept{Erasable} from \tcode{X}
+(see~\ref{container.requirements.general}, below).
```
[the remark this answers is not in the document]
jensmaurer: fixed

source/containers.tex
```diff
@@ -88,218 +91,371 @@
...
+that meets the requirements of a constant iterator and
+those of a forward iterator
+with value type \tcode{T}.
+\item
+The name \tcode{X::difference_type} denotes a signed integer type,
+identical to the difference type of
+\tcode{X::iterator} and \tcode{X::const_iterator}.
+\item
+The name \tcode{X::size_type} denotes an unsigned integer type
+that can represent any non-negative value of \tcode{difference_type}.
```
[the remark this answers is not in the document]
jensmaurer: fixed

source/containers.tex
jensmaurer: fixed

source/containers.tex
jensmaurer: fixed.

source/containers.tex
jensmaurer: fixed

source/containers.tex
jensmaurer: fixed

jensmaurer:
@jwakely, as discussed in LWG on 2021-10-01 , I've used `typename X::whatever_type` in the same fashion as any other function description.

That was the only actionable result from that LWG discussion; any further cleanup should be based on this incremental improvement.

jensmaurer:
@tkoeppe, this was seen by LWG; some improvements were applied. I think this change is fully editorial, except it's large. What's the next step here?

jensmaurer:
@jwakely, Daniel Krügler has opined on the -lib reflector that "Result:" is an undefined descriptive item and that it would be better to use "Return type:" (which is similarly undefined, but at least pre-existing). However, an expression usually has a type and a value category, but not really a "return type". LWG3166 Should I just invent some words for "Result" for [structure.specification], or what would be your preferred way forward?

http://lists.isocpp.org/lib/2021/12/21327.php

jwakely:
I think it would be better to use _Result:_ (and define it) rather than try to reuse something that isn't a great fit.

Dani-Hub:
> I think it would be better to use _Result:_ (and define it) rather than try to reuse something that isn't a great fit.

I'm OK with that but would like to ask for explicitly referring to [LWG 3166](https://wg21.link/lwg3166) as part of the paper.

jensmaurer:
I've added an entry for "Result:"to [structure.specification]. I think I'm not using "Value:" anywhere, so I'm not seeing why a reference to LWG3166 is necessary.

Dani-Hub:
> I've added an entry for "Result:"to [structure.specification]. I think I'm not using "Value:" anywhere, so I'm not seeing why a reference to [LWG3166](https://wg21.link/lwg3166) is necessary.

The new entry for "Result:" is new information to me, it was not present in [the draft](https://wiki.edg.com/pub/Wg21telecons2021/LibraryWorkingGroup/d2416r1.pdf) I had reviewed and where my request came from. Given the new state I agree that quoting LWG 3166 is no longer useful.

jensmaurer:
> The new entry for "Result:" is new information to me, it was not present in the draft I had reviewed

Right, I've just added it.

PR #4462, opened 2021-01-12
[dcl.pre] Reorder paragraphs

### LINKED ISSUE #338: [dcl.pre] paragraphs are in a strange order
First we have "the simple-declaration <blah> is divided into three parts".
Then a digression on attribute-declarations and scopes.
Then back to simple-declarations again.
Then a digression on static_assert-declarations and empty-declarations.
Then, with no introduction or explanation, some words about init-declarators, and about decl-specifier-seq's. That's clearly supposed to be related to the prior stuff about simple-declarations, but we're too far away for that to be a reasonable implied context.
Oh, then a lie: "Syntactic components beyond those found in the general form of declaration are added to a function declaration to make a function-definition." This means to say "beyond those found in the general form of a simple-declaration"; a function-definition is one of the possibilities in the general form of "declaration".
And then back to decl-specifier-seq's again.

This is a mess and should be fixed.

Related: we should check which of these simple-declaration rules should also apply to function-definitions and member-declarations, and suggest to core that any discrepancies be fixed.

jensmaurer:
Group discussion of simple-declaration and
move other kinds of declarations to the end.

Fixes #338

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -92,6 +92,16 @@ the form
 \end{note}
 
 \pnum
+\indextext{declaration}%
+\indextext{scope}%
+Certain declarations contain one or more scopes\iref{basic.scope.scope}.
+Unless otherwise stated, utterances in
+\ref{dcl.dcl} about components in, of, or contained by a
+declaration or subcomponent thereof refer only to those components of
+the declaration that are \emph{not} nested within scopes nested within
+the declaration.
+
+\pnum
 A
 \grammarterm{simple-declaration} or
 \grammarterm{nodeclspec-function-declaration} of the form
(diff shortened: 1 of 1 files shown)


tkoeppe:
This looks good, but maybe @opensdh could lend a second pair of eyes?

opensdh (approved):
I might move the scope discussion up rather than down (in case now or in the future there is some text in [dcl.pre] that depends on its "utterances" rule), but other than that this looks thoroughly benign.

tkoeppe:
@jensmaurer: feel free to merge this, with or without @opensdh's minor suggestion!

tkoeppe:
@opensdh: Thank you very much!

jensmaurer:
Moved "utterances" up a little.

PR #4470, opened 2021-01-24
[iterator.concept.winc] Improve paragraph's indexing and grammar

johelegp:


source/iterators.tex
```diff
@@ -1321,7 +1321,7 @@
 \pnum
 A type \tcode{I} is an \defnadj{integer-class}{type}
 if it is in a set of implementation-defined class types
```
johelegp, line 1323:
```suggestion
if it is in a set of \impldef{integer-class types} class types
```
How does that look?

jensmaurer: make it singular inside the \impldef

jensmaurer (changes requested):

source/iterators.tex
johelegp: Fixed.

jensmaurer (approved):

CaseyCarter (approved):

PR #4473, opened 2021-01-26
[stacktrace.basic.cmp] Update 'lexicographical_­compare_­3way'

lam-work-group:
The name has been changed to `lexicographical_compare_three_way` by [P1614R2](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1614r2.html#clause-25-algorithms-library).

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -21723,7 +21723,7 @@ friend strong_ordering
 \pnum
 \returns
 \tcode{x.size() <=> y.size()} if \tcode{x.size() != y.size()};
-\tcode{lexicographical_compare_three_way(\newline
+\tcode{lexicographical_compare_three_way(
 x.begin(), x.end(), y.begin(), y.end())}
 otherwise.
 \end{itemdescr}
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -21723,7 +21723,7 @@
 \pnum
 \returns
 \tcode{x.size() <=> y.size()} if \tcode{x.size() != y.size()};
-\tcode{lexicographical_compare_3way(\newline
+\tcode{lexicographical_compare_three_way(\newline
```
jensmaurer, line 21726: The manual `\newline` needs to go, otherwise we get a bad line break.

```suggestion
\tcode{lexicographical_compare_three_way(
```

jensmaurer (changes requested):

source/utilities.tex
lam-work-group: Done

jensmaurer: Did you actually build the PDF with that change? Now there is an extra space after the `(`, so some more fixes are in order.

Further, your name shows up as

`Author: lam-work-group <77063444+lam-work-group@users.noreply.github.com>`

in the resulting git history. Is that how you wish to be cited?

jensmaurer:
See #4474 for the fixed pull request.

tkoeppe:
This was fixed by #4474. Thanks for bringing this up!

PR #4477, opened 2021-01-29
[lib] Index all uses of public concepts

### LINKED ISSUE #4383: Always use \libconcept, even in synopsis and itemdecls
There are several remaining unmarked libconcepts, particularly in ranges.tex. They should be marked so that they can be indexed.  This is a continuation of Jens first pass at this in 4f255cae.

johelegp:
Resolves #4383.

No visible diff outside the index according to diffpdf.

Remaining question: Should these be dropped?
```
support.tex:\indexlibraryglobal{three_way_comparable}%
support.tex:\indexlibraryglobal{three_way_comparable_with}%
```

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -716,7 +716,7 @@ namespace std {
     template<class I, class F>
       using for_each_n_result = in_fun_result<I, F>;
 
-    template<input_iterator I, class Proj = identity,
+    template<@\libconcept{input_iterator}@ I, class Proj = identity,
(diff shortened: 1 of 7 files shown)


jensmaurer:
> Remaining question: Should these be dropped?

I think so.

jensmaurer (approved):

johelegp:
> No visible diff outside the index according to diffpdf.

I reconfirmed this.

PR #4481, opened 2021-02-05
[stacktrace.entry.query] Fix description of `source_line`

cpplearner:
Unlike `description` and `source_file`, this function does not return a string.

This appears to be a mistake made when [P0881R7](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p0881r7.html) was applied.


jensmaurer (approved):

tkoeppe:
Thank you!

PR #4482, opened 2021-02-05
[dcl.init.general] Fix misapplied term 'block variable'

### LINKED ISSUE #4478: [dcl.init.general] p5 wording does not seem to match the intent
[\[dcl.init.general\]/5](http://eel.is/c++draft/dcl.init.general#5):

> A declaration of a block variable with linkage that has an *initializer* is ill-formed.

IIUC the intent is to prohibit

```c++
void f() {
    extern int x = 1;
}
```

But it seems that the variable `x` is not a [block variable](http://eel.is/c++draft/basic.scope.block#def:block_variable), because it does not [belong](http://eel.is/c++draft/basic.scope.scope#def:entity,belong) to the block scope, because the target scope of its declaration is the innermost enclosing namespace scope ([\[dcl.meaning.general\]/(3.5)](http://eel.is/c++draft/dcl.meaning.general#3.5)).

I think this sentence should instead talk about the *immediate scope* of the declaration.

cc @opensdh

jensmaurer:
Block-scope externs are in view here, but those are not
block variables, because their target scope is an
enclosing namespace scope.

Fixes #4478

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -4252,8 +4252,9 @@ and~\ref{stmt.dcl}.
 \end{note}
 
 \pnum
-A declaration of a block variable with linkage
-that has an \grammarterm{initializer} is ill-formed.
+A declaration $D$ of a variable with linkage
+shall not have an \grammarterm{initializer}
+if $D$ inhabits a block scope.
 
 \pnum
 \indextext{initialization!default}%


opensdh (commented):
I like this word order better than the original suggestion in the issue, although I think "it" could work fine here; the subject of the sentence is the obvious antecedent, and there isn't another meaningful interpretation available anyway.

jensmaurer:
@opensdh, thanks for the comment. Elsewhere, we had troubles with "it" being sometimes obfuscating, so let's stick with the super-obvious $D$ here.

tkoeppe:
OK, let's take this and hope that nobody asks for a definition of "inhabitiation".

jensmaurer:
"inhabit" is a well-defined term these days [basic.scope.scope] p2.3

PR #4483, opened 2021-02-05
[lib] Use 'tm', not 'struct tm'

### LINKED ISSUE #4338: Consistent (and conventional) spelling of `struct tm`
Conventionally, when C++ declares or defines a struct type, such as `struct timespec;`, it refers to that type by name without the `struct` tag when used.

However, `struct tm` appears to be something of a special case.  As this type is supplied by the C library, and appears in mostly C specified APIs, the `struct` tag (required by C compilers) has been preserved in much of our wording and library synopses.  However, the `<locale>` header is an exception, consistently referring to just `tm*` in the functions it declares.

We should be consistent in how we reference this type in C++ source, and make an active decision to either follow the C++ convention of not repeating the `struct` tag on each use, or whether we prefer to follow the C convention of requiring the `struct` tag in each case.

jensmaurer:
Using the elaborated-type-specifier implies that this type
is declared in namespace 'std', which might not be true.

Fixes #4338

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -4171,8 +4171,8 @@ namespace std {
   @\unspec@ setw(int n);
   template<class moneyT> @\unspec@ get_money(moneyT& mon, bool intl = false);
   template<class moneyT> @\unspec@ put_money(const moneyT& mon, bool intl = false);
-  template<class charT> @\unspec@ get_time(struct tm* tmb, const charT* fmt);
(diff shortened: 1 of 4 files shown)


jensmaurer:
We have declarations of "struct tm" in header synopses `<cwchar>` and `<ctime>`, showing that "struct tm" is a member of namespace `std`. That's probably not exactly great, but also likely a LWG-level larger wording issue.

PR #4486, opened 2021-02-05
[basic.start.dynamic] Clarify note about templated variables

### LINKED ISSUE #4420: Confusing wording in note on [basic.start.dynamic]/1
> An explicitly specialized non-inline static data member or variable template specialization has ordered initialization.

Since English lacks a grouping operator it is unclear which adjectives apply to both sides of the "or". I initially read it as `An (explicitly specialized non-inline static data member) or (variable template specialization)`, but I think it is supposed to be closer to `An explicitly specialized non-inline (static data member or variable template specialization)`. Can this be rephrased to make this clear?

jensmaurer:
Fixes #4420

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -6511,8 +6511,8 @@ specialization, is partially-ordered if the variable
 is an inline variable that is not an implicitly or explicitly instantiated
 specialization, and otherwise is ordered.
 \begin{note}
-An explicitly specialized non-inline static data member or
-variable template specialization has ordered initialization.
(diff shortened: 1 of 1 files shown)


jensmaurer:
Send e-mail to the CWG reflector: http://lists.isocpp.org/core/2021/02/10481.php
As per #4420, wait two weeks for objections before applying this pull request.

jensmaurer:
@tkoeppe : no objections were received; this is ready.

PR #4487, opened 2021-02-05
[iterator.concept.winc] Move concept definition near its description

### LINKED ISSUE #4472: [iterator.concept.winc] `I` used to represent both an integer-class and weakly_incrementable type
The first 2 exposition-only concepts, _`is-integer-like`_ and _`is-signed-integer-like`_, have a template parameter named `T`.

Then comes the `weakly_incrementable` concept, with a template parameter named `I`.

The following normative paragraphs use `I` to define "integer-class", a subset of the `T` above. They rightfully reintroduce `I` each time, although the first one is unused.

Then, the final normative paragraph (before the Note: and Recommended practice:) uses `I` again, but this time to describe `weakly_incrementable`.

`I` is widely used throughout [iterator] to represent iterators. I believe those describing integer-class types should be renamed to `T`. This would be consistent with the _`is-integer-like`_ concept.

Related: #3457.

jensmaurer:
Also rename the placeholder type I to T, used in the specification
of the exposition-only concepts is-integer-like and
is-signed-integer-like, to avoid a presentation conflict
with the placeholder type I referring to an iterator type.

Fixes #4472

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -1332,7 +1332,7 @@ template<class I>
 \end{codeblock}
 
 \pnum
-A type \tcode{I} is an \defnadj{integer-class}{type}
+A type \tcode{T} is an \defnadj{integer-class}{type}
 if it is in a set of \impldef{integer-class type} types
 that behave as integer types do, as defined below.
 \begin{note}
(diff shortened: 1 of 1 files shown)


CaseyCarter (approved):

jensmaurer:
@tkoeppe , rebased and force-pushed.

source/iterators.tex
```diff
@@ -1312,20 +1312,10 @@
...
-template<class I>
-  concept @\deflibconcept{weakly_incrementable}@ =
-    @\libconcept{default_initializable}@<I> && @\libconcept{movable}@<I> &&
-    requires(I i) {
-      typename iter_difference_t<I>;
-      requires @\exposid{is-signed-integer-like}@<iter_difference_t<I>>;
-      { ++i } -> @\libconcept{same_as}@<I&>;   // not required to be equality-preserving
-      i++;                      // not required to be equality-preserving
-    };
 \end{codeblock}
```
tkoeppe, line 1315: Hm, it's a bit weird to have these "bare" codeblocks, which in reality are stuck to their preceding paragraph. The status quo was sort of OK, but the new codeblock you're now introducing below seems to be stuck to the wrong paragraph now. We're lacking a mechanism to introduce something with an unnumbered block at the beginning. We have that for synopses and for grammar, and I suppose for itemdecls, but this use case is none of these. Thoughts?

CaseyCarter: FWIW, I wouldn't mind if all of these concept definitions were changed to use `itemdecl` instead of `codeblock`. I'm not sure how we landed on `codeblock` in the first place; I suspect it was a typo on my part the first time that was then replicated a couple of dozen times.

jensmaurer: Rebased. I've also moved the `\pnum` to before the codeblock.

tkoeppe: I've been adding subclause structures whenever we had more such "bare" codeblocks (of which recent papers added a lot), but maybe itemdecls are the way to go here. In any case, I'm afraid this needs another rebase.

tkoeppe: @jensmaurer would you like to take another stab at this?

PR #4489, opened 2021-02-06
[temp.expl.spec] Use 'reachable from', not 'declared before'

### LINKED ISSUE #4479: [temp.expl.spec] Should "declared before" be replaced by "reachable"?
[\[temp.expl.spec\]/7](http://eel.is/c++draft/temp.expl.spec#7.sentence-1), emphasis mine:

> If a template, a member template or a member of a class template is explicitly specialized then that specialization shall be **declared before** the first use of that specialization that would cause an implicit instantiation to take place, in every translation unit in which such a use occurs; no diagnostic is required.

Now that we have modules, this sentence should probably say "... then a declaration of that specialization shall be reachable from ..." instead.

Note that [P1787R6: Declarations and where to find them](https://github.com/cplusplus/draft/commit/c7e8927098ab5074491edb64cbb4e1258c0c2141#diff-993d162d99c97390cfd9b2c9b076a5ef904d9e48366aa8bbbe7663cfc8367588L3195-R3145) updated a similar sentence in [\[temp.spec.partial.general\]](http://eel.is/c++draft/temp.spec.partial.general#1.sentence-3), but neglected this one.

jensmaurer:
With modules, 'declared before' no longer makes sense.

Fixes #4479

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -6462,7 +6462,8 @@ template<class U> void A<short>::C<U>::f() { @\commentellip@ }    // error: \tco
 
 \pnum
 If a template, a member template or a member of a class template is explicitly
-specialized then that specialization shall be declared before the first use of
+specialized, a declaration of that specialization shall be reachable from
+every use of
 that specialization that would cause an implicit instantiation to take place,
 in every translation unit in which such a use occurs;
 no diagnostic is required.


source/templates.tex
```diff
@@ -6462,7 +6462,8 @@
 
 \pnum
 If a template, a member template or a member of a class template is explicitly
-specialized then that specialization shall be declared before the first use of
+specialized then a declaration of that specialization shall be reachable from
+the first use of
 that specialization that would cause an implicit instantiation to take place,
```
opensdh, line 6467: Can we simplify this with "every use", as in

> If a template, a member template or a member of a class template is explicitly specialized, a declaration of that specialization shall be reachable from every use of that specialization that would cause an implicit instantiation to take place; no diagnostic is required.

?

jensmaurer: Fixed and force-pushed.

source/templates.tex
```diff
@@ -6462,7 +6462,8 @@
 
 \pnum
 If a template, a member template or a member of a class template is explicitly
-specialized then that specialization shall be declared before the first use of
+specialized, a declaration of that specialization shall be reachable from
+every use of
 that specialization that would cause an implicit instantiation to take place,
 in every translation unit in which such a use occurs;
 no diagnostic is required.
```
cpplearner, line 6469:
```suggestion
that specialization that would cause an implicit instantiation to take place;
no diagnostic is required.
```

tkoeppe: @opensdh, any thoughts on this suggestion?

cpplearner: This is part of [his suggestion](https://github.com/cplusplus/draft/pull/4489#discussion_r572329651).

tkoeppe: I see, thanks!

PR #4490, opened 2021-02-10
[intro.object]/1 Turn non-normative wording into a Note; remove incorrect wording.

### LINKED ISSUE #2308: [intro.object]/1: the (interpretation of the) value of an object is determined by the type of the expressions used to access it
[[intro.object]/1](http://eel.is/c++draft/intro.object#1):
> Some objects are polymorphic ([class.virtual]); the implementation generates information associated with each such object that makes it possible to determine that object's type during program execution.
For other objects, the interpretation of the values found therein is determined by the type of the expressions ([expr.compound]) used to access them.

The last sentence doesn't make any sense.
The only way I found to give sense to this sentence is: "object" — "region of storage", "values found therein" — "pile of bits in storage", and this sentence says that when the storage is accessed by an expression, these bits are interpreted as an object representation of an object with the type of an expression.

languagelawyer:
The last sentence looks just wrong: the type of the expression doesn't affect the value of the objects.
The second to last sentence doesn't look normative, it looks like a typical implementation strategy description.

Fixes #2308

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3023,13 +3023,12 @@ name\iref{basic.pre}. An object has a storage
 duration\iref{basic.stc} which influences its
 lifetime\iref{basic.life}. An object has a
 type\iref{basic.types}.
+\begin{note}
 Some objects are
 polymorphic\iref{class.virtual}; the implementation
 generates information associated with each such object that makes it
-possible to determine that object's type during program execution. For
-other objects, the interpretation of the values found therein is
-determined by the type of the \grammarterm{expression}{s}\iref{expr.compound}
-used to access them.
+possible to determine that object's type during program execution.
+\end{note}
 
 \pnum
 \indextext{subobject}%


jensmaurer:
I think our current object model in C++ is that each object has a type; if the expression used to access the object doesn't fit, it's undefined behavior (except for the aliasing exceptions). I agree the note appears to say something else.

jensmaurer:
Editorial meeting 2021-04-16: Looks good.

kitsnet:
> 
> 
> I think our current object model in C++ is that each object has a type; if the expression used to access the object doesn't fit, it's undefined behavior (except for the aliasing exceptions). I agree the note appears to say something else.

How does this change address the aliasing exceptions?
Doesn't it _break_ them?

jensmaurer:
How so? Please look at the actual change.

kitsnet:
https://stackoverflow.com/a/54170480/5343439

jensmaurer:
This change has removed the confusing (and incorrect) wording

"For other objects, the interpretation of the values found therein is determined by the type of the \grammarterm{expression}{s}\iref{expr.compound} used to access them."

This (also given the StackOverflow discussion) appears to be a net improvement.

kitsnet:
OK, I see that the wording of [expr.reinterpret.cast] that caused the StackOverflow discussion has also changed since then, so I remove my objection.

languagelawyer:
[The wording has been changed to take into account pointer-interconvertibility](http://wg21.link/cwg2342). For non-pointer-interconvertible objects (as in the SO question), the new wording has the same meaning as the old wording.

kitsnet:
That might have been the intent. However, it also moved some otherwise objectionable wording from the normative part into a note and arguably reduced its scope.

Otherwise, adding yet another case to the use of the language that is formally UB but practically IDB is hardly an improvement.

xmh0511:
I wonder, if P1839 would be approved, Is that something is necessary to be clarified by the wording that has been a note by this thread. Probably not, Maybe we can use pointer-interconvertible to interpret that. Cancel the normative for the wording can make an object always constantly; Regardless of what the type of a "glvalue" that refers to the object is, it always has the same value as if the type is the dynamic type of that object.

PR #4495, opened 2021-02-12
[conv.lval] Determine the value of a constant without accessing an object

### LINKED ISSUE #4493: [conv.lval] doesn't specify the result for objects that are not accessed?
From https://stackoverflow.com/questions/58479267/initialization-order-of-constant-initialized-variable-with-static-storage-durati @sdkrystian 

To my understanding, nothing specifies what value `b` will have.
```c++
const int a = 42;
const int b = a;
```
Variable `a` is [usable in constant expressions](http://eel.is/c++draft/expr.const#def:usable_in_constant_expressions), so it is [not ODR-used](http://eel.is/c++draft/basic.def.odr#4.2), so [[conv.lval]/(2.2)](http://eel.is/c++draft/conv.lval#2) apply:
> 2 When an lvalue-to-rvalue conversion is applied to an expression _E_, and either
> - (2.1) _E_ is not potentially evaluated, or
> - (2.2) the evaluation of _E_ results in the evaluation of a member _E_<sub>`x`</sub> of the set of potential results of _E_, and _E_<sub>`x`</sub> names a variable `x` that is not odr-used by _E_<sub>`x`</sub>
>
> **the value contained in the referenced object is not accessed**.

I think it means (and should say) **the object** is not accessed. But this is not the main concern. [conv.lval]/3, which says «The result of the conversion is determined according to the following rules», doesn't seem to cover the case when l-to-r conversion is applied to an evaluated expression denoting an object which is not accessed. I believe the bullet (3.3) saying «Otherwise, **the object indicated by the glvalue is read ([defns.access])**, and the value contained in the object is the prvalue result» can't apply in case the object is not accessed. Or the implied reading of (3.3) is that the part about access is just ignored for non-accessed objects?

(In case there is doubt that [conv.lval] is applied to `a` in `const int b = a;`, it can be replaced with `const int b = a + 0;`)

jensmaurer:
Fixes #4493

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -599,7 +599,7 @@ descriptions of those operators and contexts.
 \pnum
 \indextext{conversion!lvalue-to-rvalue}%
 \indextext{type!incomplete}%
-A glvalue\iref{basic.lval} of a non-function, non-array type \tcode{T}
+A glvalue\iref{basic.lval} $E$ of a non-function, non-array type \tcode{T}
 can be converted to
 a prvalue.
 \begin{footnote}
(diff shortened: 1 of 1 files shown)


source/expressions.tex
```diff
@@ -661,6 +661,10 @@
 type, the conversion copy-initializes the result object from
 the glvalue.
 
+\item Otherwise, if $E$ is a constant expression that designates an object
+that is usable in constant expressions ([expr.const]),
+the result is the value of that object.
```
zygoloid, line 666: This seems much broader than the rule in /2.2. This would also remove the access in:

```
struct X { int a; };
constexpr X x = {1};
int n = (&x)->a;
```

... which currently does result in an odr-use of `x` and an access. Changing that doesn't seem editorial.

Can we make the words here more directly refer back to the /2.2 case somehow? Maybe something like:

```
\item Otherwise, if $E$ is potentially evaluated and the object designated by $E$ is not accessed, as specified above, the designated object is necessarily usable in constant expressions ([expr.const]) and the result is the value of that object.
```

(I don't like "is necessarily" here. Do we have another way to say "must be" without making ISO angry?)

zygoloid: Or maybe we could inline /2 into these bullets?

jensmaurer: This also has interactions with the very definition of "lvalue":

```
struct S
{
     static const int a = 1;
};

int main()
{
     return S::a;  // no odr-use; no definition of S::a needed
}
```

[intro.object] p1 "An object is created by a definition (6.2) ..."

[basic.lval] p1 "A glvalue is an expression whose evaluation determines the identity of an object or function.

How can you identify an object that wasn't ever created (because there is no definition)?

http://nolp.dhl.de/nextt-online-public/set_identcodes.do?lang=de&idc=579347005106

languagelawyer: @jensmaurer I think you pasted a wrong link

jensmaurer: I think it works best to integrate p2 into the "result" bullets.

languagelawyer: Maybe say `the result is the value of \tcode{x}'s initializer.` in non-odr-used variable case? Variable is not an object after all, it doesn't have value. And this

languagelawyer: BTW
```c++
struct S
{
	int m;
};

int main()
{
	&S::m;
}
```
[expr.prim.id.qual] says that a _qualified-id_ denoting data member is an lvalue, but I don't think `S::m` in `&S::m` can denote an object. Even though we have pointer-to-member, IIUC `S::m` is still an lvalue there, which makes me feel uncomfortable. Should this be a Core issue? 🤔

xmh0511: *Maybe say the result is the value of \tcode{x}'s initializer*. A variable that is usable in constant expressions does not always has an initializer 
> A variable or temporary object o is constant-initialized if. 
>> - either it has an initializer or **its default-initialization results in some initialization being performed** 
````cpp
struct A{
    int v = 1;
};
````

jensmaurer:
Editorial meeting 2021-04-16: "const int x = 10000; sizeof(char{x})" needs the value of x, but is not potentially evaluated. Fix by moving the "set of potential results" wording to the front. Then, create a diff so that CWG can sign off.

PR #4499, opened 2021-02-18
[container.requirements] Use bulleted lists to introduce identifiers

jensmaurer:
This is a first step towards dissolving the container requirements tables. This step is clearly editorial.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -72,13 +72,21 @@ aligned buffers and call \tcode{construct} to place the element into the buffer.
 \pnum
 In Tables~\ref{tab:container.req},
 \ref{tab:container.rev.req}, and
-\ref{tab:container.opt}
+\ref{tab:container.opt},
+\begin{itemize}
+\item
 \tcode{X} denotes a container class containing objects of type \tcode{T},
+\item
 \tcode{a} and \tcode{b} denote values of type \tcode{X},
+\item
 \tcode{i} and \tcode{j} denote values of type (possibly const) \tcode{X::iterator},
+\item
 \tcode{u} denotes an identifier,
+\item
 \tcode{r} denotes a non-const value of type \tcode{X}, and
+\item
 \tcode{rv} denotes a non-const rvalue of type \tcode{X}.
+\end{itemize}
 
 \begin{libreqtab5}
 {Container requirements}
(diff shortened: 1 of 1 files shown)


tkoeppe:
@jwakely, @CaseyCarter: could you perhaps take a look at this? Jens and I thought that those introductions of "meaningful names for the purpose of stating requirements" would look much more manageable as a list rather than as an every-expanding sentence.

source/containers.tex
```diff
@@ -1632,44 +1661,68 @@
...
 \tcode{a2} denotes a value of a type with nodes compatible with type
 \tcode{X} (\tref{container.node.compat}),
+\item
 \tcode{b} denotes a possibly \tcode{const} value of type \tcode{X},
+\item
 \tcode{u} denotes the name of a variable being declared,
+\item
 \tcode{a_uniq} denotes a value of type \tcode{X}
 when \tcode{X} supports unique keys,
 \tcode{a_eq} denotes a value of type \tcode{X}
```
CaseyCarter, line 1679: Are we intentionally keeping `a_uniq` and `a_eq` together in a single bullet, or was this overlooked? I find it inconsistent that `a_tran` is separate but these two share a bullet.

CaseyCarter (approved):
Overall: Yes, please! I think LWG has discussed doing this at least once before. I have one nitpick with the bulleting, which is far fewer than the number of problems I have with the pre-existing wording.

source/containers.tex
jensmaurer: This was overlooked. Fixed + force-pushed.

CaseyCarter (approved):

tkoeppe:
@CaseyCarter: yes, we have long wanted to do something about the tables, and this is only the very first step. Stay tuned for more! :-)

source/containers.tex
```diff
@@ -73,12 +73,20 @@
 In Tables~\ref{tab:container.req},
 \ref{tab:container.rev.req}, and
 \ref{tab:container.opt}
```
tkoeppe, line 75: We sometimes have a comma after "In tables A, B, C\opt{,}" and sometimes not. I think with the new list format, a comma is always good, what do you think?

tkoeppe (changes requested):

CaseyCarter:
> @CaseyCarter: yes, we have long wanted to do something about the tables, and this is only the very first step. Stay tuned for more! :-)

I was referring specifically and only to the bulleted-list-of-names change. We only dislike the paragraphs of names. The tables themselves have a _much_ more storied history of hatred and vitriol. (I've starred the notification for the table PR in my inbox to remind me to take a look at it over the weekend.)

source/containers.tex
jensmaurer: Made it a comma everywhere (the unordered container requirements had a colon).

tkoeppe: Thanks!

tkoeppe (approved):

PR #4500, opened 2021-02-21
[Editorial] Add whitespace.

mordante:
Change all occuances of '){' to ') {'. This was noticed due to the
differences in [bitmask.types].

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -6604,7 +6604,7 @@ that allows the programmer to avoid deadlocks.
 #include "a.h"
 #include "b.h"
 B b;
-A::A(){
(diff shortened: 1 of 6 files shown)


jensmaurer (approved):

jwakely:
Doesn't this go against the Core policy of intentionally using inconsistent formatting?

Making the library clauses consistent is good.

jensmaurer:
> Doesn't this go against the Core policy of intentionally using inconsistent formatting?

Yeah, sort-of, although I personally think that gratuitous spacing differences are not in scope here. (We also want consistent indentation, I believe.)

PR #4502, opened 2021-02-22
[stdatomic.h.syn] Add semicolons in using statements.

mkurdej:


jwakely (changes requested):
Thanks for the fix. The semi-colons are present in http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p0943r6.html and their omission from the draft is an editorial error.

As per https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#code-blocks the _// see below_ comments are meant to line up on a specific column, so you should remove a space when adding the `;` character.

mkurdej:
OK. I'll line up the `// see below` comments as needed.

jwakely (approved):
Thanks!

jensmaurer (approved):

PR #4503, opened 2021-02-23
[range.iota.iterator,range.iota.sentinel] Fix markup around \libconcept

CaseyCarter:
...so it doesn't appear in the rendered text.


johelegp:
Good catch. However, nothing appears (unlike using @s in inline code). N4878 looks good whereas http://eel.is/c++draft/range.iota is missing the hyperlink to the concept definition, which is what this would fix.

jensmaurer (approved):

PR #4527, opened 2021-02-25
[class.cdtor] Only objects of scalar type can be accessed

languagelawyer:


jensmaurer (approved):
I think that's right, but @zygoloid should also comment.

zygoloid (commented):
I think this is correct, but I think we're missing similar wording talking about accessing the object representation of the object.

The old wording -- in particular, the "the object or" part -- might have been read as covering that before, in some cases: an incautious reader might conclude that copying a union with a trivial copy special member results in an access of the value of the union object.

I think we want "the value of any of <del>the object or</del> its subobjects <ins>or any element of its object representation</ins>" in the first edit and "the value <del>of the object or subobject</del> thus obtained" in the second edit, though that seems to be straying out of editorial country.

tkoeppe:
Thanks, @zygoloid! Let's give this some more baking time, not for now.

tkoeppe:
@languagelawyer: please see above discussion, and also rebase?

languagelawyer:
@tkoeppe If I understand it correctly, @zygoloid suggests that this can't be fixed editorially.

tkoeppe:
@languagelawyer: could you make the proposed change regardless, and we can run this issue past CWG? If necessary we can file a core issue then, but maybe a group review will suffice.

jensmaurer:
Put on CWG's agenda: https://wiki.edg.com/bin/view/Wg21telecons2021/CoreWorkingGroup

AlisdairM:
I could not find any minutes --- were the editorial issues flagged for telecon ever discussed?

tkoeppe:
@jensmaurer Just to check, is the current state either self-evidently editorial, or did CWG otherwise have a look?

jensmaurer:
The change looks correct to me. CWG did not have a look.

tkoeppe:
OK, let's merge it then, thanks!

PR #4530, opened 2021-02-26
Make checker script more helpful

### LINKED ISSUE #4529: ci: more visibility of errors output by tools/check.sh
@burblebee commented at https://github.com/cplusplus/draft/pull/4520#issuecomment-786393620 about wanting something like that. Does the following meet your expectations? https://github.com/johelegp/draft/runs/1985020607?check_suite_focus=true#step:4:2160. Maybe what you actually want is a bot that comments the errors on the PR. Or have the errors output in a separate step, so that you don't have to scroll all the LaTeX build output.

jwakely:
Fixes #4529

### DIFF
--- a/tools/check.sh
+++ b/tools/check.sh
@@ -1,5 +1,14 @@
 #!/bin/bash
 
+# Exit the script if any command (or pipeline) ends with an error
+set -e
+
+# Filter that exits the process if its input is non-empty.
+# If $* args given, prefixes each line of input with "Error: $*: ".
+fail() {
+  awk -v str="$*" '/./ { if (length(str)) { print "Error: " str ": " $0 } } END { exit NR }'
+}
+
 # Ignore files where rules may be violated within macro definitions.
 texfiles=$(ls *.tex | grep -v macros.tex | grep -v layout.tex | grep -v tables.tex)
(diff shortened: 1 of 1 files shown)


tools/check.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/bash
 
+# Filter that exits the process if its input is non-empty.
+# If $* args given, prefixes each line of input with "Error: $*: ".
+fail() {
+  awk -v str="$*" '/./ { if (length(str)) { print "Error: " str ": " $0 } x=1 } EXIT { exit x }'
```
jensmaurer, line 6: This doesn't exit the surrounding shell, it just exits the awk script. Thus, this changes the status quo. But maybe we want all checks run, and not terminate after the first error.

But the general idea is nice. We should mix it with the "github action log commands" so that the errors are categorized as such, and colored red. Give me a little while.

jwakely: Oops, I lost the `|| exit 1` when changing it from a sed command (as demo'd at https://github.com/cplusplus/draft/issues/4529#issuecomment-786559241) to an awk one.

jwakely: Oops, but it needs to be `END` not `EXIT`

jwakely: And it still won't work, because it exits the process at the end of the pipeline, which doesn't exit the process running the script. We need `set -e` to make it exit the whole script.

jwakely:
It should work better now.

jensmaurer:
Superseded by #4533

PR #4531, opened 2021-02-26
ci: split make and check steps

johelegp:
Addresses #4529 partially and https://github.com/cplusplus/draft/issues/4529#issuecomment-786786718 completely.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -6,6 +6,10 @@ on:
   pull_request:
   push:
 
+defaults:
+  run:
+    working-directory: source
+
 jobs:
   run-checks:
     name: Run checks
@@ -15,5 +19,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: install
         run: sudo apt-get install latexmk texlive-latex-recommended texlive-latex-extra texlive-fonts-recommended texlive-generic-recommended lmodern
+      - name: make
+        run: make quiet
       - name: check
-        run: cd source && make quiet && ../tools/check.sh
+        run: ../tools/check.sh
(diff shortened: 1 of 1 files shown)


.github/workflows/check.yml
```diff
@@ -6,14 +6,20 @@ on:
...
 
+defaults:
+  run:
+    working-directory: source
+
 jobs:
   run-checks:
     name: Run checks
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
jensmaurer, line 16: Why are we fixating this to Ubuntu 20.04 ?  "ubuntu-latest" seems to be less long-term maintenance effort, unless Ubuntu decides to change the package structure.

jensmaurer (changes requested):

.github/workflows/check.yml
johelegp: It's to address this: https://github.com/cplusplus/draft/issues/4529#issuecomment-786624331. I point out what's wrong in the second part of its following comment.

johelegp: See the first note at https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on. During the migration period, using `runs-on: ubuntu-latest` could use an Ubuntu 18.04 or 20.04 runner. @jwakely's fork workflow runs (https://github.com/jwakely/draft/actions) were all on Ubuntu 20.04 (for example, see https://github.com/jwakely/draft/runs/1986966401?check_suite_focus=true#step:1:4).

jensmaurer: Ok, thanks.

PR #4533, opened 2021-02-27
[check] Split github action into two stages (source, output)

### LINKED ISSUE #4529: ci: more visibility of errors output by tools/check.sh

jensmaurer:
Also improve the visibility of error messages.

Resolves #4529

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -17,9 +17,11 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - name: check-source
+        run: ../tools/check-source.sh
       - name: install
         run: sudo apt-get install latexmk texlive-latex-recommended texlive-latex-extra texlive-fonts-recommended lmodern
       - name: make
         run: make quiet
-      - name: check
-        run: ../tools/check.sh
+      - name: check-output
+        run: ../tools/check-output.sh
--- a/tools/check-output.sh
+++ b/tools/check-output.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+# This script checks the results of the LaTeX run.
+
+failed=0
+
+function fail() {
+
+    # echo "::error file=app.js,line=10,col=15::Something went wrong"
+
+    ! sed 's/^\(.\+\.tex\):/file=\1::/;s/^/::error /' | grep .
+}
+
+
+# Discover "Overfull \[hv]box" and "Reference ... undefined" messages from LaTeX.
+sed -n '/\.tex/{s/^.*\/\([-a-z0-9]\+\.tex\).*$/\1/;h};
+/Overfull [\\][hv]box\|LaTeX Warning..Reference/{x;p;x;p}' std.log |
+    sed '/^.\+\.tex$/{N;s/\n/:/}' | fail || failed=1
+
+
+# Cross references since the previous standard.
+function indexentries() { sed 's,\\glossaryentry{\(.*\)@.*,\1,' "$1" | LANG=C sort; }
+function removals() { diff -u "$1" "$2" | grep '^-' | grep -v '^---' | sed 's/^-//'; }
+function difference() { diff -u "$1" "$2" | grep '^[-+]' | grep -v '^\(---\|+++\)'; }
+XREFDELTA="$(difference <(indexentries xrefdelta.glo) <(removals <(indexentries xrefprev) <(indexentries xrefindex.glo)))"
+if [ -n "$XREFDELTA" ]; then
+  echo "incorrect entries in xrefdelta.tex:" >&2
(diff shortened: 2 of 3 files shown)


jensmaurer:
error-sample

tools/check.sh
```diff
@@ -1,109 +1,144 @@
...
-sed -n '/\.tex/{s/^.*\/\([-a-z0-9]\+\.tex\).*$/\1/;h};
-/Overfull [\\][hv]box\|LaTeX Warning..Reference/{x;p;x;p}' std.log |
-sed '/^.\+\.tex$/{N;s/\n/:/}' | grep . && exit 1
+# Filter that reformats the error message as a "workflow command",
+# for native handling by github actions.
+# Prefixes each line of input with "$*: ".
+function fail() {
+    # echo "::error file=app.js,line=10,col=15::Something went wrong"
+
+    # For some reason, the file/line/col information is not shown in the GUI.
```
johelegp, line 18: Have you tried prepending `source/` to the file? I did, but for some reason my workflow run doesn't fail in the precheck like yours does: https://github.com/johelegp/draft/runs/1995473623?check_suite_focus=true#step:3:31.

johelegp: Ah, it must be because I didn't escape the `/`.

jensmaurer: I just tried it. On the command-line, I get

```
::error file=source/intro.tex,line=500::intro.tex:500:indented codeblock env:    \begin{codeblock}
```
which seems correct, but the github output is still ... not extra-cool. Anyway, what we have now is a nice  improvement regardless.

johelegp:
It seems that the source location in the message is still useful. Otherwise, it only appears when you look at it from ["Files changed"](https://github.com/cplusplus/draft/pull/4533/files) (see image below) and not in the workflow run logs.

jensmaurer:
Ah, nice.

jwakely:
Excellent, thanks!

tkoeppe (approved):
1. I don't understand all the detailed `sed`/`grep` magic, but please do what you tihnk is best.
2. The split is great. I might maybe call the two phases `check-source` and `check-output`?
3. Regarding the commit message: this isn't just "CI", since it's also precommit. I recommend phrasing the commit message as "Split GitHub action into two stages (source, output)".

tools/check.sh
```diff
@@ -1,109 +1,144 @@
...
+    # Make sure to leave it in the message proper.
+    ! sed 's/^\(.\+\.tex\):\([0-9]\+\):/::error file=source\/\1,line=\2::\1:\2:'"$*"': /; s/^\([^:][^:]\)/::error ::'"$*"': \1/' |
```
jwakely, line 20: I think the last sed command will be ill-formed if the `$*` message contains a `/` e.g. something like `| fail "misplaced begin/end"`

It's probably OK, but if we want to allow `/` in the errors we could use another character less likely to be used in the error string, e.g. `@`

     ! sed 's@^\(.\+\.tex\):\([0-9]\+\):@::error file=source/\1,line=\2::\1:\2:'"$*"': @; s@^\([^:][^:]\)@::error ::'"$*"': \1@' |

(N.B. this also allows `source/` instead of `source\/`)

jwakely (approved):

johelegp:
>     3. Regarding the commit message: this isn't just "CI", since it's also precommit. I recommend phrasing the commit message as "Split GitHub action into two stages (source, output)".

Regarding that. For my recent commit that did not touch the LaTeX sources, I took the liberty to use https://www.conventionalcommits.org/.

tools/check.sh
jensmaurer: The trouble is that awk's regex machinery doesn't seem to allow references to sub-matches grouped by ().

But using @ is probably a good compromise.

tkoeppe:
@jensmaurer: Great, thanks, ship it!

@johelegp: Yeah, I didn't have a good regular label at hand previously, so I just sayd "Action" a lot, but I think Jens has come up with a nice one that we can use from now on.

tkoeppe (approved):

tools/check.sh
jwakely: GNU awk's `gensub()` function supports back-refs.

jensmaurer: Great, but @burblebee already dislikes GNU-isms in sed, so better avoid them for such a central functionality.

PR #4539, opened 2021-03-07
[time.syn] add constexpr to leap_second comparisons

MattStephanson:
The comparison operators for `leap_second` are all specified as being `constexpr` in [[time.zone.leap.nonmembers]](http://eel.is/c++draft/time.zone.leap.nonmembers), but this specifier is missing in the corresponding section of [[time.syn]](http://eel.is/c++draft/time.syn). Although the current wording matches [P0355R7](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0355r7.html), I think this could qualify as an editorial issue. It's internally inconsistent and the intent seems clear, especially because these functions just call the corresponding `time_point` overrides, which are `constexpr` in both places. I'm happy to submit an LWG issue, however, if you feel differently.

### DIFF
--- a/source/time.tex
+++ b/source/time.tex
@@ -772,27 +772,27 @@ namespace std {
     // \ref{time.zone.leap}, leap second support
     class leap_second;
 
-    bool operator==(const leap_second& x, const leap_second& y);
-    strong_ordering operator<=>(const leap_second& x, const leap_second& y);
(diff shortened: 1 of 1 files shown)


jwakely (approved):
I agree this counts as editorial.

jensmaurer (approved):

PR #4543, opened 2021-03-09
[temp.variadic] Rearrange description of pack expansion

### LINKED ISSUE #4542: [temp.variadic] description of the instantiation of fold-expressions
In the current draft, there are two issues about the instantiation for a fold-expression.
The first is the standard seems not clear for stating which are the elements of the pack expansion parameters that appear in the pattern of fold-expression.  It could be assumed they're stated in paragraph 8 [temp.variadic#8](https://eel.is/c++draft/temp.variadic#8). 

However, such a layout of the rule seems that the description for what these elements of pack expansion parameters are is only applying to the case that is: `the instantiation of a pack expansion that is neither a sizeof... expression nor a fold-expression`.  

Is it necessary to take the rule out from paragraph 8 to be an individual rule?  That would clearly cover the case that what these elements of the pack expansion parameters are which appears in the fold-expression. 

The second issue is about the **binary fold-expression** when the number of the elements is zero. The whole paragraph 10 only talks about how the result is for the unary fold-expression when the number of the elements is zero. It lacks the case of a binary fold-expression. (Presumably, the result of this case is that of the expression `E`).

````cpp
#include <iostream>
struct A{
    A* node;
};
A a{};
template<typename...T>
auto fun(T...member){
    return (&a .*....* member);
}
int main() {
    a.node = &a;
    auto r = fun();
}
````
Such as this example, even though the first operand  &a would violate [expr.mptr.oper#2](https://eel.is/c++draft/expr.mptr.oper#2)

### PR DESCRIPTION
**Pull request by jensmaurer**
Fixes #4542

x1

x1

### CODE REVIEW

**Comment by jensmaurer** [comment]
@zygoloid , could you please have a look?

File: source/templates.tex [hunk 1]
```diff
@@ -2836,36 +2832,12 @@
...
-template void f<>();    // OK: \tcode{X<>} has no base classes
-                        // \tcode{x} is a variable of type \tcode{X<>} that is value-initialized
-\end{codeblock}
-\end{example}
-
 \pnum
 The instantiation of a \tcode{sizeof...} expression\iref{expr.sizeof} produces
-an integral constant containing the number of elements in the pack
-it expands.
+an integral constant with value $N$.
```
zygoloid, line 2837: Should we say that `sizeof...(Pack)` expands the elements of the pack? I think that this isn't observable, even in the weird alias template cases where you end up with `sizeof...` applied to something that's not just the name of a pack, so this is probably fine, but it seems slightly surprising. Maybe the wording consistency is worth the modicum of surprise.

source/templates.tex
```diff
@@ -2931,6 +2900,32 @@
...
+\begin{note}
+The variety of list varies with the context:
+\grammarterm{expression-list},
+\grammarterm{base-specifier-list},
+\grammarterm{template-argument-list}, etc.
+\end{note}
+When $N$ is zero, the instantiation of the expansion produces an empty list.
+Such an instantiation does not alter the syntactic interpretation of the
+enclosing construct, even in cases where omitting the list entirely would
+otherwise be ill-formed or would result in an ambiguity in the grammar.
```
zygoloid, line 2916: Do we need to say the same thing for the fold-expression case? Eg:

```
int y;
template<int ...N> int x(int((y && ... && N)));
```

... is not interpreted as a function declaration `int x(int ((y)))` when the pack is empty. I think that's pre-existing, but if we're generalizing the E_i wording to apply to fold-expressions, then maybe we should consider generalizing this too? The intent seems clear that this should apply in all cases.

zygoloid (approved):
Seems fine; I have comments but they're mostly drive-by things. This change seems like an improvement to me.

github won't let me put comments on unchanged lines, but since I'm making drive-by comments anyway, I don't like the first bullet in the "Such an element [...] is interpreted as follows" list:

> the element is a template parameter [..] designating [a] template argument

In the other bullets, we say the element is a (grammar term) designating a parameter. Saying in this case that the element is a parameter designating an argument seems wrong.

source/templates.tex
jensmaurer: p7 status quo: "A pack whose name appears within the pattern of a pack expansion is expanded by that pack expansion."

What else do you want me to say?

source/templates.tex
jensmaurer: Fixed.

jensmaurer:
> the element is a template parameter [..] designating [a] template argument

I've made this a little more regular, although the different kinds of template parameters make this a little bit more involved. Please have a look. (At least, you now have a hook where to apply your comments.)

The other two wishes have been implemented and force-pushed.

jensmaurer:
@zygoloid, is this more to your liking?

source/templates.tex
zygoloid: Hmm, I had interpreted that "is expanded by" as being a simple boolean property of the appearance of the pack, tying into the rule that all appearances of a pack need to be expanded, and not anything to do with substitution of the elements of the pack into some enclosing pattern.

But I think it's fine to leave this as you have it. Sure, we'll substitute into the *identifier* and produce a bunch of *id-expression*s / *template-name*s/*typedef-name*s naming the elements of the pack, then count them and drop them on the floor, but it doesn't matter that we do so.

zygoloid (approved):

source/templates.tex
```diff
@@ -2800,24 +2800,26 @@
...
 
 \pnum
-The instantiation of a pack expansion
-that is neither a \tcode{sizeof...} expression
-nor a \grammarterm{fold-expression}
-produces a
-list of elements
-$\mathtt{E}_1,$ $\mathtt{E}_2,$ $\cdots,$ $\mathtt{E}_N$,
+The instantiation of a pack expansion considers
+items $\mathtt{E}_1,$ $\mathtt{E}_2,$ $\cdots,$ $\mathtt{E}_N$,
```
tkoeppe, line 2804: a) should this be `$\tcode{E}_1, \tcode{E}_2` etc?

b) should this be `\dotsc`, not `cdots`?

tkoeppe: (I could do this separately if you prefer to keep the diff small, just as  a note-to-self then.)

jensmaurer: Style Guidelines:
"Use \cdots when \dotsc (or \ldots) could be confused with a pack expansion."

Updated to \dotsc and fixed the math mode.

@tkoeppe, this seems ready now.

source/templates.tex
```diff
@@ -2931,6 +2910,29 @@
 \tcode{,}       & \tcode{void()} \\
 \end{floattable}
 
+\pnum
+The instantiation of any other pack expansion
+produces a list of elements
+$\mathtt{E}_1,$ $\mathtt{E}_2,$ $\cdots,$ $\mathtt{E}_N$.
```
tkoeppe, line 2916: Here, too?

source/templates.tex
tkoeppe: Ah, I see. Was there a risk of confusion? It seems pretty clear.

source/templates.tex
```diff
@@ -2800,24 +2800,26 @@
...
+if the pack is a template parameter pack, the element is
+an \grammarterm{id-expression}
+(for a non-type template parameter pack),
+a \grammarterm{typedef-name}
+(for a type template parameter pack declared without \tcode{template}), or
+a \grammarterm{template-name}
+(for a type template parameter pack declared with \tcode{template})
+designating the $i^\text{th}$ template parameter
+that resulted from instantiation of
+the template parameter pack.
```
xmh0511, line 2822: "that resulted from instantiation of the template parameter pack." seems like it's only suitable for the template parameter pack itself is a pack expansion(it might be non-type template pack or type template pack declared with template). So, Is it necessary to add a restriction to that sentence, that is "if the template parameter pack itself is a pack expansion"?

source/templates.tex
jensmaurer: Fixed

source/templates.tex
jensmaurer: I'm not seeing your point. Could you given an example, please?

jensmaurer:
@tkoeppe, updated. I think we should merge this and postpone @xmh0511's concerns to a separate issue (if there are any left).

source/templates.tex
xmh0511: My point is that "instantiation of the template parameter pack" sounds like the template parameter pack is a pack expansion, such a description has the same manner as that of the function template parameter(which is itself a pack expansion). `template<typename...T> struct A{ template<T...id> struct B{};};`. In this example, the template parameter pack of the nested class template is a pack expansion, hence, when using the pack `id` in some context, we would say that replace `id` to designate ith template parameter **that resulted from instantiation of the template parameter pack**. However,  a type template parameter pack declared without `template` is never a pack expansion. Furthermore, such a `template<int...N> struct C{T b = {N...};};` where `int...C` is not a pack expansion and when instantiating the pattern `N`, each element merely designates the ith template parameter; I mean the result is not instantiated from something, instead, it just designates the corresponding template parameter according to the template arguments.

tkoeppe:
Sorry, @jensmaurer, it was getting late and I went ahead and made the new working draft. Let's get this in for the next round.

tkoeppe:
I'll try to take another look so we can make this mailing.

source/templates.tex
tkoeppe: (@jensmaurer: can you please "resolve" conversations that definitely don't need further attention?)

jensmaurer: ack

source/templates.tex
tkoeppe: @jensmaurer: any final thoughts on this?

tkoeppe:
Updated screenshot:

image

xmh0511:
> Updated screenshot:
> 
> image

As aforementioned, For bullet 8.1, the attributive clause("that resulted from instantiation of the template parameter pack") is not completely suitable here. For instance, give the example `template<class...T> struct A{using type = std::tuple<T...>}`, where `class...T` is not a pack expansion, How could we say that the typedef-name resulted from expanding `T...` designates the ith template parameter *that resulted from instantiation of the template parameter pack*?

jensmaurer:
The prior wording "designating the $i^\text{th}$ corresponding type or value template argument" was restored for case 8.1, fixing the regression.

xmh0511:
> The prior wording "designating the $i^\text{th}$ corresponding type or value template argument" was restored for case 8.1, fixing the regression.

I still think "if the pack was introduced by a pack expansion" should be removed for bullet 8.1. For example, `template<int...N> struct B{ B(){fun(N...)}};`, Isn't that each element instantiated in `N...` does designate the corresponding template argument? However, `int... N` is just a template parameter pack declaration instead of a pack expansion. If add the condition seems that the latter clause is not suitable for this case.

source/templates.tex
```diff
@@ -2800,24 +2800,25 @@
...
-non-type) designating the $i^\text{th}$
-corresponding type or value template argument;
+if the pack is a template parameter pack, the element is
+an \grammarterm{id-expression}
+(for a non-type template parameter pack),
+a \grammarterm{typedef-name}
+(for a type template parameter pack declared without \tcode{template}), or
+a \grammarterm{template-name}
+(for a type template parameter pack declared with \tcode{template})
+if the pack was introduced by a pack expansion,
```
xmh0511, line 2820: I still think "if the pack was introduced by a pack expansion" should be removed for bullet 8.1. For example, `template<int...N> struct B{ B(){fun(N...)}};`, Isn't that each element instantiated in `N...` does designate the corresponding template argument? However, `int... N` is just a template parameter pack declaration instead of a pack expansion. If add the condition seems that the latter clause is not suitable for this case.

PR #4545, opened 2021-03-12
[range.split.inner] Clarify paragraph 1

CaseyCarter:
... to avoid confusion about what it means for a name that is not present to denote a type.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -5866,7 +5866,7 @@ namespace std::ranges {
 \end{codeblock}
 
 \pnum
-If \tcode{Base} does not model \libconcept{forward_range}
+If \exposid{Base} does not model \libconcept{forward_range}
 there is no member \tcode{iterator_category}.
 Otherwise, the \grammarterm{typedef-name} \tcode{iterator_category} denotes:
 \begin{itemize}
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -5866,7 +5866,9 @@ namespace std::ranges {
 \end{codeblock}
 
 \pnum
-When present, the \grammarterm{typedef-name} \tcode{iterator_category} denotes:
+If \tcode{Base} does not model \libconcept{forward_range}
+there is no member \tcode{iterator_category}.
+Otherwise, the \grammarterm{typedef-name} \tcode{iterator_category} denotes:
 \begin{itemize}
 \item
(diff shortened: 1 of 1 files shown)


timsong-cpp (approved):

jensmaurer:
@jwakely , are you ok with this change?

source/ranges.tex
```diff
@@ -5866,7 +5866,7 @@
 \end{codeblock}
 
 \pnum
-The \grammarterm{typedef-name} \tcode{iterator_category} denotes:
+When present, the \grammarterm{typedef-name} \tcode{iterator_category} denotes:
```
jwakely, line 5869: This is certainly an improvement, but I'd also like the "present only if ..." comment in the synopsis to be stated in the normative prose. It's easy to miss the comment in the synopsis.

Contrast with [thread.lock.scoped] where [LWG 2731](https://wg21.link/lwg2731) added an explicit "there is no member `mutex_type`" for the case where it's absent.

So something like:

```suggestion
If `Base` does not model `forward_range` there is no member `iterator_category`.
Otherwise, the \grammarterm{typedef-name} \tcode{iterator_category} denotes:
```

If you'd prefer to do that separately, I'm happy to approve this as-is.

jwakely (approved):

source/ranges.tex
CaseyCarter: I think this suggestion is nicely explicit. Done.

source/ranges.tex
```diff
@@ -5866,7 +5866,9 @@
 \end{codeblock}
 
 \pnum
-When present, the \grammarterm{typedef-name} \tcode{iterator_category} denotes:
+If \tcode{Base} does not model \libconcept{forward_range}
```
johelegp, line 5869:
```suggestion
If \exposid{Base} does not model \libconcept{forward_range}
```

jwakely: Ah yes, good catch, thanks.

jwakely (approved):
Approved, with @johelegp's suggestion for `\exposid{Base}`

jensmaurer:
@CaseyCarter , I think we need a small update here.

CaseyCarter:
Thanks @johelegp - applied.

jwakely (approved):

jensmaurer (approved):

PR #4549, opened 2021-03-17
[lex.string] Replace string-literal with a string literal object in a note

Endilll:
I think syntactic construct `string-literal` couldn't be modified, unlike its resulting object.


jensmaurer (approved):

jensmaurer:
(Note: when merging, shorten and fix commit description.)

PR #4562, opened 2021-03-27
[namespace.def.general] Clarify inline namespaces

### LINKED ISSUE #4556: [namespace.def.general] p6 Clarify "enclosing" and inserted using-directive
The first place that needs to be precise is   
> Members of an inline namespace can be used in most respects as though they were members of the enclosing namespace. 

Since the following vague(misleading) sentence has been removed in the current draft
> The enclosing namespace set of O is the set of namespaces consisting of the innermost non-inline namespace enclosing an inline namespace O, together with any intervening inline namespaces.  

Presumably, the wording `enclosing` should assume to have the meaning in English, that is, all namespaces that declared in a namespace A, we would say A **enclosing** those namespaces.  

For the first bullet, Should it be precise to 
>  Members of an inline namespace can be used in most respects as though they were members of the enclosing namespace whose **inline namespace set comprises the inline namespace**.    

it together with  the following rule   
> The inline namespace set of N is the transitive closure of all inline namespaces in N.    

clarify this case
````
namespace A{
  namespace B{
     inline namespace C{
          int v;
     }
  }
}
A::v;  //ill-formed , Although A enclosing C
A::B::v; //well-formed
````

The second issue is   
#1 
> Specifically, the inline namespace and its enclosing namespace are both added to the set of associated namespaces used in argument-dependent lookup whenever one of them is, and **a using-directive ([namespace.udir]) that names the inline namespace is implicitly inserted into the enclosing namespace** as for an unnamed namespace ([namespace.unnamed]). 

Is it a bit misleading and useless?  According to the statement for lookup rule in the current draft, such a using-directive does not **affect the result of the lookup**, since the following rule 
> A search in a scope X for a name N from a program point P is **a single search** in X for N from P unless X is the scope of a class or class template T    

> A single search in a scope S for a name N from a program point P finds all declarations that precede P to which any name that is the same as N ([basic.pre]) is bound in S.   

Even though in the current standard, the relevant rule does also clearly state this point  
> When considering an associated namespace N, the lookup is the same as the lookup performed when N is used as a qualifier ([namespace.qual]) except that: 
>> **Any using-directives in N are ignored**.

And there's a more readable rule in [basic.lookup.argdep] that can instead of the initial sentence of the above rule at #1.   
> The associated namespaces for a call are the innermost enclosing non-inline namespaces for its associated entities as well as every element of the inline namespace set ([namespace.def]) of those namespaces.   

So, Is it necessary to remove the redundant and a bit misleading rule to follow the current draft?

jensmaurer:
Highlight 'innermost enclosing' namespace, and remove
unneeded mechanism of inserting a using-directive.

Fixes #4556

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -7220,7 +7220,7 @@ appertains to the namespace being defined or extended.
 
 \pnum
 Members of an inline namespace can be used in most respects as though they were members
-of the enclosing namespace. Specifically, the inline namespace and its enclosing
+of the innermost enclosing namespace. Specifically, the inline namespace and its enclosing
 namespace are both added to the set of associated namespaces used in
 argument-dependent lookup\iref{basic.lookup.argdep} whenever one of them is,
 and a \grammarterm{using-directive}\iref{namespace.udir} that names the inline


tkoeppe:
@opensdh: could you kindly review this?

opensdh:
The discussion on the issue mostly focused on the "enclosing" phrasing.  Isn't the *using-directive* necessary for unqualified lookup?  [namespace.qual]/1 says that qualified lookup "additionally searches every element of the inline namespace set", but unqualified lookup has only the directives ([basic.lookup.unqual]/2.2).

tkoeppe:
@jensmaurer: any thoughts?

jensmaurer:
@opensdh, you're right. I'll reduce this pull request to just talk about "innermost enclosing" then, and leave the using-directive alone.

Do you feel we could/should add an explicit bullet to basic.lookup.unqual that talks about looking in the inline namespace set? That feels a lot more straightforward than introducing a hypothetical using-directive.

jensmaurer:
@tkoeppe, this should be pretty straightforward now.

tkoeppe:
Thanks, that's indeed straight-forward. @opensdh: what do you think about following up as Jens suggests?

opensdh:
Since *using-directive*s are transitive, you would have to add not a bullet to [basic.lookup.unqual]/2 but a rule **after** those bullets that for every namespace's scope explicitly searched in them, a search is also performed in every member of its inline namespace set.  I don't know whether that's nicer than the current formulation.

tkoeppe:
@opensdh: Thanks! I find it hard to tell. @jensmaurer, feel free to propose something if you think the exposition could be made easier to follow here, and we can see if it's worth it. But it sounds like the status quo isn't all that bad, either.

PR #4563, opened 2021-03-27
[except.spec] Clarify potentially-throwing functions

### LINKED ISSUE #4561: Several issues about section [except.spec]
> any of the destructors for any of its potentially constructed subobjects is **potentially-throwing** or the destructor is virtual and the destructor of any virtual base class is **potentially-throwing**.  

> E implicitly invokes a function (...) that is potentially-throwing 

AFAIK, the property "potentially-throwing" ascribes to an expression. That is, we may say an expression is potentially-throwing or not. However, we cannot say a constructor, destructor, or function is potentially-throwing. Maybe, we should say such an entity has the **potentially-throwing exception specification**  

The statement for implicitly-declared assignment operator indicates this point  
>  if the **invocation** of any assignment operator in the implicit definition is potentially-throwing.  

So, for the first three bullets. Should we say that the **invocation** of the destructor, constructor, or function is potentially-throwing? Or, another way, such a constructor, destructor, or function has the potentially-throwing exception specification? That may make sense.  

In this case, Isn't that the last bullet can cover the former, as per   
> The immediate subexpressions of an expression E are 
>> - any function call that E implicitly invokes    

As specified below, the relevant rules do not only cover the case for `defaulted on its first declaration` but also cover the case for **implicitly-declared**. So, in order to be consistent, I think the introduction in this sentence should add the case for implicitly-declared into its exception cases.

jensmaurer:
Expressions are potentially-throwing; functions have a
potentially-throwing exception specification.

Fixes #4561


xmh0511:
Maybe miss the constructor, destructor in other paragraphs,  which has the same issue. [except.spec#7.1], [except.spec#8].

jensmaurer:
Fixed.

PR #4565, opened 2021-03-28
[time.clock.cast.sys],[time.clock.cast.utc] Use Duration2 for clarity.

### LINKED ISSUE #4564: [time.clock.cast.sys] and [time.clock.cast.utc] reuse "Duration" confusingly
[N4885](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/n4885.pdf) [time.clock.cast.utc]/2:

> ```cpp
> template<class Duration>
>   auto operator()(const time_point<SourceClock, Duration>& t) const
>     -> decltype(SourceClock::to_utc(t));
> ```
> *Mandates:* `SourceClock::to_utc(t)` returns a `utc_time<Duration>`, where `Duration` is a valid `chrono::duration` specialization.

Confusingly, the `Duration` that this `operator()` is templated on, is **not** the same as the `Duration` in the return type! What this means to say is:

*Mandates:* `SourceClock::to_utc(t)` returns a `utc_time<Duration2>`, where `Duration2` is a valid `chrono::duration` specialization.

Attempting to `static_assert` that this `operator()` returns `utc_time<Duration>` matching the template parameter exactly, will fail to compile (observe that [time.clock.tai.overview] specifies that `tai_clock::to_utc()` returns `utc_time<common_type_t<Duration, seconds>>`). Which just happened to me, hence the issue. While I was confused by the wording, the intent seems clear upon a re-read, and I believe that this is an editorial issue, not an LWG issue. Thanks to @MattStephanson for diagnosing this and explaining it to me.

This occurs 4 times:
* [time.clock.cast.sys]/2
* [time.clock.cast.sys]/5
* [time.clock.cast.utc]/2
* [time.clock.cast.utc]/5

StephanTLavavej:
Fixes #4564.

### DIFF
--- a/source/time.tex
+++ b/source/time.tex
@@ -3717,8 +3717,8 @@ template<class Duration>
 
 \pnum
 \mandates
-\tcode{SourceClock::to_sys(t)} returns a \tcode{sys_time<Duration>},
-where \tcode{Duration} is a valid \tcode{chrono::duration} specialization.
+\tcode{SourceClock::to_sys(t)} returns a \tcode{sys_time<Duration2>},
+where \tcode{Duration2} is a valid \tcode{chrono::duration} specialization.
 
 \pnum
 \returns
@@ -3748,8 +3748,8 @@ template<class Duration>
 
 \pnum
 \mandates
-\tcode{DestClock::from_sys(t)} returns a \tcode{time_point<DestClock, Duration>},
-where \tcode{Duration} is a valid \tcode{chrono::duration} specialization.
+\tcode{DestClock::from_sys(t)} returns a \tcode{time_point<DestClock, Duration2>},
+where \tcode{Duration2} is a valid \tcode{chrono::duration} specialization.
 
 \pnum
 \returns
@@ -3781,8 +3781,8 @@ template<class Duration>
 
 \pnum
 \mandates
-\tcode{SourceClock::to_utc(t)}  returns a \tcode{utc_time<Duration>},
-where \tcode{Duration} is a valid \tcode{chrono::duration} specialization.
+\tcode{SourceClock::to_utc(t)}  returns a \tcode{utc_time<Duration2>},
+where \tcode{Duration2} is a valid \tcode{chrono::duration} specialization.
 
 \pnum
 \returns
@@ -3812,8 +3812,8 @@ template<class Duration>
 
 \pnum
 \mandates
-\tcode{DestClock::from_utc(t)} returns a \tcode{time_point<DestClock, Duration>},
-where \tcode{Duration} is a valid \tcode{chrono::duration} specialization.
+\tcode{DestClock::from_utc(t)} returns a \tcode{time_point<DestClock, Duration2>},
+where \tcode{Duration2} is a valid \tcode{chrono::duration} specialization.
 
 \pnum
 \returns
--- a/source/time.tex
+++ b/source/time.tex
@@ -3749,7 +3749,7 @@ template<class Duration>
 \pnum
 \mandates
 \tcode{DestClock::from_sys(t)} returns a \tcode{time_point<DestClock, Duration2>},
-where \tcode{Duration2} is a valid \tcode{chrono::duration} specialization.
+where \tcode{Dura\-tion2} is a valid \tcode{chrono::duration} specialization.
 
 \pnum
 \returns
@@ -3813,7 +3813,7 @@ template<class Duration>
 \pnum
 \mandates
 \tcode{DestClock::from_utc(t)} returns a \tcode{time_point<DestClock, Duration2>},
-where \tcode{Duration2} is a valid \tcode{chrono::duration} specialization.
+where \tcode{Dura\-tion2} is a valid \tcode{chrono::duration} specialization.
 
 \pnum
 \returns


StephanTLavavej:
Maybe this needs to be shorter, like `Dur2`, to avoid failing checks? I am unfamiliar with the error here.

CaseyCarter:
> Maybe this needs to be shorter, like `Dur2`, to avoid failing checks? I am unfamiliar with the error here.

"Overfull \hbox" is latex telling us there's a long-and-unbreakable word at the end of a line overflowing ~10pts into the margin, and that wrapping it to the next line would leave too much whitespace on the current line. In other words, it can't find a way to flow this text that doesn't look terrible within the constraints it has been given.

The fix is for a human to look at the generated output to see what the problem word is, and either introduce a discretionary hyphen with `\-` ("If you have to break this word break it here and add a hyphen"), or a plain discretionary linebreak with `\brk{}` (ditto, but with no hyphen - generally used for code), or by rephrasing the prose so the problem word doesn't encroach into the right margin.

source/time.tex
```diff
@@ -3717,8 +3717,8 @@
 
 \pnum
 \mandates
-\tcode{SourceClock::to_sys(t)} returns a \tcode{sys_time<Duration>},
-where \tcode{Duration} is a valid \tcode{chrono::duration} specialization.
+\tcode{SourceClock::to_sys(t)} returns a \tcode{sys_time<Duration2>},
+where \tcode{Duration2} is a valid \tcode{chrono::duration} specialization.
```
CaseyCarter, line 3721: This seems like a normative change to me. Before, I read this as requiring `Duration` to be a valid specialization of `chrono::duration` and requiring `SourceClock::to_sys(t)` to return `sys_time<Duration>`. The after condition is obviously quite different, allowing `SourceClock::to_sys(t)` to return a `sys_time` parameterized with a completely different specialization of `chrono::duration`.

MattStephanson: > I read this as [...] requiring `SourceClock::to_sys(t)` to return `sys_time<Duration>`.

That's a plausible reading on its face, but based on my experience implementing this, I think it's too restrictive to be the intended meaning. There's no guarantee that the input `Duration` will be able to represent the difference between the two clock epochs, so in general there *must* be some flexibility to return a different `duration` specialization. That's why all the built-in conversions return a `common_type_t<Duration, seconds>`. Otherwise, something reasonable like `clock_cast<gps_clock>(sys_days{1d / January / 2020y})` won't work if the input and output `Duration` have to be the same.

But I'm happy to help, if desired, if people prefer to see this escalated to an LWG issue.

CaseyCarter: > But I'm happy to help, if desired, if people prefer to see this escalated to an LWG issue.

We've already gone to the trouble to prepare a PR, I suggest waiting for word from one of the editorial team before dropping this and resubmitting as an LWG issue. My instincts for what the Project Editor will consider to be an editorial change are not infallible.

jensmaurer:
The "where ..." phrase seems to imply we're talking about a fresh Duration here, but there is another Duration in scope that makes that a stretch. @jwakely , did you implement this?

jwakely:
We have not implemented this yet.

HowardHinnant:
Thanks for bringing this to my attention @jensmaurer.

This does look like a defect to me, and the fix looks correct to me.  Somehow when translating the original documentation to the proposed wording, I introduced this bug.

Here is the original documentation:  https://howardhinnant.github.io/date/tz.html#clock_cast  Search for:

"template <class Duration>
auto
operator()(const utc_time<Duration>& t) const
    -> decltype(DestClock::from_utc(t));"

> _Remarks:_ This function does not participate in overload resolution unless `DestClock::from_utc(t)` is well formed. If `DestClock::from_utc(t)` does not return `time_point<DestClock, _some duration_>` the program is ill-formed.

I.e. it was _some duration_ that was intended instead of `Duration`, and the proposed fix subs in `Duration2` where _some duration_ was in the original documentation.

jensmaurer:
@tkoeppe, @jwakely: Any opposition to fixing this editorially, given that @HowardHinnant said the pull request manifests the intended behavior?

jensmaurer:
Editorial meeting 2021-04-30: Given the stakeholders, this is considered editorial.

jwakely:
Sorry for the late comment, but couldn't this be further simplified? What is a "valid" `duration`? [time.duration.general] already says that invalid specializations are ill-formed, and `time_point` requires a `duration`. So isn't the requirement here just that it returns a time point with the right clock?

jwakely:
Specifically, I think that both instances of "where Duration2 is a valid chrono::duration specialization" could be changed to "for some Duration2" with no change in meaning.

HowardHinnant:
No opinion either way here.

jensmaurer:
I appreciate the concerns, and I sympathize with them, but given the rather invasive nature (from a "squint the wrong way" standpoint) of this pull request, I'd suggest to merge it as-is and have a separate issue to look for "valid specializations" in all of [time]. (I suspect there are more of these).

jwakely:
That works for me

jwakely (approved):

jwakely:
The only other cases are [time.clock.cast.sys] and [time.clock.cast.utc]. This change seems consistent with the language there. I think we could remove "valid" there without changing the meaning.

tkoeppe:
@jwakely Please let's follow up on the simplification suggestions. I created #4613 to track that.

PR #4570, opened 2021-04-01
replace all keywords by static

Humm42:
That one’s all we need.


cpplearner:
This is a normative change. Consider [submit a defect report](https://isocpp.org/std/submit-issue).

jensmaurer:
@Humm42, please consider attaching your real name in github. Otherwise, after a merge, only your alias will be in the permanent git tree.

PR #4574, opened 2021-04-02
[dcl.init.aggr] An initializer list is brace-enclosed

### LINKED ISSUE #4571: [dcl.init.aggr] Inconsistent meaning of initializer list
> List-initialization is initialization of an object or reference from a braced-init-list. Such an initializer is called an initializer list, and the comma-separated initializer-clauses of the initializer-list or designated-initializer-clauses of the designated-initializer-list are called the elements of the initializer list.   

That hints that the component `braced-init-list` is called an initializer list.   

However, in section [dcl.init.aggr], there are several components that are not braced-init-list are called initializer list. List in the following:  
[dcl.init.aggr#3.1]
> If the initializer list is a designated-initializer-list ... 

[dcl.init.aggr#3.2]
> If the initializer list is an initializer-list...  

[dcl.init.aggr#4.1]
> the initializer list is a designated-initializer-list...  

Should we only say that it's an initializer list by using braces to enclose them?  Such as `{initializer-list}`, `{designated-initializer-list}`   

As a contrast   
[dcl.init.aggr#3.3] 
>  Otherwise, the initializer list must be {}...  

Which is a kind of  `braced-init-list`. So it's called initializer list

jensmaurer:
Do not claim that a designated-initializer-list or
an initializer-list is an initializer list.

Fixes #4571

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -4735,14 +4735,15 @@ The \defnx{explicitly initialized elements}{explicitly initialized elements!aggr
 of the aggregate are determined as follows:
 \begin{itemize}
 \item
-If the initializer list is a \grammarterm{designated-initializer-list},
+If the initializer list is
+a brace-enclosed \grammarterm{designated-initializer-list},
 the aggregate shall be of class type,
 the \grammarterm{identifier} in each \grammarterm{designator}
 shall name a direct non-static data member of the class, and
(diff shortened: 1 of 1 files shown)


jensmaurer:
Editorial meeting 2021-04-30: Don't use "has", use "braced-enclosed /designated-initializer-list/".

PR #4575, opened 2021-04-02
[basic.stc.static] Rephrase the definition of 'static storage duration'

### LINKED ISSUE #4567: The non-normative wording "local variable" should not survive in [basic.stc.static]
Since the "non-local variable " has been completely excised after P1787. However, there's still a similar wording survived in the current draft 
> All variables which do not have dynamic storage duration, do not have thread storage duration, and are not **local** have static storage duration.

Should we say 
> All variables which do not have dynamic storage duration, do not have thread storage duration, and do not belong to block and function parameter scope have static storage duration.  

To be consistent with these prior modifications.

jensmaurer:
This clarifies and centralizes the definition and
also avoids the undefined term 'local variable'.

Fixes #4567

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3696,10 +3696,17 @@ causes a system-generated runtime fault.
 
 \pnum
 \indextext{storage duration!static}%
-All variables which do not have dynamic storage duration, do not have thread
-storage duration, and are not local
-have \defn{static storage duration}. The
-storage for these entities lasts for the duration of the
+All variables which
+\begin{itemize}
+\item
+do not have thread storage duration and
+\item
+belong to a namespace scope\iref{basic.scope.namespace} or
+are first declared with
+the \keyword{static} or \keyword{extern} keywords\iref{dcl.stc}
+\end{itemize}
+have \defn{static storage duration}.
+The storage for these entities lasts for the duration of the
 program~(\ref{basic.start.static}, \ref{basic.start.term}).
 
 \pnum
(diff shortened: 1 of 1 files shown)


tkoeppe:
(To review by group, but tentatively OK to merge.)

tkoeppe:
To be reworded, it's too obscure at the moment that static block variables are handled elsewhere.

jensmaurer:
Editorial meeting 2021-04-30: Make sure this definition is complete (considers static local variables as well). Also reword p3.

tkoeppe:
@zygoloid: your review would be most appreciated here.

source/basic.tex
```diff
@@ -3669,10 +3669,17 @@
...
-storage for these entities lasts for the duration of the
+All variables which
+\begin{itemize}
+\item
+do not have thread storage duration and
+\item
+belong to a namespace scope\iref{basic.scope.namespace} or
+are declared with
+the \keyword{static} or \keyword{extern} keywords\iref{dcl.stc}
+\end{itemize}
```
zygoloid, line 3680: We need to exclude `static thread_local` here.

jensmaurer: @zygoloid, I did. The bullet just above says "does not have thread storage duration and". "static thread_local" has thread storage duration per [basic.stc.thread], so that's covered.

zygoloid: Ah, right, OK.

zygoloid (approved):

source/basic.tex
```diff
@@ -3696,10 +3696,17 @@
...
-storage duration, and are not local
-have \defn{static storage duration}. The
-storage for these entities lasts for the duration of the
+All variables which
+\begin{itemize}
+\item
+do not have thread storage duration and
+\item
+belong to a namespace scope\iref{basic.scope.namespace} or
+are declared with
```
zygoloid, line 3705: Would it be useful to say "are first declared" here, given that `static` is inherited in redeclarations (and can't be present in a redeclaration of a class member)?

jensmaurer: Yes, that clarifies the "class member redeclaration" case.

Fixed

tkoeppe:
Looks good, I'll merge this.

tkoeppe (approved):

PR #4579, opened 2021-04-05
P2351: add [[nodiscard]] to cast-like library wrappers [WIP]

hanickadot:
bit_cast, to_integer, forward, move, move_if_noexcept, as_const, to_underlying, and identity::operator()

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -1288,7 +1288,7 @@ manipulate and process both individual bits and bit sequences.
 namespace std {
   // \ref{bit.cast}, \tcode{bit_cast}
   template<class To, class From>
-    constexpr To bit_cast(const From& from) noexcept;
+    [[nodiscard]] constexpr To bit_cast(const From& from) noexcept;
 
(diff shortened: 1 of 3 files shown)


jensmaurer:
Which part is "WIP" here?

hanickadot:
Paper is WIP. LEWG is discussing if the paper is worth of spending time on it. My idea is to have an open pull request as part of the paper. If that's not what you want here close it.

jensmaurer:
@tkoeppe, please state whether you consider the change here editorial.
(My view: There is no general or specific guidance from LWG / LEWG for this matter, so this doesn't feel editorial.)

hanickadot:
No, I don't think this is editorial (as much I would like to). I want guidance from LEWG/LWG too. But I also like the idea that the paper is accompanied with an open pull request to the draft. To make it easier for editors.

jwakely:
More specifically, because of https://isocpp.org/std/standing-documents/sd-9-library-evolution-policies, which set the policy.

PR #4538, opened 2021-03-03
[check] Terminate sed commands with semicolons

jensmaurer:
even before a closing }, to improve compatibility
with BSD-based implementations.

PR #4595, opened 2021-04-29
[check] Fixes to scripts to run on OSX

burblebee:
Patches include:
* [check] Terminate more sed commands with semicolons (#4538).

WIP - more fixes to come...

### DIFF
--- a/tools/check-output.sh
+++ b/tools/check-output.sh
@@ -13,9 +13,9 @@ function fail() {
 
 
 # Discover "Overfull \[hv]box" and "Reference ... undefined" messages from LaTeX.
-sed -n '/\.tex/{s/^.*\/\([-a-z0-9]\+\.tex\).*$/\1/;h};
-/Overfull [\\][hv]box\|LaTeX Warning..Reference/{x;p;x;p}' std.log |
-    sed '/^.\+\.tex$/{N;s/\n/:/}' | fail || failed=1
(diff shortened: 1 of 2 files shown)


tkoeppe:
Note to self: extend the precommit scripts to run on Mac.

tkoeppe:
@burblebee: that change alone isn't enough to get the script to run on MacOS. I set up a MacOS presubmit on the `beta` branch, see here: https://github.com/cplusplus/draft/runs/2766096602

If you'd like to experiment further, feel free to just push more changes into that branch. When it all works, we can apply the full change to master.

tkoeppe:
I added the missing semicolon in awk. The other semicolons don't seem to be necessary.

That said, the bigger problem is that BSD-sed doesn't appear to allow nested commands, e.g. see https://unix.stackexchange.com/questions/13711/differences-between-sed-on-mac-osx-and-other-standard-sed. That breaks https://github.com/cplusplus/draft/blob/628ded4b8e6cd4b7297f9e04a394f200423ab2d7/tools/check-source.sh#L167. I've not managed to rewrite that.

For the time being, we can install gnu-sed on MacOS (`brew install gnu-sed`); that's at least good enough for the GitHub unit tests.

I'll close this for now. Please reopen if you have found a way to rewrite the sed command.

burblebee:
>> That said, the bigger problem is that BSD-sed doesn't appear to allow nested commands

@tkoeppe  Yup, that's what I got stuck on.  Let's just document in the scripts and on github that gnu-sed is required. Thanks for looking into this!

PR #4597, opened 2021-05-02
[index] Fix entry for 'program execution'

### LINKED ISSUE #4591: Index item 'program execution' range from page 10 to 80
The index item 'program execution' is marked from page 10 ([intro.abstract], 4.1.2), to page 80 (before [intro.multithread], 6.9.2).
I think this was caused by adding [intro.multithread] section, and latter restructured by moving both [intro.execution] and [intro.multithread] into [basic] as a whole, carrying the index ending mark within them.

I think the range of 'program execution' should contain the contents of both [intro.abstract] and [basic.exec] sections.

jensmaurer:
Fixes #4591

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -5405,6 +5405,7 @@ conversions\iref{expr.arith.conv}.
 \rSec1[basic.exec]{Program execution}
 
 \rSec2[intro.execution]{Sequential execution}
+\indextext{program execution|(}
 
 \pnum
 An instance of each object with automatic storage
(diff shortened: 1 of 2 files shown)


source/basic.tex
```diff
@@ -5404,6 +5404,8 @@
 
 \rSec1[basic.exec]{Program execution}
 
+\indextext{program execution|(}%
+
 \rSec2[intro.execution]{Sequential execution}
```
tkoeppe, line 5409: If you're going to have whitespace here anyway, then no need to bother escaping the newline. Alternatively, remove the blank line.

tkoeppe (changes requested):

source/basic.tex
jensmaurer: fixed

tkoeppe (approved):

PR #4602, opened 2021-05-17
[tuple.elem] Replace "member variables" with "data members"

morinmorin:
While I use both terms — data members and member variables — I don't think "member variable" is a defined term in the Standard.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -2209,7 +2209,7 @@ the return type is \tcode{X\&}, not \tcode{const X\&}.
 However, if the element type is a non-reference type \tcode{T},
 the return type is \tcode{const T\&}.
 This is consistent with how constness is defined to work
-for member variables of reference type.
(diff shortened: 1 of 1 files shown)


jwakely (approved):
To be really correct it should be "non-static data members", but it's only a note.

source/utilities.tex
```diff
@@ -2209,7 +2209,7 @@
 However, if the element type is a non-reference type \tcode{T},
 the return type is \tcode{const T\&}.
 This is consistent with how constness is defined to work
-for member variables of reference type.
+for data members of reference type.
```
jensmaurer, line 2212: -> "non-static data members"

jensmaurer (changes requested):

source/utilities.tex
morinmorin: Done!

jensmaurer (approved):

PR #4605, opened 2021-05-20
[expr.static.cast] static_cast of pointer to derived class only handles base class subobjects

randomnetcat:
`static_cast` from `base*` to `derived*` only handles the case where the `base` is a base class subobject of a `derived`, not when `base` is a member subobject. This brings the pointer wording in line with the reference wording:
> If the object of type “cv1 B” is actually a base class subobject of an object of type D, the result refers to the enclosing object of type D.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -3887,7 +3887,7 @@ to ``pointer to \tcode{B}'' exists\iref{conv.ptr}, the program is ill-formed.
 The null pointer value\iref{basic.compound} is converted
 to the null pointer value of the destination type. If the prvalue of type
 ``pointer to \cvqual{cv1} \tcode{B}'' points to a \tcode{B} that is
-actually a subobject of an object of type \tcode{D}, the resulting
+actually a base class subobject of an object of type \tcode{D}, the resulting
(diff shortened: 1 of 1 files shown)


CaseyCarter (approved):

jensmaurer (approved):

jensmaurer:
@tkoeppe, this looks sufficiently obvious to me to make it editorial. What do you think?

tkoeppe:
I agree. I'm not sure about the history, but if I had to speculate I'd guess that that wording is old, and leans on the "subtype" analogy, and may predate the current (entirely unrelated) notion of "subobject". (Maybe someone with an ARM could check?) The change seems an outright improvement. Feel free to merge!

jensmaurer:
The ARM didn't have `static_cast`, but the text about explicit type conversions on page 68 reads:

"Such a cast from a base to a derived class assumes that the object of the base class is a sub-object of an object of the derived class; the resulting pointer points to the enclosing object of the derived class."

Note that this puts "enclosing" in sufficient context to exclude the member case.

tkoeppe:
Great, thanks for adding this historical context!

PR #4607, opened 2021-05-21
[vector.bool] reference has multiple assignment operators

CaseyCarter:


jensmaurer:
This depends on [P2321](https://wg21.link/p2321) zip (Tim Song).

CaseyCarter:
> This depends on [P2321](https://wg21.link/p2321) zip (Tim Song) [cplusplus/papers#1003](https://github.com/cplusplus/papers/issues/1003)

I don't believe so; `vector<bool>::reference` currently has two assignment operators:
```c++
constexpr reference& operator=(const bool x) noexcept;
constexpr reference& operator=(const reference& x) noexcept;
```
So use of the singular "assignment operator" in the paragraph in question is incorrect. P2321 does add a third assignment operator, but that change is orthogonal to the fix in this PR.

jensmaurer:
Ah, right.

jensmaurer (approved):

PR #4608, opened 2021-05-21
[vector.bool] Remove top level const from signature

birbacher:


tkoeppe:
Good find! Is that the only case?

birbacher:
It's the only one that “const bool” would point at. :)

PR #4610, opened 2021-05-25
[list.ops] Fix name of parameter in list::unique effects

jwakely:


jwakely:
This has been wrong since C++98.

tkoeppe:
And just like Twin Peaks, only 25 years later it is made right. Thanks!

PR #4611, opened 2021-05-26
[dcl.meaning.general] Improve clarity of presentation

zygoloid:
[dcl.meaning.general] is hard to read currently, mostly due to a dense nesting of "if" / "otherwise" clauses where it's not always clear which wording is in the scope of which "if"s, and what each "otherwise" binds to. This PR attempts to clean this up a bit.

In passing, I also fixed an issue where we use the same name `S` for two different purposes in the same scope, and front-loaded the "declarator-id is normally not looked up" rule so that the special cases where lookup does happen are more obvious.

@opensdh Your review in particular would be appreciated.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -2460,6 +2460,8 @@ void h(int *(C[10]));           // \tcode{void h(int *(*_fp)(C _parm[10]));}
 \indextext{declaration!type}%
 A declarator contains exactly one \grammarterm{declarator-id};
 it names the entity that is declared.
+The terminal name of the \grammarterm{declarator-id} is not looked up
+except as specified below.
 If the \grammarterm{unqualified-id} occurring in a \grammarterm{declarator-id}
 is a \grammarterm{template-id},
 the declarator shall appear in the \grammarterm{declaration} of a
@@ -2484,8 +2486,9 @@ the \grammarterm{declarator-id} of the \grammarterm{declarator} is
 a \grammarterm{qualified-id} or a \grammarterm{template-id}:
 \begin{itemize}
   \item
+The terminal name of the \grammarterm{declarator-id} is looked up.
 If the friend declaration is not a template declaration,
-then in the lookup for the terminal name of $E$:
+then in this name lookup:
 \begin{itemize}
     \item
 if the \grammarterm{unqualified-id} in $E$ is a \grammarterm{template-id},
@@ -2507,8 +2510,8 @@ they shall all have the same target scope, and
 the target scope of the \grammarterm{declarator} is that scope.
 \end{itemize}
 \item
-Otherwise, the terminal name of $E$ is not looked up.
-The declaration's target scope is the innermost enclosing namespace scope.
+Otherwise,
+the declaration's target scope is the innermost enclosing namespace scope.
 If the declaration is contained by a block scope,
 the declaration shall correspond to a declaration
 that inhabits the innermost block scope.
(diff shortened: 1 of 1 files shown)


source/declarations.tex
```diff
@@ -2526,15 +2529,19 @@
...
 the \grammarterm{declarator} does not bind a name.
-If it declares a class member,
-the terminal name of the \grammarterm{declarator-id} is not looked up;
-otherwise, only those lookup results that are nominable in $S$
+If such a \grammarterm{declarator} does not declare a class member,
+the terminal name of the \grammarterm{declarator-id} is looked up,
+and only those lookup results that are nominable in $S$
 are considered when identifying
-any function template specialization being declared\iref{temp.deduct.decl}.
+any function or variable template specialization
```
zygoloid, line 2543: @opensdh Was the different treatment of function and variable template specializations intended?

source/declarations.tex
```diff
@@ -2526,15 +2529,19 @@
...
+Then:
+\begin{itemize}
 \item
 If the \grammarterm{declarator} declares
 an explicit instantiation or a partial or explicit specialization,
 the \grammarterm{declarator} does not bind a name.
-If it declares a class member,
-the terminal name of the \grammarterm{declarator-id} is not looked up;
-otherwise, only those lookup results that are nominable in $S$
+If such a \grammarterm{declarator} does not declare a class member,
```
zygoloid, line 2539: @opensdh I tried to figure out why we're treating class members differently from non-class-members here and didn't succeed. Perhaps an example would help? I'm also not sure how we're supposed to handle an explicit specialization of a class member template without performing name lookup. Was the intent here instead that if the specialization or instantiation is for a non-template member of a class template specialization, no lookup is performed?

opensdh:
I will get to this, but I'm out for about a week starting now (and expect to be at least as busy with Real Work when I return as I am now).

source/declarations.tex
opensdh: It was (although that doesn't guarantee that it's _good_ to do so!).  The idea is that we do ordinary lookup for the name before the `<` (and for an explicit specialization/instantiation without an explicit template argument list, for consistency): if we find a variable template, it must be the unique lookup result, so there isn't any reason to disqualify that one result because of nominability.  I think we did talk about this in the review, but I can't find the minutes for it.

source/declarations.tex
opensdh: I think that you're right: we do have to do the lookup for the same reason that we have to do it when encountering `template<> my_namespace::my_class<` since it might continue `int>::sub_template<int> {` or just `int> {`.  You're probably right about the intent too, but I'm less clear on that at the moment.  (My plan is to, tomorrow, look at _your_ wording here and see if I see any problems with it of a similar sort; if not, it stands to reason that it reflects the intent better.)

source/declarations.tex
```diff
@@ -2507,15 +2510,15 @@
...
+Otherwise,
+the declaration's target scope is the innermost enclosing namespace scope.
+If the declaration is contained by a block scope,
 the declaration shall correspond to a declaration
 that inhabits the innermost block scope.
 \end{itemize}
 
 \pnum
-Otherwise:
+If the declaration is not a friend declaration:
```
opensdh, line 2521: As this is exhaustive together with the previous paragraph, I'd prefer to use one of our formulations that indicates that: perhaps
> Otherwise, the declaration is not a friend declaration:

or
> Otherwise (for a non-friend declaration):

source/declarations.tex
```diff
@@ -2526,15 +2529,19 @@
 \item
 Otherwise, let $S$ be the entity associated with the scope inhabited by
 the \grammarterm{declarator}.
+\end{itemize}
+Then:
+\begin{itemize}
 \item
 If the \grammarterm{declarator} declares
 an explicit instantiation or a partial or explicit specialization,
 the \grammarterm{declarator} does not bind a name.
```
opensdh, line 2538: This is actually redundant: such declarations do not introduce names ([temp.explicit]/5, [temp.expl.spec]/4, and [temp.spec.partial.general]/7), so [basic.scope.scope]/2.5 ignores them already.  This is relevant because it is an opportunity to merge into the next sentence (once we're sure exactly what we mean it to mean).

opensdh (approved):
This is a less drastic change than I expected before reading it carefully.  The change about variable templates doesn't seem eminently editorial, but neither does it necessarily need more than a cursory CWG review.

tkoeppe:
@opensdh, @zygoloid: if you can agree on a clear improvement, we can try to land this for the upcoming WD, but no rush.

PR #4614, opened 2021-05-29
[misc] Add index of Working Drafts

jensmaurer:


tkoeppe:
Looks great! I was also thinking of keeping this on a separate page, and the `papers` directory is a good place. If you're done with this, feel free to merge!

jensmaurer:
@tkoeppe , please make sure to include a rerun of that script and a check-in of the produced list in your procedure of preparing a new working draft, just after "prepare for post-X drafts" (when you replace the N number with an xxxx in the PDF). 

Yes, the link will be dangling until the mailing is released, but who cares.

PR #4615, opened 2021-05-29
[expr.prim.lambda] Replace 'unnamed' with 'unique name'

jensmaurer:
Also add a definition of 'unique name' in [intro.defs].

As discussed in the thread starting at http://lists.isocpp.org/core/2021/05/11017.php


tkoeppe:
Let's wait a few more days. CWG has been discussing this and seems to agree, but that's only recent, and not everyone may have had a chance to catch up with the discussion.

@RemindMeBot: come back June 5 please.

jensmaurer:
This needs a CWG nod: https://wiki.edg.com/bin/view/Wg21telecons2021/CoreWorkingGroup

PR #4617, opened 2021-05-30
[expr.rel] Clarify function pointer comparisons

### LINKED ISSUE #3850: [expr.rel] Unclear how function pointers are compared
The intent of the wording is that result of `p < q` for unequal function pointers is unspecified - but the way we get there is that [expr.rel]/4 defines when pointers to objects compare greater and then [expr.rel]/5 we fall through to "Otherwise, the result of each of the operators is unspecified." 

Can we clean this up by inserting a new paragraph between [expr.rel]/4 and [expr.rel]/5 as follows:

> The result of comparing unequal pointers to functions ~or pointers to member functions~ is unspecified. 

Or something?

jensmaurer:
Fixes #3850

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -6257,7 +6257,7 @@ and their class is not a union.
 \pnum
 If two operands \tcode{p} and \tcode{q} compare equal\iref{expr.eq},
 \tcode{p<=q} and \tcode{p>=q} both yield \keyword{true} and \tcode{p<q} and
-\tcode{p>q} both yield \keyword{false}. Otherwise, if a pointer \tcode{p}
+\tcode{p>q} both yield \keyword{false}. Otherwise, if a pointer to object \tcode{p}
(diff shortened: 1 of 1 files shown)


source/expressions.tex
```diff
@@ -6248,6 +6248,9 @@
 
 \end{itemize}
 
+When comparing two unequal pointers to functions,
```
tkoeppe, line 6251: Hm, this is a somewhat unusual way of saying things. How would you feel about the usual "The result of comparing two unequal pointers is unspecified"?

jensmaurer: This is a bit of a tricky two-stage process. We first define "(is required to) compare greater" and then we define the results of the comparison operators based on that (partial) order. But that's only for pointer to object.

I've updated the second sentence where the results of the actual relational operators are defined. This feels better now.

tkoeppe: Agreed, this is nice. Thanks!

PR #4620, opened 2021-05-31
Check for misordered library elements

### LINKED ISSUE #3677: Check element order programmatically and presubmit

jensmaurer:
Fixes #3677

### DIFF
--- a/tools/check-source.sh
+++ b/tools/check-source.sh
@@ -73,6 +73,10 @@ grep -ne '^\\\(constraints\|mandates\|expects\|effects\|sync\|ensures\|returns\|
 # Fixup: sed 's/^\\\(constraints\|mandates\|expects\|effects\|sync\|ensures\|returns\|throws\|complexity\|remarks\|errors\)\s*\(.\)/\\\1\n\2/'
 # Fixup: sed 's/^\\ //'
 
+# Order of library elements.
+../tools/element-order.awk *.tex |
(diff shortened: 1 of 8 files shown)


tools/check-source.sh
```diff
@@ -73,6 +73,10 @@ grep -ne '^\\\(constraints\|mandates\|expects\|effects\|sync\|ensures\|returns\|
 # Fixup: sed 's/^\\\(constraints\|mandates\|expects\|effects\|sync\|ensures\|returns\|throws\|complexity\|remarks\|errors\)\s*\(.\)/\\\1\n\2/'
 # Fixup: sed 's/^\\ //'
 
+# Order of library elements.
+../tools/element-order.awk *.tex |
+    fail '' || failed=1
```
tkoeppe, line 78: Do you not have a message here because the tool already prints its own message? Might be nice to have a generic message here regardless (if for nothing but documentation of the script itself), e.g. "incorrect ordering of library elements".

tkoeppe (approved):
This is great, thanks!

tools/check-source.sh
jensmaurer: Fixed + force-pushed.

Issue #4592, opened 2021-04-23
Nested-name-specifier is not mentioned for member definition outside its target scope


xmh0511:
In c++20 standard, the member functions or static member data which are these members  that can be defined outside its enclosing class has a normative rule for how to define them outside the enclosing class, they are 
For member functions 
[class#mfct-3](https://timsong-cpp.github.io/cppwp/n4861/class#mfct-3)
> In the definition at namespace scope, the name of the static data member shall be qualified by its class name using the ​::​operator.  
For static data member 
[class#static.data-3](https://timsong-cpp.github.io/cppwp/n4861/class#static.data-3)
> In the definition at namespace scope, the name of the static data member shall be qualified by its class name using the ​::​operator.  

Similar for the definition of members of a namespace 
[namespace.memdef#2](https://timsong-cpp.github.io/cppwp/n4861/namespace.memdef#2) 
> Members of a named namespace can also be defined outside that namespace by explicit qualification ([namespace.qual]) of the name being defined, provided that the entity being defined was already declared in the namespace and the definition appears after the point of declaration in a namespace that encloses the declaration's namespace.  

I tried to find out the alternative normative rule in the current draft that can replace these omission rules, however, I don't find them out. Is it considered a bit radical modification that excises these rules?

jensmaurer:
These permissions are, in fact, just redeclarations that happen to be definitions. The generic redeclaration rules handle these cases, I believe.

xmh0511:
> These permissions are, in fact, just redeclarations that happen to be definitions. The generic redeclaration rules handle these cases, I believe.

I doubt that these rules in the above are intended to be covered by [dcl.meaning]. However, it's not explicitly. I have to try to read [dcl.meaning] more carefully to check whether the meaning is readable.

languagelawyer:
I think what are you looking for is [expr.prim.id.qual]/2 and /3.

xmh0511:
> I think what are you looking for is [expr.prim.id.qual]/2 and /3. (Yes, it is an unexpected place for declaration rules).
[expr.prim.id.qual]/2 and /3 are basiclly introducing the concept about declarative nested-name-specifier, it is useful for determining where can the declaration inhabit. It didn't state that we should redeclare the member by using nested-name-specifier.

jensmaurer:
By what other means would you possibly refer to that member? If you omit the declarative nested-name-specifier, you're not redeclaring the entity you introduced inside the class or namespace definition.

[...]

PR #4624, opened 2021-06-03
[class.pre] Fix incorrect comment in example

jensmaurer:
Partially addresses #4592

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -106,7 +106,7 @@ namespace N {
 }
 using N::A;
 template<class T> struct A<T>::B {};    // OK
-template<> struct A<void> {};           // error: \tcode{A} not nominable in \tcode{::}
+template<> struct A<void> {};           // OK
 \end{codeblock}
 \end{example}
 


tkoeppe:
@opensdh: could you please take a look?

opensdh:
I [commented](https://github.com/cplusplus/draft/issues/4592#issuecomment-859718676) on the issue already.

tkoeppe:
@opensdh: OK, thanks! So we take this PR as a partial improvement, but I'll leave it to you (= collectively) to come up with a more comprehensive example?

opensdh:
If it's up to me, we'd make the full change I suggested in one go, since @xmh0511 correctly pointed out that merely correcting the example leaves it a bit mysterious (apparently misplaced).  I can always do the rest later, though, since some more work is already needed (as also noted).

tkoeppe:
Thanks. I agree that we should get the full rewrite, but let's take this partial fix for now so that we're at least not saying anything wrong, even if it's mysterious. But I continue looking forward to the full solution!

PR #4628, opened 2021-06-04
[range.prim.data] Use ranges::begin(t) not ranges::begin(E)

jwakely:
This is consistent with the typo fixes for ranges::size done in
aa9c660a835540117123617a13b0ba1ab6dd801e as part of #3752.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -919,7 +919,7 @@ Then:
   is a valid expression whose type models
   \libconcept{contiguous_iterator},
   \tcode{ranges::data(E)} is expression-equivalent to
-  \tcode{to_address(ranges::begin(E))}.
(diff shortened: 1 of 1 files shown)


jwakely:
As I said at https://github.com/cplusplus/draft/pull/3752#discussion_r645502395

> We check that `ranges::begin(t)` is valid, and then do `ranges::begin(E)` instead. It should have the same effects, but it's inconsistent and confusing to switch between `t` and `E`.

CaseyCarter (approved):
I have no idea why I didn't make this change in P2091 - it seems like an obvious consistency fix.

jensmaurer (approved):

jensmaurer (approved):

PR #4683, opened 2021-06-10
P2325R3 Views should not be required to be default constructible

### LINKED ISSUE #4656: [2021-06 LWG Motion 19] P2325R3 Views should not be required to be default constructible
[P2325R3](https://wg21.link/p2325r3)

jensmaurer:
Also fixes LWG3479.

Fixes  #4656
Fixes cplusplus/papers#1007

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -10866,8 +10866,7 @@ namespace std {
     class span;
 
   template<class ElementType, size_t Extent>
-    inline constexpr bool ranges::enable_view<span<ElementType, Extent>> =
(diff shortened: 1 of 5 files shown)


tkoeppe:
@jensmaurer: what was the reason for using "non-propagating-cache for split_view" (for the report)?

tkoeppe:
@jensmaurer: Oh, I see, you changed `optional<iterator_t<...>>` to `non-propagating-cache<iterator_t<...>>`, because of Motion 18. And there's even an editor's note to do so.

jensmaurer:
Yes.

PR #4691, opened 2021-06-11
[conv.general] Fix index redirect for contextual conversion to bool.

Eelis:
Here's the entry being redirected to:

https://github.com/cplusplus/draft/blob/fb3bea889899c5ea5c0f70b12fc72129f7d1692a/source/expressions.tex#L552


jensmaurer (approved):

Eelis:
Amended to also style `bool` as code in the redirecting entry.

PR #4693, opened 2021-06-11
P2386R0 Core Language Working Group "ready" Issues for the June, 2021 meeting

### LINKED ISSUE #4612: [intro.execution] Sequencing of evaluations inside called functions CWG2466
After application of CWG2466, the note that immediately follows the changed text:

[Note 7 : If A and B would not otherwise be sequenced then they are indeterminately sequenced. — end note]

is no longer good: A and B no longer exist, and we've actually changed the normative text to say "A before all B or all A before B", so "indeterminately sequenced" no longer fits directly.

The note should be struck.

### LINKED ISSUE #4631: [2021-06 CWG Motion 1] P2386R0 Core Language Working Group "ready" Issues for the June, 2021 meeting
[P2386R0](https://wg21.link/p2386r0)

jensmaurer:
Fixes #4631 
Fixes #4612
Partially addresses #3262

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -6432,7 +6432,7 @@ a program.
 The linkage\iref{basic.link} of \tcode{main} is
 \impldef{linkage of \tcode{main}}. A program that defines \tcode{main} as
 deleted or that declares \tcode{main} to be
-\keyword{inline}, \keyword{static}, or \keyword{constexpr} is ill-formed.
+\keyword{inline}, \keyword{static}, \keyword{constexpr}, or \keyword{consteval} is ill-formed.
 The function \tcode{main} shall not be a coroutine\iref{dcl.fct.def.coroutine}.
 The \tcode{main} function shall not be declared with a
 \grammarterm{linkage-specification}\iref{dcl.link}.
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -5674,24 +5674,26 @@ void g(int i) {
 \end{example}
 
 \pnum
-When calling a function (whether or not the function is inline), every
-\indextext{value computation}%
-value computation and
-\indextext{side effects}%
-side effect associated with any argument
-expression, or with the postfix expression designating the called
-function, is sequenced before execution of every expression or statement
+When invoking a function (whether or not the function is inline),
+every argument expression and
+the postfix expression designating the called function
+are sequenced before every expression or statement
 in the body of the called function.
-For each function invocation \placeholder{F},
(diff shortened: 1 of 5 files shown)


languagelawyer:
Shouldn't CWG2458 resolution also modify [basic.lval] wording, as in #3262?

jensmaurer:
> Shouldn't CWG2458 resolution also modify [basic.lval] wording, as in #3262?

Maybe it should, but it's not part of the motion, so we should revisit after we're done with the motions.
I think `C::f`, when it is an lvalue, still "identifies a function" per [basic.lval].

languagelawyer:
> I think `C::f`, when it is an lvalue, still "identifies a function"

This is fine. The thing is that `C::f` identifies a function when it is a prvalue, whilst [basic.lval] says that evaluated expressions denoting functions are glvalues. And I think class member access expression `e.C::f` is where `C::f` is evaluated and is a prvalue.

jensmaurer:
> And I think class member access expression e.C::f is where C::f is evaluated and is a prvalue.

Ok, but then the proposed change in #3262 isn't actually what we want post-CWG2458.
Instead, we want to augment the definition of "prvalue" such that `C::f` fits.

languagelawyer:
>  Instead, we want to augment the definition of "prvalue" such that `C::f` fits.

The current definition — A _prvalue_ is an expression whose evaluation ... computes the value of an operand of an operator — looks fine to me, since `C::f` computes the value of the right operand of the `.` or `->` operator. The definition of «glvalue» should say that some expressions denoting a non-static member functions are not glvalues.

jensmaurer:
The "glvalue" and "prvalue" definitions do not exhaustively describe which expressions yield which kinds. I agree that `C::f` is the operand of an operator. But `C::f` also designates a function (in case we take its address). So, this seems fine both ways.

tkoeppe (approved):

languagelawyer:
> The "glvalue" and "prvalue" definitions do not exhaustively describe which expressions yield which kinds.

I'm fine with it, but let me remind that there was [CWG1076](http://wg21.link/cwg1076) issue about value category inconsistency:
> The taxonomy of value categories in 7.2.1 [basic.lval] classifies temporaries as prvalues. However, some temporaries are explicitly referred to as lvalues (cf 14.2 [except.throw] paragraph 3).

which can be reformulated for CWG2458 resolution:
> The taxonomy of value categories in [basic.lval] classifies function designators as glvalues. However, some non-static member function designators are explicitly referred to as prvalues

PR #4695, opened 2021-06-12
P2385R0 C++ Standard Library Issues to be moved in Virtual Plenary, June 2021

### LINKED ISSUE #1793: I do not think [fs.req.namespaces] means what it thinks it means LWG2818

### LINKED ISSUE #4221: [forwardlist.ops] bogus specification of forward_list::unique LWG2997
The specification for `forward_list::unique` says:

> Erases all but the first element from every consecutive group of equal elements referred to by the iterator `i` in the range `[first + 1, last)` for which `*i == *(i-1)` (for the version with no arguments) or `pred(*i, *(i - 1))` (for the version with a predicate argument) holds.

Err, what?

Given `{1, 1, 1, 1, 1, 2, 2, 2, 2, 2}`, we:

1) consider the range `[first + 1, last)`: `{1, 1, 1, 1, 2, 2, 2, 2, 2}`.
2) identify the elements for which `*i == *(i-1)`, dropping the first `2`.
3) identify the consecutive groups of equal elements from step (2): `{1, 1, 1, 1}` and `{2, 2, 2, 2}`. 
4) erase all but the first element from each such group.

The result is `{1, 1, 2, 2}`. That doesn't seem right.

The problem here is that we express basically every part of the rules twice or three times in the same sentence, and usually at least all but one of those repetitions is wrong. (Eg, compare "equal" to the more careful description later, compare "all but the first" with the use of `[first + 1, last)` and `*i == *(i-1)`.)

Perhaps we meant:

> Erases all but the first element from every consecutive group of equal elements<ins>. That is, for a nonempty list, erases all elements</ins> referred to by the iterator `i` in the range `[first + 1, last)` for which `*i == *(i-1)` (for the version with no arguments) or `pred(*i, *(i - 1))` (for the version with a predicate argument) holds.

Oh, and while we're here: what are `first` and `last` supposed to be here? I think it means `begin()` and `end()`...

### LINKED ISSUE #4640: [2021-06 LWG Motion 3] P2385R0 C++ Standard Library Issues to be moved in Virtual Plenary, June 2021

jensmaurer:
Fixes #4640
Fixes #1793
Fixes #4221

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -4651,6 +4651,12 @@ In this subclause,
 arguments for a template parameter
 named \tcode{Predicate} or \tcode{BinaryPredicate}
 shall meet the corresponding requirements in \ref{algorithms.requirements}.
+The semantics of \tcode{i + n},
+where \tcode{i} is an iterator into the list and \tcode{n} is an integer,
+are the same as those of \tcode{next(i, n)}.
+The expression \tcode{i - n},
+where \tcode{i} is an iterator into the list and \tcode{n} is an integer,
+means an iterator \tcode{j} such that \tcode{j + n == i} is \tcode{true}.
 For \tcode{merge} and \tcode{sort},
 the definitions and requirements in \ref{alg.sorting} apply.
 
@@ -4784,16 +4790,24 @@ Stable\iref{algorithm.stable}.
 \indexlibrarymember{unique}{forward_list}%
 \begin{itemdecl}
 size_type unique();
-template<class BinaryPredicate> size_type unique(BinaryPredicate pred);
+template<class BinaryPredicate> size_type unique(BinaryPredicate binary_pred);
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
+Let \tcode{binary_pred} be \tcode{equal_to<>\{\}} for the first overload.
+
+\pnum
+\expects
+\tcode{binary_pred} is an equivalence relation.
+
+\pnum
 \effects
 Erases all but the first element from every consecutive
-group of equal elements referred to by the iterator \tcode{i} in the range \range{first +
-1}{last} for which \tcode{*i == *(i-1)} (for the version with no arguments) or \tcode{pred(*i,
-*(i - 1))} (for the version with a predicate argument) holds.
+group of equivalent elements.
+That is, for a nonempty list, erases all elements referred to
+by the iterator \tcode{i} in the range \range{begin() + 1}{end()}
+for which \tcode{binary_pred(*i, *(i - 1))} is \tcode{true}.
 Invalidates only the iterators and references to the erased elements.
 
 \pnum
@@ -4802,11 +4816,14 @@ The number of elements erased.
 
 \pnum
 \throws
-Nothing unless an exception is thrown by the equality comparison or the predicate.
+Nothing unless an exception is thrown by the predicate.
 
 \pnum
 \complexity
-If the range \range{first}{last} is not empty, exactly \tcode{(last - first) - 1} applications of the corresponding predicate, otherwise no applications of the predicate.
+If \tcode{empty()} is \tcode{false},
+exactly \tcode{distance(begin(), end()) - 1} applications of
+the corresponding predicate,
+otherwise no applications of the predicate.
 \end{itemdescr}
 
 \indexlibrarymember{merge}{forward_list}%
@@ -5296,6 +5313,10 @@ In this subclause,
 arguments for a template parameter
 named \tcode{Predicate} or \tcode{BinaryPredicate}
 shall meet the corresponding requirements in \ref{algorithms.requirements}.
+The semantics of \tcode{i + n} and \tcode{i - n},
+where \tcode{i} is an iterator into the list and \tcode{n} is an integer,
+are the same as those of \tcode{next(i, n)} and \tcode{prev(i, n)},
+respectively.
 For \tcode{merge} and \tcode{sort},
 the definitions and requirements in \ref{alg.sorting} apply.
 
@@ -5475,12 +5496,19 @@ template<class BinaryPredicate> size_type unique(BinaryPredicate binary_pred);
 
 \begin{itemdescr}
 \pnum
+Let \tcode{binary_pred} be \tcode{equal_to<>\{\}} for the first overload.
+
+\pnum
+\expects
+\tcode{binary_pred} is an equivalence relation.
+
+\pnum
 \effects
 Erases all but the first element from every
-consecutive group of equal elements referred to by the iterator \tcode{i} in the range
-\range{first + 1}{last} for which \tcode{*i == *(i-1)} (for the version of
-\tcode{unique} with no arguments) or \tcode{binary_pred(*i, *(i - 1))} (for the version of
-\tcode{unique} with a predicate argument) holds.
+consecutive group of equivalent elements.
+That is, for a nonempty list, erases all elements referred to
+by the iterator \tcode{i} in the range \range{begin() + 1}{end()}
+for which \tcode{binary_pred(*i, *(i - 1))} is \tcode{true}.
 Invalidates only the iterators and references to the erased elements.
 
 \pnum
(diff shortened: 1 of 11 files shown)


source/containers.tex
```diff
@@ -4651,6 +4651,13 @@
 arguments for a template parameter
 named \tcode{Predicate} or \tcode{BinaryPredicate}
 shall meet the corresponding requirements in \ref{algorithms.requirements}.
+The semantics of \tcode{i + n},
+where \tcode{i} is an iterator into the list and \tcode{n} is an integer,
+are the same as those of \tcode{next(i, n)}.
+The expression \tcode{i - n},
+where \tcode{i} is an iterator into the list and \tcode{n} is an integer,
+means an iterator \tcode{j} such that \tcode{j + n == i} is \tcode{true}.
+
```
JohelEGP, line 4660:
```suggestion
```

source/containers.tex
```diff
@@ -4784,16 +4791,24 @@
 \indexlibrarymember{unique}{forward_list}%
 \begin{itemdecl}
 size_type unique();
-template<class BinaryPredicate> size_type unique(BinaryPredicate pred);
+template<class BinaryPredicate> size_type unique(BinaryPredicate binary_pred);
 \end{itemdecl}
 
 \begin{itemdescr}
+\pnum
+Let \tcode{binary_pred} be \tcode{equal_to<>\{\}} for the first overload.
```
JohelEGP, line 4799:
```suggestion
Let \tcode{binary_pred} be \tcode{equal_to\{\}} for the first overload.
```

source/containers.tex
```diff
@@ -5474,13 +5496,20 @@
...
```
JohelEGP, line 5500:
```suggestion
Let \tcode{binary_pred} be \tcode{equal_to\{\}} for the first overload.
```

source/iostreams.tex
```diff
@@ -10470,6 +10482,21 @@
 Equivalent to: \tcode{basic_ofstream(s.c_str(), mode)}.
 \end{itemdescr}
 
+\indexlibraryctor{basic_ofstream}%
+\begin{itemdecl}
+explicit basic_ofstream(const T& s, ios_base::openmode mode = ios_base::out);
```
JohelEGP, line 10487:
```suggestion
template<class T>
  explicit basic_ofstream(const T& s, ios_base::openmode mode = ios_base::out);
```

source/containers.tex
```diff
@@ -10265,6 +10317,75 @@
...
+  priority_queue(InputIterator first, InputIterator last, const Compare& compare,
+                 const Container& cont, const Alloc& a);
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\effects
+Initializes \tcode{c} with
+\tcode{cont} as the first argument and \tcode{a} as the second argument, and
+initializes \tcode{comp} with compare;
```
JohelEGP, line 10366:
```suggestion
initializes \tcode{comp} with \tcode{compare};
```

source/ranges.tex
```diff
@@ -5150,7 +5152,8 @@
     }
 
     friend constexpr void iter_swap(const @\exposid{iterator}@& x, const @\exposid{iterator}@& y)
-      noexcept(noexcept(ranges::iter_swap(x.@\exposid{inner_}@, y.@\exposid{inner_}@)));
+      noexcept(noexcept(ranges::iter_swap(x.@\exposid{inner_}@, y.@\exposid{inner_}@)))
+      requires indirectly_swappable<@\exposid{InnerIter}@>;
```
JohelEGP, line 5156:
```suggestion
      requires \libconcept{indirectly_swappable}<@\exposid{InnerIter}@>;
```

JohelEGP (changes requested):

JohelEGP (changes requested):

source/containers.tex
jensmaurer: The incoming text has the `<>` here.

source/containers.tex
jensmaurer: The incoming text has the `<>` here.

source/containers.tex
jensmaurer: fixed

source/utilities.tex
```diff
@@ -15221,27 +15221,45 @@
...
 \item \tcode{f} is a null function pointer value.
 \item \tcode{f} is a null member pointer value.
-\item \tcode{F} is a specialization of the \tcode{function} class template, and
-  \tcode{!f}.
+\item \tcode{remove_cvref_t<F>} is
+a specialization of the \tcode{function} class template, and
+\tcode{!f} is \tcode{true}.
 \end{itemize}
 
 \pnum
```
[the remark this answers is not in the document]
jensmaurer: fixed

source/ranges.tex
```diff
@@ -2253,6 +2255,40 @@
 \exposid{bound_} with \tcode{bound}.
 \end{itemdescr}
 
+\indexlibraryctor{iota_view}%
+\begin{itemdecl}
+constexpr iota_view(iterator first, @\seebelow@ last);
```
JohelEGP, line 2260:
```suggestion
constexpr iota_view(@\exposid{iterator}@ first, @\seebelow@ last);
```

source/ranges.tex
```diff
@@ -2253,6 +2255,40 @@
...
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\effects
+Equivalent to:
+\begin{itemize}
+\item
+If \tcode{same_as<W, Bound>} is \tcode{true},
```
JohelEGP, line 2269:
```suggestion
If \tcode{\libconcept{same_as}<W, Bound>} is \tcode{true},
```

source/ranges.tex
```diff
@@ -2253,6 +2255,40 @@
...
+\item
+Otherwise, \tcode{iota_view(first.\exposid{value_}, last.\exposid{bound_})}.
+\end{itemize}
+
+\pnum
+\remarks
+The type of \tcode{last} is:
+\begin{itemize}
+\item
+If \tcode{same_as<W, Bound>} is \tcode{true}, \exposid{iterator}.
```
JohelEGP, line 2283:
```suggestion
If \tcode{\libconcept{same_as}<W, Bound>} is \tcode{true}, \exposid{iterator}.
```

source/utilities.tex
```diff
@@ -2046,14 +2046,15 @@
 Given the exposition-only function:
 \begin{codeblock}
 template<class T, class Tuple, size_t... I>
+  requires @\libconcept{is_constructible_v}@<T, decltype(get<I>(declval<Tuple>()))...>
```
JohelEGP, line 2049:
```suggestion
  requires is_constructible_v<T, decltype(get<I>(declval<Tuple>()))...>
```

source/containers.tex
```diff
@@ -10158,13 +10217,32 @@
...
+\end{itemdescr}
+
 \indexlibraryctor{priority_queue}%
 \begin{itemdecl}
 template<class InputIterator>
   priority_queue(InputIterator first, InputIterator last, const Compare& x, const Container& y);
 template<class InputIterator>
-  priority_queue(InputIterator first, InputIterator last, const Compare& x = Compare(),
-                 Container&& y = Container());
+  priority_queue(InputIterator first, InputIterator last, const Compare&, Container&& y);
```
JohelEGP, line 10245:
```suggestion
  priority_queue(InputIterator first, InputIterator last, const Compare& x, Container&& y);
```

JohelEGP (changes requested):

JohelEGP (changes requested):

source/utilities.tex
```diff
@@ -21011,8 +21071,8 @@
 \begin{codeblock}
 namespace std {
   template<class Context, class... Args>
-  struct @\placeholder{format-arg-store}@ {      // \expos
-    array<basic_format_arg<Context>, sizeof...(Args)> args;
+  class @\exposid{format-arg-store}@ {                                      // \expos
```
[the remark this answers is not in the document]
jensmaurer: Yes; postponed to https://github.com/cplusplus/draft/issues/4698

jensmaurer:
@JohelEGP , thanks. Applied + force-pushed.

tkoeppe:
Thank you both, stellar work!

Issue #1280, opened 2016-12-21
Inconsistencies with "object declarations", "declarations that declare objects" and "class members of object type"


RealLitb:
Richard asked me to add the following std-discussion content as an issue (the original question was about "A variable is introduced by the declaration of a reference other than a non-static data member or of an object.").

>> I still don't understand. I can see why a non-static data member is
>> not the declaration of an object.. because the (sub-) object is only
>> there once we create an object of the membering class. But how do
>> references differ in that regard? They too are not "there
>> immediately", but require an object of the membering class to
>> necessiate the beginning of the lifetime of a reference for each such
>> object.
>
>
> I suppose the problem is in the ambiguity of what a "declaration of a
> reference" is. I believe the intent here is that a "declaration of a
> reference" is a declaration that declares a particular name to be of
> reference type (which a non-static data member declaration of reference type
> does), not necessarily a declaration that causes the lifetime of a
> particular reference to begin. Conversely, a "declaration of an object" is
> intended to be interpreted as a declaration that declares a particular name
> to name a specific object (which a non-static data member declaration does
> not).
>
> That is:
>
>    struct A {
>      int &a;
>      int b;
>    };
>
> 'a' is a declaration of a reference, and so we need another condition to
> restrict it from being a variable. 'b' is /not/ a declaration of an object,
> so we don't need to say anything further.
>
> Suggestions on how to reword this to make it clearer would be welcome :)
>

OK, thanks I see. But, is this intent worded somewhere in the spec
aswell? As far as declarations that contain decl-specifier-seq are
concerned we have at 7p10 that says

> "If the decl-specifier-seq contains no typedef specifier, the declaration is called a function declaration if the type associated with the name is a function type ([dcl.fct]) and an object declaration otherwise.".

Am am not totally sure whether this only applies to
simple-declarations (because of the location of that paragraph), or
whether it applies to all declarations that contain
decl-specifier-seqs. Another thing is, because it indicate that an
"object declaration" is not quite the same as a "declaration of an
object", it becomes very involved. Are the following summaries correct
based on your answers and the spec?

    struct A  {
       int &b; // Not the declaration of an object. But an object declaration. And the declaration of a reference
       int c; // Not the declaration of an object. But an object declaration
   };

    extern int &d; // Not the declaration of an object. But an object declaration. And the declaration of a reference
    int e; // The declaration of an object. And an object declaration. But not the declaration of a reference

Are they correct concerning the intent? Would it not make sense to
[...]

PR #4696, opened 2021-06-12
Clarify definition of "variable"

t3nsor:
In the current definition, the special exception for non-static data members is only stated for references, which gives the misleading impression that non-reference non-static data member declarations are variable declarations. By re-wording the definition, we can make it clear that a non-static data member declaration is never a variable declaration.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -92,8 +92,11 @@ by a \grammarterm{typedef-name} introduced by a declaration specifying $E$.
 \pnum
 A \defn{variable} is introduced by the
 declaration of
-a reference other than a non-static data member or of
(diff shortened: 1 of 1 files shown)


languagelawyer:
In #1280, [it was decided that «declaration of an object» should be handled in a CWG issue](https://github.com/cplusplus/draft/issues/1280#issuecomment-549590504). I think this PR partially addresses #1280 and thus can't be editorial.

jensmaurer:
We need to properly divorce "variable" and "object"; recursive functions have a single variable, but several objects (one per activation frame). Fixing this it not editorial at all.

PR #4709, opened 2021-06-19
[basic, class, dcl] Fix ranged index entries

tkoeppe:


tkoeppe:
@jensmaurer: do the ranges look reasonable? I'm enclosing the entire subclause that follows the range-open in each case.

jensmaurer (approved):
Looks good.

PR #4714, opened 2021-06-22
Better string splitting example

cor3ntin:
With the adoption of P1989, string_view is constructible from
subrange, we can use that to provide a good example of
string splitting.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -6192,10 +6192,8 @@ the expression \tcode{views::split(E, F)} is expression-equivalent to
 \begin{example}
 \begin{codeblock}
 string str{"the quick brown fox"};
-for (span<char const> word : split(str, ' ')) {
-  for (char ch : word)
-    cout << ch;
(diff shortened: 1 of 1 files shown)


source/ranges.tex
```diff
@@ -6192,10 +6192,8 @@
 \begin{example}
 \begin{codeblock}
 string str{"the quick brown fox"};
-for (span<char const> word : split(str, ' ')) {
-  for (char ch : word)
-    cout << ch;
-  cout << '*';
+for (string_view word : split(str, ' ')) {
+  cout <<  word << '*';
```
JohelEGP, line 6196:
```suggestion
  cout << word << '*';
```

cor3ntin: Thanks

jwakely (approved):

tkoeppe:
Thank you!

For future reference, please try to prefix commit messages with the label of the affected subclause(s).

PR #4715, opened 2021-06-22
[iterator.operations, range.iter.op.distance] Reword "get to".

### LINKED ISSUE #2807: [iterator.operations] It should be stated clearly that `std::distance(first, last)` might invalidate all copies of `first` if it is not a ForwardIterator.

tkoeppe:
It was previously not explicitly stated that input iterators would
actually be incremented (which invalidates copies). The new wording is
more explicit about how the distance is measured, and in doing so
calls out more explicitly that input iterators are indeed incremented.

Fixes #2807.
Fixes #2813.

This is an alternative approach to #2813, as suggested by @jensmaurer in the comments.

@jwakely, @CaseyCarter, I'd welcome your review on this!

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -2763,11 +2763,9 @@ the \oldconcept{RandomAccessIterator} requirements and
 \pnum
 \effects
 If \tcode{InputIterator} meets the \oldconcept{RandomAccessIterator} requirements,
-returns \tcode{(last - first)}; otherwise, returns
-the number of increments needed to get from
-\tcode{first}
-to
-\tcode{last}.
+returns \tcode{(last - first)}; otherwise, increments
+\tcode{first} until \tcode{last} is reached and returns
+the number of increments.
 \end{itemdescr}
 
 \indexlibraryglobal{next}%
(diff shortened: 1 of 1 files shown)


tkoeppe:
@CaseyCarter: I don't think either version of `advance` is using similar wording, so this should make things neither more nor less consistent. The range version of advance does say "increments <foo>", though, so this seems like a reasonable way to say that an iterator really gets incremented.

jensmaurer (commented):
Looks good for this case, but the neighboring `advance` and `ranges::advance` still talks about "increments i by n" for input iterators, which should get a similar treatment; maybe just "increments i n times" except that something should separate "i" and "n".

tkoeppe:
I see -- I thought we had blanket wording for the "by n" bit, but I was thinking of [algorithms.requirements]p12. I was mainly after making sure we are saying "increment" explicitly.

I'm not sure if the "by n" part itself is controversial or in need of clarification. Thoughts?

jensmaurer:
Well, I think that some of the `ranges::advance` variants avoid the "by n" and clearly say what they mean, in terms of repeatedly incrementing the iterator. Given that `advance` is such a basic building block, I feel we should not gloss over the expressiveness differences of iterators here, and be explicit instead.

I'm ok with forking off "advance" into a separate issue.

CaseyCarter (approved):

CaseyCarter:
LWG originally chose the "increments `i` by `n`" phrasing to avoid "increments `i` `n` times." (Ditto "decrements `i` by `-n`".) I've never been enthusiastic about it, but folks felt it was "clear enough". Feel free to replace it if you have a better phrasing.

tkoeppe:
I tend to agree that "increments i by n" is clear enough; I have yet to see any confusion arising from it. Just for contrast, the linked issue shows that the lack of clarity about incrementing was indeed something that could make you wonder about what exactly is specified, but nobody has complained about "by n" yet. If anyone really wants to spend time on that, let me know, but I'd just leave it as is.

jensmaurer:
> I'd just leave it as is.

Fine with me.

jwakely (approved):

umanwizard:
Thanks for addressing this!

tkoeppe:
@umanwizard: My pleasure, thanks for reporting!

PR #4719, opened 2021-06-25
[lib] Remove space between `requires` and parameter list

JohelEGP:
Or
> [concept.booleantestable,reverse.iter.elem,common.iter.nav] Remove space between `requires` and parameter list


CaseyCarter (approved):

jensmaurer:
I think we should decide first whether we want to go there or not. The existing practice does not show an obvious preference; see #4720.

JohelEGP:
Right. I forgot about that. Although there are not many _requires-expression_s with a parameter list on the library side in total, I noticed the minority fixed on this PR.

PR #4725, opened 2021-06-29
upload the PDF for everybody to download

yecril71pl:
The PDF will be available in build job results.


tkoeppe:
Thank you, but we don't want to include PDFs in the source repository. You will notice that we attach PDFs to the labelled releases already, and PDFs appear in the public mailings as well. (We are in fact in the process of _deleting_ the existing, historic PDFs from the repository.)

yecril71pl:
> Thank you, but we don't want to include PDFs in the source repository. You will notice that we attach PDFs to the labelled releases already, and PDFs appear in the public mailings as well. (We are in fact in the process of _deleting_ the existing, historic PDFs from the repository.)

I did not add any PDF to the source repository and it will not end up there once this request is accepted.

tkoeppe:
Ah, I see. Where does the PDF go?

.github/workflows/check.yml
```diff
@@ -2,14 +2,12 @@
 
 name: check
 
-on:
-  pull_request:
-  push:
+on: [push]
```
tkoeppe, line 5: Why did this change?

.github/workflows/check.yml
```diff
@@ -2,14 +2,12 @@
...
 
 defaults:
   run:
     working-directory: source
-
+
```
tkoeppe, line 10: Unnecessary edit.

.github/workflows/check.yml
yecril71pl: I did not know how to trigger it.

.github/workflows/check.yml
```diff
@@ -25,3 +25,8 @@ jobs:
         run: make quiet
       - name: check-output
         run: ../tools/check-output.sh
+      - name: Upload output file
+        uses: actions/upload-artifact@v2
+        with:
+          name: output-log-file
+          path: source/std.pdf
```
CaseyCarter, line 32: The name "output-log-file" seems to be inconsistent with the fact that we're uploading `source/std.pdf` which is an output, but not a log. `source/std.log` is what I would consider to be the log file. Maybe it would make more sense to call this `working-draft`?

CaseyCarter:
> Ah, I see. Where does the PDF go?

It's an artifact attached to the build results, which I assume means we can easily view the result of applying a PR without needing to checkout the branch and build it locally.

jwakely:
I'm not convinced this is necessary or desirable. How long do those artefacts stay around?

Please at least squash your branch and force-push. Eight different commits for a five line change is unnecessary.

jensmaurer:
> How long do those artefacts stay around?

Quote from https://github.com/actions/upload-artifact

"Artifacts are retained for 90 days by default."

We can change that per-repo; I think 14 days should be enough.
I've just changed the repo-wide maximum to 30 days.

"Billing is based on the raw uploaded size"

It seems our plan at github has 2 GB of actions space included.

@tkoeppe, I think we should squash and merge this and try it out. If it doesn't work or costs a lot, we can remove it at any time.

tkoeppe:
Hm, the artifact is currently delivered as  zip file, e.g. see https://github.com/cplusplus/draft/runs/2962468115. Is that what you want? Is there a way to make the PDF available directly?

tkoeppe:
Hm,

> There is currently no way to download artifacts after a workflow run finishes in a format other than a zip or to download artifact contents individually.

I suppose not. OK. Let's try this out, it seems like it could be useful. Thanks!

yecril71pl:
> Hm, the artifact is currently delivered as zip file, e.g. see https://github.com/cplusplus/draft/runs/2962468115. Is that what you want? Is there a way to make the PDF available directly?

You could add another step to upload it somewhere else, of course.  I do not think it is necessary.  I just wanted to prevent a situation where everybody forks and builds just to get the PDF.

jwakely:
> I just wanted to prevent a situation where everybody forks and builds just to get the PDF.

I don't consider that a suitable use of this repo. "Everybody" should not need the latest working draft, and we do not need to provide them with it (they can already get a recent one from the monthly mailings at https://www.open-std.org/jtc1/sc22/wg21/ if they want a recent draft).

If somebody really needs the very latest draft then they should build it themselves.

yecril71pl:
> I don't consider that a suitable use of this repo. "Everybody" should not need the latest working draft, and we do not need to provide them with it (they can already get a recent one from the monthly mailings at https://www.open-std.org/jtc1/sc22/wg21/ if they want a recent draft).

If you want that to happen, you should say so in README.

jwakely:
The project's "About" already links to it, but we could make the README more explicit. 

This repo is for managing the sources, **not** publishing the drafts, and **not** for end users to easily obtain drafts.

If the purpose of this pull request is to publish the latest draft for users, then I object and want it reverted.

JohelEGP:
I've been thinking that it'd be nice to have a `diffpdf` output on PRs done by CI. This PR could be considered a step towards that.

tkoeppe:
> > Hm, the artifact is currently delivered as zip file, e.g. see https://github.com/cplusplus/draft/runs/2962468115. Is that what you want? Is there a way to make the PDF available directly?
> 
> You could add another step to upload it somewhere else, of course. I do not think it is necessary. I just wanted to prevent a situation where everybody forks and builds just to get the PDF.

Hm, all it would take here is to clone this repo. If people really fork everything at the drop of a hat, there's probably not much we can do about that.

tkoeppe:
I also changed the artifact name to "draft-snapshot", since the things on this repo aren't even working drafts. Working drafts are the PDFs that appear in the public mailing and are approved by the comittee.

jensmaurer:
It's good we make sure those PDFs are not marketed as "working drafts". Also, note that they are snapshots after applying pull requests, so become effectively stale after the pull request is merged.

It seems that offering the PDFs is a no-cost operation for everybody involved (including the editors), once we've overcome the discussion in the present pull request, so let's just do that.

(My opinion will change drastically if github actually starts charging for the PDFs in any way.)

yecril71pl:
> It's good we make sure those PDFs are not marketed as "working drafts". Also, note that they are snapshots after applying pull requests, so become effectively stale after the pull request is merged.

I do not think they should be *marketed* at all.  GitHub users will know where to find them, everybody else will have the README nicely updated by @jwakely, and that is that.

tkoeppe:
@jensmaurer: the artefacts are also available on the `main` branch actions, so there's a "post-commit version", too.

PR #4730, opened 2021-07-02
[std] Harmonize 'reference binds to an expression' phrasing

### LINKED ISSUE #4722: "binds to" versus "is bound to"
In most places in the standard, where we talk about a reference being initialized with an expression or a temporary object, we say "the reference binds to the expression" and similar things. However, we also say all of these things:

* the expression binds to the reference (in only about 3 places)
* the reference is bound to the expression
* the expression is bound to the reference

It'd be nice if we were more consistent here. I think the right viewpoint is that references bind to expressions, not the other way around, but if so, is it correct to say that expressions are bound to references or that they are bound to *by* references or something else?

jensmaurer:
Fixes #4722

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -4272,8 +4272,7 @@ the destruction of every temporary created in a default argument is
 sequenced before the construction of the next array element, if any.
 
 \pnum
-The third context is when a reference is bound to a
-temporary object.
+The third context is when a reference binds to a temporary object.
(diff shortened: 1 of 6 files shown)


zygoloid (approved):
Looks good to me, thanks!

jensmaurer:
@tkoeppe, anything else you want to do before merging?

PR #4732, opened 2021-07-03
[range.nonprop.cache] Clarify that emplace-deref does not require materialization of the result of *i before the initialization

timsong-cpp:
Based on empirical evidence, the existing wording is too subtle, so this adds a note to make the implication more obvious.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -3216,6 +3216,10 @@ template<class I>
  \mandates
 The declaration \tcode{T t(*i);} is well-formed
 for some invented variable \tcode{t}.
+\begin{note}
+When \tcode{*i} is a prvalue of type \cv{} \tcode{T},
+\tcode{is_constructible_v<T, decltype(*i)>} can be \tcode{false}.
+\end{note}
 
  \effects
 Calls \tcode{reset()}.
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -3217,8 +3217,8 @@ template<class I>
 The declaration \tcode{T t(*i);} is well-formed
 for some invented variable \tcode{t}.
 \begin{note}
-When \tcode{*i} is a prvalue of type \cv{} \tcode{T},
-\tcode{is_constructible_v<T, decltype(*i)>} can be \tcode{false}.
+If \tcode{*i} is a prvalue of type \cv{} \tcode{T},
+there is no requirement that it is movable\iref{dcl.init.general}.
 \end{note}
 
  \effects


jensmaurer (approved):

jwakely (approved):

tkoeppe:
Out of curiosity, could you give an example for which this `is_constructible` is false?

JohelEGP:
```C++
#include <new>
#include <type_traits>

struct S {
  S() = default;
  S(const S&) = delete;
  S(S&&) = delete;
};

S deref() { return S{}; }
S s{};

void emplace_deref() {
  s.~S();
  new (&s) S(deref());
}

static_assert(!std::is_constructible_v<S, S>);
```
https://godbolt.org/z/GbsxT8P34

tkoeppe:
Thanks! Should we maybe augment the note to say "... is_constructible can be false<ins>, but the above declaration of `t` can still be valid</ins>."?

jensmaurer:
Or maybe: "If `*i` is a prvalue, there is no requirement that `*i` it is copyable (dcl.init.general)."

tkoeppe:
Maybe even, "`if *i` is a prvalue, `i` can be passed to emplace_deref without a requirement that `*i` is copyable (dcl.init.general)"?

jensmaurer:
Repeating "emplace_deref doesn't seem to add anything.

jensmaurer:
Editorial meeting 2021-07-16: Go with Jens' suggestion.

@timsong-cpp , please update your pull request accordingly.

timsong-cpp:
I used "movable" instead since `emplace(*i)` would be a move rather than a copy.

jensmaurer (approved):
Thanks, looks good.

jensmaurer:
@tkoeppe , this seems ready to merge now.

timsong-cpp:
@tkoeppe The commit message on 1305740 is incorrect:

> The normatively expressed requirement in terms of an invented initialization of a variable was prone to confusion, since such a declaration would indeed require `*i` to be copyable/movable, whereas the actual use of `emplace-deref` does not
cause any intermediate materializations of prvalues.

That invented initialization does _not_ require `*i` to be copyable or movable in this case. That's exactly why it is specified this way and not with `is_constructible_v` or `constructible_from`, and the whole reason for the note's existence (and why it is a note rather than normative wording).

In the future, I would appreciate an opportunity to comment on this kind of commit messages that are going to be attributed to me _before_ they end up in main, instead of discovering them afterwards.

jensmaurer:
@tkoeppe, this looks like time for force-push of master.

tkoeppe:
@timsong-cpp Sorry about that! You're right -- could you please provide a new message (that explains the clarified subtlety), and I'll update that?

timsong-cpp:
How about:

> A new note explains that `emplace-deref` requires implementations to avoid 
> materialization of the result of `*i` before the initialization. This is implied by the 
> normatively expressed requirement in terms of an invented initialization of a 
> variable, but is easily overlooked.

tkoeppe:
Thank you, I updated that!

Issue #4662, opened 2021-06-08
Is a null pointer value of a pointer type be called a pointer?


xmh0511:
Please see this example
````cpp
 (void*)0 == (char*)0;
````
The operands of the equal operator are a null pointer value of type `void*` and a null pointer value of type `char*`, respectively as per: 
> A null pointer constant is an integer literal ([lex.icon]) with value zero or a prvalue of type std​::​nullptr_­t. A null pointer constant can be converted to a pointer type; **the result is the null pointer value of that type** ([basic.compound]) and is distinguishable from every other value of object pointer or function pointer type.   

Meanwhile,  [expr.eq#3] states that
> If at least one of the operands is **a pointer**, pointer conversions, function pointer conversions, and qualification conversions are performed on both operands to bring them to their composite pointer type. Comparing **pointers** is defined as follows: 

The two operands after conversion should both be **pointers** if one of them is original a **pointer**. However, according to the relevant rule in [basic.compound]#3 
> Except for pointers to static members, text referring to “pointers” does not apply to pointers to members. Pointers to incomplete types are allowed although there are restrictions on what can be done with them ([basic.align]). **Every value of pointer type** is one of the following: 
>> - a pointer to an object or function (the pointer is said to point to the object or function), or 
>> - a pointer past the end of an object ([expr.add]), or 
>> - the **null pointer value** for that type, or 
>> - an **invalid pointer value**.   

Except for the first two bullets, they are explicitly called **pointer**; the remaining cases are not explicitly called **pointer**. In other words, Can a null pointer value or invalid pointer value be called a pointer? It seems that there is no rule elsewhere in the standard that states any `value` of pointer type is called a pointer. If it is say, the two operands of the equal operator should be invalid in this example. 

Anyway, the issue here is that, Is an arbitrary `value of pointer type` called pointer?

jwakely:
[expr.eq] p2 says "The converted operands shall have arithmetic, enumeration, pointer, or pointer-to-member type" so it seems obvious to me that when p3 says an operand "is a pointer" it means it has pointer type.

The obvious answer to your question is yes, everything that has pointer type is a pointer. I think you are reading too much into the list in [basic.align]. I read it as "a pointer to _something_, or a pointer past the end of _something_, or ..." not as a definition of what is and isn't "a pointer".

xmh0511:
[...]

PR #4733, opened 2021-07-04
[expr.mptr.oper] Clarify pointer-to-member operators.

### LINKED ISSUE #4703: Mislead utterance of section [expr.mptr.oper]
> The binary operator .* binds its second operand, which shall be of type “pointer to member of T” to its first operand, which shall be a glvalue of class T or of a class of which T is an unambiguous and accessible base class. **The result is an object or a function of the type specified by the second operand**.

As aforementioned, the second operand has the type of “pointer to member of T”, but the rule says the result is an object or a function of the type specified by the second operand, Is it a contradiction here? 

Maybe we should say
> The result is an object or a function of the type specified by the entity to which the second operand refers. 

Can `E2` be the second operand of the class member access operator?
> Abbreviating pm-expression.*cast-expression as E1.*E2, E1 is called the object expression. If the dynamic type of E1 does not contain the member to which E2 refers, the behavior is undefined. Otherwise, the expression E1 is sequenced before the expression E2.  
>
> The restrictions on cv-qualification, and the manner in which the cv-qualifiers of the operands are combined to produce the cv-qualifiers of the result, are the same as the rules for **E1.E2** given in [expr.ref].    

Again, `E2` still has the type of “pointer to member of T” in the context of the above rule, which means `E2` is invalid to be the second operand of the expression E1.E2.

jensmaurer:
Fixes #4703

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -5883,16 +5883,15 @@ left-to-right.
 
 \pnum
 The binary operator \tcode{.*} binds its second operand, which shall be
-of type ``pointer to member of \tcode{T}'' to its first operand, which shall be
+of type ``pointer to member of class \tcode{T} of type \tcode{X}'' to its first operand, which shall be
 a glvalue
 of
 class \tcode{T} or of a class of which \tcode{T} is an unambiguous and
-accessible base class. The result is an object or a function of the type
-specified by the second operand.
+accessible base class. The result is an object or function of type \tcode{X}.
 
 \pnum
 The binary operator \tcode{->*} binds its second operand, which shall be
-of type ``pointer to member of \tcode{T}'' to its first operand, which shall be of
+of type ``pointer to member of class \tcode{T}'' to its first operand, which shall be of
 type ``pointer to \tcode{U}''
 where \tcode{U} is either \tcode{T} or
 a class of which \tcode{T}
@@ -5912,7 +5911,10 @@ Otherwise, the expression \tcode{E1} is sequenced before the expression \tcode{E
 The restrictions on cv-qualification, and the manner in which
 the cv-qualifiers of the operands are combined to produce the
 cv-qualifiers of the result, are the same as the rules for
-\tcode{E1.E2} given in~\ref{expr.ref}.
+\tcode{E1.EM} given in~\ref{expr.ref},
+where \tcode{EM} is an \grammarterm{identifier}
+naming a hypothetical non-static non-mutable member of class \tcode{T}
+declared with type \tcode{X}.
 \begin{note}
 It is not possible to use a pointer to member that refers to a
 \keyword{mutable} member to modify a const class object. For


source/expressions.tex
```diff
@@ -5795,7 +5794,10 @@
 The restrictions on cv-qualification, and the manner in which
 the cv-qualifiers of the operands are combined to produce the
 cv-qualifiers of the result, are the same as the rules for
-\tcode{E1.E2} given in~\ref{expr.ref}.
+\tcode{E1.EM} given in~\ref{expr.ref},
```
xmh0511, line 5797: Maybe, we could say that `EM` is the name of the member pointed by the value of the second operand by ignoring mutable specifier(if any).

jensmaurer: I think name lookup considerations are not really relevant here. I say "EM must name a member of T" below; plus EM has a unique name (it's hypothetical / invented), so there is no hiding in view.
All that's relevant is the type of EM and the fact it's non-static and non-mutable; everything else falls out of the expr.ref rules.

xmh0511: I would expect that the expression `E1.EM` can additionally state what the result is, after all, [expr.mptr.oper#2] merely says "The result is an object or a function of the type X", but it didn't say which the object or function is, which needs the relevant rule of `E1.EM`. Except that, the utterance for type is ok.

xmh0511: So, I think the paragraph [expr.mptr.oper#5] should be changed to that   
> The result of E1.*E2 is determined by E1.EM([expr.ref]), where EM is an identifier naming the member function pointed by the value of E2 if the value of E2 points to a member function that is virtual([class.virtual]); Otherwise, EM is a qualified-id `T::m` naming a member `m` of class T pointed by the value of E2 and the mutable specifier in the declaration of the member is ignored(if any).

Now, the [expr.mptr.oper#5] specifies all the respects regard the result of expression `E1.*E2`(e.g, which object or function does the result denote).

jwakely: Why can't we just say "E2 points to" instead of "the value of E2 points to"? It's a pointer, it points to things.

jensmaurer: E2 is an expression, so it presumably it has a value.

That said, I note that @xmh0511 also laments the absence of a specification what the result value actually is, so some massaging to the original wording is probably necessary. I would like to avoid talking in "identifiers" for anything but finding out the cv-qualification of the result type, if we can help it. There's too much name lookup / virtual function etc baggage around "identifier".

xmh0511: I want to use "the value of E2" to avoid a similar issue mentioned in https://github.com/cplusplus/draft/issues/4662#issuecomment-856884069. But, it seems you're right that the meaning is unambiguous if using "E2 points to". However, the benefit of using "the value of ..." is that it implies the `lvalue-to-rvalue` conversion shall be applied to the operand(if necessary), which is also under-specified in many sections in [expr].

xmh0511: @jensmaurer  Not strictly, a pointer to member records the information of the entity and its class, hence I think that we use `T::m` is equivalent to use the entity pointed to by the pointer to member in most respects. The exception is the virtual function, which has been listed in the proposed wording. IMHO, the wording could cover the other cases.

xmh0511: Consider the second operand may result from Pointer-to-member conversions, I think we should separate [expr.mptr.oper#2] into two parts; One part phrases type of the result, the other phrases the entity of the result. Since the wording about the type of the result in the current changed file is ok, we should pay attention to the entity designated by the result.   
> The entity designated by E1.*E2 is determined by E1.EM([expr.ref]), where EM is a qualified-id `T::m` naming the member to which the value of E2 refers, except that If the value of E2 refers to a virtual function, the function designated by E1.*E2 is the final overrider in the dynamic type of the object expression. 

Since the relevant rule defined in [conv.mem]
> The result of the conversion refers to the same member as the pointer to member before the conversion took place, but it refers to the base class member as if it were a member of the derived class. The result refers to the member in D's instance of B. 

So, even if the operand E2 is taken from the result of Pointer-to-member conversions, its value still refers to the original member, which is the entity that E1.*E2 designates.

jensmaurer:
@tkoeppe, I think I want CWG review for this eventually.

tkoeppe:
Could you please rebase?

PR #4736, opened 2021-07-05
[unique.ptr.single] Rephrase destruction

jensmaurer:
Supersedes #2962


jensmaurer:
@CaseyCarter , @jwakely, what do you think?

PR #4737, opened 2021-07-05
[stmt.return,class.{ctor,dtor}] Clarify no return operand

jensmaurer:
Highlight that constructors and destructors do not have a
return type and thus a return statement within a constructor
or destructor cannot have an operand.

Supersedes #3647

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -1208,12 +1208,13 @@ They come into effect when the constructor for the
 most derived object\iref{intro.object} ends.
 
 \pnum
-\indextext{restriction!constructor}%
-A
-\tcode{return}
-statement in the body of a constructor shall not specify a return value.
 \indextext{constructor!address of}%
 The address of a constructor shall not be taken.
+\indextext{restriction!constructor}%
+\begin{note}
+A \tcode{return} statement in the body of a constructor
+cannot specify a return value\iref{stmt.return}.
+\end{note}
 
 \pnum
 A constructor shall not be a coroutine.
@@ -2071,8 +2072,13 @@ and in particular,
 the selected destructor may be deleted\iref{dcl.fct.def.delete}.
 
 \pnum
-\indextext{restriction!destructor}%
+\indextext{destructor!address of}%
 The address of a destructor shall not be taken.
+\indextext{restriction!destructor}%
+\begin{note}
+A \tcode{return} statement in the body of a destructor
+cannot specify a return value\iref{stmt.return}.
+\end{note}
 \indextext{\idxcode{const}!destructor and}%
 \indextext{\idxcode{volatile}!destructor and}%
 A destructor can be invoked for a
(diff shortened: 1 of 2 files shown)


source/statements.tex
```diff
@@ -810,14 +810,17 @@
...
+in a function that has a \cv{}~\keyword{void} return type.
 A \tcode{return} statement with any other operand shall be used only
-in a function whose return type is not \cv{}~\keyword{void};
+in a function that has a return type other than \cv{}~\keyword{void};
 \indextext{conversion!return type}%
 the \tcode{return} statement initializes the
 glvalue result or prvalue result object of the (explicit or implicit) function call
 by copy-initialization\iref{dcl.init} from the operand.
 \begin{note}
+A constructor or destructor does not have a return type.
```
xmh0511, line 821: If we claim that the constructor of a class does not have a return type, consider [dcl.init.general] p15
>  if the function is a constructor, the call is a prvalue of the cv-unqualified version of the destination type whose result object is initialized by the constructor.  

> A function call is an lvalue if the result type is an lvalue reference type or an rvalue reference to function type, an xvalue if the result type is an rvalue reference to object type, and a prvalue otherwise.  

These rules will be conflict. How could the call of a constructor be a prvalue of the **cv-unqualified version of the destination type**?

tkoeppe: @jensmaurer Final thoughts?

jensmaurer: Constructors / destructors don't have a return type in the "function declaration" sense, and that absence makes "return something" syntactically ill-formed.  I think the quoted stuff doesn't change that; it just means that a constructor call (in the context of copy-initialization, not in general) has (is considered to have) a certain type and value category. That has no bearing on the syntactic constraints on constructor declarations.

I thinks this should go in.

source/classes.tex
```diff
@@ -2075,6 +2076,10 @@
 \pnum
 \indextext{restriction!destructor}%
```
tkoeppe, line 2077: @jensmaurer Can we make the indexing similar for constructor and destructor, i.e. have "restriction" and "address of" entries for both?

jensmaurer: Sure. Done.

source/classes.tex
```diff
@@ -2078,8 +2079,13 @@
 the selected destructor may be deleted\iref{dcl.fct.def.delete}.
 
 \pnum
-\indextext{restriction!destructor}%
+\indextext{constructor!address of}%
```
tkoeppe, line 2082: destructor

jensmaurer: Fixed

tkoeppe: Thanks -- could you rebase please? Something weird seems to have happened to this branch.

jensmaurer:
Rebased and force-pushed.

PR #4740, opened 2021-07-08
[expr.delete] "result" usage fix

languagelawyer:


jensmaurer (commented):
This feels like an improvement to me.

PR #4744, opened 2021-07-11
[array.overview] Update exceptions to account for changes in #4460

JohelEGP:


JohelEGP:
This may interfere with #4460, specifically at https://github.com/cplusplus/draft/pull/4460/files#diff-3b5851f7056b7c68cb3ba2ab51cf0b75a780c6a11e5e181770067700454ace7bL3417.

source/containers.tex
```diff
@@ -3414,7 +3414,8 @@
 \indextext{requirements!container}%
 An \tcode{array} meets all of the requirements of a container and
 of a reversible container\iref{container.requirements}, except that a default
-constructed \tcode{array} object is not empty and that \tcode{swap} does not have constant
+constructed \tcode{array} object is not empty, and
+that move construction and \tcode{swap} do not have constant
```
jensmaurer, line 3418: The requirements table for general container requirements already says (in Note A and Note B) that the complexity in the array case is linear, not constant. So, it feels we should make the "swap" exception as well as the move construction exception a note here.

jensmaurer (changes requested):

source/containers.tex
JohelEGP: Maybe after #4460, which doesn't use notes anymore. From reading https://www.iso.org/sites/directives/current/part2/index.xhtml#_idTextAnchor333, it seems that these meant to not be normative. But they're referred to by normative text right after the table. Maybe we need a better term to not reuse note.

In any case, it doesn't feel right to make the exception in the requirements themselves rather than the specification of the container. To say that `array` _does_ meet the requirements, make no exception after saying so (or sneaking that in a note), to then make exceptions in the requirements feels the wrong way to go about it.

If we were to go with your suggestion here, how about including the whole sentence to also include the default constructor, which doesn't have a note in the requirements table (in the WD nor in #4460)? I suppose that then we'd need a Note C for the default constructor in the requirements table similar to the Note A and B.

jensmaurer: Those "Note A/B" things are certainly normative, despite the spelling of "note".

JohelEGP: Done. I suppose the next step would be to move the remaining exception to the referred requirements for consistency.

jensmaurer (approved):

source/containers.tex
jensmaurer: Yes, except that's unfriendly to do while #4460 is in flight.

jensmaurer:
#4460 has landed.

JohelEGP:
There's 2 things to do:
- Mention the default constructor for `array` exception in the container requirements.
- Mention the move construction exception in `array`'s introduction.

I don't see a need to use notes anymore.

JohelEGP:
Let me amend that.

---

> * Mention the move construction exception in `array`'s introduction.

This is already present:
> [16](https://eel.is/c++draft/container.reqmts#16)
[#](http://github.com/Eelis/draft/tree/dd187fdd1eb0bd31df208223997ff766dff5fee4/source/containers.tex#L250)
Complexity: Linear for array and constant for all other standard containers[.](https://eel.is/c++draft/container.reqmts#16.sentence-1)

---

`array`'s intro says:
> [3](https://eel.is/c++draft/array#overview-3)
[#](http://github.com/Eelis/draft/tree/dd187fdd1eb0bd31df208223997ff766dff5fee4/source/containers.tex#L5724)
An array meets all of the requirements of a container ([[container.reqmts]](https://eel.is/c++draft/container.reqmts)) and of a reversible container ([[container.rev.reqmts]](https://eel.is/c++draft/container.rev.reqmts)), except that a default constructed array object is not empty and that swap does not have constant complexity[.](https://eel.is/c++draft/array#overview-3.sentence-1)

`swap`, which is already specified as
> [47](https://eel.is/c++draft/container.reqmts#47)
[#](http://github.com/Eelis/draft/tree/dd187fdd1eb0bd31df208223997ff766dff5fee4/source/containers.tex#L448)
Complexity: Linear for array and constant for all other standard containers[.](https://eel.is/c++draft/container.reqmts#47.sentence-1)

shouldn't be part of this "except that", since its rightfully part of the container requirements.

A default constructed `array` object is not always empty. That depends on `N`. This should be fixed. For simplicity, I plan to keep the "except that" and leave the default constructor container requirements untouched. The reason for this is that it's more incremental and easier to get accepted, as well that all other `array` exceptions lie in _Complexity_ elements, as well that we'd have to talk about `N`, which we'd need to introduce, and its specification as a container requirement would be much longer than the existing exceptions for `array`'s complexities.

jensmaurer:
@jwakely, this boils down to removing redundant specification (swap complexity) and fixing bad specification (zero-sized arrays are, in fact, empty).

source/containers.tex
```diff
@@ -6123,8 +6123,8 @@
 of a container\iref{container.reqmts} and
 of a reversible container\iref{container.rev.reqmts},
 except that a default
-constructed \tcode{array} object is not empty and that \tcode{swap} does not have constant
-complexity. An \tcode{array} meets some of the requirements of a sequence
+constructed \tcode{array} object is not empty if $\tcode{N} != 0$.
```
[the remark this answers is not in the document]
JohelEGP: Done.

jwakely (approved):

Issue #4747, opened 2021-07-13
[bit.count] Turn introduction of `N` into a note


JohelEGP:
`N` is only used in notes. Maybe (also) say "in the following notes" instead. This sentence is also missing a paragraph number.

jensmaurer:
I agree with the missing paragraph number.
It's also correct that "N" is only used in notes, but I'm not sure any of the proposed cures is better than the disease. In short, I think introducing N normatively and then using it only inside notes is not a problem that needs a fix.

PR #4759, opened 2021-07-18
[bit.count] Add missing paragraph number

jensmaurer:
Partially addresses #4747


tkoeppe:
(No need to make pull requests for things like this, just feel free to push the change directly.)

PR #4772, opened 2021-07-25
[std] Use \Fundescx to avoid awkward mid-sentence colons.

Eelis:
image

There is precedent here, for example:

https://github.com/cplusplus/draft/blob/b4e36f7c5b7498a01b1d068afed0b90b0d46ab32/source/lib-intro.tex#L374-L378

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -115,7 +115,7 @@ the names of template parameters are used to express type requirements.
 \end{itemize}
 
 \pnum
-If an algorithm's \effects element specifies
+If an algorithm's \effectsx{} element specifies
 that a value pointed to by any iterator passed as an argument is modified,
 then that algorithm has an additional type requirement:
 The type of that argument shall meet
--- a/source/future.tex
+++ b/source/future.tex
@@ -1,7 +1,8 @@
 %!TEX root = std.tex
 \normannex{depr}{Compatibility features}
 
-\newcommand{\requires}{\Fundesc{Requires}}
+\newcommand{\requiresx}{\Fundescx{Requires}}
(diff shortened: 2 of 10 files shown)


jensmaurer:
If we do this, I'd prefer new LaTeX macros like `\requiresx` or so, instead of introducing the chance of misspelling those items. Oh, and the definition of `\requires` should then use `\requiresx`, of course.

Eelis:
Amended as suggested.

tkoeppe:
I'm not entirely convinced that a change is needed here at all (the status quo has a certain appeal even), but if you do want to pursue this, two questions:

1. Can we have starred versions, like `\effects*`, to not have the colon?
2. Can we generate all the macros more compactly, say by defining a "meta macro" in TeX that emits multiple macros? The goal would be to only have to say each base name once, and have all the derived macros generated automatically.

tkoeppe:
I'm tempted to just live with the status quo. Please reopen if you feel strongly. At the very least we would want a system that does not require us to state every element twice in `macros.tex`.

Also, @jwakely, do you have any opinions? We'd like LWG approval for this change, too.

jwakely:
I'm happy with the status quo.

Eelis:
> the status quo has a certain appeal even

I suspect an editor of any professional publication would be fired on the spot if they let through an error as egregious as having colons in sentences without those colons actually serving as colons but instead serving as some kind of implicitly *quoted* colons because they appear unquoted elsewhere. But that's just me and I don't feel strongly about having the document look like amateur hour, so it's fine with me. :)

jensmaurer:
@tkoeppe, I guess that counts as a strong opinion.

PR #4774, opened 2021-07-25
[container.adaptors.general] Fix ref

JohelEGP:


jensmaurer:
This looks like an incorrect change. [sequences.general] p2 does define iter-value-type, and container adaptors are based on sequence containers.

JohelEGP:
You are right. I was looking for uses of "exposition" in the passing, and noticed that two places said that they used the _`iter-value-type`_ defined in 22.4.1 and 22.3.1. Turns out that the exposition-only name is equivalently defined in both places.

PR #4777, opened 2021-07-26
[defns.access] [basic.lval] Clarify what and how can be accessed

languagelawyer:
```c++
struct S { int i; } s;
auto u = (unsigned&)s; // accesses s, which is an object of class type
```
So the Notes are not accurate enough.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -319,7 +319,9 @@ for a union of type \tcode{U} with a glvalue argument
 that does not denote an object of type \cv{}~\keyword{U} within its lifetime,
 the behavior is undefined.
 \begin{note}
-Unlike in C, \Cpp{} has no accesses of class type.
+In C, an entire object of structure type can be accessed, e.g. using assignment.
+By contrast, \Cpp{} has no notion of accessing an object of class type
+through an lvalue of class type.
 \end{note}
 
 \rSec2[expr.type]{Type}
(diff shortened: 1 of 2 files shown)


jensmaurer:
[basic.lval] p11 formulates the aliasing constraints, and it uses "access". So, it would be good if "access" also applied to reads/writes made invalid by the aliasing rules.

jensmaurer (approved):

source/expressions.tex
```diff
@@ -319,7 +319,7 @@
 that does not denote an object of type \cv{}~\keyword{U} within its lifetime,
 the behavior is undefined.
 \begin{note}
-Unlike in C, \Cpp{} has no accesses of class type.
+Unlike in C, \Cpp{} has no accesses through a glvalue of class type.
```
tkoeppe, line 322: Hm, but C also has no notion of "glvalue", so even though this change has made the language more precise on one way, it's also intruduced a kind of non-sequitur comparison. Maybe we need to say this a bit more verbosely?

languagelawyer: Neither C has notion of class type. How verbose you want this to become?

tkoeppe: Well, if we're already interested in clarifying this, we should try to do the best possible job. Can you think of anything that would help?

jensmaurer: Maybe: "In C, entire structs can be accessed, e.g. using assignment. In contrast, \Cpp{} has no notion of accessing an entire struct through an lvalue of class type."

tkoeppe: That's the spirit. Thanks, Jens!

Issue #3614, opened 2020-01-08
[expr.add]/6 unfortunate example in Note


languagelawyer:
[[expr.add]/6 says](http://eel.is/c++draft/expr.add#6):
> For addition or subtraction, if the expressions `P` or `Q` have type “pointer to _cv_ `T`”, where `T` and the array element type are not similar, the behavior is undefined. [Note: In particular, a pointer to a base class cannot be used for pointer arithmetic when the array contains objects of a derived class type. — end note]

The paragraph was added in the resolution of [CWG1504](http://wg21.link/cwg1504) (the normative text has been a lil bit changed by [CWG1865](http://wg21.link/cwg1865)). Description of the issue says:
> The current wording is not sufficiently clear that a pointer to a base class subobject of an array element cannot be used in pointer arithmetic.

I suspect this meant something like
```c++
struct B {};
struct D : B {};
D d[3];
B* bp = &d[0];
bp + 0; // This should have become UB with the resolution of the issue
```
At C++11 times (when the issue was reported), [[conv.ptr]/3](https://timsong-cpp.github.io/cppwp/n3337/conv.ptr#3) was saying that `bp` would be «pointer to the base class subobject of the derived class object». In C++17, [[conv.ptr]/3](https://timsong-cpp.github.io/cppwp/n4659/conv.ptr#3) says the same thing.  
[C++11 [expr.add]/4](https://timsong-cpp.github.io/cppwp/n3337/expr.add#4) was saying that «…a pointer to a nonarray object behaves the same as a pointer to the first element of an array of length one…». We may claim that currently the Standard says mostly the same thing.  
This means that expressions `bp + 0`, `bp + 1` were/are fine (before and after the resolution of the issue). `bp + 2` was/is UB, but **not** because of the paragraph added by the resolution of the issue.  
So, the resolution of the issue changed nothing for the code above (if something like the code above was meant by the issue description). That's why I call the example in the Note unfortunate.

It **is** possible to construct an example with base and derived classes which would match [expr.add]/6 Note, however, this would be quite non-trivial, involving multiple or virtual inheritance, to make it possible to obtain a pointer of type "pointer to base" pointing to the derived class object, not the base class subobject.

I would suggest to change the Note to say, for example:
> In particular, a pointer to `int` cannot be used for pointer arithmetic when the array contains objects of `unsigned int` type.

jensmaurer:
[...]

PR #4778, opened 2021-07-26
[expr.add] Use better example

languagelawyer:
Related to #3614

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -6234,6 +6234,11 @@ points to the (possibly-hypothetical) array element
 $i - j$ of \tcode{x} if $0 \le i - j \le n$.
 \item Otherwise, the behavior is undefined.
 \end{itemize}
+\begin{note}
+Adding a value other than $0$ or $1$
+to a pointer to a base class subobject, a member subobject,
+or a complete object results in undefined behavior.
+\end{note}
 
 \pnum
 \indextext{\idxcode{ptrdiff_t}!implementation-defined type of}%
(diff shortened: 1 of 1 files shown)


tkoeppe:
Has the discussion in the linked issue been resolved?

@jensmaurer: any thoughts on this?

jensmaurer:
I think the main gist here is that using a pointer-to-base when actually pointing to a derived class is fine in most cases, but not in the array indexing case. Changing the example as suggested is losing that point.

languagelawyer:
> I think the main gist here is that using a pointer-to-base when actually pointing to a derived class is fine in most cases, but not in the array indexing case.

It is UB in most (if not all) cases (member access, member function call etc.).

Using pointer-to-base which points to base subobject is fine in all cases (even in the array indexing case). What you can violate with such a pointer, is [expr.add]/4, not [expr.add]/6.

tkoeppe:
@jensmaurer I'm not sure what to do here. Should we move this example to p4 and edit the example in p6 like proposed? Or just keep the status quo?

jensmaurer:
I like the int / unsigned int example in p6, because it is mildly surprising.

I also like to retain the base / derived situation, but maybe it's worthwhile to formulate that in terms of "base class subobject" in p4, as a special case for when the p4.2 precondition isn't satisfied.

tkoeppe:
@languagelawyer Could you take another stab at this then?

languagelawyer:
@jensmaurer not sure what you meant, but I did this. Note, the Note is outside the `itemize` environment, unlike in the next paragraph.

tkoeppe:
@jensmaurer Could you please take another look here?

jensmaurer (commented):
That looks reasonable to me.

If you can get @zygoloid  to okay these changes, we can apply them right away, otherwise I'd like to have CWG assent.

tkoeppe:
@languagelawyer could you please have a look?

PR #4783, opened 2021-07-27
[basic.types.general] Fix comment in example

### LINKED ISSUE #4779: [basic.types.general] p6 Incorrect comment in example
[basic.types#general-example-3]
````cpp
extern int arr[];               // the type of arr is incomplete
typedef int UNKA[];             // UNKA is an incomplete type
UNKA* arrp;                     // arrp is a pointer to an incomplete type
int  arr[10];                   // now the type of arr is complete
void bar() {
  arrp = &arr;                  // error: different types
}
````
The above example is a simplified example of [basic.types#general-example-3], the comment says it's an error. However, it's valid even though their types are different. A **qualification conversion** can be used to convert the type of `&arr` to the cv-unqualified type of `arrp`, which obeys [expr.ass#3]
> If the right operand is an expression, it is implicitly converted to the cv-unqualified type of the left operand.  

The type of `&arr` is `pointer to array of 10 int` whilst the type of `arrp` is `pointer to array of unknown bound of int`, as per [conv.qual#3], the former type can be converted to the later type by a qualification conversion. Hence, the comment should be "ok". [GCC](https://godbolt.org/z/7Eoj99dKz) is in compliance with these rules.

jensmaurer:
Fixes #4779

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -4628,7 +4628,7 @@ int  arr[10];                   // now the type of \tcode{arr} is complete
 X x;
 void bar() {
   xp = &x;                      // OK; type is ``pointer to \tcode{X}''
-  arrp = &arr;                  // error: different types
+  arrp = &arr;                  // OK; qualification conversion\iref{conv.qual}
   xp++;                         // OK:  \tcode{X} is complete
   arrp++;                       // error: \tcode{UNKA} can't be completed
 }


tkoeppe:
@jensmaurer: could you kindly rebase this after my force-push? (And any other branches you may have?)

jensmaurer:
Done.

"And any other branches you may have?"

That's a joke, right? Feel free to tag with "needs-rebase" any that you encounter.

tkoeppe:
Only those PRs that are based on a later commit than where I force-pushed should be affected, so I was hoping there wouldn't be too many!

jensmaurer:
Ok, let's pretend those have been handled.

PR #4789, opened 2021-08-03
[variant.variant],[variant.ctor] - clarifying the meaning of Ti

NinaRanns:
We preface some of the sections with "In the descriptions that follow, let i be in the range [0, sizeof...(Types)), and Ti be the ith type in Types.", but we do not do this for all the sections that use Ti. Similarly, there is another place where we use TI, but do not elaborate on what it means.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -4110,7 +4110,7 @@ A program that instantiates the definition of \tcode{variant} with
 no template arguments is ill-formed.
 
 \pnum
-In the descriptions that follow, let $i$ be in the range \range{0}{sizeof...(Types)},
+In the descriptions of \ref{variant}, let $i$ be in the range \range{0}{sizeof...(Types)},
 and $\tcode{T}_i$ be the $i^\text{th}$ type in \tcode{Types}.
 
 \rSec3[variant.ctor]{Constructors}
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -4880,7 +4880,7 @@ $\tcode{I} < \tcode{sizeof...(Types)}$.
 
 \pnum
 \ctype
-The type in \tcode{Types} with index \tcode{I}.
+The $\tcode{I}^\text{th}$ element of \tcode{Types}, where indexing is zero-based.
 \end{itemdescr}
 
 \rSec2[variant.get]{Value access}
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -4109,11 +4109,11 @@ the \oldconcept{Destructible} requirements (\tref{cpp17.destructible}).
 A program that instantiates the definition of \tcode{variant} with
 no template arguments is ill-formed.
 
-\rSec3[variant.ctor]{Constructors}
-
 \pnum
 In the descriptions that follow, let $i$ be in the range \range{0}{sizeof...(Types)},
 and $\tcode{T}_i$ be the $i^\text{th}$ type in \tcode{Types}.
+
+\rSec3[variant.ctor]{Constructors}
 
 \indexlibraryctor{variant}%
 \begin{itemdecl}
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -4109,12 +4109,12 @@
 A program that instantiates the definition of \tcode{variant} with
 no template arguments is ill-formed.
 
-\rSec3[variant.ctor]{Constructors}
-
 \pnum
 In the descriptions that follow, let $i$ be in the range \range{0}{sizeof...(Types)},
```
JohelEGP, line 4113:
```suggestion
In the descriptions that of \ref{variant}, let $i$ be in the range \range{0}{sizeof...(Types)},
```
[variant.variant] is not enough. [variant.relops] and [variant.specalg], which come after it, says "for all _i_" without defining _i_. This suggestion would have it defined.

JohelEGP (changes requested):
The commit needs to be prefixed with [variant.variant], according to https://github.com/cplusplus/draft/wiki/Commit-message-format#editorial-commits.

The `variant(in_place_index)` overload also uses `T`<sub>_`I`_</sub> without defining it.

source/utilities.tex
NinaRanns: perhaps
"In the descriptions <del>that</del> of \ref{variant}, let $i$ be in the range \range{0}{sizeof...(Types)}," 
?
Also, if we make this change, do we need the other suggested change ? This should cover that section too.

JohelEGP: > perhaps
> "In the descriptions ~that~ of \ref{variant}, let $i$ be in the range \range{0}{sizeof...(Types)},"
> ?

Definitely!

> Also, if we make this change, do we need the other suggested change ? This should cover that section too.

Which change are you talking about? I mentioned everything else because I don't think this would cover it.

NinaRanns: The other change in this pull request in [variant.helper] that elaborates on TI in variant_alternative. I believe your suggestion clarifies what TI means there too. On the other hand, "The Ith element of Types, where indexing is zero-based." reads much better than "The type TI". What do you think ?

JohelEGP: I think this is defining _i_ and `T`<sub>_i_</sub> specifically, and thus don't cover what `I` and `T`<sub>_`I`_</sub> might mean.

NinaRanns: review comments applied

source/utilities.tex
```diff
@@ -4880,7 +4880,7 @@
 
 \pnum
 \ctype
-The type $\tcode{T}_I$.
+The $\tcode{I}^\text{th}$ element of \tcode{Types}, where indexing is zero-based.
```
tkoeppe, line 4883: Why is this better than the original, given that you now have a widely usable definition of  T_i available? Maybe the original should also use code font for `I`.

JohelEGP: Probably due to what I said in https://github.com/cplusplus/draft/pull/4789#discussion_r681838323, but I was wrong there. [variant] also uses T<sub>_j_</sub> and T<sub>_I_</sub> (without code font `I`), so I suggest this:
```suggestion
$\tcode{T}_I$.
```

tkoeppe: But that's basically the status quo, isn't it?

JohelEGP: You're right.

tkoeppe:
I'm a bit lost what's being improved by this change. Please feel free to reopen with new information (and see the discussion above).

JohelEGP:
> We preface some of the sections with "In the descriptions that follow, let i be in the range [0, sizeof...(Types)), and Ti be the ith type in Types.", but we do not do this for all the sections that use Ti.

So the first paragraph of [variant.ctor] is moved to the last paragraph of [variant.variant.general], and is modified to apply to [variant].

PR #4793, opened 2021-08-04
dependent names

### LINKED ISSUE #4791: The qualified-id has exceeded its designated extent in the intent of certain rules
In the current draft, [basic.lookup.qual#general-2] specifies the qualified name, where it seems that it intends to use `qualified-id` to cover such as a simple-type-specifier that has the form *nested-name-specifier<sub>opt</sub>type-name*.   In other words, they all can form **nested-name-specifier identifier** on syntax.  However, they are not the same thing on grammar.   

Furthermore, [temp.res#general-5] also intends to use qualified-id to cover a simple-type-specifier that may have the aforementioned form. The meaning of the qualified-id is vague in such use, consider a using-declarator, which has the form *nested-name-specifier unqualified-id*
````cpp
using T::identifier;
```` 
Does that mean any rule that uses the qualified-id to cover the form that is not a qualified-id on grammar can also impose on the using-declarator that has the same form? It seems a defect in the current standard.  This issue is also mentioned on [SO](https://stackoverflow.com/questions/66407140/is-it-the-intent-of-the-standard-that-any-form-nested-type-specifier-followed-b)

### PR DESCRIPTION
**Pull request by jensmaurer**
Fixes #4791

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -4385,13 +4385,20 @@ template<typename T> void f() {
 \end{example}
 
 \pnum
-A \grammarterm{qualified-id}
-whose terminal name is dependent and that is in a type-only context
-is considered to denote a type.
+Each of the forms
+\begin{ncsimplebnf}
+nested-name-specifier identifier\br
+nested-name-specifier \opt{\keyword{template}} simple-template-id
+\end{ncsimplebnf}
+where the terminal name $N$ is dependent\iref{temp.dep.type}
+is interpreted as a \grammarterm{simple-type-specifier}\iref{dcl.type.simple}
+if $N$ is in a type-only context and
+as a \grammarterm{qualified-id}\iref{expr.prim.id.qual} otherwise.
 A name
 that refers to a \grammarterm{using-declarator} whose terminal name is dependent
-is interpreted as a \grammarterm{typedef-name}
-if the \grammarterm{using-declarator} uses the keyword \keyword{typename}.
+is interpreted as a \grammarterm{typedef-name}\iref{dcl.typedef}
+if the \grammarterm{using-declarator} uses the keyword \keyword{typename}
+and as an \grammarterm{unqualified-id}\iref{expr.prim.id.unqual} otherwise.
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by jensmaurer** [comment]
http://lists.isocpp.org/core/2021/08/11297.php

File: source/templates.tex [hunk 1]
```diff
@@ -4386,8 +4386,8 @@
 
 \pnum
 A \grammarterm{qualified-id}
-whose terminal name is dependent and that is in a type-only context
-is considered to denote a type.
+whose terminal name is dependent and in a type-only context
+is interpreted as a \grammarterm{simple-type-specifier}.
```
xmh0511, line 4390: If it is not in a type-only context, how to interpret it? Is it just an expression or Is it determined by the result of name lookup in the instantiation context?

source/templates.tex
```diff
@@ -4386,8 +4386,8 @@
 
 \pnum
 A \grammarterm{qualified-id}
```
xmh0511, line 4388: *qualified-id* is just an expression defined in [expr.prim.id.qual], how would it be anything others other than an expression? As we discussed in https://github.com/cplusplus/draft/issues/4791#issuecomment-892908042.

jensmaurer: A qualified-id is also used in declarative contexts, e.g.

`void N::C::f() { ..}`

`N::C::f` is a _qualified-id_.

source/templates.tex
jensmaurer: Fixed

xmh0511:
@jensmaurer  For the second form, it may be *nested-name-specifier template<sub>opt</sub> simple-template-id*. The first form may be *nested-name-specifier identifier*. Except that, I think it improves [temp.res.general-5].

jensmaurer:
Right, fixed.

jensmaurer:
Discussed by CWG https://wiki.edg.com/bin/view/Wg21telecons2021/IssuesProcessingTeleconference2021-08-24.
Needs an update using a new grammar term.

Issue #4582, opened 2021-04-09
Clarify direct-initialize and copy-initialize


xmh0511:
Presumably, "direct-initialized from E" and "copy-initialized from E"  should be distinguishable with direct-initialization and copy-initialization. The initialization might be considered as an expression, as per  
> If a language construct is defined to produce an implicit call of a function, a use of the language construct is considered to be an expression for the purposes of this definition. Conversions applied to the result of an expression in order to satisfy the requirements of the language construct in which the expression appears are also considered to be part of the full-expression. For an initializer, performing the initialization of the entity (including evaluating default member initializers of an aggregate) is also considered part of the full-expression.    

I mean that an `initialization` has been a recipe(comprises the function will be called, if any) whose evaluation will initialize the destination object.  However, `direct(copy)-initialized from E` might be a bit different. 

There are many such a usage in the standard, such as 
[expr.static.cast#4](https://eel.is/c++draft/expr.static.cast#4)  
> Otherwise, the result object is direct-initialized from E  

[dcl.dcl#dcl.init.aggr-5](https://eel.is/c++draft/dcl.dcl#dcl.init.aggr-5)
>Otherwise, if the element is not a reference, the element is copy-initialized from an empty initializer list  

So, what's the appearance of form here for copy-initialize or direct-initialize respectively? In other words, if we say that an entity `t`  is direct-initialized from an expression `E`, what's the form? 
````cpp
T t(E); // #1
T t{E}; //#2
`````
Does it refer to `#1` or `#2`? Similarly, when we say that an entity `t` is copy-initialized from an expression `E`, what's the form? 
````cpp
T t = E; //#3
T t = {E};  //#4
````
It does not specify in the standard.

jwakely:
> Presumably, "direct-initialized from E" and "copy-initialized from E" should be distinguishable with direct-initialization and copy-initialization.

Why?

> So, what's the appearance of form here for copy-initialize or direct-initialize respectively? In other words, if we say that an entity t is direct-initialized from an expression E, what's the form?

If we say it's direct-initialized, we mean it's initialized using direct-initialization. Similarly for copy-initialized. I don't understand why there would be any doubt about that.

Direct-initialized means #1, because #2 is direct-list-initialization. Copy-initialized means #3 because #4 is copy-list-initialization.

xmh0511:
> > Presumably, "direct-initialized from E" and "copy-initialized from E" should be distinguishable with direct-initialization and copy-initialization.
> 
[...]

PR #4803, opened 2021-08-12
[stmt.return,dcl.fct.def.coroutine] Avoid use of 'glvalue result'

### LINKED ISSUE #4724: Reword the wording for glvalue result of a function call CWG2495
As pointed by #4107, #3124, and #3587, the glvalue result is being unclear in the current draft. The glvalue result sometimes has the meaning(namely, anonymous reference) as mentioned in this comment <https://github.com/cplusplus/draft/pull/4107#issuecomment-667602092>, but it's not normative wording. I think we should reword the wording for "glvalue result" of a function call to make it more clear. My suggestion is
> if the function call is a prvalue, the return statement initializes the prvalue result object of the (explicit or implicit) function call by copy-initialization from the operand.
>
> Otherwise, if the function call is a glvalue, the return statement is equivalent to the following hypothetical declaration 
>>    `T t = e;`  if the operand of the return statement is a comma expression X without parenthesis, e is `(X)`, otherwise e is `X`. T is the return type of the function call;   
>
> The invented variable `t` is the result of the function call.   

Here, the invented variable `t` is the so-called anonymous reference for purpose.

jensmaurer:
Also fixes CWG2495.

Fixes #4724

As discussed in the CWG teleconference 2021-08-09.

Needs CWG review.

### DIFF
--- a/source/statements.tex
+++ b/source/statements.tex
@@ -815,7 +815,8 @@ A \tcode{return} statement with any other operand shall be used only
 in a function whose return type is not \cv{}~\keyword{void};
 \indextext{conversion!return type}%
 the \tcode{return} statement initializes the
-glvalue result or prvalue result object of the (explicit or implicit) function call
+returned reference or prvalue result object
+of the (explicit or implicit) function call
 by copy-initialization\iref{dcl.init} from the operand.
 \begin{note}
 A \tcode{return} statement can involve
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -6466,7 +6466,7 @@ results in undefined behavior\iref{stmt.return.coroutine}.
 \pnum
 The expression \tcode{\exposid{promise}.get_return_object()} is used
 to initialize
-the glvalue result or prvalue result object of a call to a coroutine.
+the returned reference or prvalue result object of a call to a coroutine.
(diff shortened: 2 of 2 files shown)


source/statements.tex
```diff
@@ -815,7 +815,8 @@
 in a function whose return type is not \cv{}~\keyword{void};
 \indextext{conversion!return type}%
 the \tcode{return} statement initializes the
-glvalue result or prvalue result object of the (explicit or implicit) function call
+returned reference or prvalue result object
```
xmh0511, line 818: If we prepare to consistently use "Y is X-initialization from Z" to specify which initialization rule will apply to that construct, in the standard. As indicated in the #4582 issue. For this purpose, I think it would be adequately clear by using this wording without using an invented declaration.

jensmaurer: The issue was that initialization of a "glvalue result" is nowhere defined, but initialization of a reference is. We also say it's copy-initialization, so I think we're good, at least for this particular situation.

jensmaurer:
Approved by CWG.

jensmaurer:

opensdh:
CWG2495 was never actually closed.

PR #4804, opened 2021-08-12
[dcl.init.ref] Avoid 'value of the expression'

### LINKED ISSUE #4753: A lvalue-to-rvalue conversion is not necessary to be required in [dcl.init.ref#5.3]
> Otherwise, if the initializer expression
>> - is an rvalue (but not a bit-field) or function lvalue and “cv1 T1” is reference-compatible with “cv2 T2”, or
>> - has a class type (i.e., T2 is a class type), where T1 is not reference-related to T2, and can be converted to an rvalue or function lvalue of type “cv3 T3”, where “cv1 T1” is reference-compatible with “cv3 T3” (see [over.match.ref]),
>
>then **the value of the initializer expression** in the first case and the result of the conversion in the second case is called the converted initializer. [....]

*the value of the initializer expression* sounds a bit strange, it seems that an lvalue-to-rvalue conversion would apply to the original initializer expression(if necessary) to convert it be a prvalue in order to read the value, as per [basic.lval#5]  
> The result of a prvalue is the value that the expression stores into its context; a prvalue that has type cv void has no result. A prvalue whose result is the value V is sometimes said to have or name the value V.  

In my mind, an "rvalue" is either a "xvalue" or a "prvalue", Its glvalue(possible-converted) could be directly bound to reference. Why do we say "the value of the initializer expression in the first case" here? We just need to say that   
> the initializer expression in the first case...  

is sufficient.   

The wording "the value of the initializer expression" keeps using since the c++11 standard, that is
> then the reference is bound to the value of the initializer expression in the first case and to
the result of the conversion in the second case (or, in either case, to an appropriate base class
subobject).  

But, in that document, the value of an expression seems to be not well-defined and has a different meaning in the current standard, hence, the wording is not suitable now.

jensmaurer:
when the properties of the expression are still relevant.

Fixes #4753

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -5475,8 +5475,8 @@ where ``\cvqual{cv1} \tcode{T1}'' is
 reference-compatible with ``\cvqual{cv3} \tcode{T3}'' (see~\ref{over.match.ref}),
 \end{itemize}
 then
-the value of the initializer expression in the first case and
-the result of the conversion in the second case
+the initializer expression in the first case and
+the converted expression in the second case
 is called the converted initializer.
 If the converted initializer is a prvalue,
 its type \tcode{T4} is adjusted to type ``\cvqual{cv1} \tcode{T4}''\iref{conv.qual}


tkoeppe:
Hm, I think this is reasonable, but I'd welcome another CWG opinion on this. For example, are we implying here that "value" should generally be understood to imply lvalue-to-rvalue conversion, and we need to review the standard for other such uses, rather than, say, considering every evaluated expression to constitute a value (of the expression's value category)?

jensmaurer:
I think the use of "value" is simply misleading in this particular case. We use "value" for quite a few things in the standard, including the "value" of template parameters (such a value might be a type). In this case, we're just looking at the expression and should avoid any idea that we're prematurely runtime-evaluating that expression in any way. I don't think a full-scope review is in order for these things; the entire [expr] section could do with a brush-up, but that's probably too large to cover in an editorial issue.

tkoeppe:
I just meant another opinion by someone from CWG, not necessarily by the actual group. Let me ask around.

source/declarations.tex
```diff
@@ -5475,7 +5475,7 @@
 reference-compatible with ``\cvqual{cv3} \tcode{T3}'' (see~\ref{over.match.ref}),
 \end{itemize}
 then
-the value of the initializer expression in the first case and
+the initializer expression in the first case and
 the result of the conversion in the second case
```
zygoloid, line 5479: I think the second case is also wrong. Suggestion:
```suggestion
the converted expression in the second case
```
A "result" of an expression is a value or object, and can't be a prvalue as required in the next sentence.

jensmaurer:
Fixed.

PR #4810, opened 2021-08-12
[over.match.best] Define 'worst conversion sequence'

### LINKED ISSUE #4809: [over.ics.list] define "worst conversion sequence" CWG2492
As agreed in CWG 2021-08-09 teleconference.

https://wiki.edg.com/bin/view/Wg21telecons2021/IssuesProcessingTeleconference2021-08-09

jensmaurer:
Fixes #4809.

DO NOT MERGE; this is obviously insufficient and incorrect.

### DIFF
--- a/source/overloading.tex
+++ b/source/overloading.tex
@@ -1811,6 +1811,14 @@ void use() {
 \end{codeblock}
 \end{example}
 
+\pnum
+Given a set of implicit conversion sequences,
+a \defnadj{worst}{conversion sequence} $S$ in the set
+is a conversion sequence
+such that $S$ is not a better conversion sequence
+than any other conversion sequence in the set and
+at least one other conversion sequence is a better conversion sequence than $S$.
+
 \rSec3[over.best.ics]{Implicit conversion sequences}%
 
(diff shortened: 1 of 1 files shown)


xmh0511:
Actually, the corner case for [over.ics.list] p5 is that
````cpp
#include <initializer_list>
struct A{
    operator short(){
        return 0;
    }
};
struct B{
    operator bool(){
        return 0;
    }
};
void fun(std::initializer_list<int>){}  //#1
void fun(std::initializer_list<bool>){}  //#2
int main(){
    fun({A{},B{}});
}
````
The worst conversion sequence selected in the implicit conversion sequence set of `#1` will directly impact the result of comparing it with the obvious worst conversion sequence (`A::short`) in the set of `#2`. Regardless of `A::short` or `B::bool`, they are all the implicit conversion sequence that comprises a user-defined conversion followed by an integral promotion. Which is not better or worse than the other. If we consider `A::short` is the worst conversion sequence in the set of `#1`, then `#1` is selected as per [over.match.best#over.ics.rank-3.3]. Otherwise, the invocation will be ambiguous.

PR #4813, opened 2021-08-14
[basic.lookup.unqual] Clarify 'unqualified name'

### LINKED ISSUE #4812: [basic.lookup.unqual] Component names in converstion-type-id are unqualified names?
[basic.lookup.unqual] p5
> An **unqualified name** that is a component name ([expr.prim.id.unqual]) of a type-specifier or ptr-operator of a conversion-type-id is looked up in the same fashion as the conversion-function-id in which it appears. If that lookup finds nothing, it undergoes unqualified name lookup; in each case, only names that denote types or templates whose specializations are types are considered.

The rule says these components names are called unqualified names, which has conflicted with [basic.lookup.unqual] p4
> An unqualified name is a name that does not follow a nested-name-specifier or the . or -> in a class member access expression ([expr.ref]), possibly after a template keyword or ~. Unless otherwise specified, such a name undergoes unqualified name lookup from the point where it appears.

Consider the example after [basic.lookup.unqual] p5
````cpp
X().operator U T::*();  // #1
X().operator U decltype(T())::*();  //#2
````
The names `U` and `T` in these two expressions all follow `.` in their class member access expressions. Hence, they shouldn't **be unqualified names**. Could we change the *unqualified name* in [basic.lookup.unqual] p5 to  *name*:
> A **name** that is a component name... 

Except for that, [basic.lookup.unqual] p5 also **does not specify how will the lookup perform on these names that are not component names**. Such as the `T` that appears in `decltype(T())` in `#2`, it is not the component name of neither a type-specifier nor ptr-operator of a conversion-type-id. Which kind of lookup will perform on that name?

jensmaurer:
Fixes #4812 

@opensdh, please have a look.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -1738,7 +1738,7 @@ from a program point performs an unqualified search in its immediate scope.
 
 \pnum
 An \defnadj{unqualified}{name} is a name
-that does not follow a \grammarterm{nested-name-specifier} or
+that does not immediately follow a \grammarterm{nested-name-specifier} or
 the \tcode{.} or \tcode{->} in a class member access expression\iref{expr.ref},
 possibly after a \keyword{template} keyword or \tcode{\~}.
 Unless otherwise specified,


opensdh (approved):
This seems trivially correct, and doesn't conflict with the proposal to define qualified names as the inverse of unqualified ones (perhaps minus names not looked up at all).

jensmaurer:
@tkoeppe, this looks ready to merge.

tkoeppe:
Thanks!

PR #4817, opened 2021-08-17
[range.dangling] Simplify syntax

JohelEGP:


CaseyCarter (approved):

jensmaurer:
Do we have a general policy here? I guess there are quite some places (e.g. `std::forward`) where "Args&&..." could be replaced by "auto&& ,.."

jwakely:
std::forward? 🤨 

I think it makes sense here because we don't need names for the types in the pack, it's just a sink accepting anything. Anywhere we need a name for the parameter (or parameter pack) using `auto` isn't an improvement.

jensmaurer:
std::forward -> std::make_unique; sorry for the confusion

JohelEGP:
That still uses `std::forward`, so we'd have to say `std::forward<decltype(args)>(args)...` instead of `std::forward<Args>(args)...`.

jensmaurer:
Ok. So, the rule is we're only using "auto" as a parameter type if we don't need names. Makes sense.

PR #4827, opened 2021-08-19
[expr.typeid] Require class type to be complete

### LINKED ISSUE #4826: [expr.typeid] glvalue of incomplete type vs typeid
````cpp
struct C;
int storage;
C* ptr = (C*)&storage;
typeid(*ptr);  //#1
````
All compilers report `#1` is an error, with the diagnosis `invalid use of incomplete type`. However, there's no relevant rule in the standard that restricts the type of the glvalue operand in [expr.typeid]. Is it an omission?

### PR DESCRIPTION
**Pull request by jensmaurer**
Fixes  #4826

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -3638,6 +3638,12 @@ of the program. Whether or not the destructor is called for the
 \tcode{std::type_info} object at the end of the program is unspecified.
 
 \pnum
+If the type of the \grammarterm{expression} or \grammarterm{type-id} operand is
+a (possibly cv-qualified) class type or
+a reference to (possibly cv-qualified) class type,
+that class shall be completely defined.
+
+\pnum
 When \keyword{typeid} is applied to a glvalue whose type is a
 polymorphic class type\iref{class.virtual}, the result refers to a
 \tcode{std::type_info} object representing the type of the most derived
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: source/expressions.tex [hunk 1]
```diff
@@ -3637,6 +3637,12 @@
 of the program. Whether or not the destructor is called for the
 \tcode{std::type_info} object at the end of the program is unspecified.
 
+\pnum
+If the \grammarterm{expression} operand is of class type, or
```
xmh0511, line 3641: (possible cv-qualified) in both cases is significant here.

jensmaurer: How so? A cv-qualified class type is also a class type.

xmh0511: Since in [basic.types#general] p9, "cv-qualified versions of these types" are all used in these cases, which seems class type does not cover "cv-qualified versions of class type", they distinct.

jensmaurer: Yeah, [basic.types] is not always good about these things.

Added "possibly cv-qualified".

jensmaurer:
Approved by CWG.

jensmaurer:
@tkoeppe , any final thoughts?

jensmaurer:
CWG teleon 2021-12-09: CWG2512 created as an accidental duplicate.

PR #4829, opened 2021-08-19
[ranges] Add missing \expos markers for nested types

### LINKED ISSUE #4819: [ranges] Inconsistent marking of definitions with `// \exposid`
https://github.com/cplusplus/draft/blob/a4e40b0ada8a15b601566af3f96bf89314e7ef60/source/ranges.tex#L2925
https://github.com/cplusplus/draft/blob/a4e40b0ada8a15b601566af3f96bf89314e7ef60/source/ranges.tex#L4794
https://github.com/cplusplus/draft/blob/a4e40b0ada8a15b601566af3f96bf89314e7ef60/source/ranges.tex#L6932
https://github.com/cplusplus/draft/blob/a4e40b0ada8a15b601566af3f96bf89314e7ef60/source/ranges.tex#L7331
Out of all nested iterator and sentinel class (templates), these are the only definitions that are marked.

jensmaurer:
Fixes #4819


CaseyCarter (approved):
FWIW, LWG removed `\expos` from these positions during review of P0896 (obviously only for the types that were introduced by P0896) because people felt we didn't need to remind people for a third time that these types are exposition-only. (I have no personal preference.)

jensmaurer:
@CaseyCarter , that's actually a good point. We use `\expos` only when introducing a name, not when using it. One could argue that the definitions of the nested types are not introducing the name (that was done when defining the class), but just using that name.

A consistent alternative approach would be to remove `\expos` from the four similar places where it currently exists. I think I like that latter approach.

tkoeppe:
Editorial decision: go with Casey's approach, remove extraneous annotations from definitions.

PR #4832, opened 2021-08-20
[expr.prim.req] Fix uses of 'unevaluated operand'

### LINKED ISSUE #4830: [expr.prim.req.nested] misuse of term "unevaluated operand"
> [expr.prim.req.nested]p2: "A local parameter shall only appear as an unevaluated operand (7.2) within the *constraint-expression*."

This allows `requires (T a) { requires sizeof(a) == 4; }` but does not allow `requires (T a) { requires sizeof(f(a)) == 4; }`, because the term "unevaluated operand" only applies to the immediate operand of various contexts. I believe the intent here is "An unqualified-id naming a local parameter shall not be potentially evaluated." or similar.

jensmaurer:
Fixes #4830

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -2694,8 +2694,6 @@ or by checking properties of types and expressions.
 \pnum
 A \grammarterm{requires-expression} is a prvalue of type \tcode{bool}
 whose value is described below.
-Expressions appearing within a \grammarterm{requirement-body}
-are unevaluated operands\iref{term.unevaluated.operand}.
 
 \pnum
 \begin{example}
@@ -2779,6 +2777,7 @@ requires {
 \pnum
 A \grammarterm{simple-requirement} asserts
 the validity of an \grammarterm{expression}.
+The \grammarterm{expression} is an unevaluated operand\iref{expr.context}.
 \begin{note}
 The enclosing \grammarterm{requires-expression} will evaluate to \keyword{false}
 if substitution of template arguments into the \grammarterm{expression} fails.
@@ -2848,7 +2847,9 @@ does not require that type to be complete\iref{basic.types}.
 
 \pnum
 A \grammarterm{compound-requirement} asserts properties
-of the \grammarterm{expression} $E$. Substitution
+of the \grammarterm{expression} $E$.
+The \grammarterm{expression} is an unevaluated operand\iref{expr.context}.
+Substitution
 of template arguments (if any) and verification of
 semantic properties proceed in the following order:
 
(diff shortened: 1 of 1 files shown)


zygoloid (commented):
Generally this looks good to me, but I have one concern:

I wonder if there are any complications here that would suggest we treat *nested-requirement*s as unevaluated operands. I ask because we do treat the *constraint-logical-or-expression* of a *requires-clause* as an unevaluated operand, and it might make sense to be consistent. But I'm not sure how to square that with the disallowance of a *requires-expression* parameter appearing in a potentially-evaluated context.

Maybe we need CWG to take a look. It seems like we're overloading "unevaluated operand" to mean two different things here, and a nested requirement is unevaluated in only one of the two senses.

source/expressions.tex
```diff
@@ -2706,6 +2704,7 @@
 \pnum
 A \grammarterm{simple-requirement} asserts
 the validity of an \grammarterm{expression}.
+The \grammarterm{expression} is an unevaluated operand\iref{expr.prop}.
```
JohelEGP, line 2707:
```suggestion
The \grammarterm{expression} is an unevaluated operand\iref{expr.context}.
```

source/expressions.tex
```diff
@@ -2775,7 +2774,9 @@
 
 \pnum
 A \grammarterm{compound-requirement} asserts properties
-of the \grammarterm{expression} $E$. Substitution
+of the \grammarterm{expression} $E$.
+The \grammarterm{expression} is an unevaluated operand\iref{expr.prop}.
```
JohelEGP, line 2778:
```suggestion
The \grammarterm{expression} is an unevaluated operand\iref{expr.context}.
```

JohelEGP (changes requested):

jensmaurer:
@zygoloid: Well, as it stands with my rewrite, the _constraint-expression_ of a _nested-requirement_ is NOT an unevaluated operand. Which allows us to talk about "potentially evaluated". I'm unsure what we want here.

Is there a minor fix that we could apply so that my patch can go forward, and we punt the rest to CWG?

zygoloid:
I think we can fix the first part of #4830 editorially, since that seems to just be confusion between two similar terms:

[expr.prim.req.nested]p2: "<del>A local parameter shall only appear as an unevaluated operand (7.2) within the constraint-expression.</del> <ins>An expression naming a local parameter shall not be potentially evaluated.</ins>"

However, given the larger issue here -- which seems core-ish to me -- I'm not sure that's the right approach. So perhaps we should punt the whole thing to core?

PR #4833, opened 2021-08-21
[basic.pre,basic.link] Clarify 'declaration of entity' vs. namespace-alias

### LINKED ISSUE #4787: [basic.scope#scope-4]The meaning of "denote different entities" is vague
> Two declarations potentially conflict if they correspond and cause their shared name to denote **different entities ([basic.link])**. 

Since [basic.link] follows "different entities", it hints that the meaning of "different entities" is given in [basic.link#8]. Consider this example
````cpp
void fun();
extern int fun;
````
According to [basic.scope.scope#3], such two declarations correspond, and according to [basic.link#8], they declare the **same entity**.  According to [basic#pre-5]  
> An entity E is denoted by the name (if any) that is introduced by a declaration of E or by a typedef-name introduced by a declaration specifying E.   

Since these two declarations declare the same entity, does it mean the names `fun` in the two declarations denote the same entity? However, according to [basic#pre-3]  

> An entity is a value, object, reference, structured binding, function, enumerator, type, class member, bit-field, template, template specialization, namespace, or pack. 

How can a function and a variable be the same entity? It seems result in [this](https://stackoverflow.com/questions/65566185/why-a-structured-binding-does-not-potentially-conflict-with-a-variable) issue. That is, what's the exact meaning of "denote different entities"? Merely determined by [basic.link#8] or should base it on by adding some other restrictions?

### PR DESCRIPTION
**Pull request by jensmaurer**
Fixes #4787

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -85,9 +85,16 @@ implicit declaration of an injected-class-name\iref{class.pre}.
 The interpretation of a \grammarterm{for-range-declaration} produces
 one or more of the above\iref{stmt.ranged}.
 \end{note}
-An entity $E$ is denoted by the name (if any)
-that is introduced by a declaration of $E$ or
-by a \grammarterm{typedef-name} introduced by a declaration specifying $E$.
+An entity $E$ is denoted by
+\begin{itemize}
+\item
+the name (if any) that is introduced by a declaration of $E$, or
+\item
+a \grammarterm{typedef-name} introduced by a declaration specifying $E$, or
+\item
+a \grammarterm{namespace-alias} introduced by
+a \grammarterm{namespace-alias-definition} denoting $E$.
+\end{itemize}
 
 \pnum
 A \defn{variable} is introduced by the
@@ -2716,7 +2723,15 @@ they both declare names with module linkage and are attached to the same module,
 they both declare names with external linkage.
 \end{itemize}
 \begin{note}
-There are other circumstances in which declarations declare the same entity%
+An \grammarterm{alias-declaration}\iref{dcl.typedef},
+a \grammarterm{using-declarator}\iref{namespace.udecl}, or
+a \grammarterm{namespace-alias-definition}\iref{namespace.alias}
+does not declare any entities;
+a \grammarterm{typedef-name} or \grammarterm{namespace-alias}
+is not an entity\iref{basic.pre}.
+\end{note}
+\begin{note}
+There are other circumstances in which declarations declare the same entity
 \iref{dcl.link,temp.type,temp.spec.partial}.
 \end{note}
 

### CODE REVIEW

**Comment by jensmaurer** [comment]
@opensdh, any opinion here?

**Comment by xmh0511** [comment]
Maybe, change the link at the end of the first sentence of [basic.scope.scope] p4 to let it links to [basic.pre] p5. It is clearer.

> Two declarations potentially conflict if they correspond and cause their shared name to denote different entities (**[basic.pre]**).

Since it talks about what's entity the name denotes. [basic.pre] p5 is exactly talking about what the entity the corresponding name introduced by the declaration denotes. 

For *an entity E is denoted by the name (if any) that is introduced by a declaration of E*, we can determine whether these two declarations declare the same entity or not as per [basic.link] in order to judge whether the names introduced by a declaration that declares an entity denote the same entity or not. For the remaining case, such as typedef-name or namespace-alias, we can judge them by checking whether the declaration specifying the same entity.

**Comment by jensmaurer** [comment]
We still need a cross-reference to basic.link for "same entity" somewhere in the (logical) vicinity.

**Comment by xmh0511** [comment]
Yes. However, the reference to [basic.link] can be indirectly referenced by [basic.pre] p5. Consider the first sentence 
> An entity E is denoted by the name (if any) that is introduced by a declaration of E. 

There are two declarations that declare entities, which both introduced the shared name "N", that sentence says "N" denote the entity E introduced by the declaration. Hence, if we want to judge whether the shared name "N" denotes the same entity, we should first determine whether the two declarations declare the same entity, which needs the help of [basic.link].

The remaining case is these names introduced by declarations that ** do not declare entities**. [basic.link] does not mention this case, but [basic.pre] p5 does.

File: source/basic.tex [hunk 1]
```diff
@@ -2680,6 +2687,14 @@
 they both declare names with external linkage.
 \end{itemize}
 \begin{note}
+An \grammarterm{alias-declaration}\iref{dcl.typedef},
```
xmh0511, line 2690: How about `using type = struct A{};`?

opensdh: It's true that an *alias-declaration* can **contain** a declaration of an entity, but I think it's clear here that it is the direct effect of the top-most declaration that is in question.

I'm not sure I understand the second question, although note that the paragraph begins by "enabling" names for linkage for purposes.

opensdh, line 2690: Why not add this note instead to [basic.pre]/5 (changed above), perhaps phrased as

> A *typedef-name* or *namespace-alias* is not itself an entity.  A *using-declarator* does not introduce an entity.

?  I think *namespace-alias-definition* is then covered by implication.  (The non-parallelism in that wording results from the lack of a pseudo-grammar term like *using-name* to distinguish identifiers with such a declaration, which I regard as no great omission.)

If we prefer to leave it here, I think it goes better after the "other circumstances" note (which is more directly related to the paragraph).

opensdh (approved):

source/basic.tex
xmh0511: I think `alias-declaration` and `declaration specified with typedef` should have the same treatment since they are merely different ways to declare the identifier to be a typedef-name. In line 2690, we just say `alias-declaration` but do not mention the declaration with typedef specifier, it's the second question.

xmh0511:
@opensdh How about https://github.com/cplusplus/draft/pull/4833#issuecomment-903078195 this opinion? I just think [basic.scope#scope-4] is talking about what's the entity the **name denotes**. [basic.pre] p5 is directly related to [basic.scope#scope-4]. Instead, [basic.link] p8 is just an indirect rule about whether the name denotes the same entity if the declaration declares the entity.

opensdh:
> @opensdh How about [#4833 (comment)](https://github.com/cplusplus/draft/pull/4833#issuecomment-903078195) this opinion?

I think we're generally in agreement.

AlisdairM:
Has this PR become moot with the notion of underlying entity in C++26?

The _underlying entity_ of an entity is that entity unless otherwise specified. A name _denotes_ the
underlying entity of the entity declared by each declaration that introduces the name.

[Note 5: Type aliases and namespace aliases have underlying entities that are distinct from themselves. —end note]

jensmaurer:
Yes, I think "underlying entity" has sufficiently changed the landscape here that this change is no longer appropriate.

PR #4834, opened 2021-08-21
[dcl.spec.general,dcl.fct.spec] Clarify duplication of decl-specifiers

### LINKED ISSUE #4796: [dcl.spec.general] p2 prevents the valid type combination
````cpp
unsigned  long  long int const volatile a = 0;
````
In this declaration, the complete decl-specifier-seq is `unsigned  long  long int const volatile`. Where *unsigned*, *long*, *long*, *int* are all *simple-type-specifier*s, which in turn are *type-specifier*s, which are also *defining-type-specifier*s; While *const* and *volatile* are both *cv-qualifier*s, which are *type-specifier*s, which are also *defining-type-specifier*s. Hence, in this complete  *decl-specifier-seq*, the defining-type-specifier appears six times. 

However, [dcl.spec.general] p2 says  
> Each decl-specifier **shall** appear at most once in a complete decl-specifier-seq, except that long may appear twice.

Obviously, *defining-type-specifier* appears more than once that are not only *long*.   I think the intent of this rule says each indivisible specifier shall appear at most once except that *long* may appear twice, I would like to name the indivisible specifier as **atomic** specifier(e.g., int, long, short, const, virtual, explicit ...). But It's not formal wording, maybe we can use some terms to precisely phrase such the meaning?

jensmaurer:
Fixes #4796

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -285,9 +285,9 @@ affects the type only for the declaration it appears in, not other declarations 
 same type.
 
 \pnum
-Each \grammarterm{decl-specifier}
-shall appear at most once in a complete \grammarterm{decl-specifier-seq},
-except that \tcode{long} may appear twice.
+At most one of each of the \grammarterm{decl-specifier}s
+\keyword{friend}, \keyword{typedef}, or \keyword{inline}
+shall appear in a \grammarterm{decl-specifier-seq}.
 At most one of
 the \keyword{constexpr}, \keyword{consteval}, and \keyword{constinit} keywords
 shall appear in a \grammarterm{decl-specifier-seq}.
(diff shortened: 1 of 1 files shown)


tkoeppe:
This looks reasonable, but  I would welcome a quick Core cross-check, given the ever-increasing complexity of decl-specifiers.

jensmaurer:
Put onto core's inbox: https://wiki.edg.com/bin/view/Wg21telecons2021/CoreWorkingGroup

tkoeppe:
CWG meeting consensus Nov 11: This is an improvement. We acknowledge that this wording strategy comes with a maintenance burden, but it's better than the status quo.

jensmaurer:
@tkoeppe, what are we  waiting for with this fix?

tkoeppe:
I think it's fine to go, thanks!

PR #4843, opened 2021-08-26
[basic.scope.block] Fix regression introduced by P1787R6 CWG2502

### LINKED ISSUE #4841: [basic.scope.block] p1 The note should be as a normative rule
> Each
>> - selection or iteration statement ([stmt.select], [stmt.iter]),
>> - substatement of such a statement,
>> - handler ([except.pre]), or
>> - compound statement ([stmt.block]) that is not the compound-statement of a handler
>
> introduces a block scope that includes that statement or handler.
>
>[Note 1: A substatement that is **also a block has only one scope**. — end note]

Isn't that note should be phrased as a normative rule? My proposal is that
> Each
>> - [...]
>> - substatement of such a statement that is **not a compound-statement**,
>> - [...]
>> - compound statement ([stmt.block]) that is not the compound-statement of a handler
>
> introduces a block scope that includes that statement or handler.
>
>[Note 1: A substatement that is also a block has only one scope. — end note]. 

````cpp
if(true) {
}
````
The substatement is a compound-statement, hence the second bullet does not apply but the last bullet does, which means the substatement only introduced one block scope.

jensmaurer:
"if (int a = 1) if (int a = 1) ..." is intended to be valid.

Fixes #4841

http://lists.isocpp.org/core/2021/08/11372.php

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -1190,7 +1190,8 @@ If a declaration whose target scope is the block scope $S$ of a
 \grammarterm{compound-statement} of a \grammarterm{lambda-expression},
 \grammarterm{function-body}, or \grammarterm{function-try-block},
 \item
-substatement of a selection or iteration statement, or
+substatement of a selection or iteration statement
+that is not itself a selection or iteration statement, or
 \item
 \grammarterm{handler} of a \grammarterm{function-try-block}
 \end{itemize}


tkoeppe:
Does everyone agree that this was the original intention, and that this is really editorial? @opensdh, @zygoloid?

opensdh:
I can confirm that I didn't intend to forbid that valid C++20 code.  It wasn't immediately clear to me when it was brought up on the reflector whether or not that was a "happy accident" that it prohibited bad code.  Richard pointed out that use in `else if` is not nearly so strange, so I'm happy with Jens' fix.

tkoeppe:
Yes, I like `else if`, too, one of my better-liked kinds of statement. Until recentlly I think the Google style guide had banned it (presumably inadvertently!), but I fixed that :-)

PR #4848, opened 2021-08-30
[expr.delete] Clarify treatment of arrays of classes

jensmaurer:
http://lists.isocpp.org/core/2021/08/11368.php

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -5540,10 +5540,13 @@ versions\iref{class.free}.
 \pnum
 If the keyword \keyword{delete} in a \grammarterm{delete-expression}
 is not preceded by the unary \tcode{::} operator and the type of the operand is
-a pointer to a (possibly cv-qualified) class type \tcode{T}:
+a pointer to a (possibly cv-qualified) class type \tcode{T}
+or (possibly multidimensional) array thereof:
 \begin{itemize}
 \item
-If \tcode{T} has a virtual destructor,
+For a single-object delete expression,
+if the operand is a pointer to \cv{} \tcode{T} and
+\tcode{T} has a virtual destructor,
 the deallocation function is the one selected at the point of definition of
 the dynamic type's virtual destructor\iref{class.dtor}.
 \item


source/expressions.tex
```diff
@@ -5540,7 +5540,8 @@
 \pnum
 If the keyword \keyword{delete} in a \grammarterm{delete-expression}
 is not preceded by the unary \tcode{::} operator and the type of the operand is
-a pointer to a (possibly cv-qualified) class type \tcode{T}:
+a pointer to a (possibly cv-qualified) class type \tcode{T}
```
xmh0511, line 5543: For the first bullet, it's relevant to [class.dtor] p16, which is defined as
> At the point of definition of a virtual destructor (including an implicit definition), the **non-array deallocation function** is determined as if for the expression delete this appearing in a non-virtual destructor of the destructor's class (see [expr.delete]).   

However, the pointer may point to the initial element of such type of an array.

jensmaurer: Arrays cannot contain polymorphic data, thus it's fine to use the deallocation function from the definition of the virtual destructor, because the static and the dynamic type must be the same here.

xmh0511: In [class.dtor] p16, it only specifies the approach about how to determine the **non-array deallocation function**. However, when an array object is being deleted, as per [expr.delete] p8 

> If a deallocation function is called, it is operator delete for a single-object delete expression or operator delete[] for an array delete expression.  

We should find `delete[]` for this purpose. 
````cpp
struct A{
    void operator delete[](void* ptr){
        std::cout<<"#1\n";
    }
    void operator delete(void* ptr){
      std::cout<<"#2\n";
    }
    virtual ~A(){}
};
int main(){
   auto ptr = new A[2][3];
   delete [] ptr;
}
````
`#1` invokes

Comment (jensmaurer) [reply source/expressions.tex:5543]: Ah, you're saying that [class.dtor] p16 only determines a non-array operator delete, but [expr.delete] p9.1 as changed by this pull request also wants to use some `operator delete[]` presumably found that way.

Comment (jensmaurer) [reply source/expressions.tex:5543]: Fixed.

File: source/expressions.tex [hunk 2]
```diff
@@ -5540,10 +5540,12 @@
...
+or (possibly multidimensional) array thereof:
 \begin{itemize}
 \item
-If \tcode{T} has a virtual destructor,
+If the operand is a pointer to \cv[} \tcode{T} and
```
JohelEGP, line 5547:
```suggestion
If the operand is a pointer to \cv{} \tcode{T} and
```

source/expressions.tex
```diff
@@ -5540,10 +5540,12 @@
 \pnum
 If the keyword \keyword{delete} in a \grammarterm{delete-expression}
 is not preceded by the unary \tcode{::} operator and the type of the operand is
-a pointer to a (possibly cv-qualified) class type \tcode{T}:
+a pointer to a (possibly cv-qualified) class type \tcode{T}
+or (possibly multidimensional) array thereof:
 \begin{itemize}
 \item
-If \tcode{T} has a virtual destructor,
+If the operand is a pointer to \cv{} \tcode{T} and
```
xmh0511, line 5547:
````cpp
auto ptr = new A[2];
delete[] ptr;
`````
`ptr` is a pointer to the initial element of the array. Hence, we could say it is a pointer to `T`. However, I think the lookup deallocation function in this case is still `operator delete[]`.

Comment (jensmaurer) [reply source/expressions.tex]: Fixed

File: source/expressions.tex [hunk 4]
```diff
@@ -5540,10 +5540,13 @@
...
 If the keyword \keyword{delete} in a \grammarterm{delete-expression}
 is not preceded by the unary \tcode{::} operator and the type of the operand is
-a pointer to a (possibly cv-qualified) class type \tcode{T}:
+a pointer to a (possibly cv-qualified) class type \tcode{T}
+or (possibly multidimensional) array thereof:
 \begin{itemize}
 \item
-If \tcode{T} has a virtual destructor,
+For a single-object delete expression,
+if the operand is a pointer to \cv{} \tcode{T} and
```
xmh0511, line 5548: I still want to refer this example, but a bit difference
````cpp
auto ptr = new A[2];
ptr++; // now, ptr point to the `1` element of the array, and its an object of type `A`
delete ptr; 
````
It seems `delete ptr; ` will cause UB in most implementations. The operand of the single-object delete expression should be restricted to that produced by single-object new-expression, however it didn't represent in [expr.delete#2].

jensmaurer: I agree this should say "complete object". That's not this issue, though.

jensmaurer:
CWG 2021-09-21: Approved.

jensmaurer:

PR #4859, opened 2021-09-02
[expr.delete] Clarify constraints on non-array delete

### LINKED ISSUE #4853: [expr.delete] p2 The restriction on operand of a single-object delete expression
> In a single-object delete expression, the value of the operand of delete may be a null pointer value, a pointer to a non-array object created by a previous new-expression, or a pointer to a subobject representing a base class of such an object. If not, the behavior is undefined. 

Which previous *new-expression* does the rule refer to? Presumably, it shall be a new-expression that allocated an object of non-class type. Hence, the above rule should be that  

> In a single-object delete expression, the value of the operand of delete may be a null pointer value, a pointer to a non-array object created by a previous new-expression **that allocated an object of non-class type**, or a pointer to a subobject representing a base class of such an object. If not, the behavior is undefined.  

Since we although said "a pointer to a non-array object", however, it is not sufficient, a pointer to an initial element of an array is also considered as "a pointer to a non-array object "

jensmaurer:
Fixes #4853

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -5431,11 +5431,12 @@ type by calling the above-mentioned conversion function, and the
 converted operand is used in place of the original operand for the
 remainder of this subclause.
 In a single-object delete expression, the value of the operand of
-\keyword{delete} may be a null pointer value, a pointer to a non-array object
-created by a previous \grammarterm{new-expression},
-or a pointer to a
-subobject\iref{intro.object} representing a base class of such an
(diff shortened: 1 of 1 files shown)


jensmaurer:
http://lists.isocpp.org/core/2021/08/11409.php

jensmaurer:
CWG 2021-09-21: Approved.

jensmaurer:

PR #4860, opened 2021-09-02
[conv.integral,over.best.ics.general,over.ics.user] Ignore cv-qualification differences

### LINKED ISSUE #4851: [conv.qual] identity conversion and similar type
In [dcl.init.ref] p4, the "reference-related" is defined as
> Given types “cv1 T1” and “cv2 T2”, “cv1 T1” is reference-related to “cv2 T2” if T1 is **similar** ([conv.qual]) to T2, or T1 is a base class of T2.   

Consider this example
````cpp
int a;
int&& rf = a;  // ill-formed
````
Presumably, it is restricted by [dcl.init.ref#5.4.4] 
> If T1 is **reference-related** to T2:
>> - [...]
>> - if the reference is an **rvalue reference**, the **initializer expression shall not be an lvalue**.

Is the type `int` similar to `int`? It should be determined by [conv.qual] p2
> Two types T1 and T2 are similar if they have qualification-decompositions with the same n such that corresponding  **P<sub>i</sub> components are either the same or one is “array of N<sub>i</sub>" and the other is “array of unknown bound of”**, and the types denoted by U are the same.

However, the qualification-decomposition  of type `int` does not have any **P<sub>i</sub>** component. Maybe, one could argue since they are empty, they're the same.  

*** 
From a prvalue of type `int` to type `int`, or from a prvalue of type `int` to `int const volatile`, Is it a qualified conversion? It seems [conv.qual] p3 can apply to these two conversions except that whether [conv.qual#3.2] is suitable for empty P<sub>i</sub>

***
For class type, it is a bit different, since a prvalue of cv class type can retain the cv-qualification. So, assume `T` is a class type, from a prvalue of type `volatile const T` to type `T` is a user-defined conversion? Why I say it is a **user-defined conversion** because of [over.best.ics#general-6] and [over.ics.user]  

> When the parameter has a class type and the argument expression has **the same type**, the implicit conversion sequence is an identity conversion. When the parameter has a class type and the argument expression has a derived class type, the implicit conversion sequence is a derived-to-base conversion from the derived class to the base class.

> A conversion of an expression of class type to **the same class type** is given Exact Match rank, and a conversion of an expression of class type to a base class of that type is given Conversion rank, in spite of the fact that a constructor (i.e., a user-defined conversion function) is called for those cases.  

Since `T` and `volatile const T` are not the same type, hence the conversion is user-defined conversion?

### PR DESCRIPTION
**Pull request by jensmaurer**
Fixes #4851

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -899,8 +899,9 @@ This conversion is called \defn{floating-point promotion}.
 
 \pnum
 \indextext{conversion!integral}%
-A prvalue of an integer type can be converted to a prvalue of another
-integer type. A prvalue of an unscoped enumeration type can be converted to
+A prvalue of an integer type \tcode{T} can be converted to
+a prvalue of another integer type not similar\iref{conv.qual} to \tcode{T}.
+A prvalue of an unscoped enumeration type can be converted to
 a prvalue of an integer type.
 
 \pnum
@@ -924,8 +925,10 @@ of integral conversions.
 
 \pnum
 \indextext{conversion!floating-point}%
-A prvalue of floating-point type can be converted to a prvalue of
-another floating-point type. If the source value can be exactly
+A prvalue of a floating-point type \tcode{T} can be converted to
+a prvalue of another floating-point type
+not similar\iref{conv.qual} to \tcode{T}.
+If the source value can be exactly
 represented in the destination type, the result of the conversion is
 that exact representation. If the source value is between two adjacent
 destination values, the result of the conversion is an
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -780,7 +780,7 @@ then \keyword{const} is added to every $\cv{}^3_k$ for $0 < k < i$,
 where $\cv{}^j_i$ and $P^j_i$ are the components of
 the qualification-decomposition of $\tcode{T}j$.
 A prvalue of type \tcode{T1}
-can be converted to type \tcode{T2}
+can be converted to another type \tcode{T2}
 if the qualification-combined type of \tcode{T1} and \tcode{T2} is \tcode{T2}.
 \begin{note}
 If a program could assign a pointer of type \tcode{T**} to a pointer of
--- a/source/overloading.tex
+++ b/source/overloading.tex
@@ -1920,8 +1920,9 @@ contains no ``conversion'' from
 to
 \tcode{A}.
 \end{example}
-When the parameter has a class type and the argument expression has the
-same type, the implicit conversion sequence is an identity conversion.
+When the parameter has a class type and the argument expression has a
+similar type\iref{conv.qual},
+the implicit conversion sequence is an identity conversion.
 When the parameter has a class type and the argument expression has a
 derived class type, the implicit conversion sequence is a
 derived-to-base
@@ -2085,7 +2086,7 @@ the second standard conversion sequence shall have exact match rank.
 
 \pnum
 A conversion of an expression of class type
-to the same class type is given Exact Match rank, and
+to a similar\iref{conv.qual} class type is given Exact Match rank, and
 a conversion of an expression of class type
 to a base class of that type is given Conversion rank,
 in spite of the

### CODE REVIEW

**Comment by xmh0511** [comment]
In terms of the opportunity to clarify these concept, consider this example
````cpp
void fun(int&){}  // #1
void fun(int){}   // #2
int a = 0;
fun(a);
````
The **identity conversion** is vague here. Since using a glvalue `a` to initialized the parameter at `#2` with type int, the conversion comprises "lvalue-to-rvalue conversion" as per the order defined in [conv.general] p1, now the result is a prvalue of type int, since the definition of [conv.qual], from a `prvalue of type int` to a `prvalue of type int` is qualification conversion. Instead, for `#1`, we explicitly say it is an identity conversion [over.ics.ref] p1. Hence, as per [over.ics.rank#3.2.1], `#1` is the best viable function? Actually, the implementations report ambiguity.

**Comment by jensmaurer** [comment]
I think the intent is that all conversions described in [conv] are not "empty".

This is clear in the formulation of integral and floating-point conversions, where we say "another ... type". Which didn't properly exclude mere differences in cv-qualification (first patch in this pull request). Similarly, [conv.qual] should only allow conversions to a different type, excluding a conversion from T to T (second patch in this pull request).

**Comment by xmh0511** [comment]
Please still consider [conv.qual], give such two example
````cpp
const volatile int v = 0;  // #1
const volatile Class obj = Class();  // #2
````
The target type of `#1` and `#2` are all cv-qualified types.  Consider `#1`, as per [dcl.init#general-15.9], a standard conversion sequence will be used **if necessary**, although the fixed [conv.integral] has rejected to be used, however, [conv.qual] still permit to be used if they are different types. Is there **necessary** to convert `int` to `const volatile int` by a qualification conversion? Presumably, it's not necessary, `#2` is also similar. In these cases, a prvalue of cv1 T is sufficient to initialize the object of cv2 T, a qualification conversion is not necessary here. If it would be permitted, it will impact [over.rank].  More generally, I think [conv.qual] is meaningful to be used only in the case that the cv-qualifiers **other than cv0** of two similar types are different.

**Comment by jensmaurer** [comment]
I think there are situations where the need for a top-level qualification-conversion is significant:
```
int f(int &x);
int f(const int& x);
int a;
int y = f(a);  // chooses f(int&)
```

Note also that forming a conversion sequence in overload resolution is different from the actual initialization in [dcl.init]. For the actual initialization, [dcl.init] should (does it?) ignore top-level cv-qualifiers on the target.

**Comment by xmh0511** [comment]
IMO, regardless of the first or second overload, the parameters are all of the **reference type**, a valid reference type cannot be cv-qualified, in this case, the rank is impacted by [over.ics.rank#3.2.6], that rule explicitly says we could judge from their **referenced type**. I think the qualification conversion is almost prepared for pointer types and their top-level cv-qualifiers do not have an effect.  
````cpp
int* ptr = 0;
int* const ptr1 = ptr // lvalue-to-rvalue conversion is sufficient

int const* ptr2 = ptr // lvalue-to-rvalue conversion followed by a qualification conversion
````
Hence, we can generally say qualification conversion is not necessary to apply to two similar types where their difference only in **cv0**(i.e. the top-level cv-qualification of that type).

File: source/expressions.tex [hunk 1]
```diff
@@ -780,7 +780,7 @@
 where $\cv{}^j_i$ and $P^j_i$ are the components of
 the qualification-decomposition of $\tcode{T}j$.
 A prvalue of type \tcode{T1}
-can be converted to type \tcode{T2}
+can be converted to another type \tcode{T2}
```
xmh0511, line 783: Since we have excluded the identity conversion from qualification conversion, [expr.const.cast] p7 should also have to be changed to adhere this modification

[expr.const.cast] p7 says
> A conversion from a type T1 to a type T2 casts away constness if T1 and T2 are different, there is a qualification-decomposition ([conv.qual]) of T1 yielding n such that T2 has a qualification-decomposition of the form
>> - [...]
>
> and there is **no qualification conversion** that converts T1 to
>> - [...]  named T3
> 

Consider this example:
````cpp
int* ptr = 0;
static_cast<void*>(ptr); 
````
In this example, `T1` is "pointer to int" while `T2` is "pointer to void". When n=1, `T3` is `pointer to void`, since `T2` and `T3` are the same types, hence there is no qualification conversion that converts `T2` to `T3`([conv.qual] p3 excludes identify conversions).  This is not the intent of the definition of the cast-away constness. 

[expr.const.cast] p7 should be changed to
> A conversion from a type T1 to a type T2 casts away constness if T1 and T2 are different, there is a qualification-decomposition ([conv.qual]) of T1 yielding n such that T2 has a qualification-decomposition of the form
>> - [...]
>
> and there is **no qualification conversion** that converts T1 to
>> - [...]  
>> 
> **if they are different types**
>

PR #4862, opened 2021-09-03
[basic.string.general] Fix basic_string::npos

jensmaurer:
Use constexpr and an explicit conversion to size_type,
for consistency with basic_string_view.

Requested by http://lists.isocpp.org/lib-ext/2021/09/19813.php

### DIFF
--- a/source/strings.tex
+++ b/source/strings.tex
@@ -807,7 +807,7 @@ namespace std {
     using const_iterator         = @\impdefx{type of \tcode{basic_string::const_iterator}}@; // see \ref{container.requirements}
     using reverse_iterator       = std::reverse_iterator<iterator>;
     using const_reverse_iterator = std::reverse_iterator<const_iterator>;
-    static const size_type npos  = -1;
+    static constexpr size_type npos = size_type(-1);
 
     // \ref{string.cons}, construct/copy/destroy
     constexpr basic_string() noexcept(noexcept(Allocator())) : basic_string(Allocator()) { }


jensmaurer:
@jwakely , please have a look. Is the proposed change editorial from an LWG perspective?

CaseyCarter (approved):

jwakely (approved):
Yes, I think this has no normative impact and qualifies as editorial.

tkoeppe:
@jensmaurer: could you please make the commit description more specific (not just "fix", but actually say what changed)? Then feel free to commit.

jensmaurer:
@tkoeppe, I'm sorry, but this _is_ in fact just for consistency with string_view. "const" vs. "constexpr" doesn't matter in this particular case (static class member of integral type).

I've slightly amended the commit description to avoid "fix".

tkoeppe:
Yes, it's for consistency abstractly, but as I understand, this change _should_ have been made by some previous paper, where it was forgotten? That's the kind of background that would be nice to know (if true).

jensmaurer:
@tkoeppe: No. basic_string is ages old, then came along basic_string_view, which used all the new stuff such as "constexpr". Nobody ever thought about upgrading basic_string to follow. This pull request does it.

tkoeppe:
OK, thanks. I thought the reflector discussion suggested that this was an oversight in http://wg21.link/P0980R1, but that might not actually have been true.

jensmaurer:
Well, that paper did add "constexpr" to basic_string member functions, which is a feature add-on. I'm hard pressed viewing "constexpr" on npos as in-scope of that paper; "constexpr" on a member function is quite different from constexpr on a static variable that was already usable in constant expressions.

tkoeppe:
Yeah, I was thinking that, too -- this is ultimately really nobody's fault.

PR #4869, opened 2021-09-06
[utility.requirements,exception.propagation] Clarify the thrown exceptions that have requirements and guarantees imposed LWG3640

### LINKED ISSUE #4863: [lib] Extend exception propagation to cover the whole call
# Problem and solution

The wording for the specification of facilities that propagate exceptions inadvertently exclude some cases of thrown exceptions. The ones I've identified have the form "any exception thrown by the {(selected) constructor,construction} of {`T`,_some object_}". This formulation excludes exceptions thrown by the initialization of the constructor's parameters ([[expr.call] p7](https://eel.is/c++draft/expr.call#7)). These are _not_ thrown by the constructor nor during the construction of the object (after its construction has begun ([[class.cdtor] p1](https://eel.is/c++draft/class.cdtor))).

Implying that the initialization of the constructor's parameter are part of the effects for which exceptions are propagated can be done with these changes:
- "Any exception thrown by <ins>calling</ins> the (selected) constructor of `T`.
- "Any exception thrown by the ~~construction~~<ins>initialization</ins> of {`T`,_some object_}".

The former is already present in the standard, as exemplified below. As for the latter (emphasis mine),
> The process of initialization described in [dcl.init] applies to all initializations regardless of syntactic context, **including the initialization of a function parameter ([expr.call])**, the initialization of a return value ([stmt.return]), or when an initializer follows a declarator.
> -- https://eel.is/c++draft/dcl.init p1

# Examples

### Example, good

> Constructors and member functions of pair do not throw exceptions unless one of the element-wise operations specified to be called for that operation throws an exception. -- https://eel.is/c++draft/pairs.pair#1

Talking in terms of operations includes initialization.

### Example, bad

> For each tuple constructor, an exception is thrown only if the construction of one of the types in Types throws an exception. -- https://eel.is/c++draft/tuple.cnstr#2

We can make a `tuple` constructor throw an exception other than by constructing one of the specified types: https://godbolt.org/z/nanTbd8sc.
```C++
#include <tuple>
struct X { X(int) { } };
struct Y { operator int() { throw 0; } };
int main() { std::tuple<X>(Y{}); }
```
As described at the top, this throws before beginning the construction of an `X` subobject.

Do note that the constructor above is described to initialize:
> Effects: Initializes the elements in the tuple with the corresponding value in std​::​forward<UTypes>(u). -- https://eel.is/c++draft/tuple.cnstr#13

One could argue that [tuple.cnstr] doesn't say "if and only if". That just leaves in the air what happens in this case.

### Example, good

> Throws: Any exception thrown by the initialization of the selected alternative Tj. -- https://eel.is/c++draft/variant.ctor#18
> Throws: Any exception thrown by calling the selected constructor of T. -- https://eel.is/c++draft/variant.ctor#23

JohelEGP:
Resolves #4863.

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -1427,6 +1427,8 @@ types that support null values.
 \ref{hash.requirements} describes the requirements on hash function objects.
 \ref{allocator.requirements} describes the requirements on storage
 allocators.
+\ref{exception.propagation}
+describes additional requirements and guarantees when propagating exceptions.
 
 \rSec3[utility.arg.requirements]{Template argument requirements}
 
@@ -2217,6 +2219,24 @@ whether or not \tcode{T} is a complete type:
 \item all the member types of \tcode{allocator_traits<X>}\iref{allocator.traits}
   other than \tcode{value_type} are complete types.
 \end{itemize}
+
+\rSec3[exception.propagation]{Exception propagation requirements}
+
+\pnum
+Some functions defined in the \Cpp{} standard library
+impose requirements and guarantees when an exception is thrown from a described evaluation.
+Unless stated otherwise,
+exceptions thrown due to evaluating the subexpression
+that implements the described evaluation
+up to the complete evaluation of its enclosing full-expression
+also have those requirements and guarantees imposed.
+\begin{note}
+This includes exceptions thrown as part of
+initializing parameters,
+evaluating default arguments, and
+destroying temporaries (including discarded-value expressions)
+\iref{expr.call}.
+\end{note}
 
 \rSec2[constraints]{Constraints on programs}
 


JohelEGP:
Looks like only the first bullet of `std::async` might need clarification with this formulation.

tkoeppe:
I would like some LWG input on both this PR and the underlying issue.

JohelEGP:
It seems to me that some select `cwg`-labeled issues reach CWG's plate, but that such is not the case for `lwg`. Should I submit a LWG issue, or is there some process behind these labels?

tkoeppe:
@JohelEGP: If you're unsure if something is editorial, then it's definitely more reliable to start with an LWG issue, and LWG can decide that it's editorial. Otherwise I'll do my best to get LWG's attention, but that process is not terribly reliable. Generally, the less certain I am that something is obviously editorial, the less I'm willing to make changes unilaterally and will try to get some expert input.

CWG has a different work load and has been able to review editorial issues from GitHub reguarly recently, but fundamentally the same applies there, too: if something isn't obviously editorial, it's safer to go via the issues lists. We'll do our best here, but just in case you're looking for hard guarantees.

JohelEGP:
Thank you. I will consider submitting a LWG issue soon.

> We'll do our best here, but just in case you're looking for hard guarantees.

I certainly wouldn't like this to go stale. The same might apply to others who have their issues labeled. Do you think that information is fit for the wiki?

tkoeppe:
Maybe? We already say, "If you are not sure whether your proposed edit falls into the above categories, you should consider submitting it as a technical issue instead.", but we could add that we'll try our best to triage and redirect, but without guarantees?

JohelEGP:
Well, it's been a long time since I've read that. I suppose it's already clear enough.

jensmaurer:
If you do get an LWG issue number, please let us know in a comment.

JohelEGP:
This is now https://wg21.link/LWG3640.

PR #4871, opened 2021-09-06
[unique.ptr.single] Remove redundant preconditions after 079f7d3b02d

### LINKED ISSUE #4872: [unique.ptr.single] Throwing after 079f7d3b02d can be well-defined
079f7d3b02d added `noexcept` to certain constructors that had the precondition that some initialization didn't throw. Before, the effects was that throwing results in undefined behavior according to [structure.specifications]. 079f7d3b02d didn't remove the preconditions, but an exception throw within a `noexcept` function is well-defined. Could this have been an oversight?

JohelEGP:
Resolves #4872.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -8761,9 +8761,6 @@ unique_ptr(unique_ptr&& u) noexcept;
 If \tcode{D} is not a reference type,
 \tcode{D} meets the \oldconcept{MoveConstructible}
 requirements (\tref{cpp17.moveconstructible}).
-Construction
-of the deleter from an rvalue of type \tcode{D} does not
-throw an exception.
 
 \pnum
 \effects
@@ -8871,9 +8868,7 @@ If \tcode{D} is not a reference type, \tcode{D} meets the
 \oldconcept{MoveAssignable} requirements (\tref{cpp17.moveassignable}) and assignment
 of the deleter from an rvalue of type \tcode{D} does not throw an exception.
 Otherwise, \tcode{D} is a reference type;
-\tcode{remove_reference_t<D>} meets the \oldconcept{CopyAssignable}
-requirements and assignment of the deleter from an
-lvalue of type \tcode{D} does not throw an exception.
+\tcode{remove_reference_t<D>} meets the \oldconcept{CopyAssignable} requirements.
 
 \pnum
 \effects


jensmaurer:
This doesn't look right.

"noexcept" means no exception escapes, but violating a precondition (e.g. when moving a deleter throws) is still undefined behavior.

JohelEGP:
There's none. My assumption was that the precondition existed to prevent users from examining the `unique_ptr` object after such a throw, or alternatively, to not have to define the state in such a case. That shouldn't be possible after it was marked `noexcept`, though. I can't think of any other reason why to declare UB, so it doesn't make sense to me that it's still a precondition.

This is definitely not editorial, though. If you agree, I'll close this PR and its issue, and consider submitting an LWG issue.

jensmaurer:
Well, for once, having to catch all exceptions to call terminate for a noexcept violation is overhead you might want to avoid. (We don't require the deleter's move operation to be declared noexcept, we just require that it does not throw an exception for this particular move invocation.)

Anyway, this is not editorial, so I'd appreciate if you closed this issue and submitted an LWG issue if you believe there is a defect in the specification (I don't believe so).

jwakely:
> Well, for once, having to catch all exceptions to call terminate for a noexcept violation is overhead you might want to avoid. (We don't require the deleter's move operation to be declared noexcept, we just require that it does not throw an exception for this particular move invocation.)

The compiler is going to provide the`noexcept` guarantee anyway, unless `unique_ptr` is decorated with some non-standard extension to tell the compiler that an exception cannot happen. In practice, nobody does that, so the terminating branch is still emitted by the compiler (although a good implementation can arrange for the terminate with no overhead anyway).

> Anyway, this is not editorial, so I'd appreciate if you closed this issue and submitted an LWG issue if you believe there is a defect in the specification (I don't believe so).

Yes, this should be an LWG issue.

Dani-Hub:
It would be nice to have the link to this draft PULL request as part of the corresponding LWG issue submission text.

JohelEGP:
> It would be nice to have the link to this draft PULL request as part of the corresponding LWG issue submission text.

Would that be the case even though there are more signatures that are `noexcept` yet have the precondition that an operation doesn't exit with an exception? This PR wouldn't suficce as a Proposed Resolution.

Dani-Hub:
I only recommend that when an LWG issue is submitted as a consequence and strong relation to an issue/pull request from this project, it would be nice to refer to that issue/pull request in the text of the LWG issue submission.

PR #4886, opened 2021-09-09
[swappable.requirements] Requires should now be Preconditions

BRevzin:


source/lib-intro.tex
```diff
@@ -1614,15 +1614,15 @@
 \begin{codeblock}
 #include <utility>
 
-// Requires: \tcode{std::forward<T>(t)} shall be swappable with \tcode{std::forward<U>(u)}.
+// Mandates: \tcode{std::forward<T>(t)} shall be swappable with \tcode{std::forward<U>(u)}.
```
tkoeppe, line 1617: Shouldn't we then also get rid of "shall", since the "mandates" style implies that?

tkoeppe (changes requested):

source/lib-intro.tex
jwakely: Yes, a Mandates condition is just stated. "is swappable with".

tkoeppe (approved):

jensmaurer (approved):

jensmaurer:
When merging, fix the commit message to mention the section label.

source/lib-intro.tex
```diff
@@ -1614,15 +1614,15 @@
 \begin{codeblock}
 #include <utility>
 
-// Requires: \tcode{std::forward<T>(t)} shall be swappable with \tcode{std::forward<U>(u)}.
+// Mandates: \tcode{std::forward<T>(t)} is swappable with \tcode{std::forward<U>(u)}.
```
CaseyCarter, line 1617: Both "is swappable with" and "swappable" have semantic requirements, violations of which an implementation cannot be expected to diagnose. To be pedantically correct, these comments both need to Mandate the syntactic requirements and Precondition the semantic requirements. (This admittedly feels like overkill for an example - I have no better idea, though.)

tkoeppe: How about "requires:"? :-)

CaseyCarter: TFW you remove a term for being too vague and then complain about having to be too precise.

tkoeppe: Given that getting all those requirements right is critical and that this example concerns a presumably very common customization point, it might be worth doing this right.

@BRevzin: would you mind adding appropriate "precondition" comments?

JohelEGP (changes requested):
Please, prefix the [stable.label] to the commit and issue title.

source/lib-intro.tex
timsong-cpp: Or we can just use _Preconditions:_, which is how we do this elsewhere in the library spec.

source/lib-intro.tex
```diff
@@ -1614,15 +1614,15 @@
 \begin{codeblock}
 #include <utility>
```
timsong-cpp, line 1615: Drive-by: since this is an unusually complete example trying to illustrate what user code would look like, perhaps we should `#include <cassert>` here?

jensmaurer:
@tkoeppe, this looks good now.

tkoeppe:
Yes, I thought I'd let @timsong-cpp take a final look.

timsong-cpp (approved):

PR #4889, opened 2021-09-10
[expr.dynamic.cast] Remove apparent condition

JohelEGP:
The preceding paragraph already introduces the
"otherwise" branch for the runtime check whose
algorithm is explain in this paragraph.


jensmaurer (approved):
@zygoloid, this looks trivially-correct.

PR #4892, opened 2021-09-12
[intro.abstract] Reference the sibling subclause

JohelEGP:
This is similar to #3192,
but recognizes that the sibling subclause
should be the destination of the reference,
which covers
conditionally-supported constructs and
locale-specific characteristics.

### DIFF
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -902,7 +902,7 @@ and behavior in these respects.
 \begin{footnote}
 This documentation also includes
 conditionally-supported constructs and locale-specific behavior.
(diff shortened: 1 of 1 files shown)


jensmaurer (approved):
The test appears to fail while installing the LaTeX environment.

JohelEGP:
Going to the parent directory shows that the same dependency with 4.3 exists. It must have to do something with an outdated local repository.

jensmaurer:
Please rebase to HEAD and force-push so that the automated test can run.

PR #4896, opened 2021-09-12
[basic.life] Use Oxford comma

JohelEGP:

PR #4897, opened 2021-09-13
[actions] Update dependencies cache before installing

jensmaurer:
because Ubuntu 20.04 can otherwise no longer install ghostscript,
which is needed for LaTeX.


pzygielo:
> because Ubuntu 20.04 can no longer install ghostscript (needed for LaTeX).

What do you mean by that? It looks like gs was installed https://github.com/cplusplus/draft/runs/3560433503#step:4:9 with no problem.

pzygielo:
I see, https://github.com/cplusplus/draft/pull/4896/checks?check_run_id=3581666007

PR #4902, opened 2021-09-13
[range.filter.iterator] Add noexcept to base() functions

hewillk:
I want to know if this is an editorial issue? If it is worthy of being an LWG, I think I will submit it for it since LWG3391 can also add noexcept.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -3469,7 +3469,7 @@ namespace std::ranges {
     @\exposid{iterator}@() requires @\libconcept{default_initializable}@<iterator_t<V>> = default;
     constexpr @\exposid{iterator}@(filter_view& parent, iterator_t<V> current);
 
-    constexpr const iterator_t<V>& base() const &;
+    constexpr const iterator_t<V>& base() const & noexcept;
     constexpr iterator_t<V> base() &&;
     constexpr range_reference_t<V> operator*() const;
     constexpr iterator_t<V> operator->() const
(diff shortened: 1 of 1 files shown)


CaseyCarter:
I'm not an editor, but I play one on TV. I think this falls just short of editorial. Obviously the Editors can investigate the code and quickly determine that it's trivial to implement these functions in a way that cannot emit exceptions, but `noexcept` is effectively a declaration that WG21 will _never_ change this function in a way that would require an implementation that might emit exceptions. That level of commitment to maintaining `noexcept` in the future is not something editors can decide trivially, but a question for the experts in LWG.

jensmaurer:
LWG3533 did not endeavor to add or remove `noexcept`, so the issue is not related.

As @CaseyCarter explained, adding `noexcept` to a function that didn't have it before without showing a paper trail that clearly establishes editorial error is not an editorial issue. Please file an LWG issue if you feel a change is warranted in that area.

hewillk:
> but `noexcept` is effectively a declaration that WG21 will never change this function in a way that would require an implementation that might emit exceptions.

@CaseyCarter. Do you mean that even for [`lazy_split_view​::​outer-iterator​::​value_type::end()`](https://eel.is/c++draft/range.adaptors#range.lazy.split.outer.value-3) which only returns `default_sentinel`, adding a `noexcept` for it is still worth an LWG issue?

I tried to combine this `end()` with the above `base() const &`s into a single LWG, but it seems more worthy of being independent.

CaseyCarter:
> @CaseyCarter. Do you mean that even for [`lazy_split_view​::​outer-iterator​::​value_type::end()`](https://eel.is/c++draft/range.adaptors#range.lazy.split.outer.value-3) which only returns `default_sentinel`, adding a `noexcept` for it is still worth an LWG issue?

Yes - that's not an editorial change.
 
> I tried to combine this `end()` with the above `base() const &`s into a single LWG, but it seems more worthy of being independent.

I'd file these as distinct issues: ideally you want an LWG issue to be as small and simple as possible so it's a no-brainer for folks to say "yes, this is an issue and the resolution is obviously correct." Doing so enables LWG to avoid lots of heavy process, making it more likely your resolution will be applied sooner rather than later.

PR #4905, opened 2021-09-14
[dcl.spec.auto.general] Remove redundant, mutually exclusive case

### LINKED ISSUE #4904: [dcl.spec.auto.general] Definition of _generic parameter type placeholder_ can be simplified
The paragraph in question, [p2](https://eel.is/c++draft/dcl.spec.auto.general#2), lists two places in the grammar where a _placeholder-type-specifier_ of the form _type-constraint_<sub>opt</sub> `auto` can appear. It leaves one unused and defines a name for the other.

The two are mutually exclusive:
- a decl-specifier of the decl-specifier-seq of a parameter-declaration of a function declaration or lambda-expression (the grammar for parameters), and
- the auto type-specifier introducing a trailing-return-type (the grammar for the return type).

All the paragraph does is define a name for the former. The latter, redundantly excluded, also complicates the surrounding wording which could otherwise read simpler.

Note that this might be concurrently modified, as the paragraph or its subclause seem to have more problems as pointed out at https://bugs.llvm.org/show_bug.cgi?id=51856#c3.

JohelEGP:
Resolves #4904.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -1678,14 +1678,11 @@ from an initializer.
 \pnum
 A \grammarterm{placeholder-type-specifier} of the form
 \opt{\grammarterm{type-constraint}} \keyword{auto}
-can be used as a \grammarterm{decl-specifier} of
+that is a \grammarterm{decl-specifier} of
 the \grammarterm{decl-specifier-seq} of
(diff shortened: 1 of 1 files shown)


JohelEGP:
Closing. See the issue.

PR #4907, opened 2021-09-15
[module.import]/3 remove redundant importing of interface partitions

RedBeard0531:
https://eel.is/c++draft/module.import#7 already describes how importing a TU implicitly imports all TUs that are `export import`ed from the TU. https://eel.is/c++draft/module.unit#3.sentence-3 ensures that all interface partitions are transitively exported by the primary interface unit. Therefore, [module.import]/3 does not need to say that it imports "all module interface units of `M`" (unless the implication is that it also imports interface partitions that are not transitively exported by the PMI). I think it is simpler if [module.import]/3 only describes the "root" interface TU being imported with /7 handling the transitive importing of partitions. This also matches the wording used at https://eel.is/c++draft/module.unit#8 to describe the implicit `import M;` that occurs after `module M;`.

### DIFF
--- a/source/modules.tex
+++ b/source/modules.tex
@@ -434,7 +434,7 @@ after the import declaration.
 \pnum
 A \grammarterm{module-import-declaration} that specifies
 a \grammarterm{module-name} \tcode{M}
-imports all module interface units of \tcode{M}.
+imports the primary interface unit of \tcode{M}.
 
 \pnum
 A \grammarterm{module-import-declaration} that specifies


jensmaurer:
@zygoloid , what's your opinion here?

zygoloid:
I think what a reader of this section wants to know is that importing a module effectively imports all module interface units -- that's the intended semantic effect, and transitive re-export via `export import :foo` is more of a vehicle by which we get that effect. While I think it's sufficient and arguably more elegant to say that we import just the primary interface unit here, if we make this change I think we should add a note saying that as a consequence, all interface partitions are also imported.

PR #4911, opened 2021-09-16
[dcl.init.general] Mention scalar type explicitly for default-initialization

maggyero:
Like it is already the case for zero-initialization.


jwakely (commented):
This is not an improvement.

jensmaurer:
"If ... otherwise if ..." ladders should terminate with an "otherwise" so that we know we have all possibilities covered and we don't accidentally fall off the end.

This is going in the opposite direction.

PR #4912, opened 2021-09-16
[dcl.init.general] break p7 into three paragraphs

jwakely:
Is [dcl.init.general] p7 really supposed to be all one paragraph, with bullet numbers continuing between two separate lists, or should there be a `\pnum` after each list? Or just after the first list, giving two paragraphs?

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -4368,6 +4368,7 @@ Otherwise,
 no initialization is performed.
 \end{itemize}
 
+\pnum
 A class type \tcode{T} is \defn{const-default-constructible} if
 default-initialization of \tcode{T} would invoke
 a user-provided constructor of \tcode{T} (not inherited from a base class)


JohelEGP:
That looks better. The stable name should be [dcl.init.general].

jwakely:
> The stable name should be [dcl.init.general].

Oops, fixed, thanks!

jensmaurer:
I agree with the addition of the first \pnum, but the second one seems not an improvement.
We're directly using the term "const-default-constructible" that we've introduced just before, and I'm generally dubious of introducing single-sentence paragraphs unless the subject matter is clearly different. Here, it is not.

@jwakely , if you would be so kind to revert the second \pnum; I'd then apply this as an obvious editorial improvement on the typo level. @tkoeppe , any thoughts here?

jwakely:
Will do

jwakely:
Done!

jensmaurer (approved):

jensmaurer:
Thanks!

tkoeppe:
Looks good, thanks. I would probably also have got rid of the source blank lines around the itemizes, so that the source code visual structure reflects the \pnums, but we have a lot of historic ballast there.

PR #4913, opened 2021-09-16
[basic.lval] Replace the misused term ‘classifications’ by ‘categories’

maggyero:
A single classification/categorization is a set of classes/categories.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -158,7 +158,7 @@ or an expression that has type \cv{}~\keyword{void}.
 \end{itemize}
 
 \pnum
-Every expression belongs to exactly one of the fundamental classifications in this
+Every expression belongs to exactly one of the fundamental categories in this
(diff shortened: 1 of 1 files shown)


tkoeppe:
I'm tempted to agree that this is an improvement, but just out of an abundance of caution, could we get CWG's approval?

tkoeppe:
Side note: There is one more use of the word "classifications" in the text, in concepts.tex, but that is unrelated ("These concepts express relationships between types, type classifications, and fundamental type properties.")

jensmaurer (changes requested):
CWG 2023-06-13: Approved in general, with one change requested.

jensmaurer (approved):

geryogam:
Thanks for the review @tkoeppe and @jensmaurer!

tkoeppe:
You're most welcome, thank you for raising the issue :-)

PR #4920, opened 2021-09-19
[std] Replace \rSec with \subclause

### LINKED ISSUE #3352: [std] Replace \rSec with plain \Sec
Quote from macros.tex:

% A convenience feature (mostly for the convenience of the Project
% Editor, to make it easy to move around large blocks of text):
% the \rSec macro is just like the \Sec macro, except that depths
% relative to a global variable, SectionDepthBase. 

We're never using that "convenience", and it would actually be harmful, because surrounding \rSec headings would be confusing as to their actual nesting level. Let's remove this globally.

JohelEGP:
Resolves #3352.

### DIFF
--- a/.ctags
+++ b/.ctags
@@ -1,6 +1,6 @@
 --langdef=latex
 --langmap=latex:.tex
---regex-latex=/\\rSec[0-9](\[[a-zA-Z0-9_.:-]+\])/\1/s,section/
+--regex-latex=/\\subclause[0-9](\[[a-zA-Z0-9_.:-]+\])/\1/s,section/
 --exclude=*.html
 --exclude=*.cpp
 --exclude=Makefile
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -1,7 +1,7 @@
 %!TEX root = std.tex
-\rSec0[algorithms]{Algorithms library}
+\subclause0[algorithms]{Algorithms library}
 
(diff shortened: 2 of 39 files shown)


JohelEGP:
Should `\rSec` be retired by this PR? There are some diff macros. Those are not used by the standard, but are used to write proposals. In that sense, retiring it would be an API break.

tkoeppe:
Thanks for this suggestion. We would prefer to generate these changes ourselves; have you got a set of instructions or a script to generate them by any chance?

JohelEGP:
Patch `macros.tex`:
```
diff --git a/source/macros.tex b/source/macros.tex
index ca967458..650db4c3 100644
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -102,6 +102,8 @@
 \s[#3]{#3\hfill[#2]}\l{#2}\addxref{#2}%
 }
 
+\def\subclause#1[#2]#3{\Sec#1[#2]{#3}}
+
 % A convenience feature (mostly for the convenience of the Project
 % Editor, to make it easy to move around large blocks of text):
 % the \rSec macro is just like the \Sec macro, except that depths
```
Then run
```
sed -i 's/^\\rSec/\\subclause/' source/*.tex
sed -i 's/rSec/subclause/g' tools/* .ctags
```

tkoeppe:
Thanks!

PR #4928, opened 2021-09-23
[class.copy.elision] Fix comment in example

### LINKED ISSUE #4926: [class.copy.elision] p2 incorrect comment in copy elision example
Paragraph 2 of [class.copy.elision] (https://eel.is/c++draft/class.copy.elision#2) contains the following example:

```
[...]

constexpr A g() {
  A loc;
  return loc;
}

constexpr A a;          // well-formed, a.p points to a
constexpr A b = g();    // error: b.p would be dangling ([expr.const])

void h() {
  A c = g();            // well-formed, c.p may point to c or to an ephemeral temporary
}
```

I believe the comment "well-formed, c.p may point to c or to an ephemeral temporary" is not entirely correct, because there is no temporary in this scenario.

This comment was originally added with the resolution of [CWG2002](http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#2022). The example used in that resolution was:
```
[...]

struct A {
    void *p;
    constexpr A(): p(this) {}
  };

  constexpr A a;        // well-formed, a.p points to a
  constexpr A b = A();  // well-formed, b.p points to b

  void g() {
    A c = A();          // well-formed, c.p may point to c or to an ephemeral temporary
  }
```
In this resolution the comment "well-formed, c.p may point to c or to an ephemeral temporary" correctly reflects the code.

After the resolution of [CWG2002](http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#2022) the code in the example was modified in [95321805a3](https://github.com/cplusplus/draft/commit/95321805a3528aa1006bf8187e39e1f4bab00c1e), but the comment was not changed to reflect this.

jensmaurer:
Fixes #4926

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -6162,7 +6162,7 @@ constexpr A a;          // well-formed, \tcode{a.p} points to \tcode{a}
 constexpr A b = g();    // error: \tcode{b.p} would be dangling\iref{expr.const}
 
 void h() {
-  A c = g();            // well-formed, \tcode{c.p} may point to \tcode{c} or to an ephemeral temporary
+  A c = g();            // well-formed, \tcode{c.p} can point to \tcode{c} or be dangling
 }
 \end{codeblock}
 Here the criteria for elision can eliminate


tkoeppe:
@zygoloid: this looks good as is, but there are interesting details here about _how_ the dangling can happen (either because a variable `loc` is not elided, or because the "trivial type may be copied anyway" rule is used), which sounds interesting enough that we may want to spell it out. Thoughts?

zygoloid (approved):

zygoloid:
While I agree that the two different ways that the pointer can dangle are (abstractly) interesting, I don't think they're what this example is trying to point out and would probably be more of a distraction than a help here.

PR #4933, opened 2021-09-24
[fstream.syn] Fix grammar

CaseyCarter:
"are only be provided" is not a valid english verb phrase. I'm not sure if the intent was "are only provided" or "are only to be provided", but I think the first more closely adheres to our style of making existential statements about library implementations.


jwakely (approved):

jwakely:
That text came from [LWG 2676](https://wg21.link/lwg2676) which said "shall only be" but when I applied it to the draft in a98ab99f5158ea214a31e3623bdecd3575bb75a9 I "fixed" it to not use shall.

PR #4935, opened 2021-09-25
[headers] Fix terminology in note

JohelEGP:
Importable C++ library headers can be imported as header units.
This currently says "as module units".
[module.unit]#1 says that
"A _module unit_ is a translation unit that contains a _module-declaration_."
[module.import]#5 says that the synthetized header unit from an `import <H>;`
is a translation unit that contains no _module-declaration_.


JohelEGP:
The wording is wonky. You don't import a header unit. Can this be reworded in a way that is correct while remaining light in terminology? Maybe just drop the "as whatever unit" and let the example show the way.

jensmaurer:
Yes, just saying "... can be imported" sounds plausible.

jensmaurer (approved):
Looks good.

The commit description has a typo, and I think something shorter would work, too:

> [headers] Fix note about importing library headers

Importable library headers are not module units, because they
do not have a module-declaration."

JohelEGP:
That is much better, especially the commit header message. Thank you.

tkoeppe:
Looks right -- but out of curiosity, how did this wording get in in the first place? By which I mean, are we really restoring the original intent here? If so, feel free to merge.

tkoeppe (approved):

jensmaurer:
@tkoeppe , it's a note, so it's fair game for editorial fixes, I believe. Further, I think this just wants to say that you can import something where the module-name looks like a standard library header name.

"git blame" would help you with the archeology; I'm pretty sure it went in with the original modules paper.

RedBeard0531:
> You don't import a header unit. 

@JohelEGP what do you mean? Isn't that exactly what https://eel.is/c++draft/module.import#5 says you do?

I don't object to not saying header unit here if there are objections to saying it, but personally I think it is both helpful and correct.

JohelEGP:
As quoted in that link, the implementation synthesizes a _header unit_ from `import <H>;` or `import "H";`. You can't specify a header unit in a _module-import-declaration_, but you _can_ specify a _header-name_, from which a header unit will be synthesized.

RedBeard0531:
This is splitting very fine linguistic hairs, but I'd still view that as "importing the header *as a*  header unit". That doesn't mean the header *is a* header unit, or just that the *import* uses the header unit mechanism.

JohelEGP:
I agree with you.

That kind of wording is used elsewhere in the standard: https://eel.is/c++draft/intro#compliance.general-5, https://eel.is/c++draft/cpp.import (the title).

It is true that "you don't import a header unit". It's also right to say that "importable C++ library headers can be imported as header units".

If it's OK, I'll revert to the line change of the original patch (keeping the improved commit message).

PR #4941, opened 2021-09-26
[std] Fix cross-references to 'unevaluated operand'

jensmaurer:


jensmaurer:
@tkoeppe, this only fixes and adds cross-references for "unevaluated operand", as a first step. It should be pretty obvious.

There is some follow-on work, because we sometimes mean "not potentially evaluated" (which also covers subexpressions). The latter needs separate work.

PR #4942, opened 2021-09-27
[lib] Rename exposition-only 'no-throw-' concepts

### LINKED ISSUE #4101: [special.mem.concepts] Remove hyphen in _no-throw_ for consistency with the type traits
See https://github.com/cplusplus/draft/issues/4050#issuecomment-647085151.

jensmaurer:
to 'nothrow-', for consistency with nothrow_constructible

Fixes #4101

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -6919,15 +6919,15 @@ namespace std {
   // \ref{specialized.algorithms}, specialized algorithms
   // \ref{special.mem.concepts}, special memory concepts
   template<class I>
-    concept @\exposconcept{no-throw-input-iterator}@ = @\seebelow@;    // \expos
+    concept @\exposconcept{nothrow-input-iterator}@ = @\seebelow@;    // \expos
   template<class I>
-    concept @\exposconcept{no-throw-forward-iterator}@ = @\seebelow@;  // \expos
+    concept @\exposconcept{nothrow-forward-iterator}@ = @\seebelow@;  // \expos
   template<class S, class I>
-    concept @\exposconcept{no-throw-sentinel-for}@ = @\seebelow@;      // \expos
+    concept @\exposconcept{nothrow-sentinel-for}@ = @\seebelow@;      // \expos
   template<class R>
-    concept @\exposconcept{no-throw-input-range}@ = @\seebelow@;       // \expos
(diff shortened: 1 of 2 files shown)


jwakely (approved):

source/utilities.tex
```diff
@@ -6919,15 +6919,15 @@
...
+    concept @\exposconcept{nothrow-forward-iterator}@ = @\seebelow@;  // \expos
   template<class S, class I>
-    concept @\exposconcept{no-throw-sentinel-for}@ = @\seebelow@;      // \expos
+    concept @\exposconcept{nothrow-sentinel-for}@ = @\seebelow@;      // \expos
   template<class R>
-    concept @\exposconcept{no-throw-input-range}@ = @\seebelow@;       // \expos
+    concept @\exposconcept{nothrow-input-range}@ = @\seebelow@;       // \expos
   template<class R>
-    concept @\exposconcept{no-throw-forward-range}@ = @\seebelow@;     // \expos
+    concept @\exposconcept{nothrow-forward-range}@ = @\seebelow@;     // \expos
```
JohelEGP, line 6930: It seems strange that you didn't have to readjust the comment to be aligned at a multiple of 4.

jensmaurer: The checking machinery ignores lines with an @ in them, because they disconnect the source file indentation from the rendered indentation anyway.

JohelEGP: Aren't the `% macro length:` comments in `macros.tex` for that?

jensmaurer: Those came afterwards and are not reflected in the checking machinery.

jensmaurer: If turns out that the alignment is actually locally consistent (but maybe not a factor-of-4), because all exposition-only concepts in the vicinity are of the nothrow-x kind. I'll not touch this for now until we actually get proper checking of @ widths.

tkoeppe: That's fine. Ultimately, the goal of the guidance is set us on an overall path of steady improvement, but there are still many parts of the text that don't follow it (or worse, contain italics), and it's not the end of the world. Thanks for spotting this, but also don't sweat it too much.

PR #4953, opened 2021-09-30
[basic.life,basic.start.term,expr.delete] Clarify destruction of scalars

### LINKED ISSUE #4944: Update the example of temporary pseudo-destruction to undefined behavior
Since Richard Smith’s proposal [P0593R6](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p0593r6.html#pseudo-destructor-calls), a pseudo-destructor call ends the lifetime of a scalar object:

> 3.6. Pseudo-destructor calls
> 
> In the current C++ language rules, "pseudo-destructor" calls may be used in generic code to allow such code to be ambivalent as to whether an object is of class type:
> 
> ```c++
> template<typename T> void destroy(T *p) { p->~T(); }
> ```
> 
> When T is, say, int, the pseudo-destructor expression p->~T() is specified as having no effect. We believe this is an error: such an expression should have a lifetime effect, ending the lifetime of the int object. Likewise, calling a destructor of a class object should always end the lifetime of that object, regardless of whether the destructor is trivial.
> 
> This change improves the ability of static and dynamic analysis tools to reason about the lifetimes of C++ objects.

So isn’t the line `0 .T::~T();` given in [[expr.prim.id.dtor]/3](https://timsong-cpp.github.io/cppwp/expr.prim.id.dtor#3) undefined behaviour now that it destroys the temporary object 0 twice?

> [*Example 1*:
> 
> ```c++
> struct C { };
> void f() {
>   C * pc = new C;
>   using C2 = C;
>   pc->C::~C2();     // OK, destroys *pc
>   C().C::~C();      // undefined behavior: temporary of type C destroyed twice
>   using T = int;
>   0 .T::~T();       // OK, no effect
>   0.T::~T();        // error: 0.T is a user-defined-floating-point-literal ([lex.ext])
> }
> ```
> 
> — *end example*]

jensmaurer:
http://lists.isocpp.org/core/2021/01/10392.php
http://lists.isocpp.org/core/2021/09/11592.php

Fixes #4944

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3383,8 +3383,8 @@ limited ways, as described below. The
 program has undefined behavior if:
 \begin{itemize}
 \item
-  the object will be or was of a class type with a non-trivial destructor
-  and the pointer is used as the operand of a \grammarterm{delete-expression},
+  the pointer is used to destroy the object
+  (\ref{expr.prim.id.dtor}, \ref{expr.delete})
 \item
   the pointer is used to access a non-static data member or call a
   non-static member function of the object, or
@@ -3447,6 +3447,7 @@ properties of the glvalue that do not depend on its value is
 well-defined. The program has undefined behavior if:
 \begin{itemize}
 \item the glvalue is used to access the object, or
+\item the glvalue is used to destroy the object\iref{expr.prim.id.dtor}, or
 \item the glvalue is used to call a non-static member function of the object, or
 \item the glvalue is bound to a reference to a virtual base class\iref{dcl.init.ref}, or
 \item the glvalue is used as the operand of a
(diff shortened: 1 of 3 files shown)


jensmaurer:
Put onto core's inbox: https://wiki.edg.com/bin/view/Wg21telecons2021/CoreWorkingGroup

languagelawyer:
So you're basically fine with https://lists.isocpp.org/core/2021/01/10409.php (meant to use `unsigned char` to provide storage)

jensmaurer:
I guess we need a special rule around "provides storage"; the examples there all suggest that if an object provides storage, using that storage to create another object does not end the lifetime of the object that provided the storage.

jensmaurer:
Ah, we have that rule: [basic.life] p1.5 says the example is fine, because the out array provides storage for the "int", and the lifetime of the outer array was not ended by creating the "int".

languagelawyer:
But the lifetime of the array elements end. And they are destroyed when the array object is destroyed.

jensmaurer:
So maybe as a special exception, the "provides storage" types (unsigned char, std::byte) shouldn't actually have lifetime-ending destruction.

languagelawyer:
Can the form of the exception (either reusing their storage doesn't end their lifetime or destroying dead objects of few types is OK) be decided editorially?

BTW, [P1839 Accessing Object Representations](http://wg21.link/p1839) says that the elements of an object representation are alive as long as the object is alive and that arrays of characters are «their own representations», so it looks like it keeps the elements of arrays providing storage alive when the new objects are created on top of them. Not sure if this was intended.

jensmaurer:
> Can the form of the exception [...] be decided editorially?

There is nothing editorial here. Even the intention for the original example in #4944 is unclear. Note the "CWG" label liberally sprinkled about.

languagelawyer:
Does the "CWG" label guarantees it is not editorial? I've thought there is "not-editorial" label for this and "CWG" means that CWG wants to take a look at this before fixing editorially.

jensmaurer:
Don't worry, CWG will tell if it's not editorial.

PR #4957, opened 2021-10-02
[range.join.view] remove duplicative condition in end() const

timsong-cpp:
This function is constrained on `is_reference_v<range_reference_t<const V>>` already so checking it again in the `if constexpr` has no effect.


jensmaurer (approved):
Looks good to me.

@jwakely , @CaseyCarter , a second opinion, please.

CaseyCarter (approved):

PR #4959, opened 2021-10-02
[stmt.pre] Remove erroneous crossreference

JohelEGP:


JohelEGP:
Since [dcl.dcl] is Clause 9 Declarations,
perhaps the crossreference was meant for the following item.

jensmaurer (approved):
I'm not sure what happened here. The status quo looks strange, though.

PR #5003, opened 2021-10-05
P2036R3 Change scope of lambda trailing-return-type

### LINKED ISSUE #4969: [2021-10 CWG Motion 8] P2036R3 Change scope of lambda trailing-return-type
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2036r3.html

jensmaurer:
Fixes #4969
Fixes cplusplus/papers#766

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -1443,21 +1443,34 @@ the component name of that construct that appears lexically last.
 \pnum
 The result is the entity denoted by
 the \grammarterm{unqualified-id}\iref{basic.lookup.unqual}.
-If the entity is a local entity\iref{basic.pre}
-and naming it from outside of an unevaluated operand
-within the scope where the \grammarterm{unqualified-id} appears
-would result in some intervening \grammarterm{lambda-expression}
-capturing it by copy\iref{expr.prim.lambda.capture},
+If the \grammarterm{unqualified-id} appears
+in a \grammarterm{lambda-expression} at program point $P$ and
+the entity is a local entity\iref{basic.pre} or a variable declared by
+an \grammarterm{init-capture}\iref{expr.prim.lambda.capture},
+then let $S$ be the \grammarterm{compound-expression} of
+the innermost enclosing \grammarterm{lambda-expression} of $P$.
+If naming the entity from outside of an unevaluated operand within $S$
+would refer to an entity
+captured by copy in some intervening \grammarterm{lambda-expression},
+then let $E$ be the innermost such \grammarterm{lambda-expression}, and:
+\begin{itemize}
+\item
+If $P$ is in $E$'s function parameter scope
+but not its \grammarterm{parameter-declaration-clause}, then
 the type of the expression is
 the type of a class member access expression\iref{expr.ref}
 naming the non-static data member
 that would be declared for such a capture
-in the object parameter\iref{dcl.fct} of the function call operator of
-the innermost such intervening \grammarterm{lambda-expression}.
+in the object parameter\iref{dcl.fct} of the function call operator of $E$.
 \begin{note}
-If that \grammarterm{lambda-expression} is not declared \keyword{mutable},
+If $E$ is not declared \keyword{mutable},
 the type of such an identifier will typically be \keyword{const} qualified.
 \end{note}
+\item
+Otherwise (if $P$ either precedes $E$'s function parameter scope or
+is in $E$'s \grammarterm{parameter-declaration-clause}),
+the program is ill-formed.
+\end{itemize}
 Otherwise, the type of the expression is the type of the result.
 \begin{note}
 If the entity is a template parameter object for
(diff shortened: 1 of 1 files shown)


source/expressions.tex
```diff
@@ -1475,12 +1488,24 @@
 \begin{codeblock}
 void f() {
   float x, &r = x;
-  [=] {
+  [=]() -> decltype((x)) {      // lambda returns \tcode{float const\&} because this lambda is not \tcode{mutable} and
```
JohelEGP, line 1491:
```suggestion
  [=]() -> decltype((x)) {      // lambda returns \tcode{\keyword{float} \keyword{const}\&} because this lambda is not \keyword{mutable} and
```

source/expressions.tex
```diff
@@ -1475,12 +1488,24 @@
...
+                                // \tcode{x} is an lvalue
     decltype(x) y1;             // \tcode{y1} has type \tcode{float}
-    decltype((x)) y2 = y1;      // \tcode{y2} has type \tcode{float const\&} because this lambda
-                                // is not \keyword{mutable} and \tcode{x} is an lvalue
+    decltype((x)) y2 = y1;      // \tcode{y2} has type \tcode{float const\&}
```
JohelEGP, line 1494:
```suggestion
    decltype((x)) y2 = y1;      // \tcode{y2} has type \tcode{\keyword{float} \keyword{const}\&}
```

source/expressions.tex
```diff
@@ -1475,12 +1488,24 @@
...
     decltype(r) r1 = y1;        // \tcode{r1} has type \tcode{float\&}
     decltype((r)) r2 = y2;      // \tcode{r2} has type \tcode{float const\&}
+    return y2;
+  };
+
+  [=]<decltype(x) P>{};         // error: \tcode{x} refers to local entity but precedes the
+                                // lambda's function parameter scope
+  [=](decltype((x)) y){};       // error: \tcode{x} refers to local entity but is in lambda's
```
JohelEGP, line 1502:
```suggestion
  [=](decltype((x)) y){};       // error: \tcode{x} refers to local entity but is in the lambda's
```

source/expressions.tex
```diff
@@ -1475,12 +1488,24 @@
...
+                                // \grammarterm{parameter-declaration-clause}
+  [=]{
+      []<decltype(x) P>{};      // OK, \tcode{x}is in the outer lambda's function parameter scope
+      [](decltype((x)) y){};    // OK, lambda takes a parameter of type \tcode{float const\&}
+      [x=1](decltype((x)) z){}; // error: \tcode{x} refers to \grammarterm{init-capture} but is in lambda's
```
JohelEGP, line 1507:
```suggestion
      []<decltype(x) P>{};      // OK: \tcode{x} is in the outer lambda's function parameter scope
      [](decltype((x)) y){};    // OK: lambda takes a parameter of type \tcode{\keyword{float} \keyword{const}\&}
      [x=1](decltype((x)) z){}; // error: \tcode{x} refers to \grammarterm{init-capture} but is in the lambda's
```

source/expressions.tex
```diff
@@ -2154,8 +2179,8 @@
 \end{example}
 
 \pnum
-An \grammarterm{init-capture} inhabits the scope of
-the \grammarterm{lambda-expression}'s \grammarterm{compound-statement}.
+An \grammarterm{init-capture} inhabits the functino parameter scope of
```
JohelEGP, line 2182:
```suggestion
An \grammarterm{init-capture} inhabits the function parameter scope of
```

source/expressions.tex
```diff
@@ -2181,6 +2206,9 @@
          }();                               // Updates \tcode{::x} to 6, and initializes \tcode{y} to 7.
 
 auto z = [a = 42](int a) { return 1; };     // error: parameter and local variable have the same name
+auto counter = [i=0]() mutable -> decltype(i) {     // OK, returns \tcode{int}
```
JohelEGP, line 2209:
```suggestion
auto counter = [i=0]() mutable -> decltype(i) {     // OK: returns \keyword{int}
```

JohelEGP (changes requested):

source/expressions.tex
jensmaurer: No; comma is consensus.

JohelEGP:
My understanding is that, since `\keyword` adds index entries, it should only be used when the meaning of the keyword is defined. That is unlike `\libconcept`, which is meant to index all uses of concepts.

In that case, changing all "_predicate_ is \tcode{true}" to use `\keyword` would have the undesirable effect of adding many index entries.

source/expressions.tex
```diff
@@ -1443,21 +1443,34 @@
...
+\item
+If $P$ is in $E$'s function parameter scope
+but not its \grammarterm{parameter-declaration-clause}, then
 the type of the expression is
 the type of a class member access expression\iref{expr.ref}
 naming the non-static data member
 that would be declared for such a capture
-in the object parameter\iref{dcl.fct} of the function call operator of
-the innermost such intervening \grammarterm{lambda-expression}.
+in the object parameter\iref{dcl.fct} of the function call operator of $E$.
```
tkoeppe, line 1464: @BRevzin: The base of the diff has changed here following the "Deducing `this`" paper; could you kindly check if the new edit is still appropriate?

BRevzin: Yeah this looks right. The point was just to name the innermost such intervening lambda `E` 👍

source/expressions.tex
```diff
@@ -1477,12 +1490,24 @@
...
     decltype((r)) r2 = y2;      // \tcode{r2} has type \tcode{float const\&}
+    return y2;
+  };
+
+  [=]<decltype(x) P>{};         // error: \tcode{x} refers to local entity but precedes the
+                                // lambda's function parameter scope
+  [=](decltype((x)) y){};       // error: \tcode{x} refers to local entity but is in the lambda's
+                                // \grammarterm{parameter-declaration-clause}
+  [=]{
+      []<decltype(x) P>{};      // OK: \tcode{x} is in the outer lambda's function parameter scope
```
tkoeppe, line 1507: @jensmaurer: should we use "OK, " here, too?

source/expressions.tex
tkoeppe: Thanks a lot!

tkoeppe (approved):

source/expressions.tex
jensmaurer: Yes. This probably needs global cleanup.
```suggestion
      []<decltype(x) P>{};      // OK, \tcode{x} is in the outer lambda's function parameter scope
```

PR #5005, opened 2021-10-06
P2372R3 Fixing locale handling in chrono formatters

### LINKED ISSUE #4973: [2021-10 LWG Motion 2] P2372R3 Fixing locale handling in chrono formatters
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2372r3.html

jensmaurer:
Fixes #4973
Fixes cplusplus/papers#1039


source/time.tex
```diff
@@ -2703,9 +2703,7 @@
 \effects
 Equivalent to:
 \begin{codeblock}
-auto const dp = floor<days>(tp);
-return os << format(os.getloc(), @\placeholder{STATICALLY-WIDEN}@<charT>("{} {}"),
-                    year_month_day{dp}, hh_mm_ss{tp-dp});
+return os << format(os.getloc(), STATICALLY-WIDEN("{:L%F %T}"), tp);
```
JohelEGP, line 2706:
```suggestion
return os << format(os.getloc(), @\placeholder{STATICALLY-WIDEN}@("{:L%F %T}"), tp);
```

source/time.tex
```diff
@@ -8609,7 +8607,7 @@
 \effects
 Equivalent to:
 \begin{codeblock}
-return os << format(os.getloc(), @\placeholder{STATICALLY-WIDEN}@<charT>("{:%T}"), hms);
+return os << format(os.getloc(), @\placeholder{STATICALLY-WIDEN}@<charT>("{::%T}"), hms);
```
JohelEGP:
```suggestion
return os << format(os.getloc(), @\placeholder{STATICALLY-WIDEN}@<charT>("{:L%T}"), hms);
```

source/time.tex
```diff
@@ -10086,14 +10084,10 @@
...
-the value returned from \tcode{t.get_local_time()}
-to \tcode{os}
-using the format \tcode{"\%F \%T \%Z"}.
-
-\pnum
-\returns
-\tcode{os}.
+Equivalent to:
+\begin{codeblock}
+return os << format(os.getloc(), STATICALLY-WIDEN("{:L%F %T %Z}"), t);
```
JohelEGP, line 10089:
```suggestion
return os << format(os.getloc(), @\placeholder{STATICALLY-WIDEN}@("{:L%F %T %Z}"), t);
```

source/time.tex
```diff
@@ -10481,15 +10475,26 @@
 represented by \fmtgrammarterm{literal-char}
 are copied unchanged to the output.
 
+\pnum
+A \defnadj{formatting}{locale} is an instance of \tcode{locale}
+used by a formatting function, that is determined as follows:
+\begin{itemize}
+\item
+the "C" locale if the \tcode{L} option
+is not present in \grammarterm{chrono-format-spec}, otherwise
```
JohelEGP, line 10484:
```suggestion
is not present in \fmtgrammarterm{chrono-format-spec}, otherwise
```

source/time.tex
```diff
@@ -10481,15 +10475,26 @@
 represented by \fmtgrammarterm{literal-char}
 are copied unchanged to the output.
 
+\pnum
+A \defnadj{formatting}{locale} is an instance of \tcode{locale}
+used by a formatting function, that is determined as follows:
+\begin{itemize}
+\item
+the "C" locale if the \tcode{L} option
```
timsong-cpp, line 10483: It looks like [localization] always uses `\tcode{"C"}` for this.
```suggestion
the \tcode{"C"} locale if the \tcode{L} option
```

source/time.tex
```diff
@@ -10481,15 +10475,26 @@
 represented by \fmtgrammarterm{literal-char}
 are copied unchanged to the output.
 
+\pnum
+A \defnadj{formatting}{locale} is an instance of \tcode{locale}
+used by a formatting function, that is determined as follows:
```
burblebee, line 10480: Problem with paper: This doesn't lead into the bullets that follow (we've ended a sentence here yet the bullets that follow are incomplete by themselves).
"A /formatting locale/ FL is an [...] that is determined as follows. FL is:"
Optionally, instead of "FL is:", each bullet could start a sentence with "FL is".
Also, should the bullets begin with otherwise?  Or end with "or"? 

I suggest the following:
A /formatting locale/ FL is an instance of locale used by a formatting function F, determined as follows.
* If the L option is not present in /chrono-format-spec/, FL is the "C" locale.
* Otherwise, if a locale was passed to F, FL is that locale.
* Otherwise,  FL is the global locale.

Please consider applying my suggested change as a separate commit, else let's open an edit issue.

source/time.tex
```diff
@@ -10764,6 +10769,7 @@
 If the \fmtgrammarterm{chrono-specs} is omitted,
 the chrono object is formatted
 as if by streaming it to \tcode{std::ostringstream os}
+with the formatting locale imbued
```
burblebee, line 10772: Problem with paper/spec: What does it mean for a locale to be "imbued"? Same for uses of "imbued" elsewhere. Should we open an issue for this?

burblebee (changes requested):
Other than 2 issues I had with the paper, just the "C" needs to be \tcoded (as already mentioned by timsong-cpp).

burblebee (approved):
"C" has been fixed, but issues with wording remain.

source/time.tex
jensmaurer: iostreams have a member function "imbue" that attaches a locale to the stream. That's what's discussed here.

source/time.tex
tkoeppe: I would just delete "determined as follows:". What do you think? @burblebee, @jensmaurer? (To be done as a later commit.)

jensmaurer: @burblebee , please create a separate editorial issue for this one.
I'll need to check more of the surroundings to form an opinion.

PR #5006, opened 2021-10-06
[macros] Preserve straight apostrophes in code.

tkoeppe:
Apostrophes are code and should not be turned into typographic single quotes.

Please let me know if there's an appetite for this change. I think it's altogether fitting to have codeblocks show actual code, but I'd like to hear other opinions.

The effect is pervasive, here's a random example screenshot:

image


jwakely (approved):
I prefer this

PR #5008, opened 2021-10-06
[utility.intcmp] Remove Mandates implied by Equivalent to

### LINKED ISSUE #5007: [utility.intcmp].9 is redundant
https://wg21.link/utility.intcmp#9 is implied by the "Equivalent to:" semantics.

JohelEGP:
Resolves #5007.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -546,11 +546,6 @@ template<class R, class T>
 
 \begin{itemdescr}
 \pnum
-\mandates
-Both \tcode{T} and \tcode{R} are standard integer types or
-extended integer types\iref{basic.fundamental}.
-
-\pnum
 \effects
 Equivalent to:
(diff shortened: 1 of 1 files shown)


timsong-cpp:
This was discussed in Prague. LWG wasn't satisfied that `numeric_limits<R>::min()` is guaranteed to be of type `R`, given that it is entirely unclear what the requirements on user specializations are.

JohelEGP:
Doesn't https://eel.is/c++draft/library#contents-3 prevent ADL? Or is it to allow some two-step dance as with `std::swap`?

timsong-cpp:
The point is that we want to ban `in_range<Meow>(42)` even if `numeric_limits<Meow>::min()` is an `int`.

JohelEGP:
Makes sense. Thank you.

PR #5011, opened 2021-10-07
P2418R2 Add support for std::generator-like types to std::format

### LINKED ISSUE #4975: [2021-10 LWG Motion 4] P2418R2 Add support for std::generator -like types to std::format
https://wiki.edg.com/pub/Wg21virtual2021-10/StrawPolls/P2418R2.html

jensmaurer:
Fixes #4975
Fixes cplusplus/papers#1092

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -61,6 +61,24 @@ auto s = std::format("{:d}", "I am not a number");      // ill-formed,
                                                         // previously threw \tcode{format_error}
 \end{codeblock}
 
+\diffref{format}
+\change
+Signature changes: \tcode{format}, \tcode{format_to}, \tcode{format_to_n},
+\tcode{formatted_size}.
+\rationale
+Enable formatting of views that are neither const-iterable nor copyable.
+\effect
+Valid \CppXX{} code that passed bit fields to formatting functions
+may become ill-formed. For example:
+\begin{codeblock}
+struct tiny {
+  int bit: 1;
+};
+
+auto t = tiny();
+std::format("{}", t.bit);       // ill-formed, previously returned "0"
+\end{codeblock}
+
 \rSec1[diff.cpp17]{\Cpp{} and ISO \CppXVII{}}
 
 \rSec2[diff.cpp17.general]{General}
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -19651,13 +19651,13 @@ namespace std {
 
   // \ref{format.functions}, formatting functions
   template<class... Args>
-    string format(@\exposid{format-string}@<Args...> fmt, const Args&... args);
+    string format(@\exposid{format-string}@<Args...> fmt, Args&&... args);
   template<class... Args>
-    wstring format(@\exposid{wformat-string}@<Args...> fmt, const Args&... args);
+    wstring format(@\exposid{wformat-string}@<Args...> fmt, Args&&... args);
   template<class... Args>
-    string format(const locale& loc, @\exposid{format-string}@<Args...> fmt, const Args&... args);
+    string format(const locale& loc, @\exposid{format-string}@<Args...> fmt, Args&&... args);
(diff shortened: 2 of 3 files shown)


source/compatibility.tex
```diff
@@ -61,6 +61,24 @@
                                                         // previously threw \tcode{format_error}
 \end{codeblock}
 
+\diffref{format}
+\change
+Signature changes: \tcode{format}, \tcode{format_to}, \tcode{format_to_n},
+\tcode{formatted_size}.
+\rationale
+Enable formatting of views that are neither const-iterable nor copyable.
```
burblebee, line 69: What is "const-iterable"? the term is not defined afaikt.

source/compatibility.tex
```diff
@@ -61,6 +61,24 @@
...
+\effect
+Valid \CppXX{} code that passed bit fields to formatting functions
+may become ill-formed. For example:
+\begin{codeblock}
+struct tiny {
+  int bit: 1;
+};
+
+auto t = tiny();
+std::format("{}", t.bit);       // ill-formed, previously returned "0"
```
burblebee, line 79: "0" should be \tcoded

source/compatibility.tex
```diff
@@ -61,6 +61,24 @@
...
 \end{codeblock}
 
+\diffref{format}
+\change
+Signature changes: \tcode{format}, \tcode{format_to}, \tcode{format_to_n},
+\tcode{formatted_size}.
+\rationale
+Enable formatting of views that are neither const-iterable nor copyable.
+\effect
+Valid \CppXX{} code that passed bit fields to formatting functions
```
burblebee, line 71: "passed" -> "passes"

burblebee (changes requested):

source/compatibility.tex
jensmaurer: Agreed, but this is non-normative, so maybe we can get away with it.

@timsong-cpp , anything we can do about "const-iterable" editorially here?

source/utilities.tex
```diff
@@ -20829,6 +20835,24 @@
 Stores the parsed format specifiers in \tcode{*this} and
 returns an iterator past the end of the parsed range.
 \\ \rowsep
+\tcode{f.format(u, fc)} &
+\tcode{FC::iterator} &
+Formats \tcode{u} according to the specifiers stored in \tcode{*this},
+writes the output to \tcode{fc.out()} and
```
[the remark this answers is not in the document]
jensmaurer: Fixed both; the one in the existing text with a separate commit.

source/compatibility.tex
jensmaurer: fixed

source/compatibility.tex
timsong-cpp: "only models `range` when non-const"? "can only be iterated over if they are not const"? "does not support iteration when const-qualified"? "only have non-const `begin`/`end` functions"?

jensmaurer: Maybe "Enable formatting of views that do not support iteration when const-qualified and are not copyable."
@burblebee , is that better?

burblebee: Sounds good to me!

jwakely: Doesn't seem like an improvement to me :disappointed: 

Is the "and are not copyable" clause a modifier to the first condition? i.e. do not support iteration when (const-qualified && !copyable) ? It's not supposed to be, but it can be read that way. "are not copyable and do not support ..." would solve that problem.

tkoeppe: I'm not sure how you can read it that way. Here's the current wording, for reference:

> Enable formatting of views that do not support iteration when const-qualified and are not copyable.

The second predicate "and are not copyable" has no subject, so it shares the subject with whatever comes before the "and"; the only matching subject here is "views". How would you read it to have any other subject?

PR #5017, opened 2021-10-09
P1147R1 Printing volatile Pointers

### LINKED ISSUE #4981: [2021-10 LWG Motion 10] P1147R1 Printing volatile Pointers
https://wiki.edg.com/pub/Wg21virtual2021-10/StrawPolls/P1147R1.html

jensmaurer:
Moved to [ostream.inserters.arithmetic] and renamed the parameter
from 'val' to 'p' for consistency with the 'const void*' overload.

Fixes #4981
Fixes cplusplus/papers#320

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -5854,6 +5854,7 @@ namespace std {
     basic_ostream& operator<<(long double f);
 
     basic_ostream& operator<<(const void* p);
+    basic_ostream& operator<<(const volatile void* p);
     basic_ostream& operator<<(nullptr_t);
     basic_ostream& operator<<(basic_streambuf<char_type, traits>* sb);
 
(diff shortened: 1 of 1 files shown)


source/iostreams.tex
```diff
@@ -6419,6 +6420,17 @@
 \tcode{*this}.
 \end{itemdescr}
 
+\indexlibrarymember{operator<<}{basic_ostream}%
+\begin{itemdecl}
+basic_ostream& operator<<(const volatile void* p);
```
burblebee, line 6425: Shouldn't we use "val" instead of "p" and combine this with the previous itemdecl, similar to how the non-volitile "val" version was?

burblebee (commented):
Let's combine \itemdecls; looks odd having the new member all by itself.  Otherwise lgtm.

source/iostreams.tex
tkoeppe: It's a bit awkward that the current `const void*` overload uses `p` in the header synopsis but `val` in the description, but regardless, using `p` in both places for the new overload seems good.

PR #5015, opened 2021-10-08
P0849R8 auto(x): decay-copy in the language

jensmaurer:
Fixes #4979
Fixes cplusplus/papers#293

PR #5039, opened 2021-10-19
[expr.type.conv] Move description of 'auto' deduction to [dcl.type.auto.deduct]

### LINKED ISSUE #4745: Awkward invented declaration in P0849
For @jensmaurer: the version approved by CWG (pending review by LWG) contains in [expr.type.conv]/1

> Otherwise, if the type is `auto`, it is replaced by the type deduced for the variable `x` in the invented declaration ([dcl.spec.auto]), which is never interpreted as a function declaration:
> ```
> auto x init;
> ```
> , where `init` is the initializer.

The appositive with the intervening relative clause and colon is a bit broken, as is the trailing comma.  My preferred means of addressing this is to replace the whole sentence with

> Otherwise, if the type is `auto`, it is replaced by the type determined by placeholder type deduction ([dcl.spec.auto.deduct]).

and add a bullet to [dcl.spec.auto.deduct]/2:

> * for an explicit type conversion ([expr.type.conv]), `T` is `auto` and `E` is the initializer.

Using `auto` explicitly rather than referring to "the specified type" avoids an ordering question with the replacement of that `auto` with the deduced type.

tkoeppe:
Fixes #4745.

See #5015 for discussion and suggestions.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -1950,6 +1950,20 @@ a single \grammarterm{assignment-expression}
 and $E$ is the \grammarterm{assignment-expression}.
 \end{itemize}
 \item
+For an explicit type conversion\iref{expr.type.conv},
+\tcode{T} is the specified type, which shall be \keyword{auto}.
+\begin{itemize}
+\item
+If the initializer is a \grammarterm{braced-init-list},
+it shall consist of a single brace-enclosed \grammarterm{assignment-expression}
+and $E$ is the \grammarterm{assignment-expression}.
+\item
+If the initializer is a parenthesized \grammarterm{expression-list},
+the \grammarterm{expression-list} shall be
+a single \grammarterm{assignment-expression}
+and $E$ is the \grammarterm{assignment-expression}.
+\end{itemize}
+\item
 For a non-type template parameter declared with a type
 that contains a placeholder type,
 \tcode{T} is the declared type of the non-type template parameter
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -3353,14 +3353,9 @@ it is replaced by the return type
 of the function selected by overload resolution
 for class template deduction\iref{over.match.class.deduct}
 for the remainder of this subclause.
-Otherwise, if the type is \tcode{auto},
-it is replaced by the type deduced for the variable \tcode{x}
-in the invented declaration\iref{dcl.spec.auto},
-which is never interpreted as a function declaration:
-\begin{codeblock}
-auto x @\placeholder{init}@;
-\end{codeblock}
-, where \placeholder{init} is the initializer.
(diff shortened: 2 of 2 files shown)


source/expressions.tex
```diff
@@ -3353,14 +3353,9 @@
...
 for the remainder of this subclause.
-Otherwise, if the type is \tcode{auto},
-it is replaced by the type deduced for the variable \tcode{x}
-in the invented declaration\iref{dcl.spec.auto},
-which is never interpreted as a function declaration:
-\begin{codeblock}
-auto x @\placeholder{init}@;
-\end{codeblock}
-, where \placeholder{init} is the initializer.
+Otherwise, if the type contains a placeholder type,,
```
JohelEGP, line 3356:
```suggestion
Otherwise, if the type contains a placeholder type,
```

source/declarations.tex
```diff
@@ -1954,6 +1954,20 @@
...
+\begin{itemize}
+\item
+If the initializer is a \grammarterm{braced-init-list},
+it shall consist of a single brace-enclosed \grammarterm{assignment-expression}
+and $E$ is the \grammarterm{assignment-expression}.
+\item
+If the initializer is a parenthesized \grammarterm{expression-list},
+the \grammarterm{expression-list} shall be
+a single \grammarterm{assignment-expression}
+and $E$ is the \grammarterm{assignment-expression}.
```
opensdh, line 1969: It's weird to say this explicitly for `auto(4)` but not for `auto x(4);`, since the latter's initializer is `(4)`.  It's preexisting, but should we adjust /2.2 (or even try to combine them)?

source/declarations.tex
```diff
@@ -1954,6 +1954,20 @@
 that contains a placeholder type,
 \tcode{T} is the declared type of the non-type template parameter
 and $E$ is the corresponding template argument.
+\item
```
opensdh, line 1957: I think this would make more sense as the second or third bullet, so that it's immediately adjacent to the very similar variable case.

opensdh (approved):

source/declarations.tex
jensmaurer: But... we say that in p2.2.3:
> If the initializer is a parenthesized expression-list, the expression-list shall be a single assignment-expression and E is the assignment-expression

source/declarations.tex
tkoeppe: (I think that would make sense.)

tkoeppe: Yes, I see it now --  I confused myself among the multiple nested lists. Done!

source/declarations.tex
tkoeppe: I'd say let's leave this for now? I'm not sure what exactly we could improve here.

PR #5040, opened 2021-10-19
[lex] P2314r4 fixes

JohelEGP:


tkoeppe:
Yes, thanks, will do after the motions. I'll squash this into one commit, too.

PR #5042, opened 2021-10-21
[utilities][range.nonprop.cache] Simplify direct-non-list-initalization phrasing

### LINKED ISSUE #5038: Wording for how "a contained value is initialized" could be simplified
From P078R8:

This _is_ the wording from the paper, but "whose contained value is direct-non-list-initialized with `invoke(std::forward<F>(f), value())`" would be clearer. LWG made a similar clarification to the `expected` paper recently. Similarly applies to lines 3512-3514.

_Originally posted by @CaseyCarter in https://github.com/cplusplus/draft/pull/5014#r727405883_

If there is a simpler expression, we should apply it to all the `optional` wording.

timsong-cpp:
Fixes #5038.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -3315,8 +3315,7 @@ there is no requirement that it is movable\iref{dcl.init.general}.
 
  \effects
 Calls \tcode{reset()}.
-Then initializes the contained value as if direct-non-list-initializing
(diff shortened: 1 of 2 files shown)


CaseyCarter (approved):
Excellent, Tim - thanks!

tkoeppe:
Very nice, thanks!

PR #5048, opened 2021-10-21
[ranges] Tweak use of braces and whitespace

tkoeppe:
For example:
* Use braces symmetrically in if+else statements.
* Consistently spell empty blocks as "{}", on the same line.
* Indent required-clauses consistently.
* Avoid unnecessary line breaks.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -85,7 +85,7 @@ namespace std::ranges {
   template<class T>
     inline constexpr bool enable_view = @\seebelow@;
 
-  struct view_base { };
+  struct view_base {};
 
   template<class T>
     concept view = @\seebelow@;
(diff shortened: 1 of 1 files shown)


tkoeppe:
This is meant to be an incremental improvement, not a perfect solution.

I have some unresolved questions around how we format functions with inline definitions, in particular when there is a requires-clause.

source/ranges.tex
```diff
@@ -5120,29 +5120,26 @@
...
     {
       const auto s = ranges::size(@\exposid{base_}@);
       const auto c = static_cast<decltype(s)>(@\exposid{count_}@);
       return s < c ? 0 : s - c;
     }
 
-    constexpr auto size() const
-      requires @\libconcept{sized_range}@<const V>
+    constexpr auto size() const requires @\libconcept{sized_range}@<const V>
     {
```
jensmaurer, line 5137: This brace should be on the previous line.

source/ranges.tex
```diff
@@ -5248,8 +5245,7 @@
 
     constexpr auto begin();
 
-    constexpr auto end()
-    { return ranges::end(@\exposid{base_}@); }
+    constexpr auto end() { return ranges::end(@\exposid{base_}@); }
```
jensmaurer, line 5248: I don't think I like a non-empty definition on the same line as the "declaration" part.

source/ranges.tex
```diff
@@ -8419,9 +8417,7 @@
 
     constexpr explicit zip_transform_view(F fun, Views... views);
 
-    constexpr auto begin() {
-      return @\exposid{iterator}@<false>(*this, @\exposid{zip_}@.begin());
-    }
+    constexpr auto begin() { return @\exposid{iterator}@<false>(*this, @\exposid{zip_}@.begin()); }
```
jensmaurer, line 8420: Same here: definition on same line seems a bit over-the-top, but maybe it improves consistency.

source/ranges.tex
tkoeppe: Fixed.

Issue #4937, opened 2021-09-26
Consider adding a macro for semantic cross references


JohelEGP:
#4936 shows that when wording is refactored, it is necessary to manually keep in sync cross references, and that unfortunate references are left semantically dangling. For that particular case, #4832 was going to copy some dangling references. #4866 also fixed references to that subclause within [basic].

It seems like there is benefit to be found in a macro for doing semantic cross references. This means that they refer not to a fixed subclause label, but something closer to what is semantically cross reference. The obvious candidate for this are indexed entities.

I am thinking of something along the lines of
```LaTeX
\reference{matter involving an \grammarterm{id-expression}}{\grammarterm{id-expression}}
```

jensmaurer:
> Is it possible to do that with LaTeX?

Conceptually, what we need is a way to add `\ref`s (or `\iref`s) to a named list of things in a first LaTeX pass, and then we expand a macro to that list in a later pass (not unlike what `\ref` does in the first place). We would then use a special marker when we use (e.g.) "unevaluated operand" in a way that complements the definition, and that special marker would add to the named list.

@godbyk , any pointers would be much appreciated.

jensmaurer:
I've misunderstood the original desire, I believe. It seems just `\label` and `\ref` will do what we want.

PR #5049, opened 2021-10-21
[basic.def.odr] Introduce label 'term.odr.use' and refer to it

jensmaurer:
instead of referring to 'basic.def.odr'.  The latter breaks if we
ever move the definition of 'odr-use', e.g. to a subclause.

This new "term.x" convention is also documented in the wiki now.
Partially addresses #4937

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -426,6 +426,7 @@ This covers
 \end{itemize}
 
 \pnum
+\label{term.odr.use}%
 A variable is named by an expression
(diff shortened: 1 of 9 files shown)


jensmaurer:
(no visible diff)

Curefab:
All links to _term.odr.use_ are broken on https://eel.is/c++draft/term.odr.use

JohelEGP:
That's https://github.com/Eelis/cxxdraft-htmlgen/issues/91.

PR #5050, opened 2021-10-21
[diff.cpp14.lex] Fix index entry for implementation-defined behavior

### LINKED ISSUE #5041: "Duplicate" index entries re: implementation-defined mapping from source
The index entries are subtly different, but they should be merged:
- mapping physical source characters to translation character set
- mapping from physical source characters to translation character set

_Originally posted by @hubert-reinterpretcast in https://github.com/cplusplus/draft/pull/5004#discussion_r732309560_

jensmaurer:
Fixes #5041


tkoeppe:
Fine, I trust your domain expertise that that's the intended meaning of (all occurrences of) this impldef.

PR #5064, opened 2021-10-22
[range.view] Move note outside of list.

tkoeppe:
Also replace "and/or" by just "/", which already means "and/or".


Flexiproof (commented):
~~spam~~

PR #5065, opened 2021-10-22
[cpp.replace.general] Rephrase function-like macro invocations

### LINKED ISSUE #5022: Consider introducing a new grammar `pp-balanced-token-seq`
Just a note from a discussion with Richard, for future consideration. We have a few places in which we spell out in prose how to form a balanced preprocessor token sequence. This could be simplified if we defined a new piece of grammar. The idea, as described by @zygoloid, might look like this:

```
pp-balanced-token:
    any preprocessor-token other than ( or )
    ( pp-balanced-token-seq[opt] )
pp-balanced-token-seq:
    pp-balanced-token
    pp-balanced-token-seq pp-balanced-token
va-opt-replacement:
    __VA_OPT__ ( pp-balanced-token-seq[opt] )
```

Then:

"""
delete [cpp.replace.general]p5 (it's misplaced anyway; this belongs in p12).
I
n p12, delete "Each subsequent [...rest of paragraph 12 and paragraph 13...]".

Add new paragraph after p12, combining parts of p5, p12, p13:

"Each instance of a function-like macro name followed by a ( as the next preprocessing token shall be a function-like macro invocation of the form:

    identifier ( pp-balanced-token-seq[opt] )

Within the sequence of preprocessing tokens making up an invocation of a function-like macro, new-line is considered a normal whitespace character. The arguments of the macro invocation are the sequences of pp-balanced-tokens separated by comma preprocessing tokens. If the identifier-list in the macro definition does not end with an ellipsis, the number of arguments (including those arguments consisting of no pp-balanced-tokens) shall equal the number of parameters in the macro definition. Otherwise, there shall be at least as many arguments in the invocation as there are parameters in the macro definition (excluding the ...). If there are sequences of preprocessing tokens within the list of arguments that would otherwise act as preprocessing directives,[Footnote] the behavior is undefined."

Then include p15 ("If there is a ... immediately preceding the )").

Then a new paragraph defining the semantics:

"Each function-like macro invocation is replaced by the replacement list in the macro definition, after performing argument substitution as described in [cpp.subst]."

Then finally p14 (the example).

I might also want to split out an "object-like macro" and a "function-like macro" section from [cpp.replace.general]
"""

jensmaurer:
Fixes #5022

### DIFF
--- a/source/preprocessor.tex
+++ b/source/preprocessor.tex
@@ -988,18 +988,6 @@ There shall be whitespace between the identifier and the replacement list
 in the definition of an object-like macro.
 
 \pnum
-If the \grammarterm{identifier-list} in the macro definition does not end with
-an ellipsis, the number of arguments (including those arguments consisting
-of no preprocessing tokens)
-in an invocation of a function-like macro shall
-equal the number of parameters in the macro definition.
-Otherwise, there shall be at least as many arguments in the invocation as there are
-parameters in the macro definition (excluding the \tcode{...}). There
-shall exist a
-\tcode{)}
-preprocessing token that terminates the invocation.
-
-\pnum
 \indextext{__va_args__@\mname{VA_ARGS}}%
 \indextext{__va_opt__@\mname{VA_OPT}}%
 The identifiers \mname{VA_ARGS} and \mname{VA_OPT}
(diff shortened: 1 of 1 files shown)


tkoeppe:
This seems to have happened in other ways since. Thanks anyway!

PR #5077, opened 2021-10-27
[lex] Use \textsc for Unicode character names

### LINKED ISSUE #5032: inconsistent formatting for \U+nnnn
For occurrences of the form "U+nnnn", we tend not to use \tcode but we're inconsistent.
We should come up with some consistent formatting for these things.

jensmaurer:
Fixes #5032

The presentation of Unicode character names is ... hurting the eye. Using \textsc improves the situation:

Before:

unicode1

After:

unicode2

We could also consider using fixed-width font for the "U+nnnn" part, but maybe it is better to limit fixed-width font to C++-level things, which "U+nnnn" most definitely is not.

@tkoeppe, I'd appreciate your opinion here.

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -243,45 +243,45 @@ the numerical value has no other meaning in this context.
 \begin{floattable}{Basic character set}{lex.charset.basic}{lll}
 \topline
 \lhdrx{2}{character} & \rhdr{glyph} \\ \capsep
-U+0009 & CHARACTER TABULATION & \\
-U+000B & LINE TABULATION & \\
-U+000C & FORM FEED (FF) & \\
-U+0020 & SPACE & \\
-U+000A & LINE FEED (LF) & new-line \\
-U+0021 & EXCLAMATION MARK & \tcode{!} \\
-U+0022 & QUOTATION MARK & \tcode{"} \\
-U+0023 & NUMBER SIGN & \tcode{\#} \\
-U+0025 & PERCENT SIGN & \tcode{\%} \\
-U+0026 & AMPERSAND  & \tcode{\&} \\
-U+0027 & APOSTROPHE & \tcode{'} \\
-U+0028 & LEFT PARENTHESIS & \tcode{(} \\
-U+0029 & RIGHT PARENTHESIS & \tcode{)} \\
-U+002A & ASTERISK & \tcode{*} \\
-U+002B & PLUS SIGN & \tcode{+} \\
-U+002C & COMMA & \tcode{,} \\
-U+002D & HYPHEN-MINUS & \tcode{-} \\
-U+002E & FULL STOP & \tcode{.} \\
-U+002F & SOLIDUS & \tcode{/} \\
-U+0030 .. U+0039 & DIGIT ZERO .. NINE & \tcode{0 1 2 3 4 5 6 7 8 9} \\
-U+003A & COLON & \tcode{:} \\
-U+003B & SEMICOLON & \tcode{;} \\
-U+003C & LESS-THAN SIGN & \tcode{<} \\
-U+003D & EQUALS SIGN & \tcode{=} \\
-U+003E & GREATER-THAN SIGN & \tcode{>} \\
-U+003F & QUESTION MARK & \tcode{?} \\
-U+0041 .. U+005A & LATIN CAPITAL LETTER A .. Z & \tcode{A B C D E F G H I J K L M} \\
-& & \tcode{N O P Q R S T U V W X Y Z} \\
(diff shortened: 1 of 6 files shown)


jensmaurer:
@tkoeppe , do you have an opinion here?

tkoeppe:
I saw that, let me think about it... it's a bit jarring to have the U+nnnn be so large by comparison -- could we perhaps make that smallcaps, too?

jensmaurer:
@tkoeppe, tried that. The problem is that the digits stay the same size:

lex

jensmaurer:
@tkoeppe, here's everything with simply `\textsmaller[1]`. This fixes the size of the digits.

jensmaurer:
This last variant looks quite nice to me.

tkoeppe:
Shame there are no real small-caps-sized numerals in our main font. I'll look a bit more. How are your scaled numerals looking for contrast and weight?

tkoeppe:
(But I think we're onto something here.)

jensmaurer:
@tkoeppe, see the latest screenshot three comments up. This has scaled digits.

tkoeppe:
Can we use actual small-caps letters, and only use scaling for the numerals? That avoids the awkward spacing of capitals (captials aren't designed to be juxtaposed and would normally call for tracking).

Can you try this:

    \newcommand{\unicode}[2]{\textsc{u}\textsmaller[1]{\protect\raisebox{.2ex}{\textsmaller[1]{+}}\uppercase{#1}} \textsc{#2}}

jensmaurer:
@tkoeppe, here's `\textsc` for everything, except scaling the digits manually:

lex

tkoeppe:
Look closely, the scaled-down caps are too tight and too thin. The proper small caps are more widely spaced and have the same weight as the other letters.

I think this is nice now, we can apply this.

tkoeppe:
Hm, I don't think we have a monospaced roman font. How do you feel about using the typewriter font for the hexadecimal digits (but leaving everything else the same)?

jensmaurer:
@tkoeppe, here is the monospace variant (including making the `U+` monospace):

lex

tkoeppe:
Yeah, I tried that, too. I'm not sure. Both versions have merit. Normal roman font blends in more nicely but doesn't give us fixed-width codepoints, whereas the codefont stands out a bit more (in the sense of being more disrupting). I think I could live with both versions -- any thoughts or preferences? @burblebee?

jensmaurer:
We usually reserve code font for, well, code, i.e. snippets that could reasonably appear in a valid C++ program. U+nnnn is definitely not of that kind. I do see the advantage of using a fixed-width font here, but that doesn't outweigh the disadvantages (more visually disruptive; does not visually connect to/align with the following character name) for me.

I do note that given the work applied to use macros for Unicode designators, we're at liberty to quickly change the layout down the road, should be so desire.

tkoeppe:
I was also thinking with _only_ making the hex digit sequence use code font, but it's also not that great. OK, let's go with the scaled roman capitals then, that definitely looked nice. And the lack of fixed would only be noticable in tabular settings.

(As long as only decimal digits appear, we _can_ use fixed-with numerals from the roman font, too --  that doesn't solve every problem, but maybe there'll be one or two situations where we can lean on that.)

jensmaurer:
Now that we've got that settled, this feels ready to me. Please review carefully that I haven't introduced an accidental change in the numbers or names while restructuring the code to use the new macros.

tkoeppe:
Thanks a lot! The smallest capitals for the biggest wins.

Eelis:
I didn't follow the complete conversation, but I just noticed that in the PDF, letters occurring as hex digits are now a bit bigger than the U at the start:

image

Looks weird, was that intentional? (If so, ignore this message.)

tkoeppe:
@Eelis: That's because the (hex) numbers are manually scaled, whereas everything else is small-caps. The problem is ultimately that our font doesn't have small-caps numerals, so we're emulating those. The details could be tweaked of course; suggestions welcome.

Eelis:
@tkoeppe For cxxdraft-htmlgen, I ended up using https://github.com/Eelis/cxxdraft-htmlgen/commit/e5f612d2c5905b4caa2fe4026d2822b23dac62b1 because scaled down capital smallcaps seems to look good in Firefox and Chromium. But I see that in LaTeX, that approach creates too widely spaced letters, so I'm at a loss, sorry.

tkoeppe:
I can make the scaled caps a little smaller still:

image

down from status quo:

image

tkoeppe:
Looks like Source Sans Pro has small-caps numerals: https://fonts.google.com/specimen/Source+Sans+Pro

Not an option for us unfortunately :-)

PR #5079, opened 2021-10-27
[macros] Allow \keyword and \grammarterm in section headings

jensmaurer:
by disabling all indexing when typesetting the table-of-contents.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -559,7 +559,7 @@ struct S {
 \end{codeblock}
 \end{example}
 
-\rSec2[dcl.typedef]{The \tcode{typedef} specifier}%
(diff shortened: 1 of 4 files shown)


JohelEGP:
Does this also mean that `\cname` is no longer necessary?

jensmaurer:
We could apply the approach to `\libconcept`, too, and thus replace `\cname` with `\libconcept` in section headings, yes.

tkoeppe:
(This actually changes surprisingly little; I suppose most keywords in headings are already indexed in the subsequent text.)

tkoeppe:
Ah yes, of course: the change _does_ add new indexing, just not in the TOC.

PR #5083, opened 2021-10-30
[basic.start.main] remove redundant phrase

timsong-cpp:
If "to end the program" is merely describing what `std::exit` does, it's redundant; if it is read as expressing purpose or intent, it's incorrect - whether a program has UB cannot depend on the programmer's intent.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -6386,7 +6386,7 @@ Terminating the program
 without leaving the current block (e.g., by calling the function
 \tcode{std::exit(int)}\iref{support.start.term}) does not destroy any
 objects with automatic storage duration\iref{class.dtor}. If
-\tcode{std::exit} is called to end a program during the destruction of
+\tcode{std::exit} is called during the destruction of
 an object with static or thread storage duration, the program has undefined
 behavior.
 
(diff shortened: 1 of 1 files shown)


source/basic.tex
```diff
@@ -6386,7 +6386,7 @@
 without leaving the current block (e.g., by calling the function
 \tcode{std::exit(int)}\iref{support.start.term}) does not destroy any
 objects with automatic storage duration\iref{class.dtor}. If
-\tcode{std::exit} is called to end a program during the destruction of
+\tcode{std::exit} is called during the destruction of
```
languagelawyer, line 6389:
```suggestion
\tcode{std::exit} is invoked during the destruction of
```
https://github.com/cplusplus/draft/issues/3079#issuecomment-544636262

languagelawyer (changes requested):

jensmaurer:
@timsong-cpp , generally agreed. Could you please apply the "called" -> "invoked" drive-by fix? ("Called" is ambiguous and could mean "named" or "termed": "I call my son Richard", with no connection to function calls.)

jensmaurer (approved):
Looks good. @tkoeppe , please squash when merging.

PR #5089, opened 2021-11-05
[check] Cosmetics

jensmaurer:
I was hoping find the "Error" bug, but that wasn't successful.

Posted https://github.com/github/feedback/discussions/7381


tkoeppe:
Nice, thanks! I've been using a very similar style in some other projects.

PR #5092, opened 2021-11-06
[except.throw,except.handle] Move lvalue specification for copies

### LINKED ISSUE #5091: [except.throw] doesn't specify the type of the lvalue denoting the exception object
[except.throw]/3:
> Throwing an exception copy-initializes a temporary object, called the _exception object_.
> An lvalue denoting the temporary is used to initialize the variable declared in the matching handler  ([except.handle]).

Is there any wording defining the type of this lvalue? Also, https://timsong-cpp.github.io/cppwp/except.handle#14 is worded in terms of initialization from object, not expression. Does it need a separate issue?

jensmaurer:
Fixes #5091

### DIFF
--- a/source/exceptions.tex
+++ b/source/exceptions.tex
@@ -260,9 +260,6 @@ Throwing an exception
 copy-initializes~(\ref{dcl.init}, \ref{class.copy.ctor}) a temporary object,
 called the
 \defnx{exception object}{exception handling!exception object}.
-An lvalue denoting the temporary is used to initialize the
-variable declared in the matching
-\grammarterm{handler}\iref{except.handle}.
 If the type of the exception object would be
 an incomplete type,
 an abstract class type\iref{class.abstract},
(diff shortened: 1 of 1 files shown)


jensmaurer:
@languagelawyer , this should address your concerns.

languagelawyer:
Still doesn't say anything about the lvalue type.

jensmaurer:
Do you want "an lvalue of the type of the exception object, designating the exception object"?

source/exceptions.tex
```diff
@@ -655,11 +652,12 @@
 of type \tcode{E}, as follows:
 \begin{itemize}
 \item
-if \tcode{T} is a base class of \tcode{E}, the variable is
-copy-initialized\iref{dcl.init} from the corresponding base class subobject
+if \tcode{T} is a base class of \tcode{E},
+the variable is copy-initialized\iref{dcl.init}
+from an lvalue designating the corresponding base class subobject
```
languagelawyer, line 657:
```suggestion
from an lvalue of type \tcode{T} designating the corresponding base class subobject
```

source/exceptions.tex
```diff
@@ -655,11 +652,12 @@
...
 of the exception object;
 \item otherwise, the variable is copy-initialized\iref{dcl.init}
-from the exception object.
+from an lvalue designating the exception object.
```
languagelawyer, line 660:
```suggestion
from an lvalue of type \tcode{E} designating the exception object.
```

languagelawyer (changes requested):

jensmaurer:
Sounds good.

PR #5095, opened 2021-11-07
[lex.string] Clarify size of string-literal

### LINKED ISSUE #5090: [lex.string]: Encoding described only for expression evaluation
In [lex.string], the length of the array type of a string literal determined from the number of encoded code units "as described below"; however, the further description only describes encoding for string literal objects, and string literal objects are only mentioned in relation to evaluating a _string-literal_.

A string may appear within unevaluated operands in such a way that its type needs to be known, for example:
```cpp
char s[sizeof("Hello, world!")];
```

As per [expr.context], "[an] unevaluated operand is not evaluated".

Perhaps, "where _n_ is the number of encoded code units <del>as described below </del><ins>in the result of an evaluation of the _string-literal_ were it to be evaluated</ins>" would work?

jensmaurer:
Fixes #5090

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -1809,7 +1809,9 @@ its associated character encoding\iref{lex.charset}
 are determined by its encoding prefix and sequence of
 \grammarterm{s-char}s or \grammarterm{r-char}s
 as defined by \tref{lex.string.literal}
-where $n$ is the number of encoded code units as described below.
+where $n$ is the number of encoded code units
+that would result from an evaluation of the \grammarterm{string-literal}
+(see below).
 
 \begin{floattable}{String literals}{lex.string.literal}
 {llp{2.6cm}p{2.3cm}p{4.7cm}}


jensmaurer:
Put on the CWG agenda: https://wiki.edg.com/bin/view/Wg21telecons2021/CoreWorkingGroup

tkoeppe:
CWG meeting consensus, Nov 11: Good.

source/lex.tex
```diff
@@ -1705,7 +1705,9 @@
 are determined by its encoding prefix and sequence of
 \grammarterm{s-char}s or \grammarterm{r-char}s
 as defined by \tref{lex.string.literal}
-where $n$ is the number of encoded code units as described below.
+where $n$ is the number of encoded code units
+in the result of an evaluation of the \grammarterm{string-literal}
+were it to be evaluated (see below).
```
opensdh, line 1710:
```suggestion
that would result from an evaluation of the \grammarterm{string-literal}
(see below).
```

AlisdairM:
Given the CWG approval, could we apply Jens's review and then land?

source/lex.tex
jensmaurer: Applied.

jensmaurer:
@tkoeppe , I think this is ready to go now.

tkoeppe:
Thanks, merging!

PR #5101, opened 2021-11-09
[lex.charset] Define 'valid encoding'

### LINKED ISSUE #4924: Follow-on to P2314R4: properly define encoding restrictions
After P2314R4 Character sets and encodings was merged:

https://wiki.edg.com/bin/view/Wg21telecons2021/PaperReviewIssuesTeleconference2021-09-21

Define a term for the encoding restrictions in lex.charset and use that term in the library section to restrict execution character sets / encodings.

jensmaurer:
and use the term for executation character sets introduced
in [character.seq].

Fixes #4924

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -346,9 +346,11 @@ The \defnadj{wide literal}{encoding} is the encoding applied
 to a wide character or string literal.
 
 \pnum
-A literal encoding or a locale-specific encoding of one of
-the execution character sets\iref{character.seq}
-encodes each element of the basic literal character set as
+An encoding is \defnx{valid}{encoding!valid} if all of the following
+conditions are satisfied:
+\begin{itemize}
+\item
+Each element of the basic literal character set is encoded as
 a single code unit with non-negative value,
 distinct from the code unit for any other such element.
 \begin{note}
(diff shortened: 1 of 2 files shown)


tkoeppe:
@jensmaurer Is this PR still appropriate after the recent round of papers that touch encodings?

jensmaurer:
I think we still want this.

tkoeppe:
OK! (I'll need a rebase then.)

tkoeppe:
Also, @tahonermann, @cor3ntin, could you kindly take a look?

cor3ntin:
The term "valid" is often use to describe a properly encoded sequence, or a sequence of byte that can be cleanly decoded with a given encoding and applying that terms to the encoding itself is confusing.
I'd suggest a different term, such as "conforming encoding", "compliant encoding".

And... all encodings describe in the standard should be conforming, tautologically.
So... I question we do need a definition at all.

`All literal and execution encodings shall` [bullet list of requirements - which is correct as modified]

I would keep `The ordinary and wide literal encodings are valid encodings, but are otherwise \impldef{ordinary and wide literal encodings}` in `lib-intro.tex`, as we have been trying to not mention them in the core wording too much. ie, in lib-intro

` The sets of additional elements (if any) are locale-specific. The encodings of the execution character sets are locale-specific and implementation-defined.`

tahonermann:
I agree with Corentin's review; I found the proposed wording with regard to valid encodings confusing. I do like the idea of naming the conditions an encoding must meet to be acceptable as a literal or execution encoding though. Perhaps "basis encoding"? The intent is to define a set of minimum requirements that can be extended; I think "basis" captures that intent. The wording might then look something like:

> A *basis encoding* is a character encoding that satisfies the following conditions
> ...
> The ordinary and wide literal encodings are \impldef{ordinary and wide literal encodings} basis encodings\iref{lex.charset}.
> ...
> Those encodings are basis encodings\iref{lex.charset}.
> ...
> The encodings of the execution character sets are locale-specific basis encodings\iref{lex.charset}.

tkoeppe:
Editorial meeting: the objections to the overly generic term "valid" are well taken, and we should look for a better term. The overall goal of factoring out the core notion in question is valuable, but we need to find the right way to express it.

Ideas welcome!

tahonermann:
@tkoeppe,
> Ideas welcome!

The [prior comment](https://github.com/cplusplus/draft/pull/5101#issuecomment-1220832911) included a suggestion.

tkoeppe:
Ah yes! Hm, "basis" is only one letter away from "basic", which may have too much historic ballast in this domain? But yes, variations on that approach occurred to us, too. E.g. "admissable encoding".

cor3ntin:
If we need to keep a terminology, "admissible", "conforming", "supported", or variations on that theme do seem like good solutions.

AlisdairM:
This PR would resolve NB comment US 63-115 (for C++26) and the confusion behind it.

PR #5104, opened 2021-11-12
Introduce term.* labels for [basic.types.general] terms

### LINKED ISSUE #4630: Imprecise cross-references for "trivially copyable type"
[refwrap.general] says:

> `reference_wrapper<T>` is a trivially copyable type (6.8).

Where 6.8 is [basic.types]. The definition of trivially copyable is in 6.8.1 [basic.types.general]. I assume we used to refer to the hanging paragraphs at the start of [basic.types] but they now have their own subclause. The cross-ref should have been updated to refer to that new subclause.

It looks like most cross-references to [basic.types] should actually be [basic.types.general]:

```
$ git grep '{basic.types}' *.tex
atomics.tex:on objects without padding bits\iref{basic.types} is
basic.tex:an incomplete type\iref{basic.types},
basic.tex:in~\ref{basic.types}.
basic.tex:type\iref{basic.types}.
basic.tex:standard-layout type\iref{basic.types} shall occupy contiguous bytes of
basic.tex:zero or more objects of implicit-lifetime types\iref{basic.types}
basic.tex:\ref{basic.types} and the subclauses thereof
basic.tex:Each set of values for any padding bits\iref{basic.types}
basic.tex:the width of that type has padding bits; see \ref{basic.types}.
basic.tex:\keyword{void}\iref{basic.types} has three corresponding cv-qualified
classes.tex:type\iref{basic.types} (or complete type) at the closing \tcode{\}} of
classes.tex:incomplete type\iref{basic.types},
classes.tex:both classes\iref{basic.types}.
classes.tex:types\iref{basic.types}.
classes.tex:\tcode{X} copies the object representation\iref{basic.types} of \tcode{X}.
classes.tex:union \tcode{X} copies the object representation\iref{basic.types} of \tcode{X}.
classes.tex:the extra bits are padding bits\iref{basic.types}.
compatibility.tex:\diffref{basic.types}
concepts.tex:  \item \tcode{T} is a literal type\iref{basic.types},
declarations.tex:is a scalar type\iref{basic.types}, the
declarations.tex:its padding bits\iref{basic.types} are initialized to zero bits and
declarations.tex:its padding bits\iref{basic.types} are initialized to zero bits and
expressions.tex:does not require that type to be complete\iref{basic.types}.
expressions.tex:and~\ref{basic.types} for the definition of object representation.
expressions.tex:thereof~(\ref{intro.object}, \ref{basic.types}, \ref{class.abstract}).
lib-intro.tex:\tcode{T, U, C}    &   any \cv-unqualified object type\iref{basic.types}       \\ \rowsep
lib-intro.tex:If an incomplete type\iref{basic.types} is used as a template
numerics.tex:\tcode{complex<long double>} are literal types\iref{basic.types}.
strings.tex:any non-array trivial standard-layout\iref{basic.types} type.
templates.tex:A template type argument can be an incomplete type\iref{basic.types}.
templates.tex:the instantiation yields an incomplete class type\iref{basic.types}.
templates.tex:classes\iref{basic.types}.
utilities.tex:representation\iref{basic.types} of \tcode{unsigned long long}.
utilities.tex:\tcode{T} is not an incomplete type\iref{basic.types}.
utilities.tex:type\iref{basic.types} that can be the type of the
utilities.tex:\tcode{reference_wrapper<T>} is a trivially copyable type\iref{basic.types}.
utilities.tex:subclause~\ref{basic.types} of the \Cpp{} standard.
utilities.tex:categories, corresponding to the descriptions given in subclause~\ref{basic.types}.
utilities.tex: \tcode{T} is an object type\iref{basic.types}                            &   \\ \rowsep
utilities.tex: \tcode{T} is a scalar type\iref{basic.types}                         &   \\ \rowsep
utilities.tex:is considered to be a trivial~(\ref{basic.types}, \ref{special}) function call
utilities.tex: \tcode{T} is a trivial type\iref{basic.types}     &
utilities.tex: \tcode{T} is a trivially copyable type\iref{basic.types} &
utilities.tex: \tcode{T} is a standard-layout type\iref{basic.types}   &
utilities.tex:  no operation that is not trivial~(\ref{basic.types}, \ref{special}). &
utilities.tex: ~(\ref{basic.types}, \ref{special}). &
utilities.tex: \tcode{T} and \tcode{U} are layout-compatible\iref{basic.types}    &
utilities.tex:is considered to be a trivial~(\ref{basic.types}, \ref{special}) function call
utilities.tex: is no greater than \tcode{Len}\iref{basic.types}.
```

jensmaurer:
For ease of review, these are all separate commits.
Feel free to squash on merge.

Fixes #4630

### DIFF
--- a/source/atomics.tex
+++ b/source/atomics.tex
@@ -1918,7 +1918,7 @@ The result of the comparison.
 \begin{note}
 For example, the effect of
 \tcode{compare_exchange_strong}
-on objects without padding bits\iref{basic.types} is
+on objects without padding bits\iref{term.padding.bits} is
 \begin{codeblock}
 if (memcmp(this, &expected, sizeof(*this)) == 0)
   memcpy(this, &desired, sizeof(*this));
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -4689,6 +4689,7 @@ a function type, not a reference type, and not \cv{}~\keyword{void}.
 \indextext{class!trivial}%
 \indextext{class!trivially copyable}%
 \indextext{class!standard-layout}%
+\label{term.scalar.type}%
 Arithmetic types\iref{basic.fundamental}, enumeration types,
 pointer types, pointer-to-member types\iref{basic.compound},
 \tcode{std::nullptr_t},
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -4684,6 +4684,7 @@ contexts incomplete types are prohibited.
 \end{note}
 
 \pnum
+\label{term.object.type}%
 An \defn{object type} is a (possibly cv-qualified) type that is not
 a function type, not a reference type, and not \cv{}~\keyword{void}.
 
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -4706,7 +4706,9 @@ types are collectively called \defnadjx{trivially copyable}{types}{type}.
 Scalar types, trivial class types\iref{class.prop},
 arrays of such types and cv-qualified versions of these
 types are collectively called
-\defnadjx{trivial}{types}{type}. Scalar types, standard-layout class
+\defnadjx{trivial}{types}{type}.
+\label{term.standard.layout.type}%
+Scalar types, standard-layout class
 types\iref{class.prop}, arrays of such types and
 cv-qualified versions of these types
 are collectively called \defnadjx{standard-layout}{types}{type}.
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -4751,9 +4751,9 @@ will be usable in a constant expression.
 \end{note}
 
 \pnum
-\indextext{layout-compatible type}%
+\label{term.layout.compatible.type}%
 Two types \cvqual{cv1} \tcode{T1} and \cvqual{cv2} \tcode{T2} are
-\defn{layout-compatible} types
+\defnadjx{layout-compatible}{types}{type}
 if \tcode{T1} and \tcode{T2} are the same type,
 layout-compatible enumerations\iref{dcl.enum}, or
 layout-compatible standard-layout class types\iref{class.mem}.
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -318,7 +318,7 @@ A class name can also be implicitly declared by an
 \indextext{type!incomplete}%
 In the definition of an object,
 the type of that object shall not be
-an incomplete type\iref{basic.types.general},
+an incomplete type\iref{term.incomplete.type},
 an abstract class type\iref{class.abstract}, or
 a (possibly multi-dimensional) array thereof.
 
@@ -4625,8 +4625,9 @@ bound or of incomplete element type, is an
 The size and layout of an instance of an incompletely-defined
 object type is unknown.
 \end{footnote}
+\label{term.incomplete.type}%
 Incompletely-defined object types and \cv{}~\keyword{void} are
-\defnx{incomplete types}{type!incomplete}\iref{basic.fundamental}.
+\defnadjx{incomplete}{types}{type}\iref{basic.fundamental}.
 \begin{note}
 Objects cannot be defined to have an incomplete type\iref{basic.def}.
 \end{note}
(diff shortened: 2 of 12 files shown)


JohelEGP (commented):
https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines says "create a \label{term.this.name} at the start of the paragraph defining the term". Perhaps it should say "sentence" rather than "paragraph".

61245eb starts with `[std}` rather than `[std]`.

jensmaurer:
@JohelEGP , fixed both issues.

tkoeppe:
@zygoloid, @opensdh: this looks very nice, but I'd welcome some more Core eyes on those new terms.

JohelEGP:
No new terms were introduced.

tkoeppe:
(I mean the internal labelling of terms. The labelling is new, I suppose, not the terms.)

jensmaurer:
In order to make `\tref` workable, we need LaTeX foo to replace the periods with spaces (or the other way round).

tkoeppe:
(Tangent: if we had a text replacement system for LaTeX, I would also have a nice improvement for the unicode-smallcaps formatting -- I would like to format numerals and letters differently, but I don't know how to do that as a macro.)

tkoeppe:
Why can't `\tref{foo.type}` work like `savebox` where it stored the actual content somewhere else?

jensmaurer:
I guess there are sufficiently many \saveboxes; I thought that's a quite finite resource.
Anyway, the definition of the term would need to translate space to period, which is approximately the same problem.

tkoeppe:
We could use a custom label to store terms. For example:

```
\documentclass[a4paper,11pt]{article}

\makeatletter
\newcommand{\termdef}[2]{%
\protected@write \@auxout {}{\string \newlabel {#1}{{#2}{}}}#2}
\makeatother

\begin{document}
The \ref{term.dcl.spec} shall not be empty.
A \termdef{term.dcl.spec}{decl-specifier} is a thing you can use.
\end{document}
```

jensmaurer:
@tkoeppe, that's not substantially different from my `\label{term...}` approach, which requires less internal knowledge about the structure of the .aux files.

We'd reap real benefits if we could make `\defn` (etc) auto-generate a label that could then be used via `\termref` or similar, giving human-readable space-separated terms in both cases, and internally mapping to a dotted form with a "term." prefix.

Ah, looks like the `xstring` package does exactly that: https://tex.stackexchange.com/questions/36266/replacing-characters-in-argument-strings

tkoeppe:
@jensmaurer, sort of, but what I'm proposing is to just explicitly spell out _both_ the label and the actual term, rather than deriving one from the other. Not only does that require less implementation magic, but it also makes it possible to change the term without changing the label, etc. And the auxfile labelling mechanism is pretty well-established and tractable.

We could make a new command to replace `\ref`, e.g. `\termref`, to make the use clearer at the callsite.

jensmaurer:
@tkoeppe , thanks for your thoughts. See #5210 for an approach that introduces term.* labels automatically.

Note that for each `\defn` (i.e. each term definition), we produce three things:
- text to be shown in the document
- an index entry
- (new) a label

We need macros so that the common case is easy, but (if need be) all three items must be influenceable independently.
For now, #5210 auto-generates labels from (the indexing portion of) \defn, \defnx, \defnadj, \defnadjx, but for index entries using special characters such as ! or @ to fine-tune the indexing, the resulting term.* label is ugly.
That doesn't prevent us from introducing a nice-looking term.* label in the vicinity, though.

In my view, \termref should show both the term and the cross-reference to where it is defined, e.g. "odr-use (6.2)", while naming the term only once.

And just to point out: Memoir creates .aux entries like this for \label:

`\newlabel{term.piecewise.construction}{{\M@TitleReference {20.4.2}{Class template \texttt  {pair}}}{579}{Class template \tcode {pair}}{lstlisting.20.-1376}{}}`

jensmaurer:
Automating the term.* definitions does detect duplicate term definitions quite nicely; see #5210

PR #5106, opened 2021-11-12
[spanstream.ctor] Fix base class name in effects

jwakely:


jwakely:
N.B. this typo was present in [P0448R4](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p0448r4.pdf) but since `basic_spanstream` doesn't have a base class of that type, it's obvious that it should be `basic_iostream`, as in the next paragraph.

jensmaurer (approved):

PR #5113, opened 2021-11-19
[index] Index keywords as '<keyword>, keyword' for differentiation

### LINKED ISSUE #5061: Mix of "see" and page numbers in the same index entry
image

These look really odd, not the least because the "see" happens in the middle of a long list of page numbers.

It looks like the "multiple encaps" warnings in `std-generalindex.ilg` can be used to detect these. (But that same warning is benign in other indices where we intentionally format one of the page numbers in bold.)

jensmaurer:
Fixes #5061

keyword1

keyword2


tkoeppe:
Nice! @zygoloid, @jwakely Any opinions?

jensmaurer:
@tkoeppe, can we move forward here, please?

tkoeppe:
Yes, thanks!

PR #5118, opened 2021-11-20
[lex.string,dcl.init.string] Use 'code unit' for initialization from a string-literal

### LINKED ISSUE #5103: [dcl.init.string] p1 Successive characters initializing array elements make nosense after P2314
> Successive characters of the value of the string-literal initialize the elements of the array.  

Since P2314 introduces a clear concept about "code unit", which is an integer value of a character type that is the type of the element of the character arrays. So, this outdated wording should be improved

> Successive code unit values of the string-literal initialize the elements of the array.

jensmaurer:
[lex.string] defines a sequence of code unit values
for a string-literal.  Use that sequence for the
initialization of an array by a string-literal.

Fixes #5103

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -5367,12 +5367,11 @@ UTF-16 string literal,
 UTF-32 string literal, or
 wide string literal,
 respectively, or by an appropriately-typed \grammarterm{string-literal} enclosed in
-braces\iref{lex.string}.
+braces.
 \indextext{initialization!character array}%
-Successive
-characters of the
-value of the \grammarterm{string-literal}
-initialize the elements of the array.
+The elements of the array are initialized
+with successive values of the sequence of code unit values
+represented by the \grammarterm{string-literal}\iref{lex.string}.
 \begin{example}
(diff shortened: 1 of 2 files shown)


source/declarations.tex
```diff
@@ -5367,12 +5367,11 @@
...
 wide string literal,
 respectively, or by an appropriately-typed \grammarterm{string-literal} enclosed in
-braces\iref{lex.string}.
+braces.
 \indextext{initialization!character array}%
-Successive
-characters of the
-value of the \grammarterm{string-literal}
-initialize the elements of the array.
+The elements of the array are initialized
```
xmh0511, line 5372: @jensmaurer  Please look at this [comment](https://github.com/cplusplus/draft/issues/5103#issuecomment-974738921). `values of the sequence of code unit values` sounds a bit vague; what value is that?

tkoeppe:
Is this change still useful after the recent encoding-related papers?

jensmaurer:
I think there is still value in a bit of wording enhancement here, but the words need a bit of work.

PR #5121, opened 2021-11-22
[allocator.adaptor.members] Fix select_on_container_copy_construction

### LINKED ISSUE #518: [allocator.adaptor.members] Imprecise wording for scoped_allocator_adaptor::select_on_container_copy_construction
[allocator.adaptor.members]/17 says:

> from the result of calling `allocator_traits<A>::select_on_container_copy_construction()` on
> the corresponding allocator in `*this`.

That function takes an argument, so displaying it with an empty set of parentheses is misleading.

Saying calling it "on the corresponding allocator" doesn't clearly say that the corresponding allocator is passed as the argument to the function.

I suggest something like:

> from the result of calling `allocator_traits<A>::select_on_container_copy_construction` with
> the corresponding allocator in `*this` as the function argument.

jensmaurer:
The description was confusing objects and types.

Fixes #518

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -16126,10 +16126,12 @@ scoped_allocator_adaptor select_on_container_copy_construction() const;
 \begin{itemdescr}
 \pnum
 \returns
-A new \tcode{scoped_allocator_adaptor} object where each allocator \tcode{A} in the
-adaptor is initialized from the result of calling
-\tcode{allocator_traits<A>::select_on_container_copy_construction()} on the
-corresponding allocator in \tcode{*this}.
+A new \tcode{scoped_allocator_adaptor} object
+where each allocator \tcode{a1} within the adaptor
+is initialized with
+\tcode{allocator_traits<A1>::select_on_container_copy_construction(a2)},
+where \tcode{A1} is the type of \tcode{a1} and
+\tcode{a2} is the corresponding allocator in \tcode{*this}.
 \end{itemdescr}
 
 \rSec2[scoped.adaptor.operators]{Operators}


jensmaurer:
@jwakely , after more than half a decade, it's about time this gets fixed. Please have a look.

source/utilities.tex
```diff
@@ -13950,10 +13950,11 @@
...
 \returns
-A new \tcode{scoped_allocator_adaptor} object where each allocator \tcode{A} in the
-adaptor is initialized from the result of calling
-\tcode{allocator_traits<A>::select_on_container_copy_construction()} on the
-corresponding allocator in \tcode{*this}.
+A new \tcode{scoped_allocator_adaptor} object
+where each allocator \tcode{a} of type \tcode{A} within the adaptor
+is initialized with
+\tcode{allocator_traits<A>::select_on_container_copy_construction(a2)},
+where \tcode{a2} is the corresponding allocator in \tcode{*this}.
```
jwakely, line 13957: I'm finding it hard not to read "each allocator `a` of type `A`" as meaning this only applies to the ones of type `A`, for some specific type `A`.

How about:


```suggestion
A new \tcode{scoped_allocator_adaptor} object
where each allocator \tcode{a} within the adaptor
is initialized with
\tcode{allocator_traits<A>::select_on_container_copy_construction(a2)},
where \tcode{A} is the type of \tcode{a} and \tcode{a2} is the corresponding allocator in \tcode{*this}.
```

jensmaurer: You're right. Fixed.

jwakely (approved):

source/utilities.tex
tkoeppe: How about instead of `a` and `a2` we have `a1` and `a2` symmetrically?

Otherwise good, thanks!

jwakely: That seems fine too.

I considered something fancy like `A`<sub>n</sub>, `a`<sub>n</sub> and `this->a`<sub>n</sub>, but it wouldn't be right as `this->a`<sub>n</sub> isn't actually a member of `*this`.

tkoeppe: @jensmaurer: I'd somewhat prefer `a1`, `a2` over `a`, `a2`, but if you feel very strongly, we can do it your way. Please feel free to merge either version!

tkoeppe: @jensmaurer: For your consideration, but fine to go in in principle.

jensmaurer: @tkoeppe, change to a1, a2. Please double-check.

tkoeppe: Thanks, I like it! Can we also make the type of `a1` be `A1`?

jensmaurer: @tkoeppe, done.

tkoeppe: Thank you!

PR #5126, opened 2021-11-24
[lib] Add 'namespace std' wrappings around class definitions

### LINKED ISSUE #1168: Which codeblocks should contain "namespace std"?
Header synopses always contain `namespace std { ...`. When we describe the header content in detail, we often repeat `namespace std {`, e.g. in "Class template vector".

However, we sometimes *don't* repeat the namespace for specifications. Examples are "optional", "polymorphic_memory_resource", and "linear_congruential_engine".

(Note that I am *not* talking about itemdecls; those never state the namespace.)

We should decide whether to state the namespace in those detailed definitions, and then apply that style consistently and record it in the wiki.

I think the majority use at the moment is to state the namespace, but that doesn't need to be a deciding factor. Note also that some class templates live in further nested namespaces, so there may be some value in being explicit.

jensmaurer:
Those were missing in a few places, notably [rand].

Also add an automated check.

Fixes #1168

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -3010,7 +3010,8 @@ is $\mathsf{GA}(\state{x}{i}) = \state{x}{i+1}$.
 \indexlibraryglobal{linear_congruential_engine}%
 \indexlibrarymember{result_type}{linear_congruential_engine}%
 \begin{codeblock}
-template<class UIntType, UIntType a, UIntType c, UIntType m>
+namespace std {
+  template<class UIntType, UIntType a, UIntType c, UIntType m>
   class linear_congruential_engine {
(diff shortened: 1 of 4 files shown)


JohelEGP:
This excludes specialization of type traits in `itemdecl` environments. Like https://wg21.link/pair.astuple.

jensmaurer:
@JohelEGP , I agree with your statement, but I won't fix that in this pull request nor do I think we need to change it at all. Partial specializations, although definitions, perform a sort-of lookup for the primary template, so there's little doubt to which namespace they belong. (We try hard to avoid annotating `itemdecl`s.)

PR #5138, opened 2021-12-04
[associative.reqmts.general] Fix confusing local use of 'r'

### LINKED ISSUE #5134: [associative.reqmts.general] introduces `r` with two different meanings
[N4901](https://isocpp.org/files/papers/N4901.pdf) [associative.reqmts.general]/8.13:

> — `r` denotes a valid dereferenceable iterator to `a`,

But then /8.19 through /8.22 introduce `r` with a totally different meaning:

> — `kl` is a value such that `a` is partitioned (25.8) with respect to `c(r, kl)`, with `r` the key value of `e` and `e` in `a`;
> — `ku` is a value such that `a` is partitioned with respect to `!c(ku, r)`;
> — `ke` is a value such that `a` is partitioned with respect to `c(r, ke)` and `!c(ke, r)`, with `c(r, ke)` implying `!c(ke, r)`;
> — `kx` is a value such that
> — `a` is partitioned with respect to `c(r, rx)` and `!c(kx, r)`, with `c(r, kx)` implying `!c(kx, r)`, and
> — `kx` is not convertible to either `iterator` or `const_iterator`; and

This is confusing; a different name should be introduced.

jensmaurer:
Fixes #5134

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -1710,21 +1710,24 @@ implicitly convertible to
 \tcode{c} denotes a possibly \keyword{const} value of type \tcode{X::key_compare};
 \item
 \tcode{kl} is a value such that \tcode{a} is partitioned\iref{alg.sorting}
-with respect to \tcode{c(r, kl)}, with \tcode{r} the key value of \tcode{e}
-and \tcode{e} in \tcode{a};
+with respect to \tcode{c(x, kl)},
+with \tcode{x} the key value of \tcode{e} and \tcode{e} in \tcode{a};
 \item
 \tcode{ku} is a value such that \tcode{a} is partitioned with respect to
-\tcode{!c(ku, r)};
+\tcode{!c(ku, x)},
+with \tcode{x} the key value of \tcode{e} and \tcode{e} in \tcode{a};
 \item
 \tcode{ke} is a value such that \tcode{a} is partitioned with respect to
-\tcode{c(r, ke)} and \tcode{!c(ke, r)}, with \tcode{c(r, ke)} implying
-\tcode{!c(ke, r)};
+\tcode{c(x, ke)} and \tcode{!c(ke, x)}, with \tcode{c(x, ke)} implying
+\tcode{!c(ke, x)} and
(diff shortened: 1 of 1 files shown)


source/containers.tex
```diff
@@ -1710,21 +1710,24 @@
 \tcode{c} denotes a possibly \keyword{const} value of type \tcode{X::key_compare};
 \item
 \tcode{kl} is a value such that \tcode{a} is partitioned\iref{alg.sorting}
-with respect to \tcode{c(r, kl)}, with \tcode{r} the key value of \tcode{e}
```
tkoeppe, line 1713: I appreciate the heavy maths jargon here, and even though it resonates on a personal level, I think it's maybe a bit unnecessary in a general text. What do you think of changing this to the more normal "where `r` is the key value of `e` and `e` is in [or "an element of"] `a`"?

jensmaurer: Well, the existing text already has the phrase; we're just replacing "r" with "x" and making sure the phrase is duplicated everywhere "x" is used.

The later bullets already contain a "with" phrase, so continuing with "where" is a bit inconsistent.

jensmaurer: (If we want to improve clarity here, adding "all" would probably also be helpful.)

tkoeppe: It's the "consistency with a bad idea" thing... OK, let's make the change as is, and maybe we can rephrase everything afterwards.

jensmaurer:
@jwakely , @StephanTLavavej , please have a look.

jwakely (approved):
As I noted in #5134 it would be a smaller edit to change the first `r` which is the one that introduced the problem (the cases being changed were there first, since C++14).

But this is also OK.

StephanTLavavej (approved):

jensmaurer:
@tkoeppe, this seems ready.

source/containers.tex
tkoeppe: (Note to selves: revisit this.)

PR #5141, opened 2021-12-05
[temp.local] Fix type-name interpretation of injected-class-name

### LINKED ISSUE #5139: [temp.local] injected-class-name as a type-name
[temp.local]/1 says that injected-class-name sometimes can be interpreted as «a _type-name_ equivalent to the _template-name_ followed by the _template-parameter_ s of the class template enclosed in `<>`». But _template-parameter_ is like `typename T`, not `T`.

Prolly it should say «followed by the template argument list» (See http://eel.is/c++draft/temp.decls.general#2.sentence-2, http://eel.is/c++draft/temp.arg.general#2)?

jensmaurer:
Fixes #5139

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -4536,7 +4536,8 @@ it is a \grammarterm{template-name} that refers to the
 class template itself.
 Otherwise, it is a \grammarterm{type-name}
 equivalent to the \grammarterm{template-name}
-followed by the \grammarterm{template-parameter}{s} of the class template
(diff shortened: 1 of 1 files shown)


languagelawyer:
I think there either should be no `\ref{temp.spec.partial.general}` in the first paragraph or the second paragraph should not mention partial specializations since they are already covered by the first one.

jensmaurer:
Good point. Removed xref in the first paragraph.

PR #5145, opened 2021-12-10
[class.virtual] Explicit object member functions cannot be virtual CWG2553

### LINKED ISSUE #5144: [class.virtual] Overriding virtual fuction through an explicit object member function
Consider this example
````cpp
struct Base{
   virtual void show(){}  // #1
};
struct Derived: Base{
    void show(this Base bptr){}  // #2
};
````

According to [basic.scope.scope] p1, `#1` corresponds to `#2` as per [basic.scope.scope] p4 if their object parameters should be corresponding. [basic.scope.scope] p3 says that
> Two non-static member functions have corresponding object parameters if:
>> - **exactly one is an implicit object member function with no ref-qualifier and the types of their object parameters ([dcl.fct]), after removing top-level references, are the same**, or
>> - their object parameters have the same type.

In this case, `#1` and `#2` satisfy the first bullet since the type of the implicit object parameter of `#1` is `Base&` while the type of the explicit object parameter of `#2` is `Base`, they have **the same type after removing the top-level references**.  Hence, `#2` can override`#1`?  Is it the intention after introducing *deducing this*?

### PR DESCRIPTION
**Pull request by jensmaurer**
including when overriding a virtual function.
Remove a prohibition focusing on 'declared virtual'
from [dcl.fct].

Fixes #5144

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -3881,6 +3881,21 @@ struct A {
 \end{example}
 
 \pnum
+A virtual function shall not be an explicit object member
+function\iref{dcl.fct}.
+\begin{example}
+\begin{codeblock}
+struct B {
+  virtual void g();             // \#1
+};
+struct D : B {
+  virtual void f(this D&);      // error: explicit object member function cannot be virtual
+  void g(this D&);              // overrides \#1; error: explicit object member function cannot be virtual
+};
+\end{codeblock}
+\end{example}
+
+\pnum
 The \grammarterm{ref-qualifier}, or lack thereof, of an overriding function
 shall be the same as that of the overridden function.
 
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -2997,6 +2997,10 @@ void f(int i) {
 \end{note}
 
 \pnum
+An allocation or deallocation function for a class
+shall not have an explicit object parameter\iref{dcl.fct}.
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: source/declarations.tex [hunk 1]
```diff
@@ -3578,7 +3578,7 @@
 A \grammarterm{member-declarator} with an explicit-object-parameter-declaration
 shall not include
 a \grammarterm{ref-qualifier} or a \grammarterm{cv-qualifier-seq} and
-shall not be declared \keyword{static} or \keyword{virtual}.
+shall not be declared \keyword{static}.
```
zygoloid, line 3581: Do we have the same problem with the implicit `static` for allocation functions?
```
struct A {
  // I'm not sure which parameter needs to be size_t, but
  // one of them does...
  void *operator new(this size_t a, size_t b);
};
```

jensmaurer: See separate commit.

source/classes.tex
```diff
@@ -6760,6 +6775,10 @@
 \end{codeblock}
 \end{note}
 
+\pnum
+An allocation or deallocation funtion for a class
```
cmeerw, line 6779:
```suggestion
An allocation or deallocation function for a class
```

jensmaurer:
CWG2553

PR #5149, opened 2021-12-14
[dcl.meaning.general] Clarify correspondence for block-scope friends

### LINKED ISSUE #5148: [dcl.meaning.general] p2.3 Correspond to a declaration that inhabits the block scope is not sufficient
[dcl.meaning.general] p2.3 says that
> Otherwise, the terminal name of E is not looked up. The declaration's target scope is the innermost enclosing namespace scope; if the declaration is contained by a block scope, **the declaration shall correspond to a declaration that inhabits the innermost block scope**.

Consider this example
````cpp
void fun(){
   struct A{
    friend void show();  // #1
   };
   void show();  // #2
}
````

`#1` does correspond to `#2` but this example is rejected by all implementations that correctly implement the c++20 standard. In that standard, we require the example to satisfy the following rules 

[class.friend] p11
> If a friend declaration appears in a local class ([class.local]) and the name specified is an unqualified name, a prior declaration is looked up without considering scopes that are outside the innermost enclosing non-class scope. For a friend function declaration, if there is no **prior declaration**, the program is ill-formed.   

the wording "correspond" does not guarantee the **order** of two declarations. It seems that we want the following rules to constrain this example 

> Otherwise, the terminal name of E is not looked up. The declaration's target scope is the innermost enclosing namespace scope; if the declaration is contained by a block scope, the declaration shall correspond to a **reachable**([module.reach#3]) declaration that inhabits the innermost block scope.

### PR DESCRIPTION
**Pull request by jensmaurer**
Fixes #5148

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -2548,7 +2548,7 @@ the target scope of the \grammarterm{declarator} is that scope.
 Otherwise, the terminal name of $E$ is not looked up.
 The declaration's target scope is the innermost enclosing namespace scope;
 if the declaration is contained by a block scope,
-the declaration shall correspond to a declaration
+the declaration shall correspond to a reachable\iref{module.reach} declaration
 that inhabits the innermost block scope.
 \end{itemize}
 

### CODE REVIEW

**Comment by jensmaurer** [comment]
@opensdh, see the linked issue. This fix seems obvious and desirable. Please have a look.

File: source/declarations.tex [hunk 1]
```diff
@@ -2548,7 +2548,7 @@
 Otherwise, the terminal name of $E$ is not looked up.
 The declaration's target scope is the innermost enclosing namespace scope;
 if the declaration is contained by a block scope,
-the declaration shall correspond to a declaration
+the declaration shall correspond to a reachable\iref{module.reach} declaration
```
opensdh, line 2551: Or just use English "an earlier", since we can't need to look into other translation units as for "precedes" or "reachable".

opensdh (approved):
Yes, this is obviously what was meant; I fell into the trap of imagining the check as happening immediately upon encountering the declaration.

source/declarations.tex
jensmaurer: Uh, but "earlier" also has a temporal meaning, which we don't want here.

jensmaurer:
@tkoeppe , this looks ready.

source/declarations.tex
opensdh: There is no time to apply here (although my initial mistake was more or less supposing that there was).

jensmaurer: Exactly, and that's why I don't want "earlier", which has a temporal aspect. In contrast, "reachable" has positional meaning (only), not temporal.

PR #5151, opened 2021-12-15
[std] Extend \iref to accept a list of cross-references

jensmaurer:


jensmaurer:
@tkoeppe , please have a look. This is for consistent application of \iref even with multiple cross-references.

tkoeppe:
The effect looks good, but I worry a little bit about creating some excessively complex implementation. Do you think this feature is worth the price?

jensmaurer:
Which price? This seems pretty straightforward in LaTeX3 terms.

Plus we get the benefit that we have a consistent handling linebreaks / spacing with xrefs for both "(1.2.3)" and "(1.2.3, 4.5.6)" cases.

jensmaurer:
@tkoeppe, rebased

tkoeppe:
OK. I've caught up on the history of LaTeX3 (it's come quite some way since I last followed it), and it seems that what you're doing here is at least somewhat normal and stable, and there's no reason to not use those new features.

Feel free to rebase and merge!

PR #5153, opened 2021-12-17
[std] Fix bad \grammarterms

jensmaurer:
Also adds a check. This essentially flags typos inside `\grammarterm`.


tkoeppe:
Very nice!

PR #5154, opened 2021-12-17
[atomics] Harmonize references to atomics operations tables

### LINKED ISSUE #5120: [atomics.types.int] Reference [tab:atomics.types.int.comp]
The table is currently alluded to like this:
> _This_ and _that_ is:
> | This | That |
> |---|---|
> | Table with _this_ | and _that_. |

I think the correspondence could be lost, as the WD is updated, if the table were to float somewhere not directly after the colon in the paragraph alluding to it.

There are 3 references to this table, so it's not a problem with the ISO Directives that requires each table to be referenced in the document.

jensmaurer:
Fixes #5120

### DIFF
--- a/source/atomics.tex
+++ b/source/atomics.tex
@@ -1218,7 +1218,7 @@ that differ from the primary template.
 
 \pnum
 The following operations perform arithmetic computations.
-The key, operator, and computation correspondence is identified
+The correspondence among key, operator, and computation is specified
 in \tref{atomic.types.int.comp}.
 
 \indexlibrarymember{fetch_add}{atomic_ref<\placeholder{integral}>}%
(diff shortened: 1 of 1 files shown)


source/atomics.tex
```diff
@@ -1340,7 +1340,7 @@
 
 \pnum
 The following operations perform arithmetic computations.
-The key, operator, and computation correspondence are identified
+The key, operator, and computation correspondence is identified
```
tkoeppe, line 1343: @jensmaurer: Instead of having these perpetual "guess the association" games, how about we say "The correspondence among key, operator, and computation is identified..."?

jensmaurer: Sure.

jwakely (approved):

tkoeppe (approved):

PR #5156, opened 2021-12-18
[format.arg] Fix the determination of T without cvref-qualifiers removed

hewillk:


jensmaurer:
This is not obviously an editorial fix. It seems the current wording would fall back to `handle(v)` for (e.g.) a `const int` argument type; with your change, it would instead initialize `value` with an int.

hewillk:
Oops, sorry for my negligence

hewillk:
> This is not obviously an editorial fix. It seems the current wording would fall back to `handle(v)` for (e.g.) a `const int` argument type; with your change, it would instead initialize `value` with an int.

It seems that using `handle` to deal with `const int` is not intentional. Is this an *acceptable* editorial fix or is it worth submitting an LWG?

CaseyCarter:
I believe this is LWG3631.

PR #5164, opened 2021-12-22
[macros] Use math font for ranges

### LINKED ISSUE #2556: Consider formatting ranges with math brackets
`\Range` and `\range` currently format the whole range, including the brackets and comma, as C++ code:

    \newcommand{\Range}[4]{\tcode{#1#3,\penalty2000{} #4#2}}
    \newcommand{\range}[2]{\Range{[}{)}{#1}{#2}}

But that's kinda weird, because something like `[eback(), gptr())` isn't actually a syntactically well-formed C++ code fragment; only the `eback()` and `gptr()` parts are. So I think it might make more sense to format the brackets (and comma) as math. What do you think?

jensmaurer:
Fixes #2556

ranges

### DIFF
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -476,7 +476,7 @@
 \newcommand{\defexposconceptnc}[1]{\ecname{#1}\indexconcept{\idxexposconcept{#1}|idxbfpage}\itcorr[-1]}  % macro length: 18
 
 %% Ranges
-\newcommand{\Range}[4]{\tcode{#1#3,\penalty2000{} #4#2}}
+\newcommand{\Range}[4]{\ensuremath{#1}\tcode{#3}, \penalty2000{} \tcode{#4}\ensuremath{#2}}
 \newcommand{\crange}[2]{\Range{[}{]}{#1}{#2}}
 \newcommand{\brange}[2]{\Range{(}{]}{#1}{#2}}
 \newcommand{\orange}[2]{\Range{(}{)}{#1}{#2}}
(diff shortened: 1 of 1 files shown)


tkoeppe:
Why is there a space in `[i, s)` and not in `[0,n)`? Is that due to the penalty? Can we ensure a certain minimal amount of space?

jensmaurer:
I think there's a bad interaction with the math environment for \countedrange that eats the space.

jensmaurer:
There were actually two spaces in `[i, s]`, it seems. And in math mode (i.e. for `\countedrange`), we get a small space after the comma.  Harmonizing to a small space everywhere looks like this:

ranges

@tkoeppe, what do you think?

jensmaurer:
@tkoeppe, your opinion is solicited here.

source/macros.tex
```diff
@@ -432,7 +432,7 @@
 \newcommand{\defexposconceptnc}[1]{\ecname{#1}\indexconcept{\idxexposconcept{#1}|idxbfpage}\itcorr[-1]}  % macro length: 18
 
 %% Ranges
-\newcommand{\Range}[4]{\tcode{#1#3,\penalty2000{} #4#2}}
+\newcommand{\Range}[4]{\ensuremath{#1}\tcode{#3}{,}\,\penalty2000{}\tcode{#4}\ensuremath{#2}}
```
tkoeppe, line 435: Should the comma also be in `\ensuremath`?

jensmaurer: Since math mode cannot see the adjacent stuff, I'm not sure it does anything.

I've done that now:

range

tkoeppe: It will at least pick the right maths font, even if that ends up being the same :-)

But yes, the interesting spacing aspects of maths mode probably don't come into play here.

Thanks!

jensmaurer:
Uh, that one still had build failures (overfull \hbox). I didn't want to address these until we had consensus on the approach.

PR #5170, opened 2021-12-23
[diff.stmt] Properly refer to function return types

jensmaurer:
Also annotate the 'void' and 'int' keywords.

Partially addresses #3908.

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -2261,7 +2261,7 @@ Besides,
 promising to return a value of a given type, and then not returning
 such a value, has always been recognized to be a questionable
 practice, tolerated only because very-old C had no distinction between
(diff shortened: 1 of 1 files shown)


source/compatibility.tex
```diff
@@ -2261,7 +2261,7 @@
 promising to return a value of a given type, and then not returning
 such a value, has always been recognized to be a questionable
 practice, tolerated only because very-old C had no distinction between
```
jwakely, line 2263: Why is "very-old" hyphenated? Is there a better way to refer to pre-ANSI C?

jensmaurer: I don't know. That's pre-existing, though.

PR #5178, opened 2022-01-02
[core] Replace 'enumerated type' with 'enumeration'

jensmaurer:
The term "enumerated type" is defined in [enumerated.types]
for use in the standard library, and is not synonymous with
"enumeration type".

Fixes #3977


jensmaurer:
This was approved by @zygoloid in #3977.

PR #5179, opened 2022-01-03
[expr] Cleanup for 'discarded-value expression'

jensmaurer:
Fixes #3909.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -471,15 +471,14 @@ following:
 Using an overloaded operator causes a function call; the
 above covers only operators with built-in meaning.
 \end{note}
-If the (possibly converted) expression is a prvalue,
-the temporary materialization conversion\iref{conv.rval} is applied.
+The temporary materialization conversion\iref{conv.rval} is applied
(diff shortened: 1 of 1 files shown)


source/expressions.tex
```diff
@@ -471,15 +471,14 @@
 Using an overloaded operator causes a function call; the
 above covers only operators with built-in meaning.
 \end{note}
-If the (possibly converted) expression is a prvalue,
-the temporary materialization conversion\iref{conv.rval} is applied.
+The temporary materialization conversion\iref{conv.rval} is applied
+if the (possibly converted) expression is a prvalue of object type.
```
tkoeppe, line 475: Does this affect whether a discarded prvalue is materialized?

E.g. with `T f();`, does `static_cast<void>(f());` materialize a `T` object?

JohelEGP: Yes. "Expression" above is discarded-value expression, and this is part of its definition.

tkoeppe: Ah I see, this is still talking about the inner expression, not the entire cast expression.

PR #5186, opened 2022-01-06
[temp.dep.general] p2.1 lacks a trialing 'or'

ChuanqiXu9:
[temp.dep.general]p2.1: http://eel.is/c++draft/temp.dep.general lacks a trailing or, which makes it a little bit confusing at the first sight.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -4709,7 +4709,7 @@ is an \grammarterm{unqualified-id} and
 \begin{itemize}
 \item
 any of the expressions in the \grammarterm{expression-list} is a pack
-expansion\iref{temp.variadic},
+expansion\iref{temp.variadic}, or
 \item
 any of the expressions
 or \grammarterm{braced-init-list}{s}


jwakely:
This changes "x, y, or z" into "x, or y, or z" which doesn't seem like an improvement. The conjunction is only needed before the last term.

jwakely:
Although since this is formatted as an itemized list, not just "x, y, or z" maybe it does help.

ChuanqiXu9:
It looks like not consistency with other part. The p2.4 in the same page is an example: http://eel.is/c++draft/temp.dep.general#2.4.

jensmaurer:
We are generally inconsistent with the placement of "and" and "or" in itemized lists. We have lots of examples where we do "x, y, or z" and we also have examples where we do "x, or y, or z".

We should decide what we want to do here and document that on the wiki for future incremental harmonization.

Suggestion: Prefer "x, y, or z" for itemized lists unless the individual items are so huge (e.g. because of sub-bullets) that the context might get lost.

jensmaurer (approved):

jwakely (commented):
Minor, but can we change the commit message to "trailing"

jensmaurer:
Of course.

PR #5191, opened 2022-01-07
[basic.lookup.unqual] Clarify conversion-function-id components lookup

languagelawyer:
What does `looked up in the same fashion as the \grammarterm{conversion-function-id}` mean when the `\grammarterm{conversion-function-id}` is not looked up? Like in declarations.
```c++
struct S
{
  using T = int;
  operator T(); // `operator T` is not looked up. What does it mean for T?
};

S::operator T() // same here
{ return {}; }
```

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -1777,13 +1777,15 @@ Unless otherwise specified,
 such a name undergoes unqualified name lookup from the point where it appears.
 
 \pnum
-An unqualified name that is a component name\iref{expr.prim.id.unqual} of
+Lookup for an unqualified name $U$
+that is a component name\iref{expr.prim.id.unqual} of
 a \grammarterm{type-specifier} or \grammarterm{ptr-operator} of
-a \grammarterm{conversion-type-id} is looked up in the same fashion
-as the \grammarterm{conversion-function-id} in which it appears.
-If that lookup finds nothing, it undergoes unqualified name lookup;
-in each case, only names
-that denote types or templates whose specializations are types are considered.
+a \grammarterm{conversion-type-id} $T$ considers only names
+that denote types or templates whose specializations are types.
+If $T$ appears in a \grammarterm{conversion-function-id} that is a qualified name
+whose lookup context\iref{basic.lookup.qual} is a class $S$,
+lookup for $U$ performs a search in the scope associated with $S$;
+if that lookup finds nothing, $U$ undergoes unqualified name lookup.
 \begin{example}
 \begin{codeblock}
 struct T1 { struct U { int i; }; };


jensmaurer:
@opensdh, your territory.

source/basic.tex
```diff
@@ -1777,12 +1777,13 @@
...
-An unqualified name that is a component name\iref{expr.prim.id.unqual} of
+If a \grammarterm{conversion-function-id} is a qualified name
+with lookup context $S$\iref{basic.lookup.qual},
+an unqualified name $U$ that is a component name\iref{expr.prim.id.unqual} of
 a \grammarterm{type-specifier} or \grammarterm{ptr-operator} of
-a \grammarterm{conversion-type-id} is looked up in the same fashion
-as the \grammarterm{conversion-function-id} in which it appears.
-If that lookup finds nothing, it undergoes unqualified name lookup;
-in each case, only names
+its \grammarterm{conversion-type-id} is looked up in $S$;
```
opensdh, line 1784: "is looked up in" is imprecise.  The formal equivalent to the previous would be to say "undergoes qualified lookup in _S_", but since none of the embellishments of qualified lookup can apply here we can follow the next paragraph and say

> …, lookup for an unqualified name that is a component name … performs a search in the scope associated with _S_.  Otherwise, or if that lookup finds nothing, it undergoes unqualified name lookup.  In each case, …

It might be profitable to first establish the "only names that denote" rule:

> Lookup for an unqualified name _U_ that is a component name of a *type-specifier* or *ptr-operator* of a *conversion-type-id* _T_ considers only names….  If _T_ appears in a *conversion-function-id* that is a qualified name…, lookup for _U_ performs…; if that lookup finds nothing, ….

If someone were feeling ambitious, they could actually combine the paragraphs by establishing two circumstances where a search happens with unqualified lookup as a backup, although you still need the restriction to types for this case.

tkoeppe:
@opensdh: What's your verdict -- is this an improvement, and, importantly, is this still editorial?

source/basic.tex
```diff
@@ -1777,13 +1777,15 @@
...
-as the \grammarterm{conversion-function-id} in which it appears.
-If that lookup finds nothing, it undergoes unqualified name lookup;
-in each case, only names
-that denote types or templates whose specializations are types are considered.
+a \grammarterm{conversion-type-id} $T$ considers only names
+that denote types or templates whose specializations are types.
+If $T$ appears in a \grammarterm{conversion-function-id} that is a qualified name
+whose lookup context\iref{basic.lookup.qual} is a class $S$,
+lookup for $U$ performs a search in the scope associated with $S$;
+if that lookup finds nothing, $U$ undergoes unqualified name lookup.
```
opensdh, line 1788: I think it would be clearer to say "[&hellip;]; **otherwise, or** if that lookup finds nothing, [&hellip;]".  It's true that the general "Unless otherwise specified" from the previous paragraph ought to apply, but it tends to sound like the alternative is that the name is not looked up at all.

opensdh:
I didn't know I issued verdicts.  I think this is definitely an improvement (and I made my one comment about it); it's certainly not strictly editorial, although one could argue that it merely stipulates the obvious thing in a place where the existing text does not specify any behavior.  It might fall into the "CWG glances at a merge request during a teleconference" category.

tkoeppe:
Thanks, that's precisely what I needed to hear! @jensmaurer: could you kindly ask the CWG chair to clear this proposed change with his people?

source/basic.tex
languagelawyer: > It's true that the general "Unless otherwise specified" from the previous paragraph ought to apply, but it tends to sound like the alternative is that the name is not looked up at all.

To me, the previous paragraph clearly applies to unqualified `\grammarterm{conversion-function-id}` :/
May I not apply the proposed change, please?

opensdh: This paragraph is about an unqualified _U_, looked up separately from the *conversion-function-id* that contains it.  I want to avoid the (nonsensical) suggestion that, if _T_ does not appear in the given context, _U_ is not looked up at all.

jensmaurer:
There is still a request for further adjustment by @opensdh.

source/basic.tex
languagelawyer: > This paragraph is about an unqualified _U_, looked up separately from the _conversion-function-id_ that contains it.

Yeah, right, but whatever, the previous paragraph applies to other unqualified names (which are not mentioned in this paragraphs) (and in any other paragraphs-exceptions).

> I want to avoid the (nonsensical) suggestion that, if _T_ does not appear in the given context, _U_ is not looked up at all.

I don't see how is it possible. I mean, not only the previous paragraph, but also the previous sentence suggest that such _U_ are looked up:
> Lookup for an unqualified name _U_ that is ... of a _conversion-type-id_ _T_ considers only names that ... If _T_ appears in a _conversion-function-id_ ...

If _U_ is not looked up when _T_ doesn't appear in _conversion-function-id_, there is no reason these sentences are in this and not the opposite order

opensdh: I don't insist on a change; it just strikes me as a trip hazard for the reader to say "If X, then Y; if Y does nothing, Z." and have Z also happen when X doesn't pertain.

PR #5013, opened 2021-10-08
P0288R9 move_only_function

jensmaurer:
Fixes #4977
Fixes cplusplus/papers#400
Partially addresses #4960

PR #5198, opened 2022-01-12
move_only_function takes one signature

zhihaoy:
The pack (`Args...`) in the specializations does not need the main template to be variadic. A signature is just one type parameter.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -14129,7 +14129,7 @@ namespace std {
     bool operator==(const function<R(ArgTypes...)>&, nullptr_t) noexcept;
 
   // \ref{func.wrap.move}, move only wrapper
-  template<class... S> class move_only_function;        // \notdef
+  template<class S> class move_only_function;        // \notdef
   template<class R, class... ArgTypes>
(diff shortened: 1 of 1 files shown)


JohelEGP:
See https://github.com/cplusplus/draft/pull/5013#discussion_r731268417.

zhihaoy:
> See [#5013 (comment)](https://github.com/cplusplus/draft/pull/5013#discussion_r731268417).

Continued the thread there. The specification (synopsis) is a non-binding declaration. If an implementation has ABI concern, it can declare the main template in whatever way compiles.

JohelEGP:
> Continued the thread there. The specification (synopsis) is a non-binding declaration. If an implementation has ABI concern, it can declare the main template in whatever way compiles.

I wonder if they considered adding a (normative) encouragement to make it variadic for implementations whose ABI would change. Not like that'd stop implementations from making it non-variadic, since it's just an encouragement. And then we'd have to get another template that's actually variadic, just like it happened with the lock guards. So it seems better to just make it variadic from the get-go.

zhihaoy:
I see; it's not editorial. `using t = move_only_function<X, Y>;` is legal code.

PR #5202, opened 2022-01-12
[func.memfn] Correct target object by fixing typo

zhihaoy:
Pointer to member has no function call expression (only pointer to member expression has). `call_args` should mean the args passed to the call wrapper.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -16013,7 +16013,7 @@ with call pattern \tcode{invoke(pmd, call_args...)}, where
 \tcode{pmd} is the target object of \tcode{fn} of type \tcode{R T::*}
 direct-non-list-initialized with \tcode{pm}, and
 \tcode{call_args} is an argument pack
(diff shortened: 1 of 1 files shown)


jensmaurer:
@jwakely, please have a look.

languagelawyer:
I'd [also] say
```diff
- A simple call wrapper\iref{func.def} \tcode{fn}
+ A simple call wrapper\iref{func.require} \tcode{fn}
```
because it is defined there.

@jensmaurer why the standard doesn't refer to definitions by their names? Aren't these mismatches annoying?
Is there some technical difficulty with tracking the subclause of a definition?

jwakely (approved):

jensmaurer:
@languagelawyer , there is no technical difficulty, and I've recently started to use term.xxx labels for defined terms.

Example: `\label{term.odr.use}`

But there is no attempt at bulk conversion. I'll address your particular case in a separate pull request.

jensmaurer (approved):

PR #5207, opened 2022-01-14
[util.smartptr.atomic] Add cross-reference to <memory> header

### LINKED ISSUE #5206: [util.smartptr.atomic] Add a note mentioning <memory>
The `atomic<shared_ptr<T>>` and `atomic<weak_ptr<T>>` partial specializations are declared by `<memory>`, but are specified in Clause 31 [atomics]. Nothing in that Clause every mentions `<memory>`. It's not at all obvious which header you need for those specializations, unless you happen to find their declarations on the eighth page of of the `<memory>` synopsis in [memory.syn].

I think we should have a note in [util.smartptr.atomic] saying they're provided by `<memory>`.

jensmaurer:
Fixes #5206

### DIFF
--- a/source/atomics.tex
+++ b/source/atomics.tex
@@ -2735,6 +2735,9 @@ Equivalent to: \tcode{return fetch_sub(1) - 1;}
 \pnum
 The library provides partial specializations of the \tcode{atomic} template
 for shared-ownership smart pointers\iref{smartptr}.
+\begin{note}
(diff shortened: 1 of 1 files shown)


jensmaurer:
@jwakely , at your service.

jwakely (approved):

jwakely:
That was quick, thanks!

PR #5211, opened 2022-01-16
[macros] In \ucode, align size of digits and letters

jensmaurer:


jensmaurer:
The + in "U+" looks oversized to me.

unicode

tkoeppe:
Yes, let's keep the `+` from what we had before -- I was glossing over that!

Thank you!

jensmaurer:
Fixed.

PR #5212, opened 2022-01-18
Less redundant 'typename' in type-id-only contexts

zhihaoy:
- default argument of a _type-parameter_ of a template cleared
- _defining-type-id_ cleared
- _type-id_ of a `static_cast`, `const_cast`, `reinterpret_cast`, or `dynamic_cast` cleared
- _trailing-return-type_ no instance
- _new-type-id_ probably no instance

~~I'm not very sure about the `static_cast<T::type*>` ones.~~

Closes: #3637

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -3355,7 +3355,7 @@ The following exposition-only alias template may appear in deduction guides for 
 
 \begin{codeblock}
 template<class InputIterator>
-  using @\placeholdernc{iter-value-type}@ = typename iterator_traits<InputIterator>::value_type;  // \expos
+  using @\placeholdernc{iter-value-type}@ = iterator_traits<InputIterator>::value_type;  // \expos
 \end{codeblock}
 
 \rSec2[array.syn]{Header \tcode{<array>} synopsis}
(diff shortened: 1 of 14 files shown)


jwakely:
> I'm not very sure about the `static_cast<T::type*>` ones.

And that's exactly why LWG haven't expressed any desire to rush into this.

One of the major implementations hasn't even implemented the feature yet.

zhihaoy:
> > I'm not very sure about the `static_cast<T::type*>` ones.
> 
> And that's exactly why LWG haven't expressed any desire to rush into this.
> 
> One of the major implementations hasn't even implemented the feature yet.

GCC and MSVC accept this. Clang's implementation under review has test cases for this case https://reviews.llvm.org/D53847:

```cpp
dynamic_cast<T::tclass *>(new B)
const_cast<const T::type *>(Ptr)
```

I think I understand what does "smallest enclosing" mean here now; in http://wg21.link/p1787 the term has been changed into "[terminal name](http://eel.is/c++draft/expr.prim.id.unqual#2),"

> The terminal name of a construct is the component name of that construct that appears lexically last.

source/containers.tex
```diff
@@ -6548,13 +6548,13 @@
 \begin{codeblock}
 template<class InputIterator>
   using @\placeholder{iter-value-type}@ =
-    typename iterator_traits<InputIterator>::value_type;                // \expos
+    iterator_traits<InputIterator>::value_type;                         // \expos
 template<class InputIterator>
   using @\placeholder{iter-key-type}@ = remove_const_t<
     typename iterator_traits<InputIterator>::value_type::first_type>;   // \expos
```
zhihaoy, line 6554: http://wg21.link/p2165r3 will get rid of the other `typename`s in this code block.

AlisdairM:
See also https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p2150r0.html
My intent is to revisit this paper as an editorial NB comment against C++23, where LWG review each option to omit typename in different contexts, before setting the editorial policy.  I expect this will be too disruptive for C++23 at this point, but hope to get the policy review to apply to the C++26 Draft immediately it is created, before any new work starts.

tkoeppe:
Could you please split this up into smaller commits or PRs? Please reopen if and when you want to resume this.

PR #5217, opened 2022-01-20
[stmt.return] Clarify flowing off the end of a function

### LINKED ISSUE #5215: [editoral] the ambiguity in [stmt.return] p4
[stmt.return] p4 says: 
> Otherwise, flowing off the end of a function other than main or a coroutine results in undefined behavior.

It seems the sentence can be read as two meanings, which are: 
> - flowing off the end of a function other than **(main or a coroutine)** results in undefined behavior.
> - flowing off the end of a function other than main or flowing off the end of a coroutine results in undefined behavior. 

we can infer that the second one is a wrong interpretation but this is based on that we get the clue after reading [dcl.fct.def.coroutine] p11.

jensmaurer:
Fixes #5215

### DIFF
--- a/source/statements.tex
+++ b/source/statements.tex
@@ -849,7 +849,7 @@ a destructor, or
 a non-coroutine function with a \cv{}~\keyword{void} return type is
 equivalent to a \tcode{return} with no operand.
 Otherwise, flowing off the end of a function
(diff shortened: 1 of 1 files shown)


jensmaurer:
@jwakely , does that help your reading skills?

jwakely (approved):
Yes, thanks. Sorry for being slow

tkoeppe:
Thanks everyone, this is an improvement.

PR #5218, opened 2022-01-20
[class.mem.general] Add cross-reference for 'layout-compatible type'

### LINKED ISSUE #5216: [class.mem.general] only second use of "layout-compatible types" has xref to definition
[class.mem.general] p23 refers to "layout-compatible types" without a cross-reference, then p25 refers to it again and cross-references [basic.types.general].

Should the xref move to p23? Or be in both p23 and p25?

jensmaurer:
Fixes #5216


jensmaurer:
@jwakely , we also have the xref in p24, so I've simply added it to p23 as well.

jwakely (approved):
Thanks, LGTM

PR #5227, opened 2022-01-21
[std] Replace hyphen with period in labels

### LINKED ISSUE #5225: inconsistent use of periods versus hyphens in [defns]
In some labels in defns, hyphens in a term translate into hyphens in the label:
* [defns.direct-non-list-init]
* [defns.expression-equivalent]

In other labels, they become periods:
* [defns.prog.def.spec]
* [defns.prog.def.type]

Finally we have some cases where the noun is dropped when abbreviating to form the label; in these cases we seem to consistently use a period rather than a hyphen:
* [defns.cond.supp]
* [defns.impl.defined]
* [defns.locale.specific]
* [defns.well.formed]

(That last category might make sense, because English style guides often say to omit the hyphens in compound adjectives when used without a noun, but we don't do that in the standard, eg: "the behavior is implementation-defined" not "the behavior is implementation defined".)

Some consistency here would be nice. Maybe the first two examples above should change from hyphens to periods?

jensmaurer:
and add a check.

Fixes #5225

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -466,7 +466,7 @@ A structured binding is odr-used if it appears as a potentially-evaluated expres
 \pnum
 \tcode{*\keyword{this}} is odr-used if \keyword{this} appears as a potentially-evaluated expression
 (including as the result of the implicit transformation in the body of a non-static
-member function~(\ref{class.mfct.non-static})).
+member function\iref{class.mfct.non.static}).
 
 \pnum
 A virtual member
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -555,7 +555,7 @@ in which case it is a \defnadj{static}{member} (see~\ref{class.static})
 of the class.
 Any other data member or member function is a \defnadj{non-static}{member}
 (a \defnadj{non-static}{data member} or
-\defnadj{non-static}{member function}~(\ref{class.mfct.non-static}), respectively).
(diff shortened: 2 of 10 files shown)


jensmaurer:
@zygoloid, at your service.

zygoloid (approved):
Thanks, I like it.

@tkoeppe How do you feel about requiring labels to match `/[0-9a-z.]*/`?

tkoeppe:
Are you referring to (visible) stable labels or to the internal term labels?

jensmaurer:
@tkoeppe, this issue is dealing with (visible) stable section labels, only. So, the question is: Are fine with hard-limiting (visible) stable section to match /[0-9a-z.]*/

tkoeppe:
I'm not fully convinced byt the resulting restrictions, since this forces us to conflate mere compound words with the hierarchical use of dots, e.g. `foo.bar`, `foo.bar.general` (subsection) would have to contend with `foo.bar.tender` (sibling section).

Practically, though, a) this particular change is very small and I have no issue with the handful of labels touched here, and b) we (or at least Jens's script) already seem to be claiming that the existing use of hyphens is "bad", so it seems like we already made that decision.

So consider this a non-enthusiastic approval.

jensmaurer:
This needs a rebase, though.

PR #5228, opened 2022-01-21
[basic.life] Add cross-reference for union member lifetime

### LINKED ISSUE #5219: [basic.life] p1 omits a cross-reference to [class.copy.assign] for union member case
> The lifetime of an object of type T begins when:
>> - [...]
>
> except that **if the object is a union member or subobject thereof**, its lifetime **only begins** if that union member is the initialized member in the union (**[dcl.init.aggr]**, **[class.base.init]**), or as described in **[class.union] and [class.copy.ctor]**, and except as described in [allocator.members].  

[class.copy.assign] p13 explicitly states
> If the source and destination of the assignment are not the same object, then for each object nested within ([intro.object]) the object that is the source of the copy, a corresponding object o nested within the destination is created, and **the lifetime of o begins** before the copy is performed.  

So, [basic.life] p1 should be: or as described in [class.union], [class.copy.ctor], and [class.copy.assign].  

***
Incidentally, I think [basic.life] should cover the complete story regards of when/how the lifetime of an object will be begun, as its subject is regarding lifetime. For instance, these operations are phrased as implicitly creating objects, which can also start the lifetimes for objects.

jensmaurer:
Fixes #5219

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3352,7 +3352,8 @@ The lifetime of an object of type \tcode{T} begins when:
 except that if the object is a union member or subobject thereof,
 its lifetime only begins if that union member is the
 initialized member in the union\iref{dcl.init.aggr,class.base.init},
-or as described in \ref{class.union} and \ref{class.copy.ctor},
+or as described in
+\ref{class.union}, \ref{class.copy.ctor}, and \ref{class.copy.assign},
 and except as described in \ref{allocator.members}.
 The lifetime of an object \placeholder{o} of type \tcode{T} ends when:
 \begin{itemize}


jensmaurer:
@zygoloid , I think this is right, but I'd appreciate your review.

languagelawyer:
Usually, when a standard defines a general rule and, in other sections, has rules overriding for some specific cases, the generic rule does not (normatively) enumerate exceptions. Why [basic.life]/1 shall be an exception?

xmh0511:
> Usually, when a standard defines a general rule and, in other sections, has rules overriding it for some specific cases, the generic rule does not (normatively) enumerate exceptions. Why [basic.life]/1 shall be an exception?

Because the general rule uses the wording: **only begins**, and it also didn't say such as *except otherwise specified*. Thus, that general rule should be responsible to enumerate all the cases.

source/basic.tex
```diff
@@ -3352,7 +3352,8 @@
 except that if the object is a union member or subobject thereof,
 its lifetime only begins if that union member is the
 initialized member in the union\iref{dcl.init.aggr,class.base.init},
-or as described in \ref{class.union} and \ref{class.copy.ctor},
+or as described in
+\ref{class.union}, \ref{class.copy.ctor}, and \ref{class.copy.assign},
```
zygoloid, line 3356: Of these cases, only [class.copy.ctor] (calling a union copy constructor) starts the lifetime of the union member subobject. Both the [class.union] (assigning to a union member) and [class.copy.assign] (assigning to a union) cases create a new object that transparently replaces the union member. So I think only [class.copy.ctor] is in scope here, although the reason for that is subtle.

zygoloid: I suppose another viewpoint is that, because the newly-created object transparently replaces a union member subobject, it is itself a member subobject ([intro.object]p2), and so it is in the scope of this rule, in which case this change seems appropriate. Either way, we should be consistent: either both [class.union] and [class.copy.assign] should be listed here or neither of them.

tkoeppe:
Thank you, @zygoloid! @jensmaurer, I suppose we'll take this as-is then? I'm sure I don't appreciate all the subtleties here, so I'll take your word for it.

jensmaurer:
@tkoeppe, I think we should take this as-is.

PR #5242, opened 2022-01-29
[derivation] Remove 'basic integral type' from footnote

jensmaurer:
Also switch footnote to note.

Partially addresses #5240


jensmaurer:
@jwakely , this fixes the footnote.

jwakely (approved):

PR #5244, opened 2022-02-01
[iterator.operations] clarify preconditions on std::distance

RedBeard0531:
The prior wording implied that the two sides of the "or" were independent, so as long as `last` was reachable from `first`, `first` didn't need to be reachable from `last` even if passing a `RandomAccessIterator`. It seems clear that it is really trying to express two different possible preconditions depending on whether a `RandomAccessIterator` is passed or not. I used the same phrasing as in the "Effects:" clause to separate the cases.

Alternatively, this precondition seems redundant with http://eel.is/c++draft/iterators#iterator.requirements.general-12, so maybe it should just be removed? This seems to be the only function in [iterators], [ranges], and [algorithms] that has an explicit precondition that iterators passed to it are reachable.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -2806,10 +2806,10 @@ template<class InputIterator>
 \begin{itemdescr}
 \pnum
 \expects
-\tcode{last} is reachable from \tcode{first}, or
-\tcode{InputIterator} meets
-the \oldconcept{RandomAccessIterator} requirements and
-\tcode{first} is reachable from \tcode{last}.
+If \tcode{InputIterator} meets the \oldconcept{RandomAccessIterator} requirements,
+\tcode{first} is reachable from \tcode{last}; otherwise,
+\tcode{last} is reachable from \tcode{first}.
+
 
 \pnum
 \effects


jwakely:
Reachable is defined in terms of `++`, so it's wrong to require that `first` is reachable from `last` for all random access iterators.

jwakely:
> The prior wording implied that the two sides of the "or" were independent, so as long as last was reachable from first, first didn't need to be reachable from last even if passing a RandomAccessIterator. 

Which is correct. Unless they are equal (in which case they are both reachable from the other by zero applications of `++`) then it's impossible for both to be reachable (except in the special case of a circular buffer with no end iterator?)

RedBeard0531:
Ahh, I think I see now. It is trying to make `distance(vector.end(), vector.begin())` well defined, without making `distance(list.end(), list.begin())` well defined. Maybe something like this would make that more clear?

> Either `[first, last)` is a valid range, or `[last, first)` is a valid range and `InputIterator`  meets the _Cpp17RandomAccessIterator_ requirements

CaseyCarter:
> Maybe something like this would make that more clear?
> 
> > Either `[first, last)` is a valid range, or `[last, first)` is a valid range and `InputIterator`  meets the _Cpp17RandomAccessIterator_ requirements

"Either" suggests these two cases are exclusive, which suggests to me that `distance(x, x)` might be invalid since both `[first, last)` and `[last, first)` are valid ranges. I _do_ think that replacing "`woof` is reachable from `meow`" with "`[meow, woof)` is a valid range" is an improvement. Could you live with this sentence without the "Either" at the beginning?

tkoeppe:
@RedBeard0531 Ping :-)

PR #5246, opened 2022-02-01
Fix order of std-format-spec field descriptions.

mordante:
Moves the wording describing the zero-padding before the description of
the width; matching the order of the fields in the std-format-spec.

The swapped order was introduced in the initial <format> paper
  P0645R10 Text formatting
Since both fields had one paragraph description it wasn't too noticeable.
  P1868R2 🦄 width: clarifying units of width and precision in std::format
expanded the wording of the width. Now it's not so easy to find the
description of the zero-padding field.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -14266,6 +14266,26 @@ In addition, for \tcode{g} and \tcode{G} conversions,
 trailing zeros are not removed from the result.
 
 \pnum
+A zero (\tcode{0}) character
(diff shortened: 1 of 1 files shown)


tkoeppe:
Could you please rebase this?

tkoeppe:
@jwakely, @CaseyCarter: thoughts on this swap?

jwakely (approved):
Seems reasonable.

PR #5254, opened 2022-02-06
[expr.prim.lambda.capture] Fix 'potentially references' for members

### LINKED ISSUE #5243: [expr.prim.lambda.capture] vs. [basic.def.odr] in their use of «names»/«named by»
[expr.prim.lambda.capture]/7:
> an _id-expression_ that names one or more non-static class members and does not form a pointer to member ([expr.unary.op]) potentially references `*this`.
> [Note 3: This occurs even if overload resolution selects a static member function for the _id-expression_. — end note]

The issue is that the normative part (plural «more non-static class members») and the Note are not compatible with [basic.def.odr], which says that
> A function is _named by_ an expression or conversion as follows:
> — A function is named by an expression or conversion if it is the selected member of an overload set

P.S. Why the first bullet in [expr.prim.lambda.capture]/7 uses semicolon? I think it should be just split into two bullets.

languagelawyer:
This editorially limits capture of `*this` to only when a non-static member of the [current class](http://eel.is/c++draft/expr.prim.this#def:class,current) is found. Currently, naming a non-static member of any unrelated class potentially references `*this`.

I'd wait for #5210 to say `\termref{current class}`.

Fixes #5243

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -2265,11 +2265,14 @@ an expression potentially references local entities as follows:
 \begin{itemize}
 \item
 An \grammarterm{id-expression} that names a local entity
-potentially references that entity;
-an \grammarterm{id-expression} that names
-one or more non-static class members
-and does not form a pointer to member\iref{expr.unary.op}
-potentially references \tcode{*\keyword{this}}.
+potentially references that entity.
+
+\item
+An \grammarterm{id-expression}
+that is not used to form a pointer to member\iref{expr.unary.op}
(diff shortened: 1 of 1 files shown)


tkoeppe:
Are you still working on this?

languagelawyer:
@tkoeppe we still do not have #5210 merged. I can rewrite w/o it

tkoeppe:
Ah yes, no worries, we can try to do the terms first!

PR #5276, opened 2022-02-09
P0323R12 std::expected

### LINKED ISSUE #5261: [2022-02 LWG Motion 2] P0323R12 std::expected
[P0323R12](https://wg21.link/p0323r12)

jensmaurer:
- Introduce "General" subclauses to avoid hanging paragraphs.
- Introduce "Returns: *this" to avoid duplication.

Fixes #5261 
Fixes cplusplus/papers#254

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -6470,7 +6470,7 @@ bool is_string(const any& operand) {
 \rSec2[expected.general]{In general}
 
 \pnum
-This subclause describes the class template \tcode{expected}
+Subclause \ref{expected} describes the class template \tcode{expected}
 that represents expected objects.
 An \tcode{expected<T, E>} object holds
 an object of type \tcode{T} or an object of type \tcode{unexpected<E>} and
@@ -6485,7 +6485,7 @@ namespace std {
   // \ref{expected.un.object}, class template \tcode{unexpected}
   template<class E> class unexpected;
 
-  // \ref{expected.bad}, class \tcode{bad_expected_access}
+  // \ref{expected.bad}, class template \tcode{bad_expected_access}
   template<class E> class bad_expected_access;
 
   // \ref{expected.bad.void}, specialization for \tcode{void}
@@ -6510,7 +6510,7 @@ namespace std {
 \rSec3[expected.un.general]{General}
 
 \pnum
-This subclause describes the class template \tcode{unexpected}
+Subclause \ref{expected.unexpected} describes the class template \tcode{unexpected}
 that represents unexpected objects stored in \tcode{expected} objects.
 
 \rSec3[expected.un.object]{Class template \tcode{unexpected}}
@@ -6595,7 +6595,7 @@ Any exception thrown by the initialization of \exposid{val}.
 \indexlibraryctor{unexpected}%
 \begin{itemdecl}
 template<class... Args>
-  constexpr explicit unexpected(in_place_t, Args&&...);
+  constexpr explicit unexpected(in_place_t, Args&&... args);
 \end{itemdecl}
 
 \begin{itemdescr}
@@ -6616,7 +6616,7 @@ Any exception thrown by the initialization of \exposid{val}.
 \indexlibraryctor{unexpected}%
 \begin{itemdecl}
 template<class U, class... Args>
-  constexpr explicit unexpected(in_place_t, initializer_list<U>, Args&&...);
+  constexpr explicit unexpected(in_place_t, initializer_list<U> il, Args&&... args);
 \end{itemdecl}
 
 \begin{itemdescr}
@@ -6627,7 +6627,7 @@ template<class U, class... Args>
 \pnum
 \effects
 Direct-non-list-initializes
-\exposid{val} with \tcode{std::forward<Args>(args)...}.
+\exposid{val} with \tcode{il, std::forward<Args>(args)...}.
 
 \pnum
 \throws
@@ -6874,7 +6874,7 @@ namespace std {
 
     // \ref{expected.object.swap}, swap
     constexpr void swap(expected&) noexcept(@\seebelow@);
-    friend constexpr void swap(expected&, expected&) noexcept(@\seebelow@);
+    friend constexpr void swap(expected& x, expected& y) noexcept(noexcept(x.swap(y)));
 
     // \ref{expected.object.obs}, observers
     constexpr const T* operator->() const noexcept;
@@ -6897,7 +6897,7 @@ namespace std {
     template<class U> constexpr T value_or(U&&) &&;
 
     // \ref{expected.object.eq}, equality operators
-    template<class T2, class E2>
+    template<class T2, class E2> requires (!is_void_v<T2>)
       friend constexpr bool operator==(const expected& x, const expected<T2, E2>& y);
     template<class T2>
       friend constexpr bool operator==(const expected&, const T2&);
@@ -7044,7 +7044,7 @@ Any exception thrown by the initialization of \exposid{val} or \exposid{unex}.
 
 \pnum
 \remarks
-The exception specification is
+The exception specification is equivalent to
 \tcode{is_nothrow_move_constructible_v<T> \&\&
 is_nothrow_move_constructible_v<E>}.
 
@@ -7185,7 +7185,7 @@ Let \tcode{GF} be \tcode{const G\&} for the first overload and
 
 \pnum
 \effects
-Direct-non-list-initializes \exposid{unex} with \tcode{std::forward<GF>(e.error())}.
+Direct-non-list-initializes \exposid{unex} with \tcode{std::forward<GF>(e.value())}.
 
 \pnum
 \ensures
@@ -7383,7 +7383,7 @@ This operator is defined as deleted unless:
 \item
 \tcode{is_copy_constructible_v<E>} is \tcode{true} and
 \item
-\tcode{is_nothrow_move_constructible_v<E> || is_nothrow_move_constructible_v<T>}
+\tcode{is_nothrow_move_constructible_v<T> || is_nothrow_move_constructible_v<E>}
 is \tcode{true}.
 \end{itemize}
 \end{itemdescr}
@@ -7434,7 +7434,7 @@ equivalent to: \tcode{has_val = rhs.has_value(); return *this;}
 
 \pnum
 \remarks
-The exception specification is
+The exception specification is equivalent to:
 \begin{codeblock}
 is_nothrow_move_assignable_v<T> && is_nothrow_move_constructible_v<T> &&
 is_nothrow_move_assignable_v<E> && is_nothrow_move_constructible_v<E>
@@ -7609,7 +7609,7 @@ See \tref{expected.object.swap}.
 \begin{floattable}{\tcode{swap(expected\&)} effects}{expected.object.swap}
 {lx{0.35\hsize}x{0.35\hsize}}
 \topline
-& \chdr{\tcode{this->has_value()}} & \rhdr{\tcode{!has_value()}} \\ \capsep
+& \chdr{\tcode{this->has_value()}} & \rhdr{\tcode{!this->has_value()}} \\ \capsep
 \lhdr{\tcode{rhs.has_value()}} &
   equivalent to: \tcode{using std::swap; swap(\exposid{val}, rhs.\exposid{val});} &
   calls \tcode{rhs.swap(*this)} \\
@@ -7654,7 +7654,7 @@ Any exception thrown by the expressions in the \Fundescx{Effects}.
 
 \pnum
 \remarks
-The exception specification is:
+The exception specification is equivalent to:
 \begin{codeblock}
 is_nothrow_move_constructible_v<T> && is_nothrow_swappable_v<T> &&
 is_nothrow_move_constructible_v<E> && is_nothrow_swappable_v<E>
@@ -7809,7 +7809,7 @@ template<class U> constexpr T value_or(U&& v) const&;
 \pnum
 \mandates
 \tcode{is_copy_constructible_v<T>} is \tcode{true} and
-\tcode{is_convertible<U, T>} is \tcode{true}.
+\tcode{is_convertible_v<U, T>} is \tcode{true}.
 
 \pnum
 \returns
@@ -7825,7 +7825,7 @@ template<class U> constexpr T value_or(U&& v) &&;
 \pnum
 \mandates
 \tcode{is_move_constructible_v<T>} is \tcode{true} and
-\tcode{is_convertible<U, T>} is \tcode{true}.
+\tcode{is_convertible_v<U, T>} is \tcode{true}.
 
 \pnum
 \returns
@@ -7855,7 +7855,7 @@ otherwise \tcode{x.error() == y.error()}.
 
 \indexlibrarymember{operator==}{expected}%
 \begin{itemdecl}
-template<class T2> constexpr bool operator==(const expected& x, const T2& v);
+template<class T2> friend constexpr bool operator==(const expected& x, const T2& v);
 \end{itemdecl}
 
 \begin{itemdescr}
@@ -7874,7 +7874,7 @@ its result is convertible to \tcode{bool}.
 
 \indexlibrarymember{operator==}{expected}%
 \begin{itemdecl}
-template<class E2> constexpr bool operator==(const expected& x, const unexpected<E2>& e);
+template<class E2> friend constexpr bool operator==(const expected& x, const unexpected<E2>& e);
 \end{itemdecl}
 
 \begin{itemdescr}
@@ -7938,7 +7938,7 @@ public:
 
   // \ref{expected.void.swap}, swap
   constexpr void swap(expected&) noexcept(@\seebelow@);
-  friend constexpr void swap(expected&, expected&) noexcept(@\seebelow@);
+  friend constexpr void swap(expected& x, expected& y) noexcept(noexcept(x.swap(y)));
 
   // \ref{expected.void.obs}, observers
   constexpr explicit operator bool() const noexcept;
@@ -8265,7 +8265,7 @@ Otherwise, equivalent to \tcode{\exposid{unex} = rhs.error()}.
 
 \pnum
 \remarks
-The exception specification is
+The exception specification is equivalent to
 \tcode{is_nothrow_move_constructible_v<E> \&\& is_nothrow_move_assignable_v<E>}.
 
 \pnum
@@ -8343,7 +8343,7 @@ See \tref{expected.void.swap}.
 \begin{floattable}{\tcode{swap(expected\&)} effects}{expected.void.swap}
 {lx{0.35\hsize}x{0.35\hsize}}
 \topline
-& \chdr{\tcode{this->has_value()}} & \rhdr{\tcode{!has_value()}} \\ \capsep
+& \chdr{\tcode{this->has_value()}} & \rhdr{\tcode{!this->has_value()}} \\ \capsep
 \lhdr{\tcode{rhs.has_value()}} &
   no effects &
   calls \tcode{rhs.swap(*this)} \\
@@ -8367,7 +8367,7 @@ Any exception thrown by the expressions in the \Fundescx{Effects}.
 
 \pnum
 \remarks
-The exception specification is
+The exception specification is equivalent to
 \tcode{is_nothrow_move_constructible_v<E> \&\& is_nothrow_swappable_v<E>}.
 \end{itemdescr}
 
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -16,6 +16,7 @@ These utilities are summarized in \tref{utilities.summary}.
 \ref{optional}              & Optional objects                  & \tcode{<optional>}    \\ \rowsep
 \ref{variant}               & Variants                          & \tcode{<variant>}     \\ \rowsep
 \ref{any}                   & Storage for any type              & \tcode{<any>}         \\ \rowsep
+\ref{expected}              & Expected objects                  & \tcode{<expected>}    \\ \rowsep
 \ref{bitset}                & Fixed-size sequences of bits      & \tcode{<bitset>}      \\ \rowsep
 \ref{memory}                & Memory                            & \tcode{<cstdlib>}, \tcode{<memory>} \\ \rowsep
 \ref{smartptr}              & Smart pointers                    & \tcode{<memory>}      \\ \rowsep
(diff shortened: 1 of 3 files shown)


source/utilities.tex
```diff
@@ -6463,6 +6464,2036 @@
...
+\tcode{is_nothrow_move_constructible_v<T> \&\&
+is_nothrow_move_constructible_v<E>}.
+
+\pnum
+This constructor is trivial if
+\begin{itemize}
+\item
+is_trivially_move_constructible_v<T> is \tcode{true}; and
+\item
+is_trivially_move_constructible_v<E> is \tcode{true}.
```
Mick235711, line 7057:
```suggestion
\tcode{is_trivially_move_constructible_v<T>} is \tcode{true}; and
\item
\tcode{is_trivially_move_constructible_v<E>} is \tcode{true}.
```

source/utilities.tex
```diff
@@ -6463,6 +6464,2036 @@
...
+namespace std {
+  template<class E>
+  class bad_expected_access : public bad_expected_access<void> {
+  public:
+    explicit bad_expected_access(E);
+    const char* what() const noexcept override;
+    E& error() & noexcept;
+    const E& error() const& noexcept;
+    E&& error() && noexcept;
+    const E&&  error() const&& noexcept;
```
Mick235711, line 6726:
```suggestion
    const E&& error() const&& noexcept;
```

source/utilities.tex
```diff
@@ -6463,6 +6464,2036 @@
...
+
+\pnum
+\returns
+\tcode{std::move(\exposid{val})}.
+\end{itemdescr}
+
+\indexlibrarymember{operator bool}{expected}%
+\indexlibrarymember{has_value}{expected}%
+\begin{itemdecl}
+\pnum
```
Mick235711, line 7727:
```suggestion
```

source/utilities.tex
Mick235711, line 6726:
```suggestion
    const E&& error() const&& noexcept;
```

source/utilities.tex
Mick235711, line 7057:
```suggestion
\tcode{is_trivially_move_constructible_v<T>} is \tcode{true}; and
\item
\tcode{is_trivially_move_constructible_v<E>} is \tcode{true}.
```

source/utilities.tex
Mick235711, line 7727:
```suggestion
```

source/utilities.tex
```diff
@@ -6463,6 +6464,2036 @@
...
+\constraints
+\tcode{is_constructible_v<E, initializer_list<U>\&, Args...>} is \tcode{true}.
+
+\pnum
+\effects
+Direct-non-list-initializes
+\exposid{val} with \tcode{std::forward<Args>(args)...}.
+
+\throws
```
Mick235711, line 6632:
```suggestion
\pnum
\throws
```

source/utilities.tex
```diff
@@ -6463,6 +6464,2036 @@
...
+\begin{itemdescr}
+\pnum
+\returns
+\exposid{val}.
+\end{itemdescr}
+
+\indexlibrarymember{value}{unexpected}%
+\begin{itemdecl}
+constexpr E&& value() && noexcept;
+constexpr const E&& value() const&& noexcept;
```
Mick235711, line 6653:
```suggestion
constexpr const E&& value() const&& noexcept;
constexpr E&& value() && noexcept;
```
Should we keep this consistent with the general synopsis? (Also many other specifications are inconsistent too, like `expected<T, E>::operator*()`, `bad_excepted_access<E>::error()` etc.. If we fix this, we may as well fix the others too...)

Another way is to fix the general synopsis to be consistent with `const&-&-&&-const&&` order, with the additional benefit of be consistent with `optional<T>::operator*()` and `optional<T>::value()` (these two are the only case in stdlib where all 4 member function qualifiers overload appears currently). However I don't feel this order is nature than, say `const&-&-const&&-&&`. (Maybe we should set a new standard for a natural overload order? i.e. fix `optional<T>` to use `const&-&-const&&-&&` order too?)

Mick235711 (changes requested):

source/utilities.tex
```diff
@@ -6463,6 +6464,2036 @@
...
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\constraints
+\tcode{is_move_constructible_v<E>} is \tcode{true}.
+
+\pnum
+\effects
+If rhs.has_value() is \tcode{false},
```
Mick235711, line 8026:
```suggestion
If \tcode{rhs.has_value()} is \tcode{false},
```

source/utilities.tex
```diff
@@ -6463,6 +6464,2036 @@
...
+\item
+\tcode{is_assignable_v<T\&, U>} is \tcode{true}; and
+\item
+\tcode{is_nothrow_constructible_v<T, U> || is_nothrow_move_constructible_v<T> ||\newline
+is_nothrow_move_constructible_v<E>}
+is \tcode{true}.
+\end{itemize}
+
+\effects
+\pnum
```
Mick235711, line 7469:
```suggestion
\pnum
\effects
```

source/utilities.tex
```diff
@@ -6463,6 +6464,2036 @@
...
+template<class... Args>
+  constexpr T& emplace(Args&&... args) noexcept;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\constraints
+\tcode{is_nothrow_constructible_v<T, Args...>} is \tcode{true}.
+
+\effects
```
Mick235711, line 7543:
```suggestion
\pnum
\effects
```

source/utilities.tex
```diff
@@ -6463,6 +6464,2036 @@
...
+  constexpr T& emplace(initializer_list<U> il, Args&&... args) noexcept;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\constraints
+\tcode{is_nothrow_constructible_v<T, initializer_list<U>\&, Args...>}
+is \tcode{true}.
+
+\effects
```
Mick235711, line 7568:
```suggestion
\pnum
\effects
```

source/utilities.tex
```diff
@@ -6463,6 +6464,2036 @@
...
+\item
+\tcode{is_assignable_v<E\&, GF>} is \tcode{true}; and
+\item
+\tcode{is_nothrow_constructible_v<E, GF> || is_nothrow_move_constructible_v<T> ||\newline
+is_nothrow_move_constructible_v<E>} is \tcode{true}.
+\end{itemize}
+
+\pnum
+\effects
+\pnum
```
Mick235711, line 7514:
```suggestion
\pnum
\effects
```

source/utilities.tex
```diff
@@ -6463,6 +6464,2036 @@
...
+\effects
+Direct-non-list-initializes \exposid{unex}
+with \tcode{std::forward<Args>(args)...}.
+
+\pnum
+\ensures
+\tcode{has_value()} is \tcode{false}.
+
+\throws
```
Mick235711, line 8156:
```suggestion
\pnum
\throws
```

source/utilities.tex
```diff
@@ -6463,6 +6464,2036 @@
...
+construct_at(addressof(@\exposid{unex}@), std::move(rhs.@\exposid{unex}@));
+destroy_at(addressof(rhs.@\exposid{unex}@));
+@\exposid{has_val}@ = false;
+rhs.@\exposid{has_val}@ = true;
+\end{codeblock}
+
+\throws
+Any exception thrown by the expressions in the \effects.
+
+\remarks
```
Mick235711, line 8366:
```suggestion
\pnum
\throws
Any exception thrown by the expressions in the \effects.

\pnum
\remarks
```

source/utilities.tex
```diff
@@ -6463,6 +6464,2036 @@
...
+constexpr E&& error() &&;
+constexpr const E&& error() const&&;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\expects
+\tcode{has_value()} is \tcode{false}.
+
+\returns
```
Mick235711, line 8457:
```suggestion
\pnum
\returns
```

source/utilities.tex
```diff
@@ -6463,6 +6464,2036 @@
...
+  constexpr bool operator==(const expected& x, const unexpected<E2>& e);
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\mandates
+The expression \tcode{x.error() == e.value()} is well-formed and
+its result is convertible to \tcode{bool}.
+
+\returns
```
Mick235711, line 8493:
```suggestion
\pnum
\returns
```

source/utilities.tex
jwakely: I thought we usually have a non-const overload first, then const. That's what we do or `begin()` and `end()` in containers and strings, but then we have `operator[]` and `front()` and `back()` in `basic_string` where the `const` overload is first.

I agree we should probably be globally consistent (and IMHO the most natural order is either `&` `const&` `&&` `const&&` or `&` `&&` `const&` `const&&`) but we are already badly inconsistent.

Making this member consistent with the class synopsis makes sense. I don't think we need to rearrange the synopsis for consistency given that we don't have any real consistency.

jwakely (commented):
All of @Mick235711's suggestions look correct.

source/utilities.tex
```diff
@@ -6463,6 +6464,2036 @@
...
+\remarks
+This constructor is defined as deleted unless
+\begin{itemize}
+\item
+\tcode{is_copy_constructible_v<T>} is \tcode{true} and
+\item
+\tcode{is_copy_constructible_v<E>} is \tcode{true}.
+\end{itemize}
+
+This constructor is trivial if
```
Mick235711, line 7005:
```suggestion
\pnum
This constructor is trivial if
```

source/utilities.tex
```diff
@@ -6463,6 +6464,2036 @@
...
+\begin{codeblock}
+@\exposid{reinit-expected}@(@\exposid{val}@, @\exposid{unex}@, std::forward<U>(v));
+@\exposid{has_val}@ = true;
+\end{codeblock}
+
+\pnum
+\returns
+\tcode{*this}.
+\end{itemize}
+\end{itemdescr}
```
Mick235711, line 7485:
```suggestion
\end{codeblock}
\end{itemize}

\pnum
\returns
\tcode{*this}.
\end{itemdescr}
```

Mick235711 (changes requested):
(I think I found all the `\pnum` issues... sorry if I'm wrong or overshoot anything, since I used LaTeX but not very familiar with C++ draft's heavily-customized flavor)

source/utilities.tex
```diff
@@ -6463,6 +6464,2036 @@
...
+Let \tcode{GF} be \tcode{const G\&} for the first overload and
+\tcode{G} for the second overload.
+
+\pnum
+\constraints
+\tcode{is_constructible_v<E, GF>} is \tcode{true}.
+
+\pnum
+\effects
+Direct-non-list-initializes \exposid{unex} with \tcode{std::forward<GF>(e.error())}.
```
jdoerrie, line 7188:
```suggestion
Direct-non-list-initializes \exposid{unex} with \tcode{std::forward<GF>(e.value())}.
```

`unexpected` calls its member accessor `value()`, not `error()`.

source/utilities.tex
```diff
@@ -6463,6 +6464,2036 @@
...
+
+\indexlibraryctor{expected}%
+\begin{itemdecl}
+constexpr expected(const expected& rhs);
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\effects
+If \tcode{rhs.has_value()} is \tcode{false}
```
Mick235711, line 7993:
```suggestion
If \tcode{rhs.has_value()} is \tcode{false},
```

source/utilities.tex
Mick235711: Actually, I'd expected `unexpected<E>`'s member accessor to be `error()` instead of `value()`... but I guess that should be a later LWG issue and is out of scope for this PR.

jwakely: Changing that would be totally out of scope for an LWG issue, because it's not a defect. It's the approved design.

It should be `error()` here though, good catch.

Mick235711: Oh, I thought LWG issues can be deferred to LEWG to discuss design changes... Does this means that I have to write a proposal to change this?

jwakely: Yes. This is not a defect. If you don't like the design, write a paper for LEWG.

Mick235711: Thank you, I think I'll write one in the weekend... BTW, the poll is saying R12, but I cannot find it anywhere ([R11 source github](https://github.com/jfbastien/papers) seems not updated yet). Is there a changelog of R12 that I can base on?

source/utilities.tex
```diff
@@ -6463,6 +6464,2036 @@
...
+
+\begin{itemdescr}
+\pnum
+\constraints
+\tcode{is_constructible_v<E, initializer_list<U>\&, Args...>} is \tcode{true}.
+
+\pnum
+\effects
+Direct-non-list-initializes
+\exposid{val} with \tcode{std::forward<Args>(args)...}.
```
jdoerrie, line 6630:
```suggestion
\exposid{val} with \tcode{il, std::forward<Args>(args)...}.
```

source/utilities.tex
```diff
@@ -6463,6 +6464,2036 @@
...
+
+\pnum
+\throws
+Any exception thrown by the initialization of \exposid{val}.
+\end{itemdescr}
+
+\indexlibraryctor{unexpected}%
+\begin{itemdecl}
+template<class U, class... Args>
+  constexpr explicit unexpected(in_place_t, initializer_list<U>, Args&&...);
```
jdoerrie, line 6619:
```suggestion
  constexpr explicit unexpected(in_place_t, initializer_list<U> il, Args&&... args);
```

source/utilities.tex
```diff
@@ -6463,6 +6464,2036 @@
...
+
+\pnum
+\throws
+Any exception thrown by the initialization of \exposid{val}.
+\end{itemdescr}
+
+\indexlibraryctor{unexpected}%
+\begin{itemdecl}
+template<class... Args>
+  constexpr explicit unexpected(in_place_t, Args&&...);
```
jdoerrie, line 6598:
```suggestion
  constexpr explicit unexpected(in_place_t, Args&&... args);
```

`args...` are mentioned in the initialize expression below.

source/utilities.tex
```diff
@@ -6463,6 +6464,2036 @@
...
+\item
+\tcode{is_constructible_v<unexpected<E>, const expected<U, G>>}
+is \tcode{false}; and
+\end{itemize}
+
+\pnum
+\effects
+If \tcode{rhs.has_value()} is \tcode{false},
+direct-non-list-initializes \exposid{unex}
+with \tcode{std::forward<GF>(rhs.er\-ror())}.
```
jdoerrie, line 8082:
```suggestion
with \tcode{std::forward<GF>(rhs.error())}.
```

source/utilities.tex
Mick235711: Nearly all ctors like this need to have named arguments, I think.

source/utilities.tex
Mick235711: Let's just float the idea: is there any interest inside WG21 editor team for an editorial PR that makes every member function (overloaded only by qualifiers) in stdlib to become a consistent order (presumably `&-const&-&&-const&&`, since the tradition is to specify first two in a group and next two (often need additional `std::move`) in other), and thus we can commit to future consistency in new proposals?

If there is interest I may be able to commit time to make a PR when I have spare time.

jensmaurer: Can we switch the entire thing to explicit object parameters and avoid the question entirely?
@jwakely , is that an ABI break of concern for the templates we talk about here?

Mick235711: I think if the calling convention of
```cpp
struct S {
    void fun(int) const &;
}
```
and
```
struct S {
    static void fun(const S&, int);
```
is completely the same (which it seems to be, testing on godbolt), then there shouldn't be an ABI break (except that the mangled name is different, and this can be solved by providing aliases).

However a possible scenario is that if we provide it as
```cpp
template<typename E>
struct unexpected {
    constexpr decltype(auto) value(this auto&& self) noexcept {return std::forward<decltype(self)>(self).unex;}
}
```
then, I think that for types derivated from `unexpected<E>` this can be an ABI break: the calling convention between calling `unexpected<E>::value() const &` (ordinary member) and calling `unexpected<E>::value<const Derived&>(this const Derived&)` (deducing this) can be different, because the former you need a cast from `const Derived&` to `const unexpected<E>&` (which can be not nop, in case of multiple inheritance), and the latter due to template you do not need any cast.

jwakely: There cannot possibly be an ABI break for `std::unexpected` because it doesn't exist in any implementation yet.

And anyway, if the mangled name is different (and it is) then it doesn't matter whether the calling convention is the same or not. It's a different symbol and existing callers of the old symbol (that already have an instantiation of the old definition) cannot be affected by the addition of a new symbol. The only potential problem is if users are relying on explicit instantiations of `std::optional`, but whether that is supported or not is unspecified and QoI.

But that's all off-topic for an editorial PR.

Mick235711: Ah, yeah, sorry I misunderstand the ABI break scenario… Then that’s great and so think we can change optional/expected to explicit this without any breakage.

Anyways, this needs a new proposal to change and is outside the scope. I think for this PR the only thing we need to done is to change every specification order to match the general synopsis…

source/utilities.tex
```diff
@@ -6463,6 +6464,2043 @@
...
+Otherwise, equivalent to \tcode{\exposid{unex} = rhs.error()}.
+\end{itemize}
+
+\pnum
+\returns
+\tcode{*this}.
+
+\pnum
+\remarks
+The exception specification is
```
JohelEGP, line 8270: If you need to make it a code block due to an overfull `\hbox`, remember to add the colon.
```suggestion
The exception specification is equivalent to
```

source/utilities.tex
```diff
@@ -6463,6 +6464,2043 @@
...
+\tcode{is_move_constructible_v<E>} is \tcode{true}.
+
+\pnum
+\effects
+See \tref{expected.void.swap}.
+
+\begin{floattable}{\tcode{swap(expected\&)} effects}{expected.void.swap}
+{lx{0.35\hsize}x{0.35\hsize}}
+\topline
+& \chdr{\tcode{this->has_value()}} & \rhdr{\tcode{!has_value()}} \\ \capsep
```
JohelEGP, line 8348: Consistency.
```suggestion
& \chdr{\tcode{this->has_value()}} & \rhdr{\tcode{!this->has_value()}} \\ \capsep
```

source/utilities.tex
```diff
@@ -6463,6 +6464,2043 @@
...
+\begin{codeblock}
+construct_at(addressof(@\exposid{unex}@), std::move(rhs.@\exposid{unex}@));
+destroy_at(addressof(rhs.@\exposid{unex}@));
+@\exposid{has_val}@ = false;
+rhs.@\exposid{has_val}@ = true;
+\end{codeblock}
+
+\pnum
+\throws
+Any exception thrown by the expressions in the \effects.
```
JohelEGP, line 8368:
```suggestion
Any exception thrown by the expressions in the \Fundescx{Effects}.
```

source/utilities.tex
```diff
@@ -6463,6 +6464,2043 @@
...
+
+\pnum
+\remarks
+The exception specification is
```
JohelEGP, line 8372: If you need to make it a code block due to an overfull `\hbox`, remember to add the colon.
```suggestion
The exception specification is equivalent to
```

source/utilities.tex
```diff
@@ -6463,6 +6464,2043 @@
...
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\throws
+\tcode{bad_expected_access(std::move(error()))}
+if \tcode{has_value()} is \tcode{false}.
+\end{itemdescr}
+
+\begin{itemdecl}
```
JohelEGP, line 8436:
```suggestion
\indexlibrarymember{error}{expected}%
\begin{itemdecl}
```

JohelEGP (changes requested):
How about indexing `void` `expected` as `expected<void>`? This should reduce confusion and back-and-forth, guess-the-correct-one when looking up members from the index.

source/utilities.tex
```diff
@@ -6463,6 +6464,2043 @@
...
+\end{codeblock}
+\item
+Otherwise, equivalent to \tcode{\exposid{unex} = std::move(rhs.error())}.
+\end{itemize}
+Then, if no exception was thrown,
+equivalent to: \tcode{has_val = rhs.has_value(); return *this;}
+
+\pnum
+\remarks
+The exception specification is
```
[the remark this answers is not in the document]
jensmaurer: No. See above.

source/utilities.tex
```diff
@@ -6463,6 +6464,2043 @@
...
+rhs.@\exposid{has_val}@ = true;
+\end{codeblock}
+
+\pnum
+\throws
+Any exception thrown by the expressions in the \effects.
+
+\pnum
+\remarks
+The exception specification is:
```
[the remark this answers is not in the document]
jensmaurer: No. See above.

source/utilities.tex
```diff
@@ -6463,6 +6464,2043 @@
...
+  bool @\exposid{has_val}@;         // \expos
+  union {
+    E @\exposid{unex}@;             // \expos
+  };
+};
+\end{codeblock}
+
+\pnum
+\tcode{E} shall meet the requirements of
+\oldconcept{Destructible} (\tref{cpp17.destructible}).
```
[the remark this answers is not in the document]
jensmaurer: No. The incoming paper has the wrong style here.

source/utilities.tex
jensmaurer: No; see above.

source/utilities.tex
jensmaurer: No. See above.

source/utilities.tex
```diff
@@ -6463,6 +6464,2043 @@
...
+\tcode{rhs.has_value()} is unchanged;
+\tcode{rhs.has_value() == this->has_value()} is \tcode{true}.
+
+\pnum
+\throws
+Any exception thrown by the initialization of \exposid{val} or \exposid{unex}.
+
+\pnum
+\remarks
+The exception specification is
```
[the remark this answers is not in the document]
JohelEGP: That's how they all currently look in the WD. That was raised at https://github.com/cplusplus/draft/pull/4105, and is a vestige of its previous utterance. What should we do?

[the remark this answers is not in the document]
jensmaurer: Ah, good point, I misremembered.

source/utilities.tex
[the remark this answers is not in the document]
JohelEGP: It seemed right to me. Perhaps because all other occurrences in this paper and some in other incoming ones also do it.

jensmaurer:
> How about indexing void expected as expected<void>? This should reduce confusion and back-and-forth, guess-the-correct-one when looking up members from the index.

Do we have an existing practice in that area?

JohelEGP:
> 

I know there are some for atomics:

jensmaurer:
Indexing for `expected<void>` is now fixed.

source/utilities.tex
```diff
@@ -6463,6 +6464,2042 @@
...
+\rSec2[expected.syn]{Header \tcode{<expected>} synopsis}
+
+\indexlibraryglobal{unexpect_t}%
+\indexlibraryglobal{unexpect}%
+\begin{codeblock}
+namespace std {
+  // \ref{expected.un.object}, class template \tcode{unexpected}
+  template<class E> class unexpected;
+
+  // \ref{expected.bad}, class \tcode{bad_expected_access}
```
tkoeppe, line 6488: @jensmaurer: should this be "class <ins>template</ins> bad_expected access"?

jensmaurer: Yes. Are you fixing that?

tkoeppe: Yes, will do, thanks!

source/utilities.tex
```diff
@@ -6463,6 +6464,2042 @@
...
+\pnum
+\returns
+If \tcode{x.has_value()} does not equal \tcode{y.has_value()}, \tcode{false};
+otherwise \tcode{x.has_value() || static_cast<bool>(x.error() == y.error())}.
+\end{itemdescr}
+
+\indexlibrarymember{operator==}{expected<void>}%
+\begin{itemdecl}
+template<class E2>
+  constexpr bool operator==(const expected& x, const unexpected<E2>& e);
```
Mick235711, line 8489:
```suggestion
  friend constexpr bool operator==(const expected& x, const unexpected<E2>& e);
```
Oops, just realized this from writing my proposal against this...

tkoeppe: Thanks! Yes, that seems to be missing from the paper. I'll add it.

PR #5282, opened 2022-02-10
P1413R3 Deprecate std::aligned_storage and std::aligned_union

### LINKED ISSUE #5265: [2022-02 LWG Motion 6] P1413R3 Deprecate std::aligned_storage and std::aligned_union
[P1413R3](https://wg21.link/p1413r3)

jensmaurer:
- Replace 'shall be' with 'is'.

Fixes #5265
Fixes cplusplus/papers#197

### DIFF
--- a/source/future.tex
+++ b/source/future.tex
@@ -1369,6 +1369,14 @@ has the following addition:
 namespace std {
   template<class T> struct is_pod;
   template<class T> inline constexpr bool is_pod_v = is_pod<T>::value;
+  template<size_t Len, size_t Align = @\exposid{default-alignment}@> // \seebelow
+    struct aligned_storage;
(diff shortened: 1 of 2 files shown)


tkoeppe:
@jensmaurer Are the removals of the table rows in tab:meta.trans.other missing? I can add those.

tkoeppe:
Ah, no, but there's a stray note that should probably also have been deleted.

jensmaurer:
Yes, the stray note should go, too.

tkoeppe:
I'm actually going to _move_ it to the Annex, given that we already have another note there.

tkoeppe:
(I'll apply the unrelated is_pod index fix after the motions.)

PR #5285, opened 2022-02-10
P2387R3 Pipe support for user-defined range adaptors

### LINKED ISSUE #5268: [2022-02 LWG Motion 9] P2387R3 Pipe support for user-defined range adaptors
[P2387R3](https://wg21.link/p2387r3)

jensmaurer:
Fixes  #5268
Fixes cplusplus/papers#1060

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -193,6 +193,11 @@ namespace std::ranges {
 
   namespace views { template<class T> inline constexpr @\unspec@ istream = @\unspec@; }
 
+  // \ref{range.adaptor.object}, range adaptor objects
+  template<class D>
+    requires is_class_v<D> && @\libconcept{same_as}@<D, remove_cv_t<D>>
+  class range_adaptor_closure { };
+
   // \ref{range.all}, all view
   namespace views {
     inline constexpr @\unspec@ all = @\unspec@;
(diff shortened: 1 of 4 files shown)


source/ranges.tex
```diff
@@ -3166,6 +3171,42 @@
 The expression \tcode{C | D} is well-formed if and only if
 the initializations of the state entities of \tcode{E} are all well-formed.
 
+\pnum
```
Mick235711, line 3174: I think the above "Given an additional range adaptor closure object" subclause probably need a `\pnum` too if this have one.

Mick235711 (changes requested):

source/ranges.tex
```diff
@@ -3166,6 +3171,42 @@
...
+Given an object \tcode{t} of type \tcode{T}, where
+\begin{itemize}
+\item
+\tcode{t} is a unary function object that accepts a \tcode{range} argument,
```
JohelEGP, line 3178:
```suggestion
\tcode{t} is a unary function object that accepts a \libconcept{range} argument,
```

source/ranges.tex
```diff
@@ -3166,6 +3171,42 @@
...
+\item
+\tcode{T} models \tcode{\libconcept{derived_from}<range_adaptor_closure<T>>},
+\item
+\tcode{T} has no other base classes of type \tcode{range_adaptor_closure<U>} for any other type \tcode{U}, and
+\item
+\tcode{T} does not model \tcode{range}
```
JohelEGP, line 3184:
```suggestion
\tcode{T} does not model \libconcept{range}
```

source/ranges.tex
jensmaurer: Well, the "additional" here makes me believe it is tightly coupled with the preceding text, so there should be no `\pnum`.

source/ranges.tex
```diff
@@ -3394,6 +3399,42 @@
...
+\item
+\tcode{t} is a unary function object that accepts a \libconcept{range} argument,
+\item
+\tcode{T} models \tcode{\libconcept{derived_from}<range_adaptor_closure<T>>},
+\item
+\tcode{T} has no other base classes of type \tcode{range_adaptor_closure<U>} for any other type \tcode{U}, and
+\item
+\tcode{T} does not model \libconcept{range}
+\end{itemize}
+then the implementation ensures
```
tkoeppe, line 3414: (I'm not wild about the use of "then" here, which is trying to make it sound as if the sentence had started with "If t is an object of type T...". Really we should drop the "then" and have a comma after the preceding item. But that's how it was given to us.)

PR #5287, opened 2022-02-10
P2533R0 Core Language Working Group "ready"

### LINKED ISSUE #5257: [2022-02 CWG Motion 1] P2533R0 (Core Language Working Group "ready" Issue
[P2533R0](https://wg21.link/p2533r0)

jensmaurer:
Fixes #5257 
Fixes #4843
Fixes #2338
Fixes cplusplus/papers#1196

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -2815,7 +2815,7 @@ The optional \grammarterm{attribute-specifier-seq} appertains
 to the entity being declared.
 A bit-field shall not be a static member.
 \indextext{bit-field!type of}%
-A bit-field shall have integral or enumeration type;
+A bit-field shall have integral or (possibly cv-qualified) enumeration type;
 the bit-field semantic property is not part of the type of the class member.
 The \grammarterm{constant-expression} shall be an integral constant expression
 with a value greater than or equal to zero and


source/classes.tex
```diff
@@ -2473,7 +2473,6 @@
 \indextext{conversion!user-defined}%
 
 \pnum
```
tkoeppe, line 2475: This is an odd (ineffective?) `\pnum` before a pure grammar block?

tkoeppe (changes requested):

source/classes.tex
```diff
@@ -2496,7 +2496,7 @@
 its \grammarterm{declarator} shall be
 a function declarator\iref{dcl.fct} of the form
 \begin{ncsimplebnf}
-ptr-declarator \terminal{(} parameter-declaration-clause \terminal{)} \opt{cv-qualifier-seq}
+ptr-declarator \terminal{(} parameter-declaration-clause \terminal{)} \opt{cv-qualifier-seq}\br
```
tkoeppe, line 2499: Is this change in the wrong commit?

source/expressions.tex
```diff
@@ -1690,7 +1690,20 @@
...
+\nontermdef{lambda-specifier}\br
+    \keyword{consteval}\br
+    \keyword{constexpr}\br
+    \keyword{mutable}
+\end{bnf}
+
+\begin{bnf}
+\nontermdef{lambda-specifier-seq}\br
+    lambda-specifier\br
+    lambda-specifier lambda-specifier-seq\br
```
tkoeppe, line 1706: Why a traling `\br` here but not above?

tkoeppe (changes requested):

source/expressions.tex
jensmaurer: There should not be a trailing `\br` here.

If you want to, I can fix that and force-push.

source/classes.tex
jensmaurer: Yes, this should have been deleted.

If you want to, I can fix that and force-push.

source/classes.tex
jensmaurer: Yes, this should be squashed into "CWG1726 Declarator operators and conversion functions".

If you want to, I can fix that and force-push.

source/classes.tex
jensmaurer: fixed

source/classes.tex
jensmaurer: fixed

source/expressions.tex

tkoeppe (approved):

source/classes.tex
```diff
@@ -2488,18 +2486,52 @@
...
+the \grammarterm{member-specification} of a class or class template
+but is not a friend declaration\iref{class.friend},
+the \grammarterm{id-expression} is a \grammarterm{conversion-function-id};
+\item
+otherwise, the \grammarterm{id-expression} is a \grammarterm{qualified-id}
+whose \grammarterm{unqualified-id} is a \grammarterm{conversion-function-id}.
+\end{itemize}
+
+\pnum
+A conversion function  shall have no non-object parameters and
```
JohelEGP, line 2518: The core issue doesn't include "non-object".

I suppose you added it to account for explicit object parameters.

Is that really enough? No need to restrict it to an explicit object parameter, but any amount of non-object parameters is OK? Is there wording elsewhere that make these cases ill-formed, and thus saying "non-object" is enough?

How about the function type specified below? Does it need a revision, too?

tkoeppe: I think the paper misrepresents the status quo? So tihs seems like a "plausible merge".

jensmaurer: A conversion function with an explicit this parameter seems plausible, but there must not be other parameters. Further cleanup (if any) should be in a separate editorial or core issue.

PR #5290, opened 2022-02-11
[module.interface] Add adjective 'exported' to dangling prons

ChuanqiXu9:
Although we could infer the declaration should refer to 'exported declaration' from the context in the previous paragraph, they are two paragraphs and split by a relatively long example. So I am confused at the first sight. I think it would be helpful to add the adjective.

### DIFF
--- a/source/modules.tex
+++ b/source/modules.tex
@@ -266,7 +266,7 @@ export using namespace N;       // error: does not declare a name
 \end{example}
 
 \pnum
-If the declaration is a \grammarterm{using-declaration}\iref{namespace.udecl}
+If an exported declaration is a \grammarterm{using-declaration}\iref{namespace.udecl}
 and is not within a header unit,
 all entities to which all of the
 \grammarterm{using-declarator}{s} ultimately refer (if any)


jensmaurer:
@opensdh, this looks like a minor improvement to me. What do you think?

opensdh:
> @opensdh, this looks like a minor improvement to me. What do you think?

I'd say it should just be "If **an** exported declaration&hellip;", since we're not trying to key off any previous mention with this approach.  (My only "relevance" to this wording is that my (unreviewed) CWG2443 wording makes the same change to /3.)

source/modules.tex
```diff
@@ -266,7 +266,7 @@
 \end{example}
 
 \pnum
-If the declaration is a \grammarterm{using-declaration}\iref{namespace.udecl}
+If the exported declaration is a \grammarterm{using-declaration}\iref{namespace.udecl}
```
jensmaurer, line 269:
```suggestion
If an exported declaration is a \grammarterm{using-declaration}\iref{namespace.udecl}
```

jensmaurer (approved):

jensmaurer:
@opensdh, if you could be so kind and hit "approve" on this pull request, in case you like it.

opensdh:
Using two commits to change two words makes me sad, but the (net) change is correct.

opensdh (approved):

tkoeppe:
I expect these commits will be squashed eventually. It's just when you edit a PR on GitHub, you get extra commits.

opensdh:
Sorry, I should have known that from previous pull requests.  My comment was a leak from other projects where people have to be prodded into doing such things even when they're "obviously correct".

jensmaurer:
@tkoeppe, on your plate now. Tagged with milestone [post-2022-02](https://github.com/cplusplus/draft/milestone/24) so that we won't forget.

ChuanqiXu9:
Thanks! I have squashed them.

tkoeppe:
It was clear in context that "the declaration" is exported, but
it is easier to understand if we restate "an exported declaration".

PR #5306, opened 2022-02-20
[lib] Add missing \pnum before descriptive elements

jensmaurer:
Also fix the ineffective check script by rewriting the
check in straightforward awk.


tkoeppe:
Would you mind putting the whitespace-only changes in strings.tex into a separate commit?

Very nice fix!

jensmaurer:
> Would you mind putting the whitespace-only changes in strings.tex into a separate commit?

Fixed

tkoeppe (approved):

PR #5309, opened 2022-02-21
[range.slide.iterator] Remove `@` in `\tcode`

cpplearner:
Without this patch:
upload


jensmaurer (approved):

tkoeppe:
Ah, thanks -- I spotted one of those cases, but not this one...

JohelEGP:
Do you think a check for this is feasible? Within `\tcode`, look for the pattern `@\macro` or `@\macro{stuff}@`.

PR #5310, opened 2022-02-21
[allocator.requirements.general] Remove unused provitions

JohelEGP:


jensmaurer:
Yes, very much so.

jwakely (approved):
We used to have an unqualified `forward` but that was replaced in ebd893b65beb4561252e19640a84d53933b17a62 for [LWG 3315](https://wg21.link/lwg3315).

jensmaurer:
@koeppe, do you want to merge this as-is or should I convert it to the non-table form?

tkoeppe:
@jensmaurer If you were able to make the conversion, I'd be very happy to immediately do that instead of this PR!

tkoeppe:
That is, we can just start with converting the intro, I mean! Not the table itself, that can wait.

PR #5311, opened 2022-02-22
[version.syn] Remove mention of nonexistent header `<priority_queue>`

cpplearner:
`std::priority_queue` is defined in `<queue>`.

The "Feature test macros" section in [P1206R7 Conversions from ranges to containers](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p1206r7.pdf) does say that the macro `__cpp_lib_containers_ranges` is present in `<priority_queue>`, but I guess it's an error rather than a proposal to add this header.


Mick235711:
Good point! (We are too getting used to the fact that nearly all containers are in their own headers (except priority_queue and *multi*)

jensmaurer (approved):

PR #5312, opened 2022-02-22
spacing of ref-qualifiers

### LINKED ISSUE #5308: How to space ref qualifiers
Should we have a convention for how to use whitespace in ref qualifiers in the library?

```
void f() const &;
void f() const&;
void f() const &&;
void f() const&&;
```

These qualifiers are still relatively rare, so we don't really have a weight of precedence here. I'm inclined to have no space between the cv and the ref qualifier (`const&`). If we want to fix a style, we should record it in the wiki.

jensmaurer:
Fixes #5308

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -965,7 +965,7 @@ struct X {
   void h(this X&, int);
   void h(int) &&;               // OK, another overload
   void j(this const X&);
-  void j() const&;              // error: redeclaration
+  void j() const &;             // error: redeclaration
   void k();
   void k(this X&);              // error: redeclaration
 };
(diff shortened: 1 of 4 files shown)


jensmaurer:
@tkoeppe, here you are

tkoeppe:
Beautiful, thank you!

source/basic.tex
```diff
@@ -965,7 +965,7 @@
   void h(this X&, int);
   void h(int) &&;               // OK, another overload
   void j(this const X&);
-  void j() const&;              // error: redeclaration
+  void j() const &;             // error: redeclaration
```
JohelEGP, line 968: This is the only occurrence in core wording (in an example). Why not check only the library sources?

jensmaurer: Sure. Done.

jensmaurer:
@tkoeppe, want to merge this?

PR #5319, opened 2022-02-24
[class.prop] Clarify definition of implicit-lifetime class

### LINKED ISSUE #5318: [class.prop] ambiguous conditions in definition of implicit-lifetime class
> A class `S` is an *implicit-lifetime class* if it is an aggregate or has at least one trivial eligible constructor and a trivial, non-deleted destructor.

It's ambiguous whether this means (A || (B && C)) or ((A || B) && C).

How about:

> A class `S` is an *implicit-lifetime class* if it has a trivial, non-deleted destructor and either it is an aggregate or it has at least one trivial eligible constructor.

jensmaurer:
Fixes #5318

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -302,10 +302,14 @@ Aggregates of class type are described in~\ref{dcl.init.aggr}.
 \end{note}
 
 \pnum
-A class \tcode{S} is an \defnadj{implicit-lifetime}{class}
-if it is an aggregate or
-has at least one trivial eligible constructor and
+A class \tcode{S} is an \defnadj{implicit-lifetime}{class} if
+\begin{itemize}
+\item
+it is an aggregate or
+\item
+it has at least one trivial eligible constructor and
 a trivial, non-deleted destructor.
+\end{itemize}
 
 \rSec1[class.name]{Class names}
 \indextext{definition!class name as type}%


jensmaurer:
It is unclear which semantics were originally intended. CWG should have a look.

zygoloid:
From P0593R6 section 3.1, the intent was to "permit implicit object creation for aggregate class types even if the aggregate contains an element with a non-trivial destructor".

How about:

> A class `S` is an *implicit-lifetime* class if <ins>either</ins> it is an aggregate or <ins>it</ins> has at least one trivial eligible constructor and a trivial, non-deleted destructor.

... or switching to bullets:

> A class `S` is an *implicit-lifetime* class if
> * it is an aggregate or
> * <ins>it</ins> has at least one trivial eligible constructor and a trivial, non-deleted destructor.

jensmaurer:
@jwakely, [P0593R6](https://wg21.link/p0593r6) section 3.1 is pretty straightforward what is meant here.

I've added bullets to clarify the and/or grouping.

opensdh:
@zygoloid: You're certainly right that P0593R6 explicitly mentions "aggregate contains an element with a non-trivial destructor", but that's not the same thing as saying that _all_ aggregates should be allowed, since
```c++
struct X {
  Y i;
  ~X();
};
```
is an aggregate and yet (regardless of the type `Y`) doesn't satisfy "Destroying an instance of the type runs no code".

jensmaurer:
CWG telecon 2022-06-17: Approved.

jensmaurer:
@tkoeppe: CWG has approved this earlier today. Any final thoughts before I merge?

tkoeppe:
Looks great, merge it! Thanks!

Issue #4702, opened 2021-06-17
Restyle expos-only member basic-formatting-string::str


tkoeppe:
See https://github.com/cplusplus/draft/issues/4680#issuecomment-862806234

We should style the exposition-only member `str` of `basic-formatting-string` in regular upright font, as we typically do for expos-only members. We should also rename it to `str_`, as we do for private expos-only members.

jensmaurer:
The private ranges members are all styled `\exposid`.

And, outside of ranges, we have a fair amount of member names not ending in `_`.

Personally, I think all exposition-only member names should be italics, and we should find a better / more pronounceable way of creating a new namespace, other than appending `_`.

tkoeppe:
Do you have a sense of how many non-ranges private members there are?

tkoeppe:
Changing the issue to decide the general style first.

jensmaurer:
> Do you have a sense of how many non-ranges private members there are?

Rough estimate, counting the `\expos` markers (ignoring `\exposid`):

$ grep -w expos *.tex | grep -v ranges.tex | wc -l
275

In other words, oodles of stuff.

tkoeppe:
OK, thanks. I feel like we're not going to resolve this for this mailing. Maybe for the next pre-meeting mailing!

jensmaurer:
Editorial meeting 2021-06-18: Consistent, global rule: All private data members have a `_` suffix. All exposition-only names (including private data members) are styled with `\exposid{...}`.

New macro `\exposdm` auto-adds the `_`.

The presence of the `_` suffix is saddening, but we don't have a better idea right now.

Dani-Hub:
Re @tkoeppe :

> Do you have a sense of how many non-ranges private members there are?

In N4885 there are 51 occurrences within Clause [ranges], specifically the following ones:
subrange: 2
single_view: 1
iota_view: 2
iota_view<W, Bound>::iterator: 1
iota_view<W, Bound>::sentinel: 1
basic_istream_view: 2
basic_istream_view<Val, CharT, Traits>::iterator: 1
ref_view: 1
filter_view: 2
filter_view<V, Pred>::iterator: 2
filter_view<V, Pred>::sentinel: 1
transform_view: 2
transform_view<V, F>::iterator: 2
transform_view<V, F>::sentinel: 1
take_view: 2
take_view<V>::sentinel: 1
take_while_view: 2
take_while_view<V, Pred>::sentinel: 2
drop_view: 2
[...]

PR #5322, opened 2022-02-24
[range.chunk.outer.value] Add missing private specifier

hewillk:


jensmaurer:
Should we make that a class instead?

tkoeppe:
Overall we seem to have a happy mix of class and struct, but `value_type` always seems to be `struct` so far.

jensmaurer:
And `lazy_split_view::outer-iterator ::value_type` uses `struct value_type` plus `private`, too.

jensmaurer (approved):

tkoeppe:
I think we have only one appearance each of "class outer-iterator" and "struct outer-iterator". Maybe we should make them all structs?

jensmaurer:
It's not about the outer-iterator; it's about the nested type `value_type`. Do you have examples where we do `class` for `value_type`?

tkoeppe:
Yes,I know -- as I was saying, all `value_type`s are already structs. But on top of that, we also have some inconsistent use of class/struct for outer-iterator and inner-iterator. All those types feel similar, so I wonder if we should make them all structs, and avoid future uncertainty if and when we get more such types.

jensmaurer:
I think the better overall approach is to make them all `class`: They all start with a bunch of private data members.

tkoeppe:
Yes, that's another option. I'm don't have a strong preference, though I think it'd be useful to fix a style.

The fact that the private members appear at the top is another localy style; elsewhere we have private expos-only members at the bottom (so the class key doesn't matter). I'm not sure it's worth mandating a stricter style for the ordering, though.

hewillk:
It is also worth noting that `iterator` and `sentinel` are structs in [range.join.with.overview], and then they are later in [range.join.with.iterator] and [range.join.with.sentinel] is defined as a `class`. Maybe I should open another pull request?

tkoeppe:
I'd like to see if we can agree on a direction first.

jensmaurer:
I think we had discussed the ordering of members (private first or public first) in passing at some point, and my general approach is to put private members last (except typedefs).  That needs lots of changes in [ranges], so this is not for now. Note also the ongoing discussion in #4702 about styling exposition-only members.

My personal preference for class vs. struct is that `struct` is used only for aggregates (e.g. the algorithm _result typse); everything else (in particular stuff with private members or any abstraction to speak of) is `class`.

I still believe that, even absent general consistency, adding "private" here is progress in both correctness of the specification and local consistency, and that's the best we can hope for given the constrained resources we have.

tkoeppe:
I recorded the issue in #5323.

PR #5327, opened 2022-02-27
[allocator.requirements.general] Dismantle requirements table

jensmaurer:
Fixed #5310


jensmaurer:
@tkoeppe, here you are. You can cherry-pick only the first commit if you don't want the whole thing.

tkoeppe:
Wonderful, thanks!

@jwakely, @CaseyCarter: would you perhaps be able to take a look at this requirements table replacement? Do you think it needs LWG oversight?

source/lib-intro.tex
```diff
@@ -1980,189 +2164,388 @@
...
+allocation_result<X::pointer>
 
 \pnum
-Note C:
+\returns
+\tcode{allocation_result<X::pointer>\{ptr, count\}}
+where \tcode{ptr} is memory allocated for an array of \tcode{count} \tcode{T}
+and such an object is created but array elements are not constructed,
+such that $\tcode{count} \geq \tcode{n}$.
 If \tcode{n == 0}, the return value is unspecified.
```
JohelEGP, line 2219: This is a _Remarks_ in `a.allocate(n)`.

source/lib-intro.tex
```diff
@@ -1980,189 +2164,388 @@
...
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\result
+\tcode{bool}
+
+\pnum
+\returns
+\tcode{!(a1 == a2)}
```
JohelEGP, line 2320:
```suggestion
\returns
\tcode{!(a1 == a2)}.
```

source/lib-intro.tex
```diff
@@ -1980,189 +2164,388 @@
...
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\result
+\tcode{bool}
+
+\pnum
+\returns
+\tcode{a == Y::rebind<T>::other(b)}
```
JohelEGP, line 2334:
```suggestion
\tcode{a == Y::rebind<T>::other(b)}.
```

source/lib-intro.tex
```diff
@@ -1980,189 +2164,388 @@
...
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\result
+\tcode{bool}
+
+\pnum
+\returns
+\tcode{!(a == b)}
```
JohelEGP, line 2348:
```suggestion
\tcode{!(a == b)}.
```

source/lib-intro.tex
```diff
@@ -1980,189 +2164,388 @@
...
+\end{itemdescr}
+
+\begin{itemdecl}
+X u(b);
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\ensures
+\tcode{Y(u) == b} and \tcode{u == X(b)}
```
JohelEGP, line 2373:
```suggestion
\tcode{Y(u) == b} and \tcode{u == X(b)}.
```

source/lib-intro.tex
```diff
@@ -1980,189 +2164,388 @@
...
+(not used)
+
+\pnum
+\effects
+Constructs an object of type \tcode{C} at \tcode{c}.
+
+\pnum
+\remarks
+Default:
+\tcode{construct_at(c,~std::forward<Args>(args)...)}
```
JohelEGP, line 2425:
```suggestion
\tcode{construct_at(c, std::forward<Args>(args)...)}
```
Is this still needed?

source/lib-intro.tex
```diff
@@ -1980,189 +2164,388 @@
...
+\result
+\tcode{X}
+
+\pnum
+\returns
+Typically returns either \tcode{a} or \tcode{X()}.
+
+\pnum
+\remarks
+Default: \tcode{return a;}
```
JohelEGP, line 2461:
```suggestion
\remarks
Default: \tcode{a}
```
Or should `max_size`'s _Default_ prepend `return ` and append `;`?

source/lib-intro.tex
```diff
@@ -1980,189 +2164,388 @@
...
+\end{itemdescr}
+
+\begin{itemdecl}
+X::propagate_on_container_copy_assignment
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\result
+Identical to or derived from \tcode{true_type} or \tcode{false_type}.
```
JohelEGP, line 2471: Preexisting: These probably needs to be more than just derived. Like "a type with a base characteristic ([meta.rqmts]) of `bool_constant<`_`B`_`>`".
In that case, the plain use of `true_type` in the descriptions should be changed accordingly.

source/lib-intro.tex
```diff
@@ -1980,189 +2164,388 @@
...
+
+\pnum
+\returns
+\tcode{true_type} only if the expression \tcode{a1 == a2} is guaranteed
+to be \tcode{true} for any two (possibly \keyword{const}) values
```
JohelEGP, line 2542:
```suggestion
to be \tcode{true} for any two (possibly \tcode{const}) values
```

JohelEGP (changes requested):

source/lib-intro.tex
```diff
@@ -1840,137 +1840,321 @@
...
+defaults for different requirements than the primary template.
+
+\begin{itemdecl}
+typename X::pointer
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\remarks
+Default: \tcode{T*}
```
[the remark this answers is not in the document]
jensmaurer: Yeah, but I think it doesn't appear in [structure.specification]. I think there's also a "Default behavior" somewhere, which sounds more generic. Leaving as-is for now.

source/lib-intro.tex
jensmaurer: Yes, but (pre-existing), `a.allocate(n)` doesn't even say what it returns in the good case, so I can't put it in the right spot.

source/lib-intro.tex
jensmaurer: I think the latter is preferable for clarity.

jensmaurer:
@JohelEGP , thanks for the review. All trivial fixes are applied.

source/lib-intro.tex
JohelEGP: Makes sense.

source/lib-intro.tex
JohelEGP: Then how about this?
> `p->m`
_Effects:_ Equivalent to `(*p).m`

For functions, that means it's `void`-returning.
Granted, the meaning of the description elements for non-functions is not formally defined in [structure.specifications].

For defaults, how about explicitly calling out what is being defaulted? Like this:
```suggestion
\remarks
Default result: \tcode{a}
```

jwakely (approved):
I don't think this needs review/approval from LWG. There is unanimous support for these changes among LWG telecon attendees.

If some minor problems do get introduced by this change, they can be fixed later. We don't need to hold the change back waiting for LWG review.

source/lib-intro.tex
tkoeppe: This is now #5348.

tkoeppe:
Thanks, @jwakely! Merging; please get in touch if there are any complaints or issues.

source/lib-intro.tex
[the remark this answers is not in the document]
jensmaurer: I don't think so. Now that we have enough use-cases, we should come up with usage guidelines for the existing stuff, and actually add it to [structure.specification] where missing.

PR #5333, opened 2022-03-06
[range.utility.conv.general] Add missing template parameter to container_inserter

hewillk:
Also, I strongly doubt that this should be a `constexpr` function, if so I'd be happy to add an LWG (if that can't be as an edit fixed) for it.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -2101,8 +2101,8 @@ constexpr bool @\exposid{container-insertable}@ =          // \expos
 \pnum
 Let \exposid{container-inserter} be defined as follows:
 \begin{codeblock}
-template<typename Ref>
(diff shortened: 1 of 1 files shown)


hewillk:
Any comments?

jensmaurer:
The words in P1206R7 Conversions from ranges to containers were correctly applied, so this was a bug in the incoming paper. Given the usage, this change looks plausible.

@jwakely , @CaseyCarter , @ericniebler , @cor3ntin : Please offer an opinion.

jensmaurer (approved):

jwakely (approved):

ericniebler:
Looks good to me.

CaseyCarter (approved):

PR #5360, opened 2022-03-24
[expected.unexpected] Rearrange std::unexpected

jwakely:
Some minor improvements to the `std::unexpected` spec.

This is the TOC before:

before

And after:

after

In the second image every "Class template foo" under 22.8 is at the same level, without the pointless "Unexpected objects" subclause, and without two "General" subclauses for one simple class template.

The other commit just reorders the `std::unexpected` synopsis to match the order of the detailed descriptions, and the order of corresponding constructors in the `std::expected` and `std::expected<cv void, E>` synopses and details descriptions.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -6541,12 +6541,12 @@ namespace std {
   public:
     constexpr unexpected(const unexpected&) = default;
     constexpr unexpected(unexpected&&) = default;
+    template<class Err = E>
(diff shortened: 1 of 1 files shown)


jensmaurer (approved):

jensmaurer:
@tkoeppe, this looks like an improvement. Do you want to merge that, or should I?

PR #5381, opened 2022-04-01
[expected.object.general] Add missing noexcept for expected::error()

hewillk:
In [[expected.object.general]](https://timsong-cpp.github.io/cppwp/expected.object.general), the function is defined as

`constexpr const E& error() const &;`

However in [[expected.object.obs]](https://timsong-cpp.github.io/cppwp/expected.object.obs), it is defined as

`constexpr const E& error() const & noexcept;`

So let me assume it's just that the author of the paper forgot to add `noexcept`.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -6905,10 +6905,10 @@ namespace std {
     constexpr T& value() &;
     constexpr const T&& value() const &&;
     constexpr T&& value() &&;
-    constexpr const E& error() const &;
-    constexpr E& error() &;
-    constexpr const E&& error() const &&;
-    constexpr E&& error() &&;
+    constexpr const E& error() const & noexcept;
+    constexpr E& error() & noexcept;
+    constexpr const E&& error() const && noexcept;
+    constexpr E&& error() && noexcept;
     template<class U> constexpr T value_or(U&&) const &;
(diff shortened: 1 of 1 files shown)


jensmaurer:
@jwakely, this looks like an oversight in the incoming paper. Do you think this is editorial, or does it need an LWG issue?

Note: squash and fix commit message when merging

jensmaurer:
@jwakely , ping.

jwakely:
I don't recall the intention or whether we discussed making these narrow contact functions noexcept. I'll check the minutes.

jwakely:
The `noexcept` on the detailed descriptions appeared in [P0323R11](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p0323r11.html#expected.expected) following the 2021-10-08 LWG review. The minutes show:

- TS: there should be noexcept on the -> and * operators although they have a precondition
- JW: we should get rid of the expected:: qualification on value()
- JW: error() can be noexcept

It looks like we agreed to add it then, but it wasn't added to the synopsis. So I think this can be fixed editorially.

jwakely (approved):

jensmaurer (approved):

PR #5382, opened 2022-04-01
[range.utility.conv.general] Add missing constexpr for container-inserter

hewillk:
This is the follow up of [pull/5333](https://github.com/cplusplus/draft/pull/5333), and I think the authors agree that the missing `constexpr` is not intended, it's more like a bug in the paper.


jwakely:
@cor3ntin @CaseyCarter ?

jwakely (approved):

cor3ntin:
@jwakely yes, this makes sense to me, it was definitively an omission on my part, sorry about that.

cor3ntin (approved):

jensmaurer (approved):

PR #5384, opened 2022-04-04
LWG3687. `expected<cv void, E>` move constructor should move

hewillk:
Which it's apparently a bug in the paper.


JohelEGP:
That's https://wg21.link/LWG3687.

jensmaurer:
Let's not track this here, too.

PR #5397, opened 2022-04-17
[string.view.deduct] Move to immediately after [string.view.cons]

### LINKED ISSUE #4039: [string.view.template] Deduction guide specified far from constructors
`array`'s is specified right after the constructors, in the same subclause.
`span`'s are specified on their own subclause after the special member functions'.
`string_view`'s are right before the non-member functions. There's a lot of material between its special member function specifications and the non-members'.

jensmaurer:
Fixes #4039

@jwakely , please have a look.


jensmaurer:
@jwakely, ping.

jwakely (approved):

PR #5399, opened 2022-04-19
[expr.call] Clarify the value of this

languagelawyer:
There is no such thing as «`this` parameter of the function», after all

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -3197,10 +3197,12 @@ template<typename ...T> int f(int n = 0, T ...t);
 int x = f<int>();               // error: no argument for second function parameter
 \end{codeblock}
 \end{example}
-If the function is an implicit object member
-function, the \keyword{this} parameter of the function\iref{expr.prim.this}
-is initialized with a pointer to the object of the call, converted
-as if by an explicit type conversion\iref{expr.cast}.
+If the function is an implicit object member function,
+the object for which it is invoked\iref{expr.prim.this}
+is the object which would be pointed to by the result
+of an explicit type conversion\iref{expr.cast}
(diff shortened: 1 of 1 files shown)


jensmaurer:
The automatic check fails.

Also, I'm not convinced this is a net clarification.

languagelawyer:
> The automatic check fails.

I've used `\termref`.

> Also, I'm not convinced this is a net clarification.

Any details?

source/expressions.tex
```diff
@@ -3197,10 +3197,12 @@
 int x = f<int>();               // error: no argument for second function parameter
 \end{codeblock}
 \end{example}
-If the function is an implicit object member
-function, the \keyword{this} parameter of the function\iref{expr.prim.this}
-is initialized with a pointer to the object of the call, converted
-as if by an explicit type conversion\iref{expr.cast}.
+If the function is an implicit object member function,
+the object for which it is invoked\iref{expr.prim.this}
+is the object which would be pointed to by the result
```
jensmaurer, line 3202: Why "would"? There is nothing hypothetical here. The member function is invoked for a certain subobject (or possibly the entire object) of the object designated by the object expression.

languagelawyer: > Why "would"? There is nothing hypothetical here.

We don't do any code transformations or anything else to make this conversion «really exist».

jensmaurer: We do convert the object expression to the type of the base class, and the result of that transformation is the value of "this" inside the member function, in my view.

frederick-vs-ja:
#6803 removed incorrect «`this` parameter». Do we still want to mention the type and value of `this` here?

PR #5405, opened 2022-04-21
[expr.prim.lambda.general] Clarify deduced lambda return type

### LINKED ISSUE #5400: [expr.prim.lambda.closure] p3 What is the return type of a function call operator(template) if trailing-return-type is omitted?
[expr.prim.lambda.closure] p3 says
> The closure type for a lambda-expression has a public inline function call operator (for a non-generic lambda) or function call operator template (for a generic lambda) ([over.call]) whose parameters and **return type** are described by the lambda-expression's parameter-declaration-clause and **trailing-return-type** respectively, and whose template-parameter-list consists of the specified template-parameter-list, if any. 

we just mention the case where *trailing-return-type* is omitted in [expr.prim.lambda.general] p5
> If the lambda-declarator does not include a trailing-return-type, **the lambda return type is auto**,

we didn't explicitly associate the return type of the function call operator(template) of a lambda with the lambda return type. we merely defined the return type of the function call operator(template) as determined from the **trailing-return-type**.  

I think we should change [expr.prim.lambda.general] p5 to
> If the lambda-declarator does not include a trailing-return-type, it is as if its trailing-return-type is `->auto`.  

I also think saying lambda return type does not make sense. Since a lambda expression is a prvalue of a closure type and only the function call operator(template) of its closure type has a return type.

jensmaurer:
Fixes #5400

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -1757,9 +1757,11 @@ it is as if \tcode{()} were inserted
 at the start of the \grammarterm{lambda-declarator}.
 If the \grammarterm{lambda-declarator}
 does not include a \grammarterm{trailing-return-type},
-the lambda return type is \keyword{auto},
-which is deduced from \keyword{return} statements
+it is considered to be \tcode{-> \keyword{auto}}.
+\begin{note}
+In that case, the return type is deduced from \keyword{return} statements
 as described in \ref{dcl.spec.auto}.
+\end{note}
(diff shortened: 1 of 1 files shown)


zygoloid (approved):

jensmaurer:
@tkoeppe, looks like we're ready to go here.

PR #5409, opened 2022-04-23
[associative] Add "i.e." in front of explanation

jensmaurer:


jensmaurer:
@jwakely , what do you think?

jwakely (approved):

PR #5413, opened 2022-04-25
[except.pre] Reword "shall not be used to" avoiding question of actual "use"

hubert-reinterpretcast:
The "shall not be used to" phrasing may be taken to refer only in cases
where the actual use occurs or is the primary intent. Instead, the
intended restriction can be written in terms of static properties of the
constructs so restricted in the style of [stmt.if].

A new term of art (control-flow-limited statement) is introduced in
[stmt.label] to express the restrictions. Both [stmt.if] and
[except.pre] are updated to use the new term.

Co-authored-by: Johel Ernesto Guerrero Peña <johelegp@gmail.com>

### DIFF
--- a/source/exceptions.tex
+++ b/source/exceptions.tex
@@ -72,8 +72,8 @@ Within this Clause
 \indextext{\idxcode{switch}!and try block}%
 \indextext{\idxcode{goto}!and handler}%
 \indextext{\idxcode{switch}!and handler}%
-A \tcode{goto} or \keyword{switch} statement shall not be used to transfer control
-into a try block or into a handler.
+The \grammarterm{compound-statement} of a try block or of a handler is a
+control-flow-limited statement\iref{stmt.label}.
 \begin{example}
 \begin{codeblock}
 void f() {
--- a/source/statements.tex
+++ b/source/statements.tex
@@ -146,6 +146,18 @@ before its introduction by a \grammarterm{labeled-statement}.
 \indextext{label!\idxcode{default}}%
 Case labels and default labels shall occur only in \keyword{switch} statements.
 
+\pnum
+A \defnadj{control-flow-limited}{statement} is a statement \tcode{S} for which:
+\begin{itemize}
+\item
+  a \keyword{case} or \keyword{default} label appearing within \tcode{S} shall
+  be associated with a \keyword{switch} statement\iref{stmt.switch} within
+  \tcode{S}, and
+\item
+  a label declared in \tcode{S} shall only be referred to by a
+  statement\iref{stmt.goto} in \tcode{S}.
+\end{itemize}
+
 
 \rSec1[stmt.expr]{Expression statement}%
 \indextext{statement!expression}
@@ -262,6 +274,8 @@ present, is a discarded statement. During the instantiation of an
 enclosing templated entity\iref{temp.pre}, if the condition is
 not value-dependent after its instantiation, the discarded substatement
 (if any) is not instantiated.
+Each substatement of a constexpr if statement is a control-flow-limited
+statement\iref{stmt.label}.
 \begin{example}
 \begin{codeblock}
 if constexpr (sizeof(int[2])) {}        // OK, narrowing allowed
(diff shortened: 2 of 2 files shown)


source/exceptions.tex
```diff
@@ -72,8 +72,11 @@
 \indextext{\idxcode{switch}!and try block}%
 \indextext{\idxcode{goto}!and handler}%
 \indextext{\idxcode{switch}!and handler}%
-A \tcode{goto} or \keyword{switch} statement shall not be used to transfer control
-into a try block or into a handler.
+A \tcode{goto} statement that transfers control to a statement within a try
+block or handler shall itself be within that try block or handler.
+A \tcode{case} or \tcode{default} label within a try block or handler shall be
+associated with a \tcode{switch} statement that is itself within that try block
+or handler.
```
jensmaurer, line 79: Hm... [stmt.if] p2 uses a slightly dfifferent phrasing:

> A case or default label appearing within such an if statement shall be associated with a switch statement
(8.5.3) within the same if statement. A label (8.2) declared in a substatement of a constexpr if statement
shall only be referred to by a statement (8.7.6) in the same substatement.

I think I like the cross-references. I also think we should express those restrictions with the same words, if we can. Bonus points for defining a term "limited control-flow statement" (or so) and just reusing that everywhere.

jensmaurer (changes requested):

source/exceptions.tex
hubert-reinterpretcast: I've added the new term to [stmt.label] and applied it to two places in [stmt.if] (in addition to the [except.pre] case).

jensmaurer (approved):
I like this.

source/statements.tex
```diff
@@ -146,6 +146,19 @@
 \indextext{label!\idxcode{default}}%
 Case labels and default labels shall occur only in \keyword{switch} statements.
 
+\pnum
+\indextext{statement!control-flow-limited}%
+A \defn{control-flow-limited statement} is a statement \tcode{S} for which:
```
JohelEGP, line 151:
```suggestion
A \defnadj{control-flow-limited}{statement} is a statement \tcode{S} for which:
```

JohelEGP (changes requested):

source/statements.tex
jensmaurer: Agreed. With this change, `\indextext{statement!control-flow-limited}%` should go, because it's now automatic / implied.

zygoloid (approved):
This looks like a nice improvement to me.

I briefly considered whether it'd be useful to say that the body of a function is also control-flow-limited, but that doesn't seem to really allow any further simplifications.  (We could maybe delete the second sentence of [stmt.goto]/1, but that seems to be about it.)

jensmaurer:
@tkoeppe , any final thoughts before I merge this?

tkoeppe:
@jensmaurer: no, this does look like a nice improvement. As long as you're CWG-happy with inventing new terms editorially here, please do feel free to merge. (Maybe CWG should be informed that we have a new term at some point? Then again, control flow is not something that we touch a lot.)

source/statements.tex
```diff
@@ -146,6 +146,18 @@
...
+\pnum
+A \defnadj{control-flow-limited}{statement} is a statement \tcode{S} for which:
+\begin{itemize}
+\item
+  a \keyword{case} or \keyword{default} label appearing within \tcode{S} shall
+  be associated with a \keyword{switch} statement\iref{stmt.switch} within
+  \tcode{S}, and
+\item
+  a label declared in \tcode{S} shall only be referred to by a
+  statement\iref{stmt.goto} in \tcode{S}.
```
tkoeppe, line 158: I realize this is original wording, but should we narrow this down to "by a _goto_ statement", not just any old statement? Which other kinds of statement could be referring to a label?

jensmaurer: gcc-style computed gotos are similarly constrained, I believe, so keeping the "goto" a bit low-profile helps that extension. This is not a strong argument.

tkoeppe: Aha, I was wondering if that was behind it! That's fine, esp. since this is pre-existing. It just looked a bit odd to me to say "statement" but then cross-reference stmt.goto.

Fine to ship!

jensmaurer: I think this just falls out of the existing wording and was not by design of the wording. But we might not want to move towards the other direction, now that we've discovered what the wording does.

tkoeppe: By "the other direction" you mean narrowing it down to just `goto` statements? So you'd just keep it as is?

jensmaurer: Yes, I'd keep it as-is.

tkoeppe: Off we go.

PR #5419, opened 2022-04-26
[ranges.syn] remove trailing `-> see below` return type from three `t…

CaseyCarter:
…o` overloads

Since there is actually no return type specification to see below in [range.utility.conv].


jensmaurer (approved):
I think an "auto" return type in the synopsis is a clear indicator this is a deduced return type. We need to be extra careful that the detailed specification actually allows deduction of the return type, but I think we get that here.

jensmaurer:
@jwakely , what do you think?

jensmaurer:
@jwakely, ping.

jwakely (approved):
If we want to use <code>-> decltype(<em>see below</em>)</code> in the synopsis we should also do it in the detailed descriptions, and we don't do it there. This makes it consistent (and I don't think it would be an improvement to be consistent in the other direction by adding the trailing return type to the detailed descriptions).

PR #5380, opened 2022-04-01
[expected.object.general] Remove explicit(see below) for expected copy/move constructor

hewillk:

PR #5424, opened 2022-04-28
[expected.object.general, expected.void.general] copy and move constr…

CaseyCarter:
…uctors of expected are never explicit

These `explicit(see below)`s in the class synopses are not present in the detailed specifications of the constructors.


JohelEGP:
This is #5380.

PR #5443, opened 2022-05-04
[thread.sema.cnt] Add missing parentheses on function call expression

### LINKED ISSUE #5435: [thread.sema.cnt] missing parens on function call expression
p15 (15.1) "Evaluates `try_acquire`"

We're talking about an actual function call here, not the name of a function, so it should be `try_acquire()`.

Extracted from #5428

jensmaurer:
Fixes #5435


jensmaurer:
@jwakely, what do you think?

jwakely (approved):

PR #5444, opened 2022-05-04
[cons.slice] Add copy constructor for 'slice' to synopsis

### LINKED ISSUE #5429: [cons.slice] lists undeclared copy constructor
The `slice(const slice&);` member is not declared in the [class.slice.overview] synopsis.

Extracted from the comments for #5428

jensmaurer:
Fixes  #5429

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -7979,6 +7979,7 @@ namespace std {
   public:
     slice();
     slice(size_t, size_t, size_t);
+    slice(const slice&);
 
     size_t start() const;
     size_t size() const;


jensmaurer:
(I'm not sure whether adding the copy constructor to the class synopsis is the right call, or just removing the copy constructor from the itemdecl, since the semantics are not actually described. But then, we should say that `slice` is semiregular or so.

jensmaurer:
@jwakely, please have a look.

jwakely (approved):
This seems like the minimal fix to make it consistent. Maybe we want another issue for your semiregular suggestion, but that might be a thread we don't want to start pulling in the ancient `valarray` wording.

PR #5446, opened 2022-05-04
[over.match.best.general] Fix typo in example

### LINKED ISSUE #5433: [over.match.best.general] example 7 should use 'c'
Example 7 says "`c` &rarr; `int` is better than `c` &rarr; `short`" but it should be the literal `'c'` not a variable `c`.

Extracted from #5428

jensmaurer:
Fixes  #5433


jensmaurer:
@jwakely, please have a look

jwakely (approved):

PR #5467, opened 2022-05-13
[stacktrace.basic.nonmem] Add missing \pnum before \recommended

### LINKED ISSUE #5466: [stacktrace.basic.nonmem] missing \pnum in p2?
Should the _Recommended practice:_ have its own paragraph number here?

Or maybe it's OK, there's one in [dcl.link] p2 that doesn't have its own paragraph. Another in [thread.req.timing] p4, and [futures.unique.future] p3, and [futures.shared.future] p3. And [futures.async] p3 is even more fun. All those look OK, but [stacktrace.basic.nonmem] looks odd.

jensmaurer:
Also augment check script

Fixes #5466

### DIFF
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -11121,7 +11121,7 @@ deferred function in the thread that called the waiting function.
 Once evaluation of \tcode{invoke(std::move(g), std::move(xyz))} begins, the function is no longer
 considered deferred.
 
-\recommended
+ \recommended
 If this policy is specified together with other policies, such as when using a
 \tcode{policy} value of \tcode{launch::async | launch::deferred}, implementations should defer
 invocation or the selection of the policy when no more concurrency can be effectively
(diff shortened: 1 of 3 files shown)


jensmaurer:
@jwakely , here you are.

source/threads.tex
```diff
@@ -11121,7 +11121,7 @@
 Once evaluation of \tcode{invoke(std::move(g), std::move(xyz))} begins, the function is no longer
 considered deferred.
 
-\recommended
+ \recommended
```
jwakely, line 11124: Is this trick of adding a space before the macro documented anywhere as the correct way to use one without starting a new paragraph?

jensmaurer: No.

I've changed my check script so that it complains about `\recommended` (and similar introducers) without `\pnum` inside `\itemdecl` only.  In the (rather exotic) case we have here, a `\pnum` isn't correct, because we're inside an `\itemize` environment. The space in front silences the check script. Maybe I should augment it even more so that stuff inside `\itemize` is ignored.

PR #5474, opened 2022-05-19
[range.join.view] Simplify `range_reference_t<V>` to `InnerRng`

hewillk:
I think this is a reasonable simplification and also makes it consistent with `end()`.


CaseyCarter (approved):
Good catch! This seems to simply have been an oversight when we replaced the other occurrences with `InnerRng`.

jensmaurer (approved):

jensmaurer:
@jwakely, this feels "obviously correct". What do you think?

jwakely (approved):

tkoeppe:
Thanks!

PR #5492, opened 2022-05-25
[range.chunk.fwd.iter] Fix unformatted semicolon

hewillk:


JohelEGP:
`\tcode{return `_`... not-a-semicolon`_`};` seems like an easy to add check with high returns.

PR #5498, opened 2022-05-27
[range.chunk.by.view] Add missing exposition only comment

hewillk:


CaseyCarter (commented):
I came to say "this isn't a thing we do in the library specification", but I see there's at least one inconsistency. That said, I don't think we want to get in the business of spamming `// exposition only` comments on every use of every exposition-only name; we've been operating under the belief that tagging the first occurrence suffices, especially so in [ranges] where exposition-only class members are all rendered in kebab-case.

I'd rather see an editorial change that _removes_ the few existing occurrences of redundant `// exposition only` comments in itemdecls than one that adds more of them.

jensmaurer:
Agreed. Patches welcome.

hewillk:
> Agreed. Patches welcome.

I am glad to. However, I still have a question about this, taking the helper `concept compatible-joinable-ranges` as an example, which first appeared in [[ranges.syn]](https://eel.is/c++draft/ranges.syn) and already has the `//exposition only` comment

```
// [range.join.with], join with view
template<class R, class P>
  concept compatible-joinable-ranges = see below; // exposition only

template<input_­range V, forward_­range Pattern>
  requires view<V> && input_­range<range_reference_t<V>>
        && view<Pattern>
        && compatible-joinable-ranges<range_reference_t<V>, Pattern>
class join_with_view;
```

So in [[range.join.with.view]](https://eel.is/c++draft/range.join.with.view), does this reappearance of `// exposition only` comment need to be removed?

```
namespace std::ranges {
  template<class R, class P>
  concept compatible-joinable-ranges =            // exposition only
      common_­with<range_value_t<R>, range_value_t<P>> &&
      common_­reference_­with<range_reference_t<R>, range_reference_t<P>> &&
      common_­reference_­with<range_rvalue_reference_t<R>, range_rvalue_reference_t<P>>;
```

jwakely:
IMHO it makes sense to keep the comment on declarations in a synopsis. Just not in a later `\itemdecl`.

PR #5500, opened 2022-05-29
[range.chunk.overview,range.slide.overview] Correct the italics of N/M in the overview

hewillk:


jensmaurer (changes requested):
I don't think we want that. N is a template parameter and thus is at the C++ source code level, and thus should be code font, not math font.

For the [range.slide.overview] changes regarding "M", this seems to be purely a mathematical placeholder not actually represented in C++ source code. Changing this to math font is fine, but maybe we want lowercase, too, to make the difference more apparent?

hewillk:
> I don't think we want that. N is a template parameter and thus is at the C++ source code level, and thus should be code font, not math font.

This change is to be *consistent* with overview section in [[range.take.overview]](https://eel.is/c++draft/range.take.overview), [[range.drop.overview]](https://eel.is/c++draft/range.drop.overview), [[range.elements.overview]](https://eel.is/c++draft/range.elements.overview), [[range.zip.transform.overview]](https://eel.is/c++draft/range.zip.transform.overview), [[range.adjacent.overview]](https://eel.is/c++draft/range.adjacent.overview), [[range.adjacent.transform.overview]](https://eel.is/c++draft/range.adjacent.transform.overview). 

Maybe they need to change too.

tkoeppe:
@jensmaurer: is N actually a template parameter here? It seems to be used quite similarly to how drop_view works. What's worse, there's a second, unrelated N in [range.slide.overview]p2.

JohelEGP:
`N` is only a template parameter in `elements_view`, `adjacent_view` and `adjacent_transform_view`. But they use an undefined _N_ in the first paragraph of their overview.

jensmaurer:
No, sorry, `N` is not a template parameter, but just some value. In the second paragraph, we say "Given subexpressions \tcode{E} and \tcode{N},"

So, this change for the first paragraph (where we talk about the abstract concept) actually looks good.

tkoeppe:
@hewillk Can you please rebase?

PR #5514, opened 2022-06-09
[range.istream.view] Add \ref for `basic_istream_view::iterator`

hewillk:


source/ranges.tex
```diff
@@ -3257,6 +3257,7 @@
     constexpr default_sentinel_t end() const noexcept;
 
   private:
+    // \ref{range.istream.iterator}, class \tcode{basic_istream_view::\exposid{iterator}}
```
CaseyCarter, line 3260: Conventionally these cross-references-in-synopsis-comments repeat the title of the referenced subclause:
```suggestion
    // \ref{range.istream.iterator}, class template \tcode{basic_istream_view::\exposid{iterator}}
```

JohelEGP: Then there's an error _there_, because this `iterator` isn't preceded by a _template-head_.

CaseyCarter (approved):

tkoeppe:
Thanks!

source/ranges.tex
CaseyCarter: For posterity: I've submitted #5515 to fix this.

PR #5516, opened 2022-06-09
[range.chunk] Add the class template description for `chunk_view`

hewillk:
Note that I also removed the "`<V>`" part in "Class template `chunk_view<V>​::​iterator` for forward ranges" to be consistent with other sections.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -10968,7 +10968,7 @@ for (auto r : v | views::chunk(2)) {
 \end{codeblock}
 \end{example}
 
-\rSec3[range.chunk.view.input]{\tcode{chunk_view} for input ranges}
+\rSec3[range.chunk.view.input]{Class template \tcode{chunk_view} for input ranges}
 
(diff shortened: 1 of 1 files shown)


hewillk:
> Note that I also removed the "`<V>`" part in "Class template `chunk_view<V>​::​iterator` for forward ranges" to be consistent with other sections.

This may not be an appropriate choice since this "`<V>`" is needed for class partial specialization.

JohelEGP:
The "for forward ranges" part is clarifying enough to me.

tkoeppe:
Please rebase?

tkoeppe:
@hewillk Ping ^^^

hewillk:
Is this acceptable?

jensmaurer (approved):

jensmaurer:
The change itself looks good; the two commits should be squashed and force-pushed, and the commit description should be fixed.

Something like "[range.chunk] Fix subclause headings" might work.

PR #5519, opened 2022-06-11
[ranges.syn] Fix the constraints order of `slide_view`

hewillk:


jensmaurer:
Why is the other order "better"?

hewillk:
I don't think it's better, I just fixed it with minimal changes.

jensmaurer:
Why is it a "fix"? The end result is two constraints, it seems.

hewillk:
The constraint orders matter, they are different things, you can not
```
namespace std::ranges {
template<view V>
  requires forward_range<V>
class slide_view;

template<forward_range V>
  requires view<V>
class slide_view { };
}
```

JohelEGP:
Because the order is relevant: https://godbolt.org/z/3MPrfeWq6. In this case, the fix orders the constraints like in its class synopsis and those of its iterator/sentinel pair.

jensmaurer:
Ah, ok, this is making the declarations consistent across the synopsis and the iterator/sentinel. Thanks.

jensmaurer (approved):

CaseyCarter (approved):

Issue #5428, opened 2022-05-02
Wholesale list of issues (with suggested fixes), as found in ISO14882:2020


naiveforce:
The list is extensive, therefore best viewed against the source document. The file attached contains markups importable (e.g. using Foxit PDF Reader) against PDF version of the Standard's text.
[isoiec_148822020.zip](https://github.com/cplusplus/draft/files/8605016/isoiec_148822020.zip)

Overall methodology of used markup primitives:
- Replacements: contain text that should be substituted;
- Inserts: contain text that should be added in the pointed out locations;
- Strikeouts: candidates for removal;
- Floating notes: editorial remarks that could not be easily attached to a specific text location;
- Highlights: editorial remarks and "general thoughts" about specific text fragments.
Aside from the specific replacement/additional text each of these can contain hints/comments delimited by square brackets.

Additionally, frequently reoccurring issues, that I feel are worth mentioning as a "preface", can be put into two general categories:

1. Frequently occurring issues that have been pointed out:
 - using 0 in place of nullptr (e.g. in assignments and comparisons involving pointer types.);
 - using sizeof(expr) instead of the (grammar appropriate) sizeof expr;
 - redundant access specifiers (e.g. struct S {public:...}; class C {private:...}; struct D : public B {...}; etc.);
 - places exhibiting formatting issues such as text indentation, missing whitespace etc.

2. Frequently occurring issues that have been left out:
 - using std::move, std::forward where move, forward would (according to context) perfectly suffice (can be easily fixed by searching/replacing using automation if need be.)

While these cannot be categorized as outright errors, I feel that fixing them would improve readability and general quality of the Standard's text.

JohelEGP:
The library part of the standard values consistency. Core wording is intentionally inconsistent in some regards.

> using 0 in place of nullptr (e.g. in assignments and comparisons involving pointer types.);

Any `0` in the core wording is probably intentional. On the library side, it can only be pre-C++11 text. New additions do use `nullptr`.

> using sizeof(expr) instead of the (grammar appropriate) sizeof expr;

The library uses the former.

> redundant access specifiers (e.g. struct S {public:...}; class C {private:...}; struct D : public B {...}; etc.);

I think it was @tkoeppe whoe recently made a comment on how it should be on the library side.

> using std::move, std::forward where move, forward would (according to context) perfectly suffice (can be easily fixed by searching/replacing using automation if need be.)

[...]

PR #5531, opened 2022-06-16
Typo and formatting fixes

jensmaurer:
Partially addresses #5428

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -9932,7 +9932,7 @@ template<class T> constexpr reference_wrapper<const T> cref(const T& t) noexcept
 \begin{itemdescr}
 \pnum
 \returns
-\tcode{reference_wrapper <const T>(t)}.
+\tcode{reference_wrapper<const T>(t)}.
 \end{itemdescr}
 
 \indexlibrarymember{cref}{reference_wrapper}%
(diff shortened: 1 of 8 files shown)


source/utilities.tex
```diff
@@ -12349,7 +12349,7 @@
 
 \indexlibraryctor{default_searcher}%
 \begin{itemdecl}
-constexpr default_searcher(ForwardIterator pat_first, ForwardIterator pat_last,
+constexpr default_searcher(ForwardIterator1 pat_first, ForwardIterator1 pat_last,
```
jwakely, line 12352: Should we change the synopsis to match?

jensmaurer: It's correct in the synopsis. (ForwardIterator1 is a template parameter of the enclosing class.)

jwakely: Right, but in the synopsis it's just called `ForwardIterator`. It's not incorrect, just inconsistent.

jensmaurer: You were talking about the header synopsis; I was looking at the class synopsis (and found not issue).

The header synopsis is now fixed.

jwakely: Yes, sorry for being unclear! I should have said header.

jwakely (approved):

jensmaurer:
@jwakely , I've pushed more fixes. Please approve again. I'll then merge and continue in a fresh pull request.

jwakely (approved):

jensmaurer:
@tkoeppe, I have no idea. My guess is that it happens only if we have a trailing underscore, so it's likely related to the underscore package.

PR #5532, opened 2022-06-16
 [common.iter.nav] Fix \exposconceptnc

hewillk:
That makes it clickable, right?

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -5055,7 +5055,7 @@ common_iterator tmp = *this;
 return tmp;
 \end{codeblock}
 Otherwise, if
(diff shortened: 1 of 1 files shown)


JohelEGP:
Concepts in code blocks are not clickable in the PDF. Perhaps you've found a bug for https://github.com/Eelis/cxxdraft-htmlgen?

hewillk:
Yep, since I can only `make` a pdf document, I have no way of knowing whether the commit _successfully_ makes the concept clickable, although it seems to be.
Note that currently at [[common.iter.nav]](https://eel.is/c++draft/common.iter.nav), it is not clickable.

PR #5534, opened 2022-06-17
use "named by" in place of "appear as" for odr-using a variable

### LINKED ISSUE #5533: [basic.def.odr] p5 "appears as" is not defined and "named by" is a better choice
[basic.def.odr] p5 says
> A variable is named by an expression if the expression is an id-expression that denotes it. A variable x **whose name appears as** a potentially-evaluated expression E is odr-used by E unless 

The intent of what `E` is here refers to the `id-expression` naming the entity. "appear as" is not a defined meaning. It is better that we just say: 
> A variable **is named by an expression** if **the expression is an id-expression** that denotes it. A variable x **named by a potentially-evaluated expression E** is odr-used by E unless  

It is more clear.

xmh0511:
Fixes #5533.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -440,7 +440,7 @@ This covers
 \label{term.odr.use}%
 A variable is named by an expression
 if the expression is an \grammarterm{id-expression} that denotes it.
-A variable \tcode{x} whose name appears as a
+A variable \tcode{x} that is named by a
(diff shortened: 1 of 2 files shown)


jensmaurer (approved):
This needs squashing and a proper commit message.

tkoeppe:
That much I can do.

Issue #4890, opened 2021-09-10
[module.private.frag] A confusion example


xmh0511:
````cpp
export module A;
// #1
export inline void fn_e();      // error: exported inline function fn_­e not defined
                                // before private module fragment
// #2
inline void fn_m();             //  OK, module-linkage inline function
static void fn_s();
export struct X;
export void g(X *x) {
  fn_s();                       // OK, call to static function in same translation unit
  fn_m();                       // OK, call to module-linkage inline function
}
export X *factory();            // OK

module :private;
struct X {};                    // definition not reachable from importers of A
X *factory() {
  return new X ();
}
void fn_e() {}
void fn_m() {}
void fn_s() {}
````
In this example, `#1` and `#2` both attach to module `A`, they have external or module linkage, respectively. In terms of this case, [basic.def.odr] p11,  [dcl.inline] p5 and p7 should apply to both `#1` and `#2`.   The definition of definition domain is defined as that  

> A definition domain is a private-module-fragment or the portion of a translation unit excluding its private-module-fragment (if any).  

I'm not sure whether the intent of this rule is that a `private-module-fragment` can divide out another definition domain in a translation unit. If it is, `#2` should be ill-formed since it is odr-used in the first definition domain but its definition appears in another domain, as per [basic.def.odr] p11  
> A definition of an inline function or variable shall be reachable from the end of every definition domain **in which it is odr-used outside of a discarded statement**.  

Except that, I'm not seeing any difference between `#1` and `#2` in terms of this example, that is, they should have the same treatment unless there are some normative rules that differentiate them in this case.

jensmaurer:
In your example, there are two definition domains: One is the private-module-fragment; the other one is the rest of the translation unit.
[basic.def.odr] p11 says that the definition of `fn_m` needs to be reachable from "the rest", but it isn't. I'd say the use inside `g` is thus ill-formed. If your compiler says otherwise, maybe you should consider posting a bug report to your vendor?

My guess is that at `#1`, the compiler just wants to be helpful and tell you that you can never odr-use that exported function, because the definition cannot ever be reachable.

xmh0511:
This example is formal case located in [module.private.frag], I just think the comments are confused here. We agree on `fn_m`, which should be ill-formed. For `fn_e`, violates [dcl.inline] p7 may the reason why it is an error.
> If an inline function or variable that is attached to a named module is declared in a definition domain, it shall be defined in that domain.

[...]

PR #5537, opened 2022-06-23
[module.private.frag] Remove misleading example and broaden note

jensmaurer:
@zygoloid, this addresses http://lists.isocpp.org/ext/2022/06/19564.php

Note that the example believes that fn_m (module linkage) is ok, but [dcl.inline] p7 makes no distinction between external and module linkage (both fn_e and fn_m are attached to the named module "A", I believe):

> If an inline function or variable that is attached to a named module is declared in a definition domain, it shall be defined in that domain.

### DIFF
--- a/source/modules.tex
+++ b/source/modules.tex
@@ -797,7 +797,7 @@ The presence of a \grammarterm{private-module-fragment} affects:
 \begin{itemize}
 \item
 the point by which the definition of
-an exported inline function
+an inline function or variable
 is required\iref{dcl.inline},
 
 \item
@@ -827,12 +827,11 @@ the reachability of declarations within it\iref{module.reach}.
 export module A;
 export inline void fn_e();      // error: exported inline function \tcode{fn_e} not defined
                                 // before private module fragment
-inline void fn_m();             // OK, module-linkage inline function
+inline void fn_m();             // error: non-exported inline function \tcode{fn_m} not defined
 static void fn_s();
 export struct X;
 export void g(X *x) {
   fn_s();                       // OK, call to static function in same translation unit
-  fn_m();                       // OK, call to module-linkage inline function
 }
 export X *factory();            // OK
 


source/modules.tex
```diff
@@ -827,12 +827,11 @@
 export module A;
 export inline void fn_e();      // error: exported inline function \tcode{fn_e} not defined
                                 // before private module fragment
-inline void fn_m();             // OK, module-linkage inline function
+inline void fn_m();             // error: module-linkage inline function \tcode{fn_m} not defined
```
zygoloid, line 830: Maybe:
```suggestion
inline void fn_m();             // error: non-exported inline function \tcode{fn_m} not defined
                                // before private module fragment
```
The current comment makes it sound like the linkage matters here, and it doesn't; making this parallel the prior comment might make it clearer that exportedness doesn't matter either?

jensmaurer: Agreed, fixed.

JohelEGP: The fix has not been pushed.

jensmaurer: Now it has. Thanks for the hint.

jensmaurer:
@tkoeppe, this seems ready.

See also endorsement by @zygoloid here: https://lists.isocpp.org/ext/2022/06/19578.php

tkoeppe:
Thanks. Has your later concern been addressed adequately by the response in http://lists.isocpp.org/ext/2022/06/19578.php?

xmh0511:
This PR seems to fix https://github.com/cplusplus/draft/issues/4890. Confusion is that [dcl.inline] p7 says
>If an inline function or variable that is attached to a named module is declared in a definition domain, it shall be defined in **that domain**.

Does "that domain" refers to the lexical portion where the inline function or variable is declared or to any same kind of definition domain that can appear in other translation units? The comment will be helpful if is 
> the function is not defined in this domain([dcl.inline])

Since [dcl.inline] unified apply to all inline functions and variables as long as they are **attached to** a named module.

jensmaurer:
As explained in #5561, "definition domain" is clearly defined lexically ("portion of a translation unit"), so I think there is no remaining issue.

Even if there were, it could be addressed separately at a later time.

tkoeppe:
(And this only changes a note/example anyway.)

PR #5538, opened 2022-06-24
[temp.inst] Clarify referent of 'declaration'

jensmaurer:
Fixes cplusplus/CWG#50


zygoloid (approved):
Looks correct, and I think this is an improvement.

jensmaurer:
@tkoeppe, this seems ready.

PR #5632, opened 2022-07-21
Post-LWG3713, use "sorted with respect to `comp`" consistently

Quuxplusone:
[LWG3713](https://cplusplus.github.io/LWG/issue3713) points out that we temporarily lost the term of art "sorted with respect to `comp`," and brings back a definition for it. However, several places in the existing draft never actually used that term of art in the first place. Fix them up so that they do. (hat tip @CaseyCarter)


CaseyCarter (approved):

jensmaurer (approved):

jensmaurer:
We should apply this only after LWG3713 was actually applied (i.e. after the July 2022 straw polls).

PR #5633, opened 2022-07-21
Substatements of if consteval

brevzin:
As noted [here](https://lists.isocpp.org/std-discussion/2022/07/1737.php), we talk about the substatements of `if consteval`, but those are currently defined just in terms of _statement_ whereas `if consteval` has a _compound-statement_.


tkoeppe:
@jensmaurer Is this editorial? It seems straight-forward enough, but it is a normative change.

jensmaurer:
I'm not seeing a normative change here, since a _compound-statement_ is a _statement_. But the fix is a welcome clarification.

tkoeppe:
I see, thanks! I'll merge this then!

PR #5634, opened 2022-07-22
[dcl.init.aggr]/1 Aggregate only cares about direct base classes CWG2610

languagelawyer:
Fixes https://github.com/cplusplus/CWG/issues/101
See also https://github.com/cplusplus/draft/pull/2892

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -4807,9 +4807,9 @@ no user-declared or inherited constructors\iref{class.ctor},
 \item
 no private or protected direct non-static data members\iref{class.access},
 \item
-no virtual functions\iref{class.virtual}, and
+no private or protected direct base classes\iref{class.mi}, and
 \item
-no virtual, private, or protected base classes\iref{class.mi}.
+no virtual functions\iref{class.virtual} or virtual base classes.
 \end{itemize}
 \begin{note}
 Aggregate initialization does not allow accessing


jensmaurer (changes requested):
I think indirect virtual base classes may be a problem (they have global constructor effects). Agreed with the private/protected changes.

source/declarations.tex
```diff
@@ -4809,7 +4809,7 @@
 \item
 no virtual functions\iref{class.virtual}, and
 \item
-no virtual, private, or protected base classes\iref{class.mi}.
+no virtual, direct private, or direct protected base classes\iref{class.mi}.
```
hubert-reinterpretcast, line 4812: Perhaps better to split this into two bullets?

languagelawyer: Or "or base classes" can be added to the previous bullet, and "virtual" removed from this one?

hubert-reinterpretcast: > Or "or base classes" can be added to the previous bullet, and "virtual" removed from this one?

Sounds good.

jensmaurer (approved):

jensmaurer:
This looks good to me now.

@hubert-reinterpretcast , please have a final look.

hubert-reinterpretcast:
> This looks good to me now.
> 
> @hubert-reinterpretcast , please have a final look.

It's good enough. I have a mild preference for repeating "virtual" and reordering the bullets so that the two "no private or protected" bullets are grouped together.

hubert-reinterpretcast (approved):
LGTM; thanks!

tkoeppe:
@jensmaurer Does CWG consider this editorial? (Feel free to merge if so!)

jensmaurer:
CWG2610

PR #5656, opened 2022-07-26
P2448R2 Relaxing some constexpr restrictions

### LINKED ISSUE #5574: [2022-07 CWG Motion 9] P2448R2 Relaxing some constexpr restrictions
[P2448R2](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2448r2.html)

jensmaurer:
Fixes #5574
Fixes cplusplus/papers#1119

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -771,12 +771,6 @@ The definition of a constexpr function shall satisfy the following
 requirements:
 \begin{itemize}
 \item
-its return type (if any) shall be a literal type;
-
-\item
-each of its parameter types shall be a literal type;
-
-\item
 it shall not be a coroutine\iref{dcl.fct.def.coroutine};
 
 \item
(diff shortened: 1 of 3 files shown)


source/declarations.tex
```diff
@@ -816,78 +810,6 @@
...
-
-struct D : B {
-  constexpr D() : B(global) { }         // ill-formed, no diagnostic required
-                                        // lvalue-to-rvalue conversion on non-constant \tcode{global}
-};
-
-constexpr int f(int x) {
-  static int n = x;
-  return n + x;                         // ill-formed, no diagnostic required
-                                        // all calls reach the static variable declaration
```
tkoeppe, line 886: This change isn't in the paper (which I suppose has a stale baseline). This example was added by https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2242r3.html.

@jensmaurer Is it OK to lose this example?

jensmaurer: I think we want to keep that example. Can you revert that hunk?

tkoeppe: We've lost the ambient paragraph, though, so where should the residual example be? Just in a paragraph of its own?

JohelEGP: But the paper normatively removes the restriction that it's IFNDR, so the example no longer has normative backing. If it does have, the example can be moved to where that is justified.

tkoeppe: I see, thanks. @jensmaurer OK to lose this then?

jensmaurer: Right, ok with keeping the example removed.

tkoeppe: Great, thank you both!

PR #5657, opened 2022-07-26
[concept.booleantestable] Remove unnecessary hyphenation hint for `boolean-testable` concept

hewillk:


source/concepts.tex
```diff
@@ -981,7 +981,7 @@
 \tcode{true_type}\iref{meta.type.synop},
 \tcode{int*}, and
 \tcode{bitset<N>::reference}\iref{template.bitset}
-model \exposconceptx{boolean\--testable}{boolean-testable}.
```
JohelEGP, line 984: Seems like this hyphenation hint is not needed anymore due to examples being a point smaller now.

jensmaurer (changes requested):
So, this is two independent changes:
- one noticing that the hyphenation hint is now superfluous
- one noticing that a `\placeholder` should become an `\exposconcept`
Two changes, two patches, please (two distinct pull requests are fine).

jensmaurer (approved):

PR #5658, opened 2022-07-26
P2266R3 Simpler implicit move

### LINKED ISSUE #5575: [2022-07 CWG Motion 10] P2266R3 Simpler implicit move
[P2266R3](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2266r3.html)

jensmaurer:
Fixes #5575
Fixes cplusplus/papers#968

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -42,6 +42,19 @@ auto c = L"a" U"b";             // was conditionally-supported; now ill-formed
 
 \rSec2[diff.cpp20.expr]{\ref{expr}: expressions}
 
+\diffref{expr.prim.id.unqual}
+\change
+Change move-eligible \grammarterm{id-expression}s from lvalues to xvalues.
+\rationale
+Simplify the rules for implicit move.
+\effect
+Valid \CppXX{} code that relies on a returned \grammarterm{id-expression}'s
+being an lvalue may change behavior or fail to compile. For example:
+\begin{codeblock}
+decltype(auto) f(int&& x) { return (x); }       // returns \tcode{int\&\&}; previously returned \tcode{int\&}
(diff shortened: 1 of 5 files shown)


source/compatibility.tex
```diff
@@ -42,6 +42,19 @@
 
 \rSec2[diff.cpp20.expr]{\ref{expr}: expressions}
 
+\diffref{expr.prim.id.unqual}
+\change
+Change move-eligible \grammarterm{id-expression}s from lvalues to xvalues.
+\rationale
+Simplify the rules for implicit move.
+\effect
+Valid \CppXX{} code that relies on a returned \grammarterm{id-expression}'s
```
JohelEGP, line 51:
```suggestion
Valid \CppXX{} code that relies on a returned \grammarterm{id-expression}
```

JohelEGP (changes requested):

source/compatibility.tex
jensmaurer: I think the existing text is fine English grammar, but I'll defer to the editor @tkoeppe for that.

tkoeppe: I agree, both are fine, and no reason to deviate from the approved wording.

JohelEGP: Is this supposed to imply "_id-expression_'s value category"?

tkoeppe: No, I think it's fine as is. "X's being a Y" is the same as "the fact that X is a Y", so you can think of this as "the id-expression is an lvalue", which seems fine - either "an expression is an lvalue" or "an expression's value category is 'lvalue'", but the former seems definitely fine, and common.

JohelEGP: I see. Thank you.

PR #5664, opened 2022-07-27
Motions 2022 07 cwg 1: P2622R0 Core Defect Reports

### LINKED ISSUE #5566: [2022-07 CWG Motion 1] P2622R0 Core Defect Reports
[P2622R0](https://wg21.link/p2622r0)

burblebee:
Fixes #5566


JohelEGP:
Also fixes the corresponding issue at https://github.com/cplusplus/papers/.

source/templates.tex
```diff
@@ -8417,6 +8441,21 @@
 \end{codeblock}
 \end{example}
 
+\pnum
+The type of \tcode{B} in the \grammarterm{noexcept-specifier}
+\tcode{noexcept(B)} of a function type is \tcode{bool}.
```
JohelEGP, line 8452: Do any of these need to use `\keyword`? I don't remember its use cases.

JohelEGP (changes requested):
CWG2507 and CWG2586 are missing.

source/templates.tex
burblebee: I don't think so.  @jensmaurer @tkoeppe, thoughts?

tkoeppe: Yes, we should say `\tcode{\keyword{noexcept}(B)}` here I think. I can make that change, thanks for flagging!

tkoeppe:
> [CWG2507](https://cplusplus.github.io/CWG/issues/2507.html) and [CWG2586](https://cplusplus.github.io/CWG/issues/2586.html) are missing.

Those are part of the next motion, CWG-2.

Mick235711:
> Also fixes the corresponding issue at https://github.com/cplusplus/papers/.

This fixes cplusplus/papers#1273.

PR #5670, opened 2022-07-28
P1467R9 Extended floating-point types and standard names

### LINKED ISSUE #5579: [2022-07 CWG Motion 14] P1467R9 Extended floating-point types and standard names
[P1467R9](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p1467r9.html)

jensmaurer:
Fixes #5579
Fixes cplusplus/papers#227

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -4995,15 +4995,23 @@ of the set of values of the type \tcode{\keyword{long} \keyword{double}}.
 The types
 \keyword{float}, \keyword{double}, and \tcode{\keyword{long} \keyword{double}},
 and cv-qualified versions\iref{basic.type.qualifier} thereof,
+are collectively termed
+\defnx{standard floating-point types}{type!floating-point!standard}.
+An implementation may also provide additional types
+that represent floating-point values and define them (and cv-qualified versions thereof) to be
+\defnx{extended floating-point types}{type!floating-point!extended}.
+The standard and extended floating-point types
 are collectively termed \defnx{floating-point types}{type!floating-point}.
-The value
-representation of floating-point types is \impldef{value representation of
-floating-point types}.
-\indextext{floating-point type!implementation-defined}%
 \begin{note}
-This document imposes no requirements on the accuracy of
-floating-point operations; see also~\ref{support.limits}.
+Any additional implementation-specific types representing floating-point values
+that are not defined by the implementation to be extended floating-point types
+are not considered to be floating-point types, and
+this document imposes no requirements on them or
+their interactions with floating-point types.
 \end{note}
+Except as specified in \ref{basic.extended.fp},
+the object and value representations and accuracy of operations
+of floating-point types is \impldef{representation of floating-point types}.
 
 \pnum
 Integral and floating-point types are collectively
@@ -5048,6 +5056,90 @@ are called \defnx{fundamental types}{fundamental type}.
 Even if the implementation defines two or more fundamental types to have the
 same value representation, they are nevertheless different types.
 \end{note}
+
+\rSec2[basic.extended.fp]{Optional extended floating-point types}
+
+\pnum
+If the implementation supports an extended floating-point type\iref{basic.fundamental}
+whose properties are specified by
+the ISO/IEC/IEEE 60559 floating-point interchange format binary16,
+then the \grammarterm{typedef-name} \tcode{std::float16_t}
+is defined in the header \libheaderref{stdfloat} and names such a type,
+the macro \mname{STDCPP_FLOAT16_T} is defined\iref{cpp.predefined}, and
+the floating-point literal suffixes \tcode{f16} and \tcode{F16}
+are supported\iref{lex.fcon}.
+
+\pnum
+If the implementation supports an extended floating-point type
+whose properties are specified by
+the ISO/IEC/IEEE 60559 floating-point interchange format binary32,
+then the \grammarterm{typedef-name} \tcode{std::float32_t}
+is defined in the header \libheader{stdfloat} and names such a type,
+the macro \mname{STDCPP_FLOAT32_T} is defined, and
+the floating-point literal suffixes \tcode{f32} and \tcode{F32} are supported.
+
+\pnum
+If the implementation supports an extended floating-point type
+whose properties are specified by
+the ISO/IEC/IEEE 60559 floating-point interchange format binary64,
+then the \grammarterm{typedef-name} \tcode{std::float64_t}
+is defined in the header \libheader{stdfloat} and names such a type,
+the macro \mname{STDCPP_FLOAT64_T} is defined, and
+the floating-point literal suffixes \tcode{f64} and \tcode{F64} are supported.
+
+\pnum
+If the implementation supports an extended floating-point type
+whose properties are specified by
+the ISO/IEC/IEEE 60559 floating-point interchange format binary128,
+then the \grammarterm{typedef-name} \tcode{std::float128_t}
+is defined in the header \libheader{stdfloat} and names such a type,
+the macro \mname{STDCPP_FLOAT128_T} is defined, and
+the floating-point literal suffixes \tcode{f128} and \tcode{F128} are supported.
+
+\pnum
+If the implementation supports an extended floating-point type
+with the properties, as specified by ISO/IEC/IEEE 60559, of
+radix ($b$) of 2,
+storage width in bits ($k$) of 16,
+precision in bits ($p$) of 8,
+maximum exponent ($emax$) of 127, and
+exponent field width in bits ($w$) of 8, then
+the \grammarterm{typedef-name} \tcode{std::bfloat16_t}
+is defined in the header \libheader{stdfloat} and names such a type,
+the macro \mname{STDCPP_BFLOAT16_T} is defined, and
+the floating-point literal suffixes \tcode{bf16} and \tcode{BF16} are supported.
+
+\pnum
+\begin{note}
+A summary of the parameters for each type is given in \tref{basic.extended.fp}.
+The precision $p$ includes the implicit 1 bit at the beginning of the mantissa,
+so the storage used for the mantissa is $p-1$ bits.
+ISO/IEC/IEEE 60559 does not assign a name for a type
+having the parameters specified for \tcode{std::bfloat16_t}.
+\end{note}
+\begin{floattable}
+{Properties of named extended floating-point types}{basic.extended.fp}{llllll}
+\topline
+\lhdr{Parameter} & \chdr{\tcode{float16_t}} & \chdr{\tcode{float32_t}} &
+\chdr{\tcode{float64_t}} & \chdr{\tcode{float128_t}} &
+\rhdr{\tcode{bfloat16_t}} \\
+\capsep
+ISO/IEC/IEEE 60559 name & binary16 & binary32 & binary64 & binary128 & \\
+$k$, storage width in bits & 16 & 32 & 64 & 128 & 16 \\
+$p$, precision in bits & 11 & 24 & 53 & 113 & 8 \\
+$emax$, maximum exponent & 15 & 127 & 1023 & 16383 & 127 \\
+$w$, exponent field width in bits & 5 & 8 & 11 & 15 & 8 \\
+\end{floattable}
+
+\pnum
+\recommended
+Any names that the implementation provides for
+the extended floating-point types described in this subsection
+that are in addition to the names defined in the \libheader{stdfloat} header
+should be chosen to increase compatibility and interoperability
+with the interchange types
+\tcode{_Float16}, \tcode{_Float32}, \tcode{_Float64}, and \tcode{_Float128}
+defined in ISO/IEC TS 18661-3 and with future versions of the C standard.
 
 \rSec2[basic.compound]{Compound types}
 
@@ -5337,7 +5429,7 @@ the type corresponding to the \grammarterm{type-id}
 has the top-level cv-qualifier \keyword{volatile}.
 \end{example}
 
-\rSec2[conv.rank]{Integer conversion rank}%
+\rSec2[conv.rank]{Conversion ranks}%
 \indextext{conversion!integer rank}
 
 \pnum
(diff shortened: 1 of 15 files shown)


source/numerics.tex
```diff
@@ -9029,519 +9034,370 @@
...
   float        cyl_bessel_jf(float nu, float x);
   long double  cyl_bessel_jl(long double nu, long double x);
 
   // \ref{sf.cmath.cyl.bessel.k}, irregular modified cylindrical Bessel functions
-  double       cyl_bessel_k(double nu, double x);
+  @\placeholder{floating-point-type}@ cyl_bessel_k(@\placeholder{floating-point-type}@ nu, @\placeholder{floating-point-type}@ x);
   float        cyl_bessel_kf(float nu, float x);
   long double  cyl_bessel_kl(long double nu, long double x);
 
   // \ref{sf.cmath.cyl.neumann}, cylindrical Neumann functions;
```
JohelEGP, line 9341:
```suggestion
  // \ref{sf.cmath.cyl.neumann}, cylindrical Neumann functions
```

source/numerics.tex
```diff
@@ -9628,15 +9477,23 @@
 \end{note}
 \end{itemdescr}
 
+\begin{itemdecl}
+constexpr @\placeholder{floating-point-type}@ abs(@\placeholder{floating-point-type}@ j);
```
JohelEGP, line 9481:
```suggestion
constexpr @\placeholder{floating-point-type}@ abs(@\placeholder{floating-point-type}@ x);
```

source/numerics.tex
```diff
@@ -459,6 +452,24 @@
...
+\indexlibraryctor{complex}%
+\begin{itemdecl}
+template<class X> constexpr explicit(@\seebelow@) complex(const complex<X>& other);
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\effects
+Initializes the real part with \tcode{other.real()} and
+the imaginary part with \tcode{other.imag()}
```
JohelEGP, line 464:
```suggestion
the imaginary part with \tcode{other.imag()}.
```

JohelEGP (changes requested):

source/basic.tex
```diff
@@ -5392,6 +5484,57 @@
...
+The subrank forms a total order among types with equal ranks.
+The types
+\tcode{std::float16_t},
+\tcode{std::float32_t},
+\tcode{std::float64_t}, and
+\tcode{std::float128_t}\iref{stdfloat.syn}
+have a greater conversion subrank than any standard floating-point type
+with equal conversion rank.
+Otherwise, the conversion subrank order is
+\impldef{floating-point conversion subrank}.
```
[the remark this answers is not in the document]
JohelEGP: After looking further into the index, that is not necessarily the case. Feel free to ignore the comment above.

source/preprocessor.tex
```diff
@@ -1724,6 +1724,45 @@
...
+if and only if the implementation supports
+the ISO/IEC/IEEE 60559 floating-point interchange format binary128
+as an extended floating-point type.
+
+\item
+\indextext{__stdcpp_float16_t__@\mname{STDCPP_BFLOAT16_T}}%
+\mname{STDCPP_BFLOAT16_T}\\
+Defined as the integer literal \tcode{1}
+if and only if the implementation supports an extended floating-point type
+with the properties described in \ref{basic.extended.fp}.
```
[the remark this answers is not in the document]
jensmaurer: I like the suggestion, but that seems to be a deviation from the paper. Leaving for @tkoeppe.

source/support.tex
```diff
@@ -27,6 +27,7 @@
 \ref{support.limits}      & Implementation properties &
   \tcode{<cfloat>}, \tcode{<climits>}, \tcode{<limits>}, \tcode{<version>}    \\ \rowsep
 \ref{cstdint}             & Integer types             &   \tcode{<cstdint>}   \\ \rowsep
+\ref{stdfloat.syn}        & Header \libheader{stdfloat} synopsis & \tcode{<stdfloat>} \\ \rowsep
```
Mick235711, line 30: I don't think there is a pre-existing example of a "Header <xxx> synopsis" being at top-level just under a main library clause (and thus appear in this table), though that's what the paper asked... (which also makes this table looks particularly weird)
Maybe we should still do a [stdfloat] "Floating-point types" with [stdfloat.syn] as subclause (and maybe a one-line general like [cstdint])? Or should we open a new convention here so that in future synopsis can directly appear under a clause?

JohelEGP: > I don't think there is a pre-existing example of a "Header <xxx> synopsis" being at second-level just under a main library clause (and thus appear in this table)

You're right.

source/preprocessor.tex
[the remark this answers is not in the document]
tkoeppe: I'd prefer handling this as a separate issue. Could you open an editorial issue please and we can get feedback from the authors, too?

source/support.tex
tkoeppe: I agree. We should either start a new subsection "Floating-point types" with additional wording, perhaps relating to C, or otherwise perhaps we can rename "Integer types" to "Numeric types" (or "Arithmetic types") and have both <cstdint> and <stdfloat> in it. I'll open an issue,

source/preprocessor.tex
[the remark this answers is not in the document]
JohelEGP: Will do.

source/iostreams.tex
```diff
@@ -4729,6 +4737,64 @@
 \end{codeblock}
 \end{itemdescr}
 
+\begin{itemdecl}
+basic_istream& operator>>(@\placeholder{extended-floating-point-type}@& val);
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
```
tkoeppe, line 4745: Why aren't we using any Elements here? Shouldn't this also be Effects:, and there should be a Returns:?

@dkolsen-pgi?

source/numerics.tex
```diff
@@ -1203,28 +1214,22 @@
...
-        \item Otherwise, if the argument has type \tcode{double} or an integer type,
-              then it is effectively cast to \tcode{complex<\brk{}double>}.
-        \item Otherwise, if the argument has type \tcode{float}, then it is
-              effectively cast to \tcode{complex<float>}.
+\item
+If the argument has a floating-point type \tcode{T},
+then it is effectively cast to \tcode{complex<T>}.
+\item
+Otherwise, if the argument has integer type,
+then it is effectively cast to \tcode{complex<double>}.
```
tkoeppe, line 1222: We should probably use `\keyword` for `double` etc., but we can fix that later.

JohelEGP: I don't think `\keyword` is used on the library clauses.

source/iostreams.tex
dkolsen-pgi: Because none of the other `operator>>` descriptions in [iostream.formatted.arithmetic] use any Elements.  I was just following the existing convention in that section.

dkolsen-pgi:
Section [complex.special] is supposed to be removed in its entirety.  The explicit specializations for `std::complex` are no longer needed with the wording changes in this paper.  There is an easily missed one sentence instruction to remove the section in https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p1467r9.html#complex-wording

source/numerics.tex
```diff
@@ -9029,519 +9034,370 @@
...
-  constexpr double nextafter(double x, double y);
-  constexpr long double nextafter(long double x, long double y);    // see \ref{library.c}
+  constexpr @\placeholder{floating-point-type}@ nextafter(@\placeholder{floating-point-type}@ x, @\placeholder{floating-point-type}@ y);
   constexpr float nextafterf(float x, float y);
   constexpr long double nextafterl(long double x, long double y);
 
-  constexpr float nexttoward(float x, long double y);       // see \ref{library.c}
-  constexpr double nexttoward(double x, long double y);
-  constexpr long double nexttoward(long double x, long double y);   // see \ref{library.c}
+  constexpr @\placeholder{floating-point-type}@ nexttoward(@\placeholder{floating-point-type}@ x, @\placeholder{floating-point-type}@ y);
```
dkolsen-pgi, line 9255: This was the result of a typo when I was doing the mass change to the `<cmath>` synopsis.  The signature of `nexttoward` should be 

`constexpr @\placeholder{floating-point-type}@ nexttoward(@\{placeholder{floating-point-type}@ x, long double y);`

This breaking design change (changing the type of the second parameter of `nexttoward`) was not in the version of the paper that was approved by LEWG.  The change was accidentally introduced without any comment or discussion and without anyone noticing during processing by LWG.

I could fix this with a NB comment.  But I think the mistake is obvious enough that it could be fixed here before the CD is finalized, especially since the author of the mistake swears that it was an honest mistake that no one noticed during the review.

source/numerics.tex
```diff
@@ -9029,519 +9034,370 @@
...
-  constexpr bool isunordered(float x, float y);
-  constexpr bool isunordered(double x, double y);
-  constexpr bool isunordered(long double x, long double y);
+  constexpr int fpclassify(@\placeholder{floating-point-type}@ x);
+  constexpr bool isfinite(@\placeholder{floating-point-type}@ x);
+  constexpr bool isinf(@\placeholder{floating-point-type}@ x);
+  constexpr bool isnan(@\placeholder{floating-point-type}@ x);
+  constexpr bool isnormal(@\placeholder{floating-point-type}@ x);
+  constexpr bool signbit(@\placeholder{floating-point-type}@ x);
+  constexpr bool isgreater(@\placeholder{floating-point-type}@ x, double y);
```
dkolsen-pgi, line 9287: The second parameter of `isgreater` should be `@\placeholder{floating-point-type}@ y`, not `double y`.

source/support.tex
dkolsen-pgi: After shuffling words around in an attempt to make things more clear (with most of the words ending up in [basic.extended.fp]), it ended up with the header `<stdfloat>` being defined entirely by its synopsis.  There were no English words left in the `<stdfloat>` section.  I found this weird, but I didn't know what to do about it.  LWG thought it was okay and didn't ask me to change it.

I have no objection to creating an almost empty section above [stdfloat.syn] so that `<stdfloat>` is organized more like all the other headers.

source/support.tex
```diff
@@ -1436,11 +1435,15 @@
...
-60559.
+\tcode{true} if and only if the type adheres to ISO/IEC/IEEE 60559.
 \begin{footnote}
-ISO/IEC/IEEE 60559:2011 is the same as IEEE 754-2008.
+ISO/IEC/IEEE 60559:2020 is the same as IEEE 754-2019.
 \end{footnote}
+\begin{note}
+The value is \tcode{true} for the types
+\tcode{float16_t}, \tcode{float32_t}, \tcode{float64_t}, or \tcode{float128_t},
+if present\iref{basic.extended.fp}.
```
dkolsen-pgi, line 1445: That's awkward English.  A phrase with "or" is usually considered to be single, but "the types" is plural.  (That's copied from P1467, so that's my fault for writing that.)

I suggest changing "for the types" to "for any of the types", or changing "or" to "and".  Either of those changes makes the English sound better.  I like the first one ("for any of the types") better.  It should be fine to make this change editorially since it is in a note and doesn't change the meaning of the sentence.

source/numerics.tex
jensmaurer: The previous comment misses that a paper is no longer "the author's" once is was approved by the various WG21 stages. @tkoeppe , for you to decide.

jensmaurer:
@tkoeppe, for your convenience, I've pushed the changes discussed by @dkolsen-pgi , except for `nexttoward`.

source/support.tex
tkoeppe: Yes, it's not entirely clear, but we'll take a look after the merging. I'm curious in particular what C is going to do and how this would look after we rebase on C23.

source/iostreams.tex
tkoeppe: Interesting, thanks. Let me open an issue for that, but this certainly makes sense as is.

source/numerics.tex
tkoeppe: Please file an LWG issue for this asap. If we can get LWG sign-off that this was an obvious oversight that can be fixed editorially ("because the intention was clear"), then maybe we can fix this for the CD, otherwise we'll have to fix it through the regular issues process. I'm not sure you even need an NB comment, since we have a standing comment to "fix all issues" anyway, but we should make sure to keep track of this.

source/numerics.tex
tkoeppe: At least not with any form of consistency. @jensmaurer Do you recall if we have a policy/plan for using `\keyword` in the library?

(No immediate action needed here.)

source/numerics.tex
Dani-Hub: BTW @tkoeppe : Is it possible to generate a public accessible link to a PDF of the working paper in flight for issues like this one, whose contents cannot reasonably be delta-ed against N4910 and provide this link to me?

source/numerics.tex
tkoeppe: > Section [complex.special] is supposed to be removed in its entirety. The explicit specializations for `std::complex` are no longer needed with the wording changes in this paper. There is an easily missed one sentence instruction to remove the section in https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p1467r9.html#complex-wording

Thank you --  I _really_ wish all wording would be presented in underline/strikethrough and coloured diffs, _all_ of it. For large changes it's OK to just show a short fragment and them say "and 20 more similar cases" or use "[...]" for the middle bits, but having a visual diff marker per edit is immensely valuable.

Thanks also, Jens!

tkoeppe:
> @tkoeppe, for your convenience, I've pushed the changes discussed by @dkolsen-pgi , except for `nexttoward`.

Thank you so much!

source/support.tex
tkoeppe: Thanks, good call!

source/numerics.tex
tkoeppe: @Dani-Hub Hi, thanks! I was just going to contact you and mention the uploaded artefacts we provide for _every_ precommit check. For example, for Motion 13, see https://github.com/cplusplus/draft/actions/runs/2806916819, and download "draft-snapshot". In general, for any commit with a precommit check, click the green tick to see the action run, then go to "Summary", and you should find the PDF there.

Does that help/work for you?

Dani-Hub: Yes, thank you very much.

tkoeppe: My great pleasure!

source/numerics.tex
jensmaurer: > Do you recall if we have a policy/plan for using \keyword in the library?

We haven't discussed this in any detail, but if the plan for C++ is approximately that of C in that we want all mentions of keywords colored a little bit (or so), then we should use `\keyword` everywhere in prose text and configure `codeblock` accordingly.

Currently, we really want all keywords in the core section to be annotated, because there will be index entries for them.

source/preprocessor.tex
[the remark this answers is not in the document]
JohelEGP: The author's suggestion from https://github.com/cplusplus/draft/issues/5715#issuecomment-1207321108.
```suggestion
if and only if the implementation supports an extended floating-point type
with the properties of the \grammarterm{typedef-name} \tcode{std::bfloat16_t} described in \ref{basic.extended.fp}.
```

PR #5678, opened 2022-07-28
P2438R2 std::string::substr() &&

### LINKED ISSUE #5611: [2022-07 LWG Motion 24] P2438R2 std::string::substr() &&
[P2438R2](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2438r2.html)

jensmaurer:
Fixes #5611
Fixes cplusplus/papers#1104

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -201,6 +201,28 @@ auto t = tiny();
 std::format("{}", t.bit);       // ill-formed, previously returned \tcode{"0"}
 \end{codeblock}
 
+\rSec2[diff.cpp20.strings]{\ref{strings}: strings library}
+
+\diffref{string.classes}
+\change
+Additional rvalue overload for the \tcode{substr} member function and
+the corresponding constructor.
+\rationale
+Improve efficiency of operations on rvalues.
+\effect
+Valid \CppXX{} code that created a substring
+by calling \tcode{substr} (or the corresponding constructor)
+on an xvalue expression with type \tcode{S}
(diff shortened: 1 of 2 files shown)


source/compatibility.tex
```diff
@@ -112,6 +112,28 @@
...
+\effect
+Valid \CppXX{} code that created a substring
+by calling \tcode{substr} (or the corresponding constructor)
+on an xvalue expression with type \tcode{S}
+that is a specialization of \tcode{basic_string}
+may change meaning in this revision of \Cpp{}.
+\begin{codeblock}
+std::string s1 = "some long string that forces allocation", s2 = s1;
+std::move(s1).substr(10, 5);
+assert(s1 == s2);       // unspecified, previously guaranteed to be true
```
JohelEGP, line 132:
```suggestion
assert(s1 == s2);       // unspecified, previously guaranteed to be \tcode{true}
```

source/compatibility.tex
```diff
@@ -112,6 +112,28 @@
...
+std::string s3(std::move(s2), 10, 5);
+assert(s1 == s2);       // unspecified, previously guaranteed to be true
```
JohelEGP, line 134:
```suggestion
assert(s1 == s2);       // unspecified, previously guaranteed to be \tcode{true}
```

JohelEGP (changes requested):

jwakely (dismissed):
I agree that `true` should use code font, otherwise this looks correct.

Issue #5605, opened 2022-07-16
[2022-07 LWG Motion 18] P2374R4 views::cartesian_product


wg21bot:
[P2374R4](https://wg21.link/p2374r4)

tkoeppe:
Note to self: fix up `tuple-or-pair` as per #5694.

PR #5694, opened 2022-07-30
P2165R4 Compatibility between tuple and tuple-like objects

### LINKED ISSUE #5599: [2022-07 LWG Motion 12] P2165R4 Compatibility between tuple and tuple-like objects
[P2165R4](https://wg21.link/p2165r4)

burblebee:
Fixes #5599

The 2nd part of this (replacing all usages of tuple-or-pair by tuple in the range clause) should be applied after P2374R4 #5605 as per the Editor's note at end of P2165R4.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -9308,14 +9308,14 @@ template<class InputIterator>
     typename iterator_traits<InputIterator>::value_type;                // \expos
 template<class InputIterator>
   using @\placeholder{iter-key-type}@ = remove_const_t<
-    typename iterator_traits<InputIterator>::value_type::first_type>;   // \expos
+    tuple_element_t<0, @\exposid{iter-value-type}@<InputIterator>>>;                // \expos
 template<class InputIterator>
   using @\placeholder{iter-mapped-type}@ =
-    typename iterator_traits<InputIterator>::value_type::second_type;   // \expos
+    tuple_element_t<1, @\exposid{iter-value-type}@<InputIterator>>;                 // \expos
 template<class InputIterator>
   using @\placeholder{iter-to-alloc-type}@ = pair<
-    add_const_t<typename iterator_traits<InputIterator>::value_type::first_type>,
-    typename iterator_traits<InputIterator>::value_type::second_type>;  // \expos
+    add_const_t<tuple_element_t<0, @\exposid{iter-value-type}@<InputIterator>>>,
+    tuple_element_t<1, @\exposid{iter-value-type}@<InputIterator>>>;                // \expos
 template<ranges::@\libconcept{input_range}@ Range>
   using @\exposid{range-key-type}@ =
     remove_const_t<typename ranges::range_value_t<Range>::first_type>;  // \expos
--- a/source/support.tex
+++ b/source/support.tex
@@ -718,6 +718,8 @@ the values of these macros with greater values.
 #define @\defnlibxname{cpp_lib_to_underlying}@                     202102L // also in \libheader{utility}
 #define @\defnlibxname{cpp_lib_transformation_trait_aliases}@      201304L // also in \libheader{type_traits}
 #define @\defnlibxname{cpp_lib_transparent_operators}@             201510L // also in \libheader{memory}, \libheader{functional}
+#define @\defnlibxname{cpp_lib_tuple_like}@                        202207L
+  // also in \libheader{utility}, \libheader{tuple}, \libheader{map}, \libheader{unordered_map}
 #define @\defnlibxname{cpp_lib_tuple_element_t}@                   201402L // also in \libheader{tuple}
 #define @\defnlibxname{cpp_lib_tuples_by_type}@                    201304L // also in \libheader{utility}, \libheader{tuple}
 #define @\defnlibxname{cpp_lib_type_identity}@                     201806L // also in \libheader{type_traits}
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -684,6 +684,8 @@ namespace std {
       constexpr explicit(@\seebelow@) pair(pair<U1, U2>&& p);
     template<class U1, class U2>
       constexpr explicit(@\seebelow@) pair(const pair<U1, U2>&& p);
+    template<@\exposconcept{pair-like}@ P>
+      constexpr explicit(@\seebelow@) pair(P&& p);
     template<class... Args1, class... Args2>
       constexpr pair(piecewise_construct_t,
                      tuple<Args1...> first_args, tuple<Args2...> second_args);
@@ -700,6 +702,10 @@ namespace std {
       constexpr pair& operator=(pair<U1, U2>&& p);
     template<class U1, class U2>
       constexpr const pair& operator=(pair<U1, U2>&& p) const;
+    template<@\exposconcept{pair-like}@ P>
+      constexpr pair& operator=(P&& p);
+    template<@\exposconcept{pair-like}@ P>
+      constexpr const pair& operator=(P&& p) const;
 
     constexpr void swap(pair& p) noexcept(@\seebelow@);
     constexpr void swap(const pair& p) const noexcept(@\seebelow@);
@@ -825,6 +831,7 @@ template<class U1, class U2> constexpr explicit(@\seebelow@) pair(pair<U1, U2>& 
 template<class U1, class U2> constexpr explicit(@\seebelow@) pair(const pair<U1, U2>& p);
 template<class U1, class U2> constexpr explicit(@\seebelow@) pair(pair<U1, U2>&& p);
 template<class U1, class U2> constexpr explicit(@\seebelow@) pair(const pair<U1, U2>&& p);
+template<@\exposconcept{pair-like}@ P> constexpr explicit(@\seebelow@) pair(P&& p);
 \end{itemdecl}
 
 \begin{itemdescr}
@@ -1057,6 +1064,70 @@ template<class U1, class U2> constexpr pair& operator=(pair<U1, U2>&& p);
 \effects
 Assigns to \tcode{first} with \tcode{std::forward<U1>(p.first)}
 and to \tcode{second} with\\ \tcode{std::forward<U2>(p.second)}.
+
+\pnum
+\returns
+\tcode{*this}.
+\end{itemdescr}
+
+\indexlibrarymember{operator=}{pair}%
+\begin{itemdecl}
+template<@\exposconcept{pair-like}@ P> constexpr pair& operator=(P&& p);
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\constraints
+\begin{itemize}
+\item
+\tcode{\exposconcept{different-from}<P, pair>}\iref{range.utility.helpers}
+is \tcode{true},
+\item
+\tcode{remove_cvref_t<P>} is not a specialization of \tcode{ranges::subrange},
+\item
+\tcode{is_assignable_v<T1\&, decltype(get<0>(std::forward<P>(p)))>}
+is \tcode{true}, and
+\item
+\tcode{is_assignable_v<T2\&, decltype(get<1>(std::forward<P>(p)))>}
+is \tcode{true}.
+\end{itemize}
+
+\pnum
+\effects
+Assigns \tcode{get<0>(std::forward<P>(p))} to \tcode{first} and
+\tcode{get<1>(std::forward<P>(p))} to \tcode{sec\-ond}.
+
+\pnum
+\returns
+\tcode{*this}.
+\end{itemdescr}
+
+\indexlibrarymember{operator=}{pair}%
+\begin{itemdecl}
+template<@\exposconcept{pair-like}@ P> constexpr const pair& operator=(P&& p) const;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\constraints
+\begin{itemize}
+\item
+\tcode{\exposconcept{different-from}<P, pair>}\iref{range.utility.helpers}
+is \tcode{true},
+\item
+\tcode{remove_cvref_t<P>} is not a specialization of \tcode{ranges::subrange},
+\item
+\tcode{is_assignable_v<const T1\&, decltype(get<0>(std::forward<P>(p)))>}
+is \tcode{true}, and
+\item
+\tcode{is_assignable_v<const T2\&, decltype(get<1>(std::forward<P>(p)))>}
+is \tcode{true}.
+\end{itemize}
+
+\pnum
+\effects
+Assigns \tcode{get<0>(std::forward<P>(p))} to \tcode{first} and
+\tcode{get<1>(std::forward<P>(p))} to \tcode{sec\-ond}.
 
 \pnum
 \returns
@@ -1359,18 +1430,19 @@ namespace std {
   template<class... Types>
     class tuple;
 
-  template<class... TTypes, class... UTypes, template<class> class TQual,
-           template<class> class UQual>
-    requires requires { typename tuple<common_reference_t<TQual<TTypes>, UQual<UTypes>>...>; }
-  struct basic_common_reference<tuple<TTypes...>, tuple<UTypes...>, TQual, UQual> {
-    using type = tuple<common_reference_t<TQual<TTypes>, UQual<UTypes>>...>;
-  };
+  // \ref{tuple.like}, concept \exposconcept{tuple-like}
+  template <typename T>
+    concept @\exposconcept{tuple-like}@ = @\seebelownc@;         // \expos
+  template <typename T>
+    concept @\defexposconcept{pair-like}@ =                     // \expos
+      @\exposconcept{tuple-like}@<T> && tuple_size_v<remove_cvref_t<T>> == 2;
 
-  template<class... TTypes, class... UTypes>
-    requires requires { typename tuple<common_type_t<TTypes, UTypes>...>; }
-  struct common_type<tuple<TTypes...>, tuple<UTypes...>> {
-    using type = tuple<common_type_t<TTypes, UTypes>...>;
-  };
+  // \ref{tuple.common.ref}, \tcode{common_reference} related specializations
+  template<@\exposconceptnc{tuple-like}@ TTuple, @\exposconceptnc{tuple-like}@ UTuple,
+           template<class> class TQual, template<class> class UQual>
+    struct basic_common_reference<TTuple, UTuple, TQual, UQual>;
+  template<@\exposconceptnc{tuple-like}@ TTuple, @\exposconceptnc{tuple-like}@ UTuple>
+    struct common_type<TTuple, UTuple>;
 
   // \ref{tuple.creation}, tuple creation functions
   inline constexpr @\unspec@ ignore;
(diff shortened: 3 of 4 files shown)


source/utilities.tex
```diff
@@ -2606,52 +2827,77 @@
...
 \pnum
 \remarks
+\begin{itemize}
+\item
 The elementary comparisons are performed in order from the
 zeroth index upwards.  No comparisons or element accesses are
 performed after the first equality comparison that evaluates to
 \tcode{false}.
+\item
+The second overload is to be found via argument-dependent lookup only.
```
JohelEGP, line 2864:
```suggestion
The second overload is to be found via argument-dependent lookup\iref{basic.lookup.argdep} only.
```

source/utilities.tex
```diff
@@ -2664,6 +2910,82 @@
...
+In the descriptions that follow:
+\begin{itemize}
+\item
+Let \tcode{TTypes} be a pack formed by
+the sequence of \tcode{tuple_element_t<$i$, TTuple>}
+for every integer $0 \leq i < \tcode{tuple_size_v<TTuple>}$.
+
+\item
+Let \tcode{UTypes} be a pack formed by
+the sequence of \tcode{tuple_element_t<$i$, UTuple>}
```
JohelEGP, line 2925: Is it OK to not explicitly say "ordered sequence", like in [tuple.cat]?

source/utilities.tex
```diff
@@ -2664,6 +2910,82 @@
...
+\tcode{is_same_v<UTuple, decay_t<UTuple>>} is \tcode{true}.
+\item
+\tcode{tuple_size_v<TTuple>} equals \tcode{tuple_size_v<UTuple>}.
+\item
+\tcode{tuple<common_reference_t<TQual<TTypes>, UQual<UTypes>>...>}
+denotes a type.
+\end{itemize}
+The member \grammarterm{typedef-name} \tcode{type} denotes the type
+\begin{codeblock}
+tuple<common_reference_t<TQual<TTypes>, UQual<UTypes>>...>.
```
JohelEGP, line 2957:
```suggestion
tuple<common_reference_t<TQual<TTypes>, UQual<UTypes>>...>
```

source/utilities.tex
```diff
@@ -2664,6 +2910,82 @@
 result of the comparison.
 \end{note}
 
+\rSec2[tuple.common.ref]{\tcode{common_reference} specialization}
```
JohelEGP, line 2913:
```suggestion
\rSec2[tuple.common.ref]{\tcode{common_reference}-related specializations}
```

source/utilities.tex
```diff
@@ -1357,18 +1427,17 @@
...
+  template <typename T>
+    concept @\defexposconcept{pair-like}@ =                     // \expos
+      @\exposconcept{tuple-like}@<T> && tuple_size_v<remove_cvref_t<T>> == 2;
 
-  template<class... TTypes, class... UTypes>
-    requires requires { typename tuple<common_type_t<TTypes, UTypes>...>; }
-  struct common_type<tuple<TTypes...>, tuple<UTypes...>> {
-    using type = tuple<common_type_t<TTypes, UTypes>...>;
-  };
+  template<@\exposconcept{tuple-like}@ TTuple, @\exposconcept{tuple-like}@ UTuple,
```
JohelEGP, line 1436:
```suggestion
  // \ref{tuple.common.ref}, \tcode{common_reference}-related specializations
  template<@\exposconcept{tuple-like}@ TTuple, @\exposconcept{tuple-like}@ UTuple,
```

source/containers.tex
```diff
@@ -9296,14 +9296,15 @@
...
 template<class InputIterator>
   using @\placeholder{iter-mapped-type}@ =
-    typename iterator_traits<InputIterator>::value_type::second_type;   // \expos
+    tuple_element_t<1, @\exposid{iter-value-type}@<InputIterator>>;                 // \expos
 template<class InputIterator>
   using @\placeholder{iter-to-alloc-type}@ = pair<
-    add_const_t<typename iterator_traits<InputIterator>::value_type::first_type>,
-    typename iterator_traits<InputIterator>::value_type::second_type>;  // \expos
+    add_const_t<tuple_element_t<0, @\exposid{iter-value-type}@<InputIterator>>>,
+    tuple_element_t<1, @\exposid{iter-value-type}@<InputIterator>>>;                // \expos
```
JohelEGP, line 9306: Why didn't these reuse _`iter-mapped-type`_?

source/containers.tex
```diff
@@ -9296,14 +9296,15 @@
...
+
```
JohelEGP, line 9307:
```suggestion
```

source/utilities.tex
```diff
@@ -1062,6 +1069,69 @@
...
+\begin{itemdescr}
+\pnum
+\constraints
+\begin{itemize}
+\item
+\tcode{@\exposconcept{different-from}@<P, pair>} is \tcode{true},
+\item
+\tcode{remove_cvref_t<P>} is not a specialization of \tcode{ranges::subrange},
+\item
+\tcode{is_assignable_v<T1\&, decltype(get<0>(std::forward<P>(p)))>}
```
JohelEGP, line 1118:
```suggestion
\tcode{is_assignable_v<const T1\&, decltype(get<0>(std::forward<P>(p)))>}
```

source/utilities.tex
```diff
@@ -1062,6 +1069,69 @@
...
+is \tcode{true}, and
+\item
+\tcode{is_assignable_v<T2\&, decltype(get<1>(std::forward<P>(p)))>}
```
JohelEGP, line 1121:
```suggestion
\tcode{is_assignable_v<const T2\&, decltype(get<1>(std::forward<P>(p)))>}
```

JohelEGP (changes requested):

source/support.tex
```diff
@@ -712,6 +712,7 @@
 #define @\defnlibxname{cpp_lib_to_underlying}@                     202102L // also in \libheader{utility}
 #define @\defnlibxname{cpp_lib_transformation_trait_aliases}@      201304L // also in \libheader{type_traits}
 #define @\defnlibxname{cpp_lib_transparent_operators}@             201510L // also in \libheader{memory}, \libheader{functional}
+#define @\defnlibxname{cpp_lib_tuple_like}@                        202207L // also in \libheader{utility}, \libheader{tuple}
```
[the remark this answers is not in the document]
burblebee: The paper contradicts itself - it mentions all 4 headers at the top, but only the 2 at the bottom under "Feature test macros".  Which is correct? 

I've added a separate edit with this change to be either applied or not.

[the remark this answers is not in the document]
cor3ntin: (Author) It should be the 4 headers. Please ignore the second out dated instructions at the bottom. Thanks

source/utilities.tex
```diff
@@ -1357,18 +1427,17 @@
...
     class tuple;
 
-  template<class... TTypes, class... UTypes, template<class> class TQual,
-           template<class> class UQual>
-    requires requires { typename tuple<common_reference_t<TQual<TTypes>, UQual<UTypes>>...>; }
-  struct basic_common_reference<tuple<TTypes...>, tuple<UTypes...>, TQual, UQual> {
-    using type = tuple<common_reference_t<TQual<TTypes>, UQual<UTypes>>...>;
-  };
+  template <typename T>
+    concept @\defexposconcept{tuple-like}@ = @\seebelownc@;         // \expos
```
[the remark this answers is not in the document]
burblebee: I'm confused - this *is* an exposition-only concept.  What am I missing?

source/utilities.tex
```diff
@@ -1448,6 +1521,23 @@
...
+template <typename T>
+  concept @\defexposconcept{tuple-like}@ = @\seebelownc@;           // \expos
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+A type \tcode{T} models and satisfies
+the exposition-only concept \exposconcept{tuple-like}
+if \tcode{std::remove_cvref_t<T>} is a specialization of
+\tcode{array}, \tcode{pair}, \tcode{tuple} or \tcode{ranges::subrange}.
```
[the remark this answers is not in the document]
burblebee: fixed

source/utilities.tex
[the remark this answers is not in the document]
JohelEGP: `\libconcept` and `\exposconcept` index uses of concepts. However, header pseudo-declarations of non exposition-only concepts don't use `\libconcept` and thus are not indexed. By using `\exposid` instead for exposition-only concepts, we achieve the same result.

source/utilities.tex
```diff
@@ -2170,6 +2324,80 @@
...
+\tcode{@\exposconcept{different-from}@<UTuple, tuple>}\iref{range.utility.helpers}
+is \tcode{true},
+
+\item
+\tcode{remove_cvref_t<UTuple>}
+is not a specialization of \tcode{ranges::subrange},
+
+\item
+\tcode{sizeof...(Types)}
+equals \tcode{tuple_size_v<remove_cvref_t<UTuple>>} and,
```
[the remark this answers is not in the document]
burblebee: done

source/utilities.tex
```diff
@@ -2170,6 +2324,80 @@
...
+
+\item
+\tcode{is_assignable_v<$\tcode{T}_1$\&, decltype(get<$i$>(std::forward<UTuple>(u)))>}
```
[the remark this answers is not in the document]
burblebee: good catch! fixed

source/utilities.tex
```diff
@@ -2170,6 +2324,80 @@
...
+\tcode{@\exposconcept{different-from}@<UTuple, tuple>}\iref{range.utility.helpers}
+is \tcode{true},
+
+\item
+\tcode{remove_cvref_t<UTuple>}
+is not a specialization of \tcode{ranges::subrange},
+
+\item
+\tcode{sizeof...(Types)}
+equals \tcode{tuple_size_v<remove_cvref_t<UTuple>>} and,
```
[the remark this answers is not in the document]
burblebee: done

source/support.tex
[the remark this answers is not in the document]
JohelEGP: Please, use my suggestion. I intentionally put them all on the next line as done for the other macros.

source/utilities.tex
```diff
@@ -2170,6 +2324,80 @@
...
+
+\item
+\tcode{is_assignable_v<const $\tcode{T}_1$\&, decltype(get<$i$>(std::forward<UTuple>(u)))>}
```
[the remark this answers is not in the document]
burblebee: fixed

source/utilities.tex
```diff
@@ -2170,6 +2324,80 @@
...
+is \tcode{true} for all $i$.
+\end{itemize}
+
+\pnum
+\effects
+For all $i$, assigns
+\tcode{get<$i$>(std::forward<UTuple>(u)) to get<$i$>(*this)}.
```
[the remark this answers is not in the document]
burblebee: fixed

source/utilities.tex
```diff
@@ -2369,9 +2589,10 @@
 \end{codeblock}
 \end{itemdescr}
 
+% FIXME: This doesn't match P2165r4
```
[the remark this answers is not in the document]
burblebee: fixed

source/utilities.tex
burblebee: fixed

source/containers.tex
burblebee: fixed

source/utilities.tex
burblebee: fixed

source/utilities.tex
burblebee: fixed

source/utilities.tex
burblebee: done

source/utilities.tex
```diff
@@ -2606,52 +2827,77 @@
...
 if (auto c = @\placeholder{synth-three-way}@(get<0>(t), get<0>(u)); c != 0) return c;
 return @$\tcode{t}_\mathrm{tail}$@ <=> @$\tcode{u}_\mathrm{tail}$@;
 \end{codeblock}
-where $\tcode{r}_\mathrm{tail}$ for some tuple \tcode{r}
+where $\tcode{r}_\mathrm{tail}$ for some \tcode{r}
 is a tuple containing all but the first element of \tcode{r}.
+
+\pnum
+\remarks
+The second overload is to be found via argument-dependent lookup only.
```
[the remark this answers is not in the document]
burblebee: done

source/containers.tex
burblebee: Not an editorial issue or part of this paper - please send to lwgchair. Ok to resolve here?

source/utilities.tex
burblebee: Added, but "`common_reference`-related" isn't a term, so didn't include the "-".

source/utilities.tex
burblebee: Added, but "`common_reference`-related" isn't a term, so didn't include the "-".

source/utilities.tex
```diff
@@ -1357,18 +1427,17 @@
   template<class... Types>
     class tuple;
 
-  template<class... TTypes, class... UTypes, template<class> class TQual,
-           template<class> class UQual>
-    requires requires { typename tuple<common_reference_t<TQual<TTypes>, UQual<UTypes>>...>; }
-  struct basic_common_reference<tuple<TTypes...>, tuple<UTypes...>, TQual, UQual> {
-    using type = tuple<common_reference_t<TQual<TTypes>, UQual<UTypes>>...>;
-  };
+  template <typename T>
```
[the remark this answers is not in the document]
burblebee: Reference added, but with \exposconcept.  Will change all \exposconcept's to \exposconcept's in a separate commit if needed.

source/support.tex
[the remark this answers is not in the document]
burblebee: Thanks for the clarification - I changed the edit to a "FIXUP" so will be part of the paper.

source/utilities.tex
burblebee: ? not an edit issue.

source/utilities.tex
```diff
@@ -2290,62 +2518,54 @@
 
 \indexlibraryglobal{tuple_cat}
 \begin{itemdecl}
-template<class... Tuples>
+template<@\exposconcept{tuple-like}@... Tuples>
   constexpr tuple<CTypes...> tuple_cat(Tuples&&... tpls);
 \end{itemdecl}
 
 \begin{itemdescr}  % NOCHECK: order
```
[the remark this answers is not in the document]
burblebee: done

source/utilities.tex
JohelEGP: Good point. I meant "Specializations related to `common_reference`", but the extensive use of hyphens in standard terms certainly makes it confusing.

source/support.tex
```diff
@@ -712,6 +712,8 @@
 #define @\defnlibxname{cpp_lib_to_underlying}@                     202102L // also in \libheader{utility}
 #define @\defnlibxname{cpp_lib_transformation_trait_aliases}@      201304L // also in \libheader{type_traits}
 #define @\defnlibxname{cpp_lib_transparent_operators}@             201510L // also in \libheader{memory}, \libheader{functional}
+#define @\defnlibxname{cpp_lib_tuple_like}@                        202207L // also in \libheader{utility}, \libheader{tuple}, \libheader{map},
```
JohelEGP, line 715: https://github.com/cplusplus/draft/pull/5694#discussion_r933822832 is still missing as suggested.

source/utilities.tex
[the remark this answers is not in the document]
JohelEGP: If you don't agree, this should at least be
```suggestion
    concept @\exposconcept{tuple-like}@ = @\seebelownc@;         // \expos
```
since it's not a definition.

source/utilities.tex
```diff
@@ -1448,6 +1521,23 @@
 }
 \end{codeblock}
 
+\rSec2[tuple.like]{Concept \exposconcept{tuple-like}}
```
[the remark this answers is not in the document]
JohelEGP: I don't remember why the indexing commands are not used in `\rSec`s.
There are not matches for
```bash
$ grep 'rSec.*libconcept' *.tex
$ grep 'rSec.*exposconcept' *.tex
```
but
```terminal
$ grep 'rSec.*cname' *.tex
concepts.tex:\rSec2[concept.same]{Concept \cname{same_as}}
concepts.tex:\rSec2[concept.derived]{Concept \cname{derived_from}}
concepts.tex:\rSec2[concept.convertible]{Concept \cname{convertible_to}}
concepts.tex:\rSec2[concept.commonref]{Concept \cname{common_reference_with}}
concepts.tex:\rSec2[concept.common]{Concept \cname{common_with}}
[...]
```
abounds, so I suggest applying https://github.com/cplusplus/draft/pull/5694/files#r933830429.

source/utilities.tex
```diff
@@ -1062,6 +1069,69 @@
...
+\begin{itemdecl}
+template<@\exposconcept{pair-like}@ P> constexpr const pair& operator=(P&& p) const;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\constraints
+\begin{itemize}
+\item
+\tcode{\exposconcept{different-from}<P, pair>} is \tcode{true},
```
JohelEGP, line 1114:
```suggestion
\tcode{\exposconcept{different-from}<P, pair>}\iref{range.utility.helpers}
is \tcode{true},
```

JohelEGP (changes requested):

source/utilities.tex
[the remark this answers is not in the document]
burblebee: I do see precedence for using \exposconcept (see similar cases in memory.tex), so I went with that.  I'd suggest opening an edit issue to have these changed to \exposid if that's what you'd prefer.

source/utilities.tex
```diff
@@ -1448,6 +1521,23 @@
 }
 \end{codeblock}
 
+\rSec2[tuple.like]{Concept \exposconcept{tuple-like}}
+
+% FIXME: \exposconcept{} is not allowed in \indexlibrarymember
+\indexlibrarymember{\placeholder{tuple-like}}{tuple}%
```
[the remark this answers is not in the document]
burblebee: done

source/utilities.tex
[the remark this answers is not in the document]
burblebee: fixed - sorry if I missed this earlier

source/support.tex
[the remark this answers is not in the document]
burblebee: comment moved to separate line (sorry, didn't see that suggestion until now)

source/support.tex
burblebee: done

source/utilities.tex
burblebee: done. Some might say the reference in the previous itemdescr is sufficient tho, so this change (as well as any deviations from the paper for that matter) could still get rejected in the final merge.

cor3ntin: As the initial author, I'm fine with this change.
As a side note, i think we might want (as an editorial thing, after c++23) move `different-from` to utility so that it's easier to use from more places

tkoeppe:
@jwakely Could you please comment on the Editor's Note in the paper that asks whether we need to modify the container requirements, and says,

> we probably want to requiere `is_constructible<value_type, T>` where `T` is either the type passed to insert, or
the InputIterator’s `value_type`. Currently, we only seem to `require convertible_to`, which may
not be sufficient?. An alternative is to add explicit insert overloads for pair-like objects

cor3ntin:
@tkoeppe Sorry, I guess i should have removed that note, it was more "lwg should discuss that" rather than actual editor guidance. I don't actually remember that it was discussed though.

I don't think there is an issue here either way, I'm struggling to imagine a case where `pair` would not be constructible from `tuple-like`. We may still want to add new overloads to associative containers for performance, maybe (i need to think about it more) but this would be nb comment material.

tkoeppe:
@cor3ntin: Thank you! I'll proceed as-is then, and we can follow up via issues later if and when necessary.

tkoeppe:
> The 2nd part of this (replacing all usages of tuple-or-pair by tuple in the range clause) should be applied after P2374R4 https://github.com/cplusplus/draft/issues/5605 as per the Editor's note at end of P2165R4.

I will add a corresponding addition to that motion when we get to it; for now, this paper is applied in its entirety.

JohelEGP:
Also resolves https://github.com/cplusplus/papers/issues/876.

PR #5702, opened 2022-08-04
Add examples in [format.string.escaped]

cor3ntin:
The s2 examples is reproduced by using the cyrillic latex
package.

The s6 example cannot be sanely reproduced with the
latex engine used.

Instead, use N + combining tilde as an example.
To make sure the LHS does not appear to be related to
normalization, spell out the combining grapheme extend
codepoint.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -25,7 +25,7 @@ jobs:
         run: sudo apt-get update
 
       - name: install
-        run: sudo apt-get install latexmk texlive-latex-recommended texlive-latex-extra texlive-fonts-recommended lmodern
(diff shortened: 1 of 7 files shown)


tkoeppe:
@cor3ntin: I think the following can work:

* Make a new document with `standalone` class, 10pt, for xelatex. Use the "NewComputerModern10" or CMU font (which should be opentype and contain cyrillic glyhs, and are based on CM and thus should match the Standard). Typeset the monospaced "hello Victor" phrase with that.
* Do something analogous for the emojis, but I haven't managed to get that to work in xelatex yet. Maybe use the "Noto Color Emoji" font?
* Then we can embed those standalone-PDFs inline with `\includegraphics`.

tkoeppe:
(I've got this working locally using lualatex and the Noto Emoji font.)

PR #5703, opened 2022-08-04
[strings.general,thread.general] Remove nonexistent header and fix some style error

Mick235711:
[strings.general]: Add `<string>` to the String classes subclause
[thread.general]: Remove nonexistent header `<atomic_ref>` and add a comma.

### DIFF
--- a/source/strings.tex
+++ b/source/strings.tex
@@ -20,7 +20,7 @@ as summarized in \tref{strings.summary}.
 \begin{libsumtab}[x{2.1in}]{Strings library summary}{strings.summary}
 \ref{char.traits}     & Character traits                    & \tcode{<string>}  \\
 \ref{string.view}     & String view classes                 & \tcode{<string_view>} \\ \rowsep
-\ref{string.classes}  & String classes                      &   \\ \rowsep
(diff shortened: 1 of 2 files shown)


jwakely (approved):

jensmaurer (changes requested):
These are two entirely different situations. Please split into two commits.

Mick235711:
> These are two entirely different situations. Please split into two commits.

Done.

jensmaurer:
In the commit description, I'm not seeing what "and fix some style error" would refer to for the second commit. It now exclusively removes a non-existing header from the intro table.

Mick235711:
> In the commit description, I'm not seeing what "and fix some style error" would refer to for the second commit. It now exclusively removes a non-existing header from the intro table.

That part of the message is for adding a comma. The rendered draft looks like
image
And that row is already missing comma as separator. So that commit, in addition to remove the header, also fix this style error.

jensmaurer (approved):

tkoeppe:
I'll take this and fix up the commit messages -- thank you!

PR #5721, opened 2022-08-08
[range.join.with.sentinel] Fix template format

hewillk:
This seems to be the only place in `<ranges>` and other libraries where there is a space between `template` and `<`.


JohelEGP:
@jensmaurer I wonder what happened to the check for this.

CaseyCarter (approved):

jensmaurer (approved):

jensmaurer:
The check triggers on `template <class`, only.

JohelEGP:
Ah, no wonder all the `template <ranges::` of the motions didn't trigger it.

PR #5725, opened 2022-08-14
[mdspan.{overview,extents.ctor}] Increase reuse of definitions

tkoeppe:
The definition of "rank index" can be used at least once more. We might even be able to use it if we can say that `extents_type` is an index space.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -15083,6 +15083,10 @@ the product of $U_i - L_i$ for each dimension $i$
 if its rank is greater than 0, and 1 otherwise.
 
 \pnum
+An integer $r$ is a \defn{rank index} of an index space $S$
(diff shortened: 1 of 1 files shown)


source/containers.tex
```diff
@@ -15331,9 +15331,13 @@
 \end{itemdecl}
 
 \begin{itemdescr}
+\pnum
+Let \tcode{N} be \tcode{sizeof...(OtherIndexTypes)},
+and let \tcode{exts_arr} be
+\tcode{array<index_type, N>\{static_cast<\\index_type>(std::move(exts))...\}}.
```
JohelEGP, line 15337: What is `\\`?

tkoeppe: A linebreak.

JohelEGP (approved):

PR #5744, opened 2022-08-18
[range.zip.transform.view] Fix typo: mmove_constructible.

Eelis:


source/ranges.tex
```diff
@@ -10039,7 +10039,7 @@
 \indexlibrarymember{size}{zip_transform_view}%
```
JohelEGP, line 10039: This looks wrong, too.

JohelEGP: Ah, it was just a lack of context.

PR #5747, opened 2022-08-19
[dcl.fct.def.default] Elaborate on the difference of two declarations

tkoeppe:
This makes it clear that T_1 and T_2 may differ because of the present
rule for the purposes of the blanket statement "other than as allowed
by the preceding rules" futher below.


tkoeppe:
@hubert-reinterpretcast: Voila ^^^. Thank you for the suggestion!

hubert-reinterpretcast (approved):
LGTM; thank you!

PR #5748, opened 2022-08-19
[range.as.rvalue.view] Fix curly brace format

hewillk:
Should this be detected?

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -4467,8 +4467,7 @@ ranges::copy(words | views::as_rvalue, back_inserter(new_words));
 namespace std::ranges {
   template<@\libconcept{input_range}@ V>
     requires @\libconcept{view}@<V>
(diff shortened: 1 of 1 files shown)


tkoeppe:
Before we do all these one by one, could you maybe have a look at the clause as a whole to see if there are other style violations like this?

> Should this be detected?

Can you describe such a check in a way we can run in the check script?

hewillk:
OK, I'll check them one by one.

> Can you describe such a check in a way we can run in the check script?

This might take some thought.

hewillk:
Also fixed inconsistent constraint ordering with synopsis.

tkoeppe:
@brevzin Could you please take a look?

brevzin:
So this swaps the two constraints to be consistent with the synopsis (oops) and moves the open brace to the start of the previous line? Looks good.

tkoeppe:
@brevzin Great, thanks! Yes, I wanted to check that that was indeed an error in the paper, and that this fix is correct.

PR #5751, opened 2022-08-19
[format.string.escaped] Fix invalid examples

mordante:
While implementing new features introduced by P2286R8 Formatting Ranges
I noticed some issues in the examples. These issues are in the paper
too.

For s3 the alternative would be to adjust the output instead of the
input.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -15367,15 +15367,15 @@ then \placeholder{C} is appended unchanged.
 %% FIXME: their Unicode characters are not available in our font (Latin Modern).
 \begin{example}
 \begin{codeblock}
-string s0 = format("[{}]", "h\tllo");                   // \tcode{s0} has value: \tcode{[h    llo]}
-string s1 = format("[{:?}]", "h\tllo");                 // \tcode{s1} has value: \tcode{["h\textbackslash tllo"]}
-string s3 = format("[{:?}] [{:?}]", '\'', '"');         // \tcode{s3} has value: \tcode{['\textbackslash '', '"']}
+string s0 = format("[{}]", "h\tllo");               // \tcode{s0} has value: \tcode{[h    llo]}
+string s1 = format("[{:?}]", "h\tllo");             // \tcode{s1} has value: \tcode{["h\textbackslash tllo"]}
+string s3 = format("[{:?}, {:?}]", '\'', '"');      // \tcode{s3} has value: \tcode{['\textbackslash '', '"']}
 
 // The following examples assume use of the UTF-8 encoding
 string s4 = format("[{:?}]", string("\0 \n \t \x02 \x1b", 9));
-                                                        // \tcode{s4} has value: \tcode{[\textbackslash u\{0\} \textbackslash n \textbackslash t \textbackslash u\{2\} \textbackslash u\{1b\}]}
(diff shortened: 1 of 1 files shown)


mordante:
@brevzin can you have a look at these changes?

source/utilities.tex
```diff
@@ -15367,15 +15367,15 @@
...
+string s0 = format("[{}]", "h\tllo");                  // \tcode{s0} has value: \tcode{[h    llo]}
+string s1 = format("[{:?}]", "h\tllo");                // \tcode{s1} has value: \tcode{["h\textbackslash tllo"]}
+string s3 = format("[{:?}, {:?}]", '\'', '"');         // \tcode{s3} has value: \tcode{['\textbackslash '', '"']}
 
 // The following examples assume use of the UTF-8 encoding
 string s4 = format("[{:?}]", string("\0 \n \t \x02 \x1b", 9));
-                                                        // \tcode{s4} has value: \tcode{[\textbackslash u\{0\} \textbackslash n \textbackslash t \textbackslash u\{2\} \textbackslash u\{1b\}]}
-string s5 = format("[{:?}]", "\xc3\x28");               // invalid UTF-8
-                                                        // \tcode{s5} has value: \tcode{["\textbackslash x\{c3\}\textbackslash x\{28\}"]}
+                                                       // \tcode{s4} has value: \tcode{["\textbackslash u\{0\} \textbackslash n \textbackslash t \textbackslash u\{2\} \textbackslash u\{1b\}"]}
```
mordante, line 15376: Adding the quotes caused the build to fail due to
"Overfull \hbox (3.15547pt too wide) in paragraph at lines 15376--15377"
removing one space fixes the issue, an alternative would be to split the result of `s4` in two lines.

JohelEGP: > removing one space fixes the issue

The checks suggest otherwise.

Try the latter.

JohelEGP: > > removing one space fixes the issue
> 
> The checks suggest otherwise.

It fixes that issue, but brings up another. That's why splitting the comment should actually work.

mordante: Removing the space gives another error, it seems the comments must be aligned on a multiple of 4.
That is detected by a different script.
But I'll take your advice and split the comment over two lines.

tkoeppe: That's right. Can you remove four spaces from all the comments maybe? If not we may need to present this a bit differently.

mordante: 4 spaces would work too. I just changed it to two lines and pushed that before I read your comment.
@tkoeppe what do you prefer the current two lines or 4 reducing the indention by 4?

tkoeppe: If the entire example can use the same column for comments (even if that's four less than before), that's perfectly fine.

brevzin:
> @brevzin can you have a look at these changes?

Looks good to me, thanks for the fixes. What happened to `s2`?

JohelEGP:
> What happened to `s2`?

See https://github.com/cplusplus/draft/pull/5700#discussion_r936371491 and https://github.com/cplusplus/draft/pull/5702.

brevzin:
> > What happened to `s2`?
> 
> See [#5700 (comment)](https://github.com/cplusplus/draft/pull/5700#discussion_r936371491) and #5702.

I find this hilarious and more than a little ironic.

PR #5757, opened 2022-08-20
[diff] Uniform style for examples

tkoeppe:
For C++ differences, examples are always preceded by the phrase "For
example:", do not use the usual [Example n: ... -- end example] style,
and always appear in the "Effects on original feature" paragraph.

For differences with C, a different set of styles is used (examples
being part of paragraphs such as "Change" and "Rationale"), and that
subclause remains unchanged by this commit.

Suggested by CD review.

I will update the wiki accordingly.

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -34,11 +34,10 @@ Removal of unimplemented conditionally-supported feature.
 Concatenation of \grammarterm{string-literal}s
 with different \grammarterm{encoding-prefix}es
 is now ill-formed.
-\begin{example}
+For example:
 \begin{codeblock}
 auto c = L"a" U"b";             // was conditionally-supported; now ill-formed
 \end{codeblock}
-\end{example}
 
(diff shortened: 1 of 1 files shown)


source/compatibility.tex
```diff
@@ -1607,13 +1623,15 @@
 \rationale
 Catches bugs.
 \effect
-Valid \CppIII{} code may fail to compile in this revision of \Cpp{}. For
-example, the following code is valid in \CppIII{} but invalid in this
-revision of \Cpp{} because \tcode{double} to \tcode{int} is a narrowing
-conversion:
+Valid \CppIII{} code may fail to compile in this revision of \Cpp{}.
+For
+example:
```
JohelEGP, line 1628:
```suggestion
For example:
```

JohelEGP (changes requested):

tkoeppe:
Wiki updated: https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#differences-with-previous-versions-of-c

PR #5758, opened 2022-08-20
[memory.syn] Add missing "// freestanding" comment to "destroy"

### LINKED ISSUE #5731: std::destroy should be marked freestanding
P1642R11 says that the non-ExecutionPolicy overload of std::destroy in the memory header should be marked as freestanding.  std::destroy_at, std::destroy_n, and the various std::ranges::destroy* functions were all marked correctly, but this one got missed.

tkoeppe:
The comment was supposed to be added by P1642R11, but was accidentally
omitted.

Fixes #5731.

### DIFF
--- a/source/memory.tex
+++ b/source/memory.tex
@@ -363,7 +363,8 @@ namespace std {
   template<class T>
     constexpr void destroy_at(T* location);                                         // freestanding
   template<class NoThrowForwardIterator>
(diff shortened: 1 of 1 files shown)


tkoeppe:
@ben-craig voila

jwakely (approved):
This matches what the incoming paper said to do.

PR #5759, opened 2022-08-20
[vector.syn, vector.bool] Add subclause structure.

### LINKED ISSUE #5719: Codeblocks from P2286R8 need a home
P2286R8 Formatting Ranges #5700 introduced several codeblocks that appear in the middle of normative wording and aren't in a synopsis or itemdecl. The codeblocks contain definitions of `formatter` and `range_formatter`, and are listed below with their location within the spec followed by a line identifying what they define (note that these may have changed as the result of P2585R1 #5706 and/or other edits):
* [vector.bool]/p8 `struct formatter<T, charT> {`
* [assoc.format]/p1 `struct formatter<@\placeholder{map-type}@<Key, T, U...>, charT> {`
* [assoc.format]/p5 `struct formatter<@\placeholder{set-type}@<Key, U...>, charT> {`
* [container.adaptors.format]/p1 `struct formatter<@\placeholder{adaptor-type}@<T, Container, U...>, charT> {`
* [format.range.formatter]/p6 `class range_formatter {`
* [format.range.formatter]/p12 `struct formatter<R, charT> {`
* [format.tuple]/p1 `struct formatter<@\placeholder{pair-or-tuple}@<Ts...>, charT> {`

A proposal for how to resolve the 1st case follows:
Split the [vector.bool] subclause into [vector.bool.general] containing paras 1-3, and [vector.bool.ref] containing paras 4-6 and the _is-vector-bool-reference_ variable template. We need to find somewhere for `hash<vector<bool, Allocator>>`, maybe after para 3 in [vector.bool.general]. Then the `formatter` specialization and its members can be in [vector.bool.formatter].
_Originally posted by @jwakely in https://github.com/cplusplus/draft/pull/5700#discussion_r937192058_

tkoeppe:
After the addition of the formatting-related specialization, the
original subclause contained several class template synopses without
any intervening separation. The header synopsis is rearranged to
follow the document order.

Fixes #5719.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -6095,23 +6095,26 @@ namespace std {
     constexpr typename vector<T, Allocator>::size_type
       erase_if(vector<T, Allocator>& c, Predicate pred);
 
-  // \ref{vector.bool}, class \tcode{vector<bool>}
-  template<class Allocator> class vector<bool, Allocator>;
+  namespace pmr {
+    template<class T>
+      using vector = std::vector<T, polymorphic_allocator<T>>;
+  }
+
+  // \ref{vector.bool}, specialization of \tcode{vector} for \tcode{bool}
+  // \ref{vector.bool.pspc}, partial class template specialization \tcode{vector<bool, Allocator>}
+  template<class Allocator>
+    class vector<bool, Allocator>;
 
   template<class T>
     constexpr bool @\exposid{is-vector-bool-reference}@ = @\seebelow@;          // \expos
-
-  template<class T, class charT> requires @\exposid{is-vector-bool-reference}@<T>
-    struct formatter<T, charT>;
 
   // hash support
   template<class T> struct hash;
   template<class Allocator> struct hash<vector<bool, Allocator>>;
 
-  namespace pmr {
-    template<class T>
-      using vector = std::vector<T, polymorphic_allocator<T>>;
-  }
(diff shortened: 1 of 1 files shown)


source/containers.tex
```diff
@@ -9112,14 +9115,14 @@
 \end{codeblock}
 \end{itemdescr}
 
-\rSec2[vector.bool]{Class \tcode{vector<bool>}}
+\rSec2[vector.bool]{Specialization for \tcode{bool}}
```
JohelEGP, line 9118: I think `vector` should be mentioned in the subclause title. [vector.bool] is a sibling of [vector], so this is unlike this [specification guideline](https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#subclause-headings):
>Do not repeat parts of subclause headings in nested headings (decided in #1242). Good: "Class template tuple" / "Construction". Bad: "Class template vector" / "vector capacity".

This is how it currently looks:



Bettter precedent:




```suggestion
\rSec2[vector.bool]{Specialization for \tcode{vector} of \tcode{bool}}
```

JohelEGP (changes requested):

source/containers.tex
tkoeppe: Yes, I already had that but forgot to push it, thanks!

jwakely (approved):

PR #5760, opened 2022-08-20
[range.as.const.view] Fix constrain order

hewillk:
Make it consistent with the synopsis.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -8657,8 +8657,8 @@ cant_touch_this(views::as_const(beat));         // will not modify the elements 
 
 \begin{codeblock}
 namespace std::ranges {
-  template<@\libconcept{input_range}@ V>
-    requires @\libconcept{view}@<V>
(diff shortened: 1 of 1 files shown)


tkoeppe:
I realize we made this change for as_rvalue_view (2f228c5cad223a5c8d686d91b054ee3bd2d2a123), but on a closer look, I don't think there's any kind of preference in the existing text for either priority. Some view templates start with `view`, others with `input_range`, and this isn't the only outlier. (E.g. look at `join_view`.)

I'd welcome a wider discussion and analysis first before jumping into piecemeal changes. Feel free to reopen with that.

hewillk:
Yep, but this is a _fix_, see #5519.

tkoeppe:
Oh OK, it was already inconsistent with its own synposis. Sounds good, thanks!

tkoeppe:
@brevzin ^^^ FYI

Issue #5735, opened 2022-08-17
[range.cartesian.view] Fix _`cartesian-product-is-common`_


JohelEGP:
What is `Const`? Why is `Vs...` unused? The refined concept only takes 1 argument.

Presumably, `Const` came from a copy-pasted _`maybe-const`_`<Const, First>`. That'd explain the extra closing `>`.

Considering the name of the concept and how it's used, this should be:
```suggestion
    (@\exposconcept{cartesian-product-common-arg}@<First> && ... &&
     @\exposconcept{cartesian-product-common-arg}@<Vs>);
```

_Originally posted by @JohelEGP in https://github.com/cplusplus/draft/pull/5647#discussion_r929081572_

tkoeppe:
I think you're right, will send a PR.

hewillk:
Is this really correct? From the [original paper](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2374r4.html#common-range):

> `cartesian_product_view` can be a common range if the first range either is common, or is sized and random access. This paper reflects this.

In my opinion this should be
```cpp
template <class First, class... Vs>
  concept cartesian-product-is-common = // exposition only
    cartesian-product-common-arg<First>;
```

@tkoeppe @JohelEGP

JohelEGP:
Then I'd also question why the `end()` overloads are constrained with `@\exposconcept{cartesian-product-is-common}@<First, Vs...>` rather than `cartesian-product-common-arg<First>` directly.

hewillk:
`cartesian-product-common-arg` might not be a good constraint name for `end`, `cartesian-product-is-common` seems to be additionally defined for consistency with `cartesian-product-is-sized`.

PR #5768, opened 2022-08-22
[range.cartesian.view] Fix cartesian-product-is-common

hewillk:
I think this should be correct.
(In fact, I implemented `cartesian_product_view` after the pre-version p2374r3 was published. There are quite a lot of editorial mistakes in that paper. From my experience, the constraints here should only be aimed at `First`)

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -14286,9 +14286,9 @@ namespace std::ranges {
       (@\libconcept{bidirectional_range}@<@\exposid{maybe-const}@<Const, Vs>>
         && @\exposconcept{cartesian-product-common-arg}@<@\exposid{maybe-const}@<Const, Vs>>));
 
-  template<class... Vs>
+  template<class First>
   concept @\defexposconcept{cartesian-product-is-common}@ =                 // \expos
-    (@\exposconcept{cartesian-product-common-arg}@<Vs> && ...);
+    @\exposconcept{cartesian-product-common-arg}@<First>;
 
   template<class... Vs>
   concept @\defexposconcept{cartesian-product-is-sized}@ =                  // \expos
@@ -14329,9 +14329,9 @@ namespace std::ranges {
 
     constexpr @\exposid{iterator}@<false> end()
       requires ((!@\exposconcept{simple-view}@<First> || ... || !@\exposconcept{simple-view}@<Vs>) &&
-        @\exposconcept{cartesian-product-is-common}@<First, Vs...>);
+        @\exposconcept{cartesian-product-is-common}@<First>);
     constexpr @\exposid{iterator}@<true> end() const
-      requires @\exposconcept{cartesian-product-is-common}@<const First, const Vs...>;
+      requires @\exposconcept{cartesian-product-is-common}@<const First>;
     constexpr default_sentinel_t end() const noexcept;
 
     constexpr @\seebelow@ size()
@@ -14383,9 +14383,9 @@ Equivalent to:
 \begin{itemdecl}
 constexpr @\exposid{iterator}@<false> end()
   requires ((!@\exposconcept{simple-view}@<First> || ... || !@\exposconcept{simple-view}@<Vs>)
-    && @\exposconcept{cartesian-product-is-common}@<First, Vs...>);
+    && @\exposconcept{cartesian-product-is-common}@<First>);
 constexpr @\exposid{iterator}@<true> end() const
-  requires @\exposconcept{cartesian-product-is-common}@<const First, const Vs...>;
+  requires @\exposconcept{cartesian-product-is-common}@<const First>;
 \end{itemdecl}
 
 \begin{itemdescr}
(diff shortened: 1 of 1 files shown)


tkoeppe:
This feels like it needs an LWG issue. Could you please file one?

hewillk:
I don't think it's worth LWG. 
The main consideration here is that the definition of `cartesian-product-is-common` in the original paper [P2374R4](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2374r4.html) is nonsense as follows

```cpp
template <class R>
  concept cartesian-product-common-arg = // exposition only
    common_range<R> || (sized_range<R> && random_access_range<R>);

template <class First, class... Vs>
  concept cartesian-product-is-common = // exposition only
    cartesian-product-common-arg<Const, First>>;
```

This is mentioned in #5750, and the reporter believes that the intent of the paper should have been to define it as

```cpp
template<class First, class... Vs>
  concept cartesian-product-is-common =                 // exposition only
    (cartesian-product-common-arg<First> && ... &&
     cartesian-product-common-arg<Vs>);
```

which has been merged into the draft.  

However, according to the description of `common_range` in [Section 3.5](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2374r4.html#common-range) of the paper

> `cartesian_product_view` can be a common range if the first range either is common, or is sized and random access. This paper reflects this.

And according to the implementation section of the paper for `cartesian_product_view::end` in [[range.cartesian.view]](https://eel.is/c++draft/ranges#range.cartesian.view-4)

> ```cpp
> constexpr iterator<false> end()
>   requires ((!simple-view<First> || ... || !simple-view<Vs>)
>     && cartesian-product-is-common<First, Vs...>);
> constexpr iterator<true> end() const
>   requires cartesian-product-is-common<const First, const Vs...>;
> ```
> Let:
> (4.1) `is-const` be `true` for the const-qualified overload, and false otherwise;
> (4.2) `is-empty` be `true` if the expression `ranges​::​empty(rng)` is `true` for any `rng` among the underlying ranges except the first one and `false` otherwise; and
> (4.3) `begin-or-first-end(rng)` be expression-equivalent to `is-empty ? ranges​::​begin(rng) : cartesian-common-arg-end(rng)` if `rng` is the first underlying range and `ranges​::​begin(rng)` otherwise.
> Effects: Equivalent to:
> ```
> iterator<is-const> it(tuple-transform(
>   [](auto& rng){ return begin-or-first-end(rng); }, bases_));
> return it;
> ```

From the above context derivation, the merged fix is obviously not the intention of the paper. I think the fix that is closest to the intention of the paper should be

```
template <class First, class... Vs>
  concept cartesian-product-is-common = // exposition only
    cartesian-product-common-arg<First>;
```

And this pull request just removes the unused `Vs...` part of the template parameter list.

Can we ask the author for their opinion?

jwakely:
> Can we ask the author for their opinion?

Paging @griwes and @TartanLlama

My feeling is that this needs an LWG issue.

griwes:
No, the words in the paper are not a mistake. We only use a sentinel for the first range in the end iterator; for the rest of the ranges, we use an iterator, because in the end iterator we are storing begin(), not end(). So only the first range needs to be common for cartesian product to be common.

griwes:
Oh, though it seems the paper was missing maybe-const in this concept?

griwes:
As for `Vs...`, I believe it's there to be consistent with the other helper concepts of cartesian product, because everywhere else we're passing around all the ranges.

tkoeppe:
@griwes Right, the initial issue #5735 by @JohelEGP was complaining both about the `Const` and the unused `Vs...`, my  8275c19b3fa9e7cb09f84049ebd6207aceb7ca80 was an attempted "fix". But it sounds like that was wrong?

tkoeppe:
@griwes Would you perhaps be able to send a PR with the correct and intended wording?

tkoeppe:
OK, I can make that change. So we go back to the original, but drop the `Const`?

griwes:
Yeah, that sounds correct to me.

jwakely:
@griwes, thanks for taking the time out from your vacation to advise.

hewillk:
> As for Vs..., I believe it's there to be consistent with the other helper concepts of cartesian product, because everywhere else we're passing around all the ranges.

This is as I guessed.

> I'd also be happy to have an LWG discussion as to whether we should drop the parameter pack from is-common or not.

```cpp
template <class First, class... Vs>
  concept cartesian-product-is-common = // exposition only
    cartesian-product-common-arg<First>;
```
In fact, at first glance, I thought it was an editing error because it looks like the fold expression was left out, which is why it was removed from the pull request since I believe this will reduce confusion for people who haven't read the original paper.
If we consider keeps them, I think a more appropriate option would be to not name the rest packs like

```cpp
template <class First, class...>
  concept cartesian-product-is-common = // exposition only
    cartesian-product-common-arg<First>;
```
This more clearly expresses the intent of ignoring non-first views.

PR #5770, opened 2022-08-22
[range.cartesian.view] Fix neglect for parent_

hewillk:
This seems editorial bug in the paper.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -14366,7 +14366,7 @@ constexpr @\exposid{iterator}@<false> begin()
 \pnum
 \effects
 Equivalent to:
-\tcode{return \exposid{iterator}<false>(\exposid{tuple-transform}(ranges::begin, \exposid{bases_}));}
+\tcode{return \exposid{iterator}<false>(*this, \exposid{tuple-transform}(ranges::begin, \exposid{bases_}));}
 \end{itemdescr}
 
 \begin{itemdecl}
@@ -14378,7 +14378,7 @@ constexpr @\exposid{iterator}@<true> begin() const
 \pnum
 \effects
 Equivalent to:
-\tcode{return \exposid{iterator}<true>(\exposid{tuple-transform}(ranges::begin, \exposid{bases_}));}
+\tcode{return \exposid{iterator}<true>(*this, \exposid{tuple-transform}(ranges::begin, \exposid{bases_}));}
 \end{itemdescr}
 
 \begin{itemdecl}
@@ -14412,7 +14412,7 @@ if \tcode{rng} is the first underlying range and
 \effects
 Equivalent to:
 \begin{codeblock}
-@\exposid{iterator}@<is-const> it(@\exposid{tuple-transform}@(
+@\exposid{iterator}@<is-const> it(*this, @\exposid{tuple-transform}@(
   [](auto& rng){ return @\exposid{begin-or-first-end}@(rng); }, @\exposid{bases_}@));
 return it;
 \end{codeblock}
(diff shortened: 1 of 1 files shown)


tkoeppe:
@jwakely Does this require an LWG issue?

hewillk:
If this is the LWG, there is *no* way to implement a functional `cartesian_view` according to the draft until the LWG makes any decisions.

tkoeppe:
> If this is the LWG, there is _no_ way to implement a functional `cartesian_view` according to the draft until the LWG makes any decisions.

That may well be so, but is not an editorial concern.

`vector` also wasn't implementable until C++23, but that also didn't get fixed editorially.

source/ranges.tex
```diff
@@ -14654,14 +14660,15 @@
...
+constexpr explicit @\exposid{iterator}@(@\exposidnc{Parent}@& parent,
+  tuple<iterator_t<@\exposid{maybe-const}@<Const, First>>,
+        iterator_t<@\exposid{maybe-const}@<Const, Vs>>...> current);
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
 \effects
-Initializes \exposid{current_} with \tcode{std::move(current)}.
+Initializes \exposid{parent_} with \tcode{addressof(parent)} and \exposid{current_} with \tcode{std::move(current)}.
```
jwakely, line 14671:
```suggestion
Initializes \exposid{parent_} with \tcode{addressof(parent)}
and \exposid{current_} with \tcode{std::move(current)}.
```

source/ranges.tex
```diff
@@ -14673,7 +14680,7 @@
 \begin{itemdescr}
 \pnum
 \effects
-Initializes \exposid{current_} with \tcode{std::move(i.\exposid{current_})}.
+Initializes \exposid{parent_} with \tcode{i.\exposid{parent_}} and \exposid{current_} with \tcode{std::move(i.\exposid{current_})}.
```
jwakely, line 14683:
```suggestion
Initializes \exposid{parent_} with \tcode{i.\exposid{parent_}}
and \exposid{current_} with \tcode{std::move(i.\exposid{current_})}.
```

jwakely (changes requested):
Use semantic line breaks:
https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#latex-source-code-formatting

jwakely:
> If this is the LWG, there is _no_ way to implement a functional `cartesian_view` according to the draft until the LWG makes any decisions.

That's fine. This is not a final standard, there is no great urgency to fix defects in a draft.

Implementors can easily do this without waiting for an LWG decision. It's common that we implement fixes before they make it into the draft.

Although the changes here seem correct, I don't think it's editorial. Please submit an LWG issue instead, thanks!

tkoeppe:
Could you kindly edit the pull request to avoid sounding disparaging of the existing work and the people who have contributed it?

hewillk:
@tkoeppe  Sorry for the misunderstanding caused by the misuse of words, I will be more careful with words in the future pull.

@jwakely, thanks for the suggestion.
The reason I insist this is an editorial fix is that it was clearly an oversight, and I believe the fix was what the author originally intended. Maybe we can ask the author again about this opinion after he is on vacation.

If this is definitely an LWG, I'd like to ask _when_ it would be appropriate to submit it to the LWG, as the C++23 draft hasn't been published yet.

jensmaurer:
Specifically addressing your question, you should submit an issue to lwgchair now, so that the matter does not get lost.

hewillk:
Closed and intends to submit LWG for this.

PR #5772, opened 2022-08-22
[flat.*.{syn,erasure}] Change return type of erase_if to member size_type

tkoeppe:
All other containers and container adapters use the member size_type,
and the flat maps already did so in the item specification, but not in
the synopsis.

The current wording follows the approved proposals, but this change
seems like an improvement.

Suggested by CD review.


jwakely (approved):
The synopses obviously need to be consistent, and using `size_type` is consistent with all other containers and with the detailed descriptions for flat maps. I don't think this needs to go via LWG, I can't imagine anybody asking for any other resolution than this.

tkoeppe:
Thanks a lot!

PR #5775, opened 2022-08-22
Fix various library typos and style issues from #5428

jwakely:
Partially addresses #5428


source/numerics.tex
```diff
@@ -7929,7 +7929,6 @@
 \begin{itemdecl}
 slice();
 slice(size_t start, size_t length, size_t stride);
-slice(const slice&);
```
tkoeppe, line 7932: Weird that those were ever there! Why have no valarray users filed bug reports...

jensmaurer: > Why have no valarray users filed bug reports...

Maybe there are no valarray users?

PR #5778, opened 2022-08-23
[range.zip, ranges.cartesian.iterator] Simplify `maybe-const<Const, Views>` into `const Views`

hewillk:
Elsewhere it is simple enough since there is no fold expression and `Base` is defined, and I think this improves readability.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -9600,8 +9600,7 @@ Initializes \exposid{current_} with \tcode{std::move(current)}.
 
 \begin{itemdecl}
 constexpr @\exposid{iterator}@(@\exposid{iterator}@<!Const> i)
-  requires Const &&
-           (@\libconcept{convertible_to}@<iterator_t<Views>, iterator_t<const Views>> && ...);
+  requires Const && (@\libconcept{convertible_to}@<iterator_t<Views>, iterator_t<const Views>> && ...);
 \end{itemdecl}
 
 \begin{itemdescr}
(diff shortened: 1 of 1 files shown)


tkoeppe:
Is this editorial?

hewillk:
Not sure, if this was LWG then I would close it.

tkoeppe:
Well, to be editorial the change cannot change the normative meaning of the text, first and foremost. I don't know whether the proposed new text is equivalent to the existing one, or if you're proposing a (simplifying) change of the observable specification. Can you elaborate?

cpplearner:
IIUC the idea is: these `maybe-const<Const, Views>` appear in the right operand of `requires Const && stuff`, so they are only used when `Const` is `true`, so they are equivalent to `maybe-const<true, Views>`, which is the same as `const Views`.

jwakely:
I believe this has no normative impact. The requires clause will short circuit so that if `Const` is false, we never try to check the `convertible_to` constraint, and so it doesn't matter whether we use `iterator_t<X>` or `iterator<const X>` as it will not even be instantiated. So the constraint using `iterator_t<const X>` is only checked when `Const` is true, in which case _maybe-const_ would have added the `const` anyway.

So I think this is a valid editorial simplification.

tkoeppe:
OK, thanks!

In that case, can you please review if any of the linebreaks can now go away?

source/ranges.tex
```diff
@@ -9520,7 +9520,7 @@
     @\exposid{iterator}@() = default;
     constexpr @\exposid{iterator}@(@\exposid{iterator}@<!Const> i)
       requires Const && (@\libconcept{convertible_to}@<iterator_t<Views>,
-                                        iterator_t<@\exposid{maybe-const}@<Const, Views>>> && ...);
+                                        iterator_t<const Views>> && ...);
```
jensmaurer, line 9523: Can we reformat so that "convertible_to" is on its own (single) line, like the other cases in this pull request?

jensmaurer (changes requested):

source/ranges.tex
```diff
@@ -9602,7 +9601,7 @@
 \begin{itemdecl}
 constexpr @\exposid{iterator}@(@\exposid{iterator}@<!Const> i)
   requires Const &&
-           (@\libconcept{convertible_to}@<iterator_t<Views>, iterator_t<@\exposid{maybe-const}@<Const, Views>>> && ...);
+           (@\libconcept{convertible_to}@<iterator_t<Views>, iterator_t<const Views>> && ...);
```
tkoeppe, line 9604: Does this fit on a single line?

jwakely (approved):

PR #5779, opened 2022-08-23
Fix adjacent_view example.

cor3ntin:
adjacent_view always yields tuples, but the example used std::pair.


jwakely (approved):
With the correct comment indentation this *just* fits on a single line.

tkoeppe:
@jwakely And it is correct as is, right? Comment at column 68?

tkoeppe:
@cor3ntin Please prefix commit messages with the affected subclause. (I'll update it this time.)

jwakely:
> @jwakely And it is correct as is, right? Comment at column 68?

After the force-push, yes :-)

PR #5783, opened 2022-08-23
[stmt.return], [stmt.return.coroutine] update notes re implicit move

Quuxplusone:
These notes probably should have been changed when P1825 was adopted.

### DIFF
--- a/source/statements.tex
+++ b/source/statements.tex
@@ -900,10 +900,10 @@ A constructor or destructor does not have a return type.
 \end{note}
 \begin{note}
 A \tcode{return} statement can involve
-an invocation of a constructor to perform a copy or move of the operand
+a copy or move of the operand
 if it is not a prvalue or if its type differs from the return type of the function.
-A copy operation associated with a \tcode{return} statement can be elided or
-converted to a move operation if an automatic storage duration variable is returned\iref{class.copy.elision}.
+A copy operation associated with a \tcode{return} statement can be elided\iref{class.copy.elision}
+or converted to a move operation if an implicitly movable entity\iref{expr.prim.id.unqual} is returned.
 \end{note}
 
 \pnum
@@ -993,6 +993,10 @@ and \placeholder{S} is defined as follows:
 \item
 If the operand is a \grammarterm{braced-init-list} or an expression of non-\keyword{void} type,
 \placeholder{S} is \placeholder{p}\tcode{.return_value(}\grammarterm{expr-or-braced-init-list}{}\tcode{)}.
+\begin{note}
+The overload resolution of \placeholder{S} is affected
+by whether the operand is an implicitly movable entity\iref{expr.prim.id.unqual}.
+\end{note}
 The expression \placeholder{S} shall be a prvalue of type \keyword{void}.
 
 \item


tkoeppe:
@jensmaurer: I'd like a core opinion for this, would you have one or could recommend someone?

source/statements.tex
```diff
@@ -921,6 +921,10 @@
 The expression \placeholder{p}\tcode{.return_void()}
 shall be a prvalue of type \keyword{void}.
 \end{itemize}
+\begin{note}
+The overload resolution associated with a \tcode{co_return} statement is affected
```
jensmaurer, line 925: Which overload resolution are we talking about? The one for `p.return_void()`?

source/statements.tex
```diff
@@ -921,6 +921,10 @@
...
+by whether the operand is an implicitly movable entity\iref{class.copy.elision}.
```
jensmaurer, line 926: This is no longer in class.copy.elision, I think.

source/statements.tex
```diff
@@ -843,10 +843,10 @@
 \end{note}
 \begin{note}
 A \tcode{return} statement can involve
-an invocation of a constructor to perform a copy or move of the operand
+a copy or move of the operand
 if it is not a prvalue or if its type differs from the return type of the function.
-A copy operation associated with a \tcode{return} statement can be elided or
-converted to a move operation if an automatic storage duration variable is returned\iref{class.copy.elision}.
+The copy operation associated with a \tcode{return} statement can be elided or
+converted to a move operation if an implicitly movable entity\iref{class.copy.elision} is returned.
```
jensmaurer, line 849: (yes, pre-existing, but while we're touching this)

It seems unclear whether the "if" applies to just the "converted to a move operation" part or to the entire "can be elised or converted to a move operation".

source/statements.tex
```diff
@@ -843,10 +843,10 @@
 \end{note}
 \begin{note}
 A \tcode{return} statement can involve
-an invocation of a constructor to perform a copy or move of the operand
+a copy or move of the operand
 if it is not a prvalue or if its type differs from the return type of the function.
-A copy operation associated with a \tcode{return} statement can be elided or
-converted to a move operation if an automatic storage duration variable is returned\iref{class.copy.elision}.
+The copy operation associated with a \tcode{return} statement can be elided or
```
jensmaurer, line 848: Why "the" copy operation? There might be none.

source/statements.tex
jensmaurer, line 849: The implicitly movable entity is no longer in class.copy.elision. The copy elision still is.

jensmaurer (changes requested):
I think there's some overall improvement lurking in the shadows for this pull request, but we're not there yet.

source/statements.tex
Quuxplusone: My thought process was: There _might_ be none, but if there are any, then there is _one_. "A copy operation" reads to me as if there might be several, and any-or-all of them might be converted. But what we mean is that there is at most one copy, and, if _the_ copy exists, then _the_ copy can be elided.

Still, I can see the logic either way, and `s/A/The/` isn't important to this patch, so I've reverted to `A`.

source/statements.tex
Quuxplusone: Oops, good catch: This note was meant to apply to `co_return` as a whole (and "the overload resolution" would be the one associated with `p.return_value(operand)`, because that's the only place "the operand" makes sense), but then it probably makes sense to move it up into the bulleted list, right after we talk about `p.return_value(operand)` and before we introduce `p.return_void()` at all. Done.

source/statements.tex
```diff
@@ -913,6 +913,10 @@
 \item
 If the operand is a \grammarterm{braced-init-list} or an expression of non-\keyword{void} type,
 \placeholder{S} is \placeholder{p}\tcode{.return_value(}\grammarterm{expr-or-braced-init-list}{}\tcode{)}.
+\begin{note}
+The overload resolution of \placeholder{S} is affected
```
Quuxplusone, line 917: s/of/inside/?

source/statements.tex
tkoeppe: @Quuxplusone PTAL?

Quuxplusone: Jens was right; but I believe I've fixed this in the update 2 months ago.

PR #5785, opened 2022-08-23
[range.drop.overview, range.take.overview] Fixed unformatted `(void)F, decay-copy(E)`

hewillk:
1. Remove the space between `(void)` and `F`, which is consistent with the zip family overview.
2. Formatted unformatted `((void)F, decay-copy(E))` in [[range.drop.overview-2.4.2]](https://eel.is/c++draft/ranges#range.drop.overview-2.4.2).
3. Extended question: Is it correct to use `auto(E)` in these contexts?


JohelEGP:
> Extended question: Is it correct to use `auto(E)` in these contexts?

I don't think so. Where it matters, it should already have been done.

PR #5787, opened 2022-08-23
[check] Flag any space between 'template' and '<'

jensmaurer:
Previously, the pattern required 'class' after '<' to fire.

This prevents the need for future piecemeal fixes such as 356fb7ff88b63d956b1109c72c5e3bf424f6ba29 or eca39f43798d7a58fdd482232c60b6db428b656f.

Tested with the state before 356fb7ff88b63d956b1109c72c5e3bf424f6ba29, which was successfully flagged.

### DIFF
--- a/tools/check-source.sh
+++ b/tools/check-source.sh
@@ -108,9 +108,9 @@ grep -Hne '^\\\(change\|rationale\|effect\|difficulty\|howwide\)\s.\+$' compatib
     fail "change marker in [diff] followed by stuff" || failed=1
 # Fixup: sed 's/^\\\(change\|rationale\|effect\|difficulty\|howwide\)\s\(.\)/\\\1\n\2/'q
 
-# "template <class" (with space) in library clause.
-grep -ne 'template\s<class' $texlib |
-    fail 'space between "template" and "<class"' || failed=1
+# "template <" (with space) in library clause.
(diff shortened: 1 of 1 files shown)


tools/check-source.sh
```diff
@@ -108,9 +108,9 @@ grep -Hne '^\\\(change\|rationale\|effect\|difficulty\|howwide\)\s.\+$' compatib
     fail "change marker in [diff] followed by stuff" || failed=1
 # Fixup: sed 's/^\\\(change\|rationale\|effect\|difficulty\|howwide\)\s\(.\)/\\\1\n\2/'q
 
-# "template <class" (with space) in library clause.
-grep -ne 'template\s<class' $texlib |
-    fail 'space between "template" and "<class"' || failed=1
+# "template <" (with space) in library clause.
+grep -ne 'template\s<' $texlib |
```
CaseyCarter, line 112: We should probably catch any number of whitespace characters instead of exactly one:
```suggestion
grep -ne 'template\s+<' $texlib |
```

jwakely: That should be `\+` because this is a POSIX BRE not ERE. Use `grep -E` for EREs.

jensmaurer: Good point. Fixed.

CaseyCarter (approved):

PR #5792, opened 2022-08-25
[move.sentinel] Remove extraneous "`std::`" qualification

hewillk:


source/iterators.tex
```diff
@@ -1346,8 +1346,8 @@
 \libconcept{indirectly_writable} has the awkward \keyword{const_cast} expressions to reject
 iterators with prvalue non-proxy reference types that permit rvalue
 assignment but do not also permit \keyword{const} rvalue assignment.
-Consequently, an iterator type \tcode{I} that returns \tcode{std::string}
-by value does not model \tcode{\libconcept{indirectly_writable}<I, std::string>}.
+Consequently, an iterator type \tcode{I} that returns \tcode{string}
+by value does not model \tcode{\libconcept{indirectly_writable}<I, string>}.
```
jwakely, line 1350: The purpose of the note is to clarify the normative text. Arguably `string` in this context is less clear than `std::string`. A clarifying note should be as clear as possible, and we can be more lax about formatting/specification rules in notes. I don't feel strongly about it though.

The other part of this fix looks fine

hewillk: Reasonable, thanks for the suggestion.

jwakely (approved):
We're not really consistent about when we do or don't qualify in examples, but this seems fine.

tkoeppe:
@jwakely I agree, though given that we're using all the other std-names unqualified here, it's probably clearer to not have one case differ.

PR #5793, opened 2022-08-25
[containers] Add `std::` for `forward`/`move`

hewillk:


jensmaurer (approved):
squash on merge

tkoeppe:
I was going to go even futher and delete some newlines, but the GitHub based editor isn't working all that well...

PR #5795, opened 2022-08-27
Improve automatic LaTeX checking of concept names

jensmaurer:
complain about undefined concepts
and about concepts in codeblocks not decorated with \libconcept
(we had quite a few late-ish fixes for the latter in the CD cycle)


tkoeppe:
Nice, thanks!

PR #5803, opened 2022-08-29
[flat.set.modifiers] Remove stray 'return' in Effects clause

### LINKED ISSUE #5728: [flat.set.modifiers] Drop `return` statement from code block in *Effects* element
```C++
template<class K> pair<iterator, bool> insert(K&& x);
template<class K> iterator insert(const_iterator hint, K&& x);
```

Part of the effects of these overloads specify:
> inserts a new element as if by: `return emplace(std::forward<K>(x));`.

These are the problems with it, first described at https://github.com/cplusplus/draft/pull/5696#discussion_r945302589:
1. The _Returns_ element fully describes the result.
2. The use of `return` has nothing to do with how the new element is inserted.
3. And also, this `return` can't work with the first overload.

jensmaurer:
Fixes #5728

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -16683,7 +16683,7 @@ an object \tcode{u}, for which \tcode{find(x) == find(u)} is true.
 If the set already contains an element equivalent to \tcode{x},
 \tcode{*this} and \tcode{x} are unchanged.
 Otherwise,
-inserts a new element as if by: \tcode{return emplace(std::forward<K>(x));}.
+inserts a new element as if by \tcode{emplace(std::forward<K>(x))}.
 
 \pnum
 \returns


source/containers.tex
```diff
@@ -16683,7 +16683,7 @@
 If the set already contains an element equivalent to \tcode{x},
 \tcode{*this} and \tcode{x} are unchanged.
 Otherwise,
-inserts a new element as if by: \tcode{return emplace(std::forward<K>(x));}.
+inserts a new element as if by: \tcode{emplace(std::forward<K>(x));}.
```
CaseyCarter, line 16686: We conventionally do not include semicolons in expressions, or offset expressions from prose with `:`:
```suggestion
inserts a new element as if by \tcode{emplace(std::forward<K>(x))}.
```

CaseyCarter (approved):

PR #5807, opened 2022-09-03
[const.iterators.alias] Rename template parameter `It` to `I`

hewillk:
This makes them consistent with the template names in the [[iterator.synopsis]](https://eel.is/c++draft/iterator.synopsis) and consistent with the `<iterator>` naming style (which basically uses `Iterator` or `I` as the template name).

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -363,7 +363,7 @@ namespace std {
   };
 
   template<@\libconcept{input_iterator}@ I>
-    constexpr const_iterator<I> make_const_iterator(I it) { return it; }            // freestanding
+    constexpr const_iterator<I> make_const_iterator(I i) { return i; }            // freestanding
 
   template<class S>
     constexpr const_sentinel<S> make_const_sentinel(S s) { return s; }              // freestanding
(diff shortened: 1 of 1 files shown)


source/iterators.tex
```diff
@@ -363,7 +363,7 @@
   };
 
   template<@\libconcept{input_iterator}@ I>
-    constexpr const_iterator<I> make_const_iterator(I it) { return it; }            // freestanding
+    constexpr const_iterator<I> make_const_iterator(I i) { return i; }            // freestanding
```
JohelEGP:
```suggestion
    constexpr const_iterator<I> make_const_iterator(I i) { return i; }              // freestanding
```

tkoeppe:
@jwakely, @CaseyCarter: have you got a general policy that iterator type parameters should be called `I`? In principle this change looks good, but maybe we can record such conventions somewhere for posterity?

tkoeppe:
@hewillk Please rebase?

@jwakely, @CaseyCarter Ping :-)

CaseyCarter:
We're inconsistent across the entire library, I think we use `I`, `It`, `Iter` and `Iterator` in various places. In Ranges we do consistently use `I`: constraints clarify the meaning of names so we don't need descriptive placeholders. The traditional `class InputIterator` becomes `input_iterator I` in Ranges things.

jwakely:
> have you got a general policy that iterator type parameters should be called `I`?

No, but what Casey said.

CaseyCarter (approved):

PR #1902, opened 2018-01-26
Replace \^ with \caret

cpplearner:

Issue #205, opened 2013-09-30
Weird special characters


StephanTLavavej:
N3690 2.3 [lex.charset]/1 has a good caret but a bad tilde. The offending LaTeX is:
_ { } [ ] # ( ) < > % : ; . ? \* + - / ^ & | @$\sim$@ ! = , @\textbackslash@ " '
so all occurrences of @$\sim$@ should be inspected.

2.13 [lex.operators] has a bad caret but a good tilde. The offending LaTeX is:
>\+ >\- >\* >/ >\% >^{} >& >| >\tilde\br
PDF search indicates that there are 25 occurrences of this.

20.10.7 [bitwise.operations]/5 has a bad tilde. The offending LaTeX is:
\tcode{operator()} returns \tcode{~{}x}.
This appears elsewhere (e.g. /9).

jensmaurer:
This seems to be fixed.

StephanTLavavej:
I confirm that [lex.charset] and [bitwise.operations] are fixed.

However, [lex.operators] is unfixed. The `^` and `^=` have weird carets. By weird/bad, I meant "doesn't copy into a plain text editor normally" (at least on Windows). Please consider reactivating this.

tkoeppe:
I think you need to replace

    \^

with

    \char`\^

However, since this appears in a few more places than the one listing mentioned here, we should probably create a macro for this and apply that consistently.

jwakely:
FWIW the "bad" one is U+02C6 rather than U+005E

tkoeppe:
Let's reopen this for now so we don't forget. Patches welcome.

StephanTLavavej:
Thank you! As someone whose job is copy-pasting from the Standard, I really appreciate this fix.

tkoeppe:
No problem, thanks for calling out the problem! And thanks to Jens for implementing the solution!

PR #5816, opened 2022-09-07
[expr] Replace `\^` with `\caret`

cpplearner:
`\^` produces the circumflex accent `ˆ` instead of the ASCII character `^`.

The `\caret` command was added after #205 points out that `\^` produces a bad caret. PR #1902 fixed two uses of `\^` that were added after #205 got fixed. But another two uses of `\^` snuck in after that.

Given that `\^` keeps reoccuring, there's currently no use of the diacritic `ˆ` in this document, and there's precedent of redefining `\~` to mean something other than the diacritic, would it make sense to redefine `\^` to mean the ASCII letter, or else add a check to prohibit `\^`?

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -6720,7 +6720,7 @@ The result is the bitwise \logop{and} function of the operands.
 \end{bnf}
 
 \pnum
-The \tcode{\^} operator groups left-to-right.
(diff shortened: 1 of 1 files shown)


tkoeppe:
Thank you! I suppose we could add a check, but then again, this is probably so rare that it's not worth the added complexity.

PR #5818, opened 2022-09-08
[basic.life]/8.3 don't put const in the middle of "complete object" words of power

RedBeard0531:
"Complete object" is a defined term (https://eel.is/c++draft/basic#def:complete_object). Putting const in the middle of that makes it unclear if we are referring to a const-qualified "complete object" or a complete const object with some other meaning of "complete" such as "with a complete type". Using the order "const complete object" avoids that ambiguity. It also matches the word order used in https://eel.is/c++draft/basic.life#10 to make it UB to create a new object on top of a "const complete object", unless it has dynamic storage duration.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3515,7 +3515,7 @@ the storage that $o_1$ occupied, and
 \item $o_1$ and $o_2$ are of the same type
 (ignoring the top-level cv-qualifiers), and
 
-\item $o_1$ is not a complete const object, and
+\item $o_1$ is not a const complete object, and
 
 \item neither $o_1$ nor $o_2$
 is a potentially-overlapping subobject\iref{intro.object}, and
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3515,7 +3515,7 @@ the storage that $o_1$ occupied, and
 \item $o_1$ and $o_2$ are of the same type
 (ignoring the top-level cv-qualifiers), and
 
-\item $o_1$ is not a const complete object, and
+\item $o_1$ is not a const, complete object, and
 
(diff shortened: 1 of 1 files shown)


jwakely (commented):
Would "complete object of const type" be even better?

jensmaurer:
@RedBeard0531, turning your argument around, [basic.type.qualifier] p1.1 defines "const object". We shouldn't put "complete" in the middle of a defined term.

Having said, that "const complete object" sounds a little better than the existing phrasing.

jensmaurer (approved):

tkoeppe:
How about "const, complete object" or "complete, const object"? Constness and completeness are both separate qualifications of "object", right? Constness does not only appertain to complete objects, nor does completeness to const objects?

tkoeppe:
@jwakely, @RedBeard0531 Could you please take another look?

jwakely:
I like the suggestion to put a comma there. Of the two, I have a weak preference for "const, complete object".

RedBeard0531:
"const, complete object" sounds good to me too. Should we also update [basic.life]/10 to match?

> Constness does not only appertain to complete objects, nor does completeness to const objects?

The line that this PR is modifying does mean that in some sense constness only "really" means anything on a complete object, since const subobjects of non-const objects can still be overwritten. But that doesn't change the fact that non-complete objects can also be marked as const, with the effects on overload resolution and such.

jensmaurer (changes requested):
Could we please add the comma and then move on with this trivial pull request?

jensmaurer (approved):

jensmaurer:
This seems to do what we wanted.

PR #5819, opened 2022-09-08
[ranges] Use plain font for ranges/view when it is not specified as concept

hewillk:
This only added \libconcept for the `view` that appears in the adapters Overview section to make them consistent. 
(Adding \libconcept for _all_ `view` keywords in `<ranges>` may be debatable)


jwakely:
I know we're inconsistent today and we already do this in many places, but I'm a little uneasy about this change.

This is implicitly defining "a `view`" as "a type which models the `view` concept", and personally I think "a view" (without the code font) is fine for that informal term. Using code font for the concept itself *and* for types which model it seems less clear than using code font for the concept alone. You need to read the context more carefully to understand what is meant.

We don't say "a `copy_constructible`" or "a `semiregular`" when we mean types that model those concepts.

I would be a little less uneasy if we used "a `view` type", but I still think "a view" is acceptable for that meaning.

jwakely:
Also, this makes the entry for `view` in the index of library concepts even more silly. I don't think we need things like "the resulting view is empty" to be indexed as a use of the concept.

hewillk:
> Also, this makes the entry for `view` in the index of library concepts even more silly. I don't think we need things like "the resulting view is empty" to be indexed as a use of the concept.

This is a persuasive point.  So, for consistency, do we need to change all `view`s in the **Overview** section to text format "view", that is, take the first step to deal with inconsistent `view`/"view" keywords?

tkoeppe:
editorial consensus: do what Jens said ^^^.

hewillk:
Do I need to make changes directly in this pull or reopen a pull request?

tkoeppe:
If you could just repurpose this PR, that'd be fine!

tkoeppe:
@jwakely, could you please take a look?

source/ranges.tex
```diff
@@ -3986,7 +3986,7 @@
 
 \pnum
 Subclause \ref{range.adaptors} defines \term{range adaptors}, which are utilities that transform a
-\libconcept{range} into a \libconcept{view} with custom behaviors. These
+\libconcept{range} into a view with custom behaviors. These
```
jwakely, line 3989: Do we want `range` in code font here? Doesn't the same principle apply?

hewillk: Yes, I should do that. Thanks for the reminder.

jwakely (approved):

tkoeppe:
I've rebased this and will commit it, thank you!

PR #5826, opened 2022-09-14
Update references to Unicode Standard & Annexes.

cor3ntin:
Currently we have a normative reference to ISO 10646 (Unicode 13.0), a floating normative reference to UAX44 (= Unicode 15.0), bibligraphy items for Unicode (14.0),
text segmentation UAX (12.0), and identifier syntax (13.0).

This align the bibligraphy items and mentions of Unicode to refer to 15.0.

Mention of Unicode 14 in the specification of print are also updated to refer to Unicode 15.

No further changes are necessary to comply to the updated standard and annexes.

### DIFF
--- a/source/back.tex
+++ b/source/back.tex
@@ -30,7 +30,7 @@
   The Unicode Consortium. Unicode Standard Annex, UAX \#31,
   \doccite{Unicode Identifier and Pattern Syntax} [online].
   Edited by Mark Davis and Robin Leroy. Revision 37; issued for Unicode 15.0.0.
-  2022-08-31 [viewed 	2022-09-14].
+  2022-08-31 [viewed 2022-09-14].
   Available from: \url{https://www.unicode.org/reports/tr31/tr31-37.html}
 \item
   The Unicode Standard Version 15.0,
--- a/source/back.tex
+++ b/source/back.tex
@@ -23,19 +23,20 @@
   %%% http://www.iec.ch/standardsdev/resources/draftingpublications/directives/principles/referencing.htm
   The Unicode Consortium. Unicode Standard Annex, UAX \#29,
   \doccite{Unicode Text Segmentation} [online].
-  Edited by Mark Davis. Revision 35; issued for Unicode 12.0.0. 2019-02-15 [viewed 2020-02-23].
-  Available from: \url{http://www.unicode.org/reports/tr29/tr29-35.html}
+  Edited by Christopher Chapman. Revision 41; issued for Unicode 15.0.0.
+  2022-08-26 [viewed 2022-09-14].
+  Available from: \url{https://www.unicode.org/reports/tr29/tr29-41.html}
 \item
   The Unicode Consortium. Unicode Standard Annex, UAX \#31,
   \doccite{Unicode Identifier and Pattern Syntax} [online].
-  Edited by Mark Davis. Revision 33; issued for Unicode 13.0.0.
-  2020-02-13 [viewed 2021-06-08].
-  Available from: \url{https://www.unicode.org/reports/tr31/tr31-33.html}
+  Edited by Mark Davis and Robin Leroy. Revision 37; issued for Unicode 15.0.0.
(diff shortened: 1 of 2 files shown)


source/back.tex
```diff
@@ -23,19 +23,20 @@ \chapter{Bibliography}
...
+  2022-08-26 [viewed 2022-09-14].
+  Available from: \url{https://www.unicode.org/reports/tr29/tr29-41.html}
 \item
   The Unicode Consortium. Unicode Standard Annex, UAX \#31,
   \doccite{Unicode Identifier and Pattern Syntax} [online].
-  Edited by Mark Davis. Revision 33; issued for Unicode 13.0.0.
-  2020-02-13 [viewed 2021-06-08].
-  Available from: \url{https://www.unicode.org/reports/tr31/tr31-33.html}
+  Edited by Mark Davis and Robin Leroy. Revision 37; issued for Unicode 15.0.0.
+  2022-08-31 [viewed 	2022-09-14].
```
JohelEGP, line 33:
```suggestion
  2022-08-31 [viewed 2022-09-14].
```

JohelEGP (changes requested):

jensmaurer:
For the bibliography references, I'd like CWG and/or LWG review / approval that we're still good with the semantics described in the updated version. (Alternatively, some review in the editorial group that Unicode did at most editorial updates to the relevant sections.)

For the references in iostreams.tex, we should instead refer to ISO 10646 and avoid the numerical section references (because it may change when updates are published).

jwakely:
I agree with Jens, changing the references is not just editorial.

cor3ntin:
> For the bibliography references, I'd like CWG and/or LWG review / approval that we're still good with the semantics described in the updated version. (Alternatively, some review in the editorial group that Unicode did at most editorial updates to the relevant sections.)
 
> For the references in iostreams.tex, we should instead refer to ISO 10646 and avoid the numerical section references (because it may change when updates are published).

But this is not a behavior described by ISO 10646. References in Unicode do not happen to change very often. I'd have not objection to remove the numerical reference though

cor3ntin:
If someone wants to put that diff in html and then back again, please feel
free to do so.

Otherwise I'll make a NB comment so that we don't loose track of it.

Cheers.

tkoeppe:
Since changing normative references is not editorial, I don't see how it would be appropriate to make this change without WG21 oversight and record keeping. But it sounds like filing an LWG issue would be a bit simpler than filing an NB comment!

jensmaurer:
@tkoeppe , the bibliography changes are not per se normative (but may still present a larger lever than what we're happy to apply editorially).

JohelEGP:
> Normatively referenced documents shall be documents published by ISO or IEC. In the absence of appropriate ISO or IEC documents, those published by other bodies may be listed as normative references provided that [...]
> -- [ISO/IEC Directives, Part 2, section 10.2 "Permitted referenced documents"](https://www.iso.org/sites/directives/current/part2/index.xhtml#_idTextAnchor130)

It seems possible to keep only the undated reference in [intro.refs] if ISO/IEC 10646 is not appropriate due to
> [P]revent[ing] the use of features from newer Unicode Standard releases for up to several years.
> -- https://github.com/cplusplus/nbballot/issues/412#issuecomment-1304696400

tahonermann:
Thank you, @JohelEGP, that reference is helpful. I agree that we have the ability to reference both in a consistent fashion. That still leaves us with the downsides of 1) requiring us to word the dependency relationship between the two standards, 2) requiring implementors to consult both standards, and 3) resolving any discrepancies that arise between the two standards.

Those downsides aren't exorbitant, but they are complications that would be avoided by solely referring to the Unicode Standard.

JohelEGP:
If those downsides make the ISO Standard not appropriate, it seems to me that the first quote of https://github.com/cplusplus/draft/pull/5826#issuecomment-1304702217 can be applied to refer only to the Unicode Standard.

Issue #5724, opened 2022-08-12
[ranges.syn] Should the *namespace-alias* `std::views` be `// freestanding`?


JohelEGP:
Related comments: https://github.com/cplusplus/draft/pull/5713#discussion_r939550414 https://github.com/cplusplus/draft/pull/5713#discussion_r939686289

https://wg21.link/p1642r11 defines
> A _freestanding entity_ is a declaration or macro definition that is present in a freestanding implementation and a hosted implementation.
> [...]
>  In a header synopsis, entities followed with a comment that includes _freestanding_ are freestanding entities.

Notably, a freestanding entity is not an entity:
> [3](https://eel.is/c++draft/basic.pre#3) [#](http://github.com/Eelis/draft/tree/f8a8938fca3df9775504cdd4d224b1f3987d1e5b/source/basic.tex#L29) An [entity](https://eel.is/c++draft/basic.pre#def:entity) is a value, object, reference, structured binding, function, enumerator, type, class member, bit-field, template, template specialization, namespace, or pack[.](https://eel.is/c++draft/basic.pre#3.sentence-1)

The paper has the following instructions:
> Change in [[ranges.syn]](https://wg21.link/ranges.syn)
Drafting note:
[...]
Instructions to the editor:
Please append a `// freestanding` comment to every entity in `<ranges>` except for the following entities:
> ```
>     basic_istream_view
>     istream_view
>     wistream_view
>     views::istream
> ```

The paper instructs to append `// freestanding` to entities in `<ranges>`, which excludes the _namespace-alias_  `std::views` by definition.

I commented:
> I suppose this begs the question, do freestanding implementations have to use `std::ranges::views` rather than `std::views`?
> -- https://github.com/cplusplus/draft/pull/5713#discussion_r940298499

Currently, by my fault, https://github.com/cplusplus/draft/pull/5713 does append `// freestanding` to `std::views`, but that currently has no effects since `std::views` is not an entity.

CaseyCarter:
The root cause here seems to be that [freestanding.entity]/3 uses "entities":

> In a header synopsis, entities followed with a comment that includes _freestanding_ are freestanding entities.

where it should instead say "declarations and macro definitions":

> In a header synopsis, declarations and macro definitions followed with a comment that includes _freestanding_ are freestanding entities.

As is fairly obvious from the following example that applies such a comment to a macro definition, which is not an entity.

JohelEGP:
[...]

PR #5396, opened 2022-04-17
Fix placement of namespace-scope declarations

jensmaurer:
Fixes #5385

PR #5839, opened 2022-09-15
[ranges.syn] Make get overloads freestanding

### LINKED ISSUE #5837: [ranges.syn] std::[ranges::]get([const ]subrange<I, S, K>&) is not marked freestanding
P1642 does not omit `get()` nor `subrange` from freestanding, so, presumably, this was an error.

jwakely:
These were moved to the synopsis after the P1642 changes to mark nearly everything in the synopsis freestanding, so they were not marked.

Fixes #5837

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -140,15 +140,15 @@ namespace std::ranges {
 
   template<size_t N, class I, class S, subrange_kind K>
     requires ((N == 0 && @\libconcept{copyable}@<I>) || N == 1)
-    constexpr auto get(const subrange<I, S, K>& r);
+    constexpr auto get(const subrange<I, S, K>& r);                                 // freestanding
 
   template<size_t N, class I, class S, subrange_kind K>
     requires (N < 2)
-    constexpr auto get(subrange<I, S, K>&& r);
+    constexpr auto get(subrange<I, S, K>&& r);                                      // freestanding
 }
 
 namespace std {
-  using ranges::get;
+  using ranges::get;                                                                // freestanding
 }
 
 namespace std::ranges {


jwakely:
Does the using-declaration to add it to `std` need the *// freestanding* comment too?

JohelEGP:
> Does the using-declaration to add it to `std` need the _// freestanding_ comment too?

This relates to https://github.com/cplusplus/draft/issues/5724#issuecomment-1224169410. The _using-declaration_ is a declaration, which can be a freestanding entity, but the literal instructions was to mark entities and not declarations and macros definitions.

jwakely:
But https://cplusplus.github.io/LWG/issue3753 changes that.

JohelEGP:
The same https://github.com/cplusplus/draft/pull/5396/ which moved `get`, also moved `memory_order` variables which are not marked freestanding. That's consistent with the instructions to mark entities, because a variable is not an entity, but inconsistent with the fact that headers marked "all freestanding" have their variables be freestanding because they are declarations. The paper even instructed to explicitly mark some variables as freestanding.

jwakely:
The intention of the paper is pretty obvious, even if the wording was imprecise. Let's not insist on inferior outcomes because of that imprecision. Expect NB comments to ensure everything is freestanding.

My question is not "did the precise wording in P1642R11 say to mark this using-declaration freestanding", it's whether it **needs** to be or not. I think it does need to be.

JohelEGP:
> Does the using-declaration to add it to `std` need the _// freestanding_ comment too?

Then _yes_ is the answer I expect it to be.

jwakely:
> "An entity is a value, object, reference, structured binding, function, enumerator, type, class member, bit-field,
template, template specialization, namespace, or pack."

Those variables are objects, no?

JohelEGP:
AFAIK, the variables designate objects.
> [6](https://eel.is/c++draft/basic#pre-6)[#](http://github.com/Eelis/draft/tree/7687ba685e61d2ea1320e90ed5bc8456df7d87a6/source/basic.tex#L92) A [variable](https://eel.is/c++draft/basic#def:variable) is introduced by the declaration of a reference other than a non-static data member or of an object[.](https://eel.is/c++draft/basic#pre-6.sentence-1)
The variable's name, if any, denotes the reference or object[.](https://eel.is/c++draft/basic#pre-6.sentence-2)

jensmaurer (approved):
Before this gets out of control: This looks good.

@jwakely , feel free to mark the _using-declarations_ as freestanding, too.

jwakely:
> @jwakely , feel free to mark the _using-declarations_ as freestanding, too.

Done.

PR #5840, opened 2022-09-16
[unord.general, container.adaptors.general] Complete the alias template that appears in the reduction guides

hewillk:
This looks like neglect of the [P1206](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p1206r7.pdf).  Is this an editorial?

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -11105,10 +11105,10 @@ The header \libheader{unordered_map} defines the class templates
 
 \pnum
 The exposition-only alias templates
-\exposid{iter\--value\--type}, \exposid{iter\--key\--type},
-\exposid{iter\--mapped\--type}, \exposid{iter\--to\--alloc\--type},
-\exposid{range\--key\--type}, \exposid{iter\--mapped\--type},
-and \exposid{range\--to\--alloc\--type}
+\exposid{iter-value-type}, \exposid{iter-key-type},
+\exposid{iter-mapped-type}, \exposid{iter-to-alloc-type},
+\exposid{range-key-type}, \exposid{range-mapped-type},
+and \exposid{range-to-alloc-type}
 defined in \ref{associative.general} may appear in deduction guides for unordered containers.
 
 \rSec2[unord.map.syn]{Header \tcode{<unordered_map>} synopsis}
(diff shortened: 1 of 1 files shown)


tkoeppe:
Yes, I think this clearly follows the intention of the paper. @CaseyCarter, @cor3ntin could you please take a look?

cor3ntin:
Yes, this looks fine to me, i missed that sentence when producing the wording.

source/containers.tex
```diff
@@ -13141,7 +13143,8 @@
 \pnum
 The exposition-only alias template \exposid{iter-value-type}
 defined in \ref{sequences.general} and
-the exposition-only alias templates \exposid{iter-key-type} and \exposid{iter-mapped-type}
+the exposition-only alias templates \exposid{iter-key-type}, \exposid{iter-mapped-type},
+\exposid{range-key-type} and \exposid{range-mapped-type}
```
jensmaurer, line 13147: Oxford comma, please

source/containers.tex
```diff
@@ -11106,7 +11106,9 @@
 \pnum
 The exposition-only alias templates
 \exposid{iter\--value\--type}, \exposid{iter\--key\--type},
-\exposid{iter\--mapped\--type}, and \exposid{iter\--to\--alloc\--type}
+\exposid{iter\--mapped\--type}, \exposid{iter\--to\--alloc\--type},
+\exposid{range\--key\--type}, \exposid{iter\--mapped\--type},
+and \exposid{range\--to\--alloc\--type}
```
jensmaurer, line 11111: Can we cut back on that jungle of hyphenation hints? Maybe we need one or two in strategic places here, but not 20.

jensmaurer (changes requested):

hewillk:
Note that there are two identical _`iter-value-type`_  in [[containers]](https://eel.is/c++draft/containers). 

Should the _`iter-value-types`_  in [[associative.general]](https://eel.is/c++draft/associative.general) refer to [[sequences.general]](https://eel.is/c++draft/sequences.general)?

source/containers.tex
hewillk: Absolutely.

jensmaurer (approved):
I don't understand what this paragraph is supposed to achieve normatively in the first place, but the update looks good to me.

PR #5841, opened 2022-09-16
[diagnostics] Remove 'shall', use 'recommended practice'

### LINKED ISSUE #5834: [syserr.general] should use "Recommended practice"
> Components described in 19.5 shall not change the value of errno (19.4). Implementations should leave the
error states provided by other libraries unchanged.

This is normative encouragement, but not using the new way to say that.

Also, change "shall" to "do".

jensmaurer:
Fixes #5834


jwakely:
Thanks!

PR #5845, opened 2022-09-16
[expected] Change stable label from "ctor" to "cons", add xref

tkoeppe:


tkoeppe:
The wiki has been updated to allow both spellings of the label.

jensmaurer:
@tkoeppe , the wiki should offer a single option so that future papers / editors know what to use, legacy aside.

JohelEGP:
Which is the preference? This PR changes from "ctor" to "cons". "cons" was also added to the wiki. If we assume that "ctor" was the preference because it was first on the wiki, this PR going backwards towards legacy.

tkoeppe:
@jensmaurer: OK, I'll make it clear that "cons" is more common and preferred.

tkoeppe:
@JohelEGP I think "cons" is more common among the C++20 material, so let's use that for new material, too.

JohelEGP:
It's certainly the more common generally, too:
```
$ grep rSec * | grep '\.cons]' | wc -l
73
$ grep rSec * | grep '\.ctor]' | wc -l
19
```

Within [expected], there's also [expected.un.ctor], which can be changed in this PR for consistency.

tkoeppe:
@jensmaurer PTAL?

jensmaurer:
https://acronyms.thefreedictionary.com/PTAL says

> PTAL = Post-Traumatic Activity Level

jensmaurer (approved):

AlisdairM:
> @JohelEGP I think "cons" is more common among the C++20 material, so let's use that for new material, too.

Sorry, coming late to the conversation.
I would like to suggest that *.ctor* was the better term, as *.cons* is a very common prefix when talking about C++:
  constructor
  constraint
  constant
  constexpr
  constinit
  consteval
  consume  (OK, I am reaching with this one!)
  
There are several others such as console, consecutive, and consistent that are less likely to be topics for labels.

We should not be using such an ambiguous common prefix as a label where we have idiomatic alternatives.

I will move this to a new issue for discussion if there is interest.

PR #5849, opened 2022-09-17
[ranges] Indexing fixes for iterator and sentinel classes

### LINKED ISSUE #5832: Some exposition-only types are not formatted as such in the index
See `chunk_view`'s _inner-iterator_ and _outer-iterator_ and _sentinel_:

image

And adjacent_transform_view and adjacent_view's _iterator_ and _sentinel_:

image

And the index entries for their `base()` member of various iterator and sentinel types:

image
image

And so on ...

image

image

image

image

image

jensmaurer:
Also add automated checks.

Fixes #5832

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -2929,7 +2929,6 @@ The expression in the \grammarterm{requires-clause} is equivalent to:
 
 \rSec3[range.iota.iterator]{Class \tcode{iota_view::\exposid{iterator}}}
 
-\indexlibraryglobal{iota_view::iterator}%
 \begin{codeblock}
 namespace std::ranges {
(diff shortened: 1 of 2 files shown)


tkoeppe:
The collation seems to go wrong. Maybe we need to remove the markup from the sort order somehow?

jensmaurer:
Oh well.

jensmaurer:
@tkoeppe , do you have a specific example? It doesn't seem to matter in the cases I looked at (e.g. elements_view::iterator / sentinel).

tkoeppe:
I think this is a pre-existing problem, indeed. Let me take a quick look.

tkoeppe:
Hm, this requires a separate fix, so it's unrelated to this patch.

PR #5851, opened 2022-09-17
[support, etc.] New subclause "Arithmetic types".

### LINKED ISSUE #5714: Subclause [stdfloat.syn] is oddly disconnected
We should either start a new subsection "Floating-point types" with additional wording, perhaps relating to C, or otherwise perhaps we can rename "Integer types" to "Numeric types" (or "Arithmetic types") and have both `<cstdint>` and `<stdfloat>` in it.

tkoeppe:
The new subclause contains both "integer types" (<cstdint>) and "extended floating-point types" (<stdfloat>).

Previously, the newly added <stdfloat> synopsis was somewhat disconnected and out of context.

Fixes #5714.

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -1441,7 +1441,7 @@ include at least the headers shown in \tref{headers.cpp.fs}.
 \ref{support.types}      & Types                     & \tcode{<cstddef>}          \\ \rowsep
 \ref{support.limits}     & Implementation properties &
   \tcode{<cfloat>}, \tcode{<climits>}, \tcode{<limits>}, \tcode{<version>} \\ \rowsep
-\ref{cstdint}            & Integer types             & \tcode{<cstdint>}          \\ \rowsep
+\ref{cstdint.syn}        & Integer types             & \tcode{<cstdint>}   \\ \rowsep
 \ref{support.start.term} & Start and termination     & \tcode{<cstdlib>}          \\ \rowsep
 \ref{support.dynamic}    & Dynamic memory management & \tcode{<new>}              \\ \rowsep
 \ref{support.rtti}       & Type identification       & \tcode{<typeinfo>}         \\ \rowsep
--- a/source/support.tex
+++ b/source/support.tex
@@ -1997,7 +1997,8 @@ respectively.
 
 \pnum
 The header \libheader{stdfloat} defines type aliases for
-the optional extended floating-point types\iref{basic.extended.fp}.
+the optional extended floating-point types that are specified in
+\ref{basic.extended.fp}.
 
 \indexheader{stdfloat}%
 \begin{codeblock}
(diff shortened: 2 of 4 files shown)


tkoeppe:
We lose both stable lables `[cstdint]` and `[cstdint.general]` in the process, which were not used very much at all.

source/lib-intro.tex
```diff
@@ -1441,7 +1441,7 @@
 \ref{support.types}      & Types                     & \tcode{<cstddef>}          \\ \rowsep
 \ref{support.limits}     & Implementation properties &
   \tcode{<cfloat>}, \tcode{<climits>}, \tcode{<limits>}, \tcode{<version>} \\ \rowsep
-\ref{cstdint}            & Integer types             & \tcode{<cstdint>}          \\ \rowsep
+\ref{support.arith.types}& Arithmetic types          & \tcode{<cstdint>}, \tcode{<stdfloat>} \\ \rowsep
```
jensmaurer, line 1444: Are we sure `<stdfloat>` was freestanding before? I'm not seeing a marker.

jensmaurer (changes requested):
Mostly harmless.

source/lib-intro.tex
tkoeppe: No, absolutely not, that was just a lack of attention.

I didn't see a good way to keep the actual section heading "Integer types", but maybe we can still use that in this table?

tkoeppe: The alternative would be to add a third nesting level and have:

```
  Arithmetic types
  +- Integer types
  |  +- General
  |  +- Synopsis
  +- Floating-point types
     +- General
     +- Synopsis

But given that each "general" section would only contain a single fluff paragrah, that seemed excessive.

Comment (jensmaurer) [reply source/lib-intro.tex]: The normative content of the "freestanding" table is the list of headers, not the headings, so I'm fine with just using "integer types" here.

**Review by jensmaurer — APPROVED** [review]

File: source/lib-intro.tex (hunk 1 above)
Comment (tkoeppe) [reply source/lib-intro.tex]: Yeah, I think that's a reasonable compromise. Thanks!

File: source/support.tex [hunk 2]
```diff
@@ -1997,7 +1993,11 @@
 respectively.
 \end{note}
 
-\rSec1[stdfloat.syn]{Header \tcode{<stdfloat>} synopsis}
+\rSec2[stdfloat.syn]{Header \tcode{<stdfloat>} synopsis}
+
+\pnum
+The header \libheader{stdfloat} defines type aliases for
+the optional extended floating-point types\iref{basic.extended.fp}.
```
tkoeppe, line 2000: @jwakely, @dkolsen-pgi: I made up this new introductory wordingl could you please have a look if it's OK?

dkolsen-pgi: May I suggest a slight change:

> The header <stdfloat> defines type aliases for the optional extended floating-point types with specified formats ([basic.extended.fp]).

or

> The header <stdfloat> defines type aliases for the optional extended floating-point types that are specified in [basic.extended.fp].

There can be other extended floating-point types beyond those specified in [basic.extended.fp] and [stdfloat.syn].  I want to avoid wording that implies that they are the only extended floating-point types.

tkoeppe: Good idea, thanks, done!

PR #5859, opened 2022-09-20
[meta, thread] Add/amend references to func.require, func.invoke

ecatmur:
In particular, not knowing where to find the definition of *INVOKE* (it's in the library index, but not the main index) is highly frustrating when reading is_invocable et al.

Corrected references in [thread] and added some more for completeness.

There's a bunch of uses of 'invoke' in [algorithm] and [ranges] that could also have references added, but that feels less necessary.

### DIFF
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -1669,7 +1669,7 @@ The following are all \tcode{true}:
 Initializes \tcode{ssource}.
 The new thread of execution executes
 \begin{codeblock}
-invoke(auto(std::forward<F>(f)), get_stop_token(),  // for invoke, see \ref{func.invoke}
+invoke(auto(std::forward<F>(f)), get_stop_token(),  // for \tcode{invoke}, see \ref{func.invoke}
        auto(std::forward<Args>(args))...)
 \end{codeblock}
 if that expression is well-formed,
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -1669,7 +1669,7 @@ The following are all \tcode{true}:
 Initializes \tcode{ssource}.
 The new thread of execution executes
 \begin{codeblock}
-invoke(auto(std::forward<F>(f)), get_stop_token(),  // see \ref{func.invoke}, \ref{thread.jthread.stop}
+invoke(auto(std::forward<F>(f)), get_stop_token(),  // for invoke, see \ref{func.invoke}
        auto(std::forward<Args>(args))...)
 \end{codeblock}
 if that expression is well-formed,
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -1338,7 +1338,7 @@ The following are all \tcode{true}:
 \effects
 The new thread of execution executes
 \begin{codeblock}
-invoke(auto(std::forward<F>(f)), auto(std::forward<Args>(args))...)  // see \ref{func.invoke}
+invoke(auto(std::forward<F>(f)), auto(std::forward<Args>(args))...) // see \ref{func.invoke}
 \end{codeblock}
 with the values produced by \tcode{auto}
 being materialized\iref{conv.rval} in the constructing thread.
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -1338,7 +1338,7 @@ The following are all \tcode{true}:
 \effects
 The new thread of execution executes
 \begin{codeblock}
-invoke(auto(std::forward<F>(f)), auto(std::forward<Args>(args))...)     // see \ref{func.invoke}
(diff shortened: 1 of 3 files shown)


source/threads.tex
```diff
@@ -1338,7 +1338,7 @@
 \effects
 The new thread of execution executes
 \begin{codeblock}
-invoke(auto(std::forward<F>(f)), auto(std::forward<Args>(args))...)
+invoke(auto(std::forward<F>(f)), auto(std::forward<Args>(args))...) // \ref{func.invoke}
```
JohelEGP, line 1341:
```suggestion
invoke(auto(std::forward<F>(f)), auto(std::forward<Args>(args))...)  // see \ref{func.invoke}
```

source/threads.tex
```diff
@@ -1669,7 +1669,7 @@
 Initializes \tcode{ssource}.
 The new thread of execution executes
 \begin{codeblock}
-invoke(auto(std::forward<F>(f)), get_stop_token(),
+    invoke(auto(std::forward<F>(f)), get_stop_token(), // \ref{func.invoke}, \ref{thread.jthread.stop}
```
JohelEGP, line 1672:
```suggestion
invoke(auto(std::forward<F>(f)), get_stop_token(),  // see \ref{func.invoke}, \ref{thread.jthread.stop}
```

JohelEGP (changes requested):

source/threads.tex
```diff
@@ -1338,7 +1338,7 @@
 \effects
 The new thread of execution executes
 \begin{codeblock}
-invoke(auto(std::forward<F>(f)), auto(std::forward<Args>(args))...)  // see \ref{func.invoke}
+invoke(auto(std::forward<F>(f)), auto(std::forward<Args>(args))...) // see \ref{func.invoke}
```
JohelEGP, line 1341: This is wrong. See https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#formatting-declarations-and-definitions.

source/threads.tex
```diff
@@ -1669,7 +1669,7 @@
 Initializes \tcode{ssource}.
 The new thread of execution executes
 \begin{codeblock}
-invoke(auto(std::forward<F>(f)), get_stop_token(),
+invoke(auto(std::forward<F>(f)), get_stop_token(),  // see \ref{func.invoke}, \ref{thread.jthread.stop}
```
jensmaurer, line 1672: We should not add a cross-reference for `get_stop_token`; it's right in this (or the next) section.

jensmaurer (changes requested):
Hm... Maybe we should add INVOKE to the main index instead.
On the other hand, INVOKE is used in a rather scattered way, so just adding cross-references seems plausible.

@tkoeppe?

AlisdairM:
Personal experience: I often look up _INVOKE_ based on conversations outside the standard.  I always start with the main index, thinking it is not an identifier in the library, and fall back on the index of library names as my second choice.  Hence, I advocate adding/moving to the main index.

tkoeppe:
Even if we do add cross-references, I don't think we need to add them repeatedly within the same subclause -- just the first time it comes up? E.g. we do this in [basic.extended.fp], where we say "extended floating-point type" many times, but only the first mention gets a cross-reference.

source/threads.tex
tkoeppe: Agreed.

It's a bit annoying that the comment is not a great place for the cross-reference, since it's not clear which bit of the code it refers to (forward? auto? invoke?). I wonder if an explicit note (e.g. "invoke is defined in \ref(...)") would be clearer -- thoughts

ecatmur: trying "for invoke, see \ref{...}", does that look OK?

source/threads.tex
```diff
@@ -1669,7 +1669,7 @@
 Initializes \tcode{ssource}.
 The new thread of execution executes
 \begin{codeblock}
-invoke(auto(std::forward<F>(f)), get_stop_token(),
+invoke(auto(std::forward<F>(f)), get_stop_token(),  // for invoke, see \ref{func.invoke}
```
JohelEGP, line 1672:
```suggestion
invoke(auto(std::forward<F>(f)), get_stop_token(),  // for \tcode{invoke}, see \ref{func.invoke}
```

JohelEGP (changes requested):

JohelEGP:
I think this should be approved.

tkoeppe:
Squashed into ef7d0e234722d2ef8b4ec069b6f929dc083d8dc7, e66269302b5b551af64c644215e6c6f40dc475fe -- thank you!

PR #5860, opened 2022-09-22
[range.as.const.overview] Remove redundant ranges namespace qualification

hewillk:


jensmaurer (approved):
The argument is that the `views::as_const` that follows in the same example does not have a `ranges::` prefix.

JohelEGP:
`std::views::as_const` is valid anyways.

PR #5869, opened 2022-09-27
[dcl.enum] Enumerators don't have "initializers"

Quuxplusone:
I might just be missing something obvious here, but the current wording of [dcl.enum/2](https://eel.is/c++draft/enum#dcl.enum-2.sentence-8) seems to use the grammar production _initializer_ when it means _constant-expression_. The grammar production [_initializer_](https://eel.is/c++draft/dcl.init.general#nt:initializer) is defined as

    initializer:
        brace-or-equal-initializer
        '(' expression-list ')'

I could have just replaced "an _initializer_" with "a _constant-expression_". But, looking at the preceding sentence, I think "with `=`" and "without `=`" are the perfect way to describe what we're talking about here. So this change is bigger (but still, IMHO, editorial).

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -6972,12 +6972,11 @@ constants, and can appear wherever constants are required.
 An \grammarterm{enumerator-definition} with \tcode{=} gives the associated
 \grammarterm{enumerator} the value indicated by the
 \grammarterm{constant-expression}.
-If the first \grammarterm{enumerator}
-has no \grammarterm{initializer}, the value of the corresponding constant
-is zero. An \grammarterm{enumerator-definition} without an
-\grammarterm{initializer} gives the \grammarterm{enumerator} the value
-obtained by increasing the value of the previous \grammarterm{enumerator}
-by one.
+An \grammarterm{enumerator-definition} without \tcode{=} gives the associated
+\grammarterm{enumerator} the value zero
+if it is the first \grammarterm{enumerator-definition},
+and the value of the previous \grammarterm{enumerator}
+increased by one otherwise.
 \begin{example}
 \begin{codeblock}
 enum { a, b, c=0 };


source/declarations.tex
```diff
@@ -6900,12 +6900,11 @@
...
-has no \grammarterm{initializer}, the value of the corresponding constant
-is zero. An \grammarterm{enumerator-definition} without an
-\grammarterm{initializer} gives the \grammarterm{enumerator} the value
-obtained by increasing the value of the previous \grammarterm{enumerator}
-by one.
+An \grammarterm{enumerator-definition} without \tcode{=} gives the associated
+\grammarterm{enumerator} the value obtained by
+increasing the value of the previous \grammarterm{enumerator} by one.
+If the first \grammarterm{enumerator-definition} has no \tcode{=}, the associated
+\grammarterm{enumerator} is given the value zero.
```
JohelEGP, line 6907: The last sentence doesn't sound like an exception to the first. It also reads better when the zero case is upfront.
```suggestion
\grammarterm{enumerator} the value zero
if it is the first \grammarterm{enumerator-definition},
and the value of the previous \grammarterm{enumerator}
increased by one otherwise.
```

Quuxplusone: FWIW, I like that suggestion.

tkoeppe: @Quuxplusone Let's apply it!

tkoeppe:
@jensmaurer I'd like some core(ish) pair of eyes if possible.

source/declarations.tex
```diff
@@ -6972,12 +6972,11 @@
 An \grammarterm{enumerator-definition} with \tcode{=} gives the associated
 \grammarterm{enumerator} the value indicated by the
 \grammarterm{constant-expression}.
-If the first \grammarterm{enumerator}
-has no \grammarterm{initializer}, the value of the corresponding constant
-is zero. An \grammarterm{enumerator-definition} without an
-\grammarterm{initializer} gives the \grammarterm{enumerator} the value
-obtained by increasing the value of the previous \grammarterm{enumerator}
-by one.
+An \grammarterm{enumerator-definition} without \tcode{=} gives the associated
```
jensmaurer, line 6975: Can we quote the entire grammar snippet here, e.g. "An enumerator-definition without = _constant-expression_ ..."?

Otherwise, a twisted soul might try to circumvent this by putting "a=b" inside the constant-expression, given a suitably overloaded operator=.

jensmaurer (changes requested):
I think the rewording is an improvement.

source/declarations.tex
jensmaurer: Uh, I just noticed the "with =" in the preceding sentence, so it seems "without =" is fine as-is.

jensmaurer (approved):
@tkoeppe , looks good to me.

Quuxplusone:
@jensmaurer this is both "approved" and labeled [cwg]; does that mean it's awaiting CWG review, or could be merged, or what? No new urgency; I just noticed it's been sitting around a while.

jensmaurer:
@tkoeppe, I don't think we need CWG approval for this; it seems fairly straightforward in a low-density part of the specification.

Can this go ahead?

PR #5873, opened 2022-09-28
[index] Remove undefined term

AlisdairM:
The notion of a basic _source_ character set has been eliminated by the work on a better specification for translating source. Remove the last definition to it in the index, that points to a definition that no longer exists.


tkoeppe:
@cor3ntin, @tahonermann could you please take a look?

cor3ntin:
This makes sense to me, that term no longer exists indeed

tahonermann:
I concur, the term is dead, long live the new term!

PR #5882, opened 2022-10-03
[class.union.general] Attach example to its note

JohelEGP:
The example relates to the note which relates to the p4's first sentence.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -3152,7 +3152,6 @@ of the union must be user-provided or it will
 be implicitly deleted\iref{dcl.fct.def.delete} for the union.
 \end{note}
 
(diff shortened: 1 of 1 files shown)


JohelEGP:
The same logic could be applied to p7, which includes a note and example for the very long p6.

tkoeppe:
I'm not sure about moving away, where did you see that? Even the "typical example" in https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#latex-source-code-formatting contains such a nested example.

JohelEGP:
You're right. I'm misremembering. A quick search reveals many such occurrences in the draft.

PR #5885, opened 2022-10-04
[intro.defs] Update hyperlink according to SC22 requirements

tkoeppe:


tkoeppe:
I'm not sure how you feel about the trailing slash: one the one hand it's what ISO said, and it avoids a pointless HTTP redirect, but on the other hand it looks a bit cluttery.

jwakely (approved):
The slash is fine IMHO.

jwakely:
Nobody looks at that page much anyway!

PR #5886, opened 2022-10-04
[conv.lval] Fix cross-reference for 'invalid pointer value'

blackteahamburger:
The definition of 'invalid pointer value' is in [basic.compound] rather than [basic.stc.dynamic.deallocation].

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -666,7 +666,7 @@ type, the conversion copy-initializes the result object from
 the glvalue.
 
 \item Otherwise, if the object to which the glvalue refers contains an invalid
-pointer value\iref{basic.stc.dynamic.deallocation}, the behavior is
+pointer value\iref{basic.compound}, the behavior is
 \impldef{lvalue-to-rvalue conversion of an invalid pointer value}.
 
 \item Otherwise, the object indicated by the glvalue is read\iref{defns.access},
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -5240,7 +5240,7 @@ a \defn{pointer past the end of} an object\iref{expr.add}, or
 the \defnx{null pointer value}{value!null pointer} for that type, or
 \item
 \indextext{invalid pointer value|see{value, invalid pointer}}
-an \defnx{invalid pointer value}{value!invalid pointer}.
+an \defnx{invalid pointer value}{value!invalid pointer}\iref{basic.stc.general}.
 \end{itemize}
 A value of a
 pointer type


JohelEGP:
With a `git reset --hard HEAD~` and `git push -f remote` you wouldn't have needed to close #5883.

blackteahamburger:
> With a `git reset --hard HEAD~` and `git push -f remote` you wouldn't have needed to close #5883.

Thanks.

tkoeppe:
> The definition of 'invalid pointer value' is in [basic.compound] rather than [basic.stc.dynamic.deallocation].

Could you please point out where that definition is exactly?

JohelEGP:
It's https://eel.is/c++draft/basic.compound#3.4 by exhaustion.

JohelEGP:
For instance, https://eel.is/c++draft/basic.stc.general#4 refers to the correct place.


Perhaps it'd be better to use a `\label`.

tkoeppe:
Right, so isn't `basic.stc.general` the actual point of definition, not `basic.compound`?

JohelEGP:
> Right, so isn't `basic.stc.general` the actual point of definition, not `basic.compound`?

I'm not sure. But at the very least, just like how https://eel.is/c++draft/basic.compound#3.2 refers to where it's specified how to form a pointer past the end of an object, p3.4 could refer to [basic.stc.general].

tkoeppe:
Yes, could you make that change then?

JohelEGP:
As a separate PR? I'm not the author of this one.

tkoeppe:
Sorry, no, I should have directed that to the PR author! @blackteahamburger ^^^

blackteahamburger:
> > Right, so isn't `basic.stc.general` the actual point of definition, not `basic.compound`?
> 
> I'm not sure. But at the very least, just like how https://eel.is/c++draft/basic.compound#3.2 refers to where it's specified how to form a pointer past the end of an object, p3.4 could refer to [basic.stc.general].

See #5893.

blackteahamburger:
> > Right, so isn't `basic.stc.general` the actual point of definition, not `basic.compound`?
> 
> I'm not sure. But at the very least, just like how https://eel.is/c++draft/basic.compound#3.2 refers to where it's specified how to form a pointer past the end of an object, p3.4 could refer to [basic.stc.general].

Done.

jensmaurer (approved):

source/expressions.tex
```diff
@@ -666,7 +666,7 @@
 the glvalue.
 
 \item Otherwise, if the object to which the glvalue refers contains an invalid
-pointer value\iref{basic.stc.dynamic.deallocation}, the behavior is
+pointer value\iref{basic.compound}, the behavior is
```
tkoeppe, line 669: What about here, shouldn't this also refer to basic.stc.general for the definition of "invalid pointer"? Why is basic.compound a good reference?

blackteahamburger: But [basic.stc.general]/4 specifies how to form an invalid pointer value, while [basic.compound]/3.4 is the definition (so [basic.stc.general]/4 refers to [basic.compound]/3.4).

And I think

> Indirection through an invalid pointer value and passing an invalid pointer value to a deallocation function have undefined behavior. Any other use of an invalid pointer value has implementation-defined behavior.

in [basic.stc.general]/4 should be better placed in [basic.compound]/3.  But [basic.compound]/3 may need to be split because it is too long.

tkoeppe: Sorry, could you explain a bit how basic.general is the definition? I don't see "invalid" mentioned in basic.general except for in that list in p3 (which doesn't seem like a definition?) and in notes.

jensmaurer: [basic.compound] p3 contains the taxonomy of pointer values. It states "there exists a pointer value that is an invalid pointer value".

[basic.stc.general] p4 tells us about one way how to obtain an invalid pointer value (maybe there are others). It also explains a bit of the semantics of an invalid pointer value, but this should probably be moved elsewhere in the long run. For example, [conv.lval] p3.3 tells us a little more about an invalid pointer value.

tkoeppe: Yes, right. In fact, [conv.lval] is the very subject of this PR. I'm not sure where the definition of invald value could be moved to improve the presentation; isn't basic.stc.general a good place? We could just add a cross reference to it from [basic.compound]/(3.4).

jensmaurer: [basic.stc.general] shouldn't talk about indirection through an invalid pointer value; that should be handled by unary "operator*".

jensmaurer: Sure, but that's converted to use "operator*" (expr.ref p2 and expr.mptr.oper p3).

jensmaurer: > Maybe the last two sentences of [[basic.stc.general]p4](https://eel.is/c++draft/basic.stc.general#4) should become a note?

Except that we seem to lack normative statements elsewhere.

tkoeppe: Yes, the normative wording would have to move somewhere more appropriate.

tkoeppe:
I think we can derive some useful cleanup opportunities here, so let's keep this open and look at it again some time.

source/expressions.tex
jensmaurer: CWG2822 moves the normative wording "somewhere else", for unrelated reasons.

tkoeppe: Shall we wait for that to land then?

tkoeppe:
This is now fixed by 5be40a6b59527e82b13a29722c623635065759bf.

PR #5887, opened 2022-10-04
[obj.lifetime] Use static_cast instead of C-style cast expression

blackteahamburger:


Dani-Hub:
This doesn't work, because this wording is supposed to be valid for pointers to cv-char as well, and the static_cast cannot do this.

blackteahamburger:
> This doesn't work, because this wording is supposed to be valid for pointers to cv-char as well, and the static_cast cannot do this.

Done.

tkoeppe:
@jwakely could you please take a look?

tkoeppe:
Thanks! So is even the status quo questionable then? Should I file an issue?

jwakely:
Yes please.

tkoeppe:
Done. This is now being handled as an LWG issue.

PR #5890, opened 2022-10-05
[format.string.escaped] Fix invalid example

mordante:
Example 5 should only have one invalid code unit. The second code unit is a valid code point. This issue was already in the paper
  P2286R8 Formatting Ranges

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -15406,7 +15406,7 @@ string s3 = format("[{:?}, {:?}]", '\'', '"');      // \tcode{s3} has value: \tc
 string s4 = format("[{:?}]", string("\0 \n \t \x02 \x1b", 9));
                                                     // \tcode{s4} has value: \tcode{["\textbackslash u\{0\} \textbackslash n \textbackslash t \textbackslash u\{2\} \textbackslash u\{1b\}"]}
 string s5 = format("[{:?}]", "\xc3\x28");           // invalid UTF-8
-                                                    // \tcode{s5} has value: \tcode{["\textbackslash x\{c3\}\textbackslash x\{28\}"]}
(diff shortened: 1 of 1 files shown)


tkoeppe:
@cor3ntin, could you please take a look?

cor3ntin:
@tkoeppe This is _borderline_ design as the wording does not specify what constitute an invalid code unit sequence
https://eel.is/c++draft/format.string.escaped#2.2 

I think the proposed change is consistent with sg-16 consensus on what the behavior should be.
Either way the change is better than status quo given the associated wording

+ @tahonermann

tahonermann:
The proposed change illustrates the intended behavior. The example was previously discussed on the SG16 mailing list (see [the thread with subject "Handling ill-formed Unicode in the library"](https://lists.isocpp.org/sg16/2022/09/3369.php)) and multiple people expressed support for the change while no one voiced opposition. Corentin is correct that the specification lacks a concrete description of how the boundaries of an ill-formed code unit sequence are determined. Addressing that will require normative changes to [\[format.string.escaped\]](https://eel.is/c++draft/format.string.escaped). I'm confident that any changes made there would be consistent with the amended example; SG16 is highly unlikely to endorse any ill-formed code unit sequence definition that is not aligned with [Unicode PR-121](http://unicode.org/review/pr-121.html).

jensmaurer (approved):

tkoeppe:
OK, thanks, everyone! @tahonermann: are you tracking the specification deficiencies in an issue elsewhere? I'll apply this change and close this PR in the meantime.

tahonermann:
@tkoeppe, I am now: https://github.com/sg16-unicode/sg16/issues/80.

PR #5892, opened 2022-10-06
[dcl.link] Fix the wording of the application of the specified language linkage

blackteahamburger:
[dcl.link]/1:

> All functions and variables **whose names have external linkage** and all function types have a language linkage.


tkoeppe:
@zygoloid Could you have a look?

zygoloid (approved):
Yes, this looks right to me.

There are cases where a _linkage-specification_ causes a name to have external linkage (either by implying an `extern` or by causing a declaration to be attached to the global module), so there's some risk of this change introducing ambiguity, but those cases rely on whether the _linkage-specification_ is syntactically present rather than whether a language linkage is applied to the name, so I think this is fine.

PR #5896, opened 2022-10-07
[vector.bool.pspc] Add a blank line between the private and public members of std::vector<bool>::reference

blackteahamburger:


tkoeppe:
This seems rather piecemeal. Similar newlines are missing in other places, even in containers.tex. Could you have a broader look?

blackteahamburger:
> This seems rather piecemeal. Similar newlines are missing in other places, even in containers.tex. Could you have a broader look?

I've done some, please help find more.

tkoeppe:
Could we maybe have one commit per file, and you can combine label references like `[foo.bar, foo.xyz] Tweak vertical whitespace`.

blackteahamburger:
> Could we maybe have one commit per file, and you can combine label references like `[foo.bar, foo.xyz] Tweak vertical whitespace`.

OK. But there are so many of them, can I just write like [ranges]?

tkoeppe:
Yes, some deepest common ancestor is fine.

PR #5897, opened 2022-10-07
Ill-formed example in [range.drop.while.overview]

huixie90:
The example in [range.drop.while.overview] is ill-formed. It attempts to use a `const char*` as a `range`, which isn't. The fix simply makes it a `const char (&)[N]`, which is a `range`

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -6270,7 +6270,7 @@ is expression-equivalent to \tcode{drop_while_view(E, F)}.
 \pnum
 \begin{example}
 \begin{codeblock}
-constexpr auto source = "  \t   \t   \t   hello there";
+constexpr decltype(auto) source = "  \t   \t   \t   hello there";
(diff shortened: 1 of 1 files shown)


source/ranges.tex
```diff
@@ -6270,7 +6270,7 @@
 \pnum
 \begin{example}
 \begin{codeblock}
-constexpr auto source = "  \t   \t   \t   hello there";
+constexpr decltype(auto) source = "  \t   \t   \t   hello there";
```
CaseyCarter, line 6273: The best fix is to make it a `string_view`:
```suggestion
constexpr auto source = "  \t   \t   \t   hello there"sv;
```
String literals should never be used as ranges because nobody has ever wanted the null terminator in their range.

CaseyCarter (changes requested):

CaseyCarter (approved):

PR #5898, opened 2022-10-10
[meta.const.eval] Add is_constant_evaluated() to index of library names

marcmutz:
Previously, only its feature-test macro was listed there.


tkoeppe:
Thanks!

PR #5907, opened 2022-10-18
[mdspan.extents.expo] remove redundant // \expos

hewillk:


tkoeppe:
I don't think this is the only place where we say `// \expos` in an itemdecl; why do you want to remove it just in these few cases? I think we might want a more general discussion/decision if we want to make any changes here, but I also don't consider that terribly pressing.

tkoeppe:
Editorial meeting consensus: See also #5498, we seem to have agreed to have `\expos` comments _only_ in the synopsis, and never later on. We should record that in the wiki and clean up. This PR is a good part of that.

tkoeppe:
Added to the wiki.

PR #5910, opened 2022-10-19
[intro.defs] Move the definition of "multibyte character" to library

cor3ntin:
This address the US-2 NB comment as per the direction voted on by SG-16.

No use of "multibyte character" remains before [multibyte.strings] (we still cross reference "multibyte string" in [basic.start.main] but it's pre-existing (and otherwise correct).

@steve-downey @jensmaurer @tahonermann

### DIFF
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -422,11 +422,6 @@ assignment of an rvalue of some object type to a modifiable lvalue of the same t
 \indexdefn{construction!move}%
 direct-initialization of an object of some type with an rvalue of the same type
 
-\indexdefn{character!multibyte}%
-\definition{multibyte character}{defns.multibyte}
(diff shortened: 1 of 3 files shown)


jensmaurer (approved):

jensmaurer:
@jwakely, SG16 thinks the definition in [defns.multibyte] is too disconnected from the relevant library clauses, and that causes confusion. It seems this move is entirely editorial.

jwakely (approved):

tahonermann:
The SG16 record of discussion is available at https://github.com/sg16-unicode/sg16-meetings#october-19th-2022. The proposed changes are consistent with the polled direction.

jensmaurer:
@tkoeppe, you're next.

PR #5915, opened 2022-10-25
[container.adaptors] uses_allocator specializations consistently use "Alloc"

Quuxplusone:
This is not intended to trigger (nor un-trigger) any blanket wording; it's purely cosmetic, aligning the parameter names for the flat_foo containers' `uses_allocator` specializations, with the parameter names for stack/queue/priority_queue.

Intuitively, `Alloc` is the more appropriate name because this type is not necessarily something that qualifies as an Allocator; it is the same kind of type as the `Alloc` constructor parameters in [queue.defn] for example.

My understanding is that the old wording's choice to use `Alloc` in one specialization and `Allocator` in another is _not_ motivated by any intentional difference in semantics between the two specializations; it's just a bit of inconsistency that we can patch up editorially.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -13285,9 +13285,8 @@ namespace std {
   inline constexpr sorted_unique_t sorted_unique{};
 
   template<class Key, class T, class Compare, class KeyContainer, class MappedContainer,
-           class Allocator>
-    struct uses_allocator<flat_map<Key, T, Compare, KeyContainer, MappedContainer>,
-                          Allocator>;
+           class Alloc>
+    struct uses_allocator<flat_map<Key, T, Compare, KeyContainer, MappedContainer>, Alloc>;
 
   // \ref{flat.map.erasure}, erasure for \tcode{flat_map}
   template<class Key, class T, class Compare, class KeyContainer, class MappedContainer,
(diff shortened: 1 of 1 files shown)


tkoeppe:
Hm, is there a clear preference?

```
$ git grep 'class Allocator\b' | wc -l
438
$ git grep 'class Alloc\b' | wc -l
179
```

Quuxplusone:
> Hm, is there a clear preference?
> 
> ```
> $ git grep 'class Allocator\b' | wc -l
> 438
> $ git grep 'class Alloc\b' | wc -l
> 179
> ```

Yes: places that require _types that qualify as allocators_ tend to use `Allocator`, and places that don't require that tend to use `Alloc`. https://eel.is/c++draft/containers#sequence.reqmts-68.3
I imagine there's still some inconsistency, but in general, my mental model is: "If it would happily accept a `std::pmr::memory_resource*`, it should be named `Alloc` not `Allocator`." (And if it wouldn't, vice versa.)

In the specific case of `uses_allocator`, I'm sure my mental model applies:
```
$ git grep 'uses_allocator<.*Allocator\b' | wc -l
      5 (after this PR, 0)
$ git grep 'uses_allocator<.*Alloc\b' | wc -l
      15 (after this PR, 20)
```

tkoeppe:
Ah, I see. Thank you! Let me take another look.

source/containers.tex
```diff
@@ -13253,9 +13253,9 @@
   inline constexpr sorted_unique_t sorted_unique{};
 
   template<class Key, class T, class Compare, class KeyContainer, class MappedContainer,
-           class Allocator>
+           class Alloc>
```
tkoeppe, line 13256: Could you please check if linebreaks can now be removed?

Quuxplusone: This has been done (since November). Ping?

PR #5919, opened 2022-10-27
[dcl.init.list]/(3.10) Specify the type of the prvalue

languagelawyer:


source/declarations.tex
```diff
@@ -5809,11 +5809,11 @@
 \end{example}
 
 \item Otherwise, if \tcode{T} is a reference type, a prvalue is generated.
-The prvalue initializes its result object by copy-list-initialization.
+The prvalue initializes its result object by copy-list-initialization from the initializer list.
```
tkoeppe, line 5812: @jensmaurer Do you prefer "initializes with" or "initializes from"?

jensmaurer: "from" sounds fine here.

PR #5922, opened 2022-10-29
[range.dangling] The `tag` of a `struct` is not a C++ term

hawkinsw:
Only in C is the term `tag` defined with respect to `struct`s. Replace the term "tag type" with simply "type" in [range.dangling].

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -2138,7 +2138,7 @@ else
 \rSec2[range.dangling]{Dangling iterator handling}
 
 \pnum
-The tag type \tcode{dangling} is used together with the template aliases
+The type \tcode{dangling} is used together with the template aliases
 \tcode{borrowed_iterator_t} and \tcode{borrowed_subrange_t}.
 When an algorithm
 that typically returns an iterator into, or a subrange of, a range argument
@@ -2146,7 +2146,7 @@ is called with an rvalue range argument
 that does not model \libconcept{borrowed_range}\iref{range.range},
 the return value possibly refers to a range whose lifetime has ended.
 In such cases,
-the tag type \tcode{dangling} is returned instead of an iterator or subrange.
+the type \tcode{dangling} is returned instead of an iterator or subrange.
 \indexlibraryglobal{dangling}%
 \begin{codeblock}
 namespace std::ranges {


jwakely:
This isn't talking about "the tag of a struct" though. It's a tag type as in "tag dispatching" https://www.fluentcpp.com/2018/04/27/tag-dispatching/

jwakely:
https://www.boost.org/community/generic_programming.html#tag_dispatching

hawkinsw:
I assumed that's *how* it was ultimately being used, there was just not
enough context for me to solidify my sense. Do you think it's worth making
that clear?

hawkinsw:
Upon further reflection, I still believe this is a good change to make. I am also concerned by the sentence 

> In such cases, the type `dangling` is returned instead of an iterator or subrange.

which seems to conflate returning a type with returning an object of a certain type. 

It seems to me that this would be better phrased as 

> In such cases, an object of type `dangling` is returned instead of an object whose type is an iterator or subrange.

JohelEGP:
There's no object involved in `return dangling{};`. See https://eel.is/c++draft/stmt.return#2.sentence-4 and https://wg21.link/LWG3805.

hawkinsw:
Thank you! I completely understand. My point was simply that it seems a bit
inaccurate to leave the language the way it is. Do you agree?

hawkinsw:
> There's no object involved in `return dangling{};`. See https://eel.is/c++draft/stmt.return#2.sentence-4 and https://wg21.link/LWG3805.

It will be interesting to see the resolution to the issue that you have open!

JohelEGP:
I'm not sure. There's also the example of https://eel.is/c++draft/std.iterator.tags#1.sentence-2.

hawkinsw:
Oh! That's exactly what we want here, then!

hawkinsw:
Well, maybe not *exactly* what we want. But, I think it highlights the
fact that we want to say something more than just "type" in this
sentence. Would you agree with that?

JohelEGP:
It already does. It says "tag type".

hawkinsw:
Which is a term not defined anywhere else in the standard. Worse, it echoes
the definition of tag from C in the context of `struct`s.

That said, I am not trying to argue and I don't mean to question your
expertise. I am just trying to be helpful.

If I can continue to refine the language, let me know. Otherwise, I'll let
it be.

JohelEGP:
> I am just trying to be helpful.

We all are. Obviously, the "something more" you meant wasn't the preexisting "tag". I shouldn't have taken it literally.

Now I'm questioning the use of "tag" standing for "tag dispatching". Unlike the iterator's tag dispatching, this one is on the way out. And the tag is actually "whether the range argument is an rvalue", whereas `dangling` is the result of the dispatch.

The
> something more than just "type"

you suggest exists in the whole paragraph, which attempts to explain what "tag" alone couldn't. So I think just dropping "tag" is an improvement in face of the potential confusion with the C term and "tag" as an adjective being an undefined term for types.

hawkinsw:
> > I am just trying to be helpful.
> 
> We all are. Obviously, the "something more" you meant wasn't the preexisting "tag". I shouldn't have taken it literally.
> 

No problem! I get nervous in these situations because everyone *else* is the expert!

> Now I'm questioning the use of "tag" standing for "tag dispatching". Unlike the iterator's tag dispatching, this one is on the way out. And the tag is actually "whether the range argument is an rvalue", whereas `dangling` is the result of the dispatch.
> 
> The
> 
> > something more than just "type"
> 
> you suggest exists in the whole paragraph, which attempts to explain what "tag" alone couldn't. So I think just dropping "tag" is an improvement in face of the potential confusion with the C term and "tag" as an adjective being an undefined term for types.

If I hear you correctly, then you think that the original PR (where we simply drop the *tag* word) is good?

Thanks again for engaging!

JohelEGP:
> If I hear you correctly, then you think that the original PR (where we simply drop the _tag_ word) is good?

I personally do. We'll have to see what _the_ experts think.

hawkinsw:
> > If I hear you correctly, then you think that the original PR (where we simply drop the _tag_ word) is good?
> 
> I personally do. We'll have to see what _the_ experts think.

Fair enough, but *I* think you're an expert! :-)

jensmaurer (approved):
This looks like an improvement to me, because it avoids the confusion with C struct tags. And it's not an iterator category tag, either (for tag dispatching).

@jwakely , what do you think?

hawkinsw:
Thank you, *@*jensmaurer

CaseyCarter (approved):
"tag" really isn't contributing anything here.

hawkinsw:
> "tag" really isn't contributing anything here.

Thank you @CaseyCarter !

jwakely:
> And it's not an iterator category tag, either (for tag dispatching).

Iterator tags aren't the only kind of tags used for tag dispatching though. I'd prefer to keep it, but then we'd have to define tag dispatching (or tag types) in the standard. So I will let it go, somewhat reluctantly.

hawkinsw:
> > And it's not an iterator category tag, either (for tag dispatching).
> 
> Iterator tags aren't the only kind of tags used for tag dispatching though. I'd prefer to keep it, but then we'd have to define tag dispatching (or tag types) in the standard. So I will let it go, somewhat reluctantly.

I'd be willing to work with you on defining that phrase and introducing it, if you'd like!! @jwakely

jensmaurer:
(While I understand the type `dangling` may be used for tag dispatching, I thought its primary purpose for users is to make careless programs ill-formed.)

jensmaurer:
@tkoeppe , on to you.

jwakely:
> its primary purpose for users is to make careless programs ill-formed.

Calling another algorithm with the result either dispatches to a function taking an iterator, or fails to compile (i.e. dispatches to nothing). Using the return type to select a valid overload or an invalid overload is a form of tag dispatching IMO.

PR #5924, opened 2022-10-30
[flat.map] s/compare/key_comp()/

Quuxplusone:
It looks to me as if the identifier `compare` is not defined in these scopes.

(I also think `key_equiv` should be replaced throughout with `@\exposid{key-equiv}@`, but exposid markup is an orthogonal issue with discussion in #5812. I _also_ think the `key_equiv(key_compare)` ctor should be `explicit`, but that's purely stylistic AFAICT.)

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -14985,7 +14985,7 @@ sorts the range \range{begin()}{end()} with respect to \tcode{value_comp()}; and
 finally erases the duplicate elements as if by:
 \begin{codeblock}
 auto zv = ranges::zip_view(c.keys, c.values);
-auto it = ranges::unique(zv, key_equiv(compare)).begin();
+auto it = ranges::unique(zv, key_equiv(key_comp())).begin();
 auto dist = distance(zv.begin(), it);
 c.keys.erase(c.keys.begin() + dist, c.keys.end());
 c.values.erase(c.values.begin() + dist, c.values.end());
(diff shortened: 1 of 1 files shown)


JohelEGP:
Is `[flat.map]` right? `flat_map` does have a `compare` non-static data member. If you expand the context of the first change towards the top, you can even see

Quuxplusone:
> Is `[flat.map]` right? `flat_map` does have a `compare` non-static data member.

Oh, bother. This whole thing might just be a subset of the "need to `\exposid` more things" issue, then. I think this PR is still a minor improvement in clarity, but maybe it's _not_ a mistake as-written. Feel free to close this PR if that's the general consensus.

I also just noticed that https://eel.is/c++draft/flat.map#overview-10 arguably should say "range that is not sorted with respect to `value_comp()`" instead of "with respect to `key_­comp()`", but don't plan to submit a PR for that.

JohelEGP:
> I also just noticed that https://eel.is/c++draft/flat.map#overview-10 arguably should say "range that is not sorted with respect to `value_comp()`" instead of "with respect to `key_­comp()`", but don't plan to submit a PR for that.

IIRC, only the keys are required to be sorted.

Quuxplusone:
> > I also just noticed that https://eel.is/c++draft/flat.map#overview-10 arguably should say "range that is not sorted with respect to `value_comp()`" instead of "with respect to `key_­comp()`", but don't plan to submit a PR for that.
> 
> IIRC, only the keys are required to be sorted.

Yes, but in
```
    template<class InputIterator>
      flat_map(sorted_unique_t s, InputIterator first, InputIterator last,
               const key_compare& comp = key_compare())
```
the range expressed by `[first, last)` is a range of `value_type` (i.e. `pair<Key,T>`) that must be sorted w.r.t. `value_comp()`. `key_comp()` will be something like `less<Key>`, which doesn't take pairs at all.  But I think you've just found _another_ typo, namely "...a container, <strike>containers,</strike> or range that is not sorted..." I don't think there's any situation where _more than one_ container needs to be sorted by _anything._ Anyway, my overall sense is that this intro wording is vague and messy and that's exactly why I'm not going to tackle submitting a PR for it. :)

Orthogonally yet again, I notice that the `from_range_t` overloads do not support e.g. `mySortedVector | views::unique | ranges::to<flat_set<int>>(sorted_unique)`. But I assume that's at least semi-intentional and thus LEWG territory.

PR #5925, opened 2022-10-30
[ratio.ratio] Use defined function for sign

### LINKED ISSUE #5879: [ratio.ratio] What is `sign`?
> [(2.1)](https://eel.is/c++draft/ratio.ratio#2.1)
num shall have the value sign(N) * sign(D) * abs(N) / gcd[.](https://eel.is/c++draft/ratio.ratio#2.1.sentence-1)

JohelEGP:
Resolves #5879.
Before and after:

Is it OK to only use math font on the relevant subexpressions?

### DIFF
--- a/source/meta.tex
+++ b/source/meta.tex
@@ -2498,7 +2498,7 @@ where \tcode{gcd} represents the greatest common divisor of the absolute values 
 \tcode{N} and \tcode{D}:
 
 \begin{itemize}
-\item \tcode{num} shall have the value \tcode{sign(N) * sign(D) * abs(N) / gcd}.
+\item \tcode{num} shall have the value \tcode{$\operatorname{sgn}(\tcode{N})$ * $\operatorname{sgn}(\tcode{D})$ * abs(N) / gcd}.
 \item \tcode{den} shall have the value \tcode{abs(D) / gcd}.
 \end{itemize}
 


source/meta.tex
```diff
@@ -2498,7 +2498,7 @@
 \tcode{N} and \tcode{D}:
 
 \begin{itemize}
-\item \tcode{num} shall have the value \tcode{sign(N) * sign(D) * abs(N) / gcd}.
+\item \tcode{num} shall have the value \tcode{$\text{sgn}(\tcode{N})$ * $\text{sgn}(\tcode{D})$ * abs(N) / gcd}.
```
tkoeppe, line 2501: This shouldn't use `\text` but something like `\operatorname`.

tkoeppe: (If we used those more often, we would add `\DeclareMathOperator` to the preamble.)

JohelEGP: Before, after and after-after:

source/meta.tex
```diff
@@ -2498,7 +2498,7 @@
 \tcode{N} and \tcode{D}:
 
 \begin{itemize}
-\item \tcode{num} shall have the value \tcode{sign(N) * sign(D) * abs(N) / gcd}.
+\item \tcode{num} shall have the value \tcode{$\operatorname{sgn}(\tcode{N})$ * $\operatorname{sgn}(\tcode{D})$ * abs(N) / gcd}.
```
tkoeppe, line 2501: Do we need _this_ many nested `\tcode`s? Can't it just be a single, all-encompassing `\tcode`?

JohelEGP: I think so, due to the math font. Grepping the sources for `'tcode.*\$'` reveals many such cases. Perhaps `\operatorname` can be used without math font.

JohelEGP: Ah, but then, the parentheses of _sgn_ would be in code font, but _sgn_ is a math function, not a code entity.

tkoeppe: Ah yes, makes sense, thanks!

PR #5926, opened 2022-10-30
[climits.syn] Correct note about types of macros

Eric41293:
This note was added apparently as a response to LWG 416 (https://cplusplus.github.io/LWG/issue416). However, the submitter of that issue misunderstood the C standard. The required types are not the types that would be produced by an unadorned integer literal, but rather those produced by the integer promotions. Thus, only the macros for (signed|unsigned)? char and (unsigned)? will have a different type. And this is what actual implementations do: https://godbolt.org/z/5E1MojhGh.

Accordingly, the note, as currently written, is at best misleading, because it implies that all the macros may have types that don't match.

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -1744,8 +1744,10 @@ namespace std {
 The header \libheader{climits} defines all macros the same as
 the C standard library header \libheader{limits.h}.
 \begin{note}
-The types of the constants defined by macros in \libheader{climits} are not
-required to match the types to which the macros refer.
+Except for \tcode{CHAR_BIT}, the types of the constants defined by macros in
+\libheader{climits} that refer to types of integer conversion rank less than
+that of \tcode{int} are the types obtained by the integral
+promotions\iref{conv.prom}.
 \end{note}
 
 \xrefc{5.2.4.2.1}


jensmaurer:
What we have is

"The types of the constants defined by macros in \libheader{climits} are not
required to match the types to which the macros refer."

Is that an incorrect summary of the C standard? Do you have a cross reference to the C standard to cross-check that statement?

Eric41293:
ISO C 5.2.4.2.1 states,

"The values given below [of the macros defined by <limits.h>] shall be replaced by constant expressions suitable for use in #if preprocessing directives.

"Moreover, except for CHAR_BIT and MB_LEN_MAX, the following shall be replaced by expressions that
have the same type as would an expression that is an object of the corresponding type converted
according to the integer promotions."

ISO C 6.3.1 describes the integer promotions:

"The following may be used in an expression wherever an `int` or `unsigned int` may be used:
"— An object or expression with an integer type (other than `int` or `unsigned int`) whose integer
conversion rank is less than or equal to the rank of `int` and `unsigned int`.
"— A bit-field of type `_Bool`, `int`, `signed int`, or `unsigned int`.
"If an `int` can represent all values of the original type (as restricted by the width, for a bit-field), the
value is converted to an `int`; otherwise, it is converted to an `unsigned int`. These are called the
_integer promotions_. All other types are unchanged by the integer promotions."

I consider the current wording to be inaccurate because some of the macros **are** required to have the same type as the type to which they refer (namely, those referring to int, long, long long, and their unsigned variants).

jwakely (changes requested):
I agree with the problem statement. Could the suggested new wording be shortened by omitting "that refer to types of integer conversion rank less than that of `int`"? The statement is true for all of them, it's just that the integer promotions don't change the type for `INT_MAX`, `LONG_MAX` etc.

And why isn't `MB_LEN_MAX` mentioned in your suggested rewording?

Eric41293:
I have removed the reference to conversion rank, but still felt that we need a phrase to explain what types the integral promotions are applied to. I also added `MB_LEN_MAX`. I had omitted it on the grounds that it does not refer to a type, and so does not need to be explicitly excluded. But I think better of that now.

jensmaurer:
I find this wordy. Also, can we make that singular for extra precision?

Suggestion:

> Except for ..., a macro referring to an integer type T defines a constant whose type is the promoted type of T (conv.prom).

jensmaurer (approved):

tkoeppe:
Looks good -- @jwakely?

jwakely (approved):
Works for me.

PR #5931, opened 2022-11-04
[execpol.unseq] Fix inconsistent indent

onihusube:
I have visually checked with [eel](http://eel.is/c++draft/execpol#unseq) and fixed the inconsistencies in the code, but have not built and checked.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -13417,6 +13417,7 @@ if the invocation of an element access function exits via an exception,
 class execution::unsequenced_policy { @\unspec@ };
 \end{itemdecl}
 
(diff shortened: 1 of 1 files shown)


jensmaurer (commented):
@tkoeppe, this seems reasonable, but it's sort-of novel to document classes (as opposed to member functions) with `itemdescr`.

tkoeppe:
Yes, usually we show a class synopsis without `\itemdecl` instead. Would that be an option here?

tkoeppe:
Sorry, we're already doing that and the `\itemdesc` is just missing.

PR #5934, opened 2022-11-05
[time.clock.system.nonmembers], [time.zone.zonedtime.nonmembers] add <charT>

jwakely:
It doesn't mean anything to use `STATICALLY-WIDEN` withOUT a template argument list.


jensmaurer (approved):
@tkoeppe, looks good to me.

PR #5939, opened 2022-11-08
[atomics] Replace integral and floating-point placeholders

jwakely:
This resolves NB comment [GB-129](https://github.com/cplusplus/nbballot/issues/438).


tkoeppe:
I'd like to apply the NB resolution merges in one batch, so let's earmark this.

tkoeppe:
We should think of a suitably uniform commit message, akin to the straw poll commits.

Either something like `NB VA-007: [foo.bar] Unindent dedeprecation negation`, or maybe as a second paragraph, `Fixes NB VA-007, CK-001`. We'll discuss in the editorial meeting.

jwakely:
N.B. _integer-type_ and _floating-point-type_ match the placeholders in [charconv].

PR #5945, opened 2022-11-10
[expected] Remove redundant 'and' in multiple conditions

hewillk:


eisenwave:
Could you rebase this please? This look like it fixes inconsistencies with how we usually present lists in quite a lot of cases, so this is good value.

PR #5947, opened 2022-11-12
[mdspan.extents.cons] "constexpr" should precede "explicit"

CaseyCarter:
In library function declarations, the `constexpr` specifier always precedes the `explicit` specifier per our standing style guide.


CaseyCarter:
It's correct in the class body, but not in the itemdecl.

JohelEGP:
What prevented the check script from noticing this?

CaseyCarter:
> What prevented the check script from noticing this?

The fact that the script doesn't check for this, which I've attempted to remedy with #5948.

JohelEGP:
Oh, I thought there already was such a check in place. Probably because Clang Format already does this.

PR #5948, opened 2022-11-12
[check] Add check for "explicit constexpr" in the library

CaseyCarter:
Per our style guide, `constexpr` should precede `explicit` in declarations. Nevertheless, the persnickety `explicit constexpr` seems to frequently sneak into the library wording.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -9238,7 +9238,7 @@ namespace std {
     constexpr iterator erase(const_iterator position);
     constexpr iterator erase(const_iterator first, const_iterator last);
     constexpr void swap(vector&);
-    constexpr static void swap(reference x, reference y) noexcept;
+    static constexpr void swap(reference x, reference y) noexcept;
     constexpr void flip() noexcept;     // flips all bits
     constexpr void clear() noexcept;
   };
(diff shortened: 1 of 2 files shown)


Mick235711:
We might also consider enforcing:
- `static constexpr` (412 occurrence) instead of `constexpr static` (7 occurrence)
- either `static inline` (2) or `inline static` (3)
- either `extern constexpr` (1) or `constexpr extern` (1)

(Just from a grep search, highly likely to not find some hidden positives like the one in [\[depr.static.constexpr\]](https://eel.is/c++draft/depr.static.constexpr), but the checking script is just doing grep anyway)

CaseyCarter:
> We might also consider enforcing:
> 
> * `static constexpr` (412 occurrence) instead of `constexpr static` (7 occurrence)
> * either `static inline` (2) or `inline static` (3)
> * either `extern constexpr` (1) or `constexpr extern` (1)

Most of the inconsistent orderings you found are in examples in core wording, which is fine and intended. Only in the library wording are we concerned about following a consistent order of specifiers in our *decl-specifier-seq*s (See https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines).

I added a checker for `static constexpr`, which occurs tons in the library and once in the wrong order `vector<bool>`. I haven't bothered for `static inline` and `extern constexpr` which don't appear in the library wording at all.

Mick235711:
Thanks for the clarification! That make sense for me.

jwakely (approved):

tkoeppe:
Thanks, Casey!

Issue #5628, opened 2022-07-17
[swappable.requirements] Consider to introduce Cpp17Swappable requirements as new editorial convenience requirement


Dani-Hub:
The by far most often used "swappable" requirement used in the current normative wording has the style

> lvalues of type T are swappable

defined in [swappable.requirements]. This is rather lengthy wording for a simple thing, and has caused editorial hiccups such as the one mentioned in #5627. I suggest introducing a new editorial vocabulary, _Cpp17Swappable_ to be introduced in subclause [swappable.requirements], for example as follows:

>A type `X` meets the _Cpp17Swappable_ requirements if lvalues of type `X` are swappable.

just before line 5 defining the _Cpp17ValueSwappable_ requirements and to replace using existing wording of the above mentioned form "lvalues of type T are swappable" by "T meets the _Cpp17Swappable_ requirements". 

This approach would be similar to the style used for other _Cpp17_ requirements and can be easily migrated into a collection of such requirements. For example, both bullets (5.1) and (5.2) of [unord.hash] could easily be merged into a single bullet of the form:

> meet the Cpp17Hash requirements (Table 36), with Key as the function call argument type, the
> Cpp17DefaultConstructible requirements (Table 29), the Cpp17CopyAssignable requirements (Table 33), the Cpp17Swappable requirements ([swappable.requirements]),

A similar and by far not last example is the potential merge of formatter.requirements] bullet (1.2) into the (1.1.x) sub-bullets, etc.

This particular wording form also matches quite well with syntactic requirements such as `is_swappable_v` versus the most general form `is_swappable_with_v`.

If this suggestion goes beyond an editorial action, please let me know, whether instead an official paper proposal would be preferred to realize these effects.

Dani-Hub:
I have decided to provide an official proposal to realize these effects. I would appreciate if we could keep the issue open until I have added a reference to the provided proposal (assuming this happens within an acceptable time frame).

Dani-Hub:
A paper exists now, [D2696R0](https://isocpp.org/files/papers/D2696R0.html), so feel free to close this issue now.

Dani-Hub:
The paper has been published now: [P2696R0](https://isocpp.org/files/papers/P2696R0.html).

PR #5978, opened 2022-11-16
2022-11 LWG Motion 14

### LINKED ISSUE #5627: [sequence.reqmts] uses atypical and unclear wording for Cpp17 requirements and swappable conditions
During the submission process of [LWG 3742](https://cplusplus.github.io/LWG/issue3742) I realized that the suggested wording violates our current wording style in the following aspects:

- For the _Cpp17MoveConstructible_ and _Cpp17MoveAssignable_ we should better use the form "`T` meets the _Cpp17XX_ requirements" instead of "`T` is _Cpp17XX_"
- Instead of saying that "`T` is swappable" we should say that "lvalues of type `T` are swappable"

The plain reason why we didn't decide for the more appropriate wording forms as part of this LWG issue was caused by the observation that the _current_ local wording style used in subclause [sequence.reqmts] follows the same unusual style with the agreement that I would open an editorial issue to take care of this problem. 

To be more specific, for example for the specification of `insert_range` we currently say:

> Preconditions: T is Cpp17EmplaceConstructible into X from *ranges::begin(rg). For vector and
> deque, T is also Cpp17MoveInsertable into X, Cpp17MoveConstructible, Cpp17MoveAssignable, and
> swappable (16.4.4.3).

According to the preferred wording forms this should better say something like this:

> Preconditions: T is Cpp17EmplaceConstructible into X from *ranges::begin(rg). For vector and
> deque, T is also Cpp17MoveInsertable into X, T meets the Cpp17MoveConstructible and 
> Cpp17MoveAssignable requirements, and lvalues of type T are swappable (16.4.4.3).

I would like to get feedback whether such a rephrasing is possible as part of editorial work and to which extend, otherwise I would open a separate LWG issue to realize such wording changes. Of-course I'm also willing to make a corresponding editorial PULL request once we have agreement that such a change is editorially appropriate.

### LINKED ISSUE #5975: [2022-11 LWG Motion 14] P2696R0 Introduce Cpp17Swappable as additional convenience requirements
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2696r0.html

tkoeppe:
P2696R0 Introduce Cpp17Swappable as additional convenience requirements

Fixes #5975.
Also fixes #5627.
Also fixes https://github.com/cplusplus/draft/issues/5628.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -2011,9 +2011,8 @@ bidirectional.iterators,random.access.iterators}.
 \pnum
 A type \tcode{X} meets the \defnoldconcept{Iterator} requirements if:
 \begin{itemize}
-\item \tcode{X} meets the \oldconcept{CopyConstructible}, \oldconcept{CopyAssignable}, and
-\oldconcept{Destructible} requirements\iref{utility.arg.requirements} and lvalues
-of type \tcode{X} are swappable\iref{swappable.requirements}, and
+\item \tcode{X} meets the \oldconcept{CopyConstructible}, \oldconcept{CopyAssignable}, \oldconcept{Swappable}, and
+\oldconcept{Destructible} requirements\iref{utility.arg.requirements,swappable.requirements}, and
 
 \item \tcode{iterator_traits<X>::difference_type} is a signed integer type or \keyword{void}, and
 
--- a/source/support.tex
+++ b/source/support.tex
@@ -3397,11 +3397,10 @@ namespace std {
 The type \tcode{source_location} meets the
 \oldconcept{DefaultConstructible},
 \oldconcept{CopyConstructible},
-\oldconcept{Copy\-Assignable}, and
+\oldconcept{Copy\-Assignable},
+\oldconcept{Swappable}, and
 \oldconcept{Destructible}
-requirements\iref{utility.arg.requirements}.
-Lvalues of type \tcode{source_location}
-are swappable\iref{swappable.requirements}.
+requirements\iref{utility.arg.requirements,swappable.requirements}.
 All of the following conditions are \tcode{true}:
 \begin{itemize}
 \item \tcode{is_nothrow_move_constructible_v<source_location>}
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -2473,6 +2473,9 @@ constexpr void swap(const tuple& rhs) const noexcept(@\seebelow@);
 
 \begin{itemdescr}
 \pnum
+Let $i$ be in the range \range{0}{sizeof...(Types)} in order.
+
+\pnum
 \mandates
 \begin{itemize}
 \item
@@ -2485,13 +2488,11 @@ For the second overload,
 
 \pnum
 \expects
-Each element in \tcode{*this} is swappable with\iref{swappable.requirements}
-the corresponding element in \tcode{rhs}.
+For all $i$, \tcode{get<$i$>(*this)} is swappable with\iref{swappable.requirements} \tcode{get<$i$>(rhs)}.
 
 \pnum
 \effects
-Calls \tcode{swap} for each element in \tcode{*this} and its
-corresponding element in \tcode{rhs}.
+For each $i$, calls \tcode{swap} for \tcode{get<$i$>(*this)} with \tcode{get<$i$>(rhs)}.
 
 \pnum
 \throws
@@ -3942,7 +3943,7 @@ constexpr void swap(optional& rhs) noexcept(@\seebelow@);
 
 \pnum
 \expects
-Lvalues of type \tcode{T} are swappable.
+\tcode{T} meets the \oldconcept{Swappable} requirements\iref{swappable.requirements}.
 
 \pnum
 \effects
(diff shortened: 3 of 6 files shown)


Dani-Hub:
What kind of thing is "\oldconcept{Swap\\-pable}"? (I'm asking about the intermediate "\\-")

tkoeppe:
@Dani-Hub That's a hyphenation hint.

Dani-Hub:
Thanks for the explanation @tkoeppe !

source/support.tex
```diff
@@ -3395,11 +3395,10 @@
...
 \oldconcept{DefaultConstructible},
 \oldconcept{CopyConstructible},
-\oldconcept{Copy\-Assignable}, and
+\oldconcept{Copy\-Assignable},
+\oldconcept{Swappable}, and
 \oldconcept{Destructible}
-requirements\iref{utility.arg.requirements}.
-Lvalues of type \tcode{source_location}
-are swappable\iref{swappable.requirements}.
+requirements\iref{utility.arg.requirements, swappable.requirements}.
```
JohelEGP, line 3401:
```suggestion
requirements\iref{utility.arg.requirements,swappable.requirements}.
```
The convention.
`grep 'iref[^}]\+, ' *.tex` has no hits, whereas
`grep 'iref[^}]\+,' *.tex` has 190.

source/iterators.tex
```diff
@@ -2005,9 +2005,8 @@
 \pnum
 A type \tcode{X} meets the \defnoldconcept{Iterator} requirements if:
 \begin{itemize}
-\item \tcode{X} meets the \oldconcept{CopyConstructible}, \oldconcept{CopyAssignable}, and
-\oldconcept{Destructible} requirements\iref{utility.arg.requirements} and lvalues
-of type \tcode{X} are swappable\iref{swappable.requirements}, and
+\item \tcode{X} meets the \oldconcept{CopyConstructible}, \oldconcept{CopyAssignable}, \oldconcept{Swappable}, and
+\oldconcept{Destructible} requirements\iref{utility.arg.requirements, swappable.requirements}, and
```
JohelEGP, line 2009:
```suggestion
\oldconcept{Destructible} requirements\iref{utility.arg.requirements,swappable.requirements}, and
```
See TBD.

JohelEGP (changes requested):
Also resolves #5627. From the paper:
> [Drafting note: The suggested changes for p37 and p41 provide an alternative resolution for [editorial issue 5627](https://github.com/cplusplus/draft/issues/5627)]

source/iterators.tex
tkoeppe: Sure, why not.

source/support.tex
tkoeppe: OK :-)

JohelEGP (approved):

Dani-Hub:
@JohelEGP @tkoeppe This paper should also resolve #5628, unfortunately, I forgot to mention this in the paper (But #5628 refers to the paper, so there exists a connection).

source/utilities.tex
```diff
@@ -2476,13 +2479,11 @@
...
 \expects
-Each element in \tcode{*this} is swappable with\iref{swappable.requirements}
-the corresponding element in \tcode{rhs}.
+For all $i$, \tcode{get<$i$>(*this)} is swappable with\iref{swappable.requirements} \tcode{get<$i$>(rhs)}.
 
 \pnum
 \effects
-Calls \tcode{swap} for each element in \tcode{*this} and its
-corresponding element in \tcode{rhs}.
+For each $i$, calls \tcode{swap} for \tcode{get<$i$>(*this)} with \tcode{get<$i$>(rhs)}.
```
jwakely, line 2486: I didn't notice this earlier, but "calls `swap` for `x` with `y`" is not how we'd usually say it. I would expect "swaps x with y" but when talking about a call, call "`swap` for `x` *and* `y`" seems better to me. This is what LWG approved though.

source/utilities.tex
```diff
@@ -15010,13 +15011,11 @@
...
-\item \oldconcept{CopyAssignable} (\tref{cpp17.copyassignable}), and
+\item \oldconcept{CopyAssignable} (\tref{cpp17.copyassignable}),
+\item \oldconcept{Swappable}\iref{swappable.requirements}, and
 \item \oldconcept{Destructible} (\tref{cpp17.destructible})
 \end{itemize}
-requirements,
-
-\item
-it is swappable\iref{swappable.requirements} for lvalues, and
+requirements, and
```
jwakely, line 15018: I think it's debatable whether we need to use an itemized list now that we only have two top-level items. Maybe it's a bit more readable to have each requirement on its own line like that, but it doesn't match any of the other lists of *Cpp17Foo* requirements being altered by this change.

jwakely:
It confused me that the edits for [container.reqmts] p37 and p41 are actually for [sequence.reqmts], sorry for not catching that in LWG.

jwakely (approved):
A couple of editorial comments, but this is a correct application of the approved changes.

source/utilities.tex
tkoeppe: Let's make this an issue!

source/utilities.tex
tkoeppe: I'll send a separate PR.

PR #5998, opened 2022-11-22
[expr.unary.op] Cleanup of {ones',two's} complement, and of "operator"

tkoeppe:
We are currently using the term "unary complement" to refer to the "~" operator without ever giving it that name.

This change removes unnecessary uses of the terms "unary" and "complement", but adds notes that make referenes to "ones'/two's complement".

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -4664,7 +4664,7 @@ The type of the result is \keyword{bool}.
 
 \pnum
 \indextext{operator!ones' complement}%
-The operand of \tcode{\~{}} shall have integral or unscoped enumeration type.
+The operand of the \tcode{\~{}} operator shall have integral or unscoped enumeration type.
 Integral promotions are performed.
 The type of the result is the type of the promoted operand.
 % FIXME: [basic.fundamental]/p5 uses $x_i$; [expr] uses $\tcode{x}_i$.
@@ -4677,7 +4677,7 @@ is 1 if $\tcode{x}_i$ is 0, and 0 otherwise.
 There is an ambiguity
 in the grammar when \tcode{\~{}} is followed by
 a \grammarterm{type-name} or \grammarterm{decltype-specifier}.
-The ambiguity is resolved by treating \tcode{\~{}} as the unary complement
+The ambiguity is resolved by treating \tcode{\~{}} as the
 operator rather than as the start of an \grammarterm{unqualified-id}
 naming a destructor.
 \begin{note}
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -4653,6 +4653,10 @@ promotion is performed on integral or enumeration operands. The negative
 of an unsigned quantity is computed by subtracting its value from $2^n$,
 where $n$ is the number of bits in the promoted operand. The type of the
 result is the type of the promoted operand.
+\begin{note}
+The result is the two's complement of the operand
+(where operand and result are considered as unsigned).
+\end{note}
 
 \pnum
 \indextext{operator!logical negation}%
@@ -4674,6 +4678,10 @@ of the promoted operand \tcode{x},
 the coefficient $\tcode{r}_i$
 of the base-2 representation of the result \tcode{r}
 is 1 if $\tcode{x}_i$ is 0, and 0 otherwise.
+\begin{note}
+The result is the ones' complement of the operand
+(where operand and result are considered as unsigned).
+\end{note}
 There is an ambiguity
 in the grammar when \tcode{\~{}} is followed by
 a \grammarterm{type-name} or \grammarterm{decltype-specifier}.


tkoeppe:
@opensdh could you please have a look?

source/expressions.tex
```diff
@@ -4683,6 +4686,9 @@
 the coefficient $\tcode{r}_i$
 of the base-2 representation of the result \tcode{r}
 is 1 if $\tcode{x}_i$ is 0, and 0 otherwise.
+\begin{note}
+The result is the ones' complement (or unary complement) function of the operand.
```
tkoeppe, line 4690: I'm actually not sure if "unary complement" on L4695 means "ones' complement", or just "the (unary) "complement"-operator" (where "complement" is not otherwise qualified), given that `~` is also spelled `compl`.

opensdh: I'm pretty sure "unary complement" is just referring to the arity of the operator: the "ones" in "ones' complement" refers to all-bits-set (0xffff), and "unary" would _count_ those bits.  That said, "complement of _x_" here just means "_X_-_x_ for some _X_ determined from context" and is at least [300 years older](https://en.wikipedia.org/wiki/Method_of_complements#Practical_uses) than its use in binary.  Since there is obviously two's complement as well, I think the note should just say
> The result is the ones' complement of the operand.

source/expressions.tex
```diff
@@ -4662,6 +4662,9 @@
 of an unsigned quantity is computed by subtracting its value from $2^n$,
 where $n$ is the number of bits in the promoted operand. The type of the
 result is the type of the promoted operand.
+\begin{note}
+The result is the two's complement function of the operand.
```
opensdh, line 4666: This should differ in only the operative word: the "considered as unsigned" still applies when we're thinking of this as a complement operation (rather than merely _commenting_ that negative numbers are represented by the two's complement in this system), and the result is not itself a function.

source/expressions.tex
tkoeppe: OK -- should we then also modify "unary complement operator" below, since we don't currently define "complement operator"?

opensdh: There isn't any other `~` operator, so I don't think it's harmful the way it is.  That said, it would be fine to just delete "unary complement" there; those two words are already available in the subclause heading and note.  (I would be tempted to start the paragraph with "The operand of the complement operator `~` shall&hellip;", but none of the other descriptions actually _name_ the operators except in the subclause names.)

tkoeppe: Middle ground: "The result of the `~` operator shal...". That's in line with the other paragraphs, but we don't say "unary" since there is no binary version.

source/expressions.tex
tkoeppe: "The sine of 0 is 0" vs "the sine applied to 0 is 0"... both work, don't they? Yes, "of" has two meanings (function declaration vs evaluation), but it's not _wrong_ to use for evaluation.

I don't quite understand the first part; could you suggest a complete note?

tkoeppe: That said, I'm happy to lose the notes entirely now that we got rid of "unary"!

source/expressions.tex
opensdh: Sure (with "operand" instead of "result").

tkoeppe: Yes, sorry, that's what I actually have in this patch now. PTAL!

source/expressions.tex
opensdh: Yes, but no one would say "the error function of 0.5" but rather "the error function{,'s value, evaluated} at 0.5" because "function of" is confusing.  (It is funny, though, that even English has declaration-follows-use here.)

I just meant
> The result is the two's complement of the operand (both considered as unsigned).

opensdh (approved):
I thought the notes were helpful (partly for people looking for the old definitions), but this small change seems to be an improvement.

source/expressions.tex
tkoeppe: Oh, OK, and for ones' complement you'd say the same, just with "ones'" instead of "two's"?

tkoeppe:
We can keep the notes if you find them useful!

source/expressions.tex
opensdh: Yes, exactly.

tkoeppe: Ah yes, that's what you suggested above. PTAL.

opensdh: It looks good to me; you might want to check that no one has qualms about "considering" things as unsigned in a note (even though the actual bit behavior should be pretty obvious to all).

tkoeppe: @jensmaurer, @burblebee, @jfbastien: any qualms about "considering" things as unsigned?

source/expressions.tex
```diff
@@ -4683,10 +4686,13 @@
 the coefficient $\tcode{r}_i$
 of the base-2 representation of the result \tcode{r}
 is 1 if $\tcode{x}_i$ is 0, and 0 otherwise.
+\begin{note}
+The result is the ones' complement of the operand (both considered as unsigned).
```
jensmaurer, line 4690: I find "both" to be referring to "operand" and "ones' complement", which seems unintended.

Maybe just drop "both" and leave it at that, considering this is a note anyway?

tkoeppe: Done, PTAL.

opensdh: I did actually mean both of those things: for a signed operand, you pretend that it's unsigned (with the same bits) and take whichever kind of complement, producing another unsigned value, but that's the result only when pretending that _it_ is unsigned in the same fashion.  That said, if people think it's clear without the "both", that works for me.

tkoeppe: How about "(result and operand considered as unsigned)"?

jensmaurer: Works for me.

tkoeppe: Done.

jensmaurer (approved):
Looks good.

@opensdh, are you ok with this?

opensdh:
> @opensdh, are you ok with this?

Very much so.

source/expressions.tex
tkoeppe: We still haven't resolved whether there are qualms!

jensmaurer: I'm fine with the phrasing. It's a note.

source/expressions.tex
zygoloid: For what it's worth: [ISO/IEC 2382:2015](https://www.iso.org/obp/ui/#iso:std:iso-iec:2382:ed-1:v2:en), which we normatively cite for definitions of terms such as this, does indeed define "complement" consistent with @opensdh's suggestion:

> ### complement
> number that is derived from a given number by subtracting it from a specified number
>
> Note 1 to entry: Example: In fixed radix notation, the specified number is typically a power of the radix or one less than a given power of the radix.
>
> Note 2 to entry: A negative of a number is often represented by its complement.

Issue #5997, opened 2022-11-21
Reword "corresponding function templates"


tkoeppe:
1. It should presumably be "two _declarations of_ function templates", since it is declarations that correspond (and the bullet immediately above says that they declare something).
1. Is the idea that we should explicitly refer to satisfaction of the constraints?  We already have in [temp.over.link]/6.4 that a *template-head* can be "satisfied".

_Originally posted by @opensdh in https://github.com/cplusplus/draft/pull/5984#discussion_r1028373547_

PR #5999, opened 2022-11-22
[temp.over.link] Reword to clarify that declarations correspond

tkoeppe:
Fixes #https://github.com/cplusplus/draft/issues/5997.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -3916,9 +3916,7 @@ declare the same name,
 have corresponding signatures\iref{basic.scope.scope},
 \item
 would declare the same entity,
-where in the determination in \ref{basic.link}
-of whether two declarations do declare the same entity,
-$A$ and $B$ are assumed to correspond, and
+when considering $A$ and $B$ to correspond in the determination per \ref{basic.link}, and
 \item
 accept and are satisfied by the same set of template argument lists,
 \end{itemize}
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -3911,12 +3911,12 @@ no diagnostic required.
 Furthermore, if two declarations $A$ and $B$ of function templates
 \begin{itemize}
 \item
-declare the same name,
+introduce the same name,
 \item
 have corresponding signatures\iref{basic.scope.scope},
 \item
 would declare the same entity,
-when considering $A$ and $B$ to correspond in the determination per \ref{basic.link}, and
+when considering $A$ and $B$ to correspond in that determination\iref{basic.link}, and
 \item
 accept and are satisfied by the same set of template argument lists,
 \end{itemize}
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -3908,10 +3908,21 @@ If the validity or meaning of the program depends on
 whether two constructs are equivalent, and they are
 functionally equivalent but not equivalent, the program is ill-formed,
 no diagnostic required.
-Furthermore, if two declarations of function templates with the same name
-and corresponding signatures\iref{basic.scope.scope} do not correspond
-but would declare the same entity\iref{basic.link} considering them to do so, and
+Furthermore, if two declarations $A$ and $B$ of function templates
+\begin{itemize}
+\item
+declare the same name,
+\item
+have corresponding signatures\iref{basic.scope.scope},
+\item
+would declare the same entity,
+where in the determination in \ref{basic.link}
+of whether two declarations do declare the same entity,
+$A$ and $B$ are assumed to correspond, and
+\item
 accept and are satisfied by the same set of template argument lists,
+\end{itemize}
+but do not correspond,
 the program is ill-formed, no diagnostic required.
 
 \pnum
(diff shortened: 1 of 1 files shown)


tkoeppe:
@burblebee, @opensdh, @jensmaurer: For your consideration, please!

W-E-Brown:
I just happened to glance at this draft rewording:  I'm sorry, but I can't understand the proposed sentence.  Please revise, preferably into two or more sentences; this one as drafted just has too many convolutions for me to decipher.  Failing that, please restore the original.

tkoeppe:
@W-E-Brown: Could you please tell me how you understand the original wording? I also find this a bit weird, but I don't think this change is at fault here -- the problem for me already resided in the original third bullet, "would declare the same entity considering them to correspond". I thought I'm just preserving that weirdness. If you tell me what the original meant, I can try and look for a clearer rewording.

W-E-Brown:
IMO, "preserving that weirdness" seems to be a questionable objective of rewording an admittedly questionable specification.  Alas, I see no improvement in clarity in the proposed text.

Having any "weirdness" in a core specification seems strong motivation for consulting CWG or even opening a core issue:  I suggest that CWG be asked to clarify the intention and its wording here, since "considering them to correspond" is (AFAIK) not a defined term of art.  Pending resolution of such inquiry, I respectfully recommend no wording change, as I find the proposed bullet-free specification much harder to parse than the existing specification with its bullet list.

tkoeppe:
The wording came in fresh  out of Core with CWG2603, and I wouldn't normally second-guess Core's deliberations and care. @jensmaurer, @opensdh Could you perhaps shed some light on how "considering them to correspond" is to be interpreted?

jensmaurer:
The idea of that bullet is to do a "specification function call" to "same entity" with the override "the declarations do correspond" (although, in fact, they don't).

"Same entity" says in 6.6 p8:

> Two declarations of entities declare the same entity if, considering declarations of unnamed types to introduce their names for linkage purposes, if any (9.2.4, 9.7.1), they correspond (6.4.1), have the same target scope that is not a function or template parameter scope, and either ...

We want to have this list of conditions checked, except that "they correspond" should be assumed to be true for that specification function all.

"considering them to correspond" seems to be on-point for that.

tkoeppe:
OK, I get that, but I still think "if two function templates [...] would declare the same entity considering them to correspond" is a pretty awkward way to say that. I think the fact that the "them" refers to something quite distant, and also to the fact that you're expected to spot that "correspond" is a detail of "same entity" make this quite a strain.

In other words, this wording makes sense if you already have your explanation in mind, but the explanation itself seems hard to discover just based off this wording.

Would it hurt to just spell this reasoning out?

"""
* would declare the same entity,<ins> where in the determination in \ref{basic.link} of whether two declarations do declare the same entity, the declarations are assumed to correspond,</ins> and
"""

tkoeppe:
Maybe we can decorate the whole sentence a bit more:

""
Furthermore, if two declarations $A$ and $B$ of function templates that do not correspond
* declare the same name,
* have corresponding signatures,
* would declare the same entity, where in the determination in \ref{basic.link} of whether two declarations do declare the same entity, $A$ and $B$ are assumed to correspond, and
* accept and are satisfied by the same set of template argument lists,

the program is ill-formed [...]
"""

tkoeppe:
We could also move the punchline (that the declarations do not correspond) to the end:

"""
Furthermore, if two declarations $A$ and $B$ of function templates [List] but do not correspond, the program is ill-formed [...]
"""

tkoeppe:
image

@jensmaurer, @opensdh What do you think of this suggestion?

jensmaurer:
I like moving the punchline.  I'm not too happy with that third bullet; it feels too wordy.

maybe

"would declare the same entity, when considering A and B to correspond in the determination per 6.6,"

tkoeppe:
Done:

image

jensmaurer:
@opensdh, what do you think?

tkoeppe:
Even more radical, and I'm not actually proposing that:

===
The program is ill-formed, no diagnostic required, if

   * the validity or meaning of the program [...] and they are functionally equivalent but not equivalent, or
   * two declarations $A$ and $B$ do not correspond, but they
       * declare the same name,
       * have corresponding signatures,
       * would declare the same entity [...], and
       * accept and are satisfied by the same set of template argument lists.

===

opensdh:
I like 58b98d1426cdbe6dccb49424d8b30393fe03ee87 (though perhaps with some slight rewordings like "declare the same name" &rarr; "introduce the same name" and "to correspond in the determination per 6.6," &rarr; "in that determination (6.6),").  The whole point here is to recompute the same-entity status with a deliberately weaker filter just as with "functionally equivalent but not equivalent".

jensmaurer (approved):
Works for me.

tkoeppe:
Thanks, everyone!

PR #6002, opened 2022-11-24
[uaxid, macros, etc.] Smaller capital-style glyphs

tkoeppe:
The "hash"/"number" sign is reduced in size and placed on the baseline. This seems to look less distracting and disruptive.

We also use the usual small-caps presentation for universal character names.

### DIFF
--- a/source/uax31.tex
+++ b/source/uax31.tex
@@ -1,14 +1,14 @@
 %!TEX root = std.tex
-\infannex{uaxid}{Conformance with UAX \#31}
+\infannex{uaxid}{Conformance with \UAX{31}}
 
 \rSec1[uaxid.general]{General}
 
 \pnum
 This Annex describes the choices made in application of
-UAX \#31 (``Unicode Identifier and Pattern Syntax'')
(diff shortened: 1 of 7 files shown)


tkoeppe:
Preview:

image

source/uax31.tex
```diff
@@ -49,8 +49,8 @@
 \rSec2[uaxid.def.rfmt]{R1a Restricted format characters}
 
 \pnum
-If an implementation of UAX \#31 wishes to allow format characters
-such as ZERO WIDTH JOINER or ZERO WIDTH NON-JOINER
+If an implementation of \UAX{31} wishes to allow format characters
+such as \uname{zero width joiner} or \uname{zero width non-joiner}
```
jensmaurer, line 53: Please use the full U+nnnn NAME style here, as we do everywhere else we refer to a Unicode character.

jensmaurer (commented):
Mostly harmless.

source/uax31.tex
tkoeppe: Done!

jensmaurer (approved):
Great, thanks.

Issue #3911, opened 2020-03-26
[std] Use \libmember and \libglobal everywhere


Eelis:
These macros were added by #3096 but are currently almost unused. That's a shame, because in principle their use enables more precise indexing and linking, which I would love to take advantage of in cxxdraft-htmlgen.

For example, a link such as http://eel.is/c++draft/string.iterators#lib:basic_string,cbegin currently points to `begin` rather than `cbegin`, because the index entry for `cbegin` is actually located just before the itemdecl that declares both `begin` and `cbegin`. Had `\libmember` been used here, I could link specifically to the declaration of `cbegin`.

Although.. that brings up the question of how these macros are intended to be used with overloaded names like `begin`. Is the intent to:

1. use the macro on all overloads
2. use the macro only on the first overload
3. not use the macro at all for overloaded names
4. something else? :)

I'd be happy to make a pull request to use these macros everywhere, but do need guidance on this issue of how to handle overloaded names. @jensmaurer do you remember whether you considered overloaded names when you made these macros?

jensmaurer:
I don't think I thought about overloads.

We do not use \libmember broadly right now, because we sometimes have `//` comments later in the same line, aligned with nearby comments. And the necessary LaTeX escape-foo plus the `\libmember` macro invocation makes the necessary alignment non-obvious in the source code, which doesn't help the editing.

So, please hold back on global `\libmember` patches.

Eelis:
Ah I see, makes sense.

An alternative idea I had was that perhaps there could be an environment that wraps a whole declaration, so that instead of

    struct X {
      int @\libmember{f}{X}@();
    }

you get something more along the lines of

    struct X {
      @\begin{libmember}{f}{X}@
      int f();
      @\end{libmember}@
    };

Bit more verbose, but easy to edit, and even better for cxxdraft-htmlgen because then I can make it highlight the whole declaration instead of just the name.

I have no idea if mixing environments with @ like this is even possible in LaTeX though.

jensmaurer:
Add-on: I think we're looking forward to `\libmember` patches where the library member is defined in the class or header synopsis and does not have a separate `\itemdecl`. (The status quo here is that we either have no index entry, or an index entry that points to the start of the synopsis. Neither option is good.)

zygoloid:
> ```
> struct X {
>   @\begin{libmember}{f}{X}@
>   int f();
>   @\end{libmember}@
> };
> ```

[...]

PR #6006, opened 2022-11-27
[utilities] Consistently index all member typedefs in the utilites he…

AlisdairM:
…ader

Audited the file `utilities.tex` for indexing of member typedefs.  Several idioms were in use, and not all typedef-names were indexed.  Adopted what seemed to be the preferred approach of using the `@\libmember{}@` style of definition within the class definition itself, so that omissions become obvious if more typedef-names are added in the future.

Please provide feedback on whether indexing the explicit template specialization for <void> types is the correct approach.  It pulls out the name that is not present for the primary template, and then the ranges library adds some non-template classes with the same name.

Noted that `char_type` for `basic_format_arg` is a private member marked as exposition only, so should really be `char-type` with the expos-id font.  After checking with the original paper, will supply a separate follow-up ticket for that --- this comment is only to confirm that the omission from the index in this patch is intentional.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -123,14 +123,14 @@ namespace std {
     requires requires { typename pair<common_reference_t<TQual<T1>, UQual<U1>>,
                                       common_reference_t<TQual<T2>, UQual<U2>>>; }
   struct basic_common_reference<pair<T1, T2>, pair<U1, U2>, TQual, UQual> {
-    using type = pair<common_reference_t<TQual<T1>, UQual<U1>>,
+    using @\libmember{type}{basic_common_reference}@ = pair<common_reference_t<TQual<T1>, UQual<U1>>,
                       common_reference_t<TQual<T2>, UQual<U2>>>;
   };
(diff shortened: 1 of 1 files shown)


AlisdairM:
This is the first of a series of patches that will cover all the standard library headers.  I will await review of this first patch before producing the rest though.

After filing the patch I became aware of #3911, which this patch would be seen as a small part of addressing.

AlisdairM:
I believe this branch has correctly been rebased, so we can lose that tag.
I am yet to review whether it is still complete, given ongoing additions to the library.

source/utilities.tex
```diff
@@ -4050,8 +4050,8 @@
 \indexlibrarymember{iterator}{optional}%
 \indexlibrarymember{const_iterator}{optional}%
```
jensmaurer, line 4051: Please review globally for now-redundant indexing like this.

jensmaurer (changes requested):

PR #6008, opened 2022-11-28
Add the SI prefixes to the index of library names

marcmutz:


jwakely (approved):

source/meta.tex
```diff
@@ -2626,6 +2646,10 @@
 \rSec2[ratio.si]{SI types for \tcode{ratio}}
 
 \pnum
+\indexlibraryglobal{yocto}%
+\indexlibraryglobal{zepto}%
+\indexlibraryglobal{zetto}%
```
JohelEGP, line 2651:
```suggestion
\indexlibraryglobal{zetta}%
```

JohelEGP (changes requested):

JohelEGP (approved):

tkoeppe:
I don't think this is editorial?!

marcmutz:
> How about using `\libglobal`?

Good idea. I have no idea what it's doing, but I see how it's used for ratio_equal_v etc, and I like how it's inline that way. Will update.

JohelEGP:
It's not mentioned at https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#indexes, but it links to the page of the declaration (or declaration itself on eel.is), rather than the beginning of the synopsis. It also avoids spelling errors.

JohelEGP (approved):

jensmaurer (approved):

PR #6016, opened 2022-12-13
[time.zone.leap.members] update note about leap seconds

jwakely:
This note about leap seconds is still true, but should be updated to be more precise. Saying something about 2019 in C++23 seems odd, and will lead readers to wonder if a negative leap seconds was inserted in 2020.

There has been no leap second inserted since 2016, and none will be inserted before 2023-06-28 at the earliest.

https://en.wikipedia.org/wiki/Leap_second

### DIFF
--- a/source/time.tex
+++ b/source/time.tex
@@ -10194,7 +10194,7 @@ constexpr seconds value() const noexcept;
 \tcode{+1s} to indicate a positive leap second or
 \tcode{-1s} to indicate a negative leap second.
 \begin{note}
-All leap seconds inserted up through 2019 were positive leap seconds.
+All leap seconds inserted up through 2022 were positive leap seconds.
 \end{note}
 \end{itemdescr}
 


jensmaurer (approved):

jensmaurer:
Somehow I feel we should change the note so that it doesn't become invalid for every standards release. But the proposed adjustment is better than no change at all.

jwakely:
It looks like the next leap second might be negative, so we would need to either remove or change it to a definite statement about the past. But there might be three standard releases before that actually happens.

tkoeppe:
I thought leap seconds are now abolished?

jwakely:
There is agreement to stop adding them after 2035, but that still has to be ratified by the ITU. Even if the ITU agrees, there could be more leap seconds added before 2035.

https://www.nature.com/articles/d41586-022-03783-5

PR #6018, opened 2022-12-15
[mdspan.mdspan.overview] Add default template parameters for mdspan

hewillk:
to make it consistent with the declaration.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -20142,7 +20142,7 @@ Equivalent to: \tcode{return p + i;}
 
 \begin{codeblock}
 namespace std {
-  template<class ElementType, class Extents, class LayoutPolicy = layout_right, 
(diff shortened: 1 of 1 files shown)


jensmaurer:
Repeated default template parameters are ill-formed.

hewillk:
> Repeated default template parameters are ill-formed.

See https://github.com/cplusplus/draft/pull/4836.

jwakely:
Identifiers with _`kebab-case`_ are ill-formed too, but we do that to make things easier to read :slightly_smiling_face: 

Nobody expects to be able to just copy & paste classes from the standard and use them as-is, so we can repeat the defaults if that makes the specification more clear. And I think it does, because often the synopsis is huge and defaults buried there are not obvious.

jensmaurer (approved):
Ok, looks good.

PR #6025, opened 2022-12-16
Replace "possibly \tcode{const}" with "possibly const"

### LINKED ISSUE #117: "possibly const" vs "possibly \tcode{const}" vs "possibly \keyword{const}"
"possibly const" appears five times in the current draft. "possibly \tcode{const}" appears ten times. Clean this up, make a rule for it.

Update added by @burblebee on 11/15/22:
We've since added several more of the above variations as well as the addition of  "possibly \keyword{const}".

tkoeppe:
The "const" here is not syntax, but just normal text. This is similar to "inline" and "public", which have previously been cleaned up similarly.

Fixes #117.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -158,7 +158,7 @@ the type \tcode{decltype(pred(*first))} shall model
 \exposconcept{boolean-testable}\iref{concept.booleantestable}.
 The function object \tcode{pred} shall not apply any non-constant function
 through its argument.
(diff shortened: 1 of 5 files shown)


tkoeppe:
@jwakely, @jensmaurer, @zygoloid, @sdutoit ^^^

jensmaurer (approved):
Consistency is good.

PR #6026, opened 2022-12-16
Isoclauses

tkoeppe:


CaseyCarter:
Is an "isoclause" a clause with all line lengths equal?

tkoeppe:
Haha, if only! I pushed the wrong thing for the wrong document here, nothing to see :-)

PR #6031, opened 2022-12-18
[lex.ccon] Add a grouping brace to clarify table cell

### LINKED ISSUE #4394: P2029R4 ("character escapes") seemingly attempts to introduce new terms
The newly added table lex.ccon.literal from #4371/P2029R4 seems to attempt to _define_ the terms "ordinary characater literal", "wide character literal", "UTF-8 character literal", etc., but the presentation makes this unclear. Usually, definitions appear in plain body text where the context and style makes it clear that a new term is being defined, but the table stands apart from the main text, and so we can no longer rely on this convention.

It seems useful to define the terms in the main text, e.g. by adding a sentence to the new paragraph 2:

"The kind of a character literal [...] <del>as defined by Table Y</del><ins>as described by Table Y. The following kinds are defined: an _ordinary character literal_ is ...".

tkoeppe:
Fixes #4394.

image

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -52,7 +52,7 @@ jobs:
           eval "$(/usr/libexec/path_helper)"
           echo "PATH=${PATH}" >> ${GITHUB_ENV}
           sudo tlmgr update --self
-          sudo tlmgr install latexmk isodate substr relsize ulem fixme rsfs environ layouts enumitem l3packages l3kernel imakeidx splitindex xstring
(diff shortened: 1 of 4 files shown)


jensmaurer (approved):
Cool.

tkoeppe:
We'd need to add a new dependency to make the `multirow` package available.

tkoeppe:
image


I'm not sure if this is still valuable now that there are only two rows.

PR #6032, opened 2022-12-19
[range.istream] Start `CharT` and `Traits` with lowercase

hewillk:
I noticed that whether it is 
- `basic_istream`/`basic_ostream` in `<iosfwd>`, 
- or `istream_iterator`/`ostream_iterator` in `<iterator>`, 
- or `string`/`string_view` in `<string>`, 
- or `formatter` in `<format>`, 

their template parameters `charT`/`traits` all start with lowercase.

We should make `istream_view` consistent with such a style.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -219,7 +219,7 @@ namespace std::ranges {
   namespace views { inline constexpr @\unspecnc@ repeat = @\unspecnc@; }            // freestanding
 
   // \ref{range.istream}, istream view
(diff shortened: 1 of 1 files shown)


JohelEGP (approved):

tkoeppe:
@jwakely Does this fit into a general policy?

jwakely:
It's consistent with our style for everything using `char_traits` since C++98, but I hate it. All other template parameters start with an uppercase letter, this is just a historical oddity IMHO.

jwakely:
Maybe ... but I should probably run that past the group.

PR #6035, opened 2022-12-19
[class.compare] Don't introduce `V` for the return value

brevzin:
In both cases, I'm not sure introducing `V` helps much - just requires name lookup for `V`. If we just say "the return value" in every case, I think that's clearer.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -6517,8 +6517,8 @@ in the expanded lists of subobjects for \tcode{x} and \tcode{y}
 until the first index $i$
 where $\tcode{x}_i\tcode{ == }\tcode{y}_i$ yields a result value which,
 when contextually converted to \tcode{bool}, yields \tcode{false}.
-If no such index exists, the return value is \tcode{true}.
-Otherwise, the return value is \tcode{false}.
+The return value is \tcode{true} if such an index exists
+and \tcode{false} otherwise.
 
 \pnum
 \begin{example}
@@ -6627,10 +6627,10 @@ until the first index $i$ where
 the synthesized three-way comparison of type \tcode{R}
 between $\tcode{x}_i$ and $\tcode{y}_i$
 yields a result value $\tcode{v}_i$ where $\tcode{v}_i \mathrel{\tcode{!=}} 0$,
-contextually converted to \tcode{bool}, yields \tcode{true};
-the return value is a copy of $\tcode{v}_i$.
-If no such index exists, the return value is
-\tcode{static_cast<R>(std::strong_ordering::equal)}.
+contextually converted to \tcode{bool}, yields \tcode{true}.
+The return value is a copy of $\tcode{v}_i$
+if such an index exists and
+\tcode{static_cast<R>(std::strong_ordering::equal)} otherwise.
 
 \pnum
 The \defn{common comparison type} \tcode{U}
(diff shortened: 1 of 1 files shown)


jensmaurer (approved):

leni536:
See also #6036 for some additional considerations for improving the wording here.

jensmaurer:
This is a clear improvement, regardless of where #6036 goes.

tkoeppe:
I'm not sure it's a "clear improvement": both versions seem equally reasonable, and one could make the case that a repeated concept should be named. I don't find this a super strong case, but am happy to hear others' position. Thoughts?

brevzin:
> one could make the case that a repeated concept should be named

One could, I guess, but the repeated concept in this case is only three words ("the return value"), is a very simple concept, and it is repeated only one time.

source/classes.tex
```diff
@@ -6509,16 +6509,16 @@
...
 with parameters \tcode{x} and \tcode{y} is determined
 by comparing corresponding elements $\tcode{x}_i$ and $\tcode{y}_i$
 in the expanded lists of subobjects for \tcode{x} and \tcode{y}
 (in increasing index order)
 until the first index $i$
 where $\tcode{x}_i\tcode{ == }\tcode{y}_i$ yields a result value which,
 when contextually converted to \tcode{bool}, yields \tcode{false}.
-If no such index exists, \tcode{V} is \tcode{true}.
-Otherwise, \tcode{V} is \tcode{false}.
+If no such index exists, the return value is \tcode{true}.
```
tkoeppe, line 6520: Could we say "The return value is `true` if no such index exists, and `false` otherwise."? Then we wouldn't have to repeat "return value".

source/classes.tex
```diff
@@ -6628,8 +6628,8 @@
 between $\tcode{x}_i$ and $\tcode{y}_i$
 yields a result value $\tcode{v}_i$ where $\tcode{v}_i \mathrel{\tcode{!=}} 0$,
 contextually converted to \tcode{bool}, yields \tcode{true};
-\tcode{V} is a copy of $\tcode{v}_i$.
-If no such index exists, \tcode{V} is
+the return value is a copy of $\tcode{v}_i$.
```
tkoeppe, line 6631: Perhaps here too we can replace the semicolon with a period (the sentence is already long enough), and then rephrase the conclusion to say "return value" only once?

tkoeppe (changes requested):

source/classes.tex
tkoeppe: @brevzin ?

tkoeppe: @brevzin Ping :-)

brevzin: I'm not sure what you mean. Rephrase how?

brevzin: Oh, there's more words after this comment lol. So something like:

>The return value is a copy of $\tcode{v}_i$ is such an index exists, otherwise \tcode{static_cast<R>(std::strong_ordering::equal)}.

?

tkoeppe: Yes, or "The return value is a copy of v_i if such an index exists, and static_cast... otherwise." See comment above.

tkoeppe (approved):
Fine.

PR #6041, opened 2022-12-30
Fix an example that got broken by P2499R0

### LINKED ISSUE #6017: [range.split] The example does not work after P2499R0
https://eel.is/c++draft/range.split.overview#3 has this example:

```cpp
string str{"the quick brown fox"};
for (string_view word : views::split(str, ' ')) {
  cout << word << '*';
}
// The above prints the*quick*brown*fox*
```

This implicitly constructs a `std::string_view` out of each subrange obtained by `split`. This does not compile any more because the corresponding constructor has been made `explicit` by P2499R0, and `split` doesn't seem to have the same shortcuts that other views have (e.g. `take` out of `string_view` yields another `string_view`).

Possible solution: make the construction explicit:

```cpp
string str{"the quick brown fox"};
for (const auto& word : views::split(str, ' ')) {
  cout << string_view(word) << '*';
}
// The above prints the*quick*brown*fox*
```

dangelog:
After P2499R0 it is no longer possible to implicitly construct a string_view out of a range (like the ones produced by views::split). Use explicit construction instead.

Fixes #6017


source/ranges.tex
```diff
@@ -8018,8 +8018,8 @@
 \begin{example}
 \begin{codeblock}
 string str{"the quick brown fox"};
-for (string_view word : views::split(str, ' ')) {
-  cout << word << '*';
+for (const auto& word : views::split(str, ' ')) {
```
hewillk, line 8021: It should be simpler to use `auto`.

dangelog: Hi,
Sorry, what do you mean? I'm using `auto`. Do you mean avoiding the string_view conversion entirely on the next line?

JohelEGP: They mean just `auto`, not `const& auto`.

dangelog: Oh, alright, makes sense. Will amend.

dangelog:
Do I need to do some further change?

CaseyCarter (approved):

tkoeppe:
Can you please update the commit message according to https://github.com/cplusplus/draft/wiki/Commit-message-format#editorial-commits?

jensmaurer:
@tkoeppe, looks like this is ready now.

PR #6046, opened 2023-01-05
Add extra notes to `ranges::c*` to mirror `ranges::*`

strega-nil-ms:
Given the new changes to `cbegin` and friends from [P2278R4][], where they are no longer simple wrappers around `begin` and friends, it is (in my opinion) confusing that they lack a similar note about how "ill-formed means substitution failure".

[P2278R4]: https://wg21.link/p2278r4

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -753,6 +753,13 @@ let \tcode{U} be \tcode{ranges::begin(\exposid{possibly-const-range}(t))}.
 
 \pnum
 \begin{note}
+Diagnosable ill-formed cases above
+result in substitution failure when \tcode{ranges::cbegin(E)}
+appears in the immediate context of a template instantiation.
+\end{note}
+
+\pnum
+\begin{note}
 Whenever \tcode{ranges::cbegin(E)} is a valid expression, its type models
 \libconcept{input_or_output_iterator} and \exposconcept{constant-iterator}.
 \end{note}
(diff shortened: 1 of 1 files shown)


CaseyCarter (approved):

tkoeppe:
What about `ranges::ssize`?

Is the reason that we don't have notes here that the items in question are "expression-equivalent" to one of the items with notes, and those those notes apply transitively?

CaseyCarter:
> Is the reason that we don't have notes here that the items in question are "expression-equivalent" to one of the items with notes, and those notes apply transitively?

Roughly. The notes (which should maybe be normative) are to clarify the cases where the wording explicitly says "`cpo_expression(meow)` is ill-formed". It's "clear enough" that "`cpo_expression(meow)` is expression-equivalent to `other_expression`" implies that `cpo_expression(meow)` is ill-formed in the immediate context of a template instantiation when and if `other_expression` is, but e.g. "If `T` is an array of unknown bound, `ranges::empty(E)` is ill-formed." could mean any kind of ill-formed.

CaseyCarter:
I suppose by my argument, "If `ranges::size(t)` is ill-formed, `ranges::ssize(E)` is ill-formed." requires a note.

tkoeppe:
Do the notes describe a consequence of existing specification, or state a novel requirement?

jwakely:
> Do the notes describe a consequence of existing specification, or state a novel requirement?

That has been the subject of some debate.

PR #6056, opened 2023-01-22
[reverse.iter.conv] Remove unclear explicit comments

hewillk:
I don't know what explicit refers to here, it seems to be more appropriate to remove. People who disagree with me are also welcome.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -3358,7 +3358,7 @@ Assigns \tcode{u.current} to \tcode{current}.
 
 \indexlibrarymember{base}{reverse_iterator}%
 \begin{itemdecl}
-constexpr Iterator base() const;          // explicit
+constexpr Iterator base() const;
 \end{itemdecl}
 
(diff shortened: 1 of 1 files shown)


CaseyCarter (approved):
This comment has been present since C++98, which had a similar comment on `basic_string_view::c_str`. I guess these comments are trying to tell us that the pertinent member functions are like explicit conversions? If we don't know what they mean today, they provide no value; I'm happy to remove them.

jensmaurer:
@jwakely, any insight here? Otherwise, this looks like good riddance to me.

jwakely (approved):
I think Casey's right, it's saying you have to call a function to convert to the wrapped type, there's no implicit conversion. Let's drop it.

CaseyCarter:
> I think Casey's right, it's saying you have to call a function to convert to the wrapped type, there's no implicit conversion. Let's drop it.

Alternatively, we could change them to `// inscrutable` ;)

PR #6058, opened 2023-01-24
Add cross-references between [basic.life],

ecatmur:
[class.cdtor] and [except.handle].

To resolve confusion noted in
https://github.com/cplusplus/CWG/issues/204

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3451,6 +3451,9 @@ program has undefined behavior if:
   the pointer is used as the operand of a
   \keyword{dynamic_cast}\iref{expr.dynamic.cast}.
 \end{itemize}
+\begin{note}
+For an object with a non-trivial constructor or destructor, see also~\ref{class.cdtor}.
+\end{note}
 \begin{example}
 \begin{codeblock}
 #include <cstdlib>
@@ -3501,6 +3504,9 @@ well-defined. The program has undefined behavior if:
 \keyword{dynamic_cast}\iref{expr.dynamic.cast} or as the operand of
 \keyword{typeid}.
 \end{itemize}
+\begin{note}
+For an object with a non-trivial constructor or destructor, see also~\ref{class.cdtor}.
+\end{note}
 
 \pnum
 If, after the lifetime of an object has ended and before the storage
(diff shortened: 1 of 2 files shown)


AlisdairM:
Now that pseudo-destructors also end the lifetime of scalar objects, is there a latent Core issue lurking here?

ecatmur:
> Now that pseudo-destructors also end the lifetime of scalar objects, is there a latent Core issue lurking here?

I don't feel there is. A scalar pseudo-destructor is necessarily trivial, so [class.cdtor] isn't needed, and [basic.life] is written in terms of destruction of the object, which is what a pseudo-destructor does https://eel.is/c++draft/expr.call#5.sentence-2 . Am I missing something?

tkoeppe:
These references may be helpful. @jensmaurer do you think there could be a more concise way of placing those references?

source/basic.tex
```diff
@@ -3451,6 +3451,9 @@
   the pointer is used as the operand of a
   \keyword{dynamic_cast}\iref{expr.dynamic.cast}.
 \end{itemize}
+\begin{note}
+For an object with a non-trivial constructor or destructor, see also~\ref{class.cdtor}.
+\end{note}
```
jensmaurer, line 3456: We already have this cross-reference in line 3425:

> For an object under construction or destruction, see~\ref{class.cdtor}.

In general, [class.cdtor] describes behavior for objects under construction or destruction, and [basic.life] describes behavior for objects that are (entirely) out-of-lifetime.

source/basic.tex
```diff
@@ -3501,6 +3504,9 @@
 \keyword{dynamic_cast}\iref{expr.dynamic.cast} or as the operand of
 \keyword{typeid}.
 \end{itemize}
+\begin{note}
+For an object with a non-trivial constructor or destructor, see also~\ref{class.cdtor}.
+\end{note}
```
jensmaurer, line 3509: Same here, see line 3492.

source/basic.tex
jensmaurer: And yes, "object" might be a subobject.

jensmaurer:
@tkoeppe, I'm not happy about the verbose notes, and I'm not sure they clarify anything.

If there is some problem with the wording (e.g. because [class.cdtor] wants to be in [basic.life]), we should entertain that as a CWG issue with a specific suggestion for a change.

tkoeppe:
@ecatmur Could you please file a CWG issue if there is a need for clarification?

PR #6060, opened 2023-01-24
[class.mem.general] Clarify ambiguous 'and'

jensmaurer:
Fixes cplusplus/CWG#229

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -605,11 +605,10 @@ the \grammarterm{member-specification} of the enclosing class.
 \end{note}
 
 \pnum
-A class is regarded as complete
-where its definition is reachable and
-within its complete-class contexts;
-otherwise it is regarded as incomplete within its own class
-\grammarterm{member-specification}.
+A class \tcode{C} is complete at a program point $P$
+if the definition of \tcode{C} is reachable from $P$\iref{module.reach}
+or if $P$ is in a complete-class context of \tcode{C}.
+Otherwise, \tcode{C} is incomplete at $P$.
 
 \pnum
 In a \grammarterm{member-declarator},


tkoeppe:
I'm not sure this is any clearer than before. THe linked issue suggests "or"; would that work? Or otherwise a larger rewrite perhaps?

jensmaurer:
Mike Miller says "or" doesn't read well, and he suggested (among other options such as "and also" and "in addition to") the words chosen here. I think "as well as" is less suggestive of and-connected conditions.

tkoeppe:
But it still admits the same erroneous interpretation as the original wording, doesn't it?

How about: "A class is regarded as complete with its complete-class contexts and also wherever its definition is reachable."?

jensmaurer:
Sounds good. Please check the force-pushed diff.

source/classes.tex
```diff
@@ -606,8 +606,8 @@
 
 \pnum
 A class is regarded as complete
-where its definition is reachable and
-within its complete-class contexts;
+within its complete-class contexts and also
+wherever its definition is reachable;
 otherwise it is regarded as incomplete within its own class
 \grammarterm{member-specification}.
```
zygoloid, line 612: It's a little strange for the "otherwise" case to not cover all cases not covered by the previous rule. This reordering makes it stranger by moving the complete-class context part away from the other member-secification pieces.

How about something like:

> A class is complete where its definition is reachable, and in complete-class contexts within its *member-specification*. Otherwise it is incomplete, including in its *member-specification*.

jensmaurer: Yeah, sounds better. Updated.

@tkoeppe?

tkoeppe: Very good, thank you!

xmh0511: I think a constructive rule is better:
> A class is complete:
>> - at the point where its definition is reachable, or
>> - in complete-class contexts within its member-specification
>
> Otherwise it is incomplete, including in other parts of its member-specification.

tkoeppe: Why do you think that's better than the previous suggestion? The sentence seems short enough to not require a list, and also I don't think the "or" is right: a class is really complete in both cases, not just in "at least one of these cases". "Or" would be appropriate if we said, "In order to be complete, a class has to [either X] or [Y]", but as written, "and" seems more appropriate.

jensmaurer: Mike M. says that "complete-class context" is already applying to "within its member-specification" per the definition, so the extra normative mention of "within its member-specification" is superfluous.

However, this is now deviating from the existing words quite a bit, so I'm wondering whether we want a brief trip through CWG.

xmh0511: I think the "within the member-specification" in otherwise case can be removed. We have completely defined what complete-class context is:
> A complete-class context of a class (template) is a 
>> - [...]
>>
>  **within the member-specification of the class or class template**.

This implies, e.g., that the function body of the member function **outside of the member-specification** is not a complete-class context. Hence, 
> the class is considered to be complete either:
>> - at the point where its definition is reachable, or
>> - in complete-class contexts within its member-specification
>>
> Otherwise, it is incomplete.  

I think this is clear.

JohelEGP: > a class is really complete in both cases, not just in "at least one of these cases". "Or" would be appropriate if we said, "In order to be complete, a class has to [either X] or [Y]", but as written, "and" seems more appropriate.

> For `and` in the updated wording, I still think "and" can result in ambiguity. How should we parse that wording?

Mind you, the current wording uses "and also", which is correct.

xmh0511: Where does "and also" appear in the current wording? The current wording is just ambiguous since its "and".

JohelEGP: Oh, I was looking at the diff of this PR, which added "also".

tkoeppe:
@jensmaurer Do you still want that trip through CWG, or is this ready to be merged?

jensmaurer:
This needs more massaging, in particular "within its \grammarterm{member-specification}" is redundant with the definition of "complete-class context". I feel the "otherwise" part then falls apart.

jensmaurer:
I now have a formulation that super-clearly picks up the defined terms (e.g. "reachable from P") and thus should be an improvement. We're now using "if", so "or" is appropriate, too, and the "otherwise" reads better.

@tkoeppe @zygoloid @JohelEGP opinions welcome.

tkoeppe (approved):

jensmaurer:
CWG telecon 2023-03-30: Approved.

PR #6061, opened 2023-01-24
[basic.life] Move definition of "before" and "after" from last paragraph

AlisdairM:
The last paragraph of this subclause changes the definition of English words used throughout the preceding paragraphs.  While it might be preferable to replace all such usage with the new definitions, that would be a Core issue, see paragraph 6 for an example of awkward usage.  Hence, we move the redefinition to the start of the subclause so we know how to read this text from the start.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3347,6 +3347,15 @@ Some functions in the \Cpp{} standard library implicitly create objects%
 \rSec2[basic.life]{Lifetime}
 
 \pnum
+In this subclause, ``before'' and ``after'' refer to the ``happens before''
+relation\iref{intro.multithread}.
+\begin{note}
+Therefore, undefined behavior results if an object that is being constructed in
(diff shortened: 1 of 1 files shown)


tkoeppe:
But if we move ths wording up, the note is a bit premature.

@jensmaurer, thoughts? It is ineed weird to only define the terms "before" and "after" at the very end.

AlisdairM:
Rebase and force-pushed.
I also moved the now-misplaced comment to a more appropriate paragraph.

AlisdairM:
How do I remove the "needs rebase" tag?

tkoeppe:
You don't -- just let us know that you've rebased.

AlisdairM:
Rebased onto main again, and fixed an accidentally moved `\indextext` that delimited the end of a section.  Force-pushed.

jensmaurer (approved):

tkoeppe:
OK, never mind, this is fine.

PR #6070, opened 2023-02-02
[intro, front] Special page header before Clause 1

tkoeppe:
ISO House Style requires a special header on page 1; the two-ruled format follows the rice model (and is what I'm doing for the LFTS).

image

The published wording will say "INTERNATIONAL STANDARD" etc.

### DIFF
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -1,6 +1,17 @@
 %!TEX root = std.tex
 
+\clearpage
+\bigskip\noindent\textlarger{\textbf{Programming languages --- \Cpp{}}}
+\bigskip\bigskip
+
+\begingroup
+\let\clearpage\relax
 \rSec0[intro.scope]{Scope}
+\endgroup
+\copypagestyle{cpppageone}{cpppage}
+\makeoddhead{cpppageone}{\textbf{WORKING DRAFT}}{}{\leaders\hrule height 2pt\hfill\kern0pt\\\textbf{\docno}}
+\makeheadrule{cpppageone}{\textwidth}{2pt}
+\thispagestyle{cpppageone}
 
 \pnum
 \indextext{scope|(}%
--- a/source/layout.tex
+++ b/source/layout.tex
@@ -10,7 +10,7 @@
 %%--------------------------------------------------
 %%  set header and footer positions and sizes
 
-\setheadfoot{\onelineskip}{4\onelineskip}
+\setheadfoot{3\onelineskip}{4\onelineskip}
 \setheaderspaces{*}{2\onelineskip}{*}
 
 %%--------------------------------------------------


source/intro.tex
```diff
@@ -1,6 +1,15 @@
 %!TEX root = std.tex
 
+\clearpage
+\bigskip\noindent\textlarger{\textbf{%
+Information Technology —\\[1.5ex]Programming languages, their environments and system software interfaces —\\[1.5ex]\Cpp{}}}
+\bigskip\bigskip
+
+\begingroup
+\let\clearpage\relax
```
tkoeppe, line 9: @Eelis I hope this won't be too annoying. It's not so easy to prevent the page break before a chapter start, so this quick hack gets the job done.

Eelis: No worries, I'll manage. Cheers!

source/styles.tex
```diff
@@ -54,6 +54,10 @@
 
 \aliaspagestyle{chapter}{cpppage}
 
+\copypagestyle{cpppageone}{cpppage}
```
jensmaurer, line 57: I understand we need to nix the existing "cpppage" style for the first page, but do we have to put the contents into a header to start with?  Or can we just use \vspace{-1cm} and write the first-page-only stuff in the main body of the text? For once, that would put all the special contents in one place as opposed to the present solution, which splits stuff between styles.tex and intro.tex.

source/layout.tex
```diff
@@ -10,7 +10,7 @@
 %%--------------------------------------------------
 %%  set header and footer positions and sizes
 
-\setheadfoot{\onelineskip}{2\onelineskip}
+\setheadfoot{3\onelineskip}{2\onelineskip}
```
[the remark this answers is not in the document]
tkoeppe: No, there's no visual diff on any other page. The actual page geometry seems to be computed by memoir based on various constraints, and it seems that we have always had leeway here.

(We need to increase the headheight in order to accommodate the special header.)

source/styles.tex
tkoeppe: The upside is that currently the special header is perfectly aligned with the normal headers. And the document title isn't really a "header". It's just a bit awkward that it has to be on the same page as Clause 1. I'm not sure if the alternaives would be any better.

We could certainly move the special logic into a macro, so that the actual intro.tex content doesn't contain it.

jensmaurer: Not an additional macro = additional indirection.

I was just asking to establish a direct relationship between the placement in the .tex source file and the output, given that we're talking about an "appears once" situation. Can we maybe redefine the page style on that very page, as opposed to elsewhere?

tkoeppe: Yes, we can move it to intro.tex, sure!

source/intro.tex
tkoeppe: Thanks!

tkoeppe:
@jensmaurer PTAL?

tkoeppe:
No visual changes apart from that page 1.

jensmaurer:
We're just called "Programming languages --- C++" these days.

https://www.iso.org/standard/83626.html

tkoeppe:
You're right, fixed!

PR #6073, opened 2023-02-08
[func.require] Use math font for non-code conditions

jwakely:
N is not a C++ entity, it refers to the N suffix of tN, and the equality is mathematical not a C++ operator.

If this was an actual C++ expression using a variable `N` we would need to say "`N == 1` is `true`" and that would be much worse :sweat_smile:


jwakely:
image

tkoeppe:
Looks reasonable, thanks :-)

jwakely:
This was suggested as a drive-by fix while reviewing [LWG 3655](https://wg21.link/lwg3655) but I said I'd do it this way instead.

CaseyCarter (approved):
Huge improvement now that we don't use both `\tcode{N}` and `$N$` to mean the same thing.

PR #6077, opened 2023-02-10
[util.smartptr.atomic.general] Add std:: qualified for move in example

hewillk:


source/threads.tex
```diff
@@ -4833,7 +4833,7 @@
     while (p && p->t != t)
       p = p->next;
 
-    return shared_ptr<node>(move(p));
+    return shared_ptr<node>(std::move(p));
```
CaseyCarter, line 4836: Why are we explicitly converting a `shared_ptr<node>` to `shared_ptr<node>` in the first place? What's wrong with `return p;`?

hewillk: `return p;` is a more reasonable approach. (I'd be happy if it could be changed to this)

CaseyCarter: Yeah, let's go ahead and change it. We like to show a variety of styles in examples, but writing something long and complicated when something short and simple will do isn't style, it's obfuscation.

CaseyCarter (approved):

jensmaurer (approved):
Looks good.

jwakely (commented):
You now need to look outside that member function to see what its return type is, because it uses `auto` for the return type and `auto` for the declaration of `p`, and now this removes the only mention of the type. While it's not hard to look at the declaration of `head`, is this actually making the example clearer?
We could change the return type from `auto` to `shared_ptr<node>` maybe.

PR #6114, opened 2023-02-14
[Motions 2023 02 lwg 13] P2679R2 Fixing std::start_lifetime_as and std::start_lifetime_as_array

### LINKED ISSUE #6103: [2023-02 LWG Motion 13] P2679R2 Fixing std::start_lifetime_as for arrays

RobertLeahy:
Fixes #6103.
Fixes cplusplus/nbballot#459.
Fixes cplusplus/nbballot#508.
Fixes cplusplus/papers#1345.

### DIFF
--- a/source/memory.tex
+++ b/source/memory.tex
@@ -881,7 +881,8 @@ template<class T>
 \begin{itemdescr}
 \pnum
 \mandates
-\tcode{T} is an implicit-lifetime type\iref{basic.types.general}.
+\tcode{T} is an implicit-lifetime type\iref{basic.types.general}
+and not an incomplete type\iref{term.incomplete.type}.
 
 \pnum
 \expects
@@ -931,14 +932,33 @@ template<class T>
 
 \begin{itemdescr}
 \pnum
+\mandates
+\tcode{T} is a complete type.
+
+\pnum
 \expects
-\tcode{n > 0} is \tcode{true}.
+\tcode{p} is suitably aligned for an array of \tcode{T} or is null.
+\tcode{n <= size_t(-1) / sizeof(T)} is \tcode{true}.
+If \tcode{n > 0} is \tcode{true},
+\range{(char*)p}{(char*)p + (n * sizeof(T))} denotes
+a region of allocated storage that is
+a subset of the region of storage
+reachable through\iref{basic.compound} \tcode{p}.
 
 \pnum
 \effects
-Equivalent to:
-\tcode{return *start_lifetime_as<U>(p);}
+If \tcode{n > 0} is \tcode{true},
+equivalent to
+\tcode{start_lifetime_as<U>(p)}
 where \tcode{U} is the type ``array of \tcode{n} \tcode{T}''.
+Otherwise, there are no effects.
+
+\pnum
+\returns
+A pointer to the first element of the created array,
+if any;
+otherwise,
+a pointer that compares equal to \tcode{p}\iref{expr.eq}.
 \end{itemdescr}
 
 \rSec2[allocator.tag]{Allocator argument tag}


frederick-vs-ja:
This also fixes cplusplus/papers#1345.

tkoeppe:
Thank you!

source/memory.tex
```diff
@@ -927,15 +928,33 @@
 \end{itemdecl}
 
 \begin{itemdescr}
+\pnum
+\mandates
+\tcode{T} is a complete type.
```
jwakely, line 933: I know this is what the approved paper says, but is there any reason that the non-array form requires "not an incomplete type" and the array form requires "is a complete type"? It seems like they could both say it the same way.

source/memory.tex
```diff
@@ -927,15 +928,33 @@
...
+
 \pnum
 \expects
-\tcode{n > 0} is \tcode{true}.
+\tcode{p} is suitably aligned for an array of \tcode{T} or is null.
+\tcode{n <= size_t(-1) / sizeof(T)} is \tcode{true}.
+\range{(char*)p}{(char*)p + (n * sizeof(T))} denotes
```
jwakely, line 939: This seems to be missing "If `n > 0` is `true`"

```suggestion
If `n > 0` is `true` \range{(char*)p}{(char*)p + (n * sizeof(T))} denotes
```

jwakely (changes requested):

source/memory.tex
RobertLeahy: Doing a quick search through the draft I see both forms used elsewhere in the standard. I'm unsure if the distinction was deliberate since I don't precisely remember the details of the completeness discussion in Core in Kona.

Interestingly "T is a complete type" doesn't seem to come with a cross reference whereas "T is not an incomplete type" seems to come with a cross reference to term.incomplete.type which I'll add to this PR in the next revision.

CaseyCarter: We're a bit inconsistent in the library. Core folk will tell you that we don't define the term "complete type", we only define "incomplete type". Library insists that we can use "complete type" with the obvious English-language meaning "a type which is not an incomplete type", so you will sometimes see "complete type" and sometimes "not an incomplete type".

RobertLeahy: Is it therefore fine for this change to be internally inconsistent (as specified in the paper) or can/should we editorially coalesce around "complete type" or "not an incomplete type" consistently?

burblebee:
Hey guys, I signed up for this motion at https://github.com/cplusplus/draft/issues/6103. My commit is in motions-2023-02-lwg-13.

source/memory.tex
```diff
@@ -878,7 +878,8 @@
 \begin{itemdescr}
 \pnum
 \mandates
-\tcode{T} is an implicit-lifetime type\iref{basic.types.general}.
+\tcode{T} is an implicit-lifetime type\iref{basic.types.general}
+and not an incomplete type\iref{term.incomplete.type}.
```
JohelEGP, line 882: According to https://github.com/cplusplus/draft/pull/6114#discussion_r1105990427
> Interestingly "T is a complete type" doesn't seem to come with a cross reference elsewhere whereas "T is not an incomplete type" does seem to (to term.incomplete.type) which I'll add to this PR in the next revision.

but I see no such reference.

source/memory.tex
```diff
@@ -927,15 +928,34 @@
...
-\tcode{return *start_lifetime_as<U>(p);}
+If \tcode{n > 0} is \tcode{true},
+equivalent to
+\tcode{start_lifetime_as<U>(p)}
 where \tcode{U} is the type ``array of \tcode{n} \tcode{T}''.
+Otherwise, there are no effects.
+
+\pnum
+\returns
+A pointer to the first element of the array,
```
JohelEGP, line 955:
```suggestion
A pointer to the first element of the created array,
```

JohelEGP (changes requested):

source/memory.tex
RobertLeahy: I'm not sure if you're pointing out that the reference isn't in the paper, or isn't in the draft created after my pull request. I confirmed it's there after my pull request:

image

JohelEGP: Yes, I'm talking about the paper.

RobertLeahy: I was adopting practice from elsewhere in the draft, especially elsewhere in `memory.tex`.

JohelEGP: That makes sense. Your phrasing of
> "T is not an incomplete type" does seem to

suggested to me that you didn't insert it yourself.

JohelEGP: Feel free to mark this as resolved.

tkoeppe:
@burblebee Very sorry about that, we'll make this clearer next time!

source/memory.tex
tkoeppe: I'd prefer to have LWG come back with some guidance, and we'll take the paper as approved for now. But it's a good question worth settling, @jwakely.

PR #6120, opened 2023-02-15
[CWG 3] P2796R0 CWG2518 Conformance requirements and #error/#warning

### LINKED ISSUE #6085: [2023-02 CWG Motion 3] P2796R0  issue 2518 Conformance requirements and #error/#warning

burblebee:
Fixes #6085.
Fixes cplusplus/papers#1366

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -281,7 +281,7 @@ void f(T t) {
 
 void g(char c) {
   f(0);                                 // OK
-  f(c);                                 // error: must be int-sized
+  f(c);                                 // error on implementations where \tcode{sizeof(int) > 1}: must be \tcode{int}-sized
 }
 \end{codeblock}
 \end{example}
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -257,14 +257,32 @@ the \grammarterm{constant-expression}
 is contextually converted to \keyword{bool} and
 the converted expression shall be a constant expression\iref{expr.const}.
 If the value of the expression when
-so converted is \tcode{true}, the declaration has no
-effect. Otherwise, the program is ill-formed, and the resulting
+so converted is \tcode{true}
+or the expression is evaluated in the context of a template definition,
+the declaration has no
+effect. Otherwise,
+the \grammarterm{static_assert-declaration} fails,
+the program is ill-formed, and the resulting
 diagnostic message\iref{intro.compliance} should include the text of
 the \grammarterm{string-literal}, if one is supplied.
 \begin{example}
 \begin{codeblock}
 static_assert(sizeof(int) == sizeof(void*), "wrong pointer size");
 static_assert(sizeof(int[2]));          // OK, narrowing allowed
+
+template <class T>
+void f(T t) {
+  if constexpr (sizeof(T) == sizeof(int)) {
+    use(t);
+  } else {
+    static_assert(false, "must be int-sized");
+  }
+}
+
+void g(char c) {
+  f(0);                                 // OK
+  f(c);                                 // error: must be int-sized
+}
 \end{codeblock}
 \end{example}
 
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -800,16 +800,16 @@ the data being processed; see \ref{intro.defs} and~\ref{intro.execution}.
 \end{footnote}
 that program.
 \item
+\indextext{behavior!undefined}%
+If a program contains a violation of a rule for which no diagnostic is required,
+this document places no requirement on implementations
+with respect to that program.
+\item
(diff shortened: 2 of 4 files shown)


source/intro.tex
```diff
@@ -792,23 +792,37 @@
...
 \end{note}
+Furthermore, a conforming implementation
+\begin{itemize}
+\item
+shall not accept a preprocessing translation unit containing
+a \tcode{\# error} preprocessing directive\iref{cpp.error},
+\item
+shall issue at least one diagnostic message for
+each \tcode{\# warning} or \tcode{\# error} preprocessing directive
+not following a \tcode{\# error} preprocessing directive in
```
JohelEGP, line 820:
```suggestion
a \tcode{\#error} preprocessing directive\iref{cpp.error},
\item
shall issue at least one diagnostic message for
each \tcode{\#warning} or \tcode{\#error} preprocessing directive
not following a \tcode{\#error} preprocessing directive in
```
Looking at `preprocessor.tex`, a space is only used in a `ncsimplebnf` environment.

source/declarations.tex
```diff
@@ -248,14 +248,33 @@
...
+  if constexpr (sizeof(T) == sizeof(int)) {
+    use(t);
+  } else {
+    static_assert(false, "must be int-sized");
+  }
+}
+
+void g(char c) {
+  f(0);                                 // OK
+  f(c);                                 // error: must be int-sized
```
JohelEGP, line 276: What's the take on `error:` comments that can be OK depending on the parameters of the abstract machine? This particular one is called out by
> [2](https://eel.is/c++draft/intro.abstract#2)
[#](http://github.com/Eelis/draft/tree/7eaad29c229ece4591b805688a9827b25d2149bf/source/intro.tex#L900)
Certain aspects and operations of the abstract machine are described in this document as implementation-defined (for example, sizeof(int))[.](https://eel.is/c++draft/intro.abstract#2.sentence-1)
> -- https://eel.is/c++draft/intro.abstract#2.sentence-1

source/templates.tex
```diff
@@ -4437,7 +4437,9 @@
 The program is ill-formed, no diagnostic required, if:
 \begin{itemize}
 \item
-no valid specialization can be generated for a template
+no valid specialization,
+ignoring \grammarterm{static_assert-declaration} that fail,
```
JohelEGP, line 4441:
```suggestion
ignoring \grammarterm{static_assert-declaration}{s} that fail,
```

JohelEGP (changes requested):

source/intro.tex
jensmaurer: Agreed. Please remove those spaces between # and the directive name.

source/declarations.tex
jensmaurer: Yes, that's a problem with the example, since `sizeof(char) == sizeof(int)` is possible.
Maybe change the comment to read "error on implementations where sizeof(int) > 1: must be int-sized".

JohelEGP: I'm positive there's a precedent here.

JohelEGP (changes requested):

JohelEGP (approved):

source/declarations.tex
tkoeppe: Nice work, thanks!

source/declarations.tex
```diff
@@ -257,14 +257,33 @@
 is contextually converted to \keyword{bool} and
 the converted expression shall be a constant expression\iref{expr.const}.
 If the value of the expression when
-so converted is \tcode{true}, the declaration has no
-effect. Otherwise, the program is ill-formed, and the resulting
+so converted is \tcode{true}
+or the expression is evaluated in the context of a template definition,
+the declaration has no
+effect. Otherwise,
+%FIXME: What does it mean for a grammar term to fail? Also in [temp.res]/p6.
```
tkoeppe, line 264: Isn't this text the _definition_ of what it means for the static_assert-declaration to fail?

@jensmaurer, any thoughts? I think this seems clear?

jensmaurer: Yes, that's the intent: We defined "fails" here.

Note that the proposed resolution in CWG2518 marks "fails" in italics to highlight the fact this is a term definition.

JohelEGP: It's not in italics in this PR. Also, does https://github.com/cplusplus/draft/pull/6120#discussion_r1115069961, which made "int" into `int`, makes sense if the error comment is supposed to be a reproduction of the _string-literal_ to the `static_assert`?

tkoeppe: Oh, good catch, we need to italicise `fails`, thank you!

PR #6121, opened 2023-02-15
[CWG 4] P2796R0 CWG2521 User-defined literals and reserved identifiers

### LINKED ISSUE #6086: [2023-02 CWG Motion 4] P2796R0 issue 2521 User-defined literals and reserved identifiers

burblebee:
[lex.name] Attempt to fix wording; identifiers (plural) can't appear as a single token, and there's no "pp-token"

Fixes #6086.
Fixes cplusplus/papers#1367

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -968,7 +968,9 @@ token as a regular \grammarterm{identifier}.
 \indextext{\idxcode{_}|see{character, underscore}}%
 \indextext{character!underscore!in identifier}%
 \indextext{reserved identifier}%
-In addition, some identifiers are reserved for use by \Cpp{}
+In addition, some identifiers
+appearing as a \grammarterm{token} or \grammarterm{preprocessing-token}
+are reserved for use by \Cpp{}
 implementations and shall
 not be used otherwise; no diagnostic is required.
 \begin{itemize}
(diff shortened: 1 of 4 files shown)


JohelEGP (approved):

source/lex.tex
```diff
@@ -968,7 +968,10 @@
 \indextext{\idxcode{_}|see{character, underscore}}%
 \indextext{character!underscore!in identifier}%
 \indextext{reserved identifier}%
-In addition, some identifiers are reserved for use by \Cpp{}
+In addition, some identifiers
+%FIXME: "appearing" isn't quite right here - reword?
+appearing as \grammarterm{token}{s} or \grammarterm{pp-tokens}
```
tkoeppe, line 973: I'm not sure I'm comfortable with this change from the approved wording. Thoughts? Maybe leave it to an issue?

tkoeppe: That said, the approved "pp-token" isn't actually an existing grammar term. @jensmaurer Are you OK with the change?

jensmaurer: We want to say that the (entire) token or pp-token is an identifier for the "reserved" rule to trigger. Except that pp-token (as in the issue write-up) doesn't exist; it's _preprocessing-token_. So, changing to `\grammarterm{preprocessing-token}{s} would be substantially better. (We mean the plural of _preprocessing-token_, not a sequence of them, which would be _pp-tokens_.

@tkoeppe, can you fix that, please?

tkoeppe: Yes, thanks! Please see what I merged, I think it both fixes this mistake and sticks as close to the approved wording as possible.

Issue #6113, opened 2023-02-13
[dcl.constexpr] awkward grammar in resolution of CWG2602


zygoloid:
CWG2602 asks for this text in [dcl.constexpr]/3:

> A function is constexpr-suitable if:
> - it is not a coroutine (9.5.4 [dcl.fct.def.coroutine]), and
> - if the function is a constructor or destructor, its class does not have any virtual base classes.

The second bullet does fit well into the enclosing context ("A function is constexpr-suitable if, [...] if the function [...]"). This would be easier to read if rephrased:

> A function is constexpr-suitable if:
> - it is not a coroutine (9.5.4 [dcl.fct.def.coroutine]), and
> - <del>if the function</del> <ins>either it</ins> is <ins>not</ins> a constructor or destructor<del>,</del> <ins>or</ins> its class does not have any virtual base classes.

or

> A function is constexpr-suitable if:
> - it is not a coroutine (9.5.4 [dcl.fct.def.coroutine]), and
> - <del>if the function</del> <ins>it</ins> is <ins>not</ins> a constructor or destructor<del>, its</del> <ins>of a</ins> class <del>does not have any</del> <ins>with</ins> virtual base classes.

PR #6124, opened 2023-02-15
[LWG  2] P2789R0 Ready and Tentatively Ready issues

### LINKED ISSUE #6092: [2023-02 LWG Motion 2] P2789R0 C++ Standard Library Issues to be moved in Issaquah, Feb. 2023

burblebee:
Fixes #6092.
Fixes cplusplus/papers#1478
Fixes cplusplus/nbballot#534

### DIFF
--- a/source/locales.tex
+++ b/source/locales.tex
@@ -555,15 +555,19 @@ explicit locale(const string& std_name);
 \begin{itemdescr}
 \pnum
 \effects
-The same as \tcode{locale(std_name.c_str())}.
+Equivalent to \tcode{locale(std_name.c_str())}.
 \end{itemdescr}
 
 \indexlibraryctor{locale}%
 \begin{itemdecl}
-locale(const locale& other, const char* std_name, category);
+locale(const locale& other, const char* std_name, category cats);
 \end{itemdecl}
 
 \begin{itemdescr}
+\pnum
+\expects
+\tcode{cats} is a valid \tcode{category} value\iref{locale.category}.
+
 \pnum
 \effects
 Constructs a locale as a copy of \tcode{other}
@@ -572,7 +576,7 @@ which instead implement the same semantics as \tcode{locale(std_name)}.
 
 \pnum
 \throws
-\tcode{runtime_error} if the argument is not valid, or is null.
+\tcode{runtime_error} if the second argument is not valid, or is null.
 
 \pnum
 \remarks
@@ -581,13 +585,13 @@ The locale has a name if and only if \tcode{other} has a name.
 
 \indexlibraryctor{locale}%
 \begin{itemdecl}
-locale(const locale& other, const string& std_name, category cat);
+locale(const locale& other, const string& std_name, category cats);
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
 \effects
-The same as \tcode{locale(other, std_name.c_str(), cat)}.
+Equivalent to \tcode{locale(other, std_name.c_str(), cats)}.
 \end{itemdescr}
 
 \indexlibraryctor{locale}%
@@ -605,7 +609,9 @@ If \tcode{f} is null, the resulting object is a copy of \tcode{other}.
 
 \pnum
 \remarks
-The resulting locale has no name.
+If \tcode{f} is null,
+the resulting locale has the same name as \tcode{other}.
+Otherwise, the resulting locale has no name.
 \end{itemdescr}
 
 \indexlibraryctor{locale}%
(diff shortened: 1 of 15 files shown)


source/ranges.tex
```diff
@@ -2298,7 +2300,10 @@
 \item
 \tcode{\libconcept{common_range}<R>} is \tcode{true},
 \item
-\tcode{\exposconcept{cpp17-input-iterator}<iterator_t<R>>} is \tcode{true}, and
+if the \grammarterm{qualified-id}
```
JohelEGP, line 2303:
```suggestion
the \grammarterm{qualified-id}
```
Not in the paper. Reads awkward like #6113. Unlike #6113, this item has no condition; rather, this item is a term in a conjunction, so removing 'if' suffices.

source/memory.tex
```diff
@@ -135,6 +135,9 @@
   template<class T, class Alloc, class U, class V>
     constexpr auto uses_allocator_construction_args(const Alloc& alloc,             // freestanding
                                                     const pair<U, V>& pr) noexcept;
+  template<class T, class Alloc, @\exposconcept{pair-like}@ P>
+    constexpr auto uses_allocator_construction_args(const Alloc& alloc,             // freestanding
+                                                    P&& p) noexcept;
   template<class T, class Alloc, class U, class V>
     constexpr auto uses_allocator_construction_args(const Alloc& alloc,             // freestanding
                                                     pair<U, V>&& pr) noexcept;
```
JohelEGP, line 143: The added signature is requested to be inserted before the signature
```C++
  template<class T, class Alloc, class U>
    constexpr auto uses_allocator_construction_args(const Alloc& alloc, U&& u) noexcept;
```

source/ranges.tex
```diff
@@ -10565,6 +10579,10 @@
     adjacent_view() requires @\libconcept{default_initializable}@<V> = default;
     constexpr explicit adjacent_view(V base);
 
+
```
JohelEGP, line 10582:
```suggestion
```

source/ranges.tex
```diff
@@ -12119,6 +12147,29 @@
...
+\begin{itemdecl}
+friend constexpr void iter_swap(const @\exposid{inner-iterator}@& x, const @\exposid{inner-iterator}@& y)
+  noexcept(noexcept(ranges::iter_swap(*x.@\exposid{parent_}@->@\exposid{current_}@, *y.@\exposid{parent_}@->@\exposid{current_}@)))
+  requires @\libconcept{indirectly_swappable}@<iterator_t<V>>;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\effects
+Equivalent to: \tcode{ranges::iter_swap(*x.\exposid{parent_}->\exposid{current_}, *y.\exposid{parent_}->\exposid{current_});}
```
JohelEGP, line 12170:
```suggestion
Equivalent to: \tcode{ranges::iter_swap(*x.\exposid{parent_}->\exposid{current_}, *y.\exposid{parent_}->\exposid{current_})}.
```

JohelEGP (changes requested):

source/locales.tex
```diff
@@ -587,7 +591,7 @@
 \begin{itemdescr}
 \pnum
 \effects
-The same as \tcode{locale(other, std_name.c_str(), cat)}.
+Equivalent to \tcode{locale(other, std_name.c_str(), cat)}.
```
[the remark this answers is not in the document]
burblebee: Thanks!

source/ranges.tex
burblebee: Actually no; when we introduce code with "Equivalent to:" (note the trailing ":"), the code is supposed to be complete, as if in a codeblock that could be substituted. Your suggestion would apply if the code had been introduced using "Equivalent to", the idea being that later reads as a statement. That said, if this should be an expression and not a statement, then the ";" should be removed. Alternatively, we could use your suggestion and introduce it with "Equivalent to".  What do you prefer?

source/ranges.tex
burblebee: It didn't read correctly to me either, but wasn't sure what was actually intended here (like, is there additional text missing?).  Added as a separate commit.

source/ranges.tex
JohelEGP: Ah, yes. Again, I forgot to remove the colon, too. Edited.

JohelEGP (approved):

source/ranges.tex
tkoeppe: I'm fine leaving it as proposed. We could make the change, but we don't _have_ to, and that way we retain some symmetry between the two instances.

PR #6139, opened 2023-02-19
[LWG 9] P2713R1 Escaping improvements in std::format

### LINKED ISSUE #6099: [2023-02 LWG Motion 9] P2713R1 Escaping improvements in std::format

burblebee:
Fixes NB US 38-098, FR 5-134 (C++23 CD).

Fixes #6099.
Fixes cplusplus/papers#1418
Fixes cplusplus/nbballot#515
Fixes cplusplus/nbballot#408

So I couldn't get the s8 example to build - maybe I don't have the right setup? Can someone else take this and see if they can get the s8 example to build for them?

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -15963,11 +15963,19 @@ Otherwise, if \placeholder{C} is not \unicode{0020}{space} and
 \begin{itemize}
 \item
 \placeholder{CE} is UTF-8, UTF-16, or UTF-32 and
-\placeholder{C} corresponds to either
-a Unicode scalar value whose Unicode property \tcode{General_Category}
-has a value in the groups \tcode{Separator} (\tcode{Z}) or \tcode{Other} (\tcode{C}) or to
-a Unicode scalar value with the Unicode property \tcode{Grapheme_Extend=Yes},
+\placeholder{C} corresponds to a Unicode scalar value
+whose Unicode property \tcode{General_Category} has a value in the groups
+\tcode{Separator} (\tcode{Z}) or \tcode{Other} (\tcode{C}),
 as described by \UAX{44} of the Unicode Standard, or
+
+\item
+\placeholder{CE} is UTF-8, UTF-16, or UTF-32 and
+\placeholder{C} corresponds to a Unicode scalar value
+with the Unicode property \tcode{Grapheme_Extend=Yes}
+as described by \UAX{44} of the Unicode Standard and
+\placeholder{C} is not immediately preceded in \placeholder{S} by
+a character \placeholder{P} appended to \placeholder{E}
+without translation to an escape sequence, or
 
 \item
 \placeholder{CE} is neither UTF-8, UTF-16, nor UTF-32 and
@@ -16050,7 +16058,8 @@ if \placeholder{C} is \unicode{0022}{quotation mark},
 then \placeholder{C} is appended unchanged.
 \end{itemize}
 
-%% FIXME: Example is incomplete; s2 and s6 are missing below;
+%% FIXME: Example is incomplete; s2 and s6 from P2286R8
+%%        and s8 (which should be s9) from P2713R1 are missing below;
 %% FIXME: their Unicode characters are not available in our font (Latin Modern).
 \begin{example}
 \begin{codeblock}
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -16018,8 +16026,9 @@
 then \placeholder{C} is appended unchanged.
 \end{itemize}
 
-%% FIXME: Example is incomplete; s2 and s6 are missing below;
+%% FIXME: Example is incomplete; s2 and s6 from P2286R8 are missing below;
 %% FIXME: their Unicode characters are not available in our font (Latin Modern).
+%% FIXME: ... same for s8 from P2713R1
```
JohelEGP, line 16031: But you're adding an `s6`. Considering that the paper doesn't have `s2` either, the intention must have been to continue the sequence.

source/utilities.tex
```diff
@@ -16030,6 +16039,8 @@
 string s4 = format("[{:?}]", string("\0 \n \t \x02 \x1b", 9));
                                                     // \tcode{s4} has value: \tcode{["\textbackslash u\{0\} \textbackslash n \textbackslash t \textbackslash u\{2\} \textbackslash u\{1b\}"]}
 string s5 = format("[{:?}]", "\xc3\x28");           // invalid UTF-8, \tcode{s5} has value: \tcode{["\textbackslash x\{c3\}("]}
+string s6 = format("[{:?}]", "\u0301");             // s6 has value: ["\u{301}"]
+string s7 = format("[{:?}]", "\\\u0301");           // s7 has value: ["\\\u{301}"]
```
JohelEGP, line 16043: See the comment on the FIXMEs above.
```suggestion
%% FIXME: string s6 = ... string s6 from P2286R8 ... // s6 has value: ...
string s7 = format("[{:?}]", "\u0301");             // s7 has value: ["\u{301}"]
string s8 = format("[{:?}]", "\\\u0301");           // s8 has value: ["\\\u{301}"]
%% FIXME: string s9 = ... string s8 from P2713R1 ... // s9 has value: ...
```

JohelEGP, line 16043:
```suggestion
string s6 = format("[{:?}]", "\u0301");             // \tcode{s6} has value: \tcode{["\u{301}"]}
string s7 = format("[{:?}]", "\\\u0301");           // \tcode{s7} has value: \tcode{["\\\u{301}"]}
```

JohelEGP (changes requested):

source/utilities.tex
burblebee: thanks!

source/utilities.tex
```diff
@@ -15931,12 +15931,20 @@
...
+\placeholder{CE} is a Unicode encoding and
+\placeholder{C} corresponds to a UCS scalar value
+which has the Unicode property \tcode{Grapheme_Extend=Yes} and
+\placeholder{C} is not immediately preceded in \placeholder{S} by
+a character \placeholder{P} appended to \placeholder{E}
+without translation to an escape sequence, or
+
 \item
 \placeholder{CE} is not a Unicode encoding and
 \placeholder{C} is one of an implementation-defined set
```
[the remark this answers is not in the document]
burblebee: I don't know how to resolve this. This PR is for the application of this paper. Merging this PR as is will cause a conflict with CWG motion 8 that @tkoeppe will have to resolve no matter what I do.  @tkoeppe, how do you want to handle this?

source/utilities.tex
burblebee: I sympathize with what you're trying to do here, but we don't want %%FIXMEs to appear in the final printed document, so unfortunately we can't put in placeholders for these missing examples as you suggested. My way of handling this was to just amend the old FIXME comment to say which s8 we were talking about (the one from the original paper). If that s8 is ever added back, it will become sN where N is the next in the sequence.

JohelEGP: > If that s8 is ever added back, it will become sN where N is the next in the sequence.

There might be value in retaining the would-have-been order if the FIXMEs weren't necessary.

source/utilities.tex
```diff
@@ -16030,6 +16039,8 @@
 string s4 = format("[{:?}]", string("\0 \n \t \x02 \x1b", 9));
                                                     // \tcode{s4} has value: \tcode{["\textbackslash u\{0\} \textbackslash n \textbackslash t \textbackslash u\{2\} \textbackslash u\{1b\}"]}
 string s5 = format("[{:?}]", "\xc3\x28");           // invalid UTF-8, \tcode{s5} has value: \tcode{["\textbackslash x\{c3\}("]}
+string s6 = format("[{:?}]", "\u0301");             // \tcode{s6} has value: \tcode{["\u{301}"]}
+string s7 = format("[{:?}]", "\\\u0301");           // \tcode{s7} has value: \tcode{["\\\u{301}"]}
```
JohelEGP, line 16043: Either way is fine.
```suggestion
string s7 = format("[{:?}]", "\u0301");             // \tcode{s7} has value: \tcode{["\u{301}"]}
string s8 = format("[{:?}]", "\\\u0301");           // \tcode{s8} has value: \tcode{["\\\u{301}"]}
```

JohelEGP (approved):

source/utilities.tex
burblebee: Agreed, especially if it would keep similar examples together. That said, we can always renumber the variables to insert the missing examples. As it is, it looks odd to have s2 skipped in the final doc.

JohelEGP: That's definitely true.

tkoeppe: The CD has examples `s2` and `s6`, and the DIS will, too.

PR #6141, opened 2023-02-19
P2614R2 Deprecate numeric_limits::has_denorm

### LINKED ISSUE #6108: [2023-02 LWG Motion 18] P2614R2 Deprecate numeric_limits::has_denorm

jensmaurer:
Fixes #6108 
Fixes cplusplus/papers#1278
Fixes cplusplus/nbballot#506

### DIFF
--- a/source/xrefdelta.tex
+++ b/source/xrefdelta.tex
@@ -110,5 +110,9 @@
 \movedxref{expos.only.func}{expos.only.entity}
 \removedxref{expos.only.types}
 
+% P2614R2 Deprecate numeric_limits::has_denorm
+\movedxref{denorm.style}{depr.numeric.limits.has.denorm}
+\removedxref{fp.style}
+
 % Deprecated features.
 %\deprxref{old.label}  (if moved to depr.old.label, otherwise use \movedxref)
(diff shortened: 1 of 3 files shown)


source/xrefdelta.tex
```diff
@@ -110,5 +110,9 @@
 \movedxref{expos.only.func}{expos.only.entity}
 \removedxref{expos.only.types}
 
+% P2614R2 Deprecate numeric_limits::has_denorm
+\movedxref{denorm.style}{depr.numeric.limits.has.denorm}
+\removedxref{fp.style}
```
JohelEGP, line 115: How about having an "split reference" command. Whereas it is true that [fp.style] was removed, it seems more useful to say it was split between [round.style] and [depr.numeric.limits.has.denorm].

JohelEGP (approved):

source/xrefdelta.tex
jensmaurer: What should the representation in the actual index be?

JohelEGP: A combination of these:

So something like
> fp.style _split_, _see_ round.style, depr.numeric.limits.has.denorm

tkoeppe: Pull requests welcome :-)

PR #6153, opened 2023-02-27
[expected.object.eq] Fix typo

timsong-cpp:
There's no `T1` in sight; the relevant template parameter of `expected` is `T`.


CaseyCarter (approved):

jwakely (approved):
This is a remnant from P0323R10 (and earlier revisions) which had:

```c++
template<class T1, class E1, class T2>
        friend constexpr bool operator==(const expected<T1, E1>&, const T2&);
```

I corrected the signature for R11 but missed this instance of T1.

jensmaurer (approved):

PR #6156, opened 2023-03-02
[temp.expl.spec] Use normal spacing after "etc." within a sentence

jwakely:
I believe LaTeX assumes that a full stop after a lower case letter ends a sentence, so it inserts slightly more space than between other words. I think this is the only occurrence of "etc." that does not end a sentence.

Before:
image
After:
image

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -993,7 +993,7 @@ that potentially conflict and one precedes the other\iref{basic.lookup}.
 \begin{note}
 Overload resolution can consider potentially conflicting declarations
 found in multiple scopes
-(e.g. via \grammarterm{using-directive}s or for operator functions),
+(e.g., via \grammarterm{using-directive}s or for operator functions),
 in which case it is often ambiguous.
 \end{note}
 \begin{example}
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -497,7 +497,7 @@ c.size()
 \pnum
 \returns
 \tcode{distance(c.begin(), c.end())},
-i.e. the number of elements in the container.
+i.e., the number of elements in the container.
 
 \pnum
 \complexity
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -6300,7 +6300,7 @@ defaulted or deleted on its first declaration. A user-provided explicitly-defaul
 is implicitly defined at the point where it is explicitly defaulted; if such a function is implicitly
 defined as deleted, the program is ill-formed.
 A non-user-provided defaulted function
-(i.e. implicitly declared or explicitly defaulted in the class)
+(i.e., implicitly declared or explicitly defaulted in the class)
 that is not defined as deleted is implicitly defined when it is odr-used\iref{basic.def.odr}
 or needed for constant evaluation\iref{expr.const}.
 \begin{note}
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -321,7 +321,7 @@ for a union of type \tcode{U} with a glvalue argument
 that does not denote an object of type \cv{}~\keyword{U} within its lifetime,
 the behavior is undefined.
 \begin{note}
-In C, an entire object of structure type can be accessed, e.g. using assignment.
+In C, an entire object of structure type can be accessed, e.g., using assignment.
 By contrast, \Cpp{} has no notion of accessing an object of class type
 through an lvalue of class type.
 \end{note}
@@ -7262,7 +7262,7 @@ right operand is a bit-field.
 \pnum
 \begin{note}
 In contexts where the comma token is given special meaning
-(e.g. function calls\iref{expr.call},
+(e.g., function calls\iref{expr.call},
 subscript expressions\iref{expr.sub},
 lists of initializers\iref{dcl.init},
 or \grammarterm{template-argument-list}{s}\iref{temp.names}),
--- a/source/styles.tex
+++ b/source/styles.tex
@@ -262,7 +262,7 @@
   \ifvmode\else\par\fi\lst@emptylinepenalty
   \vskip\parskip
   \vskip\baselineskip
-  % \lsthk@EveryLine has \lst@parshape, i.e. \parshape, which causes an \hbox
+  % \lsthk@EveryLine has \lst@parshape, i.e., \parshape, which causes an \hbox
   % \lsthk@EveryPar increments line counters; \refstepcounter balloons the PDF
   \global\advance\lst@newlines\m@ne
   \lst@newlinetrue}
--- a/source/support.tex
+++ b/source/support.tex
@@ -1155,8 +1155,7 @@ Equivalent to \tcode{FLT_RADIX}.
 For integer types, specifies the base of the
 representation.
 \begin{footnote}
-Distinguishes types with bases other than 2 (e.g.
-BCD).
+Distinguishes types with bases other than 2 (e.g., BCD).
 \end{footnote}
 
 \pnum
--- a/source/tables.tex
+++ b/source/tables.tex
@@ -63,7 +63,7 @@
 
 % General Usage: TITLE is the title of the table, XREF is the
 % cross-reference for the table. LAYOUT is a sequence of column
-% type specifiers (e.g. cp{1.0}c), without '|' for the left edge
+% type specifiers (e.g., cp{1.0}c), without '|' for the left edge
 % or right edge.
 
 % usage: \begin{floattablebase}{TITLE}{XREF}{LAYOUT}{PLACEMENT}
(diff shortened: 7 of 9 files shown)


tkoeppe:
Is this the only place we have `etc.` in the middle of a sentence? I thought LaTeX has a list of words that are known inter-sentence occurrences of full stops, but you're right that this looks wrong. Could you have a wider look at more sustainable solutions? And what are we doing with `e.g.` and `i.e.` -- I suppose those are now always followed by a comma and thus sidestep the issue?

Otherwise, in principle this is the correct solution.

jwakely:
```
basic.tex-In this document, the notation \cv{} (or
basic.tex:\cvqual{cv1}, \cvqual{cv2}, etc.), used in the description of types,
basic.tex-represents an arbitrary set of cv-qualifiers, i.e., one of
--
declarations.tex-particular form of representing names of objects and functions with
declarations.tex:external linkage, or with a particular calling convention, etc.
declarations.tex-\end{note}
--
future.tex-  public iterator<bidirectional_iterator_tag, double, long, T*, T&> {
future.tex:  // code implementing \tcode{++}, etc.
future.tex-};
--
lib-intro.tex-
lib-intro.tex:\rSec3[expos.only.entity]{Exposition-only entities, etc.}
lib-intro.tex-
--
lib-intro.tex-Here, the names $\tcode{\placeholder{C}}_0$,
lib-intro.tex:$\tcode{\placeholder{C}}_1$, etc.\ represent
lib-intro.tex-\defnx{enumerated elements}{enumerated element}
--
lib-intro.tex-Here, the names $\tcode{\placeholder{C}}_0$,
lib-intro.tex:$\tcode{\placeholder{C}}_1$, etc.\ represent
lib-intro.tex-\defnx{bitmask elements}{bitmask!element}
--
locales.tex-Thus, a facet class need not have a public
locales.tex:copy constructor, assignment, default constructor, destructor, etc.
locales.tex-\end{footnote}
--
macros.tex-
macros.tex:%% NTBS, etc.
macros.tex-\verbtocs{\StrTextsmaller}|\textsmaller[1]{|
--
modules.tex-default arguments of functions or template declarations, attributes,
modules.tex:names bound, etc.
modules.tex-Since default arguments are evaluated in the context of the call expression,
--
numerics.tex-the first two elements of the result will be value-initialized\iref{dcl.init}; the third element of the result will be assigned the value
numerics.tex:of the first element of \tcode{*this}; etc.
numerics.tex-\end{example}
--
overloading.tex-they can be called explicitly, their addresses can be
overloading.tex:taken, etc.
overloading.tex-\end{note}
--
preprocessor.tex-all identifiers either are or are not macro names ---
preprocessor.tex:there simply are no keywords, enumeration constants, etc.
preprocessor.tex-\end{footnote}
--
preprocessor.tex-Larger alignments will be passed to
preprocessor.tex:\tcode{operator new(std::size_t, std::align_val_t)}, etc.\iref{expr.new}.
preprocessor.tex-\end{note}
--
support.tex-Equivalent to \tcode{CHAR_MIN}, \tcode{SHRT_MIN},
support.tex:\tcode{FLT_MIN}, \tcode{DBL_MIN}, etc.
support.tex-\end{footnote}
--
support.tex-Equivalent to \tcode{CHAR_MAX}, \tcode{SHRT_MAX},
support.tex:\tcode{FLT_MAX}, \tcode{DBL_MAX}, etc.
support.tex-\end{footnote}
--
templates.tex-\grammarterm{base-specifier-list},
templates.tex:\grammarterm{template-argument-list}, etc.
templates.tex-\end{note}
--
templates.tex-That is, its members need
templates.tex:not have the same names, types, etc.\ as the members of a generated
templates.tex-specialization.
--
templates.tex-member function templates of member classes of class
templates.tex:templates, etc., and the placement of partial specialization declarations
templates.tex-of class templates, variable templates,
--
templates.tex-static data member templates of non-template classes, member
templates.tex:class templates of class templates, etc., can affect whether a program is
templates.tex-well-formed according to the relative positioning of the explicit specialization
--
templates.tex-function template specializations, the generation of implicitly-defined functions,
templates.tex:etc. Such effects are not in the ``immediate context'' and can result in the
templates.tex-program being ill-formed.
--
time.tex-However, if a flag refers to a ``time of day''
time.tex:(e.g. \tcode{\%H}, \tcode{\%I}, \tcode{\%p}, etc.),
time.tex-then a specialization of \tcode{duration} is interpreted as
--
time.tex-However, if a flag refers to a ``time of day''
time.tex:(e.g. \tcode{\%H}, \tcode{\%I}, \tcode{\%p}, etc.),
time.tex-then a specialization of \tcode{duration} is parsed as
```

I'm not sure about the "etc. (7.6.2.8)." in [cpp.predefined] but the others look right (and we already use \ like this in lib-intro.tex).

jwakely:
These should use a comma, I think:
```
containers.tex-\tcode{distance(c.begin(), c.end())},
containers.tex:i.e. the number of elements in the container.
containers.tex-
--
declarations.tex-A non-user-provided defaulted function
declarations.tex:(i.e. implicitly declared or explicitly defaulted in the class)
declarations.tex-that is not defined as deleted is implicitly defined when it is odr-used\iref{basic.def.odr}
--
styles.tex-  \vskip\baselineskip
styles.tex:  % \lsthk@EveryLine has \lst@parshape, i.e. \parshape, which causes an \hbox
styles.tex-  % \lsthk@EveryPar increments line counters; \refstepcounter balloons the PDF
```

jwakely:
Likewise:
```
basic.tex-found in multiple scopes
basic.tex:(e.g. via \grammarterm{using-directive}s or for operator functions),
basic.tex-in which case it is often ambiguous.
--
expressions.tex-\begin{note}
expressions.tex:In C, an entire object of structure type can be accessed, e.g. using assignment.
expressions.tex-By contrast, \Cpp{} has no notion of accessing an object of class type
--
expressions.tex-In contexts where the comma token is given special meaning
expressions.tex:(e.g. function calls\iref{expr.call},
expressions.tex-subscript expressions\iref{expr.sub},
--
support.tex-\begin{footnote}
support.tex:Distinguishes types with bases other than 2 (e.g.
support.tex-BCD).
--
tables.tex-% cross-reference for the table. LAYOUT is a sequence of column
tables.tex:% type specifiers (e.g. cp{1.0}c), without '|' for the left edge
tables.tex-% or right edge.
--
templates.tex-% Note: "\space" used below because " " inside tcode adds too much whitespace;
templates.tex:% one could optionally use mathfont inside tcode, e.g. "\tcode{($ $}".
templates.tex-\tcode{(}\space
--
time.tex-However, if a flag refers to a ``time of day''
time.tex:(e.g. \tcode{\%H}, \tcode{\%I}, \tcode{\%p}, etc.),
time.tex-then a specialization of \tcode{duration} is interpreted as
--
time.tex-However, if a flag refers to a ``time of day''
time.tex:(e.g. \tcode{\%H}, \tcode{\%I}, \tcode{\%p}, etc.),
time.tex-then a specialization of \tcode{duration} is parsed as
--
time.tex-If the century is not otherwise specified
time.tex:(e.g.  with \tcode{\%C}),
time.tex-values in the range \crange{69}{99}
```

jwakely:
Yes, this looks too spacey:
image

tkoeppe:
See also https://github.com/cplusplus/draft/pull/1512#issuecomment-342378273.

W-E-Brown:
I was taught in school (admittedly, a looong time ago!) that these Latin abbreviations (i.e., etc., e.g.) should always be followed by a comma unless ending a sentence or unless other punctuation (e.g., a closing parenthesis) interferes.  A quick check of contemporary style guides shows some variation, but I've found none says that the rule I was taught is wrong today.

I recommend we adopt that rule as the simplest resolution, and then special spacing becomes a non-issue.

(Then we can move on to all our many misplaced uses of "only".  Sigh.)

-- WEB

jwakely:
I've added the comma to the remaining i.e. and e.g. cases (including some in latex comments, which isn't necessary, but would make things simpler if we add an automated check for this one day).

Do we want the comma after etc. too? We're currently inconsistent (it's used in a couple of places in templates.tex, but not in lib-intro.tex). I think only a handful of changes would be needed, so I can do that if we want to go with Walter's recommendation.

morinmorin:
`\frenchspacing` seems a big hammer, but maintaining correct spacing in `nonfrenchspacing` is tedious. grepping finds 
```
iostreams.tex:12742:Some behavior is specified by reference to POSIX. How …
```
which should be `POSIX\@. ` and
```
locales.tex:4298:In common U.S.\ locales, this is 2.
```
which makes no harm but it's OK to use `U.S. locales` (lines 4218 and 4233 in the same file use the latter).

PR #6161, opened 2023-03-06
[dcl.pre] Add missing markup that makes "fails" a definition.

tkoeppe:
This change was missed from the application of CWG2518.

Also fixes the indexing of "static_assert" (which is a keyword, not a grammar production).


jensmaurer (approved):

jensmaurer:
Looks good, thanks.

PR #4992, opened 2021-10-04
P2334R1 Add support for preprocessing directives elifdef and elifndef

jensmaurer:
Fixes #4966
Fixes cplusplus/papers#1016

PR #6162, opened 2023-03-07
[cpp.pre] Fix grammar for #elifdef

AlisdairM:
This fix is editorial as it corrects a mis-application of the original paper, https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2334r1.pdf.

Note that the corresponding use of this grammar in [cpp.cond]p13 assumes the correct grammar, making the intent equally clear.


JohelEGP:
Looking at https://github.com/cplusplus/draft/pull/4992#discussion_r722037577, it must've been a mismerge.

JohelEGP (approved):

jensmaurer (approved):

PR #6164, opened 2023-03-08
[depr.template.template] Add cross-ref to core language

AlisdairM:
Add a cross reference to the core language paragraph that contains the deprecation notice, [temp.names] (p6).


jensmaurer (approved):

jensmaurer:
The cross-reference points to the right place, and adding cross-references for cases like these seems helpful to the reader.

Issue #4720, opened 2021-06-25
[lib] space between requires and its parameter list?


jensmaurer:
Despite #4719, it is not at all clear whether there should be a space between "requires" and its parameter list.

Statistics: 74 places with space, 46 wihout.

fgrep "requires (" *.tex | wc -l

CaseyCarter:
In the library, we consistently use:
* `requires(` (no space) to introduce a _requires-expression_ with parameters,
* `requires {` to introduce a _requires-expression_ without parameters, and
* `requires (e)` (space) to introduce a _requires-clause_ whose first term `e` is not a _primary-expression_.

This makes it easy to determine at-a-glance when `requires` is introducing a _requires-clause_ vs. a _requires-expression_.

tkoeppe:
This seems like a good rule. Perhaps we should add it to the wiki.

tkoeppe:
Editorial meeting decision: We like Casey's rule and will adopt it generally.

PR #6166, opened 2023-03-11
[over.match.funcs.general]/9 fix the style of references

blackteahamburger:
The style of references here (`T &&`) is different from that elsewhere (`T&&`).

### DIFF
--- a/source/overloading.tex
+++ b/source/overloading.tex
@@ -333,15 +333,15 @@ if the argument list has exactly one argument and
 \begin{codeblock}
 struct A {
   A();                                  // \#1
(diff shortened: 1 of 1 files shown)


JohelEGP:
Only the library wording is intentionally consistent in its coding conventions. See the last paragraph at https://github.com/cplusplus/draft/issues/4720#issuecomment-868861847.

jensmaurer:
The core language does not prescribe any particular coding style, nor is there a specific coding style recommended or approved by the standard. Thus, examples use a variety of styles intentionally.

PR #6170, opened 2023-03-11
[dcl.init.general] Clarify initialization of arrays

lprv:
[[dcl.init.general]/13](https://wg21.link/dcl.init.general#13) interferes with parenthesized initialization of arrays, and [[dcl.init.general]/16](https://wg21.link/dcl.init.general#16) does not specify the behavior of copy-initialization for array types.


jensmaurer:
There might be a bug in the standard hiding somewhere, but this change doesn't feel editorial to me.

PR #6171, opened 2023-03-11
[expr.prim.lambda.capture] Improve wording on anonymous unions in lambda captures

lprv:
The restrictions in [[expr.prim.lambda.capture]](https://wg21.link/expr.prim.lambda.capture) should refer to anonymous union variables instead of their members, as the latter aren't local entities and can't be captured anyway.


jensmaurer:
What is an "anonymous union variable"? Why would it be captured when attempting to capture a member of an anonymous union?

jensmaurer:
Since members of anonymous union variables are not local entities (as you correctly observe), they can't be captured to start with.

I think the correct fix is to remove both mentions of anonymous union members (normative duplication) and add a note after [expr.prim.lambda.capture] p1, saying that members of anonymous unions cannot be capture because they are not local entities (xref).

PR #6172, opened 2023-03-11
[conv.qual] Remove unused definition of 'cv-qualification signature'

lprv:


tkoeppe:
@jensmaurer Any thoughts on this? I think this is some of your wording? Was "cv-qualification signature" ever used?

jensmaurer:
I don't remember. I think we still have a latent bug in the area that we should explicitly talk about the "longest decomposition" in more places.

In any case, if it's indeed unused (I didn't check), the definition can go.

AlisdairM:
I ran `grep "cv-qualification signature" *.tex` and it gave just the definition above.

I then ran `grep cv-qualification *.tex` to check that there were no word-wrapping cases to fool us, and confirmed that each match, other than the definition above, was not an occurrence of this phrase.

I believe that satisfies an audit that this defined term is never used.

tkoeppe:
I think this is indeed unused, so let's apply this. Thanks!

PR #6173, opened 2023-03-12
[expr.rel] Clarify confusing wording CWG2749

lprv:
'[Pointer to an object](https://wg21.link/basic.compound#def:pointer_to)' is one of the pointer value categories defined in [basic.compound], but according to [CWG2526](https://cplusplus.github.io/CWG/issues/2526.html) the intended reading here is that `void*` values are excluded from the ordering.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -6489,7 +6489,7 @@ them to their composite pointer type\iref{expr.type}.
 After conversions, the operands shall have the same type.
 
 \pnum
-The result of comparing unequal pointers to objects
+The result of comparing unequal values of pointer-to-object type
 \begin{footnote}
 As specified in \ref{basic.compound},
 an object that is not an array element


jensmaurer:
What's wrong with the existing text?  A "pointer to void" is not a pointer to object, because void is not an object type.

tkoeppe:
@jensmaurer Any further thoughts? This seems like a corish detail I'd defer to you on.

jensmaurer:
I think this is good, but CWG should have a look because it's a bit of a delicate wording here.

t3nsor:
@jensmaurer  Is there a core issue for this?

jensmaurer:
CWG 2023-06-14: Needs research on C behavior for "pointer to void".

t3nsor:
It appears that this is now CWG2749?

jensmaurer:
Yes.

tkoeppe:
Already handled by CWG issue.

PR #6174, opened 2023-03-12
Bring pointer terminology up to date

lprv:
Pointer values are not addresses ([[basic.compound]/3](https://wg21.link/basic.compound#3.sentence-8)), this updates some of the old text to use proper terminology. The intent of the wording is clear, and this PR does not change its normative meaning.

I'm not sure if changing 'address of' to 'pointer representing the address of' is an improvement, so the following are left untouched:
* [[basic.stc.dynamic.allocation]/2](https://wg21.link/basic.stc.dynamic.allocation#2.sentence-2)
* [[expr.delete]/11](https://wg21.link/expr.delete#11.sentence-4)
* [[dcl.fct.def.coroutine]/12](https://wg21.link/dcl.fct.def.coroutine#12.sentence-6)

Also, in [[temp.arg.nontype]/3](https://wg21.link/temp.arg.nontype#3), the text should probably also include past-the-end pointers, but that's not editorial.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -2040,9 +2040,9 @@ if the function call operator
 has a non-throwing exception specification.
 If the function call operator is a static member function,
 then the value returned by this conversion function is
-the address of the function call operator.
+a pointer to the function call operator.
 Otherwise, the value returned by this conversion function
-is the address of a function \tcode{F} that, when invoked,
+is a pointer to a function \tcode{F} that, when invoked,
 has the same effect as invoking the closure type's function call operator
 on a default-constructed instance of the closure type.
 \tcode{F} is a constexpr function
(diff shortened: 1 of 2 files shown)


tkoeppe:
Why is this edtiorial? Unless it is obvious that a change is non-normative (e.g. because it changes a note or example, or fixes a typo), there should be some rationale explaining the intended meaning and why that's unambiguous, and why the change retains that meaning. Otherwise such change requests should probably be CWG issues.

source/basic.tex
```diff
@@ -3988,7 +3988,7 @@
 currently installed new-handler function\iref{new.handler}, if any.
 \begin{note}
 \indextext{\idxcode{new_handler}}%
-A program-supplied allocation function can obtain the address of the
+A program-supplied allocation function can obtain the
```
jwakely, line 3991: This part is non-normative and seems like an improvement. `std::new_handler` is defined as a pointer to a function, so what `std::get_new_handler` returns is a new_handler, not the address of one.

tkoeppe: Indeed, thanks. @lprv Feel free to send this one change as a separate PR if you like!

tkoeppe:
@jensmaurer I think some CWG input here would be good, but it seems like a nice improvement.

jensmaurer:
The first part (about lambdas) feels editorial, the other parts fix normative defects and thus should be addressed by a CWG issue.

tkoeppe:
CWG meeting consensus Nov 11: good

tkoeppe:
@jensmaurer Could you check again that the entire PR as it is now is editorial?

tkoeppe:
@jensmaurer Ping

jensmaurer (approved):
Yes, please.

PR #6177, opened 2023-03-13
[range.cartesian.view] Remove unused template parameter

hewillk:
Is it reasonable to do so?

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -14989,7 +14989,7 @@ namespace std::ranges {
       (@\libconcept{bidirectional_range}@<@\exposid{maybe-const}@<Const, Vs>>
         && @\exposconcept{cartesian-product-common-arg}@<@\exposid{maybe-const}@<Const, Vs>>));
 
-  template<class First, class... Vs>
+  template<class First, class...>
   concept @\defexposconcept{cartesian-product-is-common}@ =                 // \expos
     @\exposconcept{cartesian-product-common-arg}@<First>;
 


JohelEGP:
Part of #5735. I don't see why not.

CaseyCarter:
I think I'd rather see the unused template parameter pack removed from the exposition-only concept; there are only four uses to clean up.

hewillk:
> I think I'd rather see the unused template parameter pack removed from the exposition-only concept; there are only four uses to clean up.

https://github.com/cplusplus/draft/issues/5735#issuecomment-1222560612: "is a cartesian product of all of these ranges a common range?". That's also why it takes all the ranges as arguments, instead of just the first one.

Although it's clearly the intended design. I think it's more straightforward to just remove the template parameter pack, and I don't see how it would cause confusion, so I prefer removing it.

tkoeppe:
@CaseyCarter Would that be editorial, or should we open an LWG issue to remove the pack?

CaseyCarter (approved):

CaseyCarter:
> @CaseyCarter Would that be editorial, or should we open an LWG issue to remove the pack?

I think the change in the PR is both clearly editorial, and a minor readability improvement. Nothing says "yes, I really intend to ignore this parameter" as explicitly as does leaving it unnamed.

Any further simplifications - Should we remove the parameter pack? Should we replace uses of `cartesian-product-is-common<T, Us...>` with `cartesian-product-common-arg<T>` and remove the former altogether? - probably ought to go through LWG. The current structure was put in place for readability, and while we may not all agree that it's ideal, we should give plenty of eyes the chance to comment on any larger structural change.

tkoeppe:
@CaseyCarter Thanks!

@hewillk Please file an LWG issue if you want to remove the template parameter.

PR #6180, opened 2023-03-13
[fs.conform.9945] Use inter-sentence spacing after an (uppercase) acronym

morinmorin:


jensmaurer (approved):

morinmorin:
Hmm..., grepping with `[A-Z]\.\s+\w` and a multiline option finds this:

- compatibility.tex:2307:
  A \keyword{struct} is a scope in \Cpp{}, not in **C.**
  For example,

- support.tex:5906:
  one of these headers in order to be valid ISO **C.**
  Source files that are not intended to also be valid ISO C

- support.tex:5918:
  The following source file is both valid \Cpp{} and valid ISO **C.**
  Viewed as \Cpp{}, it declares a function with C language linkage;

- time.tex:2815:
  The UTC time standard began on 1972-01-01 00:00:10 **TAI.**
  To measure time since this epoch instead, one can add/subtract the constant

These should be "UPPERCASES\\@."

morinmorin:
We also need to fix cases like
```
Let \tcode{i} be an object of type \tcode{I}. When ...
```

I don't know if these are worth fixing or if `\nofrenchspacing` can be a realistic option.

tkoeppe:
Please check first if these actually make a difference to the presentation. I also looked at a couple of such cases where the additional annotations did not affect the visual output.

As far as I can tell, some of these changes make minor differences to the available glue, which may not result in any actual changes.

morinmorin:
And here is a diff for the second comment. In that comment, I just picked one example of bad spacing, but actually there are **TENS** of such spacing (i.e. `UPPERCASE}. …`).
[tcode.pdf](https://github.com/cplusplus/draft/files/10960100/tcode.pdf)

tkoeppe:
Could you just post a `diffpdf` screenshot?

jensmaurer:
Example:

Before:

image

After:

image

This looks like an improvement to me.

jensmaurer:
I have an automated check; 87 occurrences outstanding.

PR #6182, opened 2023-03-13
[intro.abstract] Index the you-know-what clause

AlisdairM:
Given we have been colloquially referring to this clause by name
for many years, it seems reasonable to add it to the index.                     

If I were aiming more at an Easter egg, I would have indexed as
"travel!time" but I don't think we do Easter eggs any more?

The index entry might be better served by adding a footnote that
"this freedom is often referred to as the time travel clause",
but I did not want to draw excessive attention for such a minor
edit (OK, leaving it half-way to an Easter Egg).


tkoeppe:
I'm not sure we need this. Please feel free to reopen with some stronger rationale.

PR #6186, opened 2023-03-14
[projected, alg.req.ind.{move, copy}, range.as.rvalue.overview] fix typo

morinmorin:


jensmaurer (approved):
Looks good. I'm wondering whether the commit description could be improved; the list of individual locations is quite long.

PR #6189, opened 2023-03-14
[container.alloc.reqmts] Better cross-references for allocator-aware …

AlisdairM:
…containers

There are now more allocator-aware containers in the standard than when this subclause was first written, so ensure we have call outs to all relevent subclauses.

The current wording for 'basic_stacktrace' also shows how containers can properly call out the allocator-aware container requirements, now they have their own titled subclause.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -888,8 +888,8 @@ Linear.
 \rSec3[container.alloc.reqmts]{Allocator-aware containers}
 
 \pnum
-All of the containers defined in \ref{containers} except \tcode{array},
-in \ref{stacktrace.basic}, in \ref{basic.string}, and in \ref{re.results}
+Except for \tcode{array}, all of the containers defined in \ref{containers},
+\ref{stacktrace.basic}, \ref{basic.string}, and \ref{re.results}
 meet the additional requirements of an \defnadj{allocator-aware}{container},
 as described below.
 
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -888,7 +888,8 @@ Linear.
 \rSec3[container.alloc.reqmts]{Allocator-aware containers}
 
 \pnum
-All of the containers defined in \ref{containers} and in~\ref{basic.string} except \tcode{array}
+All of the containers defined in \ref{containers} except \tcode{array},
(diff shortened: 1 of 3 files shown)


tkoeppe:
@jwakely, @CaseyCarter Could you please have a look? This looks reasonable to me.

source/containers.tex
```diff
@@ -861,7 +861,8 @@
 \rSec3[container.alloc.reqmts]{Allocator-aware containers}
 
 \pnum
-All of the containers defined in \ref{containers} and in~\ref{basic.string} except \tcode{array}
+All of the containers defined in \ref{containers} except \tcode{array},
+in \ref{stacktrace.basic}, in \ref{basic.string}, and in \ref{re.results}
```
CaseyCarter, line 865: No change requested, but something to consider: IMO this would read better if the "in containers except array" case was last, but it would seem quite odd not to list the "real" containers first.

CaseyCarter (approved):
This seems an overall improvement to me.

source/containers.tex
AlisdairM: I considered this, but to avoid all confusion I wanted "in \ref{containers} except \tcode{array}" to clearly be a single item in the list, where placing this last requires attention to the Oxford comma to understand the intent.  The difference has no impact on interpretation, as `array` *is specified in* the [containers] clause, but I feel it gives an easier parse so the reader never worries about those details.  So yes, this ordering was intentional :) but I am open to others' view on clarity/ease-of-reading.

jensmaurer: Maybe put the exception first:

"Except for array, all of the containers defined in a,b,c meet the additional requirements..."

This doesn't tell you where "array" is defined, but maybe that's not super-important here.

jensmaurer (changes requested):
We need a rebase.

jensmaurer:
I think I prefer my suggestion over the status quo, even after a night's sleep.

jwakely (approved):
I think this now has the requested change from Jens, and is ready to squerge.

jensmaurer (approved):

source/containers.tex
tkoeppe: Thank's, Jen's!

PR #6192, opened 2023-03-16
[container.reqmts] Fix cross-references to contiguous container

AlisdairM:
This edit deserves describing in detail, to establish that it is purely editorial.

As part of the mission to clean up tables in the standard, the general container requirements were split into 5 more focused leaf nodes.  Due to its location in the original text, the definition for a contiguous container fell into the subclause for reversible containers, to which it is entirely unrelated.  There is no requirement that a contiguous container be reversible, only that it has the correct category for its iterators.

Meanwhile, all 3 of the existing cross-references point to the wrong leaf node, that simply provides a key to the identifiers used throughout the specification of this clause.

The fix has two parts.  First move the definition of contiguous container, and a container with special iterators, into the [container.reqmts] clause where it best belongs.  This move is appended to the end so that there can be no ambiguity that any existing text could be confused with requirements only on contiguous
containers after the edit.  The other part is to fix up the three cross-references to point to [container.reqmts] rather than its sibling that has no information on contiguous containers.

A grep of the .tex source files confirms that these three references (array,basic_string, and vector) are the only current uses of contiguous container, even though basic_stacktrace would also meet the requirements.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -628,6 +628,14 @@ element in the other container after the swap. It is unspecified whether an iter
 with value \tcode{a.end()} before the swap will have value \tcode{b.end()} after the
 swap.
 
+\pnum
+A \defnadj{contiguous}{container}
+is a container
+whose member types \tcode{iterator} and \tcode{const_iterator}
+meet the
+\oldconcept{RandomAccessIterator} requirements\iref{random.access.iterators} and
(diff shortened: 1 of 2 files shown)


jensmaurer (approved):
This moves a self-contained definition and adjusts cross-references accordingly. This feels low-risk to me.

jwakely (approved):

tkoeppe:
Thank you, nice catch!

PR #6197, opened 2023-03-20
[range.adaptor.helpers] make as-lvalue noexcept

hewillk:
Given that it's a sibling of `std::as_const`, would adding `noexcept` here conform to a more standard style even though it's exposition only?

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -4302,7 +4302,7 @@ namespace std::ranges {
   }
 
   template<class T>
-  constexpr T& @\exposid{as-lvalue}@(T&& t) {                   // \expos
+  constexpr T& @\exposid{as-lvalue}@(T&& t) noexcept {          // \expos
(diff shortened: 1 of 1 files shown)


tkoeppe:
@CaseyCarter Could you have a look please?

tkoeppe:
@jwakely: is this editorial?

jwakely:
Yes, it seems editorial, but completely unnecessary.

There are only a few uses of _as-lvalue_ and none of them are in `noexcept` functions, and none of them depend on whether that cast can throw, and users can't observe whether or not it's noexcept. I see no reason to clutter an exposition-only function with keywords that have no effect.

Adding `noexcept` to _as-const-pointer_ in [[range.prim.cdata]](https://wg21.link/range.prim.cdata) seems like it _might_ be more useful, since we use that directly in an "expression equivalent" context that users could actually test with the `noexcept` operator.

PR #6199, opened 2023-03-21
[container.reqmts] Relocate more container requirements

AlisdairM:
As part of rewriting the container requirements tables as text, several paragraphs of general container requirements were buried in the new subsection for more specific requirements such as reversible containers or allocator-aware containers.  This seems to have happened when general container requirements followed one of the container requirements tables that are noe expressed as text.

This PR carefully restores text to the general container requirements clause, in the original order they appeared in C++20, and audits that there are no further paragraphs to move that do not have outstanding pull requests.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -629,12 +629,81 @@ with value \tcode{a.end()} before the swap will have value \tcode{b.end()} after
 swap.
 
 \pnum
+Unless otherwise specified (see~\ref{associative.reqmts.except}, \ref{unord.req.except}, \ref{deque.modifiers}, and
+\ref{vector.modifiers})
+all container types defined in this Clause meet
+the following additional requirements:
+
+\begin{itemize}
+\item
+If an exception is thrown by an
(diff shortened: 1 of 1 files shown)


tkoeppe:
@jwakely, @CaseyCarter Could you kindly have a look?

tkoeppe:
@jwakely, @CaseyCarter Could you kindly have a look?

jwakely (approved):
Yes, this definitely needs to be fixed for C++23. In C++20 the two "Unless otherwise specified" chunks of text about came after the paragraph about reversible containers: "If the iterator type of a container belongs to the bidirectional or random access iterator categories (23.3), the container is called reversible and meets the additional requirements in Table 74." In the C++23 draft, that paragraph became [container.rev.reqmts] but the following paragraphs should not have been made part of that new subclause.

Similarly, the "The behavior of certain container member functions" chunk of text came after the paragraph that became [container.alloc.reqmts], but should not have gone into that subclause.

tkoeppe:
Thanks!

PR #6200, opened 2023-03-21
Prevent double spacing in tcode

morinmorin:
Due to `nofrenchspacing`, a space after `.`, `:`, `?` are rendered as two spaces in monospaced fonts if these punctuations are preceded by non-capital characters. For example, `\tcode{a ? b : c}` is rendered as `a ?  b :  c`, and
```
\tcode{template <class... A> void f(A...) noexcept}
```
is rendered as
```
template <class...  A> void f(A...)  noexcept;
```

Currently, there are tens of these double spacings in the Standard.
This PR prevents them by adding `\frenchspacing` to `CodeStylex`.

### DIFF
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -175,7 +175,7 @@
 % This command uses the "cooked" \indeximpldef command to emit index
 % entries; thus they only work for simple index entries that do not contain
 % special indexing instructions.
-\newcommand{\impldef}[1]{\indeximpldef{#1}implementation-defined}
+\newcommand{\impldef}[1]{\indeximpldef{#1}imple\-men\-ta\-tion-defined}
 % \impldefplain passes the argument directly to the index, allowing you to
 % use special indexing instructions (!, @, |).
 \newcommand{\impldefplain}[1]{\index[impldefindex]{#1}implementation-defined}
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -219,7 +219,7 @@
 % General code style
 %%--------------------------------------------------
 \newcommand{\CodeStyle}{\ttfamily}
-\newcommand{\CodeStylex}[1]{\texttt{#1}}
+\newcommand{\CodeStylex}[1]{\texttt{\protect\frenchspacing #1}}
 
 \definecolor{grammar-gray}{gray}{0.2}
 
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -10873,7 +10873,7 @@ is \tcode{true},
 Otherwise,
 if \tcode{(\libconcept{derived_from}<Cs, bidirectional_iterator_tag> \&\& ...)}
 is \tcode{true},
-\tcode{itera\-tor_category} denotes \tcode{bidirectional_iterator_tag}.
+\tcode{iterator\-_category} denotes \tcode{bidirectional_iterator_tag}.
 \item
 Otherwise,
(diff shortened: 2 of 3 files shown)


morinmorin:
Diff screenshot.
frenchspacing

morinmorin:
Defining `\tcode` using `\lstinline` would be far better. Benefits of `\lstinline` include

- Single quotes are rendered as straight quotes (using `upquote` option).
- No escape is needed; we can just write `%`, `\`, etc.
- No double spacings.

`\lstinline` can be used with `mathescape` option. So we can use macros in `$...$` e.g. `using $\exposidnc{indirect-value-t}$ = $\seebelow$`.

However, **`\lstinline` behaves badly when used in macro arguments**. For example, backslashes are used as escape characters in the TeX world and spaces before `$` are not rendered.

jensmaurer:
This now has overfull `\hbox` issues that need to be addressed.

jensmaurer (changes requested):
Please fix overfull lines per "check-output.sh" results.

morinmorin:
Thanks for the guidance. There are three overfull boxes.

- `ranges.tex (L10873–10878, std.pdf: p. 1225+10): 0.32013pt too wide`
  * The offending word: `\tcode{iterator_category}`.
  * The overfullness is so small. This “overfull hbox” warning is suppressed by locally increasing `\hfuzz` from `0.1pt` (the default value) to `0.5pt` (this value is also used in `\sloppy`). Note that “overfull hbox” warnings are emitted only when the width is more than `\hfuzz`'s value. 

- `ranges.tex (L15112–15116, std.pdf: p. 1262+10): 141.33292pt too wide`
  * The offending “word”: `\tcode{… ? … : …}`
  * Easily fixable. It is fixed by adding `\brk` after `: `.

- `iostreams.tex (L2279–2287 std.pdf: p. 1628+10): 21.81808pt too wide`
  * The offending word: "implementation-defined" emitted from `\impldef`.
  * TeX does not hyphenate words that already contain hyphens (compound words). Adding `\-` changes this behavior. Compound words might be better not to be hyphenated, so I add `\-` conservatively (i.e. `imple\-men\-ta\-tion-defined` rather than `im\-ple\-men\-ta\-tion-de\-fined`). 

The global macro `\impldef` is changed in the third one, which introduces changes like "implementation-defined" -> "implementa-tion-defined" in other places.

morinmorin:
A screenshot for the first overfull hbox (`iterator_-`).
iterator_category

jensmaurer:
For the first transgression, can you add a hyphenation hint iterator_catego\-ry instead? Usually, the spaces are already very compressed when you get an overfull \hbox warning, and we regularly hyphenate this \tcode words if necessary.

morinmorin:
For the first transgression, the offending part is the `-` just after the highlighted "tor_" in the screenshot. `\tcode{it\-er\-a\-tor_cat\-e\-go\-ry}` doesn't make any change in the output.

The only viable hyphen-hint is `\tcode{iterator\-_category}`, which outputs "iterator-`\n`_category". Is this our preferred option?

morinmorin:
I took this route: 

> `\tcode{iterator\-_category}`

The issue is resolved now, I think.

jensmaurer:
@tkoeppe, this looks like an editorial improvement.

tkoeppe:
@jensmaurer Yes, I agree. Any final thoughts?

jensmaurer (approved):
Fine with me.

PR #6205, opened 2023-03-23
[version.syn] bump value of __cpp_lib_allocate_at_least

### LINKED ISSUE #6202: Bump value of __cpp_lib_allocate_at_least for C++23
[LWG 3887](https://cplusplus.github.io/LWG/issue3887) wants to increase the value of the `__cpp_lib_allocate_at_least` macro to distinguish the pre-P2652R0 API from the new one that will be in C++23. LWG voted this Tentatively Ready, but if we follow the usual process that will be approved in Varna, for C++26. There's not much point updating it if we only do so for C++26.

I think it can be argued that the value of the macro in the standard is editorial. The standard doesn't say what a given value means, only [SD-6](https://isocpp.org/std/standing-documents/sd-6-sg10-feature-test-recommendations#__cpp_lib_allocate_at_least) does that.

Would the editor please consider giving the value `202302L` to `__cpp_lib_allocate_at_least`, and we'll get @brevzin to update SD-6?

jwakely:
This is requested for the Tentatively Ready LWG issue 3887.

Fixes #6202

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -553,7 +553,7 @@ the values of these macros with greater values.
 #define @\defnlibxname{cpp_lib_addressof_constexpr}@               201603L // also in \libheader{memory}
 #define @\defnlibxname{cpp_lib_algorithm_iterator_requirements}@   202207L
   // also in \libheader{algorithm}, \libheader{numeric}, \libheader{memory}
-#define @\defnlibxname{cpp_lib_allocate_at_least}@                 202106L // also in \libheader{memory}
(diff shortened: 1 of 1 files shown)


jensmaurer (approved):

jwakely:
I've rebased this to get the Mac modernization change, and the checks passed now.

tkoeppe:
You can just "rerun" the checks!

tkoeppe:
Oh, but if you don't rebase, then the old checks will still fail of course, never mind.

PR #6206, opened 2023-03-23
Mac modernization: allow listings 1.9, and don't allow errors in a pipe

tkoeppe:
@jensmaurer We need to allow listings 1.9 for MacOS. Is there any point in having the version check at all?

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -33,6 +33,11 @@ jobs:
       - name: check-source.sh
         run: ../tools/check-source.sh
 
+      - name: update brew
+        if: matrix.cfg.os == 'macos-12'
+        run: |
+          brew update
+
       - name: update-apt-cache
         if: matrix.cfg.os == 'ubuntu-22.04'
         run: sudo apt-get update
@@ -41,7 +46,7 @@ jobs:
         if: matrix.cfg.os == 'ubuntu-22.04'
         run: sudo apt-get install latexmk texlive-latex-recommended texlive-latex-extra texlive-fonts-recommended lmodern
 
-      - name: install (MacOS, Part 1/2)
+      - name: install (MacOS)
         if: matrix.cfg.os == 'macos-12'
         run: |
           brew install basictex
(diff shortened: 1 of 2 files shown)


source/styles.tex
```diff
@@ -142,13 +142,13 @@
 %% (copied verbatim from listings.sty version 1.6 except where commented)
 \makeatletter
 
-\lst@CheckVersion{1.8d}{\lst@CheckVersion{1.8c}{\lst@CheckVersion{1.8b}{\lst@CheckVersion{1.7}{\lst@CheckVersion{1.6}{\lst@CheckVersion{1.5b}{
+\lst@CheckVersion{1.9}{\lst@CheckVersion{1.8d}{\lst@CheckVersion{1.8c}{\lst@CheckVersion{1.8b}{\lst@CheckVersion{1.7}{\lst@CheckVersion{1.6}{\lst@CheckVersion{1.5b}{
```
tkoeppe, line 145: What happens if we just don't check this?

jensmaurer: The problem is I'm overwriting internal macros of the listing package, which may or may not work in a future version.

But I guess that uncertainty is better than a reliable failure with each new version of the package.

jensmaurer: Let's get rid of the check.

.github/workflows/check.yml
```diff
@@ -50,9 +55,14 @@ jobs:
...
           sudo tlmgr install latexmk isodate substr relsize ulem fixme rsfs extract layouts enumitem l3packages l3kernel imakeidx splitindex xstring
 
-      - name: make
+      - name: make (Linux)
+        if: matrix.cfg.os == 'ubuntu-22.04'
         run: make quiet
 
+      - name: make (MacOS)
+        if: matrix.cfg.os == 'macos-12'
+        run: make full
```
tkoeppe, line 64: So, this one is weird: the `||` in the Makefile seems to not work on MacOS anymore.

jensmaurer (approved):

.github/workflows/check.yml
tkoeppe: Nothing is needed, but it's an odd change of behaviour, which I imagine should show up elsewhere... but I think we can live with it.

jwakely: The default shell changed from bash to zsh in modern macOS versions, but that shouldn't matter. It seems more likely that the behaviour of latexmk changed so that -silent causes it to exit with status 0 even if the build failed.

tkoeppe: Ohh, that's interesting! That means we'll see the same on Linux when that version changes, I expect!

PR #6208, opened 2023-03-25
[range.utility.conv.to] Use model instead of satisfy for range concept

hewillk:
Is this editorial? If not, is this change worthwhile?


jensmaurer:
This change is not good, I think. This is a place where we want to select different behavior depending on syntactic properties, not check semantic preconditions.

hewillk:
Close base on the feedback. Thanks for the explanation.

PR #6209, opened 2023-03-27
[over.literal] Cross-reference deprecated grammar

AlisdairM:
The core conention is to retain deprecated wording in the core clauses, but always make a forward link to Annex D where a feature, or parts of a feature, are deprecated.


jensmaurer (approved):
Yes, that's a consistency improvement.

PR #6210, opened 2023-03-27
[basic.compound] Fix incorrect cross-reference

lprv:


jensmaurer (approved):
Yes, alignment has nothing to do with incomplete types.

PR #6215, opened 2023-03-28
[dcl.decl.general] Fix cross-references

jensmaurer:
"templated function" is defined in [temp.pre], but avoid having two cross-references to the same place in the same paragraph by dropping the cross-reference for the definition of the grammar non-terminal "requires-clause".

Partially addresses llvm/llvm-project#61748


shafik:
Thank you for the fix!

Issue #6204, opened 2023-03-23
N4944 pre-CD: Remaining inconsistent inline constexpr variable templates


Dani-Hub:
After the recent adjustment to strike "inline" from constexpr variable templates, the following cases are still present (looking at the N4944 pre-CD working draft):

1) [ranges.syn]:
```
namespace views {
  template<size_t N>
    inline constexpr unspecified adjacent = unspecified; // freestanding
  [...]
}
[...]
namespace views {
  template<size_t N>
    inline constexpr unspecified adjacent = unspecified; // freestanding
  [...]
}
[...]
namespace views {
  template<size_t N>
    inline constexpr unspecified adjacent_transform = unspecified; // freestanding
  [...]
}
```
2) [range.refinements]:
```
template<class R>
  inline constexpr bool is-initializer-list = see below ; // exposition only
```

jwakely:
I agree.

tkoeppe:
Thank you both, sounds good! Does either of you fancy making a PR? :-) If
not, I'll get to it in a bit.

tkoeppe:
Thanks a lot!

frederick-vs-ja:
This should have been fixed.

AlisdairM:
This was most recently addressed by #7240.  Does that resolve this issue?

PR #6218, opened 2023-03-29
#6204 Remove remaining inline from variable templates

Dani-Hub:


tkoeppe:
@CaseyCarter, @jwakely could you please have a look?

source/ranges.tex
```diff
@@ -1584,7 +1584,7 @@
 
 \begin{itemdecl}
 template<class R>
-  inline constexpr bool @\exposidnc{is-initializer-list}@ = @\seebelow@;        // \expos
+  constexpr bool @\exposidnc{is-initializer-list}@ = @\seebelow@;        // \expos
```
JohelEGP, line 1587:
```suggestion
  constexpr bool @\exposidnc{is-initializer-list}@ = @\seebelow@;                // \expos
```
Just in case it already was aligned. The check script skips these.

JohelEGP (changes requested):

source/ranges.tex
Dani-Hub: I hope I fixed it correctly.

JohelEGP: Seems like you added one too many.

JohelEGP (approved):
All that's missing is following https://github.com/cplusplus/draft/wiki/Commit-message-format#editorial-commits.

jwakely (approved):

jensmaurer (approved):
Looks good; needs a fixed commit message when squashing.

PR #6220, opened 2023-04-04
[container.opt.reqmts] Index 3-way compare for containers

AlisdairM:
All of the standard containers are 3-way comparable, but the specification is provided under optional container requirements so does not show up in the index.

Adding these links shows just how poorly indexed the members of the 'flat_*' container adapters are, but that is properly the business on a separate PR.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -868,10 +868,6 @@ are implemented by constexpr functions.
 \indexlibrarymemberx{set}{#1}%
 \indexlibrarymemberx{multiset}{#1}%
 \indexlibrarymemberx{multimap}{#1}%
-\indexlibrarymemberx{unordered_map}{#1}%
-\indexlibrarymemberx{unordered_set}{#1}%
-\indexlibrarymemberx{unordered_multiset}{#1}%
-\indexlibrarymemberx{unordered_multimap}{#1}%
 \indexlibrarymemberx{flat_map}{#1}%
 \indexlibrarymemberx{flat_set}{#1}%
 \indexlibrarymemberx{flat_multiset}{#1}%
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -856,19 +856,8 @@ meet the constexpr iterator requirements\iref{iterator.requirements.general}
 then the operations described below
 are implemented by constexpr functions.
 
-\indexlibrarymember{array}{operator<=>}%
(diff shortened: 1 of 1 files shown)


source/containers.tex
```diff
@@ -856,6 +856,24 @@
...
+\indexlibrarymember{list}{operator<=>}%
+\indexlibrarymember{vector}{operator<=>}%
+\indexlibrarymember{map}{operator<=>}%
+\indexlibrarymember{set}{operator<=>}%
+\indexlibrarymember{multiset}{operator<=>}%
+\indexlibrarymember{multimap}{operator<=>}%
+\indexlibrarymember{unordered_map}{operator<=>}%
+\indexlibrarymember{unordered_set}{operator<=>}%
+\indexlibrarymember{unordered_multiset}{operator<=>}%
+\indexlibrarymember{unordered_multimap}{operator<=>}%
```
JohelEGP, line 871:
```suggestion
\indexcont{operator<=>}%
```

JohelEGP (changes requested):

source/containers.tex
JohelEGP: Readded `forward_list`.

JohelEGP: The difference is that `\indexcont` uses `\indexlibrarymemberx`, which only adds the member under the container, rather than both ways.

AlisdairM:
I think we now have two options: the original version of the PR, or the current state where I use a local macro to index all operations with a grouped text-name, while retaining the correct indexing in each container.

@tkoeppe @jensmaurer @jwakely could you express a preference?

JohelEGP (changes requested):

jensmaurer:
The current state with `\indexcont` seems what we do elsewhere in the container requirements sections, so I'd say "squash and merge" as-is.

tkoeppe:
@jwakely Any thoughts?

source/containers.tex
```diff
@@ -856,6 +856,26 @@
 then the operations described below
 are implemented by constexpr functions.
 
+% Local command to index names as members of all containers.
```
jwakely, line 859: I don't have a preference on which way to do this, but I would prefer:

```suggestion
% Local command to index a name as a member of all containers.
```

PR #6224, opened 2023-04-07
[format.string.std] Replace "Derived Extracted Property" with simply "property"

cpplearner:
"Derived Extracted Property" refers to the fact that the data for the `East_Asian_Width` property are extracted into the file [`DerivedEastAsianWidth.txt`]. It is not really a terminology.

[`DerivedEastAsianWidth.txt`]: https://www.unicode.org/Public/UCD/latest/ucd/extracted/DerivedEastAsianWidth.txt
[`EastAsianWidth.txt`]: https://www.unicode.org/Public/UCD/latest/ucd/EastAsianWidth.txt

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -874,8 +874,7 @@ literals, and alternative tokens containing alphabetic characters.
 \indextext{name!length of}%
 \indextext{name}%
 \begin{note}
-The character properties XID_Start and XID_Continue are Derived Core Properties
-as described by \UAX{44} of the Unicode Standard.
+The character properties XID_Start and XID_Continue are described by \UAX{44} of the Unicode Standard.
 \begin{footnote}
 On systems in which linkers cannot accept extended
 characters, an encoding of the \grammarterm{universal-character-name} can be used in
--- a/source/text.tex
+++ b/source/text.tex
@@ -6352,7 +6352,7 @@ The following code points have a field width of 2:
 \begin{itemize}
 \item
 any code point with the \tcode{East_Asian_Width="W"} or
-\tcode{East_Asian_Width="F"} Derived Extracted Property as described by
+\tcode{East_Asian_Width="F"} property as described by
 \UAX{44} of the Unicode Standard
 \item
 \ucode{4dc0} -- \ucode{4dff} (Yijing Hexagram Symbols)


jensmaurer:
@cor3ntin , @tahonermann , any opinions on the above? The rationale sounds plausible.

cor3ntin:
Unicode says 

> When referencing a Unicode character property, it is customary to prepend the word “Unicode” to the name of the property, unless it is clear from context that the Unicode Standard
is the source of the specification.

I think the change make sense but  maybe we should do something similar for http://eel.is/c++draft/lex.name#1

jensmaurer:
@cor3ntin , what do you want to change in http://eel.is/c++draft/lex.name#1 ? The "Derived Core  Properties" mention is in an note, explaining XID_Start and XID_Continue. That seems reasonable to me.

cor3ntin:
Maybe something like
The properties XID_Start and XID_Continue are described by UAX #44 of the Unicode Standard.

Ie the fact that they are derived is not really adding any information

tkoeppe:
@cpplearner Could you please have another look?

cpplearner:
I agree that the mention of "Derived Core Properties" is unneeded. But it's slightly different from the case with East_Asian_Width, which is *not* a derived property, and it's *wrong* to describe it as such.

tkoeppe:
@tahonermann Do you mean as a separate, unrelated change?

tahonermann:
> @tahonermann Do you mean as a separate, unrelated change?

@tkoeppe, preferably as part of this change since it is somewhat related.

tkoeppe:
@tahonermann Ah, OK, but it's not my pull request :-) We'll have to ask @cpplearner.

cpplearner:
I'm unable to make the change due to network issues, and since this PR has "allow edits by maintainers" checked, maintainers can push to my branch as they wish :)

cpplearner:
I finally managed to commit the change.

Ping

jensmaurer (approved):
Works for me.

cor3ntin:
LGTM

tahonermann:
Ship it!

cpplearner:
rebased.

ping @tkoeppe

jwakely (approved):

jensmaurer (approved):
Let's do that.

PR #6225, opened 2023-04-07
[expr.shift] more precise wording for arithmetic shift rounding CWG2724

Eisenwave:
## Problems with the present wording

> rounded down

is sometimes used to describe rounding towards zero, and sometimes to describe rounding towards negative infinity. For example, Java uses [java.math.RoundingMode.DOWN](https://docs.oracle.com/javase/7/docs/api/java/math/RoundingMode.html#DOWN) to describe rounding towards zero.

It is intuitively clear to the committee and to implementers that this should mirror the semantics of the "arithmetic right shift" instruction (which rounds towards negative infinity), but the wording is not so clear.

## Improvements made by the suggested wording

> rounded towards negative infinity

is unambiguous.
The wording is also quite popular, e.g. [the Wikipedia article on rounding](https://en.wikipedia.org/wiki/Rounding) uses it.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -6302,7 +6302,7 @@ vacated bits are zero-filled.
 \end{note}
 
 \pnum
-The value of \tcode{E1 >> E2} is $\tcode{E1} / 2^\tcode{E2}$, rounded down.
+The value of \tcode{E1 >> E2} is $\tcode{E1} / 2^\tcode{E2}$, rounded towards negative infinity.
 \begin{note}
 \tcode{E1} is right-shifted \tcode{E2} bit positions.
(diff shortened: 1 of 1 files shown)


Eisenwave:
I've noticed this needs a fixed commit message when merging.

tkoeppe:
Thanks for raising this, but I think this should be treated as a library issue, and not editorially. Could you please file one?

jensmaurer:
@tkoeppe, why would this be a library issue? It's certainly affecting CWG wording.

jensmaurer:
@Eisenwave , if you want your real name to appear in the core issue, please either change your github settings or send it to me via e-mail.

jensmaurer:
CWG2724

PR #6230, opened 2023-04-11
[coro.generator] Editorial fixes

Quuxplusone:
Iterators can't have been "previously obtained from" a by-value parameter; they must have been obtained elsewhere (e.g. from `x` and then `other` was move-constructed from `x`, so now we have iterators into `x`).

`std::exchange(x, nullptr)` for `unique_ptr x` is equivalent to `std::move(x)`.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -15835,11 +15835,11 @@ generator(generator&& other) noexcept;
 Initializes \exposid{coroutine_} with
 \tcode{exchange(other.\exposid{coroutine_}, \{\})} and
 \exposid{active_} with
-\tcode{exchange(\brk{}other.active_, nullptr)}.
+\tcode{std::move(\brk{}other.active_)}.
 
 \pnum
 \begin{note}
-Iterators previously obtained from \tcode{other} are not invalidated;
+Iterators previously into \tcode{other} are not invalidated;
 they become iterators into \tcode{*this}.
 \end{note}
 \end{itemdescr}
@@ -15889,7 +15889,7 @@ swap(@\exposid{active_}@, other.@\exposid{active_}@);
 
 \pnum
 \begin{note}
-Iterators previously obtained from \tcode{other} are not invalidated;
+Iterators previously into \tcode{other} are not invalidated;
 they become iterators into \tcode{*this}.
 \end{note}
 \end{itemdescr}


source/ranges.tex
```diff
@@ -15835,11 +15835,11 @@
 Initializes \exposid{coroutine_} with
 \tcode{exchange(other.\exposid{coroutine_}, \{\})} and
 \exposid{active_} with
-\tcode{exchange(\brk{}other.active_, nullptr)}.
+\tcode{std::move(\brk{}other.active_)}.
```
jensmaurer, line 15838: I think it is beneficial to keep the consistency of always using `exchange` when defining a move constructor.

source/ranges.tex
```diff
@@ -15835,11 +15835,11 @@
...
 
 \pnum
 \begin{note}
-Iterators previously obtained from \tcode{other} are not invalidated;
+Iterators previously into \tcode{other} are not invalidated;
```
jensmaurer, line 15842: The new phrasing doesn't read like English to me. Also, `other` is an rvalue reference, not a value, so it seems iterators could have been obtained from `other`.

jensmaurer (changes requested):
There's an overfull `\hbox`.

source/ranges.tex
Quuxplusone: I agree that "iterators (previously) into X" is a little vague. Other places (e.g. https://eel.is/c++draft/associative.reqmts.general#115.sentence-2 ) use "iterators pointing to elements of X", but we can't do that here because generators don't have elements. The phrase "iterators into X" is used in https://eel.is/c++draft/sequence.reqmts#37.sentence-3 , as well as on the left-hand side of this diff.

This note could be updated to say "this operation does not modify any iterator `it` in the program, regardless of whether `it.coroutine_ == other.coroutine_` before the operation," but that's _obviously_ tautological, as opposed to being _subtly_ tautological. I think it would also be reasonable to just kill this note entirely.

Re `other` here (in the move-constructor) being an rvalue reference: True, I updated this wording _only_ for consistency with the wording in `operator=`, where `other` _is_ a value (not a reference). I think if the notes are kept, they should be kept parallel, and not with different wording in different places.

source/ranges.tex
Quuxplusone: > I think it is beneficial to keep the consistency of always using exchange when defining a move constructor.

This is the only place in "ranges.tex" where `exchange(` is mentioned. Every other place where we define a move-constructor, we use `std::move`; e.g. https://eel.is/c++draft/range.elements.iterator#5.sentence-1 , https://eel.is/c++draft/range.elements#sentinel-2.sentence-1 , https://eel.is/c++draft/range.take.while.sentinel#2 , https://eel.is/c++draft/range.adjacent.sentinel#2 , ...

I'm not familiar with any other place in the standard that uses `exchange` when it doesn't need to. `git grep 'Initializes.*exchange'` has zero hits. `git grep '\(^\|with \).tcode.exchange'` has only three hits, all in this paragraph.

PR #6234, opened 2023-04-17
[range.repeat.iterator] Remove redundant period

hewillk:


jensmaurer (approved):

JohelEGP:
That's worthy of a check, whether outside or inside `\tcode`. The same for `;` outside of `\tcode`. Even if it's followed by more wording, it could be confusing.

PR #6235, opened 2023-04-19
One \grammarterm{x}, two \grammarterm{x}{s} (and likewise)

adr26:
There was inconsistent use of grouping after `\grammarterm`, `\tcode`, `\keyword`, etc.

Update plurals to consistently use `{s}`/`{es}` after these.

Also fix a missing apostrophe in [dcl.init.aggr]/16, example 11.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -7884,9 +7884,9 @@ return ranges::inplace_merge(ranges::begin(r), middle, ranges::end(r), comp, pro
 
 \pnum
 Subclause \ref{alg.set.operations} defines all the basic set operations on sorted structures.
-They also work with \tcode{multiset}s\iref{multiset}
+They also work with \tcode{multiset}{s}\iref{multiset}
 containing multiple copies of equivalent elements.
-The semantics of the set operations are generalized to \tcode{multiset}s
+The semantics of the set operations are generalized to \tcode{multiset}{s}
 in a standard way by defining \tcode{set_union}
 to contain the maximum number of occurrences of every element,
 \tcode{set_intersection} to contain the minimum, and so on.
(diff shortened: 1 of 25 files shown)


tkoeppe:
Hm, I'm not sure what the benefit of these changes is, or if there is any valuable notion of consistency at stake here. I'll happily take the missing apostrophe fix, but the rest seems to lack motivation. In particular, I don't think `\tcode{foo}{s}` is a useful spelling, compared to `\tcode{foo}s`.

jensmaurer:
Once upon a time, there was a spacing problem with the`\grammarterm` macro that required the trailing `{s}` to avoid bad behavior. The spacing problem has been fixed, and we should gravitate towards plain `s`.

adr26:
> Once upon a time, there was a spacing problem with the`\grammarterm` macro that required the trailing `{s}` to avoid bad behavior. The spacing problem has been fixed, and we should gravitate towards plain `s`.

I was worried about that, given the lack of consistency I'd spotted. If the spacing problem has been fixed, do you think it would be worth doing the reverse of the original changes and change {s} -> s throughout?

If the spacing problem is fixed, you could make the argument it was unnecessary change, but consistency is a virtue in its own right... I am happy to be guided and either leave this as-is, or open a new PR, depending on your thoughts.

Thanks for getting back to me,

Andrew R

tkoeppe:
I see neither a defect nor any value in changing anything here.

If you have spare time, we could definitely use some help in either triaging issues, or in adopting orphaned, old pull requests!

JohelEGP:
> Once upon a time, there was a spacing problem with the`\grammarterm` macro that required the trailing `{s}` to avoid bad behavior. The spacing problem has been fixed, and we should gravitate towards plain `s`.

Is that still the case for `{'s}`?
#6328 introduces some of those.

PR #6244, opened 2023-05-05
[container.alloc.reqmts] Fix incorrect change of \mandates to \expects

jwakely:
This was incorrectly changed from a \mandates to an \expects when 93ff092d1cd2b335f372b9546365b3d495caf2d8 replaced the requirements tables.

https://github.com/cplusplus/draft/commit/93ff092d1cd2b335f372b9546365b3d495caf2d8#diff-3b5851f7056b7c68cb3ba2ab51cf0b75a780c6a11e5e181770067700454ace7bL734 (click on "Load diff") shows the previous state, with `\mandates`, and https://github.com/cplusplus/draft/commit/93ff092d1cd2b335f372b9546365b3d495caf2d8#diff-3b5851f7056b7c68cb3ba2ab51cf0b75a780c6a11e5e181770067700454ace7bR1005 shows the new state, with `\expects`. That change was supposed to be editorial, but this is a normative change.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -1031,7 +1031,7 @@ typename X::allocator_type
 \tcode{A}
 
 \pnum
-\expects
+\mandates
 \tcode{allocator_type::value_type} is the same as \tcode{X::value_type}.
 \end{itemdescr}
 


jwakely:
It looks like that was the only `\mandates` in the container requirements tables, and the only one removed by the commit.

tkoeppe:
Thanks a lot for spotting this!

jwakely:
Thanks to @dani-hub for pointing out that the current draft said *Preconditions* there.

tkoeppe:
Nothing escapes the watchful eye of @Dani-Hub -- thank you!

PR #6246, opened 2023-05-07
[flat.set.modifiers] `(first, last)` should be `rg`

Quuxplusone:
Restore consistency with [[flat.map.modifiers]/13](https://eel.is/c++draft/flat.map.modifiers#13), which already uses `ranges::distance(rg)`.

However, @jwakely can I also get some LWG guidance here?  Right now, there is no `insert_range` in [flat.multiset.modifiers](https://eel.is/c++draft/flat.multiset.modifiers), and there is no [flat.multimap.modifiers](https://eel.is/c++draft/flat.multimap.modifiers) at all. My intuition is that this `insert_range` wording should appear in both places.

I'm planning P2767R0 "flat_foo omnibus" for Varna. Should I submit a pull-request _before_ Varna to add a section named [flat.multimap.modifiers], with contents similar to [flat.map.modifiers]? Should I just make P2767 introduce all of [flat.multimap.modifiers]? Or does LWG think it's "obvious" what `flat_multimap::insert_range` should do, and therefore it shouldn't be documented explicitly? See the [index of library names](https://eel.is/c++draft/libraryindex#lib:insert_range) for the status quo.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -16955,7 +16955,7 @@ from each group of consecutive equivalent elements.
 \pnum
 \complexity
 $N$ + $M \log M$, where $N$ is \tcode{size()} before the operation and $M$
-is \tcode{distance(first, last)}.
+is \tcode{ranges::distance(rg)}.
 
 \pnum
 \remarks


jensmaurer (approved):
Given that first, last are not even in-scope identifiers here, this seems good to me.

@jwakely ?

jensmaurer:
@Quuxplusone , if you write a paper introducing a section of new text, please don't post a pull request doing the same.

jwakely (approved):
The change is correct. This is just a copy&paste error in the incoming paper (P1222R4), reusing the text from the `insert(InputIterator first, InputIterator last)` function in paragraph 6.

jwakely:
> Should I submit a pull-request _before_ Varna to add a section named [flat.multimap.modifiers], with contents similar to [flat.map.modifiers]?

No, that would not be editorial.

jwakely:
> Right now, there is no `insert_range` in [flat.multiset.modifiers](https://eel.is/c++draft/flat.multiset.modifiers), and there is no [flat.multimap.modifiers](https://eel.is/c++draft/flat.multimap.modifiers) at all. My intuition is that this `insert_range` wording should appear in both places.

IIRC it's intentional that we don't repeat redundant wording:

>  Except as otherwise noted, operations on `flat_multimap` are equivalent to those of `flat_map`

>  Descriptions are provided here only for operations on `flat_multiset` that are not described in one of the
general sections or for operations where there is additional semantic information.

It would be better if we used similar wording for those two places, but I don't think we need to duplicate specifications if there is no difference. We have been doing something similar for `unique_ptr<T>` and `unique_ptr<T[]>` for 10+ years.

PR #6251, opened 2023-05-12
[intro.execution] Fix bad function call in example

jensmaurer:
`S` does not have a default constructor.


shafik:
Thank you for fixing this!

PR #6254, opened 2023-05-17
[ranges] Make the "Effects: Equivalent to" format consistent

hewillk:
- When "_Effects_: Equivalent to" is followed by a " ", the subsequent expression ends with a ".";
- When "_Effects_: Equivalent to" is followed by a ":", the subsequent expression ends with a ";".

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -8320,7 +8320,7 @@ constexpr iterator_t<V> base() const;
 \begin{itemdescr}
 \pnum
 \effects
-Equivalent to \tcode{return \exposid{cur_};}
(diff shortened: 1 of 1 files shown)


JohelEGP (approved):

JohelEGP:
See <https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#writing-effects-in-a-function-description> for the specifics.
It's actually about expression vs. statement.

hewillk:
[[input.output]](https://eel.is/c++draft/input.output) also has this inconsistency problem, and I will fix that with a new PR.

jensmaurer (approved):

PR #6255, opened 2023-05-17
Fix remaining references to container requirements, and adjust subclause naming

### LINKED ISSUE #6184: [std] Update references to [container.requirements.general]
The entirety of [container.requirements.general] is a list of items of the form
> - _X_ denotes ...

yet there are many references to this subclause, which probably need updating. For example
```terminal
strings.tex-\tcode{InputIterator} is a type that qualifies as an input
strings.tex:iterator\iref{container.requirements.general}.
strings.tex-
```

jwakely:
Related to #6184

I missed the [re.results.general] one in #6253 because I searched for `{container.requirements.general}`. The [string.cons] one is to the top-level of container requirements, but should be to the sequence requirements.

After fixing those, we can retitle/relabel some of the container requirements subclauses, as discussed in #6184

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -69,12 +69,12 @@ means, for example, that a node-based container would need to construct nodes co
 aligned buffers and call \tcode{construct} to place the element into the buffer.
 \end{note}
 
-\rSec2[container.gen.reqmts]{General containers}
+\rSec2[container.requirements.general]{General containers}
 
-\rSec3[container.requirements.general]{General}
+\rSec3[container.intro.reqmts]{Introduction}
 
 \pnum
-In subclause \ref{container.gen.reqmts},
+In subclause \ref{container.requirements.general},
 \begin{itemize}
 \item
 \tcode{X} denotes a container class containing objects of type \tcode{T},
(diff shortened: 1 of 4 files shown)


jwakely:
Would it be better to say 

> the requirements of an allocator-aware container (24.2.2.5) and of a sequence
container (24.2.4) 

instead of:

> the requirements of an allocator-aware container and of a sequence
container (24.2.2.5, 24.2.4) 

?

jwakely:
That would be consistent with [stacktrace.basic.overview], which says:

> The class template `basic_stacktrace` satisfies the requirements of a reversible container (24.2.2.3), of an
allocator-aware container (24.2.2.5), and of a sequence container (24.2.4), except that

source/containers.tex
```diff
@@ -69,12 +69,12 @@
 aligned buffers and call \tcode{construct} to place the element into the buffer.
 \end{note}
 
-\rSec2[container.gen.reqmts]{General containers}
```
tkoeppe, line 72: If the stable label already existed in C++23, we should add an entry to the stable label changes.

jwakely: So add this to xrefdeltas.tex?

```
\movedxref{container.requirements.general}{container.intro.reqmts}
\movedxref{container.gen.reqmts}{container.requirements.general}
```

tkoeppe: Yes, that's right!

jwakely: OK, added and rebased

tkoeppe: Continue rebasing until morale improves.

jwakely: incorrect entries in xrefdelta.tex:
spurious container.requirements.general

PR #6257, opened 2023-05-28
[unord.hash] [format.formatter.spec] Rephrase "specialization of X<Y>"

Quuxplusone:
Both instances here have the same form: The first paragraph is phrased correctly as "for any type Y for which the library provides a specialization of X," and talks about disabling; the second paragraph talks about enabling in terms of "a specialization of X<Y>," which is incorrect. Flip it around so we talk enabling first, and use parallel construction in the two paragraphs.

Question: Can we replace the words "explicit or partial specialization" with the one word "specialization"? Would that be wrong or confusing somehow?

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -13749,14 +13749,17 @@ disabled specializations do not.
 Each header that declares the template \tcode{hash}
 provides enabled specializations of \tcode{hash} for \tcode{nullptr_t} and
 all cv-unqualified arithmetic, enumeration, and pointer types.
+
+\pnum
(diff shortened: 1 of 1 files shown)


JohelEGP:
> Question: Can we replace the words "explicit or partial specialization" with the one word "specialization"? Would that be wrong or confusing somehow?

Yeah. "Specialization" doesn't mean explicit specialization or partial specialization. It's defined at <https://eel.is/c++draft/temp.spec.general#def:specialization>.

PR #6263, opened 2023-05-30
[mdspan.mdspan.cons] `OtherIndexType` is not a parameter pack ([mdspan.mdspan.cons]/8.2)

JMazurkiewicz:


jwakely (approved):

jwakely:
Even if it were a pack, this would be wrong because it would need to use `is_nothrow_constructible_v` instead.

tkoeppe:
Aren't we missing the `_v` in any case? @jwakely?

jwakely:
Yes indeed - not sure how I missed that despite commenting on it :facepalm:

tkoeppe:
@jwakely PTAL, like so?

jwakely (approved):

PR #6269, opened 2023-06-07
[expr.add] Clarify note on pointer subtraction

### LINKED ISSUE #6268: [expr.add] Pointer subtraction when P-Q is out of range of ptrdiff_t is unclear
> Otherwise, if P and Q point to, respectively, array elements *i* and *j* of the same array object x, the expression P - Q has the value *i - j*.

When *i - j* is not in range of `ptrdiff_t`, this seems to be undefined behaviour due to [expr.pre]. The note on 5.3 should probably be moved here, and a reference to [expr.pre] added.

vasama:
Fixes #6268.

No changes to normative wording.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -6241,13 +6241,13 @@ the result is 0.
 array elements $i$ and $j$
 of the same array object \tcode{x},
 the expression \tcode{P - Q} has the value $i - j$.
-\item Otherwise, the behavior is undefined.
 \begin{note}
 If the value $i - j$
 is not in the range of representable values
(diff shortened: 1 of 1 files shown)


jensmaurer (approved):

jensmaurer:
@vasama, if you want to have your real name show up in the commit history, please configure it in the github settings.

jensmaurer:
@vasama, that's enough. When squash-merging your commit, github will put in your real name in the "new" commit.

jensmaurer:
@tkoeppe , this moves a note one bullet up to better fit the context.

PR #6322, opened 2023-06-18
P2752R3 Static storage for braced initializers

### LINKED ISSUE #6284: [2023-06 CWG Motion 9] P2752R3 Static storage for braced initializers
https://wiki.edg.com/pub/Wg21varna/StrawPolls/P2752R3.html

jensmaurer:
Fixes #6284
Fixes cplusplus/papers#1442

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -1,6 +1,36 @@
 %!TEX root = std.tex
 \infannex{diff}{Compatibility}
 
+\rSec1[diff.cpp23]{\Cpp{} and ISO \CppXXIII{}}
+
+\rSec2[diff.cpp23.general]{General}
+
+\pnum
+\indextext{summary!compatibility with ISO \CppXXIII{}}%
+Subclause \ref{diff.cpp23} lists the differences between \Cpp{} and
+ISO \CppXXIII{} (ISO/IEC 14882:2023, \doccite{Programming Languages --- \Cpp{}}),
+by the chapters of this document.
+
+\rSec2[diff.cpp23.expr]{\ref{expr}: expressions}
+
+\diffref{dcl.init.list}
+\change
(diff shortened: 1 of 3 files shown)


cpplearner:
Does it make sense to bump the `__cpp_initializer_lists` feature-test macro?

source/compatibility.tex
```diff
@@ -1,6 +1,36 @@
...
+\effect
+Valid \CppXXIII{} code
+that relies on the result of pointer comparison between backing arrays
+may change behavior.
+For example:
+\begin{example}
+bool ne(std::initializer_list<int> a, std::initializer_list<int> b) {
+  return a.begin() != b.begin() + 1;
+}
+bool b = ne({2,3}, {1,2,3}); // unspecified result; previously false
```
JohelEGP, line 31:
```suggestion
bool b = ne({2,3}, {1,2,3}); // unspecified result; previously \tcode{false}
```

JohelEGP (changes requested):
I'm using <https://isocpp.org/files/papers/D2752R3.html> to pre-review
(there are sufficient differences that I'll need to wait for the mailing to confirm).

jensmaurer:
> Does it make sense to bump the `__cpp_initializer_lists` feature-test macro?

This is a pure optimization enabler; I don't think it warrants a "feature" announcement.

JohelEGP (approved):
> (there are sufficient differences that I'll need to wait for the mailing to confirm).

It was actually alright.
The wording in the generated PDF looks as was proposed.
But it seems there was a change in the context
and the paper's PR wasn't rebased.

tkoeppe:
It seems that the paper is just presenting the status quo in a rather loose and lacunary fashion. The PR looks good. Thanks for checking!

PR #6350, opened 2023-06-24
[LWG motion 21] P2637R3 Member visit

### LINKED ISSUE #6307: [2023-06 LWG Motion 21] P2637R3 Member visit

burblebee:
Fixes #6307.

Notes/questions:
* Is there any order to the sections in [depr]?

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -5022,6 +5022,12 @@ namespace std {
 
     // \ref{variant.swap}, swap
     constexpr void swap(variant&) noexcept(@\seebelow@);
+
+    // \ref{variant.visit}, visitation
+    template<class Self, class Visitor>
+      constexpr decltype(auto) visit(this Self&&, Visitor&&);
+    template<class R, class Self, class Visitor>
+      constexpr R visit(this Self&&, Visitor&&);
   };
 }
 \end{codeblock}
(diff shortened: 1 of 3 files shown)


JohelEGP:
> Is there any order to the sections in [depr]?

Considering the introduction,
I'd hope it mimics the main text.

JohelEGP (changes requested):
I'm using <https://isocpp.org/files/papers/P2637R3.html>.

I think you need to add some `\movedxref`s to `xrefdelta.tex`.

[3.1](https://isocpp.org/files/papers/P2637R3.html#feature-test-macro) wasn't applied.

burblebee:
> I think you need to add some `\movedxref`s to `xrefdelta.tex`.

Don't see where that's needed; don't see any moved sections.

> [3.1](https://isocpp.org/files/papers/P2637R3.html#feature-test-macro) wasn't applied.

Applied missed edit 3.1, thanks!

JohelEGP (approved):

source/utilities.tex
```diff
@@ -6140,6 +6146,44 @@
...
+
+\pnum
+\constraints
+The call to \tcode{visit} does not use
+an explicit \grammarterm{template-argument-list} that
+begins with a type \grammarterm{template-argument}.
+
+\pnum
+\effects
+Equivalent to: \tcode{return std::visit(std::forward<Visitor>(vis), (V)self);}
```
tkoeppe, line 6168: Is the use of `std::visit` here deliberate/important? Usually we don't qualify names within `std`.

CaseyCarter: Without qualification, `visit` here would refer to the member function template `std::variant::visit` instead of the namespace-scope template `std::visit`, making the equivalent code ill-formed.

tkoeppe: Ah of course, thanks!

PR #6361, opened 2023-06-27
[LWG motion 16] P1901R2 Enabling the Use of weak_ptr as Keys in Unordered Associative…

### LINKED ISSUE #6302: [2023-06 LWG Motion 16] P1901R2 Enabling the Use of weak_ptr as Keys in Unordered Associative Containers
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p1901r2.html

burblebee:
Fixes #6302.

### DIFF
--- a/source/memory.tex
+++ b/source/memory.tex
@@ -551,6 +551,12 @@ namespace std {
   // \ref{util.smartptr.ownerless}, class template \tcode{owner_less}
   template<class T = void> struct owner_less;
 
+  // \ref{util.smartptr.owner.hash}, struct \tcode{owner_hash}
+  struct owner_hash;
+
+  // \ref{util.smartptr.owner.equal}, struct \tcode{owner_equal}
+  struct owner_equal;
+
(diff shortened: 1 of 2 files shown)


source/memory.tex
```diff
@@ -4782,13 +4859,100 @@
...
+\begin{itemdescr}
+\pnum
+\returns
+\tcode{x.owner_equal(y)}.
+
+\pnum
+\begin{note}
+\tcode{x.owner_equal(y)} is \tcode{true}
+%FIXME: Who are "they"?
+if and only if they share ownership or are both empty.
```
JohelEGP, line 4952: Looking at the added `shared_ptr::owner_equal` and `weak_ptr::owner_equal`,
```suggestion
if and only if \tcode{x} and \tcode{y} share ownership or are both empty.
```

JohelEGP (changes requested):

JohelEGP (approved):

PR #6363, opened 2023-06-29
[meta.unary.prop.query, meta.trans.arr] Use `static_assert` instead of `assert` in example

hewillk:
Not sure why runtime `assert` was used on [meta] in the first place.


JohelEGP (changes requested):
Please, remove the redundant parentheses.

jwakely:
> Not sure why runtime `assert` was used on [meta] in the first place.

Because the example comes straight from TR1 which was based on C++03 and so there was no `static_assert` in the language.

tkoeppe:
Could you rerun the precommit checks?

jwakely (approved):

PR #6367, opened 2023-07-06
[namespace.std] Convert unconventional (a) and (b) notation to items

Eisenwave:
It would be more appropriate in American English to use a comma in this place.

Without the comma, the sentence reads as:
> [...] depends on the name of at least one program-defined type and [on something else]

With the comma, the sentence reads as:
> [...] depends on the name of at least one program-defined type, and [there is another requirement]

The second reading is correct here, given the rest of the sentence.

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -2837,10 +2837,11 @@ a program may add a template specialization for
 any standard library class template
 to namespace
 \tcode{std} provided that
-(a) the added declaration
-depends on at least one program-defined type
-and
-(b) the specialization meets the standard library requirements
+\begin{itemize}
+\item the added declaration
+depends on at least one program-defined type, and
+
+\item the specialization meets the standard library requirements
 for the original template.
 \begin{footnote}
 Any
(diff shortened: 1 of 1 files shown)


jwakely:
I think the use of "(a) something and (b) something else" makes it unambiguous that the second reading is the only correct one. It's not "depends on (a) something and (b) something else".

That said, the use of "(a) and (b)" here is not how we usually specify things. I think we could either replace those with `\item` to make this a bulleted list (which would still be clear that the "depends on" part only applies to the first requirement), or remove the (a) and (b) and add the comma. I think the bulleted list would be my preference between those two options.

Keeping the (a) and (b) would be my least favourite solution, and with that option adding the comma seems like putting lipstick on a pig.

Eisenwave:
> That said, the use of "(a) and (b)" here is not how we usually specify things. I think we could either replace those with `\item` to make this a bulleted list (which would still be clear that the "depends on" part only applies to the first requirement), or remove the (a) and (b) and add the comma. I think the bulleted list would be my preference between those two options.

I agree. I've implemented this change for the paragraph in question, as well as another paragraph above:

grafik

jensmaurer (approved):

jensmaurer:
@Eisenwave , feel free to squash and force-push the two commits.

jwakely (approved):

PR #6368, opened 2023-07-06
[basic.def.odr] Add serial comma

Eisenwave:
It would be more appropriate in American English to use a comma in this place.

Without the comma, the sentence reads as:
> [...] is attached to a named module and [to something else]

With the comma, the sentence reads as:
> [...] is attached to a named module, and [there is another requirement]

The second reading is correct here, given the rest of the sentence.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -632,7 +632,7 @@ do not satisfy the following requirements,
 \end{itemize}
 the program is ill-formed;
 a diagnostic is required only
-if the definable item is attached to a named module and
+if the definable item is attached to a named module, and
 a prior definition is reachable at the point where a later definition occurs.
 Given such an item,
 for all definitions of \tcode{D}, or,


jwakely:
This isn't strictly needed because "atttached to a named module and attached to a prior definition is reachable at the point where a later definition occurs" is meaningless, so only the second reading is possible.

jwakely:
That is, I think without the comma the sentence could potentially be read in two ways (not only in the first way as you imply). But since the first reading is meaningless, it has to be read the second way.

The comma avoids requiring lookahead and tentative parsing to decide which meaning is valid.

Eisenwave:
> The comma avoids requiring lookahead and tentative parsing to decide which meaning is valid.

Yes, I agree; once the sentence is read in full, only one interpretation is meaningful. I've added the comma not to remove ambiguity; it's just a minor stylistic improvement which eliminates that look-ahead.

PR #6369, opened 2023-07-06
[filebuf.virtuals] fix "if width if less than zero" typo

jwakely:
This typo has been present since C++98.


jensmaurer (commented):
Good catch!

PR #6382, opened 2023-07-16
[dcl.init] Fix brace spacing inconsistencies

Eisenwave:
[dcl.init] involves many uses of list initialization, and the style isn't consistent. Sometimes it isn't even consistent within the same example.

This PR makes the style consistent within the same section, where the chosen style is:
- spaces after `{` and before `}`
- empty brace pairs formatted as `{}`
- direct list initialization formatted as `T var{ a, b, c };`
- copy list initialization formatted as `T var = { a, b, c };`

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -5112,9 +5112,9 @@ struct A;
 extern A a;
 struct A {
   const A& a1 { A{ a, a } };    // OK
-  const A& a2 { A{ } };          // error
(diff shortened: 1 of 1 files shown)


jensmaurer:
In general, the core language examples exhibit a variety of styles to highlight the fact that C++ does not impose a particular way of formatting.

Odd duplicate spaces certainly seem to be in the "typo" arena, but other formatting should be left as-is.

PR #6383, opened 2023-07-16
[time.zone.leap.overview] Fix example

hewillk:
`leap_seconds` can only be compared with `sys_time<Duration>`, so we are in nondeduced context here? 

Correct me if I'm wrong.

### DIFF
--- a/source/time.tex
+++ b/source/time.tex
@@ -10133,7 +10133,7 @@ are constructed and stored in the time zone database when initialized.
 \begin{example}
 \begin{codeblock}
 for (auto& l : get_tzdb().leap_seconds)
-  if (l <= 2018y/March/17d)
+  if (l <= sys_days(2018y/March/17d))
     cout << l.date() << ": " << l.value() << '\n';
 \end{codeblock}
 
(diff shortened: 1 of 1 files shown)


tkoeppe:
Does the example compile?

hewillk:
CC @HowardHinnant

HowardHinnant:
The example as corrected compiles.  The uncorrected code does not.

HowardHinnant:
nitpick:  I use `{}` for construction when possible to disambiguate it visually from a function call which uses `()`.

source/time.tex
```diff
@@ -10133,7 +10133,7 @@
 \begin{example}
 \begin{codeblock}
 for (auto& l : get_tzdb().leap_seconds)
-  if (l <= 2018y/March/17d)
+  if (l <= sys_days(2018y/March/17d))
```
tkoeppe, line 10136:
```suggestion
  if (l <= sys_days{2018y/March/17d})
```

jwakely (approved):
I don't think we need to follow that specification policy in examples, and this conversion is guaranteed to work without narrowing so the braces are fine.

PR #6388, opened 2023-07-18
[priqueue.members] [queue.mod] [stack.mod] Harmonize push_range wording

Quuxplusone:
Discussion of P2767 raised the point that `forward` here is redundant (ranges are take-by-forwarding-reference but use-by-lvalue, since they can't possibly dangle at this point). Be consistent. Also, harmonize the English wording of all three container adaptors.

The flat container adaptors' `insert_range` functions are deliberately excluded from this editorial patch, because this harmonization would _not_ be editorial for them; they will be brought into line by [P2767 section 5](https://quuxplusone.github.io/draft/d2767-flat-omnibus.html#set-insert-range) once LWG has re-reviewed it.

(Attn @CaseyCarter @var-const)

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -13630,9 +13630,9 @@ template<@\exposconcept{container-compatible-range}@<T> R>
 \begin{itemdescr}
 \pnum
 \effects
-Equivalent to \tcode{c.append_range(std::forward<R>(rg))}
-if that is a valid expression,
-otherwise \tcode{ranges::copy(rg, back_inserter(c))}.
+Adds the elements of \tcode{rg} to \tcode{c}
+via \tcode{c.append_range(rg)} if that is a valid expression,
+or \tcode{ranges::copy(rg, back_inserter(c))} otherwise.
 \end{itemdescr}
 
 \rSec3[queue.ops]{Operators}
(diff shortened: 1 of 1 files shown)


source/containers.tex
```diff
@@ -13630,9 +13630,9 @@
 \begin{itemdescr}
 \pnum
 \effects
-Equivalent to \tcode{c.append_range(std::forward<R>(rg))}
-if that is a valid expression,
-otherwise \tcode{ranges::copy(rg, back_inserter(c))}.
+Adds the elements of \tcode{rg} to \tcode{c}
```
CaseyCarter, line 13633: For other reviewers: `c.append_range(rg)` is equivalent to `c.append_range(std::forward<R>(rg))` since `append_range` for a standard container doesn't care about the value category of its `input_range` argument.

Dropping the "Equivalent to" Words of Power is non-editorial, however, which I failed to point out when we discussed this change earlier. "via" doesn't have the same fully-specified meaning. Could we rephrase this as `Equivalent to \tcode{c.append_range(rg)} if that is a valid expression, otherwise equivalent to \tcode{ranges::copy(rg, back_inserter(c))}.`?

Quuxplusone: > Dropping the "Equivalent to" Words of Power is non-editorial

I was afraid of that. I do want to point out that the current wording is "Equivalent to ..." in 2 places and "Inserts ... via ..." in the 3rd place, which is again bizarrely inconsistent; and IIUC we can't harmonize all 3 in _either_ direction without one of the changes being non-editorial in this sense. I can bring this back in a real paper if we need me to.

Once I'm not trying to be editorial, I'll go even further and suggest replacing the word "via" with "as if by".

PR #6390, opened 2023-07-21
[version.syn] Fix header references.

Eelis:


jwakely (approved):

Eelis:
Amended to fix another header reference in the same section.

jwakely (approved):

jensmaurer (approved):

JohelEGP:
Can't we have `\libheader` check that it uses an existing reference, like `\ref` does?

PR #6391, opened 2023-07-21
[version.syn] Fix <cstdlib> header reference.

Eelis:


Eelis:
Integrated into #6390.

PR #6394, opened 2023-07-21
Replace cross-references from C++20 with cross-references from C++23.

tkoeppe:


source/xrefprev
```diff
@@ -1,2764 +1 @@
-\glossaryentry{intro.scope@ {\memgloterm{intro.scope}}{\memglodesc{(\ref  {intro.scope})}} {\memgloref{}}|memjustarg}{1}
```
jensmaurer, line 1: The wholesale removal here is wrong.

This must be a copy of xrefindex.glo (NOT: xrefdelta.glo) at the point of C++23.

jensmaurer (changes requested):
@tkoeppe, mostly good except for a fatal issue.

source/xrefprev
tkoeppe: Ah, I see, thanks!

jensmaurer (approved):

PR #6402, opened 2023-07-23
[input.output] Fix non-compliance Effects: Equivalent to format

hewillk:
This is the follow-up of #6254.
Not sure if this is 100% spec-compliant, corrections are welcome.


jensmaurer (approved):

PR #6410, opened 2023-07-24
[basic.fundamental] Clarify that table of minimum integral type widths applies only to standard integral types

randomnetcat:
Per https://github.com/cplusplus/CWG/issues/365#issuecomment-1637162904

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -4970,7 +4970,7 @@ the largest value of the corresponding unsigned type.
 \end{floattable}
 
 \pnum
-The width of each signed integer type
+The width of each standard signed integer type
 shall not be less than the values specified in \tref{basic.fundamental.width}.
 The value representation of a signed or unsigned integer type
 comprises $N$ bits, where N is the respective width.


tkoeppe:
Why is this editorial?

randomnetcat:
I don't see any alternative reading that makes sense.

I see three potential readings of the status quo:
1. The sentence constrains the widths of only the types listed in the table. This is what I think is the intended reading.
2. The sentence constrains the widths of all integer types, including extended integer types. This is the reading taken at https://github.com/cplusplus/CWG/issues/365#issuecomment-1637161344. I don't think this is intended (if it was, it would be explicitly stated, or there would be another entry in the table read "any other signed integer type" or similar). If this reading applies, it's not clear which minimum is meant to apply to an extended integer type (the linked comment assumes `8` but it could just as well be `64`).
3. The sentence prohibits the existence of extended integer types. This is the reading suggested at https://github.com/cplusplus/CWG/issues/365#issuecomment-1637162904. This is obviously not intended.

Maybe I'm wrong and 2 is what is intended?

tkoeppe:
@jensmaurer Thoughts?

jensmaurer:
The table lists widths of the particular standard signed integer types. It's editorial to clarify the reference to that table (and thus, to this particular list of integer types) to be the "standard signed integer types", because that's what they are per the definition in basic.fundamental p1.

tkoeppe:
Sounds good, thanks!

jensmaurer:
@tkoeppe, let's go for it.

PR #6411, opened 2023-07-24
[conv.ptr], [conv.mem] Remove redundant text on null pointer comparisons

lprv:
(Member) pointer equality is specified in [[expr.eq]](https://eel.is/c++draft/expr.eq).

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -1012,8 +1012,7 @@ distinguishable from every other value of
 object pointer or function pointer
 type.
 Such a conversion is called a \defnx{null pointer conversion}{conversion!null pointer}.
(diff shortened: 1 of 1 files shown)


jensmaurer (approved):
Agreed this is redundant with [expr.eq].

tkoeppe:
@jensmaurer Do you want CWG to have a look at this PR, or are you OK with it as is?

tkoeppe:
@jensmaurer Ping ^^^

jensmaurer:
I think this is fine as-is and doesn't need CWG review.

tkoeppe:
Thanks!

PR #6413, opened 2023-07-24
[over.call.object] Refer to the static type of the object expression

lprv:
'`E` evaluates to a class object' is something that happens at runtime, but the text here is clearly concerned with the static type of the expression.


tkoeppe:
I appreciate the runtime/static confusion, but @jensmaurer to double-check if the proposed new wording. Thanks!

PR #6414, opened 2023-07-24
[dcl.meaning.general] Use 'declarator-id' instead of 'name'

lprv:
A [*declarator-id*](https://eel.is/c++draft/dcl.decl.general#nt:declarator-id) is not necessarily a [name](https://eel.is/c++draft/basic.pre#def:name).


jensmaurer (approved):
(In particular, "name" excludes qualified-ids, but those are in view here.)

PR #6416, opened 2023-07-25
[optional.optional] Remove qualified destructor calls, refactor postconditions of `swap`

lprv:
`val->T::~T()` should be `val->~T()` (the former breaks if `T` has a member named `T`), but considering that '[destroys](https://eel.is/c++draft/dcl.init.general#22) the contained value' is already sufficiently clear in its meaning, I don't see why explicitly specifying the syntax for a (pseudo-)destructor call is necessary at all.

This PR also changes the presentation of postconditions in [[optional.swap]](https://eel.is/c++draft/optional.swap) by moving them out of the 'Effects' table.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -3955,17 +3955,22 @@ See \tref{optional.swap}.
 calls \tcode{swap(*(*this), *rhs)} &
 direct-non-list-initializes the contained value of \tcode{*this}
 with \tcode{std::move(*rhs)},
-then destroys the contained value of \tcode{rhs};
-postcondition is that \tcode{*this} contains a value and \tcode{rhs} does not contain a value \\
+then destroys the contained value of \tcode{rhs} \\
 \rowsep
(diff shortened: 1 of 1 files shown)


JohelEGP (approved):

jwakely:
> `val->T::~T()` should be `val->~T()` (the former breaks if `T` has a member named `T`),

Hmm, GCC and Clang compile that OK, EDG and MSVC don't.

tkoeppe:
Thanks! I'm not sure if this PR is really an improvement. Even moving the postconditions out just creates the combinatorial complexity elsewhere that the table was meant to express concisely.

Please reopen if you feel strongly, but I'd say let's just leave this.

jwakely:
> Hmm, GCC and Clang compile that OK, EDG and MSVC don't.

This is [CWG 1089](https://wg21.link/cwg1089).

PR #6417, opened 2023-07-25
[module.global.frag] Simplify wording

lprv:
* Simplify the wording of [[module.global.frag]/3.5.2](https://eel.is/c++draft/module.global.frag#3.5.2)
* Make the phrasing of 'one of *D* and *M*' consistent across following bullets

### DIFF
--- a/source/modules.tex
+++ b/source/modules.tex
@@ -639,14 +639,13 @@ for which $M$ is decl-reachable from $S$ and either
 \item
 $D$ is decl-reachable from $M$, or
 \item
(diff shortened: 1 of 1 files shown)


tkoeppe:
@opensdh Would you have thoughts on whether this is a useful (and correct) simplification?

opensdh:
Seems right to me; I think the modules wording predates having a good definition of "declare the same entity", as silly as that sounds.

jensmaurer (approved):
Looks good.

tkoeppe:
Thanks, everyone!

PR #6418, opened 2023-07-25
Switch to lualatex and New Computer Modern

tkoeppe:
*This is experimental.*

LuaTeX seems to have come a long way: it now uses harfbuzz for glyph shaping, and most microtype features seem to work with it now. We could consider switching from pdfTeX to LuaTeX.

I see two main options:

1. Just switch to LuaTeX but leave `lmodern`/`fontenc` as is. This provides a very similar result, but also doesn't give us much benefit.
2. Switch to LuaTeX, and use `fontspec` and the New Computer Modern family (which I believe is derived from Latin modern). This is what this PR does. 

**Pros:**

* Better unicode support, access to more scripts and glyphs.
* Better indexing (@godbyk, is that true?).
* Potential access to more fonts, should we care.

**Cons:**

* Slower.
* New Computer Modern doesn't have nearly as many optical sizes as Latin Modern.
* Despite microtype claiming support, protrusions seem less sophisticated than with pdfTeX.
* Hard to be sure if everything is still OK.

If you'd like to weigh in, it would be very useful if you could review the resulting PDF and see if you notice any major defects (e.g. look at some areas that you might recall as being typographically tricky in the past).

We don't have to make a decision any time soon, but it might be worth to start looking into this.


tkoeppe:
> Slower.

Actually quite a  lot slower: the github action build seems to take some 20min now, up from 5min.

PR #5693, opened 2022-07-29
P2093R14 Formatted output

burblebee:
FIxes #5598

PR #6423, opened 2023-07-30
[ostream.formatted.print] Use \impldef macro.

mordante:
The text defines implementation-defined behavior, but it doesn't use the appropriate macro. So the text is not available in "Index of implementation-defined behavior".

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -6838,8 +6838,8 @@ the function initializes an automatic variable via
 string out = vformat(os.getloc(), fmt, args);
 \end{codeblock}
 If the function is \tcode{vprint_unicode} and
-\tcode{os} is a stream that refers to a terminal capable of displaying Unicode
-which is determined in an implementation-defined manner,
+\tcode{os} is a stream that refers to \impldef{a terminal capable of displaying
+Unicode} which is determined in an implementation-defined manner,
 writes \tcode{out} to the terminal using the native Unicode API;
 if \tcode{out} contains invalid code units,
 \indextext{undefined}%


JohelEGP:
See <https://github.com/cplusplus/draft/pull/5693#discussion_r933647093> for how this macro is used.
That comment also appertains to this specific use.
It was rejected back then.

JohelEGP:
I think this is also similar to <https://github.com/cplusplus/draft/pull/6359#discussion_r1245843128>.

mordante:
Thanks for the links, searching didn't show them. After reading the links I'm still not sure why this is not implementation-defined behavior. The way I read it is "Whether `os` refers to a terminal capable of displaying Unicode is determined in an implementation-defined manner". Is this the wrong interpretation?

tkoeppe:
@mordante I think the point is that the macro expands to the words "implementation-defined", and the macro argument is the explanatory text that goes into the index.

You can just build the PDF with your changes and check what you get (both in the text and in the index)!

source/iostreams.tex
```diff
@@ -6838,8 +6838,8 @@
 string out = vformat(os.getloc(), fmt, args);
 \end{codeblock}
 If the function is \tcode{vprint_unicode} and
-\tcode{os} is a stream that refers to a terminal capable of displaying Unicode
-which is determined in an implementation-defined manner,
+\tcode{os} is a stream that refers to \impldef{a terminal capable of displaying
+Unicode} which is determined in an implementation-defined manner,
```
jensmaurer, line 6842: Specifically, this appearance of the word "implementation-defined" must be replaced by `\impldef{text to appear in the index}`

PR #6431, opened 2023-08-01
[intro.races] p13 - add missing serial comma between items

Eisenwave:


jensmaurer (commented):
The "serial comma" is traditionally applied only if there are more than two components of an "and" or "or" chain.  It's "X and Y", but "X, Y, and Z".

This case has only two components, so should be "X and Y" (no comma).

PR #6434, opened 2023-08-02
[intro.execution] "Sequenced before" should be a strict partial order

eisenwave:
A relation which is asymmetric must also be irreflexive. Asymmetry is defined as:
```cpp
A sequenced before B  =>  !(A sequenced before B)
```
If any evaluation **E** was sequenced before itself, this would yield:
```cpp
E sequenced-before E  =>  !(E sequenced before E)
true                  =>  false
```
This is a contradiction of the asymmetry requirement, meaning that *sequenced before* is implied to be irreflexive. Ideally, this should be clearly stated, and it should be stated that it is a [strict partial order](https://en.wikipedia.org/wiki/Partially_ordered_set#Strict_partial_orders) since it fulfills all requirements of one.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -5846,9 +5846,10 @@ by the call (such as the I/O itself) or by the \keyword{volatile} access
 may not have completed yet.
 
 \pnum
-\defnx{Sequenced before}{sequenced before} is an asymmetric, transitive, pair-wise relation between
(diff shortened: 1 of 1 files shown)


Eisenwave:
I initially thought that this is a core issue, but it looks to be purely editorial.

See https://github.com/cplusplus/CWG/issues/387

PR #6437, opened 2023-08-06
[conv.mem] Itemize p2

eisenwave:
This is meant to improve readability and make it possible to cite each bullet separately. This paragraph is quite a massive block of text otherwise.


Eisenwave:
@tkoeppe?

source/expressions.tex
jensmaurer: The null pointer discussion should become the second bullet.

jensmaurer (changes requested):

source/expressions.tex
eisenwave: Done. I've also made each bullet start with "Otherwise".

PR #6440, opened 2023-08-07
[dcl.meaning.general] Remove `extern` in one case to make example 3 more informative

eisenwave:
The paragraph mentions that if the declaration declares a function, or uses the `extern` specifier, then ...

However, the example only shows declarations which are both functions and use the `extern` specifier, making it less informative. This PR removes `extern` in one of the two declarations, to make it clear that `extern` isn't required, and declaring a local function in itself is enough for the paragraph to apply.


tkoeppe:
@opensdh I'd welcome your views on this!

opensdh (approved):

PR #6441, opened 2023-08-07
[basic.def.odr] Fix grammatical error in p17

eisenwave:
The paragraph presently uses:
- singular form in the first half "there is ..."
- plural form in the second half "that have the same first ..."

This PR fixes that grammatical error. Furthermore "there are more than one" is intuitively awkward wording, so it is replaced with "there are multiple".

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -804,8 +804,8 @@ denote the same closure type in each translation unit.
 
 \pnum
 If, at any point in the program,
-there is more than one
-reachable unnamed enumeration definition in the same scope
+there are multiple
+reachable unnamed enumeration definitions in the same scope
 that have the same first enumerator name and
 do not have typedef names for linkage purposes\iref{dcl.enum},
 those unnamed enumeration types shall be the same; no diagnostic required.


tkoeppe:
@opensdh An opportunity to offer further alternative wordings

tkoeppe:
(See also https://english.stackexchange.com/questions/591516/when-is-more-than-one-singular-or-plural .)

opensdh (approved):

Eisenwave:
@tkoeppe?

jwakely:
This use of "multiple" as a determiner meaning more than one is awkward for me. My dictionary does allow it, but defines it as "many, especially more than several". We don't mean "lots and lots" here, we just mean more than one.

source/basic.tex
```diff
@@ -804,8 +804,8 @@
 
 \pnum
 If, at any point in the program,
-there is more than one
-reachable unnamed enumeration definition in the same scope
+there are multiple
```
jwakely, line 807:
```suggestion
there are two or more
```

jwakely:
See https://www.englishspeechservices.com/pronunciation-pieces/multiple-multiplies/ and his recent
https://www.youtube.com/watch?v=D5DCQtz-RaE explaining why multiple doesn't work for me. I suppose you can consider "multiple reachable unnamed enumeration definitions" as aspects of a single type. So it's a type with multiple reachable definitions, but the fact that we're saying what happens if they _aren't_ the same type (IF;NDR) means they aren't necessarily always aspects of a single type.

Anyway, I think "two or more" works fine.

source/basic.tex
Eisenwave: Both your suggestion and my suggestion are grammatically correct. I'm not attached to my wording.

What about the bigger picture though? There are over 120 uses of "multiple" in N5008, and many of them are used in a "two or more" way. For example, [[dcl.init.general] p21](https://eel.is/c++draft/dcl.init#general-21) says:

> The same *identifier* shall not appear in multiple *designator*s of a *designated-initializer-list*.

If "multiple" means "one or more", then `{ .x = 0 }` is ill-formed. If it means "numerous, as in, a lot", then `{ .x = 0, .x = 0 }` is well-formed because two is not that many.

Anyhow, I feel like we should make a decision in the specification guidelines regarding this. If we decide that "multiple" is bad, then I'm happy to change this wording. Otherwise, I'm just reusing an existing wording idiom, and the change would be somewhat arbitrary.

jensmaurer:
If we mean "more than one", saying that feels a lot clearer than talking about "multiple".

Eisenwave:
> If we mean "more than one", saying that feels a lot clearer than talking about "multiple".

Yeah sure, I agree. I'm just curious what we're going to do about the 100 or so other usages of "multiple" that are synonymous with "more than one" in N5008.

We could at least agree in the spec guidelines not to create more instances of "multiple" in favor of less ambiguous quantifiers. Otherwise there's a good chance we'll talk about "multiple" in some other PR in the next weeks or months, with no overarching decision having been made.

AlisdairM:
> This use of "multiple" as a determiner meaning more than one is awkward for me. My dictionary does allow it, but defines it as "many, especially more than several". We don't mean "lots and lots" here, we just mean more than one.

Conversely, "multiple" conveys exactly the right cardinality to me --- for a common point of reference think of the multiple choice exams we have all sat at some point or other.  In those cases multiple was typically 4 or 5, and could be less.

"several" is "multiple with an expected low number of options" where "many" is multiple with a high number of options, but multiple covers the whole spread better than any other word I know, and feels less clumsy than "more than one".

PR #6442, opened 2023-08-07
[basic.link] fix unusual punctuation in p4

eisenwave:
Paragraph 4 breaks convention by excessively using semicolons instead of serial commas. This is also inconsistent with the preceding p3.

Furthermore, even though the latter bullets 4.7 through 4.9 are complete sentences, and usually would be capitalized and punctuated as such elsewhere in the standard, they are also separated by semicolons.

This PR fixes this unconventional and surprising form of capitalization and punctuation.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -2694,35 +2694,35 @@ The name of an entity that belongs to a namespace scope
 that has not been given internal linkage above
 and that is the name of
 \begin{itemize}
-\item a variable; or
-\item a function; or
+\item a variable, or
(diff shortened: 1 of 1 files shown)


tkoeppe:
@Eisenwave Please rebase?

tkoeppe:
Editorial meeting decision:

For the first list, break the longer items into multiple items, and only use commas.

For the second list, turn each item into a separate sentence.

This should remove all the semicolons.

PR #6444, opened 2023-08-07
[class.static.data] Fix - classes have no subobjects

Eisenwave:
To my understanding, a class cannot have any subobjects, only an object of class type can have subobjects. Therefore, the wording is incorrect, and should state `class object` instead of just `class`.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -2658,7 +2658,7 @@ A static member function cannot be qualified with \keyword{const},
 \indextext{member data!static}%
 
 \pnum
-A static data member is not part of the subobjects of a class. If a
+A static data member is not part of the subobjects of a class object. If a
 static data member is declared \keyword{thread_local} there is one copy of
 the member per thread. If a static data member is not declared
 \keyword{thread_local} there is one copy of the data member that is shared by all


tkoeppe:
The original wording seems to have issues, but I'm not sure if this is an improvement. I think a more thorough rewording is needed here.

tkoeppe:
We say "subobject(s) of a class" in multiple places, so let's not make a single change like this. Could you maybe start by looking for all the occurrences. (E.g. in [conv.mem] we say "the subobject of `D`", where `D` is a class.

I think "class object" is generally a phrase we can use, but we'll have to look at each individual case to see how to make it fit.

jensmaurer:
The standard is generally confused about "base class" and "data member" and "subobject" and the relationship among them. (For example, you can have multiple base classes of the same type, that may or may not ("virtual") cause different subobjects to exist in any class object.)

I don't think I'd support applying any cleanup in this area in the context of an editorial issue (as opposed to a core issue).

tkoeppe:
OK, thanks, @jensmaurer! Let's close this issue then. @Eisenwave: please file a CWG issue if you want to pursue this.

PR #6446, opened 2023-08-07
[except.terminate] Fix missing introductory comma

Eisenwave:
There is a missing comma after the conditional clause *"In some situations,"*.

### DIFF
--- a/source/exceptions.tex
+++ b/source/exceptions.tex
@@ -1015,7 +1015,7 @@ capture the currently handled exception.
 \pnum
 \indextext{\idxcode{terminate}}%
 % FIXME: What does it mean to abandon exception handling?
-In some situations exception handling is abandoned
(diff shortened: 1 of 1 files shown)


jensmaurer (approved):

tkoeppe:
"In some situations" isn't really a clause, is it?

Eisenwave:
> "In some situations" isn't really a clause, is it?

You're right, I think it's merely an *introductory phrase*, and one typically uses a comma there.

See https://owl.purdue.edu/owl/general_writing/punctuation/commas/commas_after_introductions.html

tkoeppe:
I'm not convinced that this is a _strong_ argument that the comma is "missing". I agree that it's nice to have the comma, but it was fine without the comma, too. I'll reword the commit message.

PR #6448, opened 2023-08-08
[depr.static.constexpr] Cross-reference core clauses 

AlisdairM:
There is no back-link from the Annex D wording to the  feature that is deprecated.  The deprecation is mentioned in two places, so this PR links to both of them, and adds the corresponding missing link to the list item in [basic.def]

### DIFF
--- a/source/future.tex
+++ b/source/future.tex
@@ -140,7 +140,8 @@ void park(linhenykus alvarezsauroid) {
 
 \pnum
 For compatibility with prior revisions of \Cpp{}, a \keyword{constexpr}
-static data member may be redundantly redeclared outside the class with no initializer.
+static data member may be redundantly redeclared outside the class with no
+initializer\iref{basic.def}\iref{class.static.data}.
 This usage is deprecated.
 \begin{example}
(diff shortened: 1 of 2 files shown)


source/future.tex
```diff
@@ -140,7 +140,8 @@
 
 \pnum
 For compatibility with prior revisions of \Cpp{}, a \keyword{constexpr}
-static data member may be redundantly redeclared outside the class with no initializer.
+static data member may be redundantly redeclared outside the class with no
+initializer\iref{basic.def}\iref{class.static.data}.
```
JohelEGP, line 144:
```suggestion
initializer\iref{basic.def,class.static.data}.
```

JohelEGP (changes requested):

jensmaurer (approved):

JohelEGP:
I'll pass on reviewing, since I don't know about the correctness of the PR.
My previous review was concerned about editorial changes.
That's why I sometimes don't re-review to approve.
Except for most Motions, where the Paper is the One Source of Truth.

PR #6450, opened 2023-08-09
[dlc.init.general] Fix wording of direct-initialization

eisenwave:
Firstly, this paragraph should be combining its items with `or`, not `and`. Initialization is not called *direct-initialization* if it's taking place both in a `static_cast` expression *and* in a functional notation type conversion at the same time, but rather if one *or* the other case.

Secondly, it's a little unusual how short the part after the `\itemize` section is. I think readability if the paragraph started with the defined term instead.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -4581,17 +4581,15 @@ Copy-initialization can invoke a move\iref{class.copy.ctor}.
 \end{note}
 
 \pnum
-The initialization that occurs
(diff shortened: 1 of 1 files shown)


AlisdairM:
The grammar with `and` is correct for the current definition, where we collectively mean all of the list.  Your reformulation turns this into a conditional with an `if`, effectively making the list into "one of these places" --- and that is when the `or` wording would be preferred.

eisenwave:
Low value.

PR #6453, opened 2023-08-13
[over.ics.list] Fix missing std:: in example 5

Eisenwave:


JohelEGP (approved):

source/overloading.tex
```diff
@@ -2424,9 +2424,9 @@
 f( {1.0} );                     // error: narrowing
 
 struct A {
-  A(std::initializer_list<double>);             // \#1
-  A(std::initializer_list<complex<double>>);    // \#2
-  A(std::initializer_list<std::string>);        // \#3
+  A(std::initializer_list<double>);                 // \#1
```
jensmaurer, line 2427: We want comments to be aligned to multiples of 4 (preferably, 8). This doesn't keep the alignment, I think.

jensmaurer (changes requested):

source/overloading.tex
Eisenwave: > We want comments to be aligned to multiples of 4 (preferably, 8). This doesn't keep the alignment, I think.

Yes, this is 52 columns, so it's only aligned to 4, not 8. I thought that it's enough for `check_source` to pass. Anyhow, will fix.

Eisenwave: Alright, I've added four more spaces, so it's aligned to 8, not just 4 now.

jensmaurer (approved):

jensmaurer (approved):

PR #6456, opened 2023-08-14
[lib] Avoid redundant \tcode{\exposid{...}} and add a check

jensmaurer:
Fixes #6406

### DIFF
--- a/tools/check-source.sh
+++ b/tools/check-source.sh
@@ -83,6 +83,10 @@ fgrep -n '\^' $texfiles |
 grep -n 'U+' $texfiles |
     fail 'use \\unicode or \\ucode or \\uname instead' || failed=1
 
+# Discourage double-wrapping \tcode{\exposid{data_}}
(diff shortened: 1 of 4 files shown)


tools/check-source.sh
```diff
@@ -83,6 +83,10 @@ fgrep -n '\^' $texfiles |
 grep -n 'U+' $texfiles |
     fail 'use \\unicode or \\ucode or \\uname instead' || failed=1
 
+# Discourage double-wrapping \tcode{\exposid{data_}}
+grep -n '\\tcode{\\exposid{[a-zA-Z0-9_]*}}' $texfiles |
+    fail 'double-wrapped \\exposid in \\tcode' || failed=1
```
JohelEGP, line 88: #6406 also has a commit for `\exposconcept` in `\tcode`.
There might be yet other macros similarly affected.

jensmaurer: Ok. Let's take this first and then add more.

tkoeppe:
Nice, thanks!

PR #6461, opened 2023-08-18
[mdspan.extents.cons] Fix typo (`dynamic_rank` => `rank_dynamic`)

cpplearner:
There's no definition of `dynamic_rank`. Based on the context, it seems that `rank_dynamic` is meant.

cc @crtrott


cpplearner:
ping

jensmaurer:
@jwakely , this looks reasonable to me. What do you think?

jensmaurer (approved):

PR #6464, opened 2023-08-19
[basic.fundamental] Itemize p13

Eisenwave:
image

This is a very slight readability improvement which highlights the already existing symmetry in the paragraph between what is said for the two infinities.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -5137,10 +5137,13 @@ The minimum range of representable values for a floating-point type is
 the most negative finite floating-point number representable
 in that type through
 the most positive finite floating-point number representable in that type.
-In addition, if negative infinity is representable in a type,
-the range of that type is extended to all negative real numbers;
-likewise, if positive infinity is representable in a type,
+In addition:
+\begin{itemize}
+\item If negative infinity is representable in a type,
+the range of that type is extended to all negative real numbers.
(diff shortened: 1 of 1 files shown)


jensmaurer (commented):
Hm... I'm not convinced this is a net improvement. I can certainly live with the result.
Bulleted lists in the standard are usually used for conjunctions or disjunctions or  for lists of conditions/rules (expr.const or basic.def.odr). This case doesn't feel like any of those.

@tkoeppe, what do you think?

tkoeppe:
I'm inclined to agree that this isn't really improving anything. The status quo seems entirely adequate. One could easily imagine an opposite desire to "declutter the presentation by turning trivial lists into fluent sentences", say, and we could endlessly churn on something like that. So I'd say we should only consider such rewrites if there is a very clear and unambiguous improvement. Having an underlying principle is great, too (such as "conjunctions and disjunctions can benefit from list style"), but I'd not want to suggest some absolute rule here. (The best way to present something clearly often depends on a lot of factors, and there should be room for discretion.)

JohelEGP:
With regards to clear presentation, I'm glad lists are more welcome now.
They're a powerful tool for the job.

jensmaurer:
Let's bury this, then.

PR #6465, opened 2023-08-19
[basic.fundamental] Itemize uses of `void` expressions

Eisenwave:
image

This edit improves readability of the paragraph by neatly splitting the list of uses into bullets. Furthermore, some (implicit) bullets are reordered, so that statements and expressions are grouped together.

Also, the enumeration "`typeid`, `noexcept`, or `decltype`" is split up. This helps clarify whether a `noexcept` specification, or a `noexcept` operator is referred to. Additional forward references to the relevant sections are added.


jensmaurer (approved):
Looks like a nice improvement in clarity to me.

PR #6469, opened 2023-08-19
[stmt.expr] use grammarterm for expression

Eisenwave:
In this paragraph, *expression* literally refers to when the *expresssion* grammar term is missing in the syntax (included in the same paragraph):
> *expression*<sub>opt</sub>`;`

It feels more appropriate to style it as a `grammarterm` due to this strong connection, instead of simple prose.

### DIFF
--- a/source/statements.tex
+++ b/source/statements.tex
@@ -188,7 +188,7 @@ All
 side effects from an expression statement
 are completed before the next statement is executed.
 \indextext{statement!empty}%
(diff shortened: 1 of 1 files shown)


jensmaurer (approved):

tkoeppe:
@Eisenwave Could you please rebase?

jensmaurer (approved):

PR #6470, opened 2023-08-19
[lex.icon] Itemize extended integer choice

Eisenwave:
image

This edit slightly improves readability of the three possible choices for the signedness of the extended integer type.

Furthermore, it conceptually separates the last `If ...` sentence from the previous ones, which is good, because otherwise it reads like just another bullet in a list.

Furthermore, the last bullet in the list is abbreviated to an `Otherwise,` bullet, since the list is exhaustive anyway.

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -1325,16 +1325,17 @@ Except for \grammarterm{integer-literal}{s} containing
 a \grammarterm{size-suffix},
 if the value of an \grammarterm{integer-literal}
 cannot be represented by any type in its list and
-an extended integer type\iref{basic.fundamental} can represent its value,
-it may have that extended integer type.
-If all of the types in the list for the \grammarterm{integer-literal}
-are signed,
-the extended integer type is signed.
-If all of the types in the list for the \grammarterm{integer-literal}
-are unsigned,
-the extended integer type is unsigned.
-If the list contains both signed and unsigned types,
-the extended integer type may be signed or unsigned.
+an extended integer type\iref{basic.fundamental} \tcode{T} can represent its value,
+it may have type \tcode{T}.
+\tcode{T} has the following signedness:
+\begin{itemize}
+\item If all types in the list for the \grammarterm{integer-literal}
+are signed, \tcode{T} is signed.
+\item If all types in the list for the \grammarterm{integer-literal}
+are unsigned, \tcode{T} is unsigned.
+\item If the list contains both signed and unsigned types,
+\tcode{T} may be signed or unsigned.
+\end{itemize}
 If an \grammarterm{integer-literal}
 cannot be represented by any of the allowed types,
 the program is ill-formed.


source/lex.tex
```diff
@@ -1327,14 +1327,17 @@
...
+\begin{itemize}
+\item If all types in the list for the \grammarterm{integer-literal}
 are signed,
 the extended integer type is signed.
-If all of the types in the list for the \grammarterm{integer-literal}
+\item If all types in the list for the \grammarterm{integer-literal}
 are unsigned,
 the extended integer type is unsigned.
-If the list contains both signed and unsigned types,
+\item Otherwise,
```
jensmaurer, line 1338: This "otherwise" feels like it might refer to the immediately preceding "if" only.
When we do if/otherwise ladders, we put "otherwise" on each subsequent "if".

In this case, I'd just leave the text of the third option as-is.

jensmaurer (changes requested):

source/lex.tex
Eisenwave: Done. I've somewhat simplified the bullets though, since repeating the term "extended integer type" in each is needlessly verbose.

Eisenwave:
Related follow-up CWG issue: https://github.com/cplusplus/CWG/issues/404

source/lex.tex
```diff
@@ -1327,14 +1327,15 @@
...
-If the list contains both signed and unsigned types,
-the extended integer type may be signed or unsigned.
+This extended integer type shall have the following signedness:
+\begin{itemize}
+\item If all types in the list for the \grammarterm{integer-literal}
+are signed, it is signed.
+\item If all types in the list for the \grammarterm{integer-literal}
+are unsigned, it is unsigned.
+\item If the list contains both signed and unsigned types,
+it may be signed or unsigned.
```
jensmaurer, line 1337: In order to make the standard as readable as possible, I feel that the antecedent for "it" should be the immediately preceding noun that matches grammar-wise. In this last bullet, it seems that "it" refers to the list. (Those in the know understand that a list is not supposed to be signed or unsigned, but the point of the standard is to convey knowledge, not to assume the reader has knowledge.)

(This is obviously violated in the status quo quite a bit, so this is more an aspirational comment for new edits.)

If you want to avoid the repetition, introduce "T" for the type we're talking about.

Given the editorial overlap with CWG/issues#404, I'm wondering whether the bulletization (and possibly slightly larger massagings) should be left to the CWG issue.

source/lex.tex
```diff
@@ -1327,14 +1327,15 @@
...
 it may have that extended integer type.
-If all of the types in the list for the \grammarterm{integer-literal}
-are signed,
-the extended integer type is signed.
-If all of the types in the list for the \grammarterm{integer-literal}
-are unsigned,
-the extended integer type is unsigned.
-If the list contains both signed and unsigned types,
-the extended integer type may be signed or unsigned.
+This extended integer type shall have the following signedness:
```
jensmaurer, line 1330: This uses "shall" for a requirement on the implementation. We usually avoid that (except in phrases like "the implementation shall") and reserve "shall" for diagnosable requirements on a (user-written) program.

jensmaurer (changes requested):

Eisenwave:
image

New version with the suggested changes.

source/lex.tex
Eisenwave: > If you want to avoid the repetition, introduce "T" for the type we're talking about.

Sounds like a good plan, I've done that.

> Given the editorial overlap with CWG/issues#404, I'm wondering whether the bulletization (and possibly slightly larger massagings) should be left to the CWG issue.

I've thought about it too, and don't think it's strictly necessary. The CWG issue is just about putting *implementation-defined* somewhere in the wording. The diff suggested in the core issue makes sense before and after the editorial changes here, so the two are somewhat independent.

tkoeppe:
OK, leaving this to the resolution of the CWG issue, but I've made a note there to incorporate this editorial change.

PR #6471, opened 2023-08-19
[dcl.attr.noreturn] Remove redundant note

Eisenwave:
This note is completely unnecessary, because just a few lines down, example 1 demonstrates:
```cpp
[[ noreturn ]] void f() {
  throw "error";                // OK
}
```
There is no good motivation for the reader to be informed in prose that throwing in a `[[noreturn]]` function is okay, if the example demonstrates that.


jensmaurer:
I disagree.

Sometimes, both a note and an example are useful to drive a point home.

In any case, both notes and examples are non-normative, and we sprinkle notes around if we feel they're helpful (as long as they're not wrong).

I'm not seeing any argument that the note is outright wrong, just that it is redundant, but the latter is exactly the purpose of notes.

PR #6472, opened 2023-08-19
[dcl.attr.nodiscard] Add example of nodiscard with message

eisenwave:
The current example does not, but could include a use of `[[nodiscard("reason")]]`. This edit fixes this wasted potential, and adds such a case to the existing example.


eisenwave:
Using messages is nice, but not really needed for clarifying behavior here.

PR #6475, opened 2023-08-19
[intro.memory] Remove stray bit definitions

Eisenwave:
This old paragraph defines two terms which aren't used anywhere else in the document. Maybe they have been used historically and someone forgot to remove them, or maybe they haven't ever been used.

On a personal note, I also have to say that I have never heard someone say *high-order bit* or *low-order bit*. The terminology *least significant* and *most significant* is much more popular in my experience, self-explanatory, and used in other places in the C++ standard.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3066,11 +3066,10 @@ bits,
 The number of bits in a byte is reported by the macro
 \tcode{CHAR_BIT} in the header \libheaderref{climits}.
 \end{footnote}
-the number of which is \impldef{bits in a byte}. The least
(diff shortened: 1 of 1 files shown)


AlisdairM:
I have gripped all the .tex files to confirm that these definitions are the only occurrences of "high-order" and "low order" in the standard.  To double-check, I verified that are no occurrences of "high order" nor "low order" either.

jensmaurer (approved):

jensmaurer:
I've confirmed the definitions are unused by searching through the PDF. This is ancient goo that should go.

PR #6477, opened 2023-08-20
[dcl.stc] rearrange wording, turn typedef restriction into note

### LINKED ISSUE #6081: [dcl.stc]/1 and [dcl.typedef]/1 each forbid "static typedef"
The former states a subset of the restrictions of the latter (which also forbids things like `constexpr typedef …`), so it would be sufficient to drop "there can be no `typedef` specifier in the same *decl-specifier-seq* and".

eisenwave:
image

This edit changes a few things:
- It adds a missing comma after the conditional clause "if `thread_local` appears ..."
- Fixes https://github.com/cplusplus/draft/issues/6081. This is done by turning the normative wording into a note at the bottom. The restriction on `typedef` is already defined in [dcl.typedef], and does not need to be stated here.
- It turns the last sentence into the first sentence, which seems more logical. What a *storage-class-specifier* applies to is more important than the list of restrictions.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -446,20 +446,23 @@ The storage class specifiers are
     \keyword{mutable}
 \end{bnf}
 
-At most one \grammarterm{storage-class-specifier} shall appear in a given
-\grammarterm{decl-specifier-seq}, except that \keyword{thread_local} may appear with \keyword{static} or
-\keyword{extern}. If \keyword{thread_local} appears in any declaration of
-a variable it shall be present in all declarations of that entity. If a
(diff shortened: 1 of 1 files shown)


AlisdairM:
Looks like a nice clean-up to me.  Thoughts @jensmaurer?

jensmaurer:
I like the added comma and I like the note for "static typedef".

Moving the last sentence to the front doesn't work:  It talks about "the" storage-class-specifier and about "in the list". Neither of which has been introduced with "a" yet. For the second part, "the list" means init-declarator-list or member-declarator-list, which is non-obvious.

PR #6483, opened 2023-08-20
[basic.def.odr] Greatly expand example 6

eisenwave:
image

The ODR is fairly complex, and deserving of examples which elaborate on intent and behavior in more detail. Firstly, the new style interleaves explanations and code blocks, which is necessary due to the scale of the example.

## First Addition
```cpp
static int local = 0; // OK
inline int a(int = local) { // OK
  return 0;
}
inline int b = a(0); // OK
inline int c = a(); // IFNDR
```
This example demonstrates three important mechanisms:
- semantic differences can violate the ODR, even if the syntax is the same in every TU
- using entities with internal linkage is dangerous
- default arguments in the parameter list are not subject to the requirements of the ODR
- however, uses of such default arguments are subject to the ODR

This part is much simpler than the subsequent part involving lambda-expressions, and establishes the knowledge necessary to understand the second part.

## Second Addition
The new (third) part of the example is as follows:
```cpp
inline decltype([]{}) u = {}; // IFNDR
inline auto v = []{}; // OK
inline auto w = decltype([]{}){}; // OK
```
This demonstrates that:
- using lambda expressions in declarations is a major footgun, and can violate the ODR in a more direct way
- lambda-expressions can appear in a definition, even without being contained within a block scope, and this doesn't violate the ODR

This part of the example is actually related to implementation divergence: https://godbolt.org/z/749Tf53hG / https://github.com/llvm/llvm-project/issues/64811.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -775,32 +775,72 @@ might still denote different types in different translation units.
 \pnum
 \begin{example}
 \begin{codeblock}
+static int local = 0;
+inline int a(int = local) {
+  return 0;
+}
+inline int b = a(0);
+\end{codeblock}
+If the definition of \tcode{a} or \tcode{b} appears in multiple translation units,
+the behavior of the program is as if
+there is only one definition of \tcode{a} or \tcode{b}, respectively.
+
+\begin{codeblock}
+inline int c = a();
+\end{codeblock}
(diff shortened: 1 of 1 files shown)


jensmaurer:
This is mostly CWG2782. I want that to land before applying additional editorial work, to avoid editing conflicts.

PR #6484, opened 2023-08-20
[dcl.constexpr] Modernize example of constexpr-usable functions

### LINKED ISSUE #5718: Example "uninit" in [dcl.constexpr] no longer valid
We discovered during the application of https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2448r2.html (which removes one incorrect example) that there is another incorrect example in dcl.constexpr, Example 2 in (3.4):

```
constexpr int uninit() {
  struct { int a; } s;
  return s.a;                   // error: uninitialized read of s.a
}
```

@zygoloid says, "Yeah, it's an uninitialized read, but prior to [p2448r2](http://wg21.link/p2448r2) that was IFNDR and now it's simply valid."

eisenwave:
Fixes https://github.com/cplusplus/draft/issues/5718.

image

As explained in the issue, the example has become largely noise because of how many restrictions have been lifted in `constexpr` functions. The intuition of most users is probably that `abs`, `long_max`, `prev` and others *should* be `constexpr`, unless proven otherwise. This shift in mentality calls for examples which demonstrate what *isn't* constexpr-usable, rather than what is.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -879,33 +879,26 @@ non-templated constexpr functions shall be constexpr-suitable.
 
 \begin{example}
 \begin{codeblock}
-constexpr int square(int x)
-  { return x * x; }             // OK
-constexpr long long_max()
-  { return 2147483647; }        // OK
-constexpr int abs(int x) {
-  if (x < 0)
-    x = -x;
-  return x;                     // OK
+constexpr int square(int x) {           // OK
+  return x * x;
 }
-constexpr int constant_non_42(int n) {  // OK
-  if (n == 42) {
-    static int value = n;
-    return value;
-  }
-  return n;
+constexpr std::generator<int> iota(int n = 0) {
+    while (true) co_yield n++;          // error: \tcode{co_yield} cannot be used in a constexpr function
 }
-constexpr int uninit() {
-  struct { int a; } s;
-  return s.a;                   // error: uninitialized read of \tcode{s.a}
+constexpr int stinit(int n) {           // \tcode{stinit} is constexpr-usable, but \tcode{stinit}
+  static int value = n;                 // may not be called in a constant expression
+  return value;
 }
-constexpr int prev(int x)
-  { return --x; }               // OK
-constexpr int g(int x, int n) { // OK
-  int r = 1;
-  while (--n > 0) r *= x;
-  return r;
+constexpr int uninit() {                // \tcode{uninit} is constexpr-usable, but a program which calls
+  struct { int a; } s;                  // \tcode{uninit} in a constant expression is ill-formed
+  return s.a;
 }
+struct Base {
+  constexpr Base() = default;           // OK
+};
+struct Derived : virtual Base {
+  constexpr Derived() = default;        // error: constructor cannot be constexpr in a
+};                                      // class or struct with virtual base classes
 \end{codeblock}
 \end{example}
 


source/declarations.tex
```diff
@@ -879,33 +879,26 @@
...
-    return value;
-  }
-  return n;
+constexpr std::generator<int> iota(int n = 0) {
+    while (true) co_yield n++;          // error: \tcode{co_yield} cannot be used in a constexpr function
 }
-constexpr int uninit() {
-  struct { int a; } s;
-  return s.a;                   // error: uninitialized read of \tcode{s.a}
+constexpr int stinit(int n) {           // \tcode{stinit} is constexpr-usable, but \tcode{stinit}
```
JohelEGP, line 888: The paragraph the example is attached to defines _constexpr-suitable_, so I think that's what you meant.
```suggestion
constexpr int stinit(int n) {           // \tcode{stinit} is constexpr-suitable, but \tcode{stinit}
```

source/declarations.tex
```diff
@@ -879,33 +879,26 @@
...
+  static int value = n;                 // may not be called in a constant expression
+  return value;
 }
-constexpr int prev(int x)
-  { return --x; }               // OK
-constexpr int g(int x, int n) { // OK
-  int r = 1;
-  while (--n > 0) r *= x;
-  return r;
+constexpr int uninit() {                // \tcode{uninit} is constexpr-usable, but a program which calls
```
JohelEGP, line 892:
```suggestion
constexpr int uninit() {                // \tcode{uninit} is constexpr-suitable, but a program which calls
```

source/declarations.tex
```diff
@@ -879,33 +879,26 @@
...
+  return x * x;
 }
-constexpr int constant_non_42(int n) {  // OK
-  if (n == 42) {
-    static int value = n;
-    return value;
-  }
-  return n;
+constexpr std::generator<int> iota(int n = 0) {
+    while (true) co_yield n++;          // error: \tcode{co_yield} cannot be used in a constexpr function
```
JohelEGP, line 886: I think this example can be improved.
[expr.const] bans `co_await` and `co_yield`,
but [dcl.constexpr]p3 (the paragraph the example is attached to)
bans a constexpr function that is a coroutine.
So I think it might better to use `co_return` instead,
which isn't banned by [expr.const].

JohelEGP (changes requested):

jensmaurer (changes requested):
This is about `constexpr`, not about [expr.const].
The focus should be on stuff that is ok/ill-formed because of [dcl.constexpr].
Having one (punchy) example of something that is ok by [dcl.constexpr] but ill-formed if invoked in a constant-expression context in all cases is fine. An unconditional throw-expression is good enough for that.

source/declarations.tex
```diff
@@ -879,33 +879,26 @@
...
 }
-constexpr int prev(int x)
-  { return --x; }               // OK
-constexpr int g(int x, int n) { // OK
-  int r = 1;
-  while (--n > 0) r *= x;
-  return r;
+constexpr int uninit() {                // \tcode{uninit} is constexpr-usable, but a program which calls
+  struct { int a; } s;                  // \tcode{uninit} in a constant expression is ill-formed
+  return s.a;
```
frederick-vs-ja, line 894:
```suggestion
constexpr int uninit() {                // OK, but never produces a core constant expression
  struct { int a; } s;
  return s.a;                           // undefined behavior: uninitialized read of \tcode{s.a}
```

frederick-vs-ja (commented):
I think "OK" in this example means constexpr-usable/constexpr-suitable. So perhaps we should only use the simple "OK".

source/declarations.tex
```diff
@@ -879,33 +879,26 @@
...
 }
+struct Base {
+  constexpr Base() = default;           // OK
+};
+struct Derived : virtual Base {
+  constexpr Derived() = default;        // error: constructor cannot be constexpr in a
+};                                      // class or struct with virtual base classes
```
jensmaurer, line 901: "class or struct" -> "class"

("class" covers all of union, class, struct)

jensmaurer (changes requested):

source/declarations.tex
frederick-vs-ja: This part is no longer erroneous since C++26 due to [P3533R2](https://wg21.link/p3533r2) and thus should be dropped from this PR.

PR #6487, opened 2023-08-20
[lex.phases],[forward.list.ops],[list.ops],[alg.unique] Hyphenate non-empty

Eisenwave:


jensmaurer (commented):
Hm... The wiki says we want to use "nonempty" for sets that are not empty:

https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#hyphens-nonx-vs-non-x

(We don't have empty classes anymore, I think; the term is gone.)

It feels to me that a range that has no elements would be an "empty set", and thus "nonempty range" and "nonempty list" would be correct.

@tkoeppe , @zygoloid (as the author of the hyphenation rule): What do you think?

JohelEGP:
#440 was reopened for this.

lorihughes:
I was asked to chime in on this thread. _Chicago Manual of Style_ omits the hyphen for prefixes most of the time. (We still include the hyphen when the characters are repeated, e.g., non-negative, or when necessary for clarity.) I double-checked several other popular style guides for their take: _Butcher's Copy-Editing_, _New Oxford Style Manual_, _AP Stylebook_, _NYT Manual of Style and Usage_, _Dreyer's English_. All agree that the hyphen should be omitted for nonempty.

jensmaurer:
@lorihughes , thanks for your input here!

Side note: Hyphenation seems to be one of areas where US and British English sometimes differ, and Germans usually get exposed to British English at school.  Also, technical texts possibly prefer hyphenation in some cases for visual clarity.

For me, it's "nonempty", and we should fix the offenders in the wording.

lorihughes:
> @lorihughes , thanks for your input here!
> 
> Side note: Hyphenation seems to be one of areas where US and British English sometimes differ, and Germans usually get exposed to British English at school. Also, technical texts possibly prefer hyphenation in some cases for visual clarity.
> 
> For me, it's "nonempty", and we should fix the offenders in the wording.

You're welcome!
Re the US/UK difference, that's why I consulted Oxford and Butcher's; those are both UK sources.

tkoeppe:
Editorial meeting February 2025: Jens's representation in https://github.com/cplusplus/draft/pull/6487#pullrequestreview-1586295636 of our wiki rule is on point. There are different contexts, and all the proposed changes concern the natural language meaning of "nonempty", which both our wiki rule and the style guide that @lorihughes cited say should use the unhyphenated form. The status quo (and our rule) seem to be good as is.

PR #6495, opened 2023-08-21
[stmt.do] Add grammar equivalence and reorder paragraphs

### LINKED ISSUE #6492: [stmt.do] Needs some work
In [[stmt.do]](https://eel.is/c++draft/stmt.do), it is stated that "the expression is contextually converted ...".

No grammar rule is provided in this paragraph, so it is a complete mystery to the reader what expression is being referred to. The expression is also called "condition" in [stmt.while], but this is not done in [stmt.do].

Also, unlike [stmt.while], the construction of the semantics does not happen in terms of `goto`, but is done in prose. This is another ugly asymmetry.

Last but no least, a note could be used to clarify the difference to `while`, or an example could show that a `do ... while (false)` statement is executed exactly once.

eisenwave:
Fixes #6492 for the most part.

image

This edit improves a few things:
1. The two paragraphs are swapped, which creates symmetry with how they are ordered in [stmt.while]
2. Similar to `while` and `for`, the equivalence to a `goto` construct is also provided.
  Note that the trailing semicolon issue is swept under the rug by defining the equivalence for `while (expression);` with a semicolon present
3. `\grammarterm` is used in a few more places.

### DIFF
--- a/source/statements.tex
+++ b/source/statements.tex
@@ -595,13 +595,25 @@ fails.
 \indextext{statement!\idxcode{do}}
 
 \pnum
-The expression is contextually converted to \tcode{bool}\iref{conv};
-if that conversion is ill-formed, the program is ill-formed.
+In the \keyword{do} statement the substatement is executed repeatedly
+until the value of the \grammarterm{expression} becomes \tcode{false}.
+The test takes place after each execution of the statement.
 
 \pnum
-In the \keyword{do} statement the substatement is executed repeatedly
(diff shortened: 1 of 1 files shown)


jensmaurer:
This shouldn't go in without CWG seeing it.

eisenwave:
Too much stuff happening in this area with `do` expressions, expansion statements, `break label`; etc. Maybe could be revisited in the future.

PR #6497, opened 2023-08-21
[class.conv.ctor] Fix missing space

Eisenwave:
https://github.com/cplusplus/draft/blob/1c22d62180901069128b21daa2773d40566bd983/source/classes.tex#L2357-L2369

The author follows the ` = ` style in the rest of the example, so it is obvious that the missing space here is not intentional, but a typo.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -2356,7 +2356,7 @@ Such a constructor is called a
 \begin{codeblock}
 struct X {
     X(int);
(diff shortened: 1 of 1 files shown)


jensmaurer (approved):

tkoeppe:
The core language deliberatey does not use any one fixed style, and the document has no specific "author". We can make the change, but I wouldn't call it a "fix".

PR #6502, opened 2023-08-22
[atomics] Make notify_one nondeterministic

languagelawyer:
See #6501 for explanation

Affects:
[atomics.ref.ops]
[atomics.types.operations]
[util.smartptr.atomic.shared]
[util.smartptr.atomic.weak]
[atomics.flag]

### DIFF
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -3177,7 +3177,7 @@ void notify_one() const noexcept;
 \begin{itemdescr}
 \pnum
 \effects
-Unblocks the execution of at least one atomic waiting operation on \tcode{*ptr}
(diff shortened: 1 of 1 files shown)


jensmaurer (changes requested):
This particular choice of words seems to be ambiguous. This sounds as if there is something unspecified about an "atomic waiting operation", but that's not the point here.

It's the choice among the eligible ones that is unspecified.

Something like "of an unspecified one among the atomic waiting opeations that are ..."
but that loses the "at least one" property.

Maybe having a generic sentence in [atomics.wait] about how the choice during unblocking is unspecified (both the number of unblocked operations and the choice which ones is unspecified).

PR #6503, opened 2023-08-22
[intro.multithread.general] Do not access functions

languagelawyer:


source/basic.tex
```diff
@@ -5990,7 +5990,7 @@
 new thread, not by the creating thread.
 \end{note}
 Every thread in a program can
-potentially access every object and function in a program.
+potentially access every object and invoke any function in a program.
```
jensmaurer, line 5993: What about taking the address of a function? The new wording seems to exclude that more clearly.

languagelawyer: Kk, since this is an introductory paragraph which is, AFAIK, is given some slack in its wording usage (I mean, it can use words not defined as terms, but still should not use terms, like "access" here), maybe
```suggestion
potentially reach every object and function in a program.
```
(Hm. We have http://eel.is/c++draft/module.reach#def:reachable,declaration)

jensmaurer: I think I'd be ok with "use" (second suggestion). Please update the pull request.

tkoeppe:
@jensmaurer: PTAL?

jensmaurer:
"access" is a defined term that doesn't apply to functions, so this is an improvement.

PR #6505, opened 2023-08-23
[class.conv.ctor] Turn last paragraph into a note

Eisenwave:
Paragraph 1 already states:
https://github.com/cplusplus/draft/blob/1c22d62180901069128b21daa2773d40566bd983/source/classes.tex#L2347-L2353

The reader has no reason to believe that a copy/move constructor is excluded from the rule, and it certainly doesn't need to be stated in normative wording. The whole paragraph can and should be a note.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -2403,10 +2403,9 @@ Z a6 = { 3, 4 };                // error: no implicit conversion
 
 \pnum
 \begin{note}
-A non-explicit copy/move constructor\iref{class.copy.ctor} is
-a converting constructor.
-An implicitly-declared copy/move constructor is not an explicit constructor;
-it can be called for implicit type conversions.
+A non-explicit copy/move constructor\iref{class.copy.ctor},
+including one implicitly declared,
+is a converting constructor.
 \end{note}
 
 \rSec3[class.conv.fct]{Conversion functions}%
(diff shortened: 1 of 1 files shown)


source/classes.tex
```diff
@@ -2402,9 +2402,9 @@
 \end{note}
 
 \pnum
+\begin{note}
 A non-explicit copy/move constructor\iref{class.copy.ctor} is
 a converting constructor.
-\begin{note}
 An implicitly-declared copy/move constructor is not an explicit constructor;
 it can be called for implicit type conversions.
```
opensdh, line 2409: This is now pretty heavily redundant.  How about

> A non-explicit copy/move constructor&nbsp;([class.copy.ctor]), including one implicitly declared, is a converting constructor.

tkoeppe: That's nice, thanks!

tkoeppe: @Eisenwave ^^^ could you please have a look?

Eisenwave: Yeah I'll go through all the things I have to do in this repo tomorrow.

jensmaurer:
CWG 2023-11-11: Approved.

tkoeppe:
@jensmaurer Would you prefer to take it as is, or wait for the improvements Davis has suggested?

jensmaurer:
We can go forward as-is, and address Davis' concerns at a later stage. The important part is that CWG approved changing normative text to a note; fixing the note is then purely editorial.

source/classes.tex
tkoeppe: @Eisenwave Can you please send that as a separate commit?

Issue #6498, opened 2023-08-21
[class.copy.assign] Definition and hyphenation issues


Eisenwave:
https://github.com/cplusplus/draft/blob/1c22d62180901069128b21daa2773d40566bd983/source/classes.tex#L1663-L1663

Why is this a seemingly defining *copy* as a term? Shouldn't it be pure prose, or defining *copy assignment operator* as a term?

Also, there are hyphenation issues throughout the document. Sometimes (though this is the vast minority), the hyphenated terms *copy-assignment* and *move-assignment* are used. This should be made uniform.

Eisenwave:
I see, so in https://github.com/cplusplus/draft/issues/556#issuecomment-283569920, it was voted that the whole term should be defined, not just `\term{copy}`.

If the whole term was defined, then the definition would also clarify that the non-hyphenated version should be used elsewhere to match the definition.

PR #6506, opened 2023-08-23
[basic.def.odr],[res.on.arguments],[futures] Fix hyphenation of "assignment operator"

Eisenwave:
Fixes https://github.com/cplusplus/draft/issues/6498 (in part).


tkoeppe:
Thanks! I've reworded the commits and will push them directly.

PR #6507, opened 2023-08-23
[stmt.return] format operand as `\term`

Eisenwave:
`operand` is a definition in this context, but isn't formatted as such.

The use of `\term` over `\defn` is intentional, since the [Q&A](https://github.com/cplusplus/draft/wiki/Q-&-A) suggests to use `\term` for definitions that shouldn't be indexed, and I don't believe it's necessary to index *operand*

### DIFF
--- a/source/statements.tex
+++ b/source/statements.tex
@@ -851,7 +851,8 @@ A function returns to its caller by the \tcode{return} statement.
 
 \pnum
 The \grammarterm{expr-or-braced-init-list}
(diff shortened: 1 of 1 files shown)


tkoeppe:
Our guidance may not be very clear on this, but ISO generally does not want words to be italicised, and all proper "terms" should be in Clause 3; as a compromise, we try not to italicise locally used terms anymore.

PR #6508, opened 2023-08-23
[stmt.dcl] Add example of static/thread_local block variable initialization

Eisenwave:
image

This edit adds an example which clarifies at least two potential questions:
- If there are multiple static/thread-local variables and one of them throws during initialization, are all initializations re-attempted?
- Are surrounding thread-local block variables initialized when a new thread is created?

It also raises awareness for a particularly dangerous edge case, namely:
```cpp
thread_local int n = 0;
std::jthread([] {
    // n is not initialized in here!
});

### CODE REVIEW

File: source/statements.tex [hunk 1]
```diff
@@ -1038,11 +1038,28 @@
 the first time control passes through its declaration; such a variable is
 considered initialized upon the completion of its initialization. If the
 initialization exits by throwing an exception, the initialization is not
-complete, so it will be tried again the next time control enters the
+complete, so it will be tried again the next time control passes through the
 declaration.
-If control enters the declaration concurrently while the variable is
+If control passes through the declaration concurrently while the variable is
 being initialized, the concurrent execution shall wait for completion
```
frederick-vs-ja, line 1044:
```suggestion
being initialized, the concurrent execution waits for completion
```

source/statements.tex
```diff
@@ -1038,11 +1038,28 @@
...
 of the initialization.
+\begin{example}
+\begin{codeblock}
+void f() {
+  static int x = 0;                     // OK
+  static std::string s = "hello";       // OK; if an exception is thrown during the
+                                        // initialization of \tcode{s}, \tcode{s} remains uninitialized, and
+                                        // an attempt to initialize \tcode{s} (but not \tcode{x}) will be
+                                        // made the next time \tcode{f} is called
+  thread_local int n = 0;               // OK, initializes \tcode{n} to \tcode{0} on the main thread
```
frederick-vs-ja, line 1054: `n` (of both threads) is constant-initialized here. So the current example is seemly free from UB and may be misleading, since `n` is unrelated to dynamic initialization.

Eisenwave:
image

Updated version

source/statements.tex
```diff
@@ -1038,11 +1038,30 @@
...
 of the initialization.
+\begin{example}
+\begin{codeblock}
+void f() {
+  static int x = 2;                     // OK
+  static std::string s = "hello";       // OK; if an exception is thrown during the
+                                        // initialization of \tcode{s}, \tcode{s} remains uninitialized, and
+                                        // an attempt to initialize \tcode{s} (but not \tcode{x}) will be
+                                        // made the next time \tcode{f} is called.
+  thread_local const char* w = "world"; // OK, initializes \tcode{w} to \tcode{"world"} on the main thread.
```
frederick-vs-ja, line 1054: `w` is still constant-initialized. Can we show non-constant initialization here? (e.g. `thread_local const char* w = std::system_category().name();`).

source/statements.tex
```diff
@@ -1038,11 +1038,30 @@
...
+  std::jthread([] {
+    thread_local int k = 4;
+    std::cout << k + x;                 // OK, prints \tcode{"6"}.
+    std::cout << s << ' ';              // OK, prints \tcode{"hello "}.
+    std::cout << w;                     // Undefined behavior if \tcode{w} hasn't been constant initialized\iref{basic.start.static}.
```
frederick-vs-ja, line 1059: cplusplus/CWG#372 may be involved. But if we change `w` to have dynamic initialization, it seems safe to say this line raises UB unconditionally - a zero-initialized (null) `const char*` value is sufficient to raise UB.

source/statements.tex
Eisenwave: Its not required to, and yes.

frederick-vs-ja: > Its not required to

I don't see why it is not required to. [[expr.const] p2.2](https://eel.is/c++draft/expr.const#2.2) and [p3.3](https://eel.is/c++draft/expr.const#13.3) seemly require that `w` is constant-initialized when the initializer is `"world"`.

Eisenwave: I was thinking of the rules for static initialization, which only require that dynamic initialization may (but is not required to) be omitted when the initializer is a constant expression. I wasn't aware of [expr.const] p2.2.

Eisenwave: > I don't see why it is not required to.

@frederick-vs-ja actually wait, I am right. It is true that e.g. `int x = 0;` in block scope is constant-initialized, but this is meaningless for block variables. Something being *constant-initialized* doesn't imply that it's constant initialized (https://eel.is/c++draft/basic.start.static#2). (pay attention to the hyphenation, `constant-initialized` and `constant initialization` are two completely different things)

frederick-vs-ja: In this example we're initializing a variable with thread storage duration. So if it is constant-initialized, it has constant initialization.

eisenwave:
I feel like the wording is clear enough; this is more of a tutorial.

PR #6515, opened 2023-08-25
[meta.unary.op] Expand note for has_unique_object_representations

eisenwave:
It's a common question in the community for what types `std::has_unique_object_representations_v` yields `false`, and why that happens.

I think it's worth expanding this note so that it mentions two very common failure cases:
- IEC 559 floating-point numbers, which have multiple representations for zero and NaN
- `bool` on ARM, which allows for more representations of `true` than just `0x01`

### DIFF
--- a/source/meta.tex
+++ b/source/meta.tex
@@ -1354,8 +1354,14 @@ if they have the same active member and the corresponding members have the same 
 The set of scalar types for which this condition holds is
 \impldef{which scalar types have unique object representations}.
 \begin{note}
-If a type has padding bits, the condition does not hold;
-otherwise, the condition holds true for integral types.
+If a type \tcode{T} has padding bits, the condition does not hold for \tcode{T}.
+If \tcode{T} is an integral type other than \keyword{bool}, the condition
+holds for \tcode{T}.
+The condition might not hold for \keyword{bool} because
+\keyword{true} or \keyword{false} can have multiple object representations.
+If \tcode{std::numeric_limits<T>::is_iec559} is \keyword{true}, the
+condition does not hold for \tcode{T} because there are multiple object
+representations for some values, such as zero.
 \end{note}
 
 \rSec2[meta.unary.prop.query]{Type property queries}


source/meta.tex
```diff
@@ -1356,6 +1356,11 @@
 \begin{note}
 If a type has padding bits, the condition does not hold;
 otherwise, the condition holds true for integral types.
+If \tcode{std::numeric_limits<float>::is_iec559} is \keyword{true}, the
+condition does not hold because there are multiple object representations for
+some values, such as zero.
+The condition might not hold for \keyword{bool} because
+\keyword{true} or \keyword{false} can have multiple object representations.
```
CaseyCarter, line 1363: This seems to contradict the first sentence of the note, since `bool` is an integral type.

source/meta.tex
```diff
@@ -1356,6 +1356,11 @@
 \begin{note}
 If a type has padding bits, the condition does not hold;
 otherwise, the condition holds true for integral types.
+If \tcode{std::numeric_limits<float>::is_iec559} is \keyword{true}, the
+condition does not hold because there are multiple object representations for
+some values, such as zero.
```
CaseyCarter, line 1361: This suggests that `has_unique_object_representations_v<T>` is `false` for all `T` when `std::numeric_limits<float>::is_iec559` is `true`. The three sentences of this note all use the phrase "the condition" with slightly different meanings.

source/meta.tex
Eisenwave: Good find, this actually makes the note in the current draft incorrect. I've reworded it so that `bool` is not included.

source/meta.tex
Eisenwave: Good point. I've fixed this by consistently using `T` and "holds for `T`" in the updated wording.

source/meta.tex
```diff
@@ -1354,8 +1354,14 @@
...
-If a type has padding bits, the condition does not hold;
-otherwise, the condition holds true for integral types.
+If a type \tcode{T} has padding bits, the condition does not hold for \tcode{T}.
+If \tcode{T} is an integral type other than \keyword{bool}, the condition
+holds for \tcode{T}.
+The condition might not hold for \keyword{bool} because
+\keyword{true} or \keyword{false} can have multiple object representations.
+If \tcode{std::numeric_limits<T>::is_iec559} is \keyword{true}, the
+condition does not hold for \tcode{T} because there are multiple object
+representations for some values, such as zero.
```
jensmaurer, line 1364: Do you mean plus and minus zero? Those are distinguishable.

source/meta.tex
```diff
@@ -1354,8 +1354,14 @@
 The set of scalar types for which this condition holds is
 \impldef{which scalar types have unique object representations}.
 \begin{note}
-If a type has padding bits, the condition does not hold;
-otherwise, the condition holds true for integral types.
+If a type \tcode{T} has padding bits, the condition does not hold for \tcode{T}.
+If \tcode{T} is an integral type other than \keyword{bool}, the condition
```
jensmaurer, line 1358: This has lost the "otherwise": We can have integral types with padding bits, I think.

source/meta.tex
```diff
@@ -1354,8 +1354,14 @@
...
+holds for \tcode{T}.
+The condition might not hold for \keyword{bool} because
```
jensmaurer, line 1360: We can't have "might" in notes, says ISO. "can" is ok.

jensmaurer (changes requested):

source/meta.tex
Eisenwave: With C++20 requiring two's complement, maybe not ... unless unsigned integers are padded too, and I think they can be.

source/meta.tex
Eisenwave: I believe C++ considers them to be the same value based on equality comparison. They are the same value, and positive/negative zero are two value/object representations of it.

source/meta.tex
jensmaurer: Yes, we can have a 31-bit unsigned int on CHAR_BIT=8 machine.

source/meta.tex
jensmaurer: Well, 1/-0 yields a value different from 1/+0 in IEEE, so (regardless of equality comparison), the value +0 seems to be different from -0.

The various IEEE NaN encodings seem a better example.

frederick-vs-ja: I'm not sure whether IEEE 754 floating-point formats are OK. It seems that all patterns of value representations are distinguishable in `strong_order` ([[cmp.alg] p1.3](https://eel.is/c++draft/cmp.alg#1.3)) due to the totalOrder operation.

Eisenwave: @frederick-vs-ja non-canonical representations are not ordered, and I believe they aren't otherwise distinguishable in any way: 
image

(from the IEEE-754 standard, which presumably has the same `totalOrder`)

Eisenwave: @jensmaurer I'm starting to think that even though `+0` and `-0` compare equal and are otherwise indistinguishable with the builtin relational operators, they are still distinct values.

I'll change the wording to talk about NaN encodings instead.

frederick-vs-ja: Non-canonical representations are only present for decimal floating-point formats. I'm not sure whether decimal FP is suitable here now.

Eisenwave: @frederick-vs-ja then I've misunderstood what non-canonical means.

Anyhow, all major compilers yield `false` for `std::has_unique_object_representations_v<float>` on IEC 559 platforms, so consensus is that floating point numbers don't satisfy it.

The way that `totalOrder` orders different encodings of NaN could be seen as an ordering imposed on the object representation, without all those possible NaNs being distinct values.

This lack of clarity sounds like a core issue to me though, and maybe CWG should add normative wording that decides whether IEC 559 floats have unique object representations or not.

Issue #6510, opened 2023-08-24
Section headings should use \hypertarget


t3nsor:
It would be very useful if we could directly link to a particular section in a PDF version of the standard, e.g., `https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/n4958.pdf#dcl.init.general` should open the PDF at page 211.

This could be accomplished by inserting a `\hypertarget` command, perhaps in the definition of the `\Sec` macros, or in the definition of `\addxref`.

Eelis:
That would be awesome! (Even though it would put cxxdraft-htmlgen out of business... ;-))

tkoeppe:
Can you prototype this and check that this would actually work?

JohelEGP:
> (Even though it would put cxxdraft-htmlgen out of business... ;-))

Hardly.
This may be a big use case,
but there's so much more you can do with HTML
that a PDF can't AFAIK.

t3nsor:
I'll prototype it soon when I get the chance.

The main motivation for this is for the links to be stable. If I have a paper in the mailing that links to a particular section of the standard, it should be a link that will always be to the particular working draft that was the most recent one at the time the paper was written. But I imagine people will still use eel.is most of the time when such stability isn't required.

t3nsor:
Please see the above PR. I've built it locally and confirmed that it works. I haven't uploaded the PDF, as I wouldn't want to be accused of violating ISO policy somehow, but please see more details on the PR.

leni536:
FTR https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/n4958.pdf#subsection.9.4.1 works for me on the current PDF. You can get the link by right clicking the outline element in Firefox, but AFAIK the link should work in other browsers too.

But I do like the better fragment identifiers introduced here.

JohelEGP:
> but AFAIK the link should work in other browsers too.

I remember it being the other way around.

leni536:
> > but AFAIK the link should work in other browsers too.
> 
> I remember it being the other way around.

I know two kinds of fragment id links that get resolved in some browsers:

1. Fragment id that refers to an outline element by outline name:
  * https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/n4958.pdf#9.4.1%20General
  * This works in Chrome, doesn't work in Firefox
2. Fragment id that refers to a PDF "bookmark"
[...]

PR #6516, opened 2023-08-26
Add \hypertarget to enable url links into xrefs

t3nsor:
Fixes https://github.com/cplusplus/draft/issues/6510 .

### DIFF
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -64,6 +64,7 @@
 %% Cross-references.
 %%--------------------------------------------------
 \newcommand{\addxref}[1]{%
+ \hypertarget{#1}{}%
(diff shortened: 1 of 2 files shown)


t3nsor:
I've built the PDF locally and tested that the following kinds of URL hash fragments do the right thing:
* `#dcl.init.general`: goes to [dcl.init.general] in the main text
* `#defns.static.type`: goes to the definition of "static type"
* `#implimits`: goes to Annex B.
It was necessary to move the `\addxref` command to the beginning of the expansions that it's referenced in, to ensure that the `\hypertarget` targets the end of the previous line, not the end of the line that the heading is located on. So I've also checked that the glossary entries still go to the right place. It appears that they do.

jensmaurer (approved):

CaseyCarter (approved):

PR #6518, opened 2023-08-26
[atomics.ref.ops], [atomics.types.operations], [util.smartptr.atomic.…

Dani-Hub:
…shared], [util.smartptr.atomic.weak], [atomics.flag]: Reword Preconditions on memory_order values in a positive form

### DIFF
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -2968,10 +2968,8 @@ void store(T desired, memory_order order = memory_order::seq_cst) const noexcept
 \begin{itemdescr}
 \pnum
 \expects
-The \tcode{order} argument is neither
-\tcode{memory_order::consume},
-\tcode{memory_order::acquire}, nor
-\tcode{memory_order::acq_rel}.
+The \tcode{order} parameter is \tcode{memory_order::relaxed},
+\tcode{memory_order::release}, or \tcode{memory_order::seq_cst}.
 
 \pnum
 \effects
@@ -3009,8 +3007,9 @@ T load(memory_order order = memory_order::seq_cst) const noexcept;
 \begin{itemdescr}
 \pnum
 \expects
-The \tcode{order} argument is neither
-\tcode{memory_order::release} nor \tcode{memory_order::acq_rel}.
+The \tcode{order} parameter is \tcode{memory_order::relaxed},
+\tcode{memory_order::consume}, \tcode{memory_order::acquire},
+or \tcode{memory_order::seq_cst}.
 
 \pnum
 \effects
@@ -3082,8 +3081,9 @@ bool compare_exchange_strong(T& expected, T desired,
 \begin{itemdescr}
 \pnum
 \expects
-The \tcode{failure} argument is neither
-\tcode{memory_order::release} nor \tcode{memory_order::acq_rel}.
+The \tcode{failure} parameter is \tcode{memory_order::relaxed},
+\tcode{memory_order::consume}, \tcode{memory_order::acquire},
+or \tcode{memory_order::seq_cst}.
 
 \pnum
 \effects
@@ -3146,8 +3146,9 @@ void wait(T old, memory_order order = memory_order::seq_cst) const noexcept;
 \begin{itemdescr}
 \pnum
 \expects
-\tcode{order} is
-neither \tcode{memory_order::release} nor \tcode{memory_order::acq_rel}.
+The \tcode{order} parameter is \tcode{memory_order::relaxed},
+\tcode{memory_order::consume}, \tcode{memory_order::acquire},
+or \tcode{memory_order::seq_cst}.
 
 \pnum
 \effects
@@ -3814,8 +3815,8 @@ For the \tcode{volatile} overload of this function,
 
 \pnum
 \expects
-The \tcode{order} argument is neither \tcode{memory_order::consume},
-\tcode{memory_order::acquire}, nor \tcode{memory_order::acq_rel}.
+The \tcode{order} parameter is \tcode{memory_order::relaxed},
+\tcode{memory_order::release}, or \tcode{memory_order::seq_cst}.
 
 \pnum
 \effects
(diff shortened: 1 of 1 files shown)


Dani-Hub:
I need some advice how to fix the "overfull \hbox" errors. It seems to say that the resulting text would need too much space in horizontal direction. OK, let's at the first case "lines 3154--3160": How is it possible to edit the text to make it even smaller?

JohelEGP:
For reviewing, it's not helpful that the PDF isn't uploaded because the output check failed.

Dani-Hub:
> For reviewing, it's not helpful that the PDF isn't uploaded because the output check failed.

I suspect this is not something I can fix, right?

Regarding my text changes: I tried to follow the pre-existing line-break structure. Maybe I shouldn't and I should try to provide each paragraph as filled line with breaks at around 70 characters?

JohelEGP:
> I need some advice how to fix the "overfull \hbox" errors.

If you built it locally, you can look at which syllable a word overfills.
You can guess where the margin is by looking at the surrounding text.
Then you can just add a `\-` in front of the syllable closest to the margin that isn't part of the overfill.

Dani-Hub:
> If you built it locally, you can look at which syllable a word overfills. 

Hmmh, I'm on Windows and I'm not sure how to setup my infrastructure to create the pdf locally.

If there is no other way I need to experiment with my PULL request.

JohelEGP:
> I suspect this is not something I can fix, right?

I don't think you need to, as it is a separate problem.

> Regarding my text changes: I tried to follow the pre-existing line-break structure. Maybe I shouldn't and I should try to provide each paragraph as filled line with breaks at around 70 characters?

The source line breaks are independent of the rendered text.
<https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#latex-source-code-formatting> says
> Use semantic line breaks in LaTeX source code: http://sembr.org/ and http://rhodesmill.org/brandon/2012/one-sentence-per-line/

which I wish was more strictly applied.
Those sources explain the difference well.

JohelEGP:
To get the PDF uploaded,
you can edit <https://github.com/cplusplus/draft/blob/main/.github/workflows/check.yml>
and swap the last two steps ("check-output.sh" and "upload PDF") at the end of the file.

Dani-Hub:
> To get the PDF uploaded, you can edit https://github.com/cplusplus/draft/blob/main/.github/workflows/check.yml and swap the last two steps ("check-output.sh" and "upload PDF") at the end of the file.

But isn't that a setting that should be independently set from my PULL request?

source/threads.tex
```diff
@@ -3146,8 +3144,8 @@
 \begin{itemdescr}
 \pnum
 \expects
-\tcode{order} is
-neither \tcode{memory_order::release} nor \tcode{memory_order::acq_rel}.
+\tcode{order} is either \tcode{memory_order::relaxed}, \tcode{memory_order::consume},
```
jensmaurer, line 3147: Can we harmonize this to read "The order argument is ..." (see next change)

Or maybe the other way round.

jensmaurer:
> > To get the PDF uploaded, you can edit https://github.com/cplusplus/draft/blob/main/.github/workflows/check.yml and swap the last two steps ("check-output.sh" and "upload PDF") at the end of the file.
> 
> But isn't that a setting that should be independently set from my PULL request?

No. In general, I think we can expect someone editing the draft to be able to build locally before uploading. And we only want "good" artifacts published from our continuous integration pipeline.

source/threads.tex
Dani-Hub: I harmonized it now as of your first suggestion.

source/threads.tex
```diff
@@ -2968,10 +2968,8 @@
 \begin{itemdescr}
 \pnum
 \expects
-The \tcode{order} argument is neither
-\tcode{memory_order::consume},
-\tcode{memory_order::acquire}, nor
-\tcode{memory_order::acq_rel}.
+The \tcode{order} argument is either \tcode{memory_order::relaxed},
```
jensmaurer, line 2971: Strictly speaking, "order" is not an argument, but the name of a parameter.

source/threads.tex
```diff
@@ -4970,10 +4976,8 @@
 \begin{itemdescr}
 \pnum
 \expects
-\tcode{order} is neither
-\tcode{memory_order::consume},
-\tcode{memory_order::acquire}, nor
-\tcode{memory_order::acq_rel}.
+The \tcode{order} argument is either \tcode{memory_order::relaxed},
```
jensmaurer, line 4979: I believe "either" in English preferentially refers to a choice of two things.  We have three things here.
I'd suggest to simply drop "either"; I don't think it adds anything.

source/threads.tex
Dani-Hub: So do you wish me to replace "argument" by "parameter"?

source/threads.tex
Dani-Hub: OK.

source/threads.tex
Dani-Hub: I decided to perform this replacement

jensmaurer:
I've redone the patch:  #6519

It keeps the semantic linebreaks and removes "The argument/parameter" fluff entirely, which is consistent with at least half of the occurrences.

Dani-Hub:
> I've redone the patch: #6519

Thanks for your help.

JohelEGP:
> > > To get the PDF uploaded, you can edit https://github.com/cplusplus/draft/blob/main/.github/workflows/check.yml and swap the last two steps ("check-output.sh" and "upload PDF") at the end of the file.
> > 
> > 
> > But isn't that a setting that should be independently set from my PULL request?
> 
> No. In general, I think we can expect someone editing the draft to be able to build locally before uploading. And we only want "good" artifacts published from our continuous integration pipeline.

A fine alternative if one doesn't want to build it locally, or can't, is to do it in one's own fork.
Just enable GitHub Actions there.
In case the branch already has a PR for `origin`,
branch again to create a playground that doesn't make noise on that PR for `origin`,
then rebase the original branch once the PDF looks good.

JohelEGP:
This can be closed, as it's subsumed by the merged #6519.

PR #6519, opened 2023-08-27
[atomics] Reword preconditions on memory_order values in a positive form

### LINKED ISSUE #6511: [atomics.types.operations] p7 What if the supplied arguments does not denote any enumerator?
> void store(T desired, memory_order order = memory_order::seq_cst) volatile noexcept;
> void store(T desired, memory_order order = memory_order::seq_cst) noexcept;
>
> Preconditions: The order argument is neither memory_order​::​consume, memory_order​::​acquire, nor memory_order​::​acq_rel.

The LWG side asked me to open an editorial post about this issue, even though I have already tried to report an issue on CWG, and the CWG side(https://github.com/cplusplus/CWG/issues/398) has confirmed that this issue should be reported as an LWG issue. The original issue is:
````cpp
#include <atomic>
int main(){
   std::atomic<int> v{0};
   auto invented_ordering = static_cast<std::memory_order>(1024);
   v.store(1,invented_ordering);
}
````
[enum.dcl.enum] p8 says:
>  It is possible to define an enumeration that has values not defined by any of its enumerators. 

The `invented_ordering` does not violate the preconditions of the store, however, this is obviously problematic.

jensmaurer:
Fixes #6511

### DIFF
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -2968,10 +2968,10 @@ void store(T desired, memory_order order = memory_order::seq_cst) const noexcept
 \begin{itemdescr}
 \pnum
 \expects
-The \tcode{order} argument is neither
-\tcode{memory_order::consume},
-\tcode{memory_order::acquire}, nor
-\tcode{memory_order::acq_rel}.
+\tcode{order} is
+\tcode{memory_order::relaxed},
+\tcode{memory_order::release}, or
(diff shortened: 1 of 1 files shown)


Dani-Hub (approved):
Looks good.

tkoeppe:
Thanks, everyone!

jwakely (approved):

PR #6520, opened 2023-08-27
[temp.expl.spec] Turn paragraph into note

eisenwave:
This paragraph can and should be a note. It provides no new normative information. Everything it says is already said in greater detail in the paragraph it refers to, [[temp.fct.general] p2](http://eel.is/c++draft/temp.fct.general#2).

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -6621,8 +6621,10 @@ when one or more trailing template arguments are left unspecified.
 \end{note}
 
 \pnum
+\begin{note}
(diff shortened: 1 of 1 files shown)


tkoeppe:
I don't think there's a problem with this wording not being a note. I'm also not sure if there is really no information in it: it might not be clear that a conversion from a type to itself is in fact a conversion, and so the fact that copy/move constructors count as converting constructors might well benefit from being spelled out in non-optional wording.

PR #6526, opened 2023-08-28
[vector.data],[array.members] Clarify C++ boolean expression

Eisenwave:
As far as I know, it is not enough to say
> For a non-empty array, [C++ boolean expression]

It is not obvious that this should be automatically interpreted as [C++ boolean expression] is `true` in all cases.

This edit fixes it by explicitly saying that the expression is `true`, which looks to be the intention here.


jwakely (approved):
We could also just phrase it as "`x` is equal to `y`" instead of "`x == y` is `true`", but this change is correct, which is better than the current wording, thanks.

jensmaurer (approved):

PR #6527, opened 2023-08-28
[expr.call] Add further forward references

Eisenwave:
This paragraph is reaching ahead quite a bit; talking about scoped enumerations, eligible special member functions, non-trivial special member functions, etc.

This edit adds the relevant forward references. What originally motivated me is that the current wording contains `non-trivial destructor\iref{special}`, which looked like a mistake. This forward reference is intended to apply to `eligible`.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -3408,13 +3408,13 @@ An argument that has type \cv{}~\tcode{std::nullptr_t} is converted
 to type \tcode{\keyword{void}*}\iref{conv.ptr}.
 After these conversions, if the
 argument does not have arithmetic, enumeration, pointer, pointer-to-member,
-or class type, the program is ill-formed. Passing a potentially-evaluated
-argument
-of a scoped enumeration type or
-of a class type\iref{class} having an eligible non-trivial
-copy constructor, an eligible non-trivial move constructor,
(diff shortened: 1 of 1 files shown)


source/expressions.tex
```diff
@@ -3408,13 +3408,13 @@
...
-of a scoped enumeration type or
-of a class type\iref{class} having an eligible non-trivial
-copy constructor, an eligible non-trivial move constructor,
-or a
-non-trivial destructor\iref{special},
+or class type, the program is ill-formed.
+Passing a potentially-evaluated argument
+of a scoped enumeration type\iref{dcl.enum} or
+of a class type\iref{class} having
+an eligible\iref{special} non-trivial copy constructor\iref{class.copy.ctor},
```
jensmaurer, line 3415: I don't like splitting adjective + noun with a cross-reference if we can help it.
I think it's good enough if we use \iref{special,class.copy.ctor} after the phrase.

jensmaurer (approved):
Mostly harmless, I guess.

jensmaurer:
@tkoeppe, this feels like a mild improvement to me.

tkoeppe:
Indeed, thank you!

PR #6528, opened 2023-08-28
[expr.call] Turn recursive call wording into note

Eisenwave:
This paragraph doesn't have to be normative. [basic.start.main] already explains that the `main` function can't be used.

The reader also has no reason to believe that recursive calls are not allowed, and no wording forbids them generally, making this paragraph unnecessary from a normative POV.


tkoeppe:
"are permitted" is permission, which is not permitted in notes.

PR #6529, opened 2023-08-28
[cpp] `\grammarterm` overhaul

### LINKED ISSUE #6524: [cpp] "identifier" should be `\grammarterm{identifier}` in more cases
The [[cpp]](https://eel.is/c++draft/cpp) section talks about "identifier"s frequently; often in a context where direct statements about grammar and macro expansion are made.

It would be appropriate to use `\grammarterm{identifier}` in many more places. For example:

> A preprocessing directive of the form
> 
> `#` `undef` *identifier* *new-line*
> 
> causes the specified identifier no longer to be defined as a macro name.
> It is ignored if the specified identifier is not currently defined as a macro name.

\- [[cpp.scope] p2](https://eel.is/c++draft/cpp.scope#2)

This is one of the more extreme cases: statements directly about grammar rules are made, but "identifier" is not formatted as a grammatical term.

eisenwave:
Fixes https://github.com/cplusplus/draft/issues/6524.

This edit consistently uses `\grammarterm` formatting for:
- *identifier*
- *preprocessing-token*
- *replacement-list*
- *group*
- *string-literal*
- *integer-literal*

No normative wording changes are made.

### DIFF
--- a/source/preprocessor.tex
+++ b/source/preprocessor.tex
@@ -134,10 +134,11 @@
 \end{bnf}
 
 \pnum
-A \defn{preprocessing directive} consists of a sequence of preprocessing tokens
+A \defn{preprocessing directive} consists of a sequence of
+\grammarterm{preprocessing-token}s
 that satisfies the following constraints:
 At the start of translation phase 4,
-the first token in the sequence,
+the first \grammarterm{preprocessing-token} in the sequence,
 referred to as a \defnadj{directive-introducing}{token},
 begins with the first character in the source file
 (optionally after whitespace containing no new-line characters) or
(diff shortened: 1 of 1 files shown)


jensmaurer (commented):
I think I like that, but it takes us farther away from the C formulation. I'd like to request CWG consensus for this change.

AlisdairM:
There has been a reasonable amount of work on the preprocessor clause for C++26, so I would personally love to see this rebased and considered for the FDIS as part of that work and clean-up.

eisenwave:
Too much effort to rebase at this point.

PR #6531, opened 2023-08-29
[format.string.std] Fix example

### LINKED ISSUE #6474: [format.string.std]/4: sC example doesn't have enough clowns
https://github.com/cplusplus/draft/blob/a272b7cd04dc0be488250a0c6aaead160e778774/source/utilities.tex#L15920

The string being formatted should have three clowns, not one.

tkoeppe:
Fixes #6474.

@timsong-cpp, @tahonermann FYI

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -15901,8 +15901,6 @@ are both absent.
 The \fmtgrammarterm{align} option applies to all argument types.
 The meaning of the various alignment options is as specified in \tref{format.align}.
 \begin{example}
-%FIXME: example is incomplete, sB and sC result in:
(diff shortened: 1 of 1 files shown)


tahonermann:
Thanks @tkoeppe, looks right to me as far as I can tell. My LaTeX skills are too lacking to evaluate the correctness of the markup though. Will this change be incorporated into the DIS?

JohelEGP:
You can check out the uploaded PDF in the Checks tab, and looking for "Artifacts" on the right.

tahonermann:
> You can find the uploaded PDF in the Checks tab, and then looking for "Artifacts" on the right.

Thank you. It looks right to me!

PR #6534, opened 2023-08-30
[input.output] Fix missing `is \keyword{true}` or `is \keyword{false}`

eisenwave:
Similar fixes to https://github.com/cplusplus/draft/pull/6526

I've realized halfway through just how much work there is to do, and so I'd like some feedback on how to proceed first.

My plan is to cherry-pick the commits and put each into a separate PR so that they can be examined and discussed separately. After all, there are some wording changes, it's not *just* changing formatting for some words.


source/iostreams.tex
```diff
@@ -1768,29 +1768,31 @@
...
  \tcode{+} offset   &
- \remarks With \tcode{ql = p + o;}, then: \tcode{ql - o == p}  \\ \rowsep
+ \remarks With \tcode{ql = p + o;}, then:
+   \tcode{ql - o == p} is \keyword{true}  \\ \rowsep
 \tcode{pl += o} &
  \tcode{P\&} &
  \tcode{+=} offset &
  \remarks With \tcode{ql = pl;} before the \tcode{+=}, then:
-   \tcode{pl - o == ql} \\ \rowsep
+   \tcode{pl - o == ql} is \keyword{true} \\ \rowsep
```
jwakely, line 1777:
```suggestion
   \tcode{pl - o} equals \tcode{ql} \\ \rowsep
```

And similarly below

source/iostreams.tex
```diff
@@ -2034,7 +2036,7 @@
 \tcode{ios_base::imbue(loc)}\iref{ios.base.locales}
 and
 if
-\tcode{rdbuf() != 0}
+\tcode{rdbuf() != 0} is \keyword{true}
```
jwakely, line 2039: rbduf() is null, or something like that

jwakely (commented):
Nobody likes the requirement to say "is true" in places like this, so it would be better to use wording like "a equals b" instead of the clunky "a == b is true"

eisenwave:
Too big of a change for one PR, and subject to regression.

Maybe if we have checks in the future that can be toggled within certain regions of the markup, we could do some automatic detection, but for now we probably have more important things to focus on.

PR #6539, opened 2023-08-30
[expr.unary.noexcept] Replace informative wording

eisenwave:
### Context and Problem with the Current Wording
I was briefly under the impression that `a + b` can throw for two `int`s `a` and `b`, because signed integer overflow is UB, and UB can result in throwing exceptions, and maybe the `noexcept` operator needs to consider that.

Paragraph 1 in its current form says:
> determines whether the evaluation of its operand [...] can throw an exception

An unsuspecting reader like me might assume that this includes exceptions thrown as the result of UB. However, this isn't actually possible, and this wording is purely informative. Paragraph 3 contains the *actual normative* wording, explaining that the `noexcept` operator checks whether an expression is potentially-throwing.

### Suggested Solution
This PR moves the *actual normative* wording from p3 up into p1 to avoid any future confusion. The informative wording is removed.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -5052,13 +5052,13 @@ which is an unevaluated operand\iref{term.unevaluated.operand}.
 The \keyword{noexcept} operator yields \keyword{true}
 unless the full-expression of the operand
 is potentially-throwing\iref{except.spec}.
-
-\pnum
-The result of the \keyword{noexcept} operator is a prvalue of type \keyword{bool}.
 \begin{note}
 A \grammarterm{noexcept-expression}
 is an integral constant expression\iref{expr.const}.
 \end{note}
+
+\pnum
+The result of the \keyword{noexcept} operator is a prvalue of type \keyword{bool}.
 
 \pnum
 If the operand is a prvalue,
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -5042,14 +5042,16 @@ element type.
 \pnum
 \indextext{\idxcode{noexcept}}%
 \indextext{expression!\idxcode{noexcept}}%
-The \keyword{noexcept} operator determines whether the evaluation of its operand,
-which is an unevaluated operand\iref{term.unevaluated.operand}, can throw an
-exception\iref{except.throw}.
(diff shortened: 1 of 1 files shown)


source/expressions.tex
```diff
@@ -5042,10 +5042,11 @@
...
-The \keyword{noexcept} operator determines whether the evaluation of its operand,
-which is an unevaluated operand\iref{term.unevaluated.operand}, can throw an
-exception\iref{except.throw}.
-
+The \keyword{noexcept} operator yields \keyword{true}
+unless the full-expression of the operand
+is potentially-throwing\iref{except.spec}.
+The operand of the \keyword{noexcept} operator
+is an unevaluated operand\iref{term.unevaluated.operand}.
 \begin{bnf}
```
jensmaurer, line 5050: I think there is some room for improvement in this subclause, but talking about "the operand" in detail before we've even introduced the grammar seems unhelpful.

My suggestion:
- grammar first
- then talk about the operand
- then talk about the result
- then have the note about constant expression

jensmaurer (changes requested):

Eisenwave:
image
updated version

source/expressions.tex
Eisenwave: I've implemented most of what you've suggested, but I don't understand what you mean by "constant expression". Neither the old wording nor the new wording talks about constant expressions.

jensmaurer: "[Note 1 : A noexcept-expression is an integral constant expression (7.7). — end note]"

Eisenwave: Ah yes, I wasn't looking at the second paragraph. I've moved it up now.

jensmaurer:
Sorry, but that's not exactly what I had in mind. Since the change is a little larger, I've created CWG2792 to have CWG review the change. See there for my suggestion; feel free to continue to discuss here.

AlisdairM:
The intent of this PR seems to have been resolved by adopting the resolution of https://cplusplus.github.io/CWG/issues/2792.html and should probably close it.

PR #6546, opened 2023-09-01
[class.copy.elision] Improve reference and replace informal term

Eisenwave:
The term "exception handler" isn't defined anywhere, so it is a mistake to refer to it.

Just a few bullets up, the wording refers to the term in the grammar, making the subsequent bullet inconsistent
https://github.com/cplusplus/draft/blob/1339fdd98ed04a5077e7dad9ed356eb0302be8aa/source/classes.tex#L6213-L6216

Also, `except.handle` is a more accurate reference than `except.pre`.


jensmaurer (approved):
Looks like an improvement.

PR #6547, opened 2023-09-01
[res.on.exception.handling] Use `\grammarterm` instead of informal term

Eisenwave:
The term "exception handler" isn't formally defined anywhere. It would be better to use the grammatical term *handler*, which is defined.

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -3772,7 +3772,7 @@ Any of the functions defined in the \Cpp{} standard library
 can report a failure by throwing an exception of a type
 described in its \throws paragraph,
 or of a type derived from a type named in the \throws paragraph
-that would be caught by an exception handler for the base type.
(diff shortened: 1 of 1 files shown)


jensmaurer (approved):
Looks good. Icing on the cake: add xref to [except.handle] or so.

tkoeppe:
@Eisenwave Could you apply Jens's suggestion/

Eisenwave:
@jensmaurer I'm not 100% sure if I put it in the place you've envisioned. Also I assume you mean `\ref`, not `\xref`.

JohelEGP:
It should be `\xref`.

tkoeppe:
I think it should be `\iref`. Jens meant "xref" as a shorthand for "cross reference", but the relevant macro is `\iref`. (The `\xref` macro is something else.)

PR #6557, opened 2023-09-08
[temp.dep.expr] Introduce nested bullets for clarity

### LINKED ISSUE #6553: [temp.dep.expr] Weird circumventing of nested bullets
If you read [[temp.dep.expr] p3](http://eel.is/c++draft/temp.dep.expr#3), you notice that the paragraph is actually three big bullets.

- one implicit bullet in the preamble before the current list of bullets
- the current list of bullets, which should really be nested inside a new outer list
- another implicit bullet at the end

jensmaurer:
Fixes #6553

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -5072,8 +5072,12 @@ dependent\iref{temp.dep.type}.
 
 \pnum
 An \grammarterm{id-expression} is type-dependent
-if it is a \grammarterm{template-id} that is not a concept-id and is dependent;
-or if its terminal name is
+if
+\begin{itemize}
+\item
+it is a \grammarterm{template-id} that is not a concept-id and is dependent; or
+\item
+its terminal name is
 \begin{itemize}
 \item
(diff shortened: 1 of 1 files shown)


source/templates.tex
```diff
@@ -5072,8 +5072,12 @@
 
 \pnum
 An \grammarterm{id-expression} is type-dependent
-if it is a \grammarterm{template-id} that is not a concept-id and is dependent;
-or if its terminal name is
+if
+\begin{itemize}
+\item
+it is a \grammarterm{template-id} that is not a concept-id and is dependent; or
```
tkoeppe, line 5078: @jensmaurer Aha, another list item ending in a semicolon!

jensmaurer: @tkoeppe , what do you mean?

We (now) have a nested bulleted list with comma separators and a top-level bulleted list with semicolon separators.  Feels acceptable (and has precedent).

PR #6561, opened 2023-09-11
[mdspan] require conversion results to be nonnegative

CaseyCarter:
"nonnegative" is meaningless for a value of some arbitrary type which we've only required to be convertible to the integral type `index_type`, so this wording clearly intends to constrain the result of the conversion.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -20513,7 +20513,8 @@ template<class OtherIndexType>
 \expects
 \begin{itemize}
 \item
-\tcode{s[$i$] > 0} is \tcode{true}
+The result of converting \tcode{s[$i$]} to \tcode{index_type}
+is greater than \tcode{0}
 for all $i$ in the range $[0, \exposid{rank_})$.
 \item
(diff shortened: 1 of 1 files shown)


CaseyCarter:
"`meow` is representable as a value of type `T`" also verges on meaningless for an expression `meow` of arbitrary type, but I don't think that can be fixed editorially.

CaseyCarter:
I've pushed two more commits fixing the same issue in a couple more places, with the expectation it will be easiest to review these together. The commits are fine-grained for ease of review and each follow the editorial guidelines so it should be possible to merge them cleanly without squashing.

Shout if I made the wrong call and I should split these up into three PRs.

CaseyCarter:
I should tag @jwakely for review. I know I'm near the boundary of what qualifies as editorial, and _I_ think I'm on the right side of that boundary, but someone more objective should judge whether LWG needs to look at this.

jwakely (approved):
LGTM. I don't think there's any doubt about what was intended, and this change expresses that more accurately.

jensmaurer (approved):

PR #6565, opened 2023-09-12
[mdspan.layout.stride.cons] Fix cross-reference

CaseyCarter:
The text says that "`StridedLayoutMapping` meets the layout mapping requirements" but the cross-reference is to the layout mapping _policy_ requirements.


jensmaurer (approved):

CaseyCarter:
Looks like the MacOS `make` timed out after 25 minutes. Does someone want to rerun that job?

PR #6567, opened 2023-09-19
[atomics.order] Memory operations should be definitions

eisenwave:
Related: [Stack Overflow / Where is the definition of the acquire operation in the C++ 20?](https://stackoverflow.com/q/77132246/5740428)

OP in this Q&A was confused by operations involving `memory_order::acquire` referring to "acquire operation", but this term isn't defined anywhere. It is indeed confusing because "acquire operation" is colloquially used throughout the standard (e.g. [[intro.races] p3](https://eel.is/c++draft/intro.races#3)).

I believe the proper resolution is to tie the definition to the listing of memory orders.

### DIFF
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -2483,10 +2483,10 @@ enumerated values and their meanings are as follows:
 \item \tcode{memory_order::relaxed}: no operation orders memory.
 
 \item \tcode{memory_order::release}, \tcode{memory_order::acq_rel}, and
-\tcode{memory_order::seq_cst}: a store operation performs a release operation on the
+\tcode{memory_order::seq_cst}: a store operation performs a \defn{release operation} on the
 affected memory location.
 
-\item \tcode{memory_order::consume}: a load operation performs a consume operation on the
+\item \tcode{memory_order::consume}: a load operation performs a \defn{consume operation} on the
 affected memory location.
 \begin{note}
 Prefer \tcode{memory_order::acquire}, which provides stronger guarantees
@@ -2496,7 +2496,7 @@ Specification revisions are under consideration.
 \end{note}
 
 \item \tcode{memory_order::acquire}, \tcode{memory_order::acq_rel}, and
-\tcode{memory_order::seq_cst}: a load operation performs an acquire operation on the
+\tcode{memory_order::seq_cst}: a load operation performs an \defn{acquire operation} on the
 affected memory location.
 \end{itemize}
 


languagelawyer:
Maybe they should be `\defnadj{X}{operation}`

jwakely:
I don't think [atomics.order] p1 provides definitions, it says that a store operation with one of those memory orders performs a release operation. That's not the same as saying a release operation **is** a store operation with one of those memory orders. Unlocking a mutex is a release operation, but it's not a store to an atomic variable using one of those memory orders. [intro.races] specifically says that the library defines atomic operations **and operations on mutexes** which  are synchronization operations, and then says that every synchronization operation is either a consumer operation, acquire operation, release operation, or acquire+release operation. If we use [atomics.order] as the definitions, that means that every synchronization operation (including those on mutexes) must be either a store or load of an atomic variable using an appropriate `memory_order` constant, but that's not how mutexes are defined in the standard.

I agree that we don't have a precise definition of the term "release operation" etc. but I don't think we can just add italics to turn that text into the desired definition.

Eisenwave:
@jwakely [[thread.mutex.requirements]](https://eel.is/c++draft/thread.mutex.requirements) doesn't actually mention "release operation" or "acquire operation", but it does say:

> The implementation provides lock and unlock operations, as described below.
> For purposes of determining the existence of a data race, these behave as atomic operations.

So there isn't any issue with saying that both atomic objects and mutexes use the definition in [atomics.order]. Maybe the wording which [intro.races] needs to be changed as well so it no longer mentions mutexes.

Eisenwave:
Furthermore, [[intro.races] p3](https://eel.is/c++draft/intro.races#3) doesn't look normative to me. It's just an introductory paragraph that lists the contents of various library parts. The entire thing could probably be a note, so it shouldn't make or break whether we can turn something into a definition.

jwakely:
[atomics.order] still talks about load and store operations, which aren't defined for mutexes.

Eisenwave:
> [atomics.order] still talks about load and store operations, which aren't defined for mutexes.

I don't think it's a problem with [atomics.order]; I think it's a problem with the wishy washy statement
>  For purposes of determining the existence of a data race, these behave as atomic operations.

... without any details as to what mutex operation maps onto what sort of atomic operation with what memory order. The solution is probably to elaborate on this and to say that:
- `std::mutex::unlock()` behaves like a release operation as if the mutex was an atomic object
- `std::mutex::lock()` behaves like an acquire operation as if the mutex was an atomic object

PR #6568, opened 2023-09-19
[range.utility.conv.to] Add terminating condition for first bullet

jwakely:
We currently fail to say what happens if the first bullet is true, but then none of its sub-bullets is true.


jwakely:
I think this qualifies as editorial, but let me know if you want me to handle it as an LWG issue instead.

CaseyCarter (approved):

PR #6574, opened 2023-09-20
[expr.log.and],[expr.log.or] improve wording symmetry and quality

eisenwave:
The wording for `&&` and `||` had a number of issues which this PR resolves.

1. The wording of `&&` and `||` is oddly asymmetric in a number of places. Specific wording choices like using `:` in one paragraph, and `; morever,` are asymmetric with no good reason.
2. There are notes regarding `&` left in normative wording. These are extracted into an actual note.
3. In both subclauses, the first and second paragraph contain a mixture of describing the result (type) and describing evaluation. This is fixed by clearly separating it so that:
   - the first paragraph talks about the result
   - the second paragraph talks about left-to-right evaluation and short-circuiting
4. The second paragraph needlessly says that the result is `bool`. The first paragraph already states that the result is `true` or `false`, so this information is redundant.
5. The wording contains the phrase "the expression is sequenced" which doesn't make any sense because an expression is a syntactical construct and cannot be sequenced. Only an evaluation can be sequenced.
6. The paragraphs contain a mixture of first/second terminology and left/right terminology. The PR uses left/right terminology in all places.
7. Other places in the standard contain notes about "the short-circuiting behavior" of the `&&` operator. However, this is not noted in the description of the `&&` and `||` operator. This information is added to a note.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -6810,20 +6810,21 @@ The result is the bitwise inclusive \logop{or} function of the operands.
 \end{bnf}
 
 \pnum
-The \tcode{\&\&} operator groups left-to-right. The operands are both
-contextually converted to \keyword{bool}\iref{conv}.
-The
-result is \keyword{true} if both operands are \keyword{true} and
-\keyword{false} otherwise. Unlike \tcode{\&}, \tcode{\&\&} guarantees
-left-to-right evaluation: the second operand is not evaluated if the
-first operand is \keyword{false}.
+The \tcode{\&\&} operator groups left-to-right.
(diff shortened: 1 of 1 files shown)


Eisenwave:
image

jensmaurer:
[intro.execution] defines sequencing of an expression, and that expressly means both value and side effects. Your  rewrite is unclear about that. Please revert.

Also, I really would like to retain a statement what value category and what type a logical-and-expression has. In particular, we don't get to the "value" part unless we actually evaluate the expression. Saying the expression is a prvalue of type bool would address this concern.

PR #6577, opened 2023-09-22
[temp.res.general] Fix misleading example related to syntax errors

Eisenwave:
The use of the word "may" suggests that an implementation isn't required to diagnose `+;` if it's in a template, but this is not true. Diagnostics are always required for syntax errors.

- `+;` is not a valid *[statement]*; therefore
- `{ +; }` is not a valid *[compound-statement]* and not a valid *[function-body]*; therefore
- `void g() { +; }` is not a valid *[function-definition]* and not a valid *[member-declaration]*; therefore
- the definition of `X` is not a valid *[template-declaration]* and not an element of the *[declaration-seq]* of the TU; therefore
- the entire *[translation-unit]* is invalid

As you can see, a syntax error inevitably "bubbles upwards" in the grammar, and also means that any containing template isn't a template in the first place. If there is no containing template, and there isn't even a valid *[translation-unit]*, then diagnostics are required because syntax errors fall under the umbrella of [diagnosable rules].

Unless "no diagnostic required" was explicitly stated, the following applies:

> [...] if a program contains a violation of any diagnosable rule or an occurrence of a construct described in this document as “conditionally-supported” when the implementation does not support that construct, **a conforming implementation shall issue at least one diagnostic message**.

\- [[intro.compliance.general] p2.3](https://eel.is/c++draft/intro.compliance.general#2.3)

[statement]: https://eel.is/c++draft/stmt.pre#nt:statement
[compound-statement]: https://eel.is/c++draft/stmt.block#nt:compound-statement
[function-body]: https://eel.is/c++draft/dcl.fct.def.general#nt:function-body
[function-definition]: https://eel.is/c++draft/dcl.fct.def.general#nt:function-definition
[member-declaration]: https://eel.is/c++draft/class.mem.general#nt:member-declaration
[template-declaration]: https://eel.is/c++draft/temp.pre#nt:template-declaration
[declaration-seq]: https://eel.is/c++draft/dcl.pre#nt:declaration-seq
[translation-unit]: https://eel.is/c++draft/basic.link#nt:translation-unit

[diagnosable rules]: https://eel.is/c++draft/intro.compliance.general#def:diagnosable_rules

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -4534,7 +4534,7 @@ template<class T> class X {
     X<T>::h();      // may be diagnosed even if \tcode{X::f} is not instantiated
   }
   void g(T t) {
-    +;              // may be diagnosed even if \tcode{X::g} is not instantiated
+    +;              // must be diagnosed even if \tcode{X::g} is not instantiated
   }
 };
 


eisenwave:
@jensmaurer thoughts? Does CWG need to see this or am I simply right or wrong on this?

The question boils down to: do we require diagnostics for token soup in templates?

jensmaurer:
It is my understanding that the standard wishes to support implementations that stash away a token stream (e.g. based on matching braces) and do not form an AST while parsing a template definition.  I believe that, at some point in the past, MSVC was such an implementation.

I agree that the grammar presentation is not amenable to this view, but repairing this is probably a very large effort.  Given that all implementations are now, in fact, using an AST representation during template parsing, maybe we can tighten the rules here and actually make `+;` ill-formed in template definitions.  I would like to see a normative change here.

Additional considerations: The C++ grammar is not context-free and there is a possibility that instantiation actually changes the AST slightly (as specified; maybe not as implemented). For example, a "T" template parameter is a _type-name_, but might (does it?) become a _class-name_ during instantiation. We're not very clear on what happens during instantiation anyway. (We have "typename" and "template" prefixes and various scattered rules to prevent large-scale changes to the AST during instantiation.)

In short, disallowing token-stream implementations should be in a paper and should clearly lay out where the new boundary for required vs. optional diagnostics for the template definition is, for example by amending the IFNDR rule for uninstantiated template definitions to refer to "anything that matches the grammar, but cannot be successfully instantiated" or similar.

eisenwave:
Alright, I may write a paper that addresses this problem at some point.

t3nsor:
In the current wording, the real normative issue here is that once you put

```
template<class T> class X {
  void g(T t) {
    +;
  }
};
```

somewhere in your program, it nullifies the need for the implementation to issue *any* diagnostics, even for syntax errors that are inside some unrelated non-templated entity. But, somehow, if you instantiate `X`, then the IFNDR goes away: the `+;` has to be diagnosed and so do syntax errors elsewhere in the program. Very odd.

If we said that the above template is unspeciformed instead of IFNDR, that would partially solve the problem, because IFNDR trumps all other rules requiring diagnostics but unspeciformed presumably would not have that effect. Then, it would be clear that the failure of the program to be valid C++ according to the grammar is diagnosable (since that's a rule about the template itself, not about the instantiations of the template).

PR #6579, opened 2023-09-25
Add memory order for atomic_flag_test_and_set and atomic_flag_clear.

xmcgcg:
The memory order for atomic_flag_test_and_set and atomic_flag_clear is not specified, it should be memory_order::seq_cst. (atomic_flag_test and atomic_flag_wait specify that)


jensmaurer:
This feels like an LWG issue, not an editorial one.

PR #6580, opened 2023-09-26
[temp.over.link] Fix cross-reference introduced by P1787R6

jensmaurer:
As discussed on the CWG reflector.


source/templates.tex
```diff
@@ -3685,7 +3685,7 @@
 \begin{note}
 If such a dependent name is unqualified,
 it is looked up from
-the first declaration of the function template\iref{temp.dep.candidate}.
+the first declaration of the function template\iref{temp.res.general}.
```
zygoloid, line 3688: "the" first declaration also doesn't seem quite right.
```suggestion
a first declaration of the function template\iref{temp.res.general}.
```
maybe?

jensmaurer:
@tkoeppe, I'd like to merge this rather soon.

PR #6583, opened 2023-09-29
[fs.filesystem.syn] Remove redundant inline

hewillk:
This is reasonable, right?


hewillk:
Closed with incorrect changes, thanks.

PR #6593, opened 2023-10-05
[temp.param] Introduce term to xref structural type

AlisdairM:
The current cross-references to [temp.param] appear confusing, as the structural type definition is buried a couple of pages below.  Also, this change looks clearer in the source.


tkoeppe:
Thanks!

PR #6594, opened 2023-10-05
[expr.prim.lambda.closure] insert an extra pnum

AlisdairM:
Paragraph 2 does two things.  The part before the note specifies the scope where the closure type is defined.  The part after the note places constraints on how the closure type can be implemented.  Adding a paragraph number after the note seems to be more readable, and referencable in the future.


jensmaurer (approved):
Why not.

PR #6602, opened 2023-10-06
[mdspan.layout.left] Add missing `noexcept`

hewillk:
..as declaration has `noexcept` qualifier. Is this editorial or LWG?

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -20017,7 +20017,7 @@ return ((static_cast<index_type>(i) * stride(P)) + ... + 0);
 
 \indexlibrarymember{stride}{layout_left::mapping}%
 \begin{itemdecl}
-constexpr index_type stride(rank_type i) const;
+constexpr index_type stride(rank_type i) const noexcept;
 \end{itemdecl}
 
 \begin{itemdescr}


CaseyCarter (approved):

CaseyCarter:
> Is this editorial or LWG?
If the Standard disagrees with the ingress paper ([P0009R18](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p0009r18.html)) it's editorial. In this case, they agree. That generally makes it an LWG issue to determine which one is correct, unless you can present a convincing argument to the editors.

I'll point out that both `layout_left::mapping::extents` and the exposition-only `extents::fwd-prod-of-extents` are `noexcept`, so computing the return value `extents().fwd-prod-of-extents(i)` is not potentially-throwing.

hewillk:
I saw that many functions that accept indices in `<mdspan>` have the constraints of `(is_convertible_v && is_nothrow_constructible_v)`. I wonder why not just simply use `is_nothrow_convertible_v` directly, which seems to be more accurate to me.

hewillk:
> I saw that many functions that accept indices in `<mdspan>` have the constraints of `(is_convertible_v && is_nothrow_constructible_v)`. I wonder why not just simply use `is_nothrow_convertible_v` directly, which seems to be more accurate to me.

CC @crtrott for insights.

tkoeppe:
That said, if this is not an editorial misapplication of the approved paper, I would prefer if you filed an LWG issue. If you provide rationale and a resolution, it might get approved very quickly.

tkoeppe:
Please file an LWG issue, or reopen if there's new information.

jwakely:
> I saw that many functions that accept indices in `<mdspan>` have the constraints of `(is_convertible_v && is_nothrow_constructible_v)`. I wonder why not just simply use `is_nothrow_convertible_v` directly, which seems to be more accurate to me.

But they're not equivalent:

```c++
#include <type_traits>
struct Tricksy {
    operator int() const;
    explicit operator unsigned() noexcept;
};
static_assert( std::is_convertible_v<Tricksy, unsigned> );
static_assert( std::is_nothrow_constructible_v<unsigned, Tricksy> );
static_assert( ! std::is_nothrow_convertible_v<Tricksy, unsigned> );
```

We explicitly construct `index_type` and that must be non-throwing, so `is_nothrow_constructible` is correct. But we also want the types to be implicitly convertible to `index_type` to disallow annoying, hard-to-use correctly types with weirdo conversion sequences.

But just requiring `is_nothrow_convertible` would not be accurate, because the conversion to `index_type` performed by `extents` etc. is explicit, not an implicit conversion.

PR #6603, opened 2023-10-07
[intro.scope] Clarify 'they'

jensmaurer:
Fixes ISO/CS 034 (C++23 DIS).

Fixes cplusplus/nbballot#576

### DIFF
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -6,7 +6,7 @@
 \indextext{scope|(}%
 This document specifies requirements for implementations
 of the \Cpp{} programming language. The first such requirement is that
-they implement the language, so this document also
+an implementation implements the language, so this document also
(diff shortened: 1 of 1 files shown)


jwakely (approved):
The change seems fine to me, but if we wanted to avoid "an implementation implements" we could also say "The first such requirement on implementations is that they implement the language" which makes it clear who "they" are.

jensmaurer:
I've had my suggestion approved by the CWG telecon on Friday, and everybody present felt that "they" was unambiguously referring to the first preceding plural, and thus the comment was unwarranted.

Given the general fluffiness of the text here, I feel I shouldn't deviate from CWG's approved approach.

tkoeppe:
Make sense, thanks!

PR #6604, opened 2023-10-07
[expr.const] Amend comments in example

jensmaurer:
Fixes NB JP 023 (C++23 DIS).

Fixes cplusplus/nbballot#565

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -7984,12 +7984,13 @@ int x = 0;
 
 template<class T>
 constexpr T h(T t = id(x)) {    // \tcode{h<int>} is not an immediate function
+                                // \tcode{id(x)} is not evaluated when parsing the default argument\iref{dcl.fct.default,temp.inst}
     return t;
 }
 
 template<class T>
-constexpr T hh() {              // \tcode{hh<int>} is an immediate function
-  return h<T>();
+constexpr T hh() {              // \tcode{hh<int>} is an immediate function because of the invocation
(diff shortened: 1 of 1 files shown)


source/expressions.tex
```diff
@@ -7984,12 +7984,13 @@
...
 constexpr T h(T t = id(x)) {    // \tcode{h<int>} is not an immediate function
+                                // \tcode{id(x)} is not evaluated when parsing the default argument\iref{dcl.fct.default,temp.inst}
     return t;
 }
 
 template<class T>
-constexpr T hh() {              // \tcode{hh<int>} is an immediate function
-  return h<T>();
+constexpr T hh() {              // \tcode{hh<int>} is an immediate function because of the invocation
+  return h<T>();                // of the immediate function \tcode{id} in the default argument of \tcode{h<int>}
```
cor3ntin, line 7993:
```suggestion
constexpr T hh() {              // \tcode{hh<int>} is an immediate function because the default argument of 
  return h<T>();                   // \tcode{h<int>}  contains a call to an immediate function whose evaluation would not be a constant expression
```

I think it's worse pointing out that `id(x)` is purposefully not valid at this point in this example

jensmaurer: Well, at this point, the part "whose evaluation would not be a constant expression" is not yet relevant, which seems important. We just use the "consteval" property of "id" to say that we're upgrading to an immediate function.

It's only when we try to evaluate (a little later) that we hit the "oops, it's not constexpr" situation.

PR #6611, opened 2023-10-08
Notes with requirements

jensmaurer:
Fixes cplusplus/nbballot#581


tkoeppe:
Great, thanks! Could you perhaps reorder the commits so that the tool update comes last?

PR #6614, opened 2023-10-10
[defns.{impl.limits,iostream.templates}] Use singular

tkoeppe:
Also change "argument" to "parameter" in the note.

Fixes ISO/CS 015 (C++23 DIS).

Fixes https://github.com/cplusplus/nbballot/issues/557


CaseyCarter (approved):

tkoeppe:
Let's see how that other discussion goes. I wasn't sure whether we'll delete or repair.

PR #6615, opened 2023-10-10
Replace "C Standard" with "ISO/IEC 9899:2018" and define "C standard library"

tkoeppe:
Fixes ISO/CS 035 (C++23 DIS).

Partially addresses https://github.com/cplusplus/nbballot/issues/544.
Fixes https://github.com/cplusplus/nbballot/issues/577.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -4728,9 +4728,8 @@ a set of bits in the object representation that determines a
 \defn{value}, which is one discrete element of an
 \impldef{values of a trivially copyable type} set of values.
 \begin{footnote}
-The
-intent is that the memory model of \Cpp{} is compatible
-with that of ISO/IEC 9899 Programming Language C.
+The intent is that the memory model of \Cpp{} is compatible
+with that of the C programming language.
 \end{footnote}
 
 \pnum
@@ -5276,7 +5275,7 @@ that are in addition to the names defined in the \libheader{stdfloat} header
 should be chosen to increase compatibility and interoperability
 with the interchange types
 \tcode{_Float16}, \tcode{_Float32}, \tcode{_Float64}, and \tcode{_Float128}
-defined in ISO/IEC TS 18661-3 and with future versions of the C standard.
+defined in ISO/IEC TS 18661-3 and with future versions of \IsoCUndated{}.
 
 \rSec2[basic.compound]{Compound types}
 
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -12,9 +12,8 @@ requirement appear at various places within this document.
 
 \pnum
 \Cpp{} is a general purpose programming language based on the C
-programming language as described in ISO/IEC 9899:2018
-\doccite{Programming languages --- C} (hereinafter referred to as the
-\defnx{C standard}{C!standard}). \Cpp{} provides many facilities
+programming language as described in \IsoC{}.
+\Cpp{} provides many facilities
 beyond those provided by C, including additional data types,
 classes, templates, exceptions, namespaces, operator
 overloading, function name overloading, references, free store
@@ -37,7 +36,7 @@ For undated references, the latest edition of the referenced document
 \item ISO/IEC 2382, \doccite{Information technology --- Vocabulary}
 \item ISO 8601:2004, \doccite{Data elements and interchange formats ---
 Information interchange --- Representation of dates and times}
-\item ISO/IEC 9899:2018, \doccite{Information technology --- Programming languages --- C}
+\item \IsoC{}, \doccite{Information technology --- Programming languages --- C}
 \item ISO/IEC/IEEE 9945:2009, \doccite{Information Technology --- Portable
 Operating System Interface (POSIX
 \begin{footnote}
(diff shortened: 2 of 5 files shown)


source/basic.tex
```diff
@@ -4728,9 +4728,8 @@
 \defn{value}, which is one discrete element of an
 \impldef{values of a trivially copyable type} set of values.
 \begin{footnote}
-The
-intent is that the memory model of \Cpp{} is compatible
-with that of ISO/IEC 9899 Programming Language C.
+The intent is that the memory model of \Cpp{} is compatible
+with that of \IsoCUndated{}.
```
zygoloid, line 4732: I think this is talking about the programming language, not about the ISO publication, so maybe
```suggestion
with that of the C programming language.
```

source/lib-intro.tex
```diff
@@ -3715,8 +3714,7 @@
 Functions from the C standard library shall not throw exceptions%
 \indextext{specifications!C standard library exception}%
 \begin{footnote}
-That is, the C
-library functions can all be treated as if they
+That is, the C library functions can all be treated as if they
```
zygoloid, line 3717:
```suggestion
That is, the C standard library functions can all be treated as if they
```
Use exactly the term we define in Clause 3, not a variant of it, per ISO rules.

tkoeppe: Good catch!

source/basic.tex
tkoeppe: Sounds good, thanks!

source/intro.tex
```diff
@@ -12,9 +12,9 @@
 
 \pnum
 \Cpp{} is a general purpose programming language based on the C
-programming language as described in ISO/IEC 9899:2018
-\doccite{Programming languages --- C} (hereinafter referred to as the
-\defnx{C standard}{C!standard}). \Cpp{} provides many facilities
+programming language as described in \IsoC{}
+\doccite{Programming languages --- C}.
```
jensmaurer, line 16: We're not allowed to add a title next to the number. This needs to go.

source/lib-intro.tex
```diff
@@ -150,9 +150,9 @@
 if it raises a floating-point exception other than \tcode{FE_INEXACT}.
 The semantics of a call to a C standard library function
 evaluated as a core constant expression
-are those specified in Annex F of the C standard
+are those specified in \IsoC{}, Annex F
 \begin{footnote}
-See also ISO/IEC 9899:2018 section 7.6.
+See also \IsoC{}, 7.6.
```
jensmaurer, line 155: Should we add "subclause" before "7.6" ?

source/lib-intro.tex
```diff
@@ -1599,8 +1599,7 @@
 \begin{footnote}
 The only reliable way to declare an object or
 function signature from the C standard library is by including the header that
-declares it, notwithstanding the latitude granted in 7.1.4 of the C
-Standard.
+declares it, notwithstanding the latitude granted in \IsoC{}, 7.1.4.
```
jensmaurer, line 1602: "subclause"?

There are citation examples in the House Style, I think.

source/support.tex
```diff
@@ -6164,6 +6164,6 @@
 The header \libheader{stdlib.h}
 assuredly provides the same declarations and definitions within
 the global namespace,
-much as in the C Standard. It may also provide these names within
+much as in \IsoC{}. It may also provide these names within
```
jensmaurer, line 6167: Make that undated? We don't refer to a specific clause/annex.

source/lib-intro.tex
tkoeppe: Not according to the House Style examples of citing specific elements, I think.

source/support.tex
tkoeppe: Hm, we also don't have an undated entry in the normative references, which is perhaps a problem: If we have both dated and undated ones, we might end up depending  on multiple versions of C. Maybe we can avoid the undated references entirely?

source/lib-intro.tex
tkoeppe: """
Use the order of “document number:date, element” when referring to a specific element:

“…in accordance with ISO 12345:2018, 4.3.9.”
“…in accordance with ISO/IEC TR 6789:2020, Table 3.”
“…in accordance with ISO 12345:2018/Amd 1:2019, Figure C.1.”
"""

Only things like "Table", "Figure", or "Annex" (and presumably "Clause") get a name.

source/support.tex
jensmaurer: This is a note / example.  Everywhere else we use undated refs, we are also non-normative or explanatory, I think. So it seems we're good.

source/intro.tex
tkoeppe: Done.

source/support.tex
tkoeppe: OK, let's try. We do actually use an undated C reference normatively in one place (with the floating point types) :-S

jensmaurer (approved):

PR #6619, opened 2023-10-12
[class.copy.assign] Remove a superfluous note.

Quuxplusone:
Alternatively we could have added the word "non-object"; or changed it to say "An overloaded assignment operator must be a member function"; but it doesn't seem like it needs to be here at all.

(Clang, EDG, MSVC all agree that `B& operator=(this B, int);` is legal C++23.)

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -1679,7 +1679,7 @@ More than one form of copy assignment operator can be declared for a class.
 \begin{note}
 If a class
 \tcode{X}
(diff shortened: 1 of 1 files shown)


jensmaurer (approved):
This looks fine; the intent seems obvious and we already say "non-object parameter" when we define "copy assignment operator", so explicit object parameters were in scope (and intended to be accepted) when the words were last changed.

PR #6620, opened 2023-10-13
[class.default.ctor, class.copy.ctor, class.copy.assign] No code font for "is constexpr"

Quuxplusone:
Just throwing this over the wall, in hopes that if any of these are editorial improvements they'll be taken up individually.
Right now the text is pretty inconsistent about "X is constexpr"/"X is `constexpr`", but I think the former is better. I'm also pretty sure that a function that isn't constexpr-suitable cannot possibly be constexpr, hence the biggest diff below.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -1350,8 +1350,7 @@ constructor for that class with no
 \grammarterm{compound-statement}.
 If that user-written default constructor would be ill-formed,
 the program is ill-formed.
-The implicitly-defined
-default constructor is \keyword{constexpr}.
+The implicitly-defined default constructor is constexpr.
 Before the defaulted default constructor for a class is
 implicitly defined,
 all the non-user-provided default constructors for its base classes and
@@ -1636,8 +1635,7 @@ otherwise the copy/move constructor is
 The copy/move constructor is implicitly defined even if the implementation elided
 its odr-use\iref{term.odr.use,class.temporary}.
 \end{note}
-The implicitly-defined\iref{dcl.fct.def.default}
-constructor is \keyword{constexpr}.
+The implicitly-defined\iref{dcl.fct.def.default} constructor is constexpr.
 
 \pnum
 Before the defaulted copy/move constructor for a class is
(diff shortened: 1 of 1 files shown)


source/classes.tex
```diff
@@ -1636,8 +1594,8 @@
 The copy/move constructor is implicitly defined even if the implementation elided
 its odr-use\iref{term.odr.use,class.temporary}.
 \end{note}
-The implicitly-defined\iref{dcl.fct.def.default}
-constructor is \keyword{constexpr}.
+An implicitly-defined\iref{dcl.fct.def.default} constructor is
```
frederick-vs-ja: As I said before, changes to [class.copy.ctor] and [class.copy.assign] were probably incorrect. Would you mind to rebase the branch and only preserve changes to [class.default.ctor]?

source/classes.tex
```diff
@@ -1842,22 +1805,24 @@
 has the return type
 \tcode{X\&}.
 An implicitly-declared copy/move assignment operator is an
-inline public member of its class.
+inline public member of its class,
+and is implicitly constexpr if it is constexpr-suitable.
```
Quuxplusone: At this point _every_ non-coroutine function is [_constexpr-suitable_](https://eel.is/c++draft/dcl.constexpr#def:constexpr-suitable), and an implicitly declared copy/move assignment operator is never a coroutine, so indeed the phrase `if it is constexpr-suitable` can be eliminated.

The point of this change — to replace "is `constexpr`" with "is implicitly constexpr" — remains, AFAICT.

(Also, [Note 1 in [dcl.fct.def.default]](https://eel.is/c++draft/dcl.fct.def#default-note-1) seems misleading: it should probably say that other functions _explicitly_ defined as defaulted are not implicitly constexpr.)

source/classes.tex
```diff
@@ -1273,102 +1236,98 @@
...
+the implicitly-defined
+default constructor is constexpr.
 Before the defaulted default constructor for a class is
 implicitly defined,
 all the non-user-provided default constructors for its base classes and
 its non-static data members are implicitly defined.
 \begin{note}
 An implicitly-declared default constructor has an
-exception specification\iref{except.spec}.
+implicit exception specification\iref{except.spec}.
```
Quuxplusone: I think this is still necessary, but it looks like the term of art is "_implied_ exception specification" (e.g. [[class.copy.ctor] Note 6](https://eel.is/c++draft/class.mem#class.copy.ctor-13.sentence-2)), not "_implicit_ exception specification." So I've removed this change from this PR. I might open a different PR to clean up this phrase, later, but have no plans to do it immediately.

frederick-vs-ja:
@tkoeppe This should have been sufficiently rebased and only contains formatting fixes now.

jensmaurer (approved):
Yup, that's formatting only now.

Issue #3515, opened 2019-11-26
don't use "is called" to introduce a definition


zygoloid:
We have decided to use "call" to mean a function call expression (see #3079). So we should stop using it to introduce definitions.

We should use a different phrasing to avoid this. For example, instead of:

> For every iterator type `X`, there is a corresponding signed integer-like type called the *difference type* of the iterator.

Consider:

> For every iterator type `X`, there is a corresponding signed integer-like type known as the *difference type* of the iterator.

Or:

> The *difference type* of an iterator type `X` is a corresponding signed integer-like type, determined as specified in [incrementable.traits].

jwakely:
Does this mean it's wrong to refer to "the caller" of a function? e.g. [new.handler] (2.3) "terminate execution of the program without returning to the caller."

The decision seems to mean that "call" is a static property of the source code, whereas "the caller" is a dynamic property of a particular invocation at run time. Changing to "the invoker" doesn't seem like an improvement (everybody knows what "the caller" means).

I'm also trying to determine what to change in the Networking TS, which has gems like:

> where `g` is a function object of unspecified type that, when called as `g()`, performs `fd()`.

(I'm considering "unspecified type such that the invocation of `g()` invokes `fd()`" but I don't really like that either.)

jensmaurer:
I think "caller of a function" is sufficiently unambiguous; "invoker" sounds stupid.

"calling thread" probably can bear with some improvement; maybe "the thread on which the function is invoked"? This would also fix the category error that a "thread" calls something (no, another function calls something).

"when called as g()" seems fine (this is [expr.call]); maybe simply replace "performs" with "evaluates".

PR #6623, opened 2023-10-14
[intro.defs, dcl.init.list] Move definition of direct-non-list-init

tkoeppe:
Partially fixes ISO/CS-11 (C++23 DIS).

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -5755,7 +5755,7 @@ List-initialization can occur in direct-initialization or copy-initialization co
 list-initialization in a direct-initialization context is called
 \defn{direct-list-initialization} and list-initialization in a
 copy-initialization context is called \defn{copy-list-initialization}.
-Direct-initialization that is not direct-list-initialization is called
+Direct-initialization that is not list-initialization is called
 \defn{direct-non-list-initialization}.
 \begin{note}
 List-initialization can be used
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -5755,6 +5755,8 @@ List-initialization can occur in direct-initialization or copy-initialization co
 list-initialization in a direct-initialization context is called
 \defn{direct-list-initialization} and list-initialization in a
 copy-initialization context is called \defn{copy-list-initialization}.
+Direct-initialization that is not direct-list-initialization is called
+\defn{direct-non-list-initialization}.
 \begin{note}
 List-initialization can be used
 \begin{itemize}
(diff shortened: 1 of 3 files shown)


tkoeppe:
The other terms "direct-list-initialization" etc. are already defined there, and the connection between "foo-initialization" and "x is foo-initialized" or "foo-initializes x" are never spelled out anyway, nor do we pervasively make cross-references to the other definitions of "foo-initialization". This seems to make it no worse to also have "direct-non-list-initialization" defined in that place.

source/declarations.tex
```diff
@@ -5755,6 +5755,8 @@
 list-initialization in a direct-initialization context is called
 \defn{direct-list-initialization} and list-initialization in a
 copy-initialization context is called \defn{copy-list-initialization}.
+Direct-initialization that is not direct-list-initialization is called
```
jensmaurer, line 5758: This seems more straightforward:

```suggestion
Direct-initialization that is not list-initialization is called
```

W-E-Brown:
> On 2023-10-14, at 2:08 PM, Jens Maurer ***@***.***> wrote:
> 
<snip>
> ⬇️ Suggested change
> -Direct-initialization that is not direct-list-initialization is called
> +Direct-initialization that is not list-initialization is called
Please adjust:  s/is called/is termed/
(or equivalent, e.g., "is known as").
Thank you.

source/declarations.tex
tkoeppe: I was thinking about that, too, and thought that direct-list-iniitialization is more immediately in memory, but it can go either way. Sure.

tkoeppe:
> On 2023-10-14, at 2:08 PM, Jens Maurer ***@***.***> wrote:
> <snip>
> ⬇️ Suggested change -Direct-initialization that is not direct-list-initialization is called +Direct-initialization that is not list-initialization is called
> Please adjust: s/is called/is termed/ (or equivalent, e.g., "is known as"). Thank you.

We say "is called" everywhere else before, too. I think if we don't like that, this should be a separate (editorial) issue. Care to file one?

JohelEGP:
> Please adjust: s/is called/is termed/

Related: #3515.

jensmaurer (approved):

W-E-Brown:
> On 2023-10-14, at 2:33 PM, Thomas Köppe ***@***.***> wrote:
> We say "is called" everywhere else before, too. I think if we don't like that, this should be a separate (editorial) issue. Care to file one?
> 
Why?  It seems we've had such an issue since 2019:  #3515 <https://github.com/cplusplus/draft/issues/3515>.  I doubt we really need another issue over the same policy concern.

PR #6625, opened 2023-10-15
[expr.prim.lambda.general] Added missing definition for the term "lambda".

Bekenn:
The unadorned term "lambda" is used interchangeably with _lambda-expression_ in a few places, but is never actually defined.  The term "generic lambda" is defined in terms of "lambda".  Adding the missing definition seemed like the simplest approach to resolving the issue.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -1843,8 +1843,9 @@ auto x3 = [&]()->auto&& { return j; };  // OK, return type is \tcode{int\&}
 \end{example}
 
 \pnum
-A lambda is a \defn{generic lambda}
-if the \grammarterm{lambda-expression}
+A \grammarterm{lambda-expression} is a \defn{lambda}.
+A lambda is \defnx{generic}{lambda!generic}
+if the \grammarterm{parameter-declaration-clause}
 has any generic parameter type placeholders\iref{dcl.spec.auto}, or
 if the lambda has a \grammarterm{template-parameter-list}.
 \begin{example}
@@ -2059,7 +2060,7 @@ auto f = []<typename T1, C1 T2> requires C2<sizeof(T1) + sizeof(T2)>
 \end{note}
 
 \pnum
-The closure type for a non-generic \grammarterm{lambda-expression} with no
+The closure type for a non-generic lambda with no
 \grammarterm{lambda-capture}
 whose constraints (if any) are satisfied
 has a conversion function to pointer to


source/expressions.tex
```diff
@@ -1843,8 +1843,9 @@
 \end{example}
 
 \pnum
+A \grammarterm{lambda-expression} is a \defn{lambda}.
 A lambda is a \defn{generic lambda}
```
JohelEGP, line 1847:
```suggestion
A lambda is \defnx{generic}{lambda!generic}
```

JohelEGP (changes requested):

JohelEGP (changes requested):
As for the commit message, see <https://github.com/cplusplus/draft/wiki/How-to-submit-an-editorial-issue#regardless>.

Bekenn:
Do you want an amended commit, or should I just change the PR title?

JohelEGP:
It'd be easier on the editors if the commit message was right from the start.
But they can also fix it when merging the PR.

Bekenn:
Updated.

jensmaurer:
Hm...

I'm not too fond of splitting up "generic lambda" into "lambda is generic"; I think we use "generic lambda" (in this order) whenever we us that term.

What's the count on plain "lambda" that we would need to turn into "lambda-expression"?

I understand that using "lambda-expression" on one hand and "generic lambda" on the other is a bit non-symmetric in phrasing, but I think I could live with it.

Bekenn:
Here are the cases I could find:

* [basic.def.odr]/10 uses the term "non-lambda function"
* [dcl.spec.auto.general]/2: "or the lambda is a generic lambda" (this is in a note, so maybe not important)
* [dcl.attr.depend]/1: "The attribute may be applied to a parameter of a function or lambda"
* [dcl.attr.depend]/1: "lambda call operator" (also [dcl.attr.nodiscard] and [dcl.attr.noreturn])
* [expr.prim.lambda.general]/6: "A lambda is a *generic lambda* if the *lambda-expression* has any generic parameter type placeholders, or if the lambda has a *template-parameter-list*."
* [expr.prim.lambda.closure]/4: "(for a non-generic lambda)"
* [expr.prim.lambda.closure]/5: "Given a lambda with a *lambda-capture*, the type of the explicit object parameter, if any, of the lambda’s function call operator"
* [expr.const]/17: "the call operator of a lambda"
* [defns.prog.def.type]: "non-implementation-provided lambda expression"
* [futures.async]/8: "the lambda might not be invoked" (this appears in a note)

Some of these can readily be replaced with *lambda-expression*, but others feel a little bit awkward (e.g. "non-generic lambda).  Some of these uses appear to refer to the closure type rather than the *lambda-expression* itself.

tkoeppe:
CWG discussion: please reword all the instances as discussed.

PR #6630, opened 2023-10-21
[streambuf.general] Strike "abstract" from "basic_streambuf serves as an abstract base class"

### LINKED ISSUE #6629: [streambuf.general] `std::basic_streambuf` is not an abstract class
[p1](http://eel.is/c++draft/streambuf.general#1) says "The class template `basic_streambuf` serves as an abstract base class for deriving various stream buffers". Base class I agree with, but an abstract class [has a pure virtual function](http://eel.is/c++draft/class.abstract#def:class,abstract), and `basic_streambuf` has not

PJBoy:
Fixes #6629.

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -3096,7 +3096,7 @@ namespace std {
 \pnum
 The class template
 \tcode{basic_streambuf}
-serves as an abstract base class for deriving various
+serves as a base class for deriving various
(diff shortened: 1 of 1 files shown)


jensmaurer (approved):

jensmaurer:
(shorten commit message on merge)

PR #6631, opened 2023-10-22
[std] Use more \defnadj

jensmaurer:
Partially addresses #6111


jensmaurer:
@tkoeppe, these are mild indexing improvements courtesy of @Eelis.

PR #6637, opened 2023-11-01
[expr.prim.lambda.closure] Improve some parallel grammar

Quuxplusone:
I noticed that we say "F is... and..." in two parallel places, but one says "and is" and the other says "and F is" when really they should both just say "and". The further parallelism of "the corresponding specialization" is just a drive-by but I think also an improvement.


Quuxplusone:
@tkoeppe ping!

PR #6645, opened 2023-11-08
[string.capacity] Remove parentheses from "reserve()"

jwakely:
It's very confusing to talk about `reserve()` when describing a call to `reserve(size_type)`, given that the overload `reserve()` also exists.

### DIFF
--- a/source/strings.tex
+++ b/source/strings.tex
@@ -3009,8 +3009,8 @@ constexpr void reserve(size_type res_arg);
 \effects
 A directive that informs a \tcode{basic_string} of a planned change in size,
 so that the storage allocation can be managed accordingly.
-After
-\tcode{reserve()},
+Following a call to
+\tcode{reserve},
 \tcode{capacity()}
 is greater or equal to the argument of
 \tcode{reserve}
@@ -3019,7 +3019,7 @@ equal to the previous value of
 \tcode{capacity()}
 otherwise.
 Reallocation happens at this point if and only if
-the current capacity is less than the argument of \tcode{reserve()}.
+the current capacity is less than the argument of \tcode{reserve}.
 
 \pnum
 \throws


source/strings.tex
```diff
@@ -3010,7 +3010,7 @@
 A directive that informs a \tcode{basic_string} of a planned change in size,
 so that the storage allocation can be managed accordingly.
 After
-\tcode{reserve()},
+\tcode{reserve},
```
tkoeppe, line 3013: Could we use slightly more regular wording here, something like "after a call of `reserve`" (or "an invocation of")?

jwakely: Updated to say "After a call to `reserve`". I hope it's clear that means _immediately_ after, not for the rest of the program!

tkoeppe: Then why not just say "immediately after"? Hope is not a strategy!

jwakely: Would "Following a call to `reserve`" be better?

tkoeppe: Works for me!

tkoeppe: "Following an invocation of `reserve`", does that sound more contemporary?

tkoeppe: @jwakely Did you want to do any of those ^^^ ?

jwakely: I think "Following a call to `reserve`" is my preference, updated.

source/strings.tex
```diff
@@ -3009,8 +3009,8 @@
 \effects
 A directive that informs a \tcode{basic_string} of a planned change in size,
```
tkoeppe, line 3010: (Not sure what's going on here with directives and informing objects, but that's presumably some vintage wording that we'll best leave as is.)

jwakely: :see_no_evil:

jensmaurer (approved):

Issue #2215, opened 2018-06-27
What "temporary variable" is?


languagelawyer:
[[conv]/3](http://eel.is/c++draft/conv#3), [[conv]/4](http://eel.is/c++draft/conv#4), [[conv]/6](http://eel.is/c++draft/conv#6) and [[expr.static.cast]/4](http://eel.is/c++draft/expr.static.cast#4) are using this ~term~ combination of words. What does it mean?

jwakely:
Does [class.temporary] not help?

languagelawyer:
No, it does not.

PR #6646, opened 2023-11-08
[conv.general, expr.static.cast] Remove inappropriate "temporary"

tkoeppe:
It's not clear in which sense the invented variables in question are "temporary". Since C++ already contains a notion of a "temporary (object)", and since variables are precisely _not_ temporary objects, it seems unnecessary to use the words "temporary variable" in the present sense.

See also http://thecodelesscode.com/case/234.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -550,14 +550,14 @@ See~\ref{dcl.init}, \ref{dcl.init.ref}.
 \pnum
 An expression $E$ can be
 \defnx{implicitly converted}{conversion!implicit} to a type \tcode{T} if and only if the
-declaration \tcode{T t=$E$;} is well-formed, for some invented temporary
+declaration \tcode{T t=$E$;} is well-formed, for some invented
 variable \tcode{t}\iref{dcl.init}.
 
 \pnum
 Certain language constructs require that an expression be converted to a Boolean
 value. An expression $E$ appearing in such a context is said to be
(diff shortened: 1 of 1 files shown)


tkoeppe:
@opensdh: could you please PTAL a look?

source/expressions.tex
```diff
@@ -550,14 +550,14 @@
 \pnum
 An expression $E$ can be
 \defnx{implicitly converted}{conversion!implicit} to a type \tcode{T} if and only if the
-declaration \tcode{T t=$E$;} is well-formed, for some invented temporary
+declaration \tcode{T t=$E$;} is well-formed, for some invented
```
opensdh, line 553:
```suggestion
declaration \tcode{T t=$E$;} is well-formed for some invented
```
while we're at it; similarly below.

opensdh (approved):
You should probably check with more than one person here, though.

source/expressions.tex
tkoeppe: Done, .

tkoeppe:
@jensmaurer: Could I interest you/CWG in this?

jensmaurer:
CWG 2023-11-10: We want [class.temporary] p6 to refer to "invented temporary variable" as one of the roots of "temporary expression" and say this is a "temporary object". Other cases (where we don't mean "temporary") should use "invented variable". This needs a core issue.

languagelawyer:
OP should say it addresses #2215

PR #6653, opened 2023-11-10
[algorithms] Change stable tag 'mismatch' to alg.mismatch, so that it is consistent with similar labels in the same section

ojrosten:


JohelEGP (changes requested):
This needs an entry into `xrefdelta.tex`.

ojrosten:
Thanks! And just to be clear it would look something like this?

% https://github.com/cplusplus/draft/pull/6653
\movedxref{algorithms.alg.nonmodifying.mismatch}{algorithms.alg.nonmodifying.alg.mismatch}

JohelEGP:
It's this example: `\movedxref{old.label}{new.label}`.
The old label is `mismatch`, and the new label `alg.mismatch`.

JohelEGP:
By the way, see <https://github.com/cplusplus/draft/wiki/Commit-message-format>.

ojrosten:
Noted; I'm very much a rookie ;)

ojrosten:
Hopefully the title of the PR is now fixed...

JohelEGP:
It should cover the commit message, not just the issue's title.
Also, the label should probably be [algorithms],
since you're changing both the [algorithm.syn] and [mismatch] within.

ojrosten:
The label in the title?

ojrosten:
How important is it to change the commit messages? I can of course do so, but I gather that github strongly recommends against force pushing and I'm not aware of any other way.

JohelEGP:
PRs are generally a working space that are OK to force-push to.

ojrosten:
I'll do this then :)

I'll prepend the first commit message with [algorithms]. But what about the second, i.e. the change to xrefdelta.tex?

jensmaurer:
> I'll prepend the first commit message with [algorithms]. But what about the second, i.e. the change to xrefdelta.tex?

xrefdelta is just technical fall-out; you should ignore this for the commit message.

JohelEGP:
It's all part of the bigger change, so you can rebase to fixup.
That is, `git rebase -i main`, and change the second commit from `pick` to `fixup`.

ojrosten:
Hopefully that's now done. Thank you so much for your help! (And of course I'm happy to make further changes if something isn't right.)

jensmaurer (approved):
While we generally frown upon changing stable labels, this seems to fix a rather disturbing inconsistency that outweighs other concerns. (Also, we expect few references to that section to exist.)

PR #6681, opened 2023-11-13
[LWG motion 4] P2546R5 Debugging Support

### LINKED ISSUE #6662: [2023-11 LWG Motion 4] Debugging Support

grafikrobot:
Fixes #6662.
Fixes cplusplus/papers#1207

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -619,6 +619,7 @@ the values of these macros with greater values.
   // \libheader{deque}, \libheader{queue}, \libheader{stack}, \libheader{string}
 #define @\defnlibxname{cpp_lib_copyable_function}@                 202306L // also in \libheader{functional}
 #define @\defnlibxname{cpp_lib_coroutine}@                         201902L // also in \libheader{coroutine}
+#define @\defnlibxname{cpp_lib_debugging}@                         202311L // freestanding, also in \libheader{debugging}
 #define @\defnlibxname{cpp_lib_destroying_delete}@                 201806L // freestanding, also in \libheader{new}
 #define @\defnlibxname{cpp_lib_enable_shared_from_this}@           201603L // also in \libheader{memory}
 #define @\defnlibxname{cpp_lib_endian}@                            201907L // freestanding, also in \libheader{bit}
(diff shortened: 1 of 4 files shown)


source/support.tex
```diff
@@ -619,6 +619,7 @@
   // \libheader{deque}, \libheader{queue}, \libheader{stack}, \libheader{string}
 #define @\defnlibxname{cpp_lib_copyable_function}@                 202306L // also in \libheader{functional}
 #define @\defnlibxname{cpp_lib_coroutine}@                         201902L // also in \libheader{coroutine}
+#define @\defnlibxname{cpp_lib_debugging}@                         202606L // freestanding, also in \libheader{debugging}
```
JohelEGP, line 622: Is this, maybe, part of the "emails with suggested editorial changes"?
If so, I'd like to know the reason, as this is surprising.
```suggestion
#define @\defnlibxname{cpp_lib_debugging}@                         202311L // freestanding, also in \libheader{debugging}
```

JohelEGP (changes requested):

source/support.tex
grafikrobot: > Is this, maybe, part of the "emails with suggested editorial changes"?
> If so, I'd like to know the reason, as this is surprising.

What aspect is surprising?

JohelEGP: Always, the value of the macro is the date of adoption.
I haven't read the paper, so I don't know if "we're merging this now, but adopting in 2026-06".

grafikrobot: > Always, the value of the macro is the date of adoption.

Ah, sorry I'm new to this, I didn't know that. Where is that documented?

JohelEGP: Typically, the paper's instruction says to set the value to the date of adoption.
@tkoeppe This might be worth adding to the editor's notes.

jensmaurer: Yes, absolutely: feature-test macros get the year/month of the plenary that voted them into the working draft, not some hypothetical future date when a future standard may or may not be approved (at what stage?)

Please fix as suggested.

JohelEGP (approved):

source/support.tex
tkoeppe: @JohelEGP Papers don't get to choose the values of those macros, so I don't think we'll call this out.

PR #6685, opened 2023-11-14
[Motions 2023 11 lwg 1] P3040R0 C++ Standard Library Issues

### LINKED ISSUE #6659: [2023-11 LWG Motion 1] C++ Standard Library Issues to be moved in Kona, Nov. 2023

burblebee:
Fixes #6659.

Notes:
* LWG3965 skipped; I don't know how to apply this change.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -21473,7 +21473,7 @@ is_same_v<make_index_sequence<rank()>, index_sequence<P...>>
 is \tcode{true}.
 Equivalent to:
 \begin{codeblock}
-return operator[](as_const(indices[P])...);
+return operator[](extents_type::@\exposid{index-cast}@(as_const(indices[P]))...);
 \end{codeblock}
 \end{itemdescr}
 
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -887,7 +887,7 @@ a <=> b
 
 \pnum
 \expects
-Either \tcode{<=>} is defined for values of type (possibly const) \tcode{T},
+Either \tcode{T} models \libconcept{three_way_comparable},
 or \tcode{<} is defined for values of type (possibly const) \tcode{T} and
 \tcode{<} is a total ordering relationship.
 
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -919,7 +919,9 @@ and given a container type \tcode{X} having a \tcode{value_type} identical to \t
 and an \tcode{allocator_type} identical to \tcode{allocator_traits<A>::rebind_alloc<T>}
 and given an lvalue \tcode{m} of type \tcode{A},
 a pointer \tcode{p} of type \tcode{T*},
-an expression \tcode{v} of type \tcode{T} or \tcode{const T},
+an expression \tcode{v} that denotes
+an lvalue of type \tcode{T} or \tcode{const T} or
+an rvalue of type \tcode{const T},
 and an rvalue \tcode{rv} of type \tcode{T},
 the following terms are defined. If \tcode{X}
 is not allocator-aware or is a specialization of \tcode{basic_string},
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -18355,6 +18355,12 @@ a trivially copyable type\iref{term.trivially.copyable.type}.
 \tcode{ElementType} is required to be
 a complete object type that is not an abstract class type.
 
+\pnum
+For a \tcode{span} \tcode{s},
+any operation that invalidates a pointer in
+the range \range{s.data()}{s.data() + s.size()}
+invalidates pointers, iterators, and references to elements of \tcode{s}.
+
 \rSec4[span.cons]{Constructors, copy, and assignment}
 
 \indexlibraryctor{span}%
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -19026,6 +19026,9 @@ namespace std {
   template<class LayoutMapping>
     struct submdspan_mapping_result;
 
+  struct full_extent_t { explicit full_extent_t() = default; };
+  inline constexpr full_extent_t full_extent{};
+
   template<class IndexType, class... Extents, class... SliceSpecifiers>
     constexpr auto submdspan_extents(const extents<IndexType, Extents...>&, SliceSpecifiers...);
 
--- a/source/strings.tex
+++ b/source/strings.tex
@@ -751,10 +751,10 @@ For a \tcode{basic_string_view str},
 any operation that invalidates a pointer
 in the range
 \begin{codeblock}
-@\range{str.data()}{\brk{}str.data() + str.size()}@
+@\range{str.data()}{str.data() + str.size()}@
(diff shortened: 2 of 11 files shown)


tkoeppe:
Thanks a lot! Is this PR ready then? It's still in "draft" status. Don't worry about the missing issue.

source/strings.tex
```diff
@@ -754,7 +754,7 @@
 @\range{str.data()}{\brk{}str.data() + str.size()}@
 \end{codeblock}
 invalidates pointers, iterators, and references
-returned from \tcode{str}'s member functions.
+to elements of \tcode{str}.
```
JohelEGP, line 757: This has not been applied:
> [Drafting note: The proposed wording also removes the extra code block that previously defined the range]

source/containers.tex
```diff
@@ -18355,6 +18357,12 @@
 \tcode{ElementType} is required to be
 a complete object type that is not an abstract class type.
 
+\pnum
+For a \tcode{span} \tcode{s},
+any operation that invalidates a pointer in
+the \range{s.data()}{s.data() + s.size()}
```
JohelEGP, line 18363:
```suggestion
the range \range{s.data()}{s.data() + s.size()}
```

source/utilities.tex
```diff
@@ -17756,6 +17770,9 @@
 \effects
 Parses the format specifier as a \fmtgrammarterm{range-format-spec} and
```
JohelEGP, line 17771:
```suggestion
Parses the format specifiers as a \fmtgrammarterm{range-format-spec} and
```

source/utilities.tex
```diff
@@ -8573,7 +8587,7 @@
 
 \pnum
 \constraints
-\tcode{is_constructible_v<T, decltype(**this)>} is \tcode{true}.
+\tcode{is_constructible_v<T, decltype(\exposid{val})>} is \tcode{true}.
```
JohelEGP, line 8590: There's extra parentheses in the PR.
Although that shouldn't make a difference.
Feel free to reject, but do tell Thomas,
as this might be worth raising in the editor's report.
```suggestion
\tcode{is_constructible_v<T, decltype((\exposid{val}))>} is \tcode{true}.
```

JohelEGP (changes requested):

source/strings.tex
burblebee: @JohelEGP Which codeblock is that?

JohelEGP: The one immediately before this line.

source/utilities.tex
burblebee: Leaving this one to @tkoeppe :). FYI - this is LWG3973.

source/strings.tex
burblebee: oh, I see. Thanks!

source/strings.tex
```diff
@@ -750,11 +750,9 @@
 For a \tcode{basic_string_view str},
 any operation that invalidates a pointer
 in the range
-\begin{codeblock}
-@\range{str.data()}{\brk{}str.data() + str.size()}@
-\end{codeblock}
+\range{str.data()}{str.data(\linebreak{}) + str.size()}
```
JohelEGP, line 753: Well, this will probably format ugly.
I think you should revert.
The code block was probably added in the first place for something like this.

burblebee: Exactly. I liked it better with the codeblock as well. But as you pointed out, the paper does ask that we inline this, so I relented. Guess @tkoeppe will have to decide what to do here...  (Be careful what you ask for? :) )

burblebee:
@tkoeppe, your choice to either apply or skip that last commit.

source/utilities.tex
jwakely: The extra parentheses are correct and removing them will break the spec, please don't.

`decltype(val)` here is the declared type of the member, i.e. `T`. If we wanted that we would just write `T`.

`decltype((val))` is affected by the cv- and ref-qualifiers of the member function, so it's either `T&` or `const T&` depending which overload it appears in. This is what we want.

source/strings.tex
jwakely: Removing the code block was suggested by @JohelEGP's original issue submission for [LWG3203](https://wg21.link/lwg3203) :stuck_out_tongue_winking_eye: 

Breaking between the parens does look horrible though:

image

Keeping the code block seems fine to me. There's no mention of that in the reflector discussion or the Varna minutes, so I don't think LWG requested that. @Dani-Hub, did you add the drafting note to LWG 3203 just because that's what the original proposed resolution did?

JohelEGP: The original PR [SUPERSERED] only affected [span.overview].

jwakely: Ah yes, so removing it from string view would be for consistency with that. But let's not if it looks bad.

source/utilities.tex
burblebee: fixed, thanks!

JohelEGP: Right, I wasn't thinking about that.
I guess I've adopted the mind that the Paper is Final.

source/utilities.tex
```diff
@@ -8429,7 +8443,7 @@
...
```
JohelEGP, line 8446:
```suggestion
\tcode{is_constructible_v<T, decltype((\exposid{val}))>} is \tcode{true}.
```

source/utilities.tex
```diff
@@ -8487,7 +8501,7 @@
 \begin{itemdescr}
 \pnum
 Let \tcode{U} be
-\tcode{remove_cv_t<invoke_result_t<F, decltype(**this)>>}.
+\tcode{remove_cv_t<invoke_result_t<F, decltype(\exposid{val})>>}.
```
JohelEGP, line 8504:
```suggestion
\tcode{remove_cv_t<invoke_result_t<F, decltype((\exposid{val}))>>}.
```

source/utilities.tex
```diff
@@ -8364,7 +8378,7 @@
 
 \begin{itemdescr}
 \pnum
-Let \tcode{U} be \tcode{remove_cvref_t<invoke_result_t<F, decltype(**this)>>}.
+Let \tcode{U} be \tcode{remove_cvref_t<invoke_result_t<F, decltype(\exposid{val})>>}.
```
JohelEGP, line 8381:
```suggestion
Let \tcode{U} be \tcode{remove_cvref_t<invoke_result_t<F, decltype((\exposid{val}))>>}.
```

JohelEGP (changes requested):

JohelEGP (approved):

source/strings.tex
tkoeppe: Let's retain the codeblock, so as to make this look decent. The paper isn't even quoting its alleged source N4950 correctly, so I don't have any faith in it demanding a particular formatting.

PR #6691, opened 2023-11-15
[LWG 9] P2447R6 std::span over an initializer list

### LINKED ISSUE #6667: [2023-11 LWG Motion 9] std::span over an initializer list
https://wiki.edg.com/pub/Wg21kona2023/StrawPolls/p2447r6.html

burblebee:
Fixes #6667.
Fixes cplusplus/papers#1118

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -107,6 +107,31 @@ auto s = std::to_string(1e-7);  // \tcode{"1e-07"}
                                 // changed according to the global C locale
 \end{codeblock}
 
+\rSec2[diff.cpp23.containers]{\ref{containers}: containers library}
+
+\diffref{span.overview}
+\change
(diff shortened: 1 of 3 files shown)


source/compatibility.tex
```diff
@@ -64,6 +64,31 @@
                                 // changed according to the global C locale
 \end{codeblock}
 
+\rSec2[diff.cpp23.containers]{\ref{containers}: containers library}
+
+\diffref{span.overview}
```
JohelEGP, line 69: Is it typical to only specify the class synopsis
instead of the parent subclause that also encompasses the detailed specifications?

JohelEGP (approved):

source/compatibility.tex
burblebee: They did the same thing for cpp20, so I guess so?

PR #6693, opened 2023-11-15
[LWG 14] P2819R2 Add tuple protocol to complex

### LINKED ISSUE #6672: [2023-11 LWG Motion 14] Add tuple protocol to complex
https://wiki.edg.com/pub/Wg21kona2023/StrawPolls/P2819R2.pdf

burblebee:
Fixes #6672.
Fixes cplusplus/papers#1500

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -289,6 +289,20 @@ namespace std {
   template<class T> constexpr complex<T> tan  (const complex<T>&);
   template<class T> constexpr complex<T> tanh (const complex<T>&);
 
+  // \ref{complex.tuple}, tuple interface
+  template<class T> struct tuple_size;
+  template<size_t I, class T> struct tuple_element;
+  template<class T> struct tuple_size<complex<T>>;
+  template<size_t I, class T> struct tuple_element<I, complex<T>>;
+  template<size_t I, class T>
+    constexpr T& get(complex<T>&) noexcept;
+  template<size_t I, class T>
+    constexpr T&& get(complex<T>&&) noexcept;
+  template<size_t I, class T>
+    constexpr const T& get(const complex<T>&) noexcept;
+  template<size_t I, class T>
+    constexpr const T&& get(const complex<T>&&) noexcept;
+
   // \ref{complex.literals}, complex literals
   inline namespace literals {
   inline namespace complex_literals {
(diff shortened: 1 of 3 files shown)


source/numerics.tex
```diff
@@ -1121,6 +1135,49 @@
 The complex hyperbolic tangent of \tcode{x}.
 \end{itemdescr}
 
+\rSec2[complex.tuple]{Tuple interface}
+
+\indexlibrarymember{tuple_size}{complex}%
+\indexlibrarymember{tuple_element}{complex}%
```
JohelEGP, line 1141: Feel free to reject.
```output
$ grep '\\index.*tuple_' *
grep: assets: Is a directory
containers.tex:\indexlibraryglobal{tuple_size}%
containers.tex:\indexlibraryglobal{tuple_element}%
utilities.tex:\indexlibraryglobal{tuple_size}%
utilities.tex:\indexlibraryglobal{tuple_element}%
utilities.tex:\indexlibraryglobal{tuple_cat}
utilities.tex:\indexlibrary{\idxcode{tuple_size}!in general}%
utilities.tex:\indexlibraryglobal{tuple_size}%
utilities.tex:\indexlibraryglobal{tuple_element}%
utilities.tex:\indexlibraryglobal{tuple_size}%
utilities.tex:\indexlibraryglobal{tuple_element}%
```
```suggestion
\indexlibraryglobal{tuple_size}%
\indexlibraryglobal{tuple_element}%
```

JohelEGP (changes requested):

source/numerics.tex
burblebee: I agree with you, but I'm following the convention of similar functions (e.g. tanh), so will leave as is. I've seen this weirdness elsewhere in the spec as well and have not understood the when and why. We should open an issue.

JohelEGP (approved):

source/numerics.tex
jwakely: The reason seems obvious, it means the index looks like:

```
tanh, 1491
    complex, 1429
    valarray, 1484
```

Instead of simply:

`tanh`, 1491, 1429, 1484

jensmaurer: Let's go with \indexlibraryglobal for tuple_* for consistency for now,
and let's change this globally to \indexlibrarymember in a separate change.

burblebee: Ok, done. While I agree that the additional context is helpful, calling these "members" is confusing.

jwakely: It's just the name of a macro in the source, "member" is not observable in the standard.

PR #6711, opened 2023-11-22
[class.eq] Fix the return value of a defaulted == operator function

poyaoc97:
The last sentence of [class.eq]p3 says the opposite after the refactor done in #6035.


jwakely (approved):

tkoeppe:
@brevzin could you have a look, too?

tkoeppe:
Maybe we should instead swap "true" and "false" to retain the symmetry with the next paragraph? Either seems fine.

jensmaurer (approved):
Uh, that's a clear bug introduced by the supposedly-editoral rewrite of #6035.

PR #6720, opened 2023-12-05
[std] Update references from ISO 8601:2004 to ISO 8601-1:2019

jensmaurer:
As requested by ISO/CS for C++23 DIS.


tkoeppe:
@Dani-Hub Could you please take a look? Jens looked up the new subclause number in the table of contents.

jensmaurer:
>  where the latter additionally explains how to combine the time of day in the presence of a time shift between local time and UTC.

Nowhere do we discuss the combination in C++.  %z just has the offset, which is defined in 5.3.4.1.

tkoeppe:
Great, thank you, both!

PR #6722, opened 2023-12-06
[syntax] Change "italic" to "italic, sans-serif"

tkoeppe:
We changed the grammar non-terminal font to sans-serif, so we should update the description.


jensmaurer (approved):
Sure.

PR #6726, opened 2023-12-08
[intro.defs] Remove mention of symbols from ISO 80000-2

jensmaurer:
According to ISO/CS, symbols cannot be imported by reference.


tkoeppe:
Sorry, I already did this :-S

PR #6735, opened 2023-12-14
[algorithms] reorders the fold family

cjdb:
I accidentally mixed up `fold_left_with_iter` with other fold algorithms multiple times while implementing it for libc++, and it might also be confusing reviewrs. This commit reorders them so that they appear in an order that they might be implemented.

* `fold_left_with_iter`
* `fold_left`
* `fold_left_first_with_iter`
* `fold_left_first`
* `fold_right`
* `fold_right_last`

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -1254,12 +1254,34 @@ namespace std {
       concept @\defexposconcept{indirectly-binary-right-foldable}@ =    // \expos
         @\exposconcept{indirectly-binary-left-foldable}@<@\exposid{flipped}@<F>, T, I>;
 
+    template<class I, class T>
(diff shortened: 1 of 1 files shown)


jensmaurer (commented):
Sorry, I'm not seeing that "while implementing it" is a convincing argument for any presentation order.

In general, the "simpler" algorithms are presented first, I think, and those with more extensive/complicated parameters or return types later.  That argues for moving the "with_iter" stuff to immediately before fold_right, but not any further up.

tkoeppe:
Editorial meeting decision Feb 2025: We don't quite think this is worth changing, and there are good enough reasons for the status quo.

PR #6737, opened 2023-12-17
[std] Rephrase notes containing 'must'

jensmaurer:
Fixes ISO/CS 037 (C++23 DIS).

Fixes cplusplus/nbballot#579

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -2708,7 +2708,7 @@ in \Cpp{} and
 \tcode{sizeof(char*)}
 in C.
 \difficulty
-Programs are required to add explicit casts to the appropriate rvalue.
+Programs need to add explicit casts to the appropriate rvalue.
 \howwide
 Rare.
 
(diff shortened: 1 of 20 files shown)


jensmaurer:
@tkoeppe , this is an initial sample of what might be done. Please have a look.

tkoeppe:
This is great, thank you! By and large this even looks like an improvement, don't you think?

jwakely (approved):

jensmaurer:
Major update; now all mentions of "must" (except in a literal string in an example) are gone.

tkoeppe:
Great! But please note that we can't say "required" as per drafting directives, since that's just a synonym for "shall" :-(

jensmaurer:
Yes, but "required" is mostly in Annex C, not directly in notes, so maybe the don't notice.

source/compatibility.tex
```diff
@@ -2708,7 +2708,7 @@
 \tcode{sizeof(char*)}
 in C.
 \difficulty
-Programs must add explicit casts to the appropriate rvalue.
+Programs are required to add explicit casts to the appropriate rvalue.
```
tkoeppe, line 2711: This is one of the "synonyms" for "shall" that we can't use, either (see https://www.iso.org/sites/directives/current/part2/index.xhtml#_idTextAnchor091).

source/compatibility.tex
```diff
@@ -2803,7 +2803,7 @@
 
 \diffref{dcl.typedef}
 \change
-A \Cpp{} \grammarterm{typedef-name} must be different from any class type name declared
+A \Cpp{} \grammarterm{typedef-name} is required to be different from any class type name declared
```
tkoeppe, line 2806: Similarly here and below.

tkoeppe:
I'll rebase and apply this PR after the IS has shipped. I'm pushing additonal changes I made to the IS back into this PR, too, so we don't lose them.

PR #6738, opened 2023-12-17
[version.syn] New feature test macro __cpp_lib_freestanding_numeric

tkoeppe:
This macro indicates that freestanding support for "saturation arithmetic" is available, which was added in motion LWG-2 (via P0543R3). This reverts the previous change
148e03a16d53ff8cffd219384df37efad5fd386d, which I had made subsequent to motion LWG-3 (P2407R5), on advice of LWG. A separate macro is preferable to mixing both headers under "algorithm".


jwakely (approved):

jwakely:
@brevzin @ben-craig

tkoeppe:
Let me merge this in the interest of expedience, but please do let me know if you want to do something different here, @ben-craig, @brevzin!

ben-craig:
I agree with the direction of this change.  The freestanding feature test
macros try to match up with the corresponding header, as that is an easy
partitioning for users to understand and implementers to implement.

PR #6745, opened 2023-12-25
[linalg] Add indexing

jensmaurer:


jensmaurer:
@jwakely , any opinion here?

jensmaurer:
@jwakely , ping.

source/numerics.tex
```diff
@@ -12407,6 +12420,7 @@
 
 \rSec3[linalg.conj.conjugated]{Function template \tcode{conjugated}}
 
+\indexlibraryglobal{conjugated}%
```
Eelis, line 12423: Is there a reason to use this, instead of using `@\libglobal{conjugated}@` in the actual declaration below?

jwakely (approved):

PR #6758, opened 2024-01-05
[std] Remove problematic phrases from notes

tkoeppe:
Only specific phrases involving the word "required" are problematic, namely when they appear to establish a normative requirement. They have been reworded, often by replacing "is required" with "needs", sometimes with slightly larger edits.

Also reword "necessary", "permitted", "allowed" in notes, so that they do not (inappropriately) state requirements or
permissions.

On request of ISO.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -6538,7 +6538,7 @@ For a thread of execution providing \defn{concurrent forward progress guarantees
 the implementation ensures that the thread will eventually make progress for as
 long as it has not terminated.
 \begin{note}
-This is required regardless of whether or not other threads of execution (if any)
+This is regardless of whether or not other threads of execution (if any)
 have been or are making progress. To eventually fulfill this requirement means that
 this will happen in an unspecified but finite amount of time.
 \end{note}
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -6202,15 +6202,15 @@ if
 The ``inter-thread happens before'' relation describes arbitrary
 concatenations of ``sequenced before'', ``synchronizes with'' and
 ``dependency-ordered before'' relationships, with two exceptions. The first
-exception is that a concatenation is not permitted to end with
+exception is that a concatenation never ends with
 ``dependency-ordered before'' followed by ``sequenced before''. The reason for
 this limitation is that a consume operation participating in a
 ``dependency-ordered before'' relationship provides ordering only with respect
 to operations to which this consume operation actually carries a dependency. The
 reason that this limitation applies only to the end of such a concatenation is
 that any subsequent release operation will provide the required ordering for a
-prior consume operation. The second exception is that a concatenation is not
-permitted to consist entirely of ``sequenced before''. The reasons for this
+prior consume operation. The second exception is that a concatenation never
+consist entirely of ``sequenced before''. The reasons for this
 limitation are (1) to permit ``inter-thread happens before'' to be transitively
 closed and (2) the ``happens before'' relation, defined below, provides for
 relationships consisting entirely of ``sequenced before''.
@@ -6290,7 +6290,7 @@ unspecified or undefined.
 \begin{note}
 This states that operations on
 ordinary objects are not visibly reordered. This is not actually detectable
-without data races, but it is necessary to ensure that data races, as defined
+without data races, but is needed to ensure that data races, as defined
 below, and with suitable restrictions on the use of atomics, correspond to data
 races in a simple interleaved (sequentially consistent) execution.
 \end{note}
@@ -6401,7 +6401,7 @@ interleaving. This is normally referred to as ``sequential consistency''.
 However, this applies only to data-race-free programs, and data-race-free
 programs cannot observe most program transformations that do not change
 single-threaded program semantics. In fact, most single-threaded program
-transformations continue to be allowed, since any program that behaves
+transformations remain possible, since any program that behaves
 differently as a result has undefined behavior.
 \end{note}
 
@@ -6538,7 +6538,7 @@ For a thread of execution providing \defn{concurrent forward progress guarantees
 the implementation ensures that the thread will eventually make progress for as
 long as it has not terminated.
 \begin{note}
-This is regardless of whether or not other threads of execution (if any)
+This applies regardless of whether or not other threads of execution (if any)
 have been or are making progress. To eventually fulfill this requirement means that
 this will happen in an unspecified but finite amount of time.
 \end{note}
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -1558,7 +1558,7 @@ This implies that, within a class template with a template
 \begin{codeblock}
 friend class T;
 \end{codeblock}
-is ill-formed. However, the similar declaration \tcode{friend T;} is allowed\iref{class.friend}.
+is ill-formed. However, the similar declaration \tcode{friend T;} is well-formed.\iref{class.friend}.
 \end{note}
 
 \pnum
@@ -4684,7 +4684,7 @@ but the declaration of a function taking no arguments and returning an
 \tcode{X}.
 The form
 \tcode{()}
-is permitted in certain other initialization contexts\iref{expr.new,
+can appear in certain other initialization contexts\iref{expr.new,
 expr.type.conv,class.base.init}.
 \end{note}
 
@@ -4752,7 +4752,7 @@ associated with the initialization of $e_i$
 is sequenced before those associated with the initialization of $e_j$.
 \begin{note}
 By contrast with direct-list-initialization,
-narrowing conversions\iref{dcl.init.list} are permitted,
+narrowing conversions\iref{dcl.init.list} can appear,
 designators are not permitted,
 a temporary object bound to a reference
 does not have its lifetime extended\iref{class.temporary}, and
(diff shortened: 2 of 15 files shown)


source/basic.tex
```diff
@@ -6202,15 +6202,15 @@
 The ``inter-thread happens before'' relation describes arbitrary
 concatenations of ``sequenced before'', ``synchronizes with'' and
 ``dependency-ordered before'' relationships, with two exceptions. The first
-exception is that a concatenation is not permitted to end with
+exception is that a concatenation does not end with
```
jwakely, line 6205: I feel like "never ends with" would express the original intent a bit better, without sounding like it's giving or removing permission to do something.

source/basic.tex
```diff
@@ -6202,15 +6202,15 @@
...
 ``dependency-ordered before'' followed by ``sequenced before''. The reason for
 this limitation is that a consume operation participating in a
 ``dependency-ordered before'' relationship provides ordering only with respect
 to operations to which this consume operation actually carries a dependency. The
 reason that this limitation applies only to the end of such a concatenation is
 that any subsequent release operation will provide the required ordering for a
-prior consume operation. The second exception is that a concatenation is not
-permitted to consist entirely of ``sequenced before''. The reasons for this
+prior consume operation. The second exception is that a concatenation does not
+consist entirely of ``sequenced before''. The reasons for this
```
jwakely, line 6213: Similarly, "never consists entirely of"

source/basic.tex
```diff
@@ -6290,7 +6290,7 @@
 \begin{note}
 This states that operations on
 ordinary objects are not visibly reordered. This is not actually detectable
-without data races, but it is necessary to ensure that data races, as defined
+without data races, but needed to ensure that data races, as defined
```
jwakely, line 6293:
```suggestion
without data races, but is needed to ensure that data races, as defined
```

I think this reads a little more naturally to me.

source/basic.tex
```diff
@@ -6401,7 +6401,7 @@
 However, this applies only to data-race-free programs, and data-race-free
 programs cannot observe most program transformations that do not change
 single-threaded program semantics. In fact, most single-threaded program
-transformations continue to be allowed, since any program that behaves
+remain transformations possible, since any program that behaves
```
jwakely, line 6404:
```suggestion
transformations remain possible, since any program that behaves
```
?

source/basic.tex
```diff
@@ -6538,7 +6538,7 @@
 the implementation ensures that the thread will eventually make progress for as
 long as it has not terminated.
 \begin{note}
-This is required regardless of whether or not other threads of execution (if any)
+This is regardless of whether or not other threads of execution (if any)
```
jwakely, line 6541:
```suggestion
This applies regardless of whether or not other threads of execution (if any)
```

source/expressions.tex
```diff
@@ -5378,7 +5378,7 @@
 and the deallocation function's name is
 \tcode{\keyword{operator} \keyword{delete}[]}.
 \begin{note}
-An implementation is required to provide default definitions for the global
+An implementation is expected to provide default definitions for the global
```
jwakely, line 5381:
```suggestion
An implementation provides default definitions for the global
```

Don't we require it, not just expect it to be true for most implementations?

source/expressions.tex
```diff
@@ -6080,7 +6080,7 @@
 inheritance relationship between the two classes.
 \begin{note}
 For example, if the classes were defined later in the translation unit,
-a multi-pass compiler would be permitted to interpret a cast between
+a multi-pass compiler could conformingly interpret a cast between
```
jwakely, line 6083:
```suggestion
a multi-pass compiler could validly interpret a cast between
```

The word "conformingly" seems clunky to me (it's not even in my dictionary).

source/iostreams.tex
```diff
@@ -15161,8 +15161,8 @@
 \item The \grammarterm{root-name} element, if present.
 \item The \grammarterm{root-directory} element, if present.
 \begin{note}
-The generic format is required to ensure lexicographical
-comparison works correctly.
+It is possible that the use of the generic format is needed
+to ensure correct lexicographical comparison.
```
jwakely, line 15165: Sigh. This would be so much easier to read as "The use of the generic format may be needed to ensure ..." if that wasn't a forbidden word.

source/lib-intro.tex
```diff
@@ -3605,8 +3605,8 @@
 provided by an implementation.
 \begin{note}
 In particular,
-an implementation is not allowed to provide
-an additional declaration of that function or function template
+an conforming implementation does not provide
```
jwakely, line 3608:
```suggestion
a conforming implementation does not provide
```

source/numerics.tex
```diff
@@ -9472,7 +9472,7 @@
 if \tcode{X} cannot be converted to \tcode{int}
 by integral promotion\iref{conv.prom}, the program is ill-formed.
 \begin{note}
-Arguments that can be promoted to \tcode{int} are permitted for compatibility with C.
+Arguments that can be promoted to \tcode{int} are supported for compatibility with C.
```
jwakely, line 9475: Alternatively (but I don't think it's necessarily better):

```suggestion
Arguments that can be promoted to \tcode{int} provide compatibility with C.
```

jwakely: Or "allowing arguments that can be promoted to `int` provide compatibility with C".

source/overloading.tex
```diff
@@ -2141,8 +2141,8 @@
 or when converting to the left operand of an assignment operation
 only standard conversion sequences are allowed.
 \begin{note}
-When converting to the explicit object parameter, if any,
-user-defined conversion sequences are allowed.
+A conversion to the explicit object parameter, if any,
+can include user-defined conversion sequences.
```
jwakely, line 2145:
```suggestion
When a conversion to the explicit object parameter occurs,
it can include user-defined conversion sequences.
```

?

jwakely (changes requested):
I've added some comments. Some are just ideas, some are suggested improvements.

source/basic.tex
tkoeppe: That's great, thanks!

source/basic.tex
tkoeppe: Done!

source/basic.tex
tkoeppe: Agreed, esp. given since the sentence goes on for quite a bit after this.

source/basic.tex
tkoeppe: Nice!

source/expressions.tex
tkoeppe: OK, sure!

source/lib-intro.tex
tkoeppe: Done.

source/numerics.tex
tkoeppe: Yes to the latter (it's not the arguments that provide compatibility, but this feature as a whole).

source/basic.tex
tkoeppe: Whoops, thanks!

source/iostreams.tex
tkoeppe: word :-(

source/overloading.tex
tkoeppe: Very nice!

jwakely (approved):

source/expressions.tex
tkoeppe: Well spotted. Can we make this a separate issue?

jwakely: This is now #6761

jensmaurer:
Uh, I'm a bit late to the game, it seems.  Anyway, the changes look good to me. (I dimly remember that we've excised "need" in a prior iteration, because that was considered a bad word at some point, but if that's a good word now --- fine.)

tkoeppe:
Thanks, Jens! I think "need" does not appear as an official equivalent phrase in https://www.iso.org/sites/directives/current/part2/index.xhtml#_idTextAnchor078.

For that matter, I also don't see "could" and "might" being generally forbidden; they're merely not allowed for _provisions_. I'm not sure if ISO is really consistent in what they're asking for...

I do still want to compose a general guidance document for the wording groups. I think what we can agree on so far is that a) provisions have to use the explicitly describled verbal forms, and b) notes must not contain requirements, recommendations, permissions, or instructions.

It's true that "might" should not be used for permission (instead of "may"), but that doesn't seem to ban "might" outright. I think the argument (back from the C++20 cycle) was similar to the argument against "cannot" we have now: that it is not immediately _clear_ that the word isn't accidentally trying to be a permission or requirement. And that's perhaps not a bad point.

Anyway, we're just left with all the "cannot"s now.

jensmaurer:
Do you ... want to do something about the "cannot"s? Some of them were recently introduced as workarounds for other forbidden words.

source/declarations.tex
```diff
@@ -1558,7 +1558,7 @@
 \begin{codeblock}
 friend class T;
 \end{codeblock}
-is ill-formed. However, the similar declaration \tcode{friend T;} is allowed\iref{class.friend}.
+is ill-formed. However, the similar declaration \tcode{friend T;} is well-formed.\iref{class.friend}.
```
Quuxplusone, line 1561: @tkoeppe You introduced an extra period `.` here — the one after the `\iref` should stay but the new one before the `\iref` should be deleted again.

PR #6770, opened 2024-01-12
[std] Reword "cannot" in notes to not sound like negative permission

tkoeppe:
Notes shall not contain permissions. Even though "cannot" is allowed to express negative possibility, ISO has requested that we avoid it when it might be ambiguous and could be mistaken as permission.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -407,8 +407,8 @@ Unless otherwise stated, implementations may make arbitrary copies of elements
 where \tcode{is_trivially_copy_constructible_v<T>}
 and \tcode{is_trivially_destructible_v<T>} are \tcode{true}.
 \begin{note}
-This implies that user-supplied function objects cannot rely on
-object identity of arguments for such input sequences.
+This implies that the object identity of arguments
+in user-supplied function objects is not reliable for such input sequences.
 If object identity of the arguments to these function objects
 is important, a wrapping iterator
 that returns a non-copied implementation object
@@ -541,7 +541,7 @@ The semantics of invocation with
 \tcode{execution::parallel_policy}, or
 \tcode{execution::parallel_unsequenced_policy}
 allow the implementation to fall back to sequential execution
-if the system cannot parallelize an algorithm invocation,
+if the system is unable to parallelize an algorithm invocation,
 e.g., due to lack of resources.
 \end{note}
 
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -700,7 +700,7 @@ incomplete type\iref{term.incomplete.type},
 an abstract class type\iref{class.abstract},
 or a (possibly multidimensional) array thereof.
 \begin{note}
-In particular, a class \tcode{C} cannot contain
+In particular, it is not possible for a class \tcode{C} to contain
 a non-static member of class \tcode{C},
 but it can contain a pointer or reference to an object of class \tcode{C}.
 \end{note}
@@ -1218,7 +1218,7 @@ The address of a constructor shall not be taken.
 \indextext{restriction!constructor}%
 \begin{note}
 A \tcode{return} statement in the body of a constructor
-cannot specify a return value\iref{stmt.return}.
+never specifies a return value\iref{stmt.return}.
 \end{note}
 
 \pnum
@@ -1425,7 +1425,7 @@ an initializer of type
 or
 \tcode{volatile}
 \tcode{X}
-cannot initialize an object of type
+is not able to initialize an object of type
 \cv{}~\tcode{X}.
 \begin{example}
 \begin{codeblock}
@@ -1434,7 +1434,7 @@ struct X {
   X(X&);            // copy constructor with a non-const parameter
 };
 const X cx;
-X x = cx;           // error: \tcode{X::X(X\&)} cannot copy \tcode{cx} into \tcode{x}
+X x = cx;           // error: \tcode{X::X(X\&)} unable to copy \tcode{cx} into \tcode{x}
 \end{codeblock}
 \end{example}
 \end{note}
@@ -1684,10 +1684,8 @@ If a class
 \tcode{X}
 only has a copy assignment operator with a non-object parameter of type
 \tcode{X\&},
-an expression of type const
-\tcode{X}
-cannot be assigned to an object of type
-\tcode{X}.
+it is not possible to assign an expression of type const \tcode{X}
+to an object of type \tcode{X}.
 \begin{example}
 \begin{codeblock}
 struct X {
@@ -1697,7 +1695,7 @@ struct X {
 const X cx;
 X x;
 void f() {
-  x = cx;           // error: \tcode{X::operator=(X\&)} cannot assign \tcode{cx} into \tcode{x}
+  x = cx;           // error: \tcode{X::operator=(X\&)} unable to assign \tcode{cx} into \tcode{x}
 }
 \end{codeblock}
 \end{example}
@@ -2026,7 +2024,7 @@ The address of a destructor shall not be taken.
 \indextext{restriction!destructor}%
 \begin{note}
 A \tcode{return} statement in the body of a destructor
-cannot specify a return value\iref{stmt.return}.
+never specifies a return value\iref{stmt.return}.
 \end{note}
 \indextext{\idxcode{const}!destructor and}%
 \indextext{\idxcode{volatile}!destructor and}%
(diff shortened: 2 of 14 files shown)


source/classes.tex
```diff
@@ -6265,8 +6261,8 @@
 of constructors and destructors for the object declared by the
 \grammarterm{exception-declaration}.
 \begin{note}
-There cannot be a move from the exception object because it is
-always an lvalue.
+Since the exception object because is always an lvalue,
```
JohelEGP, line 6264:
```suggestion
Since the exception object is always an lvalue,
```

JohelEGP (changes requested):

tkoeppe:
Second half is in. PTAL.

source/classes.tex
tkoeppe: Thanks!

JohelEGP (approved):

source/classes.tex
```diff
@@ -1434,7 +1434,7 @@
   X(X&);            // copy constructor with a non-const parameter
 };
 const X cx;
-X x = cx;           // error: \tcode{X::X(X\&)} cannot copy \tcode{cx} into \tcode{x}
+X x = cx;           // error: \tcode{X::X(X\&)} unable copy \tcode{cx} into \tcode{x}
```
jensmaurer, line 1437:
```suggestion
X x = cx;           // error: \tcode{X::X(X\&)} unable to copy \tcode{cx} into \tcode{x}
```

source/classes.tex
```diff
@@ -6265,8 +6261,8 @@
 of constructors and destructors for the object declared by the
 \grammarterm{exception-declaration}.
 \begin{note}
-There cannot be a move from the exception object because it is
-always an lvalue.
+Since the exception object is always an lvalue,
+it is never moved from, which would potentially be incorrect.
```
jensmaurer, line 6265: "would" is also bad in notes, I think.

jensmaurer, line 6265:
```suggestion
The exception object is always an lvalue, thus it is not implicitly moved from."
```

source/expressions.tex
```diff
@@ -98,14 +98,14 @@
 that result is then added to 5 which results in the value assigned to
 \tcode{a}. On a machine in which overflows produce an exception and in
 which the range of values representable by an \tcode{int} is
-\crange{-32768}{+32767}, the implementation cannot rewrite this
+\crange{-32768}{+32767}, a valid implementation avoids rewriting this
```
jensmaurer, line 101:
```suggestion
\crange{-32768}{+32767}, an implementation is precluded from rewriting this
```

source/expressions.tex
```diff
@@ -98,14 +98,14 @@
...
 expression as
 \begin{codeblock}
 a = ((a + b) + 32765);
 \end{codeblock}
 since if the values for \tcode{a} and \tcode{b} were, respectively,
 $-32754$ and $-15$, the sum \tcode{a + b} would produce an exception while
-the original expression would not; nor can the expression be rewritten
+the original expression would not; similarly, it avoids rewriting the expression
```
jensmaurer, line 108:
```suggestion
the original expression would not; similarly, it is precluded from rewriting the expression
```

source/expressions.tex
```diff
@@ -1821,7 +1820,7 @@
 any attributes are treated as
 \grammarterm{attribute-specifier-seq} in \grammarterm{lambda-expression}.
 \begin{note}
-Such ambiguous cases cannot have valid semantics
+Such ambiguous cases would not have valid semantics
 because the constraint expression would not have type \keyword{bool}.
```
jensmaurer, line 1824:
```suggestion
Such ambiguous cases do not have valid semantics,
because the constraint expression needs to have type \keyword{bool}.
```

source/locales.tex
```diff
@@ -500,7 +500,7 @@
 \begin{note}
 Because facets are used by iostreams,
 potentially while static constructors are running,
-their initialization cannot depend on programmed static initialization.
+their initialization would be unreliable if it depended on programmed static initialization.
```
jensmaurer, line 503:
```suggestion
it is not possible for their initialization to depend on programmed static initialization.
```

source/statements.tex
```diff
@@ -1044,7 +1044,7 @@
 being initialized, the concurrent execution shall wait for completion
 of the initialization.
 \begin{note}
-A conforming implementation cannot introduce
+A conforming implementation avoids introducing
```
jensmaurer, line 1047:
```suggestion
A conforming implementation never introduces
```

source/utilities.tex
```diff
@@ -687,7 +687,7 @@
 \expects
 \tcode{false} is \tcode{true}.
 \begin{note}
-This precondition cannot be satisfied, thus the behavior
+Since this precondition cannot be satisfied, the behavior
```
jensmaurer, line 690: Richard was avoiding "Since", because it might be interpreted temporal "since two weeks ago".  Use "because" or "thus".

jensmaurer (changes requested):
I think the "cannot -> never" changes are readable and sometimes an improvement; "cannot -> it is not possible" changes are ugly.  But so be it.

source/classes.tex
tkoeppe: I think "would" is fine, actually. What isn't allowed is the use of "would" to express a provision, but we're not doing that here.

tkoeppe: This is nice!

source/expressions.tex
tkoeppe: Let's leave this as is, since I'm not sure if "needs" is actually safe.

source/utilities.tex
tkoeppe: That's a good point, I'll add a fixup.

tkoeppe: That said, we can avoid "cannot": "The precondition is never satisfied, thus..."

source/algorithms.tex
```diff
@@ -407,8 +407,8 @@
 where \tcode{is_trivially_copy_constructible_v<T>}
 and \tcode{is_trivially_destructible_v<T>} are \tcode{true}.
 \begin{note}
-This implies that user-supplied function objects cannot rely on
-object identity of arguments for such input sequences.
+This implies that there is no reliable object identity of arguments
+in user-supplied function objects for such input sequences.
```
jwakely, line 411:
```suggestion
This implies that the object identity of arguments
in user-supplied function objects is not reliable for such input sequences.
```

source/locales.tex
```diff
@@ -500,7 +500,7 @@
 \begin{note}
 Because facets are used by iostreams,
 potentially while static constructors are running,
-their initialization cannot depend on programmed static initialization.
+it is not possible for their initialization to depend on programmed static initialization.
```
jwakely, line 503: What is programmed static initialization?

source/support.tex
```diff
@@ -2203,9 +2203,9 @@
 \begin{note}
 A function registered via \tcode{at_quick_exit}
 is invoked by the thread that calls \tcode{quick_exit},
-which can be a different thread
-than the one that registered it, so registered functions cannot rely on the identity
-of objects with thread storage duration.
+which can be a different thread than the one that registered it,
```
jwakely, line 2206: "can"?

source/algorithms.tex
tkoeppe: Nice, done!

source/support.tex
tkoeppe: How about "is invoked in the thread that calls `at_quick_exit`, a possibly different one than the one that registered it"?

jwakely: calls quick_exit not at_quick_exit, and maybe "a possibly different thread than the one" to avoid "one than the one".

source/templates.tex
```diff
@@ -8638,7 +8637,7 @@
 \pnum
 \begin{note}
 Except for reference and pointer types, a major array bound is not part of a
-function parameter type and cannot be deduced from an argument:
+function parameter type and is not possible to be deduced from an argument:
```
jwakely, line 8640:
```suggestion
function parameter type and it is not possible for it to be deduced from an argument:
```

This seems a bit nicer to me

source/support.tex
jwakely: I don't think we need a change here, I'm happy to resolve this comment

source/templates.tex
jwakely: Not essential to change

source/locales.tex
jwakely: I'll open an LWG issue for this as I have no idea what it's saying.

jwakely (approved):

PR #6771, opened 2024-01-12
[structure.specifications] clarify description of Results element

CaseyCarter:
The text "... the expression is an lvalue if the type is an lvalue reference type, an xvalue if the type is an rvalue reference type, and a prvalue otherwise" clearly indicates that a Result element describes the type and value category of an expression, yet we summarize it as only "a description of the type of the expression".


tkoeppe:
Yes, looks valuable, thanks!

PR #6772, opened 2024-01-15
[exception] Paragraph two is no longer universally true

CaseyCarter:
We recently added `bad_expected_access<void>` to the Standard Library, which derives from `exception`, but does not have "the following publicly accessible member functions, each of them having a non-throwing exception specification." For clarity, we should point out that this provision is not universal.


Dani-Hub:
Very much appreciated, @CaseyCarter !

jwakely (approved):

PR #6773, opened 2024-01-16
[alg.ends.with]: drop_view should be views::drop

Quuxplusone:
We don't want programmers doing CTAD on `ranges::drop_view(args...)`, so we shouldn't put it in the as-if-by code either. What we expect people to do (and what we intend vendors to do here) is `views::drop`.

(Compare #6373. I'm also pinging the LWG reflector for an opinion on whether this is handleable without an LWG issue.)

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -4682,7 +4682,7 @@ Let \tcode{N1} be \tcode{ranges::distance(r1)} and
 \returns
 \tcode{false} if $\tcode{N1} < \tcode{N2}$, otherwise
 \begin{codeblock}
-ranges::equal(ranges::drop_view(ranges::ref_view(r1), N1 - N2), r2, pred, proj1, proj2)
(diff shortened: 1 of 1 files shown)


Quuxplusone:
Updated (thanks Tim). The `ranges::ref_view(r1)` wrapper is necessary for `owning_view`: https://godbolt.org/z/8secz55Tn
Here I'd like to use `views::ref(r1)` but that doesn't exist; or `views::all(r1)` but that's ill-formed for `owning_view`s; I don't think there's a user-facing way to spell "I just want to look at this view, not copy or modify it."

timsong-cpp:
> We _are_ trashing the view (by iterating it)

Move-only views are not necessarily input-only.

Quuxplusone:
@tkoeppe @jwakely ping!

jwakely (approved):

jensmaurer (approved):

Issue #6775, opened 2024-01-16
[thread.sema] `std::binary_semaphore` is not indexed


Eisenwave:
See https://eel.is/c++draft/libraryindex

PR #6781, opened 2024-01-25
[semaphore.syn] add binary_semaphore to index

Eisenwave:
Fixes https://github.com/cplusplus/draft/issues/6775

### DIFF
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -9262,7 +9262,7 @@ namespace std {
   template<ptrdiff_t least_max_value = @\impdef@>
     class counting_semaphore;
 
-  using binary_semaphore = counting_semaphore<1>;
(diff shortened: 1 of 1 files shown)


source/threads.tex
```diff
@@ -9256,6 +9256,7 @@
 \rSec2[semaphore.syn]{Header \tcode{<semaphore>} synopsis}
 
 \indexheader{semaphore}%
+\indexlibraryglobal{binary_semaphore}%
```
jensmaurer, line 9259: Please move to the place where it is defined (a few lines below) and use  @\libglobal{blah}@

That way, we point to the right spot within the synopsis, not at its beginning.

jensmaurer (changes requested):

jensmaurer (approved):

PR #6785, opened 2024-01-27
[tuple.cnstr] Do not use code font for cardinal number 1

jwakely:
Consistent with '2' on line 1977 above.

https://github.com/cplusplus/draft/blob/4fe9190fa05c4fb4e83c1a1ba68aa12aa49542e9/source/utilities.tex#L1977


jensmaurer (approved):

jensmaurer:
Being locally consistent is certainly the first step here.

PR #6793, opened 2024-02-04
[iterator.requirements.general] Clarify that "constexpr iterator" is …

Dani-Hub:
…a requirement to be met

[iterator.requirements.general] p16 says "Iterators are called _constexpr iterators_ [..]", but all referencing sections say "meet the constexpr iterator requirements". For clarity, we should reword the definition to make it clear that this is a named requirement.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -709,12 +709,12 @@ is undefined.
 
 \pnum
 \indextext{iterator!constexpr}%
-Iterators are called \defn{constexpr iterators}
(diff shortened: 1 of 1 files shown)


Dani-Hub:
This solves issue #6793.

Dani-Hub:
@jwakely What do you think?

jwakely (approved):

jensmaurer (approved):

PR #6797, opened 2024-02-06
[dcl.type.elab] Don't start a paragraph with a note

Quuxplusone:
Currently looks like this: https://eel.is/c++draft/dcl.type.elab#4

I *think* the note "obviously" belongs to the previous paragraph.
Possible alternative solution: move Note 2 to the bottom of paragraph 5 instead.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -1533,17 +1533,20 @@ does not consider scopes that contain the target scope; no name is bound.
 \begin{note}
 A \grammarterm{using-directive} in the target scope is ignored
 if it refers to a namespace not contained by that scope.
-\ref{basic.lookup.elab} describes how name lookup proceeds
-in an \grammarterm{elaborated-type-specifier}.
 \end{note}
 
 \pnum
 \begin{note}
+\ref{basic.lookup.elab} describes how name lookup proceeds
+in an \grammarterm{elaborated-type-specifier}.
 An \grammarterm{elaborated-type-specifier} can be used to refer to
 a previously declared \grammarterm{class-name} or \grammarterm{enum-name}
 even if the name has been hidden by a non-type declaration.
 \end{note}
+
+\pnum
 If the \grammarterm{identifier} or \grammarterm{simple-template-id}
+in an \grammarterm{elaborated-type-specifier}
 resolves to a \grammarterm{class-name} or
 \grammarterm{enum-name}, the \grammarterm{elaborated-type-specifier}
 introduces it into the declaration the same way a
@@ -1563,7 +1566,7 @@ is ill-formed. However, the similar declaration \tcode{friend T;} is well-formed
 
 \pnum
 The \grammarterm{class-key} or \keyword{enum} keyword
-present in the
+present in an
 \grammarterm{elaborated-type-specifier} shall agree in kind with the
 declaration to which the name in the
 \grammarterm{elaborated-type-specifier} refers. This rule also applies to


jensmaurer:
I think the note should be in a numbered paragraph alone.  It doesn't fit with the preceding paragraph, which talks about friends, nor does it fit with the next paragraph, which talks about what an elaborated-type-specifier means.

Quuxplusone:
(Thinking out loud, then recording for posterity in case it's useful)
Ah, I take back my "'obviously' belongs to the previous paragraph," even in scare-quotes. I had had `friend` on the brain, and had been reading Note 2 (which appears sandwiched between p4 about friend declarations and p5 _still_ talking about that _same_ friend declaration) as fundamentally relevant to friend declarations, e.g.
```
class C {
  int S;
  friend struct S; // OK
};
```
but I see now that Note 2 could also be read as relevant to `struct S memberfunc();` etc.

But if it's _not_ talking about friend declarations, then Note 2 doesn't seem to be reinforcing anything _anywhere_ in this section, then! It's referring to a quirk of name lookup described in <a href="https://eel.is/c++draft/basic.lookup.elab#1.sentence-1">[basic.lookup.elab]</a> (as name-dropped in the preceding Note 1)... which means that another plausible fix would be to _combine_ Note 1 and Note 2 into a single Note. So that's what I've done now.

I've also removed the paragraph break between p4 and p5, since both paragraphs appear to be discussing the grammar introduced in p4. The following p6, on the other hand, is back to talking about _elaborated-type-specifiers_ in general rather than the p4 grammar in particular.

Finally, I'm unclear on the meaning of this sentence:

> If the _identifier_ or _simple-template-id_ resolves to a _class-name_ or _enum-name_, the _elaborated-type-specifier_ introduces it into the declaration the same way a _simple-type-specifier_ introduces its _type-name_ ([[dcl.type.simple]](https://eel.is/c++draft/dcl.spec#dcl.type.simple)).

The word "introduce" does not appear with any relevant meaning in [dcl.type.simple]. I don't know what it means. Therefore I don't know if this sentence is indeed talking _only_ about friend declarations (cf. the preceding and following sentences in the same paragraph) or if it's making some broader point about all _elaborated-type-specifiers_ in general, and therefore ought to be pulled down to below the friend-declaration stuff.

jensmaurer:
No.

p4 is the only paragraph here talking about friends.  Note 1 (first sentence) emphasizes a possibly non-obvious consequence of the normative utterance "does not consider scopes that contain the target scope", but that rule is limited to friends.

p5 is not exclusively about friends, but just says that a mention of "struct X" has the same semantics as a mention of plain "X" (that what "introduces into a declaration" wants to say).

Note 2 is a side effect of elaborated-type-specifier lookup pointed to as the second part of Note 1.

So, please leave p4 alone, except: I would be fine with joining the second sentence of note 1 with note 2, and putting the enlarged note 2 in a numbered paragraph of its own (this is the "name lookup" pointer, essentially).

Quuxplusone:
> No.
> 
> p4 is the only paragraph here talking about friends. Note 1 (first sentence) emphasizes a possibly non-obvious consequence of the normative utterance "does not consider scopes that contain the target scope", but that rule is limited to friends.
> 
> p5 is not exclusively about friends, but just says that a mention of "struct X" has the same semantics as a mention of plain "X" (that's what "introduces into a declaration" wants to say).

Hm. So you're saying that [p5's sentence](https://eel.is/c++draft/dcl.type.elab#5.sentence-2) "If the _identifier_ or _simple-template-id_ resolves to..." isn't talking about the most recently mentioned "_identifier_ or _simple-template-id_" (i.e. the one in the three-line grammar production in p4), but is actually meant to hark back to the four-line grammar production in p1? If we want it to read that way, then I suggest (and propose in the latest update) that it should be amended to "If the _identifier_ or _simple-template-id_ <ins>in an _elaborated-type-specifier_</ins> resolves to...".

I see now that [Note 3](https://eel.is/c++draft/dcl.type.elab#note-3) wants to make a general point, that e.g. `using T = int; class T g;` and `template<class T> struct X { class T m_; };` are ill-formed. Its usage specifically of `friend` (so soon after p4) is just asking for confusion IMO.

Finally, there's the question of why [p5 contains](https://eel.is/c++draft/dcl.type.elab#5.sentence-2) a cross-reference to [[dcl.type.simple](https://eel.is/c++draft/dcl.type.simple). You write:

> p5 is not exclusively about friends, but just says that a mention of "struct X" has the same semantics as a mention of plain "X" (that's what "introduces into a declaration" wants to say).

But [dcl.type.simple] doesn't seem to say anything about the semantics of "introducing." [This table](https://eel.is/c++draft/dcl.type.simple#tab:dcl.type.simple) does talk about "specifying a type" — but if p5 is trying to talk about that, it should say something like "If the _identifier_ or _simple-template-id_ <ins>in an _elaborated-type-specifier_</ins> resolves to a _class-name_, the _elaborated-type-specifier_ <del>introduces it into the declaration</del> <ins>specifies the named class type</ins> the same way <ins>as</ins> a _simple-type-specifier_ <del>introduces its</del> <ins>consisting of a</ins> _type-name_ [<ins>with the same spelling</ins>?] ([dcl.type.simple])." And then it's not clear to me what analogy the "same way..." subclause is trying to draw: Does "the same way" say something about the _mechanism by which_ or the _effect of_?

I'm pretty sure that this PR has turned into something above my pay grade either way. I'm updating the PR to more-or-less what you said (plus the extra disambiguators/scope-establishers that I think are needed for clarity) and not proposing any change re "introduces" nor re the example in Note 3. (Nor re changing Note 3 from a "Note" _to_ an "Example.")

jensmaurer (approved):
I think this is a small enough change so that we don't need CWG review for this.

PR #6804, opened 2024-02-12
[temp.arg.template] Clean up wording for template template argument matching

Bekenn:
I've rearranged the paragraphs describing matching in terms of partial ordering vs component-wise matching to more clearly separate the concepts.  I've also cleaned up some resulting redundant wording and moved the examples into separate paragraphs as they're fairly general within the context of this section (with the exception of the last example, which could reasonably be pulled into the paragraph on matching in terms of partial ordering).

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -1351,84 +1351,12 @@ C<A> c;             // \tcode{V<int>} within \tcode{C<A>} uses the primary templ
 \end{example}
 
 \pnum
-A \grammarterm{template-argument} matches a template
+A template \grammarterm{template-argument} \tcode{A} matches a template
 \grammarterm{template-parameter} \tcode{P} when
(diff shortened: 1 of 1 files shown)


tkoeppe:
@jensmaurer A Core perspective would be welcome here. I think we've asked James to send us this?

Bekenn:
I sent this unprompted; core has not seen this.

tkoeppe:
@jensmaurer: Could we perhaps get a slot for this during a CWG telecon (or whenever you review CWG-labelled editorial issues)? Thanks!

Bekenn:
I've reformulated this as a set of change instructions, available [here](https://github.com/Bekenn/isocpp/blob/main/pr6804.html).  The change instructions are fully equivalent to the content of this PR.

PR #6808, opened 2024-02-15
[temp.func.order] Specify to only add extra first argument if needed CWG2834

HoBoIs:
In [temp.func.order] paragraph 3, it is written that we need to add a new first parameter for all member functions for the purposes of template deduction. However, it only makes sense if we have an implicit object parameter.

If we don't have an implicit object parameter, we shouldn't add a new first parameter because all of our parameters explicitly are written.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -3904,7 +3904,7 @@ in the type of the value synthesized for a non-type template parameter
 is also a unique synthesized type.
 \end{note}
 Each function template $M$ that is a member function
-is considered to have
+with no explicit object parameter is considered to have
 a new first parameter of type $X(M)$, described below,
 inserted in its function parameter list.
 If exactly one of the function templates was considered by overload resolution


HoBoIs:
@jwiegley could you take a look, or recommend someone who can take a look?

jwiegley:
I haven't been involved with the C++ standard for 10 years now, I do not have any current recommendations.

jensmaurer:
Please squash the commits and force-push.

This looks like a CWG defect to me.

HoBoIs:
@jensmaurer
I have squashed and force-pushed them.

HoBoIs:
[This](https://www.open-std.org/jtc1/sc22/wg21/docs/cwg_active.html#2834) open cwg defect solves the same problem, by suggesting the same change (and some more for static member functions). So I think my change does not need to be a separate thing on its own.

jensmaurer:
CWG2834

PR #6814, opened 2024-02-21
[temp.pre] Add comma after introductory clause

Eisenwave:
Gramatically, there should be a comma here. Sentences such as

> In case X, Y happens.

... read very strangely when removing the comma:

> In case X Y happens.

[temp.pre] p5 suffers.


Eisenwave:
Note that there is also a CWG issue https://github.com/cplusplus/CWG/issues/506 in parallel to this change, but they affect different parts of the paragraph. Neither depends on the other.

jensmaurer (approved):

PR #6816, opened 2024-02-22
[format.parse.ctx] add comma before 'which'

Eisenwave:
Normally, you would write a comma here.

Saying
> `indexing == manual` is `true` which

suggests that `index == manual` is the `true` which ...

This wouldn't make sense since there is only one `true`, which makes it more correct to write a comma here. Of course, the sentence is unambiguous one way or the other; this is purely a readability improvement.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -17353,8 +17353,10 @@ return next_arg_id_++;
 
 \pnum
 \throws
-\tcode{format_error} if \tcode{indexing_ == manual} is \tcode{true}
-which indicates mixing of automatic and manual argument indexing.
+\tcode{format_error} if \tcode{indexing_ == manual} is \tcode{true}.
+\begin{note}
+This indicates mixing of automatic and manual argument indexing.
+\end{note}
 
 \pnum
 \remarks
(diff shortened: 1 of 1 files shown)


jwakely (changes requested):
I don't agree that without the comma it suggests "the true which". The definite article is not implicit there, and adding it changes the meaning completely. I do think this is a slight improvement, but the same change should be done on line 17383. Change both, or neither.

We could even move the explanation into a note:

> _[Note_: Calling this function when `indexing_ == manual` is `true` indicates mixing of ... _]_

So the normative part just says what+when it throws, and the note explains why the condition is an error. Just adding commas is fine though.

Eisenwave:
image

I favor the solution which converts the non-normative part into a note.

Also, I believe that it's sufficient to say "This indicates ...". Firstly, the stakes are low since this isn't normative wording anyway. Secondly, no matter whether the reader thinks that "This" refers to
- calling the function with the condition being true, or
- an error being thrown,

the reading is correct. I don't think the verbosity of "Calling this function when `indexing == manual` is `true` indicates ..." is required.

jwakely (approved):

jensmaurer (approved):

PR #6818, opened 2024-02-22
[string.view.ops] convert Effects to Returns paragraphs

Eisenwave:


source/strings.tex
```diff
@@ -1189,7 +1189,7 @@
 
 \pnum
 \effects
-Equivalent to \tcode{traits::copy(s, data() + pos, rlen)}.
+Equivalent to: \tcode{traits::copy(s, data() + pos, rlen)}.
```
jwakely, line 1192: The current form is in line with https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#writing-effects-in-a-function-description

source/strings.tex
```diff
@@ -1262,8 +1262,8 @@
 
 \begin{itemdescr}
 \pnum
-\effects
-Equivalent to: \tcode{return substr(pos1, n1).compare(str);}
+\returns
+\tcode{substr(pos1, n1).compare(str)}.
```
jwakely, line 1266: This has normative effect, because it no longer inherits the preconditions and other specification from `substr`.

e.g. what happens if `pos1` is out of range? The current wording defines that, your change doesn't.

jwakely: See [structure.specifications] p4.

Eisenwave: > See [structure.specifications] p4.

Hmm, I think you're right, but it still surprises me. Isn't there a substantial amount of `\returns` uses which are broken then?

For example [[string.conversions] p7](https://eel.is/c++draft/string.conversions#7) simply defines *Returns* for `std::to_string`, but no *Preconditions* or *Throws* specification.

However, this ultimately forwards to [`vformat`](https://eel.is/c++draft/format.functions#lib:vformat) which does have *Throws*. Does this mean that `std::to_string` never throws, or does it mean that `std::to_string` inherits its *Throws* (although this isn't specified)?

I feel like there's a larger issue. I've always assumed that *Returns* does inherit *Preconditions*, *Throws*, etc.

jwakely: > > See [structure.specifications] p4.
> 
> Hmm, I think you're right, but it still surprises me. Isn't there a substantial amount of `\returns` uses which are broken then?

No.


> For example [[string.conversions] p7](https://eel.is/c++draft/string.conversions#7) simply defines _Returns_ for `std::to_string`, but no _Preconditions_ or _Throws_ specification.

Those functions have no preconditions, so that's fine.
They are allowed to throw anything, which might be somewhat underspecified but isn't broken.

> However, this ultimately forwards to [`vformat`](https://eel.is/c++draft/format.functions#lib:vformat) which does have _Throws_. Does this mean that `std::to_string` never throws, 

No! It has to be noexcept or say "_Throws_: nothing" for that to be true. Those functions can throw anything.

> or does it mean that `std::to_string` inherits its _Throws_ (although this isn't specified)?

No. It will never throw a `format_error` for example, and it can only throw `bad_alloc` in practice. But since we don't say what it can throw, it's technically allowed to throw anything (modulo the global restriction that the library only their things derived from `exception`).

> I feel like there's a larger issue.

I don't think so. The best majority of cases are written as intended, and correct as written. There might be exceptions of course.

If we want to restrict what they're allowed to throw we could add a _Throws_ element.

> I've always assumed that _Returns_ does inherit _Preconditions_, _Throws_, etc.

Definitely not, it only tells you the value that is returned. It says nothing about how that's actually achieved.

jensmaurer:
Closing this as going in the wrong direction.

PR #6819, opened 2024-02-22
[expr.unary.op] remove redundant value category wording

Eisenwave:
https://github.com/cplusplus/draft/blob/8c8e05d7ff6cda6329ca898e7a270547a85675d7/source/expressions.tex#L4598-L4606

The paragraph above already states that *all* unary operators yield a prvalue. We can remove this redundant wording from the `&` operator.


jensmaurer (approved):

jensmaurer:
The general rule is literally in the paragraph preceding the struck text.  @tkoeppe , this feels safe.

PR #6820, opened 2024-02-22
[bit.pow.two] Remove redundant Remarks specification

Eisenwave:
This paragraph is useless and should be removed.

[[structure.specifications] p3.3](https://eel.is/c++draft/structure.specifications#3.3) already states that precondition violations are undefined behavior. Evaluations with undefined behavior are not core constant expressions as per [[expr.const] p5.8](https://eel.is/c++draft/expr.const#5.8).

It could be argued that this paragraph is not only pointless but also harmful because it suggests that the precondition violations don't *generally* disqualify expressions from being constant expressions, only in this case, and this is surely not intended.

Even if we read this paragraph as
> If a precondition is violated, the behavior is undefined, but it's not allowed for the undefined behavior to be a core constant expression.

... then what's the point of that? It just seems confusing.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -18961,12 +18961,6 @@ $N$.
 \pnum
 \throws
 Nothing.
-
-\pnum
-\remarks
-A function call expression
-that violates the precondition in the \expects element
-is not a core constant expression\iref{expr.const}.
 \end{itemdescr}
 
 \indexlibraryglobal{bit_floor}%


jensmaurer:
[[expr.const] p5.8](https://eel.is/c++draft/expr.const#5.8) is restricted to core-language undefined behavior.

Here, the remarks in the library section establish a special case that this particular library undefined behavior is on par with core language undefined behavior, as far as const evaluation goes.

jwakely:
Right, not all violations of library preconditions are guaranteed to produce an error during constant evaluation, e.g. using `lower_bound` on an unsorted range. This requirement is guaranteeing that the violation will be detected and diagnosed.

In an ideal world, maybe all violations would be diagnosed. But removing this paragraph does not get us closer to that ideal, instead it removes a useful guarantee and so makes the standard worse.

Eisenwave:
@jensmaurer I have brought shame upon myself. Thanks for clearing up my misconception.

PR #6827, opened 2024-02-28
P0732R2 Add `__cpp_nontype_template_parameter_class`

Eisenwave:
See https://github.com/cplusplus/CWG/issues/508 for discussion.

[[P0732R2]](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0732r2.pdf) proposed to add a `__cpp_nontype_template_parameter_class` predefined macro, and this has been accepted into C++20.

According to @jensmaurer, it's an editorial oversight that it wasn't added to the working draft.


cpplearner:
https://wg21.link/sd6 indicates that this macro was removed when P1907R1 was adopted. Instead,  `__cpp_nontype_template_args` was bumped.

PR #6832, opened 2024-03-01
[headers] Strike 'C standard library headers' and turn paragraph into note

### LINKED ISSUE #6789: [headers] has incorrect name for subclause xref
[headers] p9 says:

> 17.14, C standard library headers, describes the effects of using the name.h (C header) form in a C++
program.

Where 17.14 is a cross-ref to [support.c.headers], but that subclause's title is "C headers", not "C standard library headers" as shown in [headers] p9.

Should we just strike "C standard library headers" from p9? Maybe change it to:

> Subclause 17.14 describes the effects of using the name.h (C header) form in a C++ program.

Eisenwave:
Fixes #6789.

Furthermore, this paragraph has no normative effect and can be turned into a note. It's merely a *"this other section specifies ..."* paragraph.

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -1313,7 +1313,7 @@ standard header \libheader{iso646.h} has no effect.
 \end{footnote}
 
 \pnum
-\ref{support.c.headers}, C standard library headers, describes the effects of using
+Subclause \ref{support.c.headers} describes the effects of using
 the \tcode{\placeholder{name}.h} (C header) form in a \Cpp{} program.
 \begin{footnote}
  The


jwakely:
NACK. We have lots of "semi-normative" introductory paragraphs like this, they shouldn't be notes.

Eisenwave:
> NACK. We have lots of "semi-normative" introductory paragraphs like this, they shouldn't be notes.

Fine, I won't die on this hill.

jensmaurer:
@jwakely  , in #6789 you suggested what this pull request does.

Have you changed your mind, or is this pull request good as-is (ignoring the suggestion about introducing more notes)?

jwakely (approved):
I like the change to the cross-ref, I just didn't like changing it to a note.

jensmaurer (approved):

PR #6835, opened 2024-03-01
[expr.mul] Reword 'truncation towards zero' in footnote

### LINKED ISSUE #6790: "truncation towards zero" could be "truncation"
https://github.com/cplusplus/draft/blob/70b99af0dfcb9cae82bcd97c7b24a2e84edfb2cd/source/expressions.tex#L6233

> For integral operands, the \tcode{/} operator yields the algebraic quotient with
> any fractional part discarded;
> \begin{footnote}
> ***This is often called truncation towards zero.***
> \end{footnote}

One can find various definitions of truncation:
* *In simplest terms, truncation means to chop off the decimal portion of a number*
* *truncation is limiting the number of digits right of the decimal point*
* *To truncate a number, we miss off digits past a certain point in the number, filling-in zeros if necessary*

It is common in all of them that they all place zero digits somewhere in a number, and that generally brings the resulting number closer to zero (unless it is left in place, when the number equals to its truncated form). That is not a choice, that is what is happening.
Also, the *often* part might not be true today, if it ever has been. If we accept that the number of hits in internet search engines do represent the frequency of usage of a term to some extent, a couple hundred thousands hits for ["truncation towards zero"](https://www.google.com/search?q=%22truncation+towards+zero%22) does not feel often when compared to 100+ million hits discussing [number truncation](https://www.google.com/search?q=number+truncation), presumably also bringing their result closer to zero.

eisenwave:
Fixes #6790.

I agree with @Tevemadar; *"truncation towards zero"* sounds redundant. However, I have frequently heard
- rounding towards zero
- truncation

The new footnote mentions both.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -6228,14 +6228,14 @@ expression by the second.
 \indextext{zero!undefined division by}%
 If the second operand of \tcode{/} or \tcode{\%} is zero, the behavior is
 undefined.
-For integral operands, the \tcode{/} operator yields the algebraic quotient with
-any fractional part discarded;
-\begin{footnote}
-This is often called truncation towards zero.
-\end{footnote}
-if the quotient \tcode{a/b} is representable in the type of the result,
-\tcode{(a/b)*b + a\%b} is equal to \tcode{a}; otherwise, the behavior
-of both \tcode{a/b} and \tcode{a\%b} is undefined.
+For integral operands, the \tcode{/} operator yields the algebraic quotient
+rounded towards zero.
+\begin{note}
+If the quotient \tcode{a/b} is representable in the type of the result,
+\tcode{(a/b)*b + a\%b} is equal to \tcode{a};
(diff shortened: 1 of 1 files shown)


eisenwave:
I've made a few more changes, upon further thought:

1. Our cleaner, modern way of wording rounding (e.g. https://eel.is/c++draft/expr.shift#3) is to say "rounded towards X".
   Saying "truncation towards zero" is redundant, and we don't need clarifying footnotes if we just word it properly.
2. The following explanation of `a / b` = `(a/b)*b + a%b` is totally tutorial-like. That's simply a consequence of `%` being the remainder of the division with said rounding mode, as the paragraph states. It also adds no new information to say that the behavior is undefined if the result cannot be represented in the result type; that's a general rule for [expr]. I think it's best to just keep this wording in a note, even if it's giving the user a math tutorial and restating something broadly known. `\ref{expr}` may be helpful though, lemme add that.

PR #6836, opened 2024-03-01
[dcl.fct.default] Correct note on 'this' in default arguments

### LINKED ISSUE #6558: [dcl.fct.default] Incorrect note on `this` in default arguments
[[dcl.fct.default] p8](https://eel.is/c++draft/dcl.fct.default#8) states:
> [Note [5](https://eel.is/c++draft/dcl.fct.default#note-5): The keyword this cannot appear in a default argument of a member function; see [[expr.prim.this]](https://eel.is/c++draft/expr.prim.this).
> ```cpp
> class A {
>   void f(A* p = this) { }           // error
> };
> ```

This is incorrect, because the following program is well-formed and the keyword `this` appears in a default argument:
```cpp
class A {
  void f(auto = [] {
    struct B { void g() { this; } }; // keyword this appeared in a default argument
  }) { }
};
```
Syntactically, the entire *initializer-clause* is the *default-argument*, and the `this` keyword obviously does appear inside of it. See [[dcl.fct.default] p1](https://eel.is/c++draft/dcl.fct.default#1)

### Solutions

We could say "cannot directly appear" but this is wishy washy, though correct according to some reader's imaginations.

Wording it in terms of "current class" and whatnot is quite difficult, so maybe you could just says:
> Except within the block scope of a *lambda-expression*, the `this` keyword cannot ...

It's just a note, so we don't need to make it cover all the cases in all the detail; we just need to make it correct.

eisenwave:
Fixes #6558.

This implements the wording suggested by @jensmaurer.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -4229,9 +4229,8 @@ void f() {
 
 \pnum
 \begin{note}
-The keyword
-\keyword{this}
-cannot appear in a default argument of a member function;
+\keyword{this} cannot appear as a subexpression of
+a default argument of a member function;
 see~\ref{expr.prim.this}.
 \begin{example}
 \begin{codeblock}


jensmaurer:
Can we use `sizeof(this)` in a default argument?  If so, "subexpression" is not the right term.  Maybe we're looking at "potentially-evaluated" or "odr-usable". (We should be correct in notes, but not necessarily exhaustive.)

Eisenwave:
> Can we use `sizeof(this)` in a default argument? If so, "subexpression" is not the right term. Maybe we're looking at "potentially-evaluated" or "odr-usable". (We should be correct in notes, but not necessarily exhaustive.)

No, we cannot. [[expr.prim.this] p3](https://eel.is/c++draft/expr.prim.this#3) defines `this` only between the optional *cv-qualifier-seq* and the end of the declaration/definition. Default arguments are in the *parameter-declaration-clause* of [*parameters-and-qualifiers*](https://eel.is/c++draft/dcl.decl.general#nt:parameters-and-qualifiers), which precedes the *cv-qualifier-seq*.

PR #6839, opened 2024-03-01
[rand.req.seedeq], [rand.req.dist], [char.traits.require] Eliminate "compile-time" complexity

Eisenwave:
Related to https://github.com/cplusplus/draft/issues/5641.

In short, the C++ standard doesn't have a clearly-defined notion of "compile-time" or "run-time".

Instead of specifying that these type aliases have "compile-time complexity" (whatever that's supposed to mean) we can just leave the cell empty. No evaluation takes place here, so it's meaningless to talk about complexity.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -1711,7 +1711,7 @@ In \tref{rand.req.seedseq} and throughout this subclause:
   & \tcode{T}
   & \tcode{T} is an unsigned integer type\iref{basic.fundamental}
     of at least 32 bits.
-  & compile-time
+  &
   \\ \rowsep
 \tcode{S()}%
   &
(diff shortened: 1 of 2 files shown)


tkoeppe:
@jwakely, @jensmaurer: I'm sympathetic to improving this, but can we do better than just deleting "compile-time"? Can we maybe say something along the lines of something being constant-evaluated?

jensmaurer:
@tkoeppe, these are typedefs.  There is no evaluation going on at all, not even constant evaluation. We should just merge this.

jensmaurer:
Suggest to improve commit description to say "... for typedefs"

tkoeppe:
OK, makes sense. @Eisenwave Could you please update the commit messages? Also please just say "remove" instead of the overly lyrical "eliminate".

Eisenwave:
I'm on vacation. If it's not urgent, I can do it in two weeks when I'm back.

tkoeppe:
@Eisenwave Ping :palm_tree:

PR #6840, opened 2024-03-01
[std.iterator.tags] Reword iterator tag description

eisenwave:
Related to #5641.

This section has a few issues:

1. It contains wording "so that the function can select". This doesn't make much sense, since a function simply exists and never selects in itself. The intent is presumably that a *function call* selects the most efficient algorithm. That's how tag dispatch works anyway.

2. The wording contains "compile time", which isn't a concept that really exists in C++. It is better to talk about overload resolution.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -2741,11 +2741,12 @@ several classes and functions.
 It is often desirable for a
 function template specialization
 to find out what is the most specific category of its iterator
-argument, so that the function can select the most efficient algorithm at compile time.
+argument, so that a function call can select the most efficient algorithm
+during overload resolution.
 To facilitate this, the
 library introduces
 \defn{category tag}
-classes which are used as compile time tags for algorithm selection.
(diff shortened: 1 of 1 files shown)


jwakely:
I disagree with the first point. A function such as `std::distance` can indeed select the best algorithm, by making a function call using the tag. That is, `distance(a, b)` calls `distance_impl(a, b, cat(a))`. Or maybe it instead does:

```c++
if constexpr (derived_from<cat_t<decltype(a)>, random_access_iterator_tag>)
  return b - a;
```
This selects the optimal implementation without a function call.

Eisenwave:
Hmm, I guess you could view it that way as well. The way I see it, it still wouldn't be that the function selects the algorithm per-se, but the `if constexpr` block would end up calling the desired function directly by discarding all other options.

In practical terms, I'd also call `if constexpr` a form of "implementation selection" though. Perhaps the first part can be reverted and we can keep just 2.

PR #6841, opened 2024-03-01
[type.traits] Strike 'at compile time'

eisenwave:
Related to #5641.

C++ doesn't have a notion of "compile-time" or "run-time". We can simply say that type traits query the properties of a type, period.

If we really wanted to, we could say something about constant expression here. However, I see this as pointless; these paragraphs are informative anyway.

### DIFF
--- a/source/meta.tex
+++ b/source/meta.tex
@@ -76,7 +76,7 @@ The type \tcode{make_integer_sequence<T, N>} is an alias for the type
 Subclause \ref{type.traits} describes components used by \Cpp{} programs, particularly in
 templates, to support the widest possible range of types, optimize
 template code usage, detect type related user errors, and perform
-type inference and transformation at compile time. It includes type
+type inference and transformation. It includes type
 classification traits, type property inspection traits, and type
 transformations. The type classification traits describe a complete taxonomy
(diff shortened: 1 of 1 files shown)


jwakely:
If we wanted to retain this in some form, we could say "during translation". I don't think we lose anything by striking it completely though.

Eisenwave:
> I don't think we lose anything by striking it completely though.

Me neither. The paragraphs already mention that the subclauses contain templates, not utility functions or something that would suggest run-time over compile-time.

> If we wanted to retain this in some form, we could say "during translation".

If we added "during translation", that may beg the question "during translation of what?" Would that justify an `\iref{lex.phases}`? Honestly it's not worth the trouble to replace the wording with an equivalent.

jwakely:
During translation of the program. That's how we refer to compilation, so something that happens at compile time happens during translation. I don't think anything more would be needed.

AlisdairM:
I will add that I like the change, the emphasis on compile-time buys us nothing.  Note that I added a hyphen though, if you search for "compile-time" you will find multiple matches in library wording, including subclause titles, that I think would be harder to replace.

If we were to keep these uses I would recommend hyphenating like elsewhere, but would rather strike these 4 uses as extraneous.

PR #6845, opened 2024-03-02
[handler.functions] Add reference to [intro.races]

Eisenwave:
This is the first mention of "data race" or "synchronize with" in Clause [library], so adding an `\iref{intro.races}` seems helpful and appropriate here.

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -3340,8 +3340,8 @@ functions:
 \end{itemize}
 
 \pnum
-Calling the \tcode{set_*} and \tcode{get_*} functions shall not incur a data race. A call to
-any of the \tcode{set_*} functions shall synchronize with subsequent calls to the same
+Calling the \tcode{set_*} and \tcode{get_*} functions shall not incur a data race\iref{intro.races}.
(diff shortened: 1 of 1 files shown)


jwakely (approved):

source/lib-intro.tex
```diff
@@ -3340,8 +3340,8 @@
 \end{itemize}
 
 \pnum
-Calling the \tcode{set_*} and \tcode{get_*} functions shall not incur a data race. A call to
-any of the \tcode{set_*} functions shall synchronize with subsequent calls to the same
+Calling the \tcode{set_*} and \tcode{get_*} functions shall not incur a data race\iref{intro.races}.
+A call to any of the \tcode{set_*} functions shall synchronize with subsequent calls to the same
```
jensmaurer, line 3344: The additional xref is an improvement, but "shall synchronize" should just be "synchronizes"
and `get_*` isn't really the name of a function (shell globs are not a defined thing in the C++ standard).

jensmaurer (approved):

source/lib-intro.tex
Eisenwave: Fixed by https://github.com/cplusplus/draft/pull/6850.

PR #6847, opened 2024-03-03
[charconv.syn] Clarify what types match integer-type

Eisenwave:
The current wording in [[charconv.syn]](https://eel.is/c++draft/charconv.syn) is confusing at a surface level, but this is purely an editorial problem:

> ... the implementation provides overloads for all cv-unqualified signed and unsigned integer types and `char` in lieu of *integer-types*

Firstly, the reader may ask whether this means:
- (signed and unsigned integer types) and `char`, or
- (signed and unsigned) (integer types and `char`)

In other words, does this include `signed char` and `unsigned char`, or does it include `char`? However, `signed char` and `unsigned char` are already signed/unsigned integer types, so mentioning `char` separately here would be totally redundant. Only the former interpretation is plausible to me.

Secondly, the reader may ask whether "cv-unqualified" applies to `char` as well, but this question is totally pointless since `char` is already cv-unqualified.

Overall, we can reduce mental complexity here by simply putting `char` in front. Each of these questions has a clear answer (which is why this edit is not normative), but it would be better if those questions were eliminated completely.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -15215,7 +15215,7 @@ The header \libheader{execution} declares global objects associated with each ty
 When a function is specified
 with a type placeholder of \tcode{\placeholder{integer-type}},
 the implementation provides overloads
-for all cv-unqualified signed and unsigned integer types and \tcode{char}
+for \tcode{char} and all cv-unqualified signed and unsigned integer types
 in lieu of \tcode{\placeholder{integer-type}}.
 When a function is specified
 with a type placeholder of \tcode{\placeholder{floating-point-type}},


jwakely:
Another reason the second interpretation is wrong is just that we never say "signed `char`" let alone "signed and unsigned `char`". That would just be very confusing (does it mean `char` when that happens to be signed/unsigned on a given architecture?)

The types are `signed char` and `unsigned char`, not "signed and unsigned `char`".

So there's no ambiguity here. The change seems harmless though, and maybe clearer.

jensmaurer (approved):
Let's do it.

PR #6848, opened 2024-03-03
[charconv.from.chars] Clarify the role of a `0x` prefix in `from_chars`

eisenwave:
The current wording

> ... the prefix `"0x"` or `"0X"` is assumed.

... is ambiguous. It could either mean
- the prefix is assumed to be at the start of the string (inclusive prefix), or
- the prefix is assumed to precede the string (exclusive prefix).

In the English language, I don't believe that the inclusive/exclusive nature of prefixes is undisputed.

However, the following example with `0x123` disambiguates this, and this example has been present in [P0067R5](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0067r5.html) (which was accepted). 

Therefore, I believe intent is clear, and this issue is editorial. However, it would be better for the normative wording to be clear in itself instead of relying on a clarifying example.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -15535,7 +15535,7 @@ but not \tcode{chars_format::scientific},
 the optional exponent part shall not appear; and
 \item
 if \tcode{fmt} is \tcode{chars_format::hex},
-the prefix \tcode{"0x"} or \tcode{"0X"} is assumed.
+the prefix \tcode{"0x"} or \tcode{"0X"} shall not appear.
 \begin{example}
 The string \tcode{0x123}
 is parsed to have the value


jwakely:
I think it would be better to not use the word "assumed" at all. What does that assumption mean? Can the function subtract 2 from the pointer and assume that is a valid pointer and the characters "0x" will be present?

Can we phrase it in terms of the result being determined as if the prefix preceded the string, not assuming it actually does?

Eisenwave:
> Can we phrase it in terms of the result being determined as if the prefix preceded the string, not assuming it actually does?

Well, it's a tough nut to crack. I've thought about it for quite a while and I've settled on "shall not appear", which is in line with the wording for the other bullets.

We don't assume anything anymore, we just say what shall and shall not be part of the subject sequence. I think that's an overall improvement and it's clear.

However, I feel like the wording for all three bullets is a bit messy; it feels clunky to use "shall" here instead of describing with "softer" wording what we consider to be part of the subject sequence.

What do you think?

jwakely:
> Well, it's a tough nut to crack.

That's why I didn't suggest any specific wording ;-)

I don't think the new suggestion works at all. The current wording says the result is as though using `strtod` on a string that begins with "0x", so a hex float is parsed. If you remove that (and also say no 0x is allowed) then there's no way to parse a hex float. Nothing actually says that the `hex` format causes a hex float to be parsed. The assumed-to-be-present-but-not-actually-present 0x prefix causes that today.

jwakely:
We could describe it in terms of transforming the argument into a pattern in the form strtod expects, i.e. inserting the "0x" prefix and then parsing that. But then we'd need to be clear that the `ptr` member of the result gets to the original argument string not this imaginary one we talk about. I don't like this suggestion much.

eisenwave:
Superseded by:
- https://github.com/cplusplus/papers/issues/2734

PR #6849, opened 2024-03-03
[pairs.pair] Replace `std::forward` with `std::move` where equivalent

Eisenwave:
The use of `std::forward` for the following assignment operators is unnecessary:
- `operator=(pair&& p)`
- `operator=(pair&& p) const`.

In both cases, we know 100% that the types of `this->first` and `p.first` and the type of `this->second` and `p.second` match, and we are working with an rvalue reference, not a forwarding reference.

In such cases, it's not necessary to use `std::forward`, and it arguably misleads the reader into thinking that this is ever copying instead of moving, which isn't the case (unless the two are equivalent).

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -1076,8 +1076,8 @@ constexpr pair& operator=(pair&& p) noexcept(@\seebelow@);
 
 \pnum
 \effects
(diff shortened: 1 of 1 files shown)


Eisenwave:
Nevermind, I have missed the possibility of reference members.

PR #6850, opened 2024-03-03
[handler.functions], [mem.res.global] Replace 'shall synchronize with' with 'synchronizes with'

Eisenwave:
Fixes https://github.com/cplusplus/draft/pull/6845#discussion_r1510205634, implementing the change suggested by @jensmaurer.

While we're at it, we can apply the same change to [mem.res.global] and add an `\iref{intro.races}` since this is the first and only mention of "synchronizes with" in [mem].

### DIFF
--- a/source/memory.tex
+++ b/source/memory.tex
@@ -6156,7 +6156,7 @@ The previous value of the default memory resource pointer.
 Calling the \tcode{set_default_resource} and
 \tcode{get_default_resource} functions shall not incur a data race.
 A call to the \tcode{set_default_resource} function
-shall synchronize with subsequent calls to
+synchronizes with\iref{intro.races} subsequent calls to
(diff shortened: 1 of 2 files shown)


eisenwave:
@jensmaurer review please.

source/memory.tex
```diff
@@ -6156,7 +6156,7 @@
 Calling the \tcode{set_default_resource} and
 \tcode{get_default_resource} functions shall not incur a data race.
 A call to the \tcode{set_default_resource} function
-shall synchronize with subsequent calls to
+synchronizes with\iref{intro.races} subsequent calls to
```
jensmaurer: Please move the added  \iref to after "data race", for consistency with get_new_handler (above).

jensmaurer (approved):

PR #6851, opened 2024-03-04
[class.member.lookup,basic.lookup.argdep,basic.life] Remove extra whi…

Andreas-Krug:
…tespace

[class.member.lookup] 6.5.2-9 example 3
[basic.lookup.argdep] 6.5.4-2 note 1
[basic.life] 6.7.3-8.5 example 3

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -1716,11 +1716,11 @@ unique instance of the name that hides all the others.
 \end{note}
 \begin{example}
 \begin{codeblock}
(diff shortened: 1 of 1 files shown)


jensmaurer:
This is a case where our default answer in the past has been: "The core language examples don't follow a particular style, in order to demonstrate that style is irrelevant for the interpretation of the language".

@tkoeppe , does this apply to these intra-line whitespace changes?

PR #6852, opened 2024-03-08
[atomics.order] Make out-of-thin-air prevention Recommended practice

Eisenwave:
The wording heavily suggests that this paragraph is merely a recommendation. The following example even says "this recommendation".

Therefore, the paragraph should be formatted as *Recommended Practice*.

### DIFF
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -2600,6 +2600,7 @@ with respect to other atomic operations performed by the same thread.
 \end{note}
 
 \pnum
+\recommended
 Implementations should ensure that no ``out-of-thin-air'' values are computed that
 circularly depend on their own computation.
 


FrankHB:
Recommendations are expressed in "should" idiomatically. This is actually [an ISO/IEC rule](https://www.iso.org/sites/directives/current/part2/index.xhtml).

Note that normative text does not mean it is mandated. Under the linked ISO/IEC rules above, by definition, recommendations are also parts of provisions as the normative elements of a draft standard. Even it is not a part of the *mandatory requirements* of conformance, it is not explicitly marked as "informative" and it has effects on positive ("allowed") judgement of conformance on some optional settings of the standard features (similar to those like conditionally supported features), hence normative.

A quick search of `\recommended` reveals some similar uses (with "should"), so the rejection is problematic. @jensmaurer please confirm if other instance of `\recommended` should follow the decision here. I doubt the should not, because I see the history in [N4867](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/n4867.html) suggesting it is "to address ISO/CS 017 (C++20 DIS)", and this seems nothing different here (except that the text here is never in some note). Adding the format is about QoI of consistency in the current draft. How to change the wording a mandatory requirement is another story (and largely non-editorial). Before the change is adopted, if we have the consensus to make the rule mandatory as the future direction, better add one more statement about it in the note section following the normative text.

jensmaurer:
Edited my earlier comment to say "normative requirement" and "editorial patches".

That means: This is a sensitive area of wording. I don't want to accidentally do something editorially here that isn't the intent of SG1 and/or CWG.  If you have constructive ideas how to improve the situation around forbidding out-of-thing-air values, feel free to write a paper to SG1.

PR #6857, opened 2024-03-15
[utilities.swap] Replace unusual "stored in two locations" wording

jwakely:
This addresses the only part of LWG 4056 that LWG wanted to change. See https://cplusplus.github.io/LWG/issue4056


tkoeppe:
@jwakely Would you mind force-rebasing this to get the tests to rerun?

PR #6894, opened 2024-03-23
P2809R3 Trivial infinite loops are not Undefined Behavior

### LINKED ISSUE #6868: [2024-03 CWG Motion 6] Trivial infinite loops are not Undefined Behavior
https://wiki.edg.com/pub/Wg21tokyo2024/StrawPolls/P2809R3.html

jensmaurer:
Fixes #6868
Fixes cplusplus/papers#1496

### DIFF
--- a/source/statements.tex
+++ b/source/statements.tex
@@ -546,6 +546,36 @@ while (--x >= 0) {
 Thus after the \keyword{while} statement, \tcode{i} is no longer in scope.
 \end{example}
 
+\pnum
+A \defnadj{trivially empty}{iteration statement} is
+an iteration statement matching one of the following forms:
+\begin{itemize}
+\item \tcode{while (} \grammarterm{expression} \tcode{) ;}
+\item \tcode{while (} \grammarterm{expression} \tcode{) \{ \}}
+\item \tcode{do ; while (} \grammarterm{expression} \tcode{) ;}
+\item \tcode{do \{ \} while (} \grammarterm{expression} \tcode{) ;}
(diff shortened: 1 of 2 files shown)


source/statements.tex
```diff
@@ -543,9 +543,40 @@
...
+\item \tcode{for (} \grammarterm{init-statement} \opt{\grammarterm{expression}} \tcode{; ) \{ \}}
+\end{itemize}
+The \defnadj{controlling}{expression} of a trivially empty iteration statement
+is the \grammarterm{expression} of
+a \tcode{while}, \tcode{do}, or \tcode{for} statement
+(or \tcode{true}, if the \tcode{for} statement has no \grammarterm{expression}).
+A \defnadj{trivial infinite}{loop} is a trivially empty iteration statement
+for which the converted controlling expression is a constant expression,
+when interpreted as a \grammarterm{constant-expression}\iref{expr.const}, and
+evaluates to \tcode{true}.
```
cpplearner, line 568: Does this "interpreted as a *constant-expression*" wording mean that the following will be treated as a trivial infinite loop?

(Note that `std::is_constant_evaluated()` returns `true` when evaluated in a *constant-expression*, and `false` when not manifestly constant-evaluated.)

```c++
#include  <type_traits>
int main() {
    while (std::is_constant_evaluated());
}
```

jensmaurer: Yes, that is a "trivial infinite loop".  That means, optimizations are reduced around that loop and each step of the loop execution will call `yield`. Except that the body of the loop will never actually be executed at runtime.

Issue #6875, opened 2024-03-23
[2024-03 LWG Motion 4] Remove Deprecated shared_ptr Atomic Access APIs from C++26


jensmaurer:
https://wiki.edg.com/pub/Wg21tokyo2024/StrawPolls/P2869R4.pdf

AlisdairM:
Could this issue be assigned to me?  I don't see the UI to assign issues.

AlisdairM:
This PR has been consolidated into #6921 that resolves merge conflicts among the 4 deprecation removal papers.

PR #6902, opened 2024-03-24
P2869R4 Remove deprecated atomic free function API for shared_ptr

AlisdairM:
The Annex C wording is modeled on the compatibility with C form, and should be revised without the "difficulty of converting" clause, probably with a simple code example in "Effects".

Fixes https://github.com/cplusplus/draft/issues/6875
Fixes https://github.com/cplusplus/papers/issues/1525


AlisdairM:
Fixes https://github.com/cplusplus/draft/issues/6875
Fixes https://github.com/cplusplus/papers/issues/1525

PR #6905, opened 2024-03-26
[variant.relops] Fix \exposid in \codeblock

jwakely:


jwakely:
image

jwakely:
This was introduced by c44868c85b01c7f7abfae278fa5b56fa6a0e2b5e so is not present in C++23

PR #6920, opened 2024-04-03
[check] Fix overly aggressive 'constexpr static' check

jensmaurer:


jensmaurer:
@tkoeppe, this is need for the pull requests of at least two papers.

PR #6923, opened 2024-04-09
Correctly render packages installs

CosminGGeorgescu:


jensmaurer:
This adds additional (duplicate) colons and adds trailing whitespace in one instance.

The former seems to satisfy https://docutils.sourceforge.io/docs/user/rst/quickref.html#literal-blocks , but the latter is a bit undermotivated. Care to elaborate?

CosminGGeorgescu:
> but the latter is a bit undermotivated

Just a tabbing tic, seemed to go over my head when reviewing.

I can add a commit removing it (a bit of an eye sore), squash them together (which would need a force :[ ) or close it and let this be a post-it note for the beholder to squeeze in their next PR.

jwakely:
Please force push, there's no problem with doing that in a pull request. Nobody else is fetching this branch.

PR #6924, opened 2024-04-16
[expos.only.entity] Add missing \expos for two exposition-only names

hewillk:
..Also fixed an issue where the first exposition-only comment wasn't quite aligned.


jensmaurer:
Looks better. Please squash the two commits and force-push.

PR #6925, opened 2024-04-16
[zombie.names] Turn lists of zombie names into tables

### LINKED ISSUE #6898: Refactor list of zombie names as a table
The list of zombie names his over a page long, and is growing further after the Tokyo meeting.

Once the Tokyo polls have landed, the list should be refactored into a table, similar to the table of reserved words, to present the information more compactly, in a form that is easier to read.

AlisdairM:
Fixes #6898

As the list of zombie names has grown so long, it is hard to check without scrolling.  Rendering as a table makes it easier to read in one place, as well as saving on our page count!

Pre-emptively turned all the other zombie name lists into tables too, for consistency.  We have precedent for such short tables with the table of contextual keywords.

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -3021,107 +3021,132 @@ undefined.%
 \indextext{brains!names that want to eat your}%
 
 \pnum
-In namespace \tcode{std}, the following names are reserved for previous standardization:
-\begin{itemize}
-\item \indexlibraryzombie{auto_ptr} \tcode{auto_ptr},
-\item \indexlibraryzombie{auto_ptr_ref} \tcode{auto_ptr_ref},
-\item \indexlibraryzombie{binary_function} \tcode{binary_function},
-\item \indexlibraryzombie{binary_negate} \tcode{binary_negate},
-\item \indexlibraryzombie{bind1st} \tcode{bind1st},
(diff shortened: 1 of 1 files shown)


jensmaurer (approved):
I've compared the names one-by-one. The diff seems fine.

Please squash the fixup.

source/lib-intro.tex
```diff
@@ -3022,106 +3022,128 @@
 
 \pnum
 In namespace \tcode{std}, the following names are reserved for previous standardization:
```
timsong-cpp, line 3024: Since the tables are floats, we probably want to refer to them by number/reference rather than "the following"?

AlisdairM:
do not merge until I address the comment above by @timsong-cpp

source/lib-intro.tex
jensmaurer: Good catch. That's even an ISO requirement.

Also, the table labels should start with the section label "zombie.names", always.  So, it's "zombie.names.om" for object macros, maybe. And "zombie.names.fm" for function macros.

AlisdairM:
The latest commit should have squashed updates that explicitly reference the tables, and give them appropriate
\[zombie.names.tag\] labels.

jensmaurer (approved):

PR #6929, opened 2024-04-17
[range.concat.view] Formatting code to match the current style

hewillk:


jensmaurer (approved):

jensmaurer:
@jwakely , this feels like an improvement to me. What do you think?

jwakely (approved):
Yeah, this seems like an improvement

PR #6932, opened 2024-04-18
[range.utility.conv.general] Fix typo

cpplearner:
This appears to be a mistake when applying [LWG 4016](https://cplusplus.github.io/LWG/issue4016).

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -2268,10 +2268,10 @@ constexpr bool @\exposid{container-appendable}@ =          // \expos
 \end{codeblock}
 
 \pnum
-Let \exposid{container-appendable} be defined as follows:
+Let \exposid{container-append} be defined as follows:
 \begin{codeblock}
 template<class Ref, class Container>
-constexpr auto @\exposid{container-appendable}@(Container& c) {                // \expos
+constexpr auto @\exposid{container-append}@(Container& c) {                // \expos
   return [&c]<class Ref>(Ref&& ref) {
     if constexpr (requires { c.emplace_back(declval<Ref>()); })
       c.emplace_back(std::forward<Ref>(ref));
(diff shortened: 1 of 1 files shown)


source/ranges.tex
```diff
@@ -2268,10 +2268,10 @@
 \end{codeblock}
 
 \pnum
-Let \exposid{container-appendable} be defined as follows:
+Let \exposid{container-append} be defined as follows:
 \begin{codeblock}
 template<class Ref, class Container>
-constexpr auto @\exposid{container-appendable}@(Container& c) {                // \expos
+constexpr auto @\exposid{container-append}@(Container& c) {                // \expos
```
tkoeppe, line 2274: Please move the comment back appropriately to retain its column (which should be on one of the preferred places 20, 32, 40, ... (see #include "semantic/understanding/rene/flumes/understand.h").

tkoeppe:
Thanks, good catch!

source/ranges.tex
```diff
@@ -2268,10 +2268,10 @@
 \end{codeblock}
 
 \pnum
-Let \exposid{container-appendable} be defined as follows:
+Let \exposid{container-append} be defined as follows:
 \begin{codeblock}
 template<class Ref, class Container>
```
cpplearner, line 2273:
```suggestion
template<class Container>
```
https://github.com/cplusplus/draft/commit/1958e5063b6c5c6893ba509eff02bf933e1a5e02 says:

> [range.utility.conv.general]p5 Template parameters are reversed from what the paper has; resolution applied to body but not to remove Ref.

Nonetheless, I think the only reasonable interpretation of [LWG 4016](https://cplusplus.github.io/LWG/issue4016) is to remove the template parameter `Ref`.

cc @jwakely

jwakely: Yes, obviously it needs to be removed. It is no longer used in the function body, and an explicit template argument is no longer present when that function is called on line 2274. Since it's not deducable, that makes line 2274 ill-formed.

The change in the issue is clearly to remove it from the function's template parameters, even if I fudged the order of those parameters.

jwakely (approved):

jensmaurer (approved):

PR #6933, opened 2024-04-19
[range.concat.view] Add missing \exposid to `make-unsigned-like-t`

hewillk:


hewillk:
Not sure if this is worth a LWG since `(CT(sizes) + ...)` may overflow.

jensmaurer (approved):

PR #6934, opened 2024-04-19
[range.concat.iterator] Remove @ in \exposid

hewillk:


jensmaurer:
The change looks good, but we now have an overfull \hbox:

Overfull \hbox (18.69255pt too wide) in paragraph at lines 8841--8845

Please fix that, squash and force-push.

PR #6936, opened 2024-04-19
LWG3031 Fix missing code-formatting of const

Eisenwave:
https://cplusplus.github.io/LWG/issue3031 proposed approved wording that includes:
> (possibly `const`)

However, all of these occurrences were formatted as plain "(possibly const)".

On top of being an editorial mistake, it's also somewhat misleading, as this shifts the intuition of the reader towards "const objects", and away from "`const` on `T`". It's clear from the surrounding wording that this `const` refers to types, but it's still easier to understand the wording with intended formatting in my opinion.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -159,7 +159,7 @@ the type \tcode{decltype(pred(*first))} shall model
 \exposconcept{boolean-testable}\iref{concept.booleantestable}.
 The function object \tcode{pred} shall not apply any non-constant function
 through its argument.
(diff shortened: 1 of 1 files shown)


jwakely (approved):

jensmaurer (approved):

Eisenwave:
@jensmaurer I probably should have noticed this sooner, but while the wording change is approved by LWG, I believe the editorial decision in https://github.com/cplusplus/draft/issues/117 supersedes that.

I guess that's why it hasn't been edited into the draft with `const` included.

PR #6938, opened 2024-04-19
Revert "LWG3031 Fix missing code-formatting of const"

jensmaurer:
As agreed in #117


tkoeppe:
Let me just force-push the removal.

tkoeppe:
Yes, I had been wondering whether unmarked-up "possibly const" isn't the normal way we say things...

PR #6939, opened 2024-04-19
[range.access] Clarify notes on SFINAE for CPOs

AlisdairM:
Readers need a deep knowledge of the standard to understand why the notes on diagnosable ill-formed code are already specified, rather than serving as a normative Remark.

Here we call out that the SFINAE constraints are specified for the function call operator of all customization point objects.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -670,7 +670,7 @@ Then:
 
 \pnum
 \begin{note}
-Diagnosable ill-formed cases above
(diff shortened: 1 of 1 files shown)


jensmaurer:
I'm sorry, I don't think "for customization point objects" adds anything useful at all to the SFINAE consideration. We already say "above" and refer to the bulleted lists, which seems clear before and after. If the SFINAE situation needs highlighting, then that needs different words.

PR #6941, opened 2024-04-20
[utility.arg.requirements] Break down requirement tables

### LINKED ISSUE #6033: Dismantling requirements tables 29-36
To convert the lib-intro requirements tables 29-36 into paragraphs, we don't currently have any section, and no good way to reference them.

Idea:

1. A new subsection "16.4.4.?: Basic type requirements", after [16.4.4.2, utility.arg.requirements] and before [16.4.4.3, swappable.requirements].
2. New nested subsections for each individual requirement, so that we can reference them individually (which is currently possible via the tables).

Thoughts?

AlisdairM:
Fixes #6033, late, but as promised.

First draft that will probably need some consistent rewording.

Break down the requirements tables into regular specification text, following the lead of breaking down the container requirements.  This also solves the problem of floating tables floating into unrelated clauses, as there will be no more floating tables.

For simplicity, new subsections have the same stable label as the table they replace, which simplifies the diff but are probably not the best names.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -3458,10 +3458,10 @@ template<class InputIterator, class Function>
 \pnum
 \expects
 \tcode{Function} meets
-the \oldconcept{MoveConstructible} requirements (\tref{cpp17.moveconstructible}).
+the \oldconcept{MoveConstructible} requirements\iref{cpp17.moveconstructible}.
 \begin{note}
 \tcode{Function} need not meet the requirements of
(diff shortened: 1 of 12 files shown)


jensmaurer:
This should be reviewed by LWG.

AlisdairM:
This PR is badly structured, and should have one commit per table that is deconstructed, allowing for easier review.

Note that https://github.com/cplusplus/draft/pull/6034 is another attempt to solve the same problem.

PR #6942, opened 2024-04-21
[range.concat.iterator] Remove redundant \expos

hewillk:
..which was already there when it was declared.


CaseyCarter (approved):
LGTM. This is consistent with Library style.

jwakely (approved):

jensmaurer (approved):

PR #6956, opened 2024-04-28
[alg.rand.generate] Remove redundant `}`

hewillk:
..which is a typo.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -11776,7 +11776,7 @@ ranges::generate(std::forward<R>(r), [&d, &g] { return invoke(d, g); });
 \remarks
 The effects of \tcode{generate_random(r, g, d)} shall be equivalent to
 \begin{codeblock}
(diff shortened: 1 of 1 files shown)


hewillk:
Some thoughts:

_Remarks_: The effects of `generate_random(r, g, d)` shall be equivalent to
`ranges::generate(std::forward<R>(r), [&d, &g] { return invoke(d, g); })`

Is it more accurate to change the above to the following?

_Remarks_: The effects of `generate_random(std::forward<R>(r), g, d)` shall be equivalent to
`ranges::generate(std::forward<R>(r), [&d, &g] { return invoke(d, g); })`

or:

_Remarks_: The effects of `generate_random(r, g, d)` shall be equivalent to
`ranges::generate(r, [&d, &g] { return invoke(d, g); })`

Not sure if this is editorial.

jensmaurer (approved):

PR #6964, opened 2024-05-07
[memory.syn][specialized.algorithms] Prefer trailing returns on complex signatures

AlisdairM:
This proposed change follows the observation that when function template signatures get complicated, especially with long names for return types, and non-trivial requires clauses, it can be very hard to spot the function name and return type among the sea of tokens, in order to start mentally parsing the function.

Here we propose using the trailing return type form of function declaration in those cases, and apply it consistently to the special algorithms in the '<memory>' header as a non-trivial sample size to gauge interest in a more complete PR that would cover the whole ranges library.


jensmaurer (commented):
I must say, I'm underimpressed.  Some changes seem outright non-improvements, such as `I f(...)` -> `auto f(...) -> I`.

`pair<ForwardIterator, NoThrowForwardIterator>` seems to be the most complicated return type I can spot, and that doesn't feel overly complicated.

In any case, I want the opinion of LWG on a global change like that.

@jwakely ?

AlisdairM:
This PR has gone stale, and was not selling well, so as author I am closing it.
I have not given up the original idea completely, but as Jens pointed out, at the very least taking a blanket policy per the original PR is has at least some regressions rather than improvements.

Any future PR trying to solve the problem of easier reading of complication function declarations will be more precisely targeted and a clear improvement, or will not be submitted.

PR #6965, opened 2024-05-07
[memory.syn] Move specification macros to appropriate subclause

AlisdairM:
The specification macros _POINTER_OF_ and _POINTER_OF_OR_ do not appear in the synopsis of the `<memory>` header, but are in fact used only to specify behavior of the smart pointer adapters.  Created a new Definitions heading in that subclause, and moved the text there.

In addition, add the two specification macros to the main index, and also the index of library names, following the example set by _INVOKE_. Given the choice, I would rather all of these entries were in just the main index, but following precedent.

### DIFF
--- a/source/memory.tex
+++ b/source/memory.tex
@@ -41,30 +41,6 @@ The header also defines the templates
 \tcode{out_ptr_t}, \tcode{inout_ptr_t}, and various function
 templates that operate on objects of these types\iref{smartptr}.
 
-\pnum
-Let \tcode{\exposid{POINTER_OF}(T)} denote a type that is
-\begin{itemize}
-\item
-\tcode{T::pointer} if the \grammarterm{qualified-id} \tcode{T::pointer}
-is valid and denotes a type,
-\item
-otherwise, \tcode{T::element_type*}
-if the \grammarterm{qualified-id} \tcode{T::element_type}
-is valid and denotes a type,
-\item
-otherwise, \tcode{pointer_traits<T>::element_type*}.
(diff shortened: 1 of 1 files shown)


source/memory.tex
```diff
@@ -5082,6 +5058,36 @@
 
 \rSec2[smartptr.adapt]{Smart pointer adaptors}
 
+\rSec3[smartptr.adapt.defns]{Definitions}
```
jensmaurer, line 5061: ISO wants only a single section "Terms and Definition" (the one in clause 3); not sure whether this is too close for their taste.

Do we have other such "Definitions" sections elsewhere?

source/memory.tex
```diff
@@ -5082,6 +5058,36 @@
...
+
+\pnum
+\indextext{pointer_of@\tcode{\placeholder{POINTER_OF}}}%
+\indexlibrary{pointer_of@\tcode{\placeholder{POINTER_OF}}}%
```
jensmaurer, line 5065: The text for the index of library names says "Constructions whose name appears in italics are for exposition only."

so we seem to have resolved to put exposition-only stuff there (at least more widely used things).

jensmaurer: There should be a record of that decision somewhere...

AlisdairM:
Wondering if @jwakely could weigh in on this?

source/memory.tex
tkoeppe: Indeed, let's not make this move, that seems to invite trouble.

PR #6966, opened 2024-05-09
[range.reverse.overview] Replace 'equivalent to' with 'then'

hewillk:
since those 'equivalent to' s seems superfluous.


jwakely (approved):
I agree they're superfluous given the "is expression-equivalent to:" before the list.

See also [range.take.overview] which has exactly the same form as proposed here.

jensmaurer (approved):

PR #6969, opened 2024-05-11
[class.derived.general] Restore accidental reversal of P2662R3 change

jensmaurer:
P2662R3 contained a rename of the grammar production "class-or-decltype" to "class-or-computed-type-specifier".  That was editorially reverted with commit c831ec0b8aac369aa61ce392783865ff04b84b19, but that commit also accidentally reverted the change from "decltype-specifier" to "computed-type-specifier" as one of the options for "class-or-decltype".

This commit restores the latter change, as intended by P2662R3.


tkoeppe (approved):

tkoeppe:
Thanks!

PR #6970, opened 2024-05-12
[format.context] Fix incorrect example

hewillk:
`isdigit()` is not a `contexpr` function so it cannot be used in `parse()`, and the alignment seems to be in the wrong direction.
Demo of the original example: https://godbolt.org/z/qbs6dvq9q
fixed: https://godbolt.org/z/sjWdxxzWc


jwakely (approved):

hewillk:
Sorry, I think there is still something wrong with this example, because` if (value < 0 || value > numeric_limits<int>::max())` will trigger the warning of `comparison of constant '0' with boolean expression is always false`.
@jwakely, any thoughts?

jwakely:
I agree, we don't need to fix warnings. The `bool` case works correctly despite the warnings (and isn't actually used at runtime anyway).

jensmaurer (approved):

PR #6972, opened 2024-05-12
[meta.const.eval] Fix `is_within_lifetime` example

MitalAshok:
`operator*` is invoked on a `const` object below (`static_assert(*engaged);`) but the member function is not `const`-qualified


MitalAshok:
Completed with #7234

Issue #6976, opened 2024-05-13
[ISO/CS C++2023] Remove the word "subclause"


jensmaurer:
- [ ] 13.5.1 General remove the word "subclause". Use the subclause number only to cross-reference a subclause.
- [ ] 19.5.1 General remove the word "subclause"
- [ ] 19.6.1 General remove word "subclause"
- [ ] 24.2.2.1 Introduction remove the word "subclause"
- [ ] 24.5.5.1 Overview Please remove the word "subclause" - the subclause number sufficies as a cross-reference. Update accordingly throughout the document.
- [ ] 29.12 Formatting please remove the word "subclause"
- [ ] 33.5.1 General remove the word "subclause"
- [ ] 33.6.1 General please remove the word "subclause"
- [ ] 33.9.1 General remove "subclause"
- [ ] C.7.4 Modifications to declarations Remove "subclause"

tkoeppe:
I've asked ISO whether it's OK to retain "Subclause 1.2.3 describes feature X" to start a sentence, since otherwise we'd come up against the natural language style rule to not start a sentence with a symbol. Please hold off on changing that.

Interior changes like "see ~subclause~ 1.2.3" are fine.

tkoeppe:
I think we've done what we can here.

PR #6982, opened 2024-05-13
[std] Remove mid-sentence 'subclause' introducer

jensmaurer:

Partially addresses #6976

### DIFF
--- a/source/concepts.tex
+++ b/source/concepts.tex
@@ -1222,7 +1222,7 @@ fundamental types like \tcode{int} and that are comparable with
 \rSec2[concepts.callable.general]{General}
 
 \pnum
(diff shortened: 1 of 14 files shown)


tkoeppe:
Thank you - I think we can live with this, and it's actually a fairly reasonable change. If we all agree, we should make this standing practice (and update the wiki). @jwakely?

jwakely:
I agree this one is reasonable, and actually an improvement in a couple of cases.

jwakely (approved):

tkoeppe:
Thanks! Let's do it. And we are allowed to keep "Subclause 1.2.3" in places where natural language style benefits from leading with a word instead of a symbol.

PR #6991, opened 2024-05-14
[util.smartptr.weak.general] Clarify when a `weak_ptr` is empty

eisenwave:
It is not entirely clear whether an expired (but not null) `weak_ptr` is said to be "empty".

@t3nsor had to ask about this two years ago (https://stackoverflow.com/q/73253624/5740428) and I found it difficult answering another Q&A (https://stackoverflow.com/q/78475345/5740428) because it isn't obvious when a `weak_ptr` is empty. I could only be confident in my answer once I've looked into the libstdc++ implementation of `compare_exchange_weak` for `weak_ptr`.

There are ways to obtain an answer indirectly by digging through various wording or implementations, but it should be more obvious. [[util.smartptr.shared.general] paragraph 1](http://eel.is/c++draft/util.smartptr.shared.general#1) already states what it means for `shared_ptr` to be empty, so why don't we add corresponding wording for `weak_ptr`?

### DIFF
--- a/source/memory.tex
+++ b/source/memory.tex
@@ -4508,6 +4508,8 @@ The \tcode{weak_ptr} class template stores a weak reference to an object
 that is already managed by a \tcode{shared_ptr}. To access the object, a
 \tcode{weak_ptr} can be converted to a \tcode{shared_ptr} using the member
 function \tcode{lock}.
+A \tcode{weak_ptr} may become empty by expiring,
+without a call to one if its member functions.
 
 \begin{codeblock}
 namespace std {


Eisenwave:
No, I think I am wrong on this. Especially after [LWG 3195](https://cplusplus.github.io/LWG/lwg-defects.html#3195), the wording implies that a `weak_ptr` being empty and a `weak_ptr` storing a null pointer value are two separate properties.

jensmaurer:
@jwakely , this feels like a mild nudge in the right direction to me. However, if there is some clarity missing, should this maybe be covered by an LWG issue?

jwakely (commented):
Although it's obvious to you and me that "expiring" means "transitioning to a state such that `expired()` is `true`", the standard doesn't actually define that term.

And I think this suggested change still doesn't define what "empty" means for a `weak_ptr`, via something equivalent to "A `shared_ptr` is said to be empty if it does not own a pointer."

And I don't agree with the implied meaning here that an expired `weak_ptr` becomes empty. I think it doesn't. When a non-empty `weak_ptr` expires, the `weak_ptr` still has a pointer to a shared control block with a non-zero weak count. That's not equivalent to a default constructed `weak_ptr`. Making a copy of that `weak_ptr` does not create an empty object, but one that shares ownership with it (increasing the weak count).

I think this change creates more questions than it answers, and is arguably just wrong.

jwakely:
https://compiler-explorer.com/z/dbbd7KPhP demonstrates that an expired `weak_ptr` that stores a null pointer is not equivalent to an empty `weak_ptr` that stores a null pointer.

The GDB pretty printer for `weak_ptr` shows the following distinct states for the `atomic<weak_ptr<int>>`:

```
$1 = std::atomic<std::weak_ptr<int>> (empty) = {get() = 0x0}
$2 = std::atomic<std::weak_ptr<int>> (use count 2, weak count 1) = {get() = 0x0}
$3 = std::atomic<std::weak_ptr<int>> (expired, weak count 1) = {get() = 0x0}
```

The expired state is not the same as the empty state.

jwakely:
I think a more accurate statement would be that a `weak_ptr` is empty if it does not share ownership with any `shared_ptr` or `weak_ptr` objects. The result of calling `expired()` on a `weak_ptr` can change as a result of operations on objects it shares ownership with, without calling any member functions on the `weak_ptr` itself. But that's also true of `weak_ptr::use_count()` and `shared_ptr::use_count()` so I'm not sure we want to call it out explicitly for `expired()` but not the two `use_count()` functions.

t3nsor:
> I think a more accurate statement would be that a `weak_ptr` is empty if it does not share ownership with any `shared_ptr` or `weak_ptr` objects.
That doesn't seem quite right either, because an expired `weak_ptr` doesn't share ownership with any `shared_ptr`s (although it used to) but that still doesn't make it empty.

jwakely:
Good point, that's why `shared_ptr` defines it in terms of "owns a pointer", which is a euphemism for "owns a control block" but that's not wording we can use in the standard. When an object `p` shares ownership with `r` that means they both own the _same_ control block. But what matters is not being co-owner of a control block, just being an owner, including when the sole owner.

jwakely:
Maybe we should just change "A shared_ptr is said to be empty if it does not own a pointer." to mention `weak_ptr` too. The relevant `weak_ptr` constructors say they share ownership, which is the same wording that `shared_ptr` constructors use to say they share ownership with an object that "owns a pointer".

t3nsor:
> Good point, that's why `shared_ptr` defines it in terms of "owns a pointer", which is a euphemism for "owns a control block" but that's not wording we can use in the standard.

Maybe we can invent a better euphemism. For example:

The raw pointer constructor of `shared_ptr` creates a "title" to a pointer and the `shared_ptr` refers to that title

The destructor of `shared_ptr` has an effect only if the `shared_ptr` refers to a title that no other `shared_ptr` refers to

The copy constructor of `shared_ptr` makes a `shared_ptr` that refers to the same title as the source object (if any)

A `weak_ptr` is empty if it doesn't refer to a title

Eisenwave:
> https://compiler-explorer.com/z/dbbd7KPhP demonstrates that an expired `weak_ptr` that stores a null pointer is not equivalent to an empty `weak_ptr` that stores a null pointer.

I am not saying that the two are interchangeable. I am saying that we should consider both to be empty. My initial intuition was also (similar to GDB) that a null `weak_ptr` is empty, and an expired `weak_ptr` isn't, but this does not match the remaining wording.

> I think a more accurate statement would be that a `weak_ptr` is empty if it does not share ownership with any `shared_ptr` or `weak_ptr` objects.

If we want to be very specific despite this being a semi-normative paragraph, this also raises some questions. Afaik "sharing ownership" is also not formally defined, so it's not any better than saying that an expired pointer is also empty.

If we *really* want to be precise, we can say that a `weak_ptr` `p` is empty if `p.expired()` is `true`.

> But that's also true of `weak_ptr::use_count()` and `shared_ptr::use_count()` ...

Yes, `use_count()` can change without interaction with the object for either type. However, `weak_ptr` can become empty/expired/useless underneath your feet, and I felt this was special enough to point out in the semi-normative introductory paragraph. Anyhow, it's not the hill to die on, we don't need to state it. Implementors already know.

> I think this change creates more questions than it answers, and is arguably just wrong.

My choice of wording may now have been ideal, but I don't see how it is wrong. There is clearly a distinction between an empty `weak_ptr` and one which stores a null pointer. The wording added by [LWG 3195](https://cplusplus.github.io/LWG/lwg-defects.html#3195) clearly supports this.

If only null pointers and not expired pointers are empty, then a lot of the current wording is simply redundant. We can simply remove all uses of "empty" and simplify those to "stores null pointer value". However, this makes no sense given present wording. For example, the move constructor has these postconditions:

> *Postconditions:* `this` contains the old value of `r`. `r` is empty, stores a null pointer value, and `r.use_count() == 0`.

If all empty weak pointers stored a null pointer value, then you wouldn't need to explicitly say that the moved-from object is empty. That leaves two more states that could be empty:
- Stores a non-null pointer value, but `expired()` is `false`.
- Stores a non-null pointer value, but `expired()` is `true`.

Obviously, a non-null non-expired `weak_ptr` isn't empty, so it has to be the latter. I don't see any other state which the wording could describe as empty.

Eisenwave:
Also seeing how contentious and difficult this is, it may be a good idea to put LWG on this. It's rapidly escaping editorial territory.

jwakely:
> If we _really_ want to be precise, we can say that a `weak_ptr` `p` is empty if `p.expired()` is `true`.

But that would be inconsistent with the meaning for `shared_ptr`, because a `weak_ptr` can still own a control block after it's expired.

Empty means if doesn't own a control block.

> However, weak_ptr can become empty/expired/useless underneath your feet

Well, yeah, that's kinda the whole point of using it. And it's not useless, because you can safely detect the case where the stored pointer has become invalid, unlike a raw pointer or `observer_ptr`.

And I still dispute it can become empty that way. expired != empty.

> If only null pointers and not expired pointers are empty

That's not what I'm saying though.


> If all empty weak pointers stored a null pointer value, then you wouldn't need to explicitly say that the moved-from object is empty. 

And if all expired weak pointers were empty, the postcondition wouldn't need to say "is empty" and **also** say that the use_count() is zero. So by your own logic, expired != empty.

>  I don't see any other state which the wording could describe as empty.

You're not looking hard enough. As I've said already in the comment you replied to, it's empty when it doesn't own a control block. The wording doesn't describe the property "owns a control block", but that doesn't mean it doesn't exist.

The way the standard refers to "owns a control block" is "owns a pointer", and it explicitly says a `shared_ptr` is empty when it doesn't own a pointer. The constructors that create a non-empty `shared_ptr` all say something like either "owns the pointer p" or "shares ownership with r". That ownership property is what determines being non-empty. And it's the same for `weak_ptr`. The constructors either create an empty `weak_ptr`, or one that "shares ownership with r". The use count dropping to zero **does not end the ownership**. It still owns a pointer (even if that pointer is now invalid) and the associated control block which stores that pointer and the ref counts.

expired is not empty.

Eisenwave:
> Empty means it doesn't own a control block.

Well, it took me a while to wrap my head around this, but I get what you're saying now. The obvious follow-up question is whether this even makes sense to have in wording though.

The control block isn't user-accessible, and it doesn't seem possible to put a `weak_ptr` into a state where it's empty in a way that isn't synonymous with something else.

The ways in which a `weak_ptr` can become empty are:
- [`weak_ptr()`](https://eel.is/c++draft/util.smartptr.weak#lib:weak_ptr,constructor) constructs an empty `weak_ptr` that stores a null pointer value, and `use_count() == 0`.
- [`weak_ptr(const weak_ptr&)`](https://eel.is/c++draft/util.smartptr.weak#lib:weak_ptr,constructor_) possibly constructs an empty `weak_ptr` that stores a null pointer value.
- [`weak_ptr(weak_ptr&&)`](https://eel.is/c++draft/util.smartptr.weak#lib:weak_ptr,constructor__) leaves the moved-from pointer in an empty state and storing null pointer value, with `use_count() == 0`.
- `swap` exchanges contents and assignment operators are defined as construct & swap equivalences.
- atomic `exchange`, `compare_exchange` et al. are defined to assign or swap.

Unless I've missed some way to obtain an empty pointer, *every case* where an empty `weak_ptr` is obtained is also a case where this `weak_ptr` stores a null pointer value, and vice versa. Storing a null pointer value, `use_count() == 0`, and "empty" appear to be equivalent in the current wording, but only if emptiness isn't also entered through expiring.

How do you obtain a non-null empty `weak_ptr`, i.e. a `weak_ptr` that does not have a control block but isn't null? None of its member functions seem to achieve that.

jwakely:
> Unless I've missed some way to obtain an empty pointer, _every case_ where an empty `weak_ptr` is obtained is also a case where this `weak_ptr` stores a null pointer value,

Yes.

> and vice versa.

No. See the compiler explorer link I provided earlier. That has a non-empty `weak_ptr` with a null pointer.

> Storing a null pointer value, `use_count() == 0`, and "empty" appear to be equivalent in the current wording, but only if emptiness isn't also entered through expiring.

They are not all equivalent. See the compiler explorer link which demonstrates that a null pointer value can be present in an empty weak_ptr, a non-empty non-expired weak_ptr, and a non-empty expired weak_ptr.

> How do you obtain a non-null empty `weak_ptr`, i.e. a `weak_ptr` that does not have a control block but isn't null? None of its member functions seem to achieve that.

I don't think you can. The `shared_ptr` aliasing constructor allows a non-null empty `shared_ptr`, but if you create a `weak_ptr` referring to that then it has a null pointer instead.

But just because you can't create a non-null empty weak_ptr, doesn't mean the inverse is true. You can create a non-empty null weak_ptr.

Can we please take this somewhere else. It's gone far from an editorial tweak now, and I'm repeating thigns I said 7 hours ago.

AlisdairM:
I do not understand the problem --- the standard does not talk about empty `weak_ptr`s because the term does not apply to weak pointers and never has.  If we are to apply the shared_ptr definition, `weak_ptr`s are *always* empty, because their whole purpose is to be non-owning, and empty is defined in terms of ownership.

jwakely:
> the standard does not talk about empty `weak_ptr`s because the term does not apply to weak pointers and never has.

It certainly does, see the spec for `weak_ptr::use_count()` for example.

> If we are to apply the shared_ptr definition, `weak_ptr`s are _always_ empty, because their whole purpose is to be non-owning, and empty is defined in terms of ownership.

As I wrote above, "owning a pointer" is a euphemism for "owns a control block" and a non-empty weak_ptr participates in shared ownership of a control block, via the weak count.

An empty weak_ptr is one that does not have (weak) ownership of a control block.

Issue #6981, opened 2024-05-13
[ISO/CS C++2023] Bad use of italics and bold


jensmaurer:
-  [x] The italics here "recommended practice" should be removed, as they are used for emphasis. The exceptional use of italics in this document has only been granted for a specific use, i.e. to show the function of certain text, not for emphasis. Please make this normal text.
- [x] 30.4.3.2.3 Virtual functions please remove bold font from "Stage 1", "Stage 2" etc. Bold is only permitted for table titles, figure titles, subclause titles etc.
- [x] 30.4.3.3.3 Virtual functions please remove bold font from "Stage 1", "Stage 2" etc. Bold is only permitted for table titles, figure titles, subclause titles etc
- [x] 33.10.10.3 Globals Annex A is in bold with "(informative)" below, not in bold. The text should also be centred.  See example in the Model Standard: https://www.iso.org/files/live/sites/isoorg/files/developing_standards/docs/en/model_document-rice_model.pdf  
- [x] B Implementation quantities Annex B is in bold with "(normative)" below, not in bold.  See example in the Model Standard: https://www.iso.org/files/live/sites/isoorg/files/developing_standards/docs/en/model_document-rice_model.pdf 
- [x] C.1.2 Clause 5: lexical conventions General comment for this entire Annex Please remove bold text here - bold is only used for specific reasons (i.e. subclause titles, figure titles etc.). If this text needs to stand out, it can be made into a list format with bullet points.

tkoeppe:
We might keep this specific to the IS.

tkoeppe:
Done for the IS.

PR #6996, opened 2024-05-14
[styles] Redesign Annex titles per Rice Model Standard

jensmaurer:

Partially addresses #6981


jensmaurer:
annex

tkoeppe:
Delightful, thanks!

PR #6997, opened 2024-05-14
[meta.unary.prop] [class.default.ctor] [class.dtor] Indexing and xrefs for the word "trivial"

Quuxplusone:
The definition of "trivial type" isn't relevant to the notion of "non-trivial function" (which isn't really defined anywhere, I'm finding).
We have a bogus index entry for "non-trivial constructor" (should have been "non-trivial default constructor," which already exists), and lack an entry (and `\defnx` italicization) for "trivial destructor".

FWIW I checked that [P3247 "Deprecate the notion of trivial types"](https://open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3247r0.html) didn't already touch either of these; I think this cleanup is pretty orthogonal to P3247.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -1208,7 +1208,6 @@ A constructor shall not have an explicit object parameter\iref{dcl.fct}.
 
 \pnum
 \indextext{constructor!inheritance of}%
-\indextext{constructor!non-trivial}%
 A \defnadj{default}{constructor} for a class \tcode{X}
 is a constructor of class \tcode{X}
 for which each parameter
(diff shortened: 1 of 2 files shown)


jensmaurer (approved):

Quuxplusone:
To be fair, [[special]](https://eel.is/c++draft/special) also says nothing about the word "trivial"; perhaps that crossref should _also_ be removed. Which would make it a little clearer that the notion of "trivial function" isn't defined anywhere.

PR #7000, opened 2024-05-14
[std] Fix colons in front of bulleted lists

### LINKED ISSUE #6999: [ISO/CS C++2023] Colons before bulleted lists
ISO/CS believes that a colon should precede a bulleted list.

We believe that bulleted lists are typographical whitespace that aids in conveying the structure of a sentence, but does not alter punctuation rules for the sentence.  That means a colon in front of a bulleted list is only acceptable if each bullet contains a complete sentence.

We should come to a decision and document it on the "style guidelines" wiki page.

jensmaurer:
Fixes #6999

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -4403,7 +4403,7 @@ otherwise, returns \tcode{false}.
 
 \pnum
 \complexity
-No applications of the corresponding predicate and projections if:
+No applications of the corresponding predicate and projections if
 \begin{itemize}
 \item
 for the first overload,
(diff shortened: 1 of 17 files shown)


tkoeppe:
Thank you, but we shan't need this as the status quo is acceptable.

jensmaurer:
@tkoeppe 

We do need this, because the status quo is inconsistent: Our rule is "bullets are whitespace", and this patch thus removes the colons where none are needed.  (ISO wants us to add more colons, which we don't want to do.)

tkoeppe:
Aha, OK, let me take a closer look, thanks!

PR #7002, opened 2024-05-14
[std] Replace 'through' with 'to' for clause ranges

### LINKED ISSUE #7001: [ISO/CS C++2023] Change 'through' to 'to'
- [x] 3 Terms and definitions edit "....in Clause 4 through to Clause 15..."
- [x] 4.1.1 General edit "....in Clause 4 to Clause 15..." This appears several times throughout the document - for clarity, please correct using "to" instead of "through". 
- [x] 16.3.3.4 Functions within classes See comments above - please change to "Clause 17 to Clause 33". Modify throughout the text.

jensmaurer:

Fixes #7001 

maybe NOT for the Working Draft


W-E-Brown:
I object to this change, as it in each case transforms (the equivalent of) a closed interval into (the equivalent of) a half-open interval.  "Because English" certainly is insufficient motivation, as we're (at best) introducing ambiguity where there was none. 

If we must s/through/to/ in these places, then we must also inject, post-positively, the adjective "inclusive".  I would withdraw my objection once that's done.

AlisdairM:
I concur with Walter, although if I squint I can read this text as intended.  Rather than add "inclusive" in each case though, I might add a clarifying sentence to the Introduction that says when we refer to a range of clauses, that range is inclusive.

tkoeppe:
Worry not, I expect this won't make it into the working draft.

tkoeppe:
I've asked if we can retain the status quo, and will close this until I get pushback.

tkoeppe:
This was "just a recommendation" and is not needed, and we shan't do it.

PR #7005, opened 2024-05-15
[input.output] Add cross-references to header synopses

jwakely:
Several of the synopses are not adjacent to the types they declare. Adding cross-references makes it easier to find the relevant definitions of classes and class templates.

Also make the title of [ostream.manip] more specific to its content.


jensmaurer (approved):

PR #7017, opened 2024-05-16
[namespace.future] Replace 'this International Standard' with 'this document'

jensmaurer:


jwakely (approved):
Sigh, this seems worse to me, but whatever.

tkoeppe:
I'll put this in the IS only.

PR #7018, opened 2024-05-16
Annex C fixes for '(ISO) C++ 20xx' and 'this revision of C++'

jensmaurer:
~Based on pull request #7007~

[macros,diff] Replace '(ISO) C++ 20xx' with the full document identifier

[diff] Replace 'this revision of C++' with 'this document'

DO NOT MERGE

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -1,14 +1,14 @@
 %!TEX root = std.tex
 \infannex{diff}{Compatibility}
 
-\rSec1[diff.cpp23]{\Cpp{} and ISO \CppXXIII{}}
+\rSec1[diff.cpp23]{\Cpp{} and \CppXXIII{}}
 
 \rSec2[diff.cpp23.general]{General}
 
 \pnum
-\indextext{summary!compatibility with ISO \CppXXIII{}}%
-Subclause \ref{diff.cpp23} lists the differences between \Cpp{} and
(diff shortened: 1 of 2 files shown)


source/compatibility.tex
```diff
@@ -151,8 +151,8 @@
   int tag;
 };
 
-static_assert(!allocator_traits<MyAlloc<int>>::is_always_equal);        // Error in \CppXXIII{},
-                                                                        // OK in \CppXXVI{}
+static_assert(!allocator_traits<MyAlloc<int>>::is_always_equal);    // Error in \CppXXIII{},
+                                                                    // OK in \CppXXVI{}
```
jwakely, line 155: CppCCVI still expands to "C++ 2026". Will they complain that it should be "this document"?

tkoeppe: Good point. Comparisons to the status quo are going to be less clear like tihs.

I'm inclined to not put this change in the WD. We can say "this document" in the IS.

jensmaurer: This entire hunk doesn't apply to C++23, the document currently with ISO/CS.

jwakely (approved):

tkoeppe:
Let's do this only for the IS. I'll put the other commits that are only for the IS in here, too.

tkoeppe:
Thanks, applied!

PR #7021, opened 2024-05-16
[lex.ccon,expr.prim.lambda.capture] Excise 'ISO' prefix

jensmaurer:

- [ ] What is ISO C++ 2014? Please use the ISO/IEC document identifier number


tkoeppe:
Hm, this isn't going to be good enough, since we also can't say "C++ 2014".

tkoeppe:
However, we can take this for the WD, and use the full document number in the IS.

PR #7024, opened 2024-05-16
[macros] Avoid small caps for cross-references to C.

jensmaurer:

- [ ] Please remove all caps. Change to "See also..." or make this NOTE 2, if it is necessary to use capital letters for this information.
- [ ] Remove all caps and make this a full sentence i.e."See ISO/IEC 9899...."
- [ ] Please remove all caps and change to a full sentence i.e. "See also ISO/IEC 9899...."
- [ ] remove all caps and change to a full sentence "See ISO/IEC ..."
- [ ] idem: change to "See also ISO/IEC 9899:2018, 7.8."
- [ ] Please modify to a full sentence with no all caps text i.e. "See also ECMA-262, 15.10."

### DIFF
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -371,8 +371,8 @@
 \newcommand{\templalias}{\Fundesc{Alias template}}
 
 %% Cross-reference
-\newcommand{\xref}{\textsc{See also:}\space}
-\newcommand{\xrefc}[1]{\xref{} \IsoC{}, #1}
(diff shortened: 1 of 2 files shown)


jensmaurer:
@tkoeppe , the first patch should be applied to the IS and the WD; the second patch only to the IS.

tkoeppe:
Thank you, done!

PR #7025, opened 2024-05-16
[fs.class.path.general] Defuse cross-reference to POSIX

jensmaurer:

- [ ] where is section 4.12? Is this in the POSIX document? If so, please be more specific i.e. "The pathname resolution mechanism is described in POSIX, section 4.12."

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -13627,7 +13627,8 @@ is the operating system dependent mechanism for resolving
 a pathname to a particular file in a file hierarchy. There may be multiple
 pathnames that resolve to the same file.
 \begin{example}
-POSIX specifies the mechanism in section 4.12, Pathname resolution.
+For POSIX-based operating systems,
+this mechanism is specified in POSIX, section 4.12, Pathname resolution.
 \end{example}
 
 \begin{codeblock}


source/iostreams.tex
```diff
@@ -13592,11 +13592,8 @@
 \pnum
 \defnx{Pathname resolution}{pathname resolution}
 is the operating system dependent mechanism for resolving
-a pathname to a particular file in a file hierarchy. There may be multiple
-pathnames that resolve to the same file.
-\begin{example}
-POSIX specifies the mechanism in section 4.12, Pathname resolution.
-\end{example}
+a pathname to a particular file in a file hierarchy (see POSIX, section 4.12).
```
jwakely, line 13595: This seems to suggest that POSIX defines the mechanism, but that's only true for POSIX-based operating systems. i.e., this seems to contradict the first half of the sentence.

```suggestion
a pathname to a particular file in a file hierarchy.
\begin{example}
For POSIX-based operating systems, this mechanism is specified in POSIX section 4.12).
\end{example}
```

I'm a little sad to lose the "Pathname Resolution" part, because while "POSIX 4.12" unambiguously means "ISO/IEC/IEEE 9945:2009 4.12" in this document, when POSIX 2024 gets published in the near future the correct section will be different (it's 4.16 in the current 202x draft 4 doc). If we also name the section then somebody consulting the latest POSIX spec will be able to find it more easily, without having to check what 4.12 in older spec corresponds to. I realise that checking the older spec is more correct, since that's our normative reference. But people often want to consult the latest spec, since that might be what their implementation's libc actually conforms to.

jensmaurer: Yeah, this is less of a change. Applied.

jwakely (approved):

PR #7026, opened 2024-05-16
[macros,numerics] Add and use numbered 'formula' environment

jensmaurer:

- [ ] This is a formula which should be labelled Formula (1).  See  HYPERLINK "https://www.iso.org/sites/directives/current/part2/index.xhtml#_idTextAnchor410"ISO/IEC Directives Part 2, 2021, 27.4  "Please number mathematical formulae in the document, for cross-referencing purposes. Arabic numbers in parentheses shall be used, starting with 1. If a mathematical formula is numbered, it should also be referred to in the text and its purpose should be made clear by its context, for example, with an introductory proposition, e.g.  "see 10.1, Formula (3)".
- [ ] see comment above regarding Formulae
- [ ] this appears to be a formulae - please follow the ISO/IEC Directives Part 2, Clause 27

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -4074,11 +4074,10 @@ Let
 \end{itemize}
 An \defn{attempt} is $k$ invocations of \tcode{g()}
 to obtain values $g_0, \dotsc, g_{k-1}$, respectively,
-and the calculation of a quantity
- \[
-   S = \sum_{i=0}^{k-1} (g_i - \tcode{g.min()})
-                        \cdot R^i
- \]
+and the calculation of a quantity $S$ given by \eqref{rand.gencanonical}:
+\begin{formula}{rand.gencanonical}
+S = \sum_{i=0}^{k-1} (g_i - \tcode{g.min()}) \cdot R^i
+\end{formula}
 
 \pnum
 \effects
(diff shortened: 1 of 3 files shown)


tkoeppe:
(Have you got a screenshot by any chance? :-) )

jensmaurer:
image

image

tkoeppe:
Looks great, thanks!

source/numerics.tex
```diff
@@ -9648,13 +9679,16 @@
 
 \pnum
 \returns
-\[ \mathsf{L}_n^m(x) =
-   (-1)^m \frac{\mathsf{d} ^ m}{\mathsf{d}x ^ m} \, \mathsf{L}_{n+m}(x)
-   \text{ ,\quad for $x \ge 0$,} \]
+The value given by \eqref{sf.cmath.assoc.laguerre},
```
tkoeppe, line 9682: Maybe we need to say more here. The formula isn't a "value" as such. Maybe "the value L_n^m(x), where L_n^m is given by Formula ..."?

jensmaurer: Maybe we can omit the "The value given by" and just say the function returns "L_n^m(x)` ..." and then the rest from your suggestion.

jensmaurer: Updated; please have a look.

jensmaurer:
image

jensmaurer:
When we do this, we might replace the "k is k" stuff by directly putting in the monospace-font C++ variables into the "value", e.g. "Pi(nu, k)".  What do you think?

tkoeppe:
> When we do this, we might replace the "k is k" stuff by directly putting in the monospace-font C++ variables into the "value", e.g. "Pi(nu, k)". What do you think?

This is something we've been wanting to do for a while, but this requires some thought. Let's do this for the WD only, and talk about it separately.

@jwakely ^^^ I think you had some thoughts here, too.

tkoeppe (approved):
This is great, thanks!

jwakely:
> @jwakely ^^^ I think you had some thoughts here, too.

More than just thoughts, see #4295

PR #7027, opened 2024-05-16
[futures.state] Turn note into example

jensmaurer:

- [ ] Please make this a full sentence, or incorporate this text into the sentence above. It does not make sense on its own. If it is giving an example, consider changing is to "EXAMPLE Promises and tasks"

### DIFF
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -10259,9 +10259,9 @@ An \defn{asynchronous provider} is an object that provides a result to a shared
 state.
 The result of a shared state is set by
 respective functions on the asynchronous provider.
-\begin{note}
-Such as promises or tasks.
-\end{note}
+\begin{example}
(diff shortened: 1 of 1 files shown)


Dani-Hub:
@jensmaurer  Should that say "providers" instead of "provides"?

jensmaurer:
Yeah, fixed.

source/threads.tex
```diff
@@ -10259,9 +10259,9 @@
 state.
 The result of a shared state is set by
 respective functions on the asynchronous provider.
-\begin{note}
-Such as promises or tasks.
-\end{note}
+\begin{example}
+Examples for asynchronous providers include promises or tasks.
```
jwakely, line 10263:
```suggestion
Promises and tasks are examples of asynchronous providers.
```

jwakely (approved):

PR #7030, opened 2024-05-17
[uaxid.general] Replace 'C++' with 'this document'

jensmaurer:

DO NOT MERGE INTO THE WORKING DRAFT

This feels not like an improvement to me.


tkoeppe:
I made a milder change to apply "this document" only in certain places. Let's see if that's enough. If not, then we'll take this change for only the IS.

PR #7031, opened 2024-05-17
[uaxid] Replace 'this requirement' with a specific reference

jensmaurer:

This seems ok for the WD, in the interest of avoiding further bifurcation.


tkoeppe:
Thanks, I rolled this into a patch that also changes C++ to "this document".

PR #7033, opened 2024-05-17
[lib] Excise Note A, Note B, etc. designations

jensmaurer:

- [ ] What is "Note A"? And likewise for the text below, what is "Note B?" If these are meant to be "notes" according to the ISO/IEC Directives Part 2, then please change them to "NOTE x" as appropriate
- [ ] see earlier comment: what is Note A and Note B?

### DIFF
--- a/source/meta.tex
+++ b/source/meta.tex
@@ -1934,8 +1934,8 @@ argument passing.
 
  \tcode{template<class... T>} \tcode{struct common_type;}
  &
- Unless this trait is specialized (as specified in Note B, below),
- the member \tcode{type} is defined or omitted as specified in Note A, below.
+ Unless this trait is specialized,
+ the member \tcode{type} is defined or omitted as specified below.
  If it is omitted, there shall be no member \tcode{type}.
  Each type in the template parameter pack \tcode{T} shall be
  complete, \cv{}~\keyword{void}, or an array of unknown bound. \\ \rowsep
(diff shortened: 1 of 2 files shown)


source/utilities.tex
```diff
@@ -2840,10 +2840,10 @@
     get(tuple<Types...>& t) noexcept;
 template<size_t I, class... Types>
   constexpr tuple_element_t<I, tuple<Types...>>&&
-    get(tuple<Types...>&& t) noexcept;        // Note A
+    get(tuple<Types...>&& t) noexcept;        // see Note 1
```
jwakely, line 2843: There's a small risk that we'll insert another note and "Note 1" would have to be changed to "Note 2" but that seems acceptable.

The notes are just restating how references work, so aren't adding anything normative. They're fine to become `note` environments.

source/meta.tex
```diff
@@ -2066,7 +2066,6 @@
 \tcode{\placeholdernc{COMMON-REF}(A, B)} is ill-formed.
 
 \pnum
-Note A:
```
jwakely, line 2069: I think these changes are definite improvements, I've wanted to be rid of these normative "notes" for a long  time.

jwakely (approved):

source/utilities.tex
tkoeppe: OK, sure. This is a bit of a novel thing to keep track of, but this seems mangeable.

Maybe an alternative would be to say something like `// #1`, and then say "For the overload marked `#1`" in the note. Thoughts? @jwakely, @jensmaurer

jwakely: Works for me

jwakely: We could even just change the notes to say "for the second overload" and "for the third overload"

jwakely: Or we could strike the notes, and add an example that demonstrates the properties explained in the notes. Let me know if you want me to prepare such an example for the WD. For the ISO C++23 doc, any solution is fine with me.

tkoeppe: Yes, please!

tkoeppe: I'll apply this PR to the WD in any case, but please do go ahead with a better change.

PR #7034, opened 2024-05-17
[intro.compliance.general] Refer to Annex B normatively

jensmaurer:

DO NOT MERGE INTO THE WORKING DRAFT

The numerical values of the limits are not normative (per the status quo); this change makes them appear more normative.  CWG2891 seeks to move the normative parts out of Annex B.


tkoeppe:
I'm sorry, I applied this change already --  I will revert that change as part of the motion application.

Issue #7035, opened 2024-05-19
[util.smartptr.shared.cast] "Will eventually" is so sure


zhihaoy:
There're four occurrences of "will eventually" in this section, for instance

> _[Note 1 :_ The seemingly equivalent expression `shared_ptr<T>(static_cast<T*>(r.get()))` will eventually
result in undefined behavior, attempting to delete the same object twice. _--end note]_

But what does it mean? If `r.get()` compares equal to `nullptr`, the program won't delete the same object twice. Moreover, in `dynamic_pointer_cast`'s case, the program may run "correctly" sometimes if the users handled the case when `dynamic_cast` gives `nullptr`...

Suggestion:

```diff
- will eventually result in undefined behavior,
+ can result in undefined behavior,
```

jwakely:
Yes, "can" seems better

PR #7037, opened 2024-05-19
[util.smartptr.shared.cast] Properly describe a bad outcome in notes

zhihaoy:
The notes may want to warn the users with the phrase" will eventually result in undefined behavior," but the saying is inaccurate.

Closes: cplusplus/draft#7035


jwakely (approved):

jensmaurer (approved):
ISO/CS can be excited.

PR #7044, opened 2024-06-04
[defns.erroneous], [basic.indet] Index 'erroneous behavior' and 'erroneous value'

### LINKED ISSUE #7043: [defns.erroneous] Erroneous behavior is not indexed in `/generalindex`
[defns.undefined] indexes "undefined behavior"/"behavior, undefined" as follows:
https://github.com/cplusplus/draft/blob/4b3f32ae814c8da3faccc0dc307904bd250371d9/source/intro.tex#L629-L632

[defns.erroneous] only contains a `\definition`:
https://github.com/cplusplus/draft/blob/4b3f32ae814c8da3faccc0dc307904bd250371d9/source/intro.tex#L251-L252

Eisenwave:
Fixes #7043.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3748,8 +3748,6 @@ thread without adequate synchronization.
 \rSec2[basic.indet]{Indeterminate and erroneous values}
 
 \pnum
-\indextext{value!indeterminate}%
-\indextext{indeterminate value}%
 When storage for an object with automatic or dynamic storage duration
 is obtained,
 the bytes comprising the storage for the object
@@ -3762,7 +3760,7 @@ whose first declaration is marked with
 the \tcode{[[indeterminate]]} attribute\iref{dcl.attr.indet},
 the bytes have \defnadjx{indeterminate}{values}{value};
 \item
-otherwise, the bytes have erroneous values,
+otherwise, the bytes have \defnadjx{erroneous}{values}{value},
 where each value is determined by the implementation
 independently of the state of the program.
 \end{itemize}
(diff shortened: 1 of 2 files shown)


source/basic.tex
```diff
@@ -3750,6 +3750,8 @@
 \pnum
 \indextext{value!indeterminate}%
 \indextext{indeterminate value}%
+\indextext{value!erroneous}%
```
jensmaurer, line 3753: I'm wondering whether we should make "erroneous value" a defined term and simply use \defnadjx.  That feels a lot more straightforward.

Eisenwave: I could do that too. If we're going to do that, I think we should make the same change for indeterminate values.

That's also why I've just perpetuated whatever is there right now; perhaps the conversion of both terms to `\denfadjx` could be done in a follow-up PR.

Eisenwave: Alright, so I've noticed that "indeterminate value" has both the `\defnadjx` definition and the `\indextext` up there.

Now, it's the same for erroneous values. Is there a reason why indeterminate values have both? Can/should we just get rid of all those `\indextext` entries?

Eisenwave: When removing `\indextext`, all index entries still exist, but we lose the link within the PDF and the page number.
image

With `\indextext`, there's a `68` at the end.

jensmaurer: Yup, that's intentional: we want the "see" entry, not a direct page number.

jensmaurer: So, please remove \indextext and use \defnadjx

jensmaurer (approved):

PR #7046, opened 2024-06-04
[basic.indet] Convert reference to [conv.lval] into note

### LINKED ISSUE #7045: [conv.lval], [basic.indet] Circular cross-referencing for "erroneous value

Eisenwave:
Fixes #7045.

It is unclear what the purpose of `erroneous value~\iref(conv.lval}` is here. [conv.lval| is not the subclause which defines "erroneous value" but this cross-reference makes it look like it is. This is especially nonsensical given that [conv.lval] cross-references back to [basic.indet], making this a circular reference.

I assume that the intent here is to draw some extra attention to lvalue-to-rvalue conversion for specific erroneous values. Therefore, I have turned this into a note.

Alternatively, we could also eliminate the `\iref` completely and leave everything as is otherwise.


tkoeppe:
Originally (as per P2795R5) the reference was indeed meant to refer to the potential UB is the erroneous value is not valid. However, that wording was changed by https://cplusplus.github.io/CWG/issues/2899.html, and now the connection isn't as clear anymore.

I agree that the note is helpful in making that connection. @jensmaurer?

source/basic.tex
```diff
@@ -3772,11 +3772,15 @@
 If any bit in the value representation has an indeterminate value,
 the object has an indeterminate value;
 otherwise, if any bit in the value representation has an erroneous value,
-the object has an erroneous value\iref{conv.lval}.
+the object has an erroneous value.
 \begin{note}
 Objects with static or thread storage duration are zero-initialized,
 see~\ref{basic.start.static}.
 \end{note}
+\begin{note}
```
jensmaurer, line 3780: Please move the note to the first place in the list of notes here.  The "static or thread" note refers to something entirely different.

source/basic.tex
```diff
@@ -3772,11 +3772,15 @@
...
+The behavior of lvalue-to-rvalue conversion\iref{conv.lval} is undefined
+if the erroneous value of an object is not valid for its type.
```
jensmaurer: I think I want the cross-reference at the end of that sentence, because conv.lval discusses not just lvalue-to-rvalue conversion, but also the "bad" situation the second part of the sentence talks about.

jensmaurer: And can we just say "Lvalue-to-rvalue conversion has undefined behavior if ..." ?  Seems more straightforward.

PR #7049, opened 2024-06-05
[conv.lval] Add example of erroneous 'trap representation' being read

### LINKED ISSUE #7047: [conv.lval] Add example of indeterminate values that are not valid for the type
[[conv.lval] p3.4 sentence 2](https://eel.is/c++draft/conv.lval#3.4.sentence-2) says:

> If the result is an erroneous value ([basic.indet]) and the bits in the value representation are not valid for the object's type, the behavior is undefined.

It's not obvious how you would run into this case given that generally, memory is initialized for erroneous values, and this initialization can typically done so that values are valid.

To be honest, I don't understand how you can run into this case, and it's not going to be obvious to others who follow, so we should add an example of how this can happen.

Eisenwave:
Fixes #7047.

image

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -678,6 +678,20 @@ and the value contained in the object is the prvalue result.
 If the result is an erroneous value\iref{basic.indet} and
 the bits in the value representation are not valid for the object's type,
 the behavior is undefined.
+\begin{example}
+\begin{codeblock}
+int main() {
+  bool a;
+  bool b = a;
+}
+\end{codeblock}
+If the values of the bytes of the object representation of \tcode{a}
+are determined by the implementation so that the
+bits of the value representation of \tcode{a}
+correspond to \tcode{true} or \tcode{false},
+the lvalue-to-rvalue conversion during the initialization of \tcode{b}
+has erroneous behavior, and undefined behavior otherwise.
+\end{example}
 \end{itemize}
 
 \pnum


Eisenwave:
Now presumably based on CWG2899.

If CWG2899 is accepted, this example would need to be shifted up by one bullet, and it may make sense to extend the example to show how this happens without erroneous values, like in the issue.

jensmaurer:
I take this as a request to add an example to CWG2899. Done.

Eisenwave:
@jensmaurer thanks, it makes sense to add it as an example there.

However, the reason why I've originally suggested the example was that it wasn't obvious how erroneous values can trigger this form of UB, so ideally, CWG2899 should show both scenarios:
```diff
bool f() {
+  bool e;            // e has erroneous value ([basic.indet])
+  bool u = e;        // undefined behavior if the value of e is not a valid value representation for bool

  bool b = true;
  char c = 42;
  memcpy(&b, &c, 1);  // implicitly creates objects at b (23.5.3 [[cstring.syn](https://wg21.link/cstring.syn)], 6.7.2 [[intro.object](https://wg21.link/intro.object)])
  return b;           // undefined behavior if 42 is not a valid value representation for bool
}
```

eisenwave:
I can live with what was added by CWG2899. There are bigger fish to fry.

PR #7050, opened 2024-06-05
[expr.static.cast], [over.call.object] Replace 'greater cv-qualification' with 'more cv-qualified'

### LINKED ISSUE #7048: [expr.static.cast], [basic.type.qualifier] "More cv-qualified" vs "greater cv-qualification"
There is a partial ordering of cv-qualifiers, and the defined term we have is [more cv-qualified](https://eel.is/c++draft/basic.type.qualifier#def:more_cv-qualified).

However, a number of places use a different term in reference; namely "greater cv-qualification" (e.g. https://eel.is/c++draft/expr.static.cast#14). We should be consistent about how we talk about the ordering of cv-qualifiers.

We could also primarily define everything in terms of cv-qualifier ordering, and say that:
> A type *cv1* `T1` is more cv-qualified than type *cv2* `T2` if *cv1* is greater than *cv2*.

eisenwave:
Fixes https://github.com/cplusplus/draft/issues/7048.

Generally, the strategy is to replace "same cv-qualification, or greater cv-qualification" with "at least as cv-qualified", which is much shorter and more natural. "At least as" is arguably not a defined term (unlike "more cv-qualified"), but I believe it naturally stems from the defined term. If we really wanted to, we could defined "at least as cv-qualified" and "more cv-qualified" more strictly, but I don't believe it's entirely necessary to do so.

Note that there is precedent for "at least as cv-qualified" in https://eel.is/c++draft/expr.cond#4.3.1

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -4049,9 +4049,8 @@ away constness\iref{expr.const.cast}.
 An lvalue of type ``\cvqual{cv1} \tcode{B}'', where \tcode{B} is a class
 type, can be cast to type ``reference to \cvqual{cv2} \tcode{D}'', where
 \tcode{D} is a complete class derived\iref{class.derived} from \tcode{B},
-if \cvqual{cv2} is the
-same cv-qualification as, or greater cv-qualification than,
-\cvqual{cv1}. If \tcode{B} is a virtual base class of \tcode{D}
+if \cvqual{cv2}~\tcode{D} is equally as or more cv-qualified than \cvqual{cv1}~\tcode{B}.
+If \tcode{B} is a virtual base class of \tcode{D}
 or a base class of a virtual base class of \tcode{D},
 or if no valid standard conversion from ``pointer to \tcode{D}''
 to ``pointer to \tcode{B}'' exists\iref{conv.ptr}, the program is ill-formed.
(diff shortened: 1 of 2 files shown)


source/overloading.tex
```diff
@@ -546,10 +546,9 @@
 \begin{ncsimplebnf}
 \keyword{operator} conversion-type-id \terminal{(\,)} \opt{cv-qualifier-seq} \opt{ref-qualifier} \opt{noexcept-specifier} \opt{attribute-specifier-seq} \terminal{;}
 \end{ncsimplebnf}
-where the optional
-\grammarterm{cv-qualifier-seq}
-is the same cv-qualification as, or a greater cv-qualification than,
-\cv{},
+where if \grammarterm{cv-qualifier-seq} is provided,
+\grammarterm{cv-qualifier-seq}~\grammarterm{conversion-type-id}
+at least as cv-qualified as \cv{}~\tcode{T},
```
jensmaurer, line 551: missing "is"

source/expressions.tex
```diff
@@ -4049,9 +4049,8 @@
 An lvalue of type ``\cvqual{cv1} \tcode{B}'', where \tcode{B} is a class
 type, can be cast to type ``reference to \cvqual{cv2} \tcode{D}'', where
 \tcode{D} is a complete class derived\iref{class.derived} from \tcode{B},
-if \cvqual{cv2} is the
-same cv-qualification as, or greater cv-qualification than,
-\cvqual{cv1}. If \tcode{B} is a virtual base class of \tcode{D}
+if \cvqual{cv2}~\tcode{D} is at least as cv-qualified as \cvqual{cv1}~\tcode{B}.
```
AlisdairM, line 4052: I am concerned that `const int` "is at least as cv-qualified" as `volatile int`, but not "the same or greater".

jensmaurer: Agreed.  We only have a partial order here.

Eisenwave: The PR now uses the wording "equally as or more cv-qualified than" (also fixes [expr.cond]), which should be unambiguous even with only a partial order.

AlisdairM: Progress, but I remain concerned that "equally as cv-qualified" does not mean "has the same cv-qualification" due to our partial ordering.

PR #7053, opened 2024-06-06
[dcl.array] No longer explain array subscript in terms of array-to-pointer conversion

Eisenwave:
Fixes https://github.com/cplusplus/draft/issues/7052.

image

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -3452,25 +3452,17 @@ can reasonably appear in an expression.
 The expression
 \tcode{x3d[i]}
 is equivalent to
-\tcode{*(x3d + i)};
-in that expression,
-\tcode{x3d}
-is subject to the array-to-pointer conversion\iref{conv.array}
-and is first converted to
-a pointer to a 2-dimensional
-array with rank
-$5 \times 7$
-that points to the first element of \tcode{x3d}.
-Then \tcode{i} is added,
-which on typical implementations involves multiplying
-\tcode{i} by the
-length of the object to which the pointer points,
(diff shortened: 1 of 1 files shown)


jensmaurer:
I'm not seeing a problem with the original formulation. 

> The expression x3d[i] is equivalent to *(x3d + i); in that expression, x3d is subject to the array-to-pointer conversion...

It's certainly true that in the expression "*(x3d + i)", x3d is subject to the array-to-pointer conversion.  A small improvement might be to add a cross-reference to [expr.sub] after "equivalent to blah", to get the caveats visible.

Eisenwave:
> I'm not seeing a problem with the original formulation.

I think you're right, at least with the current wording. If we say that it's *equivalent* to `*(x3d + i)`, then array-to-pointer conversion really takes places.

If CWG2548 had been accepted, it would definitely be wrong though, as the wording would be

> The **result** of the expression `E1[E2]` is identical (by definition) to that of `*((E1)+(E2))`

If only the result, but not the remaining semantics are equivalent, then array-to-pointer conversion doesn't really take place.

Even though you're right right now, I still don't favor the current wording. The note is intended to explain multi-dimensional subscript math, and this side tangent on array-to-pointer conversion isn't making it easier to understand. Depending on how [expr.sub] changes (and there have been a fair amount of historical/proposed changes), this side tangent becomes less relevant or simply wrong.

jensmaurer:
In short, the status quo of this pull request is not what we want.

Eisenwave:
I'm closing this because [dcl.array] already has a reference to [expr.sub], so neither the proposed changes nor the possible small alternative can be completed here.

PR #7055, opened 2024-06-10
[expr.new] Extend example for new-expressions with zero size arrays

Eisenwave:
Someone has told me that they've been reading [[expr.new] p6](http://eel.is/c++draft/expr.new#6) wrong; namely in a way that implies `new int[0]` is ill-formed.

It's not immediately obvious that it is okay even though the wording says that something shall not be zero.

image

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -5285,6 +5285,15 @@ Given the definition \tcode{int n = 42},
 \grammarterm{expression} of a \grammarterm{noptr-new-declarator}), but
 \tcode{new float[5][n]} is ill-formed (because \tcode{n} is not a
 constant expression).
+Furthermore,
+\tcode{new float[0]} is well-formed
+(because \tcode{0} is the \grammarterm{expression}
+of a \grammarterm{noptr-new-declarator},
+where a value of zero results in the allocation of an array with no elements),
+but \tcode{new float[n][0]} is ill-formed
+(because \tcode{0} is the \grammarterm{constant-expression}
+of a \grammarterm{noptr-new-declarator},
+where only values greater than zero are allowed).
 \end{example}
 
 \pnum


source/expressions.tex
```diff
@@ -5285,6 +5285,11 @@
 \grammarterm{expression} of a \grammarterm{noptr-new-declarator}), but
 \tcode{new float[5][n]} is ill-formed (because \tcode{n} is not a
 constant expression).
+Furthermore,
+\tcode{new float[0]} is well-formed
+(because \tcode{0} is an \grammarterm{expression}), but
```
jensmaurer, line 5290: please use the same phrasing as the other sentence: "because 0 is the expression of a noptr-new-declarator"

source/expressions.tex
```diff
@@ -5285,6 +5285,11 @@
...
+\tcode{new float[n][0]} is ill-formed
+(because \tcode{0} is a \grammarterm{constant-expression}).
```
jensmaurer, line 5292: same here, and highlight that it's the value that's troublesome

jensmaurer (changes requested):

source/expressions.tex
```diff
@@ -5285,6 +5285,15 @@
 \grammarterm{expression} of a \grammarterm{noptr-new-declarator}), but
 \tcode{new float[5][n]} is ill-formed (because \tcode{n} is not a
 constant expression).
+Furthermore,
+\tcode{new float[0]} is well-formed
+(because \tcode{0} is the \grammarterm{expression}
+of a \grammarterm{noptr-new-declarator},
+where a value of zero results in the allocation of an array with no elements), but
```
jensmaurer, line 5292: Can we get semantic linebreaks and move ", but" to the next line?

tkoeppe:
@jensmaurer could you have another look?

jensmaurer (approved):

tkoeppe:
This is nice, thanks!

PR #7066, opened 2024-06-16
[implimits] Reorder Annex B by clause number

AlisdairM:
Reorder the entries in the list of implementation limits according to their clause number.  This ordering neatly groups related items such as all the preprocessor limits, and segregates library from core language limits.  It loses some of the local organisation of the original ordering, such as all parameter and argument list lengths being colocated.

A second commit splits lines that defined several quantities into separate lines that can then be correctly ordered.  The secondary benefit is that these quantities may now be updated independently.  The new risk is that these quantities that were previously guaranteed to be the same can now vary in a future standard.

### DIFF
--- a/source/limits.tex
+++ b/source/limits.tex
@@ -10,23 +10,19 @@ a potential minimum value for that quantity.
 \item%
 Characters in one logical source line\iref{lex.phases} [65\,536].
 \item%
-Number of
-characters in an internal identifier\iref{lex.name}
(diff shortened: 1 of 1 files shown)


jensmaurer:
CWG should have a say on whether they like it or not.

jensmaurer:
CWG 2025-11-08 in Kona: No objections to the reordering.

PR #7067, opened 2024-06-17
[lex.charset] Extract universal-character-name grammar to new subclause

AlisdairM:
The grammar for universal-character-name is oddly sandwiched into the middle of the subcluase talking about the different character sets used by the standard.  To improve the flow, extract that grammar into its own subclause.

In the extraction, I make two other clarifying changes.  First, describe this new subclause as 'a way to name any element of the of the tranlation character set using just the basic character set' rather than simply 'a way to name other characters'.  Secondly, remove the 'one of' in the grammar where there is only one option to choose.

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -249,7 +249,9 @@ needed for execution in its execution environment.%
 \indextext{translation!phases|)}
 \end{enumerate}
 
-\rSec1[lex.charset]{Character sets}
+\rSec1[lex.char]{Characters}%
+
+\rSec2[lex.charset]{Character sets}
 
 \pnum
 \indextext{character set|(}%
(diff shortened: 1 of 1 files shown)


source/lex.tex
```diff
@@ -358,6 +416,10 @@
     named-universal-character
 \end{bnf}
 
+\pnum
+The \grammarterm{universal-character-name} construct provides a way to name
+any element in the translation character set using just the basic character set.
```
jensmaurer, line 421: That rephrasing is not correct: You can't use a UCN to name a basic character outside of a string-literal.

source/lex.tex
```diff
@@ -320,11 +320,69 @@
...
+\indextext{UTF-32}%
+For a UTF-8, UTF-16, or UTF-32 literal,
+the implementation shall encode
+the Unicode scalar value
+corresponding to each character of the translation character set
+as specified in the Unicode Standard
+for the respective Unicode encoding form.
+\indextext{character set|)}
+
+\rSec1[lex.universal.char]{Universal Character Names}
```
jensmaurer, line 382: Hm... That makes another, rather small, top-level subclause.   Maybe have a top-level subclause "characters" with "character sets" and "universal character names" as sub-subclauses?

Also, we don't use title case, but sentence case, in subclause headings.

jensmaurer (changes requested):

source/lex.tex
AlisdairM: I tested the following program to declare the identifier pi as a variable, and it was accepted by all 4 major front ends:
```
double \u03c0 = 3.14;
```

Does this compile for some other reason?  Or are all 4 implementations buggy in allowing UCNs outside of character and string literals?

https://godbolt.org/z/hnbWr9qeo

source/lex.tex
AlisdairM: Fixed.

source/lex.tex
jensmaurer: I said "basic character". You're not naming one of the basic characters.
Try naming "A" with a UCN outside of a literal.

AlisdairM: In that case, I still think the text is good as written.  This is an introductory sentence, not a definition.  As far as I can tell it is still possible to name elements of the basic character set, but we limit where those particular names can be used to character and string literals --- it is not forming the name that fails, but how we use it after.  That all seems reasonable for the introductory paragraph.

jensmaurer (changes requested):
I'm not yet sure whether we'll squash some of this when merging, but please fix the commit description "Introduced parent [lax.char] clause for character sets and UCNs".
Make it imperative mood ("Introduce") and fix typo "lax" in section label.

AlisdairM:
Squashed all three commits into one, and update the commit message as requested.

jensmaurer:
The commit message still says [lax.char].  Note the laxative there.

jensmaurer:
@tkoeppe, I think this is reasonable and should go in. Feel free to fix the commit message when merging.

AlisdairM:
Sorry, force-pushed a fix to the commit message; no other changes

jensmaurer (approved):

tkoeppe:
@jensmaurer Are you OK with striking the "one of" from the grammar?

jensmaurer:
@tkoeppe , yes. "One of any member..." makes no sense.

PR #7071, opened 2024-06-19
[func.require] Add missing formatting of variable index

j6t:
In paragraph [func.require#1](https://eel.is/c++draft/func.require#1), all variables use subscripted indices in text that is formatted as code, except in two cases. These look like an unintended omission. Adjust them to subscripts.


jwakely (approved):
Well spotted, thanks

jwakely:
FWIW, I made them use the correct font in 367dfefdd7e032f1f7dd8414aeb2b28bd7fb773c but that got reverted in f8a9adc76467258ca4f2188c008ef2bada17bb9c

jensmaurer (approved):

PR #7117, opened 2024-07-03
[istream.unformatted] add missing semi-colon to list item

jwakely:


jensmaurer (approved):
It's consistent; I'm not sure I like the thrust.

jwakely:
Yeah it's not great, but we either needed this change, or to remove the semi-colon from the second bullet (and then make them full sentences?)

PR #7122, opened 2024-07-08
[basic.pre] Defragment specification of names and entities

AlisdairM:
The current contents of [basic.pre] jump between specifying different things.  This PR moves all the specification of names to the front, followed by the specification of entities.

There are two main benefits: (1) the specification for when two names are the same is a list of 4 rules that correspond to the 4 things than can form a name --- the connection is much clearer when the paragraphs are adjacent and the list is sorted to the same order; (2) in this form, even though all the words are the same, the reordering and merging of paragraphs a fit on a single page.  The very last paragraph was forced over a page-break in the original layout.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -32,17 +32,21 @@ Clauses.
 \indextext{storage class}%
 \indextext{scope}%
 \indextext{linkage}%
-An \defn{entity} is a value, object, reference,
-structured binding,
-function, enumerator, type,
-class member, bit-field, template, template specialization, namespace, or
-pack.
-
-\pnum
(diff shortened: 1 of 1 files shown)


AlisdairM:
Re-applied edits on the latest working draft, to ensure that any wording updates in this subclause are preserved.  Requested rebasing complete.

jensmaurer (approved):
The rationale "fits on one page" is of course bogus, because CWG will update this as it sees fit.

I think this is an improvement; I can't see the potential for harm given that text is moved around locally a bit.

@tkoeppe?

PR #7125, opened 2024-07-09
[classes] Turn ad-hoc examples into proper examples

tkoeppe:
Also addresses an ISO comment about not having "For example," as the last part of the normal sentence.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -950,7 +950,7 @@ definition, if they are defined at all.
 A member function can be declared (but not defined) using a typedef for
 a function type. The resulting member function has exactly the same type
 as it would have if the function declarator were provided explicitly,
-see~\ref{dcl.fct}. Also see~\ref{temp.arg}.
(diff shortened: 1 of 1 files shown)


source/classes.tex
```diff
@@ -948,8 +950,8 @@
 A member function can be declared (but not defined) using a typedef for
 a function type. The resulting member function has exactly the same type
 as it would have if the function declarator were provided explicitly,
-see~\ref{dcl.fct}. For example,
-
+see~\ref{dcl.fct}. Also see~\ref{temp.arg}.
```
jensmaurer, line 953:
```suggestion
see~\ref{dcl.fct} and \ref{temp.arg}.
```

jensmaurer (changes requested):
mostly straightforward

source/classes.tex
tkoeppe: Yes, nice.

jensmaurer (approved):

PR #7129, opened 2024-07-10
[coro.generator] Rename the `generator`'s template parameter `V` to `Val`

### LINKED ISSUE #6451: [coro.generator.class] Name the generator's second template parameter Val?
The current synopsis of [`std::generator`](https://eel.is/c++draft/coro.generator.class#lib:generator) is:

```
namespace std {
  template<class Ref, class V = void, class Allocator = void>
  class generator : public ranges::view_interface<generator<Ref, V, Allocator>> {
  // ...
```

When I first saw the naming of the second template parameter, I didn't quite understand what the `V` here was referring to and it seemed to be just an abbreviation for` void`. After I looked at it a few more times, I finally realized that it is actually the corresponding value type relative to the first template parameter `Ref`.

The `class V = void` here seems to be so simplified that it even made me misunderstand that it is just a common template name like `U` and has no meaning. It may be more friendly to change it to `class Val = void`. 

Comments are welcome.

hewillk:
Fixes #6451.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -16773,12 +16773,12 @@ void f() {
 \begin{codeblock}
 namespace std {
   // \ref{coro.generator.class}, class template \tcode{generator}
-  template<class Ref, class V = void, class Allocator = void>
+  template<class Ref, class Val = void, class Allocator = void>
     class generator;
 
   namespace pmr {
-    template<class R, class V = void>
-      using generator = std::generator<R, V, polymorphic_allocator<>>;
+    template<class Ref, class Val = void>
+      using generator = std::generator<Ref, Val, polymorphic_allocator<>>;
   }
 }
 \end{codeblock}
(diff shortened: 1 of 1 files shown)


jensmaurer:
Names of template parameters rarely convey lots of meaning, it seems to me.  So, this change feels rather undermotivated to me.

@jwakely ?

jwakely (approved):
While I agree that template parameter names aren't always meaningful, for `generator` we have `Ref` and `V` not just `T` and `U` or `T1` and `T2`, so there's already some attempt at making them meaningful.

And I think there's also scope for confusion due to `class V = void` and `is_void_v<V>` making it seem like V is for void.

I think this change does give some positive value (no pun intended).

jwakely:
> And I think there's also scope for confusion due to `class V = void` and `is_void_v<V>` making it seem like V is for void.

Ha, just realised I've repeated the argument given in #6451, so I guess that means I agree with it.

jensmaurer (approved):
Let's do that, then.

PR #7136, opened 2024-07-16
[deque, forward.list, list, vector] Fix instances of "FooInsertable into *this"

### LINKED ISSUE #7133: Bad phrasing "FooInsertable into *this"
We have several instances of the phrasing "`T` is FooInsertable into `*this`", which is non-sensical since "`FooInsertable` into `X`" is only defined [container.alloc.reqmts] when `X` is a type.

We can either fix this by changing `*this` to the relevant type name, or by adding blanket wording that whenever `X` is an expression, we mean the type of the expression.

Quuxplusone:
Fixes #7133

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -6716,7 +6716,7 @@ explicit deque(size_type n, const Allocator& = Allocator());
 \begin{itemdescr}
 \pnum
 \expects
-\tcode{T} is \oldconcept{DefaultInsertable} into \tcode{*this}.
+\tcode{T} is \oldconcept{DefaultInsertable} into \tcode{deque}.
 
 \pnum
 \effects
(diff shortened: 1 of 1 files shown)


source/containers.tex
```diff
@@ -6716,7 +6716,7 @@
 \begin{itemdescr}
 \pnum
 \expects
-\tcode{T} is \oldconcept{DefaultInsertable} into \tcode{*this}.
+\tcode{T} is \oldconcept{DefaultInsertable} into \tcode{deque}.
```
Quuxplusone, line 6719: Should I change `T` to `value_type` in each case as well?

jwakely: I don't think that's necessary (and would require changes in places that this PR doesn't already touch).

PR #7139, opened 2024-07-17
[list] [vector] [inplace.vector] Consistent comma in "If X, there are no effects"

Quuxplusone:
There is exactly one place in containers.tex where we have "If X, *then* there are no effects." But the "X" in that case is very convoluted, so I think the "then" is kind of helpful.

Notice the several novel instances of "there are no effects *on `*this`*" in [inplace.vector]. I want to fix those too, for consistency with [vector] and [list]; but I think that probably requires a formal LWG issue.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -6912,7 +6912,7 @@ and causes a single call to a constructor of
 If an exception is thrown other than by the
 copy constructor, move constructor,
 assignment operator, or move assignment operator of
-\tcode{T}
+\tcode{T},
 there are no effects.
 If an exception is thrown while inserting a single element at either end,
 there are no effects.
@@ -8271,7 +8271,7 @@ to the number of elements inserted.
 \pnum
 \remarks
 Does not affect the validity of iterators and references.
-If an exception is thrown there are no effects.
+If an exception is thrown, there are no effects.
 \end{itemdescr}
 
 \indexlibrarymember{erase}{list}%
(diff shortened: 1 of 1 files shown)


jensmaurer (approved):

jwakely (approved):

jwakely:
"no effects on `*this`" seems preferable though. We can't say that inserting elements has no effects, because the element constructor might log before it throws, or other observable side effects. What we mean is that it doesn't change the container's size or contents. We should fix all other places that claim to have no effects period, because they're impossible to implement.

Quuxplusone:
(Purely a tangent, not relevant to this PR which should be merged AFAIC...)

> "no effects on `*this`" seems preferable though. We can't say that inserting elements has no effects, because the element constructor might log before it throws, or other observable side effects. What we mean is that it doesn't change the container's size or contents.

...or capacity, I think.
Obviously "no effects" is _physically_ inaccurate, because if nothing else we'll spend some extra nanoseconds and generate some extra waste heat (and, like, throw an exception, which is a major "effect" in itself). But I think it's _semantically_ accurate, assuming (as the STL always does) that `T`'s constructor does nothing but construct, `T`'s destructor does nothing but destroy, and so on. If we construct a copy of a value and then destroy that copy again, that's _semantically_ "no effect" on the entire universe, not merely on `*this`.

Vice versa, of course it's _physically_ possible to write a type such that constructing an element of that type _does_ have an observable effect on `*this` ([Godbolt](https://godbolt.org/z/bsMW8PE9r)):
```
struct S {
    int i_ = 42;
    explicit S() = default;
    explicit S(int *p) { *p = 99; throw "oops"; }
};
~~~~
    assert(v.front().i_ == 42);
    try {
        v.emplace_back(&v[0].i_);
    } catch (...) {}
    assert(v.front().i_ == 99);
```
But there's some philosophical sense in which the STL doesn't expect user-programmers to do that. (And not just _that_, but _anything_ "like that" — anything where the constructor does more than construct, or the destructor does more than destroy, or the assignment operator does more than assign, or the comparison does more than compare...)

Quuxplusone:
Gentle ping: someone can merge this, right?

PR #7140, opened 2024-07-18
[alg.search] Replace "the following corresponding conditions"

jwakely:
It's not clear which conditions correspond to each overload.

Let me know if you'd prefer this to be done as an LWG issue.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -4551,11 +4551,15 @@ The type \tcode{Size}
 is convertible to an integral type\iref{conv.integral,class.conv}.
 
 \pnum
+Let $E$ be \tcode{pred(*(i + n), value) != false}
+for the overloads with a parameter \tcode{pred},
+and \tcode{*(i + n) == value} otherwise.
+
+\pnum
 \returns
 The first iterator \tcode{i} in the range \range{first}{last-count}
 such that for every non-negative integer \tcode{n} less than \tcode{count}
-the following corresponding conditions hold:
-\tcode{*(i + n) == value, pred(*(i + n), value) != false}.
+the condition $E$ is true.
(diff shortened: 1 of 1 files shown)


jensmaurer (approved):
Looks good and in line with phrasing elsewhere.

source/algorithms.tex
```diff
@@ -4550,12 +4550,16 @@
 The type \tcode{Size}
 is convertible to an integral type\iref{conv.integral,class.conv}.
 
+\pnum
+Let $E$ be \tcode{pred(*(i + n), value)}
```
jensmaurer, line 4554: We have `!= false` in the existing text. That does not matter if `pred` is required to return `bool`, but is it?

jwakely: It's required to model _`boolean-testable`_ which means that `!= false` is already wrong.

This is the subject of the recently opened [LWG 4127](https://cplusplus.github.io/LWG/issue4127). I can incorporate this change into the proposed resolution of that issue if you prefer.

jwakely: The LWG issue was only opened after I created this pull request (possibly as a result of seeing it). Since we now have an LWG issue touching that text, maybe we shouldn't change it editorially.

jensmaurer: side note: LWG4127 should use "is true" everywhere.

I think I'm good with applying the fix here editorially, although I'd prefer to retain the `!= false` in the editorial rewrite and fix that part in the LWG issue. But we don't have to burden the LWG issue with editorial rewrites such as this one.

jwakely: OK, I've pushed a change to restore the `!= false`

jensmaurer (approved):

PR #7150, opened 2024-07-22
[inplace.vector.erasure] Added missing return statement

hewillk:


jwakely (approved):
This was present in the incoming paper and missed from the PR that added it to the draft.

PR #7152, opened 2024-07-22
[rand.eng.philox] Make the round states explicit.

tkoeppe:
This clarifies which state is the final result, and avoids the use of the vaguely defined variable $X'$. It changes the index variable $q$ to be 1-based. The single sequence $V$ is replaced with the sequence of sequences $V^{(q)}$.

Currently, the definition of the "r-round network", "rounds", and how they fit together, is somewhat informal and imprecise. In particular, the statement that `Philox` "returns the sequence $Y = X'$" is needlessly ambiguous.

It would be clearer if all involved mathematical entities and operations were clearly spelled out.

---

**Preview:**

image

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -3186,27 +3186,24 @@ The state transition is performed as if by the following algorithm:
 \begin{codeblock}
 @$i$@ = @$i$@ + 1
 if (@$i$@ == @$n$@) {
-  @$Y$@ = Philox(@$K$@, @$X$@) // \seebelow
-  @$Z$@ = @$Z$@ + 1
+  @$Y$@ = Philox(@$K$@, @$X$@)        // \seebelow
+  @$Z$@ = @$Z$@ + 1                // this updates $X$
   @$i$@ = 0
 }
 \end{codeblock}
 
 \pnum
 The \tcode{Philox} function maps the length-$n/2$ sequence $K$ and
-the length-$n$ sequence $X$ into a length-$n$ output sequence $Y$.
+the length-$n$ sequence $X$ into a length-$n$ output sequence.
 Philox applies an $r$-round substitution-permutation network to the values in $X$.
-A single round of the generation algorithm performs the following steps:
+That is, there are intermediate values $X^{(0)}, X^{(1)}, \dotsc, X^{(r)}$,
+where $X^{(0)} \cedef X$, and for each round $q$ (with $q = 1, \dotsc, r$),
+$X^{(q)}$ is computed from $X^{(q - 1)}$ as follows. The output sequence is $X^{(r)}$.
 \begin{itemize}
 \item
-The output sequence $X'$ of the previous round
-($X$ in case of the first round)
-is permuted to obtain the intermediate state $V$:
-\begin{codeblock}
-@$V_j = X'_{f_n(j)}$@
-\end{codeblock}
-where $j = 0, \dotsc, n - 1$ and
-$f_n(j)$ is defined in \tref{rand.eng.philox.f}.
+An intermediate state $V^{(q)}$ is obtained by permuting the previous output,
+$V^{(q)}_j \cedef X^{(q - 1)}_{f_n(j)}$,
+where $j = 0, \dotsc, n - 1$, and $f_n(j)$ is defined in \tref{rand.eng.philox.f}.
 
 \begin{floattable}{Values for the word permutation $\bm{f}_{\bm{n}}\bm{(j)}$}{rand.eng.philox.f}
 {l|l|l|l|l|l}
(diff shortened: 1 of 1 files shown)


tkoeppe:
@villevoutilainen, @iburyl, PTAL?

villevoutilainen:
This looks so non-editorial to me that it isn't even funny. :D

tkoeppe:
@villevoutilainen: Hm :-) Do you think this changes the normative content? Which part specifically?

villevoutilainen:
@tkoeppe I don't want to figure out whether it does change the normative content, and if so, how. I want LWG to do that.

tkoeppe:
Got it, thanks!

iburyl:
My problem with this change is that X is now being used 
1) without indices, 
2) with upper indices, 
3) with lower indices and 
4) with upper and lower indices.
Also not sure why q is being used with parenthesis for V and X and without for key.
While rounds are defined a tiny bit clearer, but it adds quite a bit of complexity to X definition.
Given the wording was not really broken before the changes, I am not sure that it is worth it.

villevoutilainen:
Right. The change is a spec revamp. While I have the utmost confidence that @tkoeppe can do it correctly, this sort of changes are in principle something that the spec review group, in this case, LWG, needs to ack.

tkoeppe:
To be clear:

* $X$ and $X^{(q)}$ are two distinct objects.
* Both are sequences:
  * $X = (X_i)_{i=0}^{n-1}$
   * $X^{(q)} = (X_{i}^{(q)})_{i=0}^{n-1}$
   * If you will, $(X^{(q)})_{q=1}^r$ is a sequence of sequences.
* For any sequence $S$, the name of the sequence names the entire sequence, and subscripting it names an individual element, so $S_i$ is a specific element, and $S$ is shorthand for $(S_i)_{i=a}^b$.

tkoeppe:
> Given the wording was not really broken before the changes, I am not sure that it is worth it.

What I find unclear in the approved wording is what the result of `Philox` is. It says $X'$, but what is that?

tkoeppe:
> My point re paranthesis was, if we start using that for _**X**_ and _**V**_, they should also be used for the _**key**_. Now usage is inconsistent.

OK, sure, we can either drop the parentheses, or also write $\mathit{key}^{(q)}_k$ if you like, that's fine!

iburyl:
> One could argue further that the term "$r$-round substitution-permutation network" isn't defined :-)

I can agree with that, but it is more a light reference to original paper rather than something really used in the definition. This sentence can be removed without changing the algorithm definition.

Currently it is defined as:

> A single round of the generation algorithm performs the following steps:
> ...
> The output sequence X’ of the previous round (X in case of the first round)
> ...
> After r applications of the single-round function, Philox returns the sequence Y = X’.

"single round of the generation algorithm" and "single-round function" could probably be called more consistently though.

tkoeppe:
Yes, indeed, what is a "round", and why is there a "first round", etc. etc. I think "$r$-round network" is a good evocative name _after the fact_, but the name alone doesn't mean anything. In maths you should generally say what something is, and then you can give it a name for later reference. But names don't make meaning.

We can also use a totally different symbol instead of $X^{(q)}$, e.g. $W^{(q)}$, if you prefer.

Oh, by the way, note that the drafting directives [forbid multiletter variable names](https://www.iso.org/sites/directives/current/part2/index.xhtml#_idTextAnchor119), so we'll probably have to do something about $\mathit{key}$ at some point :-)

tkoeppe:
> Given the wording was not really broken before the changes, I am not sure that it is worth it.

I would definitely call into question what the result of `Philox` is in the approved wording. The paper says "returns the sequence $Y = X'$", but $X'$ is defined as "the output sequence of the previous round". Again, "previous" seems a bit disconnected in this context, and in any case one could read this as "previous means that there's something else coming after, so the final round is just discarded". But surely that's now what we mean here?

tkoeppe:
Good idea, done!

image

iburyl:
Bullet list:
* K^{(q)}_k is the k^{th} ... <- singular
* K_k _are the elements_ ... <- plural
* M_k is multipliers[k], and ... <- singular

Since everything around became singular...
* K_k is the k^{th} element of the key sequence K

iburyl:
Other than that looks ok for me. But as @villevoutilainen said, there should be some bless from LWG on that change.

tkoeppe:
> * K_k is the k^{th} element of the key sequence K

Yes, good call, done. I might even apply this change separately already, since ISO will require the renaming anyway.

> Other than that looks ok for me. But as @villevoutilainen said, there should be some bless from LWG on that change.

Yes, indeed, but I wanted to get some sense of agreement first and flesh out the details -- thanks!

iburyl (approved):
ok

tkoeppe:
I've updated the PR after all the LWG issues have gone in:

image

tkoeppe:
@Dani-Hub, @jwakely  I think it would be prudent to handle this as an LWG issue. What do you think? If so, how would you like me to present the proposed changes? Would you by any chance be able to derive them from this pull request?

Dani-Hub:
It sounds reasonable to me to open an LWG issue.

jwakely:
I agree

Dani-Hub:
> Thanks, will do! @Dani-Hub, how would you like me to describe the proposed wording in the email? Would a link to this PR suffice, or shoudl I transcribe it?

A link would suffice, maybe you could add some introductory words

tkoeppe:
Of course, yes, I was just asking about the wording! (The intro will be shamelessly copied from the this PR, of course, but I'll handle that.)

tkoeppe:
@Dani-Hub Looks good, thank you!

Issue #2252, opened 2018-07-07
Reorder [basic] before [lex]


zygoloid:
[basic] would make more sense if it appeared prior to [lex] rather than after.  Conversely, [lex.separate] and [lex.phases] really belong in [basic], not [lex]. (We should also consider moving [cpp] to immediately after [lex] so that the top-level clauses are in a more logical phase-of-translation order, but keeping it in exile -- as an appendix of sorts -- has some merit too.)

jensmaurer:
Literally minutes after releasing a draft with the large C++20 renumbering applied do we come up with even further large-scale rearrangements.

tkoeppe:
Well, there's always force-pushing :-)

jensmaurer:
Maybe also move the preprocessor section near lex or merge both into a new "lexical processing" clause.

[basic] has a some details that need to move later, e.g. allocation/deallocation function details should go to "Declarations".

Operator overloading [over.oper] should partially move to [expr] and to "Declarations".

jensmaurer:
I have reservations about moving [basic] before [lex]; this needs more internal restructuring.

AlisdairM:
I was working on a PR/paper that would perform the suggested restructuring, and it quickly ended up redistributing most of [lex] across [basic] and [cpp].  In hindsight, the contents of [lex] truly should precede the contents of [basic].

One reorganisation that I believe did help was moving [cpp] to directly follow [lex], per the phases of translation, and then moving [modules] directly after the relocated [cpp] to put all the program composition and source transformation together, before starting on trying to interpret its actual contents.  That involved also creating and deploying `\firstcoreclause` and `\lastcoreclause` to match `\firstlibraryclause` and `\lastlibraryclause` in `config.tex`.

cor3ntin:
lex.phases and lex.separate don't belong in lex, but lex really is better before basic so I don't see a good solution.
I really would be against splitting the rest of lex. lex is starting to be in good shape.

[modules] is in sort of a weird place - but it needs to be after [dcl] to make sense (or at least after basic). [cpp] does belong after [lex]. I think it was pushed to the end because well, it's [cpp].

We could do either of these somewhat targeted changes:
 - Move [modules] right before [cpp]
 - Move [cpp] right after [lex]
 - Move [cpp] right after [lex] and [modules] right after [basic] - giving us  lex/cpp/basic/modules - which kind of make sense

The status quo does not prevent me to sleep, this would be a marginal improvement

[...]

PR #7153, opened 2024-07-23
[DO NOT MERGE] Reorder core basics

AlisdairM:
This PR shows a vision for re-ordering the Core clauses around program construction for C++26.  This is based on the pre-St Louis draft, NOT the current working draft.  The intent is to show direction, and re-apply useful edits if this direction gains approval.

There are a variety of small edits, but the significant ones are:

- give each phase of translation a stable label so they can be individually referenced
- arrange the lex subclauses to better follow the phases of translation
- merge the specification of comments directly into phase 3, as that is the only place they are used
- split the literals subclause into 2, to reflect string literals are in the preprocessor, but arithmetic literals are phase 7
- move all the preprocessor tokens (phase 3) below a new subtitle to better group them
- MERGE [cpp] INTO PLACE AS PHASE 4
- move modules adjacent to [lex], preceding [basic]

This is a response to #2252 where in practice I could not find a satisfactory way to integrate the primitive parts of [lex] with [basic], so instead made the best attempt to clarify the source aspects of program creation that I could.

Overall, I find this proposed structure very helpful, as I better understood a number of parsing issues, mostly preprocessing issues that I did not always realise were preprocessing, but cannot be sure how much of that understanding comes from being involved in making the transformation itself.


AlisdairM:
While I like this rearrangement, the placing of the predefined macro names feels odd, as I got too used to it being the very last part of the Core specification, before the Library intro, and it feels lost buried in the middle of this combined clause.

jensmaurer:
Not having looked at the details, do we finally differentiate the grammar for preprocessor tokens from those for phase 7 tokens?  "identifier" does double-duty here.

AlisdairM:
Closing this PR in favor of #7272 that does not try to mess with the text for phases of translation, but more freely reorganizes the core clauses.

PR #7160, opened 2024-07-26
[func.search.bm] Remove superfluous the

hewillk:


source/utilities.tex
```diff
@@ -14808,10 +14808,8 @@
 \begin{itemdescr}
 \pnum
 \expects
-The value type of \tcode{RandomAccessIterator1} meets
-the \oldconcept{DefaultConstructible},
-the \oldconcept{Copy\-Constructible}, and
-the \oldconcept{CopyAssignable} requirements.
+The value type of \tcode{RandomAccessIterator1} meets the \oldconcept{DefaultConstructible},
+\oldconcept{Copy\-Constructible}, and \oldconcept{CopyAssignable} requirements.
```
jensmaurer, line 14812: Why did you reflow the list?  The three items in the list don't fit in one line, so should stay as three lines, to show the parallelism.  See also https://sembr.org/ .

Also, is the hyphenation hint for CopyConstructible still needed, now that there are fewer words in that particular line?

jensmaurer (changes requested):

jwakely:
This change doesn't seem necessary, the existing text isn't wrong.

jwakely:
Yes. If that's the rationale for the change, it would have been helpful to say so.

jwakely (approved):
Looks OK to me if Jens is happy with the semantic breaks

jensmaurer (approved):

PR #7169, opened 2024-07-28
[execution.syn] Declare read_env with the right name

Eelis:


jensmaurer:
This is a defect present in P2300R10, but the fix seems obviously correct.

@jwakely  ?

jwakely (approved):
I think `read` was renamed to `read_env` in R10 of the paper, so the synopsis was apparently missed when renaming it.

jensmaurer (approved):

PR #7180, opened 2024-07-28
[lex.separate][module.unit] move definitions of program and translation unit

AlisdairM:
The definition of program at the top of [basic.link] should move to the front of [lex.separate] so that it is defined before its first usage, and also clarifies what the phases of translation produce.

Similarly, move the definition of the grammar production translation-unit to the top of the first clause to actually use it, [module.unit].

Finally, retitle [basic.link] as just Linkage, rather than prgrams and linkage.

### DIFF
--- a/source/modules.tex
+++ b/source/modules.tex
@@ -6,6 +6,10 @@
 \rSec1[module.unit]{Module units and purviews}
 
 \begin{bnf}
+\nontermdef{translation-unit}\br
+    \opt{declaration-seq}\br
+    \opt{global-module-fragment} module-declaration \opt{declaration-seq} \opt{private-module-fragment}
+
 \nontermdef{module-declaration}\br
     \opt{export-keyword} module-keyword module-name \opt{module-partition} \opt{attribute-specifier-seq} \terminal{;}
 \end{bnf}
@@ -25,6 +29,10 @@
     identifier \terminal{.}\br
     module-name-qualifier identifier \terminal{.}
 \end{bnf}
+
+\pnum
+\indextext{translation unit}%
+A translation unit consists of a sequence of declarations.
 
 \pnum
 A \defn{module unit} is a translation unit that contains
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -2649,23 +2649,10 @@ during the lookup for a \grammarterm{namespace-name} or for a name in a
 only namespace names are considered.%
 \indextext{lookup!name|)}%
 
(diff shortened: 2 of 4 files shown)


AlisdairM:
Just to confirm that I looked at every reference for program and translation unit to see if they had a cross reference that should be updated.   These terms occur quite frequently, but rarely have a cross-reference, and in those places that do have a cross reference, they are frequently cross referencing a concept in the whole sentence rather than a definition of either term.

AlisdairM:
Note to CWG chair: this PR, if accepted, would seem to be making progress towards resolving http://wg21.link/cwg2670.

source/modules.tex
```diff
@@ -6,6 +6,10 @@
 \rSec1[module.unit]{Module units and purviews}
 
 \begin{bnf}
+\nontermdef{translation-unit}\br
+    \opt{declaration-seq}\br
+    \opt{global-module-fragment} module-declaration \opt{declaration-seq} \opt{private-module-fragment}
+
```
jensmaurer: A _translation-unit_ is not necessarily a module unit, so I find this new place not good.  I agree having this production in [lex] is not good, either, because this is a phase 7 grammar production.  I agree that [basic.link] isn't a great place, so it's either [basic.pre] or a new subclause [basic.program].

AlisdairM: Looking again, I think the best location is to introduce [basic.def].  What you you think?

AlisdairM:
Rebased with wording better targeted at relevant clauses, and clearly addressing a resolution to CWG2670.

jensmaurer:
This does more than moving stuff around (which makes review harder), and will have a merge conflict with the approved CWG3091.

AlisdairM:
Note that it does apply the resolution to [CWG3091](https://cplusplus.github.io/CWG/issues/3091.html) in the words that it moves.  Other than moving words --- one paragraph split into two locations --- it updates all relevant links to where the words used to occur.  Some of those cross-references are simply removed, not redirected, as the words land close to the location that was making a cross-reference.

I believe the only other change I made was to drop "Programs and" from the subclause header "Programs and linkage", without touching any of the stable labels.

I also believe the new locations for both definitions are much more suited than where they were, or I would not have submitted the PR ;)

AlisdairM:
I also expect to push an update with conflict resolution after the Kona motions land.

tkoeppe:
Editiorial meeting 2026-03-26: this is not an improvement, let's leave it.

PR #7191, opened 2024-07-30
[config.tex] Create and apply macros denoting first and last core chapters

AlisdairM:
This PR should show no difference in the rendered document, but will make it much easier to refactor the Core clauses in the future --- just as we can refactor Library clauses today.

### DIFF
--- a/source/config.tex
+++ b/source/config.tex
@@ -8,6 +8,9 @@
 %% Release date
 \newcommand{\reldate}{\today}
 
+%% Core chapters
+\newcommand{\lastcorechapter}{cpp}
+
 %% Library chapters
 \newcommand{\firstlibchapter}{support}
 \newcommand{\lastlibchapter}{exec}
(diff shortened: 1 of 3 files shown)


source/config.tex
```diff
@@ -8,6 +8,10 @@
 %% Release date
 \newcommand{\reldate}{\today}
 
+%% Library chapters
```
jensmaurer, line 11: "Library" chapters, really?

jensmaurer (changes requested):

AlisdairM:
Oops.  Fixed comment, rebased onto main, and force-pushed for a clean history.

jensmaurer (approved):

jensmaurer:
This looks reasonable, although I notice we sometimes refer to [intro] and sometimes to [lex] when we refer to a range of clauses.

PR #7192, opened 2024-07-30
[lex.phases] replace term 'input file' with 'source file' in phase 1

AlisdairM:
The only part of the standard that refers to input files is phase 1 of the phases of translation.  Everywhere else, we use the term source file, that has a clear definition in the preceding [lex.separate] clarifying that we do not necessarily mean text files in a hosted filing system.

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -69,19 +69,19 @@ occur, although in practice different phases can be folded together.
 \begin{enumerate}
 \item
 \indextext{character!source file}%
-An implementation shall support input files
+An implementation shall support source files
 that are a sequence of UTF-8 code units (UTF-8 files).
 It may also support
-an \impldef{supported input files} set of other kinds of input files, and,
-if so, the kind of an input file is determined in
-an \impldef{determination of kind of input file} manner
-that includes a means of designating input files as UTF-8 files,
+an \impldef{supported source files} set of other kinds of source files, and,
+if so, the kind of a source file is determined in
+an \impldef{determination of kind of source file} manner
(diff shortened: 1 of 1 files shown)


jensmaurer:
CWG and/or SG16 spent quite some time massaging [lex.phases] p1 into its present form (for the UTF-8 source support); I won't touch that wording editorially.

Maybe "input file" is the subject of phase 1, and "source file" is what's handled in phases 2 and 3.

AlisdairM:
I am very happy to defer this to Core, but at the moment "input file" is not a defined term in the standard, and occurs in only this one paragraph.  We are very careful to define "source file" as not necessarily implying a file in a filing system, but have no such notion of what "input file" should mean.  I believe we need clarity here.

jensmaurer:
Where in [lex.separate] do we expressly say we don't imply a "file"?  I think we just do so by omission, and "input file" is no better or worse here.

AlisdairM:
I stand corrected.  The only reference I can find to "not necessarily a file" is a footnote in [headers], 16.4.2.3, and that applies to headers, not to source files.

jensmaurer:
Can we please _not_ do this change, then?

AlisdairM:
I would still like the discussion in Core, because the term "input file" is not defined, unlike the term "source file".  If you would prefer to create a core issue and close this PR, I would be content.

AlisdairM:
Jens stated that such discussion belongs in Core.  I will start a thread there this week, pointing back to this commit.

AlisdairM:
This will be addressed with the pending paper P3556

PR #7193, opened 2024-07-30
[lex] Reorganize contents to follow grammar and phases of translation

AlisdairM:
This PR purely moves existing words around, and does not create any new content.  It would be the precursor to a larger change set that might integrate [cpp] into lex, or move it adjactent, and similarly move [modules] adjactent.

This may be more significant than we want to land in C++26, but I offer it while time is still available, and as inspiration for the C++29 reorg otherwise.

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -465,29 +465,47 @@ as specified in the Unicode Standard
 for the respective Unicode encoding form.
 \indextext{character set|)}
 
+\rSec1[lex.comment]{Comments}
+
+\pnum
+\indextext{comment|(}%
+\indextext{comment!\tcode{/*} \tcode{*/}}%
+\indextext{comment!\tcode{//}}%
+The characters \tcode{/*} start a comment, which terminates with the
+characters \tcode{*/}. These comments do not nest.
+\indextext{comment!\tcode{//}}%
+The characters \tcode{//} start a comment, which terminates immediately before the
+next new-line character. If there is a form-feed or a vertical-tab
+character in such a comment, only whitespace characters shall appear
+between it and the new-line that terminates the comment; no diagnostic
+is required.
+\begin{note}
+The comment characters \tcode{//}, \tcode{/*},
+and \tcode{*/} have no special meaning within a \tcode{//} comment and
+are treated just like other characters. Similarly, the comment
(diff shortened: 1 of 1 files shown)


AlisdairM:
Is this kind of internal reorganization of a clause still in scope for C++26?

If not, please flag with the C++29 milestone.
If it is in scope, I will work with editors to get it into shape to land.

If only parts are in scope (I am thinking specifically of moving [lex.comment]) I can break them out into separate PRs.

AlisdairM:
Also, needs a rebase!

jensmaurer:
I think shuffling around stuff inside [lex] is still possible for C++26 if we want to do that. Moving [lex.comment] earlier seems like a no-brainer; we can do that right away if we wanted to.  For the rest: @AlisdairM , can you show the new section order in a comment here, e.g. as a screenshot of the table of contents or so?  It's hard to reconstruct that from the github diff view.

tkoeppe:
> I think shuffling around stuff inside [lex] is still possible for C++26 if we want to do that. Moving [lex.comment] earlier seems like a no-brainer; we can do that right away if we wanted to. 

Let's land that first then and appraise the remainder of the proposal afterwards.

AlisdairM:
See #7315 for moving [lex.comment], I will remove that part before rebasing and pushing this PR again, and presenting the new clause ordering more clearly. per Jens comment above.

AlisdairM:
This branch has now been rebased so can lose the "needs rebase" tag.

AlisdairM:
I have updated the initial comment with the intended section outline, and an alternative that I prefer that simple adds another level of grouping, with no additional text.

jensmaurer (approved):
Ok, trusting the "nothing but text reshuffling" presentation, I think this is an improvement.

The additional subclause nesting suggested in the intro text (but not actually realized in the pull request) seems questionable without also splitting _identifier_ into _pp-identifier_.

AlisdairM:
Trust, but verify!
I tried to repeat the re-organization to verify that there were no intervening commits to the cut and pasted text, and discovered that I had also moved a few paragraphs around.  My attempt to push the simple fix accidentally force-pushed a copy of main (as I forgot to commit changes) and that closed this PR.  I am going to open a new PR that really is just a cut and paste for the section ordering and nothing more.

PR #7194, opened 2024-07-30
[lex.pptoken] Consistent use of preprocessing vs processing

AlisdairM:
There are three cases here all doing the same thing.  Two refer to preprocessing a directive, while the first refers to processing without the pre.


jensmaurer (approved):

jensmaurer:
Feels like a typo fix to me.

PR #7195, opened 2024-07-31
[ios.syn] Fix typo'd xref to [fpos] introduced in 4b3f32ae

Quuxplusone:
@jwakely LGTY, right? https://eel.is/c++draft/fpos exists.


jwakely (approved):
Not sure what happened there - maybe I scrolled _down_ from the definition of `fpos` to the next rSec instead of scrolling up to the previous one?!

jensmaurer (approved):

PR #7196, opened 2024-08-01
[dcl.fct] Fix obsolete phrasing when defining 'function type'

jensmaurer:
Fixes cplusplus/CWG#588

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -3588,16 +3588,14 @@ the \grammarterm{parameter-declaration-clause} as described below and
 the optional \keyword{noexcept} is present if and only if
 the exception specification\iref{except.spec} is non-throwing.
 \end{itemize}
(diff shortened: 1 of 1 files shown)


jensmaurer:
@tkoeppe, this feels like an editorial adjustment of outdated wording, from a time when we had two paragraphs talking about function types.  Now, we only have one.

jensmaurer:
@t3nsor , please have a look (no, it's not perfect, but should fix the first-level concern).

t3nsor:
Yeah, it's fine, I guess. It would be nice if the definition could also be moved above

"The declared return type \tcode{U} of the function type is determined as follows:"

but that would require some more extensive rephrasing and it may not be worth the effort.

tkoeppe:
Yes, indeed, the status quo looks off.

PR #7212, opened 2024-08-10
[stmt.for] Fix scope issue in for-statement translation.

### LINKED ISSUE #7211: [stmt.for] Missing curly brackets in for-statement translation
According to [[stmt.for]](https://eel.is/c++draft/stmt.for),

    for (; false; ++i) int i = 5;

is equivalent to:

    {
        ;
        while (false) {
            int i = 5;
            ++i;
        }
    }

which is well-formed. But this is clearly not the intent, and all compilers I tried reject the `for` loop above.

Eelis:
Fixes #7211.


tkoeppe:
@jensmaurer: it's be worth fixing this, but this is such classic wording that I'd like CWG to confirm that we want to change it this way.

PR #7213, opened 2024-08-10
[cpp.predefined] Sort prefined macros

AlisdairM:
This change also puts the text requiring the feature macros to be predefined adjacent to the table of feature macros, at the end of the list.  It also moves the `__STDCPP_THREADS` macro from the conditionally defined list into the main list, as it is specified in exactly the same form as the five macros for extended floating point types in the main list.

Note that `__STDCPP_BFLOAT16` remains out of order, to maintain a consistent block for the extended floating point types.


jensmaurer:
Once upon a time, there was an issue with pagination that was best resolved by putting the feature-test macros first.

I think the correct approach is to move the floating-point macros into the "conditionally-defined" section (where they belong), and not moving `__STDCPP_THREADS` to the regular section.

AlisdairM:
I think the upon-times have changes, and sorting the features paragraph next to the tables renders better now.

For the "optional" macros, I think they come in two categories, and prefer my suggested organization, although happy to move the threads/floats macros back to the conditional section if preferred.

My reasoning on `__stdcpp_threads` is that the specification for this macro is well defined, and non-optional for a hosted implementations as hosted implementations are required to support multiple threads, in which case the value of the macro is fully defined.  For free-standanding implementations the support for multiple threads is optional, but the value and presence, or lack, of the macro is similarly fully specified.

For the float macros, there is similarly a full spec that we either have an implementation will full support for the optional types, or we do not.   There is no middle ground.

For the __STDC_* macros there is no specification, we are merely allowing an implementation to choose a C compatibility layer in any manner that it chooses, including completely ignoring such concerns.  That feels a different category of more strictly optional, where the standard says almost nothing about these macros other than "you are allowed to define them, and should document them if you do".

I would also be open to adding another paragraph for just the "fully-specified-but-conditionally-supported" macros, which are the set I unified.  If we do so, I can also alpha-sort the `__stdcpp_float` macro to the top without interleaving with other macros.

jensmaurer:
My guess is that the "conditionally defined" list of macros is intended to help the user figure out easily whether they can directly do `#if macro == x" or whether they must do `#if defined(macro) && macro == x`.

Whether something is "fully specified" doesn't change the situation that you need to do the "#if defined(macro)" dance if you want to write a portable program.  So, I'd say we've simply added the floating-point macros to the wrong list.

jensmaurer:
I note that `__STDC_MB_MIGHT_NEQ_WC__` is still in the "conditionally-defined" section, even though it is also fairly well-defined.

I continue to believe that the "optional" section should contain all macros where you need to do `#if defined(macro)` before use (for a portable program), which is actually a more precise dividing line than any of your criteria of "well-specifiedness".  Please reorder accordingly.

PR #7223, opened 2024-08-14
[dcl.init.list] Add comma

Andreas-Krug:


jensmaurer:
This doesn't look like an improvement to me. This is a two-prong "or" where usually no comma is applied.  Also, the superordinate bulleted list uses ", or", so there is a bit of a confusion chance here.

PR #7237, opened 2024-08-20
[text.encoding.aliases] Add note about what isn't required

jwakely:
Make it explicit that `copyable` is intended, rather than `semiregular`, and that the return types of `begin()` and `end()` can differ.

Also remove a FIXME comment for something that doesn't need fixing. These requirements on a type are specified in our usual form, it would be wrong to use _Remarks_: or _Requires_: here.

### DIFF
--- a/source/locales.tex
+++ b/source/locales.tex
@@ -5063,6 +5063,11 @@ struct text_encoding::aliases_view : ranges::view_interface<text_encoding::alias
 \tcode{ranges::\libconcept{view}},
 \tcode{ranges::\libconcept{random_access_range}}, and
 \tcode{ranges::\libconcept{borrowed_range}}.
+\begin{note}
+\tcode{text_encoding::aliases_view} is not required to satisfy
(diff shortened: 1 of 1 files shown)


jwakely:
This came up in a GCC bug report where it was suggested that the presentation of `aliases_view` as an apparently empty class suggests it should be default initializable. It was also questioned whether the _implementation defined_ return types need to be the same type or not.

@cor3ntin confirmed the intent of the [P1885R12](https://wg21.link/p1885r12) design, which is that it models `copyable` but not necessarily `default_initializable`, and models `range` but not necessarily `common_range`. Add a note to make it clear that the normative requirements are what's required, and no more.

Please say if you'd rather have an LWG issue for this. I'd be inclined to vote NAD on such an issue, but at least we'd have the issues list recording that the normative requirements are correct as written.

cor3ntin:
I'm happy with the change

jensmaurer (approved):

PR #7239, opened 2024-08-23
[expr.prim.lambda.capture] Incorporate ellipsis into "captured by copy" definition

Endilll:
It would be quite strange if captures starting with `&` would suddenly be captured by copy because of presence of ellipsis.


jensmaurer (approved):
This correctly updates the quotes of the relevant grammar snippets.

tkoeppe:
Yes, indeed -- thanks!

PR #7240, opened 2024-08-23
(library-wide) Remove `inline` from variable templates

CaseyCarter:
Removes extraneous `inline` from the definitions of variable templates in [format.syn], [format.tuple], [time.format], [execution.syn], and [exec.util.cmplsig].

### DIFF
--- a/source/exec.tex
+++ b/source/exec.tex
@@ -464,7 +464,7 @@ namespace std::execution {
   enum class forward_progress_guarantee;
   inline constexpr get_forward_progress_guarantee_t get_forward_progress_guarantee{};
   template<class CPO>
-    inline constexpr get_completion_scheduler_t<CPO> get_completion_scheduler{};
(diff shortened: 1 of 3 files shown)


jensmaurer (approved):
Seems straightforward.

@jwakely , any thoughts?

jwakely:
I think this was made unnecessary by CWG 2387, right? If so, I'm fine with dropping them now.

GCC only implemented that three weeks ago, which might have contributed to my confusion about whether these `inline` keywords were needed.

jensmaurer:
We've had earlier incarnations of inline removal for these cases, e.g. 9568c66abd12eadd696b9f16bb4884908e2aad66 and d2ba9e375911c7994454fcb72c7d2f3cdb7ba7d4 and 9ac55553459e15c84db6d8072c93787d41ef7ccf .

PR #7241, opened 2024-08-24
[dcl.type.elab] Remove normative duplication about name binding

Endilll:
[[dcl.type.elab]/4](https://eel.is/c++draft/dcl.type.elab#4.sentence-1) goes out of its way to normatively specify that elaborated type specifiers in friend declarations do not bind names, despite a blanket statement in [[dcl.meaning.general]/2.1](https://eel.is/c++draft/dcl.meaning.general#2.1) that (all) friend declarations do not bind names.

I suggest to turn this into a note referencing [dcl.meaning], or remove it altogether.


Endilll:
Please let me know if I'm following the wrong process here, and should instead email CWG or Jens, or do something else entirely.

Endilll:
Never mind, [dcl.meaning] does not cover friend-type-specifiers.

PR #7248, opened 2024-09-05
[utility.syn, flat.map.defn] Remove all [[nodiscard]] from library wording

Mick235711:
After extensive discussion about [[nodiscard]] policy in LEWG, it was agreed that [[nodiscard]] should not appear in the library wording (see [SD-9](https://isocpp.org/std/standing-documents/sd-9-library-evolution-policies)). However, the wording removal paper, P2422R1 (approved in St. Louis, #7110), does not seem to remove all occurrences of [[nodiscard]] in the library clauses due to some recently added [[nodiscard]] annotation (such as those on flat containers’ `empty()`) being missed.

Therefore, this PR removes all occurrences of [[nodiscard]] in N4988. Those are:
- 22.2.1 [utility.syn]
- 24.6.9.2 [flat.map.defn]
- 24.6.10.2 [flat.multimap.defn]
- 24.6.11.2 [flat.set.defn]
- 24.6.12.2 [flat.multiset.defn]
- 24.7.3.6.1 [mdspan.mdspan.overview] and 24.7.3.6.3 [mdspan.mdspan.members]

After the change, the only mention of [[nodiscard]] in the standard is in examples and Core wording.

(I assumed that this is editorial since it seems likely that these are just missed by P2422's wording...)

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -15999,7 +15999,7 @@ namespace std {
     const_reverse_iterator crend() const noexcept;
 
     // \ref{flat.map.capacity}, capacity
-    [[nodiscard]] bool empty() const noexcept;
+    bool empty() const noexcept;
     size_type size() const noexcept;
     size_type max_size() const noexcept;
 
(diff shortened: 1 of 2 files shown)


jensmaurer (approved):

jensmaurer:
@jwakely , this seems like an editorially-obvious sync operation. What do you think?

jwakely (approved):

PR #7249, opened 2024-09-06
[associative.reqmts.general],[unord.req.general] Fix cross-references to [container.alloc.reqmts] and [container.reqmts]

CaseyCarter:
Both paragraphs incorrectly point to [container.reqmts] instead of [container.alloc.reqmts] for "the requirements of an allocator-aware container".

Both paragraphs state "the requirements placed on `value_type` in [container.alloc.reqmts] apply instead to `key_type` and `mapped_type`" for `map`, `multimap`, `set`, and `multiset`. This would be nonsense, since those containers never allocate or (or allocator-construct) `key_type` or `map_type` objects: _do_ allocator-construct `value_type` objects. The examples in each paragraph also make the error obvious: there are no `CopyAssignable` requirements in [container.alloc.reqmts].

I suspect these pairs of xrefs in each paragraph were transposed sometime in the past and we just haven't noticed.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -2726,12 +2726,12 @@ and \tcode{nh} denotes a non-const rvalue of type \tcode{X::node_type}.
 \pnum
 A type \tcode{X} meets the \defnadj{associative}{container} requirements
 if \tcode{X} meets all the requirements of an allocator-aware
-container\iref{container.reqmts} and
+container\iref{container.alloc.reqmts} and
 the following types, statements, and expressions are well-formed and
 have the specified semantics,
 except that for
 \tcode{map} and \tcode{multimap}, the requirements placed on \tcode{value_type}
-in \ref{container.alloc.reqmts} apply instead to \tcode{key_type}
+in \ref{container.reqmts} apply instead to \tcode{key_type}
 and \tcode{mapped_type}.
 \begin{note}
 For example, in some cases \tcode{key_type} and \tcode{mapped_type}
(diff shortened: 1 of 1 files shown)


cjdb:
We may want to defer this to LWG first: I don't see any other wording that prohibits `map<K&, V>` (but that has issues), and it would be strange to allow `map<K&&, V>`, but not any of `map<K&, V>`, `unordered_map<K&, V>`, or `unordered_map<K&&, V>`. It might also be worth raising whether we want to continue supporting cv-qualified `key_type`s, especially since `MapLike<K const, V>` is the same as `MapLike<K, V>`, but with additional instantiations.

jwakely:
> I suspect these pairs of xrefs in each paragraph were transposed sometime in the past and we just haven't noticed.

Yes, d09a77e09a5210c50994b89d999f3cbdfcc9a436 fixed another case of this.

And dd32b7eeb21e0ca8be7aae1996814a2ffda0ed83, and 4fac9f97a2c25d39a01f75cf198d0783bfa8deda

The subclause [container.alloc.requirements] was added by 93ff092d1cd2b335f372b9546365b3d495caf2d8, before that they didn't have their own subclause, so referring to [container.reqmts] was correct.

CaseyCarter:
> We may want to defer this to LWG first: I don't see any other wording that prohibits `map<K&, V>` (but that has issues), and it would be strange to allow `map<K&&, V>`, but not any of `map<K&, V>`, `unordered_map<K&, V>`, or `unordered_map<K&&, V>`. It might also be worth raising whether we want to continue supporting cv-qualified `key_type`s, especially since `MapLike<K const, V>` is the same as `MapLike<K, V>`, but with additional instantiations.

Changing "the requirements placed on value_type in [container.alloc.reqmts] apply instead to `key_type` and `mapped_type`" to "the requirements placed on `value_type` in [container.reqmts] apply instead to `key_type` and `mapped_type`" doesn't affect this. The first requirement on `value_type` in [container.reqmts] is:

> `typename X::value_type`
> 
> [-2-](https://eel.is/c++draft/container.reqmts#2) _Result_: `T`
> [-3-](https://eel.is/c++draft/container.reqmts#3) _Preconditions_: `T` is _`Cpp17Erasable`_ from `X` (see [[container.alloc.reqmts]](https://eel.is/c++draft/container.alloc.reqmts), below).

which still implicitly requires `value_type*` to be a valid type, and therefore `key_type`* and `map_type`*. This still forbids `key_type` and `map_type` to be reference types just as effectively as before. It's not terribly explicit - and like I said seems silly, since containers don't pass pointers of those types to `Allocator::construct` and `Allocator::destroy` - but swapping these cross-references doesn't change this property.

cjdb:
Thanks for clarifying!

jensmaurer (approved):

PR #7251, opened 2024-09-12
[sf.cmath.assoc.laguerre,sf.cmath.assoc.legendre] Add reference to eq

mattkretz:
The associated Laguerre/Legendre functions build on the Laguerre/Legendre functions, which are defined in different equations. Point to them from the associated functions.

Also use the correct \ell as used in the formula.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -9949,6 +9949,7 @@ of their respective arguments
 \returns
 $\mathsf{L}_n^m(x)$,
 where $\mathsf{L}_n^m$ is given by \eqref{sf.cmath.assoc.laguerre},
+$\mathsf{L}_{n+m}$ is given by \eqref{sf.cmath.laguerre},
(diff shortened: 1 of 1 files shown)


jensmaurer (approved):

mattkretz:
Oh, there are more of such formula reuses. The other functions typically use a "See also" paragraph. My change makes the [sf.cmath] section less consistent.

Options:
1. Undo the formula reference and add a "See also" paragraph instead.
2. Add a "See also" paragraph and add formula references to all the other functions.
3. Add formula references and remove "See also" paragraphs.

Personally, I find the formula reference to be clearer / more discoverable.

jensmaurer:
Since we normatively refer to ISO 80000-2, where all these symbols are introduced, I don't think we need to invest extra effort into cross-references. A "see also" seems good enough.

tkoeppe:
Thanks!

PR #7252, opened 2024-09-14
[inplace.vector.cons] "Constructs an object" is redundant

CaseyCarter:
So we don't need this `from_range_t` constructor (and only this constructor) to tell us that the `inplace_vector` it constructs happens to be an object. I should also point out that this is inconsistent with the other constructors which simply say "constructs an `inplace_vector` with [properties]".

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -9768,7 +9768,7 @@ template<@\exposconcept{container-compatible-range}@<T> R>
 \begin{itemdescr}
 \pnum
 \effects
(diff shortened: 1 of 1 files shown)


jwakely (approved):

jensmaurer (approved):
Frankly, I don't think we need the redundant statement that a constructor constructs (an object of) its type to start with, but this consistency fix is an initial step in right direction.

PR #7258, opened 2024-09-17
[over.match.best.general] Minor formatting fixes

Bekenn:
$F_1$ and $F_2$ are introduced at the start of this long paragraph, but have morphed to `F1` and `F2` by the time we reach the third bullet point.  This change preserves the original formatting for these names throughout the paragraph.


tkoeppe:
@jensmaurer This seems reasonable, but could you have a look, too?

jensmaurer (approved):
Most wonderful.

jwakely (approved):

PR #7260, opened 2024-09-19
[basic.stc.dynamic.general] Strike "during program execution"

Eisenwave:
Since C++20, it is possible to use `new` during program translation, not just during execution, so it is misleading to specify "during program execution" here.

Note that this text has not been changed since C++98.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3992,8 +3992,7 @@ specified in~\ref{class.copy.elision}.
 \indextext{storage duration!dynamic|(}
 
 \pnum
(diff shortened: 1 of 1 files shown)


Eisenwave:
Now looking at it, maybe we should also say
> objects with dynamic storage duration can be created

... rather than
> objects can be created dynamically

... considering that "created dynamically" doesn't really mean anything.

Eisenwave:
Superseded by https://github.com/cplusplus/draft/pull/7330

PR #7263, opened 2024-09-25
[ios.init] Remove unused Init::init_cnt static member

jwakely:
The text that made use of this variable was removed by LWG1123 and has not been present in the WP since N3090. The effects of `Init` construction and destruction are specified entirely without the use of this variable, so it serves no purpose now.

LWG1123 even noted that this variable could be dropped, stating "Some editorial changes are expected (in addition to the proposed wording) to remove `init_cnt` from `Init`." Let's do that. Better late than never.

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -1132,9 +1132,6 @@ namespace std {
     Init(const Init&) = default;
     ~Init();
     Init& operator=(const Init&) = default;
-
-  private:
-    static int init_cnt;        // \expos
   };
 }
 \end{codeblock}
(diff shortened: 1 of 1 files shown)


tkoeppe:
Is that the "Schwartz counter"?

jwakely:
> Is that the "Schwartz counter"?

Yup, and it's still a valid implementation strategy ... if you can make it threadsafe. Doing that correctly is non-trivial and would be out of scope for [ios.init] so it's good that we no longer try to describe the `ios_base::Init` effects in terms of a simple counter.

jensmaurer (approved):

PR #7264, opened 2024-09-26
[tab:lex.charset.literal] Shorten table heading

jensmaurer:


tkoeppe:
Do we want this change in the mainline?

jensmaurer:
Yes, we want this change in mainline; the table heading looks a bit awkward there, too.  Let me add a few more commits for other tables that I haven't identified yet (because numbers have changed between the current draft and C++23).

jensmaurer:
@tkoeppe, that's all for table headings in mainline.  The other two tables are in c++23 only (because Annex D).

jwakely (approved):

PR #7268, opened 2024-09-26
[styles] Format title of \codeblocktu using 'caption' package

jensmaurer:
This restores the C++20 status of the formatting.

### DIFF
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -587,22 +587,6 @@
 
 \lstnewenvironment{codeblock}{\CodeBlockSetup}{}
 
-% Left-align listings titles
-\makeatletter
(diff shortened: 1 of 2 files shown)


jensmaurer:
Apparently, our prior attempt at redefining some internal stuff of the listings package got neutered by later versions of said package.

Use the right tool for the job instead.

tkoeppe:
@jensmaurer Currently, this allows page breaks between the title and the code block. I couldn't find a general way to fix this easily and am adding a manual break hint before "Translation unit #2" in [basic.link]. I was trying adding a negative penalty _before_ the title to encourage page breaks, but that ends up adding unnecessary breaks.

jensmaurer:
Yeah, getting the page break penalties right is probable for another day.

PR #7276, opened 2024-10-01
[except.uncaught] Tidy the specification for uncaught exceptions

AlisdairM:
Several concurrent fixes.  First include the normative wording that 'uncaught_exceptions' returns the number of uncaught exceptions *on the current thread*.  This wording is present in the core language.

Then move the core wording for when an exception is uncaught directly into the text that talks about caught and uncaught exceptions.  In the process, turn the reference to into a note, so that there is only one normative specification.

Finally, remove [except.uncaught] as it is now empty, and add the missing descriptive sentence to the intro paragraph of [except.special.general].  This happens to produce quite a pleasing page-break, but that is just luck, not design.

### DIFF
--- a/source/exceptions.tex
+++ b/source/exceptions.tex
@@ -1023,8 +1023,10 @@ needed.
 The function \tcode{std::terminate}\iref{except.terminate}
 is used by the exception
 handling mechanism for coping with errors related to the exception handling
-mechanism itself. The function
-\tcode{std::current_exception()}\iref{propagation} and the class
+mechanism itself.
+The function \tcode{std::uncaught_exceptions}\iref{uncaught.exceptions}
+reports how many exceptions are uncaught in the current thread.
+The function \tcode{std::current_exception}\iref{propagation} and the class
 \tcode{std::nested_exception}\iref{except.nested} can be used by a program to
 capture the currently handled exception.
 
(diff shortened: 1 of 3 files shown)


AlisdairM:
One last tweak to update the comment in xrefdelta.tex to reference the correct PR number, and to move the first paragraph on uncaught exceptions up one, so that there is a much clearer flow from creating the exception object, being uncaught, then caught, then rethrown, before talking about what happens when an exception is uncaught.

I should be done pushing updates to this PR until I get reviews now :)

source/exceptions.tex
```diff
@@ -317,6 +317,15 @@
 the selected constructor is odr-used\iref{basic.def.odr} and
 the destructor of \tcode{T} is potentially invoked\iref{class.dtor}.
 
+\pnum
+An exception is considered uncaught
```
Eisenwave, line 321: While we're at it, this looks like a definition and presumably should be formatted and indexed as one (exception, uncaught).

AlisdairM: Good point!  Added indexing, rebased, and force-pushed.

source/xrefdelta.tex
```diff
@@ -95,9 +95,13 @@
 \movedxref{stoptoken.cons}{stopsource}
 \movedxref{stoptoken.nonmembers}{stopsource}
 
+% https://github.com/cplusplus/draft/pull/7276
+\movedxref{except.uncaught}{except.throw}
+
 % https://github.com/cplusplus/draft/pull/7345
 \movedxref{basic.stc.inherit}{basic.stc.general}
 
+
```
tkoeppe, line 104: Extraneous blank line.

PR #7277, opened 2024-10-01
[ostream.inserters.arithmetic] Put use_facet use on fewer lines

tkoeppe:
This is more compact, keeps the whole template-id on a single line, and avoids some awkward page breaks.


tkoeppe:
image

jwakely (approved):

PR #7278, opened 2024-10-02
[except.handle] Remove confusing comparison to variadic functions

AlisdairM:
The analogy that the ellipsis in an exception handler was similar to an ellipsis in a function declaration may have made sense at one time, but the comparison with a syntax using a macro based API calling 'va_arg' to access its contents --- something that is not possible for an exception handler --- seems more confusing than helpful today.


jensmaurer (approved):
This removes superfluous (and strenuous) explanatory handwaving from normative text.  Looks good to me.

PR #7279, opened 2024-10-02
[depr.lit] Fix grammar according to P2361R6

AlisdairM:
P2361R6 introduced the notion of unevaluated strings, and updated the grammar for literal operator function accodingly. Unfortunely, the corresponding grammar reference that was deprecated was not similarly updated.


jensmaurer (approved):
Agreed, and certainly an editorially-obvious fix to me.

PR #7281, opened 2024-10-02
[except.terminate] Better describe the  function `terminate`

AlisdairM:
While `std:terminate` was originally conceived as the way to report failures in the exception handling machinery, it has evolved to become a more general tool for reporting unrecoverable failures in the C++ runtime.  This rewording attempts to address that evolving design, and in doing so addresses the outstanding %FIXME% that the current text is not adequately descriptive in the first place.

### DIFF
--- a/source/exceptions.tex
+++ b/source/exceptions.tex
@@ -1012,9 +1012,10 @@ capture the currently handled exception.
 
 \pnum
 \indextext{\idxcode{terminate}}%
-% FIXME: What does it mean to abandon exception handling?
-In some situations, exception handling is abandoned
-for less subtle error handling techniques.
+Some errors in a program cannot be recovered from, such as when an exception
+is not handled or a \tcode{std::thread} object is destroyed while its thread
+function is still executing. In such cases,
+the function \tcode{std::terminate}\iref{exception.terminate} is invoked.
 \begin{note}
 These situations are:
 \indextext{\idxcode{terminate}!called}%
@@ -1108,8 +1109,6 @@ whose promise type has an \tcode{unhandled_stopped} member function.
 
 \pnum
 \indextext{\idxcode{terminate}}%
-In such cases,
-the function \tcode{std::terminate} is invoked\iref{exception.terminate}.
 In the situation where no matching handler is found, it is
 \impldef{stack unwinding before invocation of \tcode{std::terminate}}
 whether or not the stack is unwound


jwakely (commented):
I don't think "the C++ runtime" is defined. Maybe just "the program"?

Or "Some errors in a program cannot be recovered from, ..." so that we don't talk about "encountering an error", which sounds like it might just be bad luck rather than well-defined control flow that results from the program logic.

AlisdairM:
*Much* better! Thanks!

Applied your suggestion, followed by minor grammar tweaks, and force-pushed to allow a clean merge.

source/exceptions.tex
```diff
@@ -1016,9 +1016,9 @@
 
 \pnum
 \indextext{\idxcode{terminate}}%
-% FIXME: What does it mean to abandon exception handling?
-In some situations, exception handling is abandoned
-for less subtle error handling techniques.
+Some errors in a program cannot be recovered from, such as when an exception is
+not handled or a thread is destroyed while still running. In such cases,
```
jensmaurer, line 1020: What does it mean to "destroy a thread"?

AlisdairM: Updated to refer to a `std::thread` object.  Force-pushed again to support a clean merge.

AlisdairM:
How do we feel about the proposed description now?

AlisdairM:
Rebased and force-pushed for good measure.  I believe all feedback is addressed, but do we want to flag this for an LWG review as well?

jwakely (approved):
I don't think LWG needs to see this, it seems fairly simple and uncontroversial.

source/exceptions.tex
```diff
@@ -1108,8 +1109,6 @@
 
 \pnum
 \indextext{\idxcode{terminate}}%
```
tkoeppe, line 1111: Is it useful to add another index entry here?

AlisdairM: Taking a quick look at the indexing of `std::terminate` in general, I would like to defer that question as it is not a change introduced by this PR.  We are removing only the first reference to `std::terminate` in this paragraph, not all as another two remain.

If we look at the indexing of `std::terminate`, we might want to do a more focussed pass, including the library indexing as well.

jensmaurer (approved):

source/exceptions.tex
jensmaurer: Agreed, let's fix the indexing later.

PR #7282, opened 2024-10-02
[except.ctor] Retitle subclause as 'stack unwinding'

AlisdairM:
The purpose of this subclause is to define stack unwinding, which in specified in terms of the lifetime of objects, not just class types.  Hence, while much of the text is addressing interactions with contructors and destructors (the original subclause title) it does more than just that.

### DIFF
--- a/source/exceptions.tex
+++ b/source/exceptions.tex
@@ -359,7 +359,7 @@ If a destructor directly invoked by stack unwinding exits via an exception,
 \end{note}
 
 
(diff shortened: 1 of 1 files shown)


jwakely (approved):

jensmaurer:
http://lists.isocpp.org/core/2024/10/16367.php

jensmaurer (approved):

Issue #5226, opened 2022-01-21
[text] Create "Text processing library" clause


jensmaurer:
This issue is further to #5124.

The proposal is to create a top-level clause entitled "Text processing library" [text] in the Working Draft at the current location of [localization] that contains the following, in order:

- [charconv]
- [localization]
- [format]
- [re]
- C library
   - [cctype.syn]
   - [cwctype.syn]
   - [cwchar.syn]
   - [cuchar.syn]
   - [c.mb.wcs]

About 90 pages in total.

Mick235711:
After this change, [strings] will only have ~40 pages and be the second-to-smallest library clause (just above [concepts]). Did string classes really fit in text formatting? At least they are tightly related.

I think probably either merge [strings] with [text] or at least move these two clauses together. Currently purposed [localization] position seems to be too far from [strings], I'd say... (But just personal opinion)

jwakely:
> Did string classes really fit in text formatting?

No. They are containers of characters, not necessarily text. I can have a `std::string` containing invalid UTF-8, for example.

tahonermann:
> No. They are containers of characters, not necessarily text. I can have a std::string containing invalid UTF-8, for example.

I don't find that a persuasive argument for separating strings and text. The string related features are clearly intended to hold and work with text despite the lack of invariants to ensure well-formedness with respect to any particular encoding. I imagine that if we introduce additional text related containers in the future that do have strong encoding associations, they will likewise eschew enforcement of well-formedness due to performance considerations. In those cases, we'll probably relegate violations to library UB. Error handling (or lack there of) doesn't strike me as a good basis for library separation.

That being said, I don't have strong opinions regarding this organization so long as it continues not to impact header or module naming.

jensmaurer:
Postponed to C++26.

tkoeppe:
Given that we'll only need to send the new draft in the March mailing, I wouldn't be entirely opposed to still doing this now, but we should feel unreservedly that we're making an improvement, without any caveats or regrets.

I'd be happy to hear alternative suggestions (e.g. how about merging strings and text), and also positions on the status quo (@tahonermann?).

jensmaurer:
I'm ok with moving all of [strings] (~50 pages) into [text], with the idea that they are intended to represent text. This gets the total to ~140 pages, with plans to grow further (e.g. regex v2, encoding conversion facilities, possibly more from the scope of ICU).

[...]

PR #7283, opened 2024-10-02
[except.handle] group all paragraphs on searching for a handler

AlisdairM:
This commit moves all of the paragraphs involved in the search for a handler for an exception into a single logical sequence.

After this change, [except.spec] deals only with specifying the `noexcept` function decorator and its interaction with the `noexcept` operator, and contains no text regarding exceptions themselves.  It might be appropriate to move that subclause into the [dcl] structure at a future date.

### DIFF
--- a/source/exceptions.tex
+++ b/source/exceptions.tex
@@ -619,6 +619,40 @@ handler continues in a dynamically surrounding try block
 of the same thread.
 
 \pnum
+\indextext{exception handling!terminate called@\tcode{terminate} called}%
+\indextext{\idxcode{terminate}!called}%
+If the search for a handler
+encounters the outermost block of a function with a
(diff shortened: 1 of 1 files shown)


tkoeppe:
@jensmaurer Any thoughts on whether this improves the order of presentation?

AlisdairM:
I have just pushed #7308 which is the main change I wanted to make for exceptions in the clause reorganization.  I have pushed it as a follow-up, based on _this_ PR.

Understood it is late to be bringing this to the table --- that is why I tried to ping the edit list early :)

tkoeppe:
Exceptions aren't part of the planned clause reorganizations (#5226), though, so no worries. This smaller cleanup is independent.

PR #7286, opened 2024-10-03
[array.creation] Clarify Mandates for `to_array` overloads

CaseyCarter:
It's confusing that these `to_array` overloads require `T` to be constructible from various types, when they actually construct `remove_cv_t<T>` objects. We experts know that initialization doesn't depend on the cv-qualification of the target type ([dcl.init.general]/16) but there's no need to make readers jump through hoops to understand the spec.


jwakely (approved):

jensmaurer (approved):
Call me lukewarm in favor.

PR #7288, opened 2024-10-04
[container.requirements] Index missing members of flat containers

AlisdairM:
Add the missing containers to index the operations defined by the container requirements clauses.


source/containers.tex
```diff
@@ -879,6 +887,25 @@
 }
 
 \indexcont{operator<=>}%
+\indexlibrarymember{array}{operator<=>}%
```
jensmaurer, line 890: `\indexcont` almost does this; if we feel we want both orderings for indexing, please fix `\indexcont` instead of near-duplicating its definition.

But, maybe we don't, and all that's missing is `\indexlibrarymemberx` for `basic_string`.

jensmaurer (approved):

PR #7292, opened 2024-10-05
[cpp.import] Active macro directive -> definition

AlisdairM:
The term 'active macro directive' is defined in p6, but never used. Meanwhile, there are multiple uses of 'active macro definition' but that term is never defined, including in the very sentence following the definition of the unused term, and in other clauses that cross-reference to this clause for their definition.

This seems clear enough to be an editorial fix, rather than a Core issue, but I defer to the CWG chair to make that call.


jensmaurer (approved):
Agreed.

PR #7293, opened 2024-10-05
[cpp.subst] proprocessing file -> translation unit

AlisdairM:
The term 'preprocessing translation unit' is defined in [lex.separate] while the term 'preprocessing file' is never defined, and is not used anywhere else in the standard.  Prefer to use the specified term, as it reasonably covers this case.


jensmaurer (approved):
Looks reasonable.

PR #7294, opened 2024-10-05
[depr] Reorder clauses by origin of deprecation

AlisdairM:
Reorders the deprecated features annex to follow the order of the main clauses that the deprecates feature refers to. Where multiple clauses are references, use the one named by the [depr.XXX] stable label.

### DIFF
--- a/source/future.tex
+++ b/source/future.tex
@@ -16,6 +16,26 @@ but having been identified as a candidate for removal from future revisions.
 An implementation may declare library names and entities described in this Clause with the
 \tcode{deprecated} attribute\iref{dcl.attr.deprecated}.
 
+\rSec1[depr.local]{Non-local use of TU-local entities}
+
+\pnum
+A declaration of a non-TU-local entity that is an exposure\iref{basic.link}
+is deprecated.
+\begin{note}
+Such a declaration in an importable module unit is ill-formed.
+\end{note}
+\begin{example}
+\begin{codeblock}
+namespace {
+  struct A {
+    void f() {}
+  };
+}
+A h();                          // deprecated: not internal linkage
+inline void g() {A().f();}      // deprecated: inline and not internal linkage
+\end{codeblock}
+\end{example}
+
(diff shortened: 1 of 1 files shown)


AlisdairM:
Rebased, updated for the creation of the new [text] clause, and force-pushed.

tkoeppe:
Thanks!

source/future.tex
```diff
@@ -1,4 +1,4 @@
-%!TEX root = std.tex
+nnn%!TEX root = std.tex
```
tkoeppe, line 1: What's `nnn`?

AlisdairM: A keyboard gone rogue.  Force-pushed the fix.

source/future.tex
```diff
@@ -674,34 +674,6 @@
 \tcode{current}.
 \end{itemdescr}
 
-\rSec1[depr.format]{Deprecated formatting}
-
-\rSec2[depr.format.syn]{Header \tcode{<format>} synopsis}
-
-\pnum
-The header \libheaderref{format}{format.syn} has the following additions:
```
tkoeppe, line 682: This reference was broken, I fixed it at head.

AlisdairM: I believe I have corrected this now, and force-pushed again.

AlisdairM:
Due to papers targeting Annex D, this PR will go stale quickly!

Could we flag it with either a C++26 or a C++29 milestone so that I can appropriately prioritize keeping it up to date?

tkoeppe:
Is this just moving text around without any changes? We can take this now, but I'm not sure this is generally worthwhile tracking.

AlisdairM:
Yes, this is entirely a copy/paste re-ordering of the clause.

When features are deprecated they try to insert themselves at the appropriate location in this clause, but over time that ordering becomes stale.  One it is correctly ordered it is easy for subsequent additions to preserve the ordering, until the next Great Clause Renumbering.

Issue #6005, opened 2022-11-27
Consider making a new index for macro names


AlisdairM:
Macros are a very special kind of library name, that become more special in C++23.

First, by their nature as macros, they directly rewrite user source rather than participating in name lookup.  It would be good to have such names collected in a single place to reference.

Secondly, the library is not the only source of macro names, given the compiler supplies some predefined macros.

Third, as of C++23, the user will not see any of those macros (other than compiler predefines) if they `import std;` rather than directly include library headers.  I think this third point now presses the case for separating macros out of the library/compiler index, and into an index of their own.

Eric41293:
I see there are as yet no comments on this issue. Are people for/against this change?

For my part, I am in favor of it.

jensmaurer:
@tkoeppe, having a separate macro names index sounds reasonable to me. What do you think?

AlisdairM:
If someone can set up the basic structure for the index, and some macro-indexing LaTeX macros, I volunteer to do the tedious work of marking up all the standard macros for the new index.

JohelEGP:
We can work with https://github.com/JohelEGP/draft/tree/index_macros. Tell me how you'd like to proceed. I think you should open a PR here once you've reviewed that the index and its macros are OK. And until then, we should work on getting to that point on one of our forks, pushing and opening PRs for them.

jensmaurer:
We should have buy-in from @tkoeppe before investing any effort into this.

jensmaurer:
Do we have any exposition-only macros anywhere, @AlisdairM ?

JohelEGP:
There should be none. Is it about my branch https://github.com/JohelEGP/draft/tree/index_macros? I was thinking of weird ALL CAPS names like _`STATICALLY-WIDEN`_ when I drafted it, since I associate ALL CAPS with macros.

jensmaurer:
I think STATICALLY-WIDEN and similar are not really macros, but some special meta-beast that shouldn't appear in a macro index.

In short, it seems we don't need to allow for exposition-only macro names, thus we don't need LaTeX macros dealing with such.

JohelEGP:
It should be fixed now.

AlisdairM:
Checking in: if we want this for C++23 I am prepared to do the work of consistently adding the indexing macros on short notice.  However, I am assuming it is more likely C++26 than adding a new index at this stage.

Eric41293:
@tkoeppe What do you think about this proposal?

[...]

PR #7297, opened 2024-10-08
Create index for macro names

JohelEGP:
See #6005.

### DIFF
--- a/source/back.tex
+++ b/source/back.tex
@@ -159,6 +159,15 @@ Concepts whose name appears in \exposid{italics} are for exposition only.\\}
 }
 
 \clearpage
+\renewcommand{\preindexhook}{The bold page number for each entry is the page
+where the macro is defined.
+Other page numbers refer to pages where the macro is mentioned in the general text.\\}
+{
+\raggedright
+\printindex[macroindex]
+}
+
+\clearpage
 \renewcommand{\preindexhook}{The entries in this index are rough descriptions; exact
 specifications are at the indicated page in the general text.\\}
 {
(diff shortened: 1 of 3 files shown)


AlisdairM:
If this lands, I will do the work to populate the index over the pending holiday season.
I might send Joel a PR for his branch, if we are in favor of this work, so the index can land initially populated.

source/back.tex
```diff
@@ -158,6 +158,15 @@
 \printindex[conceptindex]
 }
 
+\clearpage
+\renewcommand{\preindexhook}{The bold page number for each entry is the page
+where the macro is defined.
+Other page numbers refer to pages where the macro is mentioned in the general text.}
```
AlisdairM, line 164: Needs a trait \\ before the closing brace, to ensure vertical space before the first item in the index.

AlisdairM, line 164: The text describing the index is not accurate, as the LaTeX macros do not support annotating a primary definition.  It is not clear that the primary definition language will be useful for C++ macros either.

I failed to come up with a better introductory paragraph myself though, as all of the other indexes start with specific text that does not apply here either.

jensmaurer: > The text describing the index is not accurate, as the LaTeX macros do not support annotating a primary definition.

Well, for the library concepts index, we do have a bold page number for the primary definition. (I haven't looked at the actual LaTeX macros proposed here.)

JohelEGP: Added `\defmacroname` for where the macro semantics is defined
(might be in prose, not necessarily a `#define`).

PR #7299, opened 2024-10-11
[stoptoken.concepts] Remove redundant 'swappable<Token>' clause from 'stoppable_token' concept

lewissbaker:
The `stoppable_token<Token>` concept requires both `copyable<Token>` to be satisfied and `swappable<Token>` to be satisfied.

However, the `copyable<Token>` requirement already subsumes `movable<Token>`, which subsumes `swappable<Token>`.

Therefore the requirement for `swappable<Token>` can be removed from the `stoppable_token` concept definition with no semantic change.

### DIFF
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -692,8 +692,7 @@ template<class Token>
       { Token(tok) } noexcept;                  // see implicit expression variations\iref{concepts.equality}
     } &&
     @\libconcept{copyable}@<Token> &&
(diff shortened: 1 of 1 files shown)


jensmaurer (approved):

jensmaurer:
@jwakely , what do you think?

CaseyCarter (approved):

jwakely (approved):

PR #7304, opened 2024-10-15
[exec.snd.expos] Fix markup of exposition-only identifiers in `basic-state` constructor

lewissbaker:
The initialization of the exposition-only `recv` and `state`
data-members in the `basic-state` constructor should have marked-up
the member identifiers using \exposid.

Also, the `recv` argument passed to `get-state` was supposed to be
referring to the exposition-only data-member rather than the parameter
of the same name as otherwise it would be referring to a moved-from
object.
So markup the `recv` argument to `get-state` using \exposid as well
to make this (slightly) clearer.


jensmaurer (changes requested):
Using two variables in scope that have the same name except for italics feels unfortunate. But oh well.

Could you please annotate "state" with \exposid in the mem-initializer-list as well?

jensmaurer (approved):

PR #7308, opened 2024-10-16
[except.spec] Exception specifications are declarations

AlisdairM:


source/declarations.tex
```diff
@@ -4422,6 +4422,272 @@
 \indextext{declaration!default argument|)}%
 \indextext{declarator!meaning of|)}
```
jensmaurer, line 4423: This should be after the added text.

source/declarations.tex
```diff
@@ -4422,6 +4422,272 @@
...
 
+\rSec3[except.spec]{Exception specifications}%
```
jensmaurer, line 4425: I'm a bit uneasy putting this under "meaning of declarators", which focuses on the recursive decomposition.

Right after "function definitions" (two levels up) feels better to me.   We talk an awful lot about a "function declaration having a non-throwing exception declaration" in the text, and a declarator isn't a complete declaration, so this feels like the wrong spot.

jensmaurer (changes requested):

source/declarations.tex
AlisdairM: How about after [dcl.fct.spec] instead?

jensmaurer: This is the section about _decl-specifier_s. A _noexcept-specifier_ isn't a _decl-specifier_. It's more like an attribute, considering its trailing placement.

AlisdairM:
Is this PR still viable for C++26, moving a subclause (verbatim) from clause 14, exceptions, to clause 9, delcarations?

If not, please give it the C++29 milestone.
I will continue to respond to Jens feedback so this PR can be ready to land whenever it is appropriate (hopefully still for C++26!)

tkoeppe:
I'm not sure why this change should be tied to any particular publication schedule? We should discuss it on the merits, and then either proceed or drop it.

I'm missing a bit of a motivation and justification here, beyond the Core feedback that Jens has already started giving. Could you update the opening post with some justification?

source/declarations.tex
AlisdairM: After carefully hunting through [decl], I think the best place for this right now is immediately following [dcl.fct] that has the grammar for the optional _noexcept-specifier_ on a function (type) declaration.

I will get this branch into a non-conflicting state, and suggest we discuss further in Wrocław.

jensmaurer:
Instead of piecewise moving meat from [except], I think the logical next step is to actually dissolve [except] and distribute its contents to where it belongs.  Which is a clause re-org, for C++29.

tkoeppe:
If you think that it's realistic to fully dissolve the clause, then do feel free to start a new issue with C++29 milestone!

AlisdairM:
Issue opened, tentatively aiming for C++29.  See #7317.

AlisdairM:
Now that we have a new issue tracking the C++29 work, I have created a new branch with the full rework and dissolve: #7320.  Further reviews and progress will happen on that PR.

PR #7309, opened 2024-10-16
[charconv.syn] Use higher-level indexing macros

jensmaurer:
This avoids false positives with the automated checks.


tkoeppe:
Thanks! Would you by any chance be able to reapply this in text.tex which I'm about to put this into?

tkoeppe:
I've got it, thanks!

PR #7321, opened 2024-10-18
[inplace.vector] Move reserve+shrink_to_fit from #modifiers to #capacity

Quuxplusone:
> In the synopsis [inplace.vector.overview], reserve() and shrink_to_fit() are under size/capacity, but their actual descriptions are in [inplace.vector.modifiers], not [inplace.vector.capacity].

And the English title of [inplace.vector.capacity] doesn't quite match [vector.capacity], either. I think this is editorial and easy to patch.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -9638,7 +9638,7 @@ namespace std {
     constexpr const_reverse_iterator crbegin() const noexcept;
     constexpr const_reverse_iterator crend()   const noexcept;
 
-    // \ref{inplace.vector.capacity}, size/capacity
+    // \ref{inplace.vector.capacity}, capacity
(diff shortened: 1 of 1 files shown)


jensmaurer (approved):

jensmaurer:
@jwakely , what do you think?

jensmaurer:
@jwakely , ping.

Quuxplusone:
Data point: I, also, think it's weird that `resize` isn't under "modifiers," and would be totally fine if someone went and fixed _all the other_ containers to put `resize`, `reserve`, and `shrink_to_fit` under "modifiers" instead. :)  But consistent one way or the other.

PR #7323, opened 2024-10-19
[basic.align] Move the Alignment clause adjacent to the Object model

AlisdairM:
[Edit]
Simplified to move just the Alignment subclause.

[Original, for reference]
Reorder the subclauses of [basic.memobj], Memory and objects, into a logical progression where each clause augments the one before.

* Memory model specifies the notion of memory for C++
* Object model specifies objects, and how they occupy memory
* Alignment restricts where objects can reside in memory
* Storage duration describes how objects acquire well-aligned memory
* Intermediate and erroneous values specifies the initial contents of acquired memory
* Lifetime describes the validity of objects within their acquired storage
* Temporary objects have a lifetime of their own.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3453,6 +3453,100 @@ Some functions in the \Cpp{} standard library implicitly create objects%
 \end{note}
 \indextext{object model|)}
 
+\rSec2[basic.align]{Alignment}
+
+\pnum
+Object types have \defnx{alignment requirements}{alignment requirement!implementation-defined}\iref{basic.fundamental,basic.compound}
+which place restrictions on the addresses at which an object of that type
+may be allocated. An \defn{alignment} is an \impldef{alignment}
+integer value representing the number of bytes between successive addresses
+at which a given object can be allocated. An object type imposes an alignment
+requirement on every object of that type; stricter alignment can be requested
+using the alignment specifier\iref{dcl.align}.
+Attempting to create an object\iref{intro.object} in storage that
+does not meet the alignment requirements of the object's type
+is undefined behavior.
+
+\pnum
+A \defnadj{fundamental}{alignment} is represented by an alignment
+less than or equal to the greatest alignment supported by the implementation in
+all contexts, which is equal to
+\tcode{\keyword{alignof}(std::max_align_t)}\iref{support.types}.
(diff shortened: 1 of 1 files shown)


jensmaurer:
I don't think micro-optimizing various small-ish subclauses in the standard is a good way to spend our time.

That said, I disagree with the suggested ordering.  In particular, storage duration is more about when the destructor is automatically called in various circumstances (or not at all), and not so much about the actual duration of the storage. I think we first should know about lifetime in general before we dig into that.

Further, indeterminate / erroneous values is mostly discussing what happens within the lifetime of an object when no initialization has happened, and not about storage without in-lifetime objects.

In short, I don't find the current ordering "obviously wrong".

AlisdairM:
How would you feel about moving just the alignment clause to follow Object model?

jensmaurer:
> How would you feel about moving just the alignment clause to follow Object model?

I'm ok with that.  Alignment is an aspect of the object model, I think.

jensmaurer (commented):
Please rebase; there's a stray commit here.

AlisdairM:
I believe the stay commit _is_ a consequence of rebasing?

jensmaurer:
I hope that isn't true.  I just merged the change that is your second commit to "main", so I'm a bit surprised to see it pop up here (again).

AlisdairM:
I just rebased, and the commit ID of that intermediate commit has changed.  I rebased again and that should have fixed it.  I have no idea what the commit ID would change, but I guess that is what I get for jumping on a change that just landed?

tkoeppe:
Please simplify the commit description, too (or at least during squashing).

AlisdairM:
The commit description has been greatly simplified.  I retained the original description above _only_ in the GitHub conversation, so that Jens's comments retained their context.

AlisdairM:
Oh, unless you mean the one-line description, I can fix that too.

AlisdairM:
I hope that the PR is now in a simple and clean state, ready for landing.

AlisdairM:
This should be properly rebased again.

tkoeppe:
Applied via 2981bd94f25ea2199fd6b8af7aa76e03cf427697, thank you!

Issue #6523, opened 2023-08-27
[dcl.constinit] p1 "Declaration of a variable" is broken wording


Eisenwave:
> The `constinit` specifier shall be applied only to a declaration of a variable with static or thread storage duration.

\- [[dcl.constinit] p1](https://eel.is/c++draft/dcl.constinit#1)

This is broken wording because [[basic.pre]](https://eel.is/c++draft/basic.pre#def:variable) says:
> A *variable* is introduced by the declaration of a reference other than a non-static data member or of an object.
The variable's name, if any, denotes the reference or object.

What the former paragraph says is thus:
> The `constinit` specifier shall be applied only to a declaration of a declaration of an object or reference other than a non-static data member or of an object, with static or thread storage duration.

Inlining the definition gives us "declaration of a declaration" which is either meaningless or recursive.

Eisenwave:
Actually this wording issue extends beyond that; there are 11 uses of "declaration of a variable" in the standard right now.

JohelEGP:
> Inlining the definition gives us "declaration of a declaration" which is either meaningless or recursive.

I think this is wrong.
A variable is not a declaration.
It is introduced by a declaration.

Eisenwave:
> A variable is not a declaration.
> It is introduced by a declaration.

Honestly that might make the wording issue even more severe, because it's unclear what a variable even is. In the declaration `int x`, a *variable* is introduced ... but what exactly is the variable?

The variable isn't the thing which is declared (the object or reference is) for sure, so *declaration of a variable* is meaningless wording.

JohelEGP:
> that might make the wording issue even more severe

I'm not suggesting a change here.
Just pointing out that variable isn't defined to be a declaration.
But alluding to its definition as a term, that it is introduced by one.
So I think that your argument I quoted is wrong, even if the other parts of the issue aren't.

jensmaurer:
We have a long-standing problem with the wording around "variable" that some variables can denote different objects at different times.

For example, `thread_local int x;" denotes a different object in each thread.
A local variable "int y;" denotes a different object in each (possibly recursive) invocation.

I don't think the cosmetic adjustment from "declaration of a variable" to "variable declaration" is worth the bother, but anyone tackling the full problem (see above) is most welcome to do so; probably that should be in the form of a paper.

RealLitb:
[...]

PR #7330, opened 2024-10-20
[basic.stc.dynamic.general] Fix definition/explanation of dynamic storage duration

eisenwave:
As part of https://github.com/cplusplus/draft/issues/7261.

This PR fixes numerous issues with [[basic.stc.dynamic.general] p1](https://eel.is/c++draft/basic.stc.dynamic.general#1):

1. "dynamically" is not defined in the standard, and the paragraph does not define it (judging by formatting), so this term is meaningless.

2. The paragraph specifically states "during program execution", but C++20 allows the user to create dynamic storage duration objects during *program translation*, i.e. within constant evaluation.

3. The explanation given here is not exhaustive. Normative wording in [[basic.stc.general] p2](https://eel.is/c++draft/basic.stc.general#2.sentence-2) already states "The dynamic storage duration is associated with objects created by a *new-expression* ([expr.new]) or with implicitly created objects ([intro.object]).". Notably, implicitly created objects are not mentioned in [basic.stc.dynamic.general].

It is worth noting that the new definition explicitly gives dynamic storage duration to *objects*, which is inconsistent with the other definitions. For example, [[basic.stc.static]](https://eel.is/c++draft/basic.stc.static) says:

> All variables which [...] have *static storage duration*.

@jensmaurer has stated in https://github.com/cplusplus/draft/issues/6523#issuecomment-1694706321 that we have long-standing problems with the term "variable". At first sight, this may be a blocker for this edit.

However, this editorial issue does not introduce a novel problem because [[expr.new] p10](https://eel.is/c++draft/expr.new#10) already states:

> Objects created by a *new-expression* have dynamic storage duration ([basic.stc.dynamic]).

In other words, this PR makes no normative changes. It aggregates existing wording from [expr.new] and [basic.stc.general] and replaces old incorrect or incomplete wording.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3992,12 +3992,16 @@ specified in~\ref{class.copy.elision}.
 \indextext{storage duration!dynamic|(}
 
 \pnum
-Objects can be created dynamically during program
-execution\iref{intro.execution}, using
+Objects created using
 \indextext{\idxcode{new}}%
-\grammarterm{new-expression}{s}\iref{expr.new}, and destroyed using
+\grammarterm{new-expression}{s}\iref{expr.new}
+and implicitly created objects\iref{intro.object}
+have \defnadj{dynamic}{storage duration}.
+Objects created using \grammarterm{new-expression}{s}
+may be destroyed using
 \indextext{\idxcode{delete}}%
-\grammarterm{delete-expression}{s}\iref{expr.delete}. A \Cpp{} implementation
+\grammarterm{delete-expression}{s}.
+A \Cpp{} implementation
 provides access to, and management of, dynamic storage via
 the global \defnx{allocation functions}{allocation function}
 \tcode{\keyword{operator} \keyword{new}} and


source/basic.tex
```diff
@@ -3989,15 +3989,15 @@
 \rSec3[basic.stc.dynamic]{Dynamic storage duration}%
 
 \rSec4[basic.stc.dynamic.general]{General}%
-\indextext{storage duration!dynamic|(}
```
jensmaurer, line 3992: Why is this being removed?

source/basic.tex
```diff
@@ -3989,15 +3989,15 @@
...
 
 \pnum
-Objects can be created dynamically during program
-execution\iref{intro.execution}, using
+Object with \defnadj{dynamic}{storage duration} are created using
```
jensmaurer, line 3994: This seems an English grammar bug: "Object ... are"

jensmaurer, line 3994: Can we reverse the phrasing here?  The storage duration is a consequence of the syntax used to create the object, not the other way round: "Objects created using X and Y have dynamic storage duration."  That seems to be the way we phrase it for the other storage durations (never mind object vs. variable).

jensmaurer (changes requested):

source/basic.tex
Eisenwave: The PR adds `\defnadj` within the paragraph below, which presumably obsoletes this.

source/basic.tex
Eisenwave: Done. The only reasonable way to accommodate the note on *delete-expression*s was to pull it into a completely separate sentence though, which is why I originally refrained from doing that.

I think it's overall for the better now though; good suggestion.

source/basic.tex
jensmaurer: There's a similar thing with a closing paren at the end of the section (I think), which pairs with this one.  You can't just remove one and leave the other. I'd like to point out that remove those parens gets rid of a "n-m" index entry, which we might want to keep.

eisenwave: I've added it back in.

PR #7333, opened 2024-10-21
Use \libglobal for indexing

jensmaurer:


jensmaurer:
Bad output; needs more work.

jensmaurer:
Bad output was fixed.

jensmaurer:
"great" is the new normal, I guess.

PR #7336, opened 2024-10-21
[lex.header] Modernize text around header names

AlisdairM:
The footnote better belongs in the main text as a regular note.                 
To make the notes flow consistently, switch the order of the                    
note and normative text in the first paragraph to lead with the                  
normative text.

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -661,16 +661,16 @@ violates a constraint on increment operators, even though the parse
 \end{bnf}
 
 \pnum
+The sequences in both forms of \grammarterm{header-name}{s} are mapped in an
+\impldef{mapping header name to header or external source file} manner to headers or to
+external source file names as specified in~\ref{cpp.include}.
 \begin{note}
-Header name preprocessing tokens only appear within
+Header name preprocessing tokens appear only within
 a \tcode{\#include} preprocessing directive,
 a \tcode{__has_include} preprocessing expression, or
 after certain occurrences of an \tcode{import} token
 (see~\ref{lex.pptoken}).
 \end{note}
-The sequences in both forms of \grammarterm{header-name}{s} are mapped in an
-\impldef{mapping header name to header or external source file} manner to headers or to
-external source file names as specified in~\ref{cpp.include}.
 
 \pnum
 The appearance of either of the characters \tcode{'} or \tcode{\textbackslash} or of
@@ -680,12 +680,12 @@ is conditionally-supported with \impldef{meaning of \tcode{'}, \tcode{\textbacks
 \tcode{/*}, or \tcode{//} in a \grammarterm{q-char-sequence} or an
 \grammarterm{h-char-sequence}} semantics, as is the appearance of the character
 \tcode{"} in an \grammarterm{h-char-sequence}.
-\begin{footnote}
+\begin{note}
 Thus, a sequence of characters
 that resembles an escape sequence can result in an error, be interpreted as the
 character corresponding to the escape sequence, or have a completely different meaning,
 depending on the implementation.
-\end{footnote}
+\end{note}
 \indextext{header!name|)}
 
 \rSec1[lex.ppnumber]{Preprocessing numbers}


source/lex.tex
```diff
@@ -660,31 +660,36 @@
...
+\tcode{/*} or \tcode{//} in a \grammarterm{q-char-sequence} is
+conditionally-supported with
+\impldef{meaning in a \grammarterm{q-char-sequence} of \tcode{'},
+\tcode{\textbackslash}, \tcode{/*}, or \tcode{//}} semantics.
+The appearance of any of the characters \unicode{0022}{quotation mark},
+\unicode{0027}{apostrophe}, \unicode{005c}{reverse solidus}, or of either of the
+character sequences \tcode{/*} or \tcode{//} in a \grammarterm{h-char-sequence}
+is conditionally-supported with
+\impldef{meaning in a \grammarterm{h-char-sequence} of \tcode{""}, \tcode{'},
+\tcode{\textbackslash}, \tcode{/*}, or \tcode{//}} semantics.
```
jensmaurer, line 686: I like the \unicode stuff, I like the reordering, I like the footnote -> note change.

I don't like repeating the list of characters.  Can we just extend the original \impldef to say "... in a q-char-sequence or h-char-sequence" ?  Yes, we omit " that way, but the text in the impldef annex doesn't need to be comprehensive.  Or maybe say "meaning of certain characters in a q-char-sequence or h-char-sequence" to cover everything?

jensmaurer (changes requested):

source/lex.tex
AlisdairM: I prefer this duplication, as it will allow me to file a CWG issue that neither of the two comment forms can appear in a _h-char-sequence_ as they will be processed as comments, unlike in the _q-char-sequence_.

jensmaurer: > I prefer this duplication, as it will allow me to file a CWG issue that ...

No amount of editorial refactoring can affect the validity of a CWG issue submission. Either there is a CWG-level problem, or there isn't, or the supposedly-editorial refactoring isn't actually editorial.

AlisdairM: Yes, my edit above is purely editorial.  However, the CWG issue I want to file will make it clear that the sets of characters have even less of an overlap, as the two comment-openers are an issue for only _h-char_sequences_ and the embedded quote is a problem for only _q-char-sequences_, and the overlap is just single-quote and backslash.

I can file the CWG issue without this refactoring, but it would look much simpler with it, and I believe the refactored version that removes the comment-openers from _h-char-sequence_ consideration would also be our preferred final outcome.

I can switch back to a common sentence form is that remains your preferred rendering before I file the CWG issue though.

jensmaurer: May I suggest to limit this editorial pull request to these non-contentious items:

> I like the \unicode stuff, I like the reordering, I like the footnote -> note change.

After the merge, you'll post your CWG issue, and its suggested resolution (by happenstance) can disentangle the sentence to give \impldef some joy.

AlisdairM: Posted the more minimal edit, and changed the commit message accordingly.  Updated the description at the top of this PR too.

source/lex.tex
```diff
@@ -660,31 +660,32 @@
 \end{bnf}
 
 \pnum
+The sequences in both forms of \grammarterm{header-name}{s} are mapped in an
+\impldef{mapping header name to header or external source file} manner to headers or to
+external source file names as specified in~\ref{cpp.include}.
```
cor3ntin, line 665: Can we get rid of "external" or would that not be editorial @jensmaurer ?

AlisdairM:
Flagging that I believe I have addressed the changes requested by Jens.

jensmaurer (approved):

AlisdairM:
Removed the last part of the Unicode markup changes from the original commit message:
" Finally, take this opportunity to clarify the  specific characters called out using the Unicode nomenclature  used throughout this clause."

jensmaurer (approved):

PR #7342, opened 2024-10-22
[basic.link] Add oxford comma

Eisenwave:
The lack of commas in [[basic.link] p4](https://eel.is/c++draft/basic.link#4) really had me stumped for a few minutes because I thought it should be read as:
```
The name of an entity that belongs to a
    namespace scope that has not been given internal linkage above
and that is the name of ...
```
... when the intended reading is
```
The name of an entity
- that belongs to a namespace scope,
- that has not been given internal linkage above, and
- that is the name of ...
```
This PR adds an Oxford comma to make it immediately obvious which reading is intended.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -2714,8 +2714,8 @@ can have external or module linkage, even if not declared \keyword{extern}.
 \pnum
 An unnamed namespace or a namespace declared directly or indirectly within an
 unnamed namespace has internal linkage. All other namespaces have external linkage.
-The name of an entity that belongs to a namespace scope
-that has not been given internal linkage above
+The name of an entity that belongs to a namespace scope,
+that has not been given internal linkage above,
 and that is the name of
 \begin{itemize}
 \item a variable; or


jensmaurer (approved):

tkoeppe:
The PR seems to do more than add an Oxford comma, and also adds an interior comma.

@jensmaurer Is it actually clear that this is editorial? The original wording could be taken to mean that that only "namespace scopes with internal linkage" are being considered. I realize that's probably not intended, but it sounds like that could be a valid reading.

Eisenwave:
It would not be a meaningful reading (albeit grammatically valid) because [[basic.link] p4.9](https://eel.is/c++draft/basic.link#4.9) states:

> otherwise, if the enclosing namespace has internal linkage, the name has internal linkage;

This is nonsensical if we read the whole paragraph as

> [...] namespace scope that has not been given internal linkage above [...]

... because this bullet would never apply to anything.

Furthermore, the word "above" would have to apply to *the sentence* above, not to a prior paragraph above, which is more typical.

jensmaurer:
@tkoeppe, I agree the other reading is possible, but it yields a meaningless statement, so I think this is one of those "pellucidly obvious" cases.

tkoeppe:
OK, makes sense, thanks!

PR #7345, opened 2024-10-22
[basic.stc.inherit] Dissolve paragraph into [basic.stc.general]

AlisdairM:
The whole subclause [basic.stc.inherit] is a single sentence that belongs adjacent to the material in [basic.std.general] that specifies how entities acquire a storage duration, wheras all the remaining subclauses below [basic.stc] describe specific storage durations.  Folding that sentence directly into the general clause is even clearer.

### DIFF
--- a/source/xrefdelta.tex
+++ b/source/xrefdelta.tex
@@ -95,6 +95,9 @@
 \movedxref{stoptoken.cons}{stopsource}
 \movedxref{stoptoken.nonmembers}{stopsource}
 
+% https://github.com/cplusplus/draft/pull/7345
+\movedxref{basic.stc.inherit}{basic.stc.general}
+
 %%% Deprecated features.
 %%% Example:
 %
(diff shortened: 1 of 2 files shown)


source/xrefdelta.tex
```diff
@@ -95,6 +95,9 @@
 \movedxref{stoptoken.cons}{stopsource}
 \movedxref{stoptoken.nonmembers}{stopsource}
 
+% https://github.com/cplusplus/draft/pull/7???
```
jensmaurer, line 98: please fix the number

jensmaurer (changes requested):
Mostly harmless, I'd say.

source/xrefdelta.tex
AlisdairM: Done

jensmaurer (approved):

PR #7346, opened 2024-10-23
[lex.pptoken] Do not use terms before they are defined

AlisdairM:
This PR purely moves existing words around, and does not create any new content.  First move p1 below p2, so that we do not refer to preprocessing tokens before they are defined.  Then move p4 up, as it is splitting some unrelated examples, neither of which use its contents.


jensmaurer (approved):
Good cleanup.

PR #7338, opened 2024-10-21
[locale.codecvt.virtuals] Fix garbled sentence

jensmaurer:
Fixes #7337

PR #7347, opened 2024-10-23
[locale.codecvt.virtuals] Improve wording

cpplearner:
Pre-C++98 draft N1146 [[lib.locale.codecvt.virtuals]/2](https://www.open-std.org/jtc1/sc22/wg21/docs/wp/html/nov97-2/lib-locales.html#lib.locale.codecvt.virtuals) mentions:

> [Note: If no translation is needed (returns `noconv`), sets `to_next` equal to argument `to`, and `from_next` equal to argument `from`.  --end note]

[LWG 19](http://cplusplus.github.io/LWG/issue19) proposes to "change the Note in paragraph 2 to normative text", but in doing so, it introduces a few problems:

- grammar error (fixed in #7338);
- the meaning of `noconv` is specified twice (once in this paragraph, once in the table [[tab:locale.codecvt.inout]](https://eel.is/c++draft/locale.codecvt.virtuals#tab:locale.codecvt.inout-row-5));
- the effect on `from_next` is lost.

This PR fixes the last two problems.

### DIFF
--- a/source/text.tex
+++ b/source/text.tex
@@ -2161,10 +2161,8 @@ Stops if it encounters a character it cannot convert.
 It always leaves the \tcode{from_next} and \tcode{to_next} pointers
 pointing one beyond the last element successfully converted.
 If it returns \tcode{noconv},
-\tcode{internT} and \tcode{externT} are the same type, and
-the converted sequence is identical to
-the input sequence \range{from}{from\textunderscore\nobreak next},
 \tcode{to_next} is set equal to \tcode{to},
+\tcode{from_next} is set equal to \tcode{from},
 the value of \tcode{state} is unchanged, and
 there are no changes to the values in \range{to}{to_end}.
 


jwakely:
I think this whole function could do with better specification, at least p3 and the table. If the result is `noconv` then there is no converted sequence, so it doesn't really make sense to talk about "the converted sequence is identical to the input sequence  [`from`, `from_next`)". 

For the noconv case, the converted sequence is hypothetical, i.e. the converted sequence _would be_ identical to the input sequence _if an actual conversion was performed_. But that doesn't happen, no conversion is done.

> It always leaves the `from_next` and `to_next` pointers pointing one beyond the last element successfully converted.

If this was stated more precisely/clearly, we wouldn't need to say what happens to `from_next` and `to_next` in the noconv case.

The addition of "`from_next` is set equal to `from`" isn't clearly sequenced relative to the use of `from_next` earlier in the sentence when referring to "the input sequence [from, from_next)".

This now becomes is very non-editorial, but I'd prefer to get rid of the table and rewrite the effects, e.g. something like

> If `internT` and `externT` are the same type and the converted sequence would be identical to the input sequence [from, from_next), then no elements are converted, the value of `state` is unchanged, there are no changes to the values in [`to`, `to_end`), and the result is `noconv`. Otherwise, if a character in [from,from_end) cannot be converted, conversion stops at that character and the result is `error`.  Otherwise, if all input characters are successfully converted and placed in the output range, the result is `ok`. Otherwise, if the output range [to,to_end) is reached before converting all characters from the input sequence, the result is `partial`. In all cases, `from_next` is set to point to the first element of the input that was not converted, `to_next` is set to point to the first unchanged element in the output. [_Note_: When the result is `noconv`, `from_next` points to `from` and `to_next` points to `to` -- _end note_]

> _Returns_: The result as described above.

This still has lots of room for improvement, but I think it would be better than what we have now, even with this PR.

cpplearner:
I suspect that "the input sequence [from, from_next)" should be "[from, from_end)". It doesn't make sense to use `from_next` since it's an out parameter.

Can we get Nathan Myers (@ncm) to comment on this issue?

tkoeppe:
Nathan concurs with Jon's suggestions above.

Could you file an LWG issue, please?

cpplearner:
Okay

jwakely:
thanks, email to lwgchair received

PR #7351, opened 2024-10-24
[lex.separate] Redistribute second comment

AlisdairM:
The second comment in [lex.separate] is both too specific, and not specific enough.  Move the last sentence about use of separately translated TUs to the end of the first note, where it seems most appropriate.  Move the rest of the note to the end of translation phase 8, where we are in a position to talk about preserving the separately translated TUs *and instantiation units* for linking in phase 9.

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -39,17 +39,8 @@ if any, is
 called a \defnadj{preprocessing}{translation unit}.
 \begin{note}
 A \Cpp{} program need not all be translated at the same time.
-\end{note}
-
-\pnum
-\begin{note}
-Previously translated translation units can be preserved individually or in libraries.
-The separate translation units of a program communicate\iref{basic.link} by (for example)
-calls to functions whose identifiers have external or module linkage,
-manipulation of objects whose identifiers have external or module linkage, or
-manipulation of data files. Translation units can be separately
-translated and then later linked to produce an executable
-program\iref{basic.link}.
+Translation units can be separately translated and then later linked
+to produce an executable program\iref{basic.link}.
 \end{note}
 \indextext{compilation!separate|)}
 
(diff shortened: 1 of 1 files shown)


source/lex.tex
```diff
@@ -39,19 +39,8 @@
...
-units can be preserved individually or in libraries. The separate
-translation units of a program communicate\iref{basic.link} by (for
-example)
-calls to functions whose identifiers have external or module linkage,
-manipulation of objects whose identifiers have external or module linkage, or
-manipulation of data files. Translation units can be separately
-translated and then later linked to produce an executable
-program\iref{basic.link}.
+Translation units can be separately translated and then later linked to
+produce an executable program\iref{basic.link}.
```
Eisenwave, line 43:
```suggestion
Translation units can be separately translated and then linked together to
produce a program\iref{basic.link}.
```
This change would bring the note more in line with the choice of words in https://eel.is/c++draft/basic.link#1

Eisenwave:
"Redistribure" typo in commit message and PR title.

source/lex.tex
AlisdairM: For now I was deliberately restricting myself to moving approved words around.

I have no objections to further refining the existing text, but would rather do that as a follow-up once the text is in its new place.

AlisdairM:
Corrected commit message, rebased, and force-pushed.

source/lex.tex
AlisdairM: For the record, my own rewording would be more like:
"Source files can be separately translated to produce translation units and instantiation units that are later linked to produce an executable program."

jensmaurer: side note: Of course, "instantiation units" no longer exist, really, given that we (must) have interleaved instantiation and constant evaluation.

jensmaurer (approved):

tkoeppe:
This text is subject to active work (P2996), please revisit after motions.

source/lex.tex
tkoeppe: Are we still OK with this move then?

tkoeppe: Or should we wait for Reflection to land first?

jensmaurer:
I think I want reflection to land first.  It's probably not changing this exact sentence, but words in the vicinity.

source/lex.tex
AlisdairM: I think this would have made a good addition for N5001, so that the reflection folks could have had this slightly cleaner base to work from.  As it missed P5001, I suggest we refer the authors of P2996 to look at this and pick up anything they find useful.

AlisdairM:
Redistributed following Reflection landing.  Probably needs re-reviewed rather than inheriting its approved status.

jensmaurer (approved):
Moving words around... Seems fine.

tkoeppe:
Alright then, thanks!

PR #7353, opened 2024-10-25
[bit.cast] Rename "behaviour" to "behavior" for consistency

Andreas-Krug:
There exist 579 "behavior" in the draft, but only this one "behaviour".

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -15152,7 +15152,7 @@ otherwise has an erroneous value.
 \item
 Otherwise, if $b$ is indeterminate, the behavior is undefined.
 \item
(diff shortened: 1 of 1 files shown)


jwakely (approved):

tkoeppe:
My bad, thanks for catching!

FrankHB:
Probably worth noting: we still don't follow the [ISO rule of English spelling](https://www.iso.org/ISO-house-style.html) in this case yet, just the consistency takes over precedence.

tkoeppe:
At least we're using ISO-standard paper sizes!

PR #7356, opened 2024-10-27
[conv.general] Remove duplicate cross-reference to [dcl.init]

Eisenwave:
The previous paragraph already contains the same `\iref{dcl.init}`, and normally, we don't duplicate these.

I'm also not sure whether this `\iref` is all that good in general, but I don't see any obvious improvement.


jensmaurer:
I'd like to keep the cross-reference here, although it might look duplicate, because it's a "normative" pointer to the interpretation of the initialization syntax.  The syntax is different in the preceding paragraph, and it's not necessarily clear that both syntaxes are covered by the same subclause.

PR #7357, opened 2024-10-27
Replace "of [...] storage duration" with "with [...] storage duration"

Eisenwave:
We currently have two different prepositions in relation to "storage duration"
- with [...] storage duration
- of [...] storage duration

For the purpose of uniformity and grepability, we should settle on one convention. To say "with [...] storage duration" is overwhelmingly more common, so it should be preferred.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -4675,7 +4675,7 @@ of reference type is ill-formed.
 
 \pnum
 \begin{note}
-For every object of static storage duration,
+For every object with static storage duration,
 static initialization\iref{basic.start.static} is performed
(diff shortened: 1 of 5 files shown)


jwakely:
No objections from me for the library parts.

tkoeppe:
Could you please put quotation marks into the commit messages to distinguish the phrase in question from the message itself?

jensmaurer (approved):
Works for me.

tkoeppe:
@Eisenwave Ping?

Eisenwave:
@tkoeppe Pong.

PR #7359, opened 2024-10-28
[lex] Better specify whitespace characters

AlisdairM:
This commit defines a grammar term for _whitespace-character_ and uses it consistently where the plain text term whitespace character is used.  A whitespace character is defined as one of the five characters that are mentioned in the text closest to provifing a defifinition.  The unicode character name is (mostly) consistently used to name these characters, and for consistency, similar changes were made to name unicode characters rather than render specified characters in code font throughout [lex].  The one exception is backslash, which is retained as-is to avoid making more issues for P2348.  Note that this commit is not a replacement for P2348, merely a clearer statement of the existing specification without any normative changes.


source/lex.tex
```diff
@@ -494,6 +503,7 @@
 
 \indextext{token!preprocessing|(}%
 \begin{bnf}
+
```
jensmaurer, line 506: Why this extra empty line?

source/lex.tex
```diff
@@ -178,7 +178,7 @@
 \item
 Adjacent \grammarterm{string-literal} tokens are concatenated\iref{lex.string}.
 
-\item Whitespace characters separating tokens are no longer
+\item \grammarterm{whitespace-character}s separating tokens are no longer
```
jensmaurer, line 181: Starting a sentence with a (lowercase) grammar non-terminal feels odd.  Can we rephrase to avoid that?

jensmaurer (changes requested):
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2348r3.pdf

has a paper touching the same area.  It seems the paper was abandoned by its author.

source/lex.tex
AlisdairM: Carelessness.  Fixed.

source/lex.tex
AlisdairM: I could not find a good way to turn the sentence around, so just started with an extra word, "Any *whitespace-character*s separating..."

AlisdairM:
> http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2348r3.pdf
> 
> has a paper touching the same area. It seems the paper was abandoned by its author.

I refer to this paper in my commit message, and how I leave room for it by deliberately not replacing textual "new-line" with U+000A LINE FEED, in order to leave room for that paper's specific treatment of new-lines and line-breaks.

This PR is entirely editorial, or there is an error in the PR.  I believe P2348 intends to make small but normative changes in our handling of whitespace.  P2348 should be a much simpler paper to update if this PR lands.

Paging @cor3ntin for further comments.

source/lex.tex
```diff
@@ -469,16 +470,25 @@
...
 \indextext{comment|(}%
+\begin{bnf}
+\nontermdef{whitespace-character}\br
+    \unicode{0009}{character tabulation}\br
+    \textnormal{new-line}\br
+    \unicode{000b}{line tabulation}\br
+    \unicode{000c}{form feed}\br
+    \unicode{0020}{space}\br
+\end{bnf}
+
```
jensmaurer, line 482: Please move that to the very top of [lex.pptoken].  While the grammar is used here first, whitespace-characters are much more relevant for separating pptokens than for comments. Yes, we'll have a forward reference then. Hm. Not good, either.

Alternatively, a new subclause under [lex.char] would be appropriate: It's a kind of character (set), after all.
But having a section that contains just grammar is not good.  Can we add some more meat? Maybe not editorially.

jensmaurer (commented):
I don't think I'm ready to add a grammar non-terminal editorially. This needs a nod from CWG.

source/lex.tex
AlisdairM: My first draft (not submitted) did indeed add the grammar to the top of [lex.pptoken], but it felt even more wrong there.

I think a real solution, leaning more in the direction of P2348 but remaining editorial, would be to add a new level above [lex.comment] for "Whitespace" the covers both whitespace characters, comprising the existing [lex.comment], and a new subclause with the grammar and some text describing whitespace characters.  This would be a sibling to [lex.char] and [lex.pptoken], similar to when we introduced [lex.char] for character sets.

source/lex.tex
```diff
@@ -673,13 +680,13 @@
...
 \pnum
-The appearance of either of the characters \tcode{'} or \tcode{\textbackslash} or of
+The appearance of either of the characters \unicode{0027}{apostrophe} or \unicode{005c}{reverse solidus} or of
 either of the character sequences \tcode{/*} or \tcode{//} in a
 \grammarterm{q-char-sequence} or an \grammarterm{h-char-sequence}
 is conditionally-supported with \impldef{meaning of \tcode{'}, \tcode{\textbackslash},
 \tcode{/*}, or \tcode{//} in a \grammarterm{q-char-sequence} or an
 \grammarterm{h-char-sequence}} semantics, as is the appearance of the character
-\tcode{"} in an \grammarterm{h-char-sequence}.
+\unicode{0022}{quotation mark} in an \grammarterm{h-char-sequence}.
```
cor3ntin, line 689: I don't think we want to use unicode notation for printable basic characters (that introduces inconsistencies even within that paragraph)

source/lex.tex
AlisdairM: I have pushed a more substantial update that creates a new Whitespace clause for both whitespace tokens and comments.  With appropriate cross-references this feels like a better approach, and closer to P2348.  The main difference with that paper is that it defines a full grammar for whitespace, whereas my PR is limited to _whitespace-character_.  The full whitespace grammar in P2348 is more detailed, and has a special treatment for line-breaks that my PR deliberately leaves room for.

I believe my changes remain purely editorial with no semantic/technical changes, although Jens correctly points out that introducing a new grammar term should be seen in Core.  And while I would like to make this change for C++26, introducing the new Whitespace clause after our main clause re-org might make this C++29 material.

I have also retained my original branch that this replaces, should we wish to revert to the earlier approach.

source/lex.tex
jensmaurer: I concur regarding this paragraph.  Please revert.

AlisdairM: Reverted and force-pushed.

AlisdairM: Note that this change seems to be an accidental repetition of the same change in #7336 which is marked as approved.  Should we revisit that PR too?

jensmaurer: On one hand, having lone special characters such as ' appear in text as (just) the glyph is a bit non-obvious, so the \unicode stuff helps for that; on the other hand, we don't make an attempt to present anything but glyphs for `//` or `/*`, which feels totally fine to me in isolation.

I'm not really sure which way we should go.  @tkoeppe ?

AlisdairM: As I have reverted that chance in this branch, and it is a duplicate of #7336, I suggest we have that conversation over there and keep the simplified PR here.

Alternatively, I could split #7336 in two so that this question has a PR all of its own.

AlisdairM:
Rebased, and revised wording.
Submitted paper https://wg21.link/P3657R0 for Core Working Group to review. as requested.

That paper is a little simpler than this PR as it merges part of paper P2843 and a resolution to CWG2002 to remove words rather than update them.  This PR does the extra work in the hope we can get CWG approval to accept the substantive  changes editorially.

AlisdairM:
This PR is entirely subsumed by https://wg21.link/p3657 and will be pursued through Core.

PR #7362, opened 2024-11-01
[basic.compound] Update introduction

Eisenwave:
This PR cleans up a few wording discrepancies in the introduction to compound types:

1. We don't like to say that "functions return references". Any such case would mean that the function is declared with reference return type, but the reference would be transformed into an lvalue or xvalue designating the object upon the call. If we just say "a result", we keep it vague and correct.

2. The list of what classes contain tries to be exhaustive without much benefit. We can just say that classes contain class members, which does not miss many things not listed here, such as `static_assert`, `template`, etc.


jensmaurer (approved):
Yeah.  Intro stuff that nobody looks at to get normative guidance.

PR #7366, opened 2024-11-02
[defns.valid] Elaborate on cause in example and add cross-references

Eisenwave:
To be honest, all I wanted to do is add a forward reference to [lib.types.movedfrom] to make the relevant section in the library intro easier to find, but that also required some wording changes to the example to make it less awkward.

### DIFF
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -665,8 +665,10 @@ value of an object that is not specified except that the object's invariants are
 met and operations on the object behave as specified for its type
 
 \begin{example}
-If an object \tcode{x} of type \tcode{std::vector<int>} is in a
-valid but unspecified state, \tcode{x.empty()} can be called unconditionally,
+If an object \tcode{x} of type \tcode{std::vector<int>} enters a
+valid but unspecified state through say,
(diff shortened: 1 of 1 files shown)


jensmaurer:
I think it's irrelevant how the "valid but unspecified state" was reached; you can always call `x.empty()`.  Thus, I think you can just add the list of cross-references and avoid the rephrasing.

jensmaurer:
Your rephrasing gives me the impression that it's important that the "valid but unspecified state" was reached via a move operation, in order to be allowed to call `x.empty()`.  I don't think that impression is helpful.

Eisenwave:
@jensmaurer is it better with the insertion of "say"? This should clarify that move operations are just one way that the state could have been entered.

jensmaurer:
I don't like "say" from a stylistic viewpoint.

You can add a second sentence, though (and leave the first one essentially alone): "The class std::vector<int> is in a valid but unspecified state after being moved from (xrefs)."

jwakely (commented):
I think the original example was better by virtue of being simpler, and I don't think the cross-reference adds anything.

PR #7367, opened 2024-11-03
[ptr.align] Use math-font 'X' for the object

### LINKED ISSUE #7363: [ptr.align] uses uppercase X for an object
The library conventionally uses lowercase letters for objects. This was pointed out by Alisdair during the review (San Diego 2018) but the author said the choice of `X` was based on the wording for `launder`. However, [ptr.launder] uses a math font _X_ which is a little more obvious it's not using our usual convention for types.

Can we either make it `x` or _X_, but not `X`?

jensmaurer:
Fixes #7363

### DIFF
--- a/source/memory.tex
+++ b/source/memory.tex
@@ -839,9 +839,9 @@ template<size_t N, class T>
 
 \pnum
 \expects
-\tcode{ptr} points to an object \tcode{X} of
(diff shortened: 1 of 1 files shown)


jwakely (approved):
Thanks!

PR #7378, opened 2024-11-07
[expr.await] Define 'suspend' as a term of power for coroutines

### LINKED ISSUE #7377: Is `suspended` a term of power for coroutines?
Important parts of the coroutines specification talk about whether a coroutine is suspended or not, but there is no clear definition of what that means.  Or rather, there is no `\defn` that indicates where we define "suspended" as a term of power.

AlisdairM:
This PR fixes #7377.

Many parts of the standard refer to whether a coroutine is suspended as if it were a term of power, but there is no \defn that defines it as such. This PR places the definition on the only use that I could find where a suspended state was not defined in terms of other suspended states, so may be the transition into suspension.  It is likely that folks with a better eye to Core wording will know a better place, that may need a little rewording to sneak in the word 'suspended' to properly define it.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -5437,7 +5437,7 @@ For postfix increment and decrement, see~\ref{expr.post.incr}.
 \indextext{\idxcode{co_await}}%
 
 \pnum
-The \keyword{co_await} expression is used to suspend evaluation of a
+The \keyword{co_await} expression is used to \defnx{suspend}{suspend a coroutine} evaluation of a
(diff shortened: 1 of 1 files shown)


eisenwave:
Marking this CWG. See the linked issue for details.

tkoeppe:
(It looks like there is no way to mark a pull request so that _closing_ it (not merging it) also closes linked issue. When we have the CWG issue, we should close both GitHub items.)

PR #7381, opened 2024-11-09
[lex.pptoken] Separate out keyword placeholders

AlisdairM:
The _preprocessing-token_ term is defined as one of another set of grammar productions, with the three placeholder tokens standing out as specific single tokens.

Moving the placeholder tokens into their own production simplifies both the _preprocessing-token_ and _keyword_ grammar productions, plus it guarantees a consistency for the set of placeholder between both places.  It also makes it easier for [cpp] to directly cross-reference where needed.

Pedantically, the term placeholder token itself is problematic as tokens are a phase 7 concern, where they become keywords.  Within the earlier phases where these placeholders are used the correct term is a preprocessing token, so strictly these should be placeholder preprocessing tokens, which is a mouthful.  So renaming the term to placeholder keyword, which is hopefully less of a concern as there are no keywords in phase 3--6 to confuse with phase 7 keywords.

Finally, clean up the informative list in text of the various kinds of preprocessing tokens, rendering as a list in the source (one kind per line) and adding a cross-reference to each such item.

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -534,29 +534,30 @@ characters \tcode{//} and \tcode{/*} have no special meaning within a
 \indextext{token!preprocessing|(}%
 \begin{bnf}
 \nontermdef{preprocessing-token}\br
+    keyword-placeholder\br
     header-name\br
-    import-keyword\br
-    module-keyword\br
-    export-keyword\br
-    identifier\br
     pp-number\br
+    preprocessing-op-or-punc\br
+    identifier\br
     character-literal\br
     user-defined-character-literal\br
     string-literal\br
     user-defined-string-literal\br
-    preprocessing-op-or-punc\br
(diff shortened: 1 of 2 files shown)


jensmaurer:
Are these "placeholder keywords" or rather "keyword placeholders"?

AlisdairM:
> Are these "placeholder keywords" or rather "keyword placeholders"?

That is a good question.  In phases 3 and 4 the emphasis is on placeholder.  In phase 7 the emphasis is on keyword.  In both cases we are dealing with the same ephemeral. notion of token-but-not-otherwise-representable-in-the-grammar.

I believe the refactoring of my PR is helpful, and am happy if we find better vocabulary to revise it.

AlisdairM:
Thought about it some more, and agreed with Jens.  Applied updated wording.

jensmaurer:
New grammar production -> CWG should have a look.

PR #7386, opened 2024-11-14
[dcl.link] p10 objects vs. entities

xmh0511:
[dcl.link] p10 says: 
>Linkage from C++ to objects defined in other languages and to objects defined in C++ from other languages is implementation-defined and language-dependent[.](https://eel.is/c++draft/dcl.link#10.sentence-1)
Only where the object layout strategies of two language implementations are similar enough can such linkage be achieved[.](https://eel.is/c++draft/dcl.link#10.sentence-2)

The objects should be changed to entities, which includes both objects and functions for which the linkage is applied.

> Linkage from C++ to entities defined in other languages and to entities defined in C++ from other languages is implementation-defined and language-dependent.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -8699,8 +8699,8 @@ which the resulting lvalue refers is considered a C function.
 \pnum
 \indextext{object!linkage specification}%
 \indextext{linkage!implementation-defined object}%
-Linkage from \Cpp{} to objects defined in other languages and to objects
-defined in \Cpp{} from other languages is \impldef{linkage of objects between \Cpp{} and other languages} and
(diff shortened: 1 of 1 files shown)


jwakely:
Please use the usual style as requested in https://github.com/cplusplus/draft/wiki/How-to-submit-an-editorial-issue

xmh0511:
> Please use the usual style as requested in https://github.com/cplusplus/draft/wiki/How-to-submit-an-editorial-issue

updated.

jensmaurer (approved):

tkoeppe:
It is the commit message that we'd like to see updated, not the PR title. (The PR title is not terribly important.)

PR #7387, opened 2024-11-15
[fs.op.remove] clarify Returns element

CaseyCarter:
To avoid confusion as in microsoft/STL#5088.

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -18425,9 +18425,10 @@ A symbolic link is itself removed, rather than the file it
 
 \pnum
 \returns
-\tcode{false} if \tcode{p} did not exist,
-  otherwise \tcode{true}. The signature with argument \tcode{ec}
-  returns \tcode{false} if an error occurs.
+\tcode{true} if a file \tcode{p} has been removed and \tcode{false} otherwise.
+\begin{note}
+Absence of a file \tcode{p} is not an error.
+\end{note}
 
 \pnum
 \throws


source/iostreams.tex
```diff
@@ -18425,9 +18425,11 @@
...
 \pnum
 \returns
-\tcode{false} if \tcode{p} did not exist,
-  otherwise \tcode{true}. The signature with argument \tcode{ec}
-  returns \tcode{false} if an error occurs.
+The signature with argument \tcode{ec}
+returns \tcode{false} if an error occurs.
+If no error occurs,
+returns \tcode{true} if \tcode{p} did exist and has been removed,
+and \tcode{false} if there is no file \tcode{p}.
```
jwakely, line 18432: If no error occurs, returns true if X and false if Y ... and what otherwise?

I realise there is no otherwise, because any other condition would be an error, but the way it's phrased seems to imply some other case could happen.

How about:

The signature with argument `ec` returns `false` if an error occurs.
If no error occurs, returns `true` if `p` was removed, and `false` otherwise.

[Note: If `p` does not exist before the call, the function returns `false` and does not report an error.]


Or keep the original wording and add a note. We have a "this case is not an error" note for `status` in [fs.ops.status].

jwakely: Maybe this would be better:

[Note: Returning `false` without reporting an error means that `p` did not exist.]

CaseyCarter: How about:

_Returns_: `true` if a file `p` has been removed and `false` otherwise. [_Note_: Absence of a file `p` is not an error.-_end note_]

(In before AINAE becomes the new thing for cool C++ kids.)

CaseyCarter: (I _knew_ it was a mistake to touch this paragraph, but I thought I'd try anyway.)

jwakely: I like that. Pithy yet precise.

jwakely (approved):

jensmaurer (approved):

PR #7388, opened 2024-11-16
[conv.rank] Remove mention of `char` from the first bullet

lprv:
According to [[basic.fundamental]/1](https://eel.is/c++draft/basic.fundamental#1), `char` is not a signed integer type.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -5704,9 +5704,8 @@ has the top-level cv-qualifier \keyword{volatile}.
 Every integer type has an \term{integer conversion rank} defined as follows:
 
 \begin{itemize}
-\item No two signed integer types other than \keyword{char} and \tcode{\keyword{signed}
-\keyword{char}} (if \keyword{char} is signed) have the same rank, even if they have
-the same representation.
+\item No two signed integer types have the same rank,
+even if they have the same representation.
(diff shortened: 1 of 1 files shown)


AlisdairM:
This reading sounds strictly correct, but if we are taking that interpretation we should also update [basic.fundamental]p11:
" The character types, bool, the signed and unsigned integer types, and cv-qualified versions (6.8.5) thereof, are
collectively termed integral types."

I have not checked how much of the standard expects `char` to be an integer type --- it would be a big surprise for a lot of the standard library it if were not.

jensmaurer:
@AlisdairM , I don't understand your comment in reaction to the submission.  The core wording says that "integer type" is not just a union of "signed integer type" and "unsigned integer type".  There is more: char and bool fit into neither category, but are integer types.  What's wrong with that?

The submitter is correct that "char" is not a signed integer type, thus does not need to be excluded when talking about "signed integer type".  I'd like to see more of the surroundings to understand that "char" and "un/signed char" have the same rank, per the specification.

AlisdairM:
Never mind -- I triple read before citing my quote, but missed the very first three words "the character types,".  Concern withdrawn with embarrassment.

PR #7396, opened 2024-11-18
[defns.argument] Mention braced-init-list

lprv:


jensmaurer (approved):
Fine with me.

tkoeppe:
That's useful. Function arguments aren't expressions, but initializers, as @geoffromer once put it very succinctly.

PR #7397, opened 2024-11-18
[temp.constr.normal] Use `\cdots` to produce an ellipsis

lprv:


jensmaurer (approved):
This can be confused with a pack expansion, so \cdots is appropriate.

tkoeppe:
The default for comma-separating dots would be AmsMath's `\dotsc`; can we use that?

tkoeppe:
@lprv Friendly ping?

PR #7404, opened 2024-11-23
[lex] Provide unicode name for all control characters

AlisdairM:
This commit does not touch the new-line character as paper P2348.  It resricts itself to consistent use of the unicode character name for space, horizontal tab, and vertical tab. Compared to PR #7359 it deliberately does not touch the grammar that would necessitate a review by core review. The intent is to rebase that PR if this one lands.

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -140,9 +140,9 @@ or \tcode{>}. A partial comment
 would arise from a source file ending with an unclosed \tcode{/*}
 comment.
 \end{footnote}
-Each comment\iref{lex.comment} is replaced by one space character. New-line characters are
+Each comment\iref{lex.comment} is replaced by one \unicode{0020}{space} character. New-line characters are
 retained. Whether each nonempty sequence of whitespace characters other
-than new-line is retained or replaced by one space character is
+than new-line is retained or replaced by one \unicode{0020}{space} character is
 unspecified.
 As characters from the source file are consumed
 to form the next preprocessing token
@@ -477,7 +477,7 @@ The characters \tcode{/*} start a comment, which terminates with the
 characters \tcode{*/}. These comments do not nest.
 \indextext{comment!\tcode{//}}%
 The characters \tcode{//} start a comment, which terminates immediately before the
-next new-line character. If there is a form-feed or a vertical-tab
+next new-line character. If there is a \unicode{0009}{character tabulation} or a \unicode{000b}{line tabulation}
 character in such a comment, only whitespace characters shall appear
 between it and the new-line that terminates the comment; no diagnostic
 is required.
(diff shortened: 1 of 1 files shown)


AlisdairM:
Note that I also did a review of all usage of the word "space" in [lex] and [cpp].  Some uses were not relevant, such as "name space".  Others applied to the contents of literals, which might not be in the basic source encoding.  This PR is considered to completely address the concern for control characters throughout the standard.

source/lex.tex
```diff
@@ -477,7 +477,7 @@
 characters \tcode{*/}. These comments do not nest.
 \indextext{comment!\tcode{//}}%
 The characters \tcode{//} start a comment, which terminates immediately before the
-next new-line character. If there is a form-feed or a vertical-tab
+next new-line character. If there is a \unicode{0009}{character tabulation} or a \unicode{000b}{line tabulation}
```
jensmaurer, line 480: That's wrong. Form feed is 0x0c .

cor3ntin (approved):

source/lex.tex
```diff
@@ -824,7 +824,7 @@
 \end{footnote}
 operators, and other separators.
 \indextext{whitespace}%
-Blanks, horizontal and vertical tabs, newlines, formfeeds, and comments
+Comments, \unicode{0020}{space}s, \unicode{0009}{character tabulation}s and \unicode{0009}{line tabulation}s, \unicode{000c}{form feed}s, and newlines
```
jensmaurer, line 827: I don't think plural works for the way we present Unicode chars.  Please rephrase.

Also, we seem to use "new-line" elsewhere (e.g. near the changes above). Note the hyphen.

cor3ntin:
This looks reasonable and correct

jensmaurer (changes requested):

source/lex.tex
AlisdairM: I specifically copied newlines from the original list, as I had seen both terms in the standard.  Now the I search again, there are only 4 remaining, and half are in code comments.  I will fix this one here, and file a separate PR to fix the rest.

AlisdairM: Rewording for plurals attempted.

source/lex.tex
AlisdairM: Thanks --- a foolish error well spotted.

AlisdairM:
Responded to Jens's comments and force-pushed to avoid badness in the history.

source/lex.tex
AlisdairM: And it turns out the remaining three uses of "newline" are all in library clauses, which outnumber the one use of "new-line" in the library, so I will defer that question for now.  The fix for new-line above is now in this PR though.

source/lex.tex
```diff
@@ -882,7 +882,8 @@
 \end{footnote}
 operators, and other separators.
 \indextext{whitespace}%
-Blanks, horizontal and vertical tabs, newlines, formfeeds, and comments
+Comments and the characters \unicode{0020}{space}, \unicode{0009}{character tabulation},
+\unicode{0009}{line tabulation}, \unicode{000c}{form feed}, and new-line
```
tkoeppe: Why does 0009 appear twice?

AlisdairM: Thought I had caught all typos.  line tabulation should be 000b.

tkoeppe:
note to self: lex.phases, lex.token

tkoeppe:
@jensmaurer, any final thoughts?

PR #7439, opened 2024-11-24
P3247R2 Deprecate the notion of trivial types

### LINKED ISSUE #7409: [2024-11 CWG Motion 5] P3247R2 Deprecate the notion of trivial types
[P3247R2](https://wiki.edg.com/pub/Wg21wroclaw2024/StrawPolls/p3247r2.html) (Deprecate the notion of trivial types

jensmaurer:
Fixes #7409
Fixes cplusplus/papers#1889

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -4959,11 +4959,6 @@ types are collectively called
 Scalar types, trivially copyable class types\iref{class.prop},
 arrays of such types, and cv-qualified versions of these
 types are collectively called \defnadjx{trivially copyable}{types}{type}.
-\label{term.trivial.type}%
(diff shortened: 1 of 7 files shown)


xmcgcg:
[inplace.vector.overview] p6 says:
>  Let `IV` denote a specialization of `inplace_vector<T, N>`. If `N` is zero, then `IV` is both **trivial** and empty.

The “trivial” here also needs to be replaced.

tkoeppe:
@jensmaurer Could you please send a PR for after motions where we address the N=0 case as per the above comment? I think it's fine as editorial, but should be a separate commit.

PR #7447, opened 2024-11-27
P2862R1 text_encoding::name() should never return null values

### LINKED ISSUE #7418: [2024-11 LWG Motion 6] P2862R1 text_encoding::name() should never return null value
[P2862R1](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2862r1.html) (text_encoding::name() should never return null values

jensmaurer:
Fixes #7418 
Also fixes cplusplus/papers#1521


Dani-Hub:
Looks good to me.

tkoeppe (approved):

PR #7450, opened 2024-11-27
Avoid incorrect sentence parsing option on types

mrussoLuxoft:
This PR is related to the **Issue** #7453 proposing multiple changes to clarify the text in [dcl.spec.auto.general]

In **[dcl.spec.auto.general]-p2**, the verbal form
_The type of a parameter-declaration of a function declaration, lambda-expression, or template-
parameter ..._
may be ambiguous in the sense that it may be interpreted both as the type of any of three things (incorrect) or the type of a parameter of any of three things (correct).
In a discussion on _std-discussion_, it was highlighted that the former interpretation does not make sense as a lambda expression has never an explicit type, however it was still agreed that the verbal syntax should be improved (https://lists.isocpp.org/std-discussion/2024/11/2724.php).

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -1771,9 +1771,11 @@ from an initializer.
 
 \pnum
 The type of a \grammarterm{parameter-declaration} of a
-function declaration\iref{dcl.fct},
-\grammarterm{lambda-expression}\iref{expr.prim.lambda}, or
-\grammarterm{template-parameter}\iref{temp.param}
+\begin{itemize}
+  \item function declaration\iref{dcl.fct},
+  \item \grammarterm{lambda-expression}\iref{expr.prim.lambda}, or
+  \item \grammarterm{template-parameter}\iref{temp.param}
+\end{itemize}
 can be declared using
 a \grammarterm{placeholder-type-specifier} of the form
 \opt{\grammarterm{type-constraint}} \keyword{auto}.


source/declarations.tex
```diff
@@ -1770,10 +1770,11 @@
 from an initializer.
 
 \pnum
-The type of a \grammarterm{parameter-declaration} of a
-function declaration\iref{dcl.fct},
+The type of a \grammarterm{parameter-declaration} of
+any of the following three:
```
jensmaurer, line 1774: This isn't an improvement.

I think just adding bullets for the three options is good enough to disambiguate.

jensmaurer (changes requested):

source/declarations.tex
mrussoLuxoft: My opinion is that the proposal with column ":" followed by a list, was anyway able to disambiguate.

However, bullets are for sure a good way, let me amend the commit.

jensmaurer:
Please fix the commit title per https://github.com/cplusplus/draft/wiki/Commit-message-format
and force-push.

mrussoLuxoft:
commit message amended

source/declarations.tex
```diff
@@ -1771,9 +1771,11 @@
 
 \pnum
 The type of a \grammarterm{parameter-declaration} of a
-function declaration\iref{dcl.fct},
-\grammarterm{lambda-expression}\iref{expr.prim.lambda}, or
-\grammarterm{template-parameter}\iref{temp.param}
+\begin{itemize}
+  \item function declaration\iref{dcl.fct},
+  \item \grammarterm{lambda-expression}\iref{expr.prim.lambda},
```
jensmaurer, line 1776: "or" missing at the end.

jensmaurer (changes requested):

source/declarations.tex
mrussoLuxoft: ok, I wasn't sure whether to add, I will refine soon.

source/declarations.tex
```diff
@@ -1771,9 +1771,11 @@
 
 \pnum
 The type of a \grammarterm{parameter-declaration} of a
-function declaration\iref{dcl.fct},
-\grammarterm{lambda-expression}\iref{expr.prim.lambda}, or
-\grammarterm{template-parameter}\iref{temp.param}
+\begin{itemize}
+  \item function declaration\iref{dcl.fct}, or
```
jensmaurer, line 1775: one "or" too many.  The original text did not have an "or" here.

Remember, we're just adding bullets and not otherwise changing the sentence structure.

jensmaurer (changes requested):

source/declarations.tex
mrussoLuxoft: Ok, sure, done

jensmaurer (approved):

PR #7452, opened 2024-11-27
Immediate understanding of non-type templ. param.

mrussoLuxoft:
This PR is related to the **Issue** #7453 proposing multiple changes to clarify the text in [dcl.spec.auto.general]

In **[dcl.spec.auto.general]-p2**, both occurrences of _template-parameter_ may be improved as _non-type template-parameter_.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -1773,7 +1773,7 @@ from an initializer.
 The type of a \grammarterm{parameter-declaration} of a
 function declaration\iref{dcl.fct},
 \grammarterm{lambda-expression}\iref{expr.prim.lambda}, or
-\grammarterm{template-parameter}\iref{temp.param}
+non-type \grammarterm{template-parameter}\iref{temp.param}
 can be declared using
 a \grammarterm{placeholder-type-specifier} of the form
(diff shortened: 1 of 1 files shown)


source/declarations.tex
```diff
@@ -1773,7 +1773,7 @@
 The type of a \grammarterm{parameter-declaration} of a
 function declaration\iref{dcl.fct},
 \grammarterm{lambda-expression}\iref{expr.prim.lambda}, or
-\grammarterm{template-parameter}\iref{temp.param}
+non-type \grammarterm{template-parameter}\iref{temp.param}
```
jensmaurer, line 1776: Since this directly refers to a first-level decomposition of the grammar non-terminal _template-parameter_, I think it's actually clearer without "non-type" (which suggests a prior semantic interpretation instead of a syntax-based grammar match).

mrussoLuxoft: ok, I was also considering this part as a minor point in Issue #7453,
even because parameter-declaration is not possible in case of a type-parameter.

So, I fully trust your opinion to reject it.

Issue #7453, opened 2024-11-27
Clarifications in [dcl.spec.auto.general]


mrussoLuxoft:
Based on the discussion starting with https://lists.isocpp.org/std-discussion/2024/11/2710.php,

I resulted with some changes to propose for the text, as follows:

**1)** in **p2**, the verbal form
_The type of a parameter-declaration of a function declaration, lambda-expression, or template-
parameter ..._
may be ambiguous in the sense that it may be interpreted both as the type of any of three things (incorrect) or the type of a parameter of any of three things (correct).
In the discussion on _std-discussion_, it was highlighted that the former interpretation does not make sense as a lambda expression has never an explicit type, however it was still agreed that the verbal syntax should be improved (https://lists.isocpp.org/std-discussion/2024/11/2724.php).

Corresponding change proposed through **PR** #7450 is:
_The type of a parameter-declaration of one of the following three: a function declaration, lambda-expression, or template-
parameter, ..._

**2)** in **p2**, both occurrences of _template-parameter_ may be improved as _non-type template-parameter_.
Corresponding change has been proposed through **PR** #7452.
I see this is a minor change, not strictly necessary.

Other parts to be added ...TBD...

jensmaurer:
For the third change, do you have an example that highlights the difference?

mrussoLuxoft:
yes :

void foo(int (*)(auto));

A function declaration with a parameter which has function pointer type.
To define the type of the parameter we are "using" a placeholder type, but it is a decl-specifier related to a nested parameter.
Such a code is clearly illegal, it would mimic a template template parameter.

PR #7454, opened 2024-11-27
Avoid ambiguity of placeholder on nested parameter

mrussoLuxoft:
This PR is related to the Issue https://github.com/cplusplus/draft/issues/7453 proposing multiple changes to clarify the text in [dcl.spec.auto.general]

In **[dcl.spec.auto.general]-p2** and **p5**, an additional text about the case of function type, may avoid the general interpretation of using a placeholder type for nested parameters.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -1779,8 +1779,13 @@ a \grammarterm{placeholder-type-specifier} of the form
 \opt{\grammarterm{type-constraint}} \keyword{auto}.
 The placeholder type shall appear
 as one of the \grammarterm{decl-specifier}{s} in
-the \grammarterm{decl-specifier-seq} or
-as one of the \grammarterm{type-specifier}{s} in
+the \grammarterm{decl-specifier-seq}
+directly related to the parameter
+\begin{note}
+  not to a nested parameter in case of
+  function pointer or reference type
+\end{note}
+or as one of the \grammarterm{type-specifier}{s} in
 a \grammarterm{trailing-return-type}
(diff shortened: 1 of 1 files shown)


source/declarations.tex
```diff
@@ -1779,7 +1779,8 @@
 \opt{\grammarterm{type-constraint}} \keyword{auto}.
 The placeholder type shall appear
 as one of the \grammarterm{decl-specifier}{s} in
-the \grammarterm{decl-specifier-seq} or
+the \grammarterm{decl-specifier-seq} or,
+in case of function pointer type,
```
jensmaurer, line 1783: I'm not convinced this is a good change. At the very least, reference-to-function types should be included, too.

jensmaurer (changes requested):

source/declarations.tex
mrussoLuxoft: Proposed a new way, not sure whether for p5 it was really needed to repeat the same note.

jensmaurer:
Hm... I think this entire change is unnecessary. We clearly say "the type of a parameter-declaration of a function declaration" (not "function type").  A nested function type is not a function declaration.

jensmaurer:
For the "IN" decl-specifier-seq, "of" would work well (and be slightly better), but (as you correctly observe), there is no recursion going on via the decl-specifier-seq.

For "THE decl-specifier-seq", I think there is none in sight here where "THE" could refer to other than the one of the parameter-declaration.

PR #7460, opened 2024-11-28
Use template-parameter and template parameter more consistently

cor3ntin:
Try to use template-parameter only when we refer to a grammar construct, and to 'template parameter' everywhere else.

Adopt the same logic to template-argument/template argument.

This change might not be  exhaustive.

The aim is to editorially adopt some of the wording changes made in P2841R5 to ease its review in core.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -773,7 +773,7 @@ is the \grammarterm{declaration} of a \grammarterm{template-declaration}.
 \indextext{class name!\idxcode{typedef}}%
 A \grammarterm{simple-template-id} is only a \grammarterm{typedef-name}
 if its \grammarterm{template-name} names
-an alias template or a template \grammarterm{template-parameter}.
+an alias template or a template template parameter.
 \begin{note}
 A \grammarterm{simple-template-id} that names a class template specialization
 is a \grammarterm{class-name}\iref{class.name}.
@@ -1652,8 +1652,8 @@ naming a structured binding\iref{dcl.struct.bind},
 the specification of the structured binding declaration;
 
 \item otherwise, if $E$ is an unparenthesized \grammarterm{id-expression}
-naming a non-type \grammarterm{template-parameter}\iref{temp.param},
-\tcode{decltype($E$)} is the type of the \grammarterm{template-parameter}
+naming a non-type template parameter\iref{temp.param},
+\tcode{decltype($E$)} is the type of the template parameter
 after performing any necessary
 type deduction\iref{dcl.spec.auto,dcl.type.class.deduct};
 
@@ -3994,7 +3994,7 @@ one or more generic parameter type placeholders\iref{dcl.spec.auto}.
 An abbreviated function template is equivalent to
 a function template\iref{temp.fct}
 whose \grammarterm{template-parameter-list} includes
-one invented type \grammarterm{template-parameter}
+one invented \grammarterm{type-parameter}
 for each generic parameter type placeholder
 of the function declaration, in order of appearance.
 For a \grammarterm{placeholder-type-specifier} of the form \keyword{auto},
@@ -4004,7 +4004,7 @@ For a \grammarterm{placeholder-type-specifier} of the form
 \grammarterm{type-constraint} \keyword{auto},
 the invented parameter is a \grammarterm{type-parameter} with
 that \grammarterm{type-constraint}.
-The invented type \grammarterm{template-parameter} is
+The invented \grammarterm{type-parameter} is
 a template parameter pack
 if the corresponding \grammarterm{parameter-declaration}
 declares a function parameter pack.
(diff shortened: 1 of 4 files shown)


source/templates.tex
```diff
@@ -1146,10 +1143,10 @@
 A template type argument can be an incomplete type\iref{term.incomplete.type}.
 \end{note}
 
-\rSec2[temp.arg.nontype]{Template non-type arguments}
+\rSec2[temp.arg.nontype]{non-type template arguments}
```
jensmaurer, line 1146: Capitalize first letter: "Non-type template arguments"

source/templates.tex
```diff
@@ -4376,7 +4371,7 @@
 \indextext{type concept|see{concept, type}}%
 A \defnx{type concept}{concept!type}
 is a concept whose prototype parameter
-is a type \grammarterm{template-parameter}.
+is a type template-parameter.
```
jensmaurer, line 4374: remove hyphen, too

jensmaurer (changes requested):

jensmaurer (approved):
CWG off-line review from 2024-12-28 to 2025-01-23 passed with no comments. CWG was alerted of the merge in the 2025-01-24 telecon, with no objections.

@tkoeppe, do you have any final comments on this?

source/overloading.tex
```diff
@@ -4100,7 +4099,8 @@
 A \defnx{string literal operator template}{literal!operator!template string}
 is a literal operator template whose \grammarterm{template-parameter-list}
 comprises
-a single non-type \grammarterm{template-parameter} of class type.
+a single \grammarterm{parameter-declaration} that declares a
```
mizvekov, line 4102: Isn't a `parameter-declaration` a grammar production for function parameters? That doesn't seem to fit here, it can't declare an NTTP..

jensmaurer: I disagree.   See the grammar in [temp.param].

mizvekov: I think I understand what this is trying to say now. It's a little awkward that this doesn't give enough context that this is talking about a `\grammarterm{template-parameter}` produced from a single `\grammarterm{parameter-declaration}`. What do you think about phrasing that something along these lines?

".. is a literal operator template whose \grammarterm{template-parameter-list}
comprises a single \grammarterm{template-parameter} produced from a \grammarterm{parameter-declaration}, which declares a ..."

jensmaurer: I don't think we use "produce" in these contexts.  And when you look at [temp.param] and  [temp.pre] and the meaning of "comprises", I can't see how you can misunderstand the sentence here.  Please re-read the text with the a-priori understanding that a _parameter-declaration_ can naturally appear in a _template-parameter-list_, when declaring a non-type template parameter.

source/templates.tex
```diff
@@ -310,17 +310,17 @@
...
-Here, the template \tcode{f} has a \grammarterm{type-parameter}
+Here, the template \tcode{f} has a type template parameter
 called \tcode{T}, rather than an unnamed non-type
-\grammarterm{template-parameter} of class \tcode{T}.
+template parameter of class \tcode{T}.
 \end{example}
-A \grammarterm{template-parameter} declaration shall not
-have a \grammarterm{storage-class-specifier}.
-Types shall not be defined in a \grammarterm{template-parameter}
+A \grammarterm{parameter-declaration} of a \grammarterm{template-parameter}
```
hubert-reinterpretcast, line 321:
```suggestion
The \grammarterm{parameter-declaration} of a \grammarterm{template-parameter}
```

source/declarations.tex
```diff
@@ -4021,7 +4021,7 @@
 An abbreviated function template is equivalent to
 a function template\iref{temp.fct}
 whose \grammarterm{template-parameter-list} includes
-one invented type \grammarterm{template-parameter}
+one invented \grammarterm{type-parameter}
```
hubert-reinterpretcast, line 4024: The removal of "type" here opens the possibility that the _type-parameter_ is a template template parameter for the unconstrained case.

source/overloading.tex
mizvekov: Still, as an implementor, I was a bit confused here.

Another suggestion, using similar-to-existing wording:

".. is a literal operator template whose \grammarterm{template-parameter-list}
comprises a single \grammarterm{parameter-declaration} of a \grammarterm{template-parameter}, which declares a ..."

source/declarations.tex
cor3ntin: I am not sure I understand what you mean

jensmaurer: The grammar for _type-parameter_ covers both type template parameters and template template parameters.  The old wording clearly ruled out template template parameters due to the "type" prefix.  The new wording does not.  You could say "type _type-parameter_" or just leave the existing wording in this one place.

cor3ntin: OH. Right, that's something that should be done in the paper, not here.

PR #7471, opened 2024-12-03
[iterator.requirements.general] Revert `indirectly_writable` to *writ…

### LINKED ISSUE #7470: Partially misapplied 2019 Belfast LWG Motion 9 - [iterator.requirements.general] p1 should keep using *writable*
During the 2019 Belfast meeting LWG motion 9 was accepted to apply [P1878R1](https://wiki.edg.com/pub/Wg21belfast/StrawPolls/P1878R1.pdf). Among the changes [iterator.requirements.general] p1 was changed by the project editors to

> An output iterator `i` has a non-empty set of types that are` indirectly_writable` to the iterator; [...].

Note that the paper instructions said ([P1878R1 p.7](https://wiki.edg.com/pub/Wg21belfast/StrawPolls/P1878R1.pdf#page=7)):

> [ Editor’s note: Change stable name [iterator.concept.writable] to [iterator.concept.indirectly.writable] and globally
replace all occurrences of “`writable`” (the concept) with “`indirectly_writable`”. ]

Problem is that the pre-Belfast working draft [N4835 p.911](https://wg21.link/N4835#page=911) said:

> An output iterator `i` has a non-empty set of types that are *writable* to the iterator; [...].

At this wording part  *writable* does **not** denote the concept `writable`, but is a definition on its own, so this change should not have been applied here. And the application had lead to a problematic situation described in the recently opened issue [LWG 4178](https://cplusplus.github.io/LWG/issue4178).

According to the LWG reflector discussion (and confirmed by @jwakely) we would like to ask to solve that LWG issue editorially by correcting the wording state to that one described in the accepted paper. I will create a PULL request with these effects.

Dani-Hub:
…able* definition

This fixes a misapplication of the 2019 Belfast meeting LWG motion 9 (P1878R1), which erroneously replaced the *writable* definition by the `indirectly_writable` concept (See issue #7470).

Fixes #7470

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -539,7 +539,7 @@ called the
 \defn{value type}
 of the iterator.
 An output iterator \tcode{i} has a non-empty set of types that are
-\libconcept{indirectly_writable} to the iterator;
+\defn{writable} to the iterator;
 for each such type \tcode{T}, the expression \tcode{*i = o}
 is valid where \tcode{o} is a value of type \tcode{T}.
 For every iterator type


jwakely (approved):
The paper only said to replace the uses of the `writable` concept, not all cases of  the word "writable"

jwakely:
The change was part of 333b6eb8c147fce1a9ddcd905ee3bb1a09b91e30 i.e. #3451

It also changed:

```
@ -1258,16 +1272,16 @@ [iterator.concept.writable]
 \pnum
 \begin{note}
 The only valid use of an \tcode{operator*} is on the left side of the assignment statement.
-Assignment through the same value of the writable type happens only once.
+Assignment through the same value of the indirectly writable type happens only once.
 \end{note}
```

but I think this is OK.

CaseyCarter (approved):

jensmaurer (approved):

PR #7475, opened 2024-12-06
[basic.pre,basic.lookup.general] Cleanup definition of term 'name lookup'

### LINKED ISSUE #7472: Definition of "name lookup" is misleading
The definition of "name lookup" is given in [basic.pre]/8:

> Some names denote types or templates. In general, whenever a name is encountered it is necessary to determine whether that name denotes one of these entities before continuing to parse the program that contains it. The process that determines this is called *name lookup* ([basic.lookup]).

This is one purpose of name lookup, not the only one.

I suggest to convert this paragraph to a note and edit it as shown:

> **[*Note:*** Some names denote types or templates. In general, whenever a name is encountered it is necessary to **look it up ([basic.lookup]) to** determine whether that name denotes one of these entities before continuing to parse the program that contains it. ~The process that determines this is called *name lookup* ([basic.lookup]).~ **&mdash;*end note*]**

Then, in [basic.lookup.general], move the second sentence to the beginning and make it the definition.

> ***Name lookup* associates the use of a name with a set of declarations ([basic.def]) of that name.** The name lookup rules apply uniformly to all names (including typedef-names ([dcl.typedef]), namespace-names ([basic.namespace]), and class-names ([class.name])) wherever the grammar allows such names in the context discussed by a particular rule. ~Name lookup associates the use of a name with a set of declarations ([basic.def]) of that name.~

jensmaurer:
Fixes #7472

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -103,11 +103,13 @@ whose corresponding variable is such an entity,
 or the \tcode{*\keyword{this}} object\iref{expr.prim.this}.
 
 \pnum
-Some names denote types or templates. In general,
-whenever a name is encountered it is necessary to determine whether that name denotes
-one of these entities before continuing to parse the program that contains it. The
-process that determines this is called
-\defnx{name lookup}{lookup!name}\iref{basic.lookup}.
+\begin{note}
+Some names denote types or templates.
+In general, whenever a name is encountered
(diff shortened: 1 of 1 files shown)


jensmaurer:
@tkoeppe, this turns tutorial-style wording into a note and moves a sentence a little earlier.  Please have a look.

PR #7481, opened 2024-12-07
[locale.ctype.virtuals] Fix a decade-old typo

cpplearner:
This sentence comes from [LWG 379](https://cplusplus.github.io/LWG/issue379), but it seems that there's a mistake when applying the proposed resolution to the draft standard: the proposed resolution says "For any named ctype category with a ctype&lt;char> facet ctc", but the draft standard says "For any named ctype category with a ctype<char<ins>***T***</ins>> facet ctc".

### DIFF
--- a/source/text.tex
+++ b/source/text.tex
@@ -1615,7 +1615,7 @@ The only characters for which unique transformations are required
 are those in the basic character set\iref{lex.charset}.
 
 For any named \tcode{ctype} category with
-a \tcode{ctype<charT>} facet \tcode{ctc} and
+a \tcode{ctype<char>} facet \tcode{ctc} and
 valid \tcode{ctype_base::mask} value \tcode{M},
 \tcode{(ctc.\brk{}is(M, c) || !is(M, do_widen(c)) )} is \tcode{true}.
 \begin{footnote}


jensmaurer (approved):
Seems a correct change, based on LWG379.

cpplearner:
ping @jwakely

jwakely:
Isn't the entire subject of LWG379 that the standard says charT and should say char? So if applying the issue failed to make that change, what _did_ it do, just add a comma? Maybe the editor misread the note "[Kona: Minor edit. Added a comma after the M for clarity.]" as saying that was the required fix.

jwakely:
Ah it also changed `ctw` to `ctc`

jwakely (approved):
I've convinced my self that this change (as originally proposed by LWG379) is correct. It's strange that  the original "typo" corrected by LWG379 also involved the name "ctw" which does strongly suggest that a ctype facet for **W**ide characters was intended, in contrast to `do_narrow` where it was always `ctype<char>` and `ctc`. If it was just a typo, you would not have expected the name "ctw" to also be a typo.

But since the argument to `ctw.is(M, c)` is a `char` it does make sense that `ctype<charT> ctw` was supposed to be `ctype<char> ctc` all along.

PR #7179, opened 2024-07-28
[dcl.dcl][stmt.stmt] Remove tautonyms from top level stable labels

AlisdairM:
Per comment on issue #5315.
Thanks to Jonathan Wakely for the notion of tautonyms.

Quickly force-pushed a correction to the comment identifying the PR number.

PR #7485, opened 2024-12-12
[stmt,dcl] Rename labels

jensmaurer:


AlisdairM:
Is this a duplicate of #7179?

jensmaurer:
Almost.  I'm not trying to rename the diff.* labels. For @tkoeppe to choose.

tkoeppe:
Yes, only the two labels we discussed, please.

tkoeppe:
@StephanTLavavej What's your take on this? How disruptive is this change? We do have the "Cross-references from C++-Prev" chapter at the end; does that help?

StephanTLavavej:
This isn't terribly disruptive, but I would *really* like for the cross-references to be permanently accumulating, instead of emptied out between Standards. The scenario here is "I'm looking at compiler/library sources that were written ten years ago and are mentioning \[garfield.lasagna\], and I need to find out that it's now \[orange.cat.favorite.food\]". This is rarely needed but when it is, it's really annoying to dig up the info.

If the purges could be retroactively undone (by gathering up each previous Standard's sections and restoring them all), that would be super awesome.

tkoeppe:
Yes, I see your point. @jensmaurer, do you think such an eternal history would be manageable?

jensmaurer:
Well, we already have Annex C with eternal history, so I don't see why not.

Do we want one single list (which doesn't say what happened in each revision of the standard) or per-standard-version lists as in Annex C?  Maybe one single list is most useful.

StephanTLavavej:
A single list would be fine as far as I'm concerned - I don't care when a rename happened, I just need to find the wording.

(And as I mentioned on the reflector, it's okay to omit the rare rename where an abbreviation was embarrassing)

AlisdairM:
Jens has submitted #7561 that should address the need for a long-lived cross-reference of stable label evolution.

tkoeppe:
@jensmaurer Could you please rebase (and also insert a space after the comma in the label list in the commit message)?

jensmaurer:
@tkoeppe , done

PR #7492, opened 2024-12-16
P3222R0 Fix C++26 by adding transposed special cases for P2642 layouts

### LINKED ISSUE #7421: [2024-11 LWG Motion 9] P3222R0 Fix C++26 by adding transposed special cases for P2642 layouts
[P3222R0](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3222r0.html) (Fix C++26 by adding transposed special cases for P2642 layouts

tkoeppe:
Fixes https://github.com/cplusplus/papers/issues/1870.
Fixes #7421.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -13127,7 +13127,7 @@ for some \tcode{size_t} value \tcode{PaddingValue};
 otherwise,
 \begin{codeblock}
 R(a.data_handle(), ReturnMapping(@\exposid{transpose-extents}@(a.mapping().extents()),
-  a.stride(0)), a.accessor())
+  a.mapping().stride(0)), a.accessor())
 \end{codeblock}
 if \tcode{Layout} is \tcode{layout_right_padded<PaddingValue>}
 for some \tcode{size_t} value \tcode{PaddingValue};
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -13066,6 +13066,14 @@ where \tcode{ReturnLayout} is:
 \item
 otherwise, \tcode{layout_left} if \tcode{Layout} is \tcode{layout_right};
 \item
+otherwise, \tcode{layout_right_padded<PaddingValue>} if \tcode{Layout} is\newline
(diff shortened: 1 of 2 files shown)


jwakely (approved):

source/numerics.tex
```diff
@@ -13108,11 +13116,26 @@
...
+R(a.data_handle(), ReturnMapping(@\exposid{transpose-extents}@(a.mapping().extents()),
+  a.mapping().stride(1)), a.accessor())
+\end{codeblock}
+if \tcode{Layout} is \tcode{layout_left_padded<PaddingValue>}
+for some \tcode{size_t} value \tcode{PaddingValue};
+\item
+otherwise,
+\begin{codeblock}
+R(a.data_handle(), ReturnMapping(@\exposid{transpose-extents}@(a.mapping().extents()),
+  a.stride(0)), a.accessor())
```
jwakely, line 13130: Actually, shouldn't this be:

```suggestion
  a.mapping().stride(0)), a.accessor())
```

@mhoemmen

jwakely: If we can fix this editorially now that would save going through the issue process. There is no `stride` member of `mdspan` so the approved wording is nonsensical.

tkoeppe: I see now, yes, makes sense. Will do!

jwakely (approved):
I'm confident the editorial fix is correct.

I hope @mhoemmen will confirm.

source/numerics.tex
mhoemmen: @jwakely @tkoeppe Hi! : - )  Good catch!  Somehow I missed that one for the `layout_right_padded` case, even though I got it right for the `layout_left_padded` case right above it.  `a.mapping().stride(0)` is absolutely what I intended to write there.

jwakely: Thanks!

mhoemmen:
@jwakely I have reviewed the changes and I confirm the fix.  Thank you! : - )

PR #7493, opened 2024-12-17
P3050R2 Fix C++26 by optimizing linalg::conjugated for noncomplex value types

### LINKED ISSUE #7422: [2024-11 LWG Motion 10] P3050R2 Fix C++26 by optimizing linalg::conjugated for noncomplex value types
[P3050R2](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3050r2.html) (Fix C++26 by optimizing linalg::conjugated for noncomplex value types)

tkoeppe:
Fixes https://github.com/cplusplus/papers/issues/1719.
Fixes #7422.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -12808,21 +12808,40 @@ constexpr typename offset_policy::data_handle_type offset(data_handle_type p, si
 \begin{itemdescr}
 \pnum
 Let \tcode{A} be
+\begin{itemize}
+\item
 \tcode{remove_cvref_t<decltype(a.accessor().nested_accessor())>}
-if \tcode{Accessor} is a specialization of \tcode{conjugated_accessor}, and
-otherwise \tcode{conjugated_accessor<Accessor>}.
+if \tcode{Accessor} is a specialization of \tcode{conjugated_accessor}; otherwise,
+\item
+\tcode{Accessor} if \tcode{remove_cvref_t<ElementType>} is an arithmetic type; otherwise,
+\item
+\tcode{Accessor} if the expression \tcode{conj(E)} is not valid for any subexpression \tcode{E}
+whose type \tcode{T} is expression-equivalent to \tcode{remove_cvref_t<ElementType>}
+with overload resolution performed in a context that includes the declaration
+\tcode{template<class T> conj(const T\&) = delete;}; and otherwise,
+\item
+\tcode{conjugated_accessor<Accessor>}.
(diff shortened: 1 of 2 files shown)


source/numerics.tex
```diff
@@ -12808,21 +12808,40 @@
...
+\item
 \tcode{remove_cvref_t<decltype(a.accessor().nested_accessor())>}
-if \tcode{Accessor} is a specialization of \tcode{conjugated_accessor}, and
-otherwise \tcode{conjugated_accessor<Accessor>}.
+if \tcode{Accessor} is a specialization of \tcode{conjugated_accessor}; otherwise,
+\item
+\tcode{Accessor} if \tcode{remove_cvref_t<ElementType>} is an arithmetic type; otherwise,
+\item
+\tcode{Accessor} if the expression \tcode{conj(E)} is not valid for any subexpression \tcode{E}
+whose type \tcode{T} is expression-equivalent to \tcode{remove_cvref_t<ElementType>}
```
jwakely, line 12819: A type cannot be expression-equivalent. I'm not sure what the paper's intent actually was here, E is simply not defined. Should it really be an rvalue, as this wording kinda implies? That doesn't seem right to me. I think we need input from @mhoemenn or LWG.

This also introduces two types `T`, the one that is the type of `E` and the template parameter of the `conj` poison pill, which is a bit confusing, but that problem is already present throughout [linalg.helpers].

```suggestion
\tcode{Accessor} if the expression \tcode{conj(E)} is not valid for any subexpression \tcode{E}
whose type \tcode{T} is expression-equivalent to \tcode{remove_cvref_t<ElementType>}
```

jwakely: I opened #7494 for the T reuse.

tkoeppe: Do let's follow up asap, but I'll merge the motion application as is for now.

jwakely: @mhoemmen in case you missed this one (I think I spelled your username wrong at first and I don't think github resends emails after comments are edited)

tkoeppe: @jwakely Please see https://github.com/cplusplus/draft/pull/7497!

mhoemmen: @jwakely FYI, this issue refers to R2, but it was R3 that was approved by LWG on 2024-10-25.  That might have wording differences.

PR #7495, opened 2024-12-17
P3539R1 Consistent Function Label Naming for Sections

Benio101:
Continuation of #7483.


tkoeppe:
Sorry, where is this coming from, who asked for this? We don't usually change stable labels, since that would defeat half of their purpose.

PR #7496, opened 2024-12-17
[basic.scope.scope] Rename to [basic.scope.general].

Benio101:
Reason: Consistency with other "General" labels, like [any.general].


tkoeppe:
I don't think we want to do that. Who suggested we should change anything?

Benio101:
> I don't think we want to do that. Who suggested we should change anything?

This is a followup of CWG relabeling suggestions.
See: https://lists.isocpp.org/edit/2024/12/1040.php

tkoeppe:
> > I don't think we want to do that. Who suggested we should change anything?
> 
> This is a followup of CWG relabeling suggestions. See: https://lists.isocpp.org/edit/2024/12/1040.php

Sorry, could you elaborate a bit -- that just looks like an email where you suggest to do more relabelling. Is there any request from CWG or the committee to change anything?

jwakely (commented):
This seems to be unnecessary churn, destabilising stable names for questionable benefit.

Benio101:
> Sorry, could you elaborate a bit -- that just looks like an email where you suggest to do more relabelling. Is there any request from CWG or the committee to change anything?

Sorry, I'm a bit confused. It's just a PR from my reflected suggestion. I did not made a paper proposal for this as per recent request: https://github.com/cplusplus/draft/pull/7483#issuecomment-2538929725. Should I make one anyway?

tkoeppe:
No. I think we should just close this. I don't think we want to make this change. Is that OK?

jensmaurer:
We agree not all stable labels are great phrasing-wise. However, changing any of them (because of the fall-out) requires more than a broad editorial consistency argument.

And no, I don't want to see papers suggesting CWG label renames; that's a total waste of time for CWG. I do entertain suggestions to change stable labels from long-term committee members and then informally seek agreement from CWG (as has happened for [stmt.stmt] and [dcl.dcl]), but this should stay a rare exception.

PR #7497, opened 2024-12-17
Fix linalg conj ADL wording

mhoemmen:
@tkoeppe @jwakely Per our offline discussion


tkoeppe:
@jensmaurer for sanity checking?

jensmaurer (approved):
Looks ok.

jwakely (approved):

mhoemmen:
Thankfully, this is how we spelled it in [P3050R3](https://isocpp.org/files/papers/P3050R3.html#wording).

jwakely:
> On the other hand, R3 takes a slightly different wording tack, to speak of "is valid" rather than "is not valid."

See #7506

PR #7500, opened 2024-12-17
[simd.alg] Fix range syntax

Andreas-Krug:


tkoeppe:
@mattkretz PTAL?

tkoeppe:
Taken as be0a25c9a2f2c1f498b0ff84a33c28adae41863e, thanks!

mattkretz:
Oh, that was a search-and-replace mistake (from the extra macros I had in P1928). Thanks for the catch!

PR #7503, opened 2024-12-17
[basic.fundamental] Ensure consistency with [conv.ptr]

jensmaurer:


tkoeppe:
I sense an ulterior motive...

PR #7506, opened 2024-12-19
[linalg.conj.conjugated] Rearrange to match P3050R3

jwakely:
This was the wording requested by LWG and approved in P3050R3, but I mistakenly put P3050R2 in the straw polls.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -12822,7 +12822,7 @@ otherwise,
 if the expression \tcode{conj(E)} is valid for any subexpression \tcode{E}
 whose type is \tcode{remove_cvref_t<ElementType>}
 with overload resolution performed in a context that includes the declaration
-\tcode{template<class U> conj(const U\&) = delete;};
+\tcode{template<class U> U conj(const U\&) = delete;};
 \item
 otherwise,
 \tcode{Accessor}.
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -12812,41 +12812,35 @@ Let \tcode{A} be
 \begin{itemize}
 \item
 \tcode{remove_cvref_t<decltype(a.accessor().nested_accessor())>}
-if \tcode{Accessor} is a specialization of \tcode{conjugated_accessor}; otherwise,
+if \tcode{Accessor} is a specialization of \tcode{conjugated_accessor};
 \item
-\tcode{Accessor} if \tcode{remove_cvref_t<ElementType>} is an arithmetic type; otherwise,
+otherwise,
+\tcode{Accessor} if \tcode{remove_cvref_t<ElementType>} is an arithmetic type;
 \item
-\tcode{Accessor} if the expression \tcode{conj(E)} is not valid for any subexpression \tcode{E}
+otherwise,
+\tcode{conjugated_accessor<Accessor>}
+if the expression \tcode{conj(E)} is valid for any subexpression \tcode{E}
 whose type is \tcode{remove_cvref_t<ElementType>}
 with overload resolution performed in a context that includes the declaration
-\tcode{template<class T> conj(const T\&) = delete;}; and otherwise,
+\tcode{template<class U> conj(const U\&) = delete;};
 \item
-\tcode{conjugated_accessor<Accessor>}.
+otherwise,
+\tcode{Accessor}.
 \end{itemize}
 
 \pnum
 \returns
+Let \tcode{MD} be \tcode{mdspan<typename A::element_type, Extents, Layout, A>}.
 \begin{itemize}
 \item
-The value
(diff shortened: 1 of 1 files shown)


jwakely:
@mhoemmen another fix for an incorrect straw poll

source/numerics.tex
```diff
@@ -12812,41 +12812,35 @@
...
+\tcode{Accessor} if \tcode{remove_cvref_t<ElementType>} is an arithmetic type;
 \item
-\tcode{Accessor} if the expression \tcode{conj(E)} is not valid for any subexpression \tcode{E}
+otherwise,
+\tcode{conjugated_accessor<Accessor>}
+if the expression \tcode{conj(E)} is valid for any subexpression \tcode{E}
 whose type is \tcode{remove_cvref_t<ElementType>}
 with overload resolution performed in a context that includes the declaration
-\tcode{template<class T> conj(const T\&) = delete;}; and otherwise,
+\tcode{template<class U> conj(const U\&) = delete;};
```
tkoeppe, line 12825: It's a shame we're losing the penultimate "and", though, isn't it?

mhoemmen: I wouldn't mind if you put the penultimate "and" back in : - ) .

mhoemmen (approved):
I compared this against [P3050R3](https://isocpp.org/files/papers/P3050R3.html#wording); it looks good!  Thanks! : - )

source/numerics.tex
tkoeppe: @jwakely Thoughts?

mhoemmen: Actually, thinking about it, I don't mind the lack of a penultimate "and" here.  This is an OR list (disjunction), not an AND list (conjunction).

jensmaurer: We don't put "and"s into "if... otherwise" ladders, which this is.

source/numerics.tex
```diff
@@ -12812,41 +12812,35 @@
...
-\tcode{a} if \tcode{remove_cvref_t<ElementType>} is an arithmetic type; otherwise,
+\tcode{MD(a.data_handle(), a.mapping(), a.accessor().nested_accessor())}
+if \tcode{Accessor} is a\newline specialization of \tcode{conjugated_accessor};
 \item
-\tcode{a} if the expression \tcode{conj(E)} is not valid for any subexpression \tcode{E}
-whose type is \tcode{remove_cvref_t<ElementType>}
-with overload resolution performed in a context that includes the declaration
-\tcode{template<class T> conj(const T\&) = delete;}; and otherwise,
+otherwise,
+\tcode{a}, if \tcode{is_same_v<A, Accessor>} is \tcode{true};
```
jensmaurer, line 12840: How is this editorially-obviously related to the struck text talking about `conj(E)` ?

jwakely: You need to read the preceding paragraph too. In R2 (as applied to the draft) the text in p2 talking about conj(E) was duplicating exactly the condition in 1.3, which was just unnecessarily verbose and repetitive.

We don't need to repeat exactly the two conditions in 1.2 and 1.3 as 2.2 and 2.3, we can just say that if the type `A` determined in p2 is `Accessor`, then we return `a`.

So this is replacing everything in p2 that repeats conditions from p1 by specifying the return value in terms of what was determined in p1.

jwakely: In other words, previously we had:

- (1.2) `Accessor` if `remove_cvref_t<ElementType>` is an arithmetic type; otherwise,

- (1.3) `Accessor` if the expression `conj(E)` is not valid for any subexpression E whose type `T` is expression-equivalent to `remove_cvref_t<ElementType>` with overload resolution performed in a context that includes the declaration `template<class T> conj(const T&) = delete;`; otherwise,

and 

- (2.2) `a` if `remove_cvref_t<ElementType>` is an arithmetic type; otherwise,

- (2.3) `a` if the expression `conj(E)` is not valid for any subexpression E whose type `T` is expression-equivalent to `remove_cvref_t<ElementType>` with overload resolution performed in a context that includes the declaration `template<class T> conj(const T&) = delete;`; otherwise,

But now p2 just says we return `a` if it's already the right type and doesn't need to be transformed.

jensmaurer (approved):
Thanks for the explanation.

source/numerics.tex
timsong-cpp, line 12825:
```suggestion
\tcode{template<class U> U conj(const U\&) = delete;};
```

source/numerics.tex
mhoemmen: @jwakely Hi!  Is there anything else you would like me to do so we can merge this?  Thanks! : - )

PR #7515, opened 2024-12-23
Fix concept markup

jensmaurer:


tkoeppe:
Thanks, that's great!

PR #7517, opened 2024-12-25
[tab:cpp17.destructible] Use the correct placeholder in requirement expression

lprv:
`u` is defined simply as "[an identifier](https://eel.is/c++draft/utility.arg.requirements#1.4)", but here it is clearly meant to denote a `T` object.


tkoeppe:
This looks good, thanks, but @jwakely to double-check.

PR #7519, opened 2024-12-25
[temp.deduct.call] Add missing words

lprv:


tkoeppe:
@jensmaurer I think this is right, but I'd appreciate a second opinion.

jensmaurer (approved):
This is ok.

PR #7520, opened 2024-12-26
[class.expl.init] Fix incorrect note

lprv:


jensmaurer (approved):
"initializing declaration" is defined in [dcl.init.general] p23. This looks like a correct change to me.

PR #7524, opened 2024-12-29
[expr.ass] Rename problematic stable label

AlisdairM:
As more folks from outside the technically focused WG21 committee pay attention to our standard, some language starts to gain attention as problematic.  This commit fixes one such case.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -674,7 +674,7 @@ type \tcode{T}, or
 is defined\iref{basic.def} or called\iref{expr.call}, or
 \item a class with a base class of type \tcode{T} is
 defined\iref{class.derived}, or
-\item an lvalue of type \tcode{T} is assigned to\iref{expr.ass}, or
+\item an lvalue of type \tcode{T} is assigned to\iref{expr.assign}, or
 \item the type \tcode{T} is the subject of an
 \keyword{alignof} expression\iref{expr.alignof}, or
 \item an \grammarterm{exception-declaration} has type \tcode{T}, reference to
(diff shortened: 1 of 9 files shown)


jensmaurer (approved):
Some members of CWG have expressed (unsolicited) support for this rename while CWG was discussing [stmt] and [dcl].

@tkoeppe , does that suffice for you, or do you want a more "official" statement from CWG?

shafik:
> Some members of CWG have expressed (unsolicited) support for this rename while CWG was discussing [stmt] and [dcl].
> 
> @tkoeppe , does that suffice for you, or do you want a more "official" statement from CWG?

Jens, I think I would like to have a wider discussion of the impact of this change to the wider C++ community.

This change has a wider blast radius considering how many Stackoverflow, blogs posts and other online community content we have out there that relies solely on the stable names to refer to standard. While folks more intimately connected to the WG21 process likely will catch on quickly enough. It is unlikely for most of the existing content out there to be updated quickly or ever at all.

As someone who came to WG21 slowly over time by participating in various C++ communities I see how opaque the WG21 process can be to those not involved. So this has the potential to make a lot of content a lot less useful for folks far away from the process.

Having spent also a lot of time in C community early on, I have seen how not having stable names has made a lot of old C content much less usable. So the impact is very real, they are stable names for a reason.

I don't think we should be willing to willy nilly go about updating stable names without some deeper thought.

Maybe we want to accept that cost but we should discuss that.

jensmaurer:
Ack.  Sending e-mail to EWG and CWG.

tkoeppe:
Thanks, applied (with updated commit messages).

PR #7525, opened 2024-12-30
[too many to list] Index macros directly from definitions

AlisdairM:
This commit uses the  macro to index macro definitions directly from the header synopses where they are defined, thus removing the need for a separate indexing macro that can get out of sync.

This commit does NOT address function-like macros, or macros that use placemarker syntax in their naming.

White this PR might have been more easily reviewed as one commit for each file, it seems better to apply the changes consistently. all at once, so that any future edits have a consistent pattern to follow.


jensmaurer:
> White this PR might have been more easily reviewed as one commit for each file, it seems better to apply the changes consistently. all at once, ...

That is a non-sequitur.  One commit for each file does not mean one pull request for each file, i.e. github provides for an environment where multiple commits can be naturally applied in an all-or-nothing fashion.

source/future.tex
```diff
@@ -221,34 +221,28 @@
 
 \pnum
 The header \libheaderref{stdalign.h} has the following macros:
-\indexlibraryglobal{__alignas_is_defined}%
 \begin{codeblock}
-#define @\xname{alignas_is_defined}@ 1
-#define @\xname{alignof_is_defined}@ 1
+#define @\libglobal{alignas_is_defined}@ 1
```
jensmaurer, line 225: This drop the double-underscore prefix  Did you actually run "diffpdf" on the resulting, rendered document?

source/diagnostics.tex
```diff
@@ -530,158 +530,83 @@
...
-#define EPROTOTYPE @\seebelow@                                           // freestanding
-#define ERANGE @\seebelow@                                               // freestanding
-#define EROFS @\seebelow@                                                // freestanding
-#define ESPIPE @\seebelow@                                               // freestanding
-#define ESRCH @\seebelow@                                                // freestanding
-#define ETIMEDOUT @\seebelow@                                            // freestanding
-#define ETXTBSY @\seebelow@                                              // freestanding
-#define EWOULDBLOCK @\seebelow@                                          // freestanding
-#define EXDEV @\seebelow@                                                // freestanding
+#define @\libglobal{errno}@ @\seebelow@
```
jensmaurer, line 534: I think I want a new LaTeX macro "\libmacro" that is currently defined as "\libglobal", but could be defined to something else.  For example, if we ever want to have a separate index of macro names or color macro definitions differently.

jensmaurer (changes requested):
I support the general approach, but please split at least the large changes into separate comm its (NOT separate pull requests), with a proper section label in each commit description.

source/future.tex
AlisdairM: Verified with diffpdf before committing the latest update.

source/diagnostics.tex
AlisdairM: Added both \libmacro and \libmacrox.  They work for both object-like and function-like macros, but I have not managed so solve the issue with macro names using placeholders.

With this change, we are close to being able to create an index of macros, that would also want the core language predefined macros to be included too.

AlisdairM:
> That is a non-sequitur. One commit for each file does not mean one pull request for each file, i.e. github provides for an environment where multiple commits can be naturally applied in an all-or-nothing fashion by grouping them into a single pull request.

I have broken up the PR into many more specific commits.

AlisdairM:
All the requested changes should now be applied.  Ready for re-review.

source/compatibility.tex
```diff
@@ -3734,10 +3734,9 @@
 Subclause \ref{csetjmp.syn} describes the changes.
 
 \rSec3[diff.offsetof]{Macro \tcode{offsetof(\placeholder{type}, \placeholder{member-designator})}}
-\indexlibraryglobal{offsetof}%
 
 \pnum
-The macro \tcode{offsetof}, defined in
+The macro \tcode{\libmacro{offsetof}}, defined in
```
jensmaurer, line 3739: \libmacro should make immediately surrounding \tcode superfluous, if defined appropriately.

Please clean up.

jensmaurer (commented):
Please typo-fix this commit description:

 [support] Better index the many macros in the [support] clauses

"Not that there are additional edits in other clauses to ensure that
macrods are consistently indexed using only through \libmacro to
support future evolution of that LaTeX macro."

Not -> Note

macrods -> macros

"using only through" feels non-English to me.

source/macros.tex
```diff
@@ -214,6 +214,11 @@
 \newcommand{\libmember}[2]{\indexlibrarymember{#1}{#2}#1}
 \newcommand{\libspec}[2]{\indexlibrarymemberx{#1}{#2}#1}
 
+% index for macros
+% use in-place defining the macro in header synopses
+% initially just a global entry in the library index
```
jensmaurer, line 219: If you want to write an explanatory sentence, please make that a real sentence (full stop at the end, capital letter at the start).  For example

% index for macros
% Use where the macro is defined in the header synposis.

"% initially just a global entry in the library index"

Can go, I think; we don't want to commit to anything happening in the future.

source/macros.tex
```diff
@@ -214,6 +214,11 @@
...
+\newcommand{\libmacro}[1]{\indexlibraryglobal{#1}#1}
```
jensmaurer, line 220: So, please put this commit plus the definition of \libmacrox (which I can't find right now) into a separate starter commit with label [macros].  Then, we mostly have nice follow-up commits for individual header files, and those commits can carry meaningful labels.

Oh, and \libmacrox is a bit of a misnomer, because "...x" for the indexing macros usually means you can select the stuff that goes into the index separately from the stuff that is shown where the macro appears in LaTeX.  But your \libmacrox is not like that; it prepends double underscores (I guess?).  So, please use a different suffix. Maybe \libmacrous ("us" for underscore) or whatever.

There is also the option to define a local LaTeX macro just for the `<version>` #defines , which also includes the "cpp_lib" prefix and saves redundancy there. But maybe that's a bit too much for this round of edits.

jensmaurer (changes requested):

source/compatibility.tex
AlisdairM: Done.

source/macros.tex
AlisdairM: Done

source/macros.tex
AlisdairM: Done

AlisdairM:
Again, latest push should have addressed all the open change requests.

source/support.tex
```diff
@@ -6019,8 +5932,7 @@
 the definition in~\ref{expr.call} applies.
 \item
 The restrictions that C places on the second parameter to the
-\indexlibraryglobal{va_start}%
-\tcode{va_start} macro in header \libheader{stdarg.h}
+\tcode{\libmacro{va_start}} macro in header \libheader{stdarg.h}
```
jensmaurer, line 5935: Drop surrounding \tcode, because it's redundant with \libmacro

And please do grep for this situation and fix all offenders; it happens often.

jensmaurer (changes requested):

jensmaurer:
I've talked to @tkoeppe, and we agree that we'd prefer to see one commit per header.  This means we can start the title of the respective commit with "[cerrno.syn]" or similar, which gives a nice scope to each commit.

I understand that sometimes, macros are referenced from various places (e.g. va_start), and you want to deal with all macros from a given header uniformly.  I'd suggest you factor the "obvious" candidates such as [cerrno.syn] (that have no references to those macros elsewhere) as a first iteration step, and we can have a deeper look at the rest later.

AlisdairM:
Could you confirm whether I hit all uses of a macro in multiple headers in a single commit, when resolving a header synopsis, or revisit with a separate commit or set of commits?  You mention `<cerrno>` as having no references, but it is actually the most referenced single header due to cross-references to `errno`.  Of course, there is every chance that I am adding too many `errno` cross-references too.

Could I land a smaller patch that adds the library macros, and just a few headers to get started?  I expect to revisit this PR on Sunday otherwise.

jensmaurer:
We'd hope there would be some headers with macros that are not mentioned elsewhere in the standard.  If `<cerrno>` is not such a header, fine, let's leave for later.

So, the ask is to start with those headers whose macros are not mentioned elsewhere, and have commits with label [cblah.syn] and changes for that header only, as a starter.  We can then have a look at the rest; maybe it's best to summarize those under [support] or so.

AlisdairM:
I have reapplied the changes one header at a time, with the exception of the Annex D updates that I applied as one commit at the end; I have not applied any changes to \[support\] yet as that has many headers to update; I have removed all additional indexing of macros from headers that are not the immediate subject of a commit to the same .tex file.

I have also rebased onto `main` so this PR should apply cleanly.

jensmaurer (changes requested):
Ok, this looks like a very nice first step to me.

Could you please update the commit titles from "Better index the macros in this header" to "Improve indexing for macros" ?  This betters the English, I think. :-)

Since this is a no-op as far as the actual text contents of the standard is concerned, I'd apply this soon afterwards.  @tkoeppe , FYI.

AlisdairM:
Commit messages should be updated now.

jensmaurer (approved):

jensmaurer:
@tkoeppe , this should be ready now.

tkoeppe:
Thanks, this is a nice simplification!

source/macros.tex
```diff
@@ -214,6 +214,11 @@
 \newcommand{\libmember}[2]{\indexlibrarymember{#1}{#2}#1}
 \newcommand{\libspec}[2]{\indexlibrarymemberx{#1}{#2}#1}
 
+% index for macros
+% Use in-place defining the macro in header synopses.
```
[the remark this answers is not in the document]
AlisdairM: Updated to provide clear comment for both new macros.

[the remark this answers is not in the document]
tkoeppe: Thanks!

PR #7529, opened 2025-01-01
[expected.bad.void] Editorial fix - syntax error in `bad_expected_access<void>`

Bronek:
This fixes obviously invalid syntax in the definition of `bad_expected_access<void>` class specialization
```c++
bad_expected_access& operator=(const bad_expected_access& noexcept);
```


jwakely (approved):
This was introduced by 8c997445c176c81a334e77f9344b91abc72b2772 in #6915

jensmaurer (approved):

PR #7537, opened 2025-01-03
escape semantically meaningful braces in [exec.getcomplsigs]

ericniebler:
[P2300R10 [exec.getcomplsigs]](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p2300r10.html#spec-execution.getcomplsigs) para 1 contains the following in code font:

```
transform_sender(decltype(get-domain-late(sndr, env)){}, sndr, env)
```

In the current working draft, this is rendered as:

```
transform_sender(decltype(get-domain-late(sndr, env)), sndr, env)
```

Note the missing braces. This is an editing mistake where the braces are not escaped for LaTeX, and hence are ignored.

### DIFF
--- a/source/exec.tex
+++ b/source/exec.tex
@@ -2373,7 +2373,7 @@ such that \tcode{decltype((sndr))} is \tcode{Sndr}, and
 let \tcode{env} be an expression
 such that \tcode{decltype((env))} is \tcode{Env}.
 Let \tcode{new_sndr} be the expression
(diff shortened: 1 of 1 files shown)


jensmaurer:
@ericniebler 

Now that we have the braces in the output, the resulting line gets too long. Because this is all code, LaTeX has troubles finding good linebreaking opportunities.  (Actually, the next line has a problem, because LaTeX doesn't hyphenate and already-hyphenated word.)

I've fixed this here: #7541

PR #7550, opened 2025-01-09
[expr.ref] Remove redundant bullet

lprv:
`E2` is grammatically an *id-expression*, which can't denote a type.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -3774,9 +3774,6 @@ is ignored\iref{expr.prim.paren}.
 \end{note}
 \end{itemize}
 
-\item If \tcode{E2} is a nested type, the expression \tcode{E1.E2} is
-ill-formed.
-
 \item If \tcode{E2} is a member enumerator and the type of \tcode{E2}
 is \tcode{T}, the expression \tcode{E1.E2} is a prvalue of type \tcode{T}
 whose value is the value of the enumerator.


jensmaurer:
A grammatical "id-expression" might not be an expression, because it has not been "suitably declared".  Which is fine, because E2 does not need to be (and sometimes isn't) a valid expression in its own right, for example when E2 names a pseudo-destructor or an overload set.

jensmaurer:
(The "suitably declared" rule was a digression, I think. Sorry.)

An _id-expression_ grammatically appears as a part of a _declarator-id_. There, it is not an expression in the sense of [expr].  However, the rules in [expr.prim.id.general] do not differentiate between _id-expression_ used in a _declarator-id_ and those used in expression contexts (such as stand-alone or as part of a class member access). Maybe they should.

Also note that overload resolution is not directly applied to the _id-expression_ appearing in a class member access, but to the entire class member access expression (for example, E1 defines the implicit object argument).

Given this mess, I'm not agreeing to editorially remove normative words for class member access.

My preferred path forward is to introduce separate grammar non-terminals for the thing that appears in a _declarator-id_, for the thing that is E2 in a class member access, and for the thing that is one option of a primary expression.  That would properly scope some of the rules we have, at which point we might be able to remove the "nested type" bullet you seek to strike.

PR #7553, opened 2025-01-10
[locale.time.put.members] Remove outdated footnote

cpplearner:
C99 defines two modifiers for `strftime`, namely `'E'` and `'O'`.


jwakely:
This is handled in my paper to rebase on C23.

jensmaurer:
Since this the footnote is already wrong for C99, I think we can strike this editorially prior to your C23 paper landing. What do you think, @jwakely ?

jwakely (approved):

jensmaurer (approved):

PR #7558, opened 2025-01-11
[temp.mem.enum] Fix consistency of example with [temp.inst] (editorial)

thebrandre:
The example is inconsistent with [[temp.inst] p3](https://wg21.link/temp.spec#temp.inst-3), which states the following: 

> The implicit instantiation of a class template specialization causes
> - the implicit instantiation of the declarations, but not of the definitions, of [...] scoped member enumerations [...]; and
> - the implicit instantiation of the definitions of [...] unscoped member enumerations [...]

Hence, the example can be fixed by either (1) omitting the instantiation or (2) by replacing the unscoped with a scoped enumeration. Since the implicit instantiation does not contribute to the point of [[temp.mem.enum]](https://wg21.link/temp.mem.enum), the inconsistency can be solved most easily with alternative (1).

The example was taken exactly as is from [DR1206](http://open-std.org/JTC1/SC22/WG21/docs/cwg_defects.html#1206), where it already seems to be an accidental copy-paste mishap because it is not important for the statement it is trying to make.
If one is willing to assume that, it is just an editorial issue.

Note: the example as is does not compile with any of the major compilers.
See here on Compiler Explorer: https://godbolt.org/z/xjM15MjP8

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -2659,7 +2659,6 @@ template definition.
 template<class T> struct A {
   enum E : T;
 };
-A<int> a;
 template<class T> enum A<T>::E : T { e1, e2 };
 A<int>::E e = A<int>::e1;
 \end{codeblock}


jensmaurer (approved):
Looks good.

jensmaurer:
@thebrandre, if you wish for your real name to appear in the commit history, you need to configure it in github.

thebrandre:
@jensmaurer Thanks for pointing that out! I just updated my GitHub profile.

thebrandre:
@lprv To my understanding, [CWG Issue 1485](https://cplusplus.github.io/CWG/issues/1485.html) addresses an issue unrelated to templates:

```cpp
struct S {
    // opaque-enum-declaration in class scope
    enum E : int;
};

// enum-specifier in namespace scope
enum S::E : int { X, Y, Z };

int x = X;           // (1) correct if enumerator declared in namespace scope?
int y = S::Y;        // (2) correct if enumerator declared in the scope of the enum-head?
```

This might have lead to the change from C++11 (N3242) [dcl.enum] p11

> Each enum-name and each unscoped enumerator is declared in the scope that immediately contains the enum-specifier.

to [C++23 (N4950) [dcl.enum] p12](https://eel.is/c++draft/enum#dcl.enum-12):

> The name of each unscoped enumerator is also bound in the scope that immediately contains the enum-specifier.

Interestingly, this changed only recently from C++20 to C++23 considering it was brought up on 2012-03-26.

Hence, in the code example above, both (1) and (2) should be correct as of C++23 with (2) having much broader compiler support regardless on the choice of `-std=c++11` or `-std=c++23`. In fact, all users, as long as they don't use Clang only, currently need to use (2), as you can see [Compiler Explorer](https://godbolt.org/z/orGPf1MGT).

**My proposed change should therefore not be affected.**

To me, it rather seems that way that [CWG Issue 1485](https://cplusplus.github.io/CWG/issues/1485.html) might have even pointed out the issue that would be fixed by this PR.

<hr/>

What astonishes me is the note on [CWG Issue 1485](https://cplusplus.github.io/CWG/issues/1485.html):

> Notes from the October, 2012 meeting:
> CWG agreed that an unscoped opaque enumeration in class scope should be forbidden.

This seems like a harsh decision, which, as far as I can tell, does not seem to be reflected by the current standard documents.
Additionally, it has far more potential to break existing code.

PR #7559, opened 2025-01-12
move para about write-env into write-env's itemdescr

ericniebler:
a paragraph about _`write-env`_ from P3396R1 got accidentally added outside of _`write-env`_'s `itemdescr`. move it inside.

### DIFF
--- a/source/exec.tex
+++ b/source/exec.tex
@@ -1901,9 +1901,9 @@ returns an object \tcode{e} such that
 \tcode{decltype(e)} models \exposconcept{queryable} and
 \item
 given a query object \tcode{q},
-the expression \tcode{e.query(q)} is expression-equivalent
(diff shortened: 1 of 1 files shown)


jensmaurer (approved):

jensmaurer:
The change seems fine, but it produces an overfull \hbox due to added indentation:

2025-01-12T04:18:15.0341238Z ##[error]Overfull \hbox (7.76248pt too wide) in paragraph at lines 1903--1908
2025-01-12T04:18:16.7916091Z ##[error]Process completed with exit code 1.

@ericniebler , please fix.

ericniebler:
i tried to reproduce the problem locally and i cannot. "it works on my machine." then i tried splatting a bunch of hyphenation markup in some of the longer words. that didn't work. so i'm at a loss.

jensmaurer:
Here's the overfull hbox:

Untitled

I'll fix it.

PR #7569, opened 2025-01-17
[type.info] Remove comments explaining deleted members

jwakely:


tkoeppe:
Who piped the Google style guide into the standard...

PR #7587, opened 2025-01-25
Rename non-type template parameter/argument to "constant"

cor3ntin:
As previously discussed by CWG.
The aim is to editorially adopt some of the wording changes made in P2841R5 to ease its review in core.

Note that not all-instance of non-type have been
mechanically replaced as [dcl] and [diff] use
the term to refer to anything that is not a type
in the context of lookup

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -296,10 +296,10 @@ Since template
 and template
 \grammarterm{template-argument}{s}
 are treated as types for descriptive purposes, the terms
-\term{non-type parameter}
+\term{constant parameter}
 and
-\term{non-type argument}
-are used to refer to non-type, non-template parameters and arguments.
+\term{constant argument}
+are used to refer to constant, non-template parameters and arguments.
 \end{footnote}
 \grammarterm{parameter-declaration}.
 A \grammarterm{template-parameter} of the form
@@ -315,7 +315,7 @@ template<class T, T i> void f(T t) {
 }
 \end{codeblock}
 Here, the template \tcode{f} has a type template parameter
-called \tcode{T}, rather than an unnamed non-type
+called \tcode{T}, rather than an unnamed constant
 template parameter of class \tcode{T}.
 \end{example}
 The \grammarterm{parameter-declaration} of a \grammarterm{template-parameter}
@@ -389,7 +389,7 @@ template<C3<int>... T> struct s5;       // associates \tcode{(C3<T, int> \&\& ..
 \end{example}
 
 \pnum
-A non-type template parameter
+A constant template parameter
 shall have one of the following (possibly cv-qualified) types:
 \begin{itemize}
 \item a structural type (see below),
@@ -420,7 +420,7 @@ structural types or (possibly multidimensional) arrays thereof.
 
 \pnum
 An \grammarterm{id-expression} naming
-a non-type template parameter of class type \tcode{T}
+a constant template parameter of class type \tcode{T}
 denotes a static storage duration object of type \tcode{const T},
 known as a \defn{template parameter object},
 which is template-argument-equivalent\iref{temp.type} to
@@ -430,7 +430,7 @@ to the type of the template parameter\iref{temp.arg.nontype}.
 No two template parameter objects are template-argument-equivalent.
 \begin{note}
 If an \grammarterm{id-expression} names
-a non-reference non-type template parameter,
+a non-reference constant template parameter,
 then it is a prvalue if it has non-class type.
 Otherwise, if it is of class type \tcode{T},
 it is an lvalue and has type \tcode{const T}\iref{expr.prim.id.unqual}.
@@ -454,7 +454,7 @@ template<const X& x, int i, A a> void f() {
 
 \pnum
 \begin{note}
-A non-type template parameter
+A constant template parameter
 cannot be declared to have type \cv{} \keyword{void}.
 \begin{example}
 \begin{codeblock}
@@ -465,7 +465,7 @@ template<void* pv> class Y;     // OK
 \end{note}
 
 \pnum
-A non-type template parameter
+A constant template parameter
 \indextext{array!template parameter of type}%
 of type ``array of \tcode{T}'' or
 \indextext{function!template parameter of type}%
@@ -485,7 +485,7 @@ S<v> z;                         // OK due to both adjustment and conversion
 \end{example}
 
 \pnum
-A non-type template parameter declared with a type that
+A constant template parameter declared with a type that
 contains a placeholder type with a \grammarterm{type-constraint}
 introduces the immediately-declared constraint
 of the \grammarterm{type-constraint}
@@ -560,7 +560,7 @@ template<class... T, class U> void g() { }      // error
 \indextext{\idxcode{<}!template and}%
 \pnum
 When parsing a default template argument
-for a non-type template parameter,
+for a constant template parameter,
 the first non-nested \tcode{>} is taken as
 the end of the \grammarterm{template-parameter-list}
 rather than a greater-than operator.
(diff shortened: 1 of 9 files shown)


source/templates.tex
```diff
@@ -296,10 +296,10 @@
 and template
 \grammarterm{template-argument}{s}
 are treated as types for descriptive purposes, the terms
-\term{non-type parameter}
+\term{constant parameter}
```
jensmaurer, line 299: We seem to call these "constant template parameter" and "constant template argument" throughout, so we should introduce that very term.

jensmaurer (changes requested):

source/templates.tex
cor3ntin: Hum. that footnote is in a bit of a weird state... should we get rid of it now? (the paper does that otherwise)

cor3ntin: P2841 has the following

```
A template parameter has one of the following forms:
• A type template parameter is a template parameter introduced by a type-parameter.
• A constant template parameter is a template parameter introduced by a parameter-declaration.
• A type template template parameter is a template parameter introduced by a type-tt-parameter.
• A variable template template parameter is a template parameter introduced by a variable-tt-parameter.
• A concept template parameter is a template parameter introduced by a concept-tt-parameter.

Type template template parameters, variable template template parameters, and concept
template parameters are collectively referred to as template template parameters.
```

That surgery seems best left to the paper (as it modifies the grammar - by splitting type-parameter and type-tt-parameter.

We could just remove the footnote by 
> A constant template parameter is a template parameter introduced by a parameter-declaration

and add the rest of the paragraph later if you think it would produce a more consistent intermediate state.

source/templates.tex
```diff
@@ -296,10 +296,10 @@
 and template
 \grammarterm{template-argument}{s}
 are treated as types for descriptive purposes, the terms
-\term{non-type parameter}
+\term{constant template parameter}
 and
-\term{non-type argument}
-are used to refer to non-type, non-template parameters and arguments.
+\term{constant template argument}
+are used to refer to constant, non-template parameters and arguments.
```
jensmaurer, line 302: I think I want to keep "non-type" here.

source/templates.tex
jensmaurer: I would prefer to keep the changes here as non-inventive as possible, and leave the rest to the paper.

cor3ntin: Agreed

source/templates.tex
```diff
@@ -3453,18 +3453,18 @@
 \end{example}
 
 \pnum
-A non-type argument is non-specialized if it is the name of a non-type
-parameter.
-All other non-type arguments are specialized.
+A constant template argument is non-specialized if it is the name of a constant
+template parameter.
+All other constant arguments are specialized.
```
hubert-reinterpretcast, line 3458:
```suggestion
All other constant template arguments are specialized.
```

source/templates.tex
```diff
@@ -4598,7 +4598,7 @@
 \item \grammarterm{parameter-declaration} in a \grammarterm{lambda-declarator}
 or \grammarterm{requirement-parameter-list},
 unless that \grammarterm{parameter-declaration} appears in a default argument, or
-\item \grammarterm{parameter-declaration} of a (non-type) \grammarterm{template-parameter}.
+\item \grammarterm{parameter-declaration} of a (constant) \grammarterm{template-parameter}.
```
hubert-reinterpretcast, line 4601:
```suggestion
\item \grammarterm{parameter-declaration} of a \grammarterm{template-parameter} that declares a constant template parameter.
```

source/templates.tex
```diff
@@ -7568,7 +7568,7 @@
 \item
 the specified member is not a template where a template is required, or
 \item
-the specified member is not a non-type where a non-type is required.
+the specified member is not a constant where a constant is required.
```
hubert-reinterpretcast, line 7571: The sentence applies to the following (which EDG and MSVC seems to get wrong: https://godbolt.org/z/j1hxG4f8e):
```cpp
template <typename T>
auto *f() {
  static constexpr auto x = sizeof (T::M); // sizeof _unary-expression_
  return &x;
}
struct Q { using M = int; };
void g() {
  f<Q>();
}
```

```suggestion
the specified member is not a non-type, non-template where a non-type, non-template is required.
```

hubert-reinterpretcast (changes requested):

source/templates.tex
```diff
@@ -7590,7 +7590,7 @@
   // Deduction fails in each of these cases:
   f<A>(0);          // \tcode{A} does not contain a member \tcode{Y}
   f<B>(0);          // The \tcode{Y} member of \tcode{B} is not a type
-  g<C>(0);          // The \tcode{N} member of \tcode{C} is not a non-type
+  g<C>(0);          // The \tcode{N} member of \tcode{C} is not a constant
```
hubert-reinterpretcast, line 7593:
```suggestion
  g<C>(0);          // The \tcode{N} member of \tcode{C} is not a non-type, non-template name
```

source/templates.tex
```diff
@@ -8325,7 +8325,7 @@
 is compared with an actual type (call it
 \tcode{A}),
 and an attempt is made to find template argument values (a type for a type
-parameter, a value for a non-type parameter, or a template for a
+parameter, a value for a constant template parameter, or a template for a
 template parameter) that will make
```
hubert-reinterpretcast, line 8329: The inconsistent use of "template parameter" versus "parameter" following the change is very confusing.

```suggestion
template parameter, a value for a constant template parameter, or a template for a
template template parameter) that will make
```

source/templates.tex
cor3ntin: I considered that (and also to remove the parenthetical entirely). parameter-declaration implies a constant template parameter, I'm afraid your wording might imply the existence of some alternative

source/templates.tex
cor3ntin: Good catch, I completely missed that

source/templates.tex
```diff
@@ -296,9 +296,9 @@
 and template
 \grammarterm{template-argument}{s}
 are treated as types for descriptive purposes, the terms
-\term{non-type parameter}
+\term{constant template parameter}
```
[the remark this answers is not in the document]
cor3ntin: @hubert-reinterpretcast It gets struck by the paper (which updates the surrounded text), hopefully we can live with that imperfection for a few weeks.  I don't have any objection to do it now but we agreed to keep the changes minimal

source/templates.tex
hubert-reinterpretcast:
```suggestion
\item \grammarterm{parameter-declaration} of a \grammarterm{template-parameter} (which necessarily declares a constant template parameter).
```

source/templates.tex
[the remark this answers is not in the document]
hubert-reinterpretcast: > hopefully we can live with that imperfection for a few weeks

I would rather not risk needing the paper to keep things in shape.

cor3ntin:
@jensmaurer Can we make progress on this this week? I am afraid things are going to be rather messy otherwise. Thanks

source/templates.tex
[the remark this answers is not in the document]
jensmaurer: It is not clear that the concept template parameters paper makes it to the Hagenberg plenary, given that we also have contracts and reflection on the plate, which take priority.  Given Hubert's view, I think I want the bulleted list from the paper (which replaces the footnote) here, because we otherwise have no definition for the term "constant template parameter", it seems.

That's still editorial, I think; we don't change the meaning of the specification.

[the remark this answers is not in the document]
cor3ntin: The bulleted list refers to grammar production that don't exist in the current wording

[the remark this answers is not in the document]
cor3ntin: I'd rather go with Hubert suggestion and remove the first part of the sentence. The fact that we have a definition in a footnote was pre-existing. I'll do that

source/templates.tex
hubert-reinterpretcast: @cor3ntin, I think you applied the older suggestion? Jens expressed a preference for the newer one.

hubert-reinterpretcast (approved):
LGTM; thanks!

jensmaurer (approved):
Ok for me.

tkoeppe:
Right after #7485 if that's OK?

jensmaurer:
Yes, that needs to be squashed and the commit-message needs to get at least the label [std] or so.
After the other commit is fine, as long as it's happening soon.

PR #7590, opened 2025-01-26
[util.smartptr.enab], [util.smartptr.shared.const] Add `\exposid` to `weak_this`

lprv:


jensmaurer (approved):
Looks good; the commit description could use some love + care.
(We're also renaming the member, not just applying \exposid.)

PR #7591, opened 2025-01-26
[defns.argument.templ] Change 'template instantiation' to 'template'; mention braced-init-list

lprv:
Not all templates are instantiated, so limiting this definition to the process template instantiation seems wrong.

Adding *braced-init-list* updates the definition to match the changes made by [P2308](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2308r1.html).

### DIFF
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -122,12 +122,10 @@ comma-separated list bounded by the parentheses
 \defncontext{throw expression} operand of \keyword{throw}
 
 \indexdefn{argument}%
-\indexdefn{argument!template instantiation}%
+\indexdefn{argument!template}%
 \definition{argument}{defns.argument.templ}
-\defncontext{template instantiation}
-\grammarterm{constant-expression},
-\grammarterm{type-id}, or
-\grammarterm{id-expression} in the comma-separated
+\defncontext{template}
+\grammarterm{template-argument} in the comma-separated
 list bounded by the angle brackets
 
 \indexdefn{block (execution)}%


tkoeppe:
See also #7396.

tkoeppe:
@jensmaurer could you have a look? Seems reasonable to me.

source/intro.tex
```diff
@@ -118,12 +118,13 @@
...
+\indexdefn{argument!template}%
 \definition{argument}{defns.argument.templ}
-\defncontext{template instantiation}
+\defncontext{template}
 \grammarterm{constant-expression},
-\grammarterm{type-id}, or
-\grammarterm{id-expression} in the comma-separated
+\grammarterm{type-id},
+\grammarterm{id-expression}, or
+\grammarterm{braced-init-list} in the comma-separated
```
jensmaurer: This is incomplete; it could also be the name of a template (as argument for a template template parameter) and arbitrary expressions.

And I'm not seeing braced-init-list in [temp.names], where does this come in?

jensmaurer (changes requested):

source/intro.tex
lprv: Expressions are covered by *constant-expression*; *braced-init-list* was restored in [temp.names] by #8487. Template template arguments were covered by *id-expression* at the time this PR was opened, but that was changed to *template-name* in the definition of *template-argument* since then.

Rather than keeping the two definitions in sync, I updated this PR to refer to the *template-argument* production instead.

PR #7592, opened 2025-01-27
[class.prop] point to actual layout spec in [expr.rel]

Axel-Naumann:
The original ref does not define the layout (anymore?), thanks to Jens for providing me with the correct ref!


tkoeppe:
Where exactly is "standard layout" defined in expr.rel?

jensmaurer:
It's not.  This should not replace the cross-reference to class.mem, but change it to class.mem.general and expr.rel.  The latter describes that data members are laid out in source code order.

Axel-Naumann:
Ah thanks, I misunderstood Jens's suggestion - this makes perfect sense. Please let me know if the current version is acceptable.

PR #7598, opened 2025-01-31
[cmath.syn] Consolidate std namespaces

AlisdairM:
There is no ordering dependency between the two typedefs in namespace std, the macros that follow, and teh next opening of namespace std, so move the two typedefs to avoid repeatedly opening an closing the namespace.

Note that we could have done this without moving
the typedefs as macros are not bound by namespaces, but our convention very sensibly avoids confusing
readers by keeping macro definitions outside of
namespaces.


tkoeppe:
@AlisdairM Could you please rebase?

AlisdairM:
Rebased and force-pushed

PR #7602, opened 2025-02-02
Wrap keywords that were \terminal{} with \keyword{} instead

ilazaric:
My impression is that keywords should be `\keyword{<string>}` usually, no surrounding command, except in `one of`  grammar definitions, where they should be wrapped in a `\terminal{...}`

Didn't do this on a couple occurences in preprocessor.tex (`if` and `else` are in `\terminal{}` args):
```
\begin{bnf}\obeyspaces
\nontermdef{else-group}\br
    \terminal{\# else   } new-line \opt{group}
\end{bnf}
```

```
\begin{ncsimplebnf}\obeyspaces
\indextext{\idxcode{\#if}}%
\terminal{\# if     } constant-expression new-line \opt{group}\br
\indextext{\idxcode{\#elif}}%
\terminal{\# elif   } constant-expression new-line \opt{group}
\end{ncsimplebnf}
```

```
\begin{bnf}\obeyspaces
\nontermdef{if-group}\br
    \terminal{\# if     } constant-expression new-line \opt{group}\br
    \terminal{\# ifdef  } identifier new-line \opt{group}\br
    \terminal{\# ifndef } identifier new-line \opt{group}
\end{bnf}
```

I don't think these are quite the same meaning as keyword if

Also, there are many instances of keywords not being \keyword-ed in \tcode{} parts, this PR does not touch those

Additionally some minor cleanup
* expressions.tex: `\keyword{U}` -> `\tcode{U}`
* lex.tex: removed space between command and argument

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -2588,7 +2588,7 @@ named in a friend declaration\iref{class.friend}.
     identifier \opt{\terminal{...}}\br
     \terminal{\&} identifier \opt{\terminal{...}}\br
     \keyword{this}\br
-    \terminal{*} \terminal{this}
+    \terminal{*} \keyword{this}
 \end{bnf}
 
 \begin{bnf}
@@ -3449,8 +3449,8 @@ Postfix expressions group left-to-right.
     typename-specifier \terminal{(} \opt{expression-list} \terminal{)}\br
     simple-type-specifier braced-init-list\br
     typename-specifier braced-init-list\br
-    postfix-expression \terminal{.} \opt{\terminal{template}} id-expression\br
-    postfix-expression \terminal{->} \opt{\terminal{template}} id-expression\br
+    postfix-expression \terminal{.} \opt{\keyword{template}} id-expression\br
+    postfix-expression \terminal{->} \opt{\keyword{template}} id-expression\br
     postfix-expression \terminal{++}\br
     postfix-expression \terminal{--}\br
     \keyword{dynamic_cast} \terminal{<} type-id \terminal{>} \terminal{(} expression \terminal{)}\br
@@ -5153,7 +5153,7 @@ transfers control to its caller or resumer.
 
 \begin{bnf}
 \nontermdef{await-expression}\br
-    \terminal{co_await} cast-expression
+    \keyword{co_await} cast-expression
 \end{bnf}
 
 \pnum
@@ -7473,8 +7473,8 @@ that type and the other is a null pointer constant. The result is of type
 
 \begin{bnf}
   \nontermdef{yield-expression}\br
-  \terminal{co_yield} assignment-expression\br
-  \terminal{co_yield} braced-init-list
+  \keyword{co_yield} assignment-expression\br
+  \keyword{co_yield} braced-init-list
 \end{bnf}
 
 \pnum
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -2079,8 +2079,8 @@ its interpretation depends on the context in which it appears.
 \indextext{literal!boolean}%
 \begin{bnf}
 \nontermdef{boolean-literal}\br
-    \terminal{false}\br
-    \terminal{true}
+    \keyword{false}\br
+    \keyword{true}
 \end{bnf}
 
 \pnum
@@ -2093,7 +2093,7 @@ Such literals have type \tcode{bool}.
 \indextext{literal!pointer}%
 \begin{bnf}
 \nontermdef{pointer-literal}\br
-    \terminal{nullptr}
+    \keyword{nullptr}
 \end{bnf}
 
 \pnum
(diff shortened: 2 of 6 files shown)


source/lex.tex
```diff
@@ -1679,7 +1679,7 @@
 \topline
 \lhdr{\grammarterm{floating-point-suffix}} & \rhdr{type} \\ \capsep
 none & \keyword{double} \\
-\tcode{f} or \tcode{F} & \keyword {float} \\
+\tcode{f} or \tcode{F} & \keyword{float} \\
```
jensmaurer, line 1682: Please move this to whitespace change a separate commit.

source/expressions.tex
```diff
@@ -322,7 +322,7 @@
 If a program invokes
 a defaulted copy/move constructor or copy/move assignment operator
 for a union of type \tcode{U} with a glvalue argument
-that does not denote an object of type \cv{}~\keyword{U} within its lifetime,
+that does not denote an object of type \cv{}~\tcode{U} within its lifetime,
```
jensmaurer, line 325: Please move this to a separate commit; it doesn't fit the theme "add \keyword". (It's actually "remove incorrect \keyword".)

source/declarations.tex
```diff
@@ -6815,16 +6815,16 @@
...
-\bnfindent\bnfindent \terminal{if (!\exposid{initial-await-resume-called})}\br
-\bnfindent\bnfindent\bnfindent \terminal{throw} \terminal{;}\br
+\bnfindent \terminal{\}} \keyword{catch} \terminal{( ... ) \{}\br
+\bnfindent\bnfindent \keyword{if} \terminal{(!\exposid{initial-await-resume-called})}\br
+\bnfindent\bnfindent\bnfindent \keyword{throw} \terminal{;}\br
 \bnfindent\bnfindent \terminal{\exposid{promise}.unhandled_exception()} \terminal{;}\br
 \bnfindent \terminal{\}}\br
 \exposid{final-suspend} \terminal{:}\br
-\bnfindent \terminal{co_await} \terminal{\exposid{promise}.final_suspend()} \terminal{;}\br
+\bnfindent \keyword{co_await} \terminal{\exposid{promise}.final_suspend()} \terminal{;}\br
```
jensmaurer, line 6827: This whole section is not a grammar snippet, but equivalent code.  It should be in a "codeblock" environment, not in a BNF environment.  Make that a separate commit if you want to keep working on that, please.

jensmaurer (changes requested):

source/lex.tex
ilazaric: Reverted

source/expressions.tex
ilazaric: Reverted, will push a different PR for this

source/declarations.tex
ilazaric: Reverted, need to understand this more first  
Thanks for the feedback, apologies for the time it took me to integrate it

ilazaric:
Last commit fixes spacing in overloading.tex, bnf for operator
I originally didn't realize the spacing was important
Rendered and it looks good:
image

jensmaurer (approved):
Looks good.

jensmaurer:
Please rebase.

tkoeppe:
@ilazaric: Could you please rebase?

ilazaric:
Apologies
I've rebased and resolved conflicts
I regenerated std.pdf and spot checked the affected locations, looks good to me, example:
image

jensmaurer (approved):

jensmaurer:
Looks good.

tkoeppe:
Could you tell me the correctly phrased commit message? It's OK to keep things in one commit, but we should state the affected subclauses. So something like "[foo.bar, xy.zw] Use "\keyword" instead of "\terminal"". (Don't worry about changing the commit, I can do that at application time. I would just appreciate you looking up the references.)

ilazaric:
[expr.prim.lambda.capture], [expr.post.general]/1, [expr.await]/1, [expr.yield], [lex.bool], [lex.nullptr], [over.call.object]/2, [over.oper.general]/1, [cpp.include]/7, [stmt.return.coroutine], [stmt.return.coroutine]/2, [temp.res.general]/2

Note [stmt.return.coroutine] is changed in 0th and 2nd paragraph
Just FYI, if I didn't write a /number, it is because the change occurs before the first paragraph
I was considering writing /0 , but looking at git log this doesn't seem to be common practice

Alternate format:
[expr.prim.lambda.capture, expr.post.general/1, expr.await/1, expr.yield, lex.bool, lex.nullptr, over.call.object/2, over.oper.general/1, cpp.include/7, stmt.return.coroutine, stmt.return.coroutine/2, temp.res.general/2]

Alternate without paragraph numbers:
[expr.prim.lambda.capture, expr.post.general, expr.await, expr.yield, lex.bool, lex.nullptr, over.call.object, over.oper.general, cpp.include, stmt.return.coroutine, temp.res.general]

PR #7606, opened 2025-02-02
[editorial] ranges "which" hunt

ericniebler:
replaces incorrect uses of "which" with "that" in the specification of `enable_view`. it should be possible to drop a phrase introduced with "which" without changing the meaning of the sentence. if the phrase is essential to the point, "that" is more correct.

(this is a pet peeve of mine.)


jwakely (approved):

jensmaurer (approved):
Could you please fix the commit description to use the format described here:

https://github.com/cplusplus/draft/wiki/Commit-message-format#editorial-commits

ericniebler:
This is now done

jensmaurer (approved):

PR #7611, opened 2025-02-03
[futures.state] Add \tcode for true

Andreas-Krug:


source/threads.tex
```diff
@@ -10696,7 +10696,7 @@
 the return object or provider gives up its reference to its shared state; and
 
 \item these actions will not block for the shared state to become ready, except that it
-may block if all of the following are true: the shared state was created by a call to
+may block if all of the following are \tcode{true}: the shared state was created by a call to
```
jensmaurer, line 10699: That's not correct; we're not talking about the result of a C++ expression of type `bool`, but about the truth of English statements.

jensmaurer (changes requested):
This is a bad change, in my view.

PR #7613, opened 2025-02-04
[memory.syn] Fix indentation

Andreas-Krug:


jwakely (approved):
It looks like they weren't reindented when `constexpr` was added to them.

jensmaurer (approved):

PR #7615, opened 2025-02-04
Index `<cstarg>` macros

AlisdairM:
Following Jens's suggestion, adding the improved indexing of the macros more widely used in the standard as separate PRs.


jensmaurer (changes requested):
Looks mostly good.

The commit descriptions need work:

The first commit should use the usual phrasing "Improve indexing for macros"
The second commit uses backticks, the third one uses usual single quotes.  Please consistencify. (no that's not a word)

AlisdairM:
Cleaned up the final commit message, but the first commit is intentional as it performs an indexing cleanup on the `va_list` typedef as well as the macros, treating the header as a whole, so I did not want to call it out as just a macro indexing PR.

jensmaurer (approved):

PR #7622, opened 2025-02-07
[fs.op.funcs] Remove empty parens when referring to functions by name

jwakely:
As per the Specification Style Guidelines.

https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#describing-function-calls

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -17515,7 +17515,7 @@ bool filesystem::create_directories(const path& p, error_code& ec);
 \begin{itemdescr}
 \pnum
 \effects
-Calls \tcode{create_directory()} for each element of \tcode{p}
+Calls \tcode{create_directory} for each element of \tcode{p}
   that does not exist.
 
 \pnum
@@ -17608,7 +17608,7 @@ void filesystem::create_directory_symlink(const path& to, const path& new_symlin
 \begin{itemdescr}
 \pnum
 \effects
-Establishes the postcondition, as if by POSIX \tcode{symlink()}.
+Establishes the postcondition, as if by POSIX \tcode{symlink}.
 
 \pnum
 \ensures
@@ -17623,7 +17623,7 @@ As specified in~\ref{fs.err.report}.
 \begin{note}
 Some operating systems require symlink creation to
 identify that the link is to a directory.
-Thus, \tcode{create_symlink()} (instead of \tcode{create_directory_symlink()})
+Thus, \tcode{create_symlink} (instead of \tcode{create_directory_symlink})
 cannot be used reliably to create directory symlinks.
 \end{note}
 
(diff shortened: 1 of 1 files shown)


jensmaurer (commented):
The specification style guidelines expressly say that "call X" is fine, where X has parentheses (= is an expression).  It's just "calling the function Y" needs to have parentheses removed.

Maybe we want to fine-tune the rules a little, but the changes proposed here don't seem to be "obviously in line" to me.

jwakely:
> The specification style guidelines expressly say that "call X" is fine, where X has parentheses (= is an expression).

Yes but it has to be a valid expression, so "calls `create_directory()`" is wrong, because it needs arguments.

> It's just "calling the function Y" needs to have parentheses removed.

I don't see a distinction between "calls f()" and "calling the function f()" in the case where f() is not a valid expression.

tkoeppe:
So, we accept `calls f(1, 2, 3)` (instead of "invokes `f` with arguments `1`, `2`, `3`"), but if we don't give arguments, then it must be "calls `f`", not "calls `f()`", right? We can make that clearer in the wiki, but that seems like a good rule.

jwakely:
Yes, that's what I thought the rule was. I see that the wiki is actually ambiguous and can be interpreted differently, but that different interpretation seems useless to me. A rule that allows "call `sin()`" but not "calling the function `sin()`" seems like a completely pointless rule. You cannot "call `sin()`" because that's not the name of a function, and calling one of the overloads of `sin` would requires an argument to make it a valid expression.

jwakely:
If we do want to allow "call `sin()`" then IMHO we might as well strike the rule entirely, and just allow empty parens on any function name! (This would match POSIX, which intentionally puts empty parens when referring to any function, but I guess that can be argued to be OK in ISO C89 where `fopen()` declares a function without saying what its parameters are).

tkoeppe:
Let's fix the wiki instead!

jensmaurer:
"call `sin()`" should not be valid. If the wiki needs fixing, so be it.

tkoeppe:
Actually, is the wiki really ambiguous? It says (emphasis mine),

"""
When we say "Calls blah", *blah should be an expression that results in a function call*. Eg, "calls operator new(size)" or "calls `current_exception() and stores the returned value".
"""

and since `sin()` is not an expression that results in a function call, it already doesn't seem to be allowed? We can certainly clarify the wording a bit, but the intention seems clear already.

jwakely:
Yeah I checked it again yesterday and cross-referenced it with #2150, and the wiki seems pretty clear. And all the changes in this PR are cases where `foo()` is not a valid expression.

tkoeppe:
OK, great, thanks!

PR #7636, opened 2025-02-13
[basic.types.general] Change ordering to "non-variant non-static"

hubert-reinterpretcast:
The definition of literal type is the only place where "non-static non-variant data member" is used as opposed to "non-variant non-static data member".

Change to use the canonical ordering.


tkoeppe:
Thanks! Let's apply it after the motions, just so it doesn't get overwritten by accident.

jensmaurer (approved):

PR #7639, opened 2025-02-13
[range.concat.iterator] Improve awkward wording

lprv:


tkoeppe:
@jwakely Could you have a look? I appreciate that the current wording is awkward, but I want to double-check that we can talk about types of a pack as an expansion of a pack of types.

An alternative might be to introduce a placeholder _ITS_ and define:

```
ITS := std::declval<const iterator_t<maybe-const<Const, Views>>&>()
```

PR #7642, opened 2025-02-14
Follow-up to P2546, "Debugging Support"

hubert-reinterpretcast:
Follow-up to https://github.com/cplusplus/draft/pull/6681 (P2546R5) from thread beginning with edit/2023/11/0870.

- The notes in the paper make statements of fact about behaviour that, normatively, is implementation-defined. These should instead be statements of intent.
- The description of `breakpoint` assumes that a debugger is present (and should instead cover both cases).
- The odd restriction on how the debugger resumes execution of the program needs elaboration.
- The description of the implementation strategy of `is_debugger_present` for POSIX makes assumptions that are more appropriate for the LSB.
- The terminology used in the LSB is "tracing process", not "tracer parent process".
- The note in `is_debugger_present` implies the wrong trade-off when it is unknown whether a debugger is present. Clarify to prefer returning `false`.

### DIFF
--- a/source/diagnostics.tex
+++ b/source/diagnostics.tex
@@ -2497,9 +2497,9 @@ It is intended that, when tracing the execution of a program with a debugger, an
 implementation returns \tcode{true} with the use of an immediate query, as
 needed, to determine if the program is being traced by a debugger. On Windows or
 equivalent systems, this can be achieved by calling the
-\tcode{::IsDebuggerPresent()} Win32 function. On POSIX, this can be achieved by
-checking for a tracer parent process, with best effort determination that such a
-tracer parent process is a debugger.
+\tcode{::IsDebuggerPresent()} Win32 function. For systems compatible with
+ISO/IEC 23360:2021, this can be achieved by checking for a tracing process, with
+best effort determination that such a tracing process is a debugger.
 \end{note}
 
 \end{itemdescr}
--- a/source/diagnostics.tex
+++ b/source/diagnostics.tex
@@ -2493,13 +2493,13 @@ This function has no preconditions.
 \impldef{default semantics of \tcode{is_debugger_present}}.
 
 \begin{note}
-It is intended that, when tracing the execution of a program with a debugger, an
-implementation returns \tcode{true} with the use of an immediate query, as
-needed, to determine if the program is being traced by a debugger. On Windows or
+It is intended that, using an immediate (uncached) query to determine if the
+program is being traced by a debugger, an implementation returns \tcode{true}
+only when tracing the execution of the program with a debugger. On Windows or
 equivalent systems, this can be achieved by calling the
 \tcode{::IsDebuggerPresent()} Win32 function. For systems compatible with
 ISO/IEC 23360:2021, this can be achieved by checking for a tracing process, with
-best effort determination that such a tracing process is a debugger.
+a best-effort determination that such a tracing process is a debugger.
 \end{note}
 
 \end{itemdescr}
--- a/source/diagnostics.tex
+++ b/source/diagnostics.tex
@@ -2493,12 +2493,13 @@ This function has no preconditions.
 \impldef{default semantics of \tcode{is_debugger_present}}.
 
 \begin{note}
-When tracing the execution of a program with a debugger, an implementation
-returns \tcode{true}. An implementation performs an immediate query, as needed,
-to determine if the program is traced by a debugger. On Windows or equivalent
-systems, this can be achieved by calling the \tcode{::IsDebuggerPresent()} Win32
-function. On POSIX, this can be achieved by checking for a tracer parent process,
-with best effort determination that such a tracer parent process is a debugger.
+It is intended that, when tracing the execution of a program with a debugger, an
+implementation returns \tcode{true} with the use of an immediate query, as
+needed, to determine if the program is being traced by a debugger. On Windows or
+equivalent systems, this can be achieved by calling the
+\tcode{::IsDebuggerPresent()} Win32 function. On POSIX, this can be achieved by
+checking for a tracer parent process, with best effort determination that such a
+tracer parent process is a debugger.
 \end{note}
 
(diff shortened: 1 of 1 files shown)


hubert-reinterpretcast:
@jwakely @grafikrobot, this is the follow-up from the November 2023 discussion.

Dani-Hub:
@hubert-reinterpretcast What is the meaning of "abend" in `the program can abend`? Should that be "terminate" instead? (I think we have banned the word "abort" unless we specifically refer to `std::abort`)

W-E-Brown:
"Abend" is a shortened form of "abnormal end."  The term is of fairly long standing; I recall it from at least the 1960's when it was used by and in connection with IBM 360 systems software, and it may date from even earlier.  When a program "abends," it terminates, often (typically?) for unforeseen reasons.

jwakely:
I agree it's an established term, but if it's not well known then it doesn't help in a note that's intended to add clarification. Otherwise this looks like a good improvement.

Dani-Hub:
> "Abend" is a shortened form of "abnormal end." The term is of fairly long standing; I recall it from at least the 1960's when it was used by and in connection with IBM 360 systems software, and it may date from even earlier. When a program "abends," it terminates, often (typically?) for unforeseen reasons.

Thanks for the explanation. Could we possibly replace "abends" by "ends abnormally" instead?

source/diagnostics.tex
```diff
@@ -2447,10 +2447,12 @@
...
-When invoked, the execution of the program temporarily halts and execution is
-handed to the debugger until such a time as: The program is terminated by the
-debugger, or the debugger resumes execution of the program as if the function
-was not invoked.
+It is intended that, when invoked with a debugger present, the execution of the
+program temporarily halts and execution is handed to the debugger until the
+program is either terminated by the debugger or the debugger resumes execution
+of the program as if the function was not invoked. In particular, there is no
+intent for a call to this function to accomodate resumption of the program in a
+different manner. If there is no debugger present, the program can abend.
```
hubert-reinterpretcast, line 2455: @Dani-Hub, how about:
```suggestion
different manner. If there is no debugger present, execution of the program can
end abnormally.
```
?

Dani-Hub: Looks like a great improvement to may - Thanks Hubert!

grafikrobot (approved):
Thank you @hubert-reinterpretcast for the improved wording.

Dani-Hub (approved):

source/diagnostics.tex
```diff
@@ -2491,12 +2494,13 @@
...
-to determine if the program is traced by a debugger. On Windows or equivalent
-systems, this can be achieved by calling the \tcode{::IsDebuggerPresent()} Win32
-function. On POSIX, this can be achieved by checking for a tracer parent process,
-with best effort determination that such a tracer parent process is a debugger.
+It is intended that, using an immediate (uncached) query to determine if the
+program is being traced by a debugger, an implementation returns \tcode{true}
+only when tracing the execution of the program with a debugger. On Windows or
+equivalent systems, this can be achieved by calling the
+\tcode{::IsDebuggerPresent()} Win32 function. For systems compatible with
+ISO/IEC 23360:2021, this can be achieved by checking for a tracing process, with
```
jensmaurer, line 2502: This replaces "POSIX" with "Linux Standard Base".  Is there a particular reason for that?  Does POSIX not cover the needed functionality?

jensmaurer (commented):
Did I mention that I was opposed to adding those functions to start with? A note is obviously not the right tool to specify them, but we lack the normative framework to put the specification in a "recommended practice".  (end of rant)

@jwakely should see this change.

source/diagnostics.tex
jwakely: Although `ptrace` is present in most UNIX-like systems, it isn't part of POSIX. The LSB is a much narrower scope than POSIX, but it _is_ a standard, unlike families like "BSD" or "SysV". So if we want a standard that includes `ptrace`, LSB makes sense.

However, the wording never mentions `prrace` it just talks about "tracing" and "a tracer parent process" which is still meaningful even in e.g. HP-UX which replaced `ptrace` with `ttrace`, and in Solaris through `ptrace` but also via procfs. So maybe it's OK.

jwakely (commented):
I don't love the change to refer to the LSB. What does it imply for e.g. macOS, which is not Windows and isn't compatible with the LSB either?

Not that "On POSIX" was really great wording either.

source/diagnostics.tex
hubert-reinterpretcast: > However, the wording never mentions `prrace` it just talks about "tracing" and "a tracer parent process" which is still meaningful even in e.g. HP-UX which replaced `ptrace` with `ttrace`, and in Solaris through `ptrace` but also via procfs. So maybe it's OK.

Using "POSIX" in conjunction with the handwaving implies that the handwaving is necessarily meaningful for all possible implementations of POSIX. That it is probably meaningful for extant implementations of POSIX does not make me feel any better about using "POSIX" without qualification.

hubert-reinterpretcast:
> I don't love the change to refer to the LSB. What does it imply for e.g. macOS, which is not Windows and isn't compatible with the LSB either?
> 
> Not that "On POSIX" was really great wording either.

I would like to err on the side of "saying nothing wrong". Which is to say that, saying nothing for macOS (or any other non-Windows, non-LSB platform) leaves people the freedom to see whether the LSB implementation hint can be adapted for macOS, etc. without implying that the hint is applicable in cases where it isn't.

jwakely:
Yeah that makes sense

hubert-reinterpretcast:
@jensmaurer @jwakely, are we good with the current proposed change then?

jwakely (approved):

tkoeppe:
Is all this to be squashed?

jensmaurer:
Yes, please squash and fix the commit message.

PR #7644, opened 2025-02-14
[dcl.type.auto.deduct] Change plural "occurrences" to singular

lprv:
The rules for decl-specifiers do not allow `auto` to appear more than once in the declared type of a variable, so the plural "occurrences" in this paragraph is potentially misleading.

This wording originates from [N1984](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2006/n1984.pdf). In the paper, "occurrence" is singular, but in [N2009](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2006/n2009.pdf) it somehow gained an "s".

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -2125,7 +2125,7 @@ a declaration of \tcode{std::initializer_list}
 shall precede\iref{basic.lookup.general}
 the \grammarterm{placeholder-type-specifier}.
 Obtain \tcode{P} from
-\tcode{T} by replacing the occurrences of
+\tcode{T} by replacing the occurrence of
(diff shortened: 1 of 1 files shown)


tkoeppe:
@sdutoit: would you happen to have any recollection when this was changed to plural? This was before GitHub...

sdutoit:
That's before my time as well. You’d have to check with Pete Becker I think.

sdutoit:
I suspect it was just a typo while applying the paper though.

tkoeppe:
Yes, interesting, seems like a misapplication. Thanks!

PR #7681, opened 2025-02-16
P1494R5 Partial program correctness

notadragon:
Fixes: https://github.com/cplusplus/draft/issues/7654
Also fixes: https://github.com/cplusplus/papers/issues/376

It's my first paper PR, so please be gentle :)

### DIFF
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -919,23 +919,42 @@ for a given program and a given input.
 Certain other operations are described in this document as
 undefined behavior (for example, the effect of
 attempting to modify a const object).
+
+\pnum
+\indextext{observable checkpoint}
+Certain events in the execution of a program
+are termed \defn{observable checkpoints}.
 \begin{note}
-This document imposes no requirements on the
-behavior of programs that contain undefined behavior.
+A call to \tcode{std::observable}\iref{utility.undefined}
+is an observable checkpoint.
 \end{note}
 
 \pnum
 \indextext{program!well-formed}%
 \indextext{behavior!observable}%
+The \defn{defined prefix} of an execution
+comprises the operations $O$
+for which for every undefined operation $U$
+there is an observable checkpoint $C$
+such that $O$ happens before $C$ and
+$C$ happens before $U$.
+
+\begin{note}
+The undefined behavior that arises from a data race\iref{intro.races}
+occurs on all participating threads.
+\end{note}
+
 A conforming implementation executing a well-formed program shall
-produce the same observable behavior as one of the possible executions
-of the corresponding instance of the abstract machine with the
+produce the observable behavior
+of the defined prefix
+of one of the possible executions
+of the corresponding instance
+of the abstract machine with the
 same program and the same input.
 \indextext{behavior!undefined}%
-However, if any such execution contains an undefined operation, this document places no
-requirement on the implementation executing that program with that input
-(not even with regard to operations preceding the first undefined
-operation).
+If the selected execution contains an undefined operation, 
+the implementation executing that program with that input
+may produce arbitrary additional observable behavior aferwards. 
 If the execution contains an operation specified as having erroneous behavior,
 the implementation is permitted to issue a diagnostic and
 is permitted to terminate the execution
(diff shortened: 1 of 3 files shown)


source/intro.tex
```diff
@@ -919,23 +919,42 @@
 Certain other operations are described in this document as
 undefined behavior (for example, the effect of
 attempting to modify a const object).
+
+\pnum
+\indextext{observable checkpoint}
```
jensmaurer, line 924: drop this

source/intro.tex
```diff
@@ -919,23 +919,42 @@
...
+Certain events in the execution of a program
+are termed \defn{observable checkpoints}.
```
jensmaurer, line 926: ... and use \defnadj{observable}{checkpoint}

You get the indexing for free, then.

source/intro.tex
```diff
@@ -919,23 +919,42 @@
...
-This document imposes no requirements on the
-behavior of programs that contain undefined behavior.
+A call to \tcode{std::observable}\iref{utility.undefined}
+is an observable checkpoint.
 \end{note}
 
 \pnum
 \indextext{program!well-formed}%
 \indextext{behavior!observable}%
+The \defn{defined prefix} of an execution
```
jensmaurer, line 935: \defnadj

source/intro.tex
```diff
@@ -953,23 +972,28 @@
 
 \pnum
 \indextext{conformance requirements}%
-The least requirements on a conforming implementation are:
+The following specify the
+\defnx{observable behavior}{behavior!observable}
```
jensmaurer, line 976: \defnadj does all of this, please use it.

source/utilities.tex
```diff
@@ -670,7 +671,7 @@
 \tcode{static_cast<underlying_type_t<T>>(value)}.
 \end{itemdescr}
 
-\rSec2[utility.unreachable]{Function \tcode{unreachable}}
+\rSec2[utility.undefined]{Undefined behavior}
```
jensmaurer, line 674: This needs a \movedxref entry in xrefdelta.tex

source/utilities.tex
```diff
@@ -704,6 +705,16 @@
 \end{example}
 \end{itemdescr}
 
+\indexlibraryglobal{observable}%
+\begin{itemdecl}
+void observable() noexcept;
+\end{itemdecl}
+\begin{itemdescr}
```
jensmaurer, line 712: empty line after end ... before begin.

jensmaurer (dismissed):

source/utilities.tex
notadragon: xrefprev seems to be missing utility.unreachable.   I'll add it, but it also seems to be missing about 500 other stable names from N4950 that are not in xrefprev.  would you like me to add those?  (It's unclear to me what the intent of the contents of xrefprev is --- whether it should be removed names, or just all names that were previously in use -- either way that is not what it is).

notadragon: ok.  shouldn't it be the list of stable names from any previously published standard?  Or from C++17 to C++23 inclusive?  I could grab that pretty easily if it helps (I already extracted it from C++23), but for this PR I have added utility.uncreachable manually to xrefprev.

jensmaurer: The problem is we've added the label since C++17, and now we're renaming it.  So, it's technically not correct that it was a label in C++17, which xrefprev attempts to represent.

The once-off fix feels reasonable for now.

tkoeppe: My understanding of @StephanTLavavej's request is that we don't just want the diff of the stable labels relative to one arbitrary point in the past, but rather, we want a full history of things that have happened, so that if you've been working with any one version of the document and are switching to a newer one, you can look up what happened to the label you had been working with.

tkoeppe (approved):

source/utilities.tex
notadragon: I definitely think that'd be way more useful than just "here's things that were in C++17 that might have moved".    It might be even more useful to have the index list when the name was moved (i.e, "utility.unreachable (moved in C++26, see utility.undefined)"

PR #7682, opened 2025-02-16
[2025-02 CWG Motion 6] P3475R2 Defang and deprecate memory_order::consume

### LINKED ISSUE #7656: [2025-02 CWG Motion 6] P3475R2 Defang and deprecate memory_order::consume
[P3475R2](https://wiki.edg.com/pub/Wg21hagenberg2025/StrawPolls/P3475R2.pdf) (Defang and deprecate memory_order::consume)

jensmaurer:
Fixes #7656 
Also fixes cplusplus/papers#2129

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -6376,121 +6376,16 @@ the value written'' by the last mutex release.
 \end{note}
 
 \pnum
-An evaluation $A$ \defn{carries a dependency} to an evaluation $B$ if
-\begin{itemize}
-\item
-the value of $A$ is used as an operand of $B$, unless:
-\begin{itemize}
-\item
-$B$ is an invocation of any specialization of
-\tcode{std::kill_dependency}\iref{atomics.order}, or
-\item
-$A$ is the left operand of a built-in logical \logop{and} (\tcode{\&\&},
-see~\ref{expr.log.and}) or logical \logop{or} (\tcode{||}, see~\ref{expr.log.or})
-operator, or
-\item
-$A$ is the left operand of a conditional (\tcode{?:}, see~\ref{expr.cond})
-operator, or
-\item
-$A$ is the left operand of the built-in comma (\tcode{,})
-operator\iref{expr.comma}; \end{itemize} or
-\item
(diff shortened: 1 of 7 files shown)


source/threads.tex
```diff
@@ -6432,7 +6385,6 @@
 
 using std::@\libglobal{memory_order}@;             // \seebelow
 using std::@\libglobal{memory_order_relaxed}@;     // \seebelow
-using std::@\libglobal{memory_order_consume}@;     // \seebelow
```
timsong-cpp, line 6435: Should this be moved to Annex D rather than removed outright?

jensmaurer: Maybe.  The paper contains no mention of this header; I have no idea what the LEWG-approved design intent here might be.

jensmaurer: @jwakely, what do you think?

jensmaurer: Removed that change by force-pushing.

source/basic.tex
```diff
@@ -6306,122 +6306,17 @@
...
 \item $A$ is sequenced before $B$, or
 \item $A$ synchronizes with $B$, or
-\item $A$ simply happens before $X$ and
-$X$ simply happens before $B$.
+\item $A$ happens before $X$ and $X$ happens before $B$.
 \end{itemize}
 \begin{note}
-In the absence of consume operations,
-the happens before and simply happens before relations are identical.
+An evaluation cannot happen before itself.
```
tkoeppe, line 6319: Any chance we could say "does not happen", to avoid problematic modal verbs?

tkoeppe (changes requested):
missing edit on basic.tex new line 6332 (deletion of "simply")

source/basic.tex
jensmaurer: Fixed.

jensmaurer:
> missing edit on basic.tex new line 6332 (deletion of "simply")

Fixed.

tkoeppe (approved):

source/basic.tex
tkoeppe: Thanks!

PR #7687, opened 2025-02-17
[2025-02 LWG Motion 11] P3430R3 simd issues: explicit, unsequenced, identity-element position, and members of disabled simd

### LINKED ISSUE #7670: [2025-02 LWG Motion 11] P3430R3 simd issues: explicit, unsequenced, identity-element position, and members of disabled simd
[P3430R3](https://wiki.edg.com/pub/Wg21hagenberg2025/StrawPolls/P3430R3.pdf) (simd issues: explicit, unsequenced, identity-element position, and members of disabled simd)

mattkretz:
Fixes #7670
Also fixes https://github.com/cplusplus/papers/issues/2097

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -17204,7 +17204,7 @@ namespace std::datapar {
     constexpr basic_simd() noexcept = default;
 
     // \ref{simd.ctor}, \tcode{basic_simd} constructors
-    template<class U> constexpr basic_simd(U&& value) noexcept;
+    template<class U> constexpr explicit(@\seebelow@) basic_simd(U&& value) noexcept;
     template<class U, class UAbi>
       constexpr explicit(@\seebelow@) basic_simd(const basic_simd<U, UAbi>&) noexcept;
     template<class G> constexpr explicit basic_simd(G&& gen) noexcept;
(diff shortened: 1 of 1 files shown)


source/numerics.tex
```diff
@@ -17093,10 +17098,7 @@
 
 \pnum
 \remarks
-The calls to \tcode{gen} are unsequenced with respect to each other.
-Vectorization-unsafe\iref{algorithms.parallel.defns} standard library functions
-may not be invoked by \tcode{gen}.
-\tcode{gen} is invoked exactly once for each $i$.
+\tcode{gen} is invoked exactly once for each $i$, in increasing order of $i$.
```
burblebee, line 17101: This wording is hand-wavy; is that ok because it's just a remark? Same for this wording added elsewhere...

burblebee (approved):

source/numerics.tex
mattkretz: What is hand-wavy about it? And what would you say instead?

burblebee: What is "each $i$"? The closest wording prior to this paper is "if $N = 0$ or if for each $i$ where $0 \le i < N$ in increasing order"

mattkretz: The preceding paragraph says:
> Initializes the $i^\text{th}$ element with
`static_cast<value_type>(gen(integral_constant<simd-size-type, i>()))` for all $i$ in the range of [0,`size()`).

Doesn't that sufficiently explain what "each $i$" means? (And there's no $N = 0$ case to consider here.)

PR #7697, opened 2025-02-19
[time.format] A minor update for LWG 4124

xmcgcg:
The resolution of [LWG 4124](https://open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3504r0.html#4124) added an exception rule for “omitting _chrono-specs_”, therefore [time.format] p7 should begin with “unless otherwise specified” to avoid conflicts.

This PR also added two missing commas in [time.format].

### DIFF
--- a/source/time.tex
+++ b/source/time.tex
@@ -10784,14 +10784,14 @@ the locale's alternate time representation.
 \\ \rowsep
 \tcode{\%y} &
 The last two decimal digits of the year.
-If the result is a single digit it is prefixed by \tcode{0}.
+If the result is a single digit, it is prefixed by \tcode{0}.
 The modified command \tcode{\%Oy} produces the locale's alternative representation.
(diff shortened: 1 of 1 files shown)


jwakely:
> The resolution of [LWG 4124](https://open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3504r0.html#4124) added an exception rule for “omitting _chrono-specs_”, therefore [time.format] p7 should begin with “unless otherwise specified” to avoid conflicts.

I'm not sure what you mean here, what kind of conflicts?

How would it be otherwise specified? The wording is specifying precisely the case where _chrono-specs_ is omitted in a _format-spec_ which uses this `formatter` specialization. There is no otherwise possible, is there?

jwakely:
Oh I see what you mean, p7 says it means one thing, and p18 says otherwise.

But "unless otherwise specified" is always implied. A more specific rule always overrides a more general one.

This wouldn't do any harm though.

PR #7700, opened 2025-02-19
P0472R3 Put std::monostate in <utility>

### LINKED ISSUE #7663: [2025-02 LWG Motion 4] P0472R3 Put std::monostate in ⟨utility⟩
[P0472R3](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p0472r3.pdf) (Put std::monostate in ⟨utility⟩)

Eisenwave:
Fixes #7663.
Fixes https://github.com/cplusplus/papers/issues/1993.

## New Wording
image

## Paper Wording
image

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -220,6 +220,21 @@ namespace std {
       explicit nontype_t() = default;
     };
   template<auto V> constexpr nontype_t<V> nontype{};
+
+  // \ref{variant.monostate}, class \tcode{monostate}%
+\indexlibraryglobal{monostate}
+  struct monostate;
+
+  // \ref{variant.monostate.relops}, \tcode{monostate} relational operators%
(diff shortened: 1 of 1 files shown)


jwakely:
I think there should be an index entry.

Eisenwave:
> I think there should be an index entry.

An index entry for what specifically? `monostate` already has index entries for the class, comparison operators, and the hash specialization. These are found in [variant.monostate] etc.

Do you mean that there should be redundant index entries that also point to these declarations in [utility.syn]? Do we normally index the synopsis instead of the place of definition?

jwakely:
We don't, but it's not redundant here because it is a new occurrence of the name in a different place. 

If the class is declared in two different places, I would expect two index entries.

tkoeppe:
Please make your pull request branches writable in the future :-)

Issue #7666, opened 2025-02-15
[2025-02 LWG Motion 7] P3378R2 constexpr exception types


jensmaurer:
[P3378R2](https://wiki.edg.com/pub/Wg21hagenberg2025/StrawPolls/P3378R2.html) (constexpr exception types)

Eisenwave:
I volunteer to do this.

PR #7701, opened 2025-02-19
P3378R2 constexpr exception types

Eisenwave:
Fixes https://github.com/cplusplus/draft/issues/7666.
Fixes https://github.com/cplusplus/papers/issues/2071.

### DIFF
--- a/source/diagnostics.tex
+++ b/source/diagnostics.tex
@@ -82,8 +82,8 @@ namespace std {
 namespace std {
   class logic_error : public exception {
   public:
-    explicit logic_error(const string& what_arg);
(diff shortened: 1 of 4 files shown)


Eisenwave:
@jwakely two questions:
1. `std::bad_expected_access` (both the primary template and the `<void>` specialization) are missing `\impldef` on `::what()`, unlike `std::bad_optional_access`. Should that be fixed up here?
2. Did I give the correct value to `cpp_lib_constexpr_exceptions`?

jwakely:
Yes and yes, thanks

burblebee (approved):

PR #7702, opened 2025-02-19
[stdexcept.syn] Align class declarations

Eisenwave:


jensmaurer (commented):
The alignment here wants to indicate a derived-from tree.

I don't think there is an urgent need to flatten the indentation.

@tkoeppe ?

tkoeppe:
Agreed. There are a lot of "bespoke" choices like this in the older text, and it's probably not terribly useful to start picking at them like this. Much of it "made sense at the time", and I'd only consider this after some wider survery and plan (but which I agree is really low down on the list of important things).

@Eisenwave If you're looking for things to do, it'd be super helpful to triage some of the pending issues and add some signal regarding whether they're useful, or have an obvious fix, or should just be closed.

Issue #7658, opened 2025-02-15
[2025-02 CWG Motion 8] P2786R13 Trivial Relocatability For C++26


jensmaurer:
[P2786R13](https://wiki.edg.com/pub/Wg21hagenberg2025/StrawPolls/P2786R13.html) (Trivial Relocatability For C++26)

hubert-reinterpretcast:
The Annex C change to [diff.cpp03.dcl.dcl] should be omitted as it is incorrect. This can be done editorially since the change is not to normative text.

AlisdairM:
Please assign to me.

PR #7705, opened 2025-02-19
[2025-02 CWG Motion 8] P2786R13 Trivial Relocation

AlisdairM:
Fixes https://github.com/cplusplus/draft/issues/7658.
Also fixes https://github.com/cplusplus/papers/issues/1463

Added freestanding to the lib feature-test macro, as all new parts are marked as freestanding.

Picked the most appropriate location within the type traits spec for each of the three new traits.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -36,7 +36,7 @@ make a \grammarterm{class-name}. An object of a class consists of a
 
 \begin{bnf}
 \nontermdef{class-head}\br
-    class-key \opt{attribute-specifier-seq} class-head-name \opt{class-virt-specifier} \opt{base-clause}\br
+    class-key \opt{attribute-specifier-seq} class-head-name \opt{class-property-specifier-seq} \opt{base-clause}\br
     class-key \opt{attribute-specifier-seq} \opt{base-clause}
 \end{bnf}
 
@@ -46,8 +46,15 @@ make a \grammarterm{class-name}. An object of a class consists of a
 \end{bnf}
 
 \begin{bnf}
-\nontermdef{class-virt-specifier}\br
-    \keyword{final}
+\nontermdef{class-property-specifier-seq}\br
+    class-property-specifier \opt{class-property-specifier-seq}
+\end{bnf}
+
+\begin{bnf}
+\nontermdef{class-property-specifier}\br
+    \keyword{final}\br
+    \keyword{trivially_relocatable_if_eligible}\br
+    \keyword{replaceable_if_eligible}
 \end{bnf}
 
 \begin{bnf}
@@ -120,12 +127,13 @@ A union holds the value of at most one data member at a time.
 \end{note}
 
 \pnum
-If a class is marked with the \grammarterm{class-virt-specifier} \tcode{final} and it appears
-as a \grammarterm{class-or-decltype} in a \grammarterm{base-clause}\iref{class.derived},
-the program is ill-formed. Whenever a
-\grammarterm{class-key} is followed by a \grammarterm{class-head-name}, the
-\grammarterm{identifier} \tcode{final}, and a colon or left brace, \tcode{final} is
-interpreted as a \grammarterm{class-virt-specifier}.
+Each \grammarterm{class-property-specifier} shall appear at most once
+withina single \grammarterm{class-property-specifier-seq}.
+Whenever a \grammarterm{class-key} is followed
+by a \grammarterm{class-head-name},
+the identifier \tcode{final}, \tcode{trivially_relocatable_if_eligible},
+or \tcode{replaceable_if_eligible}, and a colon or left brace,
+the identifier is interpreted as a \grammarterm{class-property-specifier}.
 \begin{example}
 \begin{codeblock}
 struct A;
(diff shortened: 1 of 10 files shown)


source/classes.tex
```diff
@@ -36,7 +36,7 @@
 
 \begin{bnf}
 \nontermdef{class-head}\br
-    class-key \opt{attribute-specifier-seq} class-head-name \opt{class-virt-specifier} \opt{base-clause}\br
+    class-key \opt{attribute-specifier-seq} class-head-name \opt{class-property-specifier-seq} \opt{base-clause}\br
```
AlisdairM, line 39: It looks like a cleaner fix would be to move the optional _class-property-specifier_ onto the _class-head-name_ production, but that is probably a bigger change than should be considered editorial.

jensmaurer: Why would that be a "cleaner fix"?  What's the "fix" part here?  Why cleaner?

AlisdairM: We have two productions, but the optional class-property-specifier-seq is valid only after a class-head-name.  We could move that optional element into the class-head-name and then the optionality of the class-head-name itself becomes more apparent.

If we move the optional class-property-specifier-seq into the class-head-name, we could even simplify the grammar further by making the class-head-name itself optional, needing only one grammar production and not two to say the same thing.

I have confirmed (to the best of my ability) that there are no other uses of the class-head-name production that would become confused by having this extra optional component suddenly showing up.

source/lib-intro.tex
```diff
@@ -3826,6 +3826,25 @@
...
+\pnum
+Unless specifically stated, it is unspecified whether any class for which
+trivial relocation (i.e., the effects of \tcode{trivially_relocate}) would be
+semantically equivalent to move-construction of the destination object followed
+by destruction of the source object is trivially relocatable.
+
+\pnum
+Unless specifically stated, it is unspecified whether any class for which move
+assignment is semantically equivalent to destroying the assigned-to object,
+then move-constructing from the source object in its place is replaceable.
```
lorihughes, line 3846: The comma after object is causing grammatical issues in this sentence. 
Unless specifically stated, it is unspecified whether X is replaceable. 
The comma separates X from its verb, "is replaceable."
My suggested correction is 
Unless specifically stated, it is unspecified whether any class for which move
assignment is semantically equivalent to destroying the assigned-to object and 
then move-constructing from the source object in its place is replaceable.

SentenceDiagram

jwakely: "Unless specifically stated" isn't what we'd usually say, is it? Was "unless otherwise specified" avoided so we don't have "specified ... unspecified"?

Even without that comma, we still have "whether any class" very separate from "is replaceable".

How about more like:

> Except where explicitly stated, it is unspecified whether any class is replaceable if that class' move assignment is semantically equivalent to destroying the assigned-to object and then move-constructing from the source object in its place.

Are "assigned-to object" and "from the source object" clear enough? It be better to introduce two objects a and b and refer to them by name?

jwakely: The previous paragraph has the same structure of course, so should match. I'm not sure my suggestion makes things any better though.

lorihughes: Yes, we were trying to avoid being repetitious with the "specified" and "unspecified".
 
In a perfect world, we would get rid of the "it is" because what is meant by that "it" is "whether any class for which move assignment is semantically equivalent to destroying the assigned-to object and then move-constructing from the source object in its place is replaceable"

Except where explicitly stated, whether any class is replaceable if that class's move assignment is semantically equivalent to destroying the assigned-to object and then move-constructing from the source object in its place is unspecified.

Another idea: 
Whether any class is replaceable if that class's move assignment is semantically equivalent to destroying the assigned-to object and then move-constructing from the source object in its place is unspecified except where explicitly stated.

source/classes.tex
jensmaurer: If we ever put more things into the class-property-specifier-seq, in particular things that need name lookup (e.g. when the "bool parameter" stuff shows up in another context), then [basic.scope.namespace] p1 and [basic.scope.class] p1 would go wrong. Also [expr.prim.id.qual] p2.1 .

Looks like such a grammar refactoring would not be forward-evolution-friendly.

To sum up, I think there is nothing to act upon in this editorial pull request.

source/lib-intro.tex
```diff
@@ -3826,6 +3826,25 @@
 side effects.
 \end{note}
 
+\rSec3[library.class.props]{Properties of library classes}
+
+\pnum
+Unless specifically stated, it is unspecified whether any class described in
```
jensmaurer, line 3832: We don't say "Unless specifically stated" elsewhere, but we do use "Unless explicitly stated otherwise," a few times.  I think I would prefer the latter phrasing, putting more emphasis on deviation ("otherwise").

lorihughes: I like "Unless explicitly stated otherwise"

source/lib-intro.tex
lorihughes: Overall, your suggestion is much better, Jens, and I will accept the pain of "it is"

jwakely: I like that, and it also avoids my next question, which was going to be whether "move assigning" means invoking a move assignment operator specifically, or just assigning an xvalue, and whether the distinction even matters here (I don't think it does).

Does it matter whether we say "move-construction from a" or "constructing from `std::move(a) in b's place"?

jensmaurer: a is already an xvalue, so we don't need `std::move` around it.  Yes, we can drop the move- from "constructing".

AlisdairM: Integrated feedback and added cross-references --- I believe this thread should be addressed now.

source/lib-intro.tex
AlisdairM: Integrated feedback and added cross-references.

PR #7710, opened 2025-02-24
[expr.const] Add non-breaking space after `\cv`

cpplearner:
`\cv \tcode{std::nullptr_t}` produces roughly "*cv*`std::nullptr_t`" with no space in-between. This PR adds a non-breaking space.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -7750,7 +7750,7 @@ an lvalue-to-rvalue conversion\iref{conv.lval} unless
 it is applied to
 \begin{itemize}
   \item
-  a glvalue of type \cv \tcode{std::nullptr_t},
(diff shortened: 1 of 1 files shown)


source/expressions.tex
```diff
@@ -7750,7 +7750,7 @@
 it is applied to
 \begin{itemize}
   \item
-  a glvalue of type \cv \tcode{std::nullptr_t},
+  a glvalue of type \cv~\tcode{std::nullptr_t},
```
tkoeppe, line 7753: The usual spelling seems to be `\cv{}~...`, could you please amend accordingly?

cpplearner: done

jensmaurer (approved):

PR #7712, opened 2025-02-25
support for core-issue 675, requires-expression as a barrier for invalid types or expressions.

ingmaurorusso:
This pull-request supports the suggested solution for the core issue https://github.com/cplusplus/CWG/issues/675

### DIFF
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -919,23 +919,41 @@ for a given program and a given input.
 Certain other operations are described in this document as
 undefined behavior (for example, the effect of
 attempting to modify a const object).
+
+\pnum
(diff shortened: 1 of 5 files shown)


jwakely:
It probably makes more sense to just keep it in your own fork rather than submit a PR for it.

ingmaurorusso:
> It probably makes more sense to just keep it in your own fork rather than submit a PR for it.

ok, I see.

If you believe it's more viable, I may close this PR and I will just refer in [issue-675](https://github.com/cplusplus/CWG/issues/675) a branch from my fork.

Sorry for learning refinement about the best way to interact.

PR #7713, opened 2025-02-25
[support.srcloc.cons] Update xref to [class.mem.general]

jwakely:
The cross-reference to [class.mem] was referring to a hanging paragraph that was fixed by 2850139be6285ba10a64fb718125a80ca967c631 so we should be referring to [class.mem.general] now.


jensmaurer (approved):

tkoeppe:
Is all this to be squashed?

PR #7729, opened 2025-03-11
 [print.fun] Follow up P1494R5

xmcgcg:
P1494R5 (and PR[#7681](https://github.com/cplusplus/draft/pull/7681)) overlooked the `vprint_nonunicode` overload taking `FILE*`. This PR fixes that.


jensmaurer:
This seems to be a normative change, not an editorial one. Please submit an LWG issue. https://isocpp.org/std/submit-issue

PR #7738, opened 2025-03-15
merge/reword "two kinds of implementations" wording from [library] into [intro.compliance]

tkoeppe:
From #7703.

### DIFF
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -833,18 +833,17 @@ translation units to form a complete \Cpp{}  program\iref{lex.phases}.%
 \indextext{conformance requirements!library|)}
 
 \pnum
-An implementation is either a
-\defnadj{hosted}{implementation} or a
-\defnadj{freestanding}{implementation}.
-A freestanding
-implementation is one in which execution may take place without the benefit of
-an operating system.
-A hosted implementation
-supports all the facilities described in this document, while
-a freestanding implementation
+A \defnadj{freestanding}{implementation}
+is one in which execution may take place
+without the benefit of an operating system.
+A \defnadj{hosted}{implementation}
+supports all the facilities described in this document,
+while a freestanding implementation
 supports the entire \Cpp{} language
 described in \ref{lex} through \ref{\lastcorechapter} and
 the subset of the library facilities described in \ref{compliance}.
+It is \impldef{whether the implementation is hosted or freestanding}
+whether the implementation is hosted or freestanding.
 
 \pnum
 It is
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -1579,13 +1579,8 @@ do not make macros visible\iref{module.import}, such as
 \indextext{implementation!freestanding|(}%
 
 \pnum
-Two kinds of implementations are defined:
-\indextext{implementation!hosted}%
-hosted and freestanding\iref{intro.compliance};
-the kind of the implementation is
-\impldef{whether the implementation is hosted or freestanding}.
-For a hosted implementation, this document
-describes the set of available headers.
+For a hosted implementation,
+this document describes the set of available headers.
 
 \pnum
 A freestanding implementation has an


tkoeppe:
@Eisenwave, @jwakely, @burblebee: I made this suggestion to remove the redundancy into a separate PR.

source/intro.tex
```diff
@@ -833,18 +833,17 @@
...
-a freestanding implementation
+A \defnadj{freestanding}{implementation}
+is one in which execution may take place
+without the benefit of an operating system.
+A \defnadj{hosted}{implementation}
+supports all the facilities described in this document,
+while a freestanding implementation
 supports the entire \Cpp{} language
 described in \ref{lex} through \ref{\lastcorechapter} and
 the subset of the library facilities described in \ref{compliance}.
```
Eisenwave, line 844: Looking at it again, I think it would read nicer if we had one sentence entirely focused on hosted implementations (that also defines that term, followed by another sentence focused entirely on freestanding implementations.

I don't like how the definition of hosted implementations also has some explanation of freestanding in it.

source/intro.tex
```diff
@@ -833,18 +833,17 @@
...
+It is \impldef{whether the implementation is hosted or freestanding}
+whether the implementation is hosted or freestanding.
```
jwakely, line 846: Can it be both? Can it be neither? With the old wording, clearly no ("An implementation is either a hosted implementation or a freestanding implementation").

source/lib-intro.tex
```diff
@@ -1579,13 +1579,8 @@
...
 \pnum
-Two kinds of implementations are defined:
-\indextext{implementation!hosted}%
-hosted and freestanding\iref{intro.compliance};
-the kind of the implementation is
-\impldef{whether the implementation is hosted or freestanding}.
-For a hosted implementation, this document
-describes the set of available headers.
+For a hosted implementation,
+this document describes the set of available headers.
```
jwakely, line 1583: It seems a bit odd to say this in the subclause titled "Freestanding implementations".

Separately from that, I think xref to [intro.compliance] for the definitions of host impl and freestanding impl would help.

source/intro.tex
jwakely: There isn't really a definition of hosted implementation, it's just one that is not freestanding. Does it have an OS? Not necessarily, all that's required is that the full set of features are available (but those could be linked directly into the program, like a unikernel that runs on bare metal or directly on a hypervisor).

It could be worded like this, but I don't love it:

```
A \defnadj{hosted}{implementation}
supports all the facilities described in this document.
A \defnadj{freestanding}{implementation}
is one in which execution may take place
without the benefit of an operating system.
A freestanding implementation supports the entire \Cpp{} language
described in \ref{lex} through \ref{\lastcorechapter} and
the subset of the library facilities described in \ref{compliance}.
```

C has separate definitions for hosted/freestanding implementations, and hosted/freestanding execution environments. But I'm not sure what advantage that has (can you use a hosted implementation to compile code for a freestanding environment? how would you use features such as I/O if the execution environment doesn't support them?).

jwakely: There isn't really a definition of hosted implementation, it's just one that is not freestanding. Does it have an OS? Not necessarily, all that's required is that the full set of features are available (but those could be linked directly into the program, like a unikernel that runs on bare metal or directly on a hypervisor).

It could be worded like this, but I don't love it:

```
A \defnadj{hosted}{implementation}
supports all the facilities described in this document.
A \defnadj{freestanding}{implementation}
is one in which execution may take place
without the benefit of an operating system.
A freestanding implementation supports the entire \Cpp{} language
described in \ref{lex} through \ref{\lastcorechapter} and
the subset of the library facilities described in \ref{compliance}.
```

C has separate definitions for hosted/freestanding implementations, and hosted/freestanding execution environments. But I'm not sure what advantage that has (can you use a hosted implementation to compile code for a freestanding environment? how would you use features such as I/O if the execution environment doesn't support them?).

PR #7746, opened 2025-03-18
[sequences.general] Hive is a sequence container

AlisdairM:


jwakely (approved):

source/containers.tex
```diff
@@ -6058,6 +6058,7 @@
 \libheaderref{array},
 \libheaderref{deque},
 \libheaderrefx{forward_list}{forward.list.syn},
+\libheaderrefx{hive}{hive.syn},
```
jensmaurer, line 6061: Since "hive.syn" is regular (derived from "hive"), please use `\libheaderref`.

jensmaurer (changes requested):

source/containers.tex
AlisdairM: Done

jensmaurer (approved):

Issue #7751, opened 2025-03-19
[variant.ctor] missing `Throws` clause on the last ctor


Tsche:
The last constructor in [variant.ctor](https://eel.is/c++draft/variant.ctor#35) is missing the `Throws` clause. I believe this inconsistency is a simple editorial issue, since the three preceeding (very similar!) constructors all explicitly say
> _Throws_: Any exception thrown by calling the selected constructor of T(i)

This should be added to the constructor
```cpp
template<size_t I, class U, class... Args>
  constexpr explicit variant(in_place_index_t<I>, initializer_list<U> il, Args&&... args);
```
as well.

tkoeppe:
To be handled as an LWG issue. Thank you for reporting!

jwakely:
This is now https://cplusplus.github.io/LWG/issue4460

tkoeppe:
Thank you!

PR #7752, opened 2025-03-20
[variant.ctor] Add missing throws clause to last ctor

Tsche:
Fixes https://github.com/cplusplus/draft/issues/7751

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -5475,6 +5475,10 @@ with \tcode{il, std::forward<Args>(\brk{}args)...}.
 \tcode{index()} is \tcode{I}.
 
 \pnum
+\throws
+Any exception thrown by calling the selected constructor of $\tcode{T}_I$.
+
+\pnum
 \remarks
 If $\tcode{T}_I$'s selected constructor is a constexpr constructor, this
 constructor is a constexpr constructor.


jensmaurer:
@Tsche , if you want your real name to appear in the git commit history, please configure it in the github profile settings.

@jwakely, does this need an LWG issue?

jwakely:
I'll do some archaeology and see if it's been discussed or just forgotten (probably the latter)

Tsche:
> @Tsche , if you want your real name to appear in the git commit history, please configure it in the github profile settings.

Thanks. I'd prefer committing it under this pseudonym instead if that's okay with you.

Tsche:
@jwakely I've done some digging but could not find any discussion about this in old meeting minutes. [P0088r3](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0088r3.html) and earlier versions thereof already miss the throws clause on that ctor.

It looks like this defect is an artifact of a change between [P0088R0](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/p0088r0.pdf) and [P0088R1](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0088r1.html). Note how in R0 neither one of the `emplaced_type_t`/`emplaced_index_t` (as they were then called) + `initializer_list` constructors have a throws clause. In R1 only one of them gained it.

This seems like a simple oversight to me. How would you like to proceed?

tkoeppe:
Discussed: let's handle this as an LWG issue.

PR #7753, opened 2025-03-20
Re-enable MacOS runner

tkoeppe:


source/macros.tex
```diff
@@ -724,6 +722,19 @@
...
+% "nc" are the non-copied versions of bnf environments; the non-"nc"
+% versions are copied to the grammar extraction file.
+\NewEnviron{bnf}{\begin{ncbnf}%
+\BODY%
+\gramWrite{\string\begin{ncbnf}\meaningbody\BODY\string\end{ncbnf}}%
+\end{ncbnf}}{}
+
+\NewEnviron{simplebnf}{\begin{ncsimplebnf}%
+\BODY%
+\gramwrite{\string\begin{ncsimplebnf}\meaningbody\BODY\string\end{ncsimplebnf}}%
```
jensmaurer, line 734:
```suggestion
\gramWrite{\string\begin{ncsimplebnf}\meaningbody\BODY\string\end{ncsimplebnf}}%
```

jensmaurer (commented):
Rebased on main and testing.

The "gramwrite" spelling error was not flagged by LaTeX; strange.  Is this ever invoked?

The commit message of the main change has a typo "exract".

jensmaurer:
"diffpdf" in "Appearance" mode shows no differences. Please merge after fixing the minor findings.

jwakely (commented):
The check.yml commit has an odd summary: "MacOS Linux runner"

Was "Linux" meant to say something else?

tkoeppe:
@jwakely Yeah, sorry, the genesis of this PR is a bit chaotic. I'll apply some of the commits directly and repurpose/rewrite the runner based things appropriately. Thanks!

tkoeppe:
> Was "Linux" meant to say something else?

(I deleted the wrong word. Initially the commit said "Update MacOS and Linux runners". I meant to delete "Linux", not "Update".)

source/macros.tex
tkoeppe: So I suppose we should rename `ncsimplebnf` and `ncrebnf`? We can drop the "nc" there if we are reasonably sure that `bnf` is the only environment we'll ever want extracted.

jensmaurer: I think it's plausible to assume we only ever want "bnf" to be extracted; everything else are either grammar quotes or non-C++ grammar things (such as regex).

PR #7766, opened 2025-03-23
[stopsource.general] Restore accidentally deleted members in class definition

timsong-cpp:
These were removed in error in the application of P2300R10.


jensmaurer (approved):
Checked against https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p2300r10.html#spec-stopsource

PR #7770, opened 2025-03-24
[ranges] Make [range.approximately.sized] close to [range.sized]

### LINKED ISSUE #7769: [range.approximately.sized]: Move it before [range.sized]?
[range.approximately.sized] is currently sandwiched between [range.prim.size.hint] and [range.prim.empty], which looks disharmonious. 
Moving it before [range.sized] seems a bit more consistent in my opinion.

hewillk:
Fix #7769.


jensmaurer:
Duplicated due to clerical error in #7807

PR #7786, opened 2025-03-27
[expr.cond] Simplify phrasing of the `nullptr_t` case

lprv:
The case where both the second and third operands have the same type is handled by the first bullet of this list.


tkoeppe:
But the first bullet is constrained to when both operands have the same value category, too, isn't it?

lprv:
I don't think so. The bullet I'm referring to is this:
> * The second and third operands have the same type; the result is of that type and the result is copy-initialized using the selected operand.

PR #7787, opened 2025-03-27
[expr.type] Remove redundant `\cvqual`

lprv:


jensmaurer (approved):
"reference-related" already strips cv-qualification.

PR #7790, opened 2025-03-27
[temp.constr.concept] Avoid talking about normalization of constraints

lprv:


jensmaurer (commented):
I think this is intentional.  We do some weird normalization here.

tkoeppe:
@hubert-reinterpretcast, @opensdh could you speak to whether the status quo needs to be cleaned up here?

tkoeppe:
Or @cor3ntin?

cor3ntin:
I think normalization is what we mean here. Yes, we will also check satisfaction recursively but this is not notable. however, the fact that concept dependent constraints would be further normalized as they become non dependent _is_ notable, as it is unusual.

status quo is therefore preferable.

tkoeppe:
Thank you!

lprv:
So am I wrong in saying that "[normalization](https://eel.is/c++draft/temp.constr.normal#def:constraint,normalization) of [...] constraints" is a category error? Because that is the main concern of this PR, and I don't think it was addressed by the above explanation.

We could find a form that keeps the term "normalization" in the sentence without using it incorrectly; my previous comment provides one suggestion.

hubert-reinterpretcast:
> So am I wrong in saying that "[normalization](https://eel.is/c++draft/temp.constr.normal#def:constraint,normalization) of [...] constraints" is a category error?

You're not wrong. We will need to see how a concept-dependent constraint is structured to see what is getting normalized.

hubert-reinterpretcast:
@lprv, please open an issue re: the category error here: https://github.com/cplusplus/CWG/issues

PR #7793, opened 2025-03-27
[temp.constr.normal] Break up a long comment

lprv:


jensmaurer (approved):
Repair commit description to "Rephrase comment in example"

PR #7794, opened 2025-03-27
[temp.constr.normal] Fix example

lprv:
* Avoid talking about the "normal form" of constraints
* Fix typo

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -2208,9 +2208,9 @@ template<typename U,
 template<typename U>
   void g() requires CC<U*, C>;
 \end{codeblock}
-The normal form of the associated constraints of \tcode{f} is
-the concept-dependent constraint \tcode{CT<T, C>}.\\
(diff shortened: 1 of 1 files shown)


source/templates.tex
```diff
@@ -2208,9 +2208,9 @@
 template<typename U>
   void g() requires CC<U*, C>;
 \end{codeblock}
-The normal form of the associated constraints of \tcode{f} is
-the concept-dependent constraint \tcode{CT<T, C>}.\\
-The normal form of the associated constraints of \tcode{g} is
+The associated constraints of \tcode{f} are
+the concept-dependent constraint \tcode{CT<U*, C>}.\\
+The associated constraints of \tcode{g} are
```
tkoeppe: The mismatch in number is a bit odd now. Maybe "The associated constraints of `g` consist of the atomic constraint `true`"?

lprv: Applied, thanks.

PR #7795, opened 2025-03-27
[temp.constr.order] Fix example

lprv:
The comment is in disagreement with the normative text.


jensmaurer (commented):
Why is `&& true` added here?

I think the intent is that the example is correct; we effectively want to ignore X&lt;T&gt; for subsumption considerations.  Maybe the normative words aren't there yet; suggestions welcome.

PR #7807, opened 2025-03-30
[range.approximately.sized] Move to before [range.sized]

### LINKED ISSUE #7769: [range.approximately.sized]: Move it before [range.sized]?

jensmaurer:
This is an unfortunate application of P2846R6.

Fixes #7769


tkoeppe:
Indeed, thanks, the context is sometimes easy to miss in review!

PR #7822, opened 2025-04-11
 [indirect.asgn] std::indirect does not have a “contained value”

xmcgcg:
[indirect.asgn] p4 says:

> If an exception is thrown during the call to `T`’s copy assignment, the state of its **contained value** is as defined by the exception safety guarantee of `T`’s copy assignment.

The wording is copied from the description of the copy assignment operator of `std::optional`, which has a contained value (defined in [optional.optional.general] p1). For `std::indirect`, there is no “contained value”, “owned object” should be used instead.


tkoeppe:
@jbcoe: Any thoughts?

jbcoe:
> owned object

"owned object" sounds correct.

jwakely (approved):

PR #7831, opened 2025-04-13
[mdspan.layout.left.obs] Add missing noexcept

hewillk:
Function declarations have noexcept, so this should be a typo.


jwakely (approved):
Having `noexcept` here violates the Lakos rule but is consistent with the other layout mappings, and with the synopsis (as noted).

jensmaurer (approved):

PR #7833, opened 2025-04-13
Remove incorrect class.union.assignment.not.start.lifetime

jensmaurer:
This references undefined behavior specified elsewhere; it does not specify additional undefined behavior.


hsutter (approved):
Looks good, thank you.

shafik (approved):
LGTM

PR #7834, opened 2025-04-13
Rename label expr.ass.overlap to expr.assign.overlap

jensmaurer:


tkoeppe:
Hm, we did end up recording the other changes, too (https://github.com/cplusplus/draft/blob/8792e5544498d262a821b2175c1fe52fd3a15156/source/xrefdelta.tex#L482-L484), so maybe we should also record this one?

tkoeppe (approved):

jensmaurer:
Note the "ub-ifndr" label: This is for preparing a pull request to create a UB/IFNDR Annex.  I'm guessing we'll just squash everything before merging to the Working Draft.
Plus, these are labels for UB, which are currently invisible in the generated PDF. So, recording changes for them in xrefdelta seems triple out-of-scope.

PR #7860, opened 2025-05-01
[polymorphic.ctor] Owned object must have cv-unqualified object type

jwakely:


jensmaurer (approved):

jwakely:
N.B. The incoming paper did say `U` here, but I think this is editorial because it's inconsistent with the rest of the `std::polymoprhic` spec. The constructors taking `in_place_type_t<U>` to specify the type all require `is_same_v<remove_cvref_t<U>, U>` to be true, so these two constructors are inconsistent. It's unclear what it would even mean to store e.g. `const X&` as the owned object. So `remove_cvref_t<U>` is the only reasonable interpretation, and that's what `UU` is.

jbcoe:
I agree with @jwakely s analysis.

Twon:
Yep, agreed, that makes sense.

PR #7861, opened 2025-05-01
[indirect.asgn], [polymorphic.asgn] Combine related bullets

jwakely:
The bullet talking about destroying the previously owned object is only applicable to the bullet immediately before it. The current structure makes it appear to be applicable to all three of the cases that are separated by "otherwise" and are mutually exclusive.


jwakely:
I've changed this to a draft because I've actually proposed a change that would make the current structure correct for move assignment operators, see https://cplusplus.github.io/LWG/issue4251

PR #7866, opened 2025-05-06
[syntax] to [intro.syntax] to match the other subclause stable names

0ABE:
The stable names for other subclauses in this clause are prepended with "intro." Make the [syntax] stable name follow the same pattern by changing it to [intro.syntax].


jwakely:
But then they won't be stable.

0ABE:
That is one way to look at it, and I wondered about that myself. But does stable mean written in stone?  I decided to find out 😀 

My impression was that these names are called stable because the LaTeX automatic numbering system will not change them with the addition/removal/reorganization of other (sub/)sections and that is why the LaTeX generated numbers are considered to be unstable.

Purposeful and deliberate changes such as this could be accepted to bring consistency to the naming scheme of the stable names.

PR #7872, opened 2025-05-13
[mdspan.layout.right.obs] Add missing constexpr to required_span_size.

1uc:
The missing `constexpr` looks like a typo, because it's present in the overview; and every other method is constexpr.


jwakely:
The constexpr was missing in the incoming paper, [P0009R18], but I agree that it looks like an editorial mistake.

mhoemmen:
@jensmaurer Thanks for the heads-up!  Thanks @1uc for pointing this out!

I agree that this must have been an editorial mistake.  It was certainly the intention for `required_span_size()` to be marked `constexpr` in `layout_right::mapping`.  It's already marked `constexpr` in the other two C++23 layout mappings (`layout_left::mapping` and `layout_stride::mapping`).

PR #7874, opened 2025-05-14
 [associative.reqmts.general] Eliminate the final usage of “multiple keys”

xmcgcg:
Associative containers possibly containing multiple elements for each key are said to support “equivalent keys”, but there is one place in [containers] that uses the term “multiple keys”. We should eliminate that.


jwakely (approved):
Yes, this is the correct terminology as defined in [associative.reqmts.general].

jensmaurer (approved):

PR #7878, opened 2025-05-16
[re.regiter.general] Move const into \tcode

Andreas-Krug:


source/text.tex
```diff
@@ -12238,8 +12238,8 @@
 constructs an end-of-sequence iterator object,
 which is the only legitimate iterator to be used for the end
```
jwakely, line 12239: This is somewhat misleading since we added comparison with `default_sentinel_t` although that's not actually an iterator so it's technically true, just misleading.

source/text.tex
```diff
@@ -12238,8 +12238,8 @@
...
 condition. The result of \tcode{operator*} on an end-of-sequence iterator is not
```
jwakely, line 12240: Most of this prose is just repeating general iterator requirements (you can't dereference the end iterator, dereferencing it returns the iterator's reference type etc).

jwakely (approved):
The change is correct, but this wording is generally in need of improvements.

PR #7890, opened 2025-05-23
[ub] small fixes to examples

notadragon:


source/ub.tex
```diff
@@ -1674,9 +1672,9 @@
 }
 void f()
 {
-  X& px = &g();
+  X* px = &g();
   px->~X();
-  int*p = px->i;    // undefined behavior
+  int p = px->i;    // undefined behavior
```
hsutter, line 1677: Minor: Maybe change this to `int j` since the name `p` might connote a pointer.

hsutter (approved):

PR #7892, opened 2025-05-26
[container.requirements] Simplify Returns specification for try_emplace

### LINKED ISSUE #7881: [container.requirements] Confusing "Returns" paragraph under `a.emplace_hint(p, args)`
>An iterator pointing to the element with the key equivalent to the newly inserted element.

* In containers with unique keys, an element is not always inserted, so there might be no "newly inserted element".
* In containers with equivalent keys, there might be more than one element "with the key equivalent to the newly inserted element".

Eisenwave:
Fixes #7881.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -3239,8 +3239,7 @@ the position just prior to \tcode{p}.
 
 \pnum
 \returns
-An iterator pointing to the element
-with the key equivalent to the newly inserted element.
+The iterator returned by \tcode{emplace}.
 
 \pnum
 \complexity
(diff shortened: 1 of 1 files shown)


jensmaurer:
@jwakely , since we already say we do as-if `emplace` in the "Effects", this change seems to be a nice clarification.  What do you think?

jwakely:
Why not just _Equivalent to_: Returns ... except that ... ?

source/containers.tex
```diff
@@ -4941,26 +4936,12 @@
...
 \effects
-Equivalent to \tcode{a.emplace(std::forward<Args>(args)...)}.
-
-\pnum
-\returns
-An iterator pointing to the element
-with the key equivalent to the newly inserted element.
-The \tcode{const_iterator} \tcode{p} is a hint
+Equivalent to: \tcode{return a.emplace(std::forward<Args>(args)...).first},
+expect that the \tcode{const_iterator} \tcode{p} is a hint
```
jwakely, line 4942:
```suggestion
except that the \tcode{const_iterator} \tcode{p} is a hint
```

source/containers.tex
```diff
@@ -3233,15 +3233,10 @@
 
 \pnum
 \effects
-Equivalent to \tcode{a.emplace(std::forward<Args>(args)...)},
+Equivalent to: \tcode{return a.emplace(std::forward<Args>(args)...).first},
```
jwakely, line 3236: I've just realised that this doesn't work, because `a_uniq.emplace` and `a_eq.emplace` return different things. It's not a pair for `a_eq`, so you can't do `.first`

So maybe what you had before was better, sorry.

jwakely (changes requested):

source/containers.tex
Eisenwave: I must not have been paying attention either; my bad. The latest commit should resolve everything.

jensmaurer (approved):

jwakely (approved):

PR #7894, opened 2025-05-27
Update README.rst

Eisenwave:
This PR makes the following changes:

- Render the suggested package installation commands in code font for increased readability.
- Suggest the alternative of using WSL on Windows. This is my personal setup, and considering that some of the scripts we have use Bash, our tooling is already biased towards Linux, so this is worth suggesting.
- Mention the `check-source.sh` and `check-output.sh` scripts and how to use them. Especially considering that our CI doesn't work properly right now, it's a relevant thing to mention.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -24,7 +24,7 @@ Install the `MacTeX distribution <http://tug.org/mactex/>`_.
 If you are on a slow network, you'll want to get the `BasicTeX package <http://tug.org/mactex/morepackages.html>`_ instead,
 then run the following command to install the other packages that the draft requires:
 
-   sudo tlmgr install latexmk isodate substr relsize ulem fixme rsfs extract layouts enumitem l3packages l3kernel imakeidx splitindex xstring
+   ``sudo tlmgr install latexmk isodate substr relsize ulem fixme rsfs extract layouts enumitem l3packages l3kernel imakeidx splitindex xstring``
 
 ---------------------------------------
 Getting Started on Debian-based Systems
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -59,6 +59,10 @@ See `sample instructions <https://bennierobinson.com/programming/2016/01/24/perl
 
 Install `MiKTeX <https://miktex.org/download>`_
 
+Alternatively, you can `install WSL <https://learn.microsoft.com/en-us/windows/wsl/install>`_
```
tkoeppe, line 62: Is there a stray space at the end here?

tkoeppe:
Please rebase.

README.rst
eisenwave: I'm not seeing one.

jwakely: It's a trailing underscore which turns the text "install WSL" into a hyperlink with that embedded URL. These docs say to use a double underscore there though, for an anonymous reference: 

https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#external-links

But a single underscore works in this case and is consistent with the link above.

jwakely:
> Render the suggested package installation commands in code font for increased readability.

Doesn't that already happen because of the indentation?

README.rst
tkoeppe: Thanks!

eisenwave:
> Doesn't that already happen because of the indentation?

It does. I'm not sure why I thought it doesn't at the time. This may be some weird Mandela effect thing - I am pretty sure that it rendered like

> a Markdown blockquote

at the time, and I added code font to have it

> `indented and in code font`.

Something must have changed about the way GitHub renders it, or I'm getting Mandela-effected pretty hard right now. Weird; lemme revert those changes.

jwakely:
> Doesn't that already happen because of the indentation?

More precisely, it's the `::` on the previous paragraph _and_ the indentation.

https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks

PR #7905, opened 2025-05-31
fix spelling of exposition-only `query-with-default` function

ericniebler:


jensmaurer (approved):
@jwakely, I'd appreciate a look.

PR #7907, opened 2025-06-01
Fix function type name in [exec.let]

ericniebler:
The specification of `let_value(sndr,fn)` and friends refers to the type of `fn` as "`Fn`". But when introducing the types, we give it the name "`F`". This fixes the inconsistency.

### DIFF
--- a/source/exec.tex
+++ b/source/exec.tex
@@ -3536,7 +3536,7 @@ except with duplicate types removed.
 Given a type \tcode{Tag} and a pack \tcode{Args},
 let \exposid{as-sndr2} be an alias template
 such that \tcode{\exposid{as-sndr2}<Tag(Args...)>} denotes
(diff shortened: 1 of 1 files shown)


jensmaurer:
Where exactly should I look to see other uses of "F" (not "Fn")?

ericniebler:
here: https://eel.is/c++draft/exec.let#3

jensmaurer:
@jwakely, this looks reasonable.  What do you think?

jwakely:
> The specification of `let_value(sndr,fn)` and friends refers to the type of `fn` as "`Fn`". But when introducing the types, we give it the name "`F`". This fixes the inconsistency.

I think "fn" in this comment should be "f", as the wording uses `F` and `f` consistently, except this one use of `Fn` (there are uses of "fn" but that's a structured binding not the same as `f` used elsewhere).

jwakely (approved):

PR #7915, opened 2025-06-13
[expected.object.cons] Reorder arguments of is_same_v for consistency

Dani-Hub:
During the discussion of LWG 4222 it was noticed that [expected.object.cons] bullet (23.2) uses an atypical order of the is_same_v arguments, especially since this issue adds another bullet. To enforce consistency, the `remove_cvref_t<U>` argument is no also used as first (instead of second) argument in that bullet.


Dani-Hub:
The suggestion on the LWG reflector was to submit this as editorial issue to harmonize the suggested proposed wording for [LWG 4222](https://cplusplus.github.io/LWG/issue4222).

jensmaurer:
Looks obviously harmless.

PR #7917, opened 2025-06-16
[rand.eng.philox] Use normal font for superscript of "ith"

jwakely:
Consistent with how we say <i>i</i><sup>th</sup> elsewhere, e.g. [tuple.cnstr].


tkoeppe:
Thanks! Even more nice notation would come in from #7152 :-)

PR #7919, opened 2025-06-16
[check] Fix MacOS build by sorting all input files for the comparison

jensmaurer:
This makes us independent from any particular sort order a tool might prefer.


tkoeppe:
Grandiose, thanks!

PR #7923, opened 2025-06-20
[expr.sub,expr.pre.incr] Add spaces around operators

Andreas-Krug:


tkoeppe:
The core language specification deliberately does not prefer any particular style, so I'm not sure we want to change this. @jensmaurer?

tkoeppe:
@jensmaurer Leave the core wording alone?

eisenwave:
We have 131 matches of `\w\+\w` in the draft (although some are in `$math$`, so they don't count), and many more for `*` and `/`.

While I prefer spaces around operators, it seems like a waste of time to peck away at a handful of these now. There are too many to fix and establish a checkable style now.

PR #7924, opened 2025-06-20
[diff.cpp23.strings] Move Clause 27 changes after Clause 23 changes

jwakely:
The 'Strings' clause was moved before the new 'Text processing' clause, so it's now after the 'Containers' clause. The order of the [diff.cpp23] subclauses should reflect that.


tkoeppe:
Nice catch!

tkoeppe (approved):

PR #7979, opened 2025-06-21
P2988R12 std::optional<T&>

### LINKED ISSUE #7939: [2025-06 LWG Motion 2] P2988R12 std::optional<‍T&‍>
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p2988r12.pdf

steve-downey:
Fixes #7939
Fixes cplusplus/papers#1661

Note -- fixed a overfull hbox issue by forcing a break for the constraints at around 5469

is_reference_v<T>  ||
(is_move_constructible_v<T> && is_swappable_v<T>)
is true.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -3196,15 +3196,21 @@ object is tracked by the optional object.
 namespace std {
   // \ref{optional.optional}, class template \tcode{optional}
   template<class T>
-    class optional;                                                     // partially freestanding
+    class optional;                                             // partially freestanding
+
+  // \ref{optional.optional.ref}, partial specialization of \tcode{optional} for lvalue reference types
+  template<class T>
+    class optional<T&>;                                         // partially freestanding
 
   template<class T>
     constexpr bool ranges::enable_view<optional<T>> = true;
   template<class T>
     constexpr auto format_kind<optional<T>> = range_format::disabled;
+  template<class T>
+    constexpr bool ranges::enable_borrowed_range<optional<T&>> = true;
 
   template<class T>
-    concept @\defexposconcept{is-derived-from-optional}@ = requires(const T& t) {       // \expos
+    concept @\defexposconceptnc{is-derived-from-optional}@ = requires(const T& t) {   // \expos
       []<class U>(const optional<U>&){ }(t);
     };
 
@@ -3379,9 +3385,14 @@ When an \tcode{optional<T>} object contains a value,
 member \tcode{val} points to the contained value.
 
 \pnum
-\tcode{T} shall be a type
-other than \cv{} \tcode{in_place_t} or \cv{} \tcode{nullopt_t}
-that meets the \oldconcept{Destructible} requirements (\tref{cpp17.destructible}).
+A type \tcode{X} is a
+\defnx{valid contained type}{valid contained type!\idxcode{optional}} for \tcode{optional}
+if \tcode{X} is an lvalue reference type or a complete non-array object type,
+and \tcode{remove_cvref_t<X>} is a type other than \tcode{in_place_t} or \tcode{nullopt_t}.
+If a specialization of \tcode{optional} is instantiated with a type \tcode{T}
+that is not a valid contained type for \tcode{optional}, the program is ill-formed.
+If \tcode{T} is an object type,
+\tcode{T} shall meet the \oldconcept{Destructible} requirements (\tref{cpp17.destructible}).
 
 \rSec3[optional.ctor]{Constructors}
 
@@ -3454,7 +3465,7 @@ constexpr optional(optional&& rhs) noexcept(@\seebelow@);
 \pnum
 \effects
 If \tcode{rhs} contains a value, direct-non-list-initializes the contained value
-with \tcode{std::move(*rhs)}.
+with \tcode{*std::move(rhs)}.
 \tcode{rhs.has_value()} is unchanged.
 
 \pnum
@@ -3618,7 +3629,7 @@ template<class U> constexpr explicit(@\seebelow@) optional(optional<U>&& rhs);
 \pnum
 \effects
 If \tcode{rhs} contains a value,
-direct-non-list-initializes the contained value with \tcode{std::move(*rhs)}.
+direct-non-list-initializes the contained value with \tcode{*std::move(rhs)}.
 \tcode{rhs.has_value()} is unchanged.
 
 \pnum
@@ -3744,8 +3755,8 @@ The result of the expression \tcode{rhs.has_value()} remains unchanged.
 {\tcode{*this} does not contain a value}
 
 \rowhdr{\tcode{rhs} contains a value} &
-assigns \tcode{std::move(*rhs)} to the contained value &
-direct-non-list-initializes the contained value with \tcode{std::move(*rhs)} \\
+assigns \tcode{*std::move(rhs)} to the contained value &
+direct-non-list-initializes the contained value with \tcode{*std::move(rhs)} \\
 \rowsep
 
 \rowhdr{\tcode{rhs} does not contain a value} &
@@ -3894,8 +3905,8 @@ The result of the expression \tcode{rhs.has_value()} remains unchanged.
 {\tcode{*this} does not contain a value}
 
 \rowhdr{\tcode{rhs} contains a value} &
-assigns \tcode{std::move(*rhs)} to the contained value &
-direct-non-list-initializes the contained value with \tcode{std::move(*rhs)} \\
+assigns \tcode{*std::move(rhs)} to the contained value &
+direct-non-list-initializes the contained value with \tcode{*std::move(rhs)} \\
 \rowsep
 
 \rowhdr{\tcode{rhs} does not contain a value} &
@@ -4319,8 +4330,7 @@ Let \tcode{U} be \tcode{remove_cv_t<invoke_result_t<F, decltype(*\exposid{val})>
 
 \pnum
 \mandates
-\tcode{U} is a non-array object type
-other than \tcode{in_place_t} or \tcode{nullopt_t}.
+\tcode{U} is a valid contained type for \tcode{optional}.
 The declaration
 \begin{codeblock}
 U u(invoke(std::forward<F>(f), *@\exposid{val}@));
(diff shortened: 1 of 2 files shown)


source/utilities.tex
```diff
@@ -3198,14 +3198,20 @@
...
   template<class T>
-    constexpr auto format_kind<optional<T>> = range_format::disabled;
+  constexpr auto format_kind<optional<T>> = range_format::disabled;
+  template<class T>
+  constexpr bool ranges::enable_borrowed_range<optional<T\&>> = true;
 
   template<class T>
-    concept @\defexposconcept{is-derived-from-optional}@ = requires(const T& t) {       // \expos
-      []<class U>(const optional<U>&){ }(t);
+    concept @\defexposconcept{is-derived-from-optional}@ = requires(const T\& t) {       // \expos
```
jensmaurer: We don't need / want quoting of ampersands inside codeblocks.   Also, these LaTeX changes were likely not asked for in the incoming paper.  Please avoid those drive-by "fixes" as much as possible when applying papers, and put them into separate pull requests.  This eases further review.

steve-downey: Fixed along with 3 other stray escapes.

source/utilities.tex
```diff
@@ -3198,10 +3198,16 @@
...
-    constexpr bool ranges::enable_view<optional<T>> = true;
+  class optional<T&>;                                                   // partially freestanding
+
+  template<class T>
+  constexpr bool ranges::enable_view<optional<T>> = true;
   template<class T>
-    constexpr auto format_kind<optional<T>> = range_format::disabled;
+  constexpr auto format_kind<optional<T>> = range_format::disabled;
+  template<class T>
+  constexpr bool ranges::enable_borrowed_range<optional<T&>> = true;
```
jensmaurer: For a non-defining declaration after a _template-head_, please indent by two spaces.
This applies predominantly to header files.  See the existing lines 3199 and 3202 for examples.

jensmaurer (changes requested):

source/utilities.tex
tkoeppe: Done.

tkoeppe:
Thanks, @steve-downey, I can take it from here!

source/utilities.tex
```diff
@@ -3379,9 +3385,10 @@
 member \tcode{val} points to the contained value.
 
 \pnum
-\tcode{T} shall be a type
-other than \cv{} \tcode{in_place_t} or \cv{} \tcode{nullopt_t}
-that meets the \oldconcept{Destructible} requirements (\tref{cpp17.destructible}).
+A type \tcode{X} is a \defnx{valid contained type}{valid contained type!\idxcode{optional}} for \tcode{optional} if \tcode{X} is an lvalue reference type
```
tkoeppe: Let's see if ISO lets us get away with this definition. If not, this is a local-enough term that we can just not mark it up, like we do elsewhere (e.g. shared_ptr's "empty").

steve-downey:
> Thanks, @steve-downey, I can take it from here!

I will keep an eye so as to improve the next one.

tkoeppe (approved):

source/utilities.tex
```diff
@@ -4442,6 +4451,551 @@
 \tcode{*this} does not contain a value.
 \end{itemdescr}
 
+\rSec2[optional.optional.ref]{Partial specialization of \tcode{optional} for reference types}
+
+\rSec3[optional.optional.ref.general]{General}
```
jwakely, line 4456: Should we call this [optional.ref.general] instead? It's a sibling of [optional.ref.ctor], [optional.ref.assign], etc.

source/utilities.tex
```diff
@@ -4442,6 +4451,551 @@
...
+      template<class U>
+        constexpr explicit(@\seebelow@) optional(U&& u) noexcept(@\seebelow@);
+      template<class U>
+        constexpr explicit(@\seebelow@) optional(optional<U>& rhs) noexcept(@\seebelow@);
+      template<class U>
+        constexpr explicit(@\seebelow@) optional(const optional<U>& rhs) noexcept(@\seebelow@);
+      template<class U>
+        constexpr explicit(@\seebelow@) optional(optional<U>&& rhs) noexcept(@\seebelow@);
+      template<class U>
+        constexpr explicit(@\seebelow@) optional(const optional<U>&& rhs) noexcept(@\seebelow@);
```
jwakely, line 4482: For optional<T> the ctors are in this order:

```
constexpr optional() noexcept;
constexpr optional(nullopt_t) noexcept;
constexpr optional(const optional&);
constexpr optional(optional&&) noexcept(see below );
template<class... Args>
  constexpr explicit optional(in_place_t, Args&&...);
template<class U, class... Args>
  constexpr explicit optional(in_place_t, initializer_list<U>, Args&&...);
template<class U = remove_cv_t<T>>
  constexpr explicit(see below ) optional(U&&);
template<class U>
  constexpr explicit(see below ) optional(const optional<U>&);
template<class U>
  constexpr explicit(see below ) optional(optional<U>&&);
```

Is there any reason to deviate from that here? I don't recall discussing this in LWG.

jwakely: Obviously some of those ctors aren't present for `optional<T&>` but for the ones that are, I think it makes sense to be consistent.

jwakely: Oh, I suppose they already are consistent, just with additional ones inserted. OK, ignore me, sorry for the noise.

source/utilities.tex
```diff
@@ -4442,6 +4451,551 @@
...
+
+\pnum
+\effects
+  Equivalent to:
+\begin{codeblock}
+if (has_value()) {
+  return invoke(std::forward<F>(f), *@\exposid{val}@);
+} else {
+  return remove_cvref_t<U>();
+}
```
jwakely, line 4912: IIRC LWG didn't want these braces here, but it's consistent with the equivalent functions for `optional<T>`. We can remove them from both places later if necessary.

source/utilities.tex
tkoeppe: Surely this is editorial? Thank you for those pointers, though :-)

tkoeppe: Yes, good call. What about [optional.optional.ref], though?

source/utilities.tex
tkoeppe: Hm, surely this is editorial? Why is LWG opposed to braces? Have braces ever decreased clarity of presentation? goto fail; goto fail.

source/utilities.tex
jensmaurer: Same.

[optional.optional.ref*] is just stupid; we spent plenty of effort to get rid of [stmt.stmt].

jwakely: We could change that too, but my main concern was the excessive length of optional.optional.ref.general

source/utilities.tex
jwakely: Yes definitely editorial. It's just shorter without the braces (only by one line, admittedly).

source/utilities.tex
tkoeppe: Hm, good point that we shouldn't immediately create a new set of problematic names that will invite ridicule and disbelief.

Can we come up with an acceptable complete set of new labels for this paper?

steve-downey: The challenge is the parallelism with the existing mess that has [optional.optional.general] and then the ctors in [optional.ctor]. I had optionalref instead of optional.ref at one point, but that is not better.

```diff
 \rSec1[optional]{Optional objects}
 \rSec2[optional.general]{General}
 \rSec2[optional.syn]{Header \tcode{<optional>} synopsis}
 \rSec2[optional.optional]{Class template \tcode{optional}}
 \rSec3[optional.optional.general]{General}
 \rSec3[optional.ctor]{Constructors}
 \rSec3[optional.dtor]{Destructor}
 \rSec3[optional.assign]{Assignment}
 \rSec3[optional.swap]{Swap}
 \rSec3[optional.iterators]{Iterator support}
 \rSec3[optional.observe]{Observers}
 \rSec3[optional.monadic]{Monadic operations}
 \rSec3[optional.mod]{Modifiers}
+\rSec2[optional.optional.ref]{Partial specialization of \tcode{optional} for reference types}
+\rSec3[optional.optional.ref.general]{General}
+\rSec3[optional.ref.ctor]{Constructors}
+\rSec3[optional.ref.assign]{Assignment}
+\rSec3[optional.ref.swap]{Swap}
+\rSec3[optional.ref.iterators]{Iterator support}
+\rSec3[optional.ref.observe]{Observers}
+\rSec3[optional.ref.monadic]{Monadic operations}
+\rSec3[optional.ref.mod]{Modifiers}
+\rSec3[optional.ref.expos]{Exposition only helper functions}
 \rSec2[optional.nullopt]{No-value state indicator}
 \rSec2[optional.bad.access]{Class \tcode{bad_optional_access}}
 \rSec2[optional.relops]{Relational operators}
 ```
 Possibly changing the [optional.optional.ref] ones to [optional.ref] would be least worst?

jwakely: > Possibly changing the [optional.optional.ref] ones to [optional.ref] would be least worst?

Good enough for me

tkoeppe: For both cases? So we'd get [optional.ref] and [optional.ref.general]?

jwakely: I don't really mind whether [optional.optional.ref] changes or not, but changing it seems more consistent with `optional<T>` which has:

[optional.optional] Class template `optional`
[optional.optional.general] General _(i.e. the class synopsis for `optional<T>`)_

And changing [optional.optional.ref] would give us:

[optional.ref] Partial specialization of `optional` for reference types
[optional.ref.general] General _(i.e. the class synopsis for `optional<T&>`)_

And changing it would also mean that all subclauses for `optional<T&>` start with [optional.ref].

So yeah, I think I've talked myself into changing it :)

tkoeppe: I was thinking the same. Yes, the existing [optional.optional{,.general}] is a bit of an odd-man-out, but the added structure works very well with "optional.ref" everywhere.

I'll add that has an editorial commit.

PR #7981, opened 2025-06-22
P2843R3 Preprocessing is never undefined

### LINKED ISSUE #7937: [2025-06 CWG Motion 11]  P2843R3 Preprocessing is never undefined
https://wiki.edg.com/pub/Wg21sofia2025/StrawPolls/p2843r3.pdf

AlisdairM:
Fixes #7937
Fixes cplusplus/papers/#1548


source/preprocessor.tex
```diff
@@ -1500,6 +1500,16 @@
...
 
+\pnum
+\indextext{unit!translation}%
+A translation unit shall not \tcode{\#define} or \tcode{\#undef}
+names lexically identical
+to keywords,
+to the identifiers listed in \tref{lex.name.special}, or
+to the \grammarterm{attribute-token}{s} described in~\ref{dcl.attr},
+except that the names \tcode{likely} and \tcode{unlikely} may be
+defined as function-like macros.
```
eisenwave, line 1540:
```suggestion
defined as function-like macros\iref{cpp.replace}.
```

source/preprocessor.tex
```diff
@@ -1783,19 +1793,15 @@
...
-\tcode{\textbackslash}
-character is inserted before each
-\tcode{"}
-and
-\tcode{\textbackslash}
-character of a \grammarterm{character-literal} or \grammarterm{string-literal}
-(including the delimiting
-\tcode{"}
-characters).
+\grammarterm{character-literal}s and \grammarterm{string-literal}s
```
eisenwave:
```suggestion
\grammarterm{character-literal}{s} and \grammarterm{string-literal}{s}
```

source/preprocessor.tex
```diff
@@ -1783,19 +1793,15 @@
...
+(including the delimiting \unicode{0022}{quotation mark} (\tcode{"}))
+contained within the preprocessing token:
+a \unicode{005c}{reverse solidus} character (\tcode{\textbackslash})
+is inserted before each \unicode{0022}{quotation mark} and
+\unicode{005c}{reverse solidus} character and each new-line character is
```
eisenwave: The PR is correct, but we should really add a comma here in post.
```suggestion
\unicode{005c}{reverse solidus} character,
and each new-line character is
```

source/preprocessor.tex
```diff
@@ -1783,19 +1793,15 @@
...
+replaced by the two-character sequence \tcode{\textbackslash n}.
 If the replacement that results is not a valid character string literal,
-the behavior is undefined. The character string literal corresponding to
+the program is ill-formed. The character string literal corresponding to
```
eisenwave: Would be nice to put the next sentence on a separate line while we're at it (semantic line breaks).

Eisenwave (changes requested):

source/preprocessor.tex
AlisdairM: I believe we established that this was an old convention that is no longer needed, and would be using the plain "s" plural going forward.

@jensmaurer, @tkoeppe do you remember the thread I am talking about?

eisenwave: I found https://github.com/cplusplus/draft/pull/6235#issuecomment-1546299977 on this, and yeah, it seems like the workaround is no longer necessary.

I thought you're applying it inconsistently because you have `{s}` added somewhere else, but that's just moved, old wording.

tkoeppe: Yes, fine without braces, and it's also fine to not touch old text unnecessarily. This doesn't affect the presentation.

source/preprocessor.tex
AlisdairM: I removed that cross-reference as this paragraph has moved *into* [cpp.replace]

eisenwave: Then there should be a fixup commit that removes that pointless reference, distinct from the first commit (which is meant to translate the paper 1:1, even if it has typos and other editorial mistakes).

See also https://github.com/cplusplus/draft/wiki/Applying-papers-to-the-Working-Draft

source/preprocessor.tex
AlisdairM: Re-ordered the two sentences, and then broke the large paragraph into several smaller ones to improve readability.

source/preprocessor.tex
```diff
@@ -1783,20 +1795,22 @@
...
+contained within the preprocessing token:
+a \unicode{005c}{reverse solidus} character (\tcode{\textbackslash})
+is inserted before each \unicode{0022}{quotation mark} and
+\unicode{005c}{reverse solidus} character,
+and each new-line character is
+replaced by the two-character sequence \tcode{\textbackslash n}.
+The character string literal corresponding to an empty stringizing argument
+is \tcode{""}.
+
+\pnum
```
eisenwave: I'm not convinced this paragraph spit here works.

> If the replacement that results

is clearly connected to the mechanisms described in the previous paragraph, so it doesn't make sense on its own. The subsequent paragraph split works, but we're not gaining a whole lot by doing that.

Just to be clear, my previous comment was just suggesting to add a semantic line break. Further readability enhancements and paragraph splitting should typically go into a separate PR, not into the motions PR.

AlisdairM:
@Eisenwave Rewrote history to correctly handle removing the circular reference as a subsequent edit.  Undid the paragraph breakdown, replacing with the single semantic line break that you suggested.

eisenwave:
That looks about right now, but the commit for adding a semantic line break isn't really necessary. That could just be a drive-by fix in the main commit since it doesn't alter the rendered output in any way.

I'm fine with all the other stuff, but I suspect @tkoeppe may want that last commit yoinked out.

Eisenwave (approved):

PR #7986, opened 2025-06-22
P1306R5 Expansion Statements

### LINKED ISSUE #7934: [2025-06 CWG Motion 8] P1306R5 Expansion Statements
https://wiki.edg.com/pub/Wg21sofia2025/StrawPolls/p1306r5.html

tkoeppe:
Fixes #7934.


source/basic.tex
```diff
@@ -4711,14 +4714,25 @@
 
 \pnum
 The fourth context is when a temporary object
-is created in the \grammarterm{for-range-initializer} of a range-based \keyword{for} statement.
+is created in the \grammarterm{for-range-initializer} of
+either a range-based \keyword{for} statement
```
eisenwave, line 4890: The paper does not put "for" in code font. Should this be a separate fixup commit?

source/statements.tex
```diff
@@ -772,13 +783,202 @@
...
+  @$S_{0}$@
+  @\vdots@
+  @$S_{N-1}$@
+}
+\end{codeblock}
+where $N$ is the result of evaluating the expression
+\begin{codeblock}
+[] consteval {
+  std::ptrdiff_t result = 0;
+  for (auto i = @\exposid{begin}@; i != @\exposid{end}@; ++i, ++result);
```
eisenwave, line 887: The paper puts a space between `)` and `;`, and I find that this helps in readability; it makes it more apparent that the loop is (intentionally) empty. Another way to go about it would be to append `{ }`.

Eisenwave (commented):
Couldn't find anything improperly translated, other than minute font and whitespace differences.

source/basic.tex
tkoeppe: Then the paper is misquoting the existing wording. It happens :-)

source/statements.tex
tkoeppe: Yes, I saw, and I think that's editorial, and this seemed fine. We can propose a change to ` {}` or even ` { continue; }` separately.

(You're up against a crowd who already wants to avoid braces on complex if statements just to save space, @jwakely ^^^, and there's a lot we could revisit in principle to improve "low-friction" readability. We should be past the cost-affecting page limit now anyway.)

source/statements.tex
```diff
@@ -832,10 +1032,11 @@
 A \keyword{break} statement shall be enclosed by\iref{stmt.pre}
 \indextext{\idxgram{iteration-statement}}%
 \indextext{statement!\idxcode{switch}}%
-an \grammarterm{iteration-statement}\iref{stmt.iter} or
-a \keyword{switch} statement\iref{stmt.switch}.
+an \grammarterm{iteration-statement}\iref{stmt.iter},
+an \grammarterm{expansion-statement}\iref{stmt.expand},
+or \keyword{switch} statement\iref{stmt.switch}.
```
jensmaurer:
```suggestion
or a \keyword{switch} statement\iref{stmt.switch}.
```

and I would put the "or" at the end of the preceding item

source/statements.tex
```diff
@@ -846,45 +1047,14 @@
...
-    // ...
-  }
-@\exposid{contin}@: ;
-}
-\end{codeblock}
-\end{minipage}
-
-a \keyword{continue} not contained in an enclosed iteration statement is
-equivalent to \tcode{goto} \exposid{contin}.
+\grammarterm{iteration-statement} or an \grammarterm{expansion-statement}.
```
jensmaurer: semantic linebreaks: move "an" from previous line

jensmaurer (changes requested):

source/basic.tex
jakubjelinek: The rest of this bullet only talks about for-range-initializer which enumerating expansion statement doesn't have.  There doesn't even have to be any reference, in the enumerating case Si has just for-range-declaration = Ei; where for-range-declaration very well could be just auto name.  So is the intent that this is done only if for-range-declaration is say auto & or auto && or int & and the lifetime is then extended until end of the Si compound statement?

jakubjelinek: I've filed https://github.com/cplusplus/CWG/issues/724 for this.

source/statements.tex
tkoeppe: Nice catch, thanks!

source/statements.tex
tkoeppe: Done.

Issue #7943, opened 2025-06-21
[2025-06 LWG Motion 6] P3179R9 Parallel Range Algorithms


jensmaurer:
https://wiki.edg.com/pub/Wg21sofia2025/StrawPolls/P3179R9.html

eisenwave:
I'll do it

PR #7987, opened 2025-06-22
P3179R9 C++ parallel range algorithms

Eisenwave:
Closes https://github.com/cplusplus/draft/issues/7943.
Also closes https://github.com/cplusplus/papers/issues/1815.

Took me a few minutes to draft this up. Not quite done yet.

I'm not sure who's up for reviewing this, but there are quite a lot of FIXMEs in here (some non-editorial) which will have to be addressed. I could need some feedback on those.


tkoeppe:
Thanks a lot!

eisenwave:
Have fun reviewing this 3000-line diff, lol.

Some further notes for reviewers:
- I still don't know what the `exposidnc` and other `nc` macros do, so I've not used any of them anywhere. It's an easy fix to use those if need be.
- There are a bunch of FIXMEs, some non-editorial. Those should be resolved before merging. However, my impression is that it would be much easier to first review the changes of the paper EXACTLY as is, prior to starting any work on fix-ups.
- I have done this change by first making the `<algorithm>` synopsis, then copying and pasting into the descriptions below and adding `ranges::`. If there is something wrong in the synopsis, it will be in the descriptions too, and if you're proof-reading the declarations in the descriptions to see if they match the synopsis, you're wasting your time.
- I have created a tool that automatically transforms an existing `ranges::` declaration into one that also takes an `execution-policy`. If I did something wrong with this tool, the transformation is almost certainly wrong everywhere.

eisenwave:
I have received confirmation from one of the authors that due to an editorial oversight
- `partition`,
- `remove`, and
- `remove_if`

are missing `requires permutable`.

However, this defect is non-editorial and will have to be addressed in an LWG issue.

source/algorithms.tex
```diff
@@ -7708,6 +9732,17 @@
   requires @\libconcept{permutable}@<iterator_t<R>>
   constexpr borrowed_subrange_t<R>
     ranges::partition(R&& r, Pred pred, Proj proj = {});
+
+template<@\exposconcept{execution-policy}@ Ep, @\libconcept{random_access_iterator}@ I, @\libconcept{sized_sentinel_for}@<I> S,
+         class Proj = identity, @\libconcept{indirect_unary_predicate}@<projected<I, Proj>> Pred>
+  subrange<I>
+    // FIXME: LWG issue, I'm pretty sure there should be requires permutable<I> here.
```
tkoeppe: @jwakely: Note the potential issue.

source/algorithms.tex
```diff
@@ -6286,31 +7997,49 @@
...
 \end{itemize}
+\begin{note}
+For the parallel algorithm overloads in namespace \tcode{std},
+there can be a performance cost
+if the value type of \tcode{ForwardIterator1} does not meet both the
+\oldconcept{CopyConstructible} and \oldconcept{CopyAssignable} requirements.
+For the parallel algorithm overloads in namespace \tcode{ranges},
+there can be a performance cost if \tcode{iter_value_t<I>}
+does not model both \libconcept{copy_constructible}
+%FIXME: There exists no such concept, but the note refers to it. LWG input needed.
```
tkoeppe: @jwakely: Note the potential issue.

source/algorithms.tex
eisenwave: An LWG issue for the missing `permutable<I>` has already been filed.

tkoeppe: Thanks!

source/algorithms.tex
eisenwave: To be clear here: `copy_constructible` exists, but `copy_assignable` is a hallucination.

eisenwave: https://lists.isocpp.org/lib/2025/07/32289.php I've asked here now.

tkoeppe: Thanks!

PR #7992, opened 2025-06-23
P2830R10 Standardized Constexpr Type Ordering

### LINKED ISSUE #7954: [2025-06 LWG Motion 17] P2830R10 Standardized Constexpr Type Ordering
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p2830r10.html

AlisdairM:
Fixes #7954
Fixes cplusplus/papers#1507

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -861,6 +861,7 @@ the values of these macros with greater values.
   // also in \libheader{utility}, \libheader{tuple}, \libheader{map}, \libheader{unordered_map}
 #define @\defnlibxname{cpp_lib_tuples_by_type}@                    201304L // freestanding, also in \libheader{utility}, \libheader{tuple}
 #define @\defnlibxname{cpp_lib_type_identity}@                     201806L // freestanding, also in \libheader{type_traits}
+#define @\defnlibxname{cpp_lib_type_order}@                        202506L // also in \libheader{compare}
 #define @\defnlibxname{cpp_lib_type_trait_variable_templates}@     201510L // freestanding, also in \libheader{type_traits}
 #define @\defnlibxname{cpp_lib_uncaught_exceptions}@               201411L // freestanding, also in \libheader{exception}
 #define @\defnlibxname{cpp_lib_unordered_map_try_emplace}@         201411L // also in \libheader{unordered_map}
@@ -4791,6 +4792,7 @@ the three-way comparison operator\iref{expr.spaceship}.
 \indexlibraryglobal{is_geq}%
 \indexlibraryglobal{is_gteq}%
 \indexlibraryglobal{common_comparison_category_t}%
+\indexlibraryglobal{type_order_v}%
 \begin{codeblock}
 // all freestanding
 namespace std {
@@ -4839,6 +4841,13 @@ namespace std {
     inline constexpr @\unspec@ compare_weak_order_fallback = @\unspec@;
     inline constexpr @\unspec@ compare_partial_order_fallback = @\unspec@;
   }
+
+  // \ref{compare.type}, type ordering
+  template<class T, class U>
+    struct type_order;
+
+  template<class T, class U>
+    constexpr strong_ordering type_order_v = type_order<T, U>::value;
 }
 \end{codeblock}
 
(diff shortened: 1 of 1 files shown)


source/support.tex
```diff
@@ -4813,6 +4815,13 @@
     inline constexpr @\unspec@ compare_weak_order_fallback = @\unspec@;
     inline constexpr @\unspec@ compare_partial_order_fallback = @\unspec@;
   }
+
+  // \ref{compare.type}, type ordering
+  template<class T, class U>
+  struct type_order;
```
tkoeppe: indent "struct" by two

source/support.tex
```diff
@@ -5668,6 +5677,42 @@
 of a template instantiation.
 \end{note}
 
+\rSec2[compare.type]{Type Ordering}
+\pnum
```
tkoeppe, line 5708: blank line after \rSec2

source/support.tex
```diff
@@ -5668,6 +5677,42 @@
...
+There is an \impldef{total ordering of all types} total ordering of all types.
+For any (possibly incomplete) types \tcode{X} and \tcode{Y},
+the expression \tcode{\exposid{TYPE-ORDER}(X, Y)} is a constant expression \iref{expr.const}
```
tkoeppe: stray space before `\iref`

source/support.tex
```diff
@@ -5668,6 +5677,42 @@
...
+\end{note}
+\begin{note}
+The ordering of TU-local types from different translation units is not observable,
+because the necessary specialization of \tcode{type_order} is impossible to name.
+\end{note}
+
+\indexlibraryglobal{type_order}%
+\begin{itemdecl}
+template<class T, class U>
+struct type_order;
```
tkoeppe: indent by two or remove linebreak

source/support.tex
```diff
@@ -5668,6 +5677,42 @@
...
+\end{itemdecl}
+
+\pnum
```
tkoeppe, line 5735: There's no \itemdescr here?

source/support.tex
AlisdairM: That would be inconsistent with `struct common_comparison_category` above.  Should I indent both?

source/support.tex
AlisdairM: Done.

source/support.tex
tkoeppe: Don't worry about existing material, let's just get the new wording right. We can clean up existing text separately if there's a pressing need.

source/support.tex
tkoeppe: Thanks!

source/support.tex
```diff
@@ -5668,6 +5677,45 @@
...
+\begin{itemdecl}
+template<class T, class U>
+struct type_order;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+The name \tcode{type_order} denotes a \oldconcept{BinaryTypeTrait}\iref{meta.rqmts}
+with a base characteristic of
+\tcode{integral_constant<strong_ordering, \exposid{TYPE-ORDER}(X, Y)>}.
```
jwakely:
```suggestion
\tcode{integral_constant<strong_ordering, \exposid{TYPE-ORDER}(T, U)>}.
```

(the incoming paper is wrong)

jwakely: We also need an LWG issue for this, because `integral_constant<strong_ordering, O>` is only valid if `strong_ordering` is a structural type, which is not true for libc++ or libstdc++ (and should not be true according to the exposition-only private member shown in the standard). So this code can't even be compiled.

Once this PR is merged, I'll open an issue.

Bronek: Does this need to be picked by LEWG ? (hope not)

jwakely: The design in the paper is unimplementable, so maybe.

source/support.tex
AlisdairM: Applied

source/support.tex
AlisdairM: Applied

source/support.tex
tkoeppe: Thanks!

source/support.tex
```diff
@@ -5668,6 +5677,45 @@
...
+\begin{itemdecl}
+template<class T, class U>
+  struct type_order;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+The name \tcode{type_order} denotes a \oldconcept{BinaryTypeTrait}\iref{meta.rqmts}
+with a base characteristic of
+\tcode{integral_constant<strong_ordering, \exposid{TYPE-ORDER}(T, U)>}.
```
eisenwave: @AlisdairM can you fix the overfull hbox here?

If you sprinkle in `\brk{}` or a `\-` in the right place, surely that takes care of it.

Eisenwave (changes requested):

PR #7994, opened 2025-06-23
Tools: Simple Bikeshed to LaTeX search and replace

mattkretz:
* bs_to_tex.sh does a first pass on the wording sources of a Bikeshed paper.
* add_libconcept.sh can be used as a second pass to index concept names.

### DIFF
--- a/tools/add_libconcept.sh
+++ b/tools/add_libconcept.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 usage() {
   cat <<EOF
--- a/tools/add_libconcept.sh
+++ b/tools/add_libconcept.sh
@@ -1,0 +1,41 @@
+#!/bin/sh
+
+usage() {
+  cat <<EOF
+Usage: $0 <LaTeX source>
+
+Greedy application of \\libconcept and \\exposconcept macros to anything that
+looks like a concept.
+EOF
+}
+
+case "$1" in
+  -h|--help|"")
+    usage
+    exit 1
+    ;;
+esac
+
+source_dir="${0%/*}/../source"
+
+concepts=$(grep -oh 'deflibconcept{[a-z_:]*}' "$source_dir"/*.tex \
+  | sed -e 's/^deflibconcept{\(.*\)}$/\1/' \
+  | sed -e ':a; N; $!ba; s/\n/\\|/g'
+)
+
+exposconcepts=$(grep -oh 'defexposconceptn\?c\?{[a-z-]*}' "$source_dir"/*.tex \
+  | sed -e 's/^defexposconceptn\?c\?{\(.*\)}$/\1/' \
+  | sed -e ':a; N; $!ba; s/\n/\\|/g'
+)
+
+sed -e 's,\([^\\{.-]\)\<\('"$concepts"'\)\>\([^-]\),\1@\\libconcept{\2}@\3,g' \
+  -e 's,\([^\\{.-]\)\<\('"$exposconcepts"'\)\>\([^-]\),\1@\\exposconcept{\2}@\3,g' \
+  -e 's,\\exposid\(nc\)\?{\('"$concepts"'\)},\\libconcept{\2},g' \
+  -e 's,\\exposid\(nc\)\?{\('"$exposconcepts"'\)},\\exposconcept\1{\2},g' \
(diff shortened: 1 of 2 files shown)


eisenwave:
This seems like a useful utility, but I don't see why it needs to live in this repository.

As you may have noticed, it can take a little while until PRs here are reviewed, you're making things a bit inconvenient for yourself if you want to keep this script here, and update it here.

Maybe we could mention such useful tools in the README, and you could host this somewhere where you can make changes easily yourself?

tkoeppe:
I'm fine having this tool here. I just haven't had a chance to had a closer look.

tools/bs_to_tex.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/zsh
```
tkoeppe: Do we need to have zsh for this? Is this available on MacOS?

tools/add_libconcept.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
+
+usage() {
+  cat <<EOF
+Usage: $0 <LaTeX source>
```
tkoeppe, line 5: could we use curly braces for all variable references, e.g. `${0}`?

tools/bs_to_tex.sh
mattkretz: zsh is probably the most common shell after bash. I have no experience with MacOS, but I read "most versions of macOS ship zsh by default".
I'm a zsh user, which is why I know its extensions (over plain sh). I'm sure it can be changed to use bash extensions instead or possibly even plain sh. Installing zsh is a simple command (if at all). Porting to bash might take a lot more time (though, I have no idea, because I never know what syntax is common between bash and zsh).

tools/add_libconcept.sh
mattkretz: `${0}` specifically reads very strange to me and unconditional use of braces is inconsistent with the rest of our shell scripts. I very much prefer to only use braces when they serve a purpose.

tools/bs_to_tex.sh
jwakely: I think everything used here behaves the same in bash, but I would need to double check the `=~` behaviour (quoting the `=~` operands always seems to work in ways that surprise me for bash).

tkoeppe: OK -- that's fine, but could we maybe have a comment at the top that calls out the need for zsh and its specific behaviour/difference from Posix/Bash, so that it's easy to remember this? I can otherwise imagine future frustration when someone trying to understand/debug this misses that detail.

And to check, the other script is fine with just "sh"?

tools/add_libconcept.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
```
jensmaurer: This script needs bash; plain "sh" is not enough, because it uses `x=$(blah)` syntax.

tools/bs_to_tex.sh
jensmaurer: Please switch to bash; this is a trivially small script.

The one thing that strikes me as odd (not being a bash expert) in the following is "((itemize))".  Please switch this to more basic [ or [[ syntax.

jensmaurer (changes requested):

tools/bs_to_tex.sh
tkoeppe: `((` is Bash, too, e.g. `for ((a = 0; a != 10; ++a)); do echo ${a}; done`. That part is fine I think.

jensmaurer: Indeed; I've never come across that.

tools/add_libconcept.sh
mattkretz: Oh. I was using `dash`, hoping it is close enough to plain `sh`. But it seems only backticks would be supported. Fine. `bash` is everywhere we care about.

tools/bs_to_tex.sh
mattkretz: Sorry for that wrong comment I posted before.

So after changing to `bash` the script runs without errors but with a different output. Now I'll need to start debugging where it goes wrong. That's exactly why I don't write `bash` scripts.

edit: @jwakely got it right. I'll test with the other paper I had and if that also works then I'll update this PR accordingly.

PR #7997, opened 2025-06-23
P2876R3 Proposal to extend std::simd with more constructors and accessors

### LINKED ISSUE #7947: [2025-06 LWG Motion 10] P2876R3 Proposal to extend std::simd with more constructors and accessors
https://wiki.edg.com/pub/Wg21sofia2025/StrawPolls/P2876R3.html

mattkretz:
Fixes #7947 
Fixes cplusplus/papers#1531

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -19354,7 +19354,7 @@ Let $N$ be the width of \tcode{unsigned long long}.
 \pnum
 \expects
 \begin{itemize}
-\item \tcode{size() <= $N$} is \tcode{true}, or
+\item \tcode{size()} $\le N$, or
 \item for all $i$ in the range \range{$N$}{size()}, \tcode{operator[]($i$)} returns \tcode{false}.
 \end{itemize}
 
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -19095,6 +19095,8 @@ namespace std::datapar {
     template<size_t UBytes, class UAbi>
       constexpr explicit basic_simd_mask(const basic_simd_mask<UBytes, UAbi>&) noexcept;
     template<class G> constexpr explicit basic_simd_mask(G&& gen) noexcept;
+    constexpr basic_simd_mask(const bitset<size()>& b) noexcept;
+    constexpr explicit basic_simd_mask(@\libconcept{unsigned_integral}@ auto val) noexcept;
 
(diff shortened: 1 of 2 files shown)


source/numerics.tex
```diff
@@ -19241,6 +19247,31 @@
...
+\begin{itemdecl}
+constexpr explicit basic_simd_mask(@\libconcept{unsigned_integral}@ auto val) noexcept;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\effects
+Initializes the first $M$ elements to the corresponding bit values in
+\tcode{val}, where $M$ is the smaller of \tcode{size()} and the number of bits in
+the value representation \iref{basic.types.general} of the type of \tcode{val}. If
```
eisenwave: No space before `\iref`.

Also the semantic line breaks (https://sembr.org/) in this area could be improved; new sentences should always be on a new line at least.

source/numerics.tex
```diff
@@ -19299,6 +19330,43 @@
...
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+Let $N$ be the width of \tcode{unsigned long long}.
+
+\pnum
+\expects
+\begin{itemize}
+\item \tcode{size()} $\le N$, or
```
eisenwave, line 19357: I'm not sure why you've decided to change what is in the paper via fix-up here. We have tons of wording precedent for

> `x < y` is `true`

... as dumb as it sounds, but to my knowledge, we don't have any kind of formal decision that would eliminate this. LWG approved this wording, so we may as well keep it.

mattkretz: I changed it because that's more consistent with what's in [simd] (i.e., what we did for P1928).  I wanted to call that out as a questionable commit / a suggestion and forgot about it.
Better to drop it then?

eisenwave: I cannot answer with authority, but we have procedures in https://github.com/cplusplus/draft/wiki/Applying-papers-to-the-Working-Draft for this. To my understanding, we only use commits in the same PR for typo repairs and other things that are broken in the paper wording.

If there are stylistic/consistency improvements, those tend to go onto into a separate PR.

mattkretz: I've removed the commit from my Motions 10–13 PR and will submit a separate PR.

mattkretz:
Moved to new PR

PR #7998, opened 2025-06-23
P3480R6 std::simd is a range

### LINKED ISSUE #7948: [2025-06 LWG Motion 11] P3480R6 std::simd is a range
https://wiki.edg.com/pub/Wg21sofia2025/StrawPolls/P3480R6.pdf

mattkretz:
Fixes #7948
Fixes cplusplus/papers#2134


source/numerics.tex
```diff
@@ -17198,6 +17454,14 @@
...
     using mask_type = basic_simd_mask<sizeof(T), Abi>;
     using abi_type = Abi;
+    using iterator = \exposid{simd-iterator}<basic_simd>;
+    using const_iterator = \exposid{simd-iterator}<const basic_simd>;
+
+    constexpr iterator begin() noexcept { return {*this, 0}; }
+    constexpr const_iterator begin() const noexcept { return {*this, 0}; }
+    constexpr const_iterator cbegin() const noexcept { return {*this, 0}; }
+    constexpr default_sentinel_t end() const noexcept { return {}; }
+    constexpr default_sentinel_t cend() const noexcept { return {}; }
```
eisenwave, line 17464: Seeing that these are entirely defined within the synopsis, we should have some `\indexlibrarymember` macros above.

Same for `basic_simd_mask`.

jensmaurer: We have \libmember for inline indexing inside codeblocks.  "grep" for examples.

mattkretz: The members of *simd-iterator* do not get indexed, right? (because it's an exposition-only type? OTOH its members are public API.)

mattkretz: Also, looking at some more examples of indexing, there's a lot of index macros missing for [simd]. I guess I should add all those as editorial commit after the motions.

mattkretz: Looking at `ranges.tex`, the exposition-only iterators are fully indexed. So I guess I could do that as well.
I'll close this PR in favor of a single PR that includes all 4 simd motions ASAP.

PR #7999, opened 2025-06-23
P1317R2 Remove return type deduction in std::apply

### LINKED ISSUE #7972: [2025-06 LWG Motion 35] P1317R2 Remove return type deduction in std::apply
https://wiki.edg.com/pub/Wg21sofia2025/StrawPolls/P1317R2.pdf

jwakely:
Fixes #7972

### DIFF
--- a/source/meta.tex
+++ b/source/meta.tex
@@ -295,6 +295,9 @@ namespace std {
   template<class Fn, class... ArgTypes> struct is_nothrow_invocable;
   template<class R, class Fn, class... ArgTypes> struct is_nothrow_invocable_r;
 
+  template<class Fn, class Tuple> struct is_applicable;
+  template<class Fn, class Tuple> struct is_nothrow_applicable;
+
   // \ref{meta.trans.cv}, const-volatile modifications
   template<class T> struct remove_const;
   template<class T> struct remove_volatile;
@@ -367,6 +370,7 @@ namespace std {
   template<class... T> struct common_reference;
   template<class T> struct underlying_type;
   template<class Fn, class... ArgTypes> struct invoke_result;
+  template<class Fn, class Tuple> struct apply_result;
   template<class T> struct unwrap_reference;
   template<class T> struct unwrap_ref_decay;
 
@@ -388,6 +392,8 @@ namespace std {
     using @\libglobal{underlying_type_t}@  = typename underlying_type<T>::type;
   template<class Fn, class... ArgTypes>
     using @\libglobal{invoke_result_t}@    = typename invoke_result<Fn, ArgTypes...>::type;
+  template<class Fn, class Tuple>
+    using @\libglobal{apply_result_t}@     = typename apply_result<Fn, Tuple>::type;
   template<class T>
     using unwrap_reference_t = typename unwrap_reference<T>::type;
   template<class T>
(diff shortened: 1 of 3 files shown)


jwakely:
I think we should bump `__cpp_lib_apply` for this.

source/meta.tex
```diff
@@ -1589,6 +1607,26 @@
  \tcode{Fn}, \tcode{R}, and all types in the template parameter pack \tcode{ArgTypes}
  shall be complete types, \cv{}~\keyword{void}, or
  arrays of unknown bound.                                             \\
+
+\indexlibraryglobal{is_applicable}%
+\tcode{template<class Fn, class Tuple>}\br
+ \tcode{struct is_applicable;}              &
+ \exposconcept{tuple-like} is \tcode{true} and
```
eisenwave:
```suggestion
 \tcode{\exposconcept{tuple-like}<Tuple>} is \tcode{true} and
```

source/meta.tex
```diff
@@ -1589,6 +1607,26 @@
...
+ \tcode{Fn} and \tcode{Tuple}
+ shall be complete types, \cv{}~\keyword{void}, or
+ arrays of unknown bound.                                             \\ \rowsep
+
+\indexlibraryglobal{is_nothrow_applicable}%
+\tcode{template<class Fn, class Tuple>}\br
+ \tcode{struct is_nothrow_applicable;}              &
+ \tcode{is_applicable_v<}\br\tcode{Fn, Tuple>} is \tcode{true} and
+ the expression \tcode{\placeholdernc{INVOKE}(declval<Fn>(), \placeholdernc{ELEMS-OF}(Tuple)...)}
+ is known not to throw any exceptions\iref{expr.unary.noexcept}       &
```
eisenwave:
```suggestion
 is known not to throw any exceptions\iref{expr.unary.noexcept}.       &
```

source/meta.tex
```diff
@@ -2029,6 +2067,28 @@
...
+ \begin{note}
+ The compilation of the expression can result in side effects such as
+ the instantiation of class template specializations and function
+ template specializations, the generation of implicitly-defined
+ functions, and so on. Such side effects are not in the ``immediate
+ context'' and can result in the program being ill-formed.
+ \end{note}
+ \expects \tcode{Fn} and \tcode{Tuple}
+ are complete types, \cv{}~\keyword{void}, or arrays of
+ unknown bound.\\ \rowsep
```
eisenwave: It would be nice to have semantic line breaks in this area, at least for new sentences.

Eisenwave (changes requested):

source/meta.tex
jwakely: Fixed, thanks

source/meta.tex
jwakely: This is consistent with `is_nothrow_invocable` and all related invocable traits, and the majority of rows in the table. Only a few newer rows end with a full stop (e.g. `is_virtual_base_of`, `is_pointer_interconvertible`).

source/meta.tex
jwakely: Agreed. It was copy+pasted from `is_invocable` but I'll reflow it.

source/meta.tex
eisenwave: Well, I was just going off of paper discrepancies. There is a period in the paper at that point. I would assume that there should be a fixup commit for consistency then.

jwakely: Fixed, and #8005 opened for the inconsistency.

Eisenwave (approved):

jwakely:
> I think we should bump `__cpp_lib_apply` for this.

I raised this on mattermost and got agreement. I was going to do it in a fixup commit, but I realised we might want to consider whether to add the macro to `<type_traits>`. It's currently only in `<tuple>` (and `<version>` of course).

tkoeppe:
> > I think we should bump `__cpp_lib_apply` for this.
> 
> I raised this on mattermost and got agreement. I was going to do it in a fixup commit, but I realised we might want to consider whether to add the macro to `<type_traits>`. It's currently only in `<tuple>` (and `<version>` of course).

@jwakely Is there an outcome?

jwakely:
I forgot all about it. I think it should be bumped and added to <type_traits>. We generally treat those macro values as editorial, but I'm not sure about which headers they're in. Do you want me to open an issue?

tkoeppe:
If you're OK with it, I'd be happy to just editorially include this in this PR (feel free to add a commit?) -- what do you think?

tkoeppe:
Added it, PTAL

Issue #7949, opened 2025-06-21
[2025-06 LWG Motion 12] P2664R11 Extend std::simd with permutation API


jensmaurer:
https://wiki.edg.com/pub/Wg21sofia2025/StrawPolls/P2664R11.html

mattkretz:
mine

PR #8000, opened 2025-06-23
P2664R11 Extend std::simd with permutation API

mattkretz:
Fixes: #7949
Fixes: cplusplus/papers#1383

I left a FIXME for an overfull \hbox that I don't know how to fix.


mattkretz:
moved to new PR

PR #8006, opened 2025-06-23
P3060R3 Add std::views::indices(n)

### LINKED ISSUE #7966: [2025-06 LWG Motion 29] P3060R3 Add std::views::indices(n)
https://wiki.edg.com/pub/Wg21sofia2025/StrawPolls/p3060r3.html

jwakely:
Fixed "if T models is-integer-like" which is a category error because that's a variable template not a concept.

Fixes #7966 7966

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -218,7 +218,10 @@ namespace std::ranges {
   template<class W, class Bound>
     constexpr bool @\libspec{enable_borrowed_range}{iota_view}@<iota_view<W, Bound>> = true;
 
-  namespace views { inline constexpr @\unspecnc@ iota = @\unspecnc@; }
+  namespace views {
+    inline constexpr @\unspecnc@ iota = @\unspecnc@;
+    inline constexpr @\unspecnc@ indices = @\unspecnc@;
+  }
 
   // \ref{range.repeat}, repeat view
   template<@\libconcept{move_constructible}@ T, @\libconcept{semiregular}@ Bound = unreachable_sentinel_t>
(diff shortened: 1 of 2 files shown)


source/ranges.tex
```diff
@@ -2784,6 +2787,17 @@
...
 
+\pnum
+\indexlibrarymember{indices}{views}%
+The name \tcode{views::indices} denotes a
+customization point object\iref{customization.point.object}.
+Given subexpression \tcode{E},
+let \tcode{T} be \tcode{remove_cvref_t<decltype((E))>}.
+\tcode{views::indices(E)} is expression-equivalent to
+\tcode{views::iota(T(0), E)}
+if \exposid{is-integer-like<T>} is \tcode{true},
```
eisenwave:
```suggestion
if \tcode{\exposid{is-integer-like}<T>} is \tcode{true},
```

Eisenwave (changes requested):

source/ranges.tex
jwakely: Fixed, thanks

source/ranges.tex
```diff
@@ -2784,6 +2787,17 @@
...
+\pnum
+\indexlibrarymember{indices}{views}%
+The name \tcode{views::indices} denotes a
+customization point object\iref{customization.point.object}.
+Given subexpression \tcode{E},
+let \tcode{T} be \tcode{remove_cvref_t<decltype((E))>}.
+\tcode{views::indices(E)} is expression-equivalent to
+\tcode{views::iota(T(0), E)}
+if \tcode{\exposid{is-integer-like}<T>} is \tcode{true},
+and ill-formed otherwise.
```
jwakely, line 2817: Do we want this?

```suggestion
and is ill-formed otherwise.
```

Eisenwave (approved):

source/ranges.tex
eisenwave: I think it's fine either way.

tkoeppe: I think the paper wording is fine.

PR #8008, opened 2025-06-24
P2996R13 Reflection for C++26

jensmaurer:
Feel free to cherry-pick into #8004 once @Eisenwave is done with the library part.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -2185,7 +2185,7 @@ do not contribute to this set.
 The set of entities
 is determined in the following way:
 \begin{itemize}
-\item If \tcode{T} is \tcode{std::meta::info}\iref{meta.reflection.synop},
+\item If \tcode{T} is \tcode{std::meta::info}\iref{meta.syn},
 its associated set of entities is the singleton containing
 the enumeration type \tcode{std::meta::operators}\iref{meta.reflection.operators}.
 \begin{note}
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -114,17 +114,33 @@ before continuing to parse the program that contains it.
 A \defn{variable} is introduced by the
 declaration of
 a reference other than a non-static data member or of
-an object. The variable's name, if any, denotes the reference or object.
+an object.
 
 \pnum
-An \defn{entity} is a value, object, reference,
+An \defn{entity} is a
+variable,
 structured binding,
 result binding,
-function, enumerator, type,
-class member, bit-field, template, template specialization, namespace, or
-pack. An entity $E$ is denoted by the name (if any)
-that is introduced by a declaration of $E$ or
-by a \grammarterm{typedef-name} introduced by a declaration specifying $E$.
+function,
+enumerator,
+type,
+type alias,
+non-static data member,
+bit-field,
+template,
+namespace,
+namespace alias,
+template parameter,
+function parameter, or
+\grammarterm{init-capture}.
+The \defnadj{underlying}{entity} of an entity is that entity
+unless otherwise specified.
+A name \defnx{denotes}{denote} the underlying entity of
+the entity declared by each declaration that introduces the name.
+\begin{note}
+Type aliases and namespace aliases have underlying entities
+that are distinct from themselves.
+\end{note}
 
 \pnum
 A \defnadj{local}{entity} is a variable with
@@ -153,7 +169,7 @@ one or more names and/or entities into a translation
 unit.
 If so, the
 declaration specifies the interpretation and semantic properties of these names.
-A declaration of an entity or \grammarterm{typedef-name} $X$ is
+A declaration of an entity $X$ is
 a redeclaration of $X$
 if another declaration of $X$ is reachable from it\iref{module.reach};
 otherwise, it is a \defnadj{first}{declaration}.
@@ -221,12 +237,15 @@ it is a
 \item it is
 an \grammarterm{alias-declaration}\iref{dcl.typedef},
 \item it is
-a
-\grammarterm{using-declaration}\iref{namespace.udecl},
+a \grammarterm{namespace-alias-definition}\iref{namespace.alias},
+\item it is
+a \grammarterm{using-declaration}\iref{namespace.udecl},
 \item it is
 a \grammarterm{deduction-guide}\iref{temp.deduct.guide},
 \item it is
 a \grammarterm{static_assert-declaration}\iref{dcl.pre},
+\item it is
+a \grammarterm{consteval-block-declaration},
 \item
 it is an
 \grammarterm{attribute-declaration}\iref{dcl.pre},
@@ -264,7 +283,6 @@ struct X {                      // defines \tcode{X}
 int X::y = 1;                   // defines \tcode{X::y}
 enum { up, down };              // defines \tcode{up} and \tcode{down}
 namespace N { int d; }          // defines \tcode{N} and \tcode{N::d}
-namespace N1 = N;               // defines \tcode{N1}
 X anX;                          // defines \tcode{anX}
 
 \end{codeblock}
@@ -275,6 +293,7 @@ extern const int c;             // declares \tcode{c}
 int f(int);                     // declares \tcode{f}
 struct S;                       // declares \tcode{S}
 typedef int Int;                // declares \tcode{Int}
+namespace N1 = N;               // declares \tcode{N1}
 extern X anotherX;              // declares \tcode{anotherX}
 using N::d;                     // declares \tcode{d}
 \end{codeblock}
@@ -368,8 +387,9 @@ a conversion in an initialization or conversion sequence in such a context.
 The set of \defn{potential results} of an expression $E$ is
 defined as follows:
 \begin{itemize}
-\item If $E$ is an
-\grammarterm{id-expression}\iref{expr.prim.id}, the set
+\item If $E$ is
+an \grammarterm{id-expression}\iref{expr.prim.id} or
+a \grammarterm{splice-expression}\iref{expr.prim.splice}, the set
 contains only $E$.
 \item If $E$ is a subscripting operation\iref{expr.sub} with
 an array operand, the set contains the potential results of that operand.
@@ -395,17 +415,21 @@ contains the potential results of the right operand.
 \item Otherwise, the set is empty.
 \end{itemize}
 \begin{note}
-This set is a (possibly-empty) set of \grammarterm{id-expression}{s},
+This set is a (possibly-empty) set of
+\grammarterm{id-expression}{s} and \grammarterm{splice-expression}s,
 each of which is either $E$ or a subexpression of $E$.
 \begin{example}
 In the following example, the set of potential results of the initializer
 of \tcode{n} contains the first \tcode{S::x} subexpression, but not the second
 \tcode{S::x} subexpression.
+The set of potential results of the initializer of \tcode{o} contains
+the subexpression \tcode{[:\caret\caret S::x:]}.
 \begin{codeblock}
 struct S { static const int x = 0; };
 const int &f(const int &r);
 int n = b ? (1, S::x)           // \tcode{S::x} is not odr-used here
           : f(S::x);            // \tcode{S::x} is odr-used here, so a definition is required
+int o = [:^^S::x:];
 \end{codeblock}
 \end{example}
 \end{note}
@@ -453,7 +477,9 @@ This covers
 \pnum
 \label{term.odr.use}%
 A variable is named by an expression
-if the expression is an \grammarterm{id-expression} that denotes it.
+if the expression is an \grammarterm{id-expression} or
+\grammarterm{splice-expression}\iref{expr.prim.splice}
+that designates it.
 A variable \tcode{x} that is named by a
 potentially-evaluated expression $N$
 that appears at a point $P$
@@ -510,7 +536,12 @@ int h(bool cond) {
 \end{example}
 
 \pnum
-A structured binding is odr-used if it appears as a potentially-evaluated expression.
+A structured binding is named by an expression
+if that expression is either an \grammarterm{id-expression} or
+a \grammarterm{splice-expression}
+that designates that structured binding.
+A structured binding is odr-used
+if it is named by a potentially-evaluated expression.
 
 \pnum
 \tcode{*\keyword{this}} is odr-used if \keyword{this} appears as a potentially-evaluated expression
@@ -696,7 +727,16 @@ defined\iref{class.derived}, or
 \end{note}
 
 \pnum
-For any definable item \tcode{D} with definitions in multiple translation units,
+If a definable item \tcode{D} is defined in a translation unit
+by an injected declaration $X$\iref{expr.const} and
+another translation unit contains a definition of \tcode{D},
+that definition shall be an injected declaration
+having the same characteristic sequence as $X$;
+a diagnostic is required only if \tcode{D} is attached to a named module and
+a prior definition is reachable at the point where a later definition occurs.
+
+\pnum
+For any other definable item \tcode{D} with definitions in multiple translation units,
 \begin{itemize}
 \item
 if \tcode{D} is a non-inline non-templated function or variable, or
@@ -722,7 +762,7 @@ where the definition of a closure type
 is considered to consist of the sequence of tokens of
 the corresponding \grammarterm{lambda-expression}.
 \item In each such definition, corresponding names, looked up
-according to~\ref{basic.lookup}, shall refer to the same entity, after
+according to~\ref{basic.lookup}, shall denote the same entity, after
 overload resolution\iref{over.match} and after matching of partial
 template specializations\iref{temp.spec.partial.match}, except that a name can refer to
 \begin{itemize}
@@ -738,7 +778,7 @@ or
 \item
 a reference with internal or no linkage
 initialized with a constant expression such that
-the reference refers to the same entity in all definitions of \tcode{D}.
+the reference refers to the same object or function in all definitions of \tcode{D}.
 \end{itemize}
 
 \item In each such definition, except within
@@ -767,11 +807,17 @@ function.
 \item In each such definition,
 a default argument used by an (implicit or explicit) function call or
 a default template argument used by an (implicit or explicit)
-\grammarterm{template-id} or \grammarterm{simple-template-id}
+\grammarterm{template-id},
+\grammarterm{simple-template-id}, or
+\grammarterm{splice-specialization-specifier}
 is treated as if its token sequence
 were present in the definition of \tcode{D};
 that is, the default argument or default template argument
 is subject to the requirements described in this paragraph (recursively).
+
+\item In each such definition,
+corresponding \grammarterm{reflect-expression}s\iref{expr.reflect}
+compute equivalent values\iref{expr.eq}.
 \end{itemize}
 
 \pnum
@@ -926,6 +972,9 @@ A declaration's \defnadj{target}{scope} is the scope it inhabits.
 Any names (re)introduced by a declaration are \defnx{bound}{name!bound} to it
 in its target scope.
 \end{itemize}
+The \defnadj{host}{scope} of a declaration is
+the inhabited scope if that scope is a block scope and
+the target scope otherwise.
 An entity \defnx{belongs}{entity!belong} to a scope $S$
 if $S$ is the target scope of a declaration of the entity.
 \begin{note}
@@ -1006,7 +1055,7 @@ unless
 \item
 either is a \grammarterm{using-declarator}, or
 \item
-one declares a type (not a \grammarterm{typedef-name}) and the other declares a
+one declares a type (not a type alias) and the other declares a
 variable,
 non-static data member other than of an anonymous union\iref{class.union.anon},
 enumerator,
@@ -1663,7 +1712,7 @@ the \grammarterm{using-declarator}\iref{namespace.udecl}.
 In certain contexts, only certain kinds of declarations are included.
 After any such restriction, any declarations of classes or enumerations are discarded if any other declarations are found.
 \begin{note}
-A type (but not a \grammarterm{typedef-name} or template)
+A type (but not a type alias or template)
 is therefore hidden by any other entity in its scope.
 \end{note}
 \indextext{type-only!lookup|see{lookup, type-only}}%
@@ -1671,8 +1720,8 @@ However, if a lookup is \defnx{type-only}{lookup!type-only},
 only declarations of
 types and templates whose specializations are types are considered;
 furthermore, if declarations
-of a \grammarterm{typedef-name} and of the type to which it refers are found,
-the declaration of the \grammarterm{typedef-name} is discarded
+of a type alias and of its underlying entity are found,
+the declaration of the type alias is discarded
 instead of the type declaration.
 
 \rSec2[class.member.lookup]{Member name lookup}%
@@ -2136,7 +2185,16 @@ do not contribute to this set.
 The set of entities
 is determined in the following way:
 \begin{itemize}
-\item If \tcode{T} is a fundamental type, its associated set of
+\item If \tcode{T} is \tcode{std::meta::info}\iref{meta.reflection.synop},
+its associated set of entities is the singleton containing
+the enumeration type \tcode{std::meta::operators}\iref{meta.reflection.operators}.
+\begin{note}
+The \tcode{std::meta::info} type is a type alias,
+so an explicit rule is needed to associate calls
+whose arguments are reflections with the namespace \tcode{std::meta}.
+\end{note}
+
+\item If \tcode{T} is any other fundamental type, its associated set of
 entities is empty.
 
 \item If \tcode{T} is a class type (including unions),
@@ -2294,8 +2352,14 @@ Lookup of an \grammarterm{identifier}
 followed by a \tcode{::} scope resolution operator
 considers only
 namespaces, types, and templates whose specializations are types.
-If a name, \grammarterm{template-id}, or \grammarterm{computed-type-specifier}
+If a
+name,
+\grammarterm{template-id},
+\grammarterm{splice-scope-specifier}, or
+\grammarterm{computed-type-specifier}
 is followed by a \tcode{::},
+it shall either be
+a dependent \grammarterm{splice-scope-specifier}\iref{temp.dep.splice} or
 it shall designate a namespace, class, enumeration, or dependent type,
 and the \tcode{::} is never interpreted as
 a complete \grammarterm{nested-name-specifier}.
@@ -2344,6 +2408,7 @@ A \defnadj{qualified}{name} is
 \item a \grammarterm{typename-specifier},
 \item a \grammarterm{qualified-namespace-specifier}, or
 \item a \grammarterm{nested-name-specifier},
+\grammarterm{reflection-name},
 \grammarterm{elaborated-type-specifier}, or
 \grammarterm{class-or-decltype}
 that has a \grammarterm{nested-name-specifier}\iref{expr.prim.id.qual}.
@@ -2738,6 +2803,70 @@ during the lookup for a \grammarterm{namespace-name} or for a name in a
 only namespace names are considered.%
 \indextext{lookup!name|)}%
 
+\rSec1[basic.splice]{Splice specifiers}
+\indextext{splice|(}%
+
+\begin{bnf}
+\nontermdef{splice-specifier}\br
+  \terminal{[:} constant-expression \terminal{:]}
+\end{bnf}
+
+\begin{bnf}
+\nontermdef{splice-specialization-specifier}\br
+  splice-specifier \terminal{<} \opt{template-argument-list} \terminal{>}
+\end{bnf}
+
+\pnum
+The \grammarterm{constant-expression} of a \grammarterm{splice-specifier}
+shall be a converted constant expression of
+type \tcode{std::meta::info}\iref{expr.const}.
+A \grammarterm{splice-specifier}
+whose converted \grammarterm{constant-expression} represents
+a construct $X$ is said to \defn{designate} either
+\begin{itemize}
+\item the underlying entity of $X$ if $X$ is an entity\iref{basic.pre}, or
+\item $X$ otherwise.
+\end{itemize}
+\begin{note}
+A \grammarterm{splice-specifier} is dependent
+if the converted \grammarterm{constant-expression} is
+value-dependent\iref{temp.dep.splice}.
+\end{note}
+
+\pnum
+A non-dependent \grammarterm{splice-specifier} of
+a \grammarterm{splice-specialization-specifier} shall designate a template.
+
+\pnum
+\begin{note}
+A \tcode{<} following a \grammarterm{splice-specifier} is interpreted as
+the delimiter of a \grammarterm{template-argument-list}
+when the \grammarterm{splice-specifier} is preceded by
+the keyword \keyword{template} or the keyword \keyword{typename}, or
+when it appears in a type-only context\iref{temp.names}.
+\begin{example}
+\begin{codeblock}
+constexpr int v = 1;
+template<int V> struct TCls {
+  static constexpr int s = V + 1;
+};
+
+using alias = [:^^TCls:]<([:^^v:])>;
+  // OK, a \grammarterm{splice-specialization-specifier} with a \grammarterm{splice-expression} as a template argument
+
+static_assert(alias::s == 2);
+
+auto o1 = [:^^TCls:]<([:^^v:])>();              // error: < means less than
+auto o2 = typename [:^^TCls:]<([:^^v:])>();     // OK, \tcode{o2} is an object of type \tcode{TCls<1>}
+
+consteval int bad_splice(std::meta::info v) {
+  return [:v:];                                 // error: \tcode{v} is not constant
+}
+\end{codeblock}
+\end{example}
+\end{note}
+\indextext{splice|)}
+
 \rSec1[basic.link]{Program and linkage}%
 \indextext{linkage|(}
 
@@ -2909,6 +3038,8 @@ and either
 \item
 they appear in the same translation unit, or
 \item
+they both declare type aliases or namespace aliases that have the same underlying entity, or
+\item
 they both declare names with module linkage and are attached to the same module, or
 \item
 they both declare names with external linkage.
@@ -3021,6 +3152,16 @@ A declaration $D$ \defnx{names}{name} an entity $E$ if
 \item
 $D$ contains a \grammarterm{lambda-expression} whose closure type is $E$,
 \item
+$D$ contains
+a \grammarterm{reflect-expression} or a \grammarterm{splice-specifier}
+that, respectively, represents or designates $E$,
+\item
+$D$ is an injected declaration\iref{expr.const}
+whose characteristic sequence contains a reflection
+that represents
+a data member description ($T$, $N$, $A$, $W$, $\mathit{NUA}$)\iref{class.mem.general}
+for which $T$ is $E$,
+\item
 $E$ is not a function or function template and $D$ contains an
 \grammarterm{id-expression},
 \grammarterm{type-specifier},
@@ -3069,7 +3210,7 @@ certain explicit specializations of it would be usable in other translation unit
 An entity is \defnx{TU-local}{TU-local!entity} if it is
 \begin{itemize}
 \item
-a type, function, variable, or template that
+a type, type alias, namespace, namespace alias, function, variable, or template that
 \begin{itemize}
 \item
 has a name with internal linkage, or
@@ -3101,14 +3242,28 @@ A specialization can be produced by implicit or explicit instantiation.
 A value or object is \defnx{TU-local}{TU-local!value or object} if either
 \begin{itemize}
 \item
+it is of TU-local type,
+\item
 it is, or is a pointer to,
-a TU-local function or the object associated with a TU-local variable, or
+a TU-local function or the object associated with a TU-local variable,
 \item
 it is an object of class or array type and
 any of its subobjects or
 any of the objects or functions
 to which its non-static data members of reference type refer
-is TU-local and is usable in constant expressions.
+is TU-local and is usable in constant expressions, or
+\item
+it is a reflection value\iref{basic.fundamental} that represents
+\begin{itemize}
+\item
+an entity, value, or object that is TU-local,
+\item
+a direct base class relationship (\tcode{D}, \tcode{B})\iref{class.derived.general}
+for which either \tcode{D} or \tcode{B} is TU-local, or
+\item
+a data member description ($T$, $N$, $A$, $W$, $\mathit{NUA}$)\iref{class.mem.general}
+for which $T$ is TU-local.
+\end{itemize}
 \end{itemize}
 
 \pnum
@@ -3158,6 +3313,16 @@ namespace N {
 void adl(double);
 
 inline void h(auto x) { adl(x); }   // OK, but certain specializations are exposures
+
+constexpr std::meta::info r1 = ^^g<0>;  // OK
+namespace N2 {
+  static constexpr std::meta::info r2 = ^^g<1>;     // OK, \tcode{r2} is TU-local
+}
+constexpr std::meta::info r3 = ^^f;                 // error: \tcode{r3} is an exposure of \tcode{f}
+
+constexpr auto ctx = std::meta::access_context::current();
+constexpr std::meta::info r4 =
+  std::meta::members_of(^^N2, ctx)[0];              // error: \tcode{r4} is an exposure of \tcode{N2::r2}
 \end{codeblocktu}
 \begin{codeblocktu}{Translation unit \#2}
 module A;
@@ -5021,7 +5186,7 @@ a function type, not a reference type, and not \cv{}~\keyword{void}.
 \label{term.scalar.type}%
 Arithmetic types\iref{basic.fundamental}, enumeration types,
 pointer types, pointer-to-member types\iref{basic.compound},
-\tcode{std::nullptr_t},
+\tcode{std::meta::\brk{}info}, \tcode{std::nullptr_t},
 and
 cv-qualified\iref{basic.type.qualifier} versions of these
 types are collectively called
@@ -5092,6 +5257,21 @@ Two types \cvqual{cv1} \tcode{T1} and \cvqual{cv2} \tcode{T2} are
 if \tcode{T1} and \tcode{T2} are the same type,
 layout-compatible enumerations\iref{dcl.enum}, or
 layout-compatible standard-layout class types\iref{class.mem}.
+
+\pnum
+A type is \defn{consteval-only} if it is either
+\tcode{std::meta::info} or
+a type compounded from a consteval-only type\iref{basic.compound}.
+Every object of consteval-only type shall be
+\begin{itemize}
+\item
+the object associated with a constexpr variable or a subobject thereof,
+\item
+a template parameter object\iref{temp.param} or a subobject thereof, or
+\item
+an object whose lifetime begins and ends
+during the evaluation of a core constant expression.
+\end{itemize}
 
 \rSec2[basic.fundamental]{Fundamental types}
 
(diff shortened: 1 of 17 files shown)


tkoeppe:
@jensmaurer did you link the correct PR, or did you mean #8004?

jensmaurer:
Fixed link, thanks.

tkoeppe:
We could certainly move all the commits into a single PR and the test and review that again. Let's wait for a first round of reviews on both current PRs to settle, then we'll take it from there.

eisenwave:
@tkoeppe by the way, you've put the milestone date for 2025-06 as 06-31. Is that actually correct? I was under the impression that we would want to have the updated working draft in the post-Sofia mailing, if not sooner so that the ISO review process can start ASAP.

tkoeppe:
@Eisenwave Sorry, no idea, I don't recall making that label or how it was made. Don't worry about it. We're not going to stop the process because of what a label says.

source/lex.tex
```diff
@@ -743,7 +794,7 @@
 \begin{bnf}
 \microtypesetup{protrusion=false}
 \nontermdef{operator-or-punctuator} \textnormal{one of}\br
-    \terminal{\{ \ \ \ \ \ \ \ \} \ \ \ \ \ \ \ [ \ \ \ \ \ \ \ ] \ \ \ \ \ \ \ ( \ \ \ \ \ \ \ )}\br
+    \terminal{\{ \ \ \ \ \ \ \ \} \ \ \ \ \ \ \ [ \ \ \ \ \ \ \ ] \ \ \ \ \ \ \ ( \ \ \ \ \ \ \ ) \ \ \ \ \ \ \ [: \ \ \ \ \ \ :]}\br
     \terminal{<\% \ \ \ \ \ \ \%> \ \ \ \ \ \ <: \ \ \ \ \ \ :> \ \ \ \ \ \ ; \ \ \ \ \ \ \ : \ \ \ \ \ \ \ ...}\br
     \terminal{? \ \ \ \ \ \ \ :: \ \ \ \ \ \ . \ \ \ \ \ \ \ .* \ \ \ \ \ \ -> \ \ \ \ \ \ ->* \ \ \ \ \ \~}\br
     \terminal{! \ \ \ \ \ \ \ + \ \ \ \ \ \ \ - \ \ \ \ \ \ \ * \ \ \ \ \ \ \ / \ \ \ \ \ \ \ \% \ \ \ \ \ \ \ \caret{} \ \ \ \ \ \ \ \& \ \ \ \ \ \ \ |}\br
```
rogerorr, line 807: The double caret added in P2996R13 appears to be missing from this line

source/basic.tex
```diff
@@ -114,17 +114,32 @@
...
+type alias,
+non-static data member,
+bit-field,
+template,
+namespace,
+namespace alias,
+template parameter,
+function parameter, or
+\grammarterm{init-capture}.
+The underlying entity of an entity is that entity unless otherwise specified.
```
rogerorr: In P2996R13 this first mention of "underlying entity" is a definition.

source/basic.tex
```diff
@@ -1664,16 +1712,16 @@
...
 is therefore hidden by any other entity in its scope.
 \end{note}
 \indextext{type-only!lookup|see{lookup, type-only}}%
 However, if a lookup is \defnx{type-only}{lookup!type-only},
 only declarations of
 types and templates whose specializations are types are considered;
 furthermore, if declarations
-of a \grammarterm{typedef-name} and of the type to which it refers are found,
-the declaration of the \grammarterm{typedef-name} is discarded
+of a type alias and of the type to which it refers are found,
```
rogerorr: P2996R13 changes 'the type to which it refers' to 'its underlying entity'.

source/basic.tex
```diff
@@ -3022,13 +3152,23 @@
 \item
 $D$ contains a \grammarterm{lambda-expression} whose closure type is $E$,
 \item
+$D$ contains
+a \grammarterm{reflect-expression} or a \grammarterm{splice-specifier}
+that, respectively, represents or designates $E$,
+\item
```
rogerorr, line 3158: I believe this item (new lines 3158-3163) should have been inserted after old line 3036. If not, the change to the "or" needs rolling back.

source/basic.tex
```diff
@@ -5022,7 +5186,7 @@
 \label{term.scalar.type}%
 Arithmetic types\iref{basic.fundamental}, enumeration types,
 pointer types, pointer-to-member types\iref{basic.compound},
-\tcode{std::nullptr_t},
+\tcode{std::meta::\linebreak info}, \tcode{std::nullptr_t},
```
rogerorr: I don't know whether the \linebreak here is expected - this may be my lack of knowledge?

source/basic.tex
```diff
@@ -5400,6 +5579,94 @@
 pointer-to-member conversions\iref{conv.ptr,conv.mem}.
 \tcode{\keyword{sizeof}(std::nullptr_t)} shall be equal to \tcode{\keyword{sizeof}(\keyword{void}*)}.
 
+\pnum
+A value of type \tcode{std::meta::info} is called a \defn{reflection}.
+There exists a unique null reflection;
```
rogerorr: In P2996R13 this is a definition of 'null reflection'

source/basic.tex
eisenwave: I would assume that it's just a way to sidestep overfull hbox problems. I find it dubious that a mandatory line break is used here though, which can result in very bad layout when the content changes eventually. `\brk{}` (our macro for discretionary line breaks) after `meta::` would be better:
```suggestion
\tcode{std::meta::\brk{}info}, \tcode{std::nullptr_t},
```

The underlying problem is that LaTeX doesn't allow line breaking in `\texttt` for the purpose of paragraph justification, so you have to sprinkle in these line breaks.

The underlying problem is also that LaTeX will happily flow the content outside the hbox rather than just breaking onto the next line at the cost of making things uglier, which is an insane design decision.

jensmaurer:
That milestone is by me.  I've moved the date to 07-15.

source/lex.tex
jensmaurer: fixed

source/basic.tex
jensmaurer: fixed

source/basic.tex
jensmaurer: fixed

source/basic.tex
jensmaurer: I've fixed the "or".   I think the new placement is a bit better, because we're in a group of bullets that start with "D"; the other group starts with "E".

source/basic.tex
jensmaurer: Fixed \linebreak -> \brk{}

source/basic.tex
jensmaurer: fixed

source/basic.tex
rogerorr: That's fine by me :)

source/expressions.tex
```diff
@@ -6360,6 +6621,232 @@
...
+\item
+If lookup finds an injected-class-name\iref{class.pre}, then:
+\begin{itemize}
+\item
+If the \grammarterm{reflection-name} is of the form
+\tcode{\grammarterm{nested-name-specifier} \keyword{template} \grammarterm{identifier}},
+then $R$ represents the class template named by the injected-class-name.
+\item
+Otherwise, the injected-class-name shall be unambiguous
+when considered as ae \grammarterm{type-name} and
```
rogerorr: s/ ae / a /

source/expressions.tex
```diff
@@ -8570,17 +9112,17 @@
...
 \item
-a potentially-evaluated \grammarterm{id-expression}
-that denotes an immediate function
-that is not a subexpression of an immediate invocation, or
+it is an \grammarterm{id-expression} or \grammarterm{splice-expression}
+that designates an immediate function,
 \item
-an immediate invocation that is not a constant expression
-and is not a subexpression of an immediate invocation.
+it is immediate invocation that is not a constant expression, or
```
rogerorr: s/it is immediate/it is an immediate/

source/declarations.tex
```diff
@@ -4393,9 +4503,19 @@
...
-unless it is used to form a pointer to member\iref{expr.unary.op}.
+A non-static member shall not be designated in a default argument unless
+\begin{itemize}
+\item
+it is designated by
+the \grammarterm{id-expression} or \grammarterm{splice-expression}
+of a class member access expression\iref{expr.ref},
+\item
+it is designated by an expression
+used to form a pointer to member\iref{expr.unary.op}.
```
rogerorr: The period should be replaced by ', or'

source/classes.tex
```diff
@@ -3515,7 +3568,12 @@
 The class denoted by the \grammarterm{class-or-decltype} of
 a \grammarterm{base-specifier} is called a
 \defnadj{direct}{base class}
-for the class being defined.
+for the class being defined;
+for each such \grammarterm{base-specifier},
+the corresponding direct base class relationship
```
rogerorr: P2996R13 appear to define 'direct base class relationship' here.

source/classes.tex
```diff
@@ -4837,10 +4906,10 @@
 left operand (considered as a pointer in the
```
rogerorr, line 4905: old line 4814 is missing a change of 'named' to 'designated'

source/templates.tex
```diff
@@ -3024,7 +3070,10 @@
 }
 \end{codeblock}
 \end{example}
-There is no syntax to form a \grammarterm{template-id}\iref{temp.names}
+An expression designating
+a particular specialization of a conversion function template
+can only be formed with a \grammarterm{splice-expression}.
+There is no analogous syntax to form a \grammarterm{template-id}\iref{temp.names}
```
rogerorr, line 3076: P2996R13 appends " for such a function" to this line

source/templates.tex
```diff
@@ -4869,17 +4931,26 @@
 (which necessarily declares a constant template parameter).
 \end{itemize}
 \end{itemize}
+A \grammarterm{splice-specifier} or
+\grammarterm{splice-specialization-specifier}\iref{basic.splice}
+is said to be in a type-only context
```
rogerorr, line 4936: P2996R13 defines "type-only context" here

source/templates.tex
```diff
@@ -5850,6 +5932,107 @@
...
+the \grammarterm{id-expression} is value-dependent.
+\end{itemize}
+
+\pnum
+A \grammarterm{splice-expression} is value-dependent
+if its \grammarterm{splice-specifier} or
+\grammarterm{splice-specialization-specifier}
+is dependent\iref{temp.dep.splice}.
+
+\rSec3[temp.dep.splice]{Dependent splice specifier}
```
rogerorr: P2996R13 has the plural: "Dependent splice specifiers". Just checking the change is deliberate

source/templates.tex
jensmaurer: Yes, but there's already a definition above, and we can only have one.

source/templates.tex
jensmaurer: Everything in the vicinity is plural, so let's go plural here, too.

source/templates.tex
jensmaurer: ... and leaves out the last part of the sentence without showing strike-out.  Fixed.

source/classes.tex
jensmaurer: fixed

source/templates.tex
jensmaurer: fixed

source/classes.tex
jensmaurer: fixed

source/declarations.tex
jensmaurer: fixed

source/expressions.tex

source/expressions.tex

source/templates.tex
rogerorr: ah, sorry - missed that :)

source/templates.tex
tkoeppe: Thanks, both!

source/overloading.tex
```diff
@@ -1698,13 +1717,76 @@
...
+\end{itemize}
+If \tcode{C} is selected as the best viable function\iref{over.match.best}:
+\begin{itemize}
+\item
+$G$ shall contain exactly one scope (call it $S$).
+\item
+If the candidates are denoted by a \grammarterm{splice-expression},
+then $S$ shall not be a block scope.
+\item
+The default arguments used in the call \tcode{C} are
```
katzdm: As discussed offline, probably should have been "the call ___to___ `C`".

tkoeppe: @jensmaurer Is that OK as an editorial follow-up? It's not in the paper.

jensmaurer: Yes.

source/compatibility.tex
```diff
@@ -253,6 +253,8 @@
 \libheaderref{hive},
 \libheaderrefx{inplace_vector}{inplace.vector.syn},
 \libheaderref{linalg},
+%FIXME: \libheaderref{meta} after renaming to meta.syn
+\libheaderrefx{meta}{meta.type.synop},
```
jensmaurer: yes, please

source/lib-intro.tex
```diff
@@ -3082,6 +3089,26 @@
 either a standard library non-static member function\iref{member.functions}
 or an instantiation of a standard library member function template.
 
+\pnum
+Let \tcode{\placeholder{F}} denote
+a standard library function or function template.
+Unless \tcode{\placeholder{F}} is designated an addressable function,
+it is unspecified if or how
+a reflection value designating the associated entity can be formed.
+%FIXME: Why is this not an example, but a note that begins with "For example"?
```
jensmaurer, line 3098: We occasionally do that.  Note there is no actual code in this note/example, so I'm fine with this being a note.  It's equally fine to use begin{example} and strike "For example," in the text.

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
 \end{example}
 \end{itemdescr}
 
+%FIXME: The paper doesn't actually specify that this subclause should exist,
+%but I read between the lines and took editorial liberty to create it.
+\rSec1[meta.reflection]{Reflection}
+
+%FIXME: Jens said this should be [meta.syn]
```
jensmaurer: Yes, please.  All headers (historic accidents notwithstanding) should be name.syn; it simplifies macro usage pointing to them so much.

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
 \end{example}
 \end{itemdescr}
 
+%FIXME: The paper doesn't actually specify that this subclause should exist,
```
jensmaurer: Sounds good, except that'd like to have a global rename of "meta.reflection*" labels to "meta.refl*" for space economy (without loss of clarity, in my view).

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+%but I read between the lines and took editorial liberty to create it.
+\rSec1[meta.reflection]{Reflection}
+
+%FIXME: Jens said this should be [meta.syn]
+\rSec2[meta.reflection.synop]{General}
+\indexheader{meta}%
+
+\begin{codeblock}
+%FIXME: I see the same in <algorithm> with a    // see \ref{initializer.list.syn} comment
```
jensmaurer: yes, let's have the "see"

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+\end{codeblock}
+\end{example}
+
+\rSec2[meta.reflection.operators]{Operator representations}
+
+%FIXME: I find it stylistically dubious that we don't say that
+%this is std::meta::operators;
+%we do repeat ranges:: in the itemdecls for ranges stuff.
+%Is this actually okay?
+%BIG fix needed in lots of places if not.
```
jensmaurer: For the ranges algorithms, there is great potential for confusion with the non-ranges algorithms with the same names, so we keep the namespace prefix in the itemdecls.

I'm not readily seeing such ranges:: prefixes in the itemdecls of [ranges], but maybe I've overlooked them.  Do you have examples from that section?

The mental model for itemdecls is that their "declarations" point to the declarations in the header or class definition and need to have sufficient information for a unique association (possibly by virtue of subclause nesting).  I'm not seeing an actual problem in that sense here.

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+
+\indexlibraryglobal{is_explicit}%
+\begin{itemdecl}
+consteval bool is_explicit(info r);
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\returns
+%FIXME: inconsistent font use for "explicit" between normative wording and note below.
```
jensmaurer: I think we want non-code font "explicit", because `explicit(true)` is totally fine and arguable not `explicit`.

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+\tcode{true} if \tcode{r} represents a member function that is declared explicit.
+Otherwise, \tcode{false}.
+\begin{note}
+If \tcode{r} represents a member function template that is declared \tcode{explicit},
+\tcode{is_explicit(r)} is still \tcode{false}
+because in general
+%FIXME: comma
```
jensmaurer: sure

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+\pnum
+\returns
+\tcode{true} if \tcode{r} represents a \tcode{noexcept} function type
+or a function with a non-throwing exception specification\iref{except.spec}.
+Otherwise, \tcode{false}.
+\begin{note}
+If \tcode{r} represents a function template that is declared \tcode{noexcept},
+\tcode{is_noexcept(r)} is still \tcode{false}
+because in general
+%FIXME: comma
```
jensmaurer: sure

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+
+\begin{itemdescr}
+\pnum
+Let $T$ be \tcode{type_of(r)} if \tcode{\exposid{has-type}(r)} is \tcode{true}.
+Otherwise, let $T$ be \tcode{dealias(r)}.
+
+\pnum
+\returns
+\tcode{true} if $T$ represents a
+%FIXME: "an", not "a"
```
jensmaurer: yes

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\returns
+\tcode{true} if \tcode{r} represents a type alias or namespace alias, respectively
+%FIXME: Very wonky note BEFORE the last sentence, inconsistent with the rest of this subclause.
+\begin{note}
+A specialization of an alias template is a type alias
+%FIXME: missing period in note.
```
jensmaurer: yes

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+\end{note}
+. %FIXME: period following note
```
jensmaurer: yes

Such punctuation mishaps can be silently fixed during the initial transcription.

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+\begin{itemdecl}
+consteval bool is_type_alias(info r);
+consteval bool is_namespace_alias(info r);
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\returns
+\tcode{true} if \tcode{r} represents a type alias or namespace alias, respectively
+%FIXME: Very wonky note BEFORE the last sentence, inconsistent with the rest of this subclause.
```
jensmaurer: Do you want to move it to just before \end{itemdescr} ?
Sure, go ahead.

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+\returns
+\tcode{true} if \tcode{r} represents a
+function template,
+class template,
+variable template,
+alias template, or
+concept.
+Otherwise, \tcode{false}.
+
+%FIXME: Why does this not get its own paragraph? Inconsistent!
```
jensmaurer: Well, there's a \pnum here, which seems totally fine.

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+\pnum
+\begin{note}
+A template specialization is not a template.
+%FIXME: begin example ? Maybe say "For example" at least?
```
jensmaurer: Yeah, "For example, "  sounds right.

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+\tcode{true} if \tcode{r} represents a
+function template,
+variable template,
+class template,
+alias template,
+conversion function template,
+literal operator template,
+constructor template, or
+concept
+%FIXME: comma
```
jensmaurer: yes

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+  \begin{itemize}
+  \item
+    If $P$ has reference type,
+    then $R$ is a reflection
+    representing the object or function referred to by $A$.
+  \item
+    Otherwise, if $P$ has class type,
+    then $R$ represents the corresponding template parameter object.
+  \item
+    %FIXME: "computed by" feels weird and unusual,
```
jensmaurer: Well, apparent "A" is an prvalue, and those "compute" values (for class prvalues),   But we've already covered class types above, so we're just talking about scalars here.  "the value of $A$" sounds good.

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+The \tcode{access_context}'s associated scope
+and designating class, respectively.
+\end{itemdescr}
+
+\begin{itemdecl}
+static consteval access_context @\libmember{current}{access_context}@() noexcept;
+\end{itemdecl}
+
+\begin{itemdescr}
+%FIXME: why are we just dumping this into the wild instead of in \remarks?
```
jensmaurer: \remarks sounds good, but needs to move later.  @jwakely ?

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+  introduced by a \grammarterm{lambda-expression} $L$,
+  the function parameter scope
+  corresponding to the call operator of the closure type of $L$.
+\item
+  Otherwise, \tcode{\exposid{ctx-scope}($S'$)}
+  %FIXME: comma
+  where $S'$ is the parent scope of $S$.
+\end{itemize}
+
+%FIXME: Why is this not \remarks?
```
jensmaurer: fine with me

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+\item Otherwise, \tcode{\exposid{PARENT-CLS}(r)}.
+\end{itemize}
+
+\pnum
+\returns
+\begin{itemize}
+\item
+  If \tcode{r} represents an unnamed bit-field $F$,
+  then \tcode{is_accessible($\tcode{r}_H$, ctx)}
+  %FIXME: comma
```
jensmaurer: yes

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+  \end{itemize}
+\item
+  a non-static data member,
+\item
+  a namesapce, or
+\item
+  a namespace alias.
+\end{itemize}
+\begin{note}
+%FIXME: begin example? The entirety of the note is an example.
```
jensmaurer: \begin{example} often comes with a codeblock, I'm fine leaving this as a note.

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+\indexlibraryglobal{enumerators_of}%
+\begin{itemdecl}
+consteval vector<info> enumerators_of(info type_enum);
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\constantwhen
+\tcode{dealias(type_enum)} represents an enumeration type
+%FIXME: comma
```
jensmaurer: yes

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+
+\begin{itemdescr}
+\pnum
+\constantwhen
+\tcode{r} represents a non-static data member,
+unnamed bit-field, or
+%FIXME: grammatically questionable;
+%the relationship itself is not a virtual base class.
+%but the base class $B$ in that relationship could be a virtual base class
+%and the derived class $D$ could be an abstract class.
```
jensmaurer: I'm not seeing a problem.  We have a "direct base class relationship" that is a pair,
and in that pair, B is constrained to have certain properties, and similar for D.

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+direct base class relationship $(D, B)$
+for which either $B$ is not a virtual base class
+or $D$ is not an abstract class.
+
+\pnum
+Let $V$ be the offset in bits from the beginning of a complete object
+%FIXME: category error.
+%parent_of(r) is not a type, it's an expression.
+%Are we trying to say "the type represented by parent_of(r)"?
```
jensmaurer: yeah.  borderline editorial, I'd say.  @jwakely ?

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+\item a non-static data member of type $T$,
+\item a data member description $(T, N, A, W, \mathit{NUA})$, or
+\item \tcode{dealias(r)} represents a type $T$,
+\end{itemize}
+then \tcode{sizeof($T$)} if $T$ is not a reference type
+and \tcode{size_of(\brk{}add_pointer(\brk{}\reflexpr{$T$}))} otherwise.
+Otherwise, \tcode{size_of(type_of(r))}.
+\begin{note}
+It is possible that while \tcode{sizeof(char)\brk{} == size_of(\reflexpr{char})}
+%FIXME: "is true"
```
jensmaurer: yes

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+%FIXME: comma
```
jensmaurer: yes

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+that \tcode{sizeof(char\&)\brk{} != size_of(\brk{}\reflexpr{char}\&)}.
+%FIXME: "is true"
```
jensmaurer: or reverse the polarity and say "is false"?

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+If \tcode{b} represents a direct base class relationship of an empty base class,
+then \tcode{size_of(b) > 0}.
+%FIXME: "is true"
```
jensmaurer: yes

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+there is a point within the evaluation context from which $T$ is not incomplete.
+
+\pnum
+\returns
+\begin{itemize}
+\item
+  If \tcode{r} represents a non-static data member
+  that is a bit-field
+  %FIXME: comma. I believe non-static data members cannot be unnamed bit-fields,
+  %so the comma would prevent the sentence from suggesting that false reality.
```
jensmaurer: I'd prefer to just reverse the two prongs of the condition.
I'm also unhappy about the "with width $W$ not clearly attaching to both prongs.

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+\rSec2[meta.reflection.extract]{Value extraction}
+
+\pnum
+The \tcode{extract} function template may be used
+to extract a value out of a reflection when its type is known.
+
+\pnum
+The following are defined for exposition only
+to aid in the specification of \tcode{extract}:
+%FIXME: does ISO like paragraphs ending in a colon?
```
jensmaurer: so-so.

I don't think a colon is needed here, so please replace with a period.

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+  \tcode{r} represents a non-static data member with type $X$,
+  that is not a bit-field,
+  that is a direct member of class \tcode{C},
+  \tcode{T} and \tcode{C} are similar types\iref{conv.qual}, and
+  \tcode{T} is \tcode{is_convertible_v<\brk{}X C::*, T>} is \tcode{true};
+\item
+  \tcode{r} represents an implicit object member function
+  with type \tcode{F} or \tcode{F noexcept}
+  %FIXME: is this even valid C++?
+  %I don't think you can just slap noexcept after an identifier ...
```
jensmaurer: I don't think you can incrementally amend function types like that.   This needs a rephrase, such as "given F is type "pointer to function", then F1 is type "pointer to noexcept function", with all the parameter types and other decoration dutifully duplicated.   There ought to be a better way to express that.  LWG issue, I'd say.

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+  that is a direct member of a class \tcode{C}
+  %FIXME: comma
```
jensmaurer: yes

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+\begin{itemdecl}
+template<class T>
+  consteval T extract(info r);
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\effects
+%FIXME: Why do we normally make a separate paragraph for Let,
+%but here we cram it into effects?
```
jensmaurer: Sometimes, we need the thing introduced by "Let" also for \remarks, so we need a global Let.
That said, I'm happy with moving this to before the \effects.

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+Let \tcode{U} be \tcode{remove_cv_t<T>}.
+Equivalent to:
+\begin{codeblock}
+%FIXME: this is inconsistent with reflect_constant,
+%which uses if constexpr,
+%and presumably, we could do the same here, with the same semantics.
```
jensmaurer: Is that a problem?  Do we maybe instantiate one of the helper functions with a bad T?

I agree "if constexpr" would be better. <del>LWG issue.</del>

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+
+template<@\libconcept{reflection_range}@ R = initializer_list<info>>
+consteval bool @\libglobal{can_substitute}@(info templ, R&& arguments);
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\constantwhen
+\tcode{templ} represents a template
+%FIXME: comma. We're not trying to say that temp represents a template and something else.
```
jensmaurer: yes

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+template<T P> struct TCls;
+\end{codeblock}
+the \grammarterm{template-id} \tcode{TCls<$V$>} would be valid.
+
+\pnum
+\returns
+\tcode{template_arguments_of(\reflexpr{TCls<$V$>})[0]}.
+\begin{note}
+This is a reflection of an object for class types
+%FIXME: comma.
```
jensmaurer: yes

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+  \begin{note}
+  The name corresponds to the spelling of an identifier~token
+  after phase~6 of translation\iref{lex.phases}.
+  Lexical constructs like
+  \grammarterm{universal-character-name}s\iref{lex.universal.char} are not processed
+  and will cause evaluation to fail.
+  %FIXME: what does it mean for evaluation to fail? Constant When violation?
+  For example, \tcode{R"(\textbackslash u03B1)"} is an invalid identifier
+  %FIXME: this shouldn't really be a mathematical alpha,
+  %but we'd need \textalpha or something from {textgreek} to make it work, I think ...
```
jensmaurer: I feel the pain, but I don't think it's worth the bother.

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+  \begin{itemize}
+  \item
+    \tcode{is_integral_type(type) || is_enumeration_type(type)} is \tcode{true},
+  \item
+    \tcode{options.alignment} does not contain a value,
+  \item
+    \tcode{options.no_unique_address} is \tcode{false}, and
+  \item
+    if $V$ equals \tcode{0}
+    %FIXME: comma
```
jensmaurer: yes

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+
+\pnum
+\returns
+\tcode{class_type}.
+\end{itemdescr}
+
+\rSec2[meta.reflection.traits]{Reflection type traits}
+
+%FIXME: There is no such defined thing as a "consteval function",
+%but there is a \tcode{consteval} function!
```
jensmaurer: I'm fine with adding code font here.

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+\pnum
+Subclause~\ref{meta.reflection.traits} specifies consteval functions to
+query the properties of types\iref{meta.unary},
+query the relationships between types\iref{meta.rel}, or
+transform types\iref{meta.trans}
+%FIXME: better to say "during program translation".
```
jensmaurer: yes

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+%FIXME: comma woudl be good here because "at compile time" applies to ALL items.
+at compile time.
```
jensmaurer: Not needed for this fluffy intro sentence, I think.

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+
+\pnum
+Ech function or function template declared above has the following behavior
+based on the signature and return type of that function or function template.
+\begin{note}
+The associated class template need not be instantiated.
+\end{note}
+
+%FIXME: This table looks terrible because the left column is too narrow
+%to fit the declarations without wrapping.
```
jensmaurer: @tkoeppe ?

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+\continuedcaption\\
+\topline
+\lhdr{Signature and Return Type} &   \rhdr{\Fundescx{Returns}} \\ \capsep
+\endhead
+
+\indexlibraryglobal{remove_const}%
+%FIXME: we have not been using meta:: in any of the item declarations FOR 3000 LINES of TeX
+%but NOW we suddenly changed our minds?
+%Maybe avoid meta:: for consistency, or use it everywhere above,
+%like we do for ranges:: declarations ...
```
jensmaurer: For this particular table, we highlight the mapping between meta:: functions to regular type traits.  I think having the `meta::` prefix helps.

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+\\ \topline
+\lhdr{Signature and Return Type} &   \rhdr{\Fundescx{Returns}} \\ \capsep
+\endfirsthead
+\continuedcaption\\
+\topline
+\lhdr{Signature and Return Type} &   \rhdr{\Fundescx{Returns}} \\ \capsep
+\endhead
+
+\indexlibraryglobal{remove_const}%
+%FIXME: we have not been using meta:: in any of the item declarations FOR 3000 LINES of TeX
```
jensmaurer: 3000 -> 5800

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+%but NOW we suddenly changed our minds?
+%Maybe avoid meta:: for consistency, or use it everywhere above,
+%like we do for ranges:: declarations ...
+\tcode{bool meta::\placeholder{UNARY}(info type);\br
+bool meta::\placeholder{UNARY}_type(info type);}  &
+%FIXME: Why are we suddenly using std:: in here?
+%I don't think we usually do this for type traits ...
```
jensmaurer: Except when it's needed for ambiguity control.  Same as with meta::, I think it helps here.  If you feel strongly, open an LWG issue.

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+bool meta::\placeholder{BINARY}_type(info type);}  &
+\tcode{std::\placeholder{BINARY}_v<$\mathit{T1}$, $\mathit{T2}$>},
+where $\mathit{T1}$ and $\mathit{T2}$ are the types or type aliases
+represented by \tcode{t1} and \tcode{t2}, respectively
+\\  \rowsep
+
+\tcode{template<reflection_range R>\br
+  bool meta::\placeholder{VARIADIC}_type(info type, R\&\& args);}  &
+\tcode{std::\placeholder{VARIADIC}_v<$T$, $U$...>}
+%FIXME: commma
```
jensmaurer: yes

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+A reflection representing the type denoted by
+\tcode{std::\placeholder{UNARY}_t<\brk{}$T$>},
+where $T$ is the type or type alias represented by \tcode{type}
+\\  \rowsep
+
+\tcode{template<reflection_range R>\br
+  info meta::\placeholder{VARIADIC}(R\&\& args);}  &
+A reflection representing the type denoted by
+\tcode{std::\placeholder{VARIADIC}_t<$T$...>}
+%FIXME: commma
```
jensmaurer: yes

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+where \tcode{$T$...} is the pack of types or type aliases
+whose elements are represented by the corresponding elements of \tcode{args}
+\\  \rowsep
+
+\tcode{template<reflection_range R>\br
+  info meta::\placeholder{VARIADIC}(info type, R\&\& args);}  &
+A reflection representing the type denoted by
+\tcode{std::\placeholder{VARIADIC}_t<$T$, $U$...>}
+%FIXME: commma
```
jensmaurer: yes

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+
+\pnum
+\begin{note}
+For those functions or function templates which return a reflection,
+that reflection always represents a type and never a type alias.
+\end{note}
+
+\pnum
+\begin{note}
+%FIXME: this is obviously an example, not a note
```
jensmaurer: As above, it's not a codeblock, so having it as a note works for me.

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+\indexlibraryglobal{tuple_size}%
+\begin{itemdecl}
+consteval size_t tuple_size(info type);
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\returns
+%FIXME: why are we suddenly starting to use math symbols for T
+%when the previous paragraph did not?
```
jensmaurer: We should not.  Types in particular are always code font. Please fix.

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+\tcode{tuple_size_v<$T$>}
+%FIXME: comma.
```
jensmaurer: yes

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+consteval info tuple_element(size_t index, info type);
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\returns
+A reflection representing
+the type denoted by \tcode{tuple_element_t<$I$, $T$>},
+where $T$ is th type represented by \tcode{dealias(type)}
+%FIXME: comma.
```
jensmaurer: yes

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+\indexlibraryglobal{variant_size}%
+\begin{itemdecl}
+consteval size_t variant_size(info type);
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\returns
+\tcode{variant_size_v<$T$>}
+%FIXME: comma
```
jensmaurer: yes

source/meta.tex
```diff
@@ -2554,6 +2570,3461 @@
...
+\begin{itemdecl}
+consteval info variant_alternative(size_t index, info type);
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\returns
+A reflection representing the type denoted by
+\tcode{variant_alternative_t<\brk{}$I$, $T$>}
+%FIXME: comma
```
jensmaurer: yes

source/meta.tex
brevzin: The wording does say `if constexpr`: https://wiki.edg.com/pub/Wg21sofia2025/StrawPolls/p2996r13.html#pnum_742

source/overloading.tex
tkoeppe: Done.

source/meta.tex
jensmaurer: Indeed.  This is just a bad application of the incoming paper.  Strike that LWG issue; this should be turned into "if constexpr" right away.

eisenwave:
I'll resolve those comments today. @jensmaurer when you said that types always use code font, does that mean that any use of e.g. `some_trait_t<$T$>` has to go, basically? The PR contains quite a lot of those currently.

The paper makes these italic, so I assume that's `\exposid{T}` then?

Also, does every FIXME require its own fixup commit? If yes, then there are like 20 incoming commits that each add one comma ...

jensmaurer:
> Also, does every FIXME require its own fixup commit? If yes, then there are like 20 incoming commits that each add one comma ...

No, one big commit is good enough.  This will be squashed anyway.

For T, in general, we use T and U as type variables in upright code font, even if they are sort-of meta-variables.  (If something could be a valid C++ snippet, we use upright code font.)  Since an expression cannot always be written \tcode{E} (value category and bit-field-ness are not expressible as "E"), we use $E$.  In contrast, all properties of a type (now: differentiation vs. type alias makes that claim dubious) are present in a typedef that happens to be called T, so we can write \tcode{T}.

I suggest you leave $T$ for now (this can notionally express differentiation between type and type alias).

source/meta.tex
jwakely: Yes, "the type represented by `parent_of(r)`" seems correct and an editorial fix.

source/meta.tex
jwakely: Given that none of the next three paragraphs begin with any _Element_: either, I'm not too concerned about this, but moving it to a _Remarks_: at the end is fine.

jwakely: There's no precedent for using _Remarks_: for this because all other cases of addressable functions in the standard and in P2996 are specified by introductory wording at the start of a subclause, rather than in a single function's \itemdescr. This can be the first.

source/meta.tex
eisenwave: I was just generally puzzled by the fact that in some reflection subclauses, basically everything gets its own paragraph, including notes. It's not super consistent, but we can figure that out after motions.

source/meta.tex
eisenwave: I'll open an LWG issue about it then. I'll keep this comment unresolved until I've done that, but I'm getting rid of the FIXME.

source/meta.tex
eisenwave: I've put both potential candidates for `\remarks` within this `\itemdescr` into a single `\remarks` paragraph.

source/meta.tex
eisenwave: Alright, I'm keeping it as is and getting rid of any FIXMEs that nag use of `meta::` or lack thereof.

source/meta.tex
eisenwave: It may be better to do this as a bulk change after motions. There are quite a lot of reflections papers that would all need individual fixups otherwise, and it's making it more complicated than it needs to be.

source/meta.tex
eisenwave: I like the idea of changing the order in the condition. I've fixed it up that way.

source/meta.tex
eisenwave: I think I goofed at the time and was using an older revision without `if constexpr` for a short time when transcribing the paper until I've noticed by mistake. I guess this one flew under the radar.

There is now `if constexpr` in the main commit. Took some history rewriting for this PR though.

eisenwave:
Alright, the PR is now in a state where all the little comma FIXME and other minor things are resolved. I'd hold off on major changes like renaming to `meta.syn` and `meta.refl` due to all the papers based on reflection being affected by this change.

source/meta.tex
tkoeppe: Done.

source/compatibility.tex
tkoeppe: Done.

tkoeppe (approved):

PR #8011, opened 2025-06-24
P3348R4 C++26 should refer to C23 not C17

### LINKED ISSUE #7940: [2025-06 LWG Motion 3] P3348R4 C++26 should refer to C23 not C17
https://wiki.edg.com/pub/Wg21sofia2025/StrawPolls/p3348r4.pdf

jensmaurer:
Avoid lone subclause [depr.ctime] in Annex D.

Fixes #7940.
Fixes cplusplus/papers#2020

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -273,6 +273,18 @@ Valid \CppXXIII{} code that defines \tcode{replaceable_if_eligible} or
 \tcode{trivially_relocatable_if_eligible} as macros is invalid
 in this revision of \Cpp{}.
 
+\rSec2[diff.cpp23.mem]{\ref{mem}: memory management library}
+
+\diffref{c.malloc}
+\change
+Calling \tcode{realloc} with a non-null pointer and zero size
+has erroneous behavior.
+\rationale
+The C standard library does not define this behavior.
+\effect
+Valid \CppXXIII{} code that calls \tcode{realloc}
+with a non-null pointer and a size of zero is erroneous and may change behavior.
+
 \rSec2[diff.cpp23.containers]{\ref{containers}: containers library}
 
 \diffref{span.overview}
@@ -3674,12 +3686,12 @@ standard library.
 
 \rSec2[diff.mods.to.definitions]{Modifications to definitions}
 
-\rSec3[diff.char16]{Types \tcode{char16_t} and \tcode{char32_t}}
+\rSec3[diff.char16]{Types \tcode{char8_t}, \tcode{char16_t}, and \tcode{char32_t}}
 
 \pnum
-The types \keyword{char16_t} and \keyword{char32_t}
+The types \keyword{char8_t}, \keyword{char16_t}, and \keyword{char32_t}
 are distinct types rather than typedefs to existing integral types.
-The tokens \keyword{char16_t} and \keyword{char32_t}
+The tokens \keyword{char8_t}, \keyword{char16_t}, and \keyword{char32_t}
 are keywords in \Cpp{}\iref{lex.key}.
 They do not appear as macro or type names defined in
 \libheaderref{cuchar}.
@@ -3695,14 +3707,6 @@ It does not appear as a macro or type name defined in any of
 \libheaderref{cstddef},
 \libheaderref{cstdlib},
 or \libheaderref{cwchar}.
-
-\rSec3[diff.header.assert.h]{Header \tcode{<assert.h>}}
-\indexhdr{assert.h}%
-
-\pnum
-The token \keyword{static_assert} is a keyword in \Cpp{}.
-It does not appear as a macro name defined
-in \libheaderref{cassert}.
 
 \rSec3[diff.header.iso646.h]{Header \tcode{<iso646.h>}}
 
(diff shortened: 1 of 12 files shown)


source/text.tex
```diff
@@ -13236,6 +13238,7 @@
 #define @\libmacro{WCHAR_MAX}@ @\seebelow@                                             // freestanding
 #define @\libmacro{WCHAR_MIN}@ @\seebelow@                                             // freestanding
 #define @\libmacro{WEOF}@ @\seebelow@                                                  // freestanding
+#define @\libmacro{WCHAR_WIDTH}@ @\seebelow@                                           // freestanding
```
eisenwave, line 13241:
```suggestion
#define @\libmacro{WCHAR_WIDTH}@ @\seebelow@                                           // freestanding
#define @\libmacro{WEOF}@ @\seebelow@                                                  // freestanding
```
The paper wants a different order for these macros.

source/time.tex
```diff
@@ -11658,7 +11667,7 @@
 \end{footnote}
 
 \pnum
-The functions \tcode{asctime}, \tcode{ctime}, \tcode{gmtime}, and
+The functions \tcode{gmtime} and
 \tcode{localtime} are not required to avoid data
 races\iref{res.on.data.races}.
```
eisenwave, line 11673: Right below, there is a missing change (GitHub won't let me comment on that line):
```diff
-\xrefc{7.27}
+\xrefc{7.29}
```

source/future.tex
```diff
@@ -220,18 +220,33 @@
...
 \end{codeblock}
+The header defines these macros the same as
+the C standard library header \libheader{float.h}.
+
+\xrefc{5.2.4.2.2, 7.33.5}
+
+\pnum
+In addition to being available via inclusion of the \libheader{cfloat} header,
+the macros \tcode{INFINITY} and \tcode{NAN} are
+available when \libheaderref{cmath} is included.
```
eisenwave, line 238: Out of curiosity, why is it `\libheader{cfloat}` but `\libheaderref{cmath}`?

source/future.tex
```diff
@@ -762,6 +777,21 @@
 Equivalent to: \tcode{return visit(std::forward<Visitor>(vis), arg.value);}
 \end{itemdescr}
 
+\rSec1[depr.ctime]{Deprecated time formatting}
```
eisenwave, line 780: The paper titles this subclause

> Header `<ctime>` synopsis

I assume the change is intentional?

Eisenwave (dismissed):

source/text.tex
jwakely: In [cstdint.syn] WCHAR_WIDTH comes immediately after WCHAR_MAX, because there's no WEOF in [cstdint.syn]. That's why I put it before WEOF here. It's not important though, certainly editorial.

source/future.tex
jwakely: No the paper has "D.? Deprecated [depr.ctime]" and then "D.?.? Header <ctime> synopsis [depr.ctime.syn]". As noted in the commit msg, the D.?.? subclause has been dissolved, so there's only D.? here, so strictly speaking the paper titled it "Deprecated".

"Deprecated" is a terrible name (I think I meant to write something there and apparently forgot).  And "Header <ctime> synopsis" only made sense for an rSec2 title, not the rSec1 title. So this change seems necessary.

source/memory.tex
```diff
@@ -2045,8 +2043,35 @@
...
 return a pointer to a suitable created object.
-In the case of \tcode{calloc} and \tcode{realloc},
-the objects are created before the storage is zeroed or copied, respectively.
+In the case of \tcode{calloc},
+the objects are created before the storage is zeroed.
+\end{itemdescr}
+
+\indexlibraryglobal{realloc}%
+\begin{itemdecl}
+void* realloc(void* ptr, size_t size)
```
jwakely:
```suggestion
void* realloc(void* ptr, size_t size);
```

source/text.tex
```diff
@@ -13279,12 +13284,11 @@
 
 \pnum
 The contents and meaning of the header \libheaderdef{cuchar}
-are the same as the C standard library header
-\libheader{uchar.h}, except that it
-declares the additional \tcode{mbrtoc8} and \tcode{c8rtomb} functions
-and does not declare types \keyword{char16_t} nor \keyword{char32_t}.
+are the same as the C standard library header \libheader{uchar.h},
+except that it does not declare types
```
jwakely, line 13288:
```suggestion
except that it does not declare the types
```

If you want to keep the "the" in [cwchar.syn] then should we add it here?

source/compatibility.tex
```diff
@@ -240,6 +240,18 @@
 \tcode{trivially_relocatable_if_eligible} as macros is invalid
 in this revision of \Cpp{}.
 
+\rSec2[diff.cpp23.mem]{\ref{mem}: memory management library}
```
jwakely, line 276: The paper has "Memory" with uppercase M but it seems that is wrong and it should indeed be lowercase. But "C.1.2 Clause 5: Lexical conventions" is wrong in the current WP.

jwakely (changes requested):

source/time.tex
tkoeppe: Done, thanks!

source/text.tex
tkoeppe: Doesn't [cwchar.syn] say "does not declare _a_ type `wchar_t`"?

source/future.tex
tkoeppe: This seems good.

source/future.tex
tkoeppe: The former already has a cross reference just earlier, but the latter does not and can have one here.

source/compatibility.tex
tkoeppe: I'll fix that, thanks!

source/memory.tex
tkoeppe: Done.

source/text.tex
jwakely: Why yes, yes it does.

So we have every variation, "does not declare types", "does not declare a type", and "does not declare the type" (twice)

jwakely (commented):
LGTM

source/text.tex
tkoeppe: I mean, all of those are grammaticaly correct... but does it make sense to not declare a definite thing? How can it be definite if it doesn't exist?

Existential musings welcome in a follow-up PR!

jwakely: Yeah I briefly had the same thought, but those types _do_ still exist in C++, they're just not declared in those headers. So "the type `wchar_t`" and "the types `char8_t`, etc." is correct.

tkoeppe: Good point, agreed!

PR #8014, opened 2025-06-25
[Motions 2025 06 lwg 1] P3742R0 C++ Standard Library Issues

### LINKED ISSUE #2371: P3742 R0 C++ Standard Library Ready Issues to be moved in Sofia, Jun. 2025
[P3742R0](https://wg21.link/p3742r0) C++ Standard Library Ready Issues to be moved in Sofia, Jun. 2025 (Jonathan Wakely)

### LINKED ISSUE #7938: [2025-06 LWG Motion 1] P3742R0 C++ Standard Library Issues to be moved in Sofia, Jun. 2025

### LINKED ISSUE #7939: [2025-06 LWG Motion 2] P2988R12 std::optional<‍T&‍>

burblebee:
Fixes #7938

Fixes cplusplus/papers#2371

### DIFF
--- a/source/exec.tex
+++ b/source/exec.tex
@@ -3612,8 +3612,11 @@ If
 is \tcode{false},
 then the expression \tcode{\exposid{let-cpo}.transform_env(sndr, env)}
 is ill-formed.
-Otherwise, it is equal to
-\tcode{\exposid{JOIN-ENV}(\exposid{let-env}(sndr), \exposid{FWD-ENV}(env))}.
+Otherwise, it is equal to:
+\begin{codeblock}
+auto& [_, _, child] = sndr;
+return @\exposid{JOIN-ENV}@(@\exposid{let-env}@(child), @\exposid{FWD-ENV}@(env));
+\end{codeblock}
 
 \pnum
 Let the subexpression \tcode{out_sndr} denote
--- a/source/exec.tex
+++ b/source/exec.tex
@@ -3567,7 +3567,7 @@ the type \tcode{\exposid{call-result-t}<F, decay_t<Args>\&...>}.
 Then \tcode{ops2_variant_t} denotes
 the type
 \begin{codeblock}
-variant<monostate, connect_result_t<@\exposid{as-sndr2}@<LetSigs>, @\exposid{receiver2}@<Rcvr, Env>>...>
+variant<monostate, connect_result_t<@\exposid{as-sndr2}@<LetSigs>, @\exposid{receiver2}@<Rcvr, env_t>>...>
 \end{codeblock}
 except with duplicate types removed.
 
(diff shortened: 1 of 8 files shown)


eisenwave (changes requested):
There are some `\linebreak`s introduced which should be `\brk{}`; otherwise I couldn't find anything wrong.

There are also some index entries added for exposition-only stuff, and I'm not sure if that's usually done.

burblebee:
> There are some `\linebreak`s introduced which should be `\brk{}`; otherwise I couldn't find anything wrong.

As I understand it, `\brk` (aka `\discretionary`) is basically `\-`, and I use `\-` where ever possible. So the fact that I resorted to `\linebreak` most likely meant that `\brk` wouldn't work in those cases. Do you know of any cases where using `\brk` would have made a difference?

eisenwave:
> As I understand it, `\brk` (aka `\discretionary`) is basically `\-`, and I use `\-` where ever possible.

Yeah, with the exception that it won't leave a hyphen at line wraps (unless `\discretionary` is configured to do so, which it isn't for `\brk`).

> So the fact that I resorted to `\linebreak` most likely meant that `\brk` wouldn't work in those cases. Do you know of any cases where using `\brk` would have made a difference?

Can you just replace any use of `\linebreak ` with `\brk{}` and see if that fixes it? I have drafted up many thousands of TeX LOCs for these motions, and I yet have to encounter a case where `\linebreak` works but `\brk` doesn't.

`\-` sometimes doesn't work because it's not in the right spot for paragraph justification to feel incentivized enough to break the line there, but if you throw in enough `\-`s in the midlde of words and `\brk{}` after punctuation etc., then it always seems to work. It's just a matter of drowning the `\tcode`s in enough line break opportunities until it works.

Jens also replaced a bunch of `\linebreak`s with `\brk`s and that worked just fine in any of his PRs.

burblebee:
> Can you just replace any use of `\linebreak ` with `\brk{}` and see if that fixes it? I have drafted up many thousands of TeX LOCs for these motions, and I yet have to encounter a case where `\linebreak` works but `\brk` doesn't.

I will do this next time. Since this is a multi-commit motion, making those changes will add more work for Thomas. That said, I have come across many cases where \brk did not work and I had to resort to \linebreak (perhaps it has happened so often that I use \linebreak more often than I should - ok, will always check in the future).

eisenwave:
There is a way to squash the `\linebreak` -> `\brk` changes into each respective commit, but it's a bit tedious to do.

I am not convinced `\linebreak` is ever totally unavoidable, but I can look into blanket-replacing the 100 or so occurrences after motions.

The rest of the PR looks good.

eisenwave (approved):

source/exec.tex
```diff
@@ -2577,6 +2577,9 @@
 \item
 Otherwise, \tcode{\exposid{connect-awaitable}(new_sndr, rcvr)}.
 \end{itemize}
+%%FIXME: This sentence should be joined with the previous sentences;
```
tkoeppe, line 2580: Yes, this looks very odd, but that's what it says in the paper. We should follow this up and reorganize this.

PR #8020, opened 2025-06-26
[2025-06 LWG 19] P3481R5 std::execution::bulk() issues

### LINKED ISSUE #7956: [2025-06 LWG Motion 19] P3481R5 std::execution::bulk() issues
https://wiki.edg.com/pub/Wg21sofia2025/StrawPolls/P3481R5.html

burblebee:
Fixes #7956.

### DIFF
--- a/source/exec.tex
+++ b/source/exec.tex
@@ -631,6 +631,8 @@ namespace std::execution {
   struct @\libglobal{let_error_t}@ { @\unspec@ };
   struct @\libglobal{let_stopped_t}@ { @\unspec@ };
   struct @\libglobal{bulk_t}@ { @\unspec@ };
+  struct @\libglobal{bulk_chunked_t}@ { @\unspec@ };
+  struct @\libglobal{bulk_unchunked_t}@ { @\unspec@ };
   struct @\libglobal{split_t}@ { @\unspec@ };
   struct @\libglobal{when_all_t}@ { @\unspec@ };
   struct @\libglobal{when_all_with_variant_t}@ { @\unspec@ };
@@ -651,6 +653,8 @@ namespace std::execution {
   inline constexpr let_error_t @\libglobal{let_error}@{};
   inline constexpr let_stopped_t @\libglobal{let_stopped}@{};
   inline constexpr bulk_t @\libglobal{bulk}@{};
+  inline constexpr bulk_chunked_t @\libglobal{bulk_chunked}@{};
+  inline constexpr bulk_unchunked_t @\libglobal{bulk_unchunked}@{};
   inline constexpr split_t @\libglobal{split}@{};
   inline constexpr when_all_t @\libglobal{when_all}@{};
   inline constexpr when_all_with_variant_t @\libglobal{when_all_with_variant}@{};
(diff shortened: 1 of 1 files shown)


source/exec.tex
```diff
@@ -3608,52 +3612,133 @@
...
+\tcode{bulk}, \tcode{bulk_chunked}, and \tcode{bulk_unchunked}
+run a task repeatedly for every index in an index space.
 
-The name \tcode{bulk} denotes a pipeable sender adaptor object.
-For subexpressions \tcode{sndr}, \tcode{shape}, and \tcode{f},
-let \tcode{Shape} be \tcode{decltype(auto(shape))}.
+\pnum
+The names \tcode{bulk}, \tcode{bulk_chunked}, and \tcode{bulk_unchunked}
+denote pipeable sender adaptor objects.
+Let \placeholder{bulk-algo} be either
```
eisenwave:
```suggestion
Let \exposid{bulk-algo} be either
```
`\placeholder` is for placeholders like `N` in the middle of an identifier `intN_t`.

source/exec.tex
```diff
@@ -3608,52 +3612,133 @@
...
 \item
+\tcode{is_execution_policy_v<Policy>} is \tcode{false}, or
+\item
 \tcode{Shape} does not satisfy \libconcept{integral}, or
 \item
-\tcode{decltype((f))} does not satisfy \exposconcept{movable-value},
+\tcode{Func} does not model \libconcept{copy_constructible},
 \end{itemize}
-\tcode{bulk(sndr, shape, f)} is ill-formed.
+\tcode{\placeholder{bulk-algo}(sndr, policy, shape, f)} is ill-formed.
```
eisenwave, line 3711: Same here, `\exposid`.

source/exec.tex
```diff
@@ -3608,52 +3612,133 @@
...
+\tcode{bulk}, \tcode{bulk_chunked}, or \tcode{bulk_unchunked}.
+For subexpressions \tcode{sndr}, \tcode{policy}, \tcode{shape}, and \tcode{f},
+let
+\tcode{Policy} be \tcode{remove_cvref_t<decltype(policy)>},
+\tcode{Shape} be \tcode{decltype(auto(shape))}, and
+\tcode{Func} be \tcode{decay_t<decltype((f))>}.
 If
 \begin{itemize}
 \item
 \tcode{decltype((sndr))} does not satisfy \libconcept{sender}, or
```
eisenwave, line 3703: This is not your fault, but we don't usually put `, or` after every bullet, just after the final bullet, unless it's a really long list.

Fixup commit should fix this up.

source/exec.tex
```diff
@@ -3608,52 +3612,133 @@
...
-\tcode{decltype((f))} does not satisfy \exposconcept{movable-value},
+\tcode{Func} does not model \libconcept{copy_constructible},
 \end{itemize}
-\tcode{bulk(sndr, shape, f)} is ill-formed.
+\tcode{\placeholder{bulk-algo}(sndr, policy, shape, f)} is ill-formed.
 
 \pnum
 Otherwise,
-the expression \tcode{bulk(sndr, shape, f)} is expression-equivalent to:
+the expression \tcode{\placeholder{bulk-algo}(sndr, policy, shape, f)}
```
eisenwave, line 3715: Same here, `\exposid`.

source/exec.tex
```diff
@@ -3608,52 +3612,133 @@
...
+is expression-equivalent to:
 
 \begin{codeblock}
-transform_sender(@\exposid{get-domain-early}@(sndr), @\exposid{make-sender}@(bulk, @\exposid{product-type}@{shape, f}, sndr))
+transform_sender(@\exposid{get-domain-early}@(sndr),
+                 @\exposid{make-sender}@(@\placeholder{bulk-algo}@,
```
eisenwave: Sam here, `\exposid`.

source/exec.tex
```diff
@@ -3608,52 +3612,133 @@
...
+
 \pnum
 The exposition-only class template \exposid{impls-for}\iref{exec.snd.general}
-is specialized for \tcode{bulk_t} as follows:
+is specialized for \tcode{bulk_chunked_t} as follows:
 \begin{codeblock}
 namespace std::execution {
   template<>
-  struct @\exposid{impls-for}@<bulk_t> : @\exposid{default-impls}@ {
+  struct @\exposid{impls-for}@<bulk_chunked_t> : @\exposid{default-impls}@ {
```
eisenwave, line 3755: Not your fault, but there should be two spaces indentation following `template`.

Maybe that can be fixed in the base paper, but that will be a merge conflict.

source/exec.tex
```diff
@@ -3669,32 +3754,109 @@
...
 The expression in the \grammarterm{requires-clause} of the lambda above
 is \tcode{true} if and only
 if \tcode{Tag} denotes a type other than \tcode{set_value_t} or
 if the expression \tcode{f(auto(shape), args...)} is well-formed.
 
 \pnum
 Let the subexpression \tcode{out_sndr} denote
-the result of the invocation \tcode{bulk(sndr, shape, f)} or
+the result of the invocation
+\tcode{\placeholder{bulk-algo}(sndr, policy, shape, f)} or
```
eisenwave, line 3822: Same here, `\exposid`.

source/exec.tex
```diff
@@ -3669,32 +3754,109 @@
...
 \item
-on a value completion operation,
-invokes \tcode{f(i, args...)}
-for every \tcode{i} of type \tcode{Shape} in \range{\tcode{0}}{\tcode{shape}},
-where \tcode{args} is a pack of lvalue subexpressions
-referring to the value completion result datums of the input sender, and
+If \tcode{sndr} has a successful completion, where
+\tcode{args} is a pack of lvalue subexpressions
+referring to the value completion result datums of \tcode{sndr}, or
+decayed copies of those values if they model copy_constructible, then:
```
eisenwave:
```suggestion
decayed copies of those values if they model \libconcept{copy_constructible}, then:
```

source/exec.tex
```diff
@@ -3669,32 +3754,109 @@
...
+
+  \begin{itemize}
```
eisenwave, line 3838: The indentation in this item list is weird. The correct style is (https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines):
```tex
\begin{itemize}
\item
  Some word
  in a sentence.
\item
  ...
  \begin{itemize}
  \item
    ...
  \end{itemize}
\end{itemize}
```
To be fair there are countless examples that don't follow this style, but we should at least try to follow it in new wording.

source/exec.tex
```diff
@@ -3669,32 +3754,109 @@
...
+    from \tcode{0} to \tcode{shape};
+
+    \item
+    for \tcode{bulk_unchunked},
+    invokes \tcode{f($i$, args...)} for every $i$ of type \tcode{Shape}
+    from \tcode{0} to \tcode{shape};
+
+    \recommended
+    The underlying scheduler should execute each iteration
+    on a distinct execution agent.
```
eisenwave, line 3855: The paper indicates that this should be a single nested bullet inside of the previous bullet, but ... not sure what to make of that. @jensmaurer or @tkoeppe probably know what to do about this. Maybe this is correct?

source/exec.tex
```diff
@@ -3669,32 +3754,109 @@
...
+
+    \item
+    for \tcode{bulk_chunked},
+    invokes \tcode{f($b$, $e$, args...)} zero or more times
+    with pairs of $b$ and $e$ of type \tcode{Shape}
+    in range \range{\tcode{0}}{\tcode{shape}},
```
eisenwave:
```suggestion
    in range \crange{0}{shape},
```
`\range` is for `[)`, `\crange` is for `[]`. See macros.tex.

Also, the range macros already apply `\tcode`.

source/exec.tex
```diff
@@ -3669,32 +3754,109 @@
...
+  \tcode{eptr} is an \tcode{exception_ptr} containing either:
+    \begin{itemize}
+    \item
+    an exception thrown by an invocation of \tcode{f}, or
+    \item
+    a \tcode{bad_alloc} exception if
+    the implementation fails to allocate required resources, or
+    \item
+    an exception derived from \tcode{runtime_error}.
+    \end{itemize}
```
eisenwave, line 3882:
```suggestion
  \begin{itemize}
  \item
    an exception thrown by an invocation of \tcode{f}, or
  \item
    a \tcode{bad_alloc} exception if
    the implementation fails to allocate required resources, or
  \item
    an exception derived from \tcode{runtime_error}.
  \end{itemize}
```
On another note, we have unconventional `, or` after every bullet, which should probably be fixed up.

source/exec.tex
```diff
@@ -3669,32 +3754,109 @@
...
+  before the stopped completion handler.
+  \end{itemize}
+
 \item
-propagates all completion operations sent by \tcode{sndr}.
+If \tcode{sndr} does not complete with \tcode{set_value}, then
+the completion is forwarded to \tcode{recv}.
+
+\item
+For \placeholder{bulk-algo},
```
eisenwave: Same here, `\exposid`.

source/exec.tex
```diff
@@ -3669,32 +3754,109 @@
...
+the parameter \tcode{policy} describes
+the manner in which
+the execution of the asynchronous operations corresponding to these algorithms
+may be parallelized and
+the manner in which
+%%FIXME: Should this be "apply to f"?
+they apply \tcode{f}.
```
eisenwave, line 3903: I read "apply" as synonymous with "invoke" here.

Maybe it would be better to say "invoke" for clarity. I don't think we usually talk about applications of functions.

source/exec.tex
```diff
@@ -3669,32 +3754,109 @@
...
+Permissions and requirements
+on parallel algorithm element access functions\iref{algorithms.parallel.exec}
+apply to \tcode{f}.
 \end{itemize}
 
+\pnum
+\begin{note}
+The asynchronous operation corresponding to
+\tcode{\placeholder{bulk-algo}(sndr, policy, shape, f)}
```
eisenwave, line 3912: Same here, `\exposid`.

source/exec.tex
burblebee: It is my understanding that \exposid is for exposition identifiers; \placeholder seems correct here from my reading of the paper. That said, it should be in \tcode - fixed now. Thanks!

source/exec.tex
burblebee: Please submit a separate issue or PR for this.

source/exec.tex
burblebee: We don't usually indent structs.

source/exec.tex
burblebee: fixed, thanks! I would have thought our checks would have caught that.

source/exec.tex
burblebee: There was a whole discussion about indenting itemized lists several years back. I also prefer the indenting in your suggestion, but I was overruled.

source/exec.tex
burblebee: good catch, thanks! fixed

source/exec.tex
burblebee: I hope it's OK with you that I'm resolving this and the ones like it that follow - I'll leave the 1st one open in case you want to comment.

source/exec.tex
burblebee: If your understanding is correct, invoke would be definitely be better here. Alas, fixing that is normative, hence outside the scope of this paper.

source/exec.tex
eisenwave: Yes, we do, or at least, we should. A *template-head* should have two space indentation. See [utility.syn] for example:
```cpp
  template<auto V>
    struct nontype_t {
      explicit nontype_t() = default;
    };
  template<auto V> constexpr nontype_t<V> nontype{};
```
This style is followed for larger classes too, like in [facet.ctype.special.general].

Especially around `template<>`, there seems to be a bit of inconsistent formatting, but https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines clearly states that a *template-head* is followed by two spaces indent.

eisenwave: It may also be fine to write
```cpp
template<> struct ... {
  // no indentation for struct itself
};
```
This has a lot of precedent and doesn't seem to be against the guidelines.

source/exec.tex
eisenwave: I don't think it's a normative change to replace "function application" with "function invocation", but I agree that it doesn't need to be fixed up here. The paper seems fine.

source/exec.tex
burblebee: We (the editors) discussed this and I remember that we decided not to indent for struct/classes. 

Just re-checked, and yes, I remember correctly: "... braced (class, enum, function) definitions do not indent after the template-head;"

eisenwave: Oh well, it seems like then spec guidelines are out-of-date with some of these decisions that have been made then. We really ought to add a clarification there that the rule doesn't apply to `struct`.

PR #8021, opened 2025-06-26
Add check target and provide pre-commit hook for 'make check'

mattkretz:
I tend to forget running `check-source.sh` before pushing. This might help.

### DIFF
--- a/source/Makefile
+++ b/source/Makefile
@@ -30,3 +30,12 @@ clean-examples:
 	rm -f $(EXAMPLES)
 
 examples: $(EXAMPLES)
+
+check: .check.stamp
+
+.check.stamp: ../tools/check-source.sh *.tex
+	@echo "Running tools/check-source.sh"
+	@../tools/check-source.sh
+	@touch $@
+
+.PHONY: default refresh refresh full quiet clean-figures figures clean-examples examples check
--- a/tools/gitconfig.sh
+++ b/tools/gitconfig.sh
@@ -4,13 +4,13 @@ git config diff.c++draft.xfuncname '\\rSec[0-9]+(\[.*\])\{'
 
 precommit="$(git rev-parse --git-dir)/hooks/pre-commit"
 
-test -f "$precommit" && exit
+test -f "${precommit}" && exit
 read -p "Install 'make check' pre-commit hook? [Y/n] " hook
-if [ -z "$hook" -o "$hook" = "y" -o "$hook" = "Y" ]; then
-  cat <<EOF > "$precommit"
+if [ -z "${hook}" -o "${hook}" = "y" -o "${hook}" = "Y" ]; then
+  cat <<EOF > "${precommit}"
 #!/bin/sh
 cd \$(git rev-parse --show-toplevel)/source
 make check
 EOF
-  chmod +x "$precommit"
+  chmod +x "${precommit}"
 fi
--- a/tools/gitconfig.sh
+++ b/tools/gitconfig.sh
@@ -1,3 +1,16 @@
 #!/bin/sh
 git config diff.orderFile .gitorder
 git config diff.c++draft.xfuncname '\\rSec[0-9]+(\[.*\])\{'
+
+precommit="$(git rev-parse --git-dir)/hooks/pre-commit"
+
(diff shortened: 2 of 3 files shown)


source/Makefile
```diff
@@ -30,3 +30,11 @@ clean-examples:
 	rm -f $(EXAMPLES)
 
 examples: $(EXAMPLES)
+
+.PHONY: check
```
tkoeppe: Aren't the "clean" targets above also "phony"? Can you move the .PHONE line up to the top?

tools/gitconfig.sh
```diff
@@ -1,3 +1,12 @@
...
 git config diff.c++draft.xfuncname '\\rSec[0-9]+(\[.*\])\{'
+
+precommit="$(git rev-parse --git-dir)/hooks/pre-commit"
+test -f "$precommit" && exit
+cat <<EOF > "$precommit"
+#!/bin/sh
+cd \$(git rev-parse --show-toplevel)/source
+make check
+EOF
+chmod +x "$precommit"
```
tkoeppe: What does this change do?

mattkretz: If no `pre-commit` script exists, it installs one that runs `make check` in the `source` directory before every commit. That way one can't commit unless `check-source` passes. 
Obviously, that's opt-in via calling this script first.

tkoeppe: You mean even locally? That seems a bit draconian, doesn't it? What if I just want to commit some random noise locally?

mattkretz: I at least don't want to commit noise that would obviously fail CI.
But yes, it might not be a feature for everybody. I could change the setup script to ask whether it should install the hook?

source/Makefile
mattkretz: One style is to have a `.PHONY` target above each target that needs it. Another style I've seen often is to list all of them at the end. I'll do a drive-by fix.

tools/gitconfig.sh
tkoeppe: I, on the other hand, as a major user of this repository, would like to commit whatever I like under whatever conditions I like...

Could you just install this script locally in your own checkout?

mattkretz: This is purely opt-in. And yes, I installed this locally. My assumption was that it's useful for others, though maybe not everyone.

tkoeppe: How is it an opt-in given...

> If no pre-commit script exists, it installs one that runs make check in the source directory before every commit.

That sounded like we'd always install the hook? Do you mean "auto opt-in"?

mattkretz: You have to run the `tools/gitconfig.sh` script manually to install the hook. If you then dislike the hook you do `rm .git/hooks/pre-commit`. But like I said, I can also make the script ask explicitly, because it should be recommended to run the script.

mattkretz: I just pushed an improved script that asks for confirmation.

mattkretz: Also, even with the hook installed, `git commit --no-verify` allows to commit failing noise :)

tools/gitconfig.sh
```diff
@@ -1,3 +1,16 @@
 #!/bin/sh
 git config diff.orderFile .gitorder
 git config diff.c++draft.xfuncname '\\rSec[0-9]+(\[.*\])\{'
+
+precommit="$(git rev-parse --git-dir)/hooks/pre-commit"
+
+test -f "$precommit" && exit
+read -p "Install 'make check' pre-commit hook? [Y/n] " hook
+if [ -z "$hook" -o "$hook" = "y" -o "$hook" = "Y" ]; then
```
tkoeppe: Can we use ${brace} syntax everywhere for variable references?

tools/gitconfig.sh
tkoeppe: OK, I see, I didn't realize that this is just an optional tool. Seems fine.

jensmaurer (approved):
`tools/gitconfig.sh` is a totally optional "whoever wants to use it" thing, and installing the hook seems totally in line with the exiting other contents.

PR #8025, opened 2025-06-27
P3111R8 Atomic Reduction Operations

### LINKED ISSUE #7965: [2025-06 LWG Motion 28] P3111R8 Atomic Reduction Operations
https://wiki.edg.com/pub/Wg21sofia2025/StrawPolls/p3111r8.html

jensmaurer:
Fixes #7965
Fixes cplusplus/papers#1902

### DIFF
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -2600,6 +2600,99 @@ namespace std {
     constexpr T atomic_fetch_min_explicit(atomic<T>*,                               // freestanding
                                           typename atomic<T>::value_type,
                                           memory_order) noexcept;
+
+  template<class T>
+    void atomic_store_add(volatile atomic<T>*,                                      // freestanding
+                          typename atomic<T>::difference_type) noexcept;
+  template<class T>
+    constexpr void atomic_store_add(atomic<T>*,                                     // freestanding
+                                    typename atomic<T>::difference_type) noexcept;
+  template<class T>
+    void atomic_store_add_explicit(volatile atomic<T>*,                             // freestanding
+                                   typename atomic<T>::difference_type, memory_order) noexcept;
+  template<class T>
+    constexpr void atomic_store_add_explicit(atomic<T>*,                            // freestanding
+                                             typename atomic<T>::difference_type,
+                                             memory_order) noexcept;
+  template<class T>
+    void atomic_store_sub(volatile atomic<T>*,                                      // freestanding
+                          typename atomic<T>::difference_type) noexcept;
+  template<class T>
+    constexpr void atomic_store_sub(atomic<T>*,                                     // freestanding
+                                    typename atomic<T>::difference_type) noexcept;
+  template<class T>
+    void atomic_store_sub_explicit(volatile atomic<T>*,                             // freestanding
+                                   typename atomic<T>::difference_type, memory_order) noexcept;
+  template<class T>
+    constexpr void atomic_store_sub_explicit(atomic<T>*,                            // freestanding
+                                             typename atomic<T>::difference_type,
+                                             memory_order) noexcept;
+  template<class T>
+    void atomic_store_and(volatile atomic<T>*,                                      // freestanding
+                          typename atomic<T>::value_type) noexcept;
+  template<class T>
+    constexpr void atomic_store_and(atomic<T>*,                                     // freestanding
+                                    typename atomic<T>::value_type) noexcept;
+  template<class T>
+    void atomic_store_and_explicit(volatile atomic<T>*,                             // freestanding
+                                   typename atomic<T>::value_type, memory_order) noexcept;
+  template<class T>
+    constexpr void atomic_store_and_explicit(atomic<T>*,                            // freestanding
+                                             typename atomic<T>::value_type,
+                                             memory_order) noexcept;
+  template<class T>
+    void atomic_store_or(volatile atomic<T>*,                                       // freestanding
+                         typename atomic<T>::value_type) noexcept;
+  template<class T>
+    constexpr void atomic_store_or(atomic<T>*,                                      // freestanding
+                                   typename atomic<T>::value_type) noexcept;
+  template<class T>
+    void atomic_store_or_explicit(volatile atomic<T>*,                              // freestanding
+                                  typename atomic<T>::value_type, memory_order) noexcept;
+  template<class T>
+    constexpr void atomic_store_or_explicit(atomic<T>*,                             // freestanding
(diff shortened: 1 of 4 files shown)


source/threads.tex
```diff
@@ -2972,6 +3065,18 @@
 (in the modification order) written before the write associated with
 the read-modify-write operation.
 
+\pnum
+An \defnadj{atomic}{modify-write operation} is
+an atomic read-modify-write operation
+with weaker synchronization requirements as specified in\iref{atomics.fences}.
```
eisenwave:
```suggestion
with weaker synchronization requirements as specified in~\ref{atomics.fences}.
```

source/threads.tex
```diff
@@ -3698,6 +3818,49 @@
...
+the result is as if \tcode{*ptr} and parameters
+were converted to their corresponding unsigned types,
+the computation performed on those types, and
+the result converted back to the signed type.
+\begin{note}
+There are no undefined results arising from the computation.
+\end{note}
+For \tcode{store_max} and \tcode{store_min},
+the maximum and minimum computation is performed
+as if by \tcode{max} and \tcode{min} algorithms\iref{alg.min.max}, respectively,
```
eisenwave, line 3860: Might want to say "as if by *the* `max` and `min` algorithms", but I guess, either is technically okay.

source/threads.tex
```diff
@@ -3698,6 +3818,49 @@
...
 
+\indexlibrarymember{store_add}{atomic_ref<\placeholder{integral-type}>}%
+\indexlibrarymember{store_and}{atomic_ref<\placeholder{integral-type}>}%
+\indexlibrarymember{store_max}{atomic_ref<\placeholder{integral-type}>}%
+\indexlibrarymember{store_min}{atomic_ref<\placeholder{integral-type}>}%
+\indexlibrarymember{store_or}{atomic_ref<\placeholder{integral-type}>}%
+\indexlibrarymember{store_sub}{atomic_ref<\placeholder{integral-type}>}%
+\begin{itemdecl}
+constexpr void store_@\placeholdernc{key}@(value_type operand,
+  memory_order order = memory_order::seq_cst) const noexcept;
```
eisenwave:
```suggestion
constexpr void store_@\placeholdernc{key}(value_type operand,
                         memory_order order = memory_order::seq_cst) const noexcept;
```

source/threads.tex
```diff
@@ -4740,6 +5050,35 @@
     constexpr @\placeholdernc{integral-type}@ fetch_min( @\placeholdernc{integral-type}@,
                             memory_order = memory_order::seq_cst) noexcept;
 
+    void store_add(@\placeholdernc{integral-type}@,
```
eisenwave, line 5115: Am I unfamiliar with something, or why is this using `placeholder` instead of `exposid`? According to the comments in macros.tex, `placeholder` is for placeholders in the middle of identifiers, like `N` in `intN_t`, and *`integral-type`* really just looks like an exposition-only type to me.

source/threads.tex
```diff
@@ -4878,6 +5217,70 @@
...
+Atomically replaces the value pointed to by \tcode{this}
+with the result of the computation applied to
+the value pointed to by \tcode{this} and the given \tcode{operand}.
+Memory is affected according to the value of \tcode{order}.
+These operations are atomic modify-write operations\iref{atomics.order}.
+
+\pnum
+\remarks
+Except for \tcode{store_max} and \tcode{store_min},
+for signed integer types the result is as if
```
eisenwave:
```suggestion
for signed integer types, the result is as if
```
That's a fixup though; the paper is missing this comma.

source/threads.tex
```diff
@@ -5037,6 +5473,98 @@
...
+may be different than the calling thread's floating-point environment.
+The arithmetic rules of floating-point atomic modify-write operations
+may be different from operations on
+floating-point types or atomic floating-point types.
+\begin{note}
+Tree reductions are permitted for atomic modify-write operations.
+\end{note}
+
+\pnum
+\begin{itemize}
```
eisenwave, line 5668: Missing `\remarks` here. I actually happened to be in LWG at the time we've looked at this wording and pointed out it's weird to me that we have two separate *Remarks* specifications in the place, but it was not seen as an issue.

@jwakely having a second *Remarks* here is intentional, right?

source/threads.tex
```diff
@@ -5037,6 +5473,98 @@
...
+\item
+For \tcode{store_max} and \tcode{store_min},
+the maximum and minimum computation is performed
+as if by \tcode{fmaximum_num} and \tcode{fminimum_num}, respectively,
+with the value pointed to by \tcode{this} and
+the first parameter as the arguments, except that:
+\begin{itemize}
+\item
+If both arguments are NaN,
+an unspecified NaN value replaces the value pointed to by \tcode{this.}
```
eisenwave:
```suggestion
an unspecified NaN value replaces the value pointed to by \tcode{this}.
```

source/threads.tex
```diff
@@ -5037,6 +5473,98 @@
...
+\item
+If exactly one argument is a NaN,
+either the other argument or an unspecified NaN value replaces
+the value pointed to by \tcode{this}, it is unspecified which.
```
eisenwave:
```suggestion
the value pointed to by \tcode{this};
it is unspecified which.
```
This is a fixup; the paper also uses a comma to separate something that feels like a second complete sentence.

source/threads.tex
jensmaurer: This phrasing is copied from fetch_min/max.  I'd suggest to make a separate pull request after the motions merge.

source/threads.tex
jensmaurer: This is a meta-placeholder, because we stamp out a full specialization for each integer type.  There is an implied "foreach" around this subclause. `exposid` isn't like that.

source/threads.tex
jensmaurer: The automatic checking won't allow duplicate elements.

Eisenwave (approved):
I'm fine with merging as is, even though it feels like there are some loose threads here. Nothing we couldn't figure out after motions.

source/threads.tex
jwakely: I don't think we need separate commits for adding commas.

source/threads.tex
jwakely: Then we can live with just one

tkoeppe: Shell scripts win over LWG any day.

PR #8028, opened 2025-06-27
[simd.expos.defn] Fix missing paragraph break

mattkretz:


jwakely (commented):
Do we need a separate paragraph there? Would it be better to just remove the `\pnum`?

(I'm surprised we don't get a paragraph number '8' in the middle of that sentence, did we fix something to prevent that? I think that's what used to happen in cases like this.)

jwakely (approved):
Oh all the surrounding exposition-only thingies use a leading first paragraph for the "Let ..." and then a separate paragraph for the actual definition of the thingy, so this matches its neighbours.

mattkretz:
I looked at `\def\pnum` and decided to not understand how it works :laughing:. But yes, this is for consistency and to restore what P1928 looked like.

jensmaurer (approved):
Obvious LaTeX-level bug fix.

PR #8034, opened 2025-07-01
[ranges.*][range.*][coro.generator] Add missing library indexes. Fixes #8015.

### LINKED ISSUE #8015: [range.to.input.view] Missing library references
Library indexes are missing for class `std::ranges::to_input_view` and its members (other than `reserve_hint`). Calls to `\indexlibrarymember` and `\indexlibraryglobal` (or `\libglobal`) should be added. Other classes in `std::ranges` should be checked for missing indexes as well.

burblebee:
Fixes #8015.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -3504,6 +3504,7 @@ friend constexpr bool operator==(const @\exposid{iterator}@& x, const @\exposid{
 Equivalent to: \tcode{return x.\exposid{value_} == y.\exposid{bound_};}
 \end{itemdescr}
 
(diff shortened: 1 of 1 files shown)


burblebee:
FYI - the build failed due to bad mac system setup. This keeps happening. Would it make sense to remove macs from our build checks?

tkoeppe:
I just restarted the action and it seems to be working now. The Mac runner is sometimes a bit fickle like that, but just restarting usually seems to work.

tkoeppe:
Thanks a lot!

PR #8038, opened 2025-07-03
[range.join.iterator] Remove spurious return

hewillk:
I suspect this is editorial rather than LWG, since the return type of `ranges::iter_swap` is explicitly defined as `void`.


jwakely (approved):

jwakely:
This was present in [P0896] so is not a misapplication of the approved wording. But I agree it's redundant and should be removed.

jensmaurer (approved):

PR #8054, opened 2025-07-14
removing hyphenation from ly adverbs

lorihughes:
Change: remove hyphenation on -ly adverbs in the text (_only_ the text) of the Standard. 

Reason: Most common writing style guides omit hyphenation for -ly adverbs; i.e., an -ly adverb + adjective combination does not create a compound adjective. 

Chicago Manual of Style, 17th ed., 5.93: "(3) A two-word phrasal adjective that begins with an adverb ending in -ly is not hyphenated {a sharply worded reprimand}" 
Chicago Manual of Style, 17th ed., 7.85: "With the exception of proper nouns (such as United States) and compounds formed by an adverb ending in ly plus an adjective, it is never incorrect to hyphenate adjectival compounds before a noun." 

New Oxford Style Manual [Oxford Univ Press, 2016], 3.3.3: "Do not hyphenate adjectival compounds where the first element is an adverb ending in -ly"

New York Times Manual of Style and Usage, 5th ed., hyphen: "Never use a hyphen after an adverb ending in ly"

AP Stylebook, 55th ed., hyphen: "No hyphen is needed to link a two-word phrase that includes the adverb very and all adverbs ending in -ly"

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -6225,11 +6225,11 @@ namespace std {
 \pnum
 \indextext{\idxcode{array}!initialization}%
 \indextext{requirements!container}%
-An \tcode{array} relies on the implicitly-declared special
+An \tcode{array} relies on the implicitly declared special
 member functions\iref{class.default.ctor,class.dtor,class.copy.ctor} to
 conform to the container requirements table in~\ref{container.requirements}.
 In addition to the requirements specified in the container requirements table,
-the implicitly-declared move constructor and move assignment operator for \tcode{array}
+the implicitly declared move constructor and move assignment operator for \tcode{array}
 require that \tcode{T} be \oldconcept{MoveConstructible} or \oldconcept{MoveAssignable},
 respectively.
 
@@ -6996,7 +6996,7 @@ Modifying any list requires access to the element preceding the first element
 of interest, but in a \tcode{forward_list} there is no constant-time way to access a
 preceding element.
 For this reason, \tcode{erase_after} and \tcode{splice_after}
-take fully-open ranges, not semi-open ranges.
+take fully open ranges, not semi-open ranges.
 \end{note}
 
 \pnum
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -256,7 +256,7 @@ that takes separate iterator and sentinel arguments.
 
 \pnum
 The well-formedness and behavior of a call to an algorithm with
-an explicitly-specified template argument list
+an explicitly specified template argument list
 is unspecified, except where explicitly stated otherwise.
 \begin{note}
 Consequently, an implementation can declare an algorithm with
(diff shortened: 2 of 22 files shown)


jwakely:
Some of these are fine, but some are technical definitions (incompletely defined object, potentially-evaluated) and changes those doesn't seem editorial to me.

lorihughes:
How do "potentially-evaluated" and "incompletely-defined object" *technically* differ from "potentially evaluated" and "incompletely defined object"?

jwakely:
They are terms that have been invented by the Core Working Group and should probably not be changed by the editor without at least checking with CWG. It's not _what_ they mean that matters, it's the fact that they are defined terms which have been chosen specifically. It's not editorial to change them.

jwakely:
I'm not saying they shouldn't be fixed, only that I think a slightly different process should be used to fix them (even if it's just asking CWG to approve the change before merging this).

source/containers.tex
```diff
@@ -6996,7 +6996,7 @@
 of interest, but in a \tcode{forward_list} there is no constant-time way to access a
 preceding element.
 For this reason, \tcode{erase_after} and \tcode{splice_after}
-take fully-open ranges, not semi-open ranges.
```
tkoeppe: This is an interesting case: I understand that, in isolation, an adverb qualifies an adjective. But here we're drawing from a coherent family of terms like "half-open", "semi-open", "closed-open", "open-open", "fully-open", "fully-closed" (some of which are synonyms), where there isn't always an adverb as the first part, but it's still useful to have a common style to show that these are all members of the same technical domain, and we don't mean "fully open" in some other, broader sense. ("You've not been fully open about your use of fully-open ranges")

(Not to be confused with "clopen", which is both open and closed at once, and [something different](https://en.wikipedia.org/wiki/Clopen_set) :-) )

lorihughes: In "You've not been fully open about your use of fully-open ranges," the function of fully is still the same in both uses: It's an adverb describing an adjective. With "half-open," you have a compound adjective, but -ly adverbs do not form compound adjectives, even when our terminology would be nicely consistent and tidy if they did.

AlisdairM:
> They are terms that have been invented by the Core Working Group and should probably not be changed by the editor without at least checking with CWG. It's not _what_ they mean that matters, it's the fact that they are defined terms which have been chosen specifically. It's not editorial to change them.

I note the following terms that are explicitly defined as hyphenated terms of power that should probably be left aside for a first patch, and addressed later in conjunction with Core:

\defn{potentially-overlapping subobject}
\defnadj{incompletely-defined}{object type}
\defnadjx{potentially-evaluated}{subexpressions}{subexpression}
\defnx{explicitly-defaulted}{definition!function!explicitly-defaulted}
\defnx{potentially-throwing exception specification}
\defnx{potentially-throwing}{potentially-throwing!expression}
\defn{potentially-constant}
\definition{conditionally-supported}{defns.cond.supp}
\defn{equivalently-valued} pointer values
\defnx{immediately-declared constraint}{constraint!immediately-declared}
\defn{possibly-reclaimable}

I fear that it would be a lot of work to filter all of those cases and would greatly reduce the scope of this PR, but not eliminate it entirely.  Hopefully applying such a filter would leave the subset of changes that can be editorially acted upon quickly.

AlisdairM:
This PR would also go a long way to resolving #440.

PR #8043, opened 2025-07-10
P3491R3 `define_static_{string,object,array}`

eisenwave:
Closes #7933.
Also closes https://github.com/cplusplus/papers/issues/2158.

PR #8063, opened 2025-07-15
[meta.reflection.array] Use "character type" in Mandates

eisenwave:
Following the discussion in https://github.com/cplusplus/draft/pull/8043#discussion_r2205802486

We already have the defined term "character type", and I find it silly not to use it but to enumerate all of its members.

Jonathan stated that there should be a cross-reference (I assume to [basic.fundamental]), which is a good idea.


jwakely (approved):

eisenwave:
@tkoeppe @jensmaurer can we merge?

PR #8069, opened 2025-07-15
[lex.separate] No more instantiation units

AlisdairM:
The term "instantiation unit" was removed by the application of P2996 for reflection, #7930.  This PR removes the last remaining reference from a note.

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -43,10 +43,8 @@ A \Cpp{} program need not all be translated at the same time.
 
 \pnum
 \begin{note}
-Previously translated translation units and instantiation
-units can be preserved individually or in libraries. The separate
-translation units of a program communicate\iref{basic.link} by (for
-example)
+Previously translated translation units can be preserved individually or in libraries.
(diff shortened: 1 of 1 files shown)


source/lex.tex
```diff
@@ -43,15 +43,13 @@
...
 \pnum
 \begin{note}
-Previously translated translation units and instantiation
-units can be preserved individually or in libraries. The separate
-translation units of a program communicate\iref{basic.link} by (for
-example)
+Previously translated translation units can be preserved individually or in libraries.
+The separate translation units of a program communicate\iref{basic.link} by (for example)
 calls to functions whose identifiers have external or module linkage,
-manipulation of objects whose identifiers have external or module linkage, or
```
tkoeppe, line 51: Yes, I realize semantics linebreaks would be better, but can we keep this edit to a minimum? Just the first part is fine.

AlisdairM: Simplified down to single commit touching only the affected lines, rebased, and force-pushed.

Issue #2076, opened 2018-05-12
Evaluate uses of `\textit` in text


godbyk:
In some cases `\textit{x}` should be one of
 * `$x$`
 * `\tcode{x}`
 * `\techterm{x}`
 * `\defn{x}`
 * `\term{x}`
 * `\placeholder{x}`
 * `\emph{x}`

instead.

jensmaurer:
This is almost a duplicate of #329.

PR #8081, opened 2025-07-17
Replace some `\textit`

eisenwave:
Related to https://github.com/cplusplus/draft/issues/2076

I'm not sure if it's worth dismantling all uses of `\textit`. There are ~150 left. This is just picking some low-hanging fruits in wording that might still be visited from time to time.

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -1712,9 +1712,9 @@ namespace std {
     static constexpr bool has_quiet_NaN     = true;
     static constexpr bool has_signaling_NaN = true;
 
-    static constexpr float infinity()      noexcept { return @\textit{value}@; }
-    static constexpr float quiet_NaN()     noexcept { return @\textit{value}@; }
-    static constexpr float signaling_NaN() noexcept { return @\textit{value}@; }
+    static constexpr float infinity()      noexcept { return @\placeholder{value}@; }
+    static constexpr float quiet_NaN()     noexcept { return @\placeholder{value}@; }
+    static constexpr float signaling_NaN() noexcept { return @\placeholder{value}@; }
     static constexpr float denorm_min()    noexcept { return min(); }
 
     static constexpr bool is_iec559  = true;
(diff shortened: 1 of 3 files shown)


source/support.tex
```diff
@@ -1956,47 +1956,47 @@
 #define __STDC_VERSION_STDINT_H__ 202311L
 
 namespace std {
-  using int8_t         = @\textit{signed integer type}@;   // optional
-  using int16_t        = @\textit{signed integer type}@;   // optional
-  using int32_t        = @\textit{signed integer type}@;   // optional
-  using int64_t        = @\textit{signed integer type}@;   // optional
+  using int8_t         = @\UNSPnc{signed integer type}@;   // optional
```
tkoeppe: Since we were using `\placeholder` for "value" in the other commit, I think this would also be better as a placeholder (i.e. `\placeholdernc`)? I don't think UNSP was meant to be reused like that, though using it for "see <elsewhere>" in the other other commit seems OK.

tkoeppe:
Thanks! Could you have a look also in text.tex, that also contains three instances of "see \ref{...}" in a \textit.

eisenwave:
I'll put it on my list of things to do once the repo is public and I can make updates to my PRs without things exploding.

jensmaurer:
@eisenwave , the (technical) blockers are gone here, I think.

source/support.tex
eisenwave: Done. Changed it to `\placeholdernc`.

jensmaurer (approved):
I'm not a big fan of surrounding "see \ref" with \placeholder, but I guess it's better than \textit.  We should have a separate macro for that.  Maybe later.

eisenwave:
@tkoeppe merge?

tkoeppe:
> Could you have a look also in text.tex, that also contains three instances of "see \ref{...}" in a \textit.

Actually, now that I think about it "see \ref{...}" isn't really a use for _placeholder_, is it -- maybe better if we made a dedicated macro `\seeref` for it. Yes, the entire expression is sort of a placeholder for a definition, but I don't think the macro name refers to the context of its use, but rather to its argument.

PR #8082, opened 2025-07-17
[exec.affine.on] Fix typo

Andreas-Krug:


tkoeppe:
There's also an "evalutation" somewhere, could you have a look?

tkoeppe:
Stealing.

PR #8084, opened 2025-07-17
Fix typo in [meta.reflection.access.context]/5.5.2

katzdm:
Typo was introduced during transcription to Latex (see draft approved in Sofia [here](https://wiki.edg.com/pub/Wg21sofia2025/CoreWorkingGroup/p2996r13.html#pnum_629)).


tkoeppe:
Thanks!

PR #8093, opened 2025-07-18
[simd.syn] Fix various formatting issues, improve aesthetics

### LINKED ISSUE #8031: [simd.syn] Wrong formatting in synopsis
The synopsis for in [[simd.syn]](https://eel.is/c++draft/simd.syn) goes against our usual formatting conventions in a few ways:

- Some of the section comments (e.g. `// [simd.bit], Bit manipulation`) have a capitalized title. To my knowledge, it should be `bit manipulation`.
- `assoc_laguerre` and `assoc_legendre` have improperly aligned parameters.
- `lerp` and a few other functions do not correctly indent the function name when the previous line is dedicated to a return type.

There are also some thing that are not incorrect, but very ugly, like:
```cpp
  template<simd-type V>
    constexpr rebind_t<make_signed_t<typename V::value_type>, V>
    countl_zero(const V& v) noexcept;
  template<simd-type V>
    constexpr rebind_t<make_signed_t<typename V::value_type>, V> countl_one(const V& v) noexcept;
```

`countl_zero` is not properly indented, but that's not what I'm pointing out here. Rather, `countl_one` is put all on one line solely because it's one character shorter. I think it would be much less ugly if we formatted `countl_zero` and `countl_one` the same.

---

Obviously, any such fixups should be done after post-Sofia motions.

Eisenwave:
Closes #8031.

This PR fixes a number of problems (some objective, some aesthetic) in [simd.syn]:
- Comments with references to subsequent sections should not capitalize the section name
- `vec` should be `\tcode{vec}`
- The two-space indentation in some places was missing
- Some code just needs to be reflown because parameters are misaligned and other things are broken

Aesthetics are also improved a bit. While it's not objectively wrong to do so, putting `countl_one` on one line and `countl_zero` onto two because of a single-character difference is ugly. It substantially improves aesthetics if we make it a bit more consistent.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -16515,7 +16515,7 @@ namespace std::simd {
   template<@\exposid{simd-size-type}@ N, class V> struct resize { using type = @\seebelow@; };
   template<@\exposid{simd-size-type}@ N, class V> using @\libmember{resize_t}{simd}@ = typename resize<N, V>::type;
 
-  // \ref{simd.flags}, Load and store flags
+  // \ref{simd.flags}, load and store flags
   template<class... Flags> struct flags;
   inline constexpr flags<> @\libmember{flag_default}{simd}@{};
   inline constexpr flags<@\exposid{convert-flag}@> @\libmember{flag_convert}{simd}@{};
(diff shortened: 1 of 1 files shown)


tkoeppe:
Thanks, that's nice!

Issue #7399, opened 2024-11-19
Main plan for clause reorganization for C++29


tkoeppe:
This issue is a living record of the developing plan for reorganizing clauses for C++29.

* Move [cpp] closer to [lex] and rework [lex]. See #7272 and #7352 for suggestions. We'd like a paper for the more profound changes in [lex].
  - #7180  
* Dissolved [except] and split it across various other core clauses. See #7317, #7352 for suggestions.

The above points need to be worked out in detail.

Further, unrelated suggestions for reorganization welcome.

AlisdairM:
Linking issue #3957

PR #8100, opened 2025-07-19
[cpp.pragma.op] Colocate the two pragma specifications

AlisdairM:
Moves the specification for the `_Pragma` operator to follow the specification for the `pragma` directive, to provide clearer context.

### DIFF
--- a/source/preprocessor.tex
+++ b/source/preprocessor.tex
@@ -2161,6 +2161,43 @@ The behavior may cause translation to fail or cause the translator or
 the resulting program to behave in a non-conforming manner.
 Any pragma that is not recognized by the implementation is ignored.
 
+\rSec1[cpp.pragma.op]{Pragma operator}%
+\indextext{macro!pragma operator}%
+\indextext{operator!pragma|see{macro, pragma operator}}
+
+\pnum
+A unary operator expression of the form:
+\begin{ncbnf}
+\terminal{_Pragma} \terminal{(} string-literal \terminal{)}
+\end{ncbnf}
+is processed as follows: The \grammarterm{string-literal} is \defnx{destringized}{destringization}
+by deleting the \tcode{L} prefix, if present, deleting the leading and trailing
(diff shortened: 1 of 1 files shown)


eisenwave:
I think there's some logic to what we have right now, which is that `_Pragma` is last because it's a different type of feature from the usual preprocessing directives.

However, I would be slightly happier if `#pragma` and `_Pragma` were next to each other.

AlisdairM:
I did consider whether it was intentional to sparate the operator specification from the directives specification.
I think that predefined macros coming last is not a concern, and maybe an improvement.

The only directive that gets split from the rest is the null directive, so I considered moving that to In order to avoid such splittage, but on balance chose to keep the initial PR minimal.  Plus, I had at least two ideas for where the null directive might move: 1) immediately before the pragma directive, so the smallest localized change 2) make it the very first directive, getting the trivial case out of the way before jumping into more detailed features.  That also inspired 3) move the error directives to the top too, giving an easy intro to the simplest directives.

All of those changes seemed more appropriate to a follow-up PR, although there was also last option that seemed too heavy handed: 4) operators and directives mix together in earlier subclauses where they are related, so open up a new parent for both the pragma directive and the pragma operator.

My personal choice would be option 3, although I have listed the 4 options in roughly the order I think would br least to most controversial :)

AlisdairM:
> Is this promised to be just-moving-text-around?

Yes.  If someone spots a difference in the text (and I reviewed after rebasing onto a recent `main`) then that should be flagged and fixed before landing.  I believe the current PR is correct as filed.

tkoeppe:
I seriously struggle to muster motivation for this change. Yeah, we could, but if the shoe were on the other foot I could just as well imagine arguing for the opposite move. I'm somewhat hesitant to perform a large number of "just because we could" churn-ish permutations that don't solve any obvious defect in clarity.

AlisdairM:
I am withdrawing this PR rather than leave it hanging in limbo.
I might revisit in the context of larger changes if we consider clause 15 a concern for #7399.

PR #8103, opened 2025-07-19
[optional.optional.ref.general] Fix indentation

Andreas-Krug:


eisenwave:
It's an improvement of the status quo, but to my understanding, definitions shouldn't add indentation after the *template-head* at all.

The spec guidelines are a bit unclear on what's correct here.

Andreas-Krug:
Thanks for the explanation.

So I create a new (hopefully correct) PR.

Issue #7214, opened 2024-08-11
[linalg] specification glitches


mhoemmen:
# [linalg] specification glitches

Many thanks to my NVIDIA colleague Ilya Burylov for discovering these issues.

We've checked this against the current C++ Working Draft N4988.

## Fix [linalg.transp.layout.transpose]

For all mdspan layout mappings in the Standard, the layout itself (e.g., `layout_right`) has no members other than the class template `mapping`, and `mapping` has all the layout-related "stuff."  However, this appears to be broken with `layout_transpose`.  There, `layout_transpose` itself has private members _`nested-mapping`_`_` and _`extents`_`_`.  Additional evidence that this is incorrect is that `layout_transpose::mapping` constructor specification refers to initializing _`nested-mapping`_`_` and _`extents`_`_`.

I am the main author of this proposal; I did not intend this.  _`nested-mapping`_`_` and _`extents`_`_` should instead be private members of `layout_transpose::mapping`.

### Fix constructor from `NestedAccessor`

The `conjugated_accessor(const NestedAccessor& acc)` constructor appears in the wording ([linalg.conj.conjugatedaccessor] para 3) but is missing from the class synopsis (para 1).  Please add it to the class synopsis, immediately after `conjugated_accessor`'s default constructor.

### Fix converting constructor from `conjugated_accessor<OtherNestedAccessor>`

For the converting constructor from `const conjugated_accessor<OtherNestedAcessor>&`, in both the condition inside `explicit` in the class synopsis (para 1) and the condition inside `explicit` in the description (just above para 4) have an extra closing `>`.

### Fix return type of `nested_accessor` member function

The `nested_accessor` member function of `conjugated_accessor` (which is defined inline in the class' synopsis) currently returns `const Accessor&`.  This is incorrect, because there is no `Accessor` type in scope.

The intent is for the function to return `const NestedAccessor&`.  As main author of this proposal, this is the intent.  At some point in revisions, the template parameter was renamed from `Accessor` to `NestedAccessor`.  It's likely that we missed changing the return type of `nested_accessor`.

## Fix [linalg.scaled.scaledaccessor]

[...]

PR #8105, opened 2025-07-20
[linalg.conj.conjugatedaccessor] Fix incorrect return type of `nested_accessor`

Eisenwave:
One part of #7214.

As @mhoemmen explains:

> ## Fix return type of nested_accessor member function
> 
> The `nested_accessor` member function of `conjugated_accessor` (which is defined inline in the class' synopsis) currently returns `const Accessor&`. This is incorrect, because there is no `Accessor` type in scope.
> 
> The intent is for the function to return `const NestedAccessor&`. As main author of this proposal, this is the intent. At some point in revisions, the template parameter was renamed from `Accessor` to `NestedAccessor`. It's likely that we missed changing the return type of `nested_accessor`.

@jwakely does this need an LWG issue or can we "just fix this"?

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -12802,7 +12802,7 @@ namespace std::linalg {
     constexpr typename offset_policy::data_handle_type
       offset(data_handle_type p, size_t i) const;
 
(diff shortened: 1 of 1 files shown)


jensmaurer (approved):

jwakely (approved):
Seems editorial

PR #8114, opened 2025-07-22
[const.wrap.class] Add \exposidnc to cw-operators

hewillk:


eisenwave:
Implemented by #8395.

PR #8115, opened 2025-07-22
[lex.pptoken] Reoder list items to avoid "or" and "and" at same level

tkoeppe:
Suggested by editorial review committee.


tkoeppe:
Thanks!

PR #8116, opened 2025-07-24
[expr.call] Itemize to clarify sequencing ambiguity

eisenwave:
The current wording is ambiguous and can be read in two ways:

> The
> - initialization of a parameter or,
> - if the implementation introduces any temporary objects to hold the values of function parameters ([class.temporary]), the initialization of those temporaries,
>
> including every associated value computation and side effect,
> is indeterminately sequenced with respect to that of any other parameter.

OR

> The
> - initialization of a parameter or,
> - if the implementation introduces any temporary objects to hold the values of function parameters ([class.temporary]), the initialization of those temporaries,
> including every associated value computation and side effect,
> 
> is indeterminately sequenced with respect to that of any other parameter.

I think it's fairly clear that we want "including every associated value computation and side effect" to apply to either case, and this also matches the original wording in https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0145r3.pdf

Besides the ambiguity, the wording is mentally demanding from the reader, and bullets help tidy it up.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -3903,10 +3903,13 @@ is not considered.
 The \grammarterm{postfix-expression} is sequenced before
 each \grammarterm{expression} in the \grammarterm{expression-list}
 and any default argument.
-The initialization of a parameter or,
-if the implementation introduces any temporary objects
+The
+\begin{itemize}
+\item initialization of a parameter or,
+\item if the implementation introduces any temporary objects
 to hold the values of function parameters\iref{class.temporary},
 the initialization of those temporaries,
+\end{itemize}
 including every associated value computation and side effect,
 is indeterminately sequenced with respect to that of any other parameter.
 These evaluations are


Eisenwave:
It seems like much of this wording is unnecessary because it's already half-covered by https://eel.is/c++draft/dcl.init#general-19

killerbee13:
As far as I can tell, the general wording in https://eel.is/c++draft/dcl.init#general-19 would mean that there is no reason to include this particular clause in this sentence to begin with, and perhaps it should simply be removed since its binding is somewhat ambiguous and the sentence is rather verbose. However, that is not quite as clear as https://eel.is/c++draft/intro.execution#8.sentence-7 is as to whether "includes" means that sequencing is also propagated downward. Unless it is intended that "initializations" can be sequenced but their side effects and value computations *not* be sequenced, which as far as I can tell it never is, I think a similarly strong statement should be made generally, rather than calling it out individually in all of those cases where initializations aren't already also full-expressions.

PR #8117, opened 2025-07-24
[lex.phases] Move dropping whitespace to end of phase 4

AlisdairM:
Phase 7 deals only with tokens, not whitespace.  For clarity move the dropping of whitespace to the end of phase 4, after preprocessing directives are deleted.  Newline characters are significant in the grammar of preprocessing directives to whitespace characters are clearly signicant until this point. Conversely, it is good to recognise that string literal preprocessing tokens are adjacted regardless of whitespace in phases 5 and 6.


source/lex.tex
```diff
@@ -165,6 +165,7 @@
 expanded\iref{cpp.replace}, and \tcode{_Pragma} unary operator expressions are executed\iref{cpp.pragma.op}.
 A \tcode{\#include} preprocessing directive\iref{cpp.include} causes the named header or
 source file to be processed from phase 1 through phase 4, recursively.
+Whitespace characters separating preprocessing tokens are no longer significant.
```
jensmaurer: At the very least, this is an unfortunate place where to put this rule, since the next sentence continues the "preprocessing directive" theme from the previous sentences.

jensmaurer (changes requested):

source/lex.tex
AlisdairM: Oops.  I thought I had fixed this before first submitting as I noticed the same.
Revised, rebased, and force-pushed.

tkoeppe:
Final thoughts, @jensmaurer?

jensmaurer:
This is not a technical concern, but more a social one.  I've sent e-mail to the core reflector so people are aware of the move.  Let's merge on Monday, unless there are objections.

AlisdairM:
Is this ready to land?

jensmaurer (approved):
This is ok.

PR #8120, opened 2025-07-26
[dcl.fct] Clarify that type adjustment does not make zero-length arrays valid CWG 3121

### LINKED ISSUE #6575: [dcl.fct] It is unclear whether zero-size arrays in function parameters are allowed
[[dcl.fct] p5](http://eel.is/c++draft/dcl.fct#5) does not sufficiently clarify the following case:
```cpp
void foo(int[0]);
```
The type of `int[0]` is adjusted to `int*`.

I believe that the existing wording in [[dcl.array]](http://eel.is/c++draft/dcl.array) is sufficiently clear for this to be an editorial issue. `int[0]` is not a valid type, so an adjustment of it cannot retroactively make it valid.

We could add the following line to the example in the mentioned paragraph:
```cpp
void j(int[0]); // ill-formed
```

eisenwave:
Closes #6575.

Currently non-editorial for triage purposes. Jens wants CWG involvement (see linked issue).

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -3871,6 +3871,8 @@ void g(char[3][3]) {}           // OK, another overload
 
 void h(int x(const int));       // \#3
 void h(int (*)(int)) {}         // defines \#3
+
+void j(int[0]);                 // error: \tcode{int[0]} is not a valid type
 \end{codeblock}
 \end{example}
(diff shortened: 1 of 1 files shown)


eisenwave:
CWG3121

PR #8126, opened 2025-07-27
[basic.pre][basic.link] Names have linkage

AlisdairM:
An essential property of names is that they have linkage. Move the statement that names have linkage up from [basic.link] to [basic.pre] where the term `name` is defined, and before the references to linkage of names in the clauses between [basic.pre] and [basic.link].


jensmaurer:
Strategically, some CWG people believe it's a bad approach that names have linkage; instead, entities should have linkage. I haven't seen a comprehensive fix for that, yet, though.

AlisdairM:
> Strategically, some CWG people believe it's a bad approach that names have linkage; instead, entities should have linkage. I haven't seen a comprehensive fix for that, yet, though.

Do you object to moving the current text to at least improve clarify of the current specification?  It would then be easy to move the parapraph down a couple of lines if and when we apply it to entities instead.

tkoeppe:
Editorial meeting 2026-03-26: Leave the linkage wording alone. It is subject to ongoing CWG considerations. Moreover, the current placement of the wording fits well and establishes the connection of names having linkage, and the proposed move of the normative wording is not an improvement.

PR #8129, opened 2025-07-27
[dcl.inline] inline specifier is for ODR

AlisdairM:
The key use of the inline specifier since C++11, if not before, has been to allow multiple declarations to satisfy the ODR rather than to provide a hint that compilers routinely ignore.

This change moves but does not change wording, in order to move the comment making the connection with the ODR more prominent than the normative wording suggestings core transformation.

It might be desirable to demote the normative coding hint to a note, but that goes beyond the remit of a simple editorial chsnge.


source/declarations.tex
```diff
@@ -1111,10 +1126,6 @@
 see~\ref{basic.link}.
 \end{note}
 
-\pnum
-A variable declaration with an \keyword{inline} specifier declares an
-\defnadj{inline}{variable}.
-
 \pnum
 The \keyword{inline} specifier shall not appear on a block scope declaration or
```
jensmaurer: If we do some ordering cleanup here at all, we should observe that we traditionally have all (syntax-ish) restrictions first before we dig into semantics. So, this paragraph should be integrated into the existing p1.

Is there any reason why we now define "inline variable" before "inline function", given that the moved note refers to both, and it says "inline function or variable" (in that order)?

AlisdairM: I thought about merging p1 and the revised p2 too, but wanted to be conservative before review.  Will make that change.  Re: order --- I must have been looking at old text before I moved p3 up, as I was sure p1 had variable before function.  Will make consistently alphabetical.

AlisdairM:
Further consolidated text per suggestion above.
Moved Comment 2 about inline not affecting linkage up to the end of p1 as well.

source/declarations.tex
jensmaurer: The "shall not appear at block scope" restriction hasn't been moved to p1, before all the fluff about defining "inline function" and "inline variable", contrary to what I thought I had asked for.

The "indicates to the implementation" note should be in p2 or later; that part is fine.

AlisdairM: Sorry, misunderstanding.
Pushed what is hopefully the requested version.

PR #8134, opened 2025-07-30
[task.promise] Clarifications and disambiguations

tkoeppe:
Cleans up multiple issues in p16:

* The name "alloc" was confusingly used both for an exposition-only member and for a locally defined variable. The latter is renamed to "alc".
* Locally defined variables are just typeset as normal code, not as placeholders.
* Some commas inserted to separate coordinate subclauses.
* Replaced "(if any)" with leading "Oherwise", which seems better in order to avoid seemingly providing two conflicting definitions of `alc`.

**Open question:** What is "`Allocator`"?

### DIFF
--- a/source/exec.tex
+++ b/source/exec.tex
@@ -7648,19 +7648,19 @@ template<class... Args>
 \begin{itemdescr}
 \pnum
 If there is no parameter with type \tcode{allocator_arg_t} then let
-\tcode{\placeholder{alloc}} be \tcode{Allocator()}.
-Let \tcode{\placeholder{arg_next}} be the parameter following the first
-\tcode{allocator_arg_t} parameter (if any) and let \tcode{\placeholder{alloc}}
-be \tcode{Allocator(\placeholder{arg_next})}.
-Then \tcode{PAlloc} is \tcode{allocator_traits<Allocator>::template
-re\-bind_alloc\brk{}<U>} where \tcode{U} is an unspecified type
-whose size and alignment are both \tcode{__STDCPP_DEFAULT_NEW_ALIGNMENT__}.
+\tcode{alloc} be \tcode{allocator_type()}.
+Otherwise, let \tcode{arg_next} be the parameter
+following the first \tcode{allocator_arg_t} parameter,
+and let \tcode{alloc} be \tcode{allocator_type(arg_next)}.
+Let \tcode{PAlloc} be \tcode{allocator_traits<allocator_type>::template
+rebind_alloc<U>}, where \tcode{U} is an unspecified type
+whose size and alignment are both \mname{STDCPP_DEFAULT_NEW_ALIGNMENT}.
 
 \pnum
 \mandates
 \begin{itemize}
 \item The first parameter of type \tcode{allocator_arg_t} (if any) is not the last parameter.
-\item \tcode{Allocator(\placeholder{arg_next})} is a valid expression if there is a parameter
+\item \tcode{allocator_type(arg_next)} is a valid expression if there is a parameter
 of type \tcode{allocator_arg_t}.
 \item \tcode{allocator_traits<PAlloc>::pointer} is a pointer type.
 \end{itemize}
@@ -7668,7 +7668,7 @@ of type \tcode{allocator_arg_t}.
 \pnum
 \effects
 Initializes an allocator \tcode{palloc} of type \tcode{PAlloc} with
-\tcode{\placeholder{alloc}}.
+\tcode{alloc}.
 Uses \tcode{palloc} to allocate storage for the
 smallest array of \tcode{U} sufficient to provide storage for a
 coroutine state of size \tcode{size}, and unspecified additional


tkoeppe:
image

tkoeppe:
@Dani-Hub Could you please have a look?

Dani-Hub:
> @Dani-Hub Could you please have a look?

Hmmh, good question. I _think_ the intention is that it is supposed to be `allocator_type` (As specified in [task.class] b (2.1) ("allocator_type is Environment::allocator_type if [...]"), that seems also be the `allocator_type` that the synopsis in [task.promise] is referring to. @dietmarkuehl : Is that correct? If this is _not_ correct I suggest that we should open an LWG issue.

AlisdairM:
The current wording with its current expos-only identifiers looks correct to me.
I think the issue is that the paragraph is poorly constructed.

If there is no `allocator_arg_t`, then _alloc_ is this; **otherwise** _alloc_ is that.

Then, regardless of "if' branch above, PAlloc is this type, and _alloc_/`alloc` are used correctly in the text that follows.

So I believe no renaming is required, but the paragraph could be better structured.

dietmarkuehl:
> > @Dani-Hub Could you please have a look?
> 
> Hmmh, good question. I _think_ the intention is that it is supposed to be `allocator_type` (As specified in [task.class] b (2.1) ("allocator_type is Environment::allocator_type if [...]"), that seems also be the `allocator_type` that the synopsis in [task.promise] is referring to. @dietmarkuehl : Is that correct? If this is _not_ correct I suggest that we should open an LWG issue.

Yes, that is correct: [task.class] p2.1 specifies how `allocator_type` is computed. The `Allocator` seems to be a left-over from earlier wording which slipped through the wording review.

source/exec.tex
```diff
@@ -7648,27 +7648,27 @@
 \begin{itemdescr}
 \pnum
 If there is no parameter with type \tcode{allocator_arg_t} then let
-\tcode{\placeholder{alloc}} be \tcode{Allocator()}.
-Let \tcode{\placeholder{arg_next}} be the parameter following the first
-\tcode{allocator_arg_t} parameter (if any) and let \tcode{\placeholder{alloc}}
-be \tcode{Allocator(\placeholder{arg_next})}.
+\tcode{alc} be \tcode{Allocator()}.
```
dietmarkuehl: This lines should read (using the type defined in [exec.task] p2.1):

    \tcode{alc} be \tcode{allocator_type()}.
    
(I didn't see a way to get a fix-it in the github UI).

source/exec.tex
```diff
@@ -7648,27 +7648,27 @@
...
+Otherwise, let \tcode{arg_next} be the parameter
+following the first \tcode{allocator_arg_t} parameter,
+and let \tcode{alc} be \tcode{Allocator(arg_next)}.
```
dietmarkuehl: Similar to above:

    and let \tcode{alc} be \tcode{allocator_type(arg_next)}.

source/exec.tex
```diff
@@ -7648,27 +7648,27 @@
...
 Then \tcode{PAlloc} is \tcode{allocator_traits<Allocator>::template
```
dietmarkuehl: ... and another one:

     Then \tcode{PAlloc} is \tcode{allocator_traits<allocator_type>::template

source/exec.tex
```diff
@@ -7648,27 +7648,27 @@
...
-whose size and alignment are both \tcode{__STDCPP_DEFAULT_NEW_ALIGNMENT__}.
+rebind_alloc<U>}, where \tcode{U} is an unspecified type
+whose size and alignment are both \mname{STDCPP_DEFAULT_NEW_ALIGNMENT}.
 
 \pnum
 \mandates
 \begin{itemize}
 \item The first parameter of type \tcode{allocator_arg_t} (if any) is not the last parameter.
-\item \tcode{Allocator(\placeholder{arg_next})} is a valid expression if there is a parameter
+\item \tcode{Allocator(arg_next)} is a valid expression if there is a parameter
```
dietmarkuehl: ... and another one:

    \item \tcode{allocator_type(arg_next)} is a valid expression

source/exec.tex
tkoeppe: Got it, will add a commit.

source/exec.tex

source/exec.tex

source/exec.tex

tkoeppe:
@Dani-Hub, @dietmarkuehl could you have another look?

Dani-Hub (approved):
Looks good to me.

tkoeppe:
Thanks, everyone!

tkoeppe:
I reverted the renaming of "alloc" to "alc". I think Alisdair has a point that there is no local ambiguity once the typography has been cleaned up, and the Otherwise rephrasing makes the local definition fairly clear.

Dani-Hub:
Agreed

tkoeppe:
I'm also changing "Then X is" to "Let X be" for the PAlloc sentence, which (I think) clears up Alisdairs other complaint.

tkoeppe:
image

Dani-Hub (approved):

PR #8136, opened 2025-07-31
[pre] No names in the preprocessor

AlisdairM:
The term "name" applies specifically to entities in phase 7 of translation.  Macros have macro names, headers are parsed as *header-name*s, etc.


source/preprocessor.tex
```diff
@@ -618,7 +618,7 @@
 Each directive's condition is checked in order.
 If it evaluates to false (zero),
 the group that it controls is skipped:
-directives are processed only through the name that determines
+directives are processed only through the macro name that determines
```
cpplearner: I think this "name" refers to the directive name.

AlisdairM: Ahah!  That makes sense, and now perhaps I can understand *how* this paragraph says *what* I know it is supposed to mean :)

tkoeppe:
I'm not sure if this requires clarification -- e.g. in "the name of the macro", isn't it clear that the name is a macro name?

@jensmaurer, any thoughts?

jensmaurer:
[cpp.replace.general] p8 does define the term "macro name".

Reading through the diff here, I think there are a few places that definitely benefit from replacing "name" with something else entirely or with "macro name".

Others, such as "macro name of a macro", sound awfully redundant, and I think we want a larger rephrasing to warrant the departure from "identical phrasing in C" for those cases.

Going through the diff one-by-one and marking those that seem redundant / need more rephrasing.

source/preprocessor.tex
```diff
@@ -500,7 +500,7 @@
 directives, and
 the \tcode{defined} conditional inclusion operator,
 shall treat \xname{has_include}, \xname{has_embed}, and \xname{has_cpp_attribute}
-as if they were the names of defined macros.
+as if they were the macro names of defined macros.
```
jensmaurer: redundant

source/preprocessor.tex
```diff
@@ -1571,7 +1571,7 @@
 An alternative token\iref{lex.digraph} is not an identifier,
 even when its spelling consists entirely of letters and underscores.
 Therefore it is not possible to define a macro
-whose name is the same as that of an alternative token.
+whose macro name is the same as that of an alternative token.
```
jensmaurer: redundant

source/preprocessor.tex
```diff
@@ -2022,12 +2022,12 @@
 \end{example}
 
 \pnum
-If the name of the macro being replaced is found during this scan of
+If the macro name of the macro being replaced is found during this scan of
```
jensmaurer: redundant

source/preprocessor.tex
```diff
@@ -2022,12 +2022,12 @@
...
 the replacement list
 (not including the rest of the source file's preprocessing tokens),
 it is not replaced.
 Furthermore,
-if any nested replacements encounter the name of the macro being replaced,
+if any nested replacements encounter the macro name of the macro being replaced,
```
jensmaurer: redundant

source/preprocessor.tex
AlisdairM: How about "as if they named defined macros", i.e., using name as a verb rather than a noun?

AlisdairM:
I have reverted the changes that Jens requested.
I made an alternative suggestion to one in the review for potential follow-up, but reverted in the PR to aid progress.

source/preprocessor.tex
jensmaurer: Sounds good.

PR #8137, opened 2025-07-31
[expr.prim.lambda.closure] Use "incomplete" instead of "not complete".

tkoeppe:
The former is a defined term.


tkoeppe:
@katzdm: FYI

jensmaurer (approved):

PR #8138, opened 2025-08-01
Redundant

tkoeppe:
Found by @timsong-cpp.

@katzdm: could you please take a look?


katzdm (approved):
Yeah, I think this is detritus from when  we wrote this case in terms of _namespace-name_ (see [R12](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p2996r12.html#pnum_196)). Agreed, remove.

jensmaurer (approved):

PR #8141, opened 2025-08-02
[locale.operators] Qualify `collate`

cpplearner:
Within `locale::operator()`, unqualified name lookup for `collate` finds `locale::collate`, which is a bit mask. Clearly the intent is to refer to the class template `std::collate`.


jensmaurer (approved):

jensmaurer:
@jwakely ?

jwakely (approved):

PR #8148, opened 2025-08-14
[cmp.common] Remove unnecessary "std::" before strong_ordering

eisenwave:
Based on my search, this is the only occurrence of `std::` before `strong_ordering`, `partial_ordering`, or `weak_ordering` in library wording.

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -5323,7 +5323,7 @@ the expanded parameter pack, or
 \keyword{void} if any element of \tcode{Ts}
 is not a comparison category type.
 \begin{note}
-This is \tcode{std::strong_ordering} if the expansion is empty.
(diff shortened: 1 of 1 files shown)


jwakely:
Since this is a note, I don't think it's a problem to qualify it. I don't really care about this either way.

eisenwave:
I think it depends a bit on whether we have ambitions for the aforementioned checks. The reason I've paid attention to it in the first place is because I've accidentally used `std::strong_ordering` in a proposed declaration in one of my papers.

So maybe there's value in ensuring such issues don't sneak their way into the standard, on the off chance that LWG doesn't catch them.

eisenwave:
We've got enough other things to do for now.

PR #8152, opened 2025-08-15
[meta.reflection.queries]/4.3 Fix syntax in example

birbacher:
The parenthesis in the example seem to do the wrong grouping. To fix, remove one set of parens. This brings the example in line with P2996R13 from which this wording originates.


jwakely (approved):

tkoeppe:
You don't need to create a novel justification for this situation: it's simply a misapplication of the paper, which is all we need to know.

PR #8153, opened 2025-08-15
[meta.reflection.queries]/1 Use info return type

birbacher:
Bring the return type of type_of in line with the synopsis and with the intention of the function and with the wording in P2996R13 from where this wording originates.


birbacher:
> that's all that needs to be said really.

Yeah, I know. I might be less verbose next time. :)

PR #8157, opened 2025-08-15
[meta] Fix bunch of typos

Rageking8:


source/meta.tex
```diff
@@ -3906,7 +3906,7 @@
 
 \indexlibraryglobal{type_of}%
 \begin{itemdecl}
-consteval bool type_of(info r);
+consteval info type_of(info r);
```
jwakely, line 3909: There's another pull request that also fixes this one.

Rageking8: > There's another pull request that also fixes this one.

Do I need to take any action (i.e. revert this change)?

jwakely: No, it will still apply cleanly

Rageking8:
Is there any further action required before this PR can be merged?

PR #8159, opened 2025-08-15
[locale.moneypunct.virtuals] latex: remove redundant backslash-space

morinmorin:
The backslash-space in "U.S.\ locales" can be replaced with a plain space, because TeX treats a period after uppercase as an acronym (which yields normal spacing). L4595, L4610 already use a plain space in "U.S. locales", so this change improves consistency.

### DIFF
--- a/source/text.tex
+++ b/source/text.tex
@@ -4673,7 +4673,7 @@ int do_frac_digits() const;
 \returns
 The number of digits after the decimal radix separator, if any.
 \begin{footnote}
-In common U.S.\ locales, this is 2.
+In common U.S. locales, this is 2.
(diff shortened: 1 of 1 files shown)


tkoeppe:
OK, but is this the only place in the entire standard where the backslash is redundant? Since this doesn't affect anything, I'd rather only consider such cleanups if it's the last and only time we'll ever have to think about this.

morinmorin:
Other than _bibliography_, I think yes. I checked with this command: `grep -P -n '[A-Z][.?!:;]\\\s' *.tex`.

tkoeppe:
Thanks!

PR #8160, opened 2025-08-15
[macros] Fix sentence spacing after `\tcode`

morinmorin:
TeX treats a word ending with an uppercase letter as an abbreviation. This makes the space after `\tcode{T}.` result in regular interword spacing. To restore proper (wider) sentence spacing, this commit appends `\@` to the `\tcode` macro.

There are dozens of intances of `\tcode{...UPPERCASE}. `

### DIFF
--- a/source/macros.tex
+++ b/source/macros.tex
@@ -242,7 +242,7 @@
 \newcommand{\GrammarStylex}[1]{\textcolor{grammar-gray}{\textsf{\textit{#1}}}}
 
 % Code and definitions embedded in text.
-\newcommand{\tcode}[1]{\CodeStylex{#1}}
+\newcommand{\tcode}[1]{\ifmmode\CodeStylex{#1}\else\CodeStylex{#1}\@\fi}
 \newcommand{\term}[1]{\textit{#1}}
 \newcommand{\gterm}[1]{\GrammarStylex{#1}}
 \newcommand{\fakegrammarterm}[1]{\gterm{#1}}


jensmaurer:
Could you paste a (small) screenshot of an example showing the spacing before/after this change?

Did you run "diffpdf" in a sufficiently detailed mode to ensure nothing gets broken?

source/macros.tex
```diff
@@ -242,7 +242,7 @@
 \newcommand{\GrammarStylex}[1]{\textcolor{grammar-gray}{\textsf{\textit{#1}}}}
 
 % Code and definitions embedded in text.
-\newcommand{\tcode}[1]{\CodeStylex{#1}}
+\newcommand{\tcode}[1]{\ifmmode\CodeStylex{#1}\else\CodeStylex{#1}\@\fi}
```
tkoeppe: Can you add a comment like "restore normal sentence spacing when not in math mode"?

tkoeppe:
This change seems to cause us to lose some index entries. Please see std-generalindex.ilg.

morinmorin:
Sorry for the breakage. I'll dig into it more and I'll attach a diff screenshot after it's resolved.

morinmorin:
OK, fixed the breakage and added a comment.

morinmorin:
There are hundreds of spacing updates.
Many are subtle/unnoticeable (typical patterns: `\tcode{…UPPERCASE}, ` and `\tcode{…UPPERCASE}; `):
diff1

Here is a typical sentence-spacing fix involving `\tcode{…UPPERCASE}. `:
diff2

In a few places, multiple lines reflow slightly:
diff3

There is only one instance where a paragraph extends by one line, causing a substantial page reflow (from [linalg.general] p4 to [linalg.algs.blas3.trmm] p8). This change only shifts a page break — confirmed with pdfdiff — so I think this is fine.
diff4

**How to read these images** (generated by diff-pdf): 
Differences are color-coded — red indicates the original and cyan indicates the revised. Any line that contains a difference is marked with a red marker in the left margin.

morinmorin:
Oops, forgot to reply to this one.

> Did you run "diffpdf" in a sufficiently detailed mode to ensure nothing gets broken?

After fixing the breakage, yes, I checked with DiffPDF (and diff-pdf) and didn't find any particular problems.

source/macros.tex
```diff
@@ -242,7 +242,11 @@
 \newcommand{\GrammarStylex}[1]{\textcolor{grammar-gray}{\textsf{\textit{#1}}}}
 
 % Code and definitions embedded in text.
-\newcommand{\tcode}[1]{\CodeStylex{#1}}
+\makeatletter
+% Append `\@` to restore proper sentence spacing in text mode. This insertion
+% happens only during normal typesetting; it is suppressed for .idx generation.
+\newcommand{\tcode}[1]{\CodeStylex{#1\ifmmode\else\ifx\protect\@typeset@protect\@\fi\fi}}
```
tkoeppe: Could we factor the common logic out into some helper macro (to be put say on L238) which can be used both here and in the two redefinitions below?

morinmorin: You're right, fixed!

tkoeppe: Thanks, that's great!

tkoeppe:
@morinmorin: Hi! Is it possible that this change has introduced a lot of new "Token not allowed in a PDF string" warnings? Are we missing some `\texorpdfstring` somewhere to exclude the new constructions from use in PDF specials?

morinmorin:
@tkoeppe Thanks, you're right that this commit does introduce those warnings.

The commit log says
> The insertion of `\@` happens only during normal typesetting; it is suppressed when writing the .idx file and when generating bookmarks.

But it fails to suppress the insertion of `\@` when generating bookmarks. A simple fix is to add
```
\pdfstringdefDisableCommands{\def\@{}}
```
to `std.tex`; I confirmed this fixes the issue.

morinmorin:
Here's PR #8432 for convenience.

PR #8162, opened 2025-08-15
[hive.erasure]/2 Have erase_if reevaluate end() LWG4318

birbacher:
The defining code must not cache the end-iterator. In case the last element of the sequence is removed, the past-the-end iterator will be invalidated. This will trigger UB in the loop condition. Instead, re-evaluate end() each time.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -8969,7 +8969,7 @@ template<class T, class Allocator, class Predicate>
 Equivalent to:
 \begin{codeblock}
 auto original_size = c.size();
-for (auto i = c.begin(), last = c.end(); i != last; ) {
+for (auto i = c.begin(); i != c.end(); ) {
   if (pred(*i)) {
     i = c.erase(i);
   } else {


tkoeppe:
This seems like a normative change. @jwakely?

birbacher:
At least it doesn't change the intended behaviour, I'd say. But if you reject, we'll file it as an NB comment.

tkoeppe:
I don't want to make that call unilaterally, but if there's agreement from LWG that it's an obvious oversight, I'd be OK to handle it editorially.

jwakely:
This seems badly normative to me, the current code is broken and avoiding repeated calls to `end()` doesn't seem like a critical part of the effects. If it really matters, we could continue caching it and do `last = c.end();` after erasing an element.

Please just file an LWG issue, there's no reason to wait for the NB ballot.

Dani-Hub:
We have now [LWG 4318](https://cplusplus.github.io/LWG/issue4318) for this.

tkoeppe:
Thanks, Daniel!

PR #8165, opened 2025-08-17
[back] Fix journal article entries in bibliography

morinmorin:
This PR applies the following corrections and formatting unifications.
- Fixes author name typo: "R. Client Whaley" should be "R. Clint Whaley". (c.f. https://doi.org/10.1145/567806.567807 )
- Adds missing page numbers.
- Unifies article title format: Roman style with sentence capitalization.
- Unifies journal title format: italicized.
- Unifies journal volume/issue format: "vol. X, no. Y".

### DIFF
--- a/source/back.tex
+++ b/source/back.tex
@@ -61,20 +61,20 @@
   Prentice-Hall, ISBN 0-13-117003-1, copyright \copyright 1995 P.\,J.\ Plauger
 \bibitem{linalg-stable}
   J.\ Demmel, I.\ Dumitriu, and O.\ Holtz,
-  \doccite{Fast linear algebra is stable},
-  Numerische Mathematik 108 (59--91), 2007.
+  Fast linear algebra is stable,
+  \doccite{Numerische Mathematik}, Vol.\ 108, No.\ 1, pp.\ 59--91, 2007.
 \bibitem{blas1}
   C.\,L.\ Lawson, R.\,J.\ Hanson, D.\ Kincaid, and F.\,T.\ Krogh,
-  \doccite{Basic linear algebra subprograms for Fortran usage}.
-  ACM Trans.\ Math.\ Soft., Vol.\ 5, pp.\ 308--323, 1979.
+  Basic linear algebra subprograms for Fortran usage.
+  \doccite{ACM Trans.\ Math.\ Soft.}, Vol.\ 5, No.\ 3, pp.\ 308--323, 1979.
 \bibitem{blas2}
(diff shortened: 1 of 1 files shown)


source/back.tex
```diff
@@ -61,20 +61,20 @@
   Prentice-Hall, ISBN 0-13-117003-1, copyright \copyright 1995 P.\,J.\ Plauger
 \bibitem{linalg-stable}
   J.\ Demmel, I.\ Dumitriu, and O.\ Holtz,
-  \doccite{Fast linear algebra is stable},
-  Numerische Mathematik 108 (59--91), 2007.
+  Fast linear algebra is stable,
+  \doccite{Numerische Mathematik}, Vol.\ 108, No.\ 1, pp.\ 59--91, 2007.
```
jensmaurer, line 65: We seem to \doccite the title of the book / article elsewhere in this list, not (necessarily) the journal name.

Agreed with the rest of the adjustments.

morinmorin: OK, will fix.
Just to be sure: is it okay to continue with the current "articles title in italic" style? In usual bibliographic styles, article titles are written in roman and journal names are written in italic. Examples (C.7.2) in ISO 690:2010 follow this convention as well.

jensmaurer: @tkoeppe, what do you want here?

tkoeppe: Yeah, I was aware that we're not following the usual conventions here. (E.g. bibtex would do it right, but when we started we just had a very small set of items.) This is a nice improvement.

tkoeppe:
Please update the commit message to be a bit more specific.

tkoeppe:
Actually, please split this up into the actual corrections (typos and missing page numbers), and separately the style changes.

morinmorin:
I split this into two commits. Please let me know if you'd like me to split this into two PRs.

tkoeppe:
Thanks!

I suppose what's somewhat unfortunate about the established style is that there's no recognizable distinction between the author list and the title. This has always irked me, and that's why I was fine with our existing style, even though it goes against some convention...

morinmorin:
Thanks for sharing your perspective! I can totally see how that lack of distinction could be annoying.

tkoeppe:
I've applied the fixes for now, and we can maybe leave this one open for discussion for a while.

tkoeppe:
Editorial meeting decision: we generally like the style we use, even though it differs from the academic style. We might even entertain the idea of also italicizing the titles of the three online references.

PR #8183, opened 2025-08-21
[cpp.replace.general] Add a cross-reference to 'see below'

AlisdairM:
The 'below' in 'see below' as actually five sublauses away. Adding a cross-reference establishes the link more clearly.


source/preprocessor.tex
```diff
@@ -1574,7 +1574,7 @@
 whose name is the same as that of an alternative token.
 \end{footnote}
 The replacement list is then rescanned for more macro names as
-specified below.
+specified below\iref{cpp.rescan}.
```
jensmaurer: Can we just say more conventionally "... as specified in \ref{cpp.rescan}."

please?  Most everything is "below".

AlisdairM: Makes sense.  I was trying to make the minimal change, and yes, "everything is below" was exactly what I was trying to address :)

jensmaurer (approved):

tkoeppe:
Yes, very helpful, thanks!

PR #8186, opened 2025-08-22
[basic.pre] Clarify definition of variable

AlisdairM:
Reorder the positive and negative terms to avoid confusion with the binding of "other than".  While grammatically the current wording is precise and correct, I have run into several folks misreading this and wondering why references to objects are not variables.


source/basic.tex
```diff
@@ -113,8 +113,8 @@
 \pnum
 A \defn{variable} is introduced by the
 declaration of
-a reference other than a non-static data member or of
-an object.
+an object or of
+a reference other than a non-static data member.
```
jensmaurer: That feels ambiguous whether "an object other than a non-static data member" is supposed to be a valid reading.  (Note that "other than a non-static data member" might be considered to attach to "declaration".)

jensmaurer: If you really want to disambiguate, just add a bulleted list with two bullets and keep the order the same.

AlisdairM: So amended.

source/basic.tex
```diff
@@ -113,8 +113,12 @@
 \pnum
 A \defn{variable} is introduced by the
 declaration of
-a reference other than a non-static data member or of
+\begin{itemize}
+\item
+a reference other than a non-static data member, or
```
jensmaurer:
```suggestion
a reference other than a non-static data member or
```

two-pronged "or" should not have a comma

jensmaurer (changes requested):

jensmaurer:
Please squash when merging.

PR #8187, opened 2025-08-22
[basic.pre] Clarify that *declaration*s are not declarations

AlisdairM:
During our recent review of the standard, much confusion arose from mistaking the grammar term _declaration_ with the textual term "declaration" defined in [basic.pre], so add a note to avoid others spending far too much time to realize the same.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -97,6 +97,9 @@ that introduces a name\iref{dcl.type.elab},
 implicit declaration of an injected-class-name\iref{class.pre}.
 \end{itemize}
 \begin{note}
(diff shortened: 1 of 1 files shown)


source/basic.tex
```diff
@@ -97,6 +97,9 @@
 implicit declaration of an injected-class-name\iref{class.pre}.
 \end{itemize}
 \begin{note}
+The term declaration is not a synonym for a \grammarterm{declaration}\iref{dcl.pre}.
```
jensmaurer:
```suggestion
The term declaration is not a synonym for the grammar non-terminal \grammarterm{declaration}\iref{dcl.pre}.
```

Give it a bit of extra punch?

PR #8190, opened 2025-08-25
[optional.ref.ctor] Format if statement

hewillk:


eisenwave:
I'm not opposed to adding a line break there from a readability standpoint, but to my knowledge, we don't have any fixed code style when it comes to if statements.

Generally speaking, code formatting is up to the author, with some exceptions, such as item declarations in the library.

jensmaurer:
We do try to impose some consistency on the library clauses.  In former times, compactness was an argument (because ISO charges per page of the standard), but we're way beyond the cap for that anyway.

@jwakely , do you want the linebreaks added here for overall code style consistency?

tkoeppe:
At this point, I would rather more worry about readability, and also suggest braces. Disentanging braceless if statements seems more mental burden than just showing the structure clearly in the first place (without requiring the reader to also be a compiler).

PR #8192, opened 2025-08-26
[basic.def] Turn list of examples into a note

AlisdairM:
The list of example side effects should neither be deemed normtaive nor exhaustive (although we will try).  It should be demoted to a note.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -173,7 +173,9 @@ A declaration of an entity $X$ is
 a redeclaration of $X$
 if another declaration of $X$ is reachable from it\iref{module.reach};
 otherwise, it is a \defnadj{first}{declaration}.
-A declaration may also have effects including:
+
+\begin{note}
+A declaration can also have effects including:
 \begin{itemize}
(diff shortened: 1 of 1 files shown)


source/basic.tex
```diff
@@ -173,14 +173,17 @@
 a redeclaration of $X$
 if another declaration of $X$ is reachable from it\iref{module.reach};
 otherwise, it is a \defnadj{first}{declaration}.
-A declaration may also have effects including:
+
+\begin{note}
+A declaration might also have effects including:
```
jensmaurer: We can't have "might" in a note, says ISO/CS.

"can" would work.

AlisdairM: Thanks, I blanked on what the correct substitute for "may" was.

tkoeppe: Looks good -- @jensmaurer?

jensmaurer: ack.

jensmaurer (approved):

PR #8198, opened 2025-09-01
[istream.sentry] Minor tweaks

cpplearner:
1. Change "`noskipws` is zero" to "`noskipws` is `false`". The latter is more conventional since `noskipws` is a `bool`.
2. Rename the variable `ctype` to `ct`. This prevents it from shadowing the class template `std::ctype`.
3. Remove redundant comparison between the result of `ct.is(…)` (which has type `bool`) and `0`.
4. Add a semicolon to make the sentence structure more clear.

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -4585,13 +4585,11 @@ the function calls
 \tcode{setstate(failbit | eofbit)}
 (which may throw
 \tcode{ios_base::failure}).
+After any preparation is completed, \exposid{ok_} is set to the value of \tcode{is.good()}.
 
 \pnum
 \remarks
-The constructor
-\begin{codeblock}
-explicit sentry(basic_istream& is, bool noskipws = false)
-\end{codeblock}
+This constructor
 uses the currently imbued locale in \tcode{is},
 to determine whether the next input character is
 whitespace or not.
@@ -4606,13 +4604,6 @@ if (ct.is(ct.space, c))
 \end{codeblock}
 
 \pnum
-If, after any preparation is completed,
-\tcode{is.good()}
-is
-\tcode{true},
-\tcode{\exposid{ok_} != false};
-otherwise,
-\tcode{\exposid{ok_} == false}.
 During preparation, the constructor may call
 \tcode{setstate(failbit)}
 (which may throw
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -4588,8 +4588,8 @@ the function calls
 
 \pnum
 The currently imbued locale in \tcode{is} is used
-to determine whether or not the next input character is whitespace.
-The character \tcode{c} is a whitespace character if
+to determine whether the next input character is whitespace:
+The character \tcode{c} is a whitespace character if and only if
 \tcode{ct.is(ct.space, c)} is \tcode{true},
 where \tcode{ct} is \tcode{use_facet<ctype<charT>>(is.getloc())}.
 
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -4571,7 +4571,7 @@ This will be possible only in functions
 that are part of the library.
 The semantics of the constructor used in user code is as specified.
 \end{footnote}
-If \tcode{noskipws} is zero and
+If \tcode{noskipws} is \tcode{false} and
 \tcode{is.flags() \& ios_base::skipws}
 is nonzero, the function extracts and discards each character as long as
 the next available input character \tcode{c} is a whitespace character.
@@ -4600,8 +4600,8 @@ whitespace or not.
 To decide if the character \tcode{c} is a whitespace character,
 the constructor performs as if it executes the following code fragment:
 \begin{codeblock}
-const ctype<charT>& ctype = use_facet<ctype<charT>>(is.getloc());
-if (ctype.is(ctype.space, c) != 0)
+const ctype<charT>& ct = use_facet<ctype<charT>>(is.getloc());
+if (ct.is(ct.space, c))
   // \tcode{c} is a whitespace character.
 \end{codeblock}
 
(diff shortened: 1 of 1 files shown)


source/iostreams.tex
```diff
@@ -4610,7 +4610,7 @@
 \tcode{is.good()}
 is
 \tcode{true},
-\tcode{\exposid{ok_} != false}
+\tcode{\exposid{ok_} != false};
```
eisenwave:
```suggestion
\tcode{\exposid{ok_}} is \tcode{false};
```

The `== false` below should also be changed for local consistency; then we have local consistency about using "is `true`"/"is `false`" in this `\itemdescr`. Not fixing this would be silly if we're already touching this line.

tkoeppe: The original wording is awfully terse. Can we spell this out a bit more, like:

"""
If, after any preparation is completed, `is.good()` is `true`, then `ok_` is `true`, otherwise `ok_` is `false`.
"""

Or could we deunpack this even further and say something like "the value of `ok_` is `is.good()`."?

@jwakely?

jwakely: Well we definitely can't change "<code><i>ok_</i> != false</code>" to "<code><i>ok_</i></code> is `false`"!

I agree with tkoeppe's suggestion, including adding "then" after the third comma.

We could also just say:

> After any preparation is completed, _`ok_`_ is set to the value of `is.good()`.

Just saying `ok_ = is.good()` seems much simpler than describing the following logic, which would never pass code review:

```
if (is.good() == true)
  ok_ = !false;
else
  ok_ = false;
```

jwakely: Anything would be better than what's in the draft today ;)

source/iostreams.tex
```diff
@@ -4600,8 +4600,8 @@
 To decide if the character \tcode{c} is a whitespace character,
```
timsong-cpp: While we are here, we can also drop the quote of the constructor signature in the previous paragraph.

Like...it's a Remarks for this constructor. Do we really need to say yet again that we are talking about this constructor?

source/iostreams.tex
timsong-cpp: Also, this sentence should probably be part of the effects instead of remarks?

source/iostreams.tex
```diff
@@ -4600,19 +4598,12 @@
...
 the constructor performs as if it executes the following code fragment:
 \begin{codeblock}
-const ctype<charT>& ctype = use_facet<ctype<charT>>(is.getloc());
-if (ctype.is(ctype.space, c) != 0)
+const ctype<charT>& ct = use_facet<ctype<charT>>(is.getloc());
+if (ct.is(ct.space, c))
   // \tcode{c} is a whitespace character.
 \end{codeblock}
```
jwakely: I think these are good improvements so far.

But these two paragraphs are saying the same thing, so I think it would be even more clear (and shorter) if we combined them:

```suggestion
The currently imbued locale in \tcode{is} is used
to determine whether or not the next input character is whitespace.
The character \tcode{c} is a whitespace character if
\tcode{ct.is(ct.space, c)} is \tcode{true},
where \tcode{ct} is \tcode{use_facet<ctype<charT>>(is.getloc())}.
```

source/iostreams.tex
```diff
@@ -4585,13 +4585,11 @@
 \tcode{setstate(failbit | eofbit)}
 (which may throw
 \tcode{ios_base::failure}).
+After any preparation is completed, \exposid{ok_} is set to the value of \tcode{is.good()}.
```
jwakely: As this describes the final step, I think it would make more sense to be right at the end, after the "During preparation" sentence (maybe as a separate paragraph).

That way the structure would be:

- The function extracts and discards each whitespace character.
- If extracting a character returns eof, set `failbit|eofbit`.
- Whitespace is determined using `ctype`.
- The constructor may set `failbit` and throw. _(for ... reasons? I guess the footnote covers that)_
- After any preparation is completed, set `ok_`.

What do you think?

jwakely (changes requested):

source/iostreams.tex
tkoeppe: Thanks @cpplearner and @jwakely for working on this -- this wording can definitely do with improvements, this is great!

source/iostreams.tex
cpplearner: IMO this makes sense if we drop this `\remarks` macro.

https://github.com/cplusplus/draft/blob/61aaf2ef8a80eda636b00df5e1d2c823af5150a4/source/iostreams.tex#L4591

jwakely: Ah yes, agreed.

jwakely (approved):
I think this is a big improvement, thanks

source/iostreams.tex
```diff
@@ -4587,32 +4587,13 @@
...
-\pnum
-To decide if the character \tcode{c} is a whitespace character,
-the constructor performs as if it executes the following code fragment:
-\begin{codeblock}
-const ctype<charT>& ctype = use_facet<ctype<charT>>(is.getloc());
-if (ctype.is(ctype.space, c) != 0)
-  // \tcode{c} is a whitespace character.
-\end{codeblock}
+The currently imbued locale in \tcode{is} is used
+to determine whether or not the next input character is whitespace.
```
tkoeppe: "Whether" already means "whether or not", so can we drop the "or not"?

On the other hand, "if" does _not_ already mean "only if", so can we say "if and only if" below?

And maybe we can also change the full stop after "is whitespace" to a colon to signal that what follows is the explanation of what whitespace is?

source/iostreams.tex
```diff
@@ -4625,6 +4606,9 @@
 \indextext{implementation-dependent}%
 implementation-dependent operations.
 \end{footnote}
+
+\pnum
+After any preparation is completed, \exposid{ok_} is set to the value of \tcode{is.good()}.
```
tkoeppe, line 4611: This is great, thanks!

PR #8202, opened 2025-09-01
[meta.reflection.queries] Add \tcode for explicit

Andreas-Krug:


jensmaurer (commented):
I don't think this helps. Using \tcode means we refer to the keyword (sort-of), and the keyword may be present even if the function doesn't end up being explicit:

`explicit(false) C(int);`

Is this function declared explicit?  Is it declared `explicit`?

PR #8209, opened 2025-09-03
[temp.deduct.general] Replace "nontype template argument" with "constant template argument"

eisenwave:
I guess this one flew under the radar because it's not spelled "non-type" template argument.


eisenwave:
@jensmaurer do we want to put down the last "non-type", or do we want to keep it as a pet?

jensmaurer:
I don't like pets.

PR #8210, opened 2025-09-04
Typo layout_left::mapping::mapping precondition.

1uc:
Here `other` is a layout mapping and layout mappings don't have this API to access the `i`th extent. They only have `extents()` to get the std::extents object. See, https://github.com/cplusplus/draft/blob/fbd1a707daf2381067fa070a902e1743a12fa55f/source/containers.tex#L22446

for the analogous statement in layout_right.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -22117,7 +22117,7 @@ then \tcode{Extents::static_extent(0)} equals
 \begin{itemize}
 \item
 If \tcode{extents_type::rank() > 1} is \tcode{true},
-then \tcode{other.stride(1)} equals \tcode{other.extents(0)}.
+then \tcode{other.stride(1)} equals \tcode{other.extents().extent(0)}.
 \item
 \tcode{other.required_span_size()} is representable as
 a value of type \tcode{index_type}.


1uc:
CI says, that now a line is too long. In hindsight, that does seem very likely, I'll fix it.

1uc:
I've add the affected section to the commit message and added the \newline. The affected page has been attached it's (11.1).

[p1138.pdf](https://github.com/user-attachments/files/22560115/p1138.pdf)

1uc:
I didn't know where to break so I searched the rest of the document; and I found a case where it broke this way.

source/containers.tex
```diff
@@ -22117,7 +22117,7 @@
 \begin{itemize}
 \item
 If \tcode{extents_type::rank() > 1} is \tcode{true},
-then \tcode{other.stride(1)} equals \tcode{other.extents(0)}.
+then \tcode{other.stride(1)} equals \tcode{other.extents().\newline extent(0)}.
```
tkoeppe, line 22120: Could we hyphenate "extents" instead (with `\-`)?

eisenwave:
@1uc sorry, but your fork got detached because the draft repo was made temporarily private.

Could you open a ticket at https://support.github.com/ (select the `Reroute a fork` option) to have your fork reattached? I believe you wouldn't be able to continue working on this PR otherwise, and would need to reopen it.

tkoeppe:
I took and applied this change as 2a7f36abe6e93e9c5f750df000ab4ea781580d39. Thank you!

PR #8213, opened 2025-09-05
[lib.intro, iterator.concept.inc, concept.regularinvocable] Avoid "annotation"; hyphenate "freestanding-deleted"

Quuxplusone:
Now that many sections use "annotation" to mean a specific core-language construct, we should avoid using "annotation" to mean anything but that construct, i.e. stop using it to mean "code comment." (There will be an NB comment to this effect.)

Drive-by hyphenate "freestanding-deleted function". Not all freestanding-deleted functions are freestanding functions, and not all freestanding-deleted functions are deleted functions; hyphenating helps the reader understand that this is a term of art that cannot meaningfully be broken down into its component notions.

### DIFF
--- a/source/concepts.tex
+++ b/source/concepts.tex
@@ -74,8 +74,8 @@ be valid for that object.
 
 \pnum
 Expressions declared in a \grammarterm{requires-expression} in the library clauses are
-required to be equality-preserving, except for those annotated with the comment
-``not required to be equality-preserving.'' An expression so annotated
+required to be equality-preserving, except for those marked with the comment
+``not required to be equality-preserving.'' An expression marked with such a comment
 may be equality-preserving, but is not required to be so.
 
 \pnum
@@ -1261,7 +1261,7 @@ The \tcode{invoke} function call expression shall be
 equality-preserving\iref{concepts.equality} and
 shall not modify the function object or the arguments.
 \begin{note}
-This requirement supersedes the annotation in the definition of
+This requirement supersedes the comment in the definition of
 \libconcept{invocable}.
 \end{note}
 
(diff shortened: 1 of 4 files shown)


source/concepts.tex
```diff
@@ -75,7 +75,7 @@
 \pnum
 Expressions declared in a \grammarterm{requires-expression} in the library clauses are
 required to be equality-preserving, except for those annotated with the comment
-``not required to be equality-preserving.'' An expression so annotated
+``not required to be equality-preserving.'' An expression annotated with such a comment
```
jwakely: We could avoid annotated entirely:
```suggestion
``not required to be equality-preserving.'' An expression marked with such a comment
```

jwakely (approved):

source/concepts.tex
jensmaurer: Or "so marked"

And line 77 has an "annotated"; that should probably be "marked", too.

jensmaurer (commented):
The rest looks good.

source/concepts.tex
Quuxplusone: Updated to replace "annotated" with "marked". I still think "marked with such a comment" is clearer than "so marked", so I left that.

jensmaurer (approved):

eisenwave:
Some of these changes have been implemented by https://github.com/cplusplus/draft/pull/8373 as requested by https://github.com/cplusplus/nbballot/issues/707

Unfortunately, we have a merge conflict now.

Quuxplusone:
@eisenwave: "Oops." Note that I was the one who submitted both the NB comment and this PR, and this PR has been sitting approved for two months now — I assume the intent was to use this PR to resolve the matching NB comment. So I _suggest_ the simplest fix is to revert your patch first, and then merge this one, and then mark the NB comment as resolved.

However, if you prefer not to revert your patch, then I can rebase this one after this week is over.

tkoeppe:
Yes, please rebase.

PR #8214, opened 2025-09-05
[flat.map.modifiers] Remove redundancy in 'insert(sorted_unique, i, j)'

jwakely:
We can specify this in terms of the overload without the `sorted_unique` tag. That is consistent with how the equivalent functions in `flat_set` and `flat_multiset` are specified.

LWG asked for this in the telecon today.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -17792,32 +17792,11 @@ template<class InputIterator>
 \begin{itemdescr}
 \pnum
 \effects
-Adds elements to \exposid{c} as if by:
-\begin{codeblock}
-for (; first != last; ++first) {
-  value_type value = *first;
-  @\exposid{c}@.keys.insert(@\exposid{c}@.keys.end(), std::move(value.first));
-  @\exposid{c}@.values.insert(@\exposid{c}@.values.end(), std::move(value.second));
-}
-\end{codeblock}
-Then, merges the sorted range of newly added elements and
-the sorted range of pre-existing elements into a single sorted range; and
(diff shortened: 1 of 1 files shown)


jwakely:
To give some more context: the wording for `insert(i, j)` is not word-for-word identical to the wording for `insert(sorted_unique, i, j)` because the latter has the additional text "sorts the range of newly inserted elements with respect to `value_comp()`;" which implies it does more work.

However, if the contents of the range [i, j) is already sorted then that step is a no-op. The implementation can elide that step, because it doesn't say the effects are equivalent to actually calling `ranges::sort`, it just says "sorts the range". If the implementation happens to know it's already sorted, it can skip that step.

So specifying `insert(sorted_unique, i, j)` in terms of `insert(i, j)` with a more restrictive complexity guarantee completely described the effects, and specifies the complexity.

And [flat.set.modifiers] p8 and [flat.multiset.modifiers] p7 already take exactly this specification approach, just deferring to `insert(i, j)` instead of duplicating 95% of its wording.

jwakely:
And [structure.specifications] p4 says that the _Complexity_: element here overrules the implied _Complexity_: that we would otherwise have from using "_Effects_: Equivalent to ..."

tkoeppe:
Thanks for the explanations! Yes, it certainly seems technically correct, and a good deduplication. The more restrictive complexity may be a bit subtle, but then again, the Standard is not a tuturial, and implementers will know how to read this.

PR #8215, opened 2025-09-06
[flat.multiset.defn] Fix format

hewillk:
It's weird to put `constexpr` after the return type, and it's also inconsistent with other places.


eisenwave:
Yeah, this is definitely an anomaly. I've already proposed an automatic check that would spot these things in #8085 ... or actually, nvm, it wouldn't have caught this one because it's a swap of return type and `constexpr`, but `constxpr` still comes after `friend`.

PR #8218, opened 2025-09-09
[meta.type.synop] Remove redundant cast in constant_wrapper declaration

jwakely:
The use of `decltype(cw-fixed-value(X))` instead of just `decltype(X)` is a workaround for a GCC bug: https://gcc.gnu.org/PR117392

There's no need for the standard to specify it this way.

The `typename` is redundant too, but that's a separate topic.

### DIFF
--- a/source/meta.tex
+++ b/source/meta.tex
@@ -169,7 +169,7 @@ namespace std {
   template<class T>
     struct @\exposidnc{cw-fixed-value}@;                                      // \expos
 
-  template<@\exposidnc{cw-fixed-value}@ X, class = typename decltype(@\exposid{cw-fixed-value}@(X))::@\exposid{type}@>
+  template<@\exposidnc{cw-fixed-value}@ X, class = typename decltype(X)::@\exposid{type}@>
     struct constant_wrapper;
 
   template<class T>


jwakely:
I've re-pushed this with a commit message that matches the NB comment format.

jensmaurer:
Sorry, but you're confused about the commit message format.

We want the traditional subject line unmodified (label + short text), and in the body of the commit message, we want "Fixes NB US 79-146 (C++26 CD)".

See fb46e16c6ce0d91c6bd21ca497d1823c997d1926 for an example.

jwakely:
I was going by what it says here: 

https://github.com/cplusplus/draft/wiki/Commit-message-format#commits-from-committee-papers

I'll fix it tomorrow

jensmaurer:
Ah, we've never done it that way.  I've fixed the wiki: https://github.com/cplusplus/draft/wiki/Commit-message-format#nb-comment-resolutions

Sorry for the confusion.

jwakely:
Should be fixed now.

Do we want the "Fixes cplusplus/nbballot/..." part in the pull request description, or do we not want the related nbballot issue to be auto-closed?

jensmaurer:
Auto-closing the nbballot issue is good; the "Fixes" establishes a link for later archeology.

jensmaurer (approved):

PR #8219, opened 2025-09-10
[func.wrap.ref.class] Use `ArgTypes` instead of `Args`

hewillk:
Use `ArgTypes` instead of `Args` to match the current class signature.
Also, `R (*thunk-ptr)(BoundEntityType, Args&&...)` should be `ArgTypes...` since there is no `Args...`.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -14960,7 +14960,7 @@ namespace std {
     template<class... T>
       static constexpr bool @\exposidnc{is-invocable-using}@ = @\seebelownc@;     // \expos
 
-    R (*@\exposidnc{thunk-ptr}@)(@\exposidnc{BoundEntityType}@, Args&&...) noexcept(@\placeholdernc{noex}@);  // \expos
+    R (*@\exposidnc{thunk-ptr}@)(@\exposidnc{BoundEntityType}@, ArgTypes&&...) noexcept(@\placeholdernc{noex}@);  // \expos
     @\exposidnc{BoundEntityType}@ @\exposidnc{bound-entity}@;                               // \expos
   };
 
@@ -14976,7 +14976,7 @@ namespace std {
 
 \pnum
 An object of class
-\tcode{function_ref<R(Args...) \cv{} noexcept(\placeholder{noex})>}
+\tcode{function_ref<R(ArgTypes...) \cv{} noexcept(\placeholder{noex})>}
 stores a pointer to function \exposid{thunk-ptr} and
 an object \exposid{bound-entity}.
 \exposid{bound-entity} has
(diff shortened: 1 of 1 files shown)


hewillk:
Need a typesetter to _suggest_ to me how to break the sentences..

tkoeppe:
I'd mildly prefer the simpler change in that case to only fix the thing that's broken. We don't need the additional verbosity.

hewillk:
> I'd mildly prefer the simpler change in that case to only fix the thing that's broken. We don't need the additional verbosity.

I don't think this is verbosity; I think it's an enhancement. Because it matches the exact signature of the function, which I did intentionally.

Other places in [func.wrap] named `Args...` are template parameters for constructing internal callable objects, not template parameters for `operator()`. 

The `function`, `copyable_function`, `move_only_function`, and `function_ref` all consistently use the template name `ArgTypes...`.

source/utilities.tex
```diff
@@ -14976,15 +14976,15 @@
 
 \pnum
 An object of class
-\tcode{function_ref<R(Args...) \cv{} noexcept(\placeholder{noex})>}
+\tcode{function_ref<R(ArgTypes...) \cv{} noexcept(\placeholder{noex})>}
 stores a pointer to function \exposid{thunk-ptr} and
 an object \exposid{bound-entity}.
 \exposid{bound-entity} has
-an unspecified trivially copyable type \exposid{BoundEntityType}, that
+an unspecified trivially copyable type \linebreak{} \exposid{BoundEntityType}, that
```
eisenwave, line 14983: By the way, please don't add forced line breaks like this.
```suggestion
an unspecified trivially copyable type \exposid{Bound\-Entity\-Type}, that
```
... would presumably work, assuming you were coping with overful hboxes due to `BoundEntityType` going off the side of the page.

tkoeppe: Let me at least split off the bug fix from the discretionary tweaks then.

source/utilities.tex
```diff
@@ -14995,7 +14995,7 @@
 Within \ref{func.wrap.ref},
 \tcode{\placeholder{call-args}} is an argument pack with elements such that
 \tcode{decltype((\placeholder{call-args}\linebreak{}))...} denote
-\tcode{Args\&\&...} respectively.
+\tcode{ArgTypes\&\&...} respectively.
```
tkoeppe, line 14998: @jwakely: this change is a bug fix, too, right? Here we're again referring to the class template synopsis.

tkoeppe:
> I don't think this is verbosity; I think it's an enhancement. Because it matches the exact signature of the function, which I did intentionally.

One could also see this the other way round in that reusing the name for a local use in a new construct "shadows" the name from the ambient context... I'm tempted to just leave that as is.

The bug fixes were of course very important; thank you for reporting!

source/utilities.tex
jwakely: ah yes, that's right

tkoeppe: Thanks!

PR #8220, opened 2025-09-10
[optional.optional.general], [optional.optional.ref.general] Use "object of type optional<T&>" correctly

eisenwave:
The current wording in p1 makes no sense to me. Firstly, it says

> An object of `optional<T&>` ...

https://github.com/cplusplus/papers/issues/1661 says

> An instance of `optional<T&>`

https://github.com/cplusplus/draft/pull/7979 editorially replaced "instance" with "object" (accidentally? I could not find discussion on this), and now this is grammatically wrong. The wording idiom is

> An object of <u>type</u>

The second sentence is also inconsistent with the paper. The paper once again uses the term "instance", but the draft wording just says

> When an `optional<T&>`

This is also incorrect because it's a category error. We use this idiom like "an `array`" or "a `vector`" when referring to types, not to objects.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -3365,8 +3365,10 @@ namespace std {
 \end{codeblock}
 
 \pnum
-Any instance of \tcode{optional<T>} at any given time either contains a value or does not contain a value.
-When an instance of \tcode{optional<T>} \defnx{contains a value}{contains a value!\idxcode{optional}},
+An object of type \tcode{optional<T>} at any given time
+either contains a value or does not contain a value.
+When an object of type \tcode{optional<T>}
+\defnx{contains a value}{contains a value!\idxcode{optional}},
 it means that an object of type \tcode{T}, referred to as the optional object's \defnx{contained value}{contained value!\idxcode{optional}},
 is nested within\iref{intro.object} the optional object.
 When an object of type \tcode{optional<T>} is contextually converted to \tcode{bool},
@@ -3374,7 +3376,7 @@ the conversion returns \tcode{true} if the object contains a value;
 otherwise the conversion returns \tcode{false}.
 
 \pnum
-When an \tcode{optional<T>} object contains a value,
+When an object of type \tcode{optional<T>} contains a value,
 member \tcode{val} points to the contained value.
 
 \pnum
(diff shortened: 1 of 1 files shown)


jwakely:
> The current wording in p1 makes no sense to me. Firstly, it says
> 
> > An object of `optional<T&>` ...
> 
> [cplusplus/papers#1661](https://github.com/cplusplus/papers/issues/1661) says
> 
> > An instance of `optional<T&>`

Where?

> #7979 editorially replaced "instance" with "object" (accidentally? I could not find discussion on this), and now this is grammatically wrong.

The paper says "An object"

> The second sentence is also inconsistent with the paper.

No it isn't, are you looking at the wrong paper? I think you're looking at an old revision.


> The paper once again uses the term "instance", but the draft wording just says
> 
> > When an `optional<T&>`

That's what the paper says.

> This is also incorrect because it's a category error. We use this idiom like "an `array`" or "a `vector`" when referring to types, not to objects.

We use it both ways, e.g. "Effects: Constructs an empty `vector`, using the specified allocator."

eisenwave:
Hmm yeah, I was in a hurry when I drafted this and looked at the pre-existing wording in https://eel.is/c++draft/optional.optional.general#1, which uses the "instance" term.

The paper was correctly applied, but

> An object of `optional<T&>`

seems bad regardless.

While we're on the subject, it's probably best to remove that weird asymmetry. We shouldn't be talking about "instances" for `optional<T>` but "objects" for `optional<T&>`.

If there's precedent, the second "an `optional<T&>`" seems fine then, but once again, that should be consistent across the wording for the primary template and the partial specialization.

jwakely:
Yes, feel free to include the pre-existing "instance of `optional<T>`" in this PR.

tkoeppe:
This current change looks fine now, but I wonder if we have or want some general policy on the use of the phrase "an instance of `T`" --  we are using that in several places at the moment, but I'm not seeing anything particularly wrong with that. I agree that "object of `T`" or "`T` object" isn't great and would better say "object of type `T`" (as in this change), but I'm not sure this needs to be some global rule. (E.g. #8221 tries something like that but seems unmotivated to me.)

jwakely:
We don't have any such policy, and I agree that there's nothing very wrong with "an instance of X". We definitely want to fix "An object of `optional<T&>`" here, and "when an `optional<T>` object" isn't great, but I don't feel strongly whether we should go with "An object of type X" over "An instance of X". Either seems OK, so taking this change because it's the one that's been offered seems fine.

tkoeppe:
Agreed, thanks!

PR #8225, opened 2025-09-11
[temp.explicit] Fix the comment in Example 2

### LINKED ISSUE #8224: [temp.explicit] Incorrect comment in Example 2
In the following comment:

https://github.com/cplusplus/draft/blob/bed3aab4b763ee5077d5cab997c020e389fc8092/source/templates.tex#L7018-L7019

`int` and `auto` should be swapped to match the code. It should read:  variable with type `auto` can be redeclared with type `int`.

ghost:
Fixes #8224.


tkoeppe:
(Thanks! Something seems stuck here with the action runners...)

PR #8235, opened 2025-09-16
[diff.class] Move text into note

Andreas-Krug:


jensmaurer:
This needs more work; the example involving declaring `struct Y` outside of `struct X` is totally confused. Maybe the note wants to be struck entirely, given we've already said "scope rules" about 13 times in that section.

PR #8243, opened 2025-09-22
Fix typo in submdspan_extents.

1uc:
The pack `Extents` must be a pack of `size_t` integers.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -21080,7 +21080,7 @@ namespace std {
   struct full_extent_t { explicit full_extent_t() = default; };
   inline constexpr full_extent_t full_extent{};
 
-  template<class IndexType, class... Extents, class... SliceSpecifiers>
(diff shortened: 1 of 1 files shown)


jwakely (changes requested):
This mistake was in the incoming paper, but I agree it has to be a pack of `size_t` because otherwise `extents<IndexType, Extents...>` is ill-formed. But it needs to be changed in the [ndspan.syn] synopsis too.

@mhoemmen

1uc:
That makes sense. I've added a commit to address the issue. If you don't want to do a "squash and merge", let me know and I'll force push a single commit.

1uc:
I've updated the commit message to mention [view.multidim] and squashed the two commits together.

mhoemmen:
@jwakely I was traveling and just got back -- is everything OK now?

jwakely:
Yes, I think this change was correct

PR #8247, opened 2025-09-23
[const.wrap.class] Add missing namespace std

hewillk:


jensmaurer:
Selecting "side-by-side diff" and "hide whitespace changes" makes the review here manageable.

This looks good to me.

tkoeppe:
Thanks, @jensmaurer, good tip! And thanks @hewillk for the fix!

PR #8248, opened 2025-09-24
Fix typo in submdspan `Effect:`.

1uc:
The class `mdspan` doesn't have a member `data`, it has `data_handle` (which returns a generalized pointer). This "data handle" is what's passed, together with an offset, to the accessor to "shift the pointer".


source/containers.tex
```diff
@@ -26118,7 +26118,7 @@
 Equivalent to:
 \begin{codeblock}
 auto sub_map_result = submdspan_mapping(src.mapping(), slices...);
-return mdspan(src.accessor().offset(src.data(), sub_map_result.offset),
+return mdspan(src.accessor().offset(src.data_handle(), sub_map_result.offset),
               sub_map_result.mapping,
               AccessorPolicy::offset_policy(src.accessor()));
```
```suggestion
              typename AccessorPolicy::offset_policy(src.accessor()));
```

These corrections are also in P3663 as drive-by fixes.

1uc:
I've added the `typename` change to the commit and checked the implementation: indeed I fixed that without triggering that it's a typo in the spec.

jensmaurer (approved):

PR #8250, opened 2025-09-26
[simd.overview] Remove spurious noexcept

hewillk:
This function is _defined_ without `noexcept`, which is reasonable since `gen` can throw.
Is this the editorial or the LWG?


eisenwave:
This is a wording bug in https://github.com/cplusplus/papers/issues/2097

The paper removes `noexcept` from those constructors, but makes no changes to the synopsis, and neither LWG nor editors noticed that. The intent is clearly to remove `noexcept` from the constructors of `basic_vec` and `basic_mask`, so this issue is editorial.

We should make corresponding changes to `basic_mask`, which also has a stray `noexcept` in the synopsis.

eisenwave:
Gotta fix the message on the new commit.

PR #8251, opened 2025-09-26
Fix dangling references to "naming class"

cor3ntin:
The term was renamed "designating class" by P2996

https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p2996r13.html#pnum_397


tkoeppe:
@cor3ntin Could you please fix the commit message to follow our usual scheme?

PR #8255, opened 2025-09-27
[concepts.callable.general] Replace "function objects" with "callable types"

### LINKED ISSUE #8241: [concepts.callable.general, concept.invocable] "function objects" and "callable type" are too restrictive
[concepts.callable.general] p1
> The concepts in 18.7 describe the requirements on **function objects**…

[concept.invocable] p1
> The invocable concept specifies a relationship between a **callable type**…

"Function objects" and "callable type" does not include important cases covered by `INVOKE`, such as references to functions. "**Callable types, function types, and references to functions**" (or just a "type") would be better. 

Since these are introductory paragraphs, I think this is an editorial issue.

eisenwave:
Fixes #8241.

The introductory paragraph is misleading: we claim that the subsequent concepts are for [function objects](https://eel.is/c++draft/function.objects#def:function_object,type), but the very next concept is `invocable`, which also accepts member function pointers and data member pointers, which are [callable types](https://eel.is/c++draft/func.def#def:type,callable) but not function objects.

This seems like an editorial issue, since it's just about some intro fluff, nothing of normative consequence.

### DIFF
--- a/source/concepts.tex
+++ b/source/concepts.tex
@@ -1222,8 +1222,8 @@ fundamental types like \tcode{int} and that are comparable with
 \rSec2[concepts.callable.general]{General}
 
 \pnum
-The concepts in \ref{concepts.callable} describe the requirements on function
-objects\iref{function.objects} and their arguments.
+The concepts in \ref{concepts.callable} describe the requirements on
+callable types\iref{func.def} and their arguments.
 
 \rSec2[concept.invocable]{Concept \cname{invocable}}
 


morinmorin:
Would "callable types, function types, and references to such types" be more accurate?

eisenwave:
I guess yeah. I'm pretty sure it's possible to put function references into `std::invocable`, but those are not defined to be "function objects".

That's a problem that also affects the contents of the subclause though, not just the intro paragraph.

morinmorin:
Got it, thanks for the comment! I'll create a small follow-up PR soon.

PR #8257, opened 2025-09-28
[containers] fix operator<=> format

hewillk:
They are currently not aligned:
截圖 2025-09-29 凌晨12 58 25
This fix puts `operator<=>` on a new line to align with the style used elsewhere.


eisenwave:
The change is clearly correct, but your commits should be squashed.

PR #8264, opened 2025-09-29
[mdspan.accessor.aligned.overview] Remove `std::` in example

hewillk:
We usually don't explicitly spell `std::` in examples unless it's really intentional.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -24525,19 +24525,18 @@ possibly less optimized function \tcode{compute_without_requiring_overalignment}
 that has no over-alignment requirement.
 \begin{codeblock}
 void compute_using_fourfold_overalignment(
-  std::mdspan<float, std::dims<1>, std::layout_right,
-    std::aligned_accessor<float, 4 * alignof(float)>> x);
+  mdspan<float, dims<1>, layout_right, aligned_accessor<float, 4 * alignof(float)>> x);
 
 void compute_without_requiring_overalignment(
-  std::mdspan<float, std::dims<1>, std::layout_right> x);
+  mdspan<float, dims<1>, layout_right> x);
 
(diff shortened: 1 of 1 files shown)


tkoeppe:
I tend to agree. @JeffGarland, any library thoughts?

jwakely:
> We usually don't explicitly spell `std::` in examples unless it's really intentional.

I think the opposite is true. The very first code example in the library clause ([alg.func.obj]) uses a qualified name.

jwakely:
See also [ptr.launder], [concept.swappable], [out.ptr.t], ...

We're not consistent though, and this one isn't a complete code example (it has no headers included for a start).

So I guess this is fine, it reduces a bit of visual noise.

tkoeppe:
Hm, I see, we do seem to use both styles. That makes this change less compelling. @hewillk Could you at least survey the rest of the mdspan material to see what other examples are doing?

hewillk:
> Hm, I see, we do seem to use both styles. That makes this change less compelling. @hewillk Could you at least survey the rest of the mdspan material to see what other examples are doing?

Most of the other examples do not use spelled namespaces, for example: https://eel.is/c++draft/views#mdspan.sub.sub-7

hewillk:
> See also [ptr.launder], [concept.swappable], [out.ptr.t], ...
> 
> We're not consistent though, and this one isn't a complete code example (it has no headers included for a start).
> 
> So I guess this is fine, it reduces a bit of visual noise.

They all have include headers, which means that namespaces are required for the entire code to be compiled correctly.
So this is what I mean by “unless it's really intentional”.
I think [alg.func.obj] is also intentional, because its example is highly relevant to ADL (i.e., namespace).

tkoeppe:
Great, thank you!

PR #8279, opened 2025-10-04
[meta.reflection.annotation] Move to before [meta.reflection.extract]

jensmaurer:
Fixes NB US 87-156 (C++26 CD).

Fixes cplusplus/nbballot#735


jensmaurer:
@tkoeppe  , rebased.

PR #8307, opened 2025-10-05
[exec.run.loop] Remove code font formatting for count and state

jensmaurer:
Fixes NB US 231-361 (C++26 CD).

Fixes cplusplus/nbballot#936

### DIFF
--- a/source/exec.tex
+++ b/source/exec.tex
@@ -6487,7 +6487,7 @@ run_loop() noexcept;
 \begin{itemdescr}
 \pnum
 \ensures
-\exposid{count} is \tcode{0} and \exposid{state} is \exposid{starting}.
+Count is \tcode{0} and state is starting.
 \end{itemdescr}
 
 \indexlibrarydtor{run_loop}%
@@ -6498,7 +6498,7 @@ run_loop() noexcept;
 \begin{itemdescr}
 \pnum
 \effects
-If \exposid{count} is not \tcode{0} or if \exposid{state} is \exposid{running},
+If count is not \tcode{0} or if state is running,
 invokes \tcode{terminate}\iref{except.terminate}.
 Otherwise, has no effects.
 \end{itemdescr}
@@ -6515,13 +6515,13 @@ Otherwise, has no effects.
 Blocks\iref{defns.block} until one of the following conditions is \tcode{true}:
 \begin{itemize}
 \item
-\exposid{count} is \tcode{0} and \exposid{state} is \exposid{finishing},
-in which case \exposid{pop-front} sets \exposid{state} to \exposid{finished}
+Count is \tcode{0} and state is finishing,
+in which case \exposid{pop-front} sets state to finished
 and returns \tcode{nullptr}; or
 \item
-\exposid{count} is greater than \tcode{0},
+count is greater than \tcode{0},
 in which case an item is removed from the front of the queue,
-\exposid{count} is decremented by \tcode{1}, and
+count is decremented by \tcode{1}, and
 the removed item is returned.
 \end{itemize}
 \end{itemdescr}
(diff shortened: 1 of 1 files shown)


jensmaurer:
@jwakely , having exposition-only data members and an exposition-only enumeration might be the better outcome here, but that leaves the atomic nature of count and state a bit out in the rain.

jwakely:
It seems a little unnatural to refer to "count" and "state" this way, when they _aren't_ actual variables. We could rephrase p2 in terms of having a counter, and then  refer to "the counter" and "the associated state" rather than just "count" and "state". And maybe capitalise the state names, Starting etc.

But I don't feel strongly about that suggestion.

tkoeppe:
I fully agree that this is now somewhat unnatural. I appreciate that we don't have actual expos-only variables and that the current state is broken in that regard, but the phrasing we use only really works when talking about variables. If we're instead talking about some general concept that was introduced locally, I think the phrasing needs to be adapted.

The terms are introduced in https://eel.is/c++draft/exec.run.loop.general#2: "A `run_loop` instance has an associated _count_ that  [...]". We could refer to this "the instance's count" and "the instance's state", say, or "the {count,state} of the `run_loop` [object]", or in a crunch even "the {count,state} of `*this`".

I think I mildly prefer the first version, but I'd like to hear other opinions. With the first version I'd also be happy to say "the instance" only once, so e.g. "The instance's count is 0 and state is starting."

Thoughts?

jwakely:
I don't love the word "instance, rather than say "the object" or "The `run_loop`" but "instance" is the word used there in the introductory prose, so it would be consistent.

I like "The instance's count" a lot more than just "Count".

tkoeppe:
Just "the `run_loop`" would be nice, but pedants might complain that that's a class and not an object. I'd be happy with "the object", too. But can we try it out in context? "The count of the object is 0 and the state is starting."? "The object's count is 0 and its state is starting."?

jwakely:
> "The object's count is 0 and its state is starting."?

I don't hate it, but I don't love it either.

Another option would be "The `run_loop` instance's count is 0 and its state is starting" which is verbose but matches the intro wording and should be pedant-proof?

There may be no perfect answer here, we can just pick on that's better than the status quo.

tkoeppe:
> Another option would be "The `run_loop` instance's count is 0 and its state is starting" which is verbose but matches the intro wording and should be pedant-proof?

Yes, I'd be fine with that, and also don't hate it. But does it scale? Will we say that same thing for every itemdescr?

tkoeppe:
@jensmaurer Could you perhaps update the PR to use that wording Jon suggested?

jensmaurer:
@tkoeppe , I've done that.

tkoeppe:
I think that looks very nice, thank you! @jwakely, what do you think?

source/exec.tex
```diff
@@ -6564,13 +6566,13 @@
...
+The \tcode{run_loop} instance's state is either starting or finishing.
 
 \pnum
 \effects
-If \exposid{state} is \exposid{starting},
-sets the \exposid{state} to \exposid{running},
-otherwise leaves \exposid{state} unchanged.
+If the \tcode{run_loop} instance's state is starting,
+sets its state to running,
+otherwise leaves its state unchanged.
```
jwakely: Line 6523 refers to "the state" after the first reference to it, but here we say "its state". I have a weak preference for "the state", because we've already specified which state we mean (the one belonging to the `run_loop`  instance).

Otherwise, this looks good to me.

tkoeppe: @jensmaurer ^^^

jensmaurer: Fixed, rebased, force-pushed.

tkoeppe: Good call, Jon, thanks!

PR #8322, opened 2025-10-05
[exec.snd.expos] Reorder specification immediately after declaration

jensmaurer:
Fixes NB US 213-353 (C++26 CD).

Fixes cplusplus/nbballot#928

### DIFF
--- a/source/exec.tex
+++ b/source/exec.tex
@@ -1731,20 +1731,6 @@ namespace std::execution {
   concept @\defexposconceptnc{completion-tag}@ =                                      // \expos
     @\libconcept{same_as}@<Tag, set_value_t> || @\libconcept{same_as}@<Tag, set_error_t> || @\libconcept{same_as}@<Tag, set_stopped_t>;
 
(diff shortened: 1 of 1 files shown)


tkoeppe:
This is just moving paragraphs around in the same subclause, right? What about the suggestion from the comment issue to introduce more subclauses? Should we do both, or aim straight for that bigger change (maybe to be worked out next week)?

jensmaurer:
Well, we can certainly try to add more subclauses, too.

tkoeppe:
Mostly for the reasons I mentioned in the other issue: it can get rather confusing to have itemdecls/itemdescrs mixed with regular paragraphs, and it's usually cleaner if those appear in subclauses of their own.

But if there isn't an obvious and clear subdivision, we can also defer that or discuss later.

tkoeppe:
Ahh, I forgot you already made this pull request. Robert has confirmed that it's best to just split up the code block and not bother with introducing subdivisions, so we can do that.

PR #8323, opened 2025-10-06
[hive.capacity], [vector.capacity] Move restrictions from Complexity to Remarks

### LINKED ISSUE #8263: [hive.capacity], [vector.capacity] Function effects incorrectly specified in Complexity
https://github.com/cplusplus/draft/blob/020140713799934d623241627c6aaef917e5039d/source/containers.tex#L8470-L8473

This is a bizarre bit of markup. Whether `reserve()` alters the size of the container clearly doesn't belong in a *Complexity* specification. The first part of the sentence should be moved down into *Remarks* or something.

eisenwave:
Fixes #8263.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -10173,11 +10173,11 @@ can throw an appropriate exception.
 
 \pnum
 \complexity
-It does not change the size of the sequence and takes at most linear
-time in the size of the sequence.
+Linear in the size of the sequence.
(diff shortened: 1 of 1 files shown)


tkoeppe:
Seems reasonable (except that I wouldn't call the specification "restrictions"). @jwakely any objections?

PR #8327, opened 2025-10-07
[allocator.requirements.general] Add namespace std for exposition-only concept

hewillk:


jwakely (commented):
We don't do this elsewhere for exposition-only helpers, do we?

hewillk:
Maybe. But this is only to consistent with [expos.only.entity] which also belongs to [library].

tkoeppe:
Hm, it's true that we put the expos-only entities in [library] in namespace std. It's only in the specific library facilities where locally used expos-only entities are declared without namespace.

Ultimately I don't think it matters for the normative effect, but I can see an consistency argument for the present case.

jwakely:
Yeah, i don't mind either way.

PR #8328, opened 2025-10-09
[atomics.types.float] Align parameters in function declarations

eisenwave:
Fixes https://github.com/cplusplus/nbballot/issues/891.

The parameter indentation in the subclause is broken in four different ways:
- In the first block with `exchange`, `fetch_max`, etc. it looks like `constexpr` was added without adding then necessary indentation to the second line of the declaration, resulting in misalignment.
- For `store_fmaximum` and some others, an *attempt* was made to align the parameters, but unsuccessfully.
- In the definition of `fetch_key`, there is one space too much indentation, and we forgot to use `placeholdernc` rather than `placeholder`.
- In the definition of `store_key`, we just used two-space indentation rather than trying to align parameters at all.

### DIFF
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -5416,56 +5416,56 @@ namespace std {
     @\placeholdernc{floating-point-type}@ exchange(@\placeholdernc{floating-point-type}@,
                                  memory_order = memory_order::seq_cst) volatile noexcept;
     constexpr @\placeholdernc{floating-point-type}@ exchange(@\placeholdernc{floating-point-type}@,
-                                 memory_order = memory_order::seq_cst) noexcept;
+                                           memory_order = memory_order::seq_cst) noexcept;
     bool compare_exchange_weak(@\placeholder{floating-point-type}@&, @\placeholdernc{floating-point-type}@,
                                memory_order, memory_order) volatile noexcept;
     constexpr bool compare_exchange_weak(@\placeholder{floating-point-type}@&, @\placeholdernc{floating-point-type}@,
-                               memory_order, memory_order) noexcept;
+                                         memory_order, memory_order) noexcept;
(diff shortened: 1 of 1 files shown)


source/threads.tex
```diff
@@ -5403,56 +5403,60 @@
...
+                         memory_order = memory_order::seq_cst) volatile noexcept;
+    constexpr @\placeholdernc{floating-point-type}@
+      fetch_fmaximum_num(@\placeholdernc{floating-point-type}@,
+                         memory_order = memory_order::seq_cst) noexcept;
+    @\placeholdernc{floating-point-type}@
+      fetch_fminimum_num(@\placeholdernc{floating-point-type}@,
+                         memory_order = memory_order::seq_cst) volatile noexcept;
+    constexpr @\placeholdernc{floating-point-type}@
+      fetch_fminimum_num(@\placeholdernc{floating-point-type}@,
+                         memory_order = memory_order::seq_cst) noexcept;
```
eisenwave: I had to make some pretty drastic changes in this place unfortunately.

```cpp
floating-point-type fetch_fmaximum_num(floating-point-type,
                                       memory_order = memory_order::seq_cst) volatile noexcept;
```
... wouldn't have onto the page, by a handful of characters.

tkoeppe: I think what we usually can and do do in these situations is to break immediately after the opening paren ( and then start parameters on the next line, indented by two. Could we do that here?

eisenwave: I haven't tried it, and I'm not able to for the next hours. Personally I don't like that style anyway.

eisenwave: It works, and it looks good.

tkoeppe:
Perfect, thanks!

PR #8329, opened 2025-10-09
[streambuf.virt.put] Replace "effects" with "affects" in footnote

eisenwave:
Fixes https://github.com/cplusplus/nbballot/issues/878.

Use of "effect" as a verb seems inappropriate here; see https://www.merriam-webster.com/dictionary/effect

"affect (verb)" (https://www.merriam-webster.com/dictionary/affect) means to "to act on and cause a change", which is what we're trying to express here. Consuming a character *acts upon* the output sequence.


jwakely:
Oops I forgot there was already a pr for this

PR #8331, opened 2025-10-09
[simd.syn] Reorder declarations to match subclause order

eisenwave:
Fixes NB US 175-281 (C++26 CD).

Fixes https://github.com/cplusplus/nbballot/issues/856.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -16532,10 +16532,24 @@ namespace std::simd {
   template<class T, @\exposid{simd-size-type}@ N = @\exposid{simd-size-v}@<T, @\exposid{native-abi}@<T>>>
     using @\libmember{vec}{simd}@ = basic_vec<T, @\exposid{deduce-abi-t}@<T, N>>;
 
-  // \ref{simd.mask.class}, class template \tcode{basic_mask}
-  template<size_t Bytes, class Abi = @\exposid{native-abi}@<@\exposid{integer-from}@<Bytes>>> class basic_mask;
-  template<class T, @\exposid{simd-size-type}@ N = @\exposid{simd-size-v}@<T, @\exposid{native-abi}@<T>>>
-    using @\libmember{mask}{simd}@ = basic_mask<sizeof(T), @\exposid{deduce-abi-t}@<T, N>>;
+  // \ref{simd.reductions}, \tcode{basic_vec} reductions
+  template<class T, class Abi, class BinaryOperation = plus<>>
+    constexpr T reduce(const basic_vec<T, Abi>&, BinaryOperation = {});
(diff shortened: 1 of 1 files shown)


eisenwave:
I also wrote down some thoughts on the issue at https://github.com/cplusplus/nbballot/issues/856#issuecomment-3385081981

We could also reorder the subclauses, but in my opinion, they're structured very logically. All the `basic_vec` stuff first, then `basic_mask`. I don't see any compelling reason to deviate, and making the synopsis match the subclauses is the easiest fix.

@mattkretz can you confirm that the mismatch between the synopsis and the subclauses is accidental?

source/numerics.tex
```diff
@@ -16715,46 +16729,13 @@
...
-  template<size_t Bytes, class Abi>
-    constexpr bool none_of(const basic_mask<Bytes, Abi>&) noexcept;
-  template<size_t Bytes, class Abi>
-    constexpr @\exposid{simd-size-type}@ reduce_count(const basic_mask<Bytes, Abi>&) noexcept;
-  template<size_t Bytes, class Abi>
-    constexpr @\exposid{simd-size-type}@ reduce_min_index(const basic_mask<Bytes, Abi>&);
-  template<size_t Bytes, class Abi>
-    constexpr @\exposid{simd-size-type}@ reduce_max_index(const basic_mask<Bytes, Abi>&);
-
   constexpr bool all_of(@\libconcept{same_as}@<bool> auto) noexcept;
```
eisenwave: Ooops, I think this should have also been moved as part of [simd.mask.reductions]?

mattkretz:
> @mattkretz can you confirm that the mismatch between the synopsis and the subclauses is accidental?

The current order is derived from what got published in the TS. And the TS order was probably fairly accidental. Making the synopsis match the subclause ordering (rather than the other way around) seems like the safest thing to do.

FWIW, I implement `basic_mask` before `basic_vec` because that's just more natural in terms of code dependencies. The order in my implementation (not asking for a subclause reordering):
1. [simd.expos]
2. [simd.traits]
3. [simd.iterator]
4. [simd.creation] + [simd.permute.static]
5. [simd.mask] including all hidden friends in [simd.mask.nonmembers]
6. [simd.flags]
7. [simd.class] including all hidden friends in [simd.nonmembers]
8. [simd.loadstore]
9. remaining [simd.permute.*]
10. [simd.mask.reductions]
11. [simd.reductions]
12. [simd.alg]
13. [simd.bit]
14. complex specializations of [simd.mask] and [simd.class]
15. [simd.math]

jensmaurer (approved):
@tkoeppe , we need this to go in so that, as a next step, we can adjust the subclause headings quoted in the synopsis to the new reality, without causing merge conflicts.

PR #8332, opened 2025-10-09
[meta.type.synop] Index the unindexed type aliases and variable templates

YexuanXiao:


tkoeppe:
Thanks a lot! I rebased and pushed this as dfe8e0b2de8df3cd890351055352e742b8ad5a5f.

PR #8335, opened 2025-10-11
[ub,intro.progress] Added entry for UB due to lack of forward progress

timuraudio:
Added `\ubdef{intro.progress.stops}` to [intro.progress]. Added matching entry with code example to [ub]. Note: the wording in [ub] already matches the new wording proposed in CWG2816. When the resolution in CWG2816 is applied to [intro.progress], make sure to not lose the `\ubdef` there.

### DIFF
--- a/source/ub.tex
+++ b/source/ub.tex
@@ -502,7 +502,7 @@ std::thread t1{f}, t2{f}, t3{f};
 \end{codeblock}
 \end{example}
 
-\rsec2[ub.intro.progress]{Forward progress}
+\rSec2[ub.intro.progress]{Forward progress}
 
 \pnum
 \ubxref{intro.progress.stops} \\
@@ -516,8 +516,8 @@ making execution steps.
 bool stop() { return false; }
 
 void busy_wait_thread() {
-  while (!stop()); // undefined behavior, thread makes no progress but the loop
-}                  // is not trivial because `stop()` is not a constant expression
+  while (!stop());  // undefined behavior, thread makes no progress but the loop
+}                   // is not trivial because `stop()` is not a constant expression
 
 int main() {
   std::thread t(busy_wait_thread);


source/ub.tex
leni536, line 1: You might want to add to the explicitly called out phrases at the beginning of [ub.general]:

https://github.com/cplusplus/draft/blob/f489b81e3e3bec041f59d473a47cd874e955a994/source/ub.tex#L6-L9

timuraudio: Thanks for the suggestion! I get where you are coming from, but I'm not sure it makes sense in this case.

The fact that the UB in [intro.progress] is _not_ called out using one of the phrases listed in [ub.general] is a wording defect which is going to be fixed by CWG2816. So I think we should leave the list of phrases in [ub.general] as-is and instead just apply the proposed resolution in CWG2816, which will change [intro.progress] to explicitly say "the behavior is undefined" instead of "the implementation may assume".

Would that be an acceptable course of action for you?

We just need to make sure that when the wording in [intro.progress] is fixed, the merge conflict between that fix and this PR will be resolved in a way that doesn't accidentally drop the new `\ubdef`.

leni536: I think it is a little bit weird for [ub.general] to be in contradiction [ub.intro.progress], even if it's temporary. If you do add the phrase to [ub.general] then it can always be removed when CWG2816 eventually gets resolved. This could be recorded under the proposed resolution of CWG2816, so it does not get forgotten.

timuraudio: The issue with that approach is that the phrase "the implementation may assume..." does not normally mean "otherwise the behaviour is undefined". The phrase, or variations of it, occur about a dozen times in the Standard, but [intro.progress] is actually the only place where it means "otherwise the behaviour is undefined". In other places, it means things like "we will ignore this case" (e.g., [over.call.func]/2), "we will treat this case as the default" (e.g., [class.access.base]/2), etc. 

Since none of the other occurrences of the phrase imply that there is UB, adding the phrase to [ub.general] as you suggest seems outright wrong and I don't think we should do that, not even temporarily. 

In other words, yes it seems a little bit weird for [ub.general] to be in contradiction of [ub.intro.progress], but if we add the phrase to [ub.general] as you suggest, then instead it will be in conctradiction with a bunch of other places, and in a way that seems worse (suggesting that there is UB where there is none).

Is there any other resolution that you can live with?

leni536: > Is there any other resolution that you can live with?

I can live with the PR as is, I am not going to die on this hill.

Another option would be to make it clear in [ub.general] that the list of phrases there are not exhaustive.

> This Annex documents undefined behavior explicitly called out in the main standard text using<del> the following</del> phrases<ins> such as</ins>:

> This Annex documents undefined behavior explicitly called out in the main standard text using<ins>, among others,</ins> the following phrases:

timuraudio: Making the list non-exhaustive loses the important property that people can reason about when the text of the Standard means "this is explicit UB" and when it doesn't. Defining a closed set of phrases (and fixing any wording that doesn't use those phrases correctly) is much more useful here in my opinion than defining an open set as you suggest.

We now have a closed set of phrases in [ub.general], and we have one instance of UB that is inconsistent with that in [intro.progress]. So I think the best approach is to fix that one inconsistency by adopting [CWG2816](https://cplusplus.github.io/CWG/issues/2816.html) and to move on, not to water down the UB Annex wording or make it more ambiguous (even if temporarily).

So my preference remains to merge this PR as-is, but I am happy for Jens to make the call.

jensmaurer:
This needs more work:

2025-10-11T12:12:57.1463233Z ##[error]comment starts in column 19: ub.tex:519:comment not aligned to multiple of 4:   while (!stop()); // undefined behavior, thread makes no progress but the loop
2025-10-11T12:12:57.1472632Z ##[error]comment starts in column 19: ub.tex:520:comment not aligned to multiple of 4: }                  // is not trivial because `stop()` is not a constant expression

PR #8341, opened 2025-10-17
[simd.permute.static] Use satisfies for satisfaction

jwakely:
Fixes NB US 182-296 (C++26 CD).

Fixes cplusplus/nbballot#871


Dani-Hub:
There seems to be a typo "satisifes" in the new text

jwakely:
> There seems to be a typo "satisifes" in the new text

Thanks, fixed now

jensmaurer (approved):

PR #8342, opened 2025-10-17
[simd.permute.mask] clarify list is in ascending order

jwakely:
Fixes NB US 183-290 (C++26 CD).

Fixes cplusplus/nbballot#865

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -18923,7 +18923,7 @@ Let:
 \begin{itemize}
 \item
 \exposid{set-indices} be a list of the index positions of \tcode{true}
-elements in \tcode{selector}.
+elements in \tcode{selector}, in ascending order.
 \item
 \tcode{\exposidnc{bit-lookup}($b$)} be a function which returns the index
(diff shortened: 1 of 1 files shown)


source/numerics.tex
```diff
@@ -18922,7 +18922,7 @@
 Let:
 \begin{itemize}
 \item
-\exposid{set-indices} be a list of the index positions of \tcode{true}
+\exposid{set-indices} be a list in ascending order of the index positions of \tcode{true}
 elements in \tcode{selector}.
```
jensmaurer: I'm sorry, I thought the proposal is to append "in ascending order".
As written, "of the index positions" applies to "order", but we want to continue to say "list of index positions".

```suggestion
\exposid{set-indices} be a list of the index positions of \tcode{true}
elements in \tcode{selector}, in ascending order.
```

jensmaurer (changes requested):

source/numerics.tex
jwakely: done

PR #8344, opened 2025-10-18
[cpp.module] Cannot treat module directive as normal text

AlisdairM:
With the application of papers restricting macros in module directives, there remain no valid transformations for treating any part of such a directive as normal text.

### DIFF
--- a/source/preprocessor.tex
+++ b/source/preprocessor.tex
@@ -1252,14 +1252,6 @@ the \grammarterm{pp-module-name} or \grammarterm{pp-module-partition}
 shall currently be defined as an object-like macro.
 
 \pnum
-Any preprocessing tokens after the \tcode{module} preprocessing token
-in the \tcode{module} directive are processed just as in normal text.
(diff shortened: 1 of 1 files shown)


jensmaurer (changes requested):
I believe this is an undesirable normative change.

Note that pp-module has pp-tokens at the end (after the pp-module-partition), and those can contain macros, in particular function-like macros (and those can expand to nothing to yield a valid phase 7 "module" thing).

AlisdairM:
Withdrawn.  I had missed the trailing _pp-tokens_ in the _pp-module_, after the module name and module partition.

Nevertheless the sentence I tried to remove is confusing, and appears to be an oversight when wording https://wg21.link/p1853 that added a further restriction against macro expansion in the _pp-module-name_ and _pp-module-partition_ when scanning for module names, i.e., the text between `module` and the trailing _pp-tokens_.

I will see if I can come back with a better clarification that does not accidentally change meaning.

PR #8348, opened 2025-10-19
[meta.unary.prop] [meta.rel] P/R for LWG3967

Quuxplusone:
https://cplusplus.github.io/LWG/issue3967

This intends to eliminate all instances of "is known not to throw any exception" from the library clauses. "Is known" is too vague a condition (known to whom?); we ought to define it explicitly in terms of potentially-throwing expressions, i.e., we expect these traits to exactly match the result of the `noexcept` operator.

### DIFF
--- a/source/meta.tex
+++ b/source/meta.tex
@@ -1425,8 +1425,8 @@ The compilation of the
  \tcode{struct is_nothrow_constructible;}   &
  \tcode{is_constructible_v<T,} \tcode{ Args...>} is \tcode{true}
  and the
- variable definition for \tcode{is_constructible}, as defined below, is known not to
- throw any exceptions\iref{expr.unary.noexcept}.
+ variable definition for \tcode{is_constructible}, as defined below,
+ contains no potentially-throwing expression\iref{except.spec}.
  &
  \tcode{T} and all types in the template parameter pack \tcode{Args}
  shall be complete types, \cv{}~\keyword{void},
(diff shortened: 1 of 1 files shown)


source/meta.tex
```diff
@@ -1425,8 +1425,8 @@
  \tcode{struct is_nothrow_constructible;}   &
  \tcode{is_constructible_v<T,} \tcode{ Args...>} is \tcode{true}
  and the
- variable definition for \tcode{is_constructible}, as defined below, is known not to
- throw any exceptions\iref{expr.unary.noexcept}.
+ variable definition for \tcode{is_constructible}, as defined below,
+ contains no potentially-throwing expression\iref{except.spec}.
```
jensmaurer, line 1429: What does "contains" mean?  The variable definition itself is not an expression.

jensmaurer (changes requested):

source/meta.tex
Quuxplusone: That's a fair question. I think what we're trying to say here is that [below](https://eel.is/c++draft/meta.unary.prop#9) we have, textually, a variable definition `T t(declval<Args>()...);`, and we're asking whether that text produces any potentially-throwing expressions... but if the English word "contains" doesn't suffice to convey that, I don't immediately have any better suggestions.

I think https://eel.is/c++draft/except.spec#6 is the closest we come to prior art here. (It doesn't say "contains.")

source/meta.tex
```diff
@@ -1824,7 +1826,7 @@
  \tcode{struct is_nothrow_convertible;}     &
  \tcode{is_convertible_v<From, To>} is \tcode{true} and
  the conversion, as defined by \tcode{is_convertible},
- is known not to throw any exceptions\iref{expr.unary.noexcept} &
+ is not a potentially-throwing expression\iref{except.spec}. &
```
Quuxplusone, line 1829: We probably have another problem here, because a conversion is not an expression.

source/meta.tex
jensmaurer: Well, the only expressions I'm seeing here are the argument expressions involving declval.  The constructor invocation is not a (grammatical) expression, and neither are the default arguments (if any).

frederick-vs-ja:
Perhaps we want to say "contains no potentially-throwing expression or potentially-throwing implicit function invocation in its immediate context". But CWG1844 is not yet resolved.

frederick-vs-ja:
I think it might be cleaner and consistent to introduce terms "potentially-throwing/non-throwing initializers" and use them for `is_nothrow_meow`. But this needs to touch the core wording...

PR #8272, opened 2025-10-03
[meta.reflection.access.context] Remove stray `static` in declaration of `via`

eisenwave:
This is a misapplication of P2996R13 https://github.com/cplusplus/papers/issues/1668.

Fixes https://github.com/cplusplus/nbballot/issues/777.

The synopsis has no `static` (which is correct).

PR #8352, opened 2025-10-20
[meta.reflection.access.context] Remove 'static' from via itemdecl

jwakely:
This was a misapplication of the incoming paper.


jwakely:
It's not static in the class synopsis above, and not static in the decl just below p11 here: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p2996r13.html#pnum_639

jensmaurer:
This is an NB comment and was addressed by #8272, with the proper NB comment linking.

PR #8353, opened 2025-10-21
[simd] Resolve LWG4420 and LWG4393

mattkretz:
This provides the proposed resolution to LWG4420, which then also resolves LWG4393.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -16232,6 +16232,12 @@ template<class T>
     bool_constant<T() == T::value>::value &&
     bool_constant<static_cast<decltype(T::value)>(T()) == T::value>::value;
 
+template<class From, class To>
+  concept @\defexposconceptnc{explicitly-convertible-to}@ =                                // \expos
+    requires {
+      static_cast<To>(declval<From>());
+    };
+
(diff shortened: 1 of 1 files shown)


jensmaurer:
This pull request has nothing to do with the current proposed resolution on the mentioned LWG issues.
Also, where did the "output_range" constraint go that would address LWG4393?
And please use the correct git commit title per https://github.com/cplusplus/draft/wiki/Commit-message-format .
And we probably have an overfull \hbox or so; the CI build is failing.

mattkretz:
Sorry @jensmaurer, I did not expect a review as long as the PR is in draft status.
I just sent a big update to LWG4420 and wanted to point to a PR from it, which is why I opened it here. I'll fix the hboxes and commit message.

mattkretz:
this is outdated and would need a new PR from my new fork anyway

tkoeppe:
All the forks have been restored -- could you have another look?

mattkretz:
Yes, I can update it with the last changes. How does this work wrt. https://github.com/cplusplus/draft/issues/8461? (It's one part of LWG Motion 2)

Issue #8249, opened 2025-09-25
alx-0014 - Refactor syntax of preprocessing directives


alejandro-colomar:
Hi!

This is an editorial refactor I'm proposing to the C Committee.  The C Committee wanted to make sure that the C++ Committee didn't have any objections before accepting it, so here I am.

The proposal below applies to the C2y working draft, but it almost applies to the C++ working draft, just changing the section numbers in the obvious way.

Would you mind applying this editorial change?

Cheers,
Alex

```
Name
	alx-0014r6 - Refactor syntax of preprocessing directives

Category
	Cosmetic refactor; readability; editorial;

Authors
	Alejandro Colomar <alx@kernel.org>

History
	<https://www.alejandro-colomar.es/src/alx/alx/wg14/alx-0014.git/>

	r4 (2025-08-29):
	-  Split from alx-0014.

	r5 (2025-08-31):
	-  Move alx-0014 to a separate proposal: alx-0065.

	r6 (2025-08-31):
	-  Merge the split proposals into a single on with subsections.

Principles
	-  Keep the language small and simple
	-  Facilitate interoperability
	-  Codify existing practice to address evident deficiencies

Rationale
	Editorial change.  This moves text around for better
	organization and readability.  No semantic changes.  This makes
	it so that any future proposals to the preprocessor will be
	easier to apply.

	The proposal is split in subsections, separated by '---', for
	better readability.

	This proposal applies as is to the C++ latest draft, N5014, only
	changing section numbers and their titles.

	This change has prior art in C++, as they've done the same exact
	thing with their pp-import directive.  I guess they haven't done
	it with the other directives for compatibility with us, so let's
	help them and do it everywhere.

    Interaction with other proposals
	This blocks alx-0003 ("Add directives #def and #enddef")
	This blocks alx-0013 ("Prohibit non-directives (other than ID directives)").

---
Name
	alx-0014A - Refactor syntax of include directives

Proposed wording
	Based on C N3550.

    6.10.1  Preprocessing directives :: General  (C++: 15.1  :: Preamble)
	@@ Syntax, p1  (C++: no 'Syntax' subtitle)
	 control-line:
	-	<b># include</b> pp-tokens new-line
	+	include-directive
		pp-import	(C++ only)
	 	<b># embed</b> pp-tokens new-line
	 	<b># define</b> identifier replacement-list new-line
	 	<b># define</b> identifier lparen identifier-list(opt) <b>)</b> replacement-list new-line
[...]

PR #8354, opened 2025-10-21
[cpp.line] Simplify specification of line directives

AlisdairM:
This change set addresses a variety of concerns to produce a simpler and clearer specification for line directives.  It makes no functional changes, but changes words sufficiently to require a Core review.  A motivation behind this PR is to simplify the changes for the proposal to allow line directives before the global module fragment, https://wg21.link/P3868.

The line directive is given its own grammar production that formally states what used to be described in text restrictions and interpretations of the generic sequence of preprocessing tokens.

The other issue to address is that the specification refers to a _string-literal_ that is not present in the grammar. However, the constraints on that string literal are sufficient to replace that quoted _s-char-sequence_ in the grammar production with _string-literal_.

### DIFF
--- a/source/preprocessor.tex
+++ b/source/preprocessor.tex
@@ -47,15 +47,15 @@
 
 \begin{bnf}
 \nontermdef{control-line}\br
-    \terminal{\# include} pp-tokens new-line\br
     pp-import\br
+    line-directive\br
+    \terminal{\# include} pp-tokens new-line\br
     \terminal{\# embed \ } pp-tokens new-line\br
     \terminal{\# define } identifier replacement-list new-line\br
     \terminal{\# define } identifier lparen \opt{identifier-list} \terminal{)} replacement-list new-line\br
     \terminal{\# define } identifier lparen \terminal{... )} replacement-list new-line\br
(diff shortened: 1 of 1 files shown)


jensmaurer:
Remark (not necessarily dispositive): The status quo still has a good match with the C phrasing; this change would kill that correspondence.

Other than that, the funny presentation of the preprocessor grammar in the status quo has at least two useful properties that we would lose with the reformulation:
- There is no question that `# line pp-tokens` is a _control-line_.  Once you factor the "of the form" rules into the grammar, the parser would simply reject a parse as a _control-line_ and instead attempt to treat that as a _text-line_.  Which is not desirable.
- As written in the pull request, the grammar seems woefully ambiguous, because `#line 100` matches both of the `#line` alternatives (`100` is a preprocessing-token, indeed).

AlisdairM:
The grammar is disambiguated the same way that we do for `#include` and `#embed` but referring to "the first form" and going to the second form only if the first does not match.

The _control-line_ issue is a genuine issue that I missed and should consider more carefully.  That should also be a concern for #8249.

PR #8361, opened 2025-10-24
[meta] reflect_constant_array of an empty range returns const array<T,0>

hanickadot:
Marek Poláček asked in SG7 Mattermost if it should return `const` or not, I think for consistency it should be. If others agree this is PR fixing it.


jwakely:
LWG discussed this in telecon and agreed with the change. https://eel.is/c++draft/temp.param#13.sentence-1 says the object has a const type.

jwakely (approved):

PR #8362, opened 2025-10-24
[meta.reflection.result] move declaration of TCls earlier

jwakely:
Fixes NB US 117-178 (C++26 CD).

Fixes https://github.com/cplusplus/nbballot/issues/747

But I wonder if we should name this something other than `TCls` seeing as that name is used for a class template in at least five examples elsewhere in reflection wording. Maybe the invented template here should have a different name to avoid any risk of confusion.

### DIFF
--- a/source/meta.tex
+++ b/source/meta.tex
@@ -6119,9 +6119,14 @@ Let $V$ be:
 \end{itemize}
 
 \pnum
+Let \tcode{TCls} be the invented template:
+\begin{codeblock}
+template<T P> struct TCls;
+\end{codeblock}
+
+\pnum
 \returns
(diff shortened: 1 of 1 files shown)


eisenwave:
What is the rationale for the name `TCIs` anyway?

Maybe something like `CTP` (constant template parameter) would work? After all, it seems like `CTP` is really just being used to wrap a constant template parameter here. If you wanted to, you could also use `std::nontype`/`std::constant_arg` and wouldn't need to invent any class template.

Anyhow, the current change looks correct and like a clear improvement.

jwakely:
It's TCls with an L not TCIs with i

It's short for "template class", but any other name except `T` would work here.

PR #8363, opened 2025-10-24
[meta.reflection.layout] change 'entity' to 'construct'

jwakely:
Fixes NB US 104-165 (C++26 CD).

Fixes https://github.com/cplusplus/nbballot/issues/745


jensmaurer (approved):
This change doesn't really increase the beauty of the standard, but it's technically correct and there is no obvious alternative.  Let's go for it.

tkoeppe:
> This change doesn't really increase the beauty of the standard, but it's technically correct and there is no obvious alternative. Let's go for it.

I think that has never stopped us before.

PR #8366, opened 2025-10-25
[lex.pptoken][module][cpp] Index keyword placeholders]

### LINKED ISSUE #6844: [lex.pptoken] import-keyword, module-keyword, export-keyword are not indexed
I believe the problem is that they don't have a `\nontermdef` section and aren't otherwise indexed.

They're defined in [[lex.pptoken] p4](https://eel.is/c++draft/lex.pptoken#4) but not in the usual way:
https://github.com/cplusplus/draft/blob/ceff4ea83b511be01a8e1756386ce6a2e06e323c/source/lex.tex#L563-L569

I guess this also means that [gram] is incomplete because it doesn't explain how an *import-keyword* is produced, and doesn't link to it.

AlisdairM:
Fixes #6844 

Index the keyword placeholders, _export-keyword_, _import-keyword_, and _module-keyword_.

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -586,6 +586,9 @@ shall have the lexical form of a keyword, an identifier, a literal,
 or an operator or punctuator.
 
 \pnum
+\indextext{keyword placeholder!export-keyword}%
+\indextext{keyword placeholder!import-keyword}%
+\indextext{keyword placeholder!module-keyword}%
 The \grammarterm{import-keyword} is produced
 by processing an \keyword{import} directive\iref{cpp.import},
 the \grammarterm{module-keyword} is produced
@@ -1014,6 +1017,9 @@ reserved to the implementation for use as a name in the global namespace.%
 \indextext{identifier|)}
 
 \rSec1[lex.key]{Keywords}
+\indextext{keyword placeholder!export-keyword}%
+\indextext{keyword placeholder!import-keyword}%
+\indextext{keyword placeholder!module-keyword}%
 
 \begin{bnf}
 \nontermdef{keyword}\br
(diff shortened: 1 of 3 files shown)


AlisdairM:
Yes, I tried several arrangements and pushed up this first one for feedback.  Thanks.

I like the grouping under "keyword placeholders" but was thinking to also have the plain term redirect to the keyword placeholders entry.

I am similarly concerned about the font, but do not know our macros well enough to know the correct or preferred way to fix that.  Am hoping to see a review comment explaining ;)

jensmaurer:
Again, what's the point here?

_module-keyword_ already appears in our index of grammar productions, so it is _already_ indexed.  Nothing to do.

AlisdairM:
Per the linked comment, the other two keyword placeholders are not indexed.  The goal is simply to make sure that all three can be easily found in the main index, and I will adjust PR accordingly to find an acceptable way to land.

If we can just get the these grammar terms to appear in the main index from their use in grammar,  like most other grammar terms, I would do that --- but I do not know how.  They are never *defined* in the grammar as they are terminals, so do not get the extra kick into the main index.

jensmaurer:
> Per the linked comment, the other two keyword placeholders are not indexed.

That is not an accurate summary of the linked issue.

All three are indexed in the "index of grammar productions".  As explained in the linked issue, there is no index entry in the general index, because they do not have a (traditional) definition.

I have no objection to adding a general index entry for those grammar productions, but that index entry should be the same style as the other grammar entries in the main index, and should point to the definition (which, in this case, is probably a prose text definition).  The pull request we  have here is nothing like that.

jensmaurer:
#8367 fixes this.

PR #8367, opened 2025-10-25
[lex.pptoken] Index keyword placeholders

### LINKED ISSUE #6844: [lex.pptoken] import-keyword, module-keyword, export-keyword are not indexed

jensmaurer:
Fixes #6844

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -586,6 +586,9 @@ shall have the lexical form of a keyword, an identifier, a literal,
 or an operator or punctuator.
 
 \pnum
+\indexgrammar{\idxgram{import-keyword}}%
+\indexgrammar{\idxgram{module-keyword}}%
+\indexgrammar{\idxgram{export-keyword}}%
 The \grammarterm{import-keyword} is produced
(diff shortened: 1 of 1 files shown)


AlisdairM:
Thanks Jens --- that is exactly what I was trying to do but did not know how.
This PR looks good to me.

jensmaurer:
> what I was trying to do but did not know how

Then, why don't you just ask instead of posting obviously misguide pull requests?

AlisdairM:
Sorry, this *was* my way of asking, with the assumption I would adjust the PR accordingly.  I was trying to make the minimal work for others, and got it wrong.  Thanks for fixing my mess.

PR #8369, opened 2025-10-25
[cpp.line] Address string-literal vs s-char-sequence

AlisdairM:
The text of this specification places a restriction on a _string-literal_ that is not present in the grammar, although the grammar is an exact match for a _string-literal_ with that restriction.

The restricted term is also used when specifying the effects of the line directive with a string, so prefer to move _string-literal_ into the grammar rather than strike that paragraph and adjust the specification for its effects.

Added a cross-reference to where character string literal is defined as it may not be obvious to all readers that this is a term of power, even though it is defined as a term and has an index entry.

### DIFF
--- a/source/preprocessor.tex
+++ b/source/preprocessor.tex
@@ -1805,7 +1805,7 @@ macro shall be followed by a parameter as the next preprocessing
 token in the replacement list.
 
 \pnum
-A \defn{character string literal} is a \grammarterm{string-literal} with no prefix.
+A \defn{character string literal} is a \grammarterm{plain-string-literal}.
 If, in the replacement list, a parameter is immediately
 preceded by a
 \tcode{\#}
@@ -2071,12 +2071,6 @@ a macro name.
 \indextext{\idxcode{\#line}|see{preprocessing directive, line control}}
 
 \pnum
-The \grammarterm{string-literal} of a
-\tcode{\#line}
-directive, if present,
-shall be a character string literal\iref{cpp.stringize}.
-
-\pnum
 The
 \defn{line number}
 of the current source line is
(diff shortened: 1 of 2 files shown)


source/preprocessor.tex
```diff
@@ -2102,7 +2102,7 @@
 \pnum
 A preprocessing directive of the form
 \begin{ncsimplebnf}
-\terminal{\# line} digit-sequence \terminal{"} \opt{s-char-sequence} \terminal{"} new-line
+\terminal{\# line} digit-sequence \terminal{"} \opt{string-literal} \terminal{"} new-line
```
jensmaurer: No.

First, a _string-literal_ includes the delimiting quotation marks, so this is a bad change.
Second, a _string-literal_ allows for an _encoding-prefix_ or rawness, but the status quo "form" match does not.  So, this is a bad change, because it would appear to allow such prefixes (and we don't know what they mean in this context).  A reference to "the character string literal" does not establish a suitable constraint.

Yes, p1 talks about a _string-literal_ with zero context, and that's a bug that needs to be fixed, but this intro sentence is at issue, not the rest of the specification without further elaboration.  What does C23 say in that area?

AlisdairM: The term "character string literal" is a well defined term establishing exactly that restriction, but if you prefer I will tackle this problem from the specification of what to do with the s-char-sequence instead.

My preferred resolution for C++29 would be to introduce _character-string-literal_ as the quoted _s-char-sequence_ in [lex.string], so that a _string-literal_ is an encoding prefix followed by either a _character-string-literal_ or a `R` raw-string-literal.  That would allow us to use _character-string-literal_ both here in [cpp.line] and [cpp.stringize], and potentially clean up other parts of the standard jumping around string literal syntax --- but that is more than I want to push into ballot review time, as that would clearly need Core review, better spent elsewhere.

jensmaurer:
CWG2693 is related.

jensmaurer:
And see CWG3076 for the introduction of _plain-string-literal_ (with no prefix), which might be useful here.

But that's not editorial territory; there is too much subtlety to go wrong.

AlisdairM:
I have pushed a revision with the basis of my broader update, and will turn this into a draft PR.  I suspect that my update will be in the direction of _plain-string-literal_ but I think you cited the wrong core issue?

Noted that the direction specifically for `#line` to move in the direction of _header-name_ does seem more appropriate.

I will leave this now as potential C++29 material.

PR #8373, opened 2025-10-28
[concept.regularinvocable], [iterator.concept.winc] Replace "annotation" with "comment"

eisenwave:
Fixes NB US 71-128 (C++26 CD).
Fixes https://github.com/cplusplus/nbballot/issues/707.

The NB comment only asks to replace "annotation" with "comment", which makes sense considering that we have a language construct called "annotation" now.

I went a step further and copied the contents of this comment here. This seems helpful because you immediately understand what the contents of those comments are without having to navigate elsewhere.

Note that "not required to be equality-preserving" is a special comment used in a large number of places, defined in https://eel.is/c++draft/concepts.equality#4 My intuition was initially to use a *Remarks* element instead, but that would be too large of a change.

### DIFF
--- a/source/concepts.tex
+++ b/source/concepts.tex
@@ -1261,8 +1261,8 @@ The \tcode{invoke} function call expression shall be
 equality-preserving\iref{concepts.equality} and
 shall not modify the function object or the arguments.
 \begin{note}
-This requirement supersedes the annotation in the definition of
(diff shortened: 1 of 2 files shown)


jwakely (approved):

tkoeppe:
Thanks, that's a good improvement!

PR #8387, opened 2025-11-02
[simd.bit] Wrap "bit-func" placeholder in `\tcode`

eisenwave:


eisenwave:
@tkoeppe are you removing tags from merged PRs in general, or do you feel like P2 was incorrect for this PR?

tkoeppe:
I'm removing them in general. Once it's resolved I figured we don't need it tagged anymore.

Separately, I'm also wondering if we might want a "P0 "ready"" tag for things which essentially require no further material work...

eisenwave:
> I'm removing them in general. Once it's resolved I figured we don't need it tagged anymore.

Hmm okay, it seems like removing the labels would create an additional task though. If you look at the open issues and PRs, it doesn't really matter whether closed items exist that still have a priority label on them. Anyhow, I just found it surprising because I'm not sure what harm there is in keeping the label.

> Separately, I'm also wondering if we might want a "P0 "ready"" tag for things which essentially require no further material work...

For PRs, I assume this would be communicated by the "approved" status from the chairs, or by you being assigned to the PR, right? The former can be searched for via `review:approved`. For issues, this would be communicated by the existence of a PR which closes the issue. Are there PRs and issues with these indicators that would not be P0?

tkoeppe:
Re removal: I don't really have a strong opinion. I was thinking of the new P tags as triaging tools, and I considered triaging to be over when the PR is applied, but that's neither deeply reasoned nor meant as guidance. I don't ultimately think it matters much either way.

Re P0: Yeah, we have other ways to signal that, too.

eisenwave:
> Re removal: I don't really have a strong opinion. I was thinking of the new P tags as triaging tools, and I considered triaging to be over when the PR is applied, but that's neither deeply reasoned nor meant as guidance. I don't ultimately think it matters much either way.

My opinion is that it's removing information for no technical benefit. I'd rather keep the priority tags.

I also don't see us pulling it off consistently, unless it's automated via GitHub. One would need to remove the priority tags from both the PR and the closed issue, in all cases. Someone will slip up eventually.

tkoeppe:
As I said, it wasn't meant to imply guidance or requirements.

PR #8388, opened 2025-11-02
[simd.expos], [simd.syn] Synchronize synopsis references with subclause headings

eisenwave:
Fixes NB US 181-294 (C++26 CD).
Fixes https://github.com/cplusplus/nbballot/issues/869.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -16277,7 +16277,7 @@ template<class BinaryOperation, class T>
 template<class T> using @\exposidnc{native-abi} = \seebelownc@;                      // \expos
 template<class T, @\exposidnc{simd-size-type}@ N> using @\exposidnc{deduce-abi-t} = \seebelownc@;  // \expos
 
-// \ref{simd.flags}, Load and store flags
+// \ref{simd.flags}, load and store flags
 struct @\exposidnc{convert-flag}@;                                                 // \expos
 struct @\exposidnc{aligned-flag}@;                                                 // \expos
 template<size_t N> struct @\exposidnc{overaligned-flag}@;                          // \expos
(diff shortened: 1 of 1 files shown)


eisenwave:
I've asked @mattkretz about it, and he likes the longer comments for the "XXX permute" subclauses more.

The direction for the initial version of the PR is to copy the subclause title as is. The default assumption should be that the subclause title is more authoritative.

Also, we're not really consistent about referencing
- "`basic_vec`"
- "`vec`"
- "`basic_vec` and `basic_mask`"
- none

I'd rather not innovate and just delete it all. We can always create prettier comments in a future version if that's a good use of our time.

tkoeppe:
Yeah, I think there was some conflict between NB comments wanting longer headings, and our overall style of not repeating the class names in headings. I'll take a look later.

jensmaurer:
The comments in the synopsis should exactly copy the subclause headings.  Yes, those are short-ish.  It looks like what this pull request attempts to do.

I don't think SIMD is so special to warrant something different here.

tkoeppe:
Indeed, and the NB comments may not have taken the wider context into consideration, so I think it's fine to accept "that we can improve things", but improve them in our own way.

jensmaurer (approved):
So, it looks like we should just go ahead with this.

tkoeppe:
@eisenwave Could you please check your commit message(s)? I think there's a blank line missing after the first line, and I've seen this in several of your commits now.

eisenwave:
@tkoeppe fixed. I wasn't even aware there's supposed to be a blank line because I've looked Jens' commits, and the GitHub web client doesn't render them (the difference only becomes visible in `git log`).

tkoeppe:
Thanks! I think this is some form of Git convention. (Emacs warns you if you have multiple lines and the second line isn't blank...)

PR #8389, opened 2025-11-03
[module.global.frag] Remove irrelevant note about preprocessor

AlisdairM:
[module.global.frag] is entirely part of phase 7 of translation, and it makes no sense to talk of preprocessing directive in the grammar term *declaration-seq*.  Strike the note rather than try to turn it into something meaningful.

### DIFF
--- a/source/modules.tex
+++ b/source/modules.tex
@@ -588,13 +588,6 @@ import M1;              // error: cyclic interface dependency $\mathtt{M3} \righ
 \end{bnf}
 
 \pnum
-\begin{note}
-Prior to phase 4 of translation,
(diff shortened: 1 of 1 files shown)


jensmaurer (dismissed):
<del>I'm ok with that, in particular there is an adjacent pull request that adds a bit of clarification to [cpp.pre].</del>

jensmaurer:
I've thought about this some more.  I think the preprocessor treatment of the global module fragment is so surprisingly different from any other context that I would want to retain a cross-reference from phase 7 [module.global.frag] to [cpp.pre] in some form.

PR #8390, opened 2025-11-03
[cpp.pre] Add note on preprocessing directives in pp-global-module-fragment

AlisdairM:
Add a note confirming that executing preprocessor directives within the pp-global-module-fragment can produce text-lines.

### DIFF
--- a/source/preprocessor.tex
+++ b/source/preprocessor.tex
@@ -302,6 +302,10 @@ with \impldef{supported forms of \#embed prefix parameters} semantics.
 At the start of phase 4 of translation,
 the \grammarterm{group} of a \grammarterm{pp-global-module-fragment} shall
 contain neither a \grammarterm{text-line} nor a \grammarterm{pp-import}.
+\begin{note}
+Executing preprocessing directives such as source file inclusion\iref{cpp.include}
+during phase 4 of translation can produce \grammarterm{text-line}s.
+\end{note}
 
 \pnum
 When in a group that is skipped\iref{cpp.cond}, the directive


source/preprocessor.tex
```diff
@@ -298,6 +298,10 @@
 At the start of phase 4 of translation,
 the \grammarterm{group} of a \grammarterm{pp-global-module-fragment} shall
 contain neither a \grammarterm{text-line} nor a \grammarterm{pp-import}.
+\begin{note}
+Executing preprocessing directives such as source file inclusion\iref{cpp.include}
```
jensmaurer, line 306: "Executing" sounds like runtime.  Do we have a better word for the phase 4 activities?  "preprocessing"?

AlisdairM: I do not like "executing" either, but it is the only term of art I could definitively find to describe how we process preprocessing directive, as per [lex.phases] phase 4:
"Preprocessing directives (Clause15) are executed, macro invocations are expanded (15.7), and `_Pragma` unary operator expressions are executed (15.13)."

I will happily trade for a better verb, especially if it would apply to the above paragraph too.

jensmaurer: Ok, then.

tkoeppe:
@jensmaurer Did I get that right that you'd only want #8389 if we also merge this PR?

jensmaurer:
I've thought about this some more.  I think the preprocessor treatment of the global module fragment is so surprisingly different from any other context that I would want to retain a cross-reference from phase 7 [module.global.frag] to [cpp.pre] in some form.  This is a comment about #8389.

And that would decouple those two pull requests.

PR #8398, opened 2025-11-03
[basic.fundamental] Use `std::` prefix consistently for library type aliases

eisenwave:
Later in [basic.fundamental], we use `std::nullptr_t`. To my understanding, we generally want to use `std::` when referring to library types from within core wording.

Perhaps there would have been an argument that `uint_least16_t` is a "C thing" and so unlike `std::nullptr_t`, it doesn't need a `std::` prefix, but now `nullptr_t` also exists in C23 which defeats any justification for this local inconsistency.


jwakely (approved):

jwakely:
I don't think the C argument would work here, because `char16_t` and `char32_t` are very different in C. It would make little sense to argue for consistency with C while defining things differently from C!

PR #8400, opened 2025-11-04
[basic.lookup.general] Italicise "follow" in [basic.lookup.general]/2

### LINKED ISSUE #8142: Italicize "follow" in [basic.lookup.general]/2
[basic.lookup.general]/2 is the definition of "follow", not the definition of "program point". "follow" should be italicized to make this clear.

TymianekPL:
Fixes #8142

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -1658,7 +1658,7 @@ are the semantic properties introduced by the declarations
 used in further processing.
 
 \pnum
-A program point $P$ is said to follow
+A program point $P$ is said to \grammarterm{follow}
 any declaration in the same translation unit
 whose locus\iref{basic.scope.pdecl} is before $P$.
 \begin{note}
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -1658,7 +1658,7 @@ are the semantic properties introduced by the declarations
 used in further processing.
 
 \pnum
-A program point $P$ is said to \grammarterm{follow}
+A program point $P$ is said to \defnx{follow}{follow!program point}
 any declaration in the same translation unit
 whose locus\iref{basic.scope.pdecl} is before $P$.
 \begin{note}


TymianekPL:
As Brian suggested, I think adding a note that states it clearly would be nice as well. Not sure if that's the right call so I'll hold on with this one for now.

source/basic.tex
```diff
@@ -1658,7 +1658,7 @@
 used in further processing.
 
 \pnum
-A program point $P$ is said to follow
+A program point $P$ is said to \grammarterm{follow}
```
frederick-vs-ja: Looks wrong to me. The "follow" is not a grammar term. Perhaps we should index it as `\defnx{follow}{follow!program point}`.

TymianekPL: You're right, it's more of a definition than a grammar term... I'll change that.

jensmaurer:
@TymianekPL , if you want the git change history to show your full name, please configure it in github.

TymianekPL:
> @TymianekPL , if you want the git change history to show your full name, please configure it in github.

> Make profile private and hide activity
> Enabling this will hide your contributions and activity from your GitHub profile and from social features like followers, stars, feeds, leaderboards and releases.

This one?

jensmaurer:
No, this one:

image

(menu in top right corner)

TymianekPL:
Ah okay, you mean the name thing

tkoeppe:
OK. There's a wider issue that ISO doesn't really want things italicised, and that _all_ definitions must come from Clause 3, but that's not really practical for us. But this does mean that we have routinely created "local definitions" that are not marked up typographically and are not indexed. (I think "empty" as the shared pointer state comes to mind: https://eel.is/c++draft/util.smartptr.shared.general#1.sentence-3)

TymianekPL:
Now we just have to decide whether follow should be local or not... I myself am unsure about that. I would say yes, as it affects the core language and a very important part of name lookup. Furthermore, it's also in [basic], which to me is kind of the clause 3 of ISO guidelines...

eisenwave:
Should this PR be closed then, or do we want the italics?

We have like ten billion definitions that don't comply with the "don't make definitions italic" ISO rule, so is that ISO rule going to stop us from adding italics here?

TymianekPL:
I think adding italics is the right move.

TymianekPL:
Unlike shared pointers' *empty*, this is not a local definition to... to a paragraph? It applies to the program and translation unit. Paragraph in question:
> A program point P is said to follow any declaration in the same translation unit whose locus ([[basic.scope.pdecl]](https://eel.is/c++draft/basic.scope.pdecl)) is before P.

[[basic.lookup.general]/2](https://eel.is/c++draft/basic.lookup.general#2.sentence-1)

TymianekPL:
I think this additional indexed definition will contribute to making C++ less ambiguous and easier to work with in future standardisation

tkoeppe:
Can you please rebase?

jensmaurer (approved):

jensmaurer:
Yes, "follow" is a defined term here, and should be italics + indexed.

PR #8403, opened 2025-11-04
[container.adaptors.general] Remove using typename from expos-only alias template

AlisdairM:
This should be the last case of `typename` in a type alias or alias template in the library, after manually inspected a search for all `typename` in the rendered pdf file.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -15581,8 +15581,7 @@ The following exposition-only alias template
 may appear in deduction guides for container adaptors:
 \begin{codeblock}
 template<class Allocator, class T>
-  using @\exposid{alloc-rebind}@ =                      // \expos
-    typename allocator_traits<Allocator>::template rebind_alloc<T>;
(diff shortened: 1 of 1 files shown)


source/containers.tex
```diff
@@ -15582,7 +15582,7 @@
 \begin{codeblock}
 template<class Allocator, class T>
   using @\exposid{alloc-rebind}@ =                      // \expos
-    typename allocator_traits<Allocator>::template rebind_alloc<T>;
+    allocator_traits<Allocator>::template rebind_alloc<T>;
```
tkoeppe: Could you check, is the linebreak still needed now?

AlisdairM: Good instinct!  Pushed as one line, but not sure of preferred spacing to the // exposition only comment.

tkoeppe: Well, the comment would need to still fit on the same line, and ideally follow the comment placement rules (on a column whose count a multiple of 4).

PR #8404, opened 2025-11-04
[exec.spawn.future] add missing \tcode

ericniebler:
"_ssource-t_" in [exec.spawn.future/p6](https://eel.is/c++draft/exec#spawn.future-6) should be in code font.

image

### DIFF
--- a/source/exec.tex
+++ b/source/exec.tex
@@ -5339,9 +5339,9 @@ namespace std::execution {
 \end{codeblock}
 
 \pnum
-Let \placeholder{ssource-t} be an unspecified type
+Let \tcode{\placeholder{ssource-t}} be an unspecified type
 that models \exposconcept{stoppable-source} and
-let \tcode{ssource} be an lvalue of type \placeholder{ssource-t}.
+let \tcode{ssource} be an lvalue of type \tcode{\placeholder{ssource-t}}.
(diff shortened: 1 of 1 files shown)


tkoeppe:
Thanks, @ericniebler! Yes, it seems quite a common gotcha that the `\placeholder` macro does not also imply "code" and that you need both. Maybe we should call that out on the wiki.

Could you please update the commit message according to our standard format https://github.com/cplusplus/draft/wiki/Commit-message-format with the positional reference? (And the change text can just say "add missing \tcode".) Thanks!

ericniebler:
I presume there is a reason why `\placeholder` does not imply `\tcode`?

eisenwave:
@ericniebler `\placeholder` is also being used in various other places as a mathematical/general placeholder, and there was a suggestion to make it render as math font even within code at some point. Unlike `\exposid`, it is a situational macro and doesn't always mean code.

I think what trips people up is that in code blocks and in places like `\tcode{int\placeholder{N}_t}`, we don't add extra code font for the placeholder; it's obtained from the context. Also, `\exposid` and `\placeholder` render the same in code, so one would assume they're interchangeable, but only `\exposid` adds code font everywhere. It's really quite surprising.

tkoeppe:
`exposid` is semantically different in that its argument is the entire identifier. `placeholder` is more a general "placeholder", which (as in the intN_t example) can be used in a variety of ways and places.

PR #8405, opened 2025-11-04
[basic.def] Fix note -- declarations cannot be empty

AlisdairM:
As noted during CWG review of another issue.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -186,9 +186,8 @@ A declaration can also have effects including:
 \begin{itemize}
 \item a static assertion\iref{dcl.pre},
 \item controlling template instantiation\iref{temp.explicit},
-\item guiding template argument deduction for constructors\iref{temp.deduct.guide},
-\item use of attributes\iref{dcl.attr}, and
-\item nothing (in the case of an \grammarterm{empty-declaration}).
+\item guiding template argument deduction for constructors\iref{temp.deduct.guide}, and
+\item use of attributes\iref{dcl.attr}.
 \end{itemize}
 \end{note}
 


jensmaurer (dismissed):
<del>Agreed.  A (prose-font) "declaration" is never an _empty-declaration_.</del> (see below)

lprv:
Why not? An *empty-declaration* can appear as a [*name-declaration*](https://eel.is/c++draft/dcl.pre#nt:name-declaration) or a [*member-declaration*](https://eel.is/c++draft/class.mem.general#nt:member-declaration), both of which are included in the definition of "declaration" given by [[basic.pre]/5](https://eel.is/c++draft/basic.pre#5.1).

jensmaurer:
Indeed.

jensmaurer:
@AlisdairM , please re-assess.

AlisdairM:
Reviewing the text again, the note is attached to a sentence that starts:
"A declaration of an entity `X` ...".  The whole paragraph is clearly talking about declarations of entities, and empty declarations do not declare entities.  The note is not super precise, allowing the reader to infer that by "declaration" we mean the set of declarations of the preceding sentence.  Likewise, the list does not claim to be complete.  I believe the initial CWG reaction was correct, and this note would be clearer without the deleted bullet.

eisenwave:
It's not clear to me whether the note is only attached to the last sentence or the paragraph as a whole. The paragraph starts with

> A declaration ([basic.pre]) may (re)introduce one or more names and/or entities into a translation unit.

Emphasis on "may". It doesn't necessarily do so (and in the case of an *empty-declaration*, which is a "declaration", it does not. Everything makes sense to me.

The last bullet (which you're not deleting) says

> use of attributes ([dcl.attr]).

This presumably refers to `[[fallthrough]];` and `[[assume(true)]];`, both of which are an *attribute-declaration*, which also doesn't declare any entity. It's basically an *empty-declaration* with some attributes slapped onto it. Why should this bullet stay then, if it also doesn't talk about entities?

AlisdairM:
Withdrawing.

After a more carefully re-reading I agree with Jan.  Everything in that list is about *not* declaring a name, the complement to the normative sentence.

PR #8407, opened 2025-11-04
[exec.cmplsig] add missing \exposid

ericniebler:
[[exec.cmplsig]/p6](https://eel.is/c++draft/exec#cmplsig-6) uses the exposition-only _`META-APPLY`_ without putting it in italics.

image


tkoeppe:
Thanks!

PR #8409, opened 2025-11-04
[lex.token] Strike mention of whitespace in phase 7

AlisdairM:
It is meaningless to talk of whitespace separating tokens in phase 7 as whitespace is discarded at the end of phase 4.

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -879,15 +879,6 @@ There are five kinds of tokens: identifiers, keywords, literals,%
 Literals include strings and character and numeric literals.
 \end{footnote}
 operators, and other separators.
-\indextext{whitespace}%
-Comments and the characters \unicode{0020}{space}, \unicode{0009}{character tabulation},
-\unicode{000b}{line tabulation}, \unicode{000c}{form feed}, and new-line
-(collectively, ``whitespace''), as described below, are ignored except
-as they serve to separate tokens.
-\begin{note}
-Whitespace can separate otherwise adjacent identifiers, keywords, numeric
-literals, and alternative tokens containing alphabetic characters.
-\end{note}
 \indextext{token|)}
 
 \rSec1[lex.name]{Identifiers}


eisenwave:
I'm not convinced we should totally remove this wording without even leaving a note. It's a helpful insight that whitespace acts as a separator between the tokens that are processed in this subclause (even if the process of separation happens in a much earlier phase).

Perhaps we could keep just the note that is deleted in this PR and explicitly point out in which phase that separation takes place, with an `\iref{lex.phases}`.

AlisdairM:
I was requested several times in review of other changes to remove this text.  It has no normative impact, and truly is nonesense as whitespace does not separate tokens, it simply does not exist for several phases of translation before tokens even exist.

We have already improved the text around phases 3 and 4 where we state that whitespace separates preprocessing tokens, and there are no parts of this text that seem useful to transfer to those other notes.

source/lex.tex
```diff
@@ -879,15 +879,6 @@
 Literals include strings and character and numeric literals.
 \end{footnote}
 operators, and other separators.
-\indextext{whitespace}%
-Comments and the characters \unicode{0020}{space}, \unicode{0009}{character tabulation},
-\unicode{000b}{line tabulation}, \unicode{000c}{form feed}, and new-line
-(collectively, ``whitespace''), as described below, are ignored except
-as they serve to separate tokens.
```
jensmaurer, line 886: Isn't this the normative part that currently defines "whitespace"?  We shouldn't strike this editorially.

AlisdairM: No, whitespace is defined by [lex.pptoken]p1:
"Preprocessing tokens can be separated by whitespace; this consists of comments (5.4), or whitespace characters (u+0020 space, u+0009 character tabulation, new-line, u+000b line tabulation, and u+000c form feed), or both."  This is the definition retained by P3657.

The rest of this paragraph, including "as described below", are nonsense.

jensmaurer (approved):

PR #8410, opened 2025-11-04
[cpp.pre] Apply unicode markup to last parts of [pre]

AlisdairM:
This should complete the process of applying unicode code point markup to denote specific characters for the grammar in the preprocessing clause.


eisenwave:
@AlisdairM did we decide to do that in some paper or issue in here? I'm asking for the purpose of P2/P3 prioritization.

AlisdairM:
This was the direction affirmed several times during the review of https://wg21.link/P2843, but has not been recorded otherwise.

PR #8412, opened 2025-11-04
[lex] Complete the use of unicode code points

### LINKED ISSUE #5502: [lex.pptoken] p2 Split the quotation characters into the corresponding terms
[lex.pptoken] p2 says
> Whitespace can appear within a preprocessing token only as part of a header name or between the **quotation characters** in a character literal or string literal.

Since we have named `"` as U+0022 QUOTATION MARK and named `'` as U+0027 APOSTROPHE, we shouldn't use "quotation characters" to generally name such two characters. We should splict it to the term to make the meaning be clearer.
> Whitespace can appear within a preprocessing token only as part of a header name or between the ~quotation characters~ U+0027 APOSTROPHE characters in a character literal or between the U+0022 QUOTATION MARK characters in a string literal.

AlisdairM:
Completes the task of applying Unicode code point markup to
refer to single characters in normative text.
    
Replace all remaining uses of backslash as glyph (\) or text
with the corresponding `unicode{005c}{reverse solidus}` markup.
    
Replace a couple of text reference to single and double quotes
with their corresponding Unicode markup, and similarly for the
glyphs ` and ".
    
This should compete the last Unicode markup in [lex].

Resolves #5502

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -99,12 +99,12 @@ representing end-of-line indicators as new-line characters.
 \indextext{line splicing}%
 If the first translation character is \unicode{feff}{byte order mark},
 it is deleted.
-Each sequence comprising a backslash character (\textbackslash)
+Each sequence comprising a \unicode{005c}{reverse solidus} character (\textbackslash)
 immediately followed by
 zero or more whitespace characters other than new-line followed by
 a new-line character is deleted, splicing
 physical source lines to form \defnx{logical source lines}{source line!logical}. Only the last
-backslash on any physical source line is eligible for being part
+\unicode{005c}{reverse solidus} on any physical source line is eligible for being part
 of such a splice.
 \begin{note}
 Line splicing can form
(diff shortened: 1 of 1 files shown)


eisenwave:
Making this P2 because the direction affirmed for P2843R3 is that we ought to use code points instead of colloquial names like "backslash".

tkoeppe:
Could you reword the commit message to describe the change made by the commit, instead of the wider project (which isn't very descriptive)?

jensmaurer:
@AlisdairM , did you intend to close this?

AlisdairM:
No, I pushed an update that went to the wrong repo --- I am disconnected.

jensmaurer:
Ack. I had to make this repo private due to the other pull request about ongoing business, and that disconnected all forks.  Quite a fall-out.

tkoeppe:
@AlisdairM: I reopened this.

AlisdairM:
I believe that this PR should be visible and open again, with the revised comment message that Thomas asked for, or at least my attempt at such a rewording.

jensmaurer (approved):

jensmaurer:
Looks good, but the commit message needs a bit of work.  "Complete X" is not descriptive of the actual changes made in this commit.

tkoeppe:
@AlisdairM Please rebase

AlisdairM:
Rebased, and should be ready to merge.

PR #8414, opened 2025-11-05
[lex.comment] Add a note that comments are replaced by a single space

AlisdairM:
Add a note about the transience of comments, that are immediately replaced by space characters as they are parsed.

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -525,6 +525,12 @@ are treated just like other characters. Similarly, the comment
 characters \tcode{//} and \tcode{/*} have no special meaning within a
 \tcode{/*} comment.
 \end{note}
+\begin{note}
+Comments are turned into \unicode{0020}{space} characters in
+phase 3 of translation as part of decomposing a source file into
+preprocessing tokens and sequences of whitespace characters.
+\end{note}
+
 \indextext{comment|)}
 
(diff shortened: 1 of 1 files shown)


source/lex.tex
```diff
@@ -525,6 +525,12 @@
 characters \tcode{//} and \tcode{/*} have no special meaning within a
 \tcode{/*} comment.
 \end{note}
+\begin{note}
+Comments are turned into \unicode{0020}{space} characters in
+phase 3 of translation as part of decomposing a source file into
+preprocessing tokens and sequences of whitespace characters.
```
jensmaurer, line 531: lex.phases p3 says a bit much in that location.

Maybe it's better to move the normative provision "Each comment (5.4) is replaced by one u+0020 space character." here or into lex.pptoken, possibly as part of the "whitespace grammar" paper.

jensmaurer (changes requested):

source/lex.tex
AlisdairM: I was reluctant to touch the phases of translation more than necessary in P3657, the whitespace grammar paper.  However, if we prefer that direction I will updated the next revision of that paper and close this PR.

AlisdairM:
Will propose more significant changes as part of P3657.

PR #8418, opened 2025-11-05
[lex.charset] Move reference to glyphs to appropriate place

AlisdairM:
The statement that glyphs are used to identify members of the basic character set does not belong separating two sentences introducing and then defining preprocessing tokens.

Also, we do not *exlusively* use glyphs for this purpose but also directly call out Unicode code points too, so tone down the phrasing to glyphs are *often* used to ...

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -317,6 +317,9 @@ A Unicode scalar value is any code point that is not a surrogate code point.
 \pnum
 The \defnadj{basic}{character set} is a subset of the translation character set,
 consisting of 99 characters as specified in \tref{lex.charset.basic}.
+In this document,
+glyphs are often used to identify
+elements of the basic character set.
 \begin{note}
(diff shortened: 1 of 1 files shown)


source/lex.tex
```diff
@@ -317,6 +317,8 @@
 \pnum
 The \defnadj{basic}{character set} is a subset of the translation character set,
 consisting of 99 characters as specified in \tref{lex.charset.basic}.
+In this document, glyphs are often used to identify elements of
+the basic character set.
```
jensmaurer: Please keep the semantic linebreaks of the original.
https://sembr.org/

tkoeppe:
@AlisdairM: Please rebase.

AlisdairM:
This has been rebased, and should be ready to merge.

jensmaurer:
@tkoeppe , this is ready now.

PR #8432, opened 2025-11-05
[std.tex] Suppress hyperref warnings introduced in #8160

morinmorin:


tkoeppe:
Brilliant, thank you!

PR #8435, opened 2025-11-06
[meta.reflection.layout] Reverse logic in specification

k3DW:
Fixes NB US 106-167 (C++26 CD).

Fixes https://github.com/cplusplus/nbballot/issues/985


jensmaurer:
LWG leadership is in unison this is editorial, apparently.

PR #8437, opened 2025-11-06
[meta.syn] Synchronize `reflect_constant`/`reflect_object` parameters with definition

### LINKED ISSUE #8244: Mismatch in reflect_constant declaration
[meta.reflection] first claims:

```
// [meta.reflection.result], expression result reflection
  template<class T>
    consteval info reflect_constant(const T& value);
```

but later in [[meta.reflection.result]](https://eel.is/c++draft/meta.reflection.result) we say:

```
template<class T>
  consteval info reflect_constant(T expr);
```

I think the latter one is correct.

eisenwave:
Fixes #8244.

This synchronizes the function signature with the definition in https://eel.is/c++draft/meta.reflection.result#lib:reflect_constant

The paper author has confirmed that the CWG intent is for the parameter to be `T value`; the use of `const T&` in the synopsis is an editorial oversight.

### DIFF
--- a/source/meta.tex
+++ b/source/meta.tex
@@ -3109,9 +3109,9 @@ namespace std::meta {
 
   // \ref{meta.reflection.result}, expression result reflection
   template<class T>
-    consteval info reflect_constant(const T& value);
+    consteval info reflect_constant(T expr);
   template<class T>
-    consteval info reflect_object(T& object);
+    consteval info reflect_object(T& expr);
   template<class T>
     consteval info reflect_function(T& fn);
 


eisenwave:
Now looking at it, we also have at least one mismatch between the parameter names in the synopsis (`value`) and in the definitions (`expr`). That's a much less severe problem than function signature differences though.

jensmaurer:
Can you rename the function parameter in the synopsis to "expr" as well?

Otherwise, looks good.

tkoeppe:
Great, thanks!

PR #8440, opened 2025-11-06
Review [lex]

AlisdairM:
Review [lex] for odd whitespace issues

### DIFF
--- a/source/meta.tex
+++ b/source/meta.tex
@@ -5563,9 +5563,9 @@ if a $Q$-members-of-eligible declaration of $M$ members-of-precedes $P$,
 and $M$ is
 \begin{itemize}
 \item
-  a class or enumeration type
+  a class or enumeration type,
 \item
-  a type alias
+  a type alias,
 \item
   a class template, function template,
   variable template, alias template, or concept,


source/classes.tex
```diff
@@ -36,7 +36,7 @@
 
 \begin{bnf}
 \nontermdef{class-head}\br
-    class-key \opt{attribute-specifier-seq} class-head-name \opt{class-property-specifier-seq} \opt{base-clause}\br
+    class-key \opt{attribute-specifier-seq} class-head-name \opt{class-virt-specifier} \opt{base-clause}\br
```
jensmaurer: I think we should retain the rename to "class-property-specifier".
To be discussed in core.

tkoeppe:
(This seems somewhat premature; we haven't even got a motion yet.)

source/classes.tex
Quuxplusone: FWIW, if it's not too late, I would go further: my experimentation indicates that vendors generally _do_ treat this thing as a `-seq` in order to diagnose mistakes like `struct A final final {}` as "repeated `final`" instead of "expected `{`". So I think it would actually be useful to keep P2786's _whole_ grammar change here, not just the change of `virt` to `property` but also the addition of `-seq` and the sentence forbidding repetitions, as (I think) it better reflects existing practice.
Yes, even though it'd be a `-seq` of only a single possible specifier and repeating that specifier would be ill-formed — it's still existing practice. Or so I think.

PR #8442, opened 2025-11-07
[utilities] Use "Result:" element in \itemdescrs of types to describe the type.

tkoeppe:
This replaces the use of the ad-hoc element "Type:" in three places with "Result", and adds "Result:" in other cases that didn't have an element at all.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -1408,7 +1408,7 @@ template<size_t I, class T1, class T2>
 $\tcode{I} < 2$.
 
 \pnum
-\ctype
+\result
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -1644,6 +1644,7 @@
 
 \begin{itemdescr}
 \pnum
+\result
```
jwakely: This isn't defining a type or an expression, it's stating the requirements of a concept.

tkoeppe: Ugh, right, yes. I noticed that initially, but not when I did the blind search and replace. Reverted!

PR #8448, opened 2025-11-08
[basic.fundamental] Remove note which explains the purpose of `int`

### LINKED ISSUE #6825: [basic.fundamental] Should we remove Note 1 on `int` having the "natural width"?
> Plain `int`s are intended to have the natural width suggested by the architecture of the execution environment; the other signed integer types are provided to meet special needs.

[[basic.fundamental] Note 1](https://eel.is/c++draft/basic.fundamental#note-1)

I feel like this note is not doing much good at this point. Firstly, this sounds a lot like a *Recommended practice* paragraph, not like a note.

Besides that, it's not what implementers typically do, or should do. The "natural width" is the width of the general purpose register or the width of pointers; anything less will usually require zero- or sign-extension:
```cpp
char* advance(char* p, int x) {
    return p + x;
}
char* advance(char* p, long long x) {
    return p + x;
}
```
Clang emits:
```asm
advance(char*, int):
  movsxd rax, esi
  add rax, rdi
  ret
advance(char*, long long):
  lea rax, [rdi + rsi]
  ret
```

If implementers respected this note, `int` should be a 64-bit type on x86_64, which would make it "natural" and drop sign extensions. I don't believe this is useful or desirable; the general approach is to cap the `int` size at 32-bit, even if this makes it less natural and requires additional operations.

eisenwave:
Fixes #6825.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -5331,11 +5331,6 @@ The range of representable values for a signed integer type is
 $-2^{N-1}$ to $2^{N-1}-1$ (inclusive),
 where $N$ is called the \defn{width} of the type.
 \indextext{integral type!implementation-defined \tcode{sizeof}}%
-\begin{note}
-Plain \tcode{int}s are intended to have
-the natural width suggested by the architecture of the execution environment;
-the other signed integer types are provided to meet special needs.
-\end{note}
 
 \pnum
 \indextext{type!\idxcode{unsigned}}%


jensmaurer:
Approved via CWG reflector review: http://lists.isocpp.org/core/2026/07/19806.php

jensmaurer (approved):

PR #8453, opened 2025-11-08
[inplace.vector.modifiers] Ensure correct type of returned iterator

### LINKED ISSUE #7486: [inplace.vector.modifiers] Revert wording for the Returns of `try_append_range`?
The wording of _Returns_ for `try_append_range` in the [original paper](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p0843r14.html) is:

> _Returns_: Iterator past last inserted element of `rg`.

It was changed after https://github.com/cplusplus/draft/pull/7115/commits/37a6c0dc80d8c2a54559a7515a4de2734f1b6242 to:

> _Returns_: An iterator pointing to the first element of `rg` that was not inserted into `*this`, or `ranges​::​end(rg)` if no such element exists.

This doesn't seem right because `ranges::end` might be different than `ranges::begin` making it impossible to convert to `borrowed_iterator_t<R>`. Given that there is no description for such a situation in the [container] as described in [[algorithm#requirements-13]](https://eel.is/c++draft/algorithms#requirements-13):

>  In the description of algorithm return values, a sentinel value s denoting the end of a range `[i, s)` is sometimes returned where an iterator is expected. In these cases, the semantics are as if the sentinel is converted into an iterator using `ranges​::​next(i, s)`.

Should we revert the current wording to make it consistent with the original wording?

@jwakely @tkoeppe

eisenwave:
Fixes #7486.

This is largely using @jwakely's proposed wording from https://github.com/cplusplus/draft/issues/7486#issuecomment-2546353089, with minor adjustments.

The PR fixes a normative defect that was introduced editorially in #7115. I assume that this needs no LWG issue because the defect is relatively new and wouldn't have existed without editorial errors anyway.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -11251,9 +11251,10 @@ Each iterator in the range \tcode{rg} is dereferenced at most once.
 
 \pnum
 \returns
-An iterator pointing to the first element of \tcode{rg}
+The first iterator in the range
+\countedrange{ranges::begin(rg)}{n}
 that was not inserted into \tcode{*this},
-or \tcode{ranges::end(rg)} if no such element exists.
+where \tcode{n} is the number of elements in \tcode{rg}.
 
 \pnum
 \complexity


jwakely:
Why not use my +[0,n) suggestion?

jwakely:
as defined by https://eel.is/c++draft/iterator.requirements.general#11

eisenwave:
I've never seen that syntax before and got spooked. I didn't even know there is a `\countedrange` macro.

I'll change it to what you've suggested.

jwakely (approved):

tkoeppe:
Thanks!

PR #8486, opened 2025-11-09
[container.requirements] Add index for inplace_vector.

YexuanXiao:


eisenwave:
Note that the commit message contains a period, and we usually don't punctuate our commit messages like that.

PR #8489, opened 2025-11-09
[lib] Adjust \indexlibrarymember{class}{member} to \indexlibrarymembe…

YexuanXiao:
…r{member}{class}

According to the [Style Guidelines](https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#indexes), the `\indexlibrarymember` command should take the member name as its first parameter and the class name as the second. After reviewing all existing `\indexlibrarymember` calls, I found the following classes violate this rule (some even use different orders within the same class): 

+ array
+ span
+ locale
+ ctype
+ codecvt
+ num_get
+ num_put
+ numpunct
+ collate
+ time_get
+ money_get
+ money_put
+ moneypunct
+ messages
+ regex_traits
+ basic_regex
+ sub_match
+ match_results
+ regex_token_iterator
+ thread

(note: "span" is also used as a member name in iostreams.tex, but the iostreams.tex file does not contain such issues).

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -6154,10 +6154,10 @@ The types \tcode{iterator} and \tcode{const_iterator} meet
 the constexpr iterator requirements\iref{iterator.requirements.general}.
 
 \indexlibraryglobal{array}%
-\indexlibrarymember{array}{begin}%
-\indexlibrarymember{array}{end}%
-\indexlibrarymember{array}{size}%
-\indexlibrarymember{array}{max_size}%
+\indexlibrarymember{begin}{array}%
+\indexlibrarymember{end}{array}%
+\indexlibrarymember{size}{array}%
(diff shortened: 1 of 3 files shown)


eisenwave:
The same inconsistency surprised me as well. However, the `\indexlibrarymember` macro is commutative anyway:
```tex
\newcommand{\indexlibrarymember}[2]{\indexlibraryboth{\idxcode{#1}}{\idxcode{#2}}}
```
... so this issue doesn't have any visible impact to the reader, even though the markup inconsistency is surprising.

YexuanXiao:
> ... so this issue doesn't have any visible impact to the reader, even though the markup inconsistency is surprising.

I noticed this issue while trying to add missing member indexes (which I plan to proceed with after this PR is merged). The inconsistency was confusing, so it would be best to correct it according to the Guidelines.

Issue #8494, opened 2025-11-09
[checks] Add check for trailing `\br` in a `bnf` environment


eisenwave:
It's quite easy to make a mistake such as:
```tex
\begin{bnf}
\nontermdef{line-directive}\br
    \terminal{\# line} pp-tokens new-line\br
\end{bnf}
```

If I recall correctly, this trailing `\br` is not ignored, but simply renders as an unwanted blank line. I have made this error during 2025 post-Sofia motions I believe.

It seems like it would be fairly easy to grep for `\br`, followed by whitespace,
followed by `\end{bnf}`.

eisenwave:
I was able to found a few violations using the following command:
```sh
pcregrep -M -n -r '\\br\s+\\end' *.tex
```

PR #8498, opened 2025-11-10
[namespace.alias], [over.oper.general] Remove trailing `\br` in grammar

eisenwave:
Using
```sh
pcregrep -M -n -r '\\br\s+\\end' *.tex
```
I was able to find a few bugs where `\br` was added at the end of a grammar. This results in a bogus line break:
image

These are the only two such occurrences.

We should also make a check that prevents this from happening again: #8494


tkoeppe:
Thank you, yes, this is quite a common mistake, and if we could check for it, that'd be helpful.

PR #8500, opened 2025-11-11
[class.union.anon] Tighten definition of "variant member"

hubert-reinterpretcast:
Rephrase the definition of variant members to address the following problems with the current wording:

- It creates a misleading reading of the "[in] addition, [ ... ]" sentence as applying only to the case where `X` is a union.
- It is not phrased as complete/exhaustive.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -3410,10 +3410,10 @@ in~\ref{dcl.init.aggr}.
 \pnum
 \indextext{class!variant member of}%
 A \defnadj{union-like}{class} is a union or a class that has an anonymous union as a direct
-member. A union-like class \tcode{X} has a set of \defnx{variant members}{variant member}.
-If \tcode{X} is a union, a non-static data member of \tcode{X} that is not an anonymous
-union is a variant member of \tcode{X}. In addition, a non-static data member of an
-anonymous union that is a member of \tcode{X} is also a variant member of \tcode{X}.
+member.
+The \defnx{variant members}{variant member} of a union-like class \tcode{X} are the variant
+members of the anonymous union members of \tcode{X} and, if \tcode{X} is a union, the
+non-static data members of \tcode{X} other than anonymous union members.
 At most one variant member of a union may have a default member initializer.
 \begin{example}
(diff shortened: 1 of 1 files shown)


source/classes.tex
```diff
@@ -3410,10 +3410,10 @@
...
 \indextext{class!variant member of}%
 A \defnadj{union-like}{class} is a union or a class that has an anonymous union as a direct
-member. A union-like class \tcode{X} has a set of \defnx{variant members}{variant member}.
-If \tcode{X} is a union, a non-static data member of \tcode{X} that is not an anonymous
-union is a variant member of \tcode{X}. In addition, a non-static data member of an
-anonymous union that is a member of \tcode{X} is also a variant member of \tcode{X}.
+member.
+The \defnx{variant members}{variant member} of a union-like class \tcode{X} are the variant
+members of the anonymous union members of \tcode{X} and, if \tcode{X} is a union, the
+non-static data members of \tcode{X} other than anonymous union members.
```
eisenwave, line 3416:
```suggestion
The \defnx{variant members}{variant member} of a union-like class \tcode{X} are
the variant members of the anonymous union members of \tcode{X}, and
if \tcode{X} is a union, the non-static data members of \tcode{X} other than anonymous union members.
```

If we're touching this section anyway, it seems beneficial to reflow the text so that if it was turned into bullets, we'd only need to add `\item`s at the start of each line.

```tex
The \defnx{variant members}{variant member} of a union-like class \tcode{X} are
\begin{itemize}
\item the variant members of the anonymous union members of \tcode{X}, and
\item if \tcode{X} is a union, the non-static data members of \tcode{X} other than anonymous union members.
\end{itemize}
```

eisenwave: Personally, I don't think this paragraph is *that* broken though. The suggestion seems like a slight upgrade.

hubert-reinterpretcast: The placement of the comma in the suggestion is less grammatical as the list is of two items and the commas are parenthetical (around the condition introduced with the "if") in nature.

PR #8501, opened 2025-11-11
[special] Add "direct" when defining "potentially constructed subobjects"

hubert-reinterpretcast:
The definition has "non-static data members". Only the direct ones are intended.


eisenwave:
P1. Normative wording defect with obvious fix.

Issue #4939, opened 2021-09-26
An anonymous union member is not a non-static data member on grammar


xmh0511:
[class.mem.general] p3 explicitly says
> A data member is a non-function member introduced by a **member-declarator**.  

While [class.union.anon] p1 says
> A union of the form *union { member-specification } ;* is called an anonymous union; it defines an unnamed type and an unnamed object of that type called an anonymous union member **if it is a non-static data member** or an anonymous union variable otherwise. 

Such two rules contradict each other.  Maybe, we should say
> An anonymous union behaves as if it were replaced by
>> *union { member-specification } unnamed;*

jensmaurer:
The fall-out for that approach is a bit wide in scope. I think just extending the sentence in [class.mem.general] p3 for the anonymous union case is good enough.

xmh0511:
Agree, that would be a minimal modification that wouldn't cause other conflictions.

PR #8502, opened 2025-11-11
[class.mem.general] Fix data member definition to include anonymous union members

hubert-reinterpretcast:
Anonymous union members are not introduced by _member-declarator_.

Fixes cplusplus/draft#4939.

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -669,8 +669,8 @@ declare at least one member name of the class
 or declare at least one unnamed bit-field.
 
 \pnum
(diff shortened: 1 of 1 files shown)


eisenwave:
Tbh I don't see how changing the definition of "data member" is editorial, but the fix is relatively obvious, and Jens already suggested to do that in #4939 without marking the issue `not-editorial`.

tkoeppe:
Thanks, @hubert-reinterpretcast! Let's give Jens a final chance to take a look.

jensmaurer (approved):

PR #8503, opened 2025-11-11
[class.base.init] Add "direct" for _mem-initializer-id_-named members

hubert-reinterpretcast:
A _mem-initializer-id_ cannot be used to initialize a base class data member from a derived class constructor; therefore, we mean _direct_ non-static data member.


eisenwave:
P1. Looks like a normative wording defect with obvious fix.

PR #8507, opened 2025-11-12
[optional.ref.assign] Add missing Returns element

eisenwave:
Notice that the return type is `T&`, but *`convert-ref-init-val`* is a `void` member function.

The intent is obviously to return `*val`; this seems like an editorial slip-up (unfortunately resulting in a normative defect).

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -4748,6 +4748,10 @@ template<class U>
 \pnum
 \effects
 Equivalent to: \tcode{\exposid{convert-ref-init-val}(std::forward<U>(u))}.
+
+\pnum
+\returns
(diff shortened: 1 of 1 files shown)


jwakely:
i.e. nothing in the current wording says it returns the result of _`convert-ref-init-val`_ so the fact it's a void expression is irrelevant.

jensmaurer (approved):

eisenwave:
You will be shocked to find out what we've just merged is *exactly* the solution in LWG4300, which was just approved in Kona.

jwakely:
> We'd usually just say _Returns:_ `*val` in cases like this.

Well at least I was consistent :sweat_smile:

PR #8518, opened 2025-11-14
[cpp.cond] Keywords are not identifiers while preprocessing

AlisdairM:
Move seemingly normative reference to keywords in #if expressions into a note.

### DIFF
--- a/source/preprocessor.tex
+++ b/source/preprocessor.tex
@@ -553,11 +553,11 @@ evaluations of
 \grammarterm{has-embed-expression}s, and
 \grammarterm{has-attribute-expression}s
 have been performed,
-all remaining identifiers (including those lexically identical to keywords)
+all remaining identifiers (including those lexically identical to keywords),
 except for
 \tcode{true}
 and
-\tcode{false}
+\tcode{false},
 are replaced with the \grammarterm{pp-number}
 \tcode{0},
 and then each preprocessing token is converted into a token.


source/preprocessor.tex
```diff
@@ -553,15 +553,19 @@
...
 and
-\tcode{false},
+\tcode{false}
 are replaced with the \grammarterm{pp-number}
 \tcode{0},
 and then each preprocessing token is converted into a token.
 \begin{note}
+Keywords are identifiers during preprocessing and are not valid macro names,
+so are always replaced by \tcode{0} before being converted into a token.
+\end{note}
```
jensmaurer: Do we need this note after we've removed "keyword" in the first place?

We do mention "true" and "false" as identifiers above, so there is a strong hint that phase 7 keywords are in view here.
If keywords were a separate thing, the note below would need to be amended with "keywords", too, I think.

AlisdairM: We mention `true` as, being an identifier that is not defined as a macro, it should convert to a `0` before we turn the preprocessor tokens into tokens in order to evaluate the `#if` predicate --- we arrange that the only tokens that convert are literals.  I think the first note covers that, and the alternative tokens note covers a different matter as alternative tokens are never identifiers.  Maybe I am too focused on identifiers as a preprocessing-token that is being converted though?

jensmaurer: I was trying to say that your new "note" can go entirely, because "true" is a keyword and is mentioned in the (revised) "identifiers except..." phrasing.  That wouldn't be necessary if "true" weren't considered a (preprocessor-level) identifier at that point.

AlisdairM: I am happy to remove the note --- should I do so now, or wait for more feedback in case others prefer to add the note?  I agree that my preference would be to remove the note as I doubt folks would be genuinely confused.

jensmaurer: For comparison, C23 says

> ... all remaining identifiers other than true (including those lexically identical to keywords such as false) are replaced with the pp-number 0, true is replaced with pp-number 1, and then each preprocessing token is converted into a token.

AlisdairM: Of course, `true` and `false` are essentially integer literals in C, but are boolean literals in C++.  By preserving the spelling of the token, we preserve the type in the expression.  I do not know how significant that change might be, but if we were to synchronize wording with C it would definitely become a CWG issue due to that type change.

jensmaurer: I'm wondering how you would observe the type change at the preprocessor `#if` level.

That said, the more important part for this pull request is the parenthetical "(including those lexically identical to keywords)" which re-emphasizes that phase 4 "identifiers" can be spelled like phase 7 keywords. That's another option, instead of the note.

hubert-reinterpretcast: If the note is kept, please leave it at "Keywords are identifiers during preprocessing and are not valid macro names."

source/preprocessor.tex
```diff
@@ -553,11 +553,11 @@
 \grammarterm{has-embed-expression}s, and
 \grammarterm{has-attribute-expression}s
 have been performed,
-all remaining identifiers and keywords,
+all remaining identifiers (including those lexically identical to keywords)
```
hubert-reinterpretcast:
```suggestion
all remaining identifiers (including those lexically identical to keywords),
```

source/preprocessor.tex
```diff
@@ -553,11 +553,11 @@
...
 except for
 \tcode{true}
 and
-\tcode{false},
+\tcode{false}
```
hubert-reinterpretcast:
```suggestion
\tcode{false},
```

jensmaurer:
I think I like the overall outcome here.

hubert-reinterpretcast (approved):
LGTM. Thanks!

PR #8520, opened 2025-11-14
P3922R1 Missing deduction guide from simd::mask to simd::vec

### LINKED ISSUE #8474: [2025-11 LWG Motion 15] P3922R1 Missing deduction guide from simd::mask to simd::vec
https://wiki.edg.com/pub/Wg21kona2025/StrawPolls/P3922R1.pdf

eisenwave:
Fixes #8474
Fixes cplusplus/nbballot#862
Also fixes cplusplus/papers#2581

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -17574,6 +17574,9 @@ namespace std::simd {
 
   template<class R, class... Ts>
     basic_vec(R&& r, Ts...) -> @\seebelow@;
+
+  template<size_t Bytes, class Abi>
+    basic_vec(basic_mask<Bytes, Abi>) -> @\seebelow@;
 }
 \end{codeblock}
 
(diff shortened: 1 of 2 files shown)


source/numerics.tex
```diff
@@ -17866,6 +17869,22 @@
 ranges::size(r)>}.
 \end{itemdescr}
 
+\begin{itemdecl}
+template<size_t Bytes, class Abi>
+  basic_vec(basic_mask<Bytes, Abi>) -> @\seebelow@;
```
jwakely:
```suggestion
  basic_vec(basic_mask<Bytes, Abi> k) -> @\seebelow@;
```

eisenwave:
> I thought "Fixup" commits were for minor editorial corrections to the paper wording, not for cases where the pull request doesn't correctly implement the paper wording.

I thought so too. Then I had a conversation last motions where it was brought up that we don't want authors to force-push on these motions because it makes reviewing harder or something, so it's better to add a Fixup commit even for things that weren't proper paper applications in the first place.

To be fair, I don't think it would have killed anyone if I force-pushed this one.

PR #8524, opened 2025-11-15
P3567R2 `flat_meow` fixes

### LINKED ISSUE #8463: [2025-11 LWG Motion 4] P3567R2 flat_meow fixes
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p3567r2.html

eisenwave:
Fixes #8463

No NB comment.

Also fixes https://github.com/cplusplus/papers/issues/2211

PROBLEM: https://github.com/cplusplus/papers/issues/2029 added `constexpr` to the members of `flat_meow`, but what we're merging doesn't have any `constexpr` on it. LWG issue? @jwakely

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -17139,10 +17139,10 @@ namespace std {
     // \ref{flat.map.cons}, constructors
     constexpr flat_map() : flat_map(key_compare()) { }
 
-    flat_map(const flat_map&);
-    flat_map(flat_map&&);
-    flat_map& operator=(const flat_map&);
-    flat_map& operator=(flat_map&&);
+    constexpr flat_map(const flat_map&);
+    constexpr flat_map(flat_map&&);
+    constexpr flat_map& operator=(const flat_map&);
+    constexpr flat_map& operator=(flat_map&&);
 
     constexpr explicit flat_map(const key_compare& comp)
       : @\exposid{c}@(), @\exposid{compare}@(comp) { }
@@ -17285,7 +17285,7 @@ namespace std {
     template<@\exposconcept{container-compatible-range}@<value_type> R>
       constexpr void insert_range(R&& rg);
     template<@\exposconcept{container-compatible-range}@<value_type> R>
-      void insert_range(sorted_unique_t, R&& rg);
+      constexpr void insert_range(sorted_unique_t, R&& rg);
 
     constexpr void insert(initializer_list<value_type> il)
       { insert(il.begin(), il.end()); }
@@ -17876,7 +17876,7 @@ Since this operation performs an in-place merge, it may allocate memory.
 \indexlibrarymember{insert_range}{flat_map}%
 \begin{itemdecl}
 template<@\exposconcept{container-compatible-range}@<value_type> R>
-  void insert_range(sorted_unique_t, R&& rg);
+  constexpr void insert_range(sorted_unique_t, R&& rg);
 \end{itemdecl}
 
 \begin{itemdescr}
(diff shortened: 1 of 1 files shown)


jwakely:
There's no need for an issue, the intended design is clear: they should all be constexpr. Please just add constexpr keywords as a fixup commit.

jensmaurer:
> Please just add constexpr keywords as a fixup commit.

Yes, please.

eisenwave:
Alright, I've made them all `constexpr` in a new commit. It's not a fixup commit because I think it's worth having some record of why we deviate from the paper in the commit history, but Thomas can squash it anyway if he doesn't think so.

source/containers.tex
```diff
@@ -19380,9 +19430,27 @@
...
+
+\pnum
+\complexity
+Linear in $N$, where $N$ is \tcode{size()} after the operation.
+\end{itemdescr}
+
 \indexlibrarymember{swap}{flat_set}%
 \begin{itemdecl}
-constexpr void swap(flat_set& y) noexcept;
+constexpr void swap(flat_set& y)
```
jwakely, line 19471: The corresponding changes to [flat.set.defn] for member `swap` and non-member `swap` seem to be missing.

source/containers.tex
```diff
@@ -18425,6 +18460,8 @@
       constexpr void insert(sorted_equivalent_t, InputIterator first, InputIterator last);
     template<@\exposconcept{container-compatible-range}@<value_type> R>
       constexpr void insert_range(R&& rg);
+    template<@\exposconcept{container-compatible-range}@<value_type> R>
+      constexpr void insert_range(sorted_unique_t, R&& rg);
```
```suggestion
      constexpr void insert_range(sorted_equivalent_t, R&& rg);
```

source/containers.tex
```diff
@@ -19702,6 +19777,8 @@
...
```
```suggestion
      constexpr void insert_range(sorted_equivalent_t, R&& rg);
```

source/containers.tex
```diff
@@ -20005,12 +20083,60 @@
...
+});
+\end{codeblock}
+Then, sorts the range of newly inserted elements with respect to \exposid{compare},
+and merges the resulting sorted range and
+the sorted range of pre-existing elements into a single sorted range.
+
+\pnum
+\complexity
+$N$ + $M \log M$, where $N$ is \tcode{size()} before the operation and $M$
+is \tcode{ranges::distance(rg)}.
```
jwakely, line 20131:
```suggestion
$N$ + $M \log M$, where $N$ is \tcode{size()} before the operation
and $M$ is \tcode{ranges::distance(rg)}.
```

(semantic break)

source/containers.tex
```diff
@@ -20005,12 +20083,60 @@
...
+\pnum
+\remarks
+Since this operation performs an in-place merge,
+it may allocate memory.
+\end{itemdescr}
+
+\indexlibrarymember{insert_range}{flat_multiset}%
+\begin{itemdecl}
+template<@\exposconcept{container-compatible-range}@<value_type> R>
+  constexpr void insert_range(sorted_unique_t, R&& rg);
```
```suggestion
  constexpr void insert_range(sorted_equivalent_t, R&& rg);
```

jwakely (changes requested):

source/containers.tex
eisenwave: The chosen line break is consistent with what we have further up for `flat_map::insert` and further below for `flat_set::insert`.

The `and` here is also a conjunction in a list, so it seems more appropriate to put it at the end of the line.

eisenwave:
Should be all fixed now.

Had to force-push because the second commit which adds `constexpr` would be hard to merge with fixups that change `sorted_unique_t` to `sorted_equivalent_` otherwise.

PR #8532, opened 2025-11-15
[exec.run.loop] update run_loop count representation to use code format

ericniebler:
[[run.loop.members]/p2](https://eel.is/c++draft/exec#run.loop.members-2) ends with two periods.

image

i notice also that the whole [exec.run.loop] section is inconsistent about putting numbers in code font. this pr addresses the issue by uniformly applying `\tcode` to numbers.

### DIFF
--- a/source/exec.tex
+++ b/source/exec.tex
@@ -6485,7 +6485,7 @@ run_loop() noexcept;
 \begin{itemdescr}
 \pnum
 \ensures
-The \tcode{run_loop} instance's count is 0 and
+The \tcode{run_loop} instance's count is \tcode{0} and
 its state is starting.
 \end{itemdescr}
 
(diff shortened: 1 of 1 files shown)


eisenwave:
The two-period issue seems to be fixed already. I'm not sure whether it's correct to use code font or not here.

Code font implies that an operation is performed with an `int` literal, and that isn't always what we want to express.

"is `0`" in particular is almost guaranteed to be wrong, unless it's certain that whatever "is `0`" is also of type `int`. It could be argued that `0ll` "equals `0`" or "is 0", but not that it "is `0`".

ericniebler:
I don't care which way we go, but I do want things to be consistent.

As a data point, we always put `true` and `false` in code font.

PR #8533, opened 2025-11-16
[array.zero] Add \tcode for array

hewillk:
I'm pretty sure this refers to `std::array` and not a language array.


hewillk:
But it will be deleted due to LWG [4276](https://cplusplus.github.io/LWG/lwg-defects.html#4276). So close.

PR #8534, opened 2025-11-16
[range.adaptors] Replace "In that case" with "If declared"

### LINKED ISSUE #5224: P2441 Fix "In that case"
Replace "In that case..." with "If defined, ..."

"In that case" is ambiguous, because it might refer to any of three preceding conditions, or the combination of them.

Fix in P2441 and the existing views::join.

eisenwave:
Fixes #5224.

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -4938,7 +4938,7 @@ satisfy the filter predicate.
 \pnum
 The member \grammarterm{typedef-name} \tcode{iterator_category} is declared
 if and only if \tcode{V} models \libconcept{forward_range}.
-In that case,
+If declared,
 \tcode{\exposid{iterator}::iterator_category} is defined as follows:
 \begin{itemize}
 \item Let \tcode{C} denote the type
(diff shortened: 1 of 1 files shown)


jwakely:
"If defined, [it] is defined a follows" is a bit repetitive. 

An alternative would be "X is not defined if V does not model C; otherwise it is defined as follows". I don't really like the double negative in that though.

eisenwave:
What about
> If these conditions are satisfied

The point of the fix is to disambiguate between whether "In that case" refers to only one of the conditions in the preceding sentence, or to all of them combined. If we refer to them in plural, that also works.

jwakely (approved):
I don't love the phrasing, but it's an improvement.

tkoeppe:
Needs some linebreaking/overfull-hbox fixes I think.

eisenwave:
Just needed a soft hyphen; please check again.

source/ranges.tex
```diff
@@ -4938,7 +4938,7 @@
 \pnum
 The member \grammarterm{typedef-name} \tcode{iterator_category} is declared
 if and only if \tcode{V} models \libconcept{forward_range}.
-In that case,
+If defined,
```
jensmaurer: [basic.def] p2.9 says typedef declarations are not definitions, so "If defined" doesn't make sense.

jensmaurer (changes requested):

source/ranges.tex
eisenwave: Right. I've changed it to "If declared".

source/ranges.tex
```diff
@@ -5467,10 +5467,10 @@
 \pnum
 The member \grammarterm{typedef-name} \tcode{iterator_category} is declared
 if and only if \exposid{Base} models \libconcept{forward_range}.
-In that case,
+If defined,
```
jensmaurer: I had hoped you'd fix "defined" -> "declared" everywhere, not just the first occurrence in this pull request.

jensmaurer (changes requested):

PR #8540, opened 2025-11-16
[simd] Add missing `\libconcept` to uses of `same_as`

eisenwave:
This follows the precedent of https://eel.is/c++draft/simd.complex.access#3.2, which says
> `same_as<...>` is modeled

We could probably also use `is_same_v` here instead, but that would create a local inconsistency with [simd.complex.access], and the three uses here are clearly wrong.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -18568,7 +18568,7 @@ Let
  \item
    \tcode{ranges::range_value_t<R>} is a vectorizable type,
  \item
-   \tcode{same_as<remove_cvref_t<V>, V>} is \tcode{true},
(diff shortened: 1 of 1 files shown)


hewillk:
There are nothing wrong with "If `same_as<A, B>` is `true`", we use a lots in `<ranges>`, see https://eel.is/c++draft/ranges#range.iota.view-10.1
The statement "If `same_as<A, B>` is model" is unnecessary because we are only checking the constraint here, and we are not asking about its semantics, which is much simpler.
I honestly don't see any improvement from `is_same_v`, except for being more verbose.

PR #8543, opened 2025-11-16
P3913R1 Optimize for std::optional in range adaptors

### LINKED ISSUE #8481: [2025-11 LWG Motion 22] P3913R1 Optimize for std::optional in range adaptors
https://wiki.edg.com/pub/Wg21kona2025/StrawPolls/P3913R1.pdf

steve-downey:
Fixes #8481
Fixes https://github.com/cplusplus/nbballot/issues/783
Also fixes cplusplus/papers#2575

Fixes NB PL-011 (C++26 CD).

### DIFF
--- a/source/ranges.tex
+++ b/source/ranges.tex
@@ -5899,6 +5899,11 @@ except that the evaluations of \tcode{E} and \tcode{F}
 are indeterminately sequenced.
 
 \item
+Otherwise, if \tcode{T} is a specialization of
+\tcode{optional} and \tcode{T} models \libconcept{view}, then
+\tcode{(static_cast<D>(F) == D() ? ((void)E, T()) : \exposidnc{decay-copy}(E))}.
+
+\item
 Otherwise, if \tcode{T} models
 \libconcept{random_access_range} and \libconcept{sized_range}
 and is a specialization of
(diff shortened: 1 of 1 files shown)


source/ranges.tex
```diff
@@ -5898,6 +5898,10 @@
 except that the evaluations of \tcode{E} and \tcode{F}
 are indeterminately sequenced.
 
+\item
+Otherwise, if \tcode{T} is a specialization of
+\tcode{optional} and \tcode{T} models \tcode{view}, then \linebreak{}
+\tcode{(static_cast<D>(F) == D() ? ((void)E, T()) : \placeholdernc{decay-copy}(E))}.
```
eisenwave:
```suggestion
\item
Otherwise, if \tcode{T} is a specialization of
\tcode{optional} and \tcode{T} models \libconcept{view}, then
\tcode{(static_cast<D>(F) == D() ? ((void)E, T()) : \exposid{decay-copy}(E))}.

```
- *`decay-copy`* is an expositon-only function, so `\exposid` is correct
- forced `\linebreak` is problematic; it's better to use a line break opportunity like `\brk{}` somewhere, or add soft hyphens (`\-`) into words to allow paragraph breaking in the right spot
- all library concepts should be wrapped in `\libconcept`
- separating blank line for consistency with other bullets

Please also apply those changes to the other added wording

steve-downey: It looks like decay-copy in the text just before is now `\exposidnc ` ?  

```
then \tcode{((void)F, \exposidnc{decay-copy}(E))},
```

Latex's natural inclination is to break just before the `?` in the ternary expression, which is less awful than when it decides be-gin is an acceptable break elsewhere. I'll try less drastic hints, otherwise I'll remove.

Will add the blank line for items.

eisenwave (approved):
The paper transciption and markup look correct now.

The commit messages aren't exactly in the right format yet (https://github.com/cplusplus/draft/wiki/Applying-papers-to-the-Working-Draft), but Thomas can easily reword those when applying the motions.

jwakely (approved):

PR #8552, opened 2025-11-18
[atomics.types.int] Use the terms "character type" and "standard integer type" instead of listing each type

eisenwave:
We have the terms "character type" and "standard integer type", so we may as well use them in the standard library instead of listing all types within those sets exhaustively.

It's really difficult to tell what the design intent is when you have a list of 15 types and need to figure out which ones are on this list, and which types are missing (but you would have expected to see).

There is also wording asymmetry: https://eel.is/c++draft/atomics.types.float#1 simply says there is a specialization for each "cv-unqualified floating type" in bulk, instead of listing `float`, `double,` and `long double`.

### DIFF
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -4961,23 +4961,10 @@ This function is an atomic notifying operation\iref{atomics.wait}.
 \indexlibrary{\idxcode{atomic<\placeholder{integral-type}>}}%
 \pnum
 There are specializations of the \tcode{atomic}
-class template for the integral types
-\tcode{char},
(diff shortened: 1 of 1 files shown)


jwakely:
> There is also wording asymmetry: https://eel.is/c++draft/atomics.types.float#1 simply says there is a specialization for each "cv-unqualified floating type" in bulk, instead of listing `float`, `double,` and `long double`.

Because that includes `std::float32_t` etc. and it's impl-defined which of the extended floating-point types are supported.

PR #8554, opened 2025-11-19
[basic.fundamental] Exhaustively list what values can be represented by floating-point types P3938

eisenwave:
This is branching off from https://github.com/cplusplus/CWG/issues/814#issuecomment-3548366174

There are various mentions of NaNs, positive and negative zeros, infinities, etc. scattered throughout the standard. We do a bad job of comprehensively listing these possibilities in one place.

It's also not obvious whether we consider floating-point types to *require* having a positive and negative zero, positive and/or negative infinity, NaN values, and whatnot. This is really difficult to figure out.

My understanding of the wording is that floating-point types are only *guaranteed* to represent some rational numbers; anything else, such as NaNs, infinities, negative zeros, etc. may or may not exist, based on the representation of `float` chosen by the implementation.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -5560,6 +5560,24 @@ the object and value representations and accuracy of operations
 of floating-point types are \impldef{representation of floating-point types}.
 
 \pnum
+A floating-point type shall at least represent a subset of rational numbers.
+Depending on the implementation-defined value representation for the type,
+it may additionally represent the non-finite values
+\begin{itemize}
+\item infinity,
+\item a set of quiet ``Not a Number'' values, and
+\item a set of signaling ``Not a Number'' values.
+\end{itemize}
+For any of the above (including zero),
+a floating-point type may either represent a single value or
+two distinct values with negative and positive sign.
+\begin{note}
+A floating-point type which adheres to \IsoFloatUndated{}
+is capable of representing a negative and positive variant
+of all the above\iref{numeric.limits.members}.
+\end{note}
+
+\pnum
 The minimum range of representable values for a floating-point type is
 the most negative finite floating-point number representable
 in that type through


source/basic.tex
```diff
@@ -5549,6 +5549,19 @@
...
 are collectively termed \defnx{floating-point types}{type!floating-point}.
 \begin{note}
+A floating-point type can represent
+\begin{itemize}
+\item a subset of rational numbers,
+\item possibly, a negative zero,
+\item possibly, a positive infinity,
+\item possibly, a negative infinity,
+\item possibly, a set of quiet ``Not a Number'' values, and
+\item possibly, a set of signaling ``Not a Number'' values.
```
jensmaurer: Hm... Can we clarify that those "possibly" are implementation-defined?
Maybe that requires changing the structure of the sentence a bit.

eisenwave: Done. It did require a bit of a rewrite though.

source/basic.tex
```diff
@@ -5549,6 +5549,20 @@
 The standard and extended floating-point types
 are collectively termed \defnx{floating-point types}{type!floating-point}.
 \begin{note}
+A floating-point type can represent at least a subset of rational numbers.
```
jwakely: Maybe mention finite here. Apart from signed zero, the special cases listed below are not finite.

If you're trying to be exhaustive, there are also subnormals (but I don't think being exhaustive is necessary).

eisenwave: I think you're onto something. We never define which values are finite and which ones are not, and having a definition (with index entry) for this would provide some clarity. Larger rewrites needed though; this can't be a note.

Including subnormals in the set of representable values would be odd; "subnormal" is just an arbitrary range within the representable rational numbers, mathematically speaking. It makes sense to draw the distinction in other places, but I'm not sure what to say when we're just listing what values can be represented.

jwakely: I don't think a definition of finite belongs here, that's what 60559 is for. This should stay note-y.

eisenwave: Hmm alright, I've struck the `\defnadjx` part.

I'd be cautious about delegating too much to 60559 because we support types that don't adhere to it, but I suppose we can at least delegate the classification of "finite values" to it, regardless.

source/basic.tex
```diff
@@ -5559,6 +5559,22 @@
 the object and value representations and accuracy of operations
 of floating-point types are \impldef{representation of floating-point types}.
 
+\pnum
+A floating-point type shall at least represent a subset of rational numbers
+called \defnadjx{finite}{values}{value}.
+Depending on the implementation-defined value representation for the type,
+it may additionally represent the finite value negative zero,
```
jwakely: Talking about negative zero as distinct from the other zero implies that the other one is naturally positive, but mathematically that's not true. It's true for IEEE floats but not mathematics. It might be better to talk about zero being signed, rather than saying there can also be negative zero.

I haven't come up with a good way to phrase that though.

jwakely: Maybe say something like there are finite values, and maybe infinities, and NaNs, and they might all have a sign, including zero.

eisenwave: N5014 already uses "negative zero" 9 times, and C has some mentions of it too. It doesn't make much mathematical sense, but I really don't see how we could avoid saying "negative zero" when that's the established term of the last 30 years; we use that anywhere else.

We also mention "negative NaN values" in [cmp.alg].

https://github.com/cplusplus/CWG/issues/817 clarifies that neither negative NaNs nor negative zero are negative values. I can see how this is all a bit paradoxical and unhelpful, but I wouldn't be comfortable with avoiding this terminology *only* in [basic.fundamental].

jwakely: My objection isn't to mentioning negative zero at all, only in this context of something definition-like where you're mentioning it _in addition to_ some other value, which either implies we have "zero and negative zero" or that it's just obvious that the other one is positive zero. But neither of those things is so obviously true that it doesn't need clarifying. A floating-point type might just have a zero value (or several equivalent zero values), without any sign, or it might have signed zeros, in which case it has both positive zero and negative zero. So I think introducing negative zero should be in the context of zero being (possibly, depending on the implementation) signed. Because that implies a positive zero too, not just "_real_ zero and some negative zero thing".

jwakely: I.e. the existence of negative zero in the IEEE formats is a direct consequence of zero (and all other values) having a sign bit. 

I've just discovered that C23 now *requires* a sign bit for all values:
https://cigix.me/c23#5.2.5.3.3.p8

C17 didn't: 
https://cigix.me/c17#5.2.4.2.2.p4

jwakely: Maybe this should be handled by a normative change for C++29 to also require that. If WG14 thinks implementations have converged sufficiently to require that, it should be ok for C++ too.

eisenwave: Makes sense. I've added the requirement that (positive) zero has to be representable now. This is evident from zero-initialization being unconditionally defined and other parts of the standard, so it's still editorial.

I think it's clear enough that in this paragraph "negative zero" is a counterpart to "(positive) zero". I've run out of concrete ideas for how to word it better. Negative zero is still a distinct value (even though not mathematically), and the more you try to treat is a single zero value with two possible signs, the more this important clarification is lost.

jwakely: I don't think it makes sense to normatively refer to positive zero, because a format that doesn't have a signed zero doesn't have positive zero, it just has zero.

Now if we follow C23 and require signed zeros, that isn't a problem - but that would be a normative change.

eisenwave: Ok, I think I figured out a way to phrase it. The new wording says that any one of those values listed may have two distinct negative/positive versions.

By using the word "distinct", it's still clear that those aren't two identical values that have a quasi-padding sign bit, but that they are technically distinct values.

jwakely: I like that.

jwakely: > I.e. the existence of negative zero in the IEEE formats is a direct consequence of zero (and all other values) having a sign bit.
> 
> I've just discovered that C23 now _requires_ a sign bit for all values: https://cigix.me/c23#5.2.5.3.3.p8
> 
> C17 didn't: https://cigix.me/c17#5.2.4.2.2.p4

I misread the new C23 wording, it still allows unsigned zero. "Floating types shall be able to represent signed zeros **or an unsigned zero**"

eisenwave: ... and I assume we don't change that in any way? The currently proposed paragraph seems to be fine then.

jensmaurer:
I hope it's clear we're not adding normative stuff ("shall") using an editorial issue.

eisenwave:
Now handled by
- https://github.com/cplusplus/papers/issues/2592

PR #8556, opened 2025-11-20
[over.match.ref] Restore resolution of CWG2267 after applying P1787R6

jensmaurer:
Paper P1787R6 shows "or 'cv2 T2'" as existing text and leaves it unchanged, even though that was struck by CWG2267.  Other amendments made by CWG2267 are correctly shown as pre-existing text.


jensmaurer:
This is wrong, and conflicts with CWG3073.

PR #8561, opened 2025-11-25
P3612R1 Harmonize proxy-reference operations

### LINKED ISSUE #8471: [2025-11 LWG Motion 12] P3612R1 Harmonize proxy-reference operations (LWG 3638 and 4187)
https://wiki.edg.com/pub/Wg21kona2025/StrawPolls/P3612R1.html

AlisdairM:
Fixes #8471
Fixes cplusplus/papers#2419

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -10482,10 +10482,10 @@ namespace std {
     public:
       constexpr reference(const reference& x) noexcept;
       constexpr ~reference();
-      constexpr operator bool() const noexcept;
       constexpr reference& operator=(bool x) noexcept;
       constexpr reference& operator=(const reference& x) noexcept;
       constexpr const reference& operator=(bool x) const noexcept;
+      constexpr operator bool() const noexcept;
       constexpr void flip() noexcept;   // flips the bit
 
       friend constexpr void swap(reference x, reference y) noexcept;
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -10523,8 +10523,8 @@ namespace std {
       constexpr reference& operator=(bool x) noexcept;              // for \tcode{b[i] = x;}
       constexpr reference& operator=(const reference& x) noexcept;  // for \tcode{b[i] = b[j];}
       constexpr const reference& operator=(bool x) const noexcept;
-      constexpr bool operator~() const noexcept;                    // flips the bit
       constexpr operator bool() const noexcept;                     // for \tcode{x = b[i];}
+      constexpr bool operator~() const noexcept;                    // flips the bit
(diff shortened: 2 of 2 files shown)


AlisdairM:
I would like to create a new subsection for each nested reference class to avoid the need to qualify each \itemdecl with the unusual `reference::` qualifier, and also to better organize these specifications out of appearing out of nowhere in the surrounding text that is talking about the surrounding class as a whole before specifying its own members in following subclauses.

source/utilities.tex
```diff
@@ -10409,13 +10409,18 @@
     // bit reference
     class reference {
     public:
-      constexpr reference(const reference&) = default;
+      constexpr reference(const reference&) noexcept;
```
eisenwave:
```suggestion
      constexpr reference(const reference& x) noexcept;
```
The paper adds an `x` here.

source/containers.tex
```diff
@@ -10480,13 +10480,17 @@
     // bit reference
     class @\libmember{reference}{vector<bool>}@ {
     public:
-      constexpr reference(const reference&) = default;
+      constexpr reference(const reference&) noexcept;
```
eisenwave:
```suggestion
      constexpr reference(const reference& x) noexcept;
```
There is also no `x` in the paper though. However, the paper does add an `x` in the synopsis of `bitset::reference`, and the whole point is harmonization, so I guess there ought to be an `x` here.

eisenwave (changes requested):

AlisdairM:
Fixed, plus applied the noted editorial comment in the paper to reorder the members consistently.  In doing so, I noticed that only `bitset::reference` has a unary `operator~`, and worse, it remains unspecified.

Should we open an LWG issue to add that specification, and following the harmonization principle, add the same to `vector<bool>::reference`?

jwakely:
> I noticed that only `bitset::reference` has a unary `operator~`, and worse, it remains unspecified.
> 
> Should we open an LWG issue to add that specification, and following the harmonization principle, add the same to `vector<bool>::reference`?

We should definitely not add it to vector bool. See the LWG mattermost on October 9 where @Bekenn observed:

> I see that paper leaves the bitset's `reference::operator~` intact; that function should be removed (or at least deprecated).  
>  
> With an actual bool, the value is promoted to int prior to the application of the ~ operation.  This means that, for instance, `~true` comes out as `-2`, not `false`.

i.e. `static_assert( ~false && ~true )` passes with the core language, so I don't think `~bitref` is a useful operator.

jwakely (approved):

PR #8566, opened 2025-11-27
[specialized.algorithms] Use `\exposid` for "voidify", not `\placeholder(nc)`

### LINKED ISSUE #8559: [specialized.algorithms] `voidify` should use `\exposid`, not `\placeholder`
Throughout the document, we use `\placeholdernc{voidify}`, but this doesn't make any sense.

https://github.com/cplusplus/draft/blob/b1ff1dfe415fdeb081c5571ad92fde2b8c506704/source/algorithms.tex#L13528-L13544

`voidify` is clearly declared as an exposition-only function template, so what is the `\placeholdernc` doing here?

eisenwave:
Fixes #8559.

This should be applied after motions; we're adding some more `\placeholdernc{voidify}` these motions.

A rebase will be needed.

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -13562,7 +13562,7 @@ Some algorithms specified in \ref{specialized.algorithms}
 make use of the following exposition-only function templates:
 \begin{codeblock}
 template<class T>
-  constexpr void* @\placeholdernc{voidify}@(T& obj) noexcept {
+  constexpr void* @\exposid{voidify}@(T& obj) noexcept {
     return addressof(obj);
   }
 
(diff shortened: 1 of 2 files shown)


jwakely (approved):

jensmaurer (approved):

source/algorithms.tex
```diff
@@ -13530,7 +13530,7 @@
 make use of the following exposition-only function templates:
 \begin{codeblock}
 template<class T>
-  constexpr void* @\placeholdernc{voidify}@(T& obj) noexcept {
```
tkoeppe, line 13557: could you retain the `nc`-ness (i.e. use `\exposidnc` where appropriate)?

eisenwave: I don't think it's appropriate anywhere. Just below this change, we have `deref-move` without `nc`.

It doesn't seem like `voidify` is ever being vertically aligned with something else, and to my understanding, that's when we would disable italics corrections.

eisenwave:
Rebased on latest `main`, added the [exec.associate] use of `voidify` added these motions.

source/algorithms.tex
jensmaurer: It's also for the spacing of (here) the parenthesis just after the italics.

tkoeppe: Indeed.

PR #8569, opened 2025-11-29
[expr.add] Simplify phrasing around ptrdiff_t and avoid redundancy

### LINKED ISSUE #8564: [expr.add] Harmonize wording for pointer subtraction/`std::ptrdiff_t` with [expr.sizeof]

jensmaurer:
Fixes #8564

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -7201,18 +7201,13 @@ or a complete object results in undefined behavior.
 \end{note}
 
 \pnum
-\indextext{\idxcode{ptrdiff_t}!implementation-defined type of}%
-\indextext{subtraction!implementation-defined pointer}%
 \indextext{\idxcode{ptrdiff_t}}%
 \indextext{comparison!undefined pointer}%
-When two pointer expressions \tcode{P} and \tcode{Q} are subtracted,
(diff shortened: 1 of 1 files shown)


eisenwave (approved):
Not sure why we're not just adjusting #8565, but this solution seems fine too.

jensmaurer:
Sorry, I overlooked the other pull request before doing mine :-(

PR #8575, opened 2025-12-02
[lex.token] Strike useless footnote

AlisdairM:
This footnote on some kinds of literals adds no value -- it does not convey anything surprising other than raising the question of how the subset of literals mentioned is chosen.


jensmaurer (approved):
I'm ok with this.

PR #8577, opened 2025-12-02
[intro.memory] Merge footnote on CHAR_BIT into main text

AlisdairM:
The footnote on `CHAR_BIT` is information relevant to the main text of the corresponding paragraph, where it can usefully be indexed.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3368,11 +3368,10 @@ and the eight-bit code units of the Unicode
 UTF-8 encoding form
 and is composed of a contiguous sequence of
 bits,
-\begin{footnote}
-The number of bits in a byte is reported by the macro
-\tcode{CHAR_BIT} in the header \libheaderref{climits}.
-\end{footnote}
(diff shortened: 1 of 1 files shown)


source/basic.tex
```diff
@@ -3366,13 +3366,10 @@
...
-bits,
-\begin{footnote}
-The number of bits in a byte is reported by the macro
-\tcode{CHAR_BIT} in the header \libheaderref{climits}.
-\end{footnote}
-the number of which is \impldef{bits in a byte}.
+and is composed of a contiguous sequence of bits,
+the number of which is \impldef{bits in a byte}
+and reported by the macro \libmacro{CHAR_BIT}
+in the header \libheaderref{climits}.
```
jensmaurer: This appears to add normative redundancy.

Can you just do

> [ Note: See the macro CHAR_BIT in the header climits. -- end note ]

which has all the same advantages for indexing, but keeps the statement non-normative.

PR #8584, opened 2025-12-04
[cpp.predefined] Tidy specification of __FILE__ and __LINE__

AlisdairM:
This PR moves the normative parentheticals into the main text, promotes the corresponding footnotes to regular notes, and adds cross-references to the #line directive.

### DIFF
--- a/source/preprocessor.tex
+++ b/source/preprocessor.tex
@@ -2219,20 +2219,20 @@ shall be supplied.
 \item
 \indextext{__file__@\mname{FILE}}%
 \mname{FILE}\\
-The presumed name of the current source file (a character string
-literal).
-\begin{footnote}
-The presumed source file name can be changed by the \tcode{\#line} directive.
-\end{footnote}
+A character string literal representing the presumed name of
+the current source file.
+\begin{note}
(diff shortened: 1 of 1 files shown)


source/preprocessor.tex
```diff
@@ -2219,20 +2219,20 @@
...
 
 \item
 \indextext{__line__@\mname{LINE}}%
 \mname{LINE}\\
-The presumed line number (within the current source file) of the current source line
-(an integer literal).
-\begin{footnote}
-The presumed line number can be changed by the \tcode{\#line} directive.
-\end{footnote}
+A decimal integer literal representing the presumed line number
```
jensmaurer: The incoming text says "an integer literal", allowing hex or binary literals.  You change that to "decimal", which feels like a normative change.

AlisdairM: Fixed.
A total foul-up on my part.  I was confusing the decimal literal of the `#line` directive with the literal the `__LINE__` macro expands to.

jensmaurer (approved):
This is shuffling words around and looks good to me.

PR #8589, opened 2025-12-08
[lex] Avoid "shall" when not stating a direct requirement

lprv:


AlisdairM:
This looks good to me.

tkoeppe:
Thanks, very nice!

PR #8590, opened 2025-12-08
[lex] Move definitions, and remove redundant use, of "[non]digit"

lprv:
Move definitions of *digit* and *nondigit* under *pp-number* where they first appear; strike their use in *identifier-start* and *identifier-continue* where they are subsumed by the definitions based on Unicode character properties.


AlisdairM:
The PR looks technically correct.
I like the change, snd believe it brings clarify to *identifier-start*.

PR #8592, opened 2025-12-08
[lex.ext] Remove redundant requirement

lprv:
This requirement is redundant with the more general provision found in [basic.lookup.general].


AlisdairM:
I believe there are bigger structural problems here.  Paragraphs 2--7 are all phase 7 material about how to lookup and dispatch a call to a literal operator, that is not needed for [lex] that does not use the actual value of the literal.  I suggest that a correct fix would move all of those paragraphs to a more appropriate location.

PR #8593, opened 2025-12-09
Tweak members_of note that not only special members may be implicitly…

jakubjelinek:
…-declared.

### DIFF
--- a/source/meta.tex
+++ b/source/meta.tex
@@ -5618,7 +5618,7 @@ Reflections of class members and unnamed bit-fields that are declared
 appear in the order in which they are declared.
 \begin{note}
 Base classes are not members.
-Implicitly-declared special members
+Implicitly-declared special members and comparison operator member functions
 appear after any user-declared members\iref{special}.
 \end{note}
 


jakubjelinek:
@jwakely @tomaszkam

In e.g.
```c
struct S { auto operator <=> (const S &) const = default; int s; };
```
`bool operator == (const S &) const = default;` is implicitly-declared and std::members_of should first list the explicitly declared members, operator <=> and s and only after that the implicitly-declared ones, 6 special members and operator ==.

jakubjelinek:
Implicitly-declared needs to apply both to special members and comparison operator member functions, not sure if it doesn't have to be repeated.

tomaszkam:
Looks good to me.

source/meta.tex
```diff
@@ -5618,7 +5618,7 @@
 appear in the order in which they are declared.
 \begin{note}
 Base classes are not members.
-Implicitly-declared special members
+Implicitly-declared special members and comparison operator member functions
```
jensmaurer: Are there any "special members" that are not "member functions"?  Why the difference in presentation vs. "comparison operator member functions"?

Maybe just "Implicitly-declared member functions appear ..." ?

(I don't think we normatively say whether implicit `operator<=>` is declared before or after the special members, so there should be no hint in the note about their relative order.)

jwakely:
Please adjust the commit message so the first line starts with the stable tag, e.g.

[meta.reflection.member.queries] tweak members_of note about implicitly-declared members

PR #8594, opened 2025-12-09
[lex.phase] Move normative handling of comments]

AlisdairM:
The specification of phase 3 of translation says too much about how things are done rather than what is done.  Per suggestions on PR #8414, move the precise handling of comments to [lex.comment]. As this move relies on the specification for whitespace, move the definition of whitespace from [lex.pptoken] to [lex.comment] too.

Note that this PR would obviate #8416, which turns the text moved from [lex.pptoken] into a comment, precisely as the moved text in this PR.

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -116,8 +116,8 @@ shall be processed as if an additional new-line character were appended
 to the file.
 
 \item The source file is decomposed into preprocessing
-tokens\iref{lex.pptoken} and sequences of whitespace characters
-(including comments). A source file shall not end in a partial
+tokens\iref{lex.pptoken} and whitespace\iref{lex.comment}.
+A source file shall not end in a partial
 preprocessing token or in a partial comment.
 \begin{footnote}
 A partial preprocessing
@@ -129,10 +129,6 @@ or \tcode{>}. A partial comment
 would arise from a source file ending with an unclosed \tcode{/*}
 comment.
 \end{footnote}
-Each comment\iref{lex.comment} is replaced by one \unicode{0020}{space} character. New-line characters are
-retained. Whether each nonempty sequence of whitespace characters other
-than new-line is retained or replaced by one \unicode{0020}{space} character is
-unspecified.
 As characters from the source file are consumed
 to form the next preprocessing token
 (i.e., not being consumed as part of a comment or other forms of whitespace),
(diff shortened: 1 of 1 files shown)


hubert-reinterpretcast (commented):
With the proposed changes, the subclause title for [lex.comment] should probably be expanded.

source/lex.tex
```diff
@@ -518,13 +514,38 @@
 \indextext{comment!\tcode{//}}%
 The characters \tcode{//} start a comment, which terminates immediately before the
 next new-line character.
+Each comment is replaced by one \unicode{0020}{space} character;
+new-line characters are retained.
```
hubert-reinterpretcast, line 518: There should be no retention of new-line characters within C-style comments. Additionally, retention of new-line characters following the termination of a `//` comment does not require mention. The termination point of a comment of either form is made clear above. My understanding of the corresponding sentence in the original text is that it is meant to be bound to the next sentence in the original text (not the previous sentence as this rendition assumes).
```suggestion
Each comment is replaced by one \unicode{0020}{space} character.
```

source/lex.tex
```diff
@@ -518,13 +514,38 @@
...
+new-line,
+\unicode{000b}{line tabulation}, and
+\unicode{000c}{form feed}), or both.
+\begin{note}
+In certain circumstances during translation phase 4, as described in \ref{cpp},
+whitespace (or the absence thereof) serves as more than
+preprocessing token separation.
+Whitespace can appear within a preprocessing token only as part of
+a \grammarterm{header-name} or
+between the quotation characters in a character literal or string literal.
```
hubert-reinterpretcast, line 542: My hope is that we would strike this from the note in the future. I think we want to move to a model, like with "UCNs inside raw strings", where "whitespace" (as we define it) does not appear except between preprocessing tokens.

source/lex.tex
```diff
@@ -518,13 +514,38 @@
...
+\end{note}
+
+\pnum
+Whether each nonempty sequence of whitespace characters other than new-line
+is retained or replaced by one \unicode{0020}{space} character is unspecified.
```
hubert-reinterpretcast, line 547: For me, this is too far out of context to be worded this way. I would like a statement that such replacement may be done by an implementation before stating that whether it is done (for any specific instance) is unspecified.

Additionally, with this presentation, it is unclear to me when this replacement occurs (which affects whether comments act as separate U+0020 characters and whether both ends of a phase 1 line splice can remain separate U+0020 characters that replaced other whitespace characters). The behaviour can be observed if whitespace is retained by an implementation when forming a _header-name_ token from `<` _h-pp-tokens_ `>`.

AlisdairM:
Thanks Hubert.  Sounds like this PR belongs in my paper where it can receive the proper Core review, as I suspected.  I will keep this open in draft until I have updated the paper accordingly, for easy access to your recommendations.

PR #8599, opened 2025-12-10
[basic.def.odr] Fix comment in example

lprv:


source/basic.tex
```diff
@@ -532,7 +532,7 @@
 int h(bool cond) {
   constexpr A a = {1};
   constexpr const volatile A& r = a;    // odr-uses \tcode{a}
-  int _ = f(cond ? a.x : r.x);          // does not odr-use \tcode{a} or \tcode{r}
+  int _ = f(cond ? a.x : r.x);          // odr-uses \tcode{a} but not \tcode{r}
```
jensmaurer, line 535: Why does this odr-use `a`  ?  `a` is sufficiently constant, I think.

tkoeppe:
@opensdh could you have a look?

opensdh:
I'm sure I could check this, but it's not wording I already know.  Maybe ask someone else first?

PR #8601, opened 2025-12-10
[basic.def.odr] Add missing "shall"

lprv:


eisenwave:
I don't think the omission of "shall" is objectively wrong, although it would seem more consistent with the other bullets to use it here.

PR #8606, opened 2025-12-11
[basic.lookup.unqual] Rephrase to avoid incorrect use of "lookup context"

lprv:
"Lookup context" defined for names; a *qualified-id* is never a name.


eisenwave:
P1, seems like a mildly normative wording defect then.

https://eel.is/c++draft/basic.lookup.qual.general#def:lookup_context

PR #8610, opened 2025-12-11
[class.qual] Apply consistent formatting to placeholder

lprv:


tkoeppe:
I'm not sure if we have a good general principle, but at least locally it seems reasonable to use mathematical variables as placeholders here. Thanks!

PR #8611, opened 2025-12-11
[namespace.qual] Remove references to undefined meta-variable

lprv:
"`S`" (previously denoting the "namespace-qualified lookup set") vanished with P1787 but the examples continue to refer to it.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -2626,22 +2626,22 @@ namespace AB {
 
 void h()
 {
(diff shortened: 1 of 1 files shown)


tkoeppe:
@opensdh This overlaps with your doing, I think, mind having a look?

opensdh:
The bug report is correct, although I suspect we could simplify the example's explanations more strongly: for instance, the first comment could just be
> search in `AB` finds `AB::g()`

and the second could be
> search in `AB` finds nothing; searches in `A` (but not `Y`) and `B` find `A::f(int)` and `B::f(char)`, the first of which is chosen by overload resolution

tkoeppe:
That sounds nice. @lprv, could you make the suggested changes?

PR #8616, opened 2025-12-14
[utility.intcmp] Rephrase integer type constraint of "standard or extended" as "signed or unsigned"

eisenwave:
"standard or extended" refers to the same set of types as "signed or unsigned". However, the latter constraint is far more common, and it would conveniently include additional sets of integers added in the future, such as bit-precise integers.

See also
- https://github.com/cplusplus/draft/pull/8546#pullrequestreview-3545939346
- https://github.com/eisenwave/cpp-proposals/issues/125#issue-3700026438

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -546,8 +546,8 @@ template<class T, class U>
 \begin{itemdescr}
 \pnum
 \mandates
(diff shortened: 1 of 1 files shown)


jwakely (approved):
Thanks

PR #8624, opened 2025-12-15
[basic.link] Reword to avoid confusion with a defined term

lprv:
"Corresponding" looks like a use of the term "[correspond](https://eel.is/c++draft/basic.scope.scope#def:correspond)", which is not what is meant here.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -2984,8 +2984,7 @@ otherwise,
 \indextext{friend function!linkage of}%
 if the entity is a function or function template
 declared in a friend declaration and
-a corresponding non-friend declaration is reachable,
-%FIXME: Which declaration is "that prior declaration"?
+a prior non-friend declaration of the entity is reachable,
 %FIXME: "prior" with respect to what? And what about dependent lookup?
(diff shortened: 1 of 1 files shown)


source/basic.tex
```diff
@@ -2984,8 +2984,7 @@
 \indextext{friend function!linkage of}%
 if the entity is a function or function template
 declared in a friend declaration and
-a corresponding non-friend declaration is reachable,
-%FIXME: Which declaration is "that prior declaration"?
+a prior non-friend declaration of the entity is reachable,
```
jensmaurer, line 2987: I think we exactly mean "correspond" from [basic.scope.scope] p4 here.  Why wouldn't we?

jensmaurer (changes requested):

source/basic.tex
opensdh: Certainly "prior" and "reachable" are at best redundant and at worst conflicting here.  I suppose it wouldn't be harmful to switch to "declaration of the entity", though, since that would still let "corresponding" do its usual work via /8 (and would look a bit like /3.2.4).

PR #8626, opened 2025-12-15
[basic.link] Strike redundant sentence

lprv:


source/basic.tex
```diff
@@ -3037,9 +3037,7 @@
 
 \pnum
 \indextext{linkage!no}%
-Names not covered by these rules have no linkage. Moreover, except as
-noted, a name declared at block scope\iref{basic.scope.block} has no
-linkage.
+Names not covered by these rules have no linkage.
```
jensmaurer, line 3040: I think I would prefer if we kept the "block-scope names (except ...) have no linkage" as a note.  Apparently, our ancestors thought it would be important to highlight that fact.

PR #8629, opened 2025-12-15
[intro.object] Clarify definition of "potentially non-unique object"

lprv:
Avoid ambiguity around whether "subobject" is transitive or not.


source/basic.tex
```diff
@@ -3625,7 +3625,7 @@
   the object introduced by a call to \tcode{std::meta::reflect_constant_array}
   or \tcode{std::meta::reflect_constant_string}\iref{meta.define.static}, or
 \item
-  a subobject thereof.
+  a subobject of a potentially non-unique object.
```
jensmaurer, line 3628: What do you mean with "transitive"?  A subobject can be a non-immediate subobject (e.g. a data member of an array element); is this unclear?  If so, we should fix the definition of "subobject" and not try to fix the fall-out here.

PR #8632, opened 2025-12-15
[basic.stc.general] Add punctuation

lprv:


eisenwave:
I do agree that something here is unconventional about these bullets. We typically use bullets in such a way that everything can still be read as a complete sentence.

Adding the commas doesn't get it into a proper sentence structure though. Either the whole thing would have to be written as a sentence with `, or` after the last bullet (like https://eel.is/c++draft/basic.fundamental#15), or each bullet should be a complete sentence ending in a period.

tkoeppe:
@Tsche our new list expert, please have a look.

Tsche:
Right, so this is not a complete sentence split into multiple list items - therefore commas seem inappropriate here, but the leading colon is necessary. Rewording this as a complete sentence (with `, or` after every item and no `one of the following:`) doesn't improve readability terribly much imo

Status quo seems fine to me

tkoeppe:
Agreed, the status quo seems the best among the various local deformations.

PR #8633, opened 2025-12-15
[basic.stc.dynamic.general] Use "the" before "global scope"

lprv:


eisenwave:
It looks like we say "in the global scope" almost everywhere else, not counting example code blocks, which tend to say "at global scope".

P2 because it seems like an unintentional inconsistency for something where we have a well-established wording idiom.

jensmaurer:
While this might improve some sort of consistency, I would prefer "at global scope" for this case.  Consider:

```
namespace N { int f(); }
```

The function `f` is obviously declared "in" the global scope, because the namespace `N` is inside the global scope and `f` is inside namespace `N`.  We usually consider "in" to be transitive and make an effort to change the phrasing if that's not what we want.

PR #8638, opened 2025-12-15
[basic.fundamental] Replace "denotes" with "is"

lprv:


jensmaurer (approved):
Yeah, "denote" is something from name lookup meanwhile.

PR #8639, opened 2025-12-15
[basic.fundamental] Rephrase note

lprv:
The current note can be misread as attaching only the `long` modifier to `bool` (and talking about just the "`signed`", "`unsigned`", and "`short`" types otherwise, all of which *are* valid *type-id*s).

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -5544,8 +5544,8 @@ an \impldef{underlying type of \tcode{bool}} unsigned integer type.
 The values of type \keyword{bool} are
 \keyword{true} and \keyword{false}.
 \begin{note}
-There are no \keyword{signed}, \keyword{unsigned},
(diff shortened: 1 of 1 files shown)


source/basic.tex
```diff
@@ -5544,8 +5544,8 @@
 The values of type \keyword{bool} are
 \keyword{true} and \keyword{false}.
 \begin{note}
-There are no \keyword{signed}, \keyword{unsigned},
-\keyword{short}, or \tcode{\keyword{long} \keyword{bool}} types or values.
+The type specifiers \keyword{signed}, \keyword{unsigned},
```
jensmaurer, line 5547: We want \grammarterm{type-specifier} here, right?

... which is introduced much later.  This is about the fundamental notion of a type, not about any particular expression (using a type-id, for instance).  As such, I find the mention of "type specifier" not great.

jensmaurer (changes requested):

PR #8641, opened 2025-12-15
[basic.stc.dynamic] Spell out compound type names

lprv:


eisenwave:
@jensmaurer I don't think we formally wrote this down anywhere, but we generally aspire to use the quoted spelling of types in core wording, don't we?

jensmaurer:
Well, first of all, something like `T*` is not a type, but a _type-id_ that needs (semantic) interpretation.  After interpretation by [dcl.decl], this becomes a quoted type string, which we take as our standard type description.  The library description tends to gloss over that difference and happily uses _type-ids_ all over the place.

Traditionally, we don't use quotes around "simple" types such as `int` or even `unsigned long int`, but those essentially don't appear in the standard at all (except where defined).

In short, the suggested changes here look good.

PR #8642, opened 2025-12-15
[basic.fundamental] Demote normatively redundant sentences to a note

lprv:


eisenwave:
I would be cautious about making changes in this area because they may conflict with CWG2966
- https://github.com/cplusplus/CWG/issues/644

PR #8647, opened 2025-12-15
[intro.races] Replace "library call" with "evaluation"; mark example

lprv:
"Synchronizes with" is sometimes applied to evaluations other than invocations of library functions.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -6747,9 +6747,12 @@ where the first operation is $A$, and
 every subsequent operation is an atomic read-modify-write operation.
 
 \pnum
-Certain library calls \defn{synchronize with} other library calls performed by
-another thread. For example, an atomic store-release synchronizes with a
+Certain evaluations \defn{synchronize with} other evaluations performed by
+another thread.
+\begin{example}
+An atomic store-release synchronizes with a
 load-acquire that takes its value from the store\iref{atomics.order}.
+\end{example}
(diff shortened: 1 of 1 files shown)


eisenwave:
How do you initiate a load-acquire or any other kind of operation without a library call? It seems to me like any atomic operation involves some library call from `<atomic>` or adjacent header.

I'm also not sure whether "operation" or "evaluation" is the better term. "Atomic operation" is an established wording idiom, for example. I can also find a number of uses like

> This operation synchronizes with ...

eisenwave:
Hmm, I guess that [intro.races] uses "evaluation" though, even in "operation" is more common in library wording.

Anyway, I'm gonna say P2 because "synchronizes with" is defined for evaluations. I think an example of an evaluation that synchronizes with something but isn't a call is thread completion. A thread (possibly detached) can complete without any library call causing that.

jensmaurer:
@eisenwave , I have a hard time seeing "thread completion" as an evaluation, without something a little more explicit saying so.

eisenwave:
It's not necessarily a library call, and I'm not sure what to call it. We define "synchronizes with" for evaluations, so when we say in [futures.async] that a thread completion synchronizes with something, I guess that means a thread completion is an evaluation?

If it's neither a library call nor an evaluation, then we have a bit of a normative wording hole and this may be LWG territory.

PR #8650, opened 2025-12-15
[intro.races], [intro.progress] Use consistent terminology

lprv:


eisenwave:
Are we sure this is editorial?

jensmaurer:
I can see that this might be correct and desirable, but this needs CWG review.

PR #8652, opened 2025-12-15
[basic.start.static] Improve readability of example

lprv:


jensmaurer (approved):

tkoeppe:
This is nice, thank you! This is a much looked-at example, and it's much better to have the explanation on normal text than in a code comment.

PR #8653, opened 2025-12-15
[basic.start.dynamic] Add cross-reference

lprv:


eisenwave:
P2 because it's a forward reference in core wording, and we ought to add cross-references for those.

PR #8655, opened 2025-12-15
[basic.start.dynamic] Fix example

lprv:
The last sentence of this example talks about `a` being initialized "after the first statement of `main`", which doesn't make sense since that statement odr-uses it (assuming `Use` is not a static member).

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -7481,8 +7481,8 @@ extern A a;
 extern B b;
 
 int main() {
-  a.Use();
   b.Use();
+  a.Use();
 }
 \end{codeblock}
 


source/basic.tex
```diff
@@ -7481,8 +7481,8 @@
 extern B b;
 
 int main() {
-  a.Use();
   b.Use();
+  a.Use();
```
jensmaurer, line 7485: Instead of changing the order here (which feels scary), I'd prefer changing the "after the first statement of main" utterance below.  What do we say normatively here?  Can we just add an empty statement as the first statement of main to defuse the situation?

PR #8656, opened 2025-12-15
[basic] Use past subjunctive after "as if"

lprv:


jensmaurer (approved):

tkoeppe:
@W-E-Brown for commentary perhaps?

tkoeppe:
I don't think the status quo is wrong; it's just a stylistic choice. We _could_ change it, but... we could change a lot of things.

W-E-Brown:
> On Dec 16, 2025, at 5:22 AM, Thomas Köppe ***@***.***> wrote:
> 
> 
> tkoeppe
>  left a comment 
> (cplusplus/draft#8656)
>  <https://github.com/cplusplus/draft/pull/8656#issuecomment-3660043238>
> @W-E-Brown <https://github.com/W-E-Brown> for commentary perhaps?
> 

Well, since you asked ...  :)

Grammatically, the change to the subjunctive following the conditional "if" seems correct to me.  However, I found [class.compare.default]/Note 2 where a similar change could (and likely should) be made.  It wouldn't surprise me to find more such places.

More generally, the intent of (the multiple occurrences) of the phrase "it is as if" is not clear to me, since I don't know (and can't seem to find) the antecedent of "it" in these contexts.  I'm sure this is not what we wanted to hear at this late stage, but (unless I'm overlooking something obvious to others) this strikes me as a substantial defect needing clarification in the current wording.

tkoeppe:
@opensdh Maybe we can figure out what "it" is?

opensdh:
"it" here is largely the same as that in "it's late": an almost completely dummy subject referring to the general situation.  I suppose one could write "The interpretation of the program is that which would arise from _situation_.", but I wouldn't recommend, er, it.

tkoeppe:
Yes... but this still seems like unusually vague and broad wording (as would "the interpretation of the program" be). Could we come up with something more localised to the topic in question?

(E.g. when you don't return from main, we don't say "It is as if the main function returned 0", but instead we have a specific phrase.)

PR #8658, opened 2025-12-15
Do not hyphenate "potentially evaluated"

lprv:
"[Potentially evaluated](https://eel.is/c++draft/basic.def.odr#def:potentially_evaluated)" is defined without a hyphen.

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -4288,7 +4288,7 @@ template <class T> void spam(decltype([]{}) (*s)[sizeof(T)]);
 \end{codeblock}
 \end{example}
 \indextext{expression!functionally equivalent|see{functionally equivalent, expressions}}%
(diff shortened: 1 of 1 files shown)


hubert-reinterpretcast:
> "[Potentially evaluated](https://eel.is/c++draft/basic.def.odr#def:potentially_evaluated)" is defined without a hyphen.

The change is correct, but the rationale should not be that it is defined without a hyphen. Rather, the rationale is that hyphens are not used for compounds modifying a noun if an adverb ending in "ly" is used. https://www.merriam-webster.com/grammar/hyphen-rules-open-closed-compound-words

tkoeppe:
Thanks, Hubert! The rationale is only in the pull request, though, and not in the commit messages, so if the changes are fine then this should be ready to go.

Issue #5904, opened 2022-10-16
[vector.bool] and [template.bitset] Divide more clearly


blackteahamburger:
Similar to LWG3638, and I suggest removing current [vector.bool.pspc]/4 and detailing each member function of std::vector\<bool\>::reference.

[template.bitset] is similar.

eisenwave:
https://github.com/cplusplus/papers/issues/2419 recently overhauled the spec of proxy references.

PR #8661, opened 2025-12-16
[template.bitset.general][vector.bool.pspc] Make subclauses for `reference` nested classes

AlisdairM:
Consistent with other nested classes within the standard library.  This formulation leads to slightly clearer specification for the member functions that now use appropriately qualified names.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -10489,21 +10489,8 @@ namespace std {
     using reverse_iterator       = std::reverse_iterator<iterator>;
     using const_reverse_iterator = std::reverse_iterator<const_iterator>;
 
-    // bit reference
-    class @\libmember{reference}{vector<bool>}@ {
-    public:
-      constexpr reference(const reference& x) noexcept;
-      constexpr ~reference();
-      constexpr reference& operator=(bool x) noexcept;
-      constexpr reference& operator=(const reference& x) noexcept;
-      constexpr const reference& operator=(bool x) const noexcept;
-      constexpr operator bool() const noexcept;
-      constexpr void flip() noexcept;   // flips the bit
-
-      friend constexpr void swap(reference x, reference y) noexcept;
-      friend constexpr void swap(reference x, bool& y) noexcept;
-      friend constexpr void swap(bool& x, reference y) noexcept;
-    };
+    // \ref{vector.bool.reference}, bit reference
+    class reference;
 
     // construct/copy/destroy
     constexpr vector() noexcept(noexcept(Allocator())) : vector(Allocator()) { }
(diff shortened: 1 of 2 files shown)


source/containers.tex
```diff
@@ -10607,13 +10594,34 @@
...
 
+\rSec3[vector.bool.reference]{Class \tcode{vector<bool>::reference}}%
 \pnum
 \tcode{reference}
 is a class that simulates a reference to a single bit in the sequence.
 
+\indexlibrarymember{reference}{vector<bool>}%
+\begin{codeblock}
+namespace std {
+  template<size_t N> class vector<bool>::reference {
```
jensmaurer: please put the _template-head_ on a separate line

source/utilities.tex
```diff
@@ -10644,12 +10632,54 @@
...
+error is associated with exceptions of type
+\tcode{overflow_error}\iref{overflow.error}.
+\indexlibraryglobal{overflow_error}%
+\end{itemize}
+
+\rSec3[bitset.reference]{Class \tcode{bitset<N>::reference}}%
+\indexlibrarymember{reference}{bitset}%
+\begin{codeblock}
+namespace std {
+  template<size_t N> class bitset<N>::reference {
```
jensmaurer: Please put the _template-head_ on a separate line.

jensmaurer (changes requested):
In general, I like this.

jensmaurer:
Can you make this two commits please (keep them in this single pull request)?

AlisdairM:
Split into two commits, making the requested change to separate out the template-head in each case.

source/containers.tex
```diff
@@ -10607,13 +10594,35 @@
...
 
+\rSec3[vector.bool.reference]{Class \tcode{vector<bool>::reference}}%
 \pnum
 \tcode{reference}
 is a class that simulates a reference to a single bit in the sequence.
 
+\indexlibrarymember{reference}{vector<bool>}%
+\begin{codeblock}
+namespace std {
+  template<size_t N>
```
jensmaurer: What's this _template-head_ doing here in the first place?

jensmaurer (changes requested):

source/containers.tex
AlisdairM: Oops --- good catch!
The template parameter should be the allocator type parameter.  Apparently I have a copy/paste issue from the bitset reference declaration.  Double-checked that there are no other copy/paste errors here.

jensmaurer (approved):

jensmaurer:
The commit descriptions (in particular the titles) aren't stellar.

"Break out into its own subclause" is missing "reference" and/or "nested class" somewhere.  In particular since the new label is [vector.bool.reference], not the quoted label.

AlisdairM:
It looks like my git command line mangled my attempts to put the class names into the commit title line.  I will fix just those commit messages and resubmit.

frederick-vs-ja:
This seemingly addresses #5904. CC @blackteahamburger.

blackteahamburger:
> This seemingly addresses #5904. CC @blackteahamburger.

Yes, thanks.

source/containers.tex
```diff
@@ -10607,13 +10594,35 @@
 of \tcode{bool} values. A space-optimized representation of bits is
 recommended instead.
 
+\rSec3[vector.bool.reference]{Class \tcode{vector<bool>::reference}}%
 \pnum
 \tcode{reference}
 is a class that simulates a reference to a single bit in the sequence.
```
tkoeppe, line 10600: Can we also reword this the same as you did below ("Class reference simulates...")? It's awkward to start a sentence with a symbol.

tkoeppe: @AlisdairM ^^^

tkoeppe: @AlisdairM Ping?

PR #8664, opened 2025-12-17
[diff.basic] Fix errors in entry about C "compatible" types

hubert-reinterpretcast:
* Use enum example instead of example outdated since C99

  C99 subclause 6.2.7, "Compatible type and composite type", had this sentence:
  > If one is declared with a tag, the other shall be declared with the same tag.

* Stop claiming, as harmless, type-based aliasing violations


jensmaurer:
I don't think we need a core issue for this; this seems simple + obvious enough and affects a non-normative Annex to start with.

PR #8667, opened 2025-12-17
[intro.refs], [numeric.limits] Remove all references to ISO/IEC 10967-1:2012

### LINKED ISSUE #8666: [intro.refs], [support] Remove all references to ISO/IEC 10967-1:2012
We have a bibliography reference to ISO/IEC 10967-1:2012 and mention it in a number of footnotes. That standard makes factually incorrect statements and is irrelevant to C and C++ following the C23 rebase. We should purge any mention of it.

## Historical background

Historically, C17 had an [Annex H](https://cigix.me/c17#H) which documented compliance with ISO/IEC 10967–1. I suppose this is why we refer to it as well. C23 no longer mentions ISO/IEC 10967–1 in any of its wording, but does contain an unused reference in its bibliography. This is most likely an editorial mistake.

C23 instead has an Annex F, G, and H that document compliance with ISO/IEC 60559. We also refer to ISO/IEC 60559, and we are based on C23 for large portions of the standard library.

## The problem with referencing ISO/IEC 10967-1

ISO/IEC 10967-1:2012 is now essentially irrelevant to us. In fact, the references we have are outright wrong and misleading:
https://github.com/cplusplus/draft/blob/ef2abfaee4a22b1765d4ef95f1feee3baa962a41/source/support.tex#L1659-L1663

C23 (https://cigix.me/c23#5.2.5.3.3.p23) says that an `FLT_ROUNDS` value of `1` indicates "to nearest, ties to even" and a value of `4` indicates "to nearest, ties away from zero"

On the contrary, ISO/IEC 10967-1 claims:

Image

This is wrong. C no longer has any ambition of complying with ISO/IEC 10967-1, and it outright contradicts it, as can be seen.

> [!NOTE]
> The "†" symbol means:
> > The operations or parameters marked “†” are not part of the language standard and must be provided by an implementation that wishes to conform to LIA-1.
>
> This is also wrong in the case of C23. The language standard does provide the value `4`.

Thankfully, no normative wording is affected by this, only a bunch of footnotes in [support].

eisenwave:
Fixes #8666.

Note that Jonathan wants SG6 to give an ack; see https://github.com/cplusplus/draft/issues/8666#issuecomment-3665372486

### DIFF
--- a/source/back.tex
+++ b/source/back.tex
@@ -7,10 +7,6 @@
 \bibitem{iso4217}
   ISO 4217:2015,
   \doccite{Codes for the representation of currencies}
-\bibitem{iso10967-1}
-  ISO/IEC 10967-1:2012,
-  \doccite{Information technology --- Language independent arithmetic ---
-    Part 1: Integer and floating point arithmetic}
 \bibitem{iso14882:2023}
   ISO/IEC 14882:2023,
   \doccite{Programming Languages --- \Cpp{}}
--- a/source/support.tex
+++ b/source/support.tex
@@ -1109,11 +1109,6 @@ Non-arithmetic standard types, such as
 \pnum
 \indextext{signal-safe!\idxcode{numeric_limits} members}%
 Each member function defined in this subclause is signal-safe\iref{support.signal}.
-\begin{note}
-\indextext{LIA-1}%
-The arithmetic specification described in ISO/IEC 10967-1:2012 is
-commonly termed LIA-1.
-\end{note}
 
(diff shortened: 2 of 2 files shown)


jwakely (approved):
We already have normative wording that some members (corresponding to infinities, NaNs, subnormals etc.) are required when `is_iec559` is true, so I don't think we lose anything by removing the footnotes about LIA-1.

tkoeppe:
Thanks everyone for checking!

PR #8669, opened 2025-12-18
[expr.const] "with [...]" => "within the evaluation"

hubert-reinterpretcast:
We use "within the evaluation of E" in various places. We also mean to use it here.


eisenwave:
This may even be a typo considering we have 11 uses of "within the evaluation" but only one use of "with the evaluation". In any case, seems like we have a clearly established wording idiom that's not being used here.

jensmaurer (approved):
Treating as a typo.

PR #8670, opened 2025-12-18
[expr.const] Definition domain fixes for "usable in constant expressions"

hubert-reinterpretcast:
- Fix index entries to identify the domains of the definitions
- Fix wording error that restricted the domain of the definition for variables

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -9223,7 +9223,9 @@ a reference member of any of the above.
 \end{itemize}
 
 \pnum
-An object or reference is \defnx{usable in constant expressions}{usable in constant expressions!object or reference} at point $P$
+An object or reference is
+\defnx{usable in constant expressions}{usable in constant expressions!object or reference}
+at point $P$
 if it is an object or reference
 that is potentially usable in constant expressions at $P$ and
 is constexpr-representable at $P$.
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -9192,7 +9192,7 @@ it has reference or non-volatile const-qualified integral or enumeration type.
 
 \pnum
 A constant-initialized potentially-constant variable $V$ is
-\defn{usable in constant expressions} at a point $P$ if
+\defnx{usable in constant expressions}{usable in constant expressions!variable} at a point $P$ if
 $V$'s initializing declaration $D$ is reachable from $P$ and
 \begin{itemize}
 \item $V$ is constexpr,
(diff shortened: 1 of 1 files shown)


source/expressions.tex
```diff
@@ -8458,7 +8459,7 @@
 \item
 a reference member of any of the above.
 \end{itemize}
-An object or reference is \defn{usable in constant expressions} at point $P$
+An object or reference is \defnx{usable in constant expressions}{usable in constant expressions!object or reference} at point $P$
```
eisenwave:
```suggestion
An object or reference is
\defnx{usable in constant expressions}{usable in constant expressions!object or reference}
at point $P$
```
This line would be well past 120 characters.

source/expressions.tex
```diff
@@ -8431,9 +8431,10 @@
 it has reference or non-volatile const-qualified integral or enumeration type.
 
 \pnum
-A constant-initialized potentially-constant variable $V$ is
-\defn{usable in constant expressions} at a point $P$ if
-$V$'s initializing declaration $D$ is reachable from $P$ and
+A variable $V$ is
+\defnx{usable in constant expressions}{usable in constant expressions!variable} at a point $P$ if
+$V$ is constant-initialized and potentially-constant,
```
eisenwave: It's not obvious that the new restriction that $V$ is constant-initialized is editorial. What does this do? Also, doesn't this conclude from the first bullet that $V$ is constexpr or something else?

eisenwave:
P2 because we have two identical `\defn` entries, and that's probably not how those macros should be used.

source/expressions.tex
hubert-reinterpretcast: > It's not obvious that the new restriction that V is constant-initialized is editorial.

That $V$ is constant-initialized is, in the strict reading of the old wording, a prerequisite for the definition to apply.

> Also, doesn't this conclude from the first bullet that V is constexpr or something else?

It does not, because in
```cpp
extern int &r0;
int &r = r0;
constexpr int &r2 = r;
```
`r` is potentially-constant and has an initializing declaration reachable from, and in the same translation unit as, the definition of `r2`.

jensmaurer (approved):
This is a good change.

@hubert-reinterpretcast , could you please rebase + force-push?

hubert-reinterpretcast:
> could you please rebase + force-push?

@jensmaurer: done

jensmaurer:
@tkoeppe , you had concerns here, so I'll leave the merge to you. We need a squash.

tkoeppe:
@jensmaurer I don't quite recall -- are you referring to the indexing perhaps, or the normative wording?

jensmaurer:
Ah, sorry, I confused the author of the (still open) comment here. @tkoeppe , just go ahead if you feel this is ok.

PR #8671, opened 2025-12-19
[stmt.pre] Streamline decl-specifier restrictions on conditions, etc.

hubert-reinterpretcast:
Avoid an abrupt shift from a mathematical introduction to the specification of additional normative requirements.

Expresses universal quantification more directly.


tkoeppe:
Definitely, thanks, Hubert!

Issue #8514, opened 2025-11-13
What is the difference between a program and a C++ program


AlisdairM:
Typically we refer to a program or the properties of a program, but in some places we refer to C++ programs and their properties.  When do we choose to qualify "program" as "C++ program"?  Is this documented? Do we know whether that rule has been applied consistently?

AFAICT, the use of "C++ Program" occurs mostly when dealing with abstract machine notion of threads and concurrency; "C++ program" is used fairly consistently through the library clauses; and "C++ program" is used widely in compatibility Annex C.

If we are to use the compound-noun more consistently, should we give it a macro the would help writing tools and text searches to distinguish correct uses of "program" distinct from the compound "C++ program"?

jwakely:
I don't think there's any intended difference in meaning.

eisenwave:
We have a definition of "program" in https://eel.is/c++draft/basic.link#def:program whereas "C++ program" seems to be an unnecessary decoration of that term.

I don't think it's objectively wrong to say "C++ program" according to our style, even though there is potential for harmonization, so P3 I suppose.

jensmaurer:
In Annex C, we occasionally contrast C vs. C++, so saying "C++ program" might be helpful there.

AlisdairM:
I guess I am mostly looking for whether we would want to harmonize towards some editorial direction.  As a starting point, I would suggest the "the program" is implicitly always a C++ program as we are talking about "the program" being translated according to this standard.  Hence "the program is ill-formed, no diagnostic required" would never become "the C++ program is ill-formed...".

Conversely, if we talk about **a** program, we might want to have some idea when to qualify that **a** program is a **C++** program?

jensmaurer:
We should carefully review where we talk about "C++ program" and whether that's really necessary.  Only C++ programs are in scope for the C++ standard, nothing else.

PR #8678, opened 2025-12-22
Remove redundant "C++" qualifier from "C++ program"

AlisdairM:
Partially addresses #8514.

All programs in the C++ Standard are C++ programs, so there should be no distinction between the terms "program" and "C++ program".  This PR addresses redundant use of "C++" in the Core language clauses, while skipping any review of clauses 1--4 where, stylistically, the C++ epithet may remain appropriate.  A more nuanced review of those clauses may follow.

Similarly, we would largely ignore use in [lib-intro] and [support] when reviewing usage in the library clauses.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3376,7 +3376,7 @@ the number of which is \impldef{bits in a byte}.
 \begin{note}
 See the macro \libmacro{CHAR_BIT} in the header \libheaderref{climits}.
 \end{note}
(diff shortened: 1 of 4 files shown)


AlisdairM:
For the initial PR I took a maximalist approach removing all "C++" qualifiers, but I am not 100% sure that this was appropriate for all the cases in [basic], especially in text around the specification of the C++ memory model --- hopefully we can find consensus in either direction in review.

I have no such concerns with the removal from the other three files.

PR #8680, opened 2025-12-22
[lex.phases] Make note on the notion of file more prominent

AlisdairM:
The notion that when the standard refers to files does not necessarily imply a file in a traditional filing system is more fundamental than its late appearance in the middle of phase 7 of translation.  Move that note right to the top of [lex] where we first talk of storing the program text in source files, where is will be understood even before the phases of translation.

### DIFF
--- a/source/lex.tex
+++ b/source/lex.tex
@@ -42,6 +42,13 @@ A \Cpp{} program need not all be translated at the same time.
 Translation units can be separately translated and then later linked
 to produce an executable program\iref{basic.link}.
 \end{note}
+\begin{note}
+Source files, translation units, and translated translation units
+need not necessarily be stored as files, nor need there be any one-to-one
+correspondence between these entities and any external representation.
+The description is conceptual only,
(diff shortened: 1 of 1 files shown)


source/lex.tex
```diff
@@ -42,6 +42,13 @@
 Translation units can be separately translated and then later linked
 to produce an executable program\iref{basic.link}.
 \end{note}
+\begin{note}
+Source files, translation
+units and translated translation units need not necessarily be stored as
+files, nor need there be any one-to-one correspondence between these
+entities and any external representation. The description is conceptual
+only, and does not specify any particular implementation.
+\end{note}
```
jensmaurer, line 51: Thanks for moving this verbatim.  However, since this is fairly small (and thus detailed review is tractable), please apply the obvious cleanup:
- semantic line breaks
- Oxford comma before "and"

AlisdairM: Done.

jensmaurer (approved):

PR #8694, opened 2026-01-02
Make checks scripts work on macOS + update readme 

hanickadot:
Scripts needs GNU sed, which is not available on macOS by default, but can be easily install with homebrew. If script detects presence of GNU sed, it will create a function to pass attributes from sed command to GNU sed.

I also updated `readme.rst` accordingly so it mentions homebrew so we don't need to point our users to download packages from internet, but they can just use the package manager. 

Also I updated `readme.rst` and made all links https.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -13,18 +13,25 @@ Get involved:
 - `How to tell if an issue is editorial <https://github.com/cplusplus/draft/wiki/How-to-tell-if-an-issue-is-editorial>`_
 - `How to submit a new issue/defect report <https://isocpp.org/std/submit-issue>`_ for non-editorial issues
 
-More information about the C++ standard can be found at `isocpp.org <http://isocpp.org/std>`_.
+More information about the C++ standard can be found at `isocpp.org <https://isocpp.org/std>`_.
 
----------------------------
-Getting Started on Mac OS X
----------------------------
+------------------------
+Getting Started on macOS
+------------------------
 
-Install the `MacTeX distribution <http://tug.org/mactex/>`_.
+Install the `MacTeX distribution <https://tug.org/mactex/>`_.
 
-If you are on a slow network, you'll want to get the `BasicTeX package <http://tug.org/mactex/morepackages.html>`_ instead,
+If you are on a slow network, you'll want to get the `BasicTeX package <https://tug.org/mactex/morepackages.html>`_ instead,
 then run the following command to install the other packages that the draft requires::
 
    sudo tlmgr install latexmk isodate substr relsize ulem fixme rsfs extract layouts enumitem l3packages l3kernel imakeidx splitindex xstring
+
+Using homebrew on macOS
+=======================
+
+Instead of downloading packages from the internet, you can use the `HomeBrew package manager <https://brew.sh>` to install all that is needed to build the draft and run checks on it:
+
+   brew install mactex gnu-sed
 
 ---------------------------------------
 Getting Started on Debian-based Systems
--- a/tools/check-output.sh
+++ b/tools/check-output.sh
@@ -11,6 +11,14 @@ function fail() {
     ! sed 's/^\(.\+\.tex\):/file=\1::/;s/^/::error /' | grep .
 }
 
+# If we detect GNU sed (on macOS installed with brew) then we will use it.
+# Otherwise the script can fail on macOS which ships with BSD sed.
+if command -v gsed >/dev/null 2>&1; then
+	GSED=$(command -v gsed)
+	sed() {
+		${GSED} "$@"
+	}
+fi
 
 # Discover "Overfull \[hv]box" and "Reference ... undefined" messages from LaTeX.
 sed -n '/\.tex/{s/^.*\/\([-a-z0-9]\+\.tex\).*$/\1/;h};
--- a/tools/check-source.sh
+++ b/tools/check-source.sh
@@ -21,6 +21,14 @@ function fail() {
     grep .
 }
 
+# If we detect GNU sed (on macOS installed with brew) then we will use it.
+# Otherwise the script can fail on macOS which ships with BSD sed.
+if command -v gsed >/dev/null 2>&1; then
+	GSED=$(command -v gsed)
+	sed() {
+		${GSED} "$@"
+	}
+fi
 
 # We require GNU tools.
 sed --version | grep -Fqe "GNU sed" || { echo "sed is not GNU sed"; exit 1; }


tools/check-output.sh
```diff
@@ -11,6 +11,14 @@ function fail() {
     ! sed 's/^\(.\+\.tex\):/file=\1::/;s/^/::error /' | grep .
 }
 
+# If we detect GNU sed (on macOS installed with brew) then we will use it.
+# Otherwise the script can fail on macOS which ships with BSD sed.
+if command -v gsed >/dev/null 2>&1; then
+	GSED=$(exec command -v gsed)
```
jensmaurer: This should work without the "exec" in between.

tools/check-source.sh
```diff
@@ -21,6 +21,14 @@ function fail() {
     grep .
 }
 
+# If we detect GNU sed (on macOS installed with brew) then we will use it.
+# Otherwise the script can fail on macOS which ships with BSD sed.
+if command -v gsed >/dev/null 2>&1; then
+	GSED=$(exec command -v gsed)
```
jensmaurer: no "exec"

README.rst
```diff
@@ -13,19 +13,26 @@ Get involved:
...
-If you are on a slow network, you'll want to get the `BasicTeX package <http://tug.org/mactex/morepackages.html>`_ instead,
+If you are on a slow network, you'll want to get the `BasicTeX package <https://tug.org/mactex/morepackages.html>`_ instead,
 then run the following command to install the other packages that the draft requires::
 
    sudo tlmgr install latexmk isodate substr relsize ulem fixme rsfs extract layouts enumitem l3packages l3kernel imakeidx splitindex xstring
 
+Using homebrew on macOS
+=======================
+
+Instead of downloading packages from internet, you can use `HomeBrew package manager <https://brew.sh>` to install all is needed to build the draft and run checks on it:
```
jensmaurer:
```suggestion
Instead of downloading packages from internet, you can use the `HomeBrew package manager <https://brew.sh>` to install all that is needed to build the draft and run checks on it:
```

jensmaurer (changes requested):
Looks mostly good.

tools/check-output.sh
hanickadot: you are right, I somehow always use `exec` and didn't consider it will work without it 🙈

README.rst
hanickadot: resolved with recent force push

jensmaurer (approved):

jensmaurer:
@tkoeppe , this looks like an improvement.  (The commit message could use an overhaul, mentioning the http -> https change as well. Maybe even in a separate commit.)

README.rst
```diff
@@ -13,19 +13,26 @@ Get involved:
...
-If you are on a slow network, you'll want to get the `BasicTeX package <http://tug.org/mactex/morepackages.html>`_ instead,
+If you are on a slow network, you'll want to get the `BasicTeX package <https://tug.org/mactex/morepackages.html>`_ instead,
 then run the following command to install the other packages that the draft requires::
 
    sudo tlmgr install latexmk isodate substr relsize ulem fixme rsfs extract layouts enumitem l3packages l3kernel imakeidx splitindex xstring
 
+Using homebrew on macOS
+=======================
+
+Instead of downloading packages from internet, you can use the `HomeBrew package manager <https://brew.sh>` to install all that is needed to build the draft and run checks on it:
```
jwakely:
```suggestion
Instead of downloading packages from the internet,
you can use the [HomeBrew package manager](https://brew.sh)
to install everything needed to build the draft and run the checking scripts:
```

hanickadot: Why did you use markdown link format? I used same format as in the rest of the document, even googled the .rst syntax when I didn't know what exactly the format should work for inline code blocks. also why adding line breaks? They are not consistently applied thru the document as it is. Not disagreeing, just wondering, personally I would convert it to markdown.

jwakely: Oh sorry, I wasn't paying attention and missed it's rst not markdown. 

The only important bit of my suggestion is "the internet" but I think the "everything needed ..." wording change is a minor improvement. 

The overlong lines aren't important and can be kept for local consistency.

hanickadot: `the internet` fixed, I force pushed rebased over latest changes from `main`.

README.rst
```diff
@@ -13,19 +13,26 @@ Get involved:
...
 
----------------------------
-Getting Started on Mac OS X
----------------------------
+------------------------
+Getting Started on macOS
+------------------------
 
-Install the `MacTeX distribution <http://tug.org/mactex/>`_.
+Install the `MacTeX distribution <https://tug.org/mactex/>`_.
```
tkoeppe: Why the trailing space?

README.rst
```diff
@@ -13,19 +13,26 @@ Get involved:
...
-If you are on a slow network, you'll want to get the `BasicTeX package <http://tug.org/mactex/morepackages.html>`_ instead,
+If you are on a slow network, you'll want to get the `BasicTeX package <https://tug.org/mactex/morepackages.html>`_ instead,
 then run the following command to install the other packages that the draft requires::
 
    sudo tlmgr install latexmk isodate substr relsize ulem fixme rsfs extract layouts enumitem l3packages l3kernel imakeidx splitindex xstring
 
+Using homebrew on macOS
+=======================
+
+Instead of downloading packages from the internet, you can use the `HomeBrew package manager <https://brew.sh>` to install all that is needed to build the draft and run checks on it:
```
tkoeppe, line 32: What is "Instead of downloading packages from the internet" referring to? Does `brew` not download anything from the internet? And is this `brew install mactex gnu-sed` a complete replacement for the `tlmgr` installation?

hanickadot: Brew validates the packages against known checksums. It's against general recommendation "thou shall not download and install stuff random from internet". Brew also manages updating it and also makes easy to uninstall, without depending on package specific way how to remove stuff. 

Yes, `brew install mactex` is a complete replacement for the `tlmgr` installation.

README.rst
hanickadot: A mistake from some reformatting.

README.rst
hanickadot: You can also install `BasicTeX` using `brew` and then use `tlmgr`, but it felt really convoluted and I felt I would end up describing all the combinations. My intention with this PR is just to provide quickest and most straightforwards way how to start with this repo, hence the auto config of `gsed` instead of asking users to change their whole environment and probably introduce some future problems.

PR #8696, opened 2026-01-04
[time.syn] Use `\placeholder` instead of `\term`

frederick-vs-ja:
"Signed integer type of at least N bits" are not terms defined in [time.syn], so perhaps we should use `\placeholder` instead.


jensmaurer (approved):
Yeah.

For later, we should probably mention "width" here, which is part of our story around integers.

PR #8704, opened 2026-01-11
[expr.prim.lambda.capture] Insert linebreak in comment

Andreas-Krug:
Linebreak analog Example 2 with "// have the same name" moved into new line


jensmaurer (approved):
Thanks for fixing this bad automatic line break!

PR #8725, opened 2026-01-29
[dcl.type.pack.index] Clarify pack-index-specifier rules

brevzin:
Currently, [dcl.type.pack.index] reads:

> ```
> pack-index-specifier:
>     typedef-name ... [ constant-expression ]
> ```
> The _typedef-name_ `P` in a _pack-index-specifier_ shall denote a pack.

A _typedef-name_ is either an _identifier_ or a _simple-template-id_, which suggests that maybe `something<Ts>...[0]` is valid in addition to `Ts...[0]`, but the first sentence here restricts that case, since a _simple-template-id_ can never denote a pack.

But to figure that out is kind of a logic puzzle. Can we just say that more directly?

> The _typedef-name_ `P` in a _pack-index-specifier_ shall <ins>be an _identifier_ that</ins> denote<ins>s</ins> a pack.

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -1537,7 +1537,7 @@ forces \tcode{char} objects to be signed; it is redundant in other contexts.
 
 \pnum
 The \grammarterm{typedef-name} $P$ in a \grammarterm{pack-index-specifier}
-shall denote a pack.
+shall be an \grammarterm{identifier} that denotes a pack.
 
 \pnum
 The \grammarterm{constant-expression} shall be


pdimov:
FWIW, I'd prefer `something<Ts>...[0]` to be valid, and GCC currently accepts it.

brevzin:
Actually, the expression wording explicitly uses the wording I'm suggesting here — which suggests maybe that `something<Ts>...[0]` is supposed to work. Although I don't see how `something<Ts>` can be said to denote a pack. 

Do we just need a different sentence entirely here?

brevzin:
Given that p2 currently talks about `sizeof...(P)` which is only valid for names, I think that we definitely mean to only support names. Otherwise, p2 would also need to change.

jensmaurer:
We definitely want an _identifier_ here; nothing else can "denote a pack".  (Pack expansions can appear in select places, but this is different.)

Parallelism with [expr.prim.pack.index] p1 suggests the amended phrasing is a consistency improvement.

jensmaurer (approved):

PR #8729, opened 2026-01-30
[cpp.replace.general] Fix singular/plural mismatch

Andreas-Krug:


source/preprocessor.tex
```diff
@@ -1606,7 +1606,7 @@
 The replaced sequence of preprocessing tokens is terminated by the matching
 \tcode{)}
 preprocessing token, skipping intervening matched pairs of left and
-right parenthesis preprocessing tokens.
+right parentheses preprocessing tokens.
```
jensmaurer, line 1609: I don't think this change is correct English.

"left and right apple slices" doesn't use plural for "apple", either.

@jwakely ?

eisenwave: "Parenthesis" acts as an attributive noun in this case. Normally, those appear in their singular form with some exceptions (childrens books, ladies man); I don't think any of these apply here.

The wording seems correct as is to me.

PR #8758, opened 2026-02-21
[const.wrap.class] Add missing this to compound assignment operators

timsong-cpp:
This corrects a misapplication of LWG4383.


jensmaurer (approved):

jensmaurer:
@tkoeppe , this is pretty important.  The change bdcfe2c9a54ca350e9bfc59227bb0285a59c635d that applied LWG4383 removed the "this" here, with no basis in the incoming issue drafting.

PR #8762, opened 2026-02-22
[basic.fundamental] Harmonize "Type X is a distinct type..." wordings

Quuxplusone:
Types are types; types do not denote types.

Here I don't know if there might be a reason we talk about cv nullptr_t being multiple distinct *types*, as if otherwise the reader might imagine const nullptr_t to be the same type as volatile nullptr_t. Otherwise, that sentence should clearly be just "std::nullptr_t is a distinct type."

Notably `std::meta::info` is *not* currently described as a distinct type. I don't think that's intentional; I think it's practically required to be a distinct type anyway. So I think this diff is editorial.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -5521,7 +5521,7 @@ signed or unsigned integer type as its underlying type.
 \indextext{\idxcode{char8_t}|see{type, \tcode{char8_t}}}%
 \indextext{type!\idxcode{char8_t}}%
 \indextext{type!underlying!\idxcode{char8_t}}%
-Type \keyword{char8_t} denotes a distinct type
+Type \keyword{char8_t} is a distinct type
 whose underlying type is \tcode{\keyword{unsigned} \keyword{char}}.
 \indextext{\idxcode{char16_t}|see{type, \tcode{char16_t}}}%
 \indextext{\idxcode{char32_t}|see{type, \tcode{char32_t}}}%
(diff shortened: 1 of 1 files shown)


eisenwave:
There are some pretty chaotic inconsistencies in there, and the use of "denotes" is pretty dubious.

source/basic.tex
```diff
@@ -5656,13 +5656,14 @@
 \end{note}
 
 \pnum
-The types denoted by \cv~\tcode{std::nullptr_t} are distinct types.
+Type \tcode{std::nullptr_t} is a distinct type.
 A prvalue of type \tcode{std::nullptr_t} is a null pointer
 constant\iref{conv.ptr}. Such values participate in the pointer and the
```
Quuxplusone, line 5661: The sentence I'm modifying here was added by [CWG2689 "Are cv-qualified std::nullptr_t fundamental types?"](https://cplusplus.github.io/CWG/issues/2689.html) in 2024.

Incidentally, I wonder whether the previous emphasis on _cv_ `nullptr_t` was supposed to imply something here: that maybe prvalues of type, say, `volatile nullptr_t` are _not_ null pointer constants? Would that be detectable in any way?

I think it might be impossible to form a prvalue of type `const nullptr_t` or `volatile nullptr_t`, though. Certainly one can't do it via [[conv.lval]](https://eel.is/c++draft/conv.lval#3.1.sentence-1): "A glvalue of a non-function, non-array type `T` can be converted to a prvalue. [...] If `T` is a non-class type, the type of the prvalue is the cv-unqualified version of `T`. [...] If `T` is _cv_ `std::nullptr_t`, the result is a null pointer constant."

eisenwave: Lvalue-to-rvalue conversions would only happen if you start out with a glvalue, but can't you create a prvalue using
```cpp
using T = volatile std::nullptr_t;
T{};
```
Also, if you didn't have a null pointer constant, I would expect something to blow up during constant evaluation. For example, you can add `0` to a null pointer despite not having a pointer to an array element. That should otherwise not be a constant expression.

PR #8763, opened 2026-02-22
[iterator.concept.winc] Refactor definition of "integer-like type"

Quuxplusone:
- "`I` models `integral<I>`" is not how we say it: the `<I>` shouldn't be there.
- The concept `integral` is modeled only by integral types. Cut out the middleman.
- Make the sentence structure less hypothetical; use "collectively called" as we usually do.
- Clearly specify that cv bool is *not* integer-like, rather than leaving it out of the premise.

### DIFF
--- a/source/iterators.tex
+++ b/source/iterators.tex
@@ -1445,15 +1445,12 @@ The width of an integer-class type is greater than
 that of every integral type of the same signedness.
 
 \pnum
-A type \tcode{I} other than \cv{}~\tcode{bool} is \defn{integer-like}
-if it models \tcode{\libconcept{integral}<I>} or
-if it is an integer-class type.
-An integer-like type \tcode{I} is \defn{signed-integer-like}
-if it models \tcode{\libconcept{signed_integral}<I>} or
-if it is a signed-integer-class type.
-An integer-like type \tcode{I} is \defn{unsigned-integer-like}
-if it models \tcode{\libconcept{unsigned_integral}<I>} or
-if it is an unsigned-integer-class type.
+Integer-class types and integral types\iref{basic.fundamental} other than \cv{}~\tcode{bool}
+are collectively called \defnadjx{integer-like}{types}{type}.
+Signed-integer-class types and types that model \tcode{\libconcept{signed_integral}}
+are collectively called \defnadjx{signed-integer-like}{types}{type}.
+Unsigned-integer-class types, and types other than \cv{}~\tcode{bool} that model \tcode{\libconcept{unsigned_integral}},
+are collectively called \defnadjx{unsigned-integer-like}{types}{type}.
 
 \pnum
 For every integer-class type \tcode{I},


source/iterators.tex
```diff
@@ -1445,15 +1445,12 @@
...
-if it is an integer-class type.
-An integer-like type \tcode{I} is \defn{signed-integer-like}
-if it models \tcode{\libconcept{signed_integral}<I>} or
-if it is a signed-integer-class type.
-An integer-like type \tcode{I} is \defn{unsigned-integer-like}
-if it models \tcode{\libconcept{unsigned_integral}<I>} or
-if it is an unsigned-integer-class type.
+Integer-class types and integral types\iref{basic.fundamental} other than \cv{}~\tcode{bool}
+are collectively called \defnadjx{integer-like}{types}{type}.
+Signed-integer-class types and signed integral types
```
jensmaurer: [concepts.arithmetic] p1 claims that "signed integral type" and `signed_integral<T>` are not isomorphic, but this change seems to assume that.

jensmaurer (changes requested):

source/iterators.tex
Quuxplusone: Oops, you're right! (`char8_t` models `unsigned_integral`, because `char8_t` is an _integral type_ and happens to be unsigned; but `char8_t` is not an _unsigned integral type_ because it is not one of the [_standard_ or _extended unsigned integral types_](https://eel.is/c++draft/basic.fundamental#2.sentence-3).)

IIUC, it _is_ true that the types that model `integral` are precisely the integral types; so that bit of the rewrite can stay. But I'll have to put back "types modeling `signed_integral`" in place of "signed integral types" and likewise for `unsigned_integral`.

source/iterators.tex
```diff
@@ -1445,15 +1445,12 @@
...
-if it models \tcode{\libconcept{signed_integral}<I>} or
-if it is a signed-integer-class type.
-An integer-like type \tcode{I} is \defn{unsigned-integer-like}
-if it models \tcode{\libconcept{unsigned_integral}<I>} or
-if it is an unsigned-integer-class type.
+Integer-class types and integral types\iref{basic.fundamental} other than \cv{}~\tcode{bool}
+are collectively called \defnadjx{integer-like}{types}{type}.
+Signed-integer-class types and types that model \tcode{\libconcept{signed_integral}}
+are collectively called \defnadjx{signed-integer-like}{types}{type}.
+Unsigned-integer-class types, and types other than \cv{}~\tcode{bool} that model \tcode{\libconcept{unsigned_integral}},
```
jensmaurer, line 1452: Where is this "bool" exclusion coming from?  It's not in the original text for unsigned-integer-like.

jensmaurer (changes requested):

source/iterators.tex
Quuxplusone: The original text is

> An integer-like type `I` is _unsigned-integer-like_ if it models `unsigned_integral<I>` or if it is an unsigned-integer-class type.

Now, is `bool` an _unsigned-integer-like_ type? Arguably the reader can't use the original sentence to say one way or the other, since the original sentence tells us anything at all only for "an integer-like type," which `bool` isn't.

But the _intent_, I think, is that the _unsigned-integer-like_ types should be a strict subset of the _integer-like_ types.

Suppose the new text was:

> Unsigned-integer-class types, and types <s>other than _cv_ `bool`</s> that model `unsigned_integral`, are collectively called _unsigned-integer-like types._

Then `bool` (which models `unsigned_integral`) would be considered _unsigned-integer-like_ but not _integer-like_. I'm sure we don't intend that.

Quuxplusone: Alternatively, how about this? (Signed-integer-class types and unsigned-integer-class types are collectively known as integer-class types: [[iterator.concept.winc]/3](https://eel.is/c++draft/iterator.concept.winc#3).)

```
Signed-integer-class types and types that model \tcode{\libconcept{signed_integral}}
are collectively called \defnadjx{signed-integer-like}{types}{type}.
Unsigned-integer-class types, and types other than \cv{}~\tcode{bool} that model \tcode{\libconcept{unsigned_integral}},
are collectively called \defnadjx{unsigned-integer-like}{types}{type}.
Signed-integer-like types and unsigned-integer-like types
are collectively called \defnadjx{integer-like}{types}{type}.
```
This puts the "except `bool`" exception in just one place instead of two, and makes it very clear that integer-like types are precisely partitioned into signed and unsigned just like integer-class types are partitioned into signed and unsigned and integral types are partitioned into signed and unsigned.

PR #8766, opened 2026-02-24
[exec.snd.expos] Fix order of make_obj_using_allocator args

jwakely:
Fixes https://github.com/cplusplus/sender-receiver/issues/353


jwakely:
The intended meaning is clear, but the wording just fails to use the `make_obj_using_allocator` function correctly. I think this qualifies as editorial, as an "internal inconsistency".

ericniebler:
Thank you @jwakely and @jensmaurer !

PR #8767, opened 2026-02-25
[format] Consistently add `constexpr` to function descriptions

frederick-vs-ja:
P3391R2 only added `constexpr` to synopses but not descriptions for some formatting functions. This PR consistently adds `constexpr` to the descriptions.


eisenwave:
@jwakely can you take a look at this and at #8768? We may want to deal with these before C++26 ships.

PR #8768, opened 2026-02-25
[vector.bool.fmt][container.adaptors.format] Add missing `constexpr`

frederick-vs-ja:
This PR completes the previously missing `constexpr` addition from P3391R2 and consistently adds `constexpr` to function descriptions.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -10735,7 +10735,7 @@ namespace std {
         parse(ParseContext& ctx);
 
     template<class FormatContext>
-      typename FormatContext::iterator
+      constexpr typename FormatContext::iterator
         format(const T& ref, FormatContext& ctx) const;
   };
 }
@@ -10756,7 +10756,7 @@ Equivalent to: \tcode{return \exposid{underlying_}.parse(ctx);}
 \indexlibrarymember{format}{formatter}%
 \begin{itemdecl}
 template<class FormatContext>
-  typename FormatContext::iterator
+  constexpr typename FormatContext::iterator
     format(const T& ref, FormatContext& ctx) const;
 \end{itemdecl}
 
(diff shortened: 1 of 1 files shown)


source/containers.tex
```diff
@@ -20308,7 +20308,7 @@
 \tcode{queue},
 \tcode{priority_queue}, and
 \tcode{stack},
-the library provides the following formatter specialization
+the library provides the following constexpr-enabled formatter specialization
```
jensmaurer, line 20311: "constexpr-enabled" does not imply "enabled" according to [format.formatter.spec], so this is a normative change.

jwakely: I think it's just a missing edit from the application of P3391, and similarly for the next change below.

jwakely: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p3391r2.html#pnum_14

jwakely: So yes, it is a normative change, but one that was approved at plenary.

jensmaurer: So, you're saying this is a missed edit from the application of P3391R2, beyond a few sprinkled "constexpr" annotations ?   That would be good to say in the commit comment.

But that causes an LWG issue, because "constexpr-enabled" doesn't imply "enabled", I think.

jwakely: We have: 

>If the library provides an explicit or partial specialization of `formatter<T, charT>`, that specialization is enabled and meets the Formatter requirements except as noted otherwise.

So I think these specializations are enabled just by virtue of being present in the library spec. We don't also need to say they're enabled

jensmaurer: Ok.

PR #8770, opened 2026-02-25
[utility.intcmp] Clarify the referenced type parameter

term-est:
in_range is worded as:
```
template<class R, class T>
  constexpr bool in_range(T t) noexcept;
```

with the mandate `Each of T and U is a signed or unsigned integer type`
The referenced U in the mandate seems to be a copy-paste error from the above `cmp_equal`


eisenwave:
Editorially broken recently in #8616.

PR #8789, opened 2026-03-14
[version.syn] Added freestanding for type_order

hewillk:
Although `<compare>` already shows `// all freestanding`, it still seems useful to add `freestanding` here, which is consistent with `__cpp_lib_three_way_comparison`, which is annotated with  `// freestanding, also in <compare>`.

Or is this LWG?


eisenwave:
Traditionally, changes to the `// freestanding` comments in [version.syn] have been done via LWG issue, such as LWG4126, but maybe @jwakely would be open to processing this one editorially.

jensmaurer:
For some reason, I can't re-run the failed tests.

PR #8793, opened 2026-03-18
[views.multidim] Index `submdspan_mapping`

### LINKED ISSUE #8792: submdspan_mapping is not indexed
It's not obvious how to index it, because it's specified a bit strangely.

Maybe one entry in [mdspan.sub.map.sliceable] where the rules on finding it via ADL are stated, and once for each friend declaration of it?

frederick-vs-ja:
Fixes #8792.

Avoid `<PaddingValue>` in indexing because we're already using `layout_{left,right}_padded::mapping` in indices.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -22203,7 +22203,7 @@ namespace std {
         -> @\seebelow@;
 
     template<class... SliceSpecifiers>
-      friend constexpr auto submdspan_mapping(
+      friend constexpr auto @\libmember{submdspan_mapping}{layout_left::mapping}@(
         const mapping& src, SliceSpecifiers... slices) {
           return src.@\exposid{submdspan-mapping-impl}@(slices...);
       }
@@ -22532,7 +22532,7 @@ namespace std {
         -> @\seebelow@;
 
     template<class... SliceSpecifiers>
-      friend constexpr auto submdspan_mapping(
+      friend constexpr auto @\libmember{submdspan_mapping}{layout_right::mapping}@(
         const mapping& src, SliceSpecifiers... slices) {
           return src.@\exposid{submdspan-mapping-impl}@(slices...);
       }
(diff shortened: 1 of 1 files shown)


source/containers.tex
```diff
@@ -23278,7 +23278,7 @@
         -> @\seebelow@;
 
     template<class... SliceSpecifiers>
-      friend constexpr auto submdspan_mapping(const mapping& src, SliceSpecifiers... slices) {
+      friend constexpr auto @\libmember{submdspan_mapping}{layout_left_padded<PaddingValue>::mapping}@(const mapping& src, SliceSpecifiers... slices) {
```
frederick-vs-ja: Do we want to use `layout_left_padded<PaddingValue>::mapping` or (incomplete but perhaps sufficient) `layout_left_padded::mapping` in the index?

frederick-vs-ja: Also, many members of these `mapping` classes are not yet indexed. Perhaps we can index them in a follow-up PR.

source/containers.tex
```diff
@@ -26013,6 +26013,7 @@
 \end{itemize}
 
 \pnum
+\indexlibraryglobal{submdspan_mapping}%
```
frederick-vs-ja, line 26016: This paragraph is perhaps the right place for "primary" `submdspan_mapping`, but it's somehow duplicate of [[contents]/3](https://eel.is/c++draft/contents#3) which also lists `submdspan_mapping`.

source/containers.tex
frederick-vs-ja: We're now using `\indexlibraryctor{layout_left_padded::mapping}` so perhaps it would be more consistent to use `\libmember{submdspan_mapping}{layout_left_padded::mapping}`.

jensmaurer (approved):

PR #8797, opened 2026-03-19
[system.error.syn] Index `is_error_condition_enum_v` and enumerators of `errc`

frederick-vs-ja:


eisenwave:
I guess we could do the same in https://eel.is/c++draft/text.encoding.id

frederick-vs-ja:
> I guess we could do the same in https://eel.is/c++draft/text.encoding.id

Seems belonging to another PR to me. I'm working on this.

Thanks to `using enum id;`, these enumerators can be referred as both <code>std::text_encoding::id::_meow_</code> and <code>std::text_encoding::_meow_</code>. Do we want to index them as members of `text_encoding::id` or `text_encoding`?

jwakely:
I think they should be indexed as members of `text_encoding` because that's the intended way to refer to them, but I don't have any preference whether they're also indexed under `id`.

jensmaurer (approved):

PR #8803, opened 2026-03-20
[dcl.struct.bind] Fix tuple-like binding index to use SB_i instead of v_i

abhinavagarwal07:
P1061R10 introduced the SB<sub>i</sub> notation for post-expansion structured bindings but this sentence in p7 (tuple-like case) was not updated. The rest of the section (p6, p8, and the end of this same paragraph) already uses SB<sub>i</sub>.

See [std-discussion thread](https://lists.isocpp.org/std-discussion/2026/03/3312.php).

### DIFF
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -7724,7 +7724,7 @@ shall be a well-formed integral constant expression
 whose value is non-negative;
 the structured binding size of \tcode{E} is equal to that value.
 Let \tcode{i} be an index prvalue of type \tcode{std::size_t}
-corresponding to $\tcode{v}_i$.
+corresponding to $\textrm{SB}_i$.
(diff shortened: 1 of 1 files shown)


eisenwave:
Looks like P1061R10 simply forgot to update the $v_i$ which was added by CWG2339 to $SB_i$. It did update $v_i$ elsewhere though.

I'm still not 100% sure how we got that specific wording (some paragraphs got rearranged/merged), but the change looks correct.

tkoeppe:
@brevzin could you have a look?

brevzin:
I'm confused as to how this happened, I guess I just missed that sentence. But yeah, we want to refer to the expanded pack here.

tkoeppe:
Yes, that seems reasonable. Thanks!

PR #8814, opened 2026-03-26
[expr.const.defns] fix comment alignment

Tsche:


source/expressions.tex
```diff
@@ -9292,7 +9292,7 @@
 constexpr int f() {
   const int n = std::is_constant_evaluated() ? 13 : 17; // \tcode{n} is 13
   int m = std::is_constant_evaluated() ? 13 : 17;       // \tcode{m} can be 13 or 17 (see below)
-  char arr[n] = {}; // char[13]
+  char arr[n] = {};                                     // char[13]
```
eisenwave:
```suggestion
  char arr[n] = {};                                     // \tcode{char[13]}
```
While we're at it.

PR #8822, opened 2026-03-28
[exec.snd.expos] Delete unused expos-only concept "completion-tag".

tkoeppe:
Also insert a missing \pnum before the now-split codeblock (see 8cef46b15b21d4ab1b96dce04c0d37e511d1ddc2).


tkoeppe:
@timsong-cpp For your consideration.

PR #8878, opened 2026-03-28
[2026-03 LWG Motion 16] P3948R1 constant_wrapper is the only tool needed

### LINKED ISSUE #2598: P3948 R0 constant_wrapper is the only tool needed for passing constant expressions
[P3948R0](https://wg21.link/p3948r0) constant_wrapper is the only tool needed for passing constant expressions (Matthias Kretz)

### LINKED ISSUE #8850: [2026-03 LWG Motion 16] P3948R1 constant_wrapper is the only tool needed
[P3948R1](https://wiki.isocpp.org/uploads/2026-03_Croydon-LWGStrawPollsStaging_-_P3948R1.pdf) (constant_wrapper is the only tool needed

eisenwave:
Fixes NB FR-019-210 and FR-021-218 (C++26 CD).

Fixes #8850

Also fixes cplusplus/papers#2598
Also fixes https://github.com/cplusplus/nbballot/issues/779
Also fixes https://github.com/cplusplus/nbballot/issues/788

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -225,15 +225,6 @@ namespace std {
     };
   template<size_t I> constexpr in_place_index_t<I> in_place_index{};
 
-  // \tcode{constant_arg} argument tag%
-\indexlibraryglobal{constant_arg_t}%
-\indexlibraryglobal{constant_arg}
-  template<auto V>
-    struct constant_arg_t {
-      explicit constant_arg_t() = default;
-    };
-  template<auto V> constexpr constant_arg_t<V> constant_arg{};
-
   // \ref{variant.monostate}, class \tcode{monostate}%
 \indexlibraryglobal{monostate}
   struct monostate;
@@ -15254,9 +15245,11 @@ namespace std {
     // \ref{func.wrap.ref.ctor}, constructors and assignment operators
     template<class F> function_ref(F*) noexcept;
     template<class F> constexpr function_ref(F&&) noexcept;
-    template<auto f> constexpr function_ref(constant_arg_t<f>) noexcept;
-    template<auto f, class U> constexpr function_ref(constant_arg_t<f>, U&&) noexcept;
-    template<auto f, class T> constexpr function_ref(constant_arg_t<f>, @\cv{}@ T*) noexcept;
+    template<auto c, class F> constexpr function_ref(constant_wrapper<c, F>) noexcept;
+    template<auto c, class F, class U>
+      constexpr function_ref(constant_wrapper<c, F>, U&&) noexcept;
+    template<auto c, class F, class T>
+      constexpr function_ref(constant_wrapper<c, F>, @\cv{}@ T*) noexcept;
 
     constexpr function_ref(const function_ref&) noexcept = default;
     constexpr function_ref& operator=(const function_ref&) noexcept = default;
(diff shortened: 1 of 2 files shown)


eisenwave:
@jwakely @mattkretz we have goofed HARD. The paper deletes `constant_arg_t` but not the deduction guides that `function_ref` has for it.

https://eel.is/c++draft/func.wrap.ref.deduct

(will deal with overfull hboxes once home btw)

source/utilities.tex
```diff
@@ -15311,21 +15302,26 @@
...
 \constraints
 \tcode{\exposid{is-invocable-using}<const F\&>} is \tcode{true}.
 
 \pnum
 \mandates
+\begin{itemize}
+\item
 If \tcode{is_pointer_v<F> || is_member_pointer_v<F>} is \tcode{true},
-then \tcode{f != nullptr} is \tcode{true}.
+then \tcode{f != nullptr} is \tcode{true}, and
```
```suggestion
then \tcode{f.value != nullptr} is \tcode{true}, and
```

jwakely (changes requested):

jwakely:
> @jwakely @mattkretz we have goofed HARD. The paper deletes `constant_arg_t` but not the deduction guides that `function_ref` has for it.


Can we just delete that (which is fine as an editorial change, because the draft is inconsistent if it refers to a type which no longer exists), or does it need to be changed to use `constant_wrapper`?

eisenwave:
I believe the purpose of the deduction guides is to make things possible like
```cpp
void f();
// CTAD
std::function_ref ref = std::nontype<f>;
```
Or now equivalently with `std::cw<f>`. If this no longer works as a result of deleting the deduction guides, that seems like an inadvertent design change.

I suppose the only editorial way to get the draft into a working state is to delete the deduction guides editorially and to reintroduce them by LWG issue. Any LWG resolutions won't be merged into the DIS anyway, right? We need to ship these motions.

zhihaoy:
Herb reminded us that accepting motion can also introduce editorial change, so I suggest just finishing the renaming here.

Mick235711:
Personally, I support treating this as an editorial wording defect in the proposal, as the intention is obviously to replace all usage of `constant_arg` mechanically. In other similar renaming proposals, if the author left over a few instances in the wording, they are also editorially applied without LWG issues as long as the intention is clear when polling the motions.

Bronek:
@Mick235711 @zhihaoy I think in this case the intention was clear, so I would agree that this change should just replace `constant_arg` with `constant_wrapper`, ensuring that this CTAD works `std::function_ref ref = std::cw<foo>;` as long as that used to work `std::function_ref ref = std::constant_arg<foo>; `, for some nullary `foo`; and treat this as a necessary editorial.

EDIT: The intention was unfortunately not captured as clearly as I would wish, but my memory is rather clear that we wanted to "replace all the uses of `constant_arg` with `constant_wrapper` and remove `constant_arg`" and this is also what the paper being accepted says in the subject ("constant_wrapper is the **only** tool needed", emphasis mine)

mattkretz:
1. Sorry!
2. I did change the deduction guide in my libstdc++ patch, so I really should have noticed by looking through the patch. :facepalm: 
3. I'll provide wording that matches a tested implementation and then we can decide whether we call that editorial or not. I agree that the intent of the change is very clear (replace `constant_arg_t` by `constant_wrapper`) and that the removal of `constant_arg_t` forces a (editorial) change to the deduction guide.

mattkretz:
This would be the necessary change for CTAD:
https://github.com/mattkretz/cplusplus-draft/commit/92a7ba1d766198220ac1e62b0eade57eb9b0becd
I implemented it and it passes all (CTAD) tests in libstdc++.

Bronek:
> This would be the necessary change for CTAD: [mattkretz@92a7ba1](https://github.com/mattkretz/cplusplus-draft/commit/92a7ba1d766198220ac1e62b0eade57eb9b0becd) I implemented it and it passes all (CTAD) tests in libstdc++.

This looks like what we voted for in LEWG, i.e. design intent.

Bronek:
> > @jwakely @mattkretz we have goofed HARD. The paper deletes `constant_arg_t` but not the deduction guides that `function_ref` has for it.
> 
> Can we just delete that (which is fine as an editorial change, because the draft is inconsistent if it refers to a type which no longer exists), or does it need to be changed to use `constant_wrapper`?

I would argue that this needs to be changed to `constant_wrapper` like @mattkretz has demonstrated above, but it is your decision whether do it immediately or use a DR instead. Just a note that change to `constant_wrapper` reflects the design intent as approved by LEWG, which is why I **personally** think DR would be suboptimal.

eisenwave:
I've incorporated Matthias' commit into the PR.

jwakely (approved):
I agree with the comments here that what LEWG approved did not include "and disable CTAD" so preserving support for CTAD is part of the intended change.

Thanks for the fixes to make the wording match that approved design.

source/utilities.tex
```diff
@@ -15200,10 +15193,10 @@
   // \ref{func.wrap.ref.deduct}, deduction guides
   template<class F>
     function_ref(F*) -> function_ref<F>;
-  template<auto f>
-    function_ref(constant_arg_t<f>) -> function_ref<@\seebelow@>;
-  template<auto f, class T>
-    function_ref(constant_arg_t<f>, T&&) -> function_ref<@\seebelow@>;
+  template<auto c, class F0>
+    function_ref(constant_wrapper<c, F>) -> function_ref<@\seebelow@>;
```
lichray: F -> F0

tkoeppe:
Thanks, everyone!

PR #8886, opened 2026-03-29
[2026-03 CWG Motion 7] P3865R3 Class template argument deduction (CTAD) for type template template parameters

### LINKED ISSUE #670: [dcl.type.simple] Significance of "nam[ing]" a deducible template for CTAD

Reference (section label): dcl.type.simple, over.match.class.deduct

Issue description:
The wording of https://eel.is/c++draft/dcl.type.simple#3 successfully achieves avoiding dependent templates for CTAD, but the wording fails to be explicit enough. In particular, naming a template template parameter _names_ a template template parameter (not a deducible template).

The over.match.class.deduct wording similarly relies on this notion of naming a deducible template (https://eel.is/c++draft/over.match.class.deduct#3 starts with "the guides of the template named by the _simple-template-id_ of the _defining-type-id_").

Implementations appear to allow deduction from a template template parameter but also appear to be unprepared for the implications of doing so.

Clang, GCC, EDG, and MSVC all accept (https://godbolt.org/z/8Kevd9dze):
```cpp
template <template <typename> class TT>
void f() { TT x(42); }

template <typename T> struct A { A(T); };

void g() { f<A>(); }
```

Clang, GCC, and EDG crash; MSVC rejects (https://godbolt.org/z/Pd9o76cWo):
```cpp
template <typename T> struct A { A(T); };

template <typename T, template <typename> class TT = A>
using Alias = TT<T>;

template <typename T>
using Alias2 = Alias<T>;

void h() { Alias2 a(42); }
void h2() { Alias a(42); }
```

Suggested resolution:
Add a note to the effect that naming a template template parameter does not name a deducible template.

### LINKED ISSUE #2443: CWG3003 Naming a deducible template for class template argument deduction
The core language specification is clear that dependent names cannot designate templates subject to class template argument deduction, thus excluding the names of template template parameters.

However, https://wg21.link/range.utility.conv.to#4 relies on such a non-existing facility.

CWG3003 seeks to add a note to highlight the status quo of the core language specification.

LWG4381 asks for removal of that unimplementable feature.

If (L)EWG wishes to add a feature to the core language that makes https://wg21.link/range.utility.conv.to#4 valid (possibly limited to situations similar to the first example in CWG3003, excluding the second example), a paper would be needed.

### LINKED ISSUE #2469: P3865 R1 Class template argument deduction (CTAD) for type template template parameters
[P3865R0](https://wg21.link/p3865r0) Class template argument deduction (CTAD) for type template template parameters (Christof Meerwald)

### LINKED ISSUE #8828: [2026-03 CWG Motion 7] P3865R3 Class template argument deduction (CTAD) for type template template parameters
[P3865R3](https://wiki.isocpp.org/uploads/2026-03_Croydon-StrawPolls_-_P3865r3.pdf) (Class template argument deduction (CTAD) for type template template parameters)

eisenwave:
Fixes #8828

Also fixes https://github.com/cplusplus/papers/issues/2469
Also fixes https://github.com/cplusplus/papers/issues/2443
Also fixes https://github.com/cplusplus/CWG/issues/670

### DIFF
--- a/source/overloading.tex
+++ b/source/overloading.tex
@@ -1443,7 +1443,7 @@ whose template parameter list is that of \tcode{P} and
 whose \grammarterm{defining-type-id} is a \grammarterm{simple-template-id}
 whose \grammarterm{template-name} designates the type template template argument and
 whose \grammarterm{template-argument-list} is the template argument list of \tcode{P}.
-\tcode{A} is then used
+The alias template \tcode{A} is then used
 instead of the original \grammarterm{template-name}
 to resolve the placeholder.
 
--- a/source/overloading.tex
+++ b/source/overloading.tex
@@ -1433,6 +1433,21 @@ J j = { "ghi" };    // error: cannot bind reference to array of \tcode{unsigned 
 \end{example}
 
 \pnum
+When resolving a placeholder for a deduced class type
+where the \grammarterm{template-name}
+designates a type template template parameter \tcode{P},
+the type template template argument for \tcode{P}
+shall be a deducible template.
+Let \tcode{A} be an alias template
+whose template parameter list is that of \tcode{P} and
+whose \grammarterm{defining-type-id} is a \grammarterm{simple-template-id}
+whose \grammarterm{template-name} designates the type template template argument and
+whose \grammarterm{template-argument-list} is the template argument list of \tcode{P}.
+\tcode{A} is then used
+instead of the original \grammarterm{template-name}
+to resolve the placeholder.
+
+\pnum
 When resolving a placeholder for a deduced class type\iref{dcl.type.simple}
 where
 the \grammarterm{template-name} or \grammarterm{splice-type-specifier}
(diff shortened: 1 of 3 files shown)


source/overloading.tex
```diff
@@ -1432,6 +1432,22 @@
...
+Let \tcode{A} be an alias template
+whose template parameter list is that of \tcode{P} and
+whose \grammarterm{defining-type-id} is a \grammarterm{simple-template-id}
+whose \grammarterm{template-name} designates the type template template argument and
+whose \grammarterm{template-argument-list} is the template argument list of \tcode{P}.
+\tcode{A} is then used instead of the original \grammarterm{template-name}
+to resolve the placeholder.
+
+\pnum
+%FIXME: pull up this iref?
```
tkoeppe: @eisenwave Yes, if you can, please add a commit that moves the reference up (or I'll do it when merging).

tkoeppe (approved):
(optional minor changes)

source/overloading.tex
eisenwave: Nevermind I've dropped that part from this PR. https://eel.is/c++draft/over.match.class.deduct contains two different `\iref`s following "deduced class type", and I feel like the first one is actually pointing to "deduced class type" and the second one is maybe a bit misplaced, and moving the second one up wouldn't make sense here.

The state of the PR should still be exactly as in the motions, with the FIXME removed.

tkoeppe: OK, thanks!

tkoeppe:
Note to self: rephrase start of sentence, as discussed with Core.

PR #8922, opened 2026-04-06
[Motions 2026 03 cwg 2] P4160R0 DR issues except <long list>

### LINKED ISSUE #89: CWG2609 [expr.sizeof] The result of applying sizeof to empty class is standard?

### LINKED ISSUE #148: [temp.param] Is a non-type template parameter that declares a reference a variable?

A *parameter-declaration* that appears in a function declaration declares a variable. How about it appears as a template parameter? The issue arises from https://stackoverflow.com/questions/62567395/is-it-a-defect-about-a-non-type-template-parameter-of-a-reference-type-that-ther?noredirect=1#comment130536482_62567395. 

The opinion in the comment on the question is that: Since a template parameter does not have storage and lifetime, it cannot be considered a variable. However, this point is not clear in the current standard.

### LINKED ISSUE #338: [intro.object]/[basic.life] Multiple objects of the same type at the same address

### LINKED ISSUE #545: [class.default.ctor] Default constructor implicitly declared for class with inherited constructors
Full name of submitter: Svetlin Totev

When is the implicit default constructor created:
11.4.5.2 Default constructors [class.default.ctor]
>1. ... If there is no user-declared constructor for class X, a non-explicit
>constructor having no parameters is implicitly declared as defaulted
>([dcl.fct.def])...

What inheriting constructors does:
9.9 The using declaration [namespace.udecl]
>13 Constructors that are named by a using-declaration are treated as though
>they were constructors of the derived class when looking up the
>constructors of the derived class ([class.qual]) or forming a set of
>overload candidates ([over.match.ctor], [over.match.copy],
>[over.match.list]).

Which constructor is called when passed `{}`:
12.2.2.8 Initialization by list-initialization [over.match.list]
>(1.1) If the initializer list is not empty or T has no default constructor,
>overload resolution is first performed where the candidate functions are
>the initializer-list constructors ([dcl.init.list]) of the class T and the
>argument list consists of the initializer list as a single argument.

Issue description:
Constructors of a derived class inherited from the base class with a using-declaration are not accounted for when checking if there are user-declared constructors. So a default constructor is implicitly declared. Even though the user explicitly declared constructors inherited from the base class.

This is an issue when the derived class must not have a default constructor. Which is likely what any user assumes would happen when inheriting other constructors. The default constructor being added to the constructor lookup is also a big problem when inheriting initializer_list constructors as the default one has higher priority when calling the constructor with `{}`.

Both gcc and clang implement this questionable behaviour.

Explicitly declaring the default constructor as deleted doesn't help as this still counts as a declaration of a default constructor. The only difference being the compilation error you get.

Example:
```c++
#include<initializer_list>
struct A { A(std::initializer_list<int> il) {}; };
struct B : A { using A::A; };
// workaround 1: manually doing what "using A::A;" should do
struct C : A { C(std::initializer_list<int> il) : A(il) {}; };
// workaround 2: declaring a useless constructor to avoid the implicit default constructor
struct D : A {
    using A::A;
    private:
    D(void*); // the argument should probably be something that can't be constructed
};

int main()
{
    // A(); // doesn't work, as expected
    A{}; // works as expected

    A a1 {};        // ok
    A a2 {1, 2, 3}; // ok
    A a3 {5};       // ok

    B b1 {};        // calls implicit B() which calls A() which doesn't exist -> fail
    B b2 {1, 2, 3}; // ok
    B b3 {5};       // ok

    C c1 {};        // ok
    C c2 {1, 2, 3}; // ok
    C c3 {5};       // ok

    D d1 {};        // ok
    D d2 {1, 2, 3}; // ok
    D d3 {5};       // ok
    return 0;
}
```

Suggested resolution:
The using-declaration inherited constructors should count as user-declared constructors as it is an explicit declaration of constructors made by the user. So the default constructor shouldn't be implicitly declared.

### LINKED ISSUE #644: [basic.fundamental] Alignment and value representation of `nullptr_t`
Full name of submitter (unless configured in github; will be published with the issue): Jiang An

Reference (section label): [basic.fundamental]

Issue description:

`nullptr_t` has several properties in C23 that is either not specified or different in C++.

[WG14 N3220](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3220.pdf) 7.21.2 p3 reads:
> 3. The size and alignment of `nullptr_t` is the same as for a pointer to character type. An object representation of the value `nullptr` is the same as the object representation of a null pointer value of type `void*`. An lvalue conversion of an object of type `nullptr_t` with such an object representation has the value `nullptr`; if the object representation is different, the behavior is undefined.<sup>303)</sup>

In C++ it's only required that `sizeof(std::nullptr_t) == sizeof(void*)`. Do we want to specify that `alignof(std::nullptr_t) == alignof(void*)`? (The superficial wording difference from C doesn't matter due to [[basic.compound] p7](https://eel.is/c++draft/basic.compound#7).)

The UB specified in that paragraph isn't present in C++. In C++, conversions from `nullptr_t` effectively touch no bit, so the value representation of a `nullptr_t` object, which is currently unspecified in C++, doesn't matter in most uses. However, the decision on value representation of `nullptr_t` can lead to different results via `bit_cast`, and there're implementation divergence (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117727).

Perhaps it's better to specifiy that the value representation (and the valid set of value bits) of `nullptr_t` is implementation-defined.

Suggested resolution:

Modify [[basic.fundamental] p16](https://eel.is/c++draft/basic.fundamental#16) as indicated:
> [...] `sizeof(std​::​nullptr_t)` shall be equal to `sizeof(void*)` <ins>**and `alignof(std::nullptr_t)` shall be equal to `alignof(void*)`**</ins>. <ins>**The value representation of `std::nullptr_t` and the non-empty set of value representations of `std::nullptr_t` that correspond to values are implementation-defined. [_Note ?_: An implementation can decide that the value representation of `std::nullptr_t` contains no bits. — _end note_]**</ins>

### LINKED ISSUE #656: [stmt.dcl] p2 Destruction of variables with automatic storage duration on coroutine exit
**Full name of submitter:** Artem Koton

**Reference (section label):** [stmt.dcl] p2

### Issue description:

[P1787](https://wg21.link/P1787) changed the wording of [[dcl.fct.def.coroutine/11]](https://wg21.link/dcl.fct.def.coroutine#11) regarding the destruction of variables with automatic storage duration from:

> In the latter case objects with automatic storage duration that are in scope at the suspend point are destroyed in the reverse order of the construction.

to:

> In the latter case, control in the coroutine is considered to be transferred _out of the function_ ([[stmt.dcl/2]](https://wg21.link/stmt.dcl#2)).

According to [[stmt.dcl/2]](https://wg21.link/stmt.dcl#2):

> Upon each transfer of control (including sequential execution of statements) _within a function_ from point P to point Q, all block variables with automatic storage duration that are active at P and not at Q are destroyed in the reverse order of their construction.

This only applies to transferring control "within a function", which does not cover the case for coroutines. As a result, there currently appears to be no wording specifying how local variables are destroyed when a coroutine is exited.

### Suggested resolution:

Clarify the wording of [[stmt.dcl/2]](https://wg21.link/stmt.dcl#2) to cover transfers of control out of a function.

### LINKED ISSUE #673: [basis.pre] Are labels names?
[[basic.pre]/3](https://eel.is/c++draft/basic.pre#3):
> A [name](https://eel.is/c++draft/basic.link#def:name) is an [identifier](https://eel.is/c++draft/lex.name#nt:identifier) ([[lex.name]](https://eel.is/c++draft/lex.name)), [conversion-function-id](https://eel.is/c++draft/class.conv.fct#nt:conversion-function-id) ([[class.conv.fct]](https://eel.is/c++draft/class.conv.fct)), [operator-function-id](https://eel.is/c++draft/over.oper.general#nt:operator-function-id) ([[over.oper]](https://eel.is/c++draft/over.oper)), or [literal-operator-id](https://eel.is/c++draft/over.literal#nt:literal-operator-id) ([[over.literal]](https://eel.is/c++draft/over.literal))[.](https://eel.is/c++draft/basic.pre#3.sentence-1)

I believe the definition for name includes `identifier` within `label` ([[stmt.label]](https://eel.is/c++draft/stmt.label#1)):
> [label:](https://eel.is/c++draft/stmt.label#nt:label)
> [attribute-specifier-seq](https://eel.is/c++draft/dcl.attr.grammar#nt:attribute-specifier-seq)opt [identifier](https://eel.is/c++draft/lex.name#nt:identifier) :

However, [[basic.pre]/5](https://eel.is/c++draft/basic.pre#5) states that:
> Every name is introduced by a declaration <...>

Which is at odds with the fact that labels fit into the definition of `name`, but are statements, and not declarations.

[P1787R6](https://wg21.link/p1787r6) "Declarations and where to find them" made major changes to the definition of name with the following intent:
> Because labels interact with so little of the language, their identifiers are excluded from names and name lookup entirely.

It seems that the definition of name should be updated to make it clear that labels are not included.

Suggested resolution (2025-02-20):
---------------------------------------
Convert [basic.pre] paragraph 3 into a bulleted list, and apply the following changes:

> A [name](https://eel.is/c++draft/basic.link#def:name) is
> - an [identifier](https://eel.is/c++draft/lex.name#nt:identifier) ([[lex.name]](https://eel.is/c++draft/lex.name)),<ins> except for _**identifier**_ of a _**label**_ ([stmt.label]),</ins>
> - <ins>a</ins> [conversion-function-id](https://eel.is/c++draft/class.conv.fct#nt:conversion-function-id) ([[class.conv.fct]](https://eel.is/c++draft/class.conv.fct)), 
> - <ins>an</ins> [operator-function-id](https://eel.is/c++draft/over.oper.general#nt:operator-function-id) ([[over.oper]](https://eel.is/c++draft/over.oper)), or 
> - <ins>a</ins> [literal-operator-id](https://eel.is/c++draft/over.literal#nt:literal-operator-id) ([[over.literal]](https://eel.is/c++draft/over.literal))[.](https://eel.is/c++draft/basic.pre#3.sentence-1)

### LINKED ISSUE #687: CWG3029 [basic.align] Note about weakest alignment for narrow char types is confusing
**Full name of submitter:** Xavier Bonaventura 

Reference: [basic.aling/6](https://eel.is/c++draft/basic.memobj#basic.align-6)

**Link to reflector thread:** https://chat.isocpp.org/general/pl/yjx3qdhrajfoxbifprpf8p16we (no reflector thread, just mattermost discussion)

**Issue description:**
https://eel.is/c++draft/basic.memobj#intro.object-3 says

> If a complete object is created ([[expr.new]](https://eel.is/c++draft/expr.new)) in storage associated with another object e of type “array of N unsigned char” or of type “array of N std​::​byte” ([[cstddef.syn]](https://eel.is/c++draft/cstddef.syn)), that array [provides storage](https://eel.is/c++draft/basic.memobj#def:provides_storage) for the created object if

However, in https://eel.is/c++draft/basic.memobj#basic.align-6 the note says

> [Note [2](https://eel.is/c++draft/basic.memobj#basic.align-note-2): This enables the ordinary character types to be used as the underlying type for an aligned memory area ([[dcl.align]](https://eel.is/c++draft/dcl.align))[.](https://eel.is/c++draft/basic.memobj#basic.align-6.sentence-3) — end note]

Implying that any narrow character type could be use to provide storage when only "unsigned char" or "std::byte" is allowed.

**Suggested resolution:**

Remove Note 2 from https://eel.is/c++draft/basic.memobj#basic.align-6

### LINKED ISSUE #706: CWG3035 [class.union.anon] Are lambda expressions disallowed in anonymous unions?

Reference (section label): class.union.anon

Issue description:
https://eel.is/c++draft/class.union.anon#1.sentence-3 says:
> Nested types, anonymous unions, and functions shall not be declared within an anonymous union.

A lambda expression declares a closure type (https://eel.is/c++draft/expr.prim.lambda#closure-2).

Does that mean that lambda expressions are disallowed in anonymous unions?

Only EDG generates a diagnostic for https://godbolt.org/z/xf96EbKTn:
```
static union {
  int x = [] { return 42; }();
};
```

Suggested resolution:
> Nested types,<ins> including closure types and</ins> anonymous unions, and functions shall not be declared within an anonymous union.

### LINKED ISSUE #749: CWG3058 "Program point" is not defined
**Full name of submitter**: Alisdair Meredith

**Link to reflector thread**: https://lists.isocpp.org/core/2025/08/18396.php

**Issue description**: See reflector thread. Alisdair may decide to submit a NB comment for this, so I'm posting it here to ensure that there's an easily accessible permalink.

### LINKED ISSUE #8823: [2026-03 CWG Motion 2] P4160R0 DR issues
except issues 3088, 3119, 3122, 3123, 3124, 3131, 3135, 3140, 3141, 3143, 3145, 3149, 3162, and 3172 in [P4160R0](https://wiki.isocpp.org/uploads/2026-03_Croydon-StrawPolls_-_p4160r0.html)

burblebee:
Fixes #8823.
Fixes cplusplus/papers#2732

Also fixes #5560

Also fixes cplusplus/CWG#89
Also fixes cplusplus/CWG#338
Also fixes cplusplus/CWG#545
Also fixes cplusplus/CWG#644
Also fixes cplusplus/CWG#656
Also fixes cplusplus/CWG#148
Also fixes cplusplus/CWG#673
Also fixes cplusplus/CWG#687
Also fixes cplusplus/CWG#706
Also fixes cplusplus/CWG#749
Also fixes cplusplus/CWG#813
Also fixes cplusplus/CWG#812
Also fixes cplusplus/CWG#801
Also fixes cplusplus/CWG#802
Also fixes cplusplus/CWG#800
Also fixes cplusplus/CWG#821
Also fixes cplusplus/CWG#826
Also fixes cplusplus/CWG#832
Also fixes cplusplus/CWG#837
Also fixes cplusplus/CWG#841
Also fixes cplusplus/CWG#838

Also fixes cplusplus/papers#2684

Also fixes cplusplus/nbballot#606
Also fixes cplusplus/nbballot#605
Also fixes cplusplus/nbballot#604

Notes:
* CWG2660: Not applied (already fixed by A. Jiang in 35904b92e).
* CWG2765: Targeted text in [expr.const] moved to [expr.const.core].
* CWG2983: The 2nd sentence of [basic.pre]p7 is no longer present.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3851,8 +3851,8 @@ The alignment requirement of a complete type can be queried using an
 the narrow character types\iref{basic.fundamental} shall have the weakest
 alignment requirement.
 \begin{note}
-This enables the ordinary character types to be used as the
-underlying type for an aligned memory area\iref{dcl.align}.
+The type \tcode{\keyword{unsigned} \keyword{char}} can be used as
+the element type of an array providing aligned storage\iref{dcl.align}.
 \end{note}
 
 \pnum
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -5986,6 +5986,28 @@ alignment requirement.
 \end{note}
 
 \pnum
+A pointer value
+pointing to a potentially non-unique object $O$\iref{intro.object} is
+\indextext{value!associated with an evaluation}%
+\defn{associated with} the evaluation of
+\begin{itemize}
+\item
+the string-literal\iref{lex.string} that resulted in the string literal object,
+\item
+the initializer list\iref{dcl.init.list} that resulted in the backing array,
+or
+\item
+the initialization of
+the template parameter object\iref{temp.arg.nontype, meta.define.static}
+\end{itemize}
+that is $O$ or of which $O$ is a subobject.
+\begin{note}
+A pointer value obtained by pointer arithmetic\iref{expr.add}
+from a pointer value associated with an evaluation $E$
+is also associated with $E$.
+\end{note}
+
+\pnum
 A pointer value $P$ is
 \indextext{value!valid in the context of an evaluation}%
 \defn{valid in the context of} an evaluation $E$
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -5666,7 +5666,12 @@ The types denoted by \cv~\tcode{std::nullptr_t} are distinct types.
 A prvalue of type \tcode{std::nullptr_t} is a null pointer
 constant\iref{conv.ptr}. Such values participate in the pointer and the
 pointer-to-member conversions\iref{conv.ptr,conv.mem}.
-\tcode{\keyword{sizeof}(std::nullptr_t)} shall be equal to \tcode{\keyword{sizeof}(\keyword{void}*)}.
+The size\iref{expr.sizeof} and alignment requirement\iref{basic.align} of
+the type \tcode{std::nullptr_t} are those of
+the type ``pointer to \keyword{void}''.
+\begin{note}
+The value representation can comprise no bits\iref{conv.lval}.
+\end{note}
 
 \pnum
 A value of type \tcode{std::meta::info} is called a \defn{reflection}.
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -1678,6 +1678,11 @@ are the semantic properties introduced by the declarations
 used in further processing.
 
 \pnum
+There is a \defnadj{program}{point}
+before the first token of the translation unit,
+at least one between every pair of adjacent tokens, and
+at least one after the last token of the translation unit.
+
 A program point $P$ is said to follow
 any declaration in the same translation unit
 whose locus\iref{basic.scope.pdecl} is before $P$.
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3675,12 +3675,6 @@ or
 \end{itemize}
 otherwise, they have distinct addresses
 and occupy disjoint bytes of storage.
-\begin{footnote}
-Under the ``as-if'' rule an
-implementation is allowed to store two objects at the same machine address or
-not store an object at all if the program cannot observe the
-difference\iref{intro.execution}.
-\end{footnote}
 \begin{example}
 \begin{codeblock}
 static const char test1 = 'x';
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -115,13 +115,15 @@ before continuing to parse the program that contains it.
 
 \pnum
 A \defn{variable} is introduced by the
-declaration of
+declaration $D$ of
 \begin{itemize}
 \item
 a reference other than a non-static data member or
 \item
-an object.
+an object,
 \end{itemize}
+where $D$ is not the \grammarterm{parameter-declaration} of
+a \grammarterm{template-parameter}.
 
 \pnum
 An \defn{entity} is a
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3079,9 +3079,7 @@ they appear in the same translation unit, or
 \item
 they both declare type aliases or namespace aliases that have the same underlying entity, or
 \item
-they both declare names with module linkage and are attached to the same module, or
-\item
-they both declare names with external linkage.
+they both declare names with module or external linkage and are attached to the same module.
 \end{itemize}
 \begin{note}
 There are other circumstances in which declarations declare
@@ -3098,9 +3096,11 @@ Such an $H$ can appear only in a header unit.
 \end{note}
 
 \pnum
-If two declarations of an entity are
-attached to different modules, the program is ill-formed;
-no diagnostic is required if neither is reachable from the other.
+\begin{note}
+If two declarations correspond but are
+attached to different modules, the program is ill-formed
+if one precedes the other\iref{basic.scope.scope}.
+\end{note}
 \begin{example}
 \begin{codeblocktu}{\tcode{"decls.h"}}
 int f();            // \#1, attached to the global module
@@ -3112,15 +3112,15 @@ module;
 #include "decls.h"
 export module M;
 export using ::f;   // OK, does not declare an entity, exports \#1
-int g();            // error: matches \#2, but attached to \tcode{M}
+int g();            // error: corresponds to \#2, but attached to \tcode{M}
 export int h();     // \#3
 export int k();     // \#4
 \end{codeblocktu}
 
 \begin{codeblocktu}{Other translation unit}
 import M;
-static int h();     // error: matches \#3
-int k();            // error: matches \#4
+static int h();     // error: conflicts with \#3
+int k();            // error: conflicts with \#4
 \end{codeblocktu}
 \end{example}
 As a consequence of these rules,
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3640,7 +3640,13 @@ that are not bit-fields
 may have the same address if
 \begin{itemize}
 \item one is nested within the other,
-\item at least one is a subobject of zero size and they are not of similar types\iref{conv.qual},
+\item
+they are both nested within some complete object $o$,
+exactly one is a subobject of $o$, and the subobject is of zero size,
+\item
+they are both subobjects of the same complete object,
+at least one is a subobject of zero size, and
+they are not of similar types\iref{conv.qual},
 or
 \item they are both potentially non-unique objects;
 \end{itemize}
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -32,10 +32,26 @@ Clauses.
 \end{note}
 
 \pnum
-A \defn{name} is an \grammarterm{identifier}\iref{lex.name},
-\grammarterm{conversion-function-id}\iref{class.conv.fct},
-\grammarterm{operator-function-id}\iref{over.oper}, or
-\grammarterm{literal-operator-id}\iref{over.literal}.
+A \defn{name} is
+\begin{itemize}
+\item an \grammarterm{identifier} token\iref{lex.token, lex.name} other than
+  \begin{itemize}
+  \item
+  the \grammarterm{identifier} of a
+  \grammarterm{label}\iref{stmt.label} or
+  \grammarterm{literal-operator-id}\iref{over.literal},
+  \item
+  the \grammarterm{identifier} following a \keyword{goto} in a
+  \grammarterm{jump-statement}\iref{stmt.jump.general},
+  \item
+  any \grammarterm{identifier} in a
+  \grammarterm{module-name}\iref{module.unit} or
+  \grammarterm{attribute-token}\iref{dcl.attr.grammar}, or
+  \end{itemize}
+\item a \grammarterm{conversion-function-id}\iref{class.conv.fct},
+\item an \grammarterm{operator-function-id}\iref{over.oper}, or
+\item a \grammarterm{literal-operator-id}\iref{over.literal}.
+\end{itemize}
 
 \pnum
 Two names are \defnx{the same}{name!same} if
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -462,6 +462,11 @@ sorts the pairs in increasing order of their \tcode{first} members:
 \end{codeblock}
 \end{example}
 
+\indexdefn{property!of the implementation}%
+\definition{property of the implementation}{defns.impl.prop}
+behavior, for a well-formed program\iref{defns.well.formed}
+construct and correct data, that depends on the implementation
+
 \definition{referenceable type}{defns.referenceable}
 \indexdefn{type!referenceable}%
 type that is either an
@@ -920,10 +925,14 @@ observable behavior of the program are produced.
 
 \pnum
 \indextext{behavior!implementation-defined}%
-Certain aspects and operations of the abstract machine are described in this
+Certain aspects and operations of the abstract machine
+constitute the parameters of the abstract machine and
+are described in this
 document as implementation-defined behavior (for example,
-\tcode{sizeof(int)}). These constitute the parameters of the abstract machine.
-Each implementation shall include documentation describing its characteristics
+\tcode{sizeof(int)})
+or as properties of the implementation (for example, padding in class types).
+For implementation-defined behavior,
+each implementation shall include documentation describing its characteristics
 and behavior in these respects.
 \begin{footnote}
 This documentation also includes
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -580,6 +580,10 @@ and each such \grammarterm{member-declaration}
 shall either
 declare at least one member name of the class
 or declare at least one unnamed bit-field.
+A \defnx{user-declared}{entity!user-declared}
+\indextext{user-declared entity|see{entity, user-declared}}
+entity is a direct member or a friend that, in either case,
+is declared by a \grammarterm{member-declaration}.
 
 \pnum
 A \defn{data member} is either a non-function member introduced by a
@@ -1273,8 +1277,9 @@ that is not a function parameter pack
 has a default argument
 (including the case of a constructor with no parameters).
 \indextext{implicitly-declared default constructor}%
-If there is no user-declared constructor or constructor template for class
-\tcode{X} and \tcode{X} is not an anonymous union,
+If a class does not have
+a user-declared constructor or constructor template,
+and the class is not an anonymous union,
 a non-explicit constructor having no parameters is implicitly declared
 as defaulted\iref{dcl.fct.def}.
 An implicitly-declared default constructor is an
@@ -1503,7 +1508,7 @@ void h() {
 \end{example}
 
 \pnum
-If the class definition does not explicitly declare a copy constructor
+If the class does not have a user-declared copy constructor
 and the class is not an anonymous union,
 a non-explicit one is declared \defnx{implicitly}{constructor!copy!implicitly declared}.
 If the class definition declares a move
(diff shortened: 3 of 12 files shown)


tkoeppe:
Thanks a lot, @burblebee! Is this ready? (The PR is still in "draft" mode.)

source/intro.tex
```diff
@@ -339,6 +339,11 @@
 \definition{implementation limit}{defns.impl.limits}
 restriction imposed upon programs by the implementation
 
+\indexdefn{prop!implementation}%
```
tkoeppe: The top-level index entry should be something like "property", not just "prop"?

And the entire entry should be in its alphabetical place.

source/basic.tex
```diff
@@ -3640,7 +3640,12 @@
 may have the same address if
 \begin{itemize}
 \item one is nested within the other,
-\item at least one is a subobject of zero size and they are not of similar types\iref{conv.qual},
+\item
+they are both nested within some complete object $o$,
+exactly one is a subobject of $o$, and the subobject is of zero size,
+\item
+they are both subobjects of the same complete object,
+at least one is a subobject of zero size and they are not of similar types\iref{conv.qual},
```
tkoeppe: Comma missing after "zero size"

source/basic.tex
```diff
@@ -5984,6 +5984,28 @@
...
+\item
+the initialization of
+the template parameter object\iref{temp.arg.nontype, meta.define.static}
+\end{itemize}
+that is $O$ or of which $O$ is a subobject.
+\begin{note}
+A pointer value obtained by pointer arithmetic\iref{expr.add}
+from a pointer value associated with an evaluation $E$
+is also associated with $E$.
+\end{note}
```
tkoeppe, line 6030: @jensmaurer Not editorial, but how come this is just a note, when it appears to be stating something normative that isn't specified normatively?

source/intro.tex
burblebee: doh! fixed, thanks

source/basic.tex
jensmaurer: If you apply pointer arithmetic, you navigate from one subobject to another.  But all such subobjects are of the same complete object, and thus are associated with the same evaluation per the definition of "associated with evaluation" just above.

source/basic.tex
burblebee: fixed

jensmaurer:
> [CWG2660](https://cplusplus.github.io/CWG/issues/2660.html): Not applied (text rewritten by [CWG2813](https://cplusplus.github.io/CWG/issues/2813.html)).

While this is superficially true, the offending phase "this parameter of the function (7.5.3 [[expr.prim.this](https://wg21.link/expr.prim.this)]) is initialized with a pointer to the object of the call" still exists post-CWG2813.  Please apply CWG2660 to that part.

jensmaurer:
I have added "fixes cplusplus/nbballot" links in the initial comment of this pull request.  It would be good to amend the corresponding commits with our usual "Fixes NB ... (C++26 CD)." text (examples all over "git log").

source/basic.tex
tkoeppe: Ah yes, thanks!

tkoeppe:
There are 32 issues in Motion 2. Issue 3127 makes no changess.

@burblebee What about CWG2660?

burblebee:
> > [CWG2660](https://cplusplus.github.io/CWG/issues/2660.html): Not applied (text rewritten by [CWG2813](https://cplusplus.github.io/CWG/issues/2813.html)).
> 
> While this is superficially true, the offending phase "this parameter of the function (7.5.3 [[expr.prim.this](https://wg21.link/expr.prim.this)]) is initialized with a pointer to the object of the call" still exists post-[CWG2813](https://cplusplus.github.io/CWG/issues/2813.html). Please apply [CWG2660](https://cplusplus.github.io/CWG/issues/2660.html) to that part.

Sorry, I referenced the wrong commit - the "this" pointer was fixed in by A. Jiang in 35904b92e. I'll fix my note.

tkoeppe:
Ah, great, thanks!

burblebee:
> I have added "fixes cplusplus/nbballot" links in the initial comment of this pull request. It would be good to amend the corresponding commits with our usual "Fixes NB ... (C++26 CD)." text (examples all over "git log").

Sorry, I totally missed those! The way they're formatted is too subtle; they look like part of the date. I seem to remember them being a lot more obvious. 

I see  - we now have a link where we used to have big, bold, black letters, right? The link is nice - thanks for doing that.

Can we do something to make the NB issues more obvious in the future? Maybe add a new, separate field for them, further away from the date? And, if possible, use a big bold font for the text of the link?

jensmaurer:
> Sorry, I totally missed those! The way they're formatted is too subtle; they look like part of the date. I seem to remember them being a lot more obvious.

We used to have links in former times, too; I've used the existing XML machinery in the issues list (the deep links to the C++26 NB comments don't actually work). Compare these examples:

- C++26: https://cplusplus.github.io/CWG/issues/3171.html
- C++23: https://cplusplus.github.io/CWG/issues/2659.html
- C++14: https://cplusplus.github.io/CWG/issues/1786.html
(maybe we didn't create core issues for NB comments in between, and instead used separate omnibus papers)

burblebee:
> > Sorry, I totally missed those! The way they're formatted is too subtle; they look like part of the date. I seem to remember them being a lot more obvious.
> 
> We used to have links in former times, too; I've used the existing XML machinery in the issues list (the links to the C++26 don't actually work). Compare these examples:
> 
> * C++26: https://cplusplus.github.io/CWG/issues/3171.html
> * C++23: https://cplusplus.github.io/CWG/issues/2659.html
> * C++14: https://cplusplus.github.io/CWG/issues/1786.html
>   (maybe we didn't create core issues for NB comments in between, and instead used separate omnibus papers)

Thanks for those references, I can see what made the NB issues more obvious - they were in the 1st column and separated from the usual fields by a newline.

jensmaurer:
> I can see what made the NB issues more obvious - they were in the 1st column and separated from the usual fields by a newline.

Again, I fail to see a material visual difference between "this round" and "last round":

image

---

image

tkoeppe:
Thanks Dawn and Jens for sorting out the NB comment associations!

burblebee:
> > I can see what made the NB issues more obvious - they were in the 1st column and separated from the usual fields by a newline.
> 
> Again, I fail to see a material visual difference between "this round" and "last round":

The browser I'm using (Safari) shows something very different for "this round" - the NB comments appear directly left of the Date: entry. Whereas I do see what you see for "last round".

I'd embed an image of what I see but I don't know how you did that (neat!) - hopefully you can see what I mean from the following:
```
3058. "Program point" is not defined

Section: 6.5.1  [[basic.lookup.general](https://wg21.link/basic.lookup.general)]     Status: ready     Submitter: Alisdair Meredith     Date: 2025-08-13 [N5028 comment US 14-029](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/n5028.pdf#US14-029)
(From submission [#749](https://github.com/cplusplus/CWG/issues/749).)

There is no definition for the term "program point".
```

For now, did I find them all? I only saw 2 issues with NB comments in this motion. 

FYI - I tried in FireFox and Brave (a Chrome-like browser) and don't see those extra newlines that you see, so I guess it's a Mac issue?? Interesting... Just looked this up and apparently a **known bug**. Now that I know I'll try to remember to look in the future. Sorry to bother you!!!

PR #8924, opened 2026-04-06
[Motions 2026 03 cwg 3] P4160R0 non-DR issues

### LINKED ISSUE #803: CWG3123 [stmt.expand] `begin(E)`/`end(E)` should perform overload resolution
Full name of submitter (unless configured in github; will be published with the issue): Ambrose T.

Reference (section label): [stmt.expand]

### Issue description:

[[stmt.expand]p3](https://eel.is/c++draft/stmt.expand#3.2) reads:
> An expression is _expansion-iterable_, if [...] 
> - argument-dependent lookups for `begin(E)` and for `end(E)` each find at least one function or function template.

This wording seems to imply that if _any_ function or function template with the name `begin`/`end` (at least one each) is found via ADL, then `E` is expansion-iterable. This seems much too broad as it doesn’t take into consideration whether `begin(E)` or `end(E)` would even be well-formed.

E.g. if `E` is of type `std::tuple<...>`, then ADL for `begin`/`end` in namespace `std` will *find* quite a few functions (or function templates), but none of them accept a parameter of type `std::tuple<...>`, and the intent is surely that expansion over a `std::tuple<...>` be destructuring rather than iterating.

### Suggested resolution:
Change the requirement from name lookup succeeding to overload resolution succeeding (notably, if overload resolution succeeds but there is no usable candidate, the program should be ill-formed).

### LINKED ISSUE #805: CWG3131 [stmt.expand] Expansion statements vs. define_static_array

### LINKED ISSUE #807: CWG3135 [dcl.struct.bind] Tuple-like decomposition with prvalues
Full name of submitter (unless configured in github; will be published with the issue): Barry Revzin

Reference (section label): dcl.struct.bind

Issue description:

In the tuple case of structured bindings, we say:

> In either case, e is an lvalue if the type of the entity e is an lvalue reference and an xvalue otherwise[.](https://eel.is/c++draft/dcl.struct.bind#7.sentence-7) Given the type `Ti` designated by `std​::​tuple_element<i, E>​::​type` and the type `Ui` designated by either `Ti&` or `Ti&&`, where `Ui` is an lvalue reference if the initializer is an lvalue and an rvalue reference otherwise, variables are introduced with unique names ri as follows:
> 
> ```
> S Ui ri = initializer ;
> ```

That is, we're always declaring either lvalue references or rvalue references — and then the initializer is either `e.get<i>()` or `get<i>(e)`. However, what if `get` returns a prvalue?

In Kona, we adopted [P1789](https://wiki.edg.com/pub/Wg21kona2025/StrawPolls/P1789R3.pdf), which adds structured bindings support to `std::integer_sequence`. Per the wording in the paper, this:
```cpp
constexpr auto [...Is] = std::make_index_sequence<2>();
```
decomposes into:
```cpp
constexpr auto e = std::integer_sequence<size_t, 0, 1>();
constexpr size_t&& Is#0 = 0;
constexpr size_t&& Is#1 = 1;
```

This means that the `Is...` are _mutable_, which is surprising, undesired, and violates [\[expr.const\]/9.5](https://eel.is/c++draft/expr.const#9.5) anyway. 

We _could_ fix this in the library wording — either by removing the `const integer_sequence` specialization or making the base specialization make the type `T const` instead of` T`. Either the way, the result there would be that the `Is...` would all be `constexpr size_t const&&` instead, which would work fine. 

But it would probably be less surprising to have the prvalue case in structured bindings simply not be references at all, so that the `constexpr` would properly propagate `const`-ness. 

Suggested resolution:

Something like this:

> Given the type `Ti` designated by `std​::​tuple_element<i, E>​::​type` and the type `Ui` designated by <del>either `Ti&` or `Ti&&`, where `Ui` is an lvalue reference if the initializer is an lvalue and an rvalue reference otherwise</del> <ins>`Ti` if the initializer is a prvalue, `Ti&` if the initializer is an lvalue, and `Ti&&` otherwise</ins>, variables are introduced with unique names ri as follows:

### LINKED ISSUE #808: CWG3122 [temp.dep.constexpr] Inadequate value-dependence for reflect-expressions
**Full name of submitter**: Dan Katz

**Reference**: [temp.dep.constexpr]

**Issue description**
[temp.dep.constexpr]/7.1.4 says that `^^operand` is value-dependent when `operand` names a dependent member of the current instantiation, e.g.,

```cpp
template <typename T>
struct TCls {
  static T mem;
  static constexpr info R = type_of(^^mem);  // '^^mem' is value-dependent.
};
```

This is good, but it doesn't go far enough: If the enclosing template is instead a function template, then the existing conditions for value-dependence fail to apply.

```cpp
template <typename T>
consteval void f() {
  T v;
  info R = type_of(^^v);
    // not currently value-dependent; really needs to be.
}
```

Annotations are also affected:

```cpp
template <int K>
int f() {
  [[=K]] int v;
  return annotations_of(^^v).size();
    // Also not value-dependent, but really ought to be (even though 'v' is neither type nor value dependent).
}
```

Applying `parent_of` in such a situation likewise proves a problem. Some of these circumstances would demand that the compiler surface a reflection that represents a dependent type; others become unnecessarily wide IFNDR traps. 

**Suggested resolution**
The reflection of any templated entity (and more: any entity with a declaration in a templated context enclosing the _`reflect-expression`_) should be value-dependent.

Modify [temp.dep.constexpr]/7 as follows:
> A _`reflect-expression`_ is value-dependent if
> - it is of the form `^^` _`reflection-name`_ and **<ins>either</ins>** the _`reflection-name`_ **<ins>is</ins>**
>   - ~~is~~ a dependent qualified name,
>   - ~~is~~ a dependent _`namespace-name`_, **<ins>or</ins>**
>   - ~~is~~ the name of a template parameter ~~, or~~
>   - ~~names a dependent member of the current instantiation ([temp.dep.type]),~~
> 
>   **<ins>or lookup for the _`reflection-name`_ finds a declaration that inhabits a scope corresponding to a templated entity,</ins>**
> - it is of the form `^^` _`type-id`_ and the _`type-id`_ denotes a dependent type, or
> - it is of the form `^^` _`id-expression`_ and the _`id-expression`_ is value-dependent.

Note that the value-dependent _`id-expression`_ bullet remains necessary to handle the case when the operand is a _`template-id`_ (which is not a _`reflection-name`_), e.g.,

```cpp
template <typename T>
static constexpr T zero = 0;

template <typename T>
unsigned f() {
  return size_of(^^zero<T>);
    // 'zero<T>' is a value-dependent id-expression that is not a reflect-expression.
}
```

### LINKED ISSUE #809: CWG3124 [dcl.attr.annotation] Disallow annotations on "remote" declarations
**Full name of submitter**: Dan Katz

**Reference**: [dcl.attr.annotation]

**Issue description**
Other than new and novel forms of friend injection, there is currently very little motivation for allowing annotations on declarations of block-scope function and extern variable declarations, or on friend declarations. Consider:

```cpp
template <int>
class Counter {
  struct S;
public:
  template <auto = []{}>
  static consteval size_t value() {
    void f(S);

    constexpr size_t result = annotations_of(^^f).size();
    return result;
  }

  template <auto = []{}>
  static consteval size_t incr() {
    [[=1]] void f(S);

    constexpr size_t result = annotations_of(^^f).size();
    return result;
  }
};

static_assert(Counter<0>::value() == 0);
static_assert(Counter<0>::incr() == 1);
static_assert(Counter<0>::incr() == 2);
static_assert(Counter<1>::incr() == 1);
static_assert(Counter<0>::incr() == 3);
static_assert(Counter<0>::value() == 3);
static_assert(Counter<1>::value() == 1);
```

(note: the above is currently IFNDR, but will be well-formed if the solution proposed for #808 is accepted; these issues are therefore related)

We should just do the conservative thing and disallow such annotations for C++26; ~nobody will miss them.

**Suggested resolution**
Modify [dcl.attr.annotation]/1 as follows:

> An annotation may be applied to ~~any~~ **<ins>a</ins>** declaration of a type, type alias, variable, function, namespace, enumerator, _`base-specifier`_, or non-static data member. **<ins>No annotation shall apply to a declaration _D_, nor to a function parameter declaration in the _`function-declarator`_ (if any) of _D_, if either</ins>**
> - **<ins>_D_ is a friend declaration ([class.friend]) or</ins>**
> - **<ins>the host scope of _D_ ([basic.scope.scope]) is not the target scope of _D_.</ins>**

### LINKED ISSUE #810: CWG3127 [expr.const] Evaluation context can be missing synthesized points
**Full name of submitter**: Dan Katz

**Reference**: [expr.const]

**Issue description**
The evaluation context ([expr.const]) determines which declarations are found by metafunctions ([meta.reflection]). That context consists of ordinary program points (typically, the point at which a manifestly constant-evaluated expression appears), together with any "synthesized points" that correspond to injected declarations (which in C++26, following the removal of `std::meta::annotate` from earlier revisions of P3394, are only produced by `define_aggregate`).

Because the evaluation of a metafunction can trigger the instantiation of a specialization, the notions of "evaluation context" and "instantiation context" are intertwined. For instance, given:

```cpp
template <typename T>
struct TCls {
  T mem;
};

struct S;
consteval {
  constexpr auto ctx = std::meta::access_context::current();

  if (nonstatic_data_members_of(substitute(^^TCls, {define_aggregate(^^S, {})}), ctx).size() > 1)
    throw "contrived, but whatever";
}
```

- The evaluation context of the expression corresponding to the consteval block initially consists of only the point following the consteval block.
- During the evaluation of that expression, a synthesized point corresponding to the definition of `S` joins the evaluation context following the evaluation of `define_aggregate(^^S)`.
- The call to `nonstatic_data_members_of` then triggers the instantiation of `TCls<S>`.
- Importantly, during the instantiation of `TCls<S>`, [[module.reach]/6](https://eel.is/c++draft/module.context#6.sentence-1) guarantees that the synthesized point associated with the definition of `S` _is_ a part of the instantiation context; this guarantees that `S` is a complete type during this instantiation:

> During the implicit instantiation of any construct that resulted from the evaluation of an expression as a core constant expression, the instantiation context contains each point in the evaluation context ([[expr.const]](https://eel.is/c++draft/expr.const)).

However, consider now the following:

```cpp
template <typename T>
struct TCls {
  T mem;

  static_assert(size_of(^^T) > 0);  // <-- problem
};

struct S;
consteval {
  constexpr auto ctx = std::meta::access_context::current();

  if (nonstatic_data_members_of(substitute(^^TCls, {define_aggregate(^^S, {})}), ctx).size() > 1)
    throw "contrived, but whatever";
}
```

The operand of the `static_assert` in `TCls<S>` is a distinct manifestly constant-evaluated expression, and nothing in the definition of "evaluation context" in [expr.const]/32 succeeds in "pushing" the synthesized point associated with `S` from the instantiation context of `TCls<S>` into the evaluation context of the static assertion - without which, `size_of(^^T)` will throw.

**Suggested resolution**
Modify [expr.const]/32 as follows:

> The _evaluation context_ is a set of program points that determines the behavior of certain functions used for reflection ([meta.reflection]). During the evaluation _V_ of an expression _E_ as a core constant expression, the evaluation context of an evaluation _X_ ([intro.execution]) consists of the following points:
> - The program point _EVAL-PT(L)_, where _L_ is the point at which _E_ appears, and where _EVAL-PT(P)_, for a point _P_, is a point _R_ determined as follows:
>   - If a potentially-evaluated subexpression ([intro.execution]) of a default member initializer _I_ appears at _P_, and a (possibly aggregate) initialization during _V_ is using _I_, then _R_ is _EVAL-PT(Q)_ where _Q_ is the point at which that initialization appears.
>   - Otherwise, if a potentially-evaluated subexpression of a default argument ([dcl.fct.default]) appears at _P_, and an invocation of a function ([expr.call]) during _V_ is using that default argument, then _R_ is _EVAL-PT(Q)_ where _Q_ is the point at which that invocation appears.
>   - Otherwise, _R_ is _P_.
> - Each synthesized point corresponding to an injected declaration produced by any evaluation sequenced before _X_ ([intro.execution]).
> - **<ins>Any synthesized point in the instantiation context of _EVAL-PT(L)_.</ins>**

### LINKED ISSUE #825: CWG3145 [dcl.attr.annotation] Which annotations are unique?

### LINKED ISSUE #828: CWG3143 [diff.dcl] Annex C statement about type of enumerators incorrect after rebase to C23

Reference (section label): [diff.dcl]

## Issue description:
https://wg21.link/diff.dcl#12 states that the type of enumerators in C is `int`. This is not true in C23 since the adoption of https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3029.htm and https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3030.htm.

## Suggested resolution:
Replace "is `int`" with "can be `int`".

### LINKED ISSUE #835: CWG3149 [stmt.expand] Not handling rvalues in destructuring expansion statement
Full name of submitter (unless configured in github; will be published with the issue): Liu Yanzuo

Reference (section label):[stmt.expand]

Link to reflector thread (if any): https://lists.isocpp.org/std-discussion/2026/01/3275.php

Issue description:

Right now, the wording for a destructuring expansion statement says that we create a possibly-constexpr structured binding declaration
```cpp
auto&& [u_0, u_1, ..., u_N-1] = expansion_initializer;
```
And then each statement of the expansion statement is:
```cpp
for_range_declaration = u_i;
```

As worded, `u_i` is always an lvalue. Which means that in an example like this:

```cpp
auto f() -> std::tuple<int, int&, int&&>;

auto g() -> void {
  template for (auto&& elem : f()) {
    ;
  }
}
```

`decltype(elem)` is `int&` on each iteration. But the [intent](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p1306r5.html#expansion-over-tuples) was that we would get `int&&` on the first and third iteration. This was never quite translated properly to the wording. 

Suggested resolution: Something to the effect of:

<code><i>for-range-declaration</i> = <del><code>u<sub>i</sub></code></del> <ins>get-expr<sub>i</i>;</code>

Where <code><i>get-expr</i><sub>i</sub></code> is <code>u<sub>i</sub></code> if either <code>decltype(u<sub>i</sub>)</code> is an lvalue reference type or <code><i>expansion-initializer</i></code> is an lvalue. Otherwise, <code>std::move(u<sub>i</sub>)</code>.

### LINKED ISSUE #2615: CWG3123  Global lookup for begin and end for expansion statements P4026

### LINKED ISSUE #8824: [2026-03 CWG Motion 3] P4160R0 non-DR issues
issues 3088, 3119, 3122, 3123, 3124, 3131, 3135, 3140, 3143, 3145, 3149, 3162, and 3172 in [P4160R0](https://wiki.isocpp.org/uploads/2026-03_Croydon-StrawPolls_-_p4160r0.html)

burblebee:
Fixes #8824.
Fixes cplusplus/papers#2732

Also fixes cplusplus/CWG#808
Also fixes cplusplus/CWG#803
Also fixes cplusplus/CWG#809
Also fixes cplusplus/CWG#805
Also fixes cplusplus/CWG#807
Also fixes cplusplus/CWG#828
Also fixes cplusplus/CWG#825
Also fixes cplusplus/CWG#835
Also fixes cplusplus/CWG#848
Also fixes cplusplus/CWG#810

Also fixes cplusplus/nbballot#683
Also fixes cplusplus/nbballot#666

Also fixes cplusplus/papers#2659
Also fixes cplusplus/papers#2615

Notes:
* CWG3162: Also fixes CWG3127.
* CWG3124: Replaced /function-declarator/ with function declarator.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -4977,7 +4977,7 @@ initialized from $E$.
 \pnum
 The sixth context is when a temporary object is created
 in the \grammarterm{expansion-initializer}
-of a destructuring expansion statement.
+of an iterating or destructuring expansion statement.
 If such a temporary object would otherwise be destroyed
 at the end of that \grammarterm{expansion-initializer},
 the object persists for the lifetime of the reference
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -222,6 +222,23 @@ Type deduction from \#1 does not succeed;
 type deductions from \#2 and \#3 both succeed.
 \end{example}
 
+\rSec2[diff.cpp23.cpp]{\ref{cpp}: preprocessing directives}
+
+\diffref{cpp.replace.general}
+\change
+Additional restrictions on macro names.
+\rationale
+Avoid hard to diagnose or non-portable constructs.
+\effect
+Keywords,
+names of identifiers with special meaning\iref{lex.name},
+and (unless otherwise specified) \grammarterm{attribute-token}{s}
+specified in \ref{dcl.attr}
+may not be used as macro names.
+For example, valid \CppXXIII{} code that
+defines \tcode{post} or \tcode{pre} as macros
+is invalid in this revision of \Cpp{}.
+
 \rSec2[diff.cpp23.library]{\ref{library}: library introduction}
 
 \diffref{headers}
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -3464,14 +3464,15 @@ Common.
 
 \diffref{dcl.enum}
 \change
-In \Cpp{}, the type of an enumerator is its enumeration. In C, the type of an enumerator is \keyword{int}.
+In \Cpp{}, the type of an enumerator is its enumeration. In C, the type of an enumerator is an integer type.
 
 \begin{example}
 \begin{codeblock}
 enum e { A };
-sizeof(A) == sizeof(int)        // in C
-sizeof(A) == sizeof(e)          // in \Cpp{}
-/* and @sizeof(int)@ is not necessarily equal to @sizeof(e)@ */
+void f() {
+  auto x = A;
+  int *p = &x;                  // valid C, invalid \Cpp{}
+}
 \end{codeblock}
 \end{example}
 
@@ -3483,10 +3484,6 @@ Change to semantics of well-defined feature.
 Semantic transformation.
 \howwide
 Seldom.
-The only time this affects existing C code is when the size of an
-enumerator is taken.
-Taking the size of an enumerator is not a
-common C coding practice.
 
 \diffref{dcl.align}
 \change
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -10256,11 +10256,8 @@ $E$ shall be a constant expression;
 the result of $E$ is the \defnadj{underlying}{constant} of the annotation.
 
 \pnum
-Each \grammarterm{annotation} produces a unique annotation.
-
-\pnum
-Substituting into an \grammarterm{annotation}
-is not in the immediate context.
+Each \grammarterm{annotation} or instantiation thereof
+produces a unique annotation.
 \begin{example}
 \begin{codeblock}
 [[=1]] void f();
@@ -10272,6 +10269,16 @@ and \tcode{g} has five annotations.
 These can be queried with metafunctions
 such as \tcode{std::\brk{}meta::\brk{}anno\-tations_of}\iref{meta.reflection.annotation}.
 \end{example}
+\begin{example}
+\begin{codeblock}
+template<int> int x [[=1]];
+static_assert(annotations_of(^^x<0>) != annotations_of(^^x<1>));    // OK
+\end{codeblock}
+\end{example}
+
+\pnum
+Substituting into an \grammarterm{annotation}
+is not in the immediate context.
 \begin{example}
 \begin{codeblock}
 template<class T>
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -9513,6 +9513,13 @@ lookup\iref{basic.lookup} is performed on any of the identifiers contained in an
 requirements on the \grammarterm{attribute-argument-clause} (if any).
 
 \pnum
+\begin{note}
+Unless otherwise specified,
+an \grammarterm{attribute-token} specified in this document cannot be used
+as a macro name\iref{cpp.replace.general}.
+\end{note}
+
+\pnum
 An \grammarterm{annotation} followed by an ellipsis
 is a pack expansion\iref{temp.variadic}.
 
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -7741,10 +7741,10 @@ In either case, \exposid{e} is an lvalue if the type of the entity \exposid{e}
 is an lvalue reference and an xvalue otherwise.
 Given the type $\tcode{T}_i$ designated by
 \tcode{std::tuple_element<i, E>::type} and
-the type $\tcode{U}_i$ designated by
-either \tcode{$\tcode{T}_i$\&} or \tcode{$\tcode{T}_i$\&\&},
-where $\tcode{U}_i$ is an lvalue reference if
-the initializer is an lvalue and an rvalue reference otherwise,
+the type $\tcode{U}_i$ defined
+as $\tcode{T}_i$ if the initializer is a prvalue,
+as ``lvalue reference to $\tcode{T}_i$'' if the initializer is an lvalue, or
+as ``rvalue reference to $\tcode{T}_i$'' otherwise,
 variables are introduced with unique names $\tcode{r}_i$ as follows:
 
 \begin{ncbnf}
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -10228,14 +10228,26 @@ if their respective types are all empty.
 \pnum
 An annotation may be applied
 to a \grammarterm{base-specifier} or
-to any declaration of a
+to a declaration $D$ of a
 type,
 type alias,
 variable,
 function,
 namespace,
 enumerator, or
-non-static data member.
+non-static data member,
+unless
+\begin{itemize}
+\item the host scope of $X$ differs from its target scope or
+\item $X$ is a non-defining friend declaration,
+\end{itemize}
+where $X$ is
+\begin{itemize}
+\item
+$D'$ if $D$ is a function parameter declaration in
+a function declarator\iref{dcl.fct} of a function declaration $D'$ and
+\item $D$ otherwise.
+\end{itemize}
 
 \pnum
 Let $E$ be the expression
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -9419,39 +9419,93 @@ consteval {                                 // \#1
 \end{example}
 
 \pnum
-The \defn{evaluation context} is a set of program points
-that determines the behavior of certain functions
-used for reflection\iref{meta.reflection}.
+During an evaluation $V$\iref{intro.execution} of
+an expression, conversion, or initialization $E$
+as a core constant expression, the
+\indextext{point of!evaluation}%
+\indextext{point!of evaluation}%
+\indextext{evaluation!point of evaluation, during}%
+\defnx{point of evaluation of $E$ during $V$}{point of evaluation during evaluation}
+is the program point $P$ determined as follows:
+\begin{itemize}
+\item
+If $E$ is a potentially-evaluated subexpression of
+a default member initializer $I$, and
+$V$ is the evaluation of $E$ in the evaluation of $I$
+as an immediate subexpression of a (possibly aggregate) initialization, then
+$P$ is the point of evaluation of that initialization.
+\begin{tailnote}
+For example,
+$E$ can be an immediate invocation in a default member initializer
+used by an aggregate initialization appearing at $P$.
+\end{tailnote}
+
+\item
+Otherwise,
+if $E$ is a potentially-evaluated subexpression of
+a default argument $A$\iref{dcl.fct.default}, and
+$V$ is the evaluation of $E$ in the evaluation of $A$ as
+an immediate subexpression of a function call\iref{expr.call}, then
+$P$ is the point of evaluation of that function call.
+
+\item
+Otherwise,
+$P$ is the point at which $E$ appears.
+\end{itemize}
 During the evaluation $V$ of an expression $E$ as a core constant expression,
-the evaluation context of an evaluation $X$\iref{intro.execution}
-consists of the following points:
+the \defnadj{evaluation}{context} of an evaluation $X$
+during $V$ is the set $C$ of program points determined as follows:
 \begin{itemize}
 \item
-The program point $\textit{EVAL-PT}(L)$,
-where $L$ is the point at which $E$ appears, and
-where $\textit{EVAL-PT}(P)$, for a point $P$,
-is a point $R$ determined as follows:
-\begin{itemize}
+If $X$ occurs during the evaluation $Y$ of
+a manifestly constant-evaluated expression,
+where $Y$ occurs during $V$, then
+$C$ is the evaluation context of $X$ during $Y$.
 \item
-If a potentially-evaluated subexpression\iref{intro.execution} of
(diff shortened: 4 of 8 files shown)


jensmaurer:
I have added "fixes cplusplus/nbballot" links in the initial comment of this pull request.  It would be good to amend the corresponding commits with our usual "Fixes NB ... (C++26 CD)." text (examples all over "git log").

PR #8926, opened 2026-04-07
[Motions 2026 03 lwg 1] P4145R0 C++ Standard Library Ready Issues

### LINKED ISSUE #2053: LWG4130 Preconditions for std::launder might be overly strict

### LINKED ISSUE #2616: LWG4492. std::generate and std::ranges::generate wording is unclear for parallel algorithms
https://cplusplus.github.io/LWG/issue4492

LWG approved this issue today, striking the new parallel overloads of `ranges::generate` and `ranges::generate_n`. As explained in the issue, there are open questions about the semantics of the existing parallel std::generate which need to be resolved before we add new parallel versions to namespace `ranges`. The conservative approach seems to be to not add new functions in C++26 until we've clarified the semantics of the existing algorithms (which will be done via a new library issue, to be opened soon). Once we know what they should do, we can decide whether to add them back again for C++29.

We would like LEWG to confirm whether removing these new functions (added to C++26 by P3179R9) is the right approach for C++26.

### LINKED ISSUE #2722: P4145 R0 C++ Standard Library Ready Issues to be moved in Croydon, Mar. 2026
[P4145R0](https://wg21.link/p4145r0) C++ Standard Library Ready Issues to be moved in Croydon, Mar. 2026 (Jonathan Wakely)

### LINKED ISSUE #8835: [2026-03 LWG Motion 1] P4145R0 C++ Standard Library Ready Issues
[P4145R0](https://wiki.isocpp.org/uploads/2026-03_Croydon-LWGStrawPollsStaging_-_P4145r0.html) C++ Standard Library Ready Issues

burblebee:
Fixes #8835.
Fixes cplusplus/papers#2722

Also fixes cplusplus/papers#2616
Also fixes cplusplus/papers#2053

Also fixes cplusplus/nbballot#708
Also fixes cplusplus/nbballot#814
Also fixes cplusplus/nbballot#879
Also fixes cplusplus/nbballot#855
Also fixes cplusplus/nbballot#830
Also fixes cplusplus/nbballot#817
Also fixes cplusplus/nbballot#815

Notes:
* LWG4474: See FIXMEs for concerns about the wording.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -8847,7 +8847,8 @@ but they now behave as iterators into \tcode{*this}, not into \tcode{x}.
 \pnum
 \throws
 \tcode{length_error} if any of \tcode{x}'s active blocks
-are not within the bounds of \exposid{current-limits}.
+are not within the bounds of \exposid{current-limits},
+as well as any exceptions thrown by the allocator.
 
 \pnum
 \complexity
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -19494,8 +19494,8 @@ template<@\exposconcept{container-compatible-range}@<value_type> R>
 \effects
 Adds elements to \exposid{c} as if by:
 \begin{codeblock}
-ranges::for_each(rg, [&](auto&& e) {
-  @\exposid{c}@.insert(@\exposid{c}@.end(), std::forward<decltype(e)>(e));
+ranges::for_each(rg, [&](value_type e) {
+  @\exposid{c}@.insert(@\exposid{c}@.end(), std::move(e));
 });
 \end{codeblock}
 Then,
@@ -20182,8 +20182,8 @@ template<@\exposconcept{container-compatible-range}@<value_type> R>
 \effects
 Adds elements to \exposid{c} as if by:
 \begin{codeblock}
-ranges::for_each(rg, [&](auto&& e) {
-  @\exposid{c}@.insert(@\exposid{c}@.end(), std::forward<decltype(e)>(e));
+ranges::for_each(rg, [&](value_type e) {
+  @\exposid{c}@.insert(@\exposid{c}@.end(), std::move(e));
 });
 \end{codeblock}
 Then, sorts the range of newly inserted elements with respect to \exposid{compare},
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -20392,10 +20392,7 @@ namespace std {
     concept @\defexposconcept{integral-constant-like}@ =                    // \expos
       is_integral_v<remove_cvref_t<decltype(T::value)>> &&
       !is_same_v<bool, remove_cvref_t<decltype(T::value)>> &&
-      @\libconcept{convertible_to}@<T, decltype(T::value)> &&
-      @\libconcept{equality_comparable_with}@<T, decltype(T::value)> &&
-      bool_constant<T() == T::value>::value &&
-      bool_constant<static_cast<decltype(T::value)>(T()) == T::value>::value;
+      @\exposconcept{constexpr-wrapper-like}@<T>;
 
   template<class T>
     constexpr size_t @\defexposconcept{maybe-static-ext}@ = dynamic_extent; // \expos
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -25030,7 +25030,7 @@ namespace std {
                 typename MappingType::layout_type>;
 
   template<class MappingType, class AccessorType>
-    mdspan(const typename AccessorType::data_handle_type&, const MappingType&,
+    mdspan(typename AccessorType::data_handle_type, const MappingType&,
            const AccessorType&)
       -> mdspan<typename AccessorType::element_type, typename MappingType::extents_type,
                 typename MappingType::layout_type, AccessorType>;
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -10652,6 +10652,18 @@ and clears it otherwise.
 \tcode{*this}.
 \end{itemdescr}
 
+\indexlibrarymember{operator bool}{vector<bool>::reference}%
+\begin{itemdecl}
+constexpr reference::operator bool() const noexcept;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\returns
+\tcode{true} if the value of the bit referred to by \tcode{*this} is one,
+\tcode{false} otherwise.
+\end{itemdescr}
+
 \indexlibrarymember{flip}{vector<bool>::reference}%
 \begin{itemdecl}
 constexpr void reference::flip() noexcept;
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -21287,12 +21287,12 @@ namespace std {
   inline constexpr full_extent_t full_extent{};
 
   template<class IndexType, size_t... Extents, class... SliceSpecifiers>
-    constexpr auto submdspan_extents(const extents<IndexType, Extents...>&, SliceSpecifiers...);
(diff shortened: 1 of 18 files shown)


jensmaurer:
I have added "fixes cplusplus/nbballot" links in the initial comment of this pull request. It would be good to amend the corresponding commits with our usual "Fixes NB ... (C++26 CD)." text (examples all over "git log").

source/support.tex
```diff
@@ -997,7 +997,11 @@
 \item
 \indexlibraryglobal{round_to_nearest}%
 \tcode{round_to_nearest}
-if the rounding style is to the nearest representable value
+if the rounding style is to the nearest representable value;
+%FIXME: The wording "equally near" seems problematic.
+if there are two equally near such values,
+%FIXME: Can multiple values have an "even least significant digit"? Or none at all?
```
frederick-vs-ja: I believe there's at least one such value. Although when the radix of the floating-point type is odd (which is theoretically permitted but doesn't seem present in any implementation), there may be two.

eisenwave: The wording is fine; I don't see any problem with the term "equally near". I could have phrased it as "equidistant" but that seemed too fancy.

An obvious example of where the "to even" part kicks in is when you divide the smallest possible `float` value by two. Zero and the subnormal minimum are equally near, and you disambiguate by choosing the number with the even least significant digit, which is zero. You get the same ambiguity halfway between all adjacent numbers with equal exponent at least.

The *really* interesting case is when even the "to even" disambiguation isn't sufficient 😄 I'm not sure about odd radix, but for even radix, this only occurs for floating-point types with a single digit mantissa digit. This case is deliberately left as a wording hole though in order to not mandate more than C's `FLT_ROUNDS` mandates. Implementers know what to do anyway.

burblebee: If there can be more than 1 "equally near" values with "even least significant digit"s, then we can't chose "the one with an even least significant digit".

eisenwave: Okay yeah that's true. I think you have to explicitly spell out the ambiguity instead of acting as if there was none for this not to be a wording defect. That belongs outside this PR and either in a follow-up editorial fix or LWG issue though, unless the fix is really obvious, which it isn't.

If you wanted to be explicit about the remaining ambiguity, you would either need to go full roundTiesToEven as per ISO/IEC 60559 and disambiguate according to that (stricter than C), or leave it implementation-defined (imposes documentation requirement), or unspecified (may be non-deterministic), and which one to choose isn't remotely obvious or editorial.

jwakely:
The fourth commit in the series (fd289eba5dcabf0c7b14cc1f3888ba5416ec4927) says "CWG" instead of "LWG" in the commit log.

source/time.tex
```diff
@@ -10788,11 +10791,17 @@
...
-The last two decimal digits of the year.
+The last two decimal digits of the year,
+regardless of the sign of the year.
 If the result is a single digit, it is prefixed by \tcode{0}.
 The modified command \tcode{\%Oy} produces the locale's alternative representation.
 The modified command \tcode{\%Ey} produces the locale's alternative representation
 of offset from \tcode{\%EC} (year only).
+\begin{tailexample}
+\begin{codeblock}
+cout << format("{:%C %y}", -1976y);     // prints \tcode{-20 76}
```
jwakely, line 10802: The issue didn't ask for a codeblock, so this is an editorial change.

source/time.tex
```diff
@@ -11478,6 +11488,12 @@
 Leading zeroes are permitted but not required.
 The modified commands \tcode{\%Ey} and \tcode{\%Oy} interpret
 the locale's alternative representation.
+\begin{tailexample}
+\begin{codeblock}
+year y;
```
jwakely, line 11493: This is also an editorial change, the issue didn't ask for a codeblock.

source/containers.tex
```diff
@@ -25846,11 +25856,12 @@
 
 \rSec4[mdspan.sub.canonical]{\tcode{submdspan} slice canonicalization}
 
-\indexlibraryglobal{submdspan_canonicalize_slices}%
+%FIXME: "Slices" is used instead of "SliceSpecifiers" in the synopsis declaration.
```
jwakely, line 25859: This should be filed as an editorial issue on github, rather than as a FIXME comment in the source.

source/meta.tex
```diff
@@ -3416,30 +3425,44 @@
...
+otherwise,
+\end{itemize}
+and $\tcode{\placeholder{it}}_i$ is an iterator to
+the $i^\text{th}$ element of \tcode{r}.
 
 \pnum
 Let $P$ be
 \begin{itemize}
 \item
   the template parameter object\iref{temp.param}
```
jwakely, line 3460: I was confused why the surrounding context didn't match what was in the issue, but it was changed by a091468825ef477a9d8110085f2963306dee9cfc

jwakely (approved):
I added a few minor comments but everything looks correct. Nice work, Dawn!

source/time.tex
tkoeppe: I don't think the presence or otherwise of a "codeblock" is something a paper can even reason about. A codeblock is merely a specific kind of whitespace, and papers can never request specific whitespace.

tkoeppe: On the other hand, the comment _is_ a change.

tkoeppe: I'd say that's still an improvement. The wording in the paper is extremely awkward in that it attempts to make an entire code snippet into the object of a sentence, and not just that: it's not just the code itself, but also some presumed execution thereof. It is much simpler to use a normal "code with comments" example style.

source/time.tex
tkoeppe: Ack, see above.

source/time.tex
jwakely: Yeah, my point was that the issue had "prose prose `inline code` prose prose `result`" and the application of the issue has a codeblock with the prose converted to a comment. It's fine if you prefer that, but it's certainly an editorial re-formulation.

tkoeppe: Got it, thanks. Yes, I think the original formulation was rather awkward. I think in the future if you wanted to say something like "the code [...] does [xyz]", it could be like: "Running the following code, [codeblock] has the effect of printing [output].", but the usual commented-code style seems to be good enough for most purposes. Maybe "After executing [codeblock], [something[ is true." would also work nicely to avoid having too much complex logic in a comment.

source/support.tex
tkoeppe: Thanks! I created #8942 to capture this.

source/meta.tex
tkoeppe: Yep, thanks for calling it out!

source/containers.tex
tkoeppe: I'll queue it up, thanks!

source/meta.tex
```diff
@@ -6650,6 +6650,9 @@
   \item
     if $V$ equals \tcode{0},
     then \tcode{options.name} does not contain a value; and
+  \item
+    if \tcode{options.name} does not contain a value, then
+    \tcode{is_const(type) || is_volatile(type)} is \tcode{false}; and
```
tkoeppe, line 6655: We're picking up a lot of "; and"s here, but that's what the paper says.

tkoeppe:
> The fourth commit in the series ([fd289eb](https://github.com/cplusplus/draft/commit/fd289eba5dcabf0c7b14cc1f3888ba5416ec4927)) says "CWG" instead of "LWG" in the commit log.

Fixed.

Issue #8872, opened 2026-03-28
[2026-03 LWG Motion 38] P4154R0 Renaming various execution things


jensmaurer:
[P4154R0](https://wiki.isocpp.org/uploads/2026-03_Croydon-LWGStrawPollsStaging_-_P4154r0_%28execution_rename%29.html) (Renaming various execution things)

eisenwave:
It's probably best if Thomas does after all other motions are merged.

PR #8927, opened 2026-04-07
[2026-03 LWG Motion 35] P3927R2 task_scheduler support for parallel bulk execution

### LINKED ISSUE #2503: LWG4336. bulk vs. task_scheduler
https://cplusplus.github.io/LWG/issue4336

LWG would like LEWG to consider this issue.

From the LWG reflector:

> [We] recommend that std::execution experts get together to talk about
sender algorithm customization -- it seems like there are several
kinds of problems that have related causes and might have a single
solution.

### LINKED ISSUE #2600: P3927 R2 task_scheduler Support for Parallel Bulk Execution
[P3927R0](https://wg21.link/p3927r0) task_scheduler Support for Parallel Bulk Execution (Eric Niebler)

### LINKED ISSUE #8869: [2026-03 LWG Motion 35] P3927R2 task_scheduler support for parallel bulk execution
[P3927R2](https://wiki.isocpp.org/uploads/2026-03_Croydon-LWGStrawPollsStaging_-_P3927R2.html) (task_scheduler support for parallel bulk execution)

jensmaurer:
Fixes NB US 238-368 (C++26 CD).

Fixes #8869
 
Also fixes cplusplus/papers#2600
Also fixes cplusplus/papers#2503
Also fixes cplusplus/nbballot#943

### DIFF
--- a/source/exec.tex
+++ b/source/exec.tex
@@ -7447,33 +7447,31 @@ the lifetime of the object to which \tcode{\placeholder{o}} refers, and
 \begin{codeblock}
 namespace std::execution {
   class @\libglobal{task_scheduler}@ {
-    class @\exposidnc{ts-sender}@;                    // \expos
+    class @\exposid{ts-domain}@;                      // \expos
 
-    template<@\libconcept{receiver}@ R>
-      class @\exposidnc{state}@;                      // \expos
+    template<@\libconcept{scheduler}@ Sch>
+      class @\exposid{backend-for}@;                  // \expos
 
   public:
     using scheduler_concept = scheduler_t;
 
     template<class Sch, class Allocator = allocator<void>>
-      requires (!@\libconcept{same_as}@<task_scheduler, remove_cvref_t<Sch>>)
-        && @\libconcept{scheduler}@<Sch>
+      requires (!@\libconcept{same_as}@<task_scheduler, remove_cvref_t<Sch>>) && @\libconcept{scheduler}@<Sch>
     explicit task_scheduler(Sch&& sch, Allocator alloc = {});
 
     task_scheduler(const task_scheduler&) = default;
     task_scheduler& operator=(const task_scheduler&) = default;
 
-    @\exposid{ts-sender}@ schedule();
+    @\seebelow@ schedule();
 
-    friend bool operator==(const task_scheduler& lhs, const task_scheduler& rhs)
-        noexcept;
+    friend bool operator==(const task_scheduler& lhs, const task_scheduler& rhs) noexcept;
     template<class Sch>
-      requires (!@\libconcept{same_as}@<task_scheduler, Sch>)
-      && @\libconcept{scheduler}@<Sch>
+      requires (!@\libconcept{same_as}@<task_scheduler, Sch>) && @\libconcept{scheduler}@<Sch>
     friend bool operator==(const task_scheduler& lhs, const Sch& rhs) noexcept;
 
   private:
-    shared_ptr<void> @\exposidnc{sch_}@; // \expos
+    // see \ref{exec.sysctxrepl.psb}
+    shared_ptr<system_context_replaceability::parallel_scheduler_backend> @\exposidnc{sch_}@;   // \expos
   };
 }
 \end{codeblock}
@@ -7484,6 +7482,10 @@ namespace std::execution {
 Given an object \tcode{s} of type \tcode{task_scheduler}, let
 \tcode{\exposid{SCHED}(s)} be the object
 pointed to by the pointer owned by \tcode{s.\exposid{sch_}}.
+The expression \tcode{get_forward_progress_guarantee(s)} is equivalent to
+\tcode{get_forward_progress_guarantee(\exposid{SCHED}(s))}.
+The expression \tcode{get_completion_domain<set_value_t>(s)} is equivalent to
+\tcode{task_scheduler::\exposid{ts-domain}()}.
 
 \indexlibraryctor{task_scheduler}
 \begin{itemdecl}
@@ -7498,8 +7500,10 @@ explicit task_scheduler(Sch&& sch, Allocator alloc = {});
 
 \pnum
 \effects
-Initialize \exposid{sch_} with
-\tcode{allocate_shared<remove_cvref_t<Sch>>(alloc,\brk{} std::forward<Sch>\brk{}(sch))}.
+Initializes \exposid{sch_} with:
+\begin{codeblock}
+allocate_shared<@\exposid{backend-for}@<remove_cvref_t<Sch>>>(alloc, std::forward<Sch>(sch))
+\end{codeblock}
 
 \pnum
 \recommended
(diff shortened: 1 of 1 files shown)


source/exec.tex
```diff
@@ -7029,33 +7029,31 @@
...
     template<class Sch>
-      requires (!@\libconcept{same_as}@<task_scheduler, Sch>)
-      && @\libconcept{scheduler}@<Sch>
+      requires (!@\libconcept{same_as}@<task_scheduler, Sch>) && @\libconcept{scheduler}@<Sch>
     friend bool operator==(const task_scheduler& lhs, const Sch& rhs) noexcept;
 
   private:
-    shared_ptr<void> @\exposidnc{sch_}@; // \expos
+    // see \ref{exec.sysctxrepl.psb}
+    shared_ptr<system_context_replaceability::parallel_scheduler_backend> @\exposidnc{sch_}@;   // \expos
```
ericniebler, line 7474: maybe you're doing this in a separate PR, but this should be `parallel_scheduler_replacement::` now.

source/exec.tex
```diff
@@ -7066,6 +7064,10 @@
 Given an object \tcode{s} of type \tcode{task_scheduler}, let
 \tcode{\exposid{SCHED}(s)} be the object
 pointed to by the pointer owned by \tcode{s.\exposid{sch_}}.
+The expression \tcode{get_forward_progress_guarantee(s)} is equivalent to
+\tcode{get_forward_progress_guarantee(SCHED(s))}.
+The expression \tcode{get_completion_domain<set_value_t>(s)} is equivalent to
+\tcode{task_scheduler::ts-domain()}.
```
ericniebler:
```suggestion
\tcode{get_forward_progress_guarantee(\exposid{SCHED}(s))}.
The expression \tcode{get_completion_domain<set_value_t>(s)} is equivalent to
\tcode{task_scheduler::\exposid{ts-domain}()}.
```

ericniebler (commented):
looks good. thanks!

source/exec.tex
jwakely: That will be done for #8872

source/exec.tex
```diff
@@ -7424,77 +7416,202 @@
 \end{itemdescr}
 
 \pnum
+For an lvalue \tcode{r} of a type derived from \tcode{receiver_proxy},
+let \tcode{\exposid{WRAP-RCVR}(r)}x be an object of a type
```
jwakely:
```suggestion
let \tcode{\exposid{WRAP-RCVR}(r)} be an object of a type
```

source/exec.tex
```diff
@@ -7424,77 +7416,202 @@
...
 \begin{codeblock}
 namespace std::execution {
-  template<@\libconcept{receiver}@ R>
-  class task_scheduler::@\exposidnc{state}@ {         // \expos
+  class task_scheduler::@\exposid{ts-domain}@ : public default_domain {     // \expos
   public:
-    using operation_state_concept = operation_state_t;
-
-    void start() & noexcept;
+    template<class BulkSndr, class Env>
```
jwakely, line 7669: The paper has:

```suggestion
    template<class BulkSndr, class Env> // \expos
```

source/exec.tex
```diff
@@ -7424,77 +7416,202 @@
...
   };
 }
 \end{codeblock}
-\exposid{state} is an exposition-only class template whose
-specializations model \libconcept{operation_state}\iref{exec.opstate}.
 
-\indexlibrarymember{start}{task_scheduler::\exposid{state}}%
 \begin{itemdecl}
-void start() & noexcept;
+template<class BulkSndr, class Env>
```
jwakely, line 7677: The paper has:
```suggestion
template<class BulkSndr, class Env> // \expos
```

jwakely (changes requested):
There's a stray `x` in the pull request.

There are two places where the paper says that `ts-domain::transform_sender` is exposition-only, which are not in the pull request. But if it's exposition-only, why isn't it called _`transform-sender`_ instead? If its name needs to be `transform_sender` then it's not exposition only, and the pull request is correct. I don't know.

source/exec.tex

source/exec.tex
tkoeppe: That'll come in a later motion, thanks!

source/exec.tex
tkoeppe: A previous expos-only class didn't annotate any of its members, neither public nor private. Since the entire class is expos-only, I don't think we gain anything from annotating members, do we? So it seems fine to omit this.

jwakely: OK, fine by me

source/exec.tex
tkoeppe: Ditto

jwakely (approved):

tkoeppe:
> There are two places where the paper says that `ts-domain::transform_sender` is exposition-only, which are not in the pull request. But if it's exposition-only, why isn't it called _`transform-sender`_ instead? If its name needs to be `transform_sender` then it's not exposition only, and the pull request is correct. I don't know.

Since *`ts-domain`* itself is an exposition-only class, I don't think it's observable whether its _members_ are also exposition-only, and we don't need to also annotate members. E.g. the members of *`backend-for`* (added in the same paper) are also not annotated. I think that's fine, I don't see what else it could mean.

jensmaurer:
> Since ts-domain itself is an exposition-only class, I don't think it's observable whether its members are also exposition-only

I disagree here.  get_completion_domain is specified to return _ts-domain_. Thus, even if this type doesn't have a known name, you can query for the presence of members using e.g. concepts or SFINAE.  And you can trivially get a name for that type with ` using T = decltype(whatever);`

tkoeppe:
Hm, good point. We should look at the paper again, then, since indeed these members aren't marked up as expos-only, and in *`backend-for`* none of the members are marked expos-only. (Weirdly enough, the _constructor_ of course needs to spell the declaration in expos-only italics, though it is (by your argument) _not_ actually expos-only -- so the choice of font is in fact only an inadequate indicator in general.

tkoeppe:
@ericniebler Could you please advice which parts of the paper are exposition-only?

jwakely:
A `transform_sender` member function on a domain is used by `std::execution::transform_sender` if it exists, so its presence or absence has normative meaning. I don't think it should have been exposition only in the paper.

jensmaurer:
I think omitting the "exposition-only" marking after "transform_sender" is good in the pull request, the paper is bad (in that regard), and life is overall splendid?

ericniebler:
the user has no way of getting their hands on a _`backend-for`_ object. that is why it is exposition only, and i was uncertain whether to annotate its members as such.

in the case of _`ts-domain`_, the type's name doesn't matter since users will not be uttering it. that is why it is exposition-only. ~~but its `transform_sender` member is _not_ marked exposition-only. it must be present.~~ whoops, mea culpa

jwakely:
> but its `transform_sender` member is _not_ marked exposition-only. it must be present.

It was marked expos-only in the paper, that's why we're asking.

ericniebler:
> It was marked expos-only in the paper, that's why we're asking.

right, i just noticed that. i agree that was an error.

tkoeppe:
Thanks, @ericniebler!

PR #8929, opened 2026-04-08
[basic.extended.fp] Reference C23 instead of "future versions"

eisenwave:


source/basic.tex
```diff
@@ -5845,9 +5845,11 @@
 the extended floating-point types described in this subsection
 that are in addition to the names declared in the \libheader{stdfloat} header
 should be chosen to increase compatibility and interoperability
-with the interchange types
+with the interchange floating types
```
frederick-vs-ja, line 5848: It's surprising that the C standard consistently says "floating type(s)" while the C++ standard consistently says "floating-point type(s)". I'm not sure whether it's better to follow the wording convention in C++.

source/basic.tex
```diff
@@ -5845,9 +5845,11 @@
...
 \tcode{_Float16}, \tcode{_Float32}, \tcode{_Float64}, and \tcode{_Float128}
-defined in ISO/IEC TS 18661-3 and with future versions of \IsoCUndated{}.
+defined in \IsoC{}.
+
+\xrefc{H.2.2}
```
frederick-vs-ja: Would it be better to say this more compactly?
```suggestion
defined in \IsoC{} H.2.2.
```

source/basic.tex
eisenwave: In this context, we are specifically talking about those types defined in C23, so it wouldn't make sense to use our own terminology.

eisenwave:
@tkoeppe it would be nice if we could do this before shipping the DIS. We should at least get cross-references to other ISO standards right.

jensmaurer (approved):

jensmaurer:
@tkoeppe , this seems important.

source/basic.tex
```diff
@@ -5920,9 +5845,9 @@
 the extended floating-point types described in this subsection
 that are in addition to the names declared in the \libheader{stdfloat} header
 should be chosen to increase compatibility and interoperability
-with the interchange types
+with the interchange floating types
```
tkoeppe, line 5848: Just "floating types", is that deliberate? Not "floating-point types"?

eisenwave: Yes, that's the C term. https://cigix.me/c23#H.2.2

PR #8930, opened 2026-04-08
[linalg.scaled.scaledaccessor, linalg.conj.conjugatedaccessor] Added missing typename

hewillk:
..which is REQUIRED.


jensmaurer:
@tkoeppe , superseded by #8961 with proper formatting and squashed commits.

PR #8932, opened 2026-04-08
[linalg.scaled.scaledaccessor] Added underline for members

hewillk:
This is consistent with the rest of [linalg].


frederick-vs-ja:
This seems to be an exact duplicate of #8424.

PR #8933, opened 2026-04-08
[diff.expr] Update incorrect comment

Halalaluyafail3:
+arr2 is a constraint violation in C, rather than being valid.

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -3077,7 +3077,7 @@ using a unary \tcode{+}.
 int arr1[5];
 int arr2[5];
 int same = arr1 == arr2;        // valid C, ill-formed C++
-int idem = arr1 == +arr2;       // valid in both C and C++
+int idem = arr1 == +arr2;       // valid C++, constraint violation in C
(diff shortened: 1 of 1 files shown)


jensmaurer:
@Halalaluyafail3 , please fix the check failure.  (See annotation in diff view.)

Halalaluyafail3:
Should the above line also be changed to use `\Cpp{}`? The `^[^%]*[^{"]C++[^"}]` pattern in check-source.sh seems to allow C++ at the beginning or end of a line but I do not see why that would be the intent.

jensmaurer:
Indeed, C23 says

> 6.5.4.3 Unary arithmetic operators
> Constraints
> The operand of the unary + or - operator shall have arithmetic type ...

For comparison, C++ says in [expr.unary.op] p7:

> The operand of the unary + operator shall be a prvalue of arithmetic, unscoped enumeration, or pointer type...

jensmaurer (approved):

jensmaurer:
@tkoeppe , I think this should go into C++26.

PR #8935, opened 2026-04-09
[Motions 2026 03 lwg 2] P4146R0 C++ Standard Library Immediate Issues

### LINKED ISSUE #2614: LWG4476. run_loop should not have a set_error completion

### LINKED ISSUE #2723: P4146 R0 C++ Standard Library Immediate Issues to be moved in Croydon, Mar. 2026
[P4146R0](https://wg21.link/p4146r0) C++ Standard Library Immediate Issues to be moved in Croydon, Mar. 2026 (Jonathan Wakely)

### LINKED ISSUE #8836: [2026-03 LWG Motion 2] P4146R0 C++ Standard Library Immediate Issues
[P4146R0](https://wiki.isocpp.org/uploads/2026-03_Croydon-StrawPolls_-_P4146r0.html) (C++ Standard Library Immediate Issues

burblebee:
Fixes #8836.
Fixes cplusplus/papers#2723

Also fixes cplusplus/papers#2614

Also fixes cplusplus/nbballot#717
Also fixes cplusplus/nbballot#728
Also fixes cplusplus/nbballot#786
Also fixes cplusplus/nbballot#787
Also fixes cplusplus/nbballot#794
Also fixes cplusplus/nbballot#803
Also fixes cplusplus/nbballot#807
Also fixes cplusplus/nbballot#821
Also fixes cplusplus/nbballot#843
Also fixes cplusplus/nbballot#877
Also fixes cplusplus/nbballot#947
Also fixes cplusplus/nbballot#954
Also fixes cplusplus/nbballot#957

### DIFF
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -10587,9 +10587,9 @@ are included in the sorted intersection.
 If, of those elements, $k$ elements from the first range
 are copied to the output range,
 then the first $k$ elements from the second range
-are considered \term{skipped}.
-If $N < M$, a non-copied element is also considered skipped
-if it compares less than the $(N + 1)^\text{th}$ element
+are considered \defn{skipped}.
+A non-copied element is also considered skipped
+if it compares less than the $\min(M, N + 1)^\text{th}$ element
 of the sorted intersection.
 Copies the first $N$ elements of the sorted intersection
 to the range \range{result}{result + $N$}.
@@ -10602,11 +10602,10 @@ to the range \range{result}{result + $N$}.
   for the overloads in namespace \tcode{std}.
 \item
   \tcode{\{last1, last2, result + $N$\}}
-  for the overloads in namespace \tcode{ranges},
-  if $N$ is equal to $M$.
+  for the non-parallel algorithm overloads in namespace \tcode{ranges}.
 \item
-  Otherwise, \tcode{\{first1 + $A$, first2 + $B$, result_last\}}
-  for the overloads in namespace \tcode{ranges},
+  Otherwise, \tcode{\{first1 + $A$, first2 + $B$, result + $N$\}}
+  for the parallel algorithm overloads in namespace \tcode{ranges},
   where $A$ and $B$ are the numbers of copied or skipped elements
   in \range{first1}{last1} and \range{first2}{last2}, respectively.
 \end{itemize}
--- a/source/algorithms.tex
+++ b/source/algorithms.tex
@@ -3633,6 +3633,8 @@ namespace std {
   namespace ranges {
     template<class I, class O>
       using @\libglobal{set_difference_result}@ = in_out_result<I, O>;
+    template<class I1, class, I2, class O>
+      using @\libglobal{set_difference_truncated_result}@ = in_in_out_result<I1, I2, O>;
 
     template<@\libconcept{input_iterator}@ I1, @\libconcept{sentinel_for}@<I1> S1, @\libconcept{input_iterator}@ I2, @\libconcept{sentinel_for}@<I2> S2,
              @\libconcept{weakly_incrementable}@ O, class Comp = ranges::less,
@@ -3653,7 +3655,7 @@ namespace std {
              @\libconcept{random_access_iterator}@ O, @\libconcept{sized_sentinel_for}@<O> OutS, class Comp = ranges::less,
              class Proj1 = identity, class Proj2 = identity>
       requires @\libconcept{mergeable}@<I1, I2, O, Comp, Proj1, Proj2>
-      set_difference_result<I1, O>
+      set_difference_truncated_result<I1, I2, O>
         set_difference(Ep&& exec, I1 first1, S1 last1, I2 first2, S2 last2,
                        O result, OutS result_last, Comp comp = {}, Proj1 proj1 = {},
                        Proj2 proj2 = {});                       // freestanding-deleted
@@ -3661,7 +3663,8 @@ namespace std {
              @\exposconcept{sized-random-access-range}@ OutR, class Comp = ranges::less,
              class Proj1 = identity, class Proj2 = identity>
       requires @\libconcept{mergeable}@<iterator_t<R1>, iterator_t<R2>, iterator_t<OutR>, Comp, Proj1, Proj2>
-      set_difference_result<borrowed_iterator_t<R1>, borrowed_iterator_t<OutR>>
+      set_difference_truncated_result<borrowed_iterator_t<R1>, borrowed_iterator_t<R2>,
+                                      borrowed_iterator_t<OutR>>
         set_difference(Ep&& exec, R1&& r1, R2&& r2, OutR&& result_r, Comp comp = {},
                        Proj1 proj1 = {}, Proj2 proj2 = {});     // freestanding-deleted
   }
@@ -10669,7 +10672,7 @@ template<@\exposconcept{execution-policy}@ Ep, @\libconcept{random_access_iterat
          @\libconcept{random_access_iterator}@ O, @\libconcept{sized_sentinel_for}@<O> OutS, class Comp = ranges::less,
          class Proj1 = identity, class Proj2 = identity>
   requires @\libconcept{mergeable}@<I1, I2, O, Comp, Proj1, Proj2>
-  ranges::set_difference_result<I1, O>
+  ranges::set_difference_truncated_result<I1, I2, O>
     ranges::set_difference(Ep&& exec, I1 first1, S1 last1,
                            I2 first2, S2 last2, O result, OutS result_last,
                            Comp comp = {}, Proj1 proj1 = {}, Proj2 proj2 = {});
@@ -10677,7 +10680,8 @@ template<@\exposconcept{execution-policy}@ Ep, @\exposconcept{sized-random-acces
          @\exposconcept{sized-random-access-range}@ OutR, class Comp = ranges::less,
          class Proj1 = identity, class Proj2 = identity>
   requires @\libconcept{mergeable}@<iterator_t<R1>, iterator_t<R2>, iterator_t<OutR>, Comp, Proj1, Proj2>
-  ranges::set_difference_result<borrowed_iterator_t<R1>, borrowed_iterator_t<OutR>>
+  ranges::set_difference_truncated_result<borrowed_iterator_t<R1>, borrowed_iterator_t<R2>,
+                                          borrowed_iterator_t<OutR>>
     ranges::set_difference(Ep&& exec, R1&& r1, R2&& r2, OutR&& result_r, Comp comp = {},
                            Proj1 proj1 = {}, Proj2 proj2 = {});
 \end{itemdecl}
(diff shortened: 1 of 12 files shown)


source/utilities.tex
```diff
@@ -8736,6 +8740,18 @@
 \tcode{is_nothrow_move_constructible_v<T> || is_nothrow_move_constructible_v<E>}
 is \tcode{true}.
 \end{itemize}
+
+%FIXME: Is this supposed to be part of the Remarks clause?
```
jwakely: Yes, but it's OK as a separate paragraph that follows the existing _Remarks_: paragraph (as discussed on mattermost)

source/containers.tex
```diff
@@ -11472,24 +11520,25 @@
...
+    remove_cvref_t<tuple_element_t<0, ranges::range_value_t<Range>>>;   // \expos
 template<ranges::@\libconcept{input_range}@ Range>
-  using @\exposid{range-mapped-type}@ = ranges::range_value_t<Range>::second_type; // \expos
+  using @\exposid{range-mapped-type}@ =
+    remove_cvref_t<tuple_element_t<1, ranges::range_value_t<Range>>>;   // \expos
 template<ranges::@\libconcept{input_range}@ Range>
   using @\exposid{range-to-alloc-type}@ =
-    pair<const typename ranges::range_value_t<Range>::first_type,
-         typename ranges::range_value_t<Range>::second_type>;           // \expos
+    pair<const @\exposid{range-key-type}@<Range>,
```
jwakely, line 11540: N.B. The P/R in the issue shows `add_const_t` here, but that was changed to `const` by #7851 so the conflict here has been resolved correctly.

source/exec.tex
```diff
@@ -7446,24 +7616,32 @@
 if that expression is valid and \tcode{scheduler_type()} otherwise.
 If neither of these expressions is valid, the program is ill-formed.
 \end{itemize}
-Let \tcode{\placeholder{st}} be \tcode{get_stop_token(get_env(\exposid{rcvr}))}.
-Initializes \tcode{\placeholder{prom}.\exposid{token}} and
-\tcode{\placeholder{prom}.\exposid{source}} such that
+After that invokes \tcode{\exposid{handle}.resume()}.
```
jwakely: I like the change to use "Finally, returns ..." below, would it make sense to be consistent here?


```suggestion
Finally, invokes \tcode{\exposid{handle}.resume()}.
```

source/exec.tex
```diff
@@ -7446,24 +7616,32 @@
...
+\end{itemdescr}
+
+\indexlibrarymember{\exposid{get-stop-token}}{task::\exposid{state}}%
```
jwakely, line 7622: Should this exposition-only member function be indexed? Huh, we do it for lots of others ... OK then.

source/exec.tex
```diff
@@ -7446,24 +7616,32 @@
...
+\begin{itemdecl}
+stop_token_type @\exposidnc{get-stop-token}@();           // \expos
+\end{itemdecl}
+\begin{itemdescr}
+\pnum
+\effects
+If
+\tcode{\libconcept{same_as}<decltype(declval<stop_source_type>().get_token()), decltype(get_\linebreak{}stop_token(get_env(\exposid{rcvr})))>}
+is \tcode{true}
```
jwakely:
```suggestion
is \tcode{true},
```

source/exec.tex
```diff
@@ -1242,6 +1239,9 @@
 Otherwise, it is expression-equivalent to
 \tcode{\exposid{MANDATE-NOTHROW}(rcvr.set_value(vs...))}.
 
+\mandates
```
jwakely, line 1242: Do we want a new paragraph for this? The issue doesn't request it, and some other places in [exec] have _Mandates_ attached to other paragraphs (or list items), so I'm not sure.

```suggestion
\pnum
\mandates
```

If we do it here, we should do it below as well.

(Maybe our library element macros like `\mandates` and `\expects` should have an implicit `\pnum`, and then have a raw form for cases like this and [structure.specifications] and [exec.get.fwd.progress] p2 where we don't want a new paragraph and a pnum.)

source/exec.tex
```diff
@@ -5567,6 +5630,87 @@
...
+      };
+    };
+
+    using @\exposid{stop-token-t}@ =                                        // \expos
+      stop_token_of_t<env_of_t<Rcvr>>;
+
+    using @\exposid{stop-callback-t}@ =                                     // \expos
+      stop_callback_for_t<@\exposid{stop-token-t}@, @\exposid{callback}@>;
+
+    struct @\exposid{receiverx}@ {                                          // \expos
```
jwakely: I don't love just appending 'x' to disambiguate this from the `receiver` concept. Is it a problem because we have scripts that complain that `\libconcept` is not used on it? Could we fix the scripts to ignore `\exposid{foo}` so we can reuse the name? Or maybe make this _`receiver-t`_ instead? That's not ideal because we also have `receiver_t` but at least they're not identical.

source/exec.tex
```diff
@@ -5567,6 +5630,87 @@
...
+    };
+
+    Rcvr @\exposid{rcvr}@;                                                  // \expos
+    StatePtr @\exposid{state}@;                                             // \expos
+    @\exposid{receiverx}@ @\exposid{inner}@;                                            // \expos
+    optional<@\exposid{stop-callback-t}@> @\exposid{stopCallback}@;                     // \expos
+
+
+    @\exposid{future-operation}@(StatePtr state, Rcvr rcvr) noexcept        // \expos
+      : @\exposid{rcvr}@(std::move(rcvr)), @\exposid{state}@(std::move(state)), @\exposid{inner}@(this)
```
jwakely, line 5684: Do we care about the shadowed named here?
```suggestion
    @\exposid{future-operation}@(StatePtr s, Rcvr r) noexcept        // \expos
      : @\exposid{rcvr}@(std::move(r)), @\exposid{state}@(std::move(s)), @\exposid{inner}@(this)
```
I don't think we need to change anything, it's fine as is.

source/exec.tex
```diff
@@ -5567,6 +5630,87 @@
...
+        @\exposid{stopCallback}@.emplace(get_stop_token(@\exposid{rcvr}@), @\exposid{callback}@(@\exposid{state}@.get()));
+      }
+      catch (...) {
+        if constexpr (!nothrow) {
+          set_error(std::move(@\exposid{rcvr}@), current_exception());
+          return;
+        }
+      }
+
+      @\exposid{state}@.release()->@\exposid{consume}@(@\exposid{innex}@);
```
jwakely:
```suggestion
      @\exposid{state}@.release()->@\exposid{consume}@(@\exposid{inner}@);
```

source/algorithms.tex
```diff
@@ -10584,9 +10587,10 @@
 If, of those elements, $k$ elements from the first range
 are copied to the output range,
 then the first $k$ elements from the second range
-are considered \term{skipped}.
-If $N < M$, a non-copied element is also considered skipped
-if it compares less than the $(N + 1)^\text{th}$ element
+\indextext{element!skipped}%
```
jwakely: This definition is local to this function, do we need to index it?

source/algorithms.tex
```diff
@@ -10716,6 +10720,13 @@
 that are equivalent to them,
 the last $\max(m - n, 0)$ elements from \range{first1}{last1}
 are included in the sorted difference, in order.
+Of those equivalent elements,
+the first $\min(m, n)$ elements in both ranges
+\indextext{element!skipped}%
```
jwakely: Same question here, do we need to index this?

source/meta.tex
```diff
@@ -3327,8 +3317,17 @@
 \pnum
 Any function in namespace \tcode{std::meta}
 whose return type is \tcode{string_view} or \tcode{u8string_view}
+%FIXME: V should not be \exposid since it's not a declared \expos varible,
+%use math font or \placeholder instead.
 returns an object \exposid{V} such that
 \tcode{\exposid{V}.data()[\exposid{V}.size()]} equals \tcode{'\textbackslash 0'}.
+Each element of the range
+\tcode{\exposid{V}.data() +} \range{0}{\exposid{V}.size()}
```
jwakely: This should be a closed range, not half-open, and the + should be math font not code font:
```suggestion
$\tcode{\exposid{V}.data()} + \crange{0}{\exposid{V}.size()}$
```

We have the `\countedrange` macro for X+[0,n) but it always uses a half-open range. We could add `\countedcrange` for X+[0,n] but I think this is the only place we need it.

jwakely (changes requested):
Some changes, some comments not requiring actions, and some questions.

source/utilities.tex
tkoeppe: I've removed the FIXMEs.

source/containers.tex
tkoeppe: Thanks!

source/exec.tex
tkoeppe: This is a weird case where we're not in an itemdescr, so our existing guidance doesn't really apply. I thought we'd just go with what's in the paper, but I agree that this sitution could do with some thought.

jwakely: [exec] has quite a lot of these cases.

tkoeppe: Alright, OK. That's certainly one less discrepancy with how it's used inside itemdecls.

tkoeppe: But let's do that later, since it touches a lot of pre-existing mandates.

source/exec.tex
tkoeppe: I'm using `rcvr-t`, which we've also used elsewhere apparently.

source/exec.tex
tkoeppe: I also don't love it, but I think this is actually quite common even in real code. We can have a look later if it comes up elsewhere, and maybe avoid it more broadly.

tkoeppe: And in any case the member is spelled in expos-only style :-)

source/exec.tex
tkoeppe: Thanks!

source/algorithms.tex
tkoeppe: There seems to be an existing parallel version of this text where it's indexed, so it seems fine to remain consistent with that, even if our overall indexing of the term "element" seems somewhat random and not particularly helpful.

source/algorithms.tex
tkoeppe: Oh, I see, the other case is introduced also by this PR. I'll remove both, thanks!

source/meta.tex
tkoeppe: Did we have to say it so weirdly? Why not `[V.data(), V.data() + V.size()]`...? I understand the appeal of "itetor plus range" notation when it's elegant and compact, but here we already have to say two member function calls of `V` anyway...

source/exec.tex
tkoeppe: Not in the paper, do you want that as an editorial drive-by fix?

source/exec.tex
tkoeppe: OK, good idea.

source/exec.tex
tkoeppe: I think I count 16 occurrences of `\mandates` without preceding `\pnum`.

PR #8939, opened 2026-04-10
Unhyphenate "potentially-evaluated"

### LINKED ISSUE #8937: "potentially evaluated expression" vs "potentially-evaluated subexpression"
#8658 changed the former to not hyphenate "potentially evaluated". [intro.execution] defines the term "potentially-evaluated subexpression", so that term is still hyphenated. And then [expr.const.defns]/2 juxtaposes the two terms for maximum contrast.

Should the latter be unhyphenated for consistency?

frederick-vs-ja:
Currently we define unhyphenated "_potentially evaluated_" in [basic.def.odr]/3 for expressions and conversions, but also define hyphenated "_potentially-evaluated subexpressions_" in [intro.execution]/4.

It's probably better to use a single form, and the hyphen shouldn't b used for compounds modifying a noun if an adverb ending in "ly" is used.

Fixes #8937.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -6449,7 +6449,7 @@ a subexpression of an immediate subexpression of $E$.
 Expressions appearing in the \grammarterm{compound-statement} of a \grammarterm{lambda-expression}
 are not subexpressions of the \grammarterm{lambda-expression}.
 \end{note}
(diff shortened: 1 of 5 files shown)


tkoeppe:
@jensmaurer I think a quick CWG checkin would be nice here, so that we can agree on a direction.

PR #8947, opened 2026-04-16
[cpp.pre] Remove confused footnote about 'lines'

### LINKED ISSUE #8934: [cpp.pre] Footnote makes unsupported claim that preprocessing directives are commonly called "lines"
[cpp.pre]/2 contains this footnote:

> 117) Thus, preprocessing directives are commonly called “lines”. These “lines” have no other syntactic significance, as all whitespace is equivalent except in certain situations during preprocessing (see the # character string literal creation operator in [cpp.stringize], for example).

I can find no evidence to refute the claim that this footnote is the only place in all of human history that the term "lines" has been used to refer specifically to preprocessing directives. I dispute that this is "common".

Also, all of raw string literals, escaped newlines, and newlines within `/*...*/` comments allow preprocessing directives to span what would be "commonly called" multiple lines. Is this footnote really adding value?

jensmaurer:
Fixes #8934

### DIFF
--- a/source/preprocessor.tex
+++ b/source/preprocessor.tex
@@ -229,15 +229,6 @@ one of the two preceding forms.
 
 The last preprocessing token in the sequence is the first preprocessing token within the sequence that
 is immediately followed by whitespace containing a new-line character.
-\begin{footnote}
-Thus,
-preprocessing directives are commonly called ``lines''.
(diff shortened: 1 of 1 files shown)


tkoeppe:
Goodbye, footnote. You did well.

PR #8952, opened 2026-04-19
Add `.vscode/` to `.gitignore`

eisenwave:


jensmaurer:
To the best of my knowledge, none of the machinery in the git repo would produce a .vscode/ directory as a temporary artifact, so why should we configure git to ignore it?  That seems unfounded.

If your local tooling produces such a directory, you might want to consider ignoring that directory in a global git config file, somewhere in your home directory.

eisenwave:
It's just a matter of convenience for those people who use VScode, either locally or by creating a codespace through GitHub.

jensmaurer:
And we entertain such pull requests for all other tools out there, with their respective local files, too?

tkoeppe:
Agreed with Jens, this is something to be solved elsewhere.

PR #8954, opened 2026-04-20
[const.wrap.class] Remove `constexpr` and `noexcept` from `operator,`

### LINKED ISSUE #8951: [const.wrap.class] The deleted `operator,` overload has meaningless `constexpr` and `noexcept`
They don't seem to make sense for a deleted overload. Perhaps it's better to remove them (after merging the PR for P3978R3).

frederick-vs-ja:
As the overloaded `operator,` is deleted, `constexpr` and `noexcept` are meaningless, so it might be better to remove them for the sake of clarity.

Fixes #8951.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -863,7 +863,7 @@ namespace std {
         { return {}; }
 
     template<@\exposconcept{constexpr-param}@ L, @\exposconcept{constexpr-param}@ R>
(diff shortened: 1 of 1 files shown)


jensmaurer:
@jwakely , are you ok with this?

jensmaurer (approved):

PR #8961, opened 2026-04-20
[linalg.scaled.scaledaccessor,linalg.conj.conjugatedaccessor] Added missing typename

jensmaurer:
Fixes #8930


jensmaurer:
(I agree the "typename" is needed here; this is not a type-only context.)

tkoeppe:
Thanks!

PR #8962, opened 2026-04-21
[numarray] Remove incorrect comments for deleted default constructors

frederick-vs-ja:
Explicitly declaring a copy constructor doesn't make the default constructor deleted, but makes it not implicitly declared at all. So it's not quite correct to say "as implied by declaring copy constructor above" for these deleted default constructors.
Moreover, the difference between deleting and not having default constructor is observable, as shown in LWG3160.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -8506,7 +8506,7 @@ namespace std {
     const slice_array& operator=(const slice_array&) const;
     void operator=(const T&) const;
 
-    slice_array() = delete;     // as implied by declaring copy constructor above
+    slice_array() = delete;
   };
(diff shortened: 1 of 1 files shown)


frederick-vs-ja:
For archaeology: These comments were introduced in [N2292](https://wg21.link/n2292), together with use of deleted functions in the standard library.

jwakely (approved):
Maybe we should not delete those, and should just have no default ctors. But that would be a design change. Given that we do have deleted constructors, we should remove the incorrect comments on them.

jensmaurer (approved):

PR #8971, opened 2026-04-22
[format.arguments, depr.format.arg] Apply `\exposid` consistently

frederick-vs-ja:
This is a "manually rebased" version of #6007.

There're conflicting changes since 2022-11-28.
- 804846a56f7e73dafe4ebd621fa81097d2e94603 relocated many contents to [text] (text.tex).
- LWG3631 largely reworded the internal constructor of `basic_format_arg`.
- [P2637R3](https://wg21.link/p2637r3) deprecated `visit_format_arg`, so `visit_format_arg` is now in [depr.format.arg].

Previously missed change(s):
- `args` in `store.args.data()` is also exposition-only (per [format.arg.store]), so it should be embraced by `\exposid{}`.

### DIFF
--- a/source/future.tex
+++ b/source/future.tex
@@ -805,7 +805,7 @@ template<class Visitor, class Context>
 \begin{itemdescr}
 \pnum
 \effects
-Equivalent to: \tcode{return visit(std::forward<Visitor>(vis), arg.value);}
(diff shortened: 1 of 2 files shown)


jwakely (approved):
This looks complete, all occurrences of these names are changed.

jensmaurer (approved):

PR #8972, opened 2026-04-22
[meta.define.static] Fix indent

hewillk:
@\seebelow@ is 11 characters long.
However, the actual displayed "see below" is 9 characters.

### DIFF
--- a/source/meta.tex
+++ b/source/meta.tex
@@ -3293,7 +3293,7 @@ using T = ranges::range_value_t<R>;
 meta::info array = meta::reflect_constant_array(r);
 if (meta::is_array_type(meta::type_of(array))) {
   return span<const T, @\seebelow@>(meta::extract<const T*>(array),
-                                    meta::extent(meta::type_of(array)));
+                                  meta::extent(meta::type_of(array)));
 } else {
   return span<const T, @\seebelow@>(static_cast<const T*>(nullptr), 0);
 }


jwakely (approved):

jwakely:
Even with this change, it doesn't really look perfect:

image

The characters on the second line aren't aligned with the first.

tkoeppe:
No, you need to use the `nc` version of the macro.

jwakely:
Ah yes, of course. Fixed at 89ffbdd802392c4f00f54c4cd2aaeda6300da9ba

tkoeppe:
Thanks!

jwakely:
Do we want to do it on the one below as well:

image

tkoeppe:
No strong opinion, it seems far enough away to not be immediately noticeable (though I would have done it myself, yes).

jwakely:
The first one looks uglier, but is correctly aligned. The second one looks prettier.

tkoeppe:
I believe there's a metaphor in there for all of C++.

hewillk:
Is there any documentation about these macros?

hewillk:
Oh, I see, they're located in macros.tex.

timsong-cpp:
> The first one looks uglier, but is correctly aligned. The second one looks prettier.

How about using `\itcorr` to add the extra space to the second line instead?

image

jwakely:
LGTM

tkoeppe:
Ugh, I'd really rather not.

PR #8974, opened 2026-04-22
[expr.ref] Do not say that the direct base relationship is (T1, B)

cpplearner:
[CWG 3114](https://cplusplus.github.io/CWG/issues/3114.html) makes it clear that the first type in the direct base relationship does not need to be `T1`.


jensmaurer (commented):
Agreed.  We later trigger on "If E2 designates a direct base class relationship" and check its properties in detail.

tkoeppe:
@jensmaurer OK to apply this then?

PR #8978, opened 2026-04-22
[meta.define.static] Use \itcorr instead of \seebelownc

timsong-cpp:
See https://github.com/cplusplus/draft/pull/8972#issuecomment-4296777253

### DIFF
--- a/source/meta.tex
+++ b/source/meta.tex
@@ -3292,8 +3292,8 @@ Equivalent to:
 using T = ranges::range_value_t<R>;
 meta::info array = meta::reflect_constant_array(r);
 if (meta::is_array_type(meta::type_of(array))) {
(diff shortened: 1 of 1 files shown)


tkoeppe:
No, I'd really rather not... if we do care about columns, we should have equally spaced columns I'd say.

timsong-cpp:
This isn't novel - we already have 14 existing uses that are exactly like this.

tkoeppe:
We also have iostreams...

timsong-cpp:
It was [the original purpose of `\itcorr`](https://github.com/cplusplus/draft/commit/5ca060ccacdcf956dca0827df6e71a84a34a05a7)...

Another possible option is to apply `\itcorr[-1]` before `\seebelow`, effectively letting these italics grow leftward instead of rightward.
image

PR #8988, opened 2026-04-24
[re.tokiter.general] Fix incorrect `suffix.match`

frederick-vs-ja:
The type of `suffix` is a `sub_match` specialization, so the intended expression is probably `suffix.matched`.


jwakely (approved):

jensmaurer (approved):

jwakely:
This has been wrong since [TR1](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2005/n1836.pdf#page=162).

PR #8992, opened 2026-04-26
[atomics.types.int] Fix indentation

Andreas-Krug:


jensmaurer (approved):
Expected fall-out from adding "constexpr" to these functions.

PR #8994, opened 2026-04-26
[simd.math] Fix indentation

Andreas-Krug:


tkoeppe:
I subsumed this in a different change, thanks!

PR #8997, opened 2026-04-26
[mdspan.sub.map.sliceable] Fix M::extent_type to M::extents_type

abhinavagarwal07:
Layout mapping types expose `extents_type`, not `extent_type` (the latter is a member of `extent_slice`). The Let clause in `[mdspan.sub.map.sliceable]` defines `IT` and `M_rank` via the undefined member `M::extent_type`, making the subsequent sliceable layout mapping requirements uninterpretable.

The same subclause uses `LayoutMapping::extents_type::rank()` and `SM::extents_type::{rank,index_type}` correctly elsewhere.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -26074,11 +26074,11 @@ Let:
 \item
   \tcode{M} denote a layout mapping class;
 \item
-  \tcode{IT} denote \tcode{M::extent_type::index_type};
+  \tcode{IT} denote \tcode{M::extents_type::index_type};
 \item
   \tcode{m} denote a value of type (possibly const) \tcode{M};
 \item
-  \tcode{M_rank} be equal to \tcode{M::extent_type::rank()};
+  \tcode{M_rank} be equal to \tcode{M::extents_type::rank()};
 \item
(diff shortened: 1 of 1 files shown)


tkoeppe:
@mhoemmen could you have a look?

jwakely (approved):

jwakely:
The error is in the [incoming paper](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p3663r3.html#new-section-mdspan.sub.sliceable-sliceable-layout-mapping-requirements) but it should definitely be `extents_type`. The `extents_slice::extent_type` alias typically refers to an integral or integral-constant-like type, so `extent_type::index_type` and `extent_type::rank()` are not valid - it should certainly be `extents_type`.

jensmaurer (approved):

mhoemmen:
@jwakely wrote:
> @mhoemmen could you have a look?

I agree, we messed up there!  In P3663, [mdspan.sub.sliceable] 1.2 should read

> `IT` denote `M::extents_type::index_type`;

instead of

> `IT` denote `M::extent_type::index_type`;

Thanks for the catch!

PR #8998, opened 2026-04-26
[mdspan.sub.map.sliceable] Use lm in sliceable-mapping concept body

abhinavagarwal07:
The `sliceable-mapping` concept body introduces `lm` (an object of type `LayoutMapping`) but then uses `m` (an object of type `M` from the surrounding requirements section, with `M` not bound to `LayoutMapping`) in the `submdspan_mapping` well-formedness check.


jwakely (approved):

mhoemmen:
Thanks for the fix! : - )

PR #9009, opened 2026-05-03
[atomics.types.int] Fix indentation

Andreas-Krug:


jensmaurer (approved):
Thanks a lot for your continued work on cleaning up the indentation!

PR #9017, opened 2026-05-12
[istream.formatted.arithmetic] remove unnecessary whitespace in codeb…

jwakely:
…locks


tkoeppe:
Andreas has trained us well!

PR #9027, opened 2026-05-18
[stmt.expand] Fix example 2

xmcgcg:
`sizeof(char[s])` does not make sense, adding `s` to `result` directly would pass the assertion.

### DIFF
--- a/source/statements.tex
+++ b/source/statements.tex
@@ -971,7 +971,7 @@ consteval int f() {
   constexpr std::array<int, 3> arr {1, 2, 3};
   int result = 0;
   template for (constexpr int s : arr) {                // OK, iterating expansion statement
-    result += sizeof(char[s]);
+    result += s;
   }
   return result;
 }


frederick-vs-ja:
> `sizeof(char[s])` does not make sense

I think it does make sense, because it also requires `s` to be a constant expression. Maybe it's clearer to use `std::cw<s>` for the purpose.

eisenwave:
To be fair, it's surprising that we would find it surprising that `constexpr int s` makes `s` usable in a constant expression, and that we feel the need to emphasize that with `sizeof(char[s])`.

What's next?

```cpp
constexpr int x = 10;
// just in case you didn't believe that 'x' is a constant expression:
constexpr int y = sizeof(char[x]);
```

I get that it's inside `template for`, but it's still spelled `constexpr int s`, so why do we think `sizeof(char[s])` adds information?

frederick-vs-ja:
`sizeof(char[s])` was initially added in brevzin/cpp_proposals#290 by @katzdm. I'd like to see author's opinion.

katzdm:
Yeah, I think I just wanted to do something that you couldn't do in a vanilla `for` loop. As long as we have some such thing, I'm not attached to the particulars of the example.

tkoeppe:
On the other hand, does this _need_ changing? What's wrong with the example as is?

eisenwave:
> On the other hand, does this need changing? What's wrong with the example as is?

A part of the example is exotic and surprising, and distracts from what the example is actually about.

In the same fashion, writing `int{1 + 1}` instead of `2` wouldn't be *wrong*, but if someone went out of their way to write that, you would think that it serves some important purpose, and it would leave you confused if that wans't the case.

eisenwave:
To be fair, I can see how we might want something that demonstrates that you can use constant expressions in loops with this feature, but making some contrived and unnecessary use inside `sizeof(char[N])` isn't ideal.

Maybe something with `std::tuple<int, long, long long>` and `std::get` would work more nicely.

PR #9041, opened 2026-05-27
[numarray] Move [valarray.range] to suitable place and level

frederick-vs-ja:
For the sake of consistency with the synopsis of `valarray` (after the modifications from P3016R6), [valarray.cassign] should be placed between [valarray.cassign] and [valarray.members]. Moreover, it should be in level `\rSec3` now.


jensmaurer (approved):
Agreed.

tkoeppe:
Nice, thanks!

PR #9055, opened 2026-06-08
[meta.reflection.queries] Add \keyword for explicit

Andreas-Krug:


source/meta.tex
```diff
@@ -4091,10 +4091,10 @@
 \begin{itemdescr}
 \pnum
 \returns
-\tcode{true} if \tcode{r} represents a member function that is declared explicit.
+\tcode{true} if \tcode{r} represents a member function that is declared \keyword{explicit}.
```
jensmaurer, line 4094: I'm not sure whether that's appropriate, because we can have ` explicit(1 == 0)` meanwhile.  That's certainly not an explicit function, but is it "declared `explicit`"?  The keyword certainly is present.

PR #9057, opened 2026-06-08
[expr.const.i{nit,mm}] Add xrefs for "manifestly constant-evaluated"

tkoeppe:
After the "constant expressions" reorganisation the definition is now in a different subclause.


opensdh (approved):
Meh.

tkoeppe:
Meh? This seems highly useful :upside_down_face:

PR #9065, opened 2026-06-13
[2026-06 CWG Motion 4] P3899R3 Clarify the behavior of floating-point overflow

### LINKED ISSUE #77: [expr] The unclear point about whether it is UB for floating-pointer type

[expr.pre] p4 says
> If during the evaluation of an expression, **the result is not mathematically defined** or not in the range of representable values for its type, **the behavior is undefined**.

[expr.mul] p4
> If the second operand of / or % is zero the behavior is undefined.

The above two rules both specify the cases where UB can be touched. 

[expr.const] p5 bullet 8
> **an operation that would have undefined behavior** as specified in **[intro] through [cpp]**;

Consider this example: 
````cpp
#include <iostream>
#include <limits>

int  main(){
    if constexpr(std::numeric_limits<float>::is_iec559){  // ensure the following expression is evaluated when the condition is true
        //constexpr float v = 2.0f / 0;  // #1
        constexpr float v2 = std::numeric_limits<float>::quiet_NaN();  // #2
    }
}
````
GCC and Clang report that `#2` is ok but `#1` is ill-formed. However, "not a number" is not mathematically defined, which is considered UB, however, it is defined under ISO/IEC/IEEE 60559. According to [expr.mul] p4, `#1` is unconditionally UB, however, it is defined under ISO/IEC/IEEE 60559. 

The inconsistent thing here is that both `#1` and `#2` are defined under ISO/IEC/IEEE 60559 but `#1` is viewed as UB and `#2` is well-defined, according to the implementation. 

### Suggested resolution

Should we relax these conditions for floating-pointer type by saying them implementation-defined?

### LINKED ISSUE #2565: P3899 R1 Clarify the behavior of floating-point overflow
[P3899R0](https://wg21.link/p3899r0) Clarify the behavior of floating-point overflow (Jan Schultke, Matthias Kretz)

### LINKED ISSUE #9071: [2026-06 CWG Motion 4] P3899R3 Clarify the behavior of floating-point overflow
[P3899R3](https://wiki.isocpp.org/uploads/2026-06_Brno-StrawPolls_-_P3899R3.html) (Clarify the behavior of floating-point overflow)

### PR DESCRIPTION
**Pull request by eisenwave**
Fixes #9071
Also fixes https://github.com/cplusplus/papers/issues/2565
Also fixes https://github.com/cplusplus/CWG/issues/77

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -62,9 +62,52 @@ see~\ref{over.match.oper}, \ref{over.built}.
 \indextext{overflow!undefined}%
 \indextext{zero!division by undefined}%
 \indextext{zero!remainder undefined}%
-If during the evaluation of an expression, the result is not
-mathematically defined or not in the range of representable values for
-its type, the behavior is undefined\ubdef{expr.expr.eval}.
+An \defnadj{arithmetic}{expression} is
+\begin{itemize}
+\item a unary plus or minus\iref{expr.unary.op},
+\item an addition\iref{expr.add},
+\item a subtraction\iref{expr.sub}, or
+\item a multiplication, division, or remainder\iref{expr.mul}
+\end{itemize}
+expression where every (possibly converted) operand is of arithmetic type.
+\begin{note}
+There exist non-arithmetic expressions such as compound assignment\iref{expr.assign}
+which are defined in terms of arithmetic expressions.
+\end{note}
+The behavior of evaluating an arithmetic expression is undefined\ubdef{expr.expr.eval}
+if the mathematical result is neither
+\begin{itemize}
+\item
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: source/expressions.tex [hunk 1]
```diff
@@ -8491,6 +8534,14 @@
 zero\iref{expr.mul}, or certain shift operations\iref{expr.shift}.
 \end{footnote}
 
+\item
+an arithmetic expression where
```
jensmaurer:
```suggestion
an arithmetic expression\iref{expr.pre} where
```

We should have a cross-reference to this definition of a novel term.

jensmaurer (approved):

PR #9107, opened 2026-06-13
[2026-06 CWG Motion 2] P3596R3 Undefined Behavior and IFNDR Annexes

### LINKED ISSUE #2623: P3596 R2 Undefined Behavior and IFNDR Annexes
[P3596R0](https://wg21.link/p3596r0) Undefined Behavior and IFNDR Annexes (Joshua Berne, Timur Doumler, Jens Maurer, Shafik Yaghmour)

### LINKED ISSUE #9069: [2026-06 CWG Motion 2] P3596R3 Undefined Behavior and IFNDR Annexes
https://wiki.isocpp.org/uploads/2026-06_Brno-StrawPolls_-_P3596R3.pdf

notadragon:
Fixes https://github.com/cplusplus/draft/issues/9069
Also fixes https://github.com/cplusplus/papers/issues/2623

This is from the branch from which the paper was generated, with a small number of editorial changes already made:
1. Most notable, after rendering the standard the annex titles were very large and wrapping to multiple lines.  After discussion, Jens and I shortened them by removing "Enumeration of" from each one.
2. One set of comments was fixed to use "IFNDR" the way the rest of the annex does instead of the (comma-less) "ill formed no diagnostic required"
3. One bug that was discovered during wording review of another paper was fixed, but that entry will also be removed when that paper lands in a later motion.
4. The \lastcorechapter macro was used where it was supposed to be used because my paper rendering couldn't handle that when the P paper was generated.

### DIFF
--- a/source/ifndr.tex
+++ b/source/ifndr.tex
@@ -1,0 +1,1059 @@
+%!TEX root = std.tex
+\infannex{ifndr}{Ill-formed, no diagnostic required}
+
+\rSec1[ifndr.general]{General}
+
+This Annex documents
+rules for which no diagnostic is required,
+called out in
+\ref{intro} through \ref{\lastcorechapter}
+using the following phrases:
+\begin{itemize}
+\item no diagnostic is required
+\item no diagnostic required
+\item a diagnostic is required only if
+\end{itemize}
+Each entry contains
+a title,
+a cross-reference,
+a summary of the circumstances,
+and code examples.
+The code examples are not intended
+to exhaustively cover all possible ways of invoking that case.
+
+\rSec1[ifndr.lex]{\ref*{lex}: Lexical conventions}
+
+\ifndrdescription{lex.name.reserved}
+
+\pnum
+Using an identifier reserved for use by \Cpp{}
+is ill-formed, no diagnostic required.
+
+\pnum
+\begin{example}
+\begin{codeblock}
+int _z; // IFNDR, \tcode{\_z} is reserved because it starts with \tcode{\_} at global scope
+
+int main() {
+    int __x;    // IFNDR, \tcode{\_\_x} is reserved because it starts with \tcode{\_\_}
+    int _Y;     // IFNDR, \tcode{\_Y} is reserved because it starts with \tcode{\_} followed by a capital letter
+    int x__y;   // IFNDR, \tcode{x\_\_y} is reserved because it contains \tcode{\_\_}
+}
+\end{codeblock}
+\end{example}
+
+\rSec1[ifndr.basic]{\ref*{basic}: Basics}
+
+\ifndrdescription{basic.link.consistent.types}
+
+\pnum
+Having multiple declarations of the same entity
+with different kinds
+when those declarations are not reachable from one another
+is ill-formed, no diagnostic required.
+
+\pnum
+\begin{example}
+\begin{codeblocktu}{Module interface of \tcode{M}}
+void g();                       // \#1
+void h();                       // \#2
+template <typename T> int j;    // \#3
+\end{codeblocktu}
+\begin{codeblocktu}{Module interface of \tcode{N}}
+int g();                        // same entity as \#1, different type
+namespace h {}                  // same entity as \#2, not both namespaces
+template <int X> int j;         // same entity as \#3, non-equivalent template heads
+\end{codeblocktu}
+\begin{codeblocktu}{Other translation unit}
+import M;
+import N;
+    // IFNDR due to the mismatched pairs above
+\end{codeblocktu}
+\end{example}
+
+\ifndrdescription{basic.def.odr.minimum.one.def}
+
+\pnum
+Not having a definition
+for a function or variable
+that is odr-used from a non-discarded statement\iref{stmt.if}
+is ill-formed, no diagnostic required.
+
+\pnum
+\begin{example}
+\begin{codeblock}
+auto f() {
+  struct A {};
+  return A{};
+}
+decltype(f()) g();
+auto x = g();   // IFNDR, function \tcode{g} is used but not defined in this translation unit, and cannot
+                // be defined in any other translation unit because its type does not have linkage
+\end{codeblock}
+\end{example}
+
+\ifndrdescription{basic.def.odr.injected.match}
+
+\pnum
+Defining a definable item \tcode{D}
+with an injected declaration\iref{expr.const.reflect}
+in one translation unit and
+a definition in a different translation unit
+when \tcode{D} is not attached to a named module
+or neither definition is reachable from the other
+is ill-formed, no diagnostic required.
+
+\pnum
+\begin{example}
+\begin{codeblocktu}{Translation unit \#1}
+struct S;
+consteval { std::meta::define_aggregate(^^S, {}); }     // \#1
+\end{codeblocktu}
+\begin{codeblocktu}{Translation unit \#2}
+struct S {};    // IFNDR, definition here, injected declaration at \#1
+\end{codeblocktu}
+\end{example}
+
+\ifndrdescription{basic.def.odr.maximum.one.def}
+
+\pnum
+If there are definitions
+in different translation units
+of a non-inline non-templated function or variable
+that are not attached to a named module
+or are not reachable from one another,
+the program is ill-formed, no diagnostic required.
+
+\pnum
+\begin{example}
+\begin{codeblocktu}{Translation unit \#1}
+export module M:A;  // module partition
+void f() {}         // \#1
+void g() {}         // \#2
+\end{codeblocktu}
+\begin{codeblocktu}{Translation unit \#2}
+export module M:B;  // module partition
+void f() {}         // IFNDR, \#1 not reachable
+\end{codeblocktu}
+\begin{codeblocktu}{Translation unit \#3}
+export module M;    // primary module interface unit
+export import :A;
+void g();           // error: \#2 is reachable
+\end{codeblocktu}
+\end{example}
+
+\ifndrdescription{basic.def.odr.definition.matches}
+
+\pnum
+If there are definitions
+in different translation units
+of a definable item \tcode{D}
+where
+\begin{itemize}
+\item \tcode{D} is not defined by an injected declaration\iref{expr.const.reflect},
+\item \tcode{D} is not an inline or templated function or variable, and
+\item \tcode{D} is not attached to a named module or the declarations are not reachable from one another,
+\end{itemize}
+that do not satisfy the matching rules described in \ref{basic.def.odr},
+the program is ill-formed, no diagnostic required.
+
+\pnum
+\begin{example}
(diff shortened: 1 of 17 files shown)


source/ub.tex
```diff
@@ -0,0 +1,1976 @@
...
+int main() {
+  unsigned long long x2 = -1;
+  float f = x2; // undefined behavior on systems where \tcode{f} does not include
+                // a representation for infinity and the maximum value for \tcode{float}
+                // is smaller than the maximum value for \tcode{unsigned long long}.
+}
+\end{codeblock}
+\end{example}
+
+\ubdescription{conv.fpint.float.not.represented}
```
shafik: Super minor but I think this one should be swapped w/ `conv.fpint.int.not.represented` b/c in `7.3.11` we have `E.3.8` first and then `E.3.7` but again super minor nit.

source/ub.tex
```diff
@@ -0,0 +1,1976 @@
...
+  B* dp = dynamic_cast<B*>(bs+1);   // undefined behavior
+}
+\end{codeblock}
+\end{example}
+
+\ubdescription{expr.dynamic.cast.glvalue.lifetime}
+
+\pnum
+Evaluating a \keyword{dynamic_cast} on a reference that
+denotes an object (of polymorphic type) of the wrong type or an object
```
shafik:
```suggestion
denotes an object (of polymorphic type) of the similar type or an object
```
I am not sure but looking at the core wording, I think that is more correct?

source/ub.tex
```diff
@@ -0,0 +1,1976 @@
...
+
+\pnum
+For addition or subtraction of two expressions P and Q,
+if P or Q have type ``pointer to cv T'', where T and the array
+element type are not similar\iref{conv.rval}, the behavior is undefined.
+
+\pnum
+\begin{example}
+\begin{codeblock}
+struct S {
```
shafik: You don't indicate the UB in the example and why didn't you use the simpler example from 7.6.6? Maybe this was my original example?

source/ub.tex
```diff
@@ -0,0 +1,1976 @@
...
+static_assert(sizeof(int) == 4 && CHAR_BIT == 8);
+int y1 = 1 << 32;       // undefined behavior, shift is equal to the bit width of \tcode{int}
+int y2 = 1 >> 32;       // undefined behavior, shift is equal to the bit width of \tcode{int}
+\end{codeblock}
+\end{example}
+
+\ubdescription{expr.assign.overlap}
+
+\pnum
+Overlap in the storage between the source and destination may result in undefined behavior.
```
shafik: Should we be using `may` here?

source/ub.tex
```diff
@@ -0,0 +1,1976 @@
...
+struct D : B {
+  void f() override;
+};
+\end{codeblock}
+\end{example}
+
+\ubdescription{class.base.init.mem.fun}
+
+\pnum
+It is undefined behavior to call a member function before all the \grammarterm{mem-initializer}s for base classes have completed.
```
shafik: It looks like there are two contracts cases added that we are missing. This feels minor and we can add afterwards too.

shafik (commented):
Finished UB section.

source/ub.tex
jensmaurer: may -> can

jensmaurer:
@notadragon, please squash all commits into the first one; we don't need the details of the genesis preserved for posterity.

jensmaurer:
@notadragon, do we have "ubx" as part of a visible label in the Annex headings now?  We shouldn't; those visible labels should be "ub:something". (Similarly with ifndrx, if any).

source/ifndr.tex
```diff
@@ -0,0 +1,1061 @@
...
+module std0;            // IFNDR, \tcode{std} followed by digits is not allowed at the beginning
+export module _Test;    // IFNDR, \tcode{_Test} is a reserved identifier
+export module te__st;   // IFNDR, \tcode{te__st} is a reserved identifier
+\end{codeblock}
+\end{example}
+
+\ifndrdescription{module.unit.named.module.no.partition}
+
+\pnum
+Having multiple primary module interface units
```
shafik, line 543: I am not sure this description matches what the wording says but I may just not be reading it correctly.

shafik (commented):
Finished reviewing IFNDR annex. Really minor comments. I think we are in good shape here.

source/ub.tex
notadragon: will rearrange.

source/ub.tex
notadragon: "of the similar type" is wrong, similar type would be fine.  "wrong" type could be interpreted broadly as "any type that is wrong for doing this" but we can be more precise and say "of a type that is not similar" .   I will change to that.

source/ub.tex
notadragon: i will compact this example slightly and add a comment about the UB.  Not sure who the example dates back to (probably you).  I would rather not rewrite examples entirely in this phase so will leave switching to clearer/simpler examples for followup efforts.

source/ub.tex
notadragon: done

source/ub.tex
notadragon: I am prepping a PR with new/missing entries and any entries that need major updating after all of the Brno motions land, I will make note to include this in that PR (and the paper that will be generated from that PR).  I think it should wait until after all of the Brno motions land.

notadragon:
> @notadragon, do we have "ubx" as part of a visible label in the Annex headings now? We shouldn't; those visible labels should be "ub:something". (Similarly with ifndrx, if any).

No, I swapped which label is "ub:" and which is "ubx:", and I also made the "stable names" shown in the annex start with "ub:" and ifndr:.

Here's some screenshots of my locally built copy of the standard for review:

image


image

source/ifndr.tex
notadragon: The wording defines "primary module interface unit" and says, effectively, "A named module shall contain one module interface unit that meets that definition, no diagnostic required".  As far as I can tell that's what the annex entry is saying as well, though i'm happy to adjust if it's wrong or there's clearer wording.  Leaving this as-is for now.

notadragon:
@jensmaurer @shafik I've addressed, I believe, all of your comments.

shafik:
> @jensmaurer @shafik I've addressed, I believe, all of your comments.

Thank you so much for all this work to get the annexes landed in the draft! I am happy with direction in your replies.

source/ifndr.tex
```diff
@@ -0,0 +1,1059 @@
...
+into a \grammarterm{requirement}
+would always result in a substitution failure,
+the program is ill-formed; no diagnostic required.
+
+\pnum
+\begin{example}
+\begin{codeblock}
+template <typename T> concept C = requires {
+  new int[-(int)sizeof(T)];     // IFNDR, the size of the allocation is required to be greater
+                                // than zero but can never be
```
frederick-vs-ja, line 300: This example seems to be well-formed per the analysis in #6055. Looks like that we incorrectly copied a legacy example.

source/ifndr.tex
```diff
@@ -0,0 +1,1059 @@
...
+inline void f() {}          // \#1
+inline void g() {}          // \#2
+inline void h() {[]{}();}   // \#3
+namespace { int i = 0; }
+inline void j() {++i;}      // \#4
+\end{codeblocktu}
+\begin{codeblocktu}{Translation unit \#2}
+inline void f() {}          // OK, same as \#1
+inline void g() {;}         // IFNDR, different sequence of tokens than \#2
+inline void h() {[]{}();}   // IFNDR, closure has different type than \#3
```
frederick-vs-ja, line 172: Such 2 definitions of `h` seem to be well-formed per CWG2300. Perhaps we should change to use default arguments.

source/ifndr.tex
notadragon: We've landed this motion, but I am going to do a pass at a followup PR to clean up any discovered issues/gaps and to reconcile with the other motions that landed in Brno.  I'll try to address this in that PR.  (Though recommended alternative examples are welcome.)

source/ifndr.tex
notadragon: Agreed, this is also not right, i will try to include a fix in the followup PR/paper.   (Concrete alternative examples are welcome)

source/ifndr.tex
frederick-vs-ja: Maybe we can adopt the proposed example in #6062.
```C++
  new decltype((void)T{});      // IFNDR, the type of the object to create is never valid
```

source/ifndr.tex
frederick-vs-ja: I think we can just change the case to `inline void h(void (*)() = []{}()) {}`.

source/ifndr.tex
```diff
@@ -0,0 +1,1059 @@
...
+\pnum
+\begin{example}
+\begin{codeblock}
+int _z; // IFNDR, \tcode{\_z} is reserved because it starts with \tcode{\_} at global scope
+
+int main() {
+    int __x;    // IFNDR, \tcode{\_\_x} is reserved because it starts with \tcode{\_\_}
+    int _Y;     // IFNDR, \tcode{\_Y} is reserved because it starts with \tcode{\_} followed by a capital letter
+    int x__y;   // IFNDR, \tcode{x\_\_y} is reserved because it contains \tcode{\_\_}
+}
```
Halalaluyafail3, line 41: Is there a reason these underscores have backslashes before them? It seems inconsistent since this is the only place that does it.

source/ifndr.tex
Halalaluyafail3: > I think we can just change the case to `inline void h(void (*)() = []{}()) {}`.

Note that the example added in CWG2300 is:
```cpp
inline void g(bool cond, void (*p)() = []{}) {
    if (cond) g(false);
}
```
I believe it is necessary to include the recursive call, if it was not then it presumably would have been omitted.

notadragon: I had gone with a local function declaration being called, but i think the recursive call is better.  I will update it.  (See https://isocpp.org/files/papers/D4284R0.pdf for the fixes i intend to land)

source/ifndr.tex
notadragon: because underscores in LaTeX are stupid.   These render the same way either way _, so removing them would be purely editorial.  (I will remove them in the PR for https://isocpp.org/files/papers/D4284R0.pdf).

Halalaluyafail3: It seems to cause issues with rendering on <https://eel.is/c++draft/ifndr:lex.name.reserved>.

tkoeppe:
We have macros to produce double underscores if that's an issue...

notadragon:
> We have macros to produce double underscores if that's an issue...

The standard renders fine without the _s, we can just remove them.  I only put them there out of personal habit.  (Though this also sounds like it is triggered by an eel.is problem, not an us problem :) )

PR #9111, opened 2026-06-14
[2026-06 LWG Motion 13] P3692R4 How to Avoid OOTA Without Really Trying

### LINKED ISSUE #2322: P3692 R0 How to Avoid OOTA Without Really Trying
[P3692R0](https://wg21.link/p3692r0) How to Avoid OOTA Without Really Trying (Paul E. McKenney, Alan Stern, Michael Wong, and Maged Michael)

### LINKED ISSUE #9100: [2026-06 LWG Motion 13] P3692R4 (How to Avoid OOTA Without Really Trying)
[P3692R4](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2026/p3692r4.pdf) (How to Avoid OOTA Without Really Trying)

eisenwave:
Fixes #9100
Also fixes https://github.com/cplusplus/papers/issues/2322

---

I feel like we could prettify the first note (in a subsequent PR) a bit by factoring out the

> An implementation that ... does not produce out-of-thin-air values for a program free of undefined behavior

This could be done in a follow-up PR that turns this into

> An implementation does not produce out-of-thin-air values for a program free of undefined behavior if
> - ..., or
> - ...

where the two bullets are the two sentences that we have now.

### DIFF
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -3009,6 +3009,29 @@ if (r2 == 42) x.store(42, memory_order::relaxed);
 \end{codeblock}
 \end{note}
 
+\begin{note}
+An implementation that
+translates to instruction sequences for a physical machine and
+treats non-volatile atomic accesses as though they were volatile
+does not produce out-of-thin-air values for a program free of undefined behavior.
+An implementation that
+performs only thread-at-a-time analysis in support of optimizations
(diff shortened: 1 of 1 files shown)


tkoeppe:
Rule 1 of WG21: Only Core gets to touch SG-1 notes :-)

PR #9114, opened 2026-06-14
[range.join.view, range.join.with.view] Make the comment more smooth.

hewillk:
The `when !forward_range<V>` here is a bit unnatural.


tkoeppe:
@jwakely This seems like an improvement.

jwakely (approved):
Smooooooooth.

This is exactly how we say it in [range.lazy.split.view]

tkoeppe:
In it goes!

PR #9120, opened 2026-06-14
[2026-06 LWG Motion 17] P3125R6 constexpr pointer tagging

### LINKED ISSUE #1903: P3125 R0 Pointer tagging
[P3125R0](https://wg21.link/p3125r0) Pointer tagging (Hana Dusíková)

### LINKED ISSUE #9104: [2026-06 LWG Motion 17] P3125R6 (constexpr pointer tagging)
[P3125R6](https://wiki.isocpp.org/uploads/2026-06_Brno-StrawPolls_-_P3125R6-pointer-tagging.html) (constexpr pointer tagging)

eisenwave:
Fixes #9104
Also fixes https://github.com/cplusplus/papers/issues/1903

### DIFF
--- a/source/memory.tex
+++ b/source/memory.tex
@@ -8,12 +8,13 @@ This Clause describes components for memory management.
 
 \pnum
 The following subclauses describe general memory management facilities,
-smart pointers, memory resources, and scoped allocators,
+smart pointers, pointer tagging, memory resources, and scoped allocators,
 as summarized in \tref{mem.summary}.
 
 \begin{libsumtab}{Memory management library summary}{mem.summary}
 \ref{memory}                & Memory                            & \tcode{<cstdlib>}, \tcode{<memory>} \\ \rowsep
 \ref{smartptr}              & Smart pointers                    & \tcode{<memory>}           \\ \rowsep
+\ref{ptrtag}                & Pointer tagging                   & \tcode{<memory>}           \\ \rowsep
 \ref{mem.composite.types}   & Types for composite class design  & \tcode{<memory>}           \\ \rowsep
 \ref{mem.res}               & Memory resources                  & \tcode{<memory_resource>}  \\ \rowsep
 \ref{allocator.adaptor}     & Scoped allocators                 & \tcode{<scoped_allocator>} \\
@@ -37,6 +38,10 @@ for containers and other template types, destroy objects, and
 construct objects in
 uninitialized memory
 buffers~(\ref{pointer.traits}--\ref{specialized.addressof} and \ref{specialized.algorithms}).
+The header also defines class template \tcode{pointer_tag_pair}
+to support storing additional information in unused bits of pointers.
+Selection of the bits is implementation-defined,
+and the size of the pointer wrapping object is the same as the size of the original pointer.
 The header also defines the templates
 \tcode{unique_ptr}, \tcode{shared_ptr}, \tcode{weak_ptr},
 \tcode{out_ptr_t}, \tcode{inout_ptr_t}, and various function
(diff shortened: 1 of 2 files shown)


eisenwave:
Need some help resolving the FIXMEs.

source/memory.tex
```diff
@@ -2119,6 +2162,428 @@
...
+\end{itemdescr}
+
+\indexlibraryglobal{pointer_bits_available}%
+\begin{itemdecl}
+constexpr unsigned pointer_bits_available(size_t alignment);
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\constant
```
frederick-vs-ja:
```suggestion
\constantwhen
```
Ditto below.

source/memory.tex
```diff
@@ -2119,6 +2162,428 @@
...
+\indexlibraryglobal{tagged_pointer_type}%
+\begin{itemdecl}
+using tagged_pointer_type = @\seebelow@;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\tcode{tagged_pointer_type} denotes \cv{}~\tcode{void*},
+%FIXME: what do we mean by "pointer denotes"? \tcode{pointer()}?
+for \cv{} such that pointer denotes \cv{}~\tcode{U*} for some type \tcode{U}.
```
frederick-vs-ja, line 2422: IIUC the "pointer" should be `pointer_type`.

I wonder whether `\result` is suitable here. Also, IIUC, we can just use `using tagged_pointer_type = common_type_t<void*, pointer_type>;`.

eisenwave: @jwakely how do we fix this? Editorially or LWG issue, and in this PR or elsewhere?

jwakely: I don't think the intended meaning is clear. R4 and R5 had:

        using tagged_pointer_type = conditional_t<is_const_v<pointer>, const void *, void *>;

But `pointer` isn't even defined, and presumably what it was meant to mean was `is_const_v<remove_pointer_t<Ptr>>`.

I think this needs an LWG issue unless @hanickadot can clarify.

hanickadot: It's supposed to be COPYCV equivalent of this https://eel.is/c++draft/meta.trans.other#3.3

so `tagged_pointer_type` has same CV-qualifiers as `pointer_type`

hanickadot: `pointer_tag_pair<const T *, N>::tagged_pointer_type` should be `const void *`

`pointer_tag_pair<volatile T *, N>::tagged_pointer_type` should be `volatile void *`

`pointer_tag_pair<T *, N>::tagged_pointer_type` should be `void *`

hanickadot: I didn't know I can use std::common_type_t as @frederick-vs-ja proposed. 

Originally it was supposed to be an unspecified type, maybe `uintptr_t`, `(CV) void *` ... as I went thru many back and forth with Jens in lib-ext@ mailing list to find something which works, he was concerned LEWG wasn't fully aware about the effects of the functionality returning basically always an invalid pointer value. Which until recently had implementation defined behavior for lvalue-to-rvalue conversion (not anymore).

tkoeppe: OK, let's follow it up later then. Thanks!

hanickadot: Sorry, I wrote replies, there, but didn't notice I need to submit them in "Files Changes" tab 🤔

tkoeppe: I leave it to you and LWG to decide if this is editorial - if so, please send a PR and I can include it in the new draft!

PR #9121, opened 2026-06-14
[2026-06 LWG Motion 18] P2019R9 Thread attributes

### LINKED ISSUE #817: P2019 Usability improvements for std::thread
[P2019R0](https://wg21.link/p2019r0) Usability improvements for std::thread (Corentin Jabot)

### LINKED ISSUE #9105: [2026-06 LWG Motion 18]  P2019R9 (Thread attributes)
[P2019R9](https://wiki.isocpp.org/uploads/2026-06_Brno-StrawPolls_-_P2019R9.pdf) (Thread attributes)

eisenwave:
Fixes #9105
Also fixes https://github.com/cplusplus/papers/issues/817

### DIFF
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -1857,8 +1857,10 @@ of \tcode{f} will be thrown in the constructing thread, not the new thread.
 \end{note}
 If the invocation of \tcode{invoke} terminates with an uncaught exception,
 \tcode{terminate} is invoked\iref{except.terminate}.
+\begin{note}
 The parameters \tcode{attrs...}
 can affect the behavior of the new thread\iref{thread.attributes}.
+\end{note}
 
 \pnum
 \sync
@@ -2238,11 +2240,10 @@ of \tcode{f} will be thrown in the constructing thread, not the new thread.
 \end{note}
 If the \tcode{invoke} expression exits via an exception,
 \tcode{terminate} is called.
+\begin{note}
 The parameters \tcode{attrs...}
-%FIXME: why do we have iref{thread.attributes} for thread for the same wording,
-%but not for jthread?
-%While we're at it, could this not be a note?
-can affect the behavior of the new thread.
+can affect the behavior of the new thread\iref{thread.attributes}.
+\end{note}
 
 \pnum
 \sync
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -1553,8 +1553,9 @@ Implementations should not store the value of the \tcode{name_hint} attribute
 in the \tcode{thread} or \tcode{jthread} object.
 
 \recommended
-%FIXME: ordinary literal encoding? Literal encoding of T? What is "THE literal encoding"?
-\exposid{name} is interpreted as a string in the literal encoding.
+\exposid{name} is interpreted as a string
+in the associated character encoding
+of \tcode{T}~(\tref{lex.string.literal}).
 
 \indexlibraryctor{thread::name_hint}%
 \begin{itemdecl}
--- a/source/threads.tex
+++ b/source/threads.tex
@@ -1474,11 +1474,22 @@ namespace std {
   public:
     // \ref{thread.thread.id}, class \tcode{thread::id}
     class id;
+
+    // \ref{thread.attributes}, thread attributes
+    template<@\libconcept{same_as}@<char> T>
+      class name_hint;
+    template<class T>
+      name_hint(const T*) -> name_hint<T>;
+    template<class T>
+      name_hint(basic_string<T>) -> name_hint<T>;
+    class stack_size_hint;
+
     using native_handle_type = @\impdefnc@;         // see~\ref{thread.req.native}
 
     // construct/copy/destroy
     thread() noexcept;
-    template<class F, class... Args> explicit thread(F&& f, Args&&... args);
+    template<class... Args>
+      explicit thread(Args&&... args);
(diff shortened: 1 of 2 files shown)


source/threads.tex
```diff
@@ -1499,6 +1506,98 @@
...
+The \tcode{name_hint} thread attribute can be used to set the name of a thread
+for debugging purposes or platform-specific display mechanisms.
+
+\recommended
+Implementations should not store the value of the \tcode{name_hint} attribute
+in the \tcode{thread} or \tcode{jthread} object.
+
+\recommended
+%FIXME: ordinary literal encoding? Literal encoding of T? What is "THE literal encoding"?
+\exposid{name} is interpreted as a string in the literal encoding.
```
frederick-vs-ja: At this moment, I believe the literal encoding must be the ordinary literal encoding because `T` must be `char`. I wish we don't need a follow-up LWG issue for this.
```suggestion
\exposid{name} is interpreted as a string in the ordinary literal encoding.
```

source/threads.tex
```diff
@@ -1675,29 +1774,64 @@
...
 
 \begin{itemdescr}
 \pnum
 \constraints
-\tcode{remove_cvref_t<F>} is not the same type as \tcode{thread}.
+\begin{itemize}
+\item
+\tcode{Args} is not an empty pack, and
+\item
+\tcode{remove_cvref_t<F Args...[0]>} is not the same type as \tcode{thread}.
```
frederick-vs-ja:
```suggestion
\tcode{remove_cvref_t<Args...[0]>} is not the same type as \tcode{thread}.
```

source/threads.tex
```diff
@@ -2006,21 +2145,52 @@
...
-template<class F, class... Args> explicit jthread(F&& f, Args&&... args);
+template<class... Args> explicit jthread(Args&&... args);
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
 \constraints
-\tcode{remove_cvref_t<F>} is not the same type as \tcode{jthread}.
+\tcode{Args} is not an empty pack, and
+\tcode{remove_cvref_t<F Args...[0]>} is not the same type as \tcode{jthread}.
```
frederick-vs-ja:
```suggestion
\tcode{remove_cvref_t<Args...[0]>} is not the same type as \tcode{jthread}.
```

source/threads.tex
```diff
@@ -1499,6 +1506,98 @@
...
+\indexlibraryglobal{thread::name_hint}%
+\begin{codeblock}
+template<@\libconcept{same_as}@<char> T>
+class thread::name_hint {
+public:
+  constexpr explicit name_hint(basic_string_view<T> n) noexcept;
+  name_hint(name_hint&&) = delete;
+  name_hint(const name_hint&) = delete;
+private:
+  basic_string_view<T> @\exposid{name}@; // \expos
```
frederick-vs-ja: Maybe we should rename this to _`name_`_, possibly in a follow-up commit or PR. Same for (_`size`_ ->) _`size_`_.

source/threads.tex
eisenwave: At this point it cannot be anything else, yeah. I'm wondering in which direction we want to resolve this though, and that depends on how likely support for more character types is in the future.

If we say "ordinary literal encoding" now, we will need to change that wording if more support is added.

eisenwave:
@jwakely some LWG input needed here.

source/threads.tex
eisenwave: I'd do it in a follow-up PR if anything.

source/threads.tex
cor3ntin: ordinary literal encoding works as a stopgap. or `associated character encoding of T [ table 12)` - There is precedent here https://eel.is/c++draft/format.string.escaped#2 

(Why did no one pinged me?)

source/threads.tex
```diff
@@ -2045,6 +2221,11 @@
 \end{note}
 If the \tcode{invoke} expression exits via an exception,
 \tcode{terminate} is called.
+The parameters \tcode{attrs...}
+%FIXME: why do we have iref{thread.attributes} for thread for the same wording,
+%but not for jthread?
+%While we're at it, could this not be a note?
```
tkoeppe: @jwakely Do you have any thoughts on this? It does feel notish.

source/threads.tex
tkoeppe: We don't have an overall style for expos-only members, and this is subject to some debate. Leave it as is for now.

source/threads.tex
tkoeppe: @jwakely Is it editorial to use the "associated character encoding" wording here?

jwakely: Yes, that seems editorial. We all agree that it can't be anything else. We need an xref to Table 12 though.

source/threads.tex
jwakely: Yeah making this a note seems good.

source/threads.tex
tkoeppe: Added as a separate commit!

source/threads.tex
jwakely: In both places, and might as well add the iref here too.

tkoeppe: Done!

tkoeppe: Done!

PR #9125, opened 2026-06-14
[2026-06 LWG Motion 5] P4206R0 Revert string support in `std::constant_wrapper`

### LINKED ISSUE #2770: P4206 R0 Revert string support in std::constant_wrapper
[P4206R0](https://wg21.link/p4206r0) Revert string support in std::constant_wrapper (Barry Revzin, Zach Laine, Matthias Kretz, Jonathan Wakely, Tomasz Kamiński)

### LINKED ISSUE #9092: [2026-06 LWG Motion 5] P4206R0 (Revert string support in std::constant_wrapper)
[P4206R0](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2026/p4206r0.html) (Revert string support in std::constant_wrapper)

eisenwave:
Fixes #9092
Also fixes https://github.com/cplusplus/papers/issues/2770

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -98,10 +98,7 @@ namespace std {
   void observable_checkpoint() noexcept;
 
   // \ref{const.wrap.class}, class template \tcode{constant_wrapper}
-  template<class T>
-    struct @\exposidnc{cw-fixed-value}@;              // \expos
-
-  template<@\exposidnc{cw-fixed-value}@ X, class = typename decltype(X)::@\exposid{type}@>
+  template<auto X, class = decltype(X)>
     struct constant_wrapper;
 
   template<class T>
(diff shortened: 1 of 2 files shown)


source/utilities.tex
```diff
@@ -914,11 +894,11 @@
         -> constant_wrapper<(T::value >>= R::value)> { return {}; }
   };
 
-  template<@\exposid{cw-fixed-value}@ X, class>
+  template<auto X, class T>
```
jwakely: We usually repeat the default template argument here, even though it was shown in the header synopsis already (e.g. the `allocator<T>` default template argument is present in both [vector.syn] and the class synopsis in [vector.overview]). The incoming paper gets this wrong. I don't care whether we fix that here, or editorially later.

jwakely (approved):

source/utilities.tex
tkoeppe: I added it here to save us the round trip.

PR #9127, opened 2026-06-14
[2026-06 LWG Motion 3] P3798R1 The unexpected in std::expected

### LINKED ISSUE #2403: P3798 R1 The unexpected in std::expected
[P3798R0](https://wg21.link/p3798r0) The unexpected in std::expected (Alex Kremer, Ayaz Salikhov)

### LINKED ISSUE #9090: [2026-06 LWG Motion 3] P3798R1 The unexpected in std::expected
[P3798R1](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p3798r1.html) (The unexpected in std::expected)

eisenwave:
Fixes #9090
Also fixes https://github.com/cplusplus/papers/issues/2403

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -8525,6 +8525,7 @@ namespace std {
     constexpr T&& operator*() && noexcept;
     constexpr explicit operator bool() const noexcept;
     constexpr bool has_value() const noexcept;
+    constexpr bool has_error() const noexcept;
     constexpr const T& value() const &;                                     // freestanding-deleted
     constexpr T& value() &;                                                 // freestanding-deleted
     constexpr const T&& value() const &&;                                   // freestanding-deleted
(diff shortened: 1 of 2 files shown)


source/utilities.tex
```diff
@@ -9417,6 +9418,20 @@
...
+constexpr bool has_error() const noexcept;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\effects
+Equivalent to:
+\begin{codeblock}
+return !has_value();
+\end{codeblock}
```
jwakely: I don't think this needs to be a codeblock, does it?

```suggestion
\tcode{return !has_value();}
```

(Also I wonder why we didn't ask for it to be just _Returns_: `!has_value()` since there are no preconditions or other special effects of `has_value()` that we need to inherit here.

source/utilities.tex
```diff
@@ -10445,6 +10461,20 @@
...
```
```suggestion
\tcode{return !has_value();}
```

jwakely (approved):
I've asked LWG about the _Effects_ vs _Returns_ question, but that can be resolved separately. This pull request matches the incoming paper, thanks.

PR #9129, opened 2026-06-14
[2026-06 LWG Motion 6] P3395R6 (Fix encoding issues and add a formatter for std::error_code)

### LINKED ISSUE #2187: P3395 R0 Formatting of std::error_code
[P3395R0](https://wg21.link/p3395r0) Formatting of std::error_code (Victor Zverovich)

### LINKED ISSUE #9093: [2026-06 LWG Motion 6] P3395R6 (Fix encoding issues and add a formatter for std::error_code)
[P3395R6](https://wiki.isocpp.org/uploads/2026-06_Brno-StrawPolls_-_P3395-error_code_formatting.html) (Fix encoding issues and add a formatter for std::error_code)

notdanhan:
Fixes: #9093 
Also Fixes https://github.com/cplusplus/papers/issues/2187

### DIFF
--- a/source/diagnostics.tex
+++ b/source/diagnostics.tex
@@ -764,6 +764,9 @@ namespace std {
     constexpr bool @\libglobal{is_error_code_enum_v}@ = is_error_code_enum<T>::value;
   template<class T>
     constexpr bool @\libglobal{is_error_condition_enum_v}@ = is_error_condition_enum<T>::value;
+
+  // \ref{syserr.fmt}, formatter
+  template<class T> struct formatter<error_code, charT>;
 }
 \end{codeblock}
 
@@ -836,7 +839,7 @@ virtual const char* name() const noexcept = 0;
 \begin{itemdescr}
 \pnum
 \returns
-A string naming the error category.
+A string in the ordinary literal encoding naming the error category.
 \end{itemdescr}
 
 \indexlibrarymember{default_error_condition}{error_category}%
@@ -880,7 +883,8 @@ virtual string message(int ev) const = 0;
 \begin{itemdescr}
 \pnum
 \returns
-A string that describes the error condition denoted by \tcode{ev}.
+A string of multibyte characters in the execution character set
+that describes the error condition denoted by \tcode{ev}.
 \end{itemdescr}
 
 \rSec3[syserr.errcat.nonvirtuals]{Non-virtual members}
@@ -1233,6 +1237,98 @@ template<class charT, class traits>
 \pnum
 \effects
 Equivalent to: \tcode{return os << ec.category().name() << ':' << ec.value();}
+\end{itemdescr}
+
+\rSec3[syserr.fmt]{Formatting}
+
+\pnum
+\indexlibraryglobal{formatter}%
+\begin{codeblock}
+template<class charT>
+struct formatter<error_code, charT> {
+  constexpr void set_debug_format();
+
+  constexpr typename basic_format_parse_context<charT>::iterator
+    parse(basic_format_parse_context<charT>& ctx);
+
+  template<class Out>
+    typename basic_format_context<Out, charT>::iterator
+      format(const error_code& ec, basic_format_context<Out, charT>& ctx) const;
+};
+\end{codeblock}
+
+\indexlibrarymember{set_debug_format}{formatter}%
+\begin{itemdecl}
+constexpr void set_debug_format();
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\effects
+Modifies the state of the \tcode{formatter} to be as if the
+\fmtgrammarterm{error-code-format-spec} parsed by the
+last call to \tcode{parse} contained the \tcode{?} option.
+\end{itemdescr}
+
+\indexlibrarymember{parse}{formatter}%
+\begin{itemdecl}
+constexpr typename basic_format_parse_context<charT>::iterator
+  parse(basic_format_parse_context<charT>& ctx);
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\effects
+Parses the format specifier as an \fmtgrammarterm{error-code-format-spec}
+and stores the parsed specifiers in \tcode{*this}.
+
+\begin{ncbnf}
+\fmtnontermdef{error-code-format-spec}\br
+    \opt{fill-and-align} \opt{width} \opt{\tcode{?}} \opt{\tcode{s}}
+\end{ncbnf}
+
+where the productions \fmtgrammarterm{fill-and-align} and \fmtgrammarterm{width} are
+described in \ref{format.string}.
+
+\pnum
+\returns
+An iterator past the end of the \fmtgrammarterm{error-code-format-spec}.
+\end{itemdescr}
+
+\indexlibrarymember{format}{formatter}%
+\begin{itemdecl}
+template<class Out>
+  typename basic_format_context<Out, charT>::iterator
+    format(const error_code& ec, basic_format_context<Out, charT>& ctx) const;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\effects
+If the \tcode{s} option is used, then:
+\begin{itemize}
+\item
+If \tcode{charT} is \tcode{char} and the ordinary literal encoding is UTF-8,
+then let \tcode{msg} be \tcode{ec.message()} transcoded to UTF-8
+with maximal subparts of ill-formed subsequences
+substituted with \unicode{fffd}{replacement character}
+per the Unicode Standard, Chapter 3.9 \unicode{fffd} Substi\-tution in Conversion.
+\item
+Otherwise, let \tcode{msg} be \tcode{ec.message()}
+transcoded to an
+\impldef{encoding of a error message during formatting}
+encoding.
+\end{itemize}
+Otherwise, let \tcode{msg} be
+\tcode{format("\{\}:\{\}", ec.category().name(), ec.value())}.
+If the \tcode{?} option is used then \tcode{msg} is formatted
+as an escaped string\iref{format.string.escaped}.
+Writes \tcode{msg} into \tcode{ctx.out()},
+adjusted according to the \fmtgrammarterm{error-code-format-spec}.
+
+\pnum
+\returns
+An iterator past the end of the output range.
 \end{itemdescr}
 
 
--- a/source/diagnostics.tex
+++ b/source/diagnostics.tex
@@ -1281,14 +1281,13 @@ constexpr typename basic_format_parse_context<charT>::iterator
 \effects
 Parses the format specifier as an \fmtgrammarterm{error-code-format-spec}
 and stores the parsed specifiers in \tcode{*this}.
-
+The syntax of format specifications is
 \begin{ncbnf}
 \fmtnontermdef{error-code-format-spec}\br
-    \opt{fill-and-align} \opt{width} \opt{\tcode{?}} \opt{\tcode{s}}
+    \opt{fill-and-align} \opt{width} \opt{\terminal{?}} \opt{\terminal{s}}
 \end{ncbnf}
-
-where the productions \fmtgrammarterm{fill-and-align} and \fmtgrammarterm{width} are
-described in \ref{format.string}.
+where the productions \fmtgrammarterm{fill-and-align} and \fmtgrammarterm{width}
+are described in \ref{format.string}.
 
 \pnum
 \returns


source/diagnostics.tex
```diff
@@ -1235,6 +1239,89 @@
 Equivalent to: \tcode{return os << ec.category().name() << ':' << ec.value();}
 \end{itemdescr}
 
+\rSec3[system.error.fmt]{Formatting}
```
eisenwave: It's not your fault, but this subclause stable name makes no sense. Everything else is called [syserr.*] except for [system.error.syn].

We probably want something like [syserr.fmt] or whatever the convention is for "formatting support" stable names.

source/diagnostics.tex
```diff
@@ -1235,6 +1239,89 @@
...
+\pnum
+\begin{codeblock}
+template<class charT> struct formatter<error_code, charT> {
```
eisenwave: Missing indexing.

source/diagnostics.tex
```diff
@@ -1235,6 +1239,89 @@
...
+  constexpr typename basic_format_parse_context<charT>::iterator
+    parse(basic_format_parse_context<charT>& ctx);
+
+  template<class Out>
+    typename basic_format_context<Out, charT>::iterator
+      format(const error_code& ec, basic_format_context<Out, charT>& ctx) const;
+};
+\end{codeblock}
+
+\pnum
```
eisenwave:
```suggestion
```

source/diagnostics.tex
```diff
@@ -1235,6 +1239,89 @@
...
+\begin{itemdecl}
+constexpr void set_debug_format();
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\effects
+Modifies the state of the \tcode{formatter} to be as if the
+\defnx{error-code-format-spec}{error-code-format-spec!syserr err fmt} passed by the
```
eisenwave:
```suggestion
\grammarterm{error-code-format-spec} passed by the
```
The paper italics are grammarterm italics, not defining.

source/diagnostics.tex
```diff
@@ -1235,6 +1239,89 @@
...
+last call to \tcode{parse} contained the \tcode{?} option.
+\end{itemdescr}
+
+\pnum
```
eisenwave:
```suggestion
```

source/diagnostics.tex
```diff
@@ -1235,6 +1239,89 @@
...
+\effects
+Parses the format specifier as a \defnx{error-code-format-spec}{error-code-format-spec!syserr err fmt}
+and stores the parsed specifiers in \tcode{*this}.
+
+\begin{ncbnf}
+\fmtnontermdef{error-code-format-spec}\br
+    \opt{fill-and-align} \opt{width} \opt{\tcode{?}} \opt{\tcode{s}}
+\end{ncbnf}
+
+where the productions and \defnx{width}{width!format string} are described in\iref{format.string}
```
eisenwave:
```suggestion
where the productions \grammarterm{fill-and-align} and \grammarterm{width} are described in\iref{format.string}.
```

source/diagnostics.tex
```diff
@@ -1235,6 +1239,89 @@
...
+\begin{itemdescr}
+\pnum
+\effects
+If the \tcode{s} option is used, then:
+\begin{itemize}
+\item If \tcode{charT} is \tcode{char} and the ordinary literal encoding is UTF-8, let
+  \tcode{msg} be \tcode{ec.message()} transcoded to UTF-8 with maximal subparts of ill-formed
+  subsequences substituted with \unicode{fffd}{replacement character} per the Unicode Standard, Chapter
+  3.9 \unicode{fffd} Substitution in Conversion.
+\item Otherwise, let \tcode{msg} be \tcode{ec.message()} transcoded to an implementation-defined encoding.
```
eisenwave: Needs some reflowing:
```suggestion
\item
If \tcode{charT} is \tcode{char} and the ordinary literal encoding is UTF-8,
let \tcode{msg} be \tcode{ec.message()} transcoded to UTF-8
with maximal subparts of ill-formed subsequences substituted with \unicode{fffd}{replacement character}
per the Unicode Standard, Chapter 3.9 \unicode{fffd} Substitution in Conversion.
\item
Otherwise, let \tcode{msg} be \tcode{ec.message()}
transcoded to an implementation-defined encoding.
```

eisenwave (changes requested):
Needs a fair amount of work, mainly converting to `\grammarterm`.

source/diagnostics.tex
```diff
@@ -1235,6 +1239,94 @@
...
+Parses the format specifier as a \fmtgrammarterm{error-code-format-spec}
+and stores the parsed specifiers in \tcode{*this}.
+
+\begin{ncbnf}
+\fmtnontermdef{error-code-format-spec}\br
+    \opt{fill-and-align} \opt{width} \opt{\tcode{?}} \opt{\tcode{s}}
+\end{ncbnf}
+
+where the productions and \fmtgrammarterm{fill-and-align} and \fmtgrammarterm{width} are
+described in\iref{format.string}.
```
eisenwave:
```suggestion
described in \ref{format.string}.
```

source/diagnostics.tex
```diff
@@ -1235,6 +1239,94 @@
...
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\effects
+If the \tcode{s} option is used, then:
+\begin{itemize}
+\item
+If \tcode{charT} is \tcode{char} and the ordinary literal encoding is UTF-8,
+let \tcode{msg} be \tcode{ec.message()} transcoded to UTF-8
```
eisenwave:
```suggestion
then let \tcode{msg} be \tcode{ec.message()} transcoded to UTF-8
```

source/diagnostics.tex
```diff
@@ -1235,6 +1239,94 @@
...
+Otherwise, let \tcode{msg} be \tcode{ec.message()}
+transcoded to an implementation-defined encoding.
+\end{itemize}
+
+\pnum
+Otherwise, let \tcode{msg} be \tcode{std::format("{}:{}", ec.category().name(), ec.value())}.
+
+\pnum
+If the \tcode{?} option is used then \tcode{msg} is formatted as an escaped string(\iref{format.string.escaped}).
+Writes \tcode{msg} into \tcode{ctx.out()}, adjusted according to the \fmtgrammarterm{error-code-format-spec}.
```
eisenwave: This all seems to be a scription of *Effects* to me, not separate paragraphs, but maybe I'm missing something. The paper does put gaps here, but idk, this is weird.
```suggestion
Otherwise, let \tcode{msg} be \tcode{std::format("{}:{}", ec.category().name(), ec.value())}.
If the \tcode{?} option is used then \tcode{msg} is formatted as an escaped string(\iref{format.string.escaped}).
Writes \tcode{msg} into \tcode{ctx.out()}, adjusted according to the \fmtgrammarterm{error-code-format-spec}.
```

source/diagnostics.tex
```diff
@@ -1235,6 +1239,94 @@
...
+\item
+Otherwise, let \tcode{msg} be \tcode{ec.message()}
+transcoded to an implementation-defined encoding.
+\end{itemize}
+
+\pnum
+Otherwise, let \tcode{msg} be \tcode{std::format("{}:{}", ec.category().name(), ec.value())}.
+
+\pnum
+If the \tcode{?} option is used then \tcode{msg} is formatted as an escaped string(\iref{format.string.escaped}).
```
eisenwave:
```suggestion
If the \tcode{?} option is used
then \tcode{msg} is formatted as an escaped string\iref{format.string.escaped}.
```

eisenwave (changes requested):
Some things are clearly wrong, but I could use some advice from @jensmaurer on the multi-paragraph `\effects` thing.

source/diagnostics.tex
jensmaurer: I think we do have multi-paragraph \effects in some places, but having a separate _numbered_ paragraph for every sentence is certainly non-optimal.

notdanhan: I have an idea

notdanhan: image
I think this works?

source/diagnostics.tex
```diff
@@ -880,7 +883,8 @@
 \begin{itemdescr}
 \pnum
 \returns
-A string that describes the error condition denoted by \tcode{ev}.
+A string of multibyte characters in the execution character set that describes the error condition
+denoted by \tcode{ev}.
```
jwakely:
```suggestion
A string of multibyte characters in the execution character set
that describes the error condition denoted by \tcode{ev}.
```

source/diagnostics.tex
```diff
@@ -1235,6 +1239,98 @@
...
+\indexlibrarymember{set_debug_format}{formatter}%
+\begin{itemdecl}
+constexpr void set_debug_format();
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\effects
+Modifies the state of the \tcode{formatter} to be as if the
+\fmtgrammarterm{error-code-format-spec} passed by the
```
jwakely:
```suggestion
\fmtgrammarterm{error-code-format-spec} parsed by the
```

source/diagnostics.tex
```diff
@@ -1235,6 +1239,98 @@
...
+\indexlibrarymember{parse}{formatter}%
+\begin{itemdecl}
+constexpr typename basic_format_parse_context<charT>::iterator
+  parse(basic_format_parse_context<charT>& ctx);
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\effects
+Parses the format specifier as a \fmtgrammarterm{error-code-format-spec}
```
jwakely:
```suggestion
Parses the format specifier as an \fmtgrammarterm{error-code-format-spec}
```

(this is wrong in the incoming paper, but we should fix it)

source/diagnostics.tex
```diff
@@ -1235,6 +1239,98 @@
...
+and stores the parsed specifiers in \tcode{*this}.
+
+\begin{ncbnf}
+\fmtnontermdef{error-code-format-spec}\br
+    \opt{fill-and-align} \opt{width} \opt{\tcode{?}} \opt{\tcode{s}}
+\end{ncbnf}
+
+where the productions and \fmtgrammarterm{fill-and-align} and \fmtgrammarterm{width} are
```
jwakely:
```suggestion
where the productions \fmtgrammarterm{fill-and-align} and \fmtgrammarterm{width} are
```

source/diagnostics.tex
```diff
@@ -1235,6 +1239,98 @@
...
+Otherwise, let \tcode{msg} be \tcode{ec.message()}
+transcoded to an implementation-defined encoding.
+\end{itemize}
+
+\item
+Otherwise, let \tcode{msg} be \tcode{std::format("{}:{}", ec.category().name(), ec.value())}.
+
+\item
+If the \tcode{?} option is used
+then \tcode{msg} is formatted as an escaped string(\iref{format.string.escaped}).
```
jwakely:
```suggestion
then \tcode{msg} is formatted as an escaped string\iref{format.string.escaped}.
```

jwakely (changes requested):

frederick-vs-ja:
Do we want a feature-test macro for `formatter<error_code, charT>` specialization?
CC @jwakely @brevzin @StephanTLavavej.

jwakely:
> Do we want a feature-test macro for `formatter<error_code, charT>` specialization? CC @jwakely @brevzin @StephanTLavavej.

Barry pinged me about this last week, suggesting that user code could polyfill `formatter<error_code, charT>` if the library doesn't support it yet. You can't legally provide exactly that specialization (it doesn't depend on a program-defined type) but you could do:

```
#ifdef __cpp_lib_format_error_code
  log("oh noes: {}", ec);
#else
  log("oh noes: {}:{}", ec.category().name(), ec.value());
#endif
```

Is that useful though? If you have to write the fallback version, just use the fallback version.

We do have `__cpp_lib_format_path` for `formatter<filesystem::path, charT>` though, or we could bump `__cpp_lib_formatters` (why didn't we do that for `path`? Worry about an ordering dependency?)

source/diagnostics.tex
```diff
@@ -1235,6 +1239,98 @@
...
+then let \tcode{msg} be \tcode{ec.message()} transcoded to UTF-8
+with maximal subparts of ill-formed subsequences substituted with \unicode{fffd}{replacement character}
+per the Unicode Standard, Chapter 3.9 \unicode{fffd} Substi\-tution in Conversion.
+\item
+Otherwise, let \tcode{msg} be \tcode{ec.message()}
+transcoded to an implementation-defined encoding.
+\end{itemize}
+
+\item
+Otherwise, let \tcode{msg} be \tcode{std::format("{}:{}", ec.category().name(), ec.value())}.
```
jwakely: No `std::` here and you need to escape the `{}` replacement fields or LaTeX eats them.

source/diagnostics.tex
jwakely: I don't love the use of a list with no other text before or after the list. If we went this form I think the final sentence should be its own bullet. "Writes `msg` into `ctx.out()`" applies to all cases, not only when `?` is used.

We could just make it all one paragraph with no paragraph breaks, but keep the innermost list for the sub-clauses of the "If the `s` options is used" case. So exactly as in the incoming paper, except without a paragraph break before "If the `?` option is used".

StephanTLavavej:
I'm in favor of a new feature-test macro. Bumping a single macro gets really annoying when multiple features are "in flight"; it's more reasonable for a once-per-Standard thing.

source/diagnostics.tex
tkoeppe: That seems entirely adequate, thanks!

jwakely:
We have LWG4605 for the macro.

PR #9138, opened 2026-06-16
[temp.arg.template] Update compatibility rules to include template template parameters

### LINKED ISSUE #9054: compatible does not cover template template parameters after CWG3080
[[temp.arg.template]/1](https://eel.is/c++draft/temp.arg.template#1) now lets a template template parameter be used as a template template argument, but the compatibility bullets in [/3](https://eel.is/c++draft/temp.arg.template#3) were never updated to match - they still only cover an argument denoting a class/alias template, variable template, or concept.

So this ordinary forwarding case has no wording making it well-formed:

```cpp
template<template<class> class TT>
struct S { };

template<template<class> class UU>
void f() {
    S<UU> s;
}
```

`UU` is permitted by /1, but /3 never makes it compatible with `TT`, so the /4 non-pack match cannot succeed. GCC, Clang, MSVC, and EDG all accept it, so this looks like a missed follow-up to [CWG3080](https://wg21.link/cwg3080) / [P2841R7](https://wg21.link/p2841r7) - the fix is just to add the template-template-parameter alternatives to the three bullets in /3, mirroring /1.

abhinavagarwal07:
Fixes the compatibility gap between paragraph 1 and paragraph 3 in [temp.arg.template].

Closes #9054

### DIFF
--- a/source/templates.tex
+++ b/source/templates.tex
@@ -1539,13 +1539,13 @@ a \grammarterm{template-argument} \tcode{A} are
 if
 \begin{itemize}
 \item
-\tcode{A} denotes a class template or an alias template and
+\tcode{A} denotes a class template, an alias template, or a type template template parameter and
 \tcode{P} is a type template template parameter,
 \item
-\tcode{A} denotes a variable template and
+\tcode{A} denotes a variable template or a variable template template parameter and
 \tcode{P} is a variable template template parameter, or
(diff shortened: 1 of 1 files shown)


abhinavagarwal07:
Filed CWG https://github.com/cplusplus/CWG/issues/923

PR #9139, opened 2026-06-16
[mdspan.sub.helpers] Fix swapped offset/extent in canonical-slice

abhinavagarwal07:
The `extent_slice` branch of `canonical-slice` ([mdspan.sub.helpers]) initializes `.offset` from `s.extent` and `.extent` from `s.offset`, reversing the intended mapping. This is inconsistent with `canonical-range-slice` and the pre-P3982R2 wording, both of which map offset → offset and extent → extent.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -25946,8 +25946,8 @@ if constexpr (is_convertible_v<S, full_extent_t>) {
   return @\exposid{canonical-index}@<IndexType>(std::move(s));
 } else if constexpr (@\exposconcept{is-extent-slice}@<S>) {
   return extent_slice{
-    .offset = @\exposid{canonical-index}@<IndexType>(std::move(s.extent)),
-    .extent = @\exposid{canonical-index}@<IndexType>(std::move(s.offset)),
+    .offset = @\exposid{canonical-index}@<IndexType>(std::move(s.offset)),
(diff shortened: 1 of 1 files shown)


eisenwave:
I'm not sure whether this is editorial. https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2026/p3982r2.html does have this problem.

Does this require an LWG issue or can we treat this as an obvious typo @jwakely

abhinavagarwal07:
Happy to file LWG if its more appropriate.

jwakely:
> Problems with the presentation of the specification, which, when fixed, would not alter the (intended) meaning of the specification. Typos and obvious inconsistencies fall into this category.

https://github.com/cplusplus/draft/wiki/How-to-tell-if-an-issue-is-editorial

This seems clearly editorial, the intended meaning is to initialize the offset with the offset and the extent with the extent.

jwakely (approved):

jensmaurer (approved):

tomaszkam:
That was a typo, libstdc++ implementation implemented the PR: https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=libstdc%2B%2B-v3/include/std/mdspan;h=f5556f35fa1497f2bbd1c4debd822841ec5f05c3;hb=HEAD#l3356. That was intent from the start.

PR #9141, opened 2026-06-16
[dcl.init.ref] Add missing comma, turn into sub-bullets

Tsche:


eisenwave:
The comma seems grammatically necessary and improves readability.

tkoeppe:
Hm, but the new comma makes it more ambiguous how the top-level ", or" chain relates to the individual list item. I don't find the status quo particularly unclear. Which misreading are you worried about?

Tsche:
The confusion was whether this is to be read as
> **(** is an rvalue (but not a bit-field) or an lvalue of function type **)** and “cv1 T1” is reference-compatible with “cv2 T2”

or
> is an rvalue (but not a bit-field) or **(** an lvalue of function type and “cv1 T1” is reference-compatible with “cv2 T2” **)**

@t3nsor suggested the extra comma to disambiguate. I think it's slightly clearer, but I see your concern with the ", or". Unfortunately I see no easy way to turn this into a proper bulleted list rather than a run-on sentence either.

Tsche:
We could turn this into two separate bullets 
> - is an rvalue (but not a bit-field) and “cv1 T1” is reference-compatible with “cv2 T2”, or
> - is an lvalue of function type and “cv1 T1” is reference-compatible with “cv2 T2”, or

but that seems strictly worse to me. I wouldn't want to repeat "and “cv1 T1” is reference-compatible with “cv2 T2”" twice.

tkoeppe:
Yes, that was what I suspected. Is this really a point of confusion, or just an academic exercise in grammar parsing? I thought it was quite clear, and if we had wanted the "and reference-compatible" part to only apply to the second option, we would have said it more elaborately.

If we must, we could entertain sub-bullets like this:

> * is either
>   * an rvalue (but not a bitfield) or
>   * an lvalue of function type, and
> * [CONTINUING SAME BULLET] cv1 etc etc reference compatible with cv2 T2", or
> * [...]

(I can't present this in markdown; there are only meant to be two top-level bullets here, not three.)

But I think I'd like to hear from at least one other person who is genuinely uncertain of the reading.

jensmaurer:
I like the additional level of bullets, which is both structurally clearer and allows the comma.

Tsche:
I've updated it to the sub-bullets variant. The CI failure on MacOS seems unrelated.

jensmaurer (approved):

PR #9142, opened 2026-06-16
[utility.arg.requirements] reformat in a style like [container.reqmts]

notdanhan:
It looks like this now

image

Additional work was needed to change some `\tref{}` to `\iref{}` across the document, along with two reflows of paragraphs to stop hbox badness

I did some mild rewording in source/lib-intro.tex to make it make sense in the new format. I hope to have not introduced any wording defects! :)

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -262,8 +262,8 @@ The library specification uses a typographical convention for naming
 requirements. Names in \textit{italic} type that begin with the prefix
 \oldconcept{} refer to sets of well-defined expression requirements typically
 presented in tabular form, possibly with additional prose semantic requirements.
-For example, \oldconcept{Destructible}~(\tref{cpp17.destructible}) is such a named
-requirement. Names in \tcode{constant width} type refer to library concepts
+For example, \oldconcept{Destructible}\iref{cpp17.destructible} is such a named requirement.
+Names in \tcode{constant width} type refer to library concepts
 which are presented as a concept definition\iref{temp}, possibly with additional
 prose semantic requirements. For example,
 \libconcept{destructible}\iref{concept.destructible}
@@ -1800,11 +1800,12 @@ allocators.
 
 \rSec3[utility.arg.requirements]{Template argument requirements}
 
+\rSec4[utility.arg.requirements.general]{General}
+
 \pnum
 The template definitions in the \Cpp{} standard library
-refer to various named requirements whose details are set out in
-Tables~\ref{tab:cpp17.equalitycomparable}--\ref{tab:cpp17.destructible}.
-In these tables,
+refer to various named requirements whose details are set out below.
+For each requirement in \ref{utility.arg.requirements}:
 \begin{itemize}
 \item
 \tcode{T} denotes an object or reference type to be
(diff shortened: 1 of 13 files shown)


source/lib-intro.tex
```diff
@@ -1824,105 +1825,217 @@
...
+A type \tcode{T} is \oldconcept{EqualityComparable} if
+the following constraints are met:
+\begin{itemdecl}
+a == b
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
+\result
+\tcode{decltype(a == b)} models \exposconceptx{boolean-testa\-ble}{boolean-testable}
```
frederick-vs-ja, line 1840: Do we need to say `decltype(expr)` in _Result_? I think _Result_ is already specifying the result type and value category of the expression in `decltype`-style.

notdanhan: I'm actually not sure, should I just keep the exposition concept here?

tkoeppe:
This is awesome, thank you very much, @notdanhan!

source/lib-intro.tex
```diff
@@ -1824,105 +1825,228 @@
...
-\hdstyle{Expression}  &   \hdstyle{Return type} &   \rhdr{Requirement} \\ \capsep
-\tcode{a == b}  &
-\tcode{decltype(a == b)} models \exposconceptx{boolean-testa\-ble}{boolean-testable} &
-\tcode{==} is an equivalence relation,
-that is, it has the following properties:
+\rSec4[cpp17.equalitycomparable]{\defnoldconcept{EqualityComparable} requirements}
+
+\pnum
+A type \tcode{T} is \oldconcept{EqualityComparable} if
+the following constraints are met:
```
tkoeppe: Could we use phrasing analogous to the one at https://eel.is/c++draft/container.reqmts#1? I think "constraints" isn't quite the right term here.

notdanhan: @tkoeppe I don't have access to google3/piper :(
Is there another place I can see as reference?

tkoeppe: @notdanhan Whoops, sorry, I mispasted the URL! Fixed.

source/lib-intro.tex
```diff
@@ -262,8 +262,8 @@
 requirements. Names in \textit{italic} type that begin with the prefix
 \oldconcept{} refer to sets of well-defined expression requirements typically
 presented in tabular form, possibly with additional prose semantic requirements.
-For example, \oldconcept{Destructible}~(\tref{cpp17.destructible}) is such a named
-requirement. Names in \tcode{constant width} type refer to library concepts
+For example, \oldconcept{Destructible} is such a named requirement.
```
tkoeppe: Why did we lose the cross-reference here?

source/lib-intro.tex
```diff
@@ -1800,11 +1800,12 @@
...
 
+\rSec4[utility.arg.requirements.general]{General}
+
 \pnum
 The template definitions in the \Cpp{} standard library
-refer to various named requirements whose details are set out in
-Tables~\ref{tab:cpp17.equalitycomparable}--\ref{tab:cpp17.destructible}.
-In these tables,
+refer to various named requirements whose details are set out below.
+For each requirement, note the following:
```
tkoeppe: Analogous to [container.intro.reqmts], how about:

"For each requirement in \ref{utility.arg.requirements}:"

source/threads.tex
```diff
@@ -4573,7 +4347,8 @@
 \begin{note}
 If the pointers point to different complete objects (or subobjects thereof),
 the \tcode{<} operator does not establish
-a strict weak ordering~(\tref{cpp17.lessthancomparable}, \ref{expr.rel}).
+a strict weak ordering
+(\ref{cpp17.lessthancomparable}, \ref{expr.rel}).
```
tkoeppe, line 4576: This can use multi-argument `\iref` I think.

source/threads.tex
```diff
@@ -6154,7 +5906,7 @@
 \begin{note}
 If the pointers point to different complete objects (or subobjects thereof),
 the \tcode{<} operator does not establish a strict weak ordering
-(\tref{cpp17.lessthancomparable}, \ref{expr.rel}).
+(\ref{cpp17.lessthancomparable}, \ref{expr.rel}).
```
tkoeppe: Likewise here, a single `\iref` should be able to handle this.

source/threads.tex
```diff
@@ -6208,7 +5960,7 @@
 \begin{note}
 If the pointers point to different complete objects (or subobjects thereof),
 the \tcode{<} operator does not establish
-a strict weak ordering~(\tref{cpp17.lessthancomparable}, \ref{expr.rel}).
+a strict weak ordering~(\iref{cpp17.lessthancomparable}, \ref{expr.rel}).
```
tkoeppe: ditto

source/numerics.tex
```diff
@@ -2486,10 +2486,10 @@
 
 \pnum
 \tcode{P} shall meet the
-\oldconcept{CopyConstructible} (\tref{cpp17.copyconstructible}),
-\oldconcept{CopyAssignable} (\tref{cpp17.copyassignable}),
+\oldconcept{CopyConstructible}\iref{cpp17.copyconstructible},
+\oldconcept{CopyAssignable}\iref{cpp17.copyassignable},
 and
-\oldconcept{Equality\-Comp\-arable} (\tref{cpp17.equalitycomparable}) requirements.
+\oldconcept{Equality\-Comp\-arable}\iref{cpp17.equalitycomparable} requirements.
```
tkoeppe: Drive-by: while you're here, could you have a look if we can get rid of those hyphenation hints now?

source/time.tex
```diff
@@ -1038,8 +1037,8 @@
 
 \item
 the types \tcode{TC::rep}, \tcode{TC::duration}, and \tcode{TC::time_point}
-meet the \oldconcept{EqualityComparable} (\tref{cpp17.equalitycomparable}) and
-\oldconcept{LessThanComparable} (\tref{cpp17.lessthancomparable})
+meet the\\ \oldconcept{EqualityComparable}\iref{cpp17.equalitycomparable} and
```
tkoeppe: instead of hard line breaks, try adding hyphenation hints first.

source/utilities.tex
```diff
@@ -16272,9 +16272,9 @@
 An enabled specialization \tcode{hash<Key>} will:
 \begin{itemize}
 \item meet the \oldconcept{Hash} requirements (\tref{cpp17.hash}),
-with \tcode{Key} as the function
-call argument type, the \oldconcept{Default\-Constructible} requirements (\tref{cpp17.defaultconstructible}),
-the \oldconcept{CopyAssignable} requirements (\tref{cpp17.copyassignable}),
+with \tcode{Key} as the function call argument type,
+the \oldconcept{Default\-Constructible} requirements\iref{cpp17.defaultconstructible}, \\
```
tkoeppe: similarly here, see if you can work with hyphenation hints instead

tkoeppe:
@notdanhan do also feel free to sqash, rebase and force-submit at some point. Having the full edit history isn't so important at this point.

source/lib-intro.tex
jwakely: Yes please, I don't like "note the following".

source/lib-intro.tex
```diff
@@ -1824,105 +1825,229 @@
...
-\hdstyle{Expression}  &   \hdstyle{Return type} &   \rhdr{Requirement} \\ \capsep
-\tcode{a == b}  &
-\tcode{decltype(a == b)} models \exposconceptx{boolean-testa\-ble}{boolean-testable} &
-\tcode{==} is an equivalence relation,
-that is, it has the following properties:
+\rSec4[cpp17.equalitycomparable]{\defnoldconcept{EqualityComparable} requirements}
+
+\pnum
+A type \tcode{T} meets the \oldconcept{EqualityComparable} requirements if the following
+statements and expressions are well-formed and have the specified semantics:
```
jwakely: Each named requirement now gets introduced with "the following statements and expressions" but some of them (like this one) have no statements and only a single expression.

source/lib-intro.tex
```diff
@@ -1824,105 +1825,229 @@
...
+\end{itemdescr}
+
+\pnum
+\tcode{rv}'s state is unspecified.
+
+\begin{note}
+\tcode{rv} must still meet the requirements of the library component that is using it.
+The operations listed in those requirements must work as specified whether \tcode{rv}
+has been moved or not.
+\end{note}
```
jwakely, line 1938: Only the first paragraph after the expression (p3) is indented, leaving the next two (p4 and its note) looking wrong:

Image

source/lib-intro.tex
```diff
@@ -1824,105 +1825,229 @@
...
+t = rv;
+\end{itemdecl}
+
+\begin{itemdescr}
+\pnum
 \begin{itemize}
+
 \item
-For all \tcode{a}, \tcode{a == a}.
+\result
```
jwakely, line 1977: I don't think we want elements like _Result_: inside an itemized list.

Image

notdanhan:
Notes:
In order to prevent hbox badness caused by `\iref` for very long clause names, I reordered Cpp17Requirements in the same paragraph, I'd assume this would not change the semantic requirements but if that's a no-no I will return them to their former order

source/utilities.tex
```diff
@@ -16274,10 +16274,10 @@
 An enabled specialization \tcode{hash<Key>} will:
 \begin{itemize}
 \item meet the \oldconcept{Hash} requirements (\tref{cpp17.hash}),
-with \tcode{Key} as the function
-call argument type, the \oldconcept{Default\-Constructible} requirements (\tref{cpp17.defaultconstructible}),
-the \oldconcept{CopyAssignable} requirements (\tref{cpp17.copyassignable}),
+with \tcode{Key} as the function call argument type,
+the \oldconcept{DefaultConstructible} requirements\iref{cpp17.defaultconstructible},
 the \oldconcept{Swappable} requirements\iref{swappable.requirements},
+the \oldconcept{CopyAssignable} requirements~\iref{cpp17.copyassignable},
```
notdanhan, line 16280: this was moved to the end to stop hbox badness

source/time.tex
```diff
@@ -1038,9 +1038,9 @@
 
 \item
 the types \tcode{TC::rep}, \tcode{TC::duration}, and \tcode{TC::time_point}
-meet the \oldconcept{EqualityComparable} (\tref{cpp17.equalitycomparable}) and
-\oldconcept{LessThanComparable} (\tref{cpp17.lessthancomparable})
-and \oldconcept{Swappable}\iref{swappable.requirements}
+meet the \oldconcept{Swappable}\iref{swappable.requirements} and
```
notdanhan, line 1041: `Swappable` and `EqualityComparable` were swapped to avoid hbox badness

source/utilities.tex
jwakely: I think it makes more sense in the original order, construction, assignment, swap.

Have you tried putting hyphenation hints instead, e.g. Default\\-Construct\\-ible and Copy\\-Assign\\-able and Swapp\\-able?

source/time.tex
jwakely: Again, I prefer the original order if it's possible to preserve it with hyphenation hints.

notdanhan:
I tried adding them and PdfLatex wouldn't take the bait, no matter what I
tried, so I had to reorder :(
I could add forced line breaks if needs be.

jensmaurer:
Yes, we should make every effort to keep the original order.  Hm... This is the first time this appears.  Strange.

PR #9143, opened 2026-06-17
[exec.let] Fix let-env argument in let-state env_t and constructor

abhinavagarwal07:
Fix two `let-state` uses of `let-env` mis-parenthesized while applying [P3826R5](https://wg21.link/p3826r5) in [#8919](https://github.com/cplusplus/draft/pull/8919). The misapplication left `get_env(...)` outside `let-env(...)`, turning both uses into comma expressions.


tkoeppe:
Presumably the missing parentheses just make the whole thing ill-formed...

tkoeppe:
I appreciate the dedication to the research, but just as some feedback: I don't think anyone would seriously have misunderstood this in the way your analysis considers. It was clearly just a simple oversight, and no analysis beyond "misapplication of the paper" is required here.

PR #9144, opened 2026-06-17
[func.wrap.ref.ctor] Restore class T in constant_wrapper constructor head

abhinavagarwal07:
Restore the missing `class T` template parameter in the out-of-line `function_ref` constructor declaration so it matches the synopsis.

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -15804,7 +15804,7 @@ is expression-equivalent\iref{defns.expression.equivalent} to
 
 \indexlibraryctor{function_ref}%
 \begin{itemdecl}
-template<auto c, class F>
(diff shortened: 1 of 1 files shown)


jensmaurer (approved):

tkoeppe:
Thanks! How was this lost in the first place...

tkoeppe:
Misapplication of P3948R1 via c8178e6873d674d7602c84d7daaae2593a909288. (Instead of adding `class F, `, the edit changed `T` to `F`.

jwakely:
The good news is that P4206R0 (Brno LWG poll 5) doesn't touch this again, so won't conflict with this fix.

PR #9145, opened 2026-06-17
[simd.math,simd.reductions] Fix undeclared identifiers in std::simd

abhinavagarwal07:
Fix undeclared identifiers in `std::simd` wording: use `V::size()` in `modf-vec` and `T` in masked `reduce_max`. Also name the previously unnamed `x` and `mask` parameters in the masked `reduce_min`/`reduce_max` declarations so they match their Returns descriptions.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -18926,7 +18926,7 @@ The value of an element \tcode{x[$j$]} for which \tcode{x[$i$] < x[$j$]} is
 \begin{itemdecl}
 template<class T, class Abi>
   constexpr T reduce_min(
-    const basic_vec<T, Abi>&, const typename basic_vec<T, Abi>::mask_type&) noexcept;
+    const basic_vec<T, Abi>& x, const typename basic_vec<T, Abi>::mask_type& mask) noexcept;
 \end{itemdecl}
 
 \begin{itemdescr}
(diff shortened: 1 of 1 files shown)


source/numerics.tex
jensmaurer: This also needs names "x" and "mask", right?

source/numerics.tex
```diff
@@ -18974,7 +18974,7 @@
 \pnum
 \returns
 If \tcode{none_of(mask)} is \tcode{true}, returns
-\tcode{numeric_limits<V::value_type>::lowest()}.
+\tcode{numeric_limits<T>::lowest()}.
```
jensmaurer, line 18977: This looks correct.  @mattkretz , please have a look.

source/numerics.tex
mattkretz: I agree.

source/numerics.tex
mattkretz: 👍

mattkretz (commented):
LGTM except for the (pre-existing) missing `x` and `mask`.

jensmaurer (approved):

source/numerics.tex
tkoeppe: Can we get these fixes in the same PR by any chance?

abhinavagarwal07: PTAL

abhinavagarwal07: Ping @tkoeppe

jensmaurer (approved):

PR #9146, opened 2026-06-17
[checks] Implement Python checks

### LINKED ISSUE #8086: [checks] Create a better `check-source.sh` in Python
This is an idea that has been on my mind for a while, and which may be a decent productivity booster. There are various contributors to our repo, some of whom only do a little bit of work seasonally, and so having high-quality checks can make contributing quite a bit easier.

With the current checks, there are a few issues:

## Maintainability problems

They are based on `sed` and `grep`, and our checking code is far from expressive:
```sh
# "Class" heading without namespace
for f in $texlib; do
    sed -n '/rSec[0-9].*{Class/,/\\end{codeblock}/{/\\begin{example}/,/\\end{example}/b;/\\begin{codeblock}/,/\(^namespace\)\|\(\\end{codeblock}\)/{s/template<[^>]*>//;/\(class\|struct\)[A-Za-z0-9_: ]*{/{=;p;};};}' $f |
    # prefix output with filename and line
    sed '/^[0-9]\+$/{N;s/\n/:/;}' | sed "s/.*/$f:&/"
done |
    fail 'No namespace around class definition' || failed=1
```

This kind of code borders on hieroglyphic. No offense to the author; it's just a limitation of shell tools.

Realistically, this should be something like:
```python
# Somewhere in loop that goes over each Tex line ...
if inside_declaration_code_block and not inside_namespace:
    if re.match(r"(struct|class)\s+[a-zA-Z_]+\b", current_line) is not None:
        raise_error("No namespace around class definition.")
```

## Error output quality

Secondly, the output quality could be improved:

```diff
-Error: future.tex:881:\iref must be flush against the preceding word, not at the start of a line: \iref{fs.path.member}:
+Error: future.tex:881:13: \iref must be flush against the preceding word, not at the start of the line:
+     | \pnum
+     | The following members are declared in addition to those members
+ 881 | specified in \iref{fs.path.member}:
+                   ^~~~~~~~~~~~~~~~~~~~~~ 
```

I find that getting a bit of a bigger code citation is helpful. It's also helpful to get suggested improvements. For example, in the `constexpr static` checks, the error message could include a code citation with `static constexpr`, i.e. with the problem fixed.

## No linter control

There are numerous checks which we could silence via something like `%NOLINTBEGIN(check-id)` within certain locations. For example, we can have a global check which outlaws raw use of `\textit` (we usually want `\exposid` or various other semantic macros), but there are still 150 or so `\textit`s, and it would take some time to dismantle them.

The easy way forward is to wrap those "bad sections" in `NOLINTBEGIN`, `NOLINTEND`, which would allow us to enable many more helpful checks in new markup, without worrying about not-yet-fixed old markup.

Other examples include enforcing `~\ref` instead of `\ref` within certain regions (possibly disabling the checks in synopses or code comments), enforcing that `\returns` has to be wrapped in `\begin{itemdescr}`, etc.

eisenwave:
Fixes #8086

I've mentioned the option to improve the checks in Croydon to @tkoeppe, and he seemed generally open to the idea.

grafik

In summary, this PR converts all the checks we have in `check-source.sh` (plus some additional once adopted from `check-output.sh`) and turns them into a Python script. The benefits are:

- Much nicer formatting for errors. You get VSCode-friendly source locations that you can use to navigate directly to the line and column in an editor. You also get a citation of the source with the affected column span highlighted, so you're not left guessing where the error is located.
- No more hieroglyphic `sed` and `grep` piping. The code is relatively simple Python. It still uses a bunch of regular expressions for a lot of checks; the greatest benefits are in the more complex checks.
- Ability to write test cases for checks using `EXPECTCHECKNEXTLINE`.
- Ability to enable and disable individual checks for source spans using `NOCHECKBEGIN` and other directives. This makes it possible for us to add much more aggressive checks that would have previously been infeasible because they generate too much noise in old TeX sources.
  - A good example of that is in this PR, where I disable the check for `\tcode{\exposid{...}}` in a part of `numerics.tex` because there are too many violations.
  - We could also use this to e.g. disallow use of `\textit` in "newer" TeX ranges for example without having to fix every single violation. We could also use it to catch many "bad words" or "bad phrases" that appear too often to be checked right now, but that we don't want perpetuated.
  - We could also use this to enforce a little bit of auto-enforcement of semantic line breaks (such as requiring a new line after a sentence period) without needing to fix the countless existing violations.
- With check suppression, it is possible to suppress checks that would have inevitably failed during motions. For example, if there is a dangling `\ref` to a section that is introduced by another paper, you can still get a green build by suppressing the `\ref` check (that is, at least for the `check-source` checks).
  
There are a few notable new checks that make things a lot easier in the future:
1. `base-env-balancing` checks for balancing of `\begin` and `\end` commands. It's quite easy to get it wrong, and when you do, it tends to blow up after minutes during the TeX build in bizarre and hard-to-debug ways.
2. `base-unknown-command` checks for use of TeX commands that have never appeared before in the code base. It's very unlikely that this is intentional in most PRs, and very likely that someone just wrote `\tocde` instead of `\tcode` or something else stupid. If we need a new TeX command, we can just add it to the list.


eisenwave:
Some more neat examples:

grafik

grafik

grafik

grafik

eisenwave:
Latest innovation:
grafik

tools/test-checks.tex
```diff
@@ -0,0 +1,262 @@
...
+
+%EXPECTCHECKNEXTLINE(base-hex-unicode-case)
+\unicode{00A0}
+%EXPECTCHECKNEXTLINE(base-hex-unicode-case)
+\unicode{00FF}
+
+%EXPECTCHECKNEXTLINE(base-tcode-exposid)
+\tcode{\exposid{snake-case}}
+%EXPECTCHECKNEXTLINE(base-tcode-exposid)
+\tcode{\exposid{kebak-case}}
```
Tsche: did you mean "kebab-case"? Also, that `snake-case` one isn't actually `snake_case`.

.gitignore
```diff
@@ -22,3 +22,4 @@ tools/sections
 *.synctex.gz
 *.synctex*
 .check.stamp
+__pycache__
```
Tsche, line 26: While it won't matter until we have multiple Python modules, you might want to also ignore `*.pyc`, `*.egg` and `*.egg-info`.

pyrightconfig.json
```diff
@@ -0,0 +1,14 @@
...
+    ],
+    "typeCheckingMode": "strict",
+    "reportMissingTypeStubs": false,
+    "reportUnknownMemberType": false,
+    "reportUnknownArgumentType": false,
+    "reportUnknownVariableType": false,
+    "reportUnknownParameterType": false,
+    "reportUnusedImport": true,
+    "reportUnusedVariable": true,
+    "pythonVersion": "3.11"
```
Tsche: why 3.11 specifically? That version of Python is EoL in about a year.

For the CI we use Ubuntu 24.04. This ships with Python 3.12 - the current Python release is 3.14

.github/workflows/check-python.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: checkout
+        uses: actions/checkout@v6
+
+      - name: setup Node
+        uses: actions/setup-node@v6
+        with:
+          node-version: "20"
+
+      - name: install pyright
+        run: npm install -g pyright
```
Tsche, line 30: I'm not sure `pyright` is a good call. There are other alternatives that do not require pulling in node as dependency.

tools/check-source.py
```diff
@@ -0,0 +1,2231 @@
...
+from collections import defaultdict
+from dataclasses import dataclass
+from enum import Enum
+from pathlib import Path
+from typing import Dict, List, Optional, Pattern, Set, Tuple
+
+
+# ==================================================================================================
+# Lookup tables
+# ==================================================================================================
```
Tsche: I'm really not a fan of those section comment headers. Would it be viable to chop this >2k LoC module up into several smaller modules?

tools/check-source.py
```diff
@@ -0,0 +1,2231 @@
...
+    "hardexpects",  # Hardened preconditions
+    "effects",  # Effects
+    "sync",  # Synchronization
+    "ensures",  # Postconditions
+    "returns",  # Returns
+    "throws",  # Throws
+    "complexity",  # Complexity
+    "remarks",  # Remarks
+    "errors",  # Error conditions
+]
```
Tsche: Can these comments be aligned? Do we really need comments that just repeat the exact string literal preceding it?

tools/check-source.py
```diff
@@ -0,0 +1,2231 @@
...
+    """`true` if a matching failure was reported."""
+
+
+# ==================================================================================================
+# Global state
+# ==================================================================================================
+
+source_dir: Path = Path()
+file_locations: Dict[str, Path] = {}
+expected_registry: Dict[Tuple[str, int], ExpectedFailure] = {}
```
Tsche: Can we avoid global state?

tools/check-source.py
```diff
@@ -0,0 +1,2231 @@
...
+def make_alt_pattern(items: List[str]) -> str:
+    return "(?:" + "|".join(re.escape(s) for s in items) + ")"
+
+
+def read_file(path: Path) -> List[str]:
+    try:
+        with open(path, encoding="utf-8", errors="replace") as f:
+            return f.read().splitlines()
+    except OSError:
+        return []
```
Tsche: You can just do `path.read_text().split(os.linesep)` instead of the context manager stuff. If you choose not to do this, please do change `open` to `path.open`.

tools/check-source.py
```diff
@@ -0,0 +1,2231 @@
...
+import os
+import re
+import sys
+import unicodedata
+from abc import ABC
+from collections import defaultdict
+from dataclasses import dataclass
+from enum import Enum
+from pathlib import Path
+from typing import Dict, List, Optional, Pattern, Set, Tuple
```
Tsche: `Dict`, `List`, `Pattern`, `Set` and `Tuple` from `typing` are all deprecated since Python 3.9. Please use the builtins (`dict`, `list` `set` and `tuple`) and `re.Pattern` instead.

`Optional[int]` can be written as `int | None` since 3.10.

tools/check-source.py
```diff
@@ -0,0 +1,2231 @@
...
+
+def read_file(path: Path) -> List[str]:
+    try:
+        with open(path, encoding="utf-8", errors="replace") as f:
+            return f.read().splitlines()
+    except OSError:
+        return []
+
+
+def format_failure(fail: Failure, lines: List[str]) -> str:
```
Tsche: If you don't care what kind of listy thing `lines` is, please annotate it with `Iterable[str]` instead of `List[str]`.

tools/check-source.py
```diff
@@ -0,0 +1,2231 @@
...
+        f" {style(f'[{fail.check_id}]', ANSI_GRAY)}"
+    )
+    if fail.line < len(lines):
+        src_line = lines[fail.line]
+        display_line = src_line.replace("\t", "    ")
+        line_num_str = str(fail.line + 1)
+        num_width = max(6, len(line_num_str))
+        pad = " " * (num_width - len(line_num_str))
+        col1 = fail.column_start + 1
+        col2 = fail.column_end + 1
```
Tsche: `col1` and `col2` aren't very good names, also you can fold "related" statements into one, ie
```py
start, end = fail.column_start + 1, fail.column_end + 1
```

tools/check-source.py
```diff
@@ -0,0 +1,2231 @@
...
+        re.sub(r"`([^`]+)`", f"{ANSI_YELLOW}\\1{ANSI_RESET}", fail.message)
+        if COLOR
+        else fail.message
+    )
+    parts.append(
+        f"{style(fail.file, ANSI_BOLD)}:"
+        f"{style(str(fail.line + 1), ANSI_BOLD)}:"
+        f"{style(str(fail.column_start + 1), ANSI_BOLD)}: "
+        f"{style('error:', ANSI_RED)} {message}"
+        f" {style(f'[{fail.check_id}]', ANSI_GRAY)}"
```
Tsche: the extra space for the last interpolated looks off. Can this be a trailing space for the preceding one to not break alignment?

tools/check-source.py
```diff
@@ -0,0 +1,2231 @@
...
+        parts.append(f"{gutter}{' ' * highlight_pad}{hl}")
+    return "\n".join(parts)
+
+
+def find_env_ranges(
+    lines: List[str],
+    env: Environment,
+) -> List[Tuple[int, int]]:
+    begin_pattern = env.begin_pattern
+    end_pattern = env.end_pattern
```
Tsche: I don't get the point of these. Why not just write `env.begin_pattern` the only time this is used?

tools/check-source.py
```diff
@@ -0,0 +1,2231 @@
...
+    stack: List[int] = []
+    for idx, line in enumerate(lines):
+        if begin_pattern.search(line):
+            stack.append(idx)
+        if end_pattern.search(line) and stack:
+            ranges.append((stack.pop(), idx + 1))
+    return ranges
+
+
+unexpected_count = 0
```
Tsche: does this have to be global?

tools/check-source.py
```diff
@@ -0,0 +1,2231 @@
...
+    begin_pattern = env.begin_pattern
+    end_pattern = env.end_pattern
+    ranges: List[Tuple[int, int]] = []
+    stack: List[int] = []
+    for idx, line in enumerate(lines):
+        if begin_pattern.search(line):
+            stack.append(idx)
+        if end_pattern.search(line) and stack:
+            ranges.append((stack.pop(), idx + 1))
+    return ranges
```
Tsche: can this be a generator that `yield`s those `tuple[int, int]` pairs instead of putting them into a temporary list?

ie 
```py
   stack: list[int] = []
    for idx, line in enumerate(lines):
        if env.begin_pattern.search(line):
            stack.append(idx)
        if env.end_pattern.search(line) and stack:
            yield stack.pop(), idx + 1
```

tools/check-source.py
```diff
@@ -0,0 +1,2231 @@
...
+        file=file,
+        line=line,
+        column_start=column_start,
+        column_end=column_end,
+        message=message,
+        check_id=check.id,
+    )
+    # Reading the file from scratch is very slow,
+    # but we don't care because this is the unhappy path anyway,
+    # and we usually don't expect failures anyway.
```
Tsche, line 208: Well.. when we have multiple check failures this'll re-read files multiple times. This function is called in two loops later in this module. Are you absolutely certain this won't trash performance of this tool?

tools/check-source.py
```diff
@@ -0,0 +1,2231 @@
...
+    print(format_failure(fail, lines), file=sys.stderr)
+    print(file=sys.stderr)
+
+
+# ==================================================================================================
+# Checks
+# ==================================================================================================
+
+
+class Check(ABC):
```
Tsche: `ABC`s without `@abstractmethod` decorations on any functions are largely pointless. You get to use virtual subclass registration via `.register()`, but this isn't used here anyway.

Generally speaking ABCs are a bit unpythonic and I don't really see the need for one here. You could also consider a protocol here.

tools/check-source.py
```diff
@@ -0,0 +1,2231 @@
...
+        """Called for each line while the check is active."""
+        pass
+
+    def end_file(self, file_path: Path) -> None:
+        """Called after processing all lines of `file_path`."""
+        pass
+
+    def end_checks(self) -> None:
+        """Called once after all files have been processed."""
+        pass
```
Tsche: If these are meant to be "pure virtual", you need to add an `@abstractmethod` decorator. Also this `pass` is superfluous, the doc-string already is a statement.

In Python we usually spell these one of these ways:
```py
@abstractmethod
def foo(self):
  raise NotImplementedError

@abstractmethod
def foo(self): ...

@abstractmethod
def foo(self): pass

@abstractmethod
def foo(self):
  """ some description """
```

I personally prefer `NotImplementedError` since that'll still give me an error even if I forget to derive from `ABC` (which makes `@abstractmethod` largely useless)

tools/check-source.py
```diff
@@ -0,0 +1,2231 @@
...
+
+    def fail(
+        self,
+        line: int,
+        column_start: int,
+        column_end: int,
+        message: str,
+    ) -> None:
+        """Report a failure at `self.file_path`."""
+        emit_check_failure(
```
Tsche, line 263: `emit_check_failure` is only called from within classes deriving from `Check`. Can you fold it into `fail` and consistently use this one?

tools/check-source.py
```diff
@@ -0,0 +1,2231 @@
...
+
+# ==================================================================================================
+# Checks
+# ==================================================================================================
+
+
+class Check(ABC):
+    """Base class for all checks."""
+
+    id: str = ""
```
Tsche: `id` is a Python builtin, please don't use it for identifiers. Also I'm not entirely sure what you're trying to do here - this `id` is a class attribute not an instance attribute.

Did you mean to say
```py
class Check(ABC):
  uid: str
  def __init__(self, uid: str = ""):
    self.uid = uid

# ...
class Foo(Check):
  def __init__(self, uid: str, **whatever):
    super().__init__(uid)
```
?

tools/check-source.py
```diff
@@ -0,0 +1,2231 @@
...
+
+
+def run_checks(tex_files: List[Path]) -> int:
+    """Run all registered checks. Returns the number of unexpected failures."""
+    global unexpected_count
+    unexpected_count = 0
+    file_locations.clear()
+
+    for fp in tex_files:
+        file_locations[os.path.relpath(fp)] = fp
```
Tsche: please don't mix `os.path` and `pathlib` usage. `pathlib.Path` exposes this functionality as `relative_to()`.

tools/check-source.py
```diff
@@ -0,0 +1,2231 @@
...
+                    else:
+                        active |= {a for a in all_ids if fnmatch.fnmatch(a, cid)}
+                elif directive == "NEXTLINE":
+                    if cid == "*":
+                        for c in all_ids:
+                            skip_next[c].add(line_num + 1)
+                    else:
+                        matched = {a for a in all_ids if fnmatch.fnmatch(a, cid)}
+                        for c in matched:
+                            skip_next[c].add(line_num + 1)
```
Tsche: you have some code duplication here

tools/check-source.py
```diff
@@ -0,0 +1,2231 @@
...
+                file=sys.stderr,
+            )
+
+    if num_failures:
+        print(
+            f"{style(str(num_failures), ANSI_RED)} error(s) emitted.",
+            file=sys.stderr,
+        )
+
+    exit_code = 1 if num_failures > 0 else 0
```
Tsche: you can just do `num_failures > 0` here. If you insist on getting an `int`, do `int(num_failures > 0)` but it's not necessary.

Tsche (changes requested):

pyrightconfig.json
eisenwave: I just blindly copied and pasted this from another project, so no particular reason.

.github/workflows/check-python.yml
eisenwave: I just use it because it integrates nicely with VSCode out of the box. What would you recommend instead?

tools/check-source.py
eisenwave: There may be some merit to pulling the lookup tables into a separate Python file just because they are a bit of clutter and make the script look a lot bigger than it actually is. The rest is just around 1K lines, and modulerization probably wouldn't improve readability much.

tools/check-source.py
eisenwave: Maybe we could; this is just what my auto-formatter did. And it's not exactly repeating the string literal. The comments are just there for the non-obvious relationships like `ensures` and `Postconditions`.

tools/check-source.py
eisenwave: Well, the expectation is that this code is never reached because checks are never supposed to fail in the first place. And if they do, having the printing of errors be a bit slower is pretty alright.

Locally, I've never noticed this being a problem, even when spamming lots of errors.

eisenwave:
@Tsche thanks a lot for the review. It'll take some time to clean it all up.

tools/check-source.py
eisenwave: I didn't even know that there is such a thing as class attributes and instance attributes, but okay. I thought the `""` is a default member initializer or something, and I think I used that at some point but then it became pointless.

source/numerics.tex
```diff
@@ -17724,6 +17725,7 @@
 
 \pnum
 \constraints
+%NOCHECKBEGIN(base-tcode-exposid)
 Every type in the parameter pack \tcode{Flags} is one of \tcode{\exposid{convert-flag}},
 \tcode{\exposid{aligned-flag}}, or \tcode{\exposid{over\-aligned-\brk{}flag}<N>}.
```
jensmaurer: I think we can just remove the \tcode wrapping around \exposid here?

source/numerics.tex
```diff
@@ -20946,7 +20949,8 @@
 \tcode{basic_mask<Bytes, Abi>} is trivially copyable.
 
 \pnum
-\recommended Implementations should support implicit conversions between
+\recommended
+Implementations should support implicit conversions between
```
jensmaurer, line 20868: Please  move all specific fixes to a separate commit.

jensmaurer (changes requested):

source/numerics.tex
eisenwave: Yeah, we could. It would be something like 10 separate fixes.

I think I'll just do a surgical fix in a separate PR for these cleanups so that this PR doesn't have to clean up any existing sources.

The problem with the existing check is that the regex pattern doesn't work on hyphenated identifiers.

eisenwave:
All the pre-existing issues are taken care of in separate PRs. Now comes rebasing on the latest motions and addressing all the feedback on the Python script.

pyrightconfig.json
eisenwave: Fixed in [6d4a54c](https://github.com/cplusplus/draft/pull/9146/commits/6d4a54cc9e74f20f1be128239943bcf71b2ac9f8)

tools/check-source.py
eisenwave: [017ddec](https://github.com/cplusplus/draft/pull/9146/commits/017ddece48fb2b4c2438a9f60c77befea0459359) split the lookup tables off into a separate file. I think that's the low-hanging fruit.

You could chop it into several more files later, but I don't see how that improves things meaningfully. You would just need to jump between files more often and manage imports.

tools/check-source.py
eisenwave: Comments removed in 8daba1d5edffc50bb68f1fa5830a563961189971

tools/check-source.py
eisenwave: Fixed in 9f2e23ee1430f6f414c08bed7fc86de0e6ad96b0

tools/check-source.py
eisenwave: Fixed in 35ed7e7853baea4f7caed69614c519304c16582f

tools/check-source.py
eisenwave: Fixed in 1a5fc29e33104e2eeb7a9fe5b4d03b71b9c9860f

tools/test-checks.tex
eisenwave: Fixed in e20070e4d7223476d14688e5ed8b2e3c79bda436

.gitignore
eisenwave: Fixed in 4513a7576f009f1971080b4a7e1974519fe4472b

tools/check-source.py
eisenwave: I use `len(lines)` and `lines[fail.line]` below, so this cannot just be an iterable.

tools/check-source.py
eisenwave: Fixed in 51114133846d4d6df93c0021a07d54d626f0fb8d

tools/check-source.py
eisenwave: They might have been useful before refactoring the function at some early stage in writing this code.

Fixed 665dbe04cae6cb619fbc9a29cccf4d168df79b2e

tools/check-source.py
eisenwave: Fixed in c48c2dd7d7d34bf546bb88056a4740629faa1d7d

tools/check-source.py
eisenwave: Fixed in 388e42cb7345447528e2ffd6e4cb98f9060f4585

tools/check-source.py
eisenwave: Fixed in 7af3d566b813c579f2f7bec092800c6dfb854922

tools/check-source.py
eisenwave: Fixed in 7af3d566b813c579f2f7bec092800c6dfb854922

tools/check-source.py
eisenwave: After the changes in 7af3d566b813c579f2f7bec092800c6dfb854922, I wouldn't want to. The difference is that `fail` is just a convenience function that provides the file path automatically, and I would want there to be a global "emit failure" function at some level, in case we expand the sources of failures in the future.

`emit_check_failure` is basically the global `console.error` counterpart, and that has kind of a right to exist even if you always use it via a wrapper. That might change at some point.

tools/check-source.py
eisenwave: Fixed in 33b9379c17fd09dcaf21857da1abc061570a7918

tools/check-source.py
eisenwave: Fixed in 147575ecb748aef62d970c9d4bf010fc3f7c4718

tools/check-source.py
eisenwave: Fixed in 0cebc60a25fb843e20420f79bceb50bed123c0fb

tools/check-source.py
eisenwave: Fixed in 7a038709d45d9af12f2d7944b72953108e6e3aed (removed `ABC`)

tools/check-source.py
eisenwave: Note that these are not actually abstract; they might just do nothing.

There are checks that only use `check_line` because they go line-by-line, in which case the other methods keep their empty implementation, which is okay. Analogously, other checks might only use `begin_file` and `end_file` and whatnot.

source/intro.tex
```diff
@@ -325,8 +326,7 @@
 behavior, for a \termref{defns.well.formed}{well-formed program}{} construct and correct data, that
 depends on the implementation and that each implementation documents
 
-\definition{implementation-defined strict total order over pointers}
-{defns.order.ptr}
+\definition{implementation-defined strict total order over pointers}{defns.order.ptr}
```
Tsche, line 329: this looks like an unrelated change, does it have to be in this patch?

pyrightconfig.json
```diff
@@ -0,0 +1,14 @@
...
+    ],
+    "typeCheckingMode": "strict",
+    "reportMissingTypeStubs": false,
+    "reportUnknownMemberType": false,
+    "reportUnknownArgumentType": false,
+    "reportUnknownVariableType": false,
+    "reportUnknownParameterType": false,
+    "reportUnusedImport": true,
+    "reportUnusedVariable": true,
+    "pythonVersion": "3.12"
```
Tsche, line 13: Is there any particular reason why this requires Python 3.12 rather than the current version (3.14)?

tools/check-source.py
```diff
@@ -0,0 +1,1408 @@
...
+from collections import defaultdict
+from dataclasses import dataclass
+from enum import Enum
+from pathlib import Path
+from re import Pattern
+
+from data import *
+
+# ===============================================================================
+# Terminal color support
```
Tsche, line 20: Please get rid of these section headers

tools/check-source.py
```diff
@@ -0,0 +1,1408 @@
...
+
+        lines = read_file(file_path)
+
+        # Always call begin_file.
+        for c in CHECKS:
+            c.begin_file(file_path, lines)
+
+        # Active-set management — controls whether check_line is called.
+        active: set[str] = set(all_ids)
+        # Per-check next-line skips: check_id → set of line numbers.
```
Tsche, line 1250: Please get rid of the LLMy em dashes and other unicode symbols (such as →) in comments. This might cause rendering issues with some editors and environments, I don't see a need to use anything but the ASCII-compatible subset here (even though Python assumes UTF-8 anyway).

tools/check-source.py
```diff
@@ -0,0 +1,1408 @@
...
+    # Such checks run on all files and identify problems like illegal characters,
+    # trailing whitespace, etc.
+    # ----------------------------------------------------------------------------------------------
+    NonAsciiCheck("text-non-ascii-char"),
+    BannedPatternCheck(
+        "text-trailing-ws",
+        re.compile(r"\s+$"),
+        "Line has trailing whitespace; remove the extra spaces.",
+    ),
+    TrailingEmptyLinesCheck("text-trailing-empty-lines"),
```
Tsche, line 958: Could you push those groups of checks into some `CheckGroup` object and attach the comment/name to it as strings? 

That way you can reuse this for the command line interface (or well, its help text) to group checks and print a short description/name for those groups.

tools/check-source.py
```diff
@@ -0,0 +1,1408 @@
...
+            self,
+            str(self.file_path.relative_to(Path.cwd(), walk_up=True)),
+            line,
+            column_start,
+            column_end,
+            message,
+        )
+
+
+class BannedPatternCheck(Check):
```
Tsche, line 273: Do you think it would make sense to break this apart into multiple modules? This file is rather large already

tools/check-source.py
```diff
@@ -0,0 +1,1408 @@
...
+                m.start(),
+                m.end(),
+                f"Non-ASCII character U+{cp:04X} {name} found; "
+                f"use the appropriate LaTeX macro instead.",
+            )
+
+
+class TrailingEmptyLinesCheck(Check):
+    def __init__(self, check_id: str):
+        super().__init__(check_id)
```
Tsche, line 372: This is not necessary (likewise in many other child classes of Check).

If you don't do any additional things in the child constructor, you don't have to define it just to defer to the base ctor.

tools/check-source.py
```diff
@@ -0,0 +1,1408 @@
...
+
+class TailnoteTailexampleCheck(Check):
+    def __init__(self, check_id: str):
+        super().__init__(check_id)
+
+    END_PATTERN = re.compile(r"\\end\{(?:example|note)\}")
+    TAIL_PATTERN = re.compile(r"- *(?:\\\\|&)")
+
+    def begin_file(self, file_path: Path, lines: list[str]) -> None:
+        super().begin_file(file_path, lines)
```
Tsche, line 420: I feel like this is not very good design if you need to explicitly invoke the base class version in every child that overrides this. 

Maybe that code in the base class can be moved elsewhere to run unconditionally _before_ calling begin_file?

tools/check-source.py
```diff
@@ -0,0 +1,1408 @@
...
+        super().__init__(check_id)
+
+    def begin_file(self, file_path: Path, lines: list[str]) -> None:
+        super().begin_file(file_path, lines)
+        self.check_file(file_path, lines)
+
+    def check_file(self, file_path: Path, lines: list[str]) -> None:
+        in_section = False
+        in_example = False
+        in_cb = False
```
Tsche, line 656: what's a cb? codeblock? Please don't abbreviate needlessly

tools/check-source.py
```diff
@@ -0,0 +1,1408 @@
...
+        super().begin_file(file_path, lines)
+        self.check_file(file_path, lines)
+
+    def check_file(self, file_path: Path, lines: list[str]) -> None:
+        for start, end in find_env_ranges(lines, Environment.ITEMDESCR):
+            if "% NOCHECK:" in lines[start] and "order" in lines[start]:
+                continue
+            prev_name: str | None = None
+            for idx in range(start, end):
+                m = self.relevant_line_pattern.match(lines[idx])
```
Tsche, line 744: please avoid single-letter variables. How about `match`?

tools/check-source.py
```diff
@@ -0,0 +1,1408 @@
...
+
+class UnbalancedBeginAndEndCheck(Check):
+    BEGIN_OR_END_PATTERN = re.compile(r"\\(begin|end)\{([^}]+)\}")
+
+    def __init__(self, check_id: str):
+        super().__init__(check_id)
+
+    def begin_file(self, file_path: Path, lines: list[str]) -> None:
+        super().begin_file(file_path, lines)
+        self.stack: list[tuple[str, int, int, int]] = []
```
Tsche, line 769: Please define a dataclass instead of this `tuple[str, int, int, int]` thing

tools/check-source.py
```diff
@@ -0,0 +1,1408 @@
...
+        self,
+        check_id: str,
+        pattern: Pattern[str],
+        message: str,
+    ):
+        super().__init__(check_id)
+        self.pattern = pattern
+        self.message = message
+
+    def check_line(self, line_num: int, line: str) -> None:
```
Tsche, line 286: Please use the `@override` decorator (from `typing`) to mark when you're overriding member functions of the base class. This gives you nice diagnostics if you ever mess up the signature.

tools/check-source.py
```diff
@@ -0,0 +1,1408 @@
...
+    print(format_failure(fail, lines), file=sys.stderr)
+    print(file=sys.stderr)
+
+
+# ==================================================================================================
+# Checks
+# ==================================================================================================
+
+
+class Check:
```
Tsche, line 220: Since this base class isn't actually doing terribly much.. have you considered a Protocol instead?

tools/check-source.py
```diff
@@ -0,0 +1,1408 @@
...
+        super().__init__(check_id)
+
+    def begin_file(self, file_path: Path, lines: list[str]) -> None:
+        super().begin_file(file_path, lines)
+        self.current_label: str | None = None
+
+    def check_line(self, line_num: int, line: str) -> None:
+        m = self.SECTION_PATTERN.match(line)
+        if m:
+            self.current_label = m.group(1)
```
Tsche, line 590: In Python's regex you can use named groups (ie `(?P<name>.*)`). This allows you to write this as `m.group("current_label")` rather than using a magic number.

tools/check-source.py
```diff
@@ -0,0 +1,1408 @@
...
+                cb_lines, cb_start = [], idx
+                continue
+            if "\\end{codeblock}" in line:
+                in_cb = False
+                cb_text = "\n".join(cb_lines)
+                stripped = self.TEMPLATE_PATTERN.sub("", cb_text)
+                if self.CLASS_DEFINITION_PATTERN.search(
+                    stripped
+                ) and not self.NAMESPACE_PATTERN.search(stripped):
+                    for ci, cline in enumerate(cb_lines):
```
Tsche, line 687: what's a ci, a cline and a cs? Can you find more self-documenting names for these?

tools/check-source.py
```diff
@@ -0,0 +1,1408 @@
...
+def run_checks(
+    tex_files: list[Path],
+    expected_tracker: ExpectedTracker,
+    source_dir: Path,
+) -> tuple[int, dict[str, Path]]:
+    """Run all registered checks.
+    Returns ``(unexpected_failures, file_locations)``.
+    """
+    file_locations: dict[str, Path] = {}
+    for fp in tex_files:
```
Tsche, line 1227: fp? path?

tools/check-source.py
```diff
@@ -0,0 +1,1408 @@
...
+        file_locations[str(fp.relative_to(Path.cwd(), walk_up=True))] = fp
+
+    for c in CHECKS:
```
Tsche, line 1230: c => check?

tools/check-source.py
```diff
@@ -0,0 +1,1408 @@
...
+    script_dir = Path(__file__).resolve().parent
+    for candidate in (script_dir.parent, Path.cwd()):
+        if (candidate / "source" / "std.tex").exists():
+            return candidate
+        if (candidate / "std.tex").exists():
+            return candidate
+    return script_dir.parent
+
+
+# ==================================================================================================
```
Tsche, line 1335: You could split the CLI into a different module and set up a proper python package for this tool. I'd prefer this to a single-file script that performs relative import of another single-file script.

.github/workflows/check-python.yml
Tsche: I (still) tend to go for mypy for my own projects, but if this gives you a better user experience I'm okay with it.

tools/check-source.py
Tsche: ok

tools/check-source.py
Tsche: ok. Have you considered a configured global logger for this instead? Python's `logging` package is very useful.

tools/check-source.py
```diff
@@ -0,0 +1,1408 @@
...
+    parser = argparse.ArgumentParser(
+        description="Check LaTeX sources for C++ standard drafting rules."
+    )
+    parser.add_argument(
+        "files",
+        nargs="*",
+        help="Specific .tex files or directories to check (default: all .tex files"
+        " under the project root).",
+    )
+    args = parser.parse_args()
```
Tsche, line 1350: Please add a command/argument that just prints all checks and exits. Alternatively you could put this into the help text, but I'd prefer to (at least additionally) have a command. 

If you do the check grouping thing I suggested this would allow filtering checks by group etc.

tools/check-source.py
```diff
@@ -0,0 +1,1408 @@
...
+    return script_dir.parent
+
+
+# ==================================================================================================
+# CLI
+# ==================================================================================================
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(
```
Tsche, line 1341: Could we make it configurable which checks are performed? This could be useful to debug a specific failure.

tools/check-source.py
```diff
@@ -0,0 +1,1408 @@
...
+            )
+
+    if num_failures:
+        print(
+            f"{style(str(num_failures), ANSI_RED)} error(s) emitted.",
+            file=sys.stderr,
+        )
+
+    exit_code = num_failures > 1
+    sys.exit(exit_code)
```
Tsche, line 1404: exit_code seems extraneous. Please fold this into the sys.exit invocation.

tools/check-source.py
```diff
@@ -0,0 +1,1408 @@
...
+    """Pre-scan ``.tex`` files for ``%EXPECTCHECKNEXTLINE(id)`` directives.
+
+    The directive must appear alone on its line; *id* is the check that is
+    expected to fire on the **next** line.
+    """
+    EXPECT_CHECK_PATTERN = re.compile(r"^\s*%EXPECTCHECKNEXTLINE\((\S+)\)\s*$")
+
+    for file_path in tex_files:
+        lines = read_file(file_path)
+        for idx in range(len(lines) - 1):
```
Tsche, line 1304: how about
`for idx, line in enumerate(lines[:-1]):`

Tsche (changes requested):
I took a brief look at this again. Please note that some of the comments apply to multiple locations - I did not repeat them.

source/intro.tex
eisenwave: If I recall correctly, it's like this because the definition-checking check uses capture groups to yoink the `defns.order.ptr` out of this command, and this regex is applied line-by-line. The checks would fail if this wasn't put onto one line.

It's easier to change just this one place in the standard markup than to redesign the checks and make them cross-line.

Tsche: Ok, but why in this patch? This seems like an okay change to do separately. 

I'd prefer if this patch doesn't touch anything that isn't directly related to the new script.

eisenwave: Because the Python checks also come with CI checks in this PR that actually run them, and you don't get a green build if you don't change this line of markup.

pyrightconfig.json
eisenwave: The current GitHub actions `ubuntu-latest` runner is on Ubuntu 24.04 according to GitHub's documentation, whose Python3 package is `3.12.3`.

Tsche: That is not a limitation. You might as well pull a newer Python version in through `uv` or something like micromamba. Also there's an official Python docker container.

3.12 is probably okay for now, but let's please switch to something newer the second we want to use a Python 3.13 or later feature.

source/intro.tex
Tsche: That I understand, but why does this need to be in this patch? Why can't that be done separately?

PR #9147, opened 2026-06-18
[Motions 2026 06 cwg 1] P4271R0 Core Language Working Group "ready" Issues

### LINKED ISSUE #660: [temp.deduct.type] Missing non-deduced context for nested-name-specifier of qualified-id of a template

Reference (section label): temp.deduct.type

Issue description:
https://wg21.link/temp.deduct.type#5.1 has, as a non-deduced context,
- The [_nested-name-specifier_](https://eel.is/c++draft/expr.prim.id.qual#nt:nested-name-specifier) of a type that was specified using a [_qualified-id_](https://eel.is/c++draft/expr.prim.id.qual#nt:qualified-id)[.](https://eel.is/c++draft/temp.deduct.type#5.1.sentence-1)

This does not cover templates named using a _qualified-id_. There is implementation divergence:
```cpp
template <template <typename> class>
struct B;

struct C {
  template <typename>
  struct Nested;
};

template <typename T>
void f(T *, B<T::template Nested> *);

void g(C *cp) {
  f(cp, 0);  // Clang rejects; `T::` in the declaration of `f` is not a non-deduced context
}
```

Online compiler link: https://godbolt.org/z/8sTvW7W94

Suggested resolution:
Modify https://wg21.link/temp.deduct.type#5.1:
- The _nested-name-specifier_ of a type <ins>or template </ins>that was specified using a _qualified-id_.

### LINKED ISSUE #700: CWG3027 [temp.type] Equivalence of pack-index-specifiers
[temp.type]  does seem to suggest any two pack-index-specifiers whose expression is dependent are equivalent if their expression are equivalent.

> If the constant-expression of a pack-index-specifier is value-dependent, two such pack-index-specifiers refer to the same type only if their constant-expressions are equivalent ([temp.over.link]). Otherwise, two such pack-index-specifiers refer to the same type only if their indexes have the same value.

quid of `U...[expr]` and `T...[expr]` ?

Suggestion

> If the constant-expression of a pack-index-specifier is value-dependent, two such pack-index-specifiers <ins>whose _typedef-name_ are equivalent</ins> refer to the same type only if their constant-expressions are equivalent ([temp.over.link]). Otherwise, two such pack-index-specifiers refer to the same type only if their indexes have the same value.

### LINKED ISSUE #732: CWG3046 [dcl.enum] Reading non-representable value from enumeration without fixed underlying type via common initial sequence rule.
Full name of submitter (unless configured in github; will be published with the issue): Benjamin Sch.

Reference (section label): [dcl.enum], [class.mem.general]

Issue description:

```cpp
#include<iostream>
#include<type_traits>

enum E { E0 } e;
enum F { F0, F1, F2, F3 };
static_assert(std::is_same_v<std::underlying_type_t<E>, std::underlying_type_t<F>>);

struct A { E e; };
struct B { F f; };
union U {
    A a;
    B b;
} u;

bool test() {
    return u.a.e == 2;
}

auto ptest = test;

int main() {
    u.b.f = F2;
    std::cout << ptest();
}
```
Assuming the `static_assert` passes, then because the two enumerations have the same underlying type, they are layout-compatible and therefore it is permitted to use the inactive `u.a.e` to read the active `u.b.f` corresponding to it in the common initial sequence. The read should happen as if `u.b.f` was nominated, meaning it should read the value `2` even if `E` is not able to represent that value. And therefore this program must print `1`.

Is this intentional? It prevents the implementation from optimizing the body of `test` to `return false;` as it otherwise could because `E` can't represent a value of `2`. Is the implementation required to special case this enumeration read through union access?

I originally reported this as bug against Clang in https://github.com/llvm/llvm-project/issues/150833.

### LINKED ISSUE #829: CWG3146 [diff.expr] Missing Annex C entry for difference in usual arithmetic conversions for values of enumeration type

Reference (section label): [diff.expr]

## Issue description:
The resolution for FR-030 in WG 14 N3148 clarifies that, in the usual arithmetic conversions, C23 converts values of an enumeration type to the underlying type of the enumeration type. For an unscoped enumeration type whose underlying type is not fixed, C++ instead performs integral promotions (https://wg21.link/conv.prom#3) that do not rely upon the implementation's choice of underlying type as part of the usual arithmetic conversions.

Note: I suspect the C committee might want to change their integral promotion rules for enumeration types instead of their change to their usual arithmetic conversions.

## Suggested resolution:
Add a new entry.
**Affected subclause:** 7.4 [expr.arith.conv]
**Change:** The usual arithmetic conversions differ, between C and C++, in the treatment of enumeration types with no fixed underlying type. In C++, values of such types are not specified as converting to the underlying type of the enumeration as part of the usual arithmetic conversions. Instead, integral promotions ([conv.prom]) based on the values of the enumeration ([dcl.enum]) are applied.
**Rationale:** Avoids the difference that can arise in C when replacing an enumeration constant in an expression with a cast of the same enumeration constant to the enumeration type. Allows C++ to treat an enumerator with behavior like that of `int` even while giving the enumerator the type of its enumeration.
**Effect on original feature:** Changes to semantics of well-defined feature. Some C expressions that have a dependence upon the implementation-defined underlying type of affected enumeration types will yield different results (as they would if a different underlying type is chosen by the C implementation).
[_Example_:
```c
typedef enum { E0 } E;
static_assert(((E)E0 - 1 < 0) == (E0 - 1 < 0), "Must pass for C++");
```
The `static_assert` may fail in C. &mdash;_end example_]
**Difficulty of converting:** Programs must add explicit casts to the underlying type.
**How widely used:** Rare.

### LINKED ISSUE #843: CWG3159 [temp.inst] When are variables with incomplete array types instantiated to complete their array type?
Reference (section label): [temp.inst]

Link to reflector thread (if any): https://lists.isocpp.org/core/2025/04/17803.php

### Issue description:

When a variable has an incomplete array type and a braced initializer, the complete array type can be determined by instantiating a definition of the variable. [temp.inst]/4 and /7 say when this happens:

> /4: "Unless a member of a templated class is a declared specialization, the specialization of the member is implicitly instantiated when the specialization is referenced in a context that requires the member definition to exist or if the existence of the definition of the member affects the semantics of the program; in particular, the initialization (and any associated side effects) of a static data member does not occur unless the static data member is itself used in a way that requires the definition of the static data member to exist."

> /7: "Unless a variable template specialization is a declared specialization, the variable template specialization is implicitly instantiated when it is referenced in a context that requires a variable definition to exist or if the existence of the definition affects the semantics of the program. A default template argument for a variable template is implicitly instantiated when the variable template is referenced in a context that requires the value of the default argument."

Moreover, [temp.inst]/11 says:

> "An implementation shall not implicitly instantiate [...] a variable template [...] or static data member of a templated class [...] unless such instantiation is required."

But it's not entirely clear what this means. Under what circumstances does "the existence of the definition of the member affect[...] the semantics of the program"?

Clang tries to determine this precisely, by checking for places where the array bound matters and only instantiating in those cases, but it sometimes gets it wrong. Other implementations seemingly instantiate a definition whenever the variable is mentioned, even in cases where the semantics of the program are unaffected by the completeness of the type of the variable. Neither approach seems entirely in line with the wording.

Some examples of divergence:
```c++
// Case 1: Clang rejects because the type of `*p` changes
// between redeclarations; Clang seems to be incorrect:
// The array bound affects the semantics of the program.
template<typename T> struct X {
  static inline int arr[] = {1, 2, 3};
};

// Clang does not instantiate a definition here.
extern decltype(X<int>::arr) *p;
// Clang does instantiate a definition here.
int n = sizeof(X<int>::arr);
decltype(X<int>::arr) *p;

// Case 2: GCC, EDG, MSVC reject because they instantiate a definition
// of `X<int>::arr` even though it's not odr-used and doesn't appear to
// affect the semantics of the program.
template<typename T> struct X {
  static inline int arr[] = {1, 2, T::error};
};

// The array bound here doesn't matter; array-to-pointer decay doesn't
// care about it.
decltype(+X<int>::arr) r;
```
What is the intent?

### Suggested resolution:

The rule used by non-Clang implementations seems a lot easier to implement. Assuming we want that, perhaps we should have an explicit rule in [temp.inst] similar to /8, eg something like:

"The existence of a definition of a variable is considered to affect the semantics of the program if it affects an array bound in the type of an id-expression ([dcl.array]), even if the semantics of the context in which the id-expression appears do not depend on the array bound."

### LINKED ISSUE #851: CWG3164 [cpp.predefined] Contradictory definitions of macros to test for standard extended floating point types
Full name of submitter (unless configured in github; will be published with the issue): Jay Ghiron

Reference (section label): [[cpp.predefined]/1](https://eel.is/c++draft/cpp.predefined#1)

Issue description:

Paragraph 1 begins with:
> The following macro names shall be defined by the implementation:

However, in the macro to test for the definition of std::float16_t it says:
> \_\_STDCPP_FLOAT16_T\_\_
>
> Defined as the integer literal 1 if and only if the implementation supports the ISO/IEC 60559 floating-point interchange format binary16 as an extended floating-point type ([basic.extended.fp]).

The text appears to say that \_\_STDCPP\_FLOAT16_T\_\_ must be defined, then it is only defined as 1 if the implementation supports std::float16_t. There are two possible interpretations of this that I can see: support for std::float16_t is mandatory, or if std::float16_t is not supported then it must be defined but as a value other than one. The former interpretation seems unintended, if it were the case such macros would be entirely useless. The latter interpretation also seems unintended, if it were the case then [[stdfloat.syn]/1](https://eel.is/c++draft/stdfloat.syn#1) would be incorrect.

The same applies to the other macros for the standard extended floating point types.

Suggested resolution:

Move the five macros \_\_STDCPP_FLOAT16_T\_\_, \_\_STDCPP_FLOAT32_T\_\_, \_\_STDCPP_FLOAT64_T\_\_, \_\_STDCPP_FLOAT128_T\_\_, and \_\_STDCPP_BFLOAT16_T\_\_ in [[cpp.predefined]/1](https://eel.is/c++draft/cpp.predefined#1) to the section below which contains conditional macros. Additionally, for each of these macros change the wording "Defined as the integer literal 1 if and only if" to "Defined, and has the value integer literal 1, if and only if" as in \_\_STDCPP_THREADS\_\_ to forbid definitions with a value other than 1 to indicate a lack of support. Alternatively, add "Otherwise, undefined." after each of theses macros.

### LINKED ISSUE #852: CWG3166 [expr.reflect] Clarify whether the protected access rule for a pointer-to-member applies to a reflect-expression
**Full name of submitter**: Aurelien Cassagnes

**Issue description**: It is not clear whether the following is valid:
```
class A {
protected:
  void f();
};

struct B : A {
    static constexpr auto r = ^^A::f;
};
```
According to [expr.reflect]/7.2, "overload resolution for the expression" `&A::f` "with no target shall select a unique function". Do the restrictions in [class.protected] on the use of the expression `&A::f` apply to the *reflect-expression*?

**Suggested resolution** (by Brian): Edit [expr.reflect]/7.2 as follows:

> Otherwise, if the id-expression denotes an overload set ~S~, ~overload resolution for the expression &S with no target shall select a unique function ([over.over])~**the expression `&` *id-expression* shall be well-formed when considered as an unevaluated operand**; R represents ~that~ **the** function **selected by overload resolution ([over.over])**.

### LINKED ISSUE #856: CWG3167 [dcl.init] No clear prohibition on initializing typedefs
Full name of submitter (unless configured in github; will be published with the issue): Bee

Reference (section label): [dcl.init], [dcl.decl.general]

Issue description: There is no clear prohibition on the declaration `typedef int x = 0;` in [dcl.init], [dcl.decl.general], or in [dcl.typedef]. [dcl.init] assumes that there is a variable to initialize, but it isn't entirely clear if that constitutes a requirement or non-applicability.

Suggested resolution: Require in [dcl.init] or [dcl.decl.general] explicitly that an initializer shall only appear on an object declaration.

### LINKED ISSUE #857: CWG3168 [conv.rank] Issues in defining integer conversion rank
Full name of submitter (unless configured in github; will be published with the issue): Jay Ghiron

Reference (section label): [[conv.rank]/1](https://eel.is/c++draft/conv.rank#1)

Issue description:

The first rule defining that no two signed integer types have the same conversion rank says that char is a signed integer type:
> No two signed integer types other than char and signed char (if char is signed) have the same rank, even if they have the same representation.

However, char is never a signed integer type even if it is signed. Even if it was a signed integer type, wchar_t can be signed too so it should be mentioned too.

The rule defining the conversion rank of bool allows it to have a greater rank than extended integer types, even those with a width greater than one.
> The rank of bool is less than the rank of all standard integer types.

It seems unintended to allow this as a possibility. C has adjusted its wording recently to require bool have the lowest conversion rank, see [N3747](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3747.pdf).

Suggested resolution:

Modify [[conv.rank]/1.1](https://eel.is/c++draft/conv.rank#1.1):
> No two signed integer types<del> other than char and signed char (if char is signed)</del> have the same rank, even if they have the same representation.

Modify [[conv.rank]/1.7](https://eel.is/c++draft/conv.rank#1.7):
> The rank of bool is less than the rank of all <del>standard</del><ins>other</ins> integer types.

Add a new footnote which is linked to at the end of [[conv.rank]/1.1](https://eel.is/c++draft/conv.rank#1.1):
> char and wchar_t are not signed integer types, even if they are signed.

### LINKED ISSUE #859: CWG3176 [intro.execution] p16 sequencing default arguments
Full name of submitter: Lénárd Szolnoki

Reference (section label): [intro.execution] p16

> During the evaluation of an expression as a core constant expression ([[expr.const]](https://eel.is/c++draft/expr.const)), evaluations of operands of individual operators and of subexpressions of individual expressions that are otherwise either unsequenced or indeterminately sequenced are evaluated in lexical order.

https://eel.is/c++draft/intro.execution#16

Issue description:

It is unclear how default arguments are sequenced according to this wording, as they don't appear lexically in a function call expression. The intent is probably to sequence these function arguments in parameter order.

My reading is that the sequencing of default arguments is either not specified by this paragraph, or they are usually sequenced before other arguments, as they lexically appear in the function declaration, which is usually lexically before the corresponding function call expression.

Suggested resolution:

> During the evaluation of an expression as a core constant expression ([[expr.const]](https://eel.is/c++draft/expr.const)), evaluations of operands of individual operators and of subexpressions of individual expressions that are otherwise either unsequenced or indeterminately sequenced are evaluated in lexical order. **For a function call ([expr.call]) the initialization of parameters are sequenced in the lexical order the parameters are declared.**

### LINKED ISSUE #1376: CWG2228 Ambiguity resolution for cast to function type
Migrating from https://isocpp.org/files/papers/P1018R18.html#issues
https://www.open-std.org/jtc1/sc22/wg21/docs/cwg_active.html#2228

Section: 9.3.3  [[dcl.ambig.res](https://wg21.link/dcl.ambig.res)]     Status: review     Submitter: Richard Smith     Date: 2016-02-02
Consider:
```
  int x = (int()) + 5; 
```
This is ill-formed, because 9.3.3 [[dcl.ambig.res](https://wg21.link/dcl.ambig.res#2)] paragraph 2 specifies:

An ambiguity can arise from the similarity between a function-style cast and a type-id. The resolution is that any construct that could possibly be a type-id in its syntactic context shall be considered a type-id.
and thus int() is interpreted as a type-id instead of as a function-style cast, so this is an ill-formed cast to a function type.

This seems to be the wrong disambiguation for all cases where there is a choice between a C-style cast and a parenthesized expression: in all those cases, the C-style cast interpretation results in a cast to a function type, which is always ill-formed.

Further, there is implementation divergence in the handling of this example:
```
  struct T { int operator++(int); T operator[](int); };
  int a = (T()[3])++; // not a cast
```

[Proposed resolution.](http://wiki.edg.com/pub/Wg21summer2020/CoreWorkingGroup/cwg2228.html?twiki_redirect_cache=5aceede89e9128019e50cb532244e26f)

C++ has a blanket disambiguation rule that says if a sequence of tokens can be interpreted as either a type-name or an expression, the type-name interpretation is chosen. This is unhelpful in cases like

```
    (T())+3
```

where the current rules make "(T())" a cast to the type "function with no parameters returning T" rather than a parenthesized value-initialization of a temporary of type T. The former interpretation is always ill-formed - you can’t cast to a function type - while the latter could be useful.

Richard’s proposed resolution, cited above, is to avoid the ambiguity by changing the grammar so that cases like "(T())" cannot be parsed as a cast. The wording in the proposal applies that change to a number of different contexts where the ambiguity can come into play. There are two contexts where the change is _not_ applied, however: 1) the operand of typeid, and 2) a template-argument. During our discussion yesterday, there was some support for the idea of applying the change to typeid, as well, although that would be a breaking change for any programs that rely on the current disambiguation to get type information for such function types. There was general agreement, however, to exclude template arguments, since they are used for things like std::function.

CWG would, therefore, like some guidance from EWG on two questions. First, should we apply the new syntax to the operand of typeid, even though it’s a breaking change?

More generally, the question was raised whether we should make this change at all. Although resolving the ambiguity in the other direction would arguably be more convenient in many cases, there is a tension between that convenience and the complexity of the language. In particular, we would be creating a situation where the exact same sequence of tokens would mean two different things, depending on the context in which they appear. Is the convenience worth the cost in complexity?

Meeting: ([note 2020-04-29](https://wiki.edg.com/bin/view/Wg21summer2020/EWG-IssueProcessing-29-Apr-2020), [notes 2020-03-23](https://wiki.edg.com/bin/view/Wg21summer2020/IssuesProcessingTeleconference2020-03-23), [note from Core summer 2020](https://wiki.edg.com/bin/view/Wg21summer2020/CoreWorkingGroup), [notes from Core Prague 2020](https://wiki.edg.com/bin/view/Wg21prague/CoreWorkingGroup), [notes from Core 2019-01-07](https://wiki.edg.com/bin/view/Wg21kona2019/IssuesProcessingTeleconference,2019-01-07), [notes from Core Kona 2019](https://wiki.edg.com/bin/view/Wg21kona2019/CoreWorkingGroup), [notes from Core Cologne 2019](https://wiki.edg.com/bin/view/Wg21cologne2019/CoreWorkingGroup)) This is an issue, we’d like to change the standard to resolve it: SF 0 F 6 N 4 A 3 SA 2

Davis [emailed EWG reflector](http://lists.isocpp.org/ext/2020/04/13473.php). Long discussion.

### LINKED ISSUE #9067: [2026-06 CWG Motion 1] P4271R0 Core Language Working Group "ready" Issues
https://wiki.isocpp.org/uploads/2026-06_Brno-StrawPolls_-_p4271r0.html

burblebee:
Fixes #9067.

Also fixes cplusplus/papers#2822

Also fixes cplusplus/papers#1376
Also fixes cplusplus/papers#2795

Also fixes cplusplus/CWG#660
Also fixes cplusplus/CWG#700
Also fixes cplusplus/CWG#732
Also fixes cplusplus/CWG#829
Also fixes cplusplus/CWG#843
Also fixes cplusplus/CWG#851
Also fixes cplusplus/CWG#852
Also fixes cplusplus/CWG#856
Also fixes cplusplus/CWG#856
Also fixes cplusplus/CWG#857
Also fixes cplusplus/CWG#859
Also fixes cplusplus/CWG#881
Also fixes cplusplus/CWG#889
Also fixes cplusplus/CWG#883
Also fixes cplusplus/CWG#886
Also fixes cplusplus/CWG#892
Also fixes cplusplus/CWG#891

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -2292,11 +2292,11 @@ The \term{associated namespaces} for a call are
 the innermost enclosing non-inline namespaces for its associated entities
 as well as every element of the inline namespace set\iref{namespace.def}
 of those namespaces.
-Argument-dependent lookup finds
-all declarations of functions and function templates that
+Argument-dependent lookup for a name $N$ finds
+all declarations of functions and function templates named $N$ that
 \begin{itemize}
 \item
-are found by a search of any associated namespace, or
+are found by a search for $N$ in any associated namespace, or
 \item
 are declared as a friend\iref{class.friend} of any class
 with a reachable definition in the set of associated entities, or
@@ -2307,7 +2307,7 @@ do not appear in the translation unit containing the point of the lookup, and
 have the same innermost enclosing non-inline namespace scope as
 a declaration of an associated entity attached to \tcode{M}\iref{basic.link}.
 \end{itemize}
-If the lookup is for a dependent name\iref{temp.dep,temp.dep.candidate},
+If $N$ is a dependent name\iref{temp.dep,temp.dep.candidate},
 the above lookup is also performed
 from each point in the instantiation context\iref{module.context} of the lookup,
 additionally ignoring any declaration that
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -6271,8 +6271,8 @@ has the top-level cv-qualifier \keyword{volatile}.
 Every integer type has an \term{integer conversion rank} defined as follows:
 
 \begin{itemize}
-\item No two signed integer types other than \keyword{char} and \tcode{\keyword{signed}
-\keyword{char}} (if \keyword{char} is signed) have the same rank, even if they have
+\item No two signed integer types
+have the same rank, even if they have
 the same representation.
 
 \item The rank of a signed integer type is greater than the rank
@@ -6290,18 +6290,12 @@ corresponding signed integer type.
 \item The rank of any standard integer type is greater than the
 rank of any extended integer type with the same width.
 
-\item The rank of \keyword{char} equals the rank of \tcode{\keyword{signed} \keyword{char}}
-and \tcode{\keyword{unsigned} \keyword{char}}.
-
 \item The rank of \tcode{bool} is less than the rank of all
-standard integer types.
+other integer types.
 
 \item
-\indextext{type!\idxcode{wchar_t}}%
-\indextext{type!\idxcode{char16_t}}%
-\indextext{type!\idxcode{char32_t}}%
-The ranks of \keyword{char8_t}, \keyword{char16_t}, \keyword{char32_t}, and
-\keyword{wchar_t} equal the ranks of their underlying
+The ranks of the character types
+equal the ranks of their corresponding underlying
 types\iref{basic.fundamental}.
 
 \item The rank of any extended signed integer type relative to another
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -7619,6 +7619,7 @@ If it is deferred,
 the initialization associated with the entity for thread \placeholder{t}
 is sequenced before the first non-initialization odr-use by \placeholder{t}
 of any non-inline variable with thread storage duration
+and dynamic initialization
 defined in the same translation unit as the variable to be initialized.
 It is \impldef{threads and program points at which deferred dynamic initialization is performed}
 in which threads and at which points in the program such deferred dynamic initialization occurs.
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -3942,7 +3942,7 @@ The lifetime of an object of type \tcode{T} begins when:
 \item storage with the proper alignment and size
   for type \tcode{T} is obtained, and
 \item its initialization (if any) is complete
-  (including vacuous initialization)\iref{dcl.init},
+  (including vacuous initialization)\iref{dcl.init,class.base.init},
 \end{itemize}
 except that if the object is a union member or subobject thereof,
 its lifetime only begins if that union member is the
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -6714,6 +6714,9 @@ evaluations of operands of individual operators and
 of subexpressions of individual expressions
 that are otherwise either unsequenced or indeterminately sequenced
 are evaluated in lexical order.
+For a function call\iref{expr.call},
+any default arguments used are evaluated in the order as if
+an argument were present for the corresponding parameter.
 
 \rSec2[intro.multithread]{Multi-threaded executions and data races}
 
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -3822,13 +3822,30 @@ see~\ref{expr.call}.
 If the
 \grammarterm{parameter-declaration-clause}
 is empty, the function takes no arguments.
-A parameter list consisting of a single unnamed non-object parameter of
-non-dependent type \keyword{void} is equivalent to an empty parameter
-list.
+A parameter list consisting of a single unnamed non-object parameter
+whose type is the same as\iref{temp.over.link} \keyword{void}
+is interpreted as an empty parameter list.
 \indextext{parameter!\idxcode{void}}%
 Except for this special case, a parameter shall not have type \cv{}~\keyword{void}.
 A parameter with volatile-qualified type is deprecated;
 see~\ref{depr.volatile.type}.
+\begin{example}
+\begin{codeblock}
+template<class T>
+struct S {
+  void f(std::void_t<T*>);      // \#1
+};
+S<int> x;                       // error: \#1 has a parameter of type \tcode{void}
+
+template<class T>
+struct S2 {
+  void f(std::void_t<int*>);    // \#2
+};
+S2<int> y;                      // OK, \#2 is a zero-parameter function
+\end{codeblock}
+\end{example}
+
+\pnum
 If the
 \grammarterm{parameter-declaration-clause}
 \indextext{argument type!unknown}%
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -2464,9 +2464,16 @@ auto j = 2.0;           // OK, \tcode{j} deduced to have type \tcode{double}
 \end{note}
 
 \pnum
-The optional \grammarterm{requires-clause} in an
+The \grammarterm{initializer} of an \grammarterm{init-declarator}
+shall not be present unless
+the declarator declares a variable and
+the host scope\iref{basic.scope.scope} of the declaration
+is the same as its target scope.
+
+\pnum
+The optional \grammarterm{requires-clause} of an
 \grammarterm{init-declarator} or \grammarterm{member-declarator}
-shall be present only if the declarator declares a
+shall not be present unlesss the declarator declares a
 templated function\iref{temp.pre}.
 %
 \indextext{trailing requires-clause@trailing \gterm{requires-clause}|see{\gterm{requires-clause}, trailing}}%
@@ -2494,8 +2501,8 @@ auto* p = new void(*)(char) requires true;  // error: not a function declaration
 
 \pnum
 The optional \grammarterm{function-contract-specifier-seq}\iref{dcl.contract.func}
-in an \grammarterm{init-declarator}
-shall be present only if
+of an \grammarterm{init-declarator}
+shall not be present unless
 the \grammarterm{declarator} declares a function.
 
 \pnum
@@ -5035,11 +5042,6 @@ Default arguments are more restricted; see~\ref{dcl.fct.default}.
 The order of initialization of variables with static storage duration is described in~\ref{basic.start}
 and~\ref{stmt.dcl}.
 \end{note}
-
-\pnum
-A declaration $D$ of a variable with linkage
-shall not have an \grammarterm{initializer}
-if $D$ inhabits a block scope.
 
 \pnum
 \indextext{initialization!default}%
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -8104,7 +8104,8 @@ as the corresponding value of the underlying type.
 
 \pnum
 Two enumeration types are \defnx{layout-compatible enumerations}{layout-compatible!enumeration}
-if they have the same underlying type.
+if they have the same underlying type
+and the same values.
 
 \pnum
 The value of an enumerator or an object of an unscoped enumeration type is
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -6707,6 +6707,11 @@ appear. That is, every value computation and side effect associated with a
 given \grammarterm{initializer-clause} is sequenced before every value
 computation and side effect associated with any \grammarterm{initializer-clause}
 that follows it in the comma-separated list of the \grammarterm{initializer-list}.
+If the \grammarterm{initializer-clause}{s} of
+the \grammarterm{initializer-list}
+are interpreted as arguments of a constructor call\iref{over.match.list},
+any default arguments used are evaluated in the order as if
+an \grammarterm{initializer-clause} were present for the corresponding parameter.
 \begin{note}
 This evaluation ordering holds regardless of the semantics of the
 initialization; for example, it applies when the elements of the
@@ -6714,6 +6719,22 @@ initialization; for example, it applies when the elements of the
 call, even though ordinarily there are no sequencing constraints on the
 arguments of a call.
 \end{note}
+\begin{example}
+\begin{codeblock}
+#include <iostream>
+struct I {
+  I(int x) { std::cout << x; }
+};
+struct S {
+  S(I, I = 2) {}
+};
+int main() {
+  S(1, 2); std::cout << '\n';   // unspecified order; prints \tcode{12} or \tcode{21}
+  S{1, 2}; std::cout << '\n';   // prints \tcode{12}
+  S{1};                         // prints \tcode{12}
+}
+\end{codeblock}
+\end{example}
 
 \pnum
 An object of type \tcode{std::initializer_list<E>} is constructed from
--- a/source/declarations.tex
+++ b/source/declarations.tex
@@ -1911,9 +1911,9 @@ a structured binding declaration\iref{dcl.struct.bind}.
 A placeholder type can also be used
 in the \grammarterm{type-specifier-seq} of
 the \grammarterm{new-type-id} or
-in the \grammarterm{type-id} of a
+in the \grammarterm{nofun-type-id} of a
 \grammarterm{new-expression}\iref{expr.new}.
-In such a \grammarterm{type-id},
+In such a \grammarterm{nofun-type-id},
 the placeholder type shall appear
 as one of the \grammarterm{type-specifier}{s} in
 the \grammarterm{type-specifier-seq} or
@@ -2288,7 +2288,7 @@ the type that replaces the placeholder shall be the same in each deduction.
 A placeholder for a deduced class type
 can also be used
 in the \grammarterm{type-specifier-seq}
-in the \grammarterm{new-type-id} or \grammarterm{type-id}
+in the \grammarterm{new-type-id} or \grammarterm{nofun-type-id}
 of a \grammarterm{new-expression}\iref{expr.new},
 as the \grammarterm{simple-type-specifier}
 in an explicit type conversion (functional notation)\iref{expr.type.conv},
@@ -2529,7 +2529,7 @@ Declarators have the syntax
 
 \begin{bnf}
 \nontermdef{trailing-return-type}\br
-    \terminal{->} type-id
+    \terminal{->} nofun-type-id
 \end{bnf}
 
 \begin{bnf}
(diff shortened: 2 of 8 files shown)


eisenwave:
```
Also fixes https://github.com/cplusplus/papers/issues/1376
Also fixes https://github.com/cplusplus/CWG/issues/660
Also fixes https://github.com/cplusplus/CWG/issues/700
Also fixes https://github.com/cplusplus/CWG/issues/732
Also fixes https://github.com/cplusplus/CWG/issues/829
Also fixes https://github.com/cplusplus/CWG/issues/843
Also fixes https://github.com/cplusplus/CWG/issues/851
Also fixes https://github.com/cplusplus/CWG/issues/852
Also fixes https://github.com/cplusplus/CWG/issues/856
Also fixes https://github.com/cplusplus/CWG/issues/856
Also fixes https://github.com/cplusplus/CWG/issues/859
Also fixes https://github.com/cplusplus/CWG/issues/881
Also fixes https://github.com/cplusplus/CWG/issues/889
Also fixes https://github.com/cplusplus/CWG/issues/883
Also fixes https://github.com/cplusplus/CWG/issues/886
Also fixes https://github.com/cplusplus/CWG/issues/892
Also fixes https://github.com/cplusplus/CWG/issues/891
Also fixes https://github.com/cplusplus/papers/issues/2795
```

source/basic.tex
```diff
@@ -3942,7 +3942,7 @@
 \item storage with the proper alignment and size
   for type \tcode{T} is obtained, and
 \item its initialization (if any) is complete
-  (including vacuous initialization)\iref{dcl.init},
+  (including vacuous initialization)\iref{dcl.init, class.base.init},
```
jensmaurer: We usually don't have a space inside a multi-prong \iref . (I didn't even know it would work.)

source/preprocessor.tex
```diff
@@ -2457,6 +2417,46 @@
 The following macro names are conditionally defined by the implementation:
 
 \begin{description}
+\item
+\indextext{__stdcpp_float16_t__@\mname{STDCPP_FLOAT16_T}}%
+\mname{STDCPP_FLOAT16_T}\\
+Defined, and has the value integer literal \tcode{1}
```
jensmaurer: add comma at end of line (it's dark green in the incoming diff)

source/preprocessor.tex
```diff
@@ -2457,6 +2417,46 @@
...
+if and only if the implementation supports
+the \IsoFloatUndated{} floating-point interchange format binary16
+as an extended floating-point type\iref{basic.extended.fp}.
+
+\item
+\indextext{__stdcpp_float32_t__@\mname{STDCPP_FLOAT32_T}}%
+\mname{STDCPP_FLOAT32_T}\\
+Defined, and has the value integer literal \tcode{1}
```
jensmaurer: add comma at end of line (it's dark green in the incoming diff)

source/preprocessor.tex
```diff
@@ -2457,6 +2417,46 @@
...
+if and only if the implementation supports
+the \IsoFloatUndated{} floating-point interchange format binary32
+as an extended floating-point type.
+
+\item
+\indextext{__stdcpp_float64_t__@\mname{STDCPP_FLOAT64_T}}%
+\mname{STDCPP_FLOAT64_T}\\
+Defined, and has the value integer literal \tcode{1}
```
jensmaurer: add comma at end of line (it's dark green in the incoming diff)

source/preprocessor.tex
```diff
@@ -2457,6 +2417,46 @@
...
+if and only if the implementation supports
+the \IsoFloatUndated{} floating-point interchange format binary64
+as an extended floating-point type.
+
+\item
+\indextext{__stdcpp_float128_t__@\mname{STDCPP_FLOAT128_T}}%
+\mname{STDCPP_FLOAT128_T}\\
+Defined, and has the value integer literal \tcode{1}
```
jensmaurer: add comma at end of line (it's dark green in the incoming diff)

source/preprocessor.tex
```diff
@@ -2457,6 +2417,46 @@
...
+if and only if the implementation supports
+the \IsoFloatUndated{} floating-point interchange format binary128
+as an extended floating-point type.
+
+\item
+\indextext{__stdcpp_bfloat16_t__@\mname{STDCPP_BFLOAT16_T}}%
+\mname{STDCPP_BFLOAT16_T}\\
+Defined, and has the value integer literal \tcode{1}
```
jensmaurer: add comma at end of line (it's dark green in the incoming diff)

source/declarations.tex
```diff
@@ -3822,13 +3822,30 @@
...
-non-dependent type \keyword{void} is equivalent to an empty parameter
-list.
+A parameter list consisting of a single unnamed non-object parameter
+whose type is the same as\iref{temp.over.link} \keyword{void}
+is interpreted as an empty parameter list.
 \indextext{parameter!\idxcode{void}}%
 Except for this special case, a parameter shall not have type \cv{}~\keyword{void}.
 A parameter with volatile-qualified type is deprecated;
 see~\ref{depr.volatile.type}.
+
```
jensmaurer: Remove the empty line here.

source/declarations.tex
```diff
@@ -3822,13 +3822,30 @@
...
+S<int> x;                       // error: \#1 has a parameter of type \tcode{void}
+
+template<class T>
+struct S2 {
+  void f(std::void_t<int*>);    // \#2
+};
+S2<int> y;                      // OK, \#2 is a zero-parameter function
+\end{codeblock}
+\end{example}
+
```
jensmaurer, line 3847: add \pnum here, the instruction says "... and split into two paragraphs"

source/expressions.tex
```diff
@@ -6872,11 +6873,32 @@
 Lookup never finds a partial or explicit specialization.
 \end{note}
 \end{itemize}
+
```
jensmaurer: We have no empty line in front of the other \items, so remove here, please.

jensmaurer (changes requested):

source/basic.tex
burblebee: heh :) fixed

source/preprocessor.tex
burblebee: thanks, fixed

source/expressions.tex
burblebee: I've assumed you also wanted the one added afterwards to be removed as well - done. Given that it's such a long \item, I would have preferred to leave them in (think of the poor editor who has to edit this in the future!), alas.

jensmaurer (approved):

source/expressions.tex
```diff
@@ -6930,9 +6930,11 @@
 then $R$ is ill-formed.
 \item
 Otherwise, if the \grammarterm{id-expression} denotes an overload set $S$,
-overload resolution for the expression \tcode{\&S} with no target
-shall select a unique function\iref{over.over};
-$R$ represents that function.
+the expression \tcode{\& grammarterm{id-expression}} shall be well-formed
```
tkoeppe: Will fix.

PR #9148, opened 2026-06-19
[CWG 2026-06-7] P2434R5 Nondeterministic pointer provenance

### LINKED ISSUE #931: P2434 [basic.types.general] [basic.compound] Pointer values are not determined by bits in the object representation
**Full name of submitter:** Andrey Erokhin

**Reference (section label):** [basic.types.general] [basic.compound]

**Link to reflector thread (if any):** -

**Issue description:**

(Wanted to file the issue for many years)

[basic.types.general] says:
> For trivially copyable types, the value representation is a set of bits in the object representation that **determines** [emphasis mine] a [value](http://eel.is/c++draft/basic.types.general#def:value), which is one discrete element of an implementation-defined set of values[.](http://eel.is/c++draft/basic.types.general#4.sentence-7)

For pointers, bits in their representation can't **determine** the value.

A simple example:
```c++
int a[2][2];

int* p1 = &a[0][2];
int* p2 = &a[1][0];
```
`p1` and `p2` usually have the same bits in their representation, but they are different values.

And, ofc, one can produce arbitrary number of pointer values with the same representation, but different values by replacing an object:
```c++
int i;

int* p1 = &i;
std::byte buf[sizeof(p1)];
std::memcpy(buf, &p1, sizeof(p1)); // save the value

int* p2 = ::new(&i) int; // rebinds p1 (is it a write to p1?)
std::memcpy(&p1, buf, sizeof(p1)); // restore p1 value
```
`p1` and `p2` point to different objects (thus have different values), but bits in their representation are usually the same

**Suggested resolution:**

Exclude pointers from trivially copyable types

### LINKED ISSUE #1364: P2434 R1 Nondeterministic pointer provenance
[P2434R0](https://wg21.link/P2434R0) Nondeterministic pointer provenance (S. Davis Herring)

### LINKED ISSUE #9074: [2026-06 CWG Motion 7] P2434R5 Nondeterministic pointer provenance
[P2434R5](https://wiki.isocpp.org/uploads/2026-06_Brno-StrawPolls_-_p2434r5.html) (Nondeterministic pointer provenance)

burblebee:
Fixes #9074.

Also fixes cplusplus/papers#1364
Also fixes cplusplus/CWG#931

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -5172,57 +5172,6 @@ or functions\iref{dcl.fct}.
 \end{note}
 
 \pnum
-\indextext{object!byte copying and|(}%
-\indextext{type!trivially copyable}%
-For any object (other than a potentially-overlapping subobject) of trivially copyable type
-\tcode{T}, whether or not the object holds a valid value of type
-\tcode{T}, the underlying bytes\iref{intro.memory} making up the
-object can be copied into an array of
-\keyword{char},
-\tcode{\keyword{unsigned} \keyword{char}}, or
-\tcode{std::byte}\iref{cstddef.syn}.
-\begin{footnote}
-By using, for example, the library
-functions\iref{headers} \tcode{std::memcpy} or \tcode{std::memmove}.
-\end{footnote}
-If the content of that array
-is copied back into the object, the object shall
-subsequently hold its original value.
-\begin{example}
-\begin{codeblock}
-constexpr std::size_t N = sizeof(T);
-char buf[N];
-T obj;                          // \tcode{obj} initialized to its original value
-std::memcpy(buf, &obj, N);      // between these two calls to \tcode{std::memcpy}, \tcode{obj} might be modified
-std::memcpy(&obj, buf, N);      // at this point, each subobject of \tcode{obj} of scalar type holds its original value
-\end{codeblock}
-\end{example}
-
-\pnum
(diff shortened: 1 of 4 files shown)


source/basic.tex
```diff
@@ -5424,6 +5369,88 @@
...
+Each trivially copyable type \tcode{T} has an
+\impldef{values of a trivially copyable type}
+set of discrete \defnx{values}{value}.
+\indextext{value!representation}%
+Each possible value representation of an object of type \tcode{T}
+corresponds to a distinct
+\impldef{subset of values of a trivially copyable type
+         for which a value representation corresponds to}
+subset of this set.
+%%FIXME: "*the* set of values" - what set of values?
```
jensmaurer: See above: "each type T has a set of (discrete) values"
(those are values in the abstract sense)

burblebee: It's not clear as is - too many "value"s have been discussed since then. Introduce a placeholder S to be that "set of discrete values" and refer to that here.

tkoeppe: In symbols: $\bigcup_R S_R = S$, where $S$ is the set of values of a type `T`, the union ranges over all object representations $R$ of the `T`, and $S_R$ is the "distinct, impl-def" subset for that representation. Morever, $\lvert S_R \rvert = 1$ when `T` is a scalar type other than a pointer type.

But I think the wording is quite clear.

PR #9155, opened 2026-06-20
[simd] Remove unnecessary `\tcode` around `\exposid`

eisenwave:
- Branching off from https://github.com/cplusplus/draft/pull/9146
- The current check doesn't work because a hyphen character is not matched.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -17724,8 +17724,8 @@ The class template \tcode{flags} acts like an integer bit-flag for types.
 
 \pnum
 \constraints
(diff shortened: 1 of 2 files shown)


jensmaurer (approved):

jensmaurer:
For the commit  "Also diagnose \tcode{\exposid{kebap-case}}, i.e. hyphenated identifiers "

please use "[check]" in front + force-push.

eisenwave:
> please use "[check]" in front + force-push.

Done.

PR #9159, opened 2026-06-21
[vector.capacity] vector::reserve is tutorial-like

### LINKED ISSUE #7401: [vector.capacity] vector::reserve is tutorial-like
> _Effects_: A directive that informs a `vector` of a planned change in size, so that it can manage the storage
allocation accordingly. 

This is not a description of what the function _does_, it's a description of what the function _is_.

Maybe just strike it.

For `hive` Matt originally copied this text, but we're suggesting a different form using "If `n <= capacity()` there are no effects." and "_Postconditions_: `capacity() >= n` is `true`. But that would be a topic for an LWG issue. Striking the tutorial-like first sentence seems editorial.

notdanhan:
Fixes #7401

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -10205,9 +10205,6 @@ constexpr void reserve(size_type n);
 
 \pnum
 \effects
-A directive that informs a
-\tcode{vector}
-of a planned change in size, so that it can manage the storage allocation accordingly.
 After
 \tcode{reserve()},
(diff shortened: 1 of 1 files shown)


source/containers.tex
notdanhan, line 10213: I find the wording here to be a bit obtuse, but that's for another time.

jwakely (approved):

PR #8035, opened 2025-07-01
[2025-06-lwg-24] P3557R3 High-Quality Sender Diagnostics with Constexpr Exceptions

burblebee:
* [exec.util.cmplsig] Subclause (now without siblings) promoted to \rSec1 (replacing [exec.util]) and renamed to [exec.cmplsig] "Completion signatures".
* [exec.snd.expos]p24 Added missing \expos comments.
* FIXED: [exec.snd.expos]p39 Unable to find paragraph to replace. Referenced paragraph starts: "Let Sndr be a (possibly const-qualified) specialization basic-sender" Inserted changes as new text after [exec.snd.expos]p41 instead.
* FIXED: [exec.write.env] Unable to apply changes - section does not exist. Referenced paragraph starts: "write_env is a sender adaptor that accepts a sender and a queryable object"

Also fixes LWG4203.

Fixes #7961.

PR #9160, opened 2026-06-21
[exec.stopped.opt] Add transform_sender tag

abhinavagarwal07:
Pass `set_value` to `stopped_as_optional.transform_sender` in the not-a-sender branch, matching the tagged call used by the guard in the same paragraph.

### DIFF
--- a/source/exec.tex
+++ b/source/exec.tex
@@ -5117,7 +5117,7 @@ is ill-formed;
 otherwise,
 if \tcode{\libconcept{sender_in}<\exposid{child-type}<Sndr>, \exposid{FWD-\brk{}ENV-\brk{}T}(Env)>}
 is \tcode{false},
-the expression \tcode{stopped_as_optional.transform_sender(sndr, env)}
+the expression \tcode{stopped_as_optional.transform_sender(set_value, sndr, env)}
 is equivalent to \tcode{\exposid{not-a-sen\-der}()};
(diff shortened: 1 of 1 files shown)


eisenwave (approved):
Problem introduced with https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p3557r3.html / https://github.com/cplusplus/draft/pull/8035

The added wording says `stopped_as_optional.transform_sender(sndr, env)`, but `transform_sender` has three parameters.

I think this came to be because P3557R3 cited an old version of pre-existing wording where it's still `transform_sender(sndr, env)`, so LWG didn't notice that the new (green) wording should not have two arguments in the call expression, but three. Or maybe it's the result of applying multiple papers during the same motions; I'd have to dig further.

eisenwave:
@abhinavagarwal07 in the future, could you provide the necessary historical context on these PRs?

It's important to know whether something is a misapplication of a paper or a problem introduced by paper/LWG issue, etc. The fixes are greatly appreciated though.

abhinavagarwal07:
@eisenwave will do. Thanks for the feedback.

jensmaurer (approved):
This looks good.

@jwakely ?

abhinavagarwal07:
gentle ping @jwakely

jwakely (approved):

PR #9161, opened 2026-06-21
[class.mem.general] Add ANN to data member description note

abhinavagarwal07:
Add the missing `ANN` bullet to the explanatory note for data member descriptions, matching the annotation component already present in the sextuple definition.


eisenwave (approved):
My impression is that when https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2026/p3795r2.html extended the definition of data member description from "pentuple" to "sextuple", it should have also updated the note below, but this was forgotten during CWG review.

source/classes.tex
jensmaurer, line 968: That "and" here needs to be stricken to keep the overall structure.

jensmaurer (changes requested):

PR #9164, opened 2026-06-27
[intro.structure] should mention Annex E

### LINKED ISSUE #7076: [intro.structure] should mention Annex E
[intro.structure] describes the structure of the standard and its annexes, but was not updated when Annex E was added.

notdanhan:
Fixes https://github.com/cplusplus/draft/issues/7076

I tried to condense the wording from [uaxid.general] without just outright copying it so, hopefully that is a fine thing to do :)

### DIFF
--- a/source/intro.tex
+++ b/source/intro.tex
@@ -1066,6 +1066,12 @@ implementations.
 published description, and explains in detail the differences between
 \Cpp{} and C\@. Certain features of \Cpp{} exist solely for compatibility
 purposes; \ref{depr} describes those features.
+
+\pnum
+\ref{uaxid} provides an outline of how \UAX{31}
+("Unicode Identifier and Pattern Syntax") has been applied to \Cpp{}.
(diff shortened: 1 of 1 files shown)


source/intro.tex
```diff
@@ -1066,6 +1066,12 @@
 published description, and explains in detail the differences between
 \Cpp{} and C\@. Certain features of \Cpp{} exist solely for compatibility
 purposes; \ref{depr} describes those features.
+
+\pnum
+\ref{uaxid} provides an outline of how \UAX{31}
+("Unicode Identifier and Pattern Syntax") has been applied to \Cpp{}.
```
tkoeppe: Can you use proper LaTeX-style quotation marks here?

tkoeppe:
```
Like ``this''.
```

jensmaurer (approved):
Annex E is on the low end of the scale, and this summary is par for the course.

Let's do it.

PR #9165, opened 2026-06-29
[namespace.std] Allow rejection of programs when they add declarations to the namespace std

TymianekPL:
Changing the behaviour from undefined to ill-formed (no diagnostics required) in [namespace.std] when a program attempts to add a declaration to the namespace std.

### DIFF
--- a/source/lib-intro.tex
+++ b/source/lib-intro.tex
@@ -3009,10 +3009,10 @@ installation of handler functions during execution\iref{handler.functions}.
 
 \pnum
 Unless otherwise specified,
-the behavior of a \Cpp{} program is undefined if it adds declarations or definitions to namespace
+a \Cpp{} program shall not add declarations or definitions to namespace
 \tcode{std}
 or to a namespace within namespace
-\tcode{std}.
+\tcode{std}; no diagnostic is required.
 
 \pnum
 Unless explicitly prohibited,


jensmaurer:
This change is not editorial.

Please file an LWG issue. (I do agree with the change, though.)

TymianekPL:
The issue has been filed. I assume I should file another LWG issue to change "declarations or definitions" to "declarations" rather than make a PR?
I already made the commit in a cplusplus/draft fork, so just making sure ^^

frederick-vs-ja:
I think there're many other occurrences of "undefined behavior" in the library wording intend to allow, but not require, implementations to reject compilation. So, "ill-formed, no diagnostic required" should be used instead.

LWG3142 is addressing one such case, although I think we should the make whole [[res.on.functions]](https://eel.is/c++draft/res.on.functions) say IFNDR.

TymianekPL:
I filed an LWG yesterday correcting namespace std, posix, future namespaces, reserved items in the standard library and header contents to say ill-formed, no diagnostics required ^^

tkoeppe:
Thanks! If this is tracked as an LWG issue, let's close this (editorial) issue then, since indeed the proposed change is not editorial.

TymianekPL:
> Thanks! If this is tracked as an LWG issue, let's close this (editorial) issue then, since indeed the proposed change is not editorial.

My LWG issue was closed as NAD
I'm working on a bigger proposal about this and many other issues related to LWG use of "undefined behaviour" and such because I certainly do not agree nor have the will to fight in e-mail about it ^^

Thank you 💖

jwakely:
To be precise, no issue was opened, because the description of the issue was incorrect and the proposed resolution made no normative change.

PR #9176, opened 2026-07-13
[dcl.contract.func] Adjust linebreak in example

Andreas-Krug:


eisenwave:
Why is this necessary? Do we get a build failure or overfull box or something without this adjustment or is it purely aesthetic?

tkoeppe:
No, the comment just breaks onto a new line, which is not intended.

jensmaurer (approved):

PR #9181, opened 2026-07-15
[bit.permute] Create notation for the popcount of the first n bits.

tkoeppe:
This avoids having to put a complex expression into sub- and superscripts in the definitions of "expand" and "compress".

### DIFF
--- a/source/utilities.tex
+++ b/source/utilities.tex
@@ -16733,12 +16733,17 @@ The number of \tcode{1} bits in the value of \tcode{x}.
 
 \pnum
 In the following descriptions,
+\begin{itemize}
+\item
 let $N$ denote the value of \tcode{numeric_limits<T>::digits},
+\item
 let $\alpha_n$ denote the value of the $n^\text{th}$ least significant bit
 in the base-2 representation of an integer $\alpha$,
-so that $\alpha$ equals $\sum_{n=0}^{N-1} \alpha_n 2^n$,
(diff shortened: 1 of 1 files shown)


source/utilities.tex
```diff
@@ -16733,10 +16733,12 @@
...
 \pnum
 In the following descriptions,
-let $N$ denote the value of \tcode{numeric_limits<T>::digits}, and
+let $N$ denote the value of \tcode{numeric_limits<T>::digits},
 let $\alpha_n$ denote the value of the $n^\text{th}$ least significant bit
 in the base-2 representation of an integer $\alpha$,
-so that $\alpha$ equals $\sum_{n=0}^{N-1} \alpha_n 2^n$.
+so that $\alpha$ equals $\sum_{n=0}^{N-1} \alpha_n 2^n$,
+and let $\sigma(\alpha, n) = \sum_{k=0}^{n-1} \alpha_k$
+be the count of one-bits among the lowest $n$ bits of $\alpha$.
```
eisenwave, line 16745: If this is going to be three relatively complex items, maybe it this should be using bullets.

tkoeppe: Done as a separate commit. I'm not sure that it makes a big difference. (Generally, what helps make complex expressions more scannable is to put them in a displayed equation, but I don't really see any good opportunities for this here.)

tkoeppe: image

jwakely (approved):

PR #9186, opened 2026-07-16
[ub:expr.mul.div.by.zero] Clarify that division by zero is always UB for floating-point types

eisenwave:


notadragon (approved):
This is fine, though there is part of me that would prefer to see any substantive changes to the annex that come now rolled into https://isocpp.org/files/papers/D4284R0.pdf so that people can use it as a reference.

PR #9187, opened 2026-07-16
[ub:expr.shift.neg.and.width] Use std::uint32_t instead of int in example

eisenwave:
Firstly, I find it unnatural to use `int` and then have to assert various properties to get 32-bit type.

Secondly, the example is incorrect (well, not really, it's still true that a 32-bit shift is UB) because `sizeof(int) == 4 && CHAR_BIT == 8` is not a sufficient assertion. It's possible that `int` is a 16-bit type, with two 8-bit padding bytes and two 8-bit value bytes.

Alternatively, we could also use `numeric_limits<int>::digits` in the assertion or just put the assumption about the width of `int` into prose.

### DIFF
--- a/source/ub.tex
+++ b/source/ub.tex
@@ -1281,12 +1281,12 @@ the bit-width of a type has undefined behavior.
 
 \pnum
 \begin{example}
+The following example assumes that \tcode{std::uint32_t} is supported\iref{cstdint.syn}.
 \begin{codeblock}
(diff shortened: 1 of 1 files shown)


notadragon (commented):
We should have the `#include <cstdint>`.   I would also prefer that we roll this into https://isocpp.org/files/papers/D4284R0.pdf for discoverability, but otherwise this is fine.

PR #9188, opened 2026-07-16
[ub:expr.add.out.of.bounds] Merge the two examples into one

eisenwave:
1. It's unclear why we need two example blocks. That seems like unnecessary fluff.
2. The name `arrs` for the first array doesn't make sense; it should just be `arr` because it's not several arrays (array of arrays), so it shouldn't have a plural name.

### DIFF
--- a/source/ub.tex
+++ b/source/ub.tex
@@ -1208,19 +1208,11 @@ Creating an out of bounds pointer has undefined behavior.
 \pnum
 \begin{example}
 \begin{codeblock}
-static const int arrs[10]{};
-
-int main() {
-  const int *y = arrs + 11;             // undefined behavior, creating an out of bounds pointer
-}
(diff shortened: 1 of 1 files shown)


notadragon (commented):
This is fine, but again i would prefer merging into https://isocpp.org/files/papers/D4284R0.pdf.

(I could also just generate that paper out of all changes we make to the annexes and we cna land the changes themselves as individual PRs... but it depends on how wide we want reviewership for those entries to be)

tkoeppe:
Yes, please, I'd much prefer if you liaise with Joshua to make this part of the post-meeting update work.

eisenwave:
Since you have a paper laying around that does these changes in bulk, I agree it's better to address these issues holistically in CWG.

PR #9191, opened 2026-07-16
[mdspan.copy] Fix singular/plural mismatch

Andreas-Krug:


tkoeppe:
Thanks! That's an error in the paper.

PR #9192, opened 2026-07-16
[thread.{thread.constr,jthread.cons}] Move the "existence of $i$ to Mandates:.

tkoeppe:
This change separates the local notational definitions from the requirement whose violation makes the program ill-formed: the latter should probably appear in one of the standard Elements rather than in a rogue paragraph.


jwakely (approved):
Typo in the commit message ("notational", I assume?) but the wording changes look good.

tkoeppe:
D'oh, thanks, fixed!

PR #9122, opened 2026-06-14
[2026-06 LWG Motion 8] P3154R3 (Deprecating signed character types in iostreams)

notdanhan:
Fixes #9095
Also Fixes https://github.com/cplusplus/papers/issues/1804

PR #9194, opened 2026-07-17
[depr.iterator] and [depr.ostream.inserters]: use \rSec instead of \Sec

notadragon:
https://github.com/cplusplus/draft/pull/9122 used \Sec1 commands instead of \rSec1, inconsistent with all other sectioning commands in the Standard.  This PR corrects that.

### DIFF
--- a/source/future.tex
+++ b/source/future.tex
@@ -1059,7 +1059,7 @@ template<class T> constexpr T kill_dependency(T y) noexcept;
 \tcode{y}.
 \end{itemdescr}
 
-\Sec1[depr.istream.extractors]{Deprecated \tcode{signed char} and \tcode{unsigned char} extraction}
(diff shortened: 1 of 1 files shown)


tkoeppe:
Good catch, thank you!

@zygoloid Did we ever use the non-relative versions of the headings for anything? Or rather, did we ever use the relative versions with anything other than the default offset?

Maybe we should just remove those redundant macros entirely.

zygoloid:
> @zygoloid Did we ever use the non-relative versions of the headings for anything? Or rather, did we ever use the relative versions with anything other than the default offset?

I don't recall them ever being anything other than a nuisance in the standard itself. Though I could imagine them getting used in a TS that wants to incorporate a section of the standard as a subsection of the TS's wording section?

PR #9196, opened 2026-07-17
[basic.start.term] Add period to end of sentence

Andreas-Krug:


tkoeppe:
The deletion of the period was explicitly requested by https://open-std.org/jtc1/sc22/wg21/docs/papers/2026/p3596r3.pdf and approved by CWG. @joshuaberne would you be able to explain this change?

notadragon:
This was a typo. We had put a \ubdef there, but because it is basically a library function precondition being restated we removed the ubdef (and the . got removed along with it accidentally).

tkoeppe:
Thanks!

PR #9198, opened 2026-07-17
[simd.mask.overview] Put content after \recommended onto a separate line

eisenwave:
This is necessary to keep the source fixes separate from #9146.


Tsche (approved):
lgtm

PR #9201, opened 2026-07-17
[simd.math] Fix various indexing problems

### LINKED ISSUE #9200: [simd.math] Stray indexing of `riemann_zeta`
https://github.com/cplusplus/draft/blob/ed83a61c91dc9e476953293f269ad016232fa94c/source/numerics.tex#L20277-L20294

Notice that there is indexing for `riemann_zeta` here, but it's not in the code block.

eisenwave:
Fixes #9200

1. There was indexing of `simd::riemann_zeta` without any declaration in the code block below.
2. Indexing for `simd::remquo` was missing.
3. Various declaration blocks were overly large for out-of-line indexing. For consistency, all blocks declaring multiple functions were converted to inline indexing. This change is in the same style as #9163 If you look at the PDF, you will find that several of the code blocks in [simd.math] span multiple pages, so the library index ends up pointing to the wrong page, sometimes off by multiple pages.

### DIFF
--- a/source/numerics.tex
+++ b/source/numerics.tex
@@ -19753,110 +19753,77 @@ lookup\iref{basic.lookup.argdep} contrary to \ref{contents}.
 
 \rSec3[simd.math]{Mathematical functions}
 
-\indexlibrarymember{ilogb}{simd}
-\indexlibrarymember{abs}{simd}
-\indexlibrarymember{fabs}{simd}
-\indexlibrarymember{ceil}{simd}
-\indexlibrarymember{floor}{simd}
-\indexlibrarymember{nearbyint}{simd}
-\indexlibrarymember{rint}{simd}
-\indexlibrarymember{lrint}{simd}
-\indexlibrarymember{llrint}{simd}
-\indexlibrarymember{round}{simd}
-\indexlibrarymember{lround}{simd}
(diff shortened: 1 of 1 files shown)


mattkretz (commented):
Looks good from what I can see on my phone (on vacation).

tkoeppe:
You're taking a vacation, in the moment where we need to index zeta functions?!

tkoeppe:
@eisenwave could you run a visual diffpdf before/after?

eisenwave:
> @eisenwave could you run a visual diffpdf before/after?

grafik

no difference

tkoeppe:
Wait, how can there be no difference if we were previously indexing something that didn't exist? Wouldn't that have shown up erroneously in the index?

tkoeppe:
Oh, you've restricted the page numbers. Can you run the full diff, and show where the index starts to differ?

eisenwave:
grafik

Just kidding; I need to rerun some stuff.

eisenwave:
grafik


The first index difference is for `simd::beta`, which got shifted from 1928 to 1929.

jensmaurer (approved):
Looks good to me.

Indexing change only, so I'll merge right away.

PR #9202, opened 2026-07-17
[container.requirements] Add consistent punctuation.

tkoeppe:


tkoeppe:
@notdanhan I realize we're currently being inconsistent, but I think everything should be punctuated. All these elements essentially sound like statements, even if some of them are very brief.

notdanhan (approved):

jensmaurer (commented):
I'm okay with these changes (although a lone keyword with a period following feels odd), but @jwakely should have a say here.

tkoeppe:
@jensmaurer We already have many lone keywords followed by a period. There was absolutely no consistency or majority before!

jwakely (approved):

PR #9212, opened 2026-07-19
[ptrtag.pair.get] Add missing template parameter `I`

hewillk:
I'm assuming this is a typo in the paper.

### DIFF
--- a/source/memory.tex
+++ b/source/memory.tex
@@ -562,7 +562,7 @@ namespace std {
       using type = TagT;
     };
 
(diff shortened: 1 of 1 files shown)


eisenwave (approved):
Seems like a typo caused by copying and pasting in https://github.com/cplusplus/papers/issues/1903 without double-checking that the *template-head* makes sense.

tkoeppe:
Yes, looks like a mistake in the paper. Thanks for spotting!

jwakely (approved):

PR #9217, opened 2026-07-20
Improve sentence structures by removing colons

tkoeppe:
We are dealing with long, complete sentences, where it does not seem appropriate to separate parts with colons.

I suggest various rewrites that incorporate the complex object of the sentence in a more natural way.

In [ranges], I would also like to change "for each" to "for all" in the description of the argument of the "logical AND".


jwakely (approved):

jensmaurer (approved):

tkoeppe:
I made a further change to a) include another case I had missed, and b) to replace the itemizations with codeblocks:

image

I think that works a lot better and also avoids the need for awkward manual linebreaks.

PR #9225, opened 2026-07-23
[tools/check] Combine grep/sed calls using "-e".

tkoeppe:
This reduces the number of subprocesses that are spawned.

### DIFF
--- a/tools/check-output.sh
+++ b/tools/check-output.sh
@@ -73,14 +73,14 @@ for f in *.tex; do
     # handle codeblock
     sed -n 's,//.*$,,;s/%.*$//;s/"[^"]*"/""/;/begin{codeblock\(tu\)\?}/,/end{codeblock\(tu\)\?}/{/[^-_a-z\]\('"$patt"'\)[^-_}a-z0-9();,]/{=;p;};}' $f |
 	# prefix output with filename and line
-	sed '/^[0-9]\+$/{N;s/\n/:/;}' | sed "s/.*/$f:&/" |
+	sed -e '/^[0-9]\+$/{N;s/\n/:/;}' -e "s/.*/$f:&/" |
 	grep -v "@.seebelow" |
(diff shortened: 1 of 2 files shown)


jwakely (approved):

tools/check-output.sh
```diff
@@ -73,14 +73,14 @@ for f in *.tex; do
...
-	sed '/^[0-9]\+$/{N;s/\n/:/;}' | sed "s/.*/$f:&/" |
+	sed -e '/^[0-9]\+$/{N;s/\n/:/;}' -e "s/.*/$f:&/" |
 	grep -v "@.seebelow" |
 	sed "s/\$/ -- concept name without markup/" |
 	fail || failed=1
     # handle itemdecl
     sed -n 's,//.*$,,;s/%.*$//;s/"[^"]*"/""/;/begin{itemdecl}/,/end{itemdecl}/{/[^-_a-z]\('"$patt"'\)[^-_a-z();,]/{/concept{[a-z_-]*}/d;=;p;};}' $f |
 	# prefix output with filename and line
-	sed '/^[0-9]\+$/{N;s/\n/:/;}' | sed "s/.*/$f:&/" |
+	sed -e '/^[0-9]\+$/{N;s/\n/:/;}' -e "s/.*/$f:&/" |
```
notadragon: You can do a grep -v with sed too.

```
	sed -e '/^[0-9]\+$/{N;s/\n/:/;}' -e  "s/.*/$f:&/" \
	  -e "/@.seebelow/d" \
	  -e "s/\$/ -- concept name without markup/" |
	fail || failed=1
```

tkoeppe: Oh, just by deleting the matching lines? Not bad!

jensmaurer (approved):

tools/check-output.sh
tkoeppe: @notadragon PTAL!

PR #9227, opened 2026-07-24
[fs.path.member] Fix indexing for `(generic_)native_encoded_string`

### LINKED ISSUE #9226: [diff] Add entry for removed std::filesystem::path::string function
[P2319R5](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p2319r5.html) was accepted which removed `std::filesystem::path::string`, but there appears to be no mention of this in [diff]. Additionally:
```tex
\indexlibrarymember{system_encoded_string}{path}%
\indexlibrarymember{wstring}{path}%
\indexlibrarymember{u8string}{path}%
\indexlibrarymember{u16string}{path}%
\indexlibrarymember{u32string}{path}%
\begin{itemdecl}
std::string native_encoded_string() const;
std::wstring wstring() const;
std::u8string u8string() const;
std::u16string u16string() const;
std::u32string u32string() const;
\end{itemdecl}
```
It appears the library index member was not adjusted to follow LWG4512.

frederick-vs-ja:
LWG4512 renamed two member functions of `path`. However, the commit applying LWG4512 did not update the indexing correspondingly.

Closes #9226.

### DIFF
--- a/source/iostreams.tex
+++ b/source/iostreams.tex
@@ -14660,7 +14660,7 @@ be performed by \tcode{a}. Conversion, if any, is specified by
 \ref{fs.path.cvt}.
 \end{itemdescr}
 
-\indexlibrarymember{system_encoded_string}{path}%
+\indexlibrarymember{native_encoded_string}{path}%
 \indexlibrarymember{wstring}{path}%
 \indexlibrarymember{u8string}{path}%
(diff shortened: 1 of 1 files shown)


tkoeppe:
Thanks a lot!

PR #9230, opened 2026-07-25
[expr.const.const] Adjust \defnadj definition for better readability

Andreas-Krug:


eisenwave:
I don't think it's just a matter of readability; three arguments to `\defnadj` aren't valid, so the `{expression}` is plain text:
grafik

jensmaurer (approved):

PR #9237, opened 2026-07-28
[simd.creation] Add missing escaping to braces in `\tcode`

### LINKED ISSUE #9236: [simd.creation] Lambda display issues
Image

The lambda expression here is missing curly braces `{}`, but LaTeX code seems to be correct. I don't know what happened.

frederick-vs-ja:
The braces would be lost in the generated PDF without escaping.

Fixes #9236.


eisenwave (approved):

tkoeppe:
Thanks!

PR #9241, opened 2026-08-02
[basic.scope.scope] Remove incorrect namespace redeclaration example CWG3214

yehudamber:
Since the adoption of P2996 "Reflection for C++26", and the changes it made to [namespace.alias], namespace aliases are no longer alternative names for namespaces but entities of their own kind. Therefore, redeclaring a namespace as a namespace alias, even denoting the original namespace, violates [basic.link]/11.3, since one declaration declares the entity as a namespace and the other does not.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -1211,7 +1211,6 @@ namespace A {}
 namespace B = A;
 namespace B = A;        // OK, no effect
 namespace B = B;        // OK, no effect
-namespace A = B;        // OK, no effect
 namespace B {}          // error: different entity for \tcode{B}
 
 void g() {


timsong-cpp:
~If they are different entities, then [basic.link]/11.3 does not apply at all. That paragraph only applies to two declarations of the same entity.~

Ah, [basic.link]/8 says that two declarations declare the same entity if they correspond (and various other things not relevant here), which they do since they introduce the same name. But that reasoning would seem to apply equally to `typedef struct T T;` and if so, that's definitely a defect.

yehudamber:
> But that reasoning would seem to apply equally to `typedef struct T T;` and if so, that's definitely a defect.

Good point. I thought [basic.scope.scope]/4.2 covers such cases, but I see now it does not apply when the other declaration declares a type alias.
So maybe the rule at [basic.link]/8 is what needs to be fixed?

katzdm:
I agree [basic.link]/8 is problematic. Maybe something like this?

> Two declarations _D1_ and _D2_ of entities declare the same entity if, considering declarations of unnamed types to introduce their names for linkage purposes, if any ([dcl.typedef], [dcl.enum]),
> - _D1_ and _D2_ correspond ([basic.scope.scope]),
> - _D1_ and _D2_ have the same target scope that is not a function or template parameter scope,
> - neither _D1_ nor _D2_ is a name-independent declaration,
> - __if either _D1_ or _D2_ declares a type alias or a namespace alias, the underlying entities declared by _D1_ and _D2_ are either both types or both namespaces__, and
> - one of the following holds:
>   - _D1_ and _D2_ appear in the same translation unit,
>   - _D1_ and _D2_ both declare type aliases or namespace aliases that have the same underlying entity, or
>   - _D1_ and _D2_ both declare names with module or external linkage and are attached to the same module.

?

jensmaurer:
@katzdm We already have "D1 and D2 both declare type aliases or namespace aliases that have the same underlying entity, or" in the next bullet, so something feels redundant here.

Also, I get confused whether type aliases are meant to have namespaces as underlying entities with this phrasing.  Maybe make that two bullets?

Also, I'm not seeing how this fixes the problem that
```
struct T;
typedef struct T T;
```
should NOT declare the same entity twice.

timsong-cpp:
"either D1 and D2 both declare type aliases or neither declares a type alias" and the same for namespace aliases?

katzdm:
@jensmaurer It's two different cases - I'm fine with two different bullets, but we should keep the existing bullet separate. So, this for instance would be fine (and equivalent) to me:

> Two declarations _D1_ and _D2_ of entities declare the same entity if, considering declarations of unnamed types to introduce their names for linkage purposes, if any ([dcl.typedef], [dcl.enum]),
> - _D1_ and _D2_ correspond ([basic.scope.scope]),
> - _D1_ and _D2_ have the same target scope that is not a function or template parameter scope,
> - neither _D1_ nor _D2_ is a name-independent declaration,
> - __if either _D1_ or _D2_ declares a type alias, both _D1_ and _D2_ declare type aliases__,
> - __if either _D1_ or _D2_ declares a namespace alias, both _D1_ and _D2_ declare namespace aliases, and__
> - one of the following holds:
>   - _D1_ and _D2_ appear in the same translation unit,
>   - _D1_ and _D2_ both declare type aliases or namespace aliases that have the same underlying entity, or
>   - _D1_ and _D2_ both declare names with module or external linkage and are attached to the same module.

Something along the lines of what @timsong-cpp said would be fine as well. No strong preference on the phrasing.

jensmaurer:
This is not an editorial issue, and removing the example does not fix anything. CWG3214

yehudamber:
@jensmaurer Seems good, thank you!
The proposed wording does affect more cases (e.g. corresponding type alias declaration and namespace alias declaration) that currently declare the same entity, but since aliases have no linkage, and declarations that both declare a type alias or a namespace alias are not affected, those case are only when both declarations appear in the same TU, and since alias declarations always bind their name in their target scope (I believe), those cases will just become name collisions between entities instead of conflicting declarations of the same entity.

PR #9249, opened 2026-08-09
[diff.expr] Strike line in example that is UB in C23

### LINKED ISSUE #9248: [diff.expr] Paragraph added by CWG2875 is not correct for C23
[[diff.expr]/5](https://eel.is/c++draft/diff.expr#5) contains the following example:
```c
void f() {
  char *p = nullptr;
  char *p2 = &*p;       // well-defined in C, undefined behavior in C++
  char *p3 = &p[0];     // well-defined in C, undefined behavior in C++
  int a[5];
  int *q = &a[5];       // well-defined in C, undefined behavior in C++
}
```
In C `&p[0]` gets translated to `p+0` which is undefined in C23, though it is valid in C2Y because of [N3322](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3322.pdf). Should the second comment be updated since C++ currently references C23?

jensmaurer:
In C23, "p + 0" is undefined behavior if p is a nullptr, regardless of whether the result is dereferenced.

Fixes #9248

### DIFF
--- a/source/compatibility.tex
+++ b/source/compatibility.tex
@@ -3161,7 +3161,6 @@ but results in undefined behavior in \Cpp{}.
 void f() {
   char *p = nullptr;
   char *p2 = &*p;       // well-defined in C, undefined behavior in \Cpp{}
-  char *p3 = &p[0];     // well-defined in C, undefined behavior in \Cpp{}
   int a[5];
   int *q = &a[5];       // well-defined in C, undefined behavior in \Cpp{}
 }


jensmaurer:
The two other lines in the example are sufficient exposition for the C incompatibility we wish to highlight.

eisenwave (approved):
Yeah, that makes sense. Keeping the deleted line would be somewhat misleading because the UB happens earlier, so that line isn't really demonstrating what the example is meant to be about.

One might ponder whether it makes sense to keep the line for C2y (since adding `0` to null would be fine), but I don't think this line is crucial for that purpose either.

jensmaurer:
@tkoeppe , you're next.

PR #9254, opened 2026-08-13
[cfloat.syn], [cmath.syn] Add missing "// optional" comments for some macros

eisenwave:
All of these macros are optionally defined, if and only if the constant is actually representable in the type.

See https://cstd.eisie.net/c23#5.2.5.3.3p28

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -1830,11 +1830,11 @@ type of \tcode{T}\iref{conv.prom}.
 #define @\libmacro{FLT_ROUNDS}@ @\seebelow@
 #define @\libmacro{FLT_EVAL_METHOD}@ @\seebelow@
 #define @\libmacro{FLT_RADIX}@ @\seebelow@
-#define @\libmacro{INFINITY}@ @\seebelow@
-#define @\libmacro{NAN}@ @\seebelow@
-#define @\libmacro{FLT_SNAN}@ @\seebelow@
-#define @\libmacro{DBL_SNAN}@ @\seebelow@
-#define @\libmacro{LDBL_SNAN}@ @\seebelow@
+#define @\libmacro{INFINITY}@ @\seebelow@          // optional
+#define @\libmacro{NAN}@ @\seebelow@               // optional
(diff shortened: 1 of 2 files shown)


source/support.tex
```diff
@@ -1830,11 +1830,11 @@
 #define @\libmacro{FLT_ROUNDS}@ @\seebelow@
 #define @\libmacro{FLT_EVAL_METHOD}@ @\seebelow@
 #define @\libmacro{FLT_RADIX}@ @\seebelow@
-#define @\libmacro{INFINITY}@ @\seebelow@
-#define @\libmacro{NAN}@ @\seebelow@
-#define @\libmacro{FLT_SNAN}@ @\seebelow@
-#define @\libmacro{DBL_SNAN}@ @\seebelow@
-#define @\libmacro{LDBL_SNAN}@ @\seebelow@
+#define @\libmacro{INFINITY}@ @\seebelow@          // optional
```
frederick-vs-ja, line 1833: We should also change [cmath.syn] for `INFINITY` and `NAN`. Also, we should indent these comments.

`INFINITY` was mandatory in the published C99/C11/C17, and then [N2848](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2848.pdf) changed it to conditionally-supported. It seems that before N2848 `INFINITY` was required to be defined as something nonsense if the implementation doesn't support infinity.

I think WG14 N2848 should be treated as DR, but I'm nor sure whether we should add some entry to Annex C.

jensmaurer, line 1833: Please fix the indentation.

jensmaurer (changes requested):

source/support.tex
eisenwave: Done.

eisenwave: I've added the changes to [cmath.syn]. If this is basically a DR, then I don't see a need for an Annex C entry. We have always had the same contents as the C header, and the C header just got some bug-fixes, so it's not really like something changed in C++.

jensmaurer (approved):
@jwakely , this seems correct. Please have a look.

PR #9256, opened 2026-08-14
[implimits] Add \grammarterm for lambda-capture

Andreas-Krug:


source/limits.tex
```diff
@@ -68,7 +68,7 @@
 \item%
 Non-static data members (including inherited ones) in a single class\iref{class.mem} [16\,384].
 \item%
-Lambda-captures in one \grammarterm{lambda-expression}\iref{expr.prim.lambda.capture} [256].
+\grammarterm{lambda-capture}s in one \grammarterm{lambda-expression}\iref{expr.prim.lambda.capture} [256].
```
jensmaurer, line 71: Is this correct? Did we maybe mean "lambda captures", including implicit ones, instead of appearances of the syntactic construct?

PR #9259, opened 2026-08-16
[expr.const.core] Fix typo

jensmaurer:
This fixes a misapplication of P4101R1, introduced by commit 154eccbe38722c6ecc67da9ce881f03188e4d2cf.


jensmaurer:
@tkoeppe , this is fairly high-priority, because there is ongoing CWG discussion in this area.

PR #9264, opened 2026-08-20
[intro.compliance.general] Fix cross-reference in footnote

jensmaurer:


tkoeppe:
(Is this also for C++26?)

jensmaurer:
Yes, that's also for C++26.

PR #9277, opened 2026-08-28
[flat.multimap] Apply "constexpr" to flat_multimap::operator=(initializer_list)

Quuxplusone:
"constexpr" was applied to flat_map::operator=(initializer_list) in commit e1502d2 "P3372R3 constexpr containers and adaptors," but it was missed here. I don't believe this was intentional.

Apologies if there's already an LWG issue for this.

### DIFF
--- a/source/containers.tex
+++ b/source/containers.tex
@@ -18635,7 +18635,7 @@ namespace std {
       constexpr flat_multimap(sorted_equivalent_t, initializer_list<value_type> il,
                               const key_compare& comp, const Alloc& a);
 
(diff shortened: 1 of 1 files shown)


frederick-vs-ja:
> Apologies if there's already an LWG issue for this.

Luckily there's no yet LWG issue for this. I just verified that the `constexpr` for `flat_multimap& operator=(initializer_list<value_type>)` is already in https://wg21.link/p3372r3#flat.multimap.defn, so missing `constexpr` is merely due to misapplication of the paper.

PR #9281, opened 2026-08-29
[class.union.anon] Correct definition of union-like class

TymianekPL:
There is a redundant *union or*, as all unions are classed.
```diff
-  union-like class is a union or a class [...]
+  union-like class is a class [...]
```

### DIFF
--- a/source/classes.tex
+++ b/source/classes.tex
@@ -3345,7 +3345,7 @@ in~\ref{dcl.init.aggr}.
 
 \pnum
 \indextext{class!variant member of}%
-A \defnadj{union-like}{class} is a union or a class that has an anonymous union as a direct
+A \defnadj{union-like}{class} is a class that has an anonymous union as a direct
 member. A union-like class \tcode{X} has a set of \defnx{variant members}{variant member}.
 If \tcode{X} is a union, a non-static data member of \tcode{X} that is not an anonymous
 union is a variant member of \tcode{X}. In addition, a non-static data member of an


eisenwave:
I don't think there's something objectively wrong here. "union" is a defined term in https://eel.is/c++draft/class.union.general#def:union

Saying "union or class" is clunky though. My preference is "class type", if we want a change, but just "class" is also fine by me.

TymianekPL:
I suppose a synonym term for "Correct" could be used? But I'm not sure which one reflects the intent better...

Halalaluyafail3:
I thought it was meant to be read like "A union-like class is a union or (a class that has an anonymous union as a direct member)." rather than "A union-like class is (a union or a class) that has an anonymous union as a direct member.".

TymianekPL:
Okay turns out that _union-like class_ is just a leftover from old wording that doesn't exist anymore

Should we remove this definition as it's unused?

eisenwave:
> I thought it was meant to be read like "A union-like class is a union or (a class that has an anonymous union as a direct member)." rather than "A union-like class is (a union or a class) that has an anonymous union as a direct member.".

Oh yeah, that reading makes much more sense.

Perhaps it could be clarified by "is a union or is a class"

eisenwave:
> Okay turns out that _union-like class_ is just a leftover from old wording that doesn't exist anymore
> 
> Should we remove this definition as it's unused?

At the end of the paragraph, it says

> At most one variant member of a union may have a default
member initializer.

So the "union-like class" term may not be used elsewhere, but it's used internally.

TymianekPL:
Oh right true, interesting...
I'm still unsure about how one should parse this sentence still. Maybe we can come up with better wording to say this intent? The current one is a bit ambiguous 😅

PR #9283, opened 2026-08-30
[basic.start.term] Describe sequencing of function calls, not of functions

### LINKED ISSUE #9280: [basic.start.term]/1 "sequenced before" followed by "function"
[basic.start.term]/1 says

> Constructed complete objects ([[dcl.init]](https://eel.is/c++draft/dcl.init)) with static storage duration are destroyed and functions registered with std​::​atexit are called as part of a call to std​::​exit ([[support.start.term]](https://eel.is/c++draft/support.start.term))[.](https://eel.is/c++draft/basic.start.term#1.sentence-1) The call to std​::​exit is **sequenced before** the destructions and the **registered functions**[.](https://eel.is/c++draft/basic.start.term#1.sentence-2)

Clearly "sequenced before registered functions" makes no sense. It shall be "sequenced before calls".

Also be careful to say calls caused by `std::exit` rather than arbitrary calls.

eisenwave:
Closes #9280

https://github.com/cplusplus/draft/issues/9280#issuecomment-5467358932 mentions that "call" should be replaced with "invoke", but I don't see how to implement that nicely and consistently.

We use "call" 20 times in [[basic.start.term]](https://eel.is/c++draft/basic.start.term), so I don't see a need to get hung up on it in only the first paragraph, especially if it decreases local consistency with the rest of the subclause.

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -7661,7 +7661,7 @@ are called as part of a call to
 \indexlibraryglobal{exit}%
 \tcode{std::exit}\iref{support.start.term}.
 The call to \tcode{std::exit} is sequenced before
-the destructions and the registered functions.
+the destructions and calls to registered functions.
 \begin{note}
 Returning from \tcode{main} invokes \tcode{std::exit}\iref{basic.start.main}.
 \end{note}


jensmaurer (approved):

eisenwave:
- https://github.com/cplusplus/CWG/issues/997

BlowingWind314:
Maybe it's pedantic but a "call (noun)" is an expression ([[expr.call]/1](https://eel.is/c++draft/expr.call#1)), which cannot be used with "sequenced before" or "(strongly) hanppens before". I guess that's why @jensmaurer talked about "invocation". I will open an issue discussing this.

PR #9296, opened 2026-09-01
[special] Index "potentially constructed subobject"

Quuxplusone:


source/classes.tex
```diff
@@ -1152,7 +1152,8 @@
 \pnum
 For a class, its direct non-static data members, its non-virtual direct base classes,
 and, if the class is not abstract\iref{class.abstract}, its virtual base
-classes are called its \term{potentially constructed subobjects}.
+classes are called its \defnadj{potentially constructed}{subobjects}{subobject}.
+\indextext{potentially constructed subobject|see{subobject, potentially constructed}}
```
frederick-vs-ja: Do we need `\indextext`?

eisenwave (changes requested):
The `\indextext` is unnecessary, and this should be `\defnadjx`.

Quuxplusone:
Updated!

jensmaurer (approved):

PR #9309, opened 2026-09-04
[cstdlib.syn] Add missing index entries and remove stray entries

eisenwave:
The indexing in [cstdlib.syn] is really broken in a number of ways:
- Several symbols are not indexed anywhere, especially the more recent C23 stuff. For example, there is no index entry for `memalignment`, `free_sized`, `strfromf`, etc. I haven't counted, but there should be at least a dozen missing entries here.
- Several index entries point to declarations, not to definitions, by virtue of having dedicated subclauses with definitions. For example, there are index entries for `rand` and `srand`, even though those are defined in [c.math.rand]. Traditionally, we do not index declarations.
- The synopsis spans at least two pages, so some of the index entries point to the wrong page.
- The big pile of `\indexlibrayglobal` makes it difficult to maintain the indexing in this area. Inline `\libglobal` is better.

### DIFF
--- a/source/support.tex
+++ b/source/support.tex
@@ -102,67 +102,19 @@ and as noted in
 \rSec2[cstdlib.syn]{Header \tcode{<cstdlib>} synopsis}
 
 \indexheader{cstdlib}%
-\indexlibraryglobal{EXIT_FAILURE}%
-\indexlibraryglobal{EXIT_SUCCESS}%
-\indexlibraryglobal{MB_CUR_MAX}%
-\indexlibraryglobal{NULL}%
-\indexlibraryglobal{RAND_MAX}%
-\indexlibraryglobal{_Exit}%
-\indexlibraryglobal{abort}%
-\indexlibraryglobal{abs}%
-\indexlibraryglobal{aligned_alloc}%
-\indexlibraryglobal{at_quick_exit}%
-\indexlibraryglobal{atexit}%
-\indexlibraryglobal{atof}%
-\indexlibraryglobal{atoi}%
-\indexlibraryglobal{atol}%
-\indexlibraryglobal{atoll}%
-\indexlibraryglobal{bsearch}%
-\indexlibraryglobal{calloc}%
-\indexlibraryglobal{div}%
-\indexlibraryglobal{div_t}%
-\indexlibraryglobal{exit}%
-\indexlibraryglobal{free}%
-\indexlibraryglobal{getenv}%
-\indexlibraryglobal{labs}%
-\indexlibraryglobal{ldiv}%
-\indexlibraryglobal{ldiv_t}%
-\indexlibraryglobal{llabs}%
-\indexlibraryglobal{lldiv}%
-\indexlibraryglobal{lldiv_t}%
-\indexlibraryglobal{malloc}%
-\indexlibraryglobal{mblen}%
-\indexlibraryglobal{mbstowcs}%
-\indexlibraryglobal{mbtowc}%
-\indexlibraryglobal{qsort}%
-\indexlibraryglobal{quick_exit}%
-\indexlibraryglobal{rand}%
-\indexlibraryglobal{realloc}%
-\indexlibraryglobal{size_t}%
-\indexlibraryglobal{srand}%
-\indexlibraryglobal{strtod}%
-\indexlibraryglobal{strtof}%
-\indexlibraryglobal{strtol}%
-\indexlibraryglobal{strtold}%
-\indexlibraryglobal{strtoll}%
-\indexlibraryglobal{strtoul}%
-\indexlibraryglobal{strtoull}%
-\indexlibraryglobal{system}%
-\indexlibraryglobal{wcstombs}%
-\indexlibraryglobal{wctomb}%
 \begin{codeblock}
 namespace std {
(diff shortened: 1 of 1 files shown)


source/support.tex
```diff
@@ -102,67 +102,19 @@
...
-\indexlibraryglobal{ldiv_t}%
-\indexlibraryglobal{llabs}%
-\indexlibraryglobal{lldiv}%
-\indexlibraryglobal{lldiv_t}%
-\indexlibraryglobal{malloc}%
-\indexlibraryglobal{mblen}%
-\indexlibraryglobal{mbstowcs}%
-\indexlibraryglobal{mbtowc}%
-\indexlibraryglobal{qsort}%
-\indexlibraryglobal{quick_exit}%
```
jensmaurer, line 138: We removed this here, but didn't add \libglobal below, it seems.

source/support.tex
```diff
@@ -181,33 +133,33 @@
...
   [[noreturn]] void quick_exit(int status) noexcept;                    // freestanding
 
-  char* getenv(const char* name);
-  int system(const char* string);
+  char* @\libglobal{getenv}@(const char* name);
+  int @\libglobal{system}@(const char* string);
 
   // \ref{c.malloc}, C library memory allocation
   void* aligned_alloc(size_t alignment, size_t size);
   void* calloc(size_t nmemb, size_t size);
```
jensmaurer, line 141: What about these?  No index entries?

source/support.tex
```diff
@@ -181,33 +133,33 @@
...
   void free(void* ptr);
-  void free_sized(void* ptr, size_t size);
-  void free_aligned_sized(void* ptr, size_t alignment, size_t size);
+  void @\libglobal{free_sized}@(void* ptr, size_t size);
+  void @\libglobal{free_aligned_sized}@(void* ptr, size_t alignment, size_t size);
   void* malloc(size_t size);
   void* realloc(void* ptr, size_t size);
```
jensmaurer, line 146: And here?  No index entries?

source/support.tex
```diff
@@ -181,33 +133,33 @@
...
+  int @\libglobal{strfroml}@(char* s, size_t n, const char* format, long double fp);
+  float @\libglobal{strtof}@(const char* nptr, char** endptr);
+  long double @\libglobal{strtold}@(const char* nptr, char** endptr);
+  long int @\libglobal{strtol}@(const char* nptr, char** endptr, int base);
+  long long int @\libglobal{strtoll}@(const char* nptr, char** endptr, int base);
+  unsigned long int @\libglobal{strtoul}@(const char* nptr, char** endptr, int base);
+  unsigned long long int @\libglobal{strtoull}@(const char* nptr, char** endptr, int base);
 
   // \ref{c.mb.wcs}, multibyte / wide string and character conversion functions
   int mblen(const char* s, size_t n);
```
jensmaurer, line 165: all of these: no index entries?

source/support.tex
```diff
@@ -238,14 +190,14 @@
...
 
-  constexpr long int labs(long int j);                                  // freestanding
-  constexpr long long int llabs(long long int j);                       // freestanding
+  constexpr long int @\libglobal{labs}@(long int j);                                  // freestanding
+  constexpr long long int @\libglobal{llabs}@(long long int j);                       // freestanding
 
-  constexpr div_t div(int numer, int denom);                            // freestanding
+  constexpr div_t @\libglobal{div}@(int numer, int denom);                            // freestanding
   constexpr ldiv_t div(long int numer, long int denom);                 // freestanding; see \ref{library.c}
   constexpr lldiv_t div(long long int numer, long long int denom);      // freestanding; see \ref{library.c}
```
jensmaurer, line 198: and here

jensmaurer (changes requested):

source/support.tex
eisenwave: This is intentional for all of the comments that you've left. All of the functions that have not received `\libglobal` formatting are defined in some other place, such as https://eel.is/c++draft/c.mb.wcs#lib:mbstowcs

We don't usually add an index entry for forward declarations, and https://eel.is/c++draft/cstdlib.syn#1 explicitly exempts the functions defined in those other subclauses from receiving the "this does the same thing as in C" treatment.

I think indexing `mbstowcs` in this place would be outright misleading because it would incorrectly imply that the function is *defined* in [cstdlib.syn] with "this does the C thing" treatment.

source/support.tex
eisenwave: https://eel.is/c++draft/c.malloc

source/support.tex
eisenwave: https://eel.is/c++draft/c.malloc

source/support.tex
eisenwave: https://eel.is/c++draft/c.mb.wcs

source/support.tex
eisenwave: I don't think we normally put `libglobal` on every member of an overload set. I think that would generate multiple quasi-identical index entries.

jensmaurer: Index entries pointing to the same page are automatically merged by the indexing machinery.

jensmaurer (approved):

PR #9312, opened 2026-09-04
[expr.prim.pack.index] Replace "in" with "of" in paragraph 1

### LINKED ISSUE #9288: [expr.prim.pack.index]/1 A nested id-expression in the pack-index-expression need not denote a pack
[expr.prim.pack.index]/1 says

> [pack-index-expression:](https://eel.is/c++draft/expr.prim.id#nt:pack-index-expression)
> [id-expression](https://eel.is/c++draft/expr.prim.id#nt:id-expression) ... [ [constant-expression](https://eel.is/c++draft/expr.const.general#nt:constant-expression) ]
>
> The [id-expression](https://eel.is/c++draft/expr.prim.id#nt:id-expression) P **in** a [pack-index-expression](https://eel.is/c++draft/expr.prim.id#nt:pack-index-expression) shall be an [identifier](https://eel.is/c++draft/lex.name#nt:identifier) that denotes a pack[.](https://eel.is/c++draft/expr.prim.id#expr.prim.pack.index-1.sentence-1)

*id-expression*s within the *constant-expression* need not denote some packs.

> The *id-expression* P <del>in</del> <ins>of</ins> a *pack-index-expression* shall be an *identifier* that denotes a pack.

eisenwave:
Closes #9288

There is an argument to be made that this is not editorial, but I personally think this is basically a typo. The combination "The" and "in" doesn't make any sense anyway here. It would need to be

> An *id-expression* $P$ in a *pack-index-expression*

to make any sense. The intent is pretty clear, especially since this paragraph refers to the grammar snippet immediately above, so it obviously refers to the *id-expression* of the *pack-index-expression* as shown above, rather than making some broad and recursive statement.

Implementers are obviously going to be doing the right thing anyway. It would be evil and unhinged to assume that every *id-expression* recursively needs to denote a pack just because it's nested somewhere in a *pack-index-expression*.

### DIFF
--- a/source/expressions.tex
+++ b/source/expressions.tex
@@ -2079,7 +2079,7 @@ and a prvalue otherwise.
 \end{bnf}
 
 \pnum
-The \grammarterm{id-expression} $P$ in a \grammarterm{pack-index-expression}
+The \grammarterm{id-expression} $P$ of a \grammarterm{pack-index-expression}
 shall be an \grammarterm{identifier} that denotes a pack.
 
 \pnum


cor3ntin:
I think we are splitting hairs here.

However, if we wanted to do that change we should do it consistently
Here https://eel.is/c++draft/dcl.type.pack.index#1
and here https://eel.is/c++draft/temp.names#3

jensmaurer (approved):

cor3ntin:
@jensmaurer you don't agree we should be consistent?

jensmaurer:
I agree we should use "of" instead of "in" when we dissect the first level of grammar.  This fixes one such (obvious) occurrence.  Patches welcome for the other two occurrences you identified.

PR #9318, opened 2026-09-06
[stdbit.h.syn], [numeric.c.ckdint] Replace misuses of `\xrefc` with `\IsoC`

eisenwave:
grafik


eisenwave:
It might be worth turning this into an automatic check at some point.

jensmaurer (approved):

PR #9324, opened 2026-09-09
[basic.fundamental] Remove bullet for reflections being function parameters

eisenwave:
Function parameters are already covered by the preceding bullet stating that a reflection may be a representation of a variable. Every function parameter is a variable, and having a distinct bullet misleads the reader into believing it is not.

This misunderstanding that function parameters are not variables led to CWG3005.

See prior discussion at:
- https://github.com/cplusplus/CWG/issues/1004

### DIFF
--- a/source/basic.tex
+++ b/source/basic.tex
@@ -5749,7 +5749,6 @@ every other reflection is a representation of
 \item a variable\iref{basic.pre},
 \item a structured binding\iref{dcl.struct.bind},
 \item a function\iref{dcl.fct},
-\item a function parameter,
 \item an enumerator\iref{dcl.enum},
 \item an annotation\iref{dcl.attr.grammar},
 \item a type alias\iref{dcl.typedef},


eisenwave:
I think you could do more than just strike the bullet; you could change the other bullet to "a variable (including function parameters)" or so, but this would de-facto admit that our wording is so confusing that every time we say "variable", we need to remind the reader of what we actually mean and that we didn't just forget about function parameters.

eisenwave:
Hmm, so it appears that [`variable_of`](https://eel.is/c++draft/meta.reflection.queries#lib:variable_of) doesn't make much sense if there are no distinct reflections of function parameters. I'm mildly confused what `variable_of` is meant to do in the first place, considering that a function parameter already is a variable, so getting the variable of a function parameter is presumably the identity function.

eisenwave:
After consulting with others, I guess we do intend for "function parameter" and "variable" to be distinct, even though the wording doesn't make that work very well. Deleting the bullet goes against that intent.

PR #9335, opened 2026-09-16
[cstdint.syn] State that INTPTR_WIDTH and UINTPTR_WIDTH are mandatory

eisenwave:
[P3248R5](https://wg21.link/p3248) removed the `// optional` comments on `INTPTR_WIDTH` and `UINTPTR_WIDTH` to indicate that these are now mandatory macros, but neglected to state in the paragraph below that this is a difference compared to the `<stdint.h>` header.


jensmaurer (approved):
Sounds good.  @jwakely , can we apply this without an LWG issue?

